PR #214, opened 2016-11-15
Support decimal -> DecimalGraphType

dlukez:
Trying to register a decimal field using a property expression currently results in an ArgumentOutOfRange exception.


joemcbride:
👍 Looks like CI is failing, possibly related to a newer version of yarn.  Will take a look at that.

joemcbride:
Not sure what is up with CI but this looks good.

joemcbride:
Published 0.13.1 with that feature.  https://www.nuget.org/packages/GraphQL/0.13.1.601

PR #216, opened 2016-11-16
Fix Interface(Type) for generic ObjectGraphType<>

dlukez:
#212 fixed the non-generic version of `ObjectGraphType` but unfortunately there's some duplicated code.... Could the non-generic `ObjectGraphType` just derive from `ObjectGraphType<object>`?

Also fixed up the test, since the bug only gets hit when calling the `Interface(Type)` overload.


tlil87:
Yep, I have this fixed locally here as well; just haven't gotten around to pushing it yet. Thanks for the reminder :-)

tlil87:
LGTM 👍

joemcbride:
I vaguely recall there being an issue with inheriting from `ObjectGraphType<T>` and that is why there is duplication.  I could be wrong.

PR #218, opened 2016-11-17
Updates for interfacing with Conventions

tlil87:
* **Add**: Added setters for `IGraphType` (needed by the Conventions library)
 * **Add**: Non-generic method for adding type to `UnionGraphType`
 * **Fix**: Validation rule threw on undefined argument type name


tlil87:
@joemcbride Dunno if you want to keep those `Console.WriteLine()` calls for `DEBUG` targets. I've commented them out for now.

joemcbride:
@tlil87 I kind of do want those debug statements, though I can add a different flag for that if/when I need them again.  They helped immensely with the Validation rules.

tlil87:
@joemcbride :+1: cool, thanks!

PR #220, opened 2016-11-21
Catch faulty tasks and rethrow exception on field resolution

tlil87:
When the middleware resolver sees a faulted task, awaiting it would enter a loop and not report the underlying exception. This branch gets around that issue by re-throwing the reported task exception.


joemcbride:
Nice find!  I'm guessing we have that same problem in the DocumentExecuter.

tlil87:
Ah, yeah, that might well be! I'll create a PR for that as well

PR #221, opened 2016-11-21
Catch faulty tasks and rethrow exception on field resolution (in DocumentExecutor)

tlil87:
#220 (part 2)

### DIFF
--- a/src/GraphQL.Tests/Execution/MutationTests.cs
+++ b/src/GraphQL.Tests/Execution/MutationTests.cs
@@ -240,7 +240,7 @@ namespace GraphQL.Tests.Execution
                 }";
 
             var result = AssertQueryWithErrors(query, expected, root: new Root(6), expectedErrorCount: 2);
-            result.Errors.First().InnerException.Message.ShouldBe("Cannot change the number");
+            result.Errors.First().InnerException.InnerException.Message.ShouldBe("Cannot change the number");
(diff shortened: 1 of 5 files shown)


src/GraphQL.Tests/Execution/MutationTests.cs
```diff
@@ -240,7 +240,7 @@ public void evaluates_mutations_correctly_in_the_presense_of_a_failed_mutation()
                 }";
 
             var result = AssertQueryWithErrors(query, expected, root: new Root(6), expectedErrorCount: 2);
-            result.Errors.First().InnerException.Message.ShouldBe("Cannot change the number");
+            result.Errors.First().InnerException.InnerException.Message.ShouldBe("Cannot change the number");
```
joemcbride: Looking at how this changes the exception, perhaps we should instead just not try to await the task if it is already faulted instead of re-throw?

```
if (!task.IsFaulted)
{
   await task.ConfigureAwait(false);
   result = task.GetProperyValue("Result");
}
```

tlil87: @joemcbride Yes, can do – reuse the logic [here](https://github.com/graphql-dotnet/graphql-dotnet/pull/221/files#diff-9ee075dfc3429298a3c5de0c1b53bef2L325) to report the `task.Exception` then you mean?

PR #222, opened 2016-11-21
better errors

jquense:
Adds a few more descriptive errors


jquense:
o this also adds support for nullable type inference...I meant to include that in a different PR oops :P

PR #243, opened 2016-12-20
Fix array inference

jquense:


joemcbride:
Published a version with this fix

https://www.nuget.org/packages/GraphQL/0.14.1.638

PR #245, opened 2016-12-23
Using Autofac instead of the GraphQL provided SimpleContainer

ghstahl:
Autofac is preferable over a custom implementation.


joemcbride:
Thanks for this, though Autofac is not preferable to me as the maintainer.  I would welcome a PR to add an example of using Autofac to the docs though.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/docs/learn.md#dependency-injection

PR #246, opened 2016-12-23
Added Instructions to use Autofac as our DI

ghstahl:


SimonCropp:
doesnt this only make sense if we also document all the other containers to the same extent?

sungam3r:
Do we still need a docs (not docs2) directory?

SimonCropp:
@sungam3r i dont think so. here is a PR https://github.com/graphql-dotnet/graphql-dotnet/pull/1417  @joemcbride thoughts?

sungam3r:
@ghstahl Could you please move these changes to https://github.com/graphql-dotnet/graphql-dotnet/blob/master/docs2/site/docs/getting-started/dependency-injection.md#autofac ?

sungam3r:
@ghstahl  Thanks for the PR. I read your changes and see that they are specific to ASP.NET MVC5 project. As @joemcbride said we have a new DI docs section with a minimal description of different containers settings. I would like to keep the simplicity of this documentation page without introducing the specifics of individual project types. If you want, you can resolve conflicts and update PR accordingly. In the meantime, I will close it.

Issue #251, opened 2017-01-06
DecimalGraphType cannot be used on inputs


LasseNisted:
An InputObjectGraphType with a DecimalGraphType field fails in the
GraphQLExtensions.AstFromValue method which does not have a case for decimal values.

joemcbride:
The fix for this looks like is more complicated than I thought.  Not sure how to "properly" fix it yet.  Thanks for the report!

ManuelDeLeon:
We solved this problem by adding a `DecimalValue` to IntValue.cs and a decimal check in GraphQLExtensions.cs

IntValue.cs:
```csharp
    public class DecimalValue : AbstractNode, IValue 
    { 
        public DecimalValue(decimal value) 
        { 
            Value = value; 
        } 
 
        public decimal Value { get; } 
 
        public override string ToString() 
        { 
            return "DecimalValue{{value={0}}}".ToFormat(Value); 
        } 
 
        protected bool Equals(DecimalValue other) 
        { 
            return Value == other.Value; 
        } 
 
        public override bool IsEqualTo(INode obj) 
        { 
            if (ReferenceEquals(null, obj)) return false; 
            if (ReferenceEquals(this, obj)) return true; 
            if (obj.GetType() != this.GetType()) return false; 
            return Equals((DecimalValue)obj); 
        } 
    } 
```

GraphQLExtensions.cs (in `AstFromValue`)
```csharp
            if (serialized is decimal) 
            { 
                return new DecimalValue((decimal)serialized); 
            } 
```

Is there a problem with this solution?

joemcbride:
> Is there a problem with this solution?

Only in that there is no `DecimalValue` in the spec only `FloatValue`.  So that starts to muddy the waters, which we've kind of already done with `LongValue`.  We can probably add `DecimalValue` in the short term.  Long term I would like to provide hooks so that anyone who wants to add custom scalar types could override these behaviors.

ManuelDeLeon:
> Long term I would like to provide hooks so that anyone who wants to add custom scalar types could override these behaviors.

That's an interesting idea. I dig it for custom types but I wonder about types native to .Net (like decimal). The target audience here are .Net developers so would it really make sense to declare "Types that translate to the GraphQL spec are handled out of the box. For the rest of the .Net types use this separate library (or turn on this flag)"?

> We can probably add DecimalValue in the short term. 

Want a PR?

joemcbride:
[...]

PR #267, opened 2017-01-25
Add DecimalValue to the AST

ManuelDeLeon:
DecimalGraphType break when used as inputs. This fixes the issue #251


joemcbride:
Thanks!  I believe the [ParseLiteral](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/DecimalGraphType.cs#L27) function of the `DecimalGraphType` then needs to be updated to support that value, otherwise it won't go through.

ManuelDeLeon:
You're right. I added the test for `ParseLiteral`.

joemcbride:
Closes #251

PR #270, opened 2017-01-30
Introduced IInputObjectGraphType and ObjectMutationTests

christostatitzikidis:
Hi,
We had a problem here doing Mutations straight on objects, which could be easily solved by checking  for IInputObjectGraphType (which derives from the original IInputGraphType interface) instead of InputObjectGraphType. That way mutations are possible on objects as well as scalar types and you can now do mutations like "immediatelyChangeTheObject(newObject:  { prop1: ""Mod1"", prop2: 1 })".

I also put some tests in  src/GraphQL.Tests/Execution/MutationTests.cs, of similar fashion as the original mutation tests.

Thanks,
Christos

### DIFF
--- a/src/GraphQL.Tests/Execution/MutationTests.cs
+++ b/src/GraphQL.Tests/Execution/MutationTests.cs
@@ -12,11 +12,27 @@ namespace GraphQL.Tests.Execution
         public int TheNumber { get; set; }
     }
 
+    public class MyObject
+    {
+        public string Prop1 { get; set; }
+        public int Prop2 { get; set; }
+    }
+
+    public class MyObjectHolder
+    {
+        public MyObject TheObject { get; set; }
+
+        public MyObjectHolder()
+        {
+            TheObject = new MyObject();
+        }
+    }
+
     public class Root
     {
         public Root(int number)
         {
-            NumberHolder = new NumberHolder {TheNumber = number};
+            NumberHolder = new NumberHolder { TheNumber = number };
         }
 
         public NumberHolder NumberHolder { get; private set; }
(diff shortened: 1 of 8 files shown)


jquense:
Not sure I understand what this gains you...mutations with object inputs were already possible, What does having an interface here enable that wasn't possible before?

christostatitzikidis:
With the current implementation the input object would have to inherit from InputObjectGraphType which inherits from ComplexGraphType<object>.

The output object type would be ObjectGraphType<T>.

So if one wanted to implement both Input and Output on the same object they would not be able to to use fluent syntax like
Field(_ => _.Id).Description("The id");
Field(_ => _.Name, true).Description("The name");

when defining an object both for input and output.

With what I implemented, one can create an object like:
TGraphType : ObjectGraphType<T>, IInputObjectGraphType which will handle both input and output with a fluent syntax.

All the rest of the code just works fine with this little tweak, so it could be a nice feature that saves having to write different objects for in and out or not using fluent syntax.

jquense:
@joemcbride can speak more intelligently about this, but I'm fairly certain that `InputObjectGraphType` and `ObjectGraphType` are intentionally discrete types per the GraphQL spec. It doesn't really make sense to have an `InputObjectGraphType<TSource>` since the source object is always a dictionary deserialized from a json object, as opposed to a data model object. I do see how it could be convienent to reuse the output types where they overlap but I'd also be concerned that it's conflating non-overlapping Types

joemcbride:
I'm in agreement with @jquense.  As he mentions they should be discrete types.  While I wouldn't be opposed to updating the code to use an interface, I do not want to promote this type of use.

christostatitzikidis:
I wouldn't say that I have studied the specification much, so if it this steps out of it, fair enough. 

Thanks for your time! :)

joemcbride:
Going to go ahead and close this.  Thanks!

harrisse:
@joemcbride @jquense I understand why you would want different types for input and output, but I think there's a valid use case for a strongly typed InputObjectGraphType:

`
public class UserInputType : InputObjectGraphType<UserInput> {
    public UserInputType() {
        Field(userInput => userInput.Id);
        Field(userInput => userInput.Name);
        ... etc.
    }
}

public class UserInput {
    public int Id { get; set; }
    ... etc.
}
`

This can also be nice when it comes to resolution of the User as you can pass your strongly typed UserInput object directly to your user resolution service:

`
Field<User>(
    "user", 
    arguments: new QueryArguments(new QueryArgument<UserInputType> { Name = "input" }),
    resolve: context => userService.FindUser(context.GetArgument<UserInput>("input"))
);
`

What are your thoughts on this usage, and would you accept a PR for InputObjectGraphType<T>?

joemcbride:
@harrisse What you propose can be done with the code already. (There is an example of that pattern in the [mutations](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/docs/getting-started.md#mutations) docs.)  What would your changes be?

harrisse:
Reviewing that example, what would be the base type of DinnerInputType? Based on my understanding it would be InputObjectGraphType<DinnerType>, but that type is not available in the current codebase.

jquense:
>  `Field(userInput => ((UserInput)userInput).Id) `

The object there is **not** a `UserInput` its a `Dictionary<string, object>` You can define types without the generics api. e.g.

```c#
public class UserInputType : InputObjectGraphType {
    public UserInputType() {
        Field("Id", StringGraphType);
        Field("Name", StringGraphType);
        ... etc.
    }
}

```

harrisse:
I understand you can define types without generics, but I would prefer to do it with generics where possible to support refactoring, type safety, etc (another aspect I'm interested in is being able to generate Typescript models using something like TypeLite so my clients can reuse the same input model as well). Do generic input types not work / not make sense from the graphql spec's perspective?

joemcbride:
You _could_ use a generic to define the `InputObjectGraphType`.  As you mention it would perhaps just make refactoring easier.  What we have to today and a generic `InputObjectGraphType` would be equivalent as far as the engine is concerned.

What was proposed earlier in this PR was to use the same `GraphType` class for both the input and output graph types, which is against the spec.

jquense:
doing the generic InputGraphType might cause errors in the resolvers as it'd expect the Source object is the `TSource` type. I dunno, I see how it'd be useful for type definition but it seems weird to key off an object Type that _isn't_ the actual backing type. Just my 2cents tho :)

joemcbride:
@jquense Agreed that could cause issues with how we have it setup. Will take some experimenting thinking upon to see if it would work without causing a bigger headache.

Issue #271, opened 2017-01-30
Issue with datetime input inside string field


mamasselin:
Hi,

I recently started working on a project with your library and just found out that there might be an issue when passing "datetime-like" strings inside string fields.

Here's an explanation of my problem:

I have a class with a string field inside, like this:

```c#
public class Item
{
    public string Value { get; set;}
}
```

With corresponding GraphType and InputGraphType as below:
```c#
public class ItemType : ObjectGraphType<Item>
{
    public ItemType()
    {
        Field(x => x.Value);
    }
}

public class ItemInputType : InputObjectGraphType
{
    public ItemInputType()
    {
        Field<StringGraphType>("value");
    }
}
```

If I send a mutation with an ISO formatted datetime string, the corresponding argument is set to null in my resolver context.

Example of ISO formatted datetime string:
```json
{
    "ItemInputType": { "value": "2017-01-27T15:19:53.000Z" }
}
```

When I send a mutation with any other random text it works just fine.

I forked the master branch and created a unit test that replicates this issue.

Please see pull request #272 

cc: @cstlaurent

joemcbride:
Because it is a date string, it is probably getting converted to a `DateValue`, which the `StringGraphType` does not know how to handle.  Are you able to use the `DateGraphType` instead of the `StringGraphType`?  Is there a reason why you want to just use the string?

mamasselin:
Thanks for the quick response.

I can't use `DateGraphType` because my field will contain any kind of string values, including date strings.

I haven't had the chance to look deeper into this yet, so I'm just throwing some ideas here:

Would it be possible for the `StringGraphType` to fallback to the `.ToString()` of the object if it's from a different type?

Alternatively, the json could be interpreted as a string when deserializing (I assume by removing the `DateFormatHandling.IsoDateFormat` setting from the serializer), and if it's going into a `DateGraphType` only then it would be parsed as a `DateTime`?

mamasselin:
I added the following setting to the JSON Serializer and it seems to solve my problem:
`DateParseHandling = DateParseHandling.None`

But now the following test fails because the date is converted to my local time for some reasons:
```c#
public void can_convert_utc_date_to_datetime_with_correct_kind()
{
    var json = @"{ 'h': '2016-10-21T13:32:15.753Z' }";
    var expected = DateTime.SpecifyKind(DateTime.Parse("2016-10-21T13:32:15.753"), DateTimeKind.Utc);

[...]

PR #272, opened 2017-01-30
bug with datetime inside string fields

mamasselin:
Fix for issue #271

### DIFF
--- a/src/GraphQL.Tests/project.json
+++ b/src/GraphQL.Tests/project.json
@@ -24,8 +24,7 @@
         "Microsoft.NETCore.App": {
           "type": "platform",
           "version": "1.0.0"
-        },
-        "Microsoft.CodeCoverage": "1.0.2"
+        }
       },
       "imports": [
         "dnxcore50",
--- a/src/GraphQL.Tests/project.json
+++ b/src/GraphQL.Tests/project.json
@@ -24,7 +24,8 @@
         "Microsoft.NETCore.App": {
           "type": "platform",
           "version": "1.0.0"
-        }
+        },
+        "Microsoft.CodeCoverage": "1.0.2"
       },
       "imports": [
         "dnxcore50",
--- a/src/GraphQL/ObjectExtensions.cs
+++ b/src/GraphQL/ObjectExtensions.cs
@@ -127,8 +127,6 @@ namespace GraphQL
 
             if (fieldType == typeof(DateTime))
             {
-                if (value is DateTime) return value;
-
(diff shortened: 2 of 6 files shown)


src/GraphQL.Tests/project.json
```diff
@@ -24,7 +24,8 @@
         "Microsoft.NETCore.App": {
           "type": "platform",
           "version": "1.0.0"
-        }
+        },
+        "Microsoft.CodeCoverage": "1.0.2"
```
huysentruitw: Don't add this dependency in the PR

src/GraphQL/ObjectExtensions.cs
```diff
@@ -123,7 +124,15 @@ public static Type GetInterface(this Type type, string name)
         public static object ConvertValue(object value, Type fieldType)
         {
             if (value == null) return null;
-            if (fieldType == typeof(DateTime) && value is DateTime) return value;
+
+            if (fieldType == typeof(DateTime))
```
joemcbride: Instead of making a change here, it should be done in the `DateTimeConverter`.  That is where all of the actual conversion of values are taking place.  After messing with it a bit, I think this is the change that is needed:

```csharp
// var success = DateTime.TryParseExact(trimmedString, "yyyy-MM-dd'T'HH:mm:ss.fffffff'Z'", CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal, out result);
var success = DateTime.TryParse(trimmedString, CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal, out result);
```

src/GraphQL.Tests/project.json
mamasselin: fixed

src/GraphQL/ObjectExtensions.cs
mamasselin: I did the change and it works. I agree that logic  should be in the DateTimeConverter class. Thanks!

src/GraphQL/ObjectExtensions.cs
```diff
@@ -123,7 +123,6 @@ public static Type GetInterface(this Type type, string name)
         public static object ConvertValue(object value, Type fieldType)
         {
             if (value == null) return null;
-            if (fieldType == typeof(DateTime) && value is DateTime) return value;
```
lngr: Bad.

Issue #275, opened 2017-02-05
First Character of Field Name becoming Lowercase


pravinhabbu4u:
@joemcbride I have node and graphql-dotnet implementation for GraphQL. graphql-dotnet implementation is automatically converting first character of field name to lower case. Is there any way to avoid this issue? I am currently using GraphQL (0.14.6.657) from Nuget. Below is an example query and output on node and dot-net core.

Node Input & Output example:

```
query TestForFieldCase{
  Identity(id:"b4d13602-57f6-46da-b071-b4e35ddd272b"){
    Data{
      IdentityID
   	  FirstName
      LastName
    }
  }
}
```

Output: Note: I also have a field named Data in response 

```
{
  "data": {
    "Identity": {
      "Data": {
        "IdentityID": "b4d13602-57f6-46da-b071-b4e35ddd272b",
        "FirstName": "PRAVIN",
        "LastName": "HABBU"
      }
    }
  }
}
```

GraphQL dot net implementation example:
Input:

```
query TestForFieldCase{
  identity(id:"b4d13602-57f6-46da-b071-b4e35ddd272b"){
    Data{
      identityID
   	  firstName
      lastName
    }
  }
}
```

Output:

```
{
  "data": {
    "identity": {
      "data": {
        "identityID": "b4d13602-57f6-46da-b071-b4e35ddd272b",
        "firstName": "PRAVIN",
        "lastName": "HABBU"
      }
    }
  }
}
```

joemcbride:
Short answer is no, that is not currently supported.  https://github.com/graphql-dotnet/graphql-dotnet/issues/169

pravinhabbu4u:
@joemcbride My hack attempt on at [https://github.com/graphql-dotnet/graphql-dotnet/issues/169#issuecomment-277821315](url) has given me another idea. Instead of renaming field with a prefix why cant we have another implementation for FIELD method as FIELD_UDC - user defined case. This method can avoid converting first character of field to lower case/ camel case. I need your advice on how you feel about this suggestion. I was able to successfully achieve the results wanted using this change. I can submit a pull request if it makes any sense to you.

       public FieldBuilder<TSourceType, TProperty> Field_UDC<TProperty>(
            Expression<Func<TSourceType, TProperty>> expression,
            bool nullable = false,
            Type type = null)
        {
            string name;
            try
            {
                name = expression.NameOf().ToUserCase();   // Retains user defined case than camel case
            }
            catch
            {
[...]

PR #277, opened 2017-02-07
Spike on supporting multi-cased fields and aliases

joemcbride:
#275

### DIFF
--- a/src/GraphQL.Tests/BasicQueryTestBase.cs
+++ b/src/GraphQL.Tests/BasicQueryTestBase.cs
@@ -1,0 +1,116 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using GraphQL.Types;
+using GraphQL.Http;
+using GraphQL.Validation;
+using GraphQLParser.Exceptions;
+using Shouldly;
+using Newtonsoft.Json.Linq;
+
+namespace GraphQL.Tests
+{
+    public class BasicQueryTestBase
+    {
+        protected readonly IDocumentExecuter Executer = new DocumentExecuter();
+        protected readonly IDocumentWriter Writer = new DocumentWriter(indent: true);
+
+        public ExecutionResult AssertQuerySuccess(
+            ISchema schema,
+            string query,
+            string expected,
+            Inputs inputs = null,
+            object root = null,
+            object userContext = null,
+            CancellationToken cancellationToken = default(CancellationToken),
+            IEnumerable<IValidationRule> rules = null)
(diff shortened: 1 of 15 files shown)


joemcbride:
Thoughts on this @tlil @jquense @pravinhabbu4u ?

Other idea I have is having a static class with these options.  That way the option wouldn't have to be passed from class to class.  Also with this implementation the "real" field name isn't determined until after the Schema is initialized.

pravinhabbu4u:
@joemcbride Thanks for the quick response. I tried running a test using casing branch but I was not able to query/ see output in user defined case. Am I supposed to pass an additional parameter while registering fields OR need to override something in my implementation.

joemcbride:
@pravinhabbu4u you will need to change the default FieldNameConverter on the config options of ExecuteAsync.  There is a PascalCaseFieldNameConverter and a DefaultFieldNameConverter.  You probably want the "default" one.  Currently the CamelCaseFieldNameConverter is registered as the default.  You can also implement your own IFieldNameConverter.

pravinhabbu4u:
@joemcbride This is great. I was able to run a quick test and achieve required results. Now I can now even have case sensitive fields
image

joemcbride:
@pravinhabbu4u Great to hear!  I need to do a little more testing on it and I'll wait for any other feedback.

pravinhabbu4u:
@joemcbride Sounds good. Thanks for your quick assistance. I hope new options will be available via nuget package soon.

tlil:
@joemcbride Sounds good to me. Might be a good idea to add a unit test to ensure that aliases aren't affected by `IFieldNameConverter`'s?

jquense:
sounds reasonable to me. might want to use more robust utils for the included case manipulations to cover a bit more edge cases. I'm partial to https://github.com/jquense/String utils :P

joemcbride:
@tlil Good idea to add more for those.  We have at least one or two tests that have aliases, though nothing that does checks around casing.

joemcbride:
@jquense Looks like your `StringUtils` library strips `_` so that messes with fields like `__type`. 😞

pravinhabbu4u:
In case anyone is looking for Default/ Pascal field name converter then use _.FieldNameConverter option on execute async:
`            return await new DocumentExecuter().ExecuteAsync(_ =>
            {
                _.Schema = schema;
                _.Query = graphqlRequest.Query;
                _.Inputs = inputs;
                _.FieldNameConverter = new DefaultFieldNameConverter();
            }).ConfigureAwait(false);`

Issue #252, opened 2017-01-06
FieldMiddleware leads to stack overflow


LasseNisted:
The DocumentExecuter.ExecuteAsync method invokes the config.FieldMiddleware on every call.
The default FieldMiddlewareBuilder then wraps all Resolve-calls in the schema.
So unless a new instance of the schema is used for each ExecuteAsync call then this inevitably leads to a stack overflow.
My understanding is that the schema should be considered static and reusable.

joemcbride:
Can you post a code example?  I have been using this without issue.  Perhaps it is related to how your system is setup?

LasseNisted:
Sure, it has to do with reusing a schema, so if I change the Execute method in GraphQL.Tests\Complexity\ComplexityTestBase.cs to:

        public async Task<ExecutionResult> Execute(ComplexityConfiguration complexityConfig, string query)
        {
            var starWarsSchema = new StarWarsTestBase().Schema;
            ExecutionResult result = null;
            for (int i = 0; i < 10000; i++)
            {
                result = await StarWarsTestBase.Executer.ExecuteAsync(
                    options =>
                    {
                        options.Schema = starWarsSchema;
                        options.Query = query;
                        options.ComplexityConfiguration = complexityConfig;
                    });
            }
            return result;
        }

To simply invoke the execution successively with the same schema, then I get a stack overflow in e.g. 
GraphQL.Tests.Complexity.ComplexityValidationTest.should_work_when_complexity_within_params

If I add the option:

                        options.FieldMiddleware = new FieldNoMiddlewareBuilder();

with the following dummy builder:

    class FieldNoMiddlewareBuilder : IFieldMiddlewareBuilder
    {
        public IFieldMiddlewareBuilder Use(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware)
        {
           throw new InvalidOperationException();
        }

        public FieldMiddlewareDelegate Build(FieldMiddlewareDelegate start = null)
        {
           throw new InvalidOperationException();
        }

        public void ApplyTo(ISchema schema)
        {
           // Do nothing.
        }
    }

Then the stack overflow disappears since it is due to the builder wrapping the resolve calls in the schema again and again.

I think the ExecuteAsync method should only apply the FieldMiddleware once (possibly by having an initialization step separate from the ExecuteAsync method).

ManuelDeLeon:
@joemcbride 

[...]

PR #282, opened 2017-02-21
Bug 252 - reuse of schema causes stack overflow

mxhammons:
As discussed in the thread, move ApplyTo call into the schema initialization section.  Simple tests to count the number of times FieldMiddlewareBuilders.ApplyTo() is called over the lifetime of an executor/schema/builder.


joemcbride:
Fixes #252

Issue #279, opened 2017-02-09
add DocumentWriter.ExposeExceptions


lanwin:
While debbuging and in test version its often very helpfull to see more infos when an error happened. Currently the DocumentWriter only exposes a simple error message like "Error trying to resolve foo". 

So I would suggest to add an option to the DocumentWriter which when set, exposes more internal information into the resolt, so its possible to guess what went wrong. This prop should be disble by default.

jquense:
I think this is a good idea generally. `graphql-express` has `formatError` which lets the user customize the mapping of GraphQL errors to the serialized ones

joemcbride:
Just as a note to anyone that may look at this, the errors are currently getting transformed via [ExecutionResultJsonConverter](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Execution/ExecutionResultJsonConverter.cs).

mahald:
If u really want to expose ur exceptions u can do this like this:

```
            #if DEBUG
            List<String> Errors = new List<string>();
            foreach (ExecutionError err in result.Errors)
            {
                Errors.Add(err.ToString());
            }
            Errors.ForEach(err => {
                ExecutionError exErr = new ExecutionError(err);
                result.Errors.Add(exErr);
            });
            #endif
```

Execute it after `DocumentExecuter().ExecuteAsync` and it will expose all exeptions on Debug Mode.

mahald:
I further improved it. I also got the wish as @jquense mentioned.
following solution let's u see fully stack trace in DEBUG mode AND Let's u throw error Messages in the Resolver that can be read:

firtst is passing any "GraphQL.ExecutionError" exception.
So u can `throw new ExecutionError("DISPLAY THIS ERROR MESSAGE");` in the resolver
second is the above and displays the full stack trace.
I replace the Errors of the result, so no double errors are in there.

```
            var result = await new DocumentExecuter().ExecuteAsync(_ => ....

            ExecutionErrors myExecutionErrors = new ExecutionErrors();

            result.Errors?.ToList().ForEach(err =>
            {
                var inner = err.InnerException?.InnerException;
                var type = inner?.GetType();
                if (type?.FullName == "GraphQL.ExecutionError")
                {
                    myExecutionErrors.Add(new ExecutionError(inner.Message));
                }
            });

            #if DEBUG
            result.Errors?.ToList().ForEach(err =>
                {
[...]

PR #295, opened 2017-03-10
added new error type in order to have a different behavior for developer triggered errors

mamasselin:
Hello,

I'am facing an issue with the error handling:
My API always return a generic error `error trying to resolve X` if external components go wrong (DB, validation error, etc).

The actual error that was raised is put in the inner exception of that generic error, but when the results are serialized in the response, we lose the inner exception.

Is there a reason to have this generic error "hiding" the orginal error? 
Because I use graphql in a nodejs app as well, and when I throw an error, it is directly sent in the response.

I added a new type of Execution error `ExternalExecutionError`. When this particular type of error is raised, it will not go in the `error trying to resolve X` scenario, it will instead take the triggered error's message.

In the end, everything functions as before, and if I throw an `ExternalExecutionError` in my code I get the behavior that I want. I also added a unit test for this.

What do you think?

Thanks,

### DIFF
--- a/src/GraphQL.Tests/Errors/CustomExecutionErrorTests.cs
+++ b/src/GraphQL.Tests/Errors/CustomExecutionErrorTests.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using GraphQL.Execution;
+using GraphQL.Types;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Errors
+{
+    public class CustomExecutionErrorTests : QueryTestBase<CustomExecutionErrorTests.TestSchema>
+    {
+        [Fact]
+        public async Task should_show_custom_error_when_manual_ExecutionError_is_thrown()
+        {
+            var body = "{\n    test\n}";
+            var result = await Executer.ExecuteAsync(
+                Schema,
+                null,
+                body,
+                null
+                );
+
+            result.Errors.Count.ShouldBe(1);
+            var error = result.Errors.First();
+            error.Locations.Count().ShouldBe(1);
+            error.Message.ShouldBe("Custom Error");
+        }
+
+        public class TestQuery : ObjectGraphType
+        {
+            public TestQuery()
+            {
+                Name = "Query";
+
+                Field<StringGraphType>()
+                    .Name("test")
+                    .Resolve(_ => { throw new ExternalExecutionError("Custom Error"); });
+            }
+        }
+
+        public class TestSchema : Schema
+        {
+            public TestSchema()
+            {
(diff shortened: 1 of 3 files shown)


joemcbride:
> Is there a reason to have this generic error "hiding" the orginal error?

Yes, the idea was to just expose generic errors as to not provide attack vectors.

What I would like to do is have a "turn on errors" feature so you can turn on returning the full error during development.  You can still send the "real" error to a log file in production.

Related: https://github.com/graphql-dotnet/graphql-dotnet/issues/279

jquense:
As it is you can already access the "real" inner exception on the Execution error right? I've just had some DEBUG switches in my http handler that formats errors differently depending on the environement.

mamasselin:
@joemcbride 
Thanks for the quick response.

To clarify my use case:

It is not for debugging purposes of Dev vs. Prod environments as you mentioned.

My use case is to have a way of displaying friendly error messages to the client applications that use my GraphQL API (the errors being sent in the response).

For example with the changes in this PR, I can do this in my item creation resolver:
`throw new ExternalExectionError("An item with the same name already exists.")`

Then in my client application where the user tried to create that item, I can display this error message "as is", or at least the client application knows why it failed.

Before my changes, any exception thrown would generate the exact same error in my client application ("error trying to resolve createItem").

As for the "attack vectors", the developer will have to specifically throw an `ExternalExecutionError` for the error to be sent "as is" to the client application. So the initial behavior still remains for all other cases.

@jquense 
Maybe I'm missing something but once the errors are serialized in the HTTP response, their innerException is lost. Maybe a configuration issue on my end?

joemcbride:
> their innerException is lost

Yes, that is on purpose.  You can override the behavior of [ExecutionResultJsonConverter](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Execution/ExecutionResultJsonConverter.cs) and have it expose the inner exception.

I don't want to introduce a special exception that developers would have to use.  I would instead like to provide what is mentioned in #279, and if that doesn't fit your scenario you can override the `ExecutionResultJsonConverter` behavior, perhaps with your own `IDocumentWriter`.

joemcbride:
I would be open to having some config settings for #279 that allowed you to just use and/or append the inner exception message and still exclude the stacktrace.  I think that would solve your issue?  Since you could still throw your own exception if you want but have the framework handle the output.

mamasselin:
Thanks, I will play around with this a bit and keep you posted on my findings.

mamasselin:
It turned our I implemented a small logic in my code to parse the errors before sending the response. This way I can receive the "real" errors in my client applications.

Thanks for your input guys, I will close this pull request.

PR #306, opened 2017-03-22
Added support for null input argument

HasAndries:
This will handle the new null input argument that will come from the parser change in this PR: https://github.com/graphql-dotnet/parser/pull/10
The issue for this change is https://github.com/graphql-dotnet/parser/issues/9

### DIFF
--- a/src/GraphQL/Language/CoreToVanillaConverter.cs
+++ b/src/GraphQL/Language/CoreToVanillaConverter.cs
@@ -222,6 +222,11 @@ namespace GraphQL.Language
                 case ASTNodeKind.ObjectValue:
                 {
                     var obj = source as GraphQLObjectValue;
+                    var scalar = source as GraphQLScalarValue;
+                    if (scalar != null && scalar.Value == null)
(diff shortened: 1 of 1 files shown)


HasAndries:
The build seems to be broken. Is this something that you are aware of?

joemcbride:
Yes, AppVeyor updated to VS2017 csproj and we're still on project.json.   This can potentially be fixed by adding a global.json until the project is updated to the new csproj format.  See #311

joemcbride:
Added the sdk to the `global.json` file, so if you rebase on master that should be fixed.

HasAndries:
@joemcbride Thanks for the fix, the build is passing. Any chance we can add this change sometime soon?

joemcbride:
Sorry for the delay on this, though I'm concerned that there would be more places that we would need to handle null, which is why it has lingered so long.  Have you looked at the JavaScript implementation to check its usage?

joemcbride:
Closing in favor of #374

PR #311, opened 2017-03-27
upgrade to netcore final sdk and VS2017, pushed netstandart to 1.3

OlegZee:
Resulting package could be referenced from netcoreapp projects with no pain. Switched projects to netcore build system (csproj).
Some tests might be failing, going to auto-check first.


jquense:
i will never understand why they removed the _finally_ hand editable/readable `.json` project file to go back to `xml` Does this require vs2017?

joemcbride:
As far as I understand, yes this would require VS2017 or VSCode.  I've been deving a lot on the Mac side of things lately so that would be VSCode or Rider.  This will also require some changes to the build scripts.  Any reason for the change to NetStandard 1.3?

jquense:
I'd selfishly prefer to not require vs2017, but i'm stuck at 2015 and dotnet 4.5.x for a while on some projects.

Still haven't gotten VSCode and the build chain to work on my Mac :P

joemcbride:
Thanks for this btw @OlegZee - My work is still on 2015 as well (for the most part), so I'm in no rush to upgrade yet.  So this may sit for a little while.  I think we may switch here fairly soon, if I had to guess I would say within a Month.

OlegZee:
In fact no reasons. I'm going to review why I did this and update PR.

OlegZee:
No, it just requires final dotnet SDK.
Don't worry, new XML projects are different, much much lighter that the old
ones.

After initial scepticism I got accomodated to it and I'd say I love it too.

OlegZee:
@joemcbride I tried to lower target fwk to netstandard1.1 and here're the findings.
The problem is that recent System.Reflection.TypeExtensions version 4.3.0 targets netstandard as lowest one. Version 4.0,0 references old tfm monickers such as netcore50.

Now I added PackageTargetFallback which is the same as `imports` in project json.
Btw here's good paper explaining project.json migration (http://www.natemcmaster.com/blog/2017/01/19/project-json-to-csproj/)

joemcbride:
Thanks for the work on this!  Having the build scripts updated in #327 - going to close this one in favor of that one.

Issue #298, opened 2017-03-14
FeatureRequest: ComplexityAnalysis - individual FieldImpact on GraphType fields


kantora:
Is is possible to set up individual FieldImpact on GraphType fields, so analysis will be more adequate? (as different field requests bring completely different load impact).

And am I right as it seems to me that now request complexity can be counted as `number of request graph nodes` * `field impact`, so defining both parameters is redundant (as you can just set FieldImpact as 1.0 and define maximum nodes number)?

joemcbride:
I'm honestly not sure, I haven't looked at that feature too in depth yet.  @dNetGuru ?

kantora:
As far as I see from the code, `ComplexityAnalyzer` recursively goes through the graph tree and have access to these fields. So adding FieldImpact to every field should not generate any additional load impact.
It can be `double?`, so for undefined case avgImpact can be used to maintain backward compatibility.

kantora:
But this will give very flexible analysis. As requesting pre-initialized scalar field costs almost nothing. Requesting something that leads to select from db with a large amount of data is complex. This select but with the use of some string filters is very complex. You just can't take an average for such complexity. They are too different.

dNetGuru:
@kantora The ```ComplexityAnalyzer```'s routine was written with two requirements in mind! To be fast (aimed for sub-millisecond runtime) and for the heuristic algorithm to make the least number of assumptions about the backend. The problem is that we need to have an approximation of how complex a given query is WITHOUT doing anything that can lock/take resources which would leave to a DoS vulnerability. The extreme case being writing some routine that takes as much time as it takes to actually execute the query! That would give us pretty accurate results, but would defeat the purpose!

I did experiment with various heuristics (some included a weight map passed into the function etc.) and ran a diverse sample of queries against them to find the best heuristic (that I could come up with) and the fastest one.

Now I do not claim that this is the absolute best approach, it's the one that satisfied the requirement that I had set forth. For this reason there exists a ```IComplexityAnalyzer``` interface if you need to implement your own custom made function that makes all the right assumptions about your sepecific backend and environment.

Making assumptions as to what operation takes more time or if the approximation that everything takes constant time on average would be too much of an assumption about the backend. Keep in mind we just need a good heuristic, not an accurate value. Unless if an operation takes 1000% more time than another I'd say that would be an okay estimation.

[...]

PR #312, opened 2017-03-29
Easier implementation of custom ComplexityAnalyzer

dNetGuru:
After discussion in #298 I tried to make it simpler to implement a custom analyzer.

Example usage is available in the test case.
In that example I implemented a **dummy** (the logic behind it is flawed for simplicity so don't use that in production) IComplexityAnalyzer type that goes through the AST and adds the value of all nodes together (values are passed through by a Dictionary object which maps field names to a value) and in absence of a value in the dictionary a default value of 1 is assumed.

You may use it as a template to build your own.

### DIFF
--- a/src/GraphQL.Tests/Complexity/CustomComplexityAnalyzerTest.cs
+++ b/src/GraphQL.Tests/Complexity/CustomComplexityAnalyzerTest.cs
@@ -1,0 +1,49 @@
+﻿using System.Collections.Generic;
+using GraphQL.Tests.Complexity.CustomComplexityAnalyzer;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Complexity
+{
+    public class CustomComplexityAnalyzerTest : CustomComplexityAnalyzerTestBase
+    {
+
+        [Fact]
+        public void should_work_when_complexity_within_params()
+        {
+            var query = @"
+                query HeroNameQuery {
+                  hero {
+                    name
+                  }
+                }
+            ";
+
+            var complexityConfiguration = new ExtendedComplexityConfig { MaxComplexity = 105, complexityMap = new Dictionary<string, double>() { { "hero", 100f } } };
+            var res = Execute(complexityConfiguration, query);
+
+            res.Result.Errors.ShouldBe(null);
+        }
+
+        [Fact]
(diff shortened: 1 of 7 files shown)


dNetGuru:
Build seems to fail due to unrelated reasons! Complaining about https://github.com/graphql-dotnet/graphql-dotnet/blame/master/src/GraphQL.Tests/GraphQL.Tests.xproj#L7 (same as another PR #306)
> dotnet build src/GraphQL.Tests  -c Release
.........SNIP.........
C:\projects\graphql-dotnet\src\GraphQL.Tests\GraphQL.Tests.xproj(7,3): error MSB4019: The imported project "C:\Program Files\dotnet\sdk\1.0.1\Microsoft\VisualStudio\v14.0\DotNet\Microsoft.DotNet.Props" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
Build FAILED.
.........SNIP.........
    0 Warning(s)
    1 Error(s)
Time Elapsed 00:00:00.16

Build environment problems?

joemcbride:
This is mostly pseudo code, though I was thinking something more like this:

```csharp
var config = new ComplexityConfig(_ =>

  _.For("hero", ctx => {
    var limit = ctx.Args<int>("limit");
    return ctx.childScore * limit;
  });

  _.Field<MyGraphType>("hero", ctx => {
    var limit = ctx.Args<int>("limit");
    return ctx.childScore * limit;
  });

);
```

I'm thinking that to target a specific field, you're going to need to know the type.  Since two different types can have the same field name.  So we may end up wanting to do something similar to validation using a `INodeVisitor` pattern to easily keep track of what type you're currently working with.

So the more I think about this it seems like the "better" route would be to add this logic to the field itself in the Schema and pull out that data while walking the tree in the complexity analysis.

dNetGuru:
I'll investigate this!

joemcbride:
Added the sdk to the `global.json` file, so if you rebase on master that should be fixed.

dNetGuru:
I was away for a while. So the main change in this PR is making the ComplexityConfiguration extensible so the user can pass through any info/object required to the analysis function they'd implement.

I'm also going through the other PR to test it out.

sungam3r:
@dNetGuru Will you continue to work on this PR?

PR #315, opened 2017-04-03
Precise complexity analyzer

kantora:
According to the discussion #298 
P.S.
I've noticed the `global.json` modification. But maybe there are some wildcards, as I am using sdk `1.0.0-preview2-003131` (vs configured `1.0.0-preview2-1-003177`). So it will be easier for developers to set up their environment.

### DIFF
--- a/src/GraphQL/Validation/PreciseComplexity/PreciseComplexityAnalyser.cs~RF437ae7.TMP
+++ b/src/GraphQL/Validation/PreciseComplexity/PreciseComplexityAnalyser.cs~RF437ae7.TMP
@@ -1,0 +1,27 @@
+﻿namespace GraphQL.Validation.PreciseComplexity
+{
+    using GraphQL.Language.AST;
+    using GraphQL.Types;
+
+    public class PreciseComplexityAnalyser
+    {
+        public Result Analyze(PreciseComplexityContext context, SelectionSet)
+        {
+            return new Result(0d, 0);
+        }
+
+        public class Result
+        {
+            /// <inheritdoc />
+            public Result(double complexity, int maxDeapth)
+            {
+                this.Complexity = complexity;
+                this.MaxDeapth = maxDeapth;
+            }
+
+            public double Complexity { get; }
+
+            public double MaxDeapth { get; }
+        }
+    }
+}
--- a/global.json
+++ b/global.json
@@ -1,3 +1,6 @@
 {
-    "projects": [ "src" ]
+    "projects": [ "src" ],
+    "sdk": {
+    	"version": "1.0.0-preview2-003131"
+    }
 }
--- a/src/GraphQL.Tests/PreciseComplexity/PreciseComplexitySchema.cs
+++ b/src/GraphQL.Tests/PreciseComplexity/PreciseComplexitySchema.cs
@@ -2,6 +2,26 @@
 {
     using GraphQL.Types;
 
+    public class RootInterface : InterfaceGraphType
+    {
+        public RootInterface()
+        {
+            this.Name = "RootInterface";
+            this.Field<StringGraphType>("string");
+            this.Field<RootInterface>("this");
+        }
+    }
+
+    public class NonRootInterface : InterfaceGraphType
+    {
+        public NonRootInterface()
+        {
+            this.Name = "NonRootInterface";
+            this.Field<NonRootInterface>("this");
+            this.Field<IntGraphType>("int");
+        }
+    }
+
     public class RootQuery : ObjectGraphType
     {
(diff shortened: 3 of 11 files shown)


joemcbride:
re `global.json`: I am doubtful that the `global.json` supports wildcards.  You can have multiple sdks installed at once so I would go that route.  That being said, all of the `.json` stuff is deprecated  in the .NET world, so once this project is upgraded to the new `csproj` all of that will be going away.

Some early comments on this - there is code already available that works with walking a query tree - validation accomplishes this with a [`BasicVisitor`](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Validation/BasicVisitor.cs) class.  Have a look at the Validation rules.  The associated [`TypeInfo`](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Validation/TypeInfo.cs) class handles things such as tracking what `GraphType` is associated with a given field.  You can re-use some of that infrastructure which would alleviate some of the issues you ran into with creating additional dependencies between classes (such as creating a dependency to the `DocumentExecutor`).  This would also allow a more declarative programming style which I think would be much easier to maintain.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Validation/DocumentValidator.cs#L47-L56

Declarative validation rule:

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Validation/Rules/ArgumentsOfCorrectType.cs#L20

kantora:
The main problem with Validators - in current solution they don't have an access to the variables values. So they can't resolve field argument values properly. Now the only correct solution to resolve arguments are in the execution context (I had to make my own resolver in this proposition). Suppose that this should be extracted to som other entity to be available also in the validators and/or complexity analyzer.

The second option - you told me that complexity analysis should be done before validation to avoid validation overload. Is this still correct?

I can try to rewrite it to Validators but will have to update them to gain access to the variables. The field complexity parameter can be also moved to field metadata with easy access through extension methods.

In case you prefer the solution from this PR, the DocumentExecutor dependency can be easily removed with DocumentExecutor refactoring. Some of its methods are really static by their body and can be moved in a separate utility class (and only such method are used in PreciseComplexityAnalyser).

So please - tell me what the way is preferable,

joemcbride:
No, I do not want this added to the validation code. I am trying to point out is that you can _write new code_ that uses some of the same classes and patterns that validation uses.  I added links to those classes in my previous post.  You can add variable/argument access to that new code.  We also have established patterns for accessing those values so those should be followed, see  [`GetArgument`](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/ResolveFieldContext.cs#L65) on the `ResolveFieldContext`.

The `TypeInfo` class should provide all of the information needed so you wouldn't need that method in the `DocumentExecutor`.  If it doesn't, the `TypeInfo` class should be updated so it does.

kantora:
Ouch, I haven't read your comment accurate. Sorry.
Thanks for response - I'll look at it an make an update.

RehanSaeed:
My understanding is that the default complexity of a field is `1`. Would it also be useful to simply be able to set the complexity of a field as a number as in `SetComplexity(2)`, as well as the delegate you've added?

src/GraphQL/Validation/PreciseComplexity/PreciseComplexityAnalyser.cs~RF437ae7.TMP
```diff
@@ -0,0 +1,27 @@
+﻿namespace GraphQL.Validation.PreciseComplexity
+{
+    using GraphQL.Language.AST;
+    using GraphQL.Types;
+
+    public class PreciseComplexityAnalyser
```
sungam3r: Tmp?

sungam3r:
@kantora Would you like to update this PR after v4 release?

Shane32:
> The main problem with Validators - in current solution they don't have an access to the variables values.

They will in v4

christianrondeau:
Is this PR dead? Being able to mark fields as more or less expensive could be very nice, in some cases we can compute values locally, and in some other cases we have to call external, potentially slow or throttled services so being to do .SetComplexity(N) would be great (average number before actual execution). If the original author is away, would you consider taking a new PR for v5 based on this?

sungam3r:
PRs are always wellcome. Just keep in mind that they can wait for a review for a long time.

sungam3r:
Closed as outdated. Also see #3159 though I agree that `ComplexityAnalyzer` still has wide room for improvements.

PR #347, opened 2017-06-07
Compile and reuse resolver expression

dlukez:
Expression compilation is apparently expensive - this change alone yielded a ~10x speed improvement on my machine (a 15" MacBook Pro 2016) while testing the [DataLoader.StarWars](https://github.com/dlukez/dataloader-dotnet/tree/master/sample/DataLoader.StarWars) sample app.


joemcbride:
Finally got this published. https://www.nuget.org/packages/GraphQL/0.16.0

Issue #302, opened 2017-03-16
Decimal and float conversions should use invariant culture (unit tests fail)


schlaup:
First, very nice work!!!

I found one issue:

When the OS is not "en-US" float and decimal values are converted incorrectly.

For example:

```
Failed: GraphQL.Tests.Types.DecimalGraphTypeTests.coerces_numeric_string_to_decimal
  type.ParseValue("12345.6579")
    should be
12345,6579m
    but was
123456579m
```

or

```
Failed: GraphQL.Tests.Bugs.Bug138DecimalPrecisionTests.double_to_decimal_does_not_lose_precision
  writtenResult
    should be
"{
  "data": {
    "set": 24.15
  }
}"
    but was
"{
  "data": {
    "set": 2415.0
  }
}"
```

To reproduce set numeric format of Windows to "German".

or set culture in the test. For example:

```
        [Fact]
        public void coerces_numeric_string_to_decimal()
        {
            System.Threading.Thread.CurrentThread.CurrentCulture = new System.Globalization.CultureInfo("de-DE");
            type.ParseValue("12345.6579").ShouldBe((decimal)12345.6579);
        }
```

(Better in an XUnit initializer for all tests. Or something like https://coding.abel.nu/2014/02/rerunning-all-unit-tests-with-different-ui-culture/)

The fix is not difficult. Use InvariantCulture when parsing decimal or floats, for example:

```
        public override object ParseValue(object value)
        {
            decimal result;
            if (decimal.TryParse(value?.ToString() ?? string.Empty, out result))
            {
                return result;
            }
            return null;
        }
```

should be:

```
        public override object ParseValue(object value)
        {
            decimal result;
            if (decimal.TryParse(value?.ToString() ?? string.Empty, NumberStyles.Any, CultureInfo.InvariantCulture, out result))
            {
                return result;
            }
            return null;
        }
```

joemcbride:
👍  Mind doing a PR to fix this?

schlaup:
Ok, the fix is easy but I am not so familiar with xUnit.
How would you want to setup additional tests?

Something like: ?
https://coding.abel.nu/2014/02/rerunning-all-unit-tests-with-different-ui-culture/

The alternative would be to write some additional tests that only cover the tests that are failing now.

joemcbride:
I'm not sure of a good way to run all tests under a specific culture, especially in the .NET Core world.  I would be fine just writing some tests for this specific issue for now.

[...]

PR #353, opened 2017-06-16
DecimalGraphType parsing fixed when thread culture decimal point is ','

mule:
Fix for issue #302  (which was closed for some reason).


joemcbride:
@mule  Thanks for the PR! #302 was not closed ... there was a duplicate report that was closed though.  Can you please add tests that run this code under a non-US culture to verify this behavior?

schlaup:
@mule Thanks for working on this! I've never gotten to it. In your commit "Improved conversion, decimal to decimal conversion now works for different cultures" you allow parsing of the current server's culture. I believe this is too much magic. For example, someone might accidently input a number with thousand-separator. (US: 11,000) If your implementation runs on a DE server it would convert this to "11.000" and allow the value. I believe, it should always throw an error. The GraphQL API should always use invariant culture and should not allow other formats if it runs on a server with a different locale setting.

pekkah:
Is this being worked on? Hitting this issue while running unit tests.

pekkah:
This can be closed now?

joemcbride:
Yes, this PR was superseded by #434

Issue #336, opened 2017-05-15
Implicitly Resolve List<T> Field Types using Field(x => x.MyField)


arianmotamedi:
Hello,

I have a field in my class that is of type `List<string>`, but GraphQL is having a hard time determining the type during field registration:

`{"The GraphQL type for Field: 'MyField' on parent type: 'MyType' could not be derived implicitly. \n"}`

I'm registering it just like any other field:

`Field(x => x.MyField, true);`

Again, type of MyField is `List<string>`. What's the correct way to register lists of simple types? Apologies if this sounds like a trivial question  but I've been stuck on it for a while now...

jquense:
The inference will work for arrays, e.g. `string[]`. You can manually specify the GQL type in your field declaration tho.

arianmotamedi:
If I try to specify the type `List<string>` or `List<>` as a parameter during registration I get another error saying `The declared Field type: List`1 should derive from GraphType, but doesn't.` Seems like it only takes type inheriting from GraphType. I'm surprised it doesn't infer `List<T>` since frankly that is more commonly used than `T[]`...

jquense:
You have to specify the GraphQLType not the data type it resolves from, so `field<ListGraphType<StringGraphType>>()`. The inference isn't there because no one has added it ;) PR's are always welcome.

arianmotamedi:
Cool thanks. I might open up a PR one of these days :smile:

huysentruitw:
When you think about it, `List` just isn't so great to use as input/output type, because you want something more immutable.

stephenlautier:
I believe this can be closed as its already implemented as @jquense suggested.

joemcbride:
@stephenlautier This is a feature request to add `List<T>` to [`GetGraphTypeFromType`](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/TypeExtensions.cs#L68) so it can be used in the [`Field(x=>x.ListProperty)`](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/ComplexGraphType.cs#L144) syntax.

stephenlautier:
@joemcbride ahh my apologies. That's even better  👍

joemcbride:
This was fixed in #359 and published in 0.16.0.

PR #359, opened 2017-07-06
List<> types supported in Field(x => x.MyList) syntax

rmc00:
Small fix for Issue #336


joemcbride:
Got this published.  https://www.nuget.org/packages/GraphQL/0.16.0

PR #362, opened 2017-07-08
Initial work for supporting a schema builder

joemcbride:
Hello World example:

```csharp
var schema = Schema.For(@"
    type Query {
      hello: String
    }
");

var root = new { Hello = "Hello World!" };
var result = schema.Execute(_ =>
{
    _.Query = "{ hello }";
    _.Root = root;
});

Console.WriteLine(result);
```

Handler/Endpoint examples:

```csharp
// classes
public class Post
{
    public Guid Id { get; set; }
    public string Body { get; set; }
}

public class Query
{
    [GraphQLMetadata("post")]
    public Post GetPost(Guid id)
    {
        return new Post { Id = id, Body = "Something" };
    }
}

var schema = Schema.For(@"
  type Post {
    id: String
    body: String
  }

  type Query {
    post(id: ID!): Post
  }
", _ => {
    _.Types.Include<Query>();
});

// execute
var id = Guid.NewGuid(); 
var result = schema.Execute(_ =>
{
    _.Query = $"{{ post(id: \"{id}\") {{ id body }} }}";
});
```


joemcbride:
After some thought, I want to keep the `SchemaBuilder` class in the core repo.  This is mainly so we can switch the "hello world" example to something much simpler and people can get started without so much overhead.

Any thoughts @jquense @tlil ?

```csharp
var schema = Schema.For(@"
    type Query {
      hello: String
    }
");

var root = new { Hello = "Hello World!" };
var result = schema.Execute(_ =>
{
    _.Query = "{ hello }";
    _.Root = root;
});

Console.WriteLine(result);
```

I have some "Authorization" code in this branch, though I'll move that to its own repo/nuget since it requires .NET 4.6 and NetStandard 1.3.

tlil:
Makes perfect sense to me @joemcbride 👍  No objections from my end

tlil:
This is starting to look good @joemcbride - will take a closer look over the weekend!

PR #374, opened 2017-07-17
Added support for null input arguments (Improved)

AndyBan:
Replaces https://github.com/graphql-dotnet/graphql-dotnet/pull/306

The other pull looks to be dead and has issues which I have resolved.

Requires https://github.com/graphql-dotnet/parser/pull/17

Build will fail until the other PR is merged, let me know if you need anything else from me.


joemcbride:
Thanks for the PR!  You can fix the build by temporarily adding the myget.org nuget source to use your changes and make sure all the tests pass.

https://www.myget.org/feed/graphql-dotnet/package/nuget/GraphQL-Parser

AndyBan:
I think your CI is broken? The myget feed hasn't updated in a few days.  Appveyor has the package but the feed asks me for a password so I guess that isn't public.  If I download it and use it in a local feed then all the tests pass.

https://i.imgur.com/VfPygy4.png

joemcbride:
Ah, yeah - its not broken, but the myget deploy is restricted by branch which I forgot about.  You could perhaps add your branch to the [`appveyor.yml`](https://github.com/graphql-dotnet/parser/blob/master/appveyor.yml#L38), but if you did it manually then I'm not worried about it.  Will have this CI re-run once a new package is officially released.

src/GraphQL/Language/AST/NullValue.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace GraphQL.Language.AST
+{
+    public class NullValue : AbstractNode, IValue
+    {
+
+        public string Value { get; }
```
joemcbride: I think we don't need this `Value` property at all?

joemcbride:
Some other places we probably want to update:

* [`AstFromValue`](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/GraphQLExtensions.cs#L309) extension method.
* [`AstPrinter`](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/AstPrinter.cs#L254)
* [`SchemaPrinter`](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/SchemaPrinter.cs) - may need to look at the graphql-js project and see what they do with `NullValue`.  Perhaps only need to worry about it with `DefaultValue` argument values?

AndyBan:
I have resolved the first two comments.  I don't think the schema printer needs to know about the NullValue AST object as it can never end up with one.  The printer will only deal with actual types or types wrapped in non null never actual values (Ignoring the default which might be null but that's the value not the AST type).

The JS project just passes null around being untyped which I did consider doing here but it does make the code a bit more messier in C# land as you end up changing int to int? etc and then having to check everywhere where you use that value that you have a value.

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -264,6 +264,12 @@ public AstPrintVisitor()
                 c.Field(x => x.Value);
                 c.Print(f => f.Arg(x => x.Value));
             });
+
+            Config<NullValue>(c =>
+            {
+                c.Print(f => null);
```
joemcbride: This should print the text `null` vs. just a null value.

```
c.Print(f => "null");
```

joemcbride:
It looks like the JS project uses [`astFromValue` to convert the `DefaultValue`](https://github.com/graphql/graphql-js/blob/master/src/utilities/schemaPrinter.js#L236).  That looks like an existing bug in our `SchemaPrinter`.  No need to fix that now if you don't want.

src/GraphQL/Utilities/AstPrinter.cs
AndyBan: Are you sure?  If you do that you end up with null in a string which correctly fails a couple of unit tests:

Message: Assert.Equal() Failure
                                 ↓ (pos 29307)
Expected: ···    "defaultValue": null\r\n          }\r\n        ]\r\n      },\r\n ···
Actual:   ···    "defaultValue": "null"\r\n          }\r\n        ]\r\n      },\r···
                                 ↑ (pos 29307)

AndyBan:
A bit pushed for time at the moment so I'll hold off on that one.

I tried changing the parser ci branch config but after looking into it they have disabled deployment for pull requests for non trusted people.

src/GraphQL/Utilities/AstPrinter.cs
joemcbride: Yes, that matches the JS project behavior.

https://github.com/graphql/graphql-js/blob/master/src/language/printer.js#L79

AndyBan: Patch pushed

AndyBan: Let me know if you need it squashing once your happy

AndyBan:
I have renamed null -> Nullvalue as reviewed on the parser repo and squashed all the commits.  Let me know if there is anything blocking this from being merged.

AndyBan:
I have pointed to the custom MyGet feed now the Parser change has been merged.  Let me know if you want that commit removing but I think it might be left in?

joemcbride:
It will need to be removed once an official release of v3 of the parser is on Nuget.org. I want to make a few tweaks to the parser project before officially releasing it. I'll do that today.

tlil:
> I want to make a few tweaks to the parser project before officially releasing it. I'll do that today.
@joemcbride did you make any progress on that?

joemcbride:
@tlil yes, see th dev branch in that project.

enkodellc:
What is the status of this? Wouldn't this reduce the need for multiple GraphQL queries for arguments. Can the feature be applied to Variables as well? Thanks for all the work on the project. I hope to learn a bit more and contribute.

AndyBan:
Its been ready for merging for 6 weeks now afaik - I think @joemcbride is waiting for some other unrelated features to do a major version bump.  I don't think this feature requires a major version bump itself.  

Variables look to be working correctly with my PR.

joemcbride:
The issue is the building of the Parser nuget package.  That build process had to be changed and I haven't been able to 100% verify its working yet.  Since this is a behavior change in the parser, yes it requires a major version bump.  Ditto for this project.

https://github.com/graphql-dotnet/parser/pull/18

joemcbride:
Got the parser project updated and added #421 which merges these changes.

src/GraphQL.Tests/StarWars/StarWarsIntrospectionTests.cs
```diff
@@ -369,7 +369,7 @@ public void allows_querying_field_args()
                             'kind': 'SCALAR'
                           }
                         },
-                        'defaultValue': null
+                        'defaultValue': 'null'
```
holm0563: This change is making GraphIQL not able to parse the introspection. Is anyone else having this problem?

holm0563: Note after reverting this change set, graph iql works as expected.

holm0563: @joemcbride should this be sent as a new issue?

joemcbride: @holm0563 may need to bump the GraphiQL version to one that supports the null keyword perhaps?

holm0563: @joemcbride i didn't look at the spec but I'd be shocked if its truly supposed to change to 'null' as a string?

joemcbride: AH, I see what you're saying.

https://github.com/graphql/graphql-js/blob/4f5e351c2afb28e4ae3a44eae76d45d895ff3aa2/src/utilities/astFromValue.js#L70-L78

The JS project treats `null` different than `undefined` and `NaN`.  We don't have the "luxury" of 3 different "nulls" in C#.  Will have to decide on the best course of action here.

PR #380, opened 2017-07-30
[WIP ] Automatically extract and pass parameters from context to resolver

Giorgi:
Inspired by ASP.NET MVC model binder I added FuncFieldModelBinderResolver which inspects parameters of the resolver function, extracts them from the context arguments (by matching parameter name to argument name) and passes them to the resolver function. As a result instead of calling `context.GetArgument<string>("id")` you can just add a `string id` parameter to resolver function.

This is still work in progress but I wanted to get feedback from community before going too far with it. I find it useful because it allows to get rid of the magic strings in resolver function. The only downside is that you need to declare and intermediate variable for the resolver function before passing it as a resolver.

So what do you think ?

### DIFF
--- a/src/GraphQL/Resolvers/FuncFieldModelBinderResolver.cs
+++ b/src/GraphQL/Resolvers/FuncFieldModelBinderResolver.cs
@@ -1,0 +1,62 @@
+using System;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using GraphQL.Types;
+
+namespace GraphQL.Resolvers
+{
+    public class FuncFieldModelBinderResolver<TSourceType, TReturnType> : IFieldResolver<Task<TReturnType>>
+    {
+        private readonly Delegate _resolver;
+
+        public FuncFieldModelBinderResolver(Delegate resolver)
+        {
+            if (resolver == null)
+            {
+                throw new ArgumentNullException("A resolver function must be specified");
+            }
+            _resolver = resolver;
+        }
+
+        public async Task<TReturnType> Resolve(ResolveFieldContext context)
+        {
+            var parameters = RuntimeReflectionExtensions.GetMethodInfo(_resolver).GetParameters();
+
+            int index = 0;
+            var arguments = new object[parameters.Length];
+
+            if (context.As<TSourceType>().GetType() == parameters[0].ParameterType)
+            {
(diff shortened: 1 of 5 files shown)


src/GraphQL/Resolvers/FuncFieldModelBinderResolver.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            if (context.As<TSourceType>().GetType() == parameters[0].ParameterType)
+            {
+                arguments[index] = context.As<TSourceType>();
+                index++;
+            }
+
+            foreach (var parameter in parameters.Skip(index))
+            {
+                arguments[index] = context.GetArgument<object>(parameter.Name);
```
joemcbride: I think this should be changed to actually pass the `Type` of the parameter so it could cast/convert it properly.  You would need to refactor `GetArgument` so you could pass in a `Type`, and have `GetArgument<T>` delegate to that new method.

Giorgi: Will give it a try tomorrow.

src/GraphQL/Resolvers/FuncFieldModelBinderResolver.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        public async Task<TReturnType> Resolve(ResolveFieldContext context)
+        {
+            var parameters = RuntimeReflectionExtensions.GetMethodInfo(_resolver).GetParameters();
+
+            int index = 0;
+            var arguments = new object[parameters.Length];
+
+            if (context.As<TSourceType>().GetType() == parameters[0].ParameterType)
```
Giorgi: Agreed, it does need more error checking.

Giorgi:
Thanks for your feedback. I added a new overload because without it I got lots of compilation errors because the compiler was unable to determine which overload to use. We can update all the code in our project but it will cause compilation error in other code too which uses this library.

If you have any suggestion how to avoid compilation error and keep the `Field` overload that will be great.

Giorgi:
@joemcbride Can you review the updated PR ?

Giorgi:
Done

joemcbride:
This is released in [0.17.2](https://github.com/graphql-dotnet/graphql-dotnet/releases/tag/v0.17.2)

PR #386, opened 2017-08-05
Added AutoRegisteringObjectGraphType

Giorgi:
Added a new subclass of `ObjectGraphType` which automatically registers scalar properties as fields. You can also override `GetPropertiesToRegister` method to filter which properties should be registered.

### DIFF
--- a/src/GraphQL.StarWars/Types/DroidType.cs
+++ b/src/GraphQL.StarWars/Types/DroidType.cs
@@ -2,22 +2,18 @@ using GraphQL.Types;
 
 namespace GraphQL.StarWars.Types
 {
-    public class DroidType : ObjectGraphType<Droid>
+    public class DroidType : AutoRegisteringObjectGraphType<Droid>
     {
         public DroidType(StarWarsData data)
         {
             Name = "Droid";
             Description = "A mechanical creature in the Star Wars universe.";
-
-            Field(d => d.Id).Description("The id of the droid.");
-            Field(d => d.Name, nullable: true).Description("The name of the droid.");
-
+            
             Field<ListGraphType<CharacterInterface>>(
                 "friends",
                 resolve: context => data.GetFriends(context.Source)
             );
             Field<ListGraphType<EpisodeEnum>>("appearsIn", "Which movie they appear in.");
-            Field(d => d.PrimaryFunction, nullable: true).Description("The primary function of the droid.");
 
             Interface<CharacterInterface>();
         }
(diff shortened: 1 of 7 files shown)


Giorgi:
The introspection test failed for two reasons:

1. The order of the fields is now different
2. Fields `name` and `primaryFunction` are now non nullable.

joemcbride:
@Giorgi I would like to keep those examples as they are and add some different ones with unit tests.  I'm also leery of calling a virtual method in a constructor.  I don't think I would want to handle the support requests on that one.  I can understand wanting to override that, though that is something someone could just add to the `ctor` of the subclass and not need to add an override for.

Giorgi:
@joemcbride Yes, I agree that it's better to add new examples instead of touching existing ones. As for virtual method in a ctor, yes it's not a best practice but if the base ctor registers all the properties it will be too late to do anything in the constructor of the subclass.

On the other hand we can add some kind of option to opt-out of bulk registration of the fields and allow to specify which properties to register in bulk. Something similar to Entity Framework OnModelCreating which allows to customize configuration.

Giorgi:
@joemcbride What do you think about the second approach that I mentioned?

joemcbride:
@Giorgi Since the amount of code to do this is fairly minimal, I don't think I want to introduce an API for this.  My guidance to users would be to just create their own class.

Giorgi:
OK, I see. Do you want to keep it as it is or inline the virtual method ?

joemcbride:
I would just inline it.

Giorgi:
@joemcbride Done.

matjazmav:
@Giorgi Good job. I have just 2 comments. I think `Guid` type resolver would be useful, just convert `Guid` to `string`. Another thing is that reflection is slow, maybe we could cache instances of  `ObjectGraphType`s.

```cs
    public class CachableDependencyResolver : IDependencyResolver
    {
        private IObjectGraphTypeCache _cache;

        public GenericDependencyResolver(IObjectGraphTypeCache cache)
        {
            _cache = cache;
        }

        public T Resolve<T>()
        {
            return (T)Resolve(typeof(T));
        }

        public object Resolve(Type type)
        {
            if(!_cache.TryGetGraphType(type, out var graphType))
            {
                graphType = type
                   .GetConstructor(new Type[] { })
                   .Invoke(new object[] { });

                _cache.AddGraphType(type, graphType);
            }

            return graphType;
        }
    }
```

`IObjectGraphTypeCache ` is just wrapper around `IMemoryCache`.

I have tested few queries with `CachableDependencyResolver ` response times are between 200 ms and 300 ms, with default `IDepandencyResolver` response times are between 500 ms and 800 ms.

PR #391, opened 2017-08-11
Added support for ExposeExceptions Option

mahald:
Added support for ExposeExceptions Option in the DocumentExecuter


joemcbride:
Hey, thanks for this PR, though this doesn't handle a few other places where the errors are set.  This transformation should be happening elsewhere.  Right now that is in the `ExecutionResultJsonConverter`.  An idea I have is to have that `ExposeExceptions` property on the `ExecutionResult` class.  That way you can check that value in the `ExecutionResultJsonConverter`.  I don't really like the idea of copying that value around, though I'm not thinking of a better idea at current.  I'm looking into this a bit more today.

PR #409, opened 2017-08-27
upgraded to dotnet core 2.0

srini85:
Upgraded project to 2.0

- Ran tests and they all passed


joemcbride:
Hey, thanks for this, though I would like to stay targeted at NetStandard1.1/net45 as long as possible.  What I would like to see is if just updating to the new sdk but staying at 1.1 will work.

May also need to add a global.json to specify the target sdk for AppVeyor.

srini85:
Hi Joe, do you think it is worthwhile creating a 2.0 branch and having it at Dotnet core 2.0 this will ensure we can get the latest features or Dotnet core 2

srini85:
Looks like build is also failing on appveyor. Seems to build fine locally :S

joemcbride:
I wouldn't want to maintain a 2.0 branch.  If I were targeting 2.0 I would have that be master and have a 0.x branch as maintenance.    You _should_ be able to target 1.1 with the 2.0 sdk.

This is an example global.json:

https://github.com/JasperFx/marten/blob/master/global.json

The build script would have to also be updated (may be more than one place):

https://github.com/graphql-dotnet/graphql-dotnet/blob/4da2d8af548b4f0d9cd620da2504781861cf6e37/tools/tasks/compile.js#L9

srini85:
@joemcbride thanks for the info

srini85:
@joemcbride - noticed the appveyor server has dotnet sdk 1.0 installed. Would have to get this upgraded if this was to be built in 2.0 (if required down the line)

joemcbride:
@srini85 That was the intent of the `global.json` file - to tell AppVeyor which sdk version to use.  There should only be a single `global.json` file in the root of the repo.  A few days ago AppVeyor was defaulting to the 2.0 sdk.  Perhaps that changed?  I don't know.

tools/tasks/compile.js
```diff
@@ -5,8 +5,8 @@ import settings from './settings';
 export default function compile() {
   const deferred = new Deferred();
 
-  const platform = process.platform === 'darwin'
-    ? '-f netcoreapp1.1'
+  const platform = process.platform === ('darwin' || 'win32')
+    ? '-f netcoreapp2.0'
     : '';
```
joemcbride: The purpose of these lines of code was to only build/test .NET Core when on macOS.  This switches it to build only .NET Core on both macOS and Windows, which isn't the wanted behavior.  When on Windows it should build both .NET Core and the full-framework .NET version (as it was previously).

tools/tasks/test.dotnet.js
```diff
@@ -5,8 +5,8 @@ import settings from './settings';
 export default function testDotnet() {
   const deferred = new Deferred();
 
-  const platform = process.platform === 'darwin'
-    ? '-f netcoreapp1.1'
+  const platform = process.platform === ('darwin' || 'win32')
+    ? '-f netcoreapp2.0'
     : '';
```
joemcbride: Same as previous comment.

src/GraphQL.Tests/global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "2.0.0"
+  }
+}
```
joemcbride: This should be moved to the root of the repo.

srini85:
Thanks for the reviews and comments. Will look at them

pekkah:
Is this going to get merged into master anytime soon?

joemcbride:
@pekkah No, it has several issues.  Most notably removing the net45 and Netstandard 1.1 support (which should be able to be multi-targeted).  I'm also stuck on 1.0.4 sdk on Windows due to some bugs in latest VS2017, though I could probably switch to 2.0 on macOS.  Just low priority at the moment.

KhalilMohammad:
Removing .Net 4.5 support should not be a problem . Most applications can be upgraded to .Net 4.6.1 seamlessly. [Migration guide for .Net 4.5 , .Net4.6 and .Net 4.7](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/)

jquense:
I'm stuck on 4.5 as well tho I am fairly certain you can multi target without issue, I've done that in the last bits of the Relay code

pekkah:
PR with standard2.0 was merged to master while keeping the old targets.

PR #425, opened 2017-09-12
support variables for types that return custom objects

csantero:
I've been trying to create a custom ScalarGraphType that would accept a JSON-encoded string, and output a deserialized object corresponding to that string, using JSON.NET. I was getting an error due to the fact that the `Conversions` class was trying to convert the value returned from the ScalarGraphType, even though there was no IConversionProvider registered for my custom type. I solved it by bypassing the attempt to convert when no provider is registered for the type.

### DIFF
--- a/src/GraphQL.Tests/Execution/VariablesTests.cs
+++ b/src/GraphQL.Tests/Execution/VariablesTests.cs
@@ -1,8 +1,10 @@
-﻿using System.Linq;
+using System.Collections.Generic;
+using System.Linq;
 using GraphQL.Language.AST;
 using GraphQL.Types;
 using GraphQL.Validation;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Shouldly;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


src/GraphQL.Tests/Execution/VariablesTests.cs
```diff
@@ -499,6 +552,34 @@ public void allows_non_nullable_inputs_to_be_set_to_a_value_directly()
...
+            {
+                ["stringProperty"] = "bear",
+                ["arrayProperty"] = new JArray {"cat", "dog", "bird"}
+            }.ToString();
+
+            var inputs = new Inputs(new Dictionary<string, object>
+            {
+                ["input"] = jsonString
+            });
+            AssertQuerySuccess(query, expected, inputs);
```
joemcbride: This is non-standard creation of `Inputs`.  All of the values should be converted from JSON.NET types to base types (string, int, etc.).  Have you ran this using the `.ToInputs` extension method on your input variables without your change?  Does that make it needed?

csantero: @joemcbride I just did it that way to make the test easier to write. I've pushed a new commit that uses ToInputs.

csantero:
@joemcbride Thanks, I added a commit to do that instead.

PR #427, opened 2017-09-13
Subscription Executer

pekkah:
# Description

Provides a `SubscriptionExecuter` which allows subscribing to a source stream provided by the `EventStreamFieldType`. Returned stream is of type `IObservable<object>` allowing easy manipulation.

## Examples

See the "unit tests": https://github.com/pekkah/graphql-dotnet/blob/features/subscription/src/GraphQL.Tests/Subscription/SubscriptionTests.cs

## Where is the transport?

I'll be pushing out a separate NuGet/repo with the transport layer for Apollo Client compatibility after this makes it into the official GraphQL NuGet so I can just add it as an dependency.


pekkah:
Have to do a rebase. Will post updated pull request later.

PR #428, opened 2017-09-13
Subscription Executer

pekkah:
# Description

Provides a `SubscriptionExecuter` which allows subscribing to a source stream provided by the `EventStreamFieldType`. Returned stream is of type `IObservable<object>` allowing easy manipulation.

## Examples

See the "unit tests": https://github.com/pekkah/graphql-dotnet/blob/features/subscription/src/GraphQL.Tests/Subscription/SubscriptionTests.cs

## Where is the transport?

I'll be pushing out a separate NuGet/repo with the transport layer for Apollo Client compatibility after this makes it into the official GraphQL NuGet so I can just add it as an dependency.

### DIFF
--- a/src/GraphQL.Tests/Subscription/SubscriptionTests.cs
+++ b/src/GraphQL.Tests/Subscription/SubscriptionTests.cs
@@ -45,6 +45,7 @@ namespace GraphQL.Tests.Subscription
             message.ShouldNotBeNull();
             message.ShouldBeOfType<ExecutionResult>();
             message.Data.ShouldNotBeNull();
+            message.Data.ShouldNotBeAssignableTo<Task>();
         }
 
         [Fact]
--- a/src/GraphQL.Tests/Subscription/SubscriptionTests.cs
+++ b/src/GraphQL.Tests/Subscription/SubscriptionTests.cs
@@ -4,7 +4,7 @@ using System.Linq;
 using System.Reactive.Linq;
 using System.Threading.Tasks;
 using GraphQL.Subscription;
-using Newtonsoft.Json.Linq;
+using Shouldly;
 using Xunit;
 
 namespace GraphQL.Tests.Subscription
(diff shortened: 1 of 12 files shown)


src/GraphQL.Tests/Subscription/SubscriptionTests.cs
```diff
@@ -0,0 +1,112 @@
...
+                Schema = schema
+            });
+
+            chat.AddMessage(addedMessage);
+
+            /* Then */
+            var stream = result.Streams.Values.FirstOrDefault();
+            var message = await stream.FirstOrDefaultAsync();
+            Assert.NotNull(message);
+            Assert.IsType<ExecutionResult>(message);
```
joemcbride: Mind switching these assertions to Shouldly?  That is what the rest of the project is using.

src/GraphQL/Subscription/SubscriptionExecuter.cs
```diff
@@ -0,0 +1,295 @@
...
+                var valueTransformer = result
+                    .Select(value =>
+                    {
+                        var r = ResolveFieldAsync(context, parentType, value, fields).GetAwaiter().GetResult();
+
+                        return new ExecutionResult()
+                        {
+                            Data = r.Value
+                        };
+                    })
```
joemcbride: I believe you can use `SelectMany` here so you can use async/await.

```csharp
.SelectMany(async value => ...)
```

src/GraphQL/Subscription/SubscriptionExecuter.cs
```diff
@@ -0,0 +1,295 @@
...
+            IDocumentBuilder documentBuilder, 
+            IDocumentValidator documentValidator, 
+            IComplexityAnalyzer complexityAnalyzer) : base(documentBuilder, documentValidator, complexityAnalyzer)
+        {
+            _documentBuilder = documentBuilder;
+            _documentValidator = documentValidator;
+            _complexityAnalyzer = complexityAnalyzer;
+        }
+
+        //todo: this is currently private in the DocumentExecuter
```
joemcbride: Can fix this todo now that this code is in the same repo.

src/GraphQL.Tests/Subscription/SubscriptionTests.cs
pekkah: Of course.

pekkah:
@joemcbride should be ready for second pass

PR #432, opened 2017-09-20
Changed IsValidLiteralValue condition from InputObjectGraphType to ICom…

Jasonkingsmill:
Changed IsValidLiteralValue condition from InputObjectGraphType to IComplexGraphType

This should now support any generic IComplexObjectGraphType as input types in mutations


joemcbride:
Unless you can prove otherwise, this won't be merged as this would break the GraphQL spec.

PR #434, opened 2017-09-20
Another attempt at fixing the decimal number parsing/printing

pekkah:
- Use when parsing values
- Use when printing values

All tests passing on fi-FI locale


pekkah:
At the moment all of the float tests fail on my machine without these changes.

joemcbride:
They run fine on my machine and CI without those changes.  So we need tests that run under the specific culture that you run the code in.  Otherwise changes can be made that will break this again and no one will know since CI won't break.

pekkah:
Just the tests against master are here https://github.com/graphql-dotnet/graphql-dotnet/pull/437

joerneu:
@pekkah: I believe you missed FloatGraphType.cs and ObjectExtensions.cs. The rest looks good.

pekkah:
@joerneu thanks :) Fixed. ObjectExtensions uses Conversions and that should be covered.

joerneu:
@pekkah:

Line 132 in ObjectExtensions uses culture dependent conversion:

```
            var text = value.ToString();
```

I changed it to:

```
            string text;
            if (value is float)
                text = ((float)value).ToString(CultureInfo.InvariantCulture);
            else if (value is double)
                text = ((double)value).ToString(CultureInfo.InvariantCulture);
            else if (value is decimal)
                text = ((decimal)value).ToString(CultureInfo.InvariantCulture);
            else
                text = value.ToString();
```

Without this change the next line does not work correctly:

```
            return _conversions.Value.Convert(fieldType, text);
```

pekkah:
@joerneu yes, you're right. Fixing it now. Too many .ToStrings and Parses.

pekkah:
@joemcbride includes now bunch of tests with xunit theories

src/GraphQL.Tests/ObjectExtensionsTest.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace GraphQL.Tests
+{
+    public class ObjectExtensionsTests
+    {
+        [Theory]
+        [ClassData(typeof(CultureList))]
+        public void convert_double(CultureInfo cultureInfo)
+        {
+            CultureInfo.DefaultThreadCurrentCulture = cultureInfo;
+            CultureInfo.DefaultThreadCurrentUICulture = cultureInfo;
```
joemcbride: I think we should use something like this [`UseCulture`](https://github.com/xunit/samples.xunit/blob/master/UseCulture/UseCultureAttribute.cs) attribute since this sets the culture but then does not revert it back to its original value.  That could then create instability between tests.

src/GraphQL/ObjectExtensions.cs
```diff
@@ -135,7 +136,16 @@ public static object ConvertValue(object value, Type fieldType)
                 return value;
             }
 
-            var text = value.ToString();
+            string text;
```
joemcbride: The point of this is to potentially convert the value to a specific type.  Perhaps there should be a more generic check if the `value` is already of the target `fieldType`?  Similar to what the `DateTime` check is doing above.  That would remove the need of the additional parse.  Of course if it does need the conversion then the below would be required.

joerneu: Yes, I agree. Not knowing the internals I went for the safest solution. For example, I did not know if the ObjectExtensions' method is used to convert between float and decimal. Or between float and strings, etc.
Depending on the requirements we could check the fileldType and use .NET's Convert which already handles float, double and decimal conversions.

pekkah:
@joemcbride created a test helper to run test with specific culture and another to run with premade set of cultures

joerneu:
Yes, many ToString() and Parse() calls. Did you see my "patch.diff.txt" at [#302](https://github.com/graphql-dotnet/graphql-dotnet/issues/302)? It should cover all ToString() and Parse() calls.
It also fixes DateTime parsing which is also culture dependent.

src/GraphQL/ObjectExtensions.cs
pekkah: I think the whole parsing, coercing etc. needs an overhaul. Currently "converting" values is happening in way too many places, but I think that kind of refactoring is outside the scope of this PR.

pekkah:
@joerneu saw the DateTime but didn't want to touch it in scope of this PR.

src/GraphQL/ObjectExtensions.cs
joemcbride: > convert between float and decimal

That is possible.  It all depends on how the data is de-serialized and what Type someone added to their property on their object.

> I think the whole parsing, coercing etc. needs an overhaul.

I agree with this - the primary purpose of some of this was to convert the JSON.NET types to actual types.  So the core framework didn't have a dependency on JSON.NET.  Otherwise you would have to be checking for JSON.NET types (JObject, JArray, etc.) throughout the codebase.

pekkah: So do we change this part somehow?

joemcbride:
For whatever reason, CI failed when this was merged.  Kicking that off again then I'll push out an updated alpha with this fix.

Issue #335, opened 2017-05-15
IInputGraphType is unused


CornedBee:
There is an interface GraphQL.Types.IInputGraphType.

This interface isn't implemented by any type (not even InputObjectGraphType), nor used by anything.

Its presence is confusing. The interface should be deleted or, alternatively, implemented by all types that are valid to use in input arguments or other input types (i.e. InputObjectGraphType and the scalar types).

PR #436, opened 2017-09-21
Allow generic InputObjectGraphType for strongly-typed field definition

Jasonkingsmill:
This should allow strongly-typed classes to be used to define input types. It should continue to adhere to GraphQL spec as all added fields must be of ScalarGraphType or InputObjectGraphType.

Usage example:
```
public class MyClass
{
    public int Id { get; set; }
    public string RandomDescription { get; set; }
}

public class MyClassInputType : InputObjectGraphType<MyClass>
{
    Field(x => x.Id);
    Field(x => x.RandomDescription);
}
```

### DIFF
--- a/src/GraphQL/Types/InputObjectGraphType.cs
+++ b/src/GraphQL/Types/InputObjectGraphType.cs
@@ -1,6 +1,109 @@
-﻿namespace GraphQL.Types
+using GraphQL.Builders;
+using GraphQL.Resolvers;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+
+namespace GraphQL.Types
 {
     public class InputObjectGraphType : ComplexGraphType<object>
     {
     }
+
+
+
+    public class InputObjectGraphType<TSourceType> : InputObjectGraphType
+    {
+
+        public FieldType Field(
+            Type type,
+            string name,
+            string description = null,
+            QueryArguments arguments = null,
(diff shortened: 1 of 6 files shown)


src/GraphQL/Types/InputObjectGraphType.cs
```diff
@@ -1,6 +1,109 @@
...
+
+namespace GraphQL.Types
 {
     public class InputObjectGraphType : ComplexGraphType<object>
     {
     }
+
+
+
+    public class InputObjectGraphType<TSourceType> : InputObjectGraphType
```
joemcbride: What we have done elsewhere is reverse this inheritance.

```csharp
public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType> {}

public class InputObjectGraphType : InputObjectGraphType<object> {}
```

Though that will probably cause problems with how we're doing checks on `InputObjectGraphType` and a number of people (including yourself) will want to use the generic `InputObjectGraphType<T>`.

There is an open issue related to this (#335) that I would like to fix for the 2.0 release.  We can use an interface to represent the input object and use that where we are now using the `InputObjectGraphType`.  Similar to what you were attempting to do with the `IComplexGraphType` but it would be input specific.

Jasonkingsmill:
Good suggestion. 
I've made changes as suggested, inverting the inheritance and removing the now-redundant duplicated Field methods inherited from ComplexGraphType.
I've updated all Input type checking to check against the IInputObjectGraphType interface.

Issue #402, opened 2017-08-22
Unable to Install Package in VS 2017 with Core 2.0


AlexNOliveira:
Hi,

   I tried to create a .NET Core 2.0 class library project in VS 2017 (15.3.2) to use GraphQL but Nuget is unable to install the package, even via Package Console. I end up with the message "package restore failed. rolling back package changes..." in the Error output console.

   I found an workaround that suggests to add the lines below to the csproj file. I did, but it didn't work for me.

"<PropertyGroup>
    <PackageTargetFallback>$(PackageTargetFallback);dotnet</PackageTargetFallback>
</PropertyGroup>"

In the page where I found the workaround, it says "This is happening because the maintainer hasn't updated the NuGet package with .NET Standard targets.". I am not sure if is the real problem..

If I change the Target Framework to .NET Core 1.1, I am able to install the package.

Thanks in Advance,
Alexandre

joemcbride:
Thanks for the heads up!  I haven't tried this yet myself.  I probably won't be getting around to looking at this till next week.  Was there a more detailed error available?  I believe you can see that in VS in the Output window, though should be able to see it for sure if you use `dotnet restore` on the command line.

mokeyish:
@joemcbride Here is the screenshot #403

srini85:
I had this issue as well. I decided to upgrade the project to dotnet core 2.0 and created a PR

https://github.com/graphql-dotnet/graphql-dotnet/pull/409

srini85:
you can also get around the issue by forcing your project to use 4.1.0 of System.Reflection.TypeExtensions

E.g.

`
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="graphql" Version="0.17.2" />
    <PackageReference Include="System.Reflection.TypeExtensions" Version="4.1.0" />
  </ItemGroup>
</Project>a

`

mpfisterUniBE:
Thanks @srini85 this helped me getting it back working.
For my netcoreapp2.0 I had to use some newer packages however, 4.1.0 of the TypeExtensions wasn't enough:

`<Project Sdk="Microsoft.NET.Sdk">
	<PropertyGroup>
		<TargetFramework>netcoreapp2.0</TargetFramework>
	</PropertyGroup>
	<ItemGroup>
[...]

PR #441, opened 2017-09-23
Add netstandard2.0 as target while keeping the old ones

pekkah:
Current targets

- NET 4.5.2
- netstandard 1.1
- netstandard 2.0 <- NEW

All tests passing on all targets and NuGet includes new assemblies for 2.0.
Global.json now requires SDK 2.0.0.


joemcbride:
I think this looks good.  May need to add using a newer `System.Reflection.TypeExtensions` for 2.0 according to #402 

```xml
<ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
    <PackageReference Include="System.Reflection.TypeExtensions" Version="4.4.0" />
</ItemGroup>
```

pekkah:
@joemcbride done

Issue #435, opened 2017-09-20
Performance when using large lists.


holm0563:
My company has several data sets that have a lot of properties and have lists of lists returned. I've noticed this library could have a few improvements in this area.

I was going to submit a pull request but it looks like this project is moving towards 2.0. I went to submit a pull request but it looks like a lot of the files like DocumentExecutor already have some major changes. I'd be happy to re-evaluate with 2.0 but want to make sure these fit in with any 2.0 objectives. So here are a summary of changes. 

The changes address the issue of running tests with a list with 10k items in a list and measuring performance. These changes result in a 120% speed increase in this setup. 

https://github.com/holm0563/graphql-dotnet/tree/performance

1. The document executor assumes all fields should be run in a new thread. Most of the time this is not true. The data is returned and usually only the NameFieldResolver to resolve the current data. Resolving these async instead results in big performance gains.

Simple example:
https://github.com/holm0563/graphql-dotnet/commit/e82565c74f4cd49ac8d005ae069e2f5b3bb4120d#diff-9ee075dfc3429298a3c5de0c1b53bef2

We definitely still want the ability to resolve fields in a thread whenever it would connect to a database or external source. The current pipeline of always using a thread and then recursively calling CompleteValueAsync -> ExecuteFieldsAsync can be optimized. I propose adding an opt in policy on the IFeildResolver to signify if it should run threaded. Also I removed the FieldMiddlewareBuilder default behavior of wrapping every field with a resolver. Allowing the resolver to be null and using the NameFieldResolver as the default helped performance. Most of the changes seen in the performance branch of my fork revolve around this idea.

2. Typically in production environments this library would often be re-running the same query over and over. Adding caching to the documentvalidator improves performance by 500% when just running simple queries repeatedly against the star wars schema. Should this be a built in feature? 

Example:
https://github.com/holm0563/graphql-dotnet/blob/performance/src/GraphQL/Validation/CachedDocumentValidator.cs

3. Reflection is slowe(er). Especially when running against lists, caching property info can improve performance. This gave about a 2% speed jump.
https://github.com/holm0563/graphql-dotnet/blob/performance/src/GraphQL/ObjectExtensions.cs

4. When running a performance analysis the TrimGraphQLTypes was on the list and shaved another 2% of the tests to cache this tests.

https://github.com/holm0563/graphql-dotnet/blob/performance/src/GraphQL/GraphQLExtensions.cs

[...]

PR #452, opened 2017-10-06
Performance2.0

holm0563:
See: https://github.com/graphql-dotnet/graphql-dotnet/issues/435

Thanks for the suggestions @joemcbride hopefully this addresses all of your thoughts.

This pull requests solves several issues:
Query's can by ran in parallel using FieldAsync
Mutations are guaranteed to run synchronously
Speed gains.
Easier to optimize for performance for production servers vs development servers.

Performance Gains:
image

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -289,7 +290,7 @@ namespace GraphQL
             return operation;
         }
 
-        public Task<Dictionary<string, object>> ExecuteOperationAsync(ExecutionContext context)
+        public Task<IDictionary<string, object>> ExecuteOperationAsync(ExecutionContext context)
         {
             var rootType = GetOperationRootType(context.Document, context.Schema, context.Operation);
             var fields = CollectFields(
@@ -302,11 +303,32 @@ namespace GraphQL
             return ExecuteFieldsAsync(context, rootType, context.RootValue, fields, new string[0]);
         }
 
-        public Task<Dictionary<string, object>> ExecuteFieldsAsync(ExecutionContext context, IObjectGraphType rootType, object source, Dictionary<string, Field> fields, IEnumerable<string> path)
+        public async Task<IDictionary<string, object>> ExecuteFieldsAsync(ExecutionContext context, IObjectGraphType rootType, object source, Dictionary<string, Field> fields, IEnumerable<string> path)
         {
-            return fields.ToDictionaryAsync<KeyValuePair<string, Field>, string, ResolveFieldResult<object>, object>(
-                pair => pair.Key,
-                pair => ResolveFieldAsync(context, rootType, source, pair.Value, path.Concat(new[]{pair.Key})));
+            var data = new ConcurrentDictionary<string, object>();
+
+            //if (context.Operation.OperationType == OperationType.Mutation)
+            {
+                foreach (var field in fields)
+                {
+                    var serialResults = await ResolveFieldAsync(context, rootType, source, field.Value,
+                        path.Concat(new[] {field.Key}));
+
+                    if (serialResults.Skip)
+                        continue;
+
+                    data.TryAdd(field.Key, serialResults.Value);
+                }
+
+                return data;
+            }
+            //else
+            //{
+                
+            //}
+            //return fields.ToDictionaryAsync<KeyValuePair<string, Field>, string, ResolveFieldResult<object>, object>(
+            //    pair => pair.Key,
+            //    pair => ResolveFieldAsync(context, rootType, source, pair.Value, path.Concat(new[]{pair.Key})));
         }
 
         public async Task<ResolveFieldResult<object>> ResolveFieldAsync(ExecutionContext context, IObjectGraphType parentType, object source, Field field, IEnumerable<string> path)
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -322,16 +322,18 @@ namespace GraphQL
 
             foreach (var fieldCollection in fields)
             {
+                var currentPath = path.Concat(new[] { fieldCollection.Key });
+
                 var field = fieldCollection.Value;
                 var fieldType = GetFieldDefinition(context.Document, context.Schema, rootType, field);
 
                 if (fieldType?.Resolver == null || !fieldType.Resolver.RunThreaded() || context.Operation.OperationType == OperationType.Mutation)
                 {
-                    await ExtractFieldAsync(context, rootType, source, field, fieldType, data, path);
+                    await ExtractFieldAsync(context, rootType, source, field, fieldType, data, currentPath);
                 }
                 else
                 {
-                    var task = Task.Run(() => ExtractFieldAsync(context, rootType, source, field, fieldType, data, path));
+                    var task = Task.Run(() => ExtractFieldAsync(context, rootType, source, field, fieldType, data, currentPath));
 
                     externalTasks.Add(task);
                 }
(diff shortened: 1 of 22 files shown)


pekkah:
For benchmarks could use this https://github.com/dotnet/BenchmarkDotNet

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -302,11 +314,152 @@ protected virtual Operation GetOperation(string operationName, Document document
...
+                }
+                else
+                {
+                    var task = Task.Run(() => ExtractFieldAsync(context, rootType, source, field, fieldType, data, currentPath));
+
+                    externalTasks.Add(task);
+                }
+            }
+
+            if (externalTasks.Any())
```
Jetski5822: Faster to use a .Count > 0 I believe?

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -29,5 +29,11 @@ public class ExecutionOptions
         public IFieldNameConverter FieldNameConverter { get; set; } = new CamelCaseFieldNameConverter();
 
         public bool ExposeExceptions { get; set; } = false;
+
+        //Note disabling will increase performance
+        public bool EnableLogging { get; set; } = true;
```
joemcbride: I think this should be renamed to `EnableMetrics`.  Since logging is not the same things as the Metrics.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -147,7 +149,10 @@ private void ValidateOptions(ExecutionOptions options)
                 {
                     using (metrics.Subject("schema", "Initializing schema"))
                     {
-                        config.FieldMiddleware.ApplyTo(config.Schema);
+                        if (config.EnableLogging)
```
joemcbride: I think this needs a different flag.  Disabling field middleware is different than disabling the metrics gathering.  You can use field middleware for more than just metrics.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -181,12 +186,19 @@ private void ValidateOptions(ExecutionOptions options)
                 IValidationResult validationResult;
                 using (metrics.Subject("document", "Validating document"))
                 {
-                    validationResult = _documentValidator.Validate(
-                        config.Query,
-                        config.Schema,
-                        document,
-                        config.ValidationRules,
-                        config.UserContext);
+                    if (config.EnableDocumentValidation)
```
joemcbride: I'm really not liking this flag.  I understand why you want to "disable" it, though I'm thinking of too many instances where this would cause problems.  I don't think this should be a feature flag but instead have this behavior in a class that does that for you (which I believe you had in your original post).  We can pass the validator more data if it needs it to properly "disable" validation.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -302,11 +314,152 @@ protected virtual Operation GetOperation(string operationName, Document document
...
+                else
+                {
+                    result = NameFieldResolver.Resolve(source, field.Name);
+                }
+
+                if (result != null)
+                {
+                    var scalarType = fieldType.ResolvedType as ScalarGraphType;
+
+                    result = scalarType?.Serialize(result);
```
joemcbride: I think the `ResolvedType` should always be a `ScalarGraphType` here (due to a previous check), though I think technically the `?` here would potentially change the `result` from non-null to null.  Is that the wanted behavior?  Should this be a hard-cast to `ScalarGraphType` since it should always be one or an error thrown after the `as` check?

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -29,5 +29,11 @@ public class ExecutionOptions
...
+
+        //Note disabling will increase performance
+        public bool EnableDocumentValidation { get; set; } = true;
```
joemcbride: I think this should be removed (see comment on where it is used).

joemcbride:
Thanks for your work on this!  Happy to get some speed gains and a better understanding of async/await + Task.Run.

src/GraphQL/Execution/DocumentExecuter.cs
holm0563: Thanks, this would be faster since its a list and not an enumerable

src/GraphQL/Execution/DocumentExecuter.cs
holm0563: @joemcbride 

If I'm understanding right you would prefer something like:
var exec = new DocumentExecutor(new GraphQLDocumentBuilder(), new SkipDocumentValidator(), new ComplexityAnalyzer());

I liked the flag idea since a SkipDocumentValidator is not part of this project.

Note this change alone was responsible for most of the 600% improvements with the starwars query. Most of the time executing small queries is in the validation and would be a huge potential savings in a production environment.

I am very curious on your thoughts. Everything I threw at it seemed to handle appropriately and just give back a more cryptic error since it wouldn't be able to parse the query. It does make me nervous as well, what kind of problems would we need to prevent without this validation?

src/GraphQL/Execution/DocumentExecuter.cs
holm0563: This is mostly due to me using Resharper which somewhat realizes a null can be returned (although it shouldn't because of the previous check) and playing it safe with a null check vs throwing an exception.

src/GraphQL/Execution/ExecutionOptions.cs
holm0563: Yeah i like this idea much better, this change was pushed.

src/GraphQL/Execution/DocumentExecuter.cs
holm0563: Yeah i like this idea much better, this change was pushed.

holm0563:
Note: My company (Navitaire) just finished load tests with these changes in place. Graph is now performing better then Rest in all categories including server CPU and response time. We typically have several web servers in a farm that handle quite a bit of traffic.

src/GraphQL/Execution/DocumentExecuter.cs
joemcbride: >If I'm understanding right you would prefer something like:
var exec = new DocumentExecutor(new GraphQLDocumentBuilder(), new SkipDocumentValidator(), new ComplexityAnalyzer());

Correct.  This turns off the [Validation feature](http://facebook.github.io/graphql/October2016/#sec-Validation) which is part of the spec.  So everything that Validation does doesn't happen.  Basically, I really don't want to support turning off that feature this easily since that is a pretty big portion of the spec being turned off.  I can understand why you want to do it, though completely turning it off is specific to your business needs and I wouldn't suggest anyone to do it unless they really understand what it means.  Meaning, it would now be easy to write invalid queries and get hard-to-understand errors or null data.

Writing a class like `SkipDocumentValidation` is very easy to do, accomplishes the same thing, and doesn't put the support burden on the core project.

holm0563: Thanks @joemcbride . I'm understanding your point now. I still like it being easier since in production everyone should care about response times but like you mentioned if developers understand the issue they can still implement a custom validator. Maybe as a future update the default validator should be evaluated from a performance standpoint and/or implement a caching strategy.

Removing this features takes the StarWarsPerformanceTest speed improvement from 642% speed gains to 15%.

For our business needs we will plan on removing it completely, since our severs handle a very high amount of traffic.

I've updated my branch. I think I have all of the changes ready that you suggested. Let me know if I'm missing anything.

joemcbride: > Maybe as a future update the default validator should be evaluated from a performance standpoint and/or implement a caching strategy.

I think that could be beneficial.  What I've heard from Facebook is they do cache quite a bit.  They have a set of "trusted" queries which I believe would skip validation (since they know it is valid) and even `Document` parsing, since they cache the parsing of the `Document`.

I believe they even go as far as only having the client send a "query name", not even a GraphQL query, and associated variables to cut down on the http traffic.  I added the basic idea of this [here](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.GraphiQL/Controllers/GraphQLController.cs#L31-L34) though it doesn't add the `Document` parsing/validation benefit.

holm0563: Cool, that is very good to know. I was playing around with the idea of named queries. Good to know that is the direction Facebook went as well.

PR #482, opened 2017-11-24
Allow Unsigned Integers

### LINKED ISSUE #473: Add UIntGraphType
Add an `GraphQL.Types.UintGraphType` to represent `uint`.

deinok:
Allow Unisgned Integers as type in parsing it as a normal integer.
This avoids casting uint to ints in the backend

Closes #473

### DIFF
--- a/src/GraphQL/TypeExtensions.cs
+++ b/src/GraphQL/TypeExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


joemcbride:
What this will actually do is convert the `uint` to either an `int` or a `long`.  That doesn't seem like it would be expected behavior?

https://github.com/graphql-dotnet/graphql-dotnet/blob/e173a61b9e8d77f55f4a007f5f7081bae622cb1f/src/GraphQL/Types/IntGraphType.cs#L27-L38

[2.0.0-alpha-811](https://github.com/graphql-dotnet/graphql-dotnet/releases/tag/v2.0.0-alpha-811) added the ability to provide `IValue` converters + custom scalars which is what I would suggest to do - a `uint` specific scalar type + `IValue`.

deinok:
@joemcbride Interesting, but does't common primitives come by default with the library? Like DateTime

PR #483, opened 2017-11-25
Enable ExecutionOptions.Indent

deinok:


joemcbride:
These extension methods are to help get you started.  Since that indent is not related to the execution I don’t want it on that class.  I would be fine with an optional parameter in the extension method.

deinok:
@joemcbride Is there a more performant way?

joemcbride:
@deinok Its less about performance and more about design.  The `ExecutionOptions` go with the `DocumentExecuter` class.  I don't want to add a property related to the `DocumentWriter` to the class that is paired with the `DocumentExecuter`.

Something more like this would be acceptable:

```csharp
public static string Execute(
    this ISchema schema,
    Action<ExecutionOptions> configure,
    bool indent = true)
{
    ...
}

public static string Execute(
    this ISchema schema,
    Action<ExecutionOptions> configure,
    Action<WritingOptions> configureWrite = null)
{
    ...
}

public class WritingOptions
{
    public Formatting Formatting { get; set; } = Formatting.None;
    public JsonSerializerSettings SerializationSettings { get; set; } =
        new JsonSerializerSettings
        {
            DateFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.FFFFFFF'Z'",
        };
}
```

deinok:
Are PRs for this accepted?

joemcbride:
Yep

PR #504, opened 2017-12-15
Add converter for DateTimeOffset type

roman-kozachenko:
That allows to pass `DateTimeOffset` type to methods marked with `[GraphQLMetadata]` attribute.

### DIFF
--- a/src/GraphQL.Tests/Utilities/DateTimeOffsetTypeTests.cs
+++ b/src/GraphQL.Tests/Utilities/DateTimeOffsetTypeTests.cs
@@ -1,0 +1,56 @@
+// <copyright file="DateTimeOffsetTypeTests.cs" company="eVote">
+//   Copyright © eVote
+// </copyright>
+
+using System;
+using GraphQL.Types;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Utilities
+{
+    public class DateTimeOffsetTypeTests : SchemaBuilderTestBase
+    {
+        [Fact]
+        public void can_use_DateTimeOffset_type()
+        {
+            var schema = Schema.For(@"
+                input DateTimeOffsetInput{
+                    value: Date
+                }
+                type Query {
+                  five(model: DateTimeOffsetInput): Date
+                }
+            ", _ =>
+            {
+                _.Types.Include<ParametersType>();
+            });
+
+            var utcNow = DateTimeOffset.UtcNow;
+
+            var result = schema.Execute(_ =>
+            {
+                _.Query = $"{{ five(model:{{ value:\"{utcNow}\"}}) }}";
+            });
+
+            var expectedResult = CreateQueryResult($"{{ 'five': \"{utcNow.AddDays(1):yyyy-MM-ddTHH:mm:ssZ}\" }}");
+            var serializedExpectedResult = Writer.Write(expectedResult);
+
+            result.ShouldBe(serializedExpectedResult);
+        }
+
+        [GraphQLMetadata("Query")]
+        class ParametersType
+        {
+            public DateTimeOffset Five(DateTimeOffsetInput model)
+            {
+                return model.Value.AddDays(1);
+            }
+        }
+    }
+
+    class DateTimeOffsetInput
+    {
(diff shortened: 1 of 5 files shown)


src/GraphQL/Conversion/DateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,81 @@
+// <copyright file="DateTimeOffsetConverter.cs" company="wut">
+//   Copyright © wut
+// </copyright>
```
joemcbride: Need to remove this copyright.

roman-kozachenko: Sure, accidentally copy-pasted it from main solution.

roman-kozachenko:
A simple test added.

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
```diff
@@ -300,6 +301,7 @@ public void can_use_context_source_usercontext()
         public void can_use_context_source_usercontext_with_params()
         {
             var schema = Schema.For(@"
+                input
```
joemcbride: I believe this is probably causing tests to fail.

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
```diff
@@ -320,6 +322,33 @@ public void can_use_context_source_usercontext_with_params()
...
+            var result = schema.Execute(_ =>
+            {
+                _.Query = $"{{ five(model:{{ value:\"{utcNow}\"}}) }}";
+            });
+
+            var expectedResult = CreateQueryResult($"{{ 'five': \"{utcNow.AddDays(1).ToString("yyyy-MM-ddTHH:mm:ssZ")}\" }}");
+            var serializedExpectedResult = Writer.Write(expectedResult);
+
+            result.ShouldBe(serializedExpectedResult);
+        }
```
joemcbride: Can you add this to its own test file?  This test is not super related to the other ones in this class/file.

src/GraphQL/ObjectExtensions.cs
```diff
@@ -131,11 +131,25 @@ public static object ConvertValue(object value, Type fieldType)
         {
             if (value == null) return null;
 
-            if (fieldType == typeof(DateTime) && value is DateTime)
+            // exact type match
+            if ((fieldType == typeof(DateTime) && value is DateTime) ||
+                (fieldType == typeof(DateTimeOffset) && value is DateTimeOffset))
```
joemcbride: Perhaps this can just be simplified to something like:

```
if (fieldType.IsInstanceOfType(value)) return value;
```

So if the given value is of the given type it just returns it.

roman-kozachenko: Yes, it should work for other types too.

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
roman-kozachenko: Sure, extracted to separate file.

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
roman-kozachenko: Missed it, thanks.

src/GraphQL.Tests/Utilities/DateTimeOffsetTypeTests.cs
```diff
@@ -0,0 +1,56 @@
+// <copyright file="DateTimeOffsetTypeTests.cs" company="eVote">
+//   Copyright © eVote
+// </copyright>
```
joemcbride: This creeped in again.

roman-kozachenko: This is a Resharper auto-generated file template. I really don/t want to have rights on that file :).
Removed.

PR #513, opened 2017-12-29
Adding TryResolveAsync helper method

glennblock:
This method is a helper for handling bubbling errors from the resolve lambda. It will invoke an async lambda provided by the user wrapping with a try/catch. If an Exception occurs it will grab the message and add it to the Errors collection of the context. The Try method also allows providing a custom error handler for more complex logic.


glennblock:
@joemcbride thoughts?

src/GraphQL/Types/ResolveFieldContext.cs
```diff
@@ -106,6 +108,27 @@ public bool HasArgument(string argumentName)
         {
             return Arguments?.ContainsKey(argumentName) ?? false;
         }
+
+        public async Task<object> TryAsyncResolve(Func<ResolveFieldContext<TSource>, Task<object>> resolve, Func<ExecutionErrors, Task<object>> error = null)
+        {
+            try
+            {
+                var result = await resolve(this);
+                return Task.FromResult(result);
```
joemcbride: I think can just return `result` here.

glennblock: Thanks use to not wrap it but it was not working right. I wlll test again.

glennblock:
@joemcbride verified that worked. Updated the code/tests. Thanks!

src/GraphQL/Types/ResolveFieldContext.cs
```diff
@@ -106,6 +108,26 @@ public bool HasArgument(string argumentName)
...
+        {
+            try
+            {
+                return await resolve(this);
+            }
+            catch (Exception ex)
+            {
+                if (error == null)
+                {
+                    Errors.Add(new ExecutionError(ex.Message));
```
pekkah: Do we lose context if only exception message is added? Should it contain the whole ex.ToString()?

pekkah: Especially in case of AggregateException etc..

global.json
```diff
@@ -1,5 +0,0 @@
-{
-  "sdk": {
-    "version": "2.0.0"
-  }
-}
```
joemcbride: Would like to keep this file as it signals to AppVeyor what SDK to use.

src/GraphQL/Types/ResolveFieldContext.cs
joemcbride: Could perhaps add the full exception as the inner exception to keep the simpler top level message.

global.json
glennblock: OK, will add it back

src/GraphQL/Types/ResolveFieldContext.cs
glennblock: @joemcbride the point of this method is to allow nicer messages. In my case for example I am throwing a contextual message to the user. I didn't include the stack trace / inner exception for that reason. I guess one option is to have a switch that lets you decide if you want the inner exception.

joemcbride: The inner exception will never been seen in the response unless you have `ExposeExceptions` set on the `ExecutionOptions`.  I'm fine with either way.

glennblock:
@pekkah @joemcbride adding setting the inner exception now, and added back the global.json.

glennblock:
👍

PR #518, opened 2018-01-04
Correct examples by adding metadata to query class

omelhoro:


joemcbride:
That should only be needed if you want to change the name of the GraphType or set some other value.  By default it should just use the name of the class if the `GraphQLMetadata` attribute is not used.

omelhoro:
Ok I'll try it again. I remember that after copy & pasting the examples it didn't work without the metadata attribute.

joemcbride:
If that is the case then that's a bug that has creeped in, though it should be optional...

omelhoro:
@joemcbride You're right. Thank! It works like you said. I've used not 'Query' as class name.

joemcbride:
```
type Post {
    id: ID!
    title: String!
}

type MyQuery {
    post(id: ID!): Post
}

schema {
  query: MyQuery
}
```

omelhoro:
Actually I noticed one other problem I had was the dependency injection, in something like that:
```
  public class Query
  {

    public IPlayerRepository PlayerRepository { get; private set; }

    public Query(IPlayerRepository playerRepository)
    {
      PlayerRepository = playerRepository;
    }
    
    [GraphQLMetadata("player")]
    public Players[] GetPlayers(int[] id)
    {
      return PlayerRepository.GetByIds(id);
    }

}
```
It seems not to work that way. The executer says there's no path 'player'.

joemcbride:
Hm, I'm doing something similar and it works.

```csharp
services.AddSingleton(sp =>
{
    var schema = Schema.For(
        SchemaTypeSystem,
        _ =>
        {
            _.DependencyResolver = new FuncDependencyResolver(sp.GetRequiredService);
            _.Types.Include<QueryType>();
            _.Types.Include<MutationType>();
        });

    return schema;
});
```

Can you use `new SchemaPrinter(schema).Print()` and see what it shows you?

omelhoro:
Thanks again!! I included the code you send, and what also was important was to include that line
`services.AddSingleton<Query>();`. Actually, I'm new to .NET and the DI  so I have no idea why this solves it. But maybe it would possible to somehow throw an error if no singleton for schema or query is configured?

joemcbride:
If you use `GetRequiredService` vs. `GetService` I believe it should throw an error when that is the case:

```csharp
_.DependencyResolver = new FuncDependencyResolver(sp.GetRequiredService);
```

That being said, if DI isn't setup like that and it can't construct the object (which in your case seems like it wouldn't be able to), I agree, it should thrown an error.

PR #525, opened 2018-01-15
Removed blocking of main thread. Passed CancellationToken into spawned threads

holm0563:
These fixes prevent thread starvation when processing a high number of graph queries that are running asynchronously. Will improve performance significantly under heavy load.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -319,13 +319,13 @@ namespace GraphQL
                 var field = fieldCollection.Value;
                 var fieldType = GetFieldDefinition(context.Document, context.Schema, rootType, field);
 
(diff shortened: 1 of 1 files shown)


holm0563:
This is failing on test: ThreadPerformanceTests.Executes_IsQuickerThanTotalTaskTime which was previously skipped. My guess is its machine specific and needs to processors to run correctly. Any suggestions? This was not changed as part of this pull request.

joemcbride:
I have seen it fail intermittently on CI.  May just need to try to run it again, or bump of the time while on CI.

joemcbride:
I kicked off a new build.

joemcbride:
Ran this a couple more times, both being > 900 and < 1000.  So may be sufficient to bump it to 1k for now.  The ~1300 was probably an outlier.

JonasZ95:
`await Task.WhenAll(externalTasks.ToArray());`

You can remove the ToArray() here aswell unless I'm wrong.

Issue #381, opened 2017-08-01
Support Apollo Tracing


benpriebe:
I am wondering if anyone has had success in instrumenting a graphql-dotnet implementation to send analytics to Apollo Data Optics service (http://www.apollodata.com/optics/)?

Apollo Optics currently have examples on how to instrument your NodeJS or Ruby GraphQL servers but not .Net. I have reached out to their support team and they suggested I reach out to this community.

joemcbride:
No, there isn't anything built in for that at present.  I looked into it after it was announced at the last GraphQL Summit, though at the time they didn't have free accounts so that made it a lot harder to integrate.  They also want the metrics in a very specific format.  I would say the current metrics are about 70% of what Apollo Optics want.

See the docs on metrics: http://graphql-dotnet.github.io/graphql-dotnet/learn#metrics

benpriebe:
Thanks for the link.

pekkah:
This will replace optics https://dev-blog.apollodata.com/introducing-apollo-engine-insights-error-reporting-and-caching-for-graphql-6a55147f63fc

joemcbride:
It looks like they've switched to wanting a new, simpler format.  This should be much easier to accomplish than what was previously required.

https://github.com/apollographql/apollo-tracing

jbaxleyiii:
@joemcbride awesome! I'll start playing around with it over the next few weeks and reach back out! Thanks for this great project 🤗

RehanSaeed:
Is there any example on how to set this up with GraphQL.NET Server? The [docs](https://graphql-dotnet.github.io/docs/getting-started/metrics) don't cover that. I've only just found out about Apollo Tracing, is there an easy way to ship these logs to Apollo to get the free metrics and charts?

csboling:
I haven't used the [server](https://github.com/graphql-dotnet/server) project but it looks to me like you should be able to configure a customized service for `IGraphQLExecuter<TSchema>`, something like:

```csharp
public class MyGraphQLExecuter<TSchema> : DefaultGraphQLExecuter<TSchema>
    where TSchema : ISchema
{
    public override async Task<ExecutionResult> ExecuteAsync(string operationName, string query, Inputs variables, object context, CancellationToken cancellationToken = default(CancellationToken))
    {
        var result = await base.ExecuteAsync(operationName, query, variables, context, cancellationToken);
        result.EnrichWithApolloTracing();
        return result;   
    }
}
```
[...]

PR #531, opened 2018-01-20
Add ability to support things like apollo tracing

mvestergaard:
In order to support the tracing proposed by https://github.com/apollographql/apollo-tracing there a few changes needed as far as i see it.

Firstly, the current time measurements in metrics is not precise enough. It is currently measured in milliseconds, but the spec expects nanoseconds.

Secondly, for things like apollo-tracing, apollo-cache-control, and other future extension points to work. There is a requirement to be able to expose an extension point in the result.

I've added this as a simple optional dictionary, that these types of extensions can choose to add to. The extensions will themselves be responsible to specify a JsonConverter in order to output expected JSON.

### DIFF
--- a/src/GraphQL/Execution/ExecutionResult.cs
+++ b/src/GraphQL/Execution/ExecutionResult.cs
@@ -1,11 +1,10 @@
+using System.Collections.Generic;
 using GraphQL.Instrumentation;
 using GraphQL.Language.AST;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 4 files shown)


mvestergaard:
Test failure looks like a false positive to me

joemcbride:
Thanks for these updates!  I re-ran CI and it passed.  That test needs to go away...

joemcbride:
#381

Issue #424, opened 2017-09-11
When an object's non-null property resolver returns null (or exception), the 'null' isn't bubbled up to the parent


diaconesq:
The spec says a **non-null** property that is resolved as null (or throws an error) should propagate up and make the whole object become null:
http://facebook.github.io/graphql/#sec-Errors

In the 'StarWars' example, if I change the `Human.Name` property to non-null in the schema, and query for 
```
{
  human(id: "IHaveNoName") {
    name
  }
}
```
and the corresponding object exists in the "database" but has the `Name` property set to `null`, the response should be:  
```
{
  "data": {
    "human": null
  },
  "errors": [
    {
      "message": "Error trying to resolve name." [...]
    }
  ]
}
```

but instead it is:

```
{
  "data": {
    "human": { // the null value should propagate to here
      "name": null //this is invalid, should never be null
    }
  },
  "errors": [
    {
      "message": "Error trying to resolve name." [...]
    }
  ]
}
```

KoditkarVedant:
@joemcbride I was going through `graphql` spec for `List and NonNull` and noticed that as of now we do not handle it in out library.
```
myField: [String!]
```
This means that the list itself can be null, but it can't have any null members. For example, in JSON:
```
myField: null // valid
myField: [] // valid
myField: ['a', 'b'] // valid
myField: ['a', null, 'b'] // error
```
Now, let's say we defined a Non-Null List of Strings:
```
myField: [String]!
```
This means that the list itself cannot be null, but it can contain null values:
```
myField: null // error
myField: [] // valid
myField: ['a', 'b'] // valid
myField: ['a', null, 'b'] // valid
```
[ref](https://graphql.github.io/learn/schema/#lists-and-non-null)

@joemcbride should we track this as separate or should we consider as part of this enhancement itself. 
I am curious to know your thoughts.

joemcbride:
@KoditkarVedant I would be pretty surprised if that is the case.  Lets track that separately and see if you can add some tests to show that its broken.

KoditkarVedant:
@joemcbride Is it okay to keep those test cases failing or should I skip them from running?

joemcbride:
@KoditkarVedant I think its fine to have them fail, then when the build passes we know we've fixed it.

KoditkarVedant:
> 
> 
> @KoditkarVedant I think its fine to have them fail, then when the build passes we know we've fixed it.

Done.

joemcbride:
[...]

PR #534, opened 2018-01-24
Add null bubbling test in case of error

roman-kozachenko:


src/GraphQL.Tests/Errors/NullBubblingTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            public TestSubObjectWithNonNullableProperty()
+            {
+                Name = "Sub";
+                Field<StringGraphType>()
+                    .Name("one");
+
+                Field<NonNullGraphType<StringGraphType>>()
+                    .Name("two")
+                    .Resolve(_ => throw new Exception("wat"));
```
jquense: is this behavior for any error or the error in which it resolves to null but the value is non-nullable?

joemcbride:
Tracking issue: #424

src/GraphQL.Tests/Errors/NullBubblingTests.cs
roman-kozachenko: If 'two' fields is nullable, then it is ok to set it to null when error occurred.
But if it is non-nullable - nearest nullable parent should be set to null in result.

roman-kozachenko:
I add simple bubbling.

joemcbride:
@roman-kozachenko Appears to be working.  Once those tests are updated I can get this merged.

roman-kozachenko:
@joemcbride I fixed tests.

src/GraphQL.Tests/Introspection/IntrospectionResult.cs
```diff
@@ -1,7 +1,7 @@
...
+        public static readonly string Data = "{\r\n  \"__schema\": null\r\n}";
```
joemcbride: Hm, this change is a little troubling.  Didn't notice this one on the failed tests.  Will look at this closer.

roman-kozachenko: You can find failed tests [here](https://ci.appveyor.com/project/graphql-dotnet-ci/graphql-dotnet/build/2.0.0.863-dxxhbeqj/tests). Failed test name is `GraphQL.Tests.Introspection.SchemaIntrospectionTests.validate_core_schema`.
As I see, `queryType` property is NonNullable, but its value is `null`. So, its value should be bubbled to nearest nullable parent - `__shema`.

roman-kozachenko:
@joemcbride, @jquense any remarks for my PR? We want to have that changes on our production app, but until that PR will be merged we have to use custom nuget package based on my branch, losing ability to rapidly get your library updates.

roman-kozachenko:
Since current pull request does not any feedback, I closed it.
All code is available in my branch https://github.com/roman-kozachenko/graphql-dotnet/tree/fix/null-bubbling. It can be merged in main repository, or any changes can be requested in future.

Issue #508, opened 2017-12-21
BeforeExecutionAwaitedAsync Executed After Awaiting


speeddm:
I am attempting to solve the N+1 problem in our project. I had tried to use Dataloader dot net, but found [it was not working out of the box](https://github.com/dlukez/graphql-dotnet-dataloader/issues/4).

My issue is that if I adapt the ExecuteBatchListener example that it doesn't hit until the tasks I have set have _already awaited_.

An example of the schema with a simple type (snippet only):

    public class TestSchema : Schema
    {
        public TestSchema()
        {
            Query = new TestQuery();
        }
    }

    public class TestQuery : ObjectGraphType
    {
        public TestQuery()
        {
            Field<HumanType>("getHuman", resolve: context =>
            {
                return Task.Run(() =>
                {
                    Thread.Sleep(TimeSpan.FromSeconds(10));
                    return new Human {GivenName = "John", FamilyName = "Citizen"};
                });
            });
        }
    }

And there is a listener:

    public class ExecuteBatchListener : DocumentExecutionListenerBase<GraphQlUserContext>
    {
        public override async Task BeforeExecutionAwaitedAsync(
            GraphQlUserContext userContext,
            CancellationToken token)
        {
            // this code does not fire until the 10 second thread sleep is up
        }
    }

Which is added here:

    var userContext = new GraphQlUserContext()
    var result = await executer.ExecuteAsync(_ =>
                {
                    _.Schema = new TestSchema();
                    _.Query = "query { getHuman { givenName familyName } } }";
                    _.UserContext = userContext;
                    _.Listeners.Add(new ExecuteBatchListener());
                }).ConfigureAwait(false);

As I have said in the comment - you do not hit the `BeforeExecutionAwaitedAsync` call until after the task in "getHuman" has run to completion - or if you are trying to batch (and trigger off the execution of the tasks in the `BeforeExecutionAwaitedAsync` method) then it never gets hit at all.

Is there a piece of the puzzle that I'm missing?

joemcbride:
This could be another victim of the most recent threading refactor.  I don’t think we have good tests around this feature so there could be a regression.

Could perhaps be this same issue?  #506

kamsar:
Funny I came here to report the same thing but with a different reproduction. In my case, I'm using `TaskCompletionSource` to block the field from completing until the `BeforeExecutionAwaitedAsync()` method is fired. This is the method that Marten, which the batching example uses, is using.

[...]

PR #536, opened 2018-01-24
Make sure Tasks from resolvers are properly awaited

johnrutherford:
This addresses issue #508. All Tasks returned by resolvers are properly awaited. Any nested Tasks are unwrapped. 

I added a simple unit test to verify.


joemcbride:
Initial review this looks good, though since there's a number of changes I'm going to pull it down and play with it a bit.

johnrutherford:
I'm not sure why the `Executes_IsQuickerThanTotalTaskTime` test is failing. The tasks are still being run in parallel and it passes on my local machine.

PR #540, opened 2018-02-01
Refactor and optimize execution strategies

johnrutherford:
This is related to #264 and #539.

This ended up being a very big overhaul, which you said was needed. 

Previously, it was not possible to split up and parallelize the execution very much because field resolution was tied to building the output. Once you started executing a field, it was then necessary to resolve all descendant fields. 

The idea now is that the execution is broken down into nodes. Each node is resolved independently and the result is stored. Once the entire tree has been executed, then we build the result data. This allows the execution to be broken up in different ways. 

So I know this is a lot of changes. Let me know if you have any questions!


joemcbride:
Indeed big changes.  Will take me a bit to wrap my head around it I think but I'm happy this is getting done!

johnrutherford:
Fixed conflicts with the master branch.

Issue #547, opened 2018-02-06
ResolveFieldContext.SubFields is null for ListGraphType items


TwitchBronBron:
The ResolveFieldContext.SubFields property is correctly populated for ObjectGraphType and subclasses, but is null for ListGraphType items. 

Here is a test I created showing the issue.

```C#
using System.Collections.Generic;
using GraphQL.Types;
using Shouldly;
using Xunit;
using GraphQL.StarWars.Types;
using GraphQL.StarWars;

namespace GraphQL.Tests.StarWars
{
    public class SubTypeIsNotNullForListGraphTypeTests : StarWarsTestBase
    {
        public SubTypeIsNotNullForListGraphTypeTests() : base()
        {
            this.RootQuery = (StarWarsQuery)this.Schema.Query;
        }

        public StarWarsQuery RootQuery;

        [Fact]
        public void subfields_is_not_null_for_ListGraphType()
        {
            RootQuery.FieldAsync<ListGraphType<HumanType>>("listOfHumans", resolve: async (ctx) =>
            {
                //this test fails
                ctx.SubFields.ShouldNotBeNull();
                ctx.SubFields.Keys.ShouldContain("id");
                ctx.SubFields.Keys.ShouldContain("friends");
                return new List<Human>();
            });
            var query = @"
                {
                    listOfHumans {
                        id
                        friends {
                            name
                        }
                    }
                }
            ";

            var expected = @"
                {
                    listOfHumans: []
                }
            ";
            AssertQuerySuccess(query, expected);
        }

        [Fact]
        public void subfields_is_not_null_for_custom_graph_type()
        {
            RootQuery.Field<HumanType>("singleHuman", resolve: (ctx) =>
            {
                ctx.SubFields.ShouldNotBeNull();
                ctx.SubFields.Keys.ShouldContain("id");
                ctx.SubFields.Keys.ShouldContain("friends");
                return null;
            });

            var query = @"
                {
                    singleHuman {
                        id
                        friends {
                            name
                        }
                    }
                }
            ";
            var expected = @"
                {
                    singleHuman: null
                }
            ";
            AssertQuerySuccess(query, expected);
        }

    }
}
```

I tracked it down, and in ExecutionHelper.cs method SubFieldsFor, it is only being applied to ObjectGraphTypes. 

```C#
 public static IDictionary<string, Field> SubFieldsFor(ExecutionContext context, IGraphType fieldType, Field field)
        {
[...]

PR #548, opened 2018-02-06
Populate subfields for all graph types that have a SelectionSet. #547

TwitchBronBron:
Fixed bug where the ResolveFieldContext.SubFields was null for ListGraphType items.

### DIFF
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -388,7 +388,9 @@ namespace GraphQL.Execution
 
         public static IDictionary<string, Field> SubFieldsFor(ExecutionContext context, IGraphType fieldType, Field field)
         {
-            if (!(fieldType is IObjectGraphType) || !field.SelectionSet.Selections.Any())
+            var selections = field?.SelectionSet?.Selections;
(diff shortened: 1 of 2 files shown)


src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -388,7 +388,9 @@ public static bool DoesFragmentConditionMatch(ExecutionContext context, string f
 
         public static IDictionary<string, Field> SubFieldsFor(ExecutionContext context, IGraphType fieldType, Field field)
         {
-            if (!(fieldType is IObjectGraphType) || !field.SelectionSet.Selections.Any())
```
TwitchBronBron: I'm not sure why there was a restriction to only IObjectGraphType. I changed this to simply verify that there are selections available.

joemcbride:
Thanks for this!  Would you mind adding another test for interfaces?  The StarWars schema has a `CharacterInterface` type.  Could also use some tests for Unions and fragments, though the StarWars schema doesn't have any Unions.  I'll pull this in after the interface test is added.

TwitchBronBron:
I added tests for both InterfaceGraphType and ListGraphType<InterfaceGraphType>.

PR #552, opened 2018-02-07
Spelling changes that affect the public API.

Grauenwolf:
This might cause some problems if people are depending on the misspelled names.


joemcbride:
Closing as most of these are fixed in other PRs.

PR #584, opened 2018-03-04
Added an error code to the errors json when the error is a ValidationError and the ErrorCode was provided

wilsonrivera:
This simple modification will add the JSON property "code" to the generated errors JSON when the error is an instance of the class ValidationError and the property ErrorCode isn't empty or whitespace.


joemcbride:
Thanks for the PR!  Pondering on this one for a little bit but wanted to let you know I've looked at it.

wilsonrivera:
Sorry for the delay, I agree, no need. closing this pull request.

PR #586, opened 2018-03-06
Adds flexible type registry

jphenow:
Thoughts?

Proposing so the library doesn't have to worry about any other scalars than listed, but it'd be great if I didn't have to `ToString()` every time I have a Guid on the API.

Happy to touch up further to match project style and add tests, just wanted to get a rough implementation down before I forgot and moved on.

### DIFF
--- a/src/GraphQL/TypeExtensions.cs
+++ b/src/GraphQL/TypeExtensions.cs
@@ -129,42 +129,6 @@ namespace GraphQL
                 }
             }
 
-
-            if (type == typeof(int))
-            {
-                graphType = typeof(IntGraphType);
-            }
-
-            if (type == typeof(long))
-            {
-                graphType = typeof(IntGraphType);
-            }
-
-            if (type == typeof(double) || type == typeof(float))
-            {
-                graphType = typeof(FloatGraphType);
-            }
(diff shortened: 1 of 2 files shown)


joemcbride:
👍 I would probably change the `Stack` to a `Dictionary` + `TryGetValue` for lookup.

jphenow:
The purpose of the Stack was simply so any added classes that overlap would default to user-added while retaining the original (in case at some point the system would want to force the system-default) but that's overkill now that I say it loud - especially since it isn't currently doing that.

Alright, yea, I'll switch it :-)

src/GraphQL/TypeExtensions.cs
```diff
@@ -179,6 +143,8 @@ public static Type GetGraphTypeFromType(this Type type, bool isNullable = false)
                 graphType = listType.MakeGenericType(elementType);
             }
 
+            graphType = GraphQL.GraphTypeRegistry.Get(type);
```
joemcbride: I believe this needs to be moved above the `type.IsArray` check to fix the tests.

jphenow:
Woot! Thanks!

Issue #582, opened 2018-02-28
Avoiding strong typing and deserialization into C# schemas as part of resolvers?


ravgupms:
I'm building a GraphQL Aggregate service which won't be owning any data by itself, instead will call a number of underlying services to fetch the data.

I'm using Handler/Endpoints approach but wanted to check is there any way to skip the strong typing and deserialization into C# schemas as part of resolvers? Something like a resolver returns data in JSON format(JObject) and GraphQL package will automatically picks necessary data out of it (assuming both uses same schema) without deserializing it to object first and then again serializing it to JSON, hence improving overall perf of our service.

joemcbride:
No this is not currently supported.  It would most likely require significant changes to the current design of the library.

ravgupms:
But are there any plans to support something like this in the near future?

joemcbride:
There is no current plan to try to support this scenario in the near future.  You could perhaps get close to something working with implementing your own `IFieldResolver` (work with `JObjects` vs. class properties + methods).  Though the internal system would still need the GraphQL Schema Model to know how to execute the GraphQL requests.

elauffenburger:
I actually came here to ask this exact question!  I couldn't find a good solution, so I'm wrapping up something I've been working on.  It's...a bit of a hack as it requires dynamic type generation, but it fetches the remote schema and generates subtypes that just proxy through a JObject.

@joemcbride: do you have any interest in me forking to show that work to see if want to take it over?

Oh, and great job on this library -- it's been really invaluable to me and the rest of my team!

joemcbride:
@elauffenburger For sure, I would be interested in seeing what you came up with.

ravgupms:
@elauffenburger : Even I want to have a look too.

elauffenburger:
Awesome! I'll get something together in the next day or two since I need to wrap this up, get it pulled out of my NuGet package, scrub it a bit (😄), and get you a PR!

If I seem to fall off the face of the planet, it's probably because I'm dealing with some fire or something, so feel free to just message me 😄

elauffenburger:
@joemcbride: So I started porting my work over, and the only real issue I'm seeing is that I use some very Full .NET FX-y APIs like `AppDomain.CurrentDomain` or `AppDomain.DefineDynamicAssembly`, which is going to be a problem for netstandard2.0 and netstandard1.3 support.

[...]

PR #592, opened 2018-03-09
Add GraphQL.Dynamic for distributed queries

elauffenburger:
@joemcbride @ravgupms: Here's the PR for our discussion in #582 

**Obviously** this isn't ready for prod (and we've talked about how to make this netstandard compliant), but this should give you an idea of what I was thinking.  I don't _think_ it should be too much work to use Roslyn, so I believe a lot of the core implementation could remain the same if we're in agreement 😄


elauffenburger:
@joemcbride: can I revive this?  I'd love to get your input!

joemcbride:
@elauffenburger Hey, sorry I haven't been able to dig into this at all yet.  Still on my "at a future date" list.  ☹️

PascalSenn:
Are you creating types for the whole remote schema?

elauffenburger:
@joemcbride all good! Just wanted to keep the conversation alive :)

@PascalSenn yep! But obviously just custom types (as in, types graphql.net isn't aware of by default)

PascalSenn:
@elauffenburger 
Alright. Alright. 

We made some thoughts about this and there are many different ways to stitch a schema together and all with different ups and downs. 

There is the way you chose with generating the schema based on the introspection. 
Ups: 
- it gives the possibility to actually stitch a schema together. That means with some engineering you could also automatically merge types. 

Downs: 
You have to resolve a type twice one in every backend. 

Another way would be to create another ComplexGraphType in which you configure the endpoint, the prefix,  argument and a root query on the endpoint. 
Then on the introspection you just "bubble" up the introspection and decorate it with a prefix. The resolver also just bubbels the resolved json up. 

Downs: 
You'd only be able to stitich together queries and not able to merge trees

Ups: 
the resolving cost would be less cost intensive. 

Conclusion: 
I guess it wouldn't be too bad to have both implementations at the end. Because the use cases are slightly different. What do you think @joemcbride ?

Some follow up questions: 
How hard is it to implement a dataloader in your solution? 
Do you prefix and translate types? 
How about arguments?

OneCyrus:
this looks nice @elauffenburger! as this PR doesn't change anything in graphql-dotnet itself maybe we can polish this as a separate package which doesn't need to live inside the main repo?

we are really interested to have proper schema stitching in graphql-dotnet. this seems to be a good start and with some effort we should be able to move the codegen to roslyn and improve on some edge cases.

src/GraphQL.Dynamic.Tests/BasicTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            Assert.NotNull(queryResult.SelectToken("githubDynamic.user.id").Value<object>());
+            Assert.NotNull(queryResult.SelectToken("githubDynamic.user.company").Value<object>());
+            Assert.NotNull(queryResult.SelectToken("githubDynamic.user.repos").Value<object>());
+            Assert.NotNull(queryResult.SelectToken("githubDynamic.user.repos[0].id").Value<object>());
+            Assert.NotNull(queryResult.SelectToken("githubDynamic.user.repos[0].name").Value<object>());
+            Assert.NotNull(queryResult.SelectToken("githubDynamic.user.repos[1].id").Value<object>());
+            Assert.NotNull(queryResult.SelectToken("githubDynamic.user.repos[1].name").Value<object>());
+        }
+
+        private string GetFileContentsForTestFile(string relativePath) => File.ReadAllText(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), relativePath));
```
OneCyrus: btw. this gives the wrong path (shadowed assembly path). using the following gives the correct path:

```Path.GetDirectoryName(Uri.UnescapeDataString(new Uri(Assembly.GetExecutingAssembly().CodeBase).AbsolutePath))```

File: src/GraphQL.Dynamic/LiteralGraphType/RemoteLiteralGraphType.cs [hunk 2]
```diff
@@ -0,0 +1,344 @@
...
+            }
+
+            return schema != null;
+        }
+
+        private static Introspection.Schema FetchRemoteSchemaViaHttp(string url)
+        {
+            using (var client = new HttpClient())
+            {
+                var query = new
```
OneCyrus: just had some issues with mediaheaders. this helps to set the accept format.

```
client.DefaultRequestHeaders
  .Accept
  .Add(new MediaTypeWithQualityHeaderValue("application/json"));
```

src/GraphQL.Dynamic/LiteralGraphType/RemoteLiteralGraphType.cs
```diff
@@ -0,0 +1,344 @@
...
+                {
+                    Query = SchemaIntrospection.IntrospectionQuery
+                };
+
+                var response = client.PostAsync(new Uri(url), new StringContent(JsonConvert.SerializeObject(query)))
```
OneCyrus: as above with the accept header, setting the content type helped to solve an issue in our environment:
```
var response = client.PostAsync(new Uri(url), new StringContent(JsonConvert.SerializeObject(query), Encoding.UTF8, "application/json"))
```

elauffenburger:
Awesome! I'm really glad this is gaining some traction (and thanks to @PascalSenn for such an in-depth analysis 😄)

I actually have a number of changes and bug fixes that I need to merge into this PR (@OneCyrus has already found most of them!)

I'll try to do that ASAP so no one is churning on this too much!

CC @joemcbride

src/GraphQL.Dynamic/LiteralGraphType/LiteralGraphTypeHelpers.cs
```diff
@@ -0,0 +1,141 @@
...
+        private static void PopulateMappings()
+        {
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.String, typeof(string), CreateFieldType<StringGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Int, typeof(int), CreateFieldType<IntGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Long, typeof(long), CreateFieldType<IntGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Float, typeof(float), CreateFieldType<FloatGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Double, typeof(double), CreateFieldType<FloatGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Boolean, typeof(bool), CreateFieldType<BooleanGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Guid, typeof(Guid), CreateFieldType<IdGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.DateTimeOffset, typeof(DateTimeOffset), CreateFieldType<DateGraphType>);
```
OneCyrus: DateTime is missing:

```
PopulateTypeMapping(LiteralGraphTypeMemberInfoType.DateTime, typeof(bool), CreateFieldType<DateGraphType>);
```

OneCyrus:
that sounds great! btw. did you look into changing to roslyn code generation yet? we are a bit stuck with the current version as we would need this in a .net core application.

and another feature would be nice to have is custom scalar types.

src/GraphQL.Dynamic/LiteralGraphType/LiteralGraphTypeHelpers.cs
PascalSenn: Also: 
I'm not so sure about this line 
```csharp
PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Guid, typeof(Guid), CreateFieldType<IdGraphType>);
```
because a GUID is not certainly a `IdGraphType` as a `IdGraphType` is not always a GUID
https://github.com/graphql-dotnet/graphql-dotnet/blob/c4bc7ca006d3ab5d7e60089c12e8e7038ee3797f/src/GraphQL/Types/IdGraphType.cs#L27

OneCyrus:
two more questions about using this library:

- is there an easy way to "generate" the query which i would send to the remote server inside the RemoteField resolver? something like generate query from resolver "ctx.FieldAST" (with children) and map this to the typeName from the remote schema.

- did you already experiment with filtering variables/arguments to only pass the correct ones to the remote server?

PascalSenn:
@OneCyrus  

- You can use the AstPrinter to Print a graphql query `AstPrinter.Print(context.FieldAst.SelectionSet)`

OneCyrus:
@PascalSenn that's nice! now we need to have an AstPrinter which can print the AST in the other direction in an other schema.

**public schema**
```
Query {
   LocalType {
      RemoteType {
         MyRemoteField
      }
      OtherRemoteType {
         OtherRemoteField
      }
   }
}
```

**schema from remote server**
```
Query {
   RemotePrimaryType {
     EvenDeeper {
        RemoteType {
           MyRemoteField
        }
        NotQueriedRemoteType {
           NotQueriedField
        }
     }
   }
}
```

**AST printer would give me**
```javascript
{
   RemoteType {
       MyRemoteField
    }
}
```

**and this is what i would need**
``` 
Query {
  RemotePrimaryType {
    EvenDeeper {
      RemoteType {
        MyRemoteField
      }
    }
  }
}
```

though it could get difficult if there are multiple times the same type in the remote schema. GraphlQL.Dynamic currently has a type as en entry point. And in this direction it's not necessary to know where this type is in the AST. but resolving it back to the remote schema we are missing the path to the type. as in the test case in this PR (https://github.com/graphql-dotnet/graphql-dotnet/pull/592/files#diff-3c2a7abce1b810eb64bdad8bfd0ea9b5R21) "GithubAPI"/GithubMoniker should probably be a path and not just a type/name.

PascalSenn:
@OneCyrus  
Funny enough. I tried to get all variable references to pass down variables for stitching and was going to refactor ASTPrinter. Would have been a good moment to add a reverse printer.
But @joemcbride  showed me the BasicVisitor. 

I guess we could use something like this. 
https://github.com/graphql-dotnet/graphql-dotnet/issues/690 

Just pattern match around the different node types should do the job

joemcbride:
May want to have a look at how `DocumentValidator` walks the tree.  It uses `TypeInfo` to keep track of context within the document tree (so you can look at parent types and such).

https://github.com/graphql-dotnet/graphql-dotnet/blob/3f2355aacd6cfd8e080dba1d802b2fa00f958f85/src/GraphQL/Validation/DocumentValidator.cs#L35-L59

OneCyrus:
@elauffenburger in the next couple of days we would like to start with adopting this approach for our project. would it be possible for you as mentioned to address the known bugs or should we fork this from the current state?

@PascalSenn sounds good!

OneCyrus:
@elauffenburger I just created a PR to change to roslyn code generation and .net standard

https://github.com/elauffenburger/graphql-dotnet/pull/1

src/GraphQL.Dynamic/LiteralGraphType/RemoteLiteralGraphType.cs
```diff
@@ -0,0 +1,344 @@
...
+
+                // get type
+                if (!TryGetFieldTypeFromRemoteSchema(schema, _name, out var type))
+                {
+                    // if no type found: fail
+                    // TODO: fail better
+                    throw new Exception($"Failed to find type '{_name}' in remote '{_remoteLocation}' schema");
+                }
+
+                Name = $"{_remoteLocation}.{_name}";
```
OneCyrus: this generates not valid names. changing to ```Name = $"{_name}";``` should fix this. or is there a reason other than duplicated types for this?

src/GraphQL.Dynamic/LiteralGraphType/RemoteLiteralGraphType.cs
```diff
@@ -0,0 +1,344 @@
...
+
+                if (!RemoteServerTypes.TryGetValue(remote, out var remoteTypes))
+                {
+                    return null;
+                }
+
+                var schemaType = remoteTypes.FirstOrDefault(t => t.Name == literalMember.TypeName);
+                var realType = literalMember.IsList
+                    ? typeof(ListGraphType<>).MakeGenericType(schemaType)
+                    : schemaType;
```
OneCyrus: this fails if the schemaType is a base GraphQL type e.g. a ListGraphType<StringGraphType>

elauffenburger:
Hey guys, sorry I dropped off the face of the Earth; a bit busy at the day job :)

I'd like to address the bugs here to unblock @onecyrus & co, so that'll be my first priority; I'll port everything over from our codebase to the PR and then address comments brought up here.

Thanks for everyone's support!

cc: @joemcbride

src/GraphQL.Dynamic/LiteralGraphType/LiteralGraphTypeHelpers.cs
elauffenburger: @OneCyrus: fixed as part of my merges; thanks for the great find 😄 
@PascalSenn: yeah, honestly I don't know the GraphQL spec well enough to know the right way to handle this.  I just know that it's supposed to be a unique value that serializes as a string -- any help would be appreciated here!

src/GraphQL.Dynamic/LiteralGraphType/RemoteLiteralGraphType.cs
elauffenburger: yep; you've got it -- it's to avoid name collisions.  This should be fixed as part of merging my changes in!

src/GraphQL.Dynamic/LiteralGraphType/RemoteLiteralGraphType.cs
elauffenburger: Again, great catch; this is fixed per my incoming fix commits

elauffenburger:
Alright, I just kinda sat down and finally did it 😄
Let's give it another look and try to wrap this in the next week or so!

cc: @PascalSenn @OneCyrus @joemcbride

OneCyrus:
@elauffenburger thanks a lot! I'll have a look. 
for the codegen PR I've some additional changes which I'll push there in the next days.

src/GraphQL.Dynamic/LiteralGraphType/LiteralGraphTypeHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Double, typeof(double), CreateFieldType<FloatGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Boolean, typeof(bool), CreateFieldType<BooleanGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Guid, typeof(Guid), CreateFieldType<IdGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.DateTime, typeof(DateTimeOffset), CreateFieldType<DateGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.DateTimeOffset, typeof(DateTimeOffset), CreateFieldType<DateGraphType>);
+        }
+
+        internal static Type GetPrimitiveGraphType(string name) => _graphTypesLookup[name]?.GetType();
+
+        internal static string GenerateRemoteTypeName(string remoteLocation, string name) => SanitizeTypeNameRegex.Replace($"{remoteLocation}.{name}", "_");
```
OneCyrus: putting the full url in front of every remote type gives a very verbose schema. the docs look very busy as well. can't we find a simpler way to solve this? e.g. using the moniker or an intelligent solution which only triggers for conflicting types.

OneCyrus:
i just merged your branch in our internal exploration codebase and played around with it. while it works good for what it does right now we still have three major limitations which we need to tackle before we can use it for real use cases.

- Arguments: currently the inputtypes are not assigned to the graphtypes. do you have any ideas how to do this the best way?

- UNION: we are using unions and would need to have some basic union support. from your comments ii get that you made some thoughts about them but they are not that easy to implement. we thought about them a bit as well and probably the only way to do this would be to bet on forcing __typename for unions in the query to the remote location. dynamic deserialization with type lookup should be possible.

- dynamic query builder: fully generate the query from all the information inside the graphql AST and remote schemas to request the data from remote location. did you implement something like that yet or are you using more or less fixed queries to request the data?

i guess we already made quite some progress with this PR and once we have the remaining limitations fixed it will be an awesome solution.

src/GraphQL.Dynamic/LiteralGraphType/LiteralGraphTypeHelpers.cs
```diff
@@ -0,0 +1,158 @@
...
+        private static void PopulateMappings()
+        {
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.String, typeof(string), CreateFieldType<StringGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Int, typeof(int), CreateFieldType<IntGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Long, typeof(long), CreateFieldType<IntGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Float, typeof(float), CreateFieldType<FloatGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Double, typeof(double), CreateFieldType<FloatGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Boolean, typeof(bool), CreateFieldType<BooleanGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.Guid, typeof(Guid), CreateFieldType<IdGraphType>);
+            PopulateTypeMapping(LiteralGraphTypeMemberInfoType.DateTime, typeof(DateTimeOffset), CreateFieldType<DateGraphType>);
```
OneCyrus: not sure if this should be DateTimeOffset or DateTime

elauffenburger:
@OneCyrus: thanks for taking a look and for giving it a semi-⭐️ of approval 😄 

So to answer your questions:
  - Arguments:
    - I'll need to take a look at this -- I thought I implemented that, but I may have missed it...
  - Unions:
    - Yeah, I haven't really event given it a thought tbh; obviously we'll have to support it, so my best approach (off the top of my head) would be exactly what you suggested.  I don't think there's an issue with deserialization (since we really have *no* idea what the remote is returning us); i think the only challenges will come from supporting them from an "infrastructure" perspective (as in, we have to float the fact that these are unions along to field resolvers).  I don't think it will be too challenging to implemented
  - Dynamic query builder:
    - I don't have a good solution for this yet 😭.  We've just been taking the selection set from the `FieldAst` and throwing that at the remote.

PascalSenn:
@elauffenburger  

- Dynamic query builder:
Thats basically what we are doing too.
I just added #710 yesterday to support inline fragments in AstPrinter.
With `BasicVisitor` it's easy to walk down the tree and get all variable references used in the selection set. 
Based on that we can get all variable definitions that map with the variable references and add them to a new document. We can also add the operation name, type and the selection set. 

To add the `__typename`'s I've got a bit of a controversial idea. 
The problem we face is, that we should not mutate the SelectionSet and I doubt it's easy to clone it. 
We could just make this protected:
https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/AstPrinter.cs#L155

Create our own Printer that extends AstPrinter that automatically includes all typenames: 
https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/AstPrinter.cs#L227-L234

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/AstPrinter.cs#L393-L399

If we request __typenames in the frontend and request them from the backend and then further from a remote schema, we would request them twice. But I guess that wouldn't really be a problem. If i request:
image

I get the response type only once. 
image

I'm not sure if they acutally get resolved multiple times. But even if, there's little overhead and we could still check if the first item of a selection set is already a __typename. (So in deeply nested schemas we would request them only twice) 

Maybe @joemcbride  has a better idea. 

Maybe interesting for https://github.com/graphql-dotnet/graphql-client/issues/42 @deinok

thivy:
@joemcbride is this part of the main package now?

joemcbride:
@thivy there are no plans to merge this at this time.  This is based on my ability to support this code in the future.  Right now that is not possible.

SimonCropp:
I might be missing something. But why cant this ship as a stand alone nuget and live in its own repo?

deinok:
@elauffenburger Can you give me a TLDR for this, as @PascalSenn said that It could be interesting for the GraphQL Client?

OneCyrus:
@deinok what @PascalSenn meant is that with introspection and code generation (what this PR here actually does), you could do a fully typed graphql client.

@PascalSenn and I are moving over to a competing solution for schema stitching in .Net. We won't invest more into this PR as well.

deinok:
Sure, code generation using something like this : Microsoft.CodeAnalysis.CSharp from the Introspection model as an MSBuild Task would be awesome

src/GraphQL.Dynamic/Introspection/Types.cs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    public static class Serialize
+    {
+        public static string ToJson(this Root self) => JsonConvert.SerializeObject(self, GraphQL.Dynamic.Types.Introspection.Converter.Settings);
+    }
+
+    internal class Converter : JsonConverter
+    {
+        public override bool CanConvert(Type t) => t == typeof(TypeElementTypeKind) || t == typeof(TypeElementTypeKind) || t == typeof(TypeElementTypeKind) || t == typeof(TypeElementTypeKind) || t == typeof(TypeElementTypeKind?) || t == typeof(TypeElementTypeKind?) || t == typeof(TypeElementTypeKind?) || t == typeof(TypeElementTypeKind?);
```
kalexii: Hey, guys. Just browsing through the code, anyone seeing anything weird here?

src/GraphQL.Dynamic/LiteralGraphType/LiteralGraphType.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            if (listType == null)
+            {
+                return null;
+            }
+
+            return listType.GetGenericArguments().FirstOrDefault();
+
+        }
+        private static bool IsListType(Type type) => type.GetInterfaces().Any(i => i == typeof(IEnumerable));
```
kalexii: string counts too?

codestellar:
This PR is open since long! Will there be any action on it?

joemcbride:
@codestellar no, this won’t be pulled in.  Schema stiching has been superseded by Apollo Federation. See #1169

Issue #560, opened 2018-02-15
DataLoader documentation


PrimeHydra:
Super excited about the DataLoader #539, but can't find anywhere how to use it. I'm new to GraphQL.NET so learning by squinting at the tests is sub-optimal. Thanks!

johnrutherford:
I can work on this. @joemcbride, where should this go? In the wiki?

joemcbride:
Docs are markdown files found in the docs folder.  https://github.com/graphql-dotnet/graphql-dotnet/tree/master/docs

Pets:
@johnrutherford Did you had the time to work on the documentation ? :)
Like @PrimeHydra I'm super excited about this feature... but it's unclear for me how to work with it.
Thank you!

johnrutherford:
I've started on it, but I'm not done yet. Haven't had much time to work on it. I'll try to finish it this week. 

You might be able to figure it out from looking at the test project:
https://github.com/graphql-dotnet/graphql-dotnet/tree/master/src/GraphQL.DataLoader.Tests

The `IDataLoaderContextAccessor` and `DataLoaderDocumentListener` need to be configured in your DI container. Then you need to add the `DataLoaderDocumentListener` instance as a listener on the `DocumentExecuter`. Then check out the `OrderType` and `UserType` to see how to get a loader instance and use it in a resolver.

sommarlov:
I've been using this and its fantastic so far. Great work!

While waiting for the documentation, I have a question. Is there a way to do .LoadAsync() on a list of Ids? Lets say for example (looking at the unit test Types) that User already has the Order ids. In this case I would like to load the orders by a list of order ids instead of the User id.

I hope my explanation makes sense, thanks!

MakaBuka:
I've just investigated implementation of DataLoader and have a few questions:
1) is it correct, that now, dataloader just cache similar queries? lets say get user (id: 1) {name} more that once
2) I haven't found solution for n+1 issue (classic case: we have 3 requests in our resolver - user(id:1){name}, user(id:2){name}, user(id:3){name} - it should be combined in the end to sth like this - users(id in (1,2,3)) {name}
--
Could u please advice how we can resolve issue that IS described above?
Thx, D.

johnrutherford:
Ok, I've finally completed a first draft of the DataLoader documentation. I'd welcome any feedback on how it could be improved or made clearer. 

@sommarlov That makes sense. There's no way to do that currently, but it could be added.

[...]

PR #598, opened 2018-03-14
Add documentation for the DataLoader implementation

johnrutherford:
Finally adding DataLoader documentation for #560.

### DIFF
--- a/src/GraphQL/DataLoader/DataLoaderDocumentListener.cs
+++ b/src/GraphQL/DataLoader/DataLoaderDocumentListener.cs
@@ -5,6 +5,10 @@ using GraphQL.Validation;
 
 namespace GraphQL.DataLoader
 {
+    /// <summary>
+    /// Used to manage the <seealso cref="DataLoaderContext"/>
+    /// and automatically dispatch data loader operations at each execution step.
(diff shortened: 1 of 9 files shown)


src/GraphQL/DataLoader/DataLoaderDocumentListener.cs
```diff
@@ -34,6 +38,8 @@ public Task BeforeExecutionAwaitedAsync(object userContext, CancellationToken to
 
         public Task AfterExecutionAsync(object userContext, CancellationToken token)
         {
+            _accessor.Context = null;
```
joemcbride: Should perhaps `DataLoaderContext` implement `IDisposable` and be disposed here, which would clear out any queue and loader references?

johnrutherford: I don’t think that would be necessary. There’s nothing that needs to be disposed of before garbage collection.

johnrutherford: This change probably isn’t necessary. I just wanted to make sure that the context would not be reused for another execution.

Issue #404, opened 2017-08-23
Support conversion to Array properties in GetArgument<T>


petrovic-ivan:
I encountered this error when tried to obtain argument from the context in mutation.

Here is the model:
`public class Job
    {
        public double[] Jobs { get; set; }
    }`

Here is GraphQL type:
```
public class JobType : ObjectGraphType<Job>
 {
        public JobType()
        {
            Description = "Something.";

            Field<ListGraphType<FloatGraphType>>("Jobs", "Many jobs.");
        }
 }
```

Here is input type:
```
public class JobInputType : InputObjectGraphType
  {
       public JobInputType()
        {
            Description = "Something.";

            Field<ListGraphType<FloatGraphType>>("Jobs", "Many jobs.");
        }
 }
```

And here is mutation:
```
public class BatchMutation : ObjectGraphType<object>
    {
        public BatchMutation()
        {
            Name = "BatchMutation";

            Field<JobType>(
                "createJob",
                arguments: new QueryArguments(
                    new QueryArgument<NonNullGraphType<JobInputType>> { Name = "batch" }
                ),
                resolve: context =>
                {
                    return context.GetArgument<Job>("batch");
                }                  
            );
        }
    }
```
Error happoened in this line `context.GetArgument<Job>("batch");` . 

If I change the type of `Jobs` from `double[]` to `List<double>` it works properly.

Am I doing something wrong or this is bug?

joemcbride:
Not a bug per say, but a missing feature.  `GetArgument<T>` calls into `ToObject`.  Right now, that doesn't support Array properties.  Would just need to add that feature.

https://github.com/graphql-dotnet/graphql-dotnet/blob/4da2d8af548b4f0d9cd620da2504781861cf6e37/src/GraphQL/ObjectExtensions.cs#L20

wiktor-golonka:
Hi, 
I did a quick fix for that inside ObjectExtensions.GetPropertyValue method. 
Code below:

```
        /// <summary>
        /// Converts the indicated value into a type that is compatible with fieldType.
        /// </summary>
        /// <param name="propertyValue">The value to be converted.</param>
        /// <param name="fieldType">The desired type.</param>
        /// <remarks>There is special handling for strings, IEnumerable&lt;T&gt;, Nullable&lt;T&gt;, and Enum.</remarks>
        public static object GetPropertyValue(this object propertyValue, Type fieldType)
        {
            // Short-circuit conversion if the property value already
[...]

PR #601, opened 2018-03-19
Support for arrays conversions #404

wiktor-golonka:
Issue page: https://github.com/graphql-dotnet/graphql-dotnet/issues/404

### DIFF
--- a/src/GraphQL.Tests/ObjectExtensionsTest.cs
+++ b/src/GraphQL.Tests/ObjectExtensionsTest.cs
@@ -8,6 +8,8 @@ using Xunit;
 
 namespace GraphQL.Tests
 {
+    using Xunit.Sdk;
+
     public class ObjectExtensionsTests
     {
         [Fact]
@@ -74,17 +76,17 @@ namespace GraphQL.Tests
             actual.ShouldBe(value);
         }
         
-                [Fact]
+        [Fact]
         public void convert_double_array_to_array()
         {
             // Arrange
             var doubles = new double[] { 1.00, 2.01, 3.14 };
 
             // Act 
-            var result = doubles.GetPropertyValue(typeof(double[]));
+            var actual = doubles.GetPropertyValue(typeof(double[]));
 
             // Assert
-            Assert.Equal(result, doubles);
+            actual.ShouldBe(doubles);
         }
 
         [Fact]
@@ -94,10 +96,10 @@ namespace GraphQL.Tests
             var doubles = new List<double>() { 1.00, 2.01, 3.14 };
 
             // Act 
-            var result = doubles.GetPropertyValue(typeof(double[]));
+            var actual = doubles.GetPropertyValue(typeof(double[]));
 
             // Assert
-            Assert.Equal(result, doubles);
+            actual.ShouldBe(doubles);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


src/GraphQL/ObjectExtensions.cs
```diff
@@ -78,6 +78,19 @@ public static object GetPropertyValue(this object propertyValue, Type fieldType)
                 var underlyingType = Nullable.GetUnderlyingType(elementType) ?? elementType;
                 var implementsIList = fieldType.GetInterface("IList") != null;
 
+                if (fieldType.IsArray)
+                {
+                    if (!(propertyValue is IList values))
```
joemcbride: Can this perhaps be changed to a more generic `IEnumerable`?

src/GraphQL/ObjectExtensions.cs
```diff
@@ -78,6 +78,19 @@ public static object GetPropertyValue(this object propertyValue, Type fieldType)
...
+                    {
+                        return null;
+                    }
+
+                    var tempArray = Array.CreateInstance(elementType, values.Count);
```
joemcbride: I think this should also use `underlyingType` vs. `elementType`, since the `elementType` could be nullable.  That would be a good test to add.

wiktor-golonka: Hi, 
I've added some unit tests and changed `IList` to `ICollection`.
If we want to have an array: `double?[]`, then I believe taking the `elementType` is the only way as it preserve information about nullablelity. 
With `underlyingType` we would get the `double[]`

src/GraphQL.Tests/ObjectExtensionsTest.cs
```diff
@@ -72,6 +72,110 @@ public void convert_single()
...
+        public void convert_double_array_to_array()
+        {
+            // Arrange
+            var doubles = new double[] { 1.00, 2.01, 3.14 };
+
+            // Act 
+            var result = doubles.GetPropertyValue(typeof(double[]));
+
+            // Assert
+            Assert.Equal(result, doubles);
```
joemcbride: This project uses `Shoudly` if you can please update these assertions to that.

`result.ShouldBe(doubles)`.

src/GraphQL/ObjectExtensions.cs
joemcbride: 👍 to using `elementType`.

Is the reason for checking for `ICollection` so that you can use `CopyTo`?  One test that we're missing is one for an array of objects (or an array of arrays).  I don't think you'll be able to use `CopyTo` with that.  So I think that check should just be against `IEnumerable` and then build up the values.

```
var valueList = propertyValue as IEnumerable;
if (valueList == null) return newArray;

valueList.Apply((listItem, idx) =>
{
  tempArray[idx] = listItem == null ? null : GetPropertyValue(listItem, elementType);
});

return tempArray;
```

wiktor-golonka:
Hi @joemcbride 
Thanks for review. You are right, I had to modify the code to support multidimensional arrays. 
Code is now just a very little bit more complex but should work as intended.
I've added tests to cover that case.

However I did found a potential issue: when querying for an object with a field: ListGraphType<ListGraphType<StringGraphType>> - it was showing always null items for me.

wiktor-golonka:
Hi @joemcbride , 
Hope you are well.
Did you had a chance to take a look on this solution?

I did it in the very similar way you proposed however, I couldn't use the Apply method because it requires IEnumerable<T> but here we have only IEnumerable. 
Also we don't have a Count() method so I had to iterate over that collection to get a length of the array before created new one.

It's passing all the tests, also I checked it on my solution.

wiktor-golonka:
@joemcbride: I cleaned this code so it should be more readable.

PR #603, opened 2018-03-22
Include symbols for the nuget packages

derwasp:
I think it could be useful for debugging purposes:
Added `--include-symbols` and `--include-source` flags to the main nuget package in order to get pdbs and sources available via the source server.
This will not affect the main package, it will just create a separate `symbols.nupkg` which will be pushed to the symbol server (See [docu](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-pack?tabs=netcore2x#description) for details)

I can't seem to find the logic which pushes to nuget.org, only to myget. Could you please help me out with that? In case you want to have this change of course.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,7 +31,7 @@ deploy:
   server: https://www.myget.org/F/graphql-dotnet/api/v2/package
   api_key:
     secure: +l1vfBMajn1WfmXkQ2LdILKxK4fQ5AHSnnU1kf11Bn1xRGUOTCdPhLwHx232piEn
-  skip_symbols: true
+  skip_symbols: false
   on:
     branch:
       - master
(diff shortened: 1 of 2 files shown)


derwasp:
Hm. I didn't touch the code. Why did the test fail? Can you please try to restart the build?

pekkah:
You can close and reopen the PR to trigger the build

joemcbride:
So, I have no problem including symbols as I can see that being useful.  My main concern is running out of space quickly on myget.org (only 500MB allowed).  In addition, right now pushing to nuget.org happens from myget.org, so also pushing symbols may be a second manual step.

What I'm thinking is: merge this and try it out.  If it becomes a problem then revert.

derwasp:
@joemcbride thank you for the merge!
One way around the limitation might be to just shovel everything to nuget ;)

PR #604, opened 2018-03-22
suppress obsolete warning inside the src

VVildVVolf:
suppress "obsolete" warning during the build

cleanup the null refferences inside dispose (https://social.msdn.microsoft.com/Forums/en-US/3532073e-816a-4f66-9922-9f527e4cca6b/setting-object-to-null-after-dispose?forum=csharplanguage). It is not so required, but why do we need to keep unrequiered code..

### DIFF
--- a/src/GraphQL/Types/Schema.cs
+++ b/src/GraphQL/Types/Schema.cs
@@ -73,7 +73,9 @@ namespace GraphQL.Types
         public Schema(IDependencyResolver dependencyResolver)
         {
             DependencyResolver = dependencyResolver;
+            #pragma warning disable 0618
             ResolveType = type => dependencyResolver.Resolve(type) as IGraphType;
+            #pragma warning restore 0618
 
             _lookup = new Lazy<GraphTypesLookup>(CreateTypesLookup);
             _additionalTypes = new List<Type>();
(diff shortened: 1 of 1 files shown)


joemcbride:
Hey, thanks for the PR though I've decided to not pull this in.  That func is marked as obsolete for users to have an easier upgrade path.  It will be removed in the next major version.  I would rather see the warnings than forget about it later.

Also, that link you posted does not apply to the purpose of setting the properties to null.  Setting the properties to null is removing those references so they can get garbage collected.

See this note from that link:

> Setting an object (reference) to null is NOT required to get the item to be garbage collected, unless failing to do so would ensure a (strong) reference to the object is kept alive.

So this is ensuring there is no strong reference (which there may be).

VVildVVolf:
Hello, thank you for the response. I would suggest to use the “todo” statement instead of keeping warning, because of warning says “there is something bad”, but todo says “do not forget to do this”. About nulls - sorry, I have provided the bad link. The main idea is that this object is not reusable (at least because it is disposable). The appropriate way of using this object is using the “using” statement (or kill it in another dispose call), so the code will not keep these links after dispose call, if it is written ok

Thank you, 
VVildVVolf

pekkah:
Using obsolete for obsolete methods is quite common and getting the warning about is quite useful so people know to migrate away from it.

VVildVVolf:
>> Using obsolete for obsolete methods is quite common and getting the warning about is quite useful so people know to migrate away from it.

I understand. But we tald about keeping in memory to remove obsolete property. Users who use this property will see the warning and look for an workaround. There is the suggestion to suppress the warning only because we cannot do not use it, but to not forget to remome it there is the "todo" statement

Thank you,
VVildVVolf

PR #612, opened 2018-03-30
Fix regression with private output class

csantero:
When upgrading my project from alpha-802 to alpha-870 I noticed some strange errors coming from `GraphQL.ExecutionHelper.UnwrapResultAsync`:

> 'System.Threading.Tasks.Task' does not contain a definition for 'Result'

I was able to determine that this occurs when you have a field that returns a `Task<T>`, where `T` is a private class. The dynamic lookup doesn't see `task` as being a `Task<T>` when the generic type is private, but rather only a `Task`, which it can't find a `Result` property on. I was able to solve the problem by using reflection to get `Result` instead of `dynamic`.

### DIFF
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Threading.Tasks;
 using GraphQL.Introspection;
 using GraphQL.Language.AST;
@@ -430,7 +431,17 @@ namespace GraphQL.Execution
                 }
                 else
                 {
-                    result = ((dynamic)task).Result;
+                    var taskType = task.GetType();
+                    if (taskType.GetGenericTypeDefinition() == typeof(Task<>))
+                    {
+                        var wrappedType = taskType.GetGenericArguments()[0];
+                        var method = UnwrapTaskOfTMethod.MakeGenericMethod(wrappedType);
+                        result = method.Invoke(null, new object[] { task });
+                    }
+                    else
+                    {
+                        result = null;
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


csantero:
Not sure what's going on with Appveyor. The tests are passing but some command failed afterward?

csantero:
Never mind, I see the specific appveyor failure now, but it doesn't seem related to this PR. I also ran `yarn test` locally and that passed.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -460,5 +471,12 @@ public static string[] AppendPath(string[] path, string pathSegment)
 
             return newPath;
         }
+
+        private static T UnwrapTaskOfT<T>(Task<T> task)
+        {
+            return task.Result;
```
OneCyrus: .Result can lead to threadpool exhausting. usually the better way is using task.GetAwaiter().GetResult() (it's already this way in the master and probably the root issue of this PR)

csantero: @OneCyrus Thanks, didn't know that. I've updated the PR.

johnrutherford:
I don't see a valid use case for this. In C#, you can't return a private class from a non-private method, so I don't see a valid reason why this would need to be supported with GraphQL.NET.

``` csharp
public class Foo
{
    private class Bar { }

    // This will not compile because of inconsistent accessibility
    public Bar GetBar()
    {
        return new Bar();
    }
}
```

johnrutherford:
Yeah, there's no way for the compiler to tell there that the private class is being exposed. Reflection was being used prior to 870 to get the `Result`. That works because reflection allows you to "break the rules" and access private fields, classes, etc., that you can't otherwise. 

When I made the change to use `dynamic`, I did some benchmarks and it was considerably faster than using reflection. So I don't think it's worth the performance hit to support this usage. 

I agree it should probably be documented somewhere.

joemcbride:
I was going to suggest to do some perf testing on it, but if @johnrutherford has already done so with those previous changes then I'm leaning towards not including this.  Admittedly that performance concern is why this was bumped from the initial 2.0 release.

csantero:
@johnrutherford Got it, thanks for the explanation. I'll go ahead and close this.

PR #624, opened 2018-04-12
Extend check for List<> to include ICollection<> and IList<>

dvdbot:
Extend check for List<T> to include ICollection<T> and IList<T>.
Because these are also valid to be used as long as T is a GraphType

### DIFF
--- a/src/GraphQL/TypeExtensions.cs
+++ b/src/GraphQL/TypeExtensions.cs
@@ -139,7 +139,9 @@ namespace GraphQL
                 graphType = listType.MakeGenericType(elementType);
             }
 
-            if (type.GetTypeInfo().IsGenericType && type.GetGenericTypeDefinition() == typeof(List<>))
+            if (type.GetTypeInfo().IsGenericType && (type.GetGenericTypeDefinition() == typeof(List<>)
+                                                    || type.GetGenericTypeDefinition() == typeof(ICollection<>) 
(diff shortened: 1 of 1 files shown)


BenjaBobs:
I think https://github.com/dvdbot/graphql-dotnet/blob/3fda05fcbe4f480e54f63c35d9043f9c9959deb1/src/GraphQL/TypeExtensions.cs#L135-L149 can be merged like this:

```c#
if (typeof(IEnumerable).IsAssignableFrom(type))
{
    var elementType = type.IsArray ? type.GetElementType() : type.GenericTypeArguments.First();
    var graphElementType = GetGraphTypeFromType(elementType, isNullable);
    var listType = typeof(ListGraphType<>);
    graphType = listType.MakeGenericType(graphElementType);
}
```

BenjaBobs:
Actually I forgot that `string` implements `IEnumerable`, so a more correct version would be
```c#
if (typeof(IEnumerable).IsAssignableFrom(type) && type != typeof(string))
{
    var elementType = type.IsArray ? type.GetElementType() : type.GenericTypeArguments.First();
    var graphElementType = GetGraphTypeFromType(elementType, isNullable);
    var listType = typeof(ListGraphType<>);
    graphType = listType.MakeGenericType(graphElementType);
}
```

joemcbride:
Closing this as an implementation was added in #700

PR #628, opened 2018-04-15
Fix SchemaPrinter bug + optional print options

tlil:
* Fix printing of default input values of enum type
 * Optional printing of descriptions
 * Optional printing of deprecation reasons

(related PRs: graphql-dotnet/conventions#82)

### DIFF
--- a/src/GraphQL/Utilities/SchemaPrinter.cs
+++ b/src/GraphQL/Utilities/SchemaPrinter.cs
@@ -12,9 +12,7 @@ namespace GraphQL.Utilities
     {
         private ISchema _schema;
 
-        private readonly bool _fieldDescriptions;
-
-        private readonly bool _fieldDeprecationReasons;
+        private readonly SchemaPrinterOptions _options;
 
         private readonly List<string> _scalars = new List<string>(
             new[]
@@ -29,8 +27,7 @@ namespace GraphQL.Utilities
         public SchemaPrinter(
             ISchema schema,
             IEnumerable<string> customScalars = null,
-            bool fieldDescriptions = false,
-            bool fieldDeprecationReasons = false)
+            SchemaPrinterOptions options = null)
         {
             _schema = schema;
 
@@ -39,8 +36,7 @@ namespace GraphQL.Utilities
                 _scalars.Fill(customScalars);
             }
 
-            _fieldDescriptions = fieldDescriptions;
-            _fieldDeprecationReasons = fieldDeprecationReasons;
+            _options = options ?? new SchemaPrinterOptions();
         }
 
         private ISchema Schema => _schema;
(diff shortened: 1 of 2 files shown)


joemcbride:
Since that is adding a couple more options, mind switching that to a Options object instead of more optional parameters?

tlil:
Sure thing – can do that.

joemcbride:
Mind adding those custom scalars to the options as well?  Then will merge.

tlil:
Okay so you're happy with me breaking the backwards compatibility for that then? If so, happy to make that change as well, yeah.

joemcbride:
Yep I’m fine with that.

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -322,9 +329,36 @@ public string FormatDefaultValue(object value)
...
+        private static bool IsEnumType(IGraphType type)
+        {
+            if (type is NonNullGraphType)
+            {
+                var nullable = (NonNullGraphType)type;
+                type = nullable.ResolvedType;
+            }
+
+            return type is EnumerationGraphType;
+        }
```
joemcbride: There is a `GraphQLExtensions.GetNamedType` that can do this unwrapping for you (and it includes multiple levels).  I think that would simplify this to:

```
private static bool IsEnumType(IGraphType type)
{
    return type.GetNamedType() is EnumerationGraphType;
}
```

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -322,9 +329,36 @@ public string FormatDefaultValue(object value)
...
+        private static object SerializeEnumValue(IGraphType type, object value)
+        {
+            if (type is NonNullGraphType)
+            {
+                var nullable = (NonNullGraphType)type;
+                type = nullable.ResolvedType;
+            }
+
+            return ((EnumerationGraphType)type).Serialize(value);
+        }
```
joemcbride: Do we not have to worry about printing the `!` for explicitly non-null enum values?  Seems like we should?  Something like the `ResolveName` function in this same file?

joemcbride:
Sorry, just noticed a couple items.

tlil:
Sure thing - will take a look at those. At first glance they both make sense. :+1:

src/GraphQL/Utilities/SchemaPrinter.cs
tlil: Confirmed working - will update.

src/GraphQL/Utilities/SchemaPrinter.cs
tlil: This is for the enum _value_, not the type. Essentially, if you have a resolver: `foo(arg: SomeEnum = X)`, it ensures that `X` is printed as the name of the enum value, i.e., the serialised form of value `X` in type `SomeEnum`.

In other words: If `SomeEnum` is defined for values `FooEnum.Foo` with name = `FOO` and `FooEum.Bar` with name `BAR` (using `FooEnum` here just to make it clear that the underlying linked value of an enum doesn't necessarily have to match with the GraphQL enum type (name of type and values)), this would print `FOO` instead of `Foo` if the value `X` was `FooEnum.Foo`. Make sense?

Issue #573, opened 2018-02-21
DateTimeOffsetTypeTests fails on computer running non US culture


pekkah:
```
Shouldly.ShouldAssertException : "{
  "data": {
    "five": "2018-02-22T09:05:41Z"
  }
}"
    should be
"{
  "data": {
    "five": "2018-02-22T09.05.41Z"
  }
}"
    but was not
    difference
Difference     |                                               |              |                                                     
               |                                              \|/            \|/                                                    
Index          | ...  34   35   36   37   38   39   40   41   42   43   44   45   46   47   48   49   50   51   52   53   54   ...  
Expected Value | ...  0    2    -    2    2    T    0    9    .    0    5    .    4    1    Z    "    \r   \n   \s   \s   }    ...  
Actual Value   | ...  0    2    -    2    2    T    0    9    :    0    5    :    4    1    Z    "    \r   \n   \s   \s   }    ...  
Expected Code  | ...  48   50   45   50   50   84   48   57   46   48   53   46   52   49   90   34   13   10   32   32   125  ...  
Actual Code    | ...  48   50   45   50   50   84   48   57   58   48   53   58   52   49   90   34   13   10   32   32   125  ...  
   at Shouldly.ShouldBeStringTestExtensions.ExecuteAssertion(IAssertion assertion, Func`1 customMessage)
   at Shouldly.ShouldBeStringTestExtensions.ShouldBe(String actual, String expected, Func`1 customMessage, StringCompareShould options)
   at Shouldly.ShouldBeStringTestExtensions.ShouldBe(String actual, String expected, StringCompareShould options)
   at Shouldly.ShouldBeStringTestExtensions.ShouldBe(String actual, String expected)
   at GraphQL.Tests.Utilities.DateTimeOffsetTypeTests.can_use_DateTimeOffset_type() in C:\projects\graphql-dotnet\graphql-dotnet\src\GraphQL.Tests\Utilities\DateTimeOffsetTypeTests.cs:line 35
```

pekkah:
Should be easy to setup for tests with this https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.Tests/CultureTestHelper.cs

PR #642, opened 2018-04-27
(#573) fix test and use culture test helper

pekkah:


joemcbride:
The build script is doing a `npm install -g npm`.  It appears the latest version of `npm` is causing an issue.  So need to update the build script to fix that.

joemcbride:
https://github.com/graphql-dotnet/graphql-dotnet/blob/58ba1ea0b769a3fc88c7e8daad45c1c1515f36cf/appveyor.yml#L14

PR #645, opened 2018-04-30
Order fields alphabetically in introspection types

### LINKED ISSUE #594: Ordering of fields and arguments in introspection
There is currently no ordering of fields and arguments in introspection, they just appear in the order they are defined. This makes it very difficult to find what you are looking for in a large schema.

Is this intended, or are you open to changing this? I don't mind sending in a PR for this.

mvestergaard:
Working with large schemas is currently difficult, because the fields appear in the order they are defined.

This change will order fields on introspection types by name

In the future, perhaps other parts of the introspection would be practical to order by name (types etc). When working with typescript on the client side, the introspection schema is downloaded, and types are generated based on it. If things appear in a predictable order, it reduces the risk of merge conflicts in the generated types when the schema changes.

Closes #594


mvestergaard:
Test failure seems to be due to a flaky test

PR #635, opened 2018-04-19
Adds extension type support for SchemaBuilder

wingsking:

Issue #629, opened 2018-04-18
hook up subscriber


wingsking:
I know that I can create a subscription like this.

```
public Subscription()
        {
            AddField(new EventStreamFieldType
            {
                Name = "commentAdded",
                Type = typeof(CommentType),
                Subscriber = new EventStreamResolver<Comment>(Subscribe)
            });
        }

private IObservable<Comment> Subscribe(ResolveEventStreamContext context)
        {
            return someIObservableObject;
        }
```

But how can I achieve the same if I want to define a subscription with GraphQL schema language?
I understand that for Query and Mutation, we can do something like below, which hooks up resolver with the GraphQLMetadata attribute. Is there a way to do the same for Subscription?

```
public class Droid
{
  public string Id { get; set; }
  public string Name { get; set; }
}

public class Query
{
  [GraphQLMetadata("hero")]
  public Droid GetHero()
  {
    return new Droid { Id = "123", Name = "R2-D2" };
  }
}

var schema = Schema.For(@"
  type Droid {
    id: String
    name: String
  }

  type Query {
    hero: Droid
  }
", _ => {
    _.Types.Include<Query>();
});

var result = schema.Execute(_ =>
{
  _.Query = "{ hero { id name } }";
});

```

wingsking:
Please take a look at the PR when you have a chance. @pekkah @joemcbride

joemcbride:
Resolved in #647

PR #647, opened 2018-05-02
Subscription with reflection

wingsking:
This is related with #629 

1. Added reflection support for subscription
- Added ResolverType attribute to GraphQLMetadataAttribute, ResolverType has values of Resolver and Subscriber, and default value is Resolver so that we don't break any existing logic. If it is set to Subscriber, it indicates that this is a subscription subscribe method. A subscriber method can be async.
- Moved FieldResoverBase.BuildArguments to ReflectionHelper.BuildArguments, since this is used multiple places.
- Deleted FieldResolverBase class, it seems to me that this class only exists because of the BuildArguments method, since the method is moved to ReflectionHelper, I think we don't need this class. Let me know if I miss anything.
2. Added unit tests for subscription with reflection
3. Please note that this pull request includes a previous PR #635, which is not merged yet


joemcbride:
I'll take a look.  I'll mention that I'm not a huge fan of attributes, so I'm hesitant with making attributes required for subscriptions, since that also makes it required for the other resolvers.  Though I honestly can't think of a betters solution yet.  It also seems easy to get wrong since you have to ensure you didn't mis-type the two method names.  Just some initial thoughts.

wingsking:
Hi @joemcbride , did you have a chance to take a look?

wingsking:
Hi @joemcbride, it's been awhile since I sent out this pull request. Do you have any concerns about this? Or if any changes I can make?

joemcbride:
Hey, sorry for going dark on this.  Would you mind rebasing on master so I can give it a proper review?

wingsking:
Thanks for looking into this. You should be good to go now.

PR #658, opened 2018-05-11
Simplify resolver interfaces and context

pekkah:
* AsyncResolver (No more returning Task<T> with null value)
* Make more sense with the inheritance chain of the ResolverFieldContext

### DIFF
--- a/src/GraphQL.Tests/Builders/FieldBuilderTests.cs
+++ b/src/GraphQL.Tests/Builders/FieldBuilderTests.cs
@@ -394,7 +394,7 @@ namespace GraphQL.Tests.Builders
             var field = objectType.Fields.First();
             field.Resolver.Resolve(new ResolveFieldContext
             {
-                Source = 12345
+                SourceObject = 12345
             });
         }
     }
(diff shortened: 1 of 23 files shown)


joemcbride:
I'll take a closer look next week, though my initial response is I'm not sure I'm comfortable with these changes.  If nothing else, due to the hard break in compatibility (this could have a very large impact).

Do you have a specific use case you are trying to solve?

pekkah:
I'm trying to simplify interfaces and the context for resolving fields so I can actually add to them features related to two other issues I opened earlier. Now the resolving just has too much magic related to the type of the source object, using new keyword in the interfaces, weird inheritance chain in the context that any changes to them causes too many complications.

joemcbride:
Still feeling uncomfortable with these changes.  Mainly with the `Source` -> `SourceObject` change and the addition of the async resolver.  I think I would prefer to break people and make returning `Task` required vs. having two different resolvers that the library has to guess which one to use.

pekkah:
Ok no problem. I'll see if I can clean it up without breaking anything.

BenjaBobs:
This pr was closed a long time ago, but I think it has some good ideas that should be considered again. In particular, a less complex inheritance chain.
Something like:
```c#
public class ResolveFieldContext
{
	//...
	public object Source { get; set; }
}

public class ResolveFieldContext<TSource> : ResolveFieldContext
{
	public new TSource Source { get => (TSource)base.Source; set => base.Source = value; }
}
```

This would increase flexibility in cases where the exact type isn't known / necessary.

pekkah:
#1440

pekkah:
#658

pekkah:
#1437

sungam3r:
@pekkah Do you mind deleting a branch?

Issue #662, opened 2018-05-17
Add support for Date, DateTime and DateTimeOffset


RehanSaeed:
The current `DateGraphType` expects a `DateTime` or `DateTimeOffset`. It's limitations are that it expects only UTC times (which you should usually be using anyway but not always, sometimes you might want to know what time something occurred with a particular offset) and when used with `DateTimeOffset` it ignores the offset.

I propose splitting the `DateGraphType` into three types:

- `DateGraphType` - Date with no time: yy-MM-dd.
- `DateTimeGraphType` - Scalar type for `System.DateTime`. Added to `GraphTypeRegistry`. Format: `2018-05-17T12:11:06.3684072Z`.
- `DateTimeOffsetGraphType` - Scalar type for `System.DateTimeOffset`. Added to `GraphTypeRegistry`: `2018-05-17T13:11:06.368408+01:00`.

Here is some example usage code:

```
public class Dates
{
    public DateTimeOffset OffsetUtc { get; set; } = DateTimeOffset.UtcNow;
    public DateTimeOffset OffsetNow { get; set; } = DateTimeOffset.Now;
    public DateTime Utc { get; set; } = DateTime.UtcNow;
    public DateTime Now { get; set; } = DateTime.Now;
}

public class DatesGraphType : ObjectGraphType<Dates>
{
    public DroidObject(IDroidRepository droidRepository)
    {
        this.Name = "Dates";

        this.Field(x => x.Utc);
        this.Field(x => x.Now);
        this.Field(x => x.OffsetUtc);
        this.Field(x => x.OffsetNow);
    }
}
```

This produces the following:

```
"utc": "2018-05-17T12:26:04.3746839Z",
"now": "2018-05-17T12:26:04.3746839Z",
"offsetUtc": "2018-05-17T12:26:04.374678+00:00",
"offsetNow": "2018-05-17T13:26:04.3746787+01:00"
```

This requires a change to the `DocumentWriter` constructor to use the default `DateFormatString`:

```
public DocumentWriter(bool indent)
            : this(
                indent ? Formatting.Indented : Formatting.None,
                new JsonSerializerSettings
                {
                    DateFormatHandling = DateFormatHandling.IsoDateFormat,
                    // DateFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.FFFFFFF'Z'",
                })
        {
        }
```

Would you like a PR?

joemcbride:
[...]

PR #665, opened 2018-05-22
Added Date, DateTime, DateTimeOffset and TimeSpan Scalar Types

RehanSaeed:
Fixes https://github.com/graphql-dotnet/graphql-dotnet/issues/662

This PR is a work in progress but I'd like to get feedback. I'm still getting my head around the code base. Once we've bottomed out a plan, I can add some tests and fix the ones I've broken.


SapphireBrand:
I am missing one detail.

If a query root has an argument of type "Date!" then that will serialize from the client to the server in ISO 8601, which possibly has time zone information.

The current behavior of GraphQL.Net is to map that argument to DateGraphType, which does not have time zone information. For instance, I get an error message from GraphQL.Validation.Rules.VariablesInAllowedPosition:

"Variable \"$from\" of type \"Date!\" used in position expecting type \"DateTimeOffset\"."

I must have to register that Date maps to DateTimeOffset somewhere. How do I do that?

(I am not requesting that this pull request map GraphQL's Date type onto C#'s DateTimeOffset, because that would be a breaking change, but it must be possible to do so or else the value of DateTimeOffsetGraphType would be limited.)

Thx

joemcbride:
@SapphireBrand You probably need to change this registration:

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/GraphTypeRegistry.cs#L24

Can you open a different issue and post more specifics of your setup if that does not fix it?

src/GraphQL/Types/TimeSpanMillisecondsGraphType.cs
```diff
@@ -0,0 +1,47 @@
+using System;
+using GraphQL.Language.AST;
+
+namespace GraphQL.Types
+{
+    public class TimeSpanMillisecondsGraphType : ScalarGraphType
+    {
+        public TimeSpanMillisecondsGraphType()
+        {
+            Name = "TimeSpan";
```
joemcbride: These `TimeSpan` types should have unique names.  Otherwise you won't be able to use both in the same schema (one would override the other).

RehanSaeed: Yes of course, I missed that. I've now called the types `Seconds` and `Milliseconds` unless you have a better name. Perhaps `TimeInSeconds` and `TimeInMilliseconds`?

If there are no more suggestions, should I start fixing and writing new tests?

src/GraphQL/Types/TimeSpanSecondsGraphType.cs
```diff
@@ -0,0 +1,42 @@
...
+        public TimeSpanSecondsGraphType()
+        {
+            Name = "Seconds";
+            Description =
+                "The `Seconds` scalar type represents a period of time represented as the total number of seconds.";
+        }
+
+        public override object Serialize(object value)
+        {
+            return ParseValue(value);
```
joemcbride: I think this need to be changed to similar to the `Milliseconds` type?  Checking for `TimeSpan` and then choosing `TimeSpan.TotalSeconds`?

src/GraphQL/Types/TimeSpanMillisecondsGraphType.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (value is TimeSpanValue)
+            {
+                return ParseValue(((TimeSpanValue)value).Value);
+            }
+
+            if (value is LongValue)
+            {
+                return ParseValue(((LongValue)value).Value);
+            }
+
```
rianjs: - All `int`s are also `long`s. How will the `IntValue` case ever be reached? Can you get rid of `int` and just use `long`?
- In .NET, `new TimeSpan(long)` corresponds to the `Ticks` property, NOT seconds or milliseconds.

rianjs:
In `DateTimeToDateTimeOffset`, you have the UTC offset hard-coded to `TimeSpan.Zero` which [does not follow the .NET convention for `DateTimeOffset` types](https://msdn.microsoft.com/en-us/library/system.datetimeoffset.op_implicit(v=vs.110).aspx).

The rules for converting `DateTime` to `DateTimeOffset` are:
* `DateTimeKind.Local` or `DateTimeKind.Unspecified` = use the local machine's time zone to figure out the UTC offset that should apply to that DateTime.
* `DateTimeKind.Utc` = use `TimeSpan.Zero`.

I would suggest following the convention, otherwise you're going to confuse a lot of .NET devs who are going to have to wrap the library to handle its non-standard behavior. In fact, you changed the `Description` property on both `Date` and `DateTime` to say:

> "The `DateTime` scalar type represents a date and time. `DateTime` expects timestamps to be formatted in accordance with the [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) standard.

...which the library itself then ignores by explicitly saying that value is UTC. ISO-8601 requires a `Z` suffix to mean UTC. Everything else is "local time".

I have similar objections to `DateTimeOffsetToDateTime` which always returns a UTC representation. I would get rid of that entirely, and only return a `DateTimeOffset` when applicable, and let the application developer decide whether they want:
- `DateTimeOffset.LocalDateTime` (that moment converted to the machine's local time)
- `DateTimeOffset.DateTime` (truncate the UTC offset)
- `DateTimeOffset.UtcDateTime` (that moment expressed as a `DateTime` with `DateTimeKind.Utc`)

...these are all subtly different, and all of them can be useful given the right circumstances.

SapphireBrand:
I agree with @rianjs regarding the use of TimeSpan.Zero in the DateTime to DateTimeOffset conversion. I wrote my own code for DateTimeOffsetGraphType, intending to use this pull request when it is merged, but now I realize that I would have to keep my own code because this implementation would not convert DateTime according to the ISO spec.

RehanSaeed:
> @rianjs In DateTimeToDateTimeOffset, you have the UTC offset hard-coded to TimeSpan.Zero which does not follow the .NET convention for DateTimeOffset types.

Agreed, I've changed it to be a case `(DateTimeOffset)dateTime` which should handle things properly.

On `DateTimeOffsetToDateTime`, I'm not wedded to it but my thinking was to provide some sensible default and using UTC, seems like a sensible default.

On to fixing and writing some tests...

RehanSaeed:
Tests are green, build is green. PR is ready to merge, unless there are further comments.

PR #673, opened 2018-05-26
Add example of how you can do it without DI

joelmandell:
I tried to make an example of how you can use DataLoader without DI. What do you think?

### DIFF
--- a/docs/src/dataloader.md
+++ b/docs/src/dataloader.md
@@ -28,8 +28,7 @@ A DataLoader helps in two ways:
 
 In the example above, a using a DataLoader will allow us to batch together all of the requests for the users. So there would be 1 request to retrieve the list of orders and 1 request to load all users associated with those orders. This would always be a total of 2 requests rather than N+1.
 
-## Setup
-
+## Setup using Dependency Injection
 1. Register `IDataLoaderContextAccessor` in your IoC container.
 2. Register `DataLoaderDocumentListener` in your IoC container.
 
@@ -52,7 +51,28 @@ var result = executer.ExecuteAsync(opts => {
 });
 ```
 
-## Usage
+## Example of Setup *without* Dependency injection
+
+1. Instantiate a `DataLoaderContextAccessor` and a `DataLoaderContext` and `DataLoaderDocumentListener`
+in the GraphQL controller. Perhaps like this:
+``` chsarp
+    var ctxAc = new DataLoader.DataLoaderContextAccessor();
+    ctxAc.Context = new DataLoaderContext();
+    var listener = new DataLoader.DataLoaderDocumentListener(ctxAc);
+
+    await new DocumentExecuter().ExecuteAsync( (s) =>
+        s.Schema = new Schema() { Query = new RootQuery(), Mutation = new RootMutation() };
+        s.Query = q.query;
+        s.Listeners.Add(listener);
+        s.Root = ctxAc;
+        if ((q.variables != null))
+            s.Inputs = JsonConvert.SerializeObject(q.variables).ToInputs;
+    ).ConfigureAwait(False)
+
+```
+__Further down you will find usage examples.__
+
+## Usage with Dependency Injection
 
 First, inject the `IDataLoaderContextAccessor` into your GraphQL type class.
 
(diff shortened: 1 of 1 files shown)


joemcbride:
I would suggest to users to put this on the UserContext vs the Root object. The Root object has other implications, whereas the UserContext is meant exactly for these kind of things.

docs/src/dataloader.md
```diff
@@ -52,7 +51,28 @@ var result = executer.ExecuteAsync(opts => {
 });
 ```
 
-## Usage
+## Example of Setup *without* Dependency injection
+
+1. Instantiate a `DataLoaderContextAccessor` and a `DataLoaderContext` and `DataLoaderDocumentListener`
+in the GraphQL controller. Perhaps like this:
+``` chsarp
+    var ctxAc = new DataLoader.DataLoaderContextAccessor();
```
joemcbride: In general I would also prefer the docs to use descriptive names vs non descriptive.  `ctxAc` is non descriptive.

docs/src/dataloader.md
```diff
@@ -168,3 +188,50 @@ public interface IUsersStore
 	Task<IEnumerable<User>> GetAllUsersAsync();
 }
 ```
+## Usage *without* Dependency Injection
+Because the DataLoaderContext is defined in the GraphQL Root object we can get it in our resolvers like this:
+```csharp
+public class ordertype : ObjectGraphType<order>
```
joemcbride: This should be uppercase `O`.

docs/src/dataloader.md
```diff
@@ -168,3 +188,50 @@ public interface IUsersStore
...
+
+        Field<IntGraphType>("id");
+        Field<StringGraphType>("userId");
+        Field<IntGraphType>("category");
+        Field<StringGraphType>("name");
+        Field<DateGraphType>("date");
+        Field<Usertype, User>.Name("user").ResolveAsync(ctx =>
+        {
+            store.SetResolveContext(ctx); //Make ctx availabele
+            var loader = ctx.RootValue.Context as DataLoaderContext.GetOrAddBatchLoader<int, User>("GetUserIds", store.GetUserIds);
```
joemcbride: This looks like invalid code, missing parentheses around the cast?  Is that a C#7 feature I’m not aware of?  Perhaps split to two lines to make it more readable?

docs/src/dataloader.md
```diff
@@ -168,3 +188,50 @@ public interface IUsersStore
...
+    }
+}
+
+```
+And we can define the UserStore like this:
+```csharp
+public class UserStore
+{
+    private object ctx;
+    public void SetResolveContext(object c)
```
joemcbride: Can remove this from the sample since it is not required.

docs/src/dataloader.md
```diff
@@ -168,3 +188,50 @@ public interface IUsersStore
...
+    {
+        ctx = c;
+    }
+    public async Task<Dictionary<int, User>> GetObjektByIdAsync(IEnumerable<int> userIds)
```
joemcbride: Should this method name match the `GetUserIds` used above?

docs/src/dataloader.md
```diff
@@ -168,3 +188,50 @@ public interface IUsersStore
...
+    {
+        var users = db.Users.. //make db call
+        Dictionary<int, User> dic = new Dictionary<int, User>();
+        
+        foreach (var i in users)
+            dic.Add(userIds(users.IndexOf(i)), i);
```
joemcbride: This code isn’t making much sense to me. I’m guessing this is a mix of C# and VB?  Can you update this to make a bit more sense?

docs/src/dataloader.md
```diff
@@ -168,3 +188,50 @@ public interface IUsersStore
...
+
+            return loader.LoadAsync(ctx.Source.userId);
+        });
+    }
+}
+
+```
+And we can define the UserStore like this:
+```csharp
+public class UserStore
```
joemcbride: The sample above uses a IUserStore interface.  I think it would help the reader to keep with that interface definition vs create an entirely new sample with basically the same name.  For example, show the class implementing that interface to keep continuity. I think the implementation is somewhat irrelevant to the sample.

joelmandell: Yes I agree that it creates inconcistency. But a question... did I just miss something or how can something happen in the interface? The interface is basically only a skeleton of what needs to be implemented? But there is no function?

docs/src/dataloader.md
joelmandell: That possibility has existed for some time ([https://stackoverflow.com/questions/4926677/c-sharp-as-cast-vs-classic-cast](url))
I will make it more readable :)

docs/src/dataloader.md
joelmandell: Yes I wil remove it.

docs/src/dataloader.md
joelmandell: Yep typo when copied from my own code in another project.

docs/src/dataloader.md
```diff
@@ -52,7 +51,28 @@ var result = executer.ExecuteAsync(opts => {
...
+    var listener = new DataLoader.DataLoaderDocumentListener(ctxAc);
+
+    await new DocumentExecuter().ExecuteAsync( (s) =>
+        s.Schema = new Schema() { Query = new RootQuery(), Mutation = new RootMutation() };
+        s.Query = q.query;
+        s.Listeners.Add(listener);
+        s.Root = ctxAc;
+        if ((q.variables != null))
+            s.Inputs = JsonConvert.SerializeObject(q.variables).ToInputs;
+    ).ConfigureAwait(False)
```
johnrutherford: This needs a semicolon at the end and false must be lowercase.

docs/src/dataloader.md
```diff
@@ -168,3 +188,50 @@ public interface IUsersStore
...
+        UserStore store = new UserStore();
+
+        Field<IntGraphType>("id");
+        Field<StringGraphType>("userId");
+        Field<IntGraphType>("category");
+        Field<StringGraphType>("name");
+        Field<DateGraphType>("date");
+        Field<Usertype, User>.Name("user").ResolveAsync(ctx =>
+        {
+            store.SetResolveContext(ctx); //Make ctx availabele
```
johnrutherford: I don’t understand why you would need to set the context on the store. Your store shouldn’t need to know anything about GraphQL.

docs/src/dataloader.md
johnrutherford: You do need parentheses around the “as” statement.

joemcbride:
Closing this as there are other updated docs for DI and there has been no movement on this for a while. Feel free to re-open with any suggested changes to the new docs.

https://graphql-dotnet.github.io/docs/getting-started/dependency-injection

PR #677, opened 2018-05-28
First Should Return DefaultPageSize

RehanSaeed:
If the `first` and `last` properties are not set, `first` should default to the `defaultPageSize`. Right now, I have to write the following code for every `Connection`:

```
var first = context.First;
if (!context.First.HasValue && !context.Last.HasValue)
{
    first = context.PageSize;
}
```

### DIFF
--- a/src/GraphQL/Builders/ResolveConnectionContext.cs
+++ b/src/GraphQL/Builders/ResolveConnectionContext.cs
@@ -26,7 +26,7 @@ namespace GraphQL.Builders
                     return _defaultPageSize;
                 }
                 
-                return first
+                return first;
             }
         }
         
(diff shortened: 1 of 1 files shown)


src/GraphQL/Builders/ResolveConnectionContext.cs
```diff
@@ -17,6 +17,20 @@ public ResolveConnectionContext(ResolveFieldContext context, bool isUnidirection
...
+        {
+            get
+            {
+                var first = FirstInternal;
+                if (!first.HasValue && !Last.HasValue)
+                {
+                    return _defaultPageSize;
+                }
+                
+                return first
```
joemcbride: Missing a semicolon here.

RehanSaeed:
Fixed the silly typo.

PR #678, opened 2018-05-28
Cursor Helper Class

RehanSaeed:
This cursor helper class can be used to convert from values of any type to base64 encoded cursor strings. It can also be used to return the first and last cursor given a collection (or null if the collection is null or empty).

### DIFF
--- a/src/GraphQL.Tests/Types/Relay/CursorTests.cs
+++ b/src/GraphQL.Tests/Types/Relay/CursorTests.cs
@@ -1,0 +1,169 @@
+using System;
+using System.Collections.Generic;
+using GraphQL.Types.Relay;
+using Xunit;
+
+namespace GraphQL.Tests.Types.Relay
+{
+    public class CursorTests
+    {
+        [Fact]
+        public void GetFirstAndLastCursor_NullCollection_ReturnsNull()
+        {
+            List<Entity> items = null;
+
+            var (firstCursor, lastCursor) = Cursor.GetFirstAndLastCursor(items, x => x.Property);
+
+            Assert.Null(firstCursor);
+            Assert.Null(lastCursor);
+        }
+
+        [Fact]
+        public void GetFirstAndLastCursor_EmptyCollection_ReturnsNull()
+        {
+            List<Entity> items = new List<Entity>();
+
+            var (firstCursor, lastCursor) = Cursor.GetFirstAndLastCursor(items, x => x.Property);
+
+            Assert.Null(firstCursor);
+            Assert.Null(lastCursor);
+        }
+
+        [Fact]
+        public void GetFirstAndLastCursor_NullGetpropertyFunc_ThrowsArgumentNullException()
+        {
+            List<Entity> items = new List<Entity>();
+
+            Assert.Throws<ArgumentNullException>(() => Cursor.GetFirstAndLastCursor<Entity, int>(items, null));
+        }
+
+        [Fact]
+        public void GetFirstAndLastCursor_ThreeItems_ReturnsTheCursorForTheFirstAndLastItems()
+        {
+            List<Entity> items = new List<Entity>()
+            {
+                new Entity() { Property = 1 },
+                new Entity() { Property = 2 },
+                new Entity() { Property = 3 },
+            };
+
+            var (firstCursor, lastCursor) = Cursor.GetFirstAndLastCursor(items, x => x.Property);
+
+            Assert.Equal("YXJyYXljb25uZWN0aW9uOjE=", firstCursor);
+            Assert.Equal("YXJyYXljb25uZWN0aW9uOjM=", lastCursor);
+        }
+
+        [Fact]
+        public void FromNullableCursor_NullValue_ReturnsNull()
+        {
+            var cursor = (string)null;
+
+            var value = Cursor.FromNullableCursor<int>(cursor);
+
+            Assert.Null(value);
+        }
+
+        [Fact]
+        public void FromNullableCursor_IntValue_ReturnsValueFromCursor()
+        {
+            var cursor = "YXJyYXljb25uZWN0aW9uOjU=";
+
+            var value = Cursor.FromNullableCursor<int>(cursor);
+
+            Assert.Equal(5, value);
+        }
+
+        [Fact]
+        public void FromNullableCursor_DateTimeValue_ReturnsValueFromCursor()
+        {
+            var cursor = "YXJyYXljb25uZWN0aW9uOjAxLzAxLzIwMDAgMDA6MDA6MDA=";
+
+            var value = Cursor.FromNullableCursor<DateTime>(cursor);
+
+            Assert.Equal(new DateTime(2000, 1, 1), value);
+        }
+
+        [Fact]
+        public void FromCursor_NullValue_ReturnsNull()
+        {
+            var cursor = (string)null;
+
+            var value = Cursor.FromCursor<string>(cursor);
+
+            Assert.Null(value);
+        }
+
+        [Fact]
+        public void FromCursor_IntValue_ReturnsValueFromCursor()
+        {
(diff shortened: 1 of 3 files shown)


src/GraphQL.Tests/Types/Relay/CursorTests.cs
```diff
@@ -0,0 +1,169 @@
...
+    {
+        [Fact]
+        public void GetFirstAndLastCursor_NullCollection_ReturnsNull()
+        {
+            List<Entity> items = null;
+
+            var (firstCursor, lastCursor) = Cursor.GetFirstAndLastCursor(items, x => x.Property);
+
+            Assert.Null(firstCursor);
+            Assert.Null(lastCursor);
```
joemcbride: This project uses Shoudly for assertions. Can you please update these to use that?

src/GraphQL.Tests/Types/Relay/CursorTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        [Fact]
+        public void ToCursor_DateTimeValue_ReturnsBase64Cursor()
+        {
+            var value = new DateTime(2000, 1, 1);
+
+            var cursor = Cursor.ToCursor(value);
+
+            Assert.Equal("YXJyYXljb25uZWN0aW9uOjAxLzAxLzIwMDAgMDA6MDA6MDA=", cursor);
```
joemcbride: This is failing on CI.

src/GraphQL.Tests/Types/Relay/CursorTests.cs
RehanSaeed: Switched to Shoudly as requested.

src/GraphQL.Tests/Types/Relay/CursorTests.cs
RehanSaeed: Now using `CultureInfo.InvariantCulture` to fix dates being converted to and from `string` producing different results in different parts of the world.

src/GraphQL/GraphQL.csproj
```diff
@@ -28,6 +28,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Include="System.Reactive.Core" Version="3.1.1" />
     <PackageReference Include="System.Reactive.Linq" Version="3.1.1" />
+    <PackageReference Include="System.ValueTuple" Version="4.4.0" />
```
joemcbride: Just noticed this addition.  I'm not seeing `ValueTuple` used anywhere (maybe I'm missing some other usage).  Can this be removed?

jquense:
this is sort of unfortunate, GraphQL Relay already adds most of this with a bit more robustness. We should perhaps consider more seriously moving all the connection stuff out of GraphQL.dotnet?

src/GraphQL/GraphQL.csproj
RehanSaeed: This allows the use C# tuples `public static (string firstCursor, string lastCursor) GetFirstAndLastCursor<TItem, TCursor>()`

RehanSaeed:
@jquense Moving makes sense to me. I saw [this](https://github.com/graphql-dotnet/relay/blob/master/src/GraphQL.Relay/Types/ConnectionUtils.cs) code in the relay project related to cursors but it didn't quite meet all my needs, so I wrote a `Cursor` static helper class.

src/GraphQL/GraphQL.csproj
RehanSaeed: The alternative if you want to avoid the dependency is to add a custom struct with these two properties.

jquense:
@RehanSaeed i'd happily take any PRs on that front, also always looking for help maintaining if that interests you 👍

RehanSaeed:
@jquense So you want this same PR, submitted to the Relay project? Thanks for the invite, much appreciated. I'm done learning GraphQL but am still getting used to this code base and have little knowledge of Relay (I can't find a good way to learn it without also learning React). I'm only interested in the connection type and other common sense types right now but that might change. I've heard you can use Relay with VueJS, is there also a .NET client? Relay sounds very useful, being closed to React is the only blocker.

jquense:
The relay dotnet project is mostly an implementation of the server bits to the relay graphql spec, which is what actually defines a Connection (it's not a core graphql concept). Generally that is paired with the relay client, but it's not required, I often use the relay server stuff with non relay clients (like apollo) since the pagination concepts are usually still helpful

src/GraphQL/Types/Relay/Cursor.cs
```diff
@@ -0,0 +1,70 @@
+namespace GraphQL.Types.Relay
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+    using System.Text;
+
+    public static class Cursor
```
jquense: This is a bit misleading, this is really an `IndexCursor`. While it's definitely helpful in simple cases, generally you probably don't actually want to use index cursors for serious implementations, as it doesn't account for ordering and can't be derived from just an item, which makes it inefficient if you are backing GQL with an ORM.

That is not say it's bad, but i worry that providing it as a default under the generic name "Cursor" suggests to much for it's use.

RehanSaeed: Thats interesting, what is the more efficient alternative? Storing the cursor in the database, so it doesn't have to be recalculated? In my head I was thinking you could use an `orderby` field that determines which property on a model to use as the cursor and which property on the model to run an `ORDER BY` on if you are using SQL.

On a related efficiency note, I asked the question ["Does GraphQL negate the need for Graph Databases"](https://stackoverflow.com/questions/50134500/does-graphql-negate-the-need-for-graph-databases) on StackOverflow. For a personal project, I was thinking of using Microsoft Orleans because it's cheap to run but CosmosDB or some other graph database might also be interesting from an efficiency standpoint. Neo4j even has some GraphQL integration.

jquense: > what is the more efficient alternative?

Building the cursor from an unique id, is the most straightforward, and doesn't create collisions. _Ideally_ any cursor should be derivable from just a single item, so you don't have to query the DB for _where_ that item is  total. Think about a `create` mutation, often you want to return the created node, as well as an `Edge<T>` for that node, in the mutation payload. It's generally not easy to efficiently insert the new item and also get its index. If you can derive it from the item tho, you don't have to do any additional lookups. The other thing is you don't collisions when fetching connections for multiple sort or filter situations.

RehanSaeed: @jquense Take a look at my [GraphQL.NET project template](https://github.com/Dotnet-Boxed/Templates) or specifically [this](https://github.com/Dotnet-Boxed/Templates/blob/8e7c051b121237564189ec2a90763842000e7a38/Source/Content/GraphQLTemplate/Schemas/QueryObject.cs#L79) relay connection for droids I'm trying to add. It talks to a [droid repository](https://github.com/Dotnet-Boxed/Templates/blob/master/Source/Content/GraphQLTemplate/Repositories/DroidRepository.cs).

The connection is ordering droids by their created date (date of birth). If you use the ID for the cursor as opposed to the created date property, this makes it more difficult to get the data from the database. Currently in the droid repository I call:

```
public Task<List<Droid>> GetDroids(
    int? first,
    DateTime? createdAfter,
    CancellationToken cancellationToken) =>
    Task.FromResult(Database
        .Droids
        .If(createdAfter.HasValue, x => x.Where(y => y.Created > createdAfter.Value))
        .If(first.HasValue, x => x.Take(first.Value))
        .ToList());
```

If the droid created date is the cursor, I can simply pass that into the query above. If the ID is the cursor, I'm not sure what the most efficient way to write that query would be (assuming we are using EF Core or some other SQL/Linq based data store). Perhaps I would have to get the created date of the droid with the given ID and then use that in the query above. Am I missing something?

PR #681, opened 2018-05-29
Allow before and after to be used together and with first or last

RehanSaeed:
The `ConnectionBuilder` was throwing exceptions the use of `before` when `first` or `after` were used. It also threw exceptions when `after` was used with `last` or `before`.

I can't see these restrictions in the [GraphQL Relay spec](https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo) but I could be wrong.

Using both `after` and `before` is used by Twitter when they add a `load more tweets` button:

1. Load some tweets.
2. User becomes inactive for a long period, not sure of the time Twitter uses.
3. User is active, load the latest tweets, then show a `load more tweets` button.
4. Clicking on `load more tweets` would load tweets `after` the latest tweets but `before` the older set.

### DIFF
--- a/src/GraphQL/Builders/ConnectionBuilder.cs
+++ b/src/GraphQL/Builders/ConnectionBuilder.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using GraphQL.Types;
 using GraphQL.Types.Relay;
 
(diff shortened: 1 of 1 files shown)


jquense:
its not strictly disallowed, but it is strongly discouraged: https://facebook.github.io/relay/graphql/connections.htm#sec-Pagination-algorithm and to some degree doesn't make a lot of sense to do, since pagination _can't_ be in both directions at once, at least not in a way that makes much sense. That said it is not prohibited so...

jquense:
actually the note is more about `first` and `last` together not either and the paired cursor. This seems like a good change 👍

RehanSaeed:
Was just about to reply, you beat me to it.

joemcbride:
@RehanSaeed My merging of your other PR created conflicts it appears.  I can merge after those conflicts are resolved.

PR #698, opened 2018-06-16
IsTypeOfFunc may be specified through GraphQLMetadata

eugenpaval:
When schema language is employed to generate GraphQL runtime wiring classes they don't contain an implementation of IsTypeOfFunc needed for types implementing interfaces. The resolving classes may then be decorated with GraphQLMetadatattribute containing a new property IsTypeOf which allows specifying what type will be used to implement the iinterface.

Fixed a minor error in TypeExtensions.GetGraphQLName() when the GraphQLmetadata attribute is being used without specifying a name

### DIFF
--- a/src/GraphQL.Tests/Utilities/CustomGraphQLAttributeTests.cs
+++ b/src/GraphQL.Tests/Utilities/CustomGraphQLAttributeTests.cs
@@ -1,4 +1,4 @@
-﻿using System.Linq;
+using System.Linq;
 using GraphQL.Types;
 using GraphQL.Utilities;
 using Shouldly;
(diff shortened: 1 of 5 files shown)


src/GraphQL/Execution/ExecutionError.cs
```diff
@@ -32,9 +32,7 @@ public ExecutionError(string message, Exception exception)
 
         public IEnumerable<string> Path { get; set; }
 
-        public Dictionary<string, object> DataAsDictionary { get; } = new Dictionary<string, object>();
-
-        public override IDictionary Data => DataAsDictionary;
+        public new Dictionary<string, object> Data { get; private set; } = null;
```
joemcbride: Need to keep these `DataAsDictionary` changes.

src/GraphQL.Tests/Utilities/CustomGraphQLAttributeTests.cs
```diff
@@ -56,5 +88,30 @@ public Post GetPostById(string id)
...
+        }
+    }
+
+    [GraphQLMetadata("IUniqueElement")]
+    public abstract class UniqueElement
+    {
+        public abstract string Id { get; }
+    }
+
+    [GraphQLMetadata(IsTypeOf = typeof(ABlog))]
```
joemcbride: Looking at this it does seem a bit redundant.  Perhaps we should provide a default implementation in the `TypeConfig`, which you can override with this `IsTypeOf` property you added to the metadata?  `TypeConfig` contains the `Type`, in this case `ABlog`, already.

eugenpaval: You're right. Let me see how I can provide the default implementation...

eugenpaval:
All requested changes provided. If the resolving type name matches GraphQL type name there is no need to add GraphQLMetadata attribute to the resolving type just to provide an implementation of IsTypeFuncOf for the GraphQL schema type.

Issue #696, opened 2018-06-15
Error - Query is too complex to validate.


dharmeshtailor:
I am getting following error, when query is too complex. But then we keep getting error for every subsequent query, irrespective of the complexity of the query. error goes off once I restart node on Azure Service Fabric.
 
![image](https://user-images.githubusercontent.com/5554788/41445250-a00db71e-708b-11e8-8e77-b6fae3e09d71.png)

There are 2 methods in class "**GraphQL.Validation.Complexity.ComplexityAnalyzer**", namely "**FragmentIterator**" and "**TreeIterator**" that has following code.

![image](https://user-images.githubusercontent.com/5554788/41445484-42b5d612-708d-11e8-99c5-5054c3e412f0.png)

I assume, **_loopCounter** should be reset to 0 before throwing error. It is done when everything works alright, but not in case if validation fails.

GraphQL version used :  2.0.0-alpha-924

joemcbride:
Thanks for the report - mind submitting a PR with a fix?

dharmeshtailor:
Sure will do it over the weekend.

Thanks.

dharmeshtailor:
Thanks, Sorry was busy over the weekend. So didn't get chance to submit the PR.

Closing the issue.

PR #699, opened 2018-06-18
#696 reset counter

MNie:
Related to #696 .

CC: @joemcbride


joemcbride:
Deployed in https://www.nuget.org/packages/GraphQL/2.0.0-alpha-927

PR #700, opened 2018-06-18
Add support to return IEnumerable as a ListGraphType

MNie:
Right now as a `ListGraphType` could be returned only a `List`. I want to be able to return `ICollection`, `IList`, `IEnumerable`, `IReadOnlyCollection` etc. as a `ListGraphType` I add some unit tests so we would be sure that everything works fine.

CC: @joemcbride


joemcbride:
Supersedes #624

joemcbride:
Deployed in https://www.nuget.org/packages/GraphQL/2.0.0-alpha-927

PR #701, opened 2018-06-19
#606 boolean graph type not converting numerical string

KoditkarVedant:
- [x] Add ability to convert `"1"` to `"0"` to `true` and `false` respectively
- [x] Change `Serialize` for `BooleanGraphType` to call `ParseValue`
- [x] Added Test cases for changes

PR #703, opened 2018-06-22
Support "1" and 0" in BooleanGraphType.

### LINKED ISSUE #606: BooleanGraphType not converting "1"
In a mutation i use a type for the arguments which has a bool property. If the client sends the argument as string "1" it gets converted to false.

```
    class MyMutation
    {
        public bool Value { get; set; }
    }
          
          //in mutation
          resolve: context =>
                {
                    var mutation = context.GetArgument<MyMutation>("request");
                }
```

I am not sure, if this is really intended or a bug. If i look at the code, the BooleanGraphType supports "1" in the ParseValue method. But it seems the result from the Serialize method gets used, which is always false when no bool is provided.

```
//BooleanGraphType
public override object Serialize(object value)
    {
      if (value is bool)
        return (object) (bool) value;
      return (object) false;
    }
```

When I look at other ScalarTypes then usually ParseValue is called in the Serialize method, which would work in my case, when the BooleanGraphType would do the same.

```
public override object Serialize(object value)
    {
      return this.ParseValue(value);
    }
```

I always wondered what is the purpose of these two methods and which method should do what excactly.

I am currently in version 0.17.

MNie:
Fixes #606 

CC: @joemcbride 

ps. I don't know if changes in ValueConverter.cs are correct(?) maybe we don't want to change here anything? If you have some suggestions I will be grateful :)

### DIFF
--- a/src/GraphQL.Tests/Types/BooleanGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/BooleanGraphTypeTests.cs
@@ -1,9 +1,11 @@
-﻿using GraphQL.Types;
+using GraphQL.Types;
 using Shouldly;
 using Xunit;
 
 namespace GraphQL.Tests.Types
 {
+    using System;
+
     public class BooleanGraphTypeTests
     {
         private BooleanGraphType type = new BooleanGraphType();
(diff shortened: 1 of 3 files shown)


joemcbride:
Thanks for the PR! Ending up having two PRs for the same thing. Going with #701 for now.

MNie:
@joemcbride I don't saw the another one :)

joemcbride:
Closing this as the other PR has now been merged.

Issue #472, opened 2017-11-11
Add UriGraphType


deinok:
Add an `GraphQL.Types.UriGraphType` to represent `Uri`.

joemcbride:
Can you talk to what a Uri specific GraphType would provide that the `StringGraphType` does not?

deinok:
The same as `DateTimeGraph`

A string with a special format

joemcbride:
Can you provide a concrete example?

deinok:
An URI, "must" have the format of this spec -> https://tools.ietf.org/html/rfc3986
Anything that cant be parsed with this rules is not a valid URI.

Imagine I have a Schema with field called "uri" of type URI.
If I try to insert a valid URI everything goes okey, if its invalid, an error is returned to the caller, cause I dont want invalid URIs in my DB.

If "uri" is of type String, by default, the two scenarios above goes okey and the DB have an unexpected value.

(Yes, we could check if a "String" is an URI in our own code, but better share this "check" with other interested people)

Also, the fact that somthing is of type URI, gives to the client more information about the schema.

PR #706, opened 2018-06-24
#472 add UriGraphType

MNie:
Add support to urigraphtype
Fixes: #472

CC: @joemcbride

### DIFF
--- a/src/GraphQL/ValueConverter.cs
+++ b/src/GraphQL/ValueConverter.cs
@@ -33,6 +33,9 @@ namespace GraphQL
             Register(typeof(long), typeof(int), LongToInt);
 
             Register(typeof(double), typeof(decimal), DoubleToDecimal);
+
+            Register(typeof(string), typeof(Uri), ParseUri);
+            Register(typeof(Uri), typeof(Uri), ParseUri);
         }
 
         private static object IntToDouble(object value)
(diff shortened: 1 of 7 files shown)


joemcbride:
Hey, thanks for the PR though this is one of the cases where I'm not sure I want to add this since it is not defined in the core GraphQL spec.

For a "proper" implementation you would need to handle the `UriValue` in the `AstPrinter`. Also at present I do not believe you would ever get a `UriValue`, since it would get converted to a `StringValue` first.  To ever get a `UriValue` you would need to [alter this code](https://github.com/graphql-dotnet/graphql-dotnet/blob/bef9e5046ee532328609b2a86486e5cfc89be4aa/src/GraphQL/GraphQLExtensions.cs#L372).

src/GraphQL/ValueConverter.cs
```diff
@@ -33,6 +33,9 @@ static ValueConverter()
             Register(typeof(long), typeof(int), LongToInt);
 
             Register(typeof(double), typeof(decimal), DoubleToDecimal);
+
+            Register(typeof(string), typeof(Uri), ParseUri);
+            Register(typeof(Uri), typeof(Uri), ParseUri);
```
joemcbride: Don’t need Uri->Uri registration.  The code checks to see if the value is of the requested type and returns if so.

MNie: fix'd

MNie:
@joemcbride ok I get it :). I fix the pr, so now all should be fine. Since I used this library in my daily work I want to contribute to it somehow so I grab some issues and try to resolve/fix them. But as you said if the URI is not available in GraphQL. I'm not 100% sure that it would be good to add it to a GraphQL-Dotnet? But on the other hand, thanks to it we have a confidence that field of type URI would be properly formatted no matters what is done on the server side.

Issue #280, opened 2017-02-20
Are DocumentExecuter and ComplexityAnalyzer supposed to be thread-safe?


niklaskallander:
Hi,

I'm wondering whether `DocumentExecuter` and `ComplexityAnalyzer` are supposed to be thread-safe? Or whether they are intended to be used one-off per request?

The `ComplexityAnalyzer` isn't thread-safe right now as it uses the field `_result` (which is replaced per request) to store data during execution/analysis (without any sync locks anywhere). In a multi-threaded environment this will lead to race-conditions (which I've been experiencing in my application).

If they aren't intended to be thread-safe I'll simply alter my code, but if it's a bug I thought I'd report it!

Thanks, great work! (:

joemcbride:
Thanks for the report!  `DocumentExecuter` should be thread safe, so looks like we need to update the `ComplexityAnalyzer` to be so as well.

joemcbride:
This should be fixed in #707 (Thanks @bittercoder!)

PR #707, opened 2018-06-25
Make complexity analysis pass threadsafe

bittercoder:
The `ComplexityAnalyzer` is currently causing us issues intermittently due to it not being thread safe.

This PR reworks the class a little to make it thread-safe by passing around the context (complexity result, fragments and loop counter) in an instance of an inner class, so we don't have these are shared private members.

This relates to https://github.com/graphql-dotnet/graphql-dotnet/issues/280


bittercoder:
Thanks for merging this @joemcbride

Issue #475, opened 2017-11-13
When registering a field, throw an error if the name does not match the spec


joemcbride:
http://facebook.github.io/graphql/October2016/#sec-Names

```
/[_A-Za-z][_0-9A-Za-z]*/
```

PR #717, opened 2018-06-28
#475 when registering a field throw an error if the name doesn't match the spec

KoditkarVedant:
Fix: #475 
- [x] Adds field name validator as stated by spec
- [x] Update `AddField` in `ComplexGraphType` and `Name` in `FieldBuilder<TSourceType, TReturnType>` to call validator
- [x] Update `FieldBuilder` and `ConnectionBuilder` to set `default` as default field name.
- [x] Added Test cases for changes

### DIFF
--- a/src/GraphQL.Tests/Builders/ConnectionBuilderTests.cs
+++ b/src/GraphQL.Tests/Builders/ConnectionBuilderTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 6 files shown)


joemcbride:
#475

KoditkarVedant:
Hello @joemcbride, I would like to understand in which case we want users to set the field name as `null`?

KoditkarVedant:
Hello @joemcbride, can you please look into this test case.
image

This test case occasionally fails, even on `master` branch.

joemcbride:
> Currently, I have considered null as a valid field name, as it is currently allowed.

A null or empty name should not be allowed.

> This test case occasionally fails, even on master branch.

Yes that is a flakey test.  It never fails on .NET Core 1.1 or NET45, but it sometimes fails on Net Core 2.0.

KoditkarVedant:
@joemcbride, Updating the validator is a no-brainer but we need to make a change in `FieldBuilder`. The simplest thing I could think of is to accept the `name` as an argument to `FielderBuilder` then it will make `Name` method impractical.

KoditkarVedant:
@joemcbride,  There is another approach I can think of is, Currently while adding the Fields using the `FilderBuilder` method we set the field name to `null` instead we can set some default name to like `_` or `default`. This will not break the existing code written in the older version and even we make sure we have valid field name if the user doesn't pass name.

KoditkarVedant:
Hello @joemcbride, Can you please have a look at this PR? I have updated the code. Now it will throw an error if the field name doesn't match the spec, A `null` or `empty` name is not be allowed.

joemcbride:
I think this looks good.  Since I merged your other PR this one now has a conflict.  If you can fix that then I'll go ahead and merge.

KoditkarVedant:
Sure.

KoditkarVedant:
Fixed conflict. Done.

Issue #587, opened 2018-03-06
InputObjectGraphType referencing an ObjectGraphType should error


tinganho:
I lost a day or so tracking down a bug in introspection. I've got this generic error:

```shell
Introspection must provide input type for arguments.
```
Turns out, `InputObjectGraphType` cannot have fields containing a `ObjectGraphType`. Couldn't the `Field<TGraphType>` be constrained to only accept `InputObjectGraphType` instead of `ObjectGraphType`?

joemcbride:
Sorry to hear about that!  I think that `Field<TGraphType` will need to stay at `IGraphType`, though could probably change `AddField` on `ComplexGraphType` to be virtual and ensure the given type does not inherit from `ObjectGraphType`.  Both `InputObjectGraphType` and `ObjectGraphType` use `ComplexGraphType`.

PR #720, opened 2018-06-29
#587 inputobjectgraphtype referencing an objectgraphyype should throw specific error

KoditkarVedant:
Fix: #587 

- [x] Update `AddFIeld` method in `ComplexGraphType` to be virtual.
- [x] Override `AddField` method in `nputObjectGraphType<TSourceType>` to throw an exception if fields contains `ObjectGraphType`.
- [x] Added Test cases for changes


joemcbride:
🎉

PR #721, opened 2018-06-30
For discussion: Documentation reorg

Perlkonig:
Here's my suggestion for restructured documentation (just the "Getting Started" page). Almost all the original text is still there, but I moved things around to be in a more logical order. I welcome any and all feedback. People can push to the commit or give me text that I can add myself. Whatever people want to do.

What I did:
* Rewrote the intro to clarify the library's scope. The audience needs to already understand basic GraphQL concepts. The library documentation should be as agnostic as possible about the database and interactivity layers.

* I reordered the top-level headings to mirror the introduction, putting things in a logical order:
  * Types & Interfaces
  * Queries & Mutations
  * Execution
    * Schema Generation
    * `DocumentExecuter`
  * Further Reading

* I added many `TODO:` tags highlighting areas that need expansion (IMO).

Here's a list of the TODO comments for discussion, in order of appearance:

* "Types" section
  * Explain more fully the `Field` function and its variants.
  * Be sure to introduce the `ListGraphType`.
  * Introduce `context.Source`.
  * Give concrete examples of more-complex field structures.

* "Resolving Ambiguity" section
  * First off, is this section even necessary? If `ObjectGraphType<Type>` already gives a default `IsTypeOf`, why bother the reader with this information? Are there situations where the default doesn't work? If so, explain.
  * Give some indication of why one should choose one method over another. Are there performance implications, for example?
  * In the `ResolveType` example, you need explain the interface constructor signature.
  * In the `IsTypeOf` example, there may be an error in the code.

* "Arguments" section
  * In the sample code, the `context.Arguments` line is never explained or used.
  * We need to introduce optional (nullable) arguments.

* "Queries" section
  * Needs to be expanded to demonstrate basics like how to search and return multiple results.

* "Mutations" section
  * Sample code should be cleaned up to focus on the mutation-specific code (delete the schema generation code, for example).
  * Talk more about error handling and validation. What will be returned if something goes wrong at the database layer?

* "Schema Generation" section
  * Needs to essentially be written from scratch. There are two approaches scattered throughout the docs, but they need to be consolidated and fully explained.

* "DocumentExecuter" section
  * Also needs to be written from scratch. Nowhere do we explicitly explain how this function works and can be modified.

### DIFF
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -11,17 +11,19 @@ As its name suggests, this library is a .NET implementation of [GraphQL](http://
 +------------------------------------+
 +------------------------------------+
 |         +-----------------------+  |
-|         | Queries and Mutations |  |
-| GraphQL +-----------------------+  |
+|         | Execution             |  |
+|         +-----------------------+  |
+| GraphQL | Queries and Mutations |  |
+|         +-----------------------+  |
 |         | Types and Interfaces  |  |
 |         +-----------------------+  |
 +------------------------------------+
 +------------------------------------+
-|              Database              |
+|                Data                |
 +------------------------------------+
 ```
 
-GraphQL sits between your database the "outside world." This library only concerns itself with this middle layer. For simplicity, the documentation uses hard-coded data, but in a real-world application your database layer would consist of some sort of ORM, like Entity Framework. And the only aspect of interactivity we will discuss here is how to actually execute the query. How specifically the query is received and the output returned is beyond the scope of this document. See the ["Further Reading"](#further-reading) section at the end of this document for examples and tutorials specific to various ORMs and deployment scenarios.
+GraphQL sits between your data and the "outside world." This library only concerns itself with this middle layer. For simplicity, the documentation uses hard-coded data, but in a real-world application your data layer would consist of some sort of ORM (like Entity Framework) or other data provider. And as for interactivity, how specifically the query is received and the output returned is beyond the scope of this document. See the ["Further Reading"](#further-reading) section at the end of this document for examples and tutorials specific to various ORMs and deployment scenarios.
 
 ## Example
 
@@ -82,7 +84,7 @@ namespace ConsoleApplication
      * of searches and filters you support.
      *
      * In this example, the data itself is hardcoded in the `resolve:` line.
-     * In the "real world," you'd be querying your database.
+     * In the "real world," you'd be querying your data provider.
      */
     public class StarWarsQuery : ObjectGraphType
     {
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -52,9 +52,10 @@ namespace ConsoleApplication
     using GraphQL.Types;
 
     /*
-     * Database Layer
+     * Data Layer
      * In this example, it's a simple object. In the "real world,"
-     * it would likely be an object tied to an ORM, like Entity Framework.
+     * it would likely be an object tied to an ORM, like Entity Framework,
+     * or a data transfer object tied to some other data provider.
      */
     public class Droid
     {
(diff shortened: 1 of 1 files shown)


docs/src/getting-started.md
```diff
@@ -3,14 +3,31 @@
...
+| GraphQL +-----------------------+  |
+|         | Types and Interfaces  |  |
+|         +-----------------------+  |
++------------------------------------+
++------------------------------------+
+|              Database              |
++------------------------------------+
+```
+
+GraphQL sits between your database the "outside world." This library only concerns itself with this middle layer. For simplicity, the documentation uses hard-coded data, but in a real-world application your database layer would consist of some sort of ORM, like Entity Framework. And the only aspect of interactivity we will discuss here is how to actually execute the query. How specifically the query is received and the output returned is beyond the scope of this document. See the ["Further Reading"](#further-reading) section at the end of this document for examples and tutorials specific to various ORMs and deployment scenarios.
```
jquense: I think this overstates the case a bit. Graphql occupies the same same space as a REST api, that may mean the level below it is a database, but more often than not its probably a service layer or other business logic  specfic thing between the DB or ORM. It just talk to the ORM directly but we shouldn't imply that's the only or correct way to use it.

docs/src/getting-started.md
```diff
@@ -31,13 +49,66 @@ namespace ConsoleApplication
...
+
+    /*
+     * Query/Mutation Layer
+     * The root query is the entry point of your GraphQL service.
+     * You must have one, and only one, root query object.
+     * This object gathers together all your defined types and configures the types
+     * of searches and filters you support.
+     *
+     * In this example, the data itself is hardcoded in the `resolve:` line.
+     * In the "real world," you'd be querying your database.
```
jquense: again this isn't really true

docs/src/getting-started.md
Perlkonig: We can certainly expand on that. Any suggestions? 

Would you prefer calling it a more general "data layer" instead of "database"? Something like:

> GraphQL sits between your data and the "outside world." ...but in a real-world application your data layer would consist of an ORM or some other data service provider...

docs/src/getting-started.md
Perlkonig: How would you prefer to say it? Even if you're not querying through a direct ORM, this *is* where you fetch the actual data, however that is provisioned. The important part to me is to signal that the `resolve:` line given is *not* representative of real-world application.

jquense: yeah "data" in the general sense is correct, im only nitpicking "database" as it connotes a specific thing. Resolve, for instance may make an http request to another REST API

Perlkonig:
I did some more extensive rewriting. There's lots I don't understand, so I'm not putting this text forward as some ideal or final state. But in my experience it's easier to edit existing text than to create something from nothing. For your consideration.

joemcbride:
I've pushed an updated docs site based on gatsbyjs.  This will make it easier to split sections up.

I've basically kept what was there for now, but I was planning on pulling out a lot of the sections into their own documents.

https://graphql-dotnet.github.io/

SimonCropp:
@joemcbride so based on your comment this should be closed?  @Perlkonig can u review the new doco in the context of your requested changes?

sungam3r:
I think the first thing worth to do is to update docs site to reflect current state of docs in master.

Perlkonig:
I'm a technical writer/editor in my day job. I teach plain language. I also happen to be developer on the side. My rewrite was undertaken after my struggles trying to understand how this library works and how to adapt it to more practical situations (in my case, implementing a GraphQL front end on an AWS lambda querying an AWS RDS backend). My goal was to make the documentation more immediately clear to a wider audience. I do not claim to fully understand all the ins and outs, but the documentation as currently written does not sufficiently guide a newcomer through what precisely this library does and does not do and how to actually use it. Right now it's a list of seemingly random pages describing various functions but no explanation of how they relate to each other.

The GraphQL project I was working on has been on hold for a while, and the documentation has diverged a fair bit since I last looked at this over a year ago. And for all I know, the way the library works has changed significantly. I don't have the time right now to revisit this. If the team wants to expand the reach of the documentation, then this PR represents my recommendation for how to approach that: have at least a single introductory page that logically explains how the major pieces fit together and explain how one would start incorporating this library into a larger project.

If that's not something the team wants to do, then that's fine. This was just a suggestion. I have found the library very helpful, and I wanted to make it easier to be adopted more widely.

PR #724, opened 2018-07-01
Add repeated subfields functionality. Fixes #602

### LINKED ISSUE #602: Repeated sub-fields quit working
To simplify query generation, we had some queries of the form

```
query {
  person(id: 1) {
    business { id name }
    business { id address { id street city state } }
  }
}
```

with repeated sub-fields (`business`) in this case.

This worked in GraphQL .NET 0.17.3, but, when I upgraded to 2.0.0-alpha-868, only the most recent sub-field was honored.

We were able to work around it by restructuring our queries, but it seems like a bug in GraphQL .NET for it to not support this use case.

Thanks!

mrtes:
Allows to merge subselections of equivalent fields during execution.

### DIFF
--- a/src/GraphQL.Tests/Execution/RepeatedSubfieldsTest.cs
+++ b/src/GraphQL.Tests/Execution/RepeatedSubfieldsTest.cs
@@ -1,0 +1,106 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using GraphQL.Execution;
+using GraphQL.Language.AST;
+using GraphQL.Types;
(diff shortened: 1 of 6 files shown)


mrtes:
Regarding AppVeyor:
The failing DataLoader test sometimes passes and sometimes fails on my machine.
Seems to be a timing/concurrency issue?

joemcbride:
#602

joemcbride:
I went ahead and added those changes in #756 since they were minor.  Thanks for the help here!

PR #726, opened 2018-07-03
fix backward compatibility in batching

MNie:
This fixes a problem introduced in a #709 . After update we could assign to a batch only IDictionary, Dictionary is not acceptable. In this pr the accepted values would be an `IEnumerable<KeyValuePair<TKey, T>>` so we could assign, Dictionary, IDictionary, IReadOnlyDictionary. But I'm not 100% sure if this change is okay since the time needed to access an element right now is O(n) instead of O(1).

cc: @joemcbride


MNie:
bump @joemcbride

MNie:
@joemcbride 
I think the API would be more flexible so we are not supposed to have one concrete return type. But I agree it would create some complication in an implementation.

PR #734, opened 2018-07-12
Add possibility to set Data in ExecutionErrors ctor

MNie:
I realize that there is no possibility to add Data to a `Data` field in an `ExecutionError` constructor explicitly. It could be set only via the Exception object. Since someone could use some `ResultType` with an `IError` which in fact doesn't have to be an `Exception` it is a little bit overhead to set an Error.Data via adding it like:
```csharp
var err = new ExecutionError("");
var err.Data.Add("key", "value");
var err.Data.Add("key1", "value");
var err.Data.Add("key2", "value");
var err.Data.Add("key3", "value");
```
so it would be more suitable to set "error data" via constructor like that:
```csharp
var errData = new Dictionary<string, object>();
var err = new ExecutionError("", errData);
```
this pr is a proposition if it doesn't align with how the library is intended to "look like" it could be closed.

CC: @joemcbride

### DIFF
--- a/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
+++ b/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


joemcbride:
Seems like a good addition to me.

PR #746, opened 2018-07-25
Performance enhancements

kamsar:
A collection of minor non-breaking code tweaks that enhances the performance and scalability of graphql-dotnet significantly.

In particular, 92fc0a520ac920d9b3813ac3a4412bd4afa8dc5e improves service time of introspection queries on larger schemas by 30% by avoiding repeatedly new-ing an object thousands of times. The rest of the tweaks are simple things like reducing casting or avoiding reallocating objects in a loop.

Let me know if you've got any questions, and thanks for `graphql-dotnet` :)

### DIFF
--- a/src/GraphQL/Resolvers/NameFieldResolver.cs
+++ b/src/GraphQL/Resolvers/NameFieldResolver.cs
@@ -1,13 +1,10 @@
 using System;
-using System.Reflection;
 using GraphQL.Types;
 
 namespace GraphQL.Resolvers
 {
     internal class NameFieldResolver : IFieldResolver
     {
-        private static readonly BindingFlags _flags = BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance;
-
         public object Resolve(ResolveFieldContext context)
         {
             return Resolve(context?.Source, context?.FieldAst?.Name);
@@ -20,7 +17,14 @@ namespace GraphQL.Resolvers
                 return null;
             }
 
-            return source.GetPropertyValue(name);
+            var propertyValue = source.GetPropertyValue(name);
+
+            if (propertyValue == null)
+            {
+                throw new InvalidOperationException($"Expected to find property {name} on {source.GetType().Name} but it does not exist.");
+            }
+
+            return propertyValue;
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/GraphQL/Resolvers/NameFieldResolver.cs
```diff
@@ -20,15 +20,7 @@ public static object Resolve(object source, string name)
...
-            var prop = source.GetType()
-                .GetProperty(name, _flags);
-
-            if (prop == null)
-            {
-                throw new InvalidOperationException($"Expected to find property {name} on {source.GetType().Name} but it does not exist.");
-            }
-
-            return prop.GetValue(source, null);
+            return source.GetPropertyValue(name);
```
joemcbride: I think this is the only change I'm mildly concerned about.  Mind moving the exception to the `GetPropertyValue` method?  I think that provides helpful debugging - perhaps the error should only be thrown in debug mode?  Maybe that would cause too much confusion with different behavior?  It appears that method wasn't being used by anything else.

src/GraphQL/Types/ComplexGraphType.cs
```diff
@@ -35,7 +37,16 @@ private set
...
         {
-            return _fields.Any(x => string.Equals(x.Name, name));
+            if (string.IsNullOrEmpty(name)) return false;
+
+            return _fields.Any(x => string.Equals(x.Name, name, StringComparison.Ordinal));
+        }
+
+        public FieldType GetField(string name)
+        {
+            if (string.IsNullOrEmpty(name)) return null;
```
joemcbride: Perhaps these two checks should be `IsNullOrWhitespace`?

kamsar:
Changes made as requested

src/GraphQL/Resolvers/NameFieldResolver.cs
```diff
@@ -20,15 +17,14 @@ public static object Resolve(object source, string name)
                 return null;
             }
 
-            var prop = source.GetType()
-                .GetProperty(name, _flags);
+            var propertyValue = source.GetPropertyValue(name);
 
-            if (prop == null)
+            if (propertyValue == null)
```
joemcbride: This change is a little off.  This change would make it impossible to return a null value (which is valid), vs. ensuring that the requested property exists on the object.

joemcbride:
This is pushed as GraphQL 2.0.0-alpha-952.  Will take Nuget.org a while to index it.

Issue #732, opened 2018-07-08
ValidationError -> errocode not serialized in response


john1452:
I am providing an error code to the validation error class

```
 context.ReportError(new ValidationError(context.OriginalQuery, "MyErrorCode",
                    "Error message."));
```
However, the error code is not serialized by json.

```
{
  "errors": [
    {
      "message": "Error message."
    }
  ]
}
```

Also, the spec doesn't mention any "errorCode" property in the response. it Leans toward extensions which can contain a code. 
http://facebook.github.io/graphql/June2018/#sec-Errors

I am using **2.0.0-aplha-938**

joemcbride:
Since an `errorCode` was not part of the spec, that was the main reason to not serialize it.  Now that there is a spec for it, we can refactor that to use `extensions` instead.

joemcbride:
At present it _should_ be showing a `code` and `data` though ... so that seems like a bit of a design issue in that both `ValidationError` and the `ExecutionError` now have two different code properties.  So that should be fixed.


https://github.com/graphql-dotnet/graphql-dotnet/blob/4e74411859c27c3e44c8b3751abfcb51f2f1adcf/src/GraphQL/Execution/ExecutionResultJsonConverter.cs#L79-L95

KoditkarVedant:
Hello @joemcbride, I would like to work on this enhancement. Is it still up for grab?

joemcbride:
@KoditkarVedant Indeed it is.

KoditkarVedant:
@joemcbride why we are keeping two different classes `ValidationError`  and the `ExecutionError`. I was thinking to combine them as `GraphQLError` or `GraphQLException`.  

Can you help understand what is the need of exposing this publicly to users? https://github.com/graphql-dotnet/graphql-dotnet/blob/1842756f1ef4b190c19994fb53f92fb80701968b/src/GraphQL/Validation/ValidationError.cs#L28-L36

joemcbride:
@KoditkarVedant One is just a more specific error.  Kind of like `NullReferenceException` vs. `Exception`.

KoditkarVedant:
@joemcbride I got it. I think we should have another constructor which do not accepts nodes. I feel it is very specific and mainly used by `IValidationRule`.

What do you think about the PR changes? Did you get time to review it?

joemcbride:
> I think we should have another constructor which do not accepts nodes.

[...]

PR #748, opened 2018-07-28
#732 validation error errocode not serialized in response

KoditkarVedant:
Fix: #732 

- [x] Fix: using different Code property in ValidationError.cs class
- [x] Update ExecutiionResultJsonConverter to use `extensions` to serialize error code.
- [x] Add test cases for changes
- [x] Fix broken test cases


src/GraphQL/Execution/ExecutionResultJsonConverter.cs
```diff
@@ -78,8 +78,11 @@ private void WriteErrors(ExecutionErrors errors, JsonWriter writer, JsonSerializ
 
                 if (!string.IsNullOrWhiteSpace(error.Code))
                 {
+                    writer.WritePropertyName("extensions");
```
joemcbride: Looking at this and the spec it appears that both `code` and `data` should be in the extensions.  Mind making that change?

KoditkarVedant: Done...

PR #760, opened 2018-08-02
update to Newtonsoft.Json 11.0.2

SimonCropp:
to avoid possible binding redirect issues

### DIFF
--- a/src/GraphQL/GraphQL.csproj
+++ b/src/GraphQL/GraphQL.csproj
@@ -24,20 +24,17 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Reference Include="Microsoft.CSharp"/>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
     <PackageReference Include="GraphQL-Parser" Version="3.0.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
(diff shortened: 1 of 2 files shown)


src/GraphQL/GraphQL.csproj
```diff
@@ -24,8 +24,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Reference Include="Microsoft.CSharp"/>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
```
joemcbride: This project multi-targets several frameworks.  These need to stay under the `net45` umbrella.

SimonCropp: Afaik netore handles this scenario correctly. Ie it will redirect to the correct assemblies at run time 

Also when i left it the 45 condition it would not compile

joemcbride: Is that a new feature with the new SDK?  It appears CI is failing specifically because of that change.  It is using `2.1.4` SDK.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/global.json

SimonCropp: Yep. I updated the sdk in my netcore 2.1 pr. I will rebase this one after netcore is merged

SimonCropp:
i will re submit this in a new PR

PR #763, opened 2018-08-02
update to netcoreapp2.1

SimonCropp:


SimonCropp:
nor sure where in the CI you say what version of netcore is required?

src/GraphQL.Tests/GraphQL.Tests.csproj
```diff
@@ -1,7 +1,7 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0;netcoreapp1.1;net452</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp1.1;net452</TargetFrameworks>
```
joemcbride: Instead of replacing `2.0` with `2.1`, I think `2.1` should just be added as another target.

SimonCropp: My intention here was not to “also test against 2.1”. It was more to get the benefits of build and runtime perf improvements of 2.1 in the context of tests

With that it mind. Do u still want to multi target?

joemcbride: Yes.  I would prefer to run the tests against all the targeted frameworks.  Call me suspicious but I don't trust that they're 100% compatible.  😀

SimonCropp:
ok changes made. i will rebase my other PRs on this one once it is merged

SimonCropp:
gentle bump

PR #768, opened 2018-08-06
Getting error with the existing query structure

EmmanuelPonnudurai:
The Query property has the string without the curly braces, which does not work and leads to an error as shown below,
```js
{
  "errors": [
    {
      "message": "Syntax Error GraphQL (1:1) Unexpected Name \"hero\"\n1: hero { id name }\n   ^\n",
      "code": "SYNTAX_ERROR"
    }
  ]
}
```


joemcbride:
Thanks!  Awesome avatar - love Kenshin.

PR #785, opened 2018-08-15
use pattern matching

SimonCropp:
less code and gives us slightly better perf via less casting


SimonCropp:
@joemcbride seems to be failing due to versioning issue? u still happy to merge?

PR #786, opened 2018-08-17
Improve error message when type is missing

freakingawesome:
Otherwise you get a null reference exception.

### DIFF
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -122,6 +122,24 @@ namespace GraphQL.Tests.Types
             type.Fields.Last().Type.ShouldBe(typeof(StringGraphType));
         }
 
+        [Fact]
+        public void throws_informative_exception_when_no_types_defined()
+        {
+            var type = new ComplexType<Droid>();
+
+            var fieldType = new FieldType()
+            {
+                Name = "name",
+                ResolvedType = null,
+                Type = null,
+            };
+
+            var exception = Should.Throw<ArgumentOutOfRangeException>(() => type.AddField(fieldType));
+
+            exception.ParamName.ShouldBe("Type");
+            exception.Message.ShouldStartWith("FieldType.Type is required when ResolvedType is unspecified.");
+        }
+
         [Theory]
         [InlineData("__id")]
         [InlineData("___id")]
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -13,9 +13,11 @@ namespace GraphQL.Tests.Types
         internal class ComplexType<T> : ComplexGraphType<T> {
             public ComplexType()
             {
-                Name = typeof(T).Name;
+                Name = typeof(T).GetFriendlyName();
             }
         }
+
+        internal class GenericFieldType<T> : FieldType { }
 
         internal class TestObject
         {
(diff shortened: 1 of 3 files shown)


src/GraphQL/Types/ComplexGraphType.cs
```diff
@@ -59,10 +59,18 @@ public virtual FieldType AddField(FieldType fieldType)
...
+                if (fieldType.Type == null)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(fieldType.Type),
+                        $"Type is required when ResolvedType is unspecified.");
+                }
+                else if (!fieldType.Type.IsGraphType())
+                {
+                    throw new ArgumentOutOfRangeException(nameof(fieldType.Type),
+                        $"The declared Field type: {fieldType.Type.Name} should derive from GraphType, but doesn't.");
+                }
```
BenjaBobs: Maybe go the extra mile and also print the name of the type and field? :)

freakingawesome: Sure thing!

freakingawesome:
I've pushed up a change to improve the error message and added a test around the error.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -122,6 +122,24 @@ public void create_field_with_func_resolver()
...
+            {
+                Name = "name",
+                ResolvedType = null,
+                Type = null,
+            };
+
+            var exception = Should.Throw<ArgumentOutOfRangeException>(() => type.AddField(fieldType));
+
+            exception.ParamName.ShouldBe("Type");
+            exception.Message.ShouldStartWith("FieldType.Type is required when ResolvedType is unspecified.");
```
joemcbride: Little bit of clarity on the request to add names.  I was thinking the name of the field  and the GraphType vs. the name of the `FieldType` class.  With the intent being it would be easier to track down which class has the error.

Some potential examples:

> The declared field 'name' on 'MyGraphQL.MyGraphType' requires a field 'Type' when no 'ResolvedType' is provided.

> The declared field 'name' on 'ComplexType\<Droid\>' requires a field 'Type' when no 'ResolvedType' is provided.

joemcbride: Link for extension method for a "friendly type name".

https://stackoverflow.com/a/26429045/279764

```
    public static string GetFriendlyName(this Type type)
    {
        string friendlyName = type.Name;
        if (type.IsGenericType)
        {
            int iBacktick = friendlyName.IndexOf('`');
            if (iBacktick > 0)
            {
                friendlyName = friendlyName.Remove(iBacktick);
            }
            friendlyName += "<";
            Type[] typeParameters = type.GetGenericArguments();
            for (int i = 0; i < typeParameters.Length; ++i)
            {
                string typeParamName = GetFriendlyName(typeParameters[i]);
                friendlyName += (i == 0 ? typeParamName : "," + typeParamName);
            }
            friendlyName += ">";
        }

        return friendlyName;
    }
```

freakingawesome:
I've pushed up another set of changes incorporating the `GetFriendlyName()` functionality (had to tweak it because `IsGenericType` isn't available in .net standard 1.x), but I've only changed the error message in this PR to incorporate the friendly type name extension. It looks like type names are used in a handful of other places and error messages, and perhaps it would be best to consolidate them all to use the friendly name, but that's probably best left to a new Issue.

freakingawesome:
Hmmm... that AppVeyor build failed on `Operation_Can_Be_Cancelled` - I don't think that was me that broke it

SimonCropp:
Yes that test is flaky

SimonCropp:
@joemcbride this one looks good to me. Ok with a merge?

joemcbride:
:+1:

PR #792, opened 2018-08-21
Expose ExecutionContext through the IDocumentExecutionListener

BenjaBobs:
With this change, more meta data about the execution is exposed, in particular `ExecutionErrors` can be useful, if one wants a service to be able to report multiple errors, and add errors without throwing.
It is a breaking change though, as it changes the interface.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -129,10 +129,7 @@ namespace GraphQL
 
                 foreach (var listener in options.Listeners)
                 {
-                    await listener.AfterValidationAsync(
-                            context,
-                            validationResult,
-                            options.CancellationToken)
+                    await listener.AfterValidationAsync(context, validationResult)
                         .ConfigureAwait(false);
                 }
 
@@ -161,7 +158,7 @@ namespace GraphQL
                     if (context.Listeners != null)
                         foreach (var listener in context.Listeners)
                         {
-                            await listener.BeforeExecutionAsync(context, context.CancellationToken)
+                            await listener.BeforeExecutionAsync(context)
                                 .ConfigureAwait(false);
                         }
 
@@ -176,7 +173,7 @@ namespace GraphQL
                     if (context.Listeners != null)
                         foreach (var listener in context.Listeners)
                         {
-                            await listener.BeforeExecutionAwaitedAsync(context, context.CancellationToken)
+                            await listener.BeforeExecutionAwaitedAsync(context)
                                 .ConfigureAwait(false);
                         }
 
@@ -185,7 +182,7 @@ namespace GraphQL
                     if (context.Listeners != null)
                         foreach (var listener in context.Listeners)
                         {
-                            await listener.AfterExecutionAsync(context, context.CancellationToken)
+                            await listener.AfterExecutionAsync(context)
                                 .ConfigureAwait(false);
                         }
                 }
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -113,10 +113,24 @@ namespace GraphQL
                         _complexityAnalyzer.Validate(document, options.ComplexityConfiguration);
                 }
 
+                context = BuildExecutionContext(
+                    options.Schema,
+                    options.Root,
+                    document,
+                    operation,
+                    options.Inputs,
+                    options.UserContext,
+                    options.CancellationToken,
+                    metrics,
+                    options.Listeners,
+                    options.ThrowOnUnhandledException,
+                    options.UnhandledExceptionDelegate,
+                    options.MaxParallelExecutionCount);
+
                 foreach (var listener in options.Listeners)
                 {
                     await listener.AfterValidationAsync(
-                            options.UserContext,
+                            context,
                             validationResult,
                             options.CancellationToken)
                         .ConfigureAwait(false);
@@ -131,20 +145,6 @@ namespace GraphQL
                         Perf = metrics.Finish()
                     };
                 }
-
-                context = BuildExecutionContext(
-                    options.Schema,
-                    options.Root,
-                    document,
-                    operation,
-                    options.Inputs,
-                    options.UserContext,
-                    options.CancellationToken,
-                    metrics,
-                    options.Listeners,
-                    options.ThrowOnUnhandledException,
-                    options.UnhandledExceptionDelegate,
-                    options.MaxParallelExecutionCount);
 
                 if (context.Errors.Count > 0)
                 {
@@ -161,7 +161,7 @@ namespace GraphQL
                     if (context.Listeners != null)
                         foreach (var listener in context.Listeners)
                         {
-                            await listener.BeforeExecutionAsync(context.UserContext, context.CancellationToken)
+                            await listener.BeforeExecutionAsync(context, context.CancellationToken)
                                 .ConfigureAwait(false);
                         }
 
@@ -176,7 +176,7 @@ namespace GraphQL
                     if (context.Listeners != null)
                         foreach (var listener in context.Listeners)
                         {
-                            await listener.BeforeExecutionAwaitedAsync(context.UserContext, context.CancellationToken)
+                            await listener.BeforeExecutionAwaitedAsync(context, context.CancellationToken)
                                 .ConfigureAwait(false);
                         }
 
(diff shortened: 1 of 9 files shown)


src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -267,20 +267,22 @@ await listener.AfterExecutionAsync(context.UserContext, context.CancellationToke
...
-            context.ThrowOnUnhandledException = throwOnUnhandledException;
+            var context = new ExecutionContext
+            {
+                Document = document,
+                Schema = schema,
+                RootValue = root,
+                UserContext = userContext,
+
+                Operation = operation,
+                Variables = GetVariableValues(document, schema, operation?.Variables, inputs),
```
joemcbride: I prefer not to call methods in object constructor syntax since if the method throws an error you do not get what line number caused the error in the stack trace, it just points to the constructor.

BenjaBobs: I rebased onto master and moved potential exceptions out of the object initializer.

BenjaBobs:
Interesting.
`Operation_Can_Be_Cancelled` test seems to be failing on appveyor.
When running tests locally, it passes just fine.
`prints_multiple_interfaces` test however fails *sometimes*.
It appears to be a race condition, sometimes resulting in 
```
type Bar implements Foo, Baaz {
  str: String
}

interface Baaz {
  int: Int
}
```

and sometimes resulting in 
```
interface Baaz {
  int: Int
}

type Bar implements Foo, Baaz {
  str: String
}
```

Shane32:
I'd like to suggest that if `ExecutionContext` is exposed, it be changed to be a read-only class, or we expose a read-only `IExecutionContext` interface.

sungam3r:
OK, we can try to see what happens.

BenjaBobs:
Ended up redoing it all instead of rebasing because my branch was so old.

sungam3r:
>  // See: https://shouldly.readthedocs.io/en/latest/assertions/shouldMatchApproved.html
            // Note: If the AssemblyName.approved.txt file doesn't match the latest publicApi value,
            // this call will try to launch a diff tool to help you out but that can fail on
            // your machine if a diff tool isn't configured/setup. 
            publicApi.ShouldMatchApproved(options => options.WithDiscriminator(type.Assembly.GetName().Name));

src/GraphQL/Execution/IExecutionContext.cs
```diff
@@ -0,0 +1,26 @@
...
+        Fragments Fragments { get; set; }
+        List<IDocumentExecutionListener> Listeners { get; set; }
+        int? MaxParallelExecutionCount { get; set; }
+        Metrics Metrics { get; set; }
+        Operation Operation { get; set; }
+        object RootValue { get; set; }
+        ISchema Schema { get; set; }
+        bool ThrowOnUnhandledException { get; set; }
+        Action<UnhandledExceptionContext> UnhandledExceptionDelegate { get; set; }
+        Variables Variables { get; set; }
```
sungam3r: > https://github.com/graphql-dotnet/graphql-dotnet/pull/792#issuecomment-593941032

Why are the setters here?

sungam3r:
Also I propose to remove generic listeners and close #1561 as duplicate.

src/GraphQL/Execution/IExecutionContext.cs
BenjaBobs: Oversight, will fix.

BenjaBobs:
Will fix this tomorrow if I have time.

src/GraphQL/Execution/DocumentExecutionListener.cs
```diff
@@ -11,61 +11,42 @@ namespace GraphQL.Execution
...
+        Task IDocumentExecutionListener.BeforeExecutionAsync(IExecutionContext context, CancellationToken token) => BeforeExecutionAsync(context, token);
 
-        Task IDocumentExecutionListener.BeforeExecutionAwaitedAsync(object userContext, CancellationToken token) => BeforeExecutionAwaitedAsync((T)userContext, token);
+        Task IDocumentExecutionListener.BeforeExecutionAwaitedAsync(IExecutionContext context, CancellationToken token) => BeforeExecutionAwaitedAsync(context, token);
 
-        Task IDocumentExecutionListener.AfterExecutionAsync(object userContext, CancellationToken token) => AfterExecutionAsync((T)userContext, token);
+        Task IDocumentExecutionListener.AfterExecutionAsync(IExecutionContext context, CancellationToken token) => AfterExecutionAsync(context, token);
 
-        Task IDocumentExecutionListener.BeforeExecutionStepAwaitedAsync(object userContext, CancellationToken token) => BeforeExecutionStepAwaitedAsync((T)userContext, token);
+        Task IDocumentExecutionListener.BeforeExecutionStepAwaitedAsync(IExecutionContext context, CancellationToken token) => BeforeExecutionStepAwaitedAsync(context, token);
```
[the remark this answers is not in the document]
sungam3r: Yes.

src/GraphQL/Execution/IExecutionContext.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Threading;
+using GraphQL.Instrumentation;
+using GraphQL.Language.AST;
+using GraphQL.Types;
+
+namespace GraphQL.Execution
+{
+    public interface IExecutionContext : IProvideUserContext
+    {
+        CancellationToken CancellationToken { get; }
```
sungam3r: Please insert newlines between properties.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -185,7 +185,7 @@ await listener.BeforeExecutionAwaitedAsync(context.UserContext, context.Cancella
                     if (context.Listeners != null)
                         foreach (var listener in context.Listeners)
                         {
-                            await listener.AfterExecutionAsync(context.UserContext, context.CancellationToken)
+                            await listener.AfterExecutionAsync(context, context.CancellationToken)
```
sungam3r: I see that second arg is not needed now.

sungam3r:
Finally merged after 1,5 years!

Shane32:
I'll update the upgrade guide

Shane32:
> Finally merged after 1,5 years!

lol !!  

👍

Shane32:
I might also issue a PR for xml comments for `IExecutionContext` now that it is public

Shane32:
For public use, I mean (rather than an class to supplement the internal workings)

sungam3r:
OK

BenjaBobs:
We did it! 😄 
I saw that the `ExecutionContext` had a lot of properties that weren't interfaced.
Should we consider making a PR interfacing them?

Shane32:
I would only to make some of them read-only, and in general wouldn't at this time.  The Metrics should probably stay read/write, of course.

However, perhaps we could make a few changes on the new `IExecutionContext` interface:
- `Fragments` to `IEnumerable<FragmentDefinition>`
- `List<IDocumentExecutionListener>` to `IEnumerable<IDocumentExecutionListener>`
- `Variables` to `IEnumerable<Variable>`

@sungam3r Would you concur?  These changes require no other changes to the code base.

To change `Document`, `FragmentDefinition`, `Operation`, and `Variable` to read-only interfaces (which rely only on other read-only interfaces) is probably a lot of work and not necessary at this time.  Those fields are all constructed for the particular request, and modifying one of them cannot have any effect on other simultaneously operating requests.

I have been working on PRs to change `IFieldType` and `IGraphType` to read-only interfaces (which only rely on other read-only interfaces), but in this case the graph/field is typically a singleton, and changing one can have an effect on any other simultaneously or subsequently operating requests.

sungam3r:
I am wary of this "readonly-infection". The main thing is to make changes meaningfully. It just makes no sense to wrap the entire implementation with interfaces.

Regarding `IEnumerable<T>` everywhere. It leads to iterator boxing every time iterating over underlying `List<T>`.

Shane32:
> It just makes no sense to wrap the entire implementation with interfaces.

Agree

> The main thing is to make changes meaningfully.

Agreed

> Regarding `IEnumerable<T>` everywhere. It leads to iterator boxing every time iterating over underlying `List<T>`.

True, but I really don't know why a document listener would need to iterate over a list of document listeners, and if so, why we need to optimize for such a scenario.


Either way is fine with me for this interface.  I really think it will be beneficial for the schema and underlying graphs and fields, but I don't think it matters much here.  If someone wants to write a document listener to modify query variables, so be it.  It's not going to affect the next request.

Issue #795, opened 2018-08-23
Multiple Threads using DataLoader


dBug404:
I recently migrated to GraphQL NET 2.0.0. 

When I implemented the dataloader in a query it worked fine, but in in more complex queries I encountered, that some other resolve methods are executed in a different thread and HttpContext.Current is null which causes problems. This disappeared, when I refactored the resolve with the dataloader back to not using dataloader.

The query is executed in a WebAPI controller in .NET 4.7.

Is this behaviour intended or is this a bug? Or can I configure, if the resolution stays in the same context so that I can use HttpContext.Current?

dBug404:
You are right and in new code this is done. But in my case GraphQL is used on top of a bigger legacy backend so that is not really an option in the near future.

In most cases HttpContext.Current is abstracted via interfaces. So I could pass the HttpContext.Current via UserContext to the resolve methods and then override the standard DI registration for each specific DI resolution in the resolve method. 

But in my opinion it would be preferable, if the user can configure if multiple threads are used or not.

johnrutherford:
The DataLoader doesn’t do anything with threads. There are unit tests that run within a synchronization context to make sure there are no issues with it. 

Is it possible that your code is creating threads or queuing work on the `ThreadPool` with `Task.Run()` or something? 

Could you create a simple app that reproduces the issue?

dBug404:
I've created a simple app https://github.com/dBug404/graphql-dotnet-issue-795

Both projects can reproduce the issue. In graphql.web.simple I removed all StructureMap DI code.

The endpoint url should be: http://localhost:8153/api/graphql

There is a queries.txt file which contains the sample query with and without debug information.

![image](https://user-images.githubusercontent.com/2844552/44657703-08d89880-a9fe-11e8-91a8-b5f3b13562e7.png)


The only place where the code does anything with Tasks is the func in the DataLoaderContextAccessor where I use Task.FromResult(...) to support non async methods in the backend. But this should not be responsible for starting new Threads, right?

`

            Field<DealerGraphType>("dealer", resolve:
                context =>
                {
                    //return dealerService.GetDealer(context.Source.DealerId);
                    var loader = new DataLoaderContextAccessor().Context.GetOrAddBatchLoader<string, Dealer>("getDealerByDealerId", dealerIds =>
                    {
                        IDictionary<string, Dealer> dictionary = new DealerService()
                            .GetDealers(dealerIds.ToList())
[...]

PR #808, opened 2018-08-31
Dataloader optimizations

johnrutherford:
Issue #795 got me looking at the DataLoader code again. I realized that the use of `ContinueWith()` in `DataLoaderBase.Dispatch()` was making it so that continuations were always run on the `ThreadPool`. So it was causing more threads to be used even when it was completely unnecessary like if the `Task` was already completed. 

So I re-wrote `Dispatch()` and `DispatchAll()` to use `async`/`await`. This results in less threads being used and improved performance. 

I also improved the DataLoader unit tests, and added a large query test that I used for profiling.


johnrutherford:
@joemcbride, looks like the AppVeyor build failed when downloading yarn packages (unrelated to my changes). Are you able to manually trigger a new build?

BenjaBobs:
IIRC, if you close and re-open the PR a new build will trigger.

cotzo:
Hey. Any ETA for when this will be published so we can also have #812?

Thanks

huysentruitw:
Any idea when this gets released? I'm currently running into dead-lock issues with dataloader in 2.1.0 when calling EF Core queries with async/await. When I call everything synchronously and return `Task.FromResult` everywhere, I don't have the dead-lock issue. I wonder if the changes in this pull-request affect this behavior.

johnrutherford:
@huysentruitw, these changes are unlikely to make a difference if you’re having deadlocks. Can you open up a new issue with some more details and tag me? I can try and help you there.

huysentruitw:
You can check-out this branch: https://github.com/huysentruitw/conventions/tree/sample/dataloader-efcore

I am using the DataLoader implementation in combination with the GraphQL Conventions library, but that shouldn't make any difference here.

Currently, the repositories use sync calls to EF Core and what I see is that I can query first and second level data with these GraphQL queries:

First level:

```
query AllMovies {
  movies {
    id
    title
    genre
    releaseDateUtc
  }
}
```

Second level:

```
query AllMovies {
	movies {
    id
    title
    genre
    releaseDateUtc
    
    actors {
      id
      name
    }
  }
}
```

But as soon as I add a third level to the query, the thing deadlocks:

```
query AllMovies {
	movies {
    id
    title
    genre
    releaseDateUtc
    
    actors {
      id
      name
      
      country {
        code
        name
      }
    }
  }
}
```

The reason is because the `DataLoaderContext.DispatchAll()` method exists before all field resolvers were able to run. I've used similar code in the past with EF6 and .NET 4.7.1 and mostly deadlocks like this were caused by a missing `ConfigureAwait(false)` somewhere (which should no longer be needed because there is no AspNetSyncrhonizationContext for .NET Core). Anyway, in that test project, I've already tried to switch to this dataloader implementation: https://github.com/dlukez/dataloader-dotnet but the results are the same (I had some hope, because that implementation contains it's own synchronization context and task scheduler, but no luck...).

huysentruitw:
I'll try my own DataLoader implementation (https://github.com/huysentruitw/yadi) later today or tomorrow and report back.

huysentruitw:
Sorry to say but, DAMN I'M GOOD! 🙂 

It works fine with my own [YADI](https://github.com/huysentruitw/yadi) implementation! See the Yadi branch: https://github.com/huysentruitw/conventions/tree/sample/yadi

And even with async EF Core calls: https://github.com/huysentruitw/conventions/tree/sample/yadi-async-repos

I think the reason is the while loop, which keeps running and checking for new loaders: https://github.com/huysentruitw/yadi/blob/master/src/Yadi/DataLoaderContext.cs#L49

Issue #798, opened 2018-08-27
IDocumentWriter - support writing to response stream directly?


DSilence:
## Summary

The current document writer only support returning a serialized string, which significantly limits the supported serializer usage. Fast json serializers for .net, like [utf8json](https://github.com/neuecc/Utf8Json) or  [SpanJson](https://github.com/Tornhoof/SpanJson) support writing data directly to stream asynchronously, which positively impacts performance, since they don't allocate a string. In addition, such approach would allow the usage of binary protocols in future.

### Proposal

Change the IDocumentWriter interface to look similar to this
```
public interface IDocumentWriter
  {
    Task WriteAsync(Stream stream, object value);
  }
```

or, even better, use generic method

```
public interface IDocumentWriter
  {
    Task WriteAsync<T>(Stream stream, T value);
  }
```
### Motivation

Better performance. This is a breaking change.

DSilence:
I'm willing to work on a PR if you guys are ok with it.

joemcbride:
Yes ... here's a thought.  `DocumentWriter` is meant to be used with `ExecutionResult`.  I can't think of any other use for it.  Perhaps we should just make it specific to that.  It would also be good to provide an upgrade path with marking the previous method `Obsolete`, which we would then remove in the next major version.

```csharp
public interface IDocumentWriter
{
  Task WriteAsync(Stream stream, ExecutionResult value);

  [Obsolete("This method is obsolete and will be removed in the next major version.  Use WriteAsync instead.")]
  string Write(object value);
}
```

DSilence:
I've given some thought to the WebSocket transport. The following solution might work:

1. IDocumentWriter should be able to write OperationMessage. This could be achieved using the original generic Write method concept (possibly with some generic restriction only allowing ExecutionResult and OperationMessage).
2. The OperationMessage payload type should be changed to object or generic. This way we can store ExecutionResult without turning it into JObject for writing purposes. 
3. -  Option A: In WebSocketWriterPipeline we create a MemoryStream and use IDocumentWriter to write OperationMessage to the stream. We can then use TryGetBuffer to get the ArraySegment and send it over socket.  (Optional) The MemoryStream instances used by WebSocketWriterPipeline could use an ArrayPool to avoid extra allocations. 
    -  Option B: We add an additional method to IDocumentWriter to write the object to ArraySegment or Memory. We can then use the produced entity to write it to a WebSocket channel.

[...]

PR #809, opened 2018-09-01
Implemented #798: IDocumentWriter - added support writing to response stream directly.

DSilence:
### Summary of changes:

- Added a new method to IDocumentWriter.
` Task WriteAsync(Stream stream, ExecutionResult value);`
- Marked the old method as obsolete
- Added implementation for DocumentWriter. 
   **Note**: the StreamWriter can be a bottleneck here, since it doesn't use ArrayPool in default implementation (see https://github.com/dotnet/corefx/issues/23874, Asp.Net has it's own implementation https://github.com/aspnet/HttpAbstractions/blob/bcb7bcbb7a6461204f4cefd9848e2a782ebb0a6b/src/Microsoft.AspNetCore.WebUtilities/HttpResponseStreamWriter.cs). This can further improve the performance.
- Added an extension method for DocumentWriter that mimics the old behavior. After some thought, I've allowed encoding to be passed.
- Updates unit tests to use the extension method in place of now obsolete method.
- Updated middleware in GraphQL.Harness project.

### The benchmark
Tested serializing of 10 000 Droid objects for 2 DocumentWriter implementations (one using Newtonsoft.Json, another using Utf8Json). The results can be found below. 
``` ini

BenchmarkDotNet=v0.11.0, OS=Windows 10.0.17134.228 (1803/April2018Update/Redstone4)
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=2.1.401
  [Host]     : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT


```
|                 Method |      Mean |     Error |    StdDev |     Gen 0 |     Gen 1 |     Gen 2 | Allocated |
|----------------------- |----------:|----------:|----------:|----------:|----------:|----------:|----------:|
| GetHeroesNewtonsoftOld | 20.740 ms | 0.4793 ms | 0.4249 ms | 1312.5000 | 1156.2500 |  656.2500 |   7.02 MB |
| GetHeroesNewtonsoftNew | 18.919 ms | 0.1921 ms | 0.1702 ms | 1062.5000 |  468.7500 |  468.7500 |   3.69 MB |
|       GetHeroesUtf8Old |  4.841 ms | 0.0919 ms | 0.1021 ms | 1000.0000 | 1000.0000 | 1000.0000 |   5.42 MB |
|       GetHeroesUtf8New |  3.417 ms | 0.0326 ms | 0.0305 ms |  656.2500 |  656.2500 |  656.2500 |   2.76 MB |

Once this is merged, I can update the usages of this API in server project.

### DIFF
--- a/src/GraphQL/Http/DocumentWriter.cs
+++ b/src/GraphQL/Http/DocumentWriter.cs
@@ -20,6 +20,7 @@ namespace GraphQL.Http
     public class DocumentWriter : IDocumentWriter
     {
         private readonly ArrayPool<byte> _pool = ArrayPool<byte>.Shared;
+        private readonly JsonArrayPool _jsonArrayPool = new JsonArrayPool(ArrayPool<char>.Shared);
         private readonly int _maxArrayLength = 1048576;
         private readonly JsonSerializer _serializer;
         private static readonly Encoding Utf8Encoding = new UTF8Encoding(false);
@@ -42,15 +43,19 @@ namespace GraphQL.Http
             _serializer.Formatting = formatting;
         }
 
-        public Task WriteAsync<T>(Stream stream, T value)
+        public async Task WriteAsync<T>(Stream stream, T value)
         {
             using (var writer = new StreamWriter(stream, Utf8Encoding, 1024, true))
-            using (var jsonWriter = new JsonTextWriter(writer))
+            using (var jsonWriter = new JsonTextWriter(writer)
+            {
+                ArrayPool = _jsonArrayPool,
+                CloseOutput = false,
+                AutoCompleteOnClose = false
+            })
             {
                 _serializer.Serialize(jsonWriter, value);
+                await jsonWriter.FlushAsync().ConfigureAwait(false);
             }
-
-            return TaskExtensions.CompletedTask;
         }
 
         public async Task<IByteResult> WriteAsync<T>(T value)
--- a/src/GraphQL/Http/DocumentWriter.cs
+++ b/src/GraphQL/Http/DocumentWriter.cs
@@ -1,16 +1,24 @@
+using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace GraphQL.Http
 {
     public interface IDocumentWriter
     {
+        Task WriteAsync(Stream stream, ExecutionResult value);
+
+        [Obsolete("This method is obsolete and will be removed in the next major version.  Use WriteAsync instead.")]
         string Write(object value);
     }
 
     public class DocumentWriter : IDocumentWriter
     {
-        private readonly Formatting _formatting;
-        private readonly JsonSerializerSettings _settings;
+        private readonly JsonSerializer _serializer;
+        private static readonly Encoding Utf8Encoding = new UTF8Encoding(false);
 
         public DocumentWriter()
             : this(indent: false)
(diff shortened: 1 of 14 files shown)


BenjaBobs:
I suggest adding the ability to write to JObject. Currently Server is using [Server](https://github.com/graphql-dotnet/server/blob/develop/src/Transports.Subscriptions.Abstractions/Subscription.cs#L66) is using `JObject.FromObject()` which can make subscription results look different from query/mutation results. If `IDocumentWriter` can write `JObject`s, it should be easy to fix.

DSilence:
Added an overload to write to JObject.

src/GraphQL/Http/DocumentWriter.cs
```diff
@@ -26,13 +34,62 @@ public DocumentWriter(bool indent)
...
+                return resolvedEncoding.GetString(buffer.Array, buffer.Offset, buffer.Count);
+#endif
+            }
+        }
+
+        public static async Task<JObject> WriteToJObjectAsync(this IDocumentWriter writer,
+            ExecutionResult value)
+        {
+            var json = await writer.WriteToStringAsync(value);
+            return JObject.Parse(json);
```
joemcbride: Did you try using `JObject.FromObject`?  Does it serialize correctly?  It may not, though is worth a shot.

```
public static JObject FromObject(
	Object o,
	JsonSerializer jsonSerializer
)
```

DSilence: It does work. 
The reason I chose not to call JObject.Parse is that
internally, JObject.FromObject just serializes into JTokenWriter anyway. 
https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JToken.cs#L1881.
Since DocumentWriter might potentially customize the serialization workflow, I chose to use this approach. In addition, there is no need to pass JsonSerializer here - the one in DocumentWriter will be reused.

joemcbride: I mentioned the one with the serializer to use if we were avoiding calling the `WriteToStringAsync`.  It would seem like calling `FromObject` would be much faster than serializing to string then re-parsing.

DSilence:
``` ini

BenchmarkDotNet=v0.11.0, OS=Windows 10.0.17134.254 (1803/April2018Update/Redstone4)
Intel Core i7-7700HQ CPU 2.80GHz (Max: 0.80GHz) (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=2.1.401
  [Host]     : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT


```
|              Method |     Mean |    Error |   StdDev |     Gen 0 |     Gen 1 |    Gen 2 | Allocated |
|-------------------- |---------:|---------:|---------:|----------:|----------:|---------:|----------:|
|          FromObject | 122.1 ms | 2.428 ms | 5.226 ms | 4800.0000 | 2000.0000 | 600.0000 |  25.81 MB |
| UsingDocumentWriter | 131.9 ms | 2.634 ms | 4.882 ms | 6000.0000 | 2500.0000 | 500.0000 |   36.3 MB |

FromObject will be faster. Both options are very slow in general and allocate a lot of memory (Compare this with the results of direct stream writing, it's using the same test set). I think we need to consider a better alternative than passing JObjects for WebSocket API.

BenjaBobs: This might very well be the case. Then maybe the `WriteToJObjectAsync` method shouldn't be included in order to not encourage it? Sorry for leading you out on a wild goose chase here. 🍕

DSilence: Yeah, I think we need to come up with something else for websockets. I'll remove the method for now.

DSilence:
Some more improvements added:

- Pooling for JsonTextWriter
- Flush to response stream async
- Used the HttpResponseStreamWriter from Microsoft.AspNetCore.WebUtilities to pool memory (in place of StreamWriter). In addition it should be more async friendly.

DSilence:
Gentle ping, the build seems to be failing because of the branch name. I can create a PR with a different branch name if needed.

johnrutherford:
@DSilence, @joemcbride, @pekkah, who wants to update the server project to use this?

DSilence:
@johnrutherford, I'm willing to take a look

johnrutherford:
I had some free time, so I went ahead and did a [pull request](https://github.com/graphql-dotnet/server/pull/173) to use this from the server project for the HTTP middleware. I did not make the change for web sockets as part of this pull request.

PR #817, opened 2018-09-07
Add the ability to filter execution nodes

gbiellem:
I needed the ability to suppress certain data based on a set of hierarchical access rules.
I couldn't find an extension point that allowed me to do this so I've introduced one.
The solution I came up with was to add an optional func to the execution options and then evaluate the func result prior to adding the data to the execution result.  

If there is another way I've missed please let me know :)


gbiellem:
What I was trying to achieve here has been addressed in Simon Cropp's GraphQL.EntityFramework project.
The [GlobalFilter](https://github.com/SimonCropp/GraphQL.EntityFramework#filters) is exactly what i was looking for.  It definitely neater than my proposed change here

PR #832, opened 2018-09-26
Added creation of DataLoaderContext in BeforeExecutionStepAwaitedAsync

tomasmcguinness:
Subscriptions weren't working when the GraphQL service had a DataLoader present in it. The Context was null. Copying the code from BeforeExecutionAsync, I initialise the Context in the method BeforeExecutionStepAwaitedAsync.


tomasmcguinness:
I will try and create a project with a simple demo of the problem over the next couple of days. An issue (https://github.com/graphql-dotnet/server/issues/162) was opened and that's how I came across the problem. That might give you some pointers. Essentially, subscriptions don't work if a DataLoader service is added.

PR #841, opened 2018-10-01
Allow custom casing on inferred enum graph types

freakingawesome:
The default behavior of EnumGraphType<TEnum> converts the enum values to uppercase. This commit allows you to specify a custom behavior for changing case.

### DIFF
--- a/src/GraphQL.Tests/Types/EnumGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/EnumGraphTypeTests.cs
@@ -23,6 +23,19 @@ namespace GraphQL.Tests.Types
             }
         }
 
+        class ColorEnumInverseCasing : EnumerationGraphType<Colors>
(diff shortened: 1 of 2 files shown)


joemcbride:
This appears to have a valid failing test, can you take a look?

freakingawesome:
Yea, that's an odd failure. I haven't been able to reproduce it locally or make any sense of how my minor change could have caused the failure. I'm merging latest from `master` to trigger another build and will see what happens.

freakingawesome:
Another failed build - but this time it's failing on what looks like nuget trying to create a an invalid version string.

freakingawesome:
I think that latest nuget failure is caused by the `/` character in my branch name. I could close this and open up a new PR against a more appropriate branch name to see if that fixes it, if you think that would be beneficial.

Issue #770, opened 2018-08-08
How to return Lists with subscriptions


Patrick-Spiegel:
## Summary

I am trying to return Lists as Subscription result. Is that even possible? 

## Relevant information

My Field Definition

            AddField(new EventStreamFieldType
            {
                Name = valueChanged_SUBSCRIPTION,
                Arguments = new QueryArguments(
                new QueryArgument<InputServiceNodeID> { Name = targets_Argument }
             ),
                Type = typeof(ListGraphType<LiveValueBase>),
                Resolver = new FuncFieldResolver<List<LiveValueBaseContainer>>(ResolveLiveValueBaseContainer),
                Subscriber = new EventStreamResolver<List<LiveValueBaseContainer>>(SubscribeLiveValueChanged)
            });

My Resolver

        private List<LiveValueBaseContainer> ResolveLiveValueBaseContainer(ResolveFieldContext arg)
        {
            var container = arg.Source as List<LiveValueBaseContainer>;
            return container;
        }

My Publisher

                _observer.OnNext(new List<LiveValueBaseContainer>
                {
                    new LiveValueBaseContainer
                    {
                        Quality = LiveValueQuality.Bad
                    }
                });

`LiveValueQuality` is an enum. `LiveValueBase` extends from `ObjectGraphType` and takes `LiveValueBaseContainer` as generic argument. the `LiveValueBaseContainer` is a simple class containing a property with this  enum. 

The value that I get as return  value is now always null. However if I remove the `List` and the `ListGraphType` I get the enum value returned correctly. This approach works in a similar way for Queries.
Did I do something wrong?

### Environment (if relevant)

Version: 2.0.0-alpha-978

Patrick-Spiegel:
Is this not supported? A reply would be kind :)

joemcbride:
You should be able to return anything from a subscription that you can return as a "normal" GraphQL query.  That being said, I haven't tried this exact scenario.  If you can provide a test that shows this not working that would be helpful in tracking down any issue there may be.

BitPuffin:
Just ran into this. Can confirm it doesn't work when you return a list (or array).

BitPuffin:
Should a new issue be created for this or should it stay here. Was trying to set up a test in a PR but struggled getting building working on linux :/

joemcbride:
@BitPuffin We can just use this issue.  Are you using the [server project](https://github.com/graphql-dotnet/server) for subscriptions?

[...]

PR #849, opened 2018-10-09
Test returning list in subscription

ztripez:
Test to demonstrate bug behind https://github.com/graphql-dotnet/graphql-dotnet/issues/770


Shane32:
These tests have been incorporated into PR #989 along with the proper fix.

PR #856, opened 2018-10-14
Avoid DataLoader dead-lock (fixes #855)

### LINKED ISSUE #855: DataLoader dead-locks
@johnrutherford 

You can check-out this branch: https://github.com/huysentruitw/conventions/tree/sample/dataloader-efcore

Running `Update-Database` in the package manager console should create the database with seed data.

I am using the DataLoader implementation in combination with the GraphQL Conventions library, but the latter shouldn't make any difference here.

Currently, the repositories use sync calls to EF Core and what I see is that I can query first and second level data with these GraphQL queries:

First level:

```
query AllMovies {
  movies {
    id
    title
    genre
    releaseDateUtc
  }
}
```

Second level:

```
query AllMovies {
  movies {
    id
    title
    genre
    releaseDateUtc
    
    actors {
      id
      name
    }
  }
}
```

But as soon as I add a third level to the query, the thing deadlocks:

```
query AllMovies {
  movies {
    id
    title
    genre
    releaseDateUtc
    
    actors {
      id
      name
      
      country {
        code
        name
      }
    }
  }
}
```

The reason is because the `DataLoaderContext.DispatchAll()` method exists before all field resolvers were able to run. I've used similar code in the past with EF6 and .NET 4.7.1 and mostly deadlocks like this were caused by a missing `ConfigureAwait(false)` somewhere (which should no longer be needed because there is no AspNetSyncrhonizationContext for .NET Core). Anyway, in that test project, I've already tried to switch to this dataloader implementation: https://github.com/dlukez/dataloader-dotnet but the results are the same (I had some hope, because that implementation contains it's own synchronization context and task scheduler, but no luck...).

When changing the repositories to async EF Core calls, you even can't query the second level without deadlocks.

huysentruitw:


src/GraphQL/DataLoader/DataLoaderBase.cs
```diff
@@ -27,15 +27,15 @@ public async Task DispatchAsync(CancellationToken cancellationToken = default)
...
+                return Task.FromResult(0);
             }
 
             try
             {
                 var result = await FetchAsync(cancellationToken)
                     .ConfigureAwait(false);
 
-                tcs.SetResult(result);
+                return Task.Run(() => tcs.SetResult(result), cancellationToken);
```
huysentruitw: I know this is not optimal, so if we can find any non-dead-locking solution without `Task.Run`, I'm all for it!

PR #873, opened 2018-10-23
Fix input types not being registered when ResolvedType is specified

rasmusjp:
I ran into a problem after upgrading to v2.2.0+ where the input types was not being registered if there where multiple arguments and they have `ResolvedType` explicit specified. Only the first type was being registered but not the rest.

Not sure why this hasn't been a problem before.


joemcbride:
Nice find, thanks!

PR #879, opened 2018-10-24
Apollo Engine Setup + Tracing

joemcbride:
Parts of this PR may be reverted before being pulled in.

### DIFF
--- a/apollo-engine.js
+++ b/apollo-engine.js
@@ -1,0 +1,27 @@
+const { ApolloEngineLauncher } = require('apollo-engine')
+
+// Define the Engine configuration.
+const launcher = new ApolloEngineLauncher({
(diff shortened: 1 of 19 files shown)


sungam3r:
@joemcbride Is it still actual? Did I understand correctly that this PR adds logging support to dataloaders?

sungam3r:
@Shane32 Do you want to review changes here? I doubt that this PR is still actual.

Shane32:
I personally would not add `ILogging` to the base data loader classes.

Regardless, I would close this PR as it has had no comments or updates in 4 years, and most of the modified code has vastly changed since then.  Most specifically, the data loaders were completely rewritten for version 3.0.

PR #890, opened 2018-11-04
More Parallelism using Dataflow Blocks

troydalldorf:
The pull request is to provide an alternative to the current ParallelStrategy. The DataflowBlocksParallelStrategy uses DataflowBlocks to "queue" tasks so that once a node has executed, it's child nodes are immediately added to the queue for processing. MaxDegreesOfParallelism provides control over the number of concurrent threads.

Some preliminary testing with my use case shows about 3-4 times improvement. Before and after charts included below. The after chart could be better, but the response time from the dependent API had a particularly slow response (AWS Lambda cold start), but the difference is still rather obvious.

The code still needs some work/validation/tests and review. @johnrutherford has pointed out a couple of possible issues:
- Possible performance overhead of creating data flow blocks on each request.
- Possible SynchronizationContext issues (see #795).
- Possible DataLoader issues with this strategy.

Also, I'm not sure where the best location would be to invoke: OnBeforeExecutionStepAwaitedAsync(context), but I understand this is necessary to prevent the DataLoader from deadlocking.

Before
parallel-strategy-before
After
parallel-strategy-after

### DIFF
--- a/src/GraphQL/Execution/DataflowBlocksParallelExecutionStrategy.cs
+++ b/src/GraphQL/Execution/DataflowBlocksParallelExecutionStrategy.cs
@@ -1,26 +1,29 @@
-using GraphQL.Execution;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
-using ExecutionContext = GraphQL.Execution.ExecutionContext;
 
 namespace GraphQL.Execution
 {
 
+    //TODO: Add unit tests.
+    //TODO: Validate performance overhead of creating data flow blocks on each request.
+    //TODO: Check for SynchronizationContext issues.
+    //TODO: Validate DataLoader can work with this strategy.
+    //TODO: Store maxDegreesOfParallelism in settings.
+    //TODO: Ensure ConfigureAwait is correct.
+    //TODO: This is needed to prevent the DataLoader from deadlocking... not sure exactly where this
+    //      needs to go...
+    //      await OnBeforeExecutionStepAwaitedAsync(context).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


sungam3r:
@Shane32 Let's look to this PR. It seems to be rather small. I understand that a lot of time has passed and we have greatly improved `ParallelExecutionStrategy`. Perhaps the new strategy doesn't make sense, but it's worth considering the ideas in the initial comment.

sungam3r:
@troydalldorf Would you like to work on this PR further? I see many `//TODO:` in your code. It would also be good to write some tests to show that the new strategy works correctly.

troydalldorf:
Hi @sungam3r. It's been a while since I've been in this space and this repository. I really wasn't pleased with the original ParallelExecutionStrategy, so I'm pleased to hear that there have been many improvements. I don't believe the concerns with DataflowBlocks were warranted since I've seen recommendations on using this similar approach for parallel loops from Microsoft. I don't have time to invest in this right now, so feel free to abandon the PR, or glean whatever you can from it. I created this PR as a proof-of-concept because there were some things I didn't understand well enough in the code when I wrote this, but wanted to validate that this provided the expected performance enhancements, which it did, and did not disappoint.

I will gladly review and updates to this PR as well. I appreciate your followup on this.

sungam3r:
OK. I looked at the code and think it's better to close the PR. The current parallel strategy has already been optimized. Also this code does not handle dataloaders.

troydalldorf:
Makes sense. Thank you.

Shane32:
Of note, this would require another dependency.  This is something I would be interested in playing around with but I don't have the time to do so now.  Also, with the new data loader implementation, it should be easy to make this properly support data loaders.  (And `OnBeforeExecutionStepAwaitedAsync` is deprecated in v4.)

Issue #889, opened 2018-11-03
More Parallelization


troydalldorf:
From what I can tell looking at Fiddler there seems to be room for improving the parallel execution. The current model seems to run one nodes children in parallel, but then wait for all of those to complete before continuing down the graph.

I would like to see each completed node immediately start work on the child nodes.

ParallelExecutionStrategy does indeed wait for each set of child nodes.

troydalldorf:
A nice-to-have addition to this would be to have the ability to process children of children in parallel where the parent id was the same or said differently where the child and the child of child have a 1:1 ratio. For example, if an organization only has one address, it would be great to have organization and its address child loaded in parallel.

troydalldorf:
The nice-to-have in mu comment is already available (#354)

troydalldorf:
@joemcbride I have a fork with the code [here](https://github.com/troydalldorf/graphql-dotnet/commit/83326b91d491f7ddb40d5a42ea765fe8530a6423).
It's working code, but pretty rough, i.e. no comments, testing, or ability to set settings, propagate task cancellation, etc. and I wasn't sure how to handle: OnBeforeExecutionStepAwaitedAsync(context).
If this is something that would be considered being merged into graphql-dotnet master, I can clean it up and add testing.
As a side note: Are there any plans to allow dependency injection of the ExecutionStrategy?

johnrutherford:
I wrote the ParallelExecutionStrategy, so I’d certainly be interested in checking this out in a pull request. 

I’d be a little concerned if there would be some scaling issues if a new Dataflow pipeline is created for each execution. There could also be issues with running in a SynchronizationContext (like with #795). I’m also not sure how it would work batching queries with the DataLoader. 

At the very least, maybe it could be available as an alternate execution strategy that users could opt-in to.

johnrutherford:
@troydalldorf, I’ll check it out. The DataLoader requires `OnBeforeExecutionStepAwaitedAsync` to be called or it will deadlock. 

Currently one can sub-class `DocumentExecuter` and override `SelectExecutionStrategy()` to use a different execution strategy.

troydalldorf:
@johnrutherford, here's the pull request:
https://github.com/graphql-dotnet/graphql-dotnet/pull/890

Still has plenty of work to do, but would appreciate the feedback. Thanks!

johnrutherford:
[...]

PR #891, opened 2018-11-05
Optimize ParallelExecutionStrategy

johnrutherford:
Based on the conversation in #889, I looked at how the `ParallelExecutionStrategy` could be improved upon. 

Currently it executes in steps where nodes are executed and then we wait for them all to complete before starting to execute the child nodes. This repeats until the edges of the graph are reached. 

The problem with this is that the currently executing nodes could finish at different times. If one node takes a long time, we have to wait for that node to complete before executing the child nodes of the nodes that have already completed. 

I made some changes so that we can start executing child nodes as soon as parent nodes complete. And this works with the DataLoader because `OnBeforeExecutionStepAwaitedAsync` is called at the appropriate times. If we can start executing child nodes as soon as parent nodes are complete, it's possible to greatly reduce execution time.

I wrote a test for this, but I know that timing/multi-threaded tests can be flaky with AppVeyor, so if it fails, I'll see if there's another way I can do it.

I still want to do a bit more work on this. I might clean up some stuff and add another test.

### DIFF
--- a/src/GraphQL.DataLoader.Tests/DataLoaderQueryTests.cs
+++ b/src/GraphQL.DataLoader.Tests/DataLoaderQueryTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


troydalldorf:
One feature I would still need is the ability to control the maximum degrees of parallelism. This would prevent a few complex queries from consuming too many threads.

SimonCropp:
@joemcbride any objections to merging this?

joemcbride:
No objects.

johnrutherford:
@joemcbride, @SimonCropp, sorry for going dark on this. I had a lot of stuff come up around the holidays. 

I found that this made some queries much slower because of the O(n) performance of `List<T>.RemoveAt()`. I re-wrote it using a `LinkedList<T>`, but haven't gotten a chance to finish it completely. 

I think the merge should be reverted, if possible, until I can finish this up.

SimonCropp:
@johnrutherford i reverted, then re-applied in a new branch+pr for you to continue your changes. sorry for jumping the gun

https://github.com/graphql-dotnet/graphql-dotnet/pull/976

johnrutherford:
@SimonCropp thanks! I’ll try to finish it up soon.

PR #893, opened 2018-11-05
Add support for float/long to double/decimal conversion (closes #840)

### LINKED ISSUE #840: Could not find conversion from System.Int64 to System.Double
## Description

When upgrading to 2.0 from 2.0.0-alpha-870 we ran into this error. Could not find conversion from System.Int64 to System.Double.

### Steps to reproduce
Our Data Models are using long and DateTimeOffset in quite a few places which are the only thing I could come up with to check regarding this error.  That being said our GraphQL model types for long properties are using FloatGraphType (Because originally I think that was the only place that conversion was supported) fields and our DateTimeOffet properties using DateTimeGraphType fields.
We switched all the fields using DateTimeGraphType to DateTimeOffsetGraphType where the matching properties were DateTimeOffset and that had no effect.

### Expected result

I am puzzled as to why this error is occurring.  Have not been able to pinpoint but it shouldn't be throwing an exception with a conversion issue.

### Actual result

GraphQL call fails when converting data model to GraphQL type result here:
https://github.com/graphql-dotnet/graphql-dotnet/blob/4b02f6e2628d3f8e02829c60938616cea6a54fee/src/GraphQL/ValueConverter.cs#L179

### Environment

We are not yet using [GraphQL Server](https://github.com/graphql-dotnet/server)

ASPNET Core 2.1/NetStandard 2
GraphQL 2.0.0
AnyCpu/x64
Windows 10/Windows Server 2016

huysentruitw:


huysentruitw:
@joemcbride - gentle ping 😊

PR #812, opened 2018-09-03
741 more scalar types

KoditkarVedant:
Enhancement:- #741
 
- Added ScalarGraphType for following types
  - [x] Short
  - [x] UShort
  - [x] UInt
  - [x] ULong
  - [x] Guid
- Added test cases.

PR #917, opened 2018-11-19
Fix scalar types added in #812

mvestergaard:
Using the scalar types added in #812 doesn't actually work. Causing all sorts of exceptions. 

First of all the types weren't registered, so it attempted to resolve the types from the dependency resolver delegate.

Once that was fixed it failed in the `GraphQL.GraphQLExtensions.AstFromValue` because it didn't actually convert the values to the various value node types that were added.

Lastly, some of the types didn't specify a proper name, causing the schema types to be very verbose.

### DIFF
--- a/src/GraphQL.Tests/Execution/RegisteredInstanceTests.cs
+++ b/src/GraphQL.Tests/Execution/RegisteredInstanceTests.cs
@@ -138,6 +138,8 @@ scalar DateTimeOffset
 
 scalar Decimal
 
+scalar Guid
+
 # The `Milliseconds` scalar type represents a period of time represented as the total number of milliseconds.
 scalar Milliseconds
 
@@ -151,6 +153,14 @@ type root {
 
 # The `Seconds` scalar type represents a period of time represented as the total number of seconds.
 scalar Seconds
+
+scalar Short
+
+scalar UInt
+
+scalar ULong
+
+scalar UShort
 ");
         }
 
@@ -177,6 +187,8 @@ scalar DateTimeOffset
 
 scalar Decimal
 
+scalar Guid
+
 # The `Milliseconds` scalar type represents a period of time represented as the total number of milliseconds.
 scalar Milliseconds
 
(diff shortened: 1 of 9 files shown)


mvestergaard:
The failing test was failing before any of my changes too.
`SchemaPrinterTests.prints_multiple_interfaces` it seems to be due to the order of the output interfaces:
Expected:
```
interface Baaz {
  int: Int
}

type Bar implements Foo, Baaz {
  str: String
}
```

Actual:
```
type Bar implements Foo, Baaz {
  str: String
}

interface Baaz {
  int: Int
}
```

mvestergaard:
I'm confused as to whether I'm doing something wrong, or there are general issues with scalar types.
I tried using the `DateTimeGraphType`, and I get similar issues:

For context, this is the exception. It's trying to resolve the scalar type from my DI container (Autofac) because the scalar type isn't registered by default:
```
GraphQL.ExecutionError: The requested service 'GraphQL.Types.DateTimeGraphType' has not been registered. To avoid this exception, either register a component to provide the service, check for service registration using IsRegistered(), or use the ResolveOptional() method to resolve an optional dependency. --
-> Autofac.Core.Registration.ComponentNotRegisteredException: The requested service 'GraphQL.Types.DateTimeGraphType' has not been registered. To avoid this exception, either register a component to provide the service, check for service registration using IsRegistered(), or use the ResolveOptional() method
 to resolve an optional dependency.
```

aoakeson:
I attempted to use ShortGraphType, and was getting Expected non-null value, resolve delegate return null for "$GraphQL.Types.ShortGraphType".  I added it to the DI container, and it started working.  Is this by design? Or should it work like the other scalar types like string. Imo I shouldnt have to register these custom types.

joemcbride:
@aoakeson Key word is "custom".  Any custom GraphType has to be registered with the Schema for it to know about it (including Scalars).  Depending on your setup, that could either be by your DI container or calling the `RegisterType` on `Schema`.  Whether this framework does it for you or not for these specific types, it has to be done.  Right now this framework is not registering them by default.  This PR changes that.

aoakeson:
Thanks for the explanation @joemcbride. It all makes sense.  Im pretty new to this project, but I'm in the process of converting my companys web api to using this. Its been great to learn and work with.  Hope to be able to contribute in the future.  Huge thumbs up.

joemcbride:
Closing in favor of #1074 which fixes the merge issues.

PR #927, opened 2018-11-27
Removed the unnecessary DocumentWriter method.

DSilence:
The initial idea was to use this method for websocket implementation. It did, however, have some design limitations
- The was a limit for max response
- The code was quite complex
As a result, the current websocket implementation planned (https://github.com/graphql-dotnet/server/pull/174) uses stream-based approach, making this method obsolete.


joemcbride:
Unfortunately we can't remove this immediately because it has now been shipped.  We need to mark it as obsolete and remove it in the next major version.

DSilence:
Gotcha, will mark it as obsolete for now in a separate pull request.

SimonCropp:
will close this one for now. on the next major all `[obsoletes]` will be reviewed and cleaned up. 

@DSilence  thanks for the other PR

PR #934, opened 2018-11-29
update newtonsoft to v12

SimonCropp:


johnrutherford:
I don’t see a reason to do this unless there is a feature in v12 that is actually needed.

SimonCropp:
mostly to prevent the binding redirect hassle

johnrutherford:
The problem I see is that it’s going to force users to update, rather than allowing them to upgrade to v12 if they want.

PR #936, opened 2018-11-30
Updateed introduction.md

GeXiaoguo:
having an instance method of schema(e.g. Execute) taking the schema instance as argument is confusing. Why do I have to pass the schema instance twice( once as this and again as Schema property)


joemcbride:
Whoops!  Thanks!

Issue #758, opened 2018-08-01
Additional Documentation


joemcbride:
* [ ] Fields - additional info around how everything is a field + resolver
* [ ] Enumerations
* [ ] `DocumentExecutor` + `DocumentWriter`
* [ ] Custom Scalars
* [ ] Relay, with pagination
* [ ] File Upload Example

fiyazbinhasan:
have plans to take a look at relay. will notify you soon :)

tsteele:
If the File Upload needs an example, doe that mean it is supported in release 2.0.?

joemcbride:
@tsteele it has always been “supported”, as in you are able to do it.  The GraphQL spec however does not specify handling files.  So there is nothing built-in specifically for file handling.  It is entirely possible to do it though.

gabrielemancofi:
@joemcbride Is it possible to have an example on how to upload a file?

SimonCropp:
@SirwanAfifi @gabrielemancofi @tsteele this might help you https://github.com/SimonCropp/GraphQL.Attachments

SimonCropp:
closing this since it is outside the scope of graphql-dotnet. the graphql spec doesnt define how attachments should be done, and graphql-dotnet does not stop you from using any of the suggested workarounds.

sungam3r:
@Shane32 you may be interested in this. I think this issue was partially solved.

PR #941, opened 2018-12-01
Add documentation for custom scalars (#758)

andrewtyped:
- Describe concepts required to implement a custom scalar
- Provide a walkthrough covering all aspects of creating a custom scalar

Let me know if you think these docs can be simplified. I feel the verbosity is warranted because there's some nuance in the process of defining a custom scalar - it took me a while to understand why all `ScalarGraphType` methods usually lead to `ParseValue`.


joemcbride:
Hey, thanks for this - I haven't done an in-depth review, though I would prefer that this custom scalars be in its own document instead of in the `schema-types` doc.  Mind adding that to its own file?

abarger-bss:
Sure. I was originally planning to make a separate PR which broke all the sections under `Schema Types` out into their own files for an easier diff, but I can do that now.

andrewtyped:
@joemcbride Hey, I started on the requested change but wanted to host the docs locally to make sure I got the site yaml right. I'm having problems with the output of the gatsby build - I can build and host the docs locally but every page I try to visit results in `TypeError: Cannot read property 'jsonName' of undefined`. I'm doing this on Windows, do you use a different platform? I get the same issue even when building docs for a fresh clone of the main repo.

PR #946, opened 2018-12-06
Improved data loader to allow resolvers to use several awaits on them.

Cooke:


johnrutherford:
@Cooke, I agree that this is an issue that should be fixed, but I think there may be a simpler way to fix it. We should also avoid `async void` methods.

Can you explain exactly how this fixes the issue? 

Maybe we could do something like putting a loop inside `DispatchAllAsync()` and keep calling `DispatchAsync()` on the loaders as long as they have more requests to dispatch (e.g., make `IsFetchNeeded()` public).

Cooke:
To be honest I have written my own data loader so I have very little interest in the included one. 

Looking at my second commit again (modifying the data loader) I agree that it is less than optimal. 

Just scrap my second commit but please merge my first. This would allow any data loader implementation to better batch load requests: "buffer" up requests until all current resolvers have awaited/finished and then start processing then pause again when all resolvers have finished.

Your suggestion "looping inside DispatchAll.. until all done" wont work on its own since you don't know when all is done. You need to continue to process load requests until the all the fields have resolved.

johnrutherford:
Adding a method to the `IDocumentExecutionListener` interface is a breaking change. We want to avoid breaking changes until the next major version. 

So I’d like to see if it’s possible to fix this issue another way.

Cooke:
Ok. I'm sorry but I cannot put more time into this now. It seems like my data loader work good enough without pausing. You can close this.

PR #950, opened 2018-12-12
IsNullable should include strings

NathanMagnus:
When using `AutoRegisteringObjectGraphType`, strings are being registered as non-nullable. This fixes this issue.


SimonCropp:
Can a unit test be added to this, to ensure we dont have a regression in the future

NathanMagnus:
Updated to include some tests. The list of tested types is not exhaustive. If you prefer a different naming scheme please let me know.

PR #967, opened 2019-01-11
Proposal: update target test FWs

SimonCropp:
IMO test suites should target the newest current release of supported runtimes. In this case that would be netcoreapp2.2 and net472. at the moment there is no consistency in the sln. eg `GraphQL.Tests.csproj` uses`netcoreapp2.1;netcoreapp2.0;netcoreapp1.1;net452`. ie 3 non-current netcore frameworks and one non-current desktop framework

### DIFF
--- a/src/GraphQL.DataLoader.Tests/GraphQL.DataLoader.Tests.csproj
+++ b/src/GraphQL.DataLoader.Tests/GraphQL.DataLoader.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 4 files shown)


SimonCropp:
@joemcbride any guesses as the the test failure? it passes locally

joemcbride:
Part of the idea is that the tests should run targeting the supported runtimes. Removing 1.1 and 452 in the tests could make it so regressions could creep in (though I do think 1.1 and 45x support should be removed in the next major released).  Do you not think that is the case?

SimonCropp:
So only support net standard 2.0?

SimonCropp:
replaced by https://github.com/graphql-dotnet/graphql-dotnet/pull/969

PR #969, opened 2019-01-11
move to only netstandard2.0

SimonCropp:


SimonCropp:
@joemcbride ok with this one?

PR #973, opened 2019-01-11
Bugfix/fixsubscription

Patrick-Spiegel:
This fixes the issue that ListGraphTypes cannot be returned in subscription. 

Test is included from ztripez (#849).


src/GraphQL/Types/ListGraphType.cs
```diff
@@ -11,7 +11,7 @@ public ListGraphType()
         }
     }
 
-    public class ListGraphType : GraphType
+    public class ListGraphType : ObjectGraphType
```
SimonCropp: i assume we consider changing a base type a breaking change?

SimonCropp:
@joemcbride LGTM. IMO it is a breaking change, but given we have the netstandard change which is also breaking, i think we merge this and declare the next pending release v3

joemcbride:
Went with #989 so closing this one.

PR #976, opened 2019-01-12
Optimize ParallelExecutionStrategy: Mark2

SimonCropp:
re apply https://github.com/graphql-dotnet/graphql-dotnet/pull/891

> Based on the conversation in #889, I looked at how the `ParallelExecutionStrategy` could be improved upon. 
> 
> Currently it executes in steps where nodes are executed and then we wait for them all to complete before starting to execute the child nodes. This repeats until the edges of the graph are reached. 
> 
> The problem with this is that the currently executing nodes could finish at different times. If one node takes a long time, we have to wait for that node to complete before executing the child nodes of the nodes that have already completed. 
> 
> I made some changes so that we can start executing child nodes as soon as parent nodes complete. And this works with the DataLoader because `OnBeforeExecutionStepAwaitedAsync` is called at the appropriate times. If we can start executing child nodes as soon as parent nodes are complete, it's possible to greatly reduce execution time.
> 
> I wrote a test for this, but I know that timing/multi-threaded tests can be flaky with AppVeyor, so if it fails, I'll see if there's another way I can do it.
> 
> I still want to do a bit more work on this. I might clean up some stuff and add another test.

### DIFF
--- a/src/GraphQL/DataLoader/DataLoaderContext.cs
+++ b/src/GraphQL/DataLoader/DataLoaderContext.cs
@@ -11,7 +11,6 @@ namespace GraphQL.DataLoader
     public class DataLoaderContext
     {
         private readonly Dictionary<string, IDataLoader> _loaders = new Dictionary<string, IDataLoader>();
-        private readonly Queue<IDataLoader> _queue = new Queue<IDataLoader>();
 
         /// <summary>
         /// Add a new data loader if one does not already exist with the provided key
@@ -38,7 +37,6 @@ namespace GraphQL.DataLoader
                     loader = dataLoaderFactory();
 
                     _loaders.Add(loaderKey, loader);
-                    _queue.Enqueue(loader);
                 }
             }
 
@@ -48,37 +46,27 @@ namespace GraphQL.DataLoader
         /// <summary>
         /// Dispatch all registered data loaders
         /// </summary>
-        /// <param name="cancellationToken">Optional <seealso cref="CancellationToken"/> to pass to fetch delegate</param>
+        /// <param name="cancellationToken">Optional <seealso cref="CancellationToken"/> to pass to fetch delegates</param>
         public async Task DispatchAllAsync(CancellationToken cancellationToken = default)
         {
             Task task;
 
             lock (_loaders)
             {
-                if (_queue.Count == 0)
-                {
+                if (_loaders.Count == 0)
                     return;
-                }
-                else if (_queue.Count == 1)
-                {
-                    var loader = _queue.Peek();
-                    task = loader.DispatchAsync(cancellationToken);
-                }
-                else
-                {
-                    var tasks = new List<Task>(_queue.Count);
 
-                    // We don't want to pop any loaders off the queue because they may get more work later
-                    foreach (var loader in _queue)
-                    {
-                        tasks.Add(loader.DispatchAsync(cancellationToken));
-                    }
+                var tasks = new List<Task>(_loaders.Count);
 
-                    task = Task.WhenAll(tasks);
+                foreach (var loader in _loaders.Values)
+                {
+                    tasks.Add(loader.DispatchAsync(cancellationToken));
                 }
+
+                task = Task.WhenAll(tasks);
             }
 
-            await task.ConfigureAwait(false);
+            await task;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


johnrutherford:
I finally got the time to finish up this PR.

sungam3r:
@SimonCropp Will you continue to work on this PR? `ParallelExecutionStrategy` has changed.

sungam3r:
@SimonCropp friendly bump

lanwin:
Did this pull request allow to await a compleat graph/tree instead only one level at the same time?

I current have the problem that I want to build an aggregation graph type which build the query and then only executes it if the all leaf nodes are awaiting (and no field can be resolved anymore). Example:

```graphql
{
  aggregation: {
     totalOrders: sum(field:ORDER)
     topOrders: topTerms(field:ORDER) {
       id
       orders: count
     }
  }
}
```

Currently that produces a deadlock. `topTerms` directly resolves but `sum` returns a task that only resolves if also `id` and `count` of `topTerms` resolve. But since the current `ParallelExecutionStrategy` only moves into `topTerms` if `sum` completes I get a deadlock.

Is it correct that this scenario will work if this pull request will getting merged?

lanwin:
@SimonCropp why did you close this? Is it not longer relevant?

Shane32:
@lanwin You should be able to accomplish your goal with a custom dataloader implementation.  Nodes that return an `IDataLoaderResult` will not complete executing until all other pending nodes have been finished (whether they are async or synchronous nodes).  Then all data loader nodes complete, and the process repeats itself - all possible nodes are executed, and when none remain, then the data loader nodes execute.  This is true for both the parallel and serial execution strategy.

For example, you could have a field "amount" that obtains the price of a widget from a database, then obtains a `WidgetTotal` scoped class via dependency injection, and increases `WidgetTotal.TotalAmount` by the amount of the widget, then returns the widget price.  `WidgetTotal` would also implement `IDataLoaderResult` and upon being requested for the total, would return the total and perhaps also clear the amount.  Your "Total" field would obtain `WidgetTotal` via dependency injection and return it (since it implements `IDataLoaderResult`).  During execution of the query, all of the `amount` fields, whether async or not, and no matter how deep in the graph, would complete executing before any `total` fields finish executing.  Note that you need to be aware of possible multithreading issues within `WidgetTotal` (in the above sample, you could use the `Interlocked` class for variable access), or use a serial execution strategy to alleviate yourself of those issues.

This type of execution has been possible since the new data loader implementation was introduced with 3.0.0.  See PR #1511

Shane32:
Note that if you use typical async/await statements in your code (versus creating a perpetually pending task that has an external trigger to continue), deadlocks should not be possible.  The use of `BeforeExecutionStepAwaitedAsync` is unnecessary and discouraged.

Finally, asynchronous field resolvers can return `Task<IDataLoaderResult>` instances, which can themselves return async (`Task<T>`) or other `IDataLoaderResult` instances, and so on.  For instance, a strongly typed response from a `IntGraphType` field could be `Task<IDataLoaderResult<Task<IDataLoaderResult<Task<int>>>>>`. Execution will all run correctly, at the right time according to the description above, and without deadlocks.

Shane32:
@sungam3r I really think we need to mark `BeforeExecutionStepAwaitedAsync` as deprecated.  Any use of it should be replaced with `IDataLoaderResult<T>` or else execution is subject to deadlocks. See #1827

Shane32:
@sungam3r By the way, I know that you've not used the data loaders, but the new design is absolutely perfect for unique situations such as the one described by @lanwin above.  The solution I suggested does not use almost any of the "data loader" implementation code, except for the interface definition of `IDataLoaderResult` (a few lines of code) and some simple changes within the execution strategy.  `IDataLoaderResult` is just a way for the execution strategy to delay the processing of nodes until all other nodes have been completed.  This appears to be perfect for his requirement, even though it has nothing to do with the typical way data loaders are used.

sungam3r:
> why did you close this? Is it not longer relevant?

I deliberately put this task on my to-do list, it's not lost.

Shane32:
The significant change with this PR is the use of WhenAny instead of WhenAll. We need to remove `BeforeExecutionStepAwaitedAsync` first or it could break implementations that rely on it. I’d be happy to write a PR.

sungam3r:
> We need to remove `BeforeExecutionStepAwaitedAsync` first or it could break implementations that rely on it. 

I suspected that this method only existed because of a flawed implementation. I'm OK to remove it.

sungam3r:
> I’d be happy to write a PR.

OK. Then tell later whether it makes sense to borrow something from this PR, so that we can finally close it and move on.

Shane32:
> > We need to remove `BeforeExecutionStepAwaitedAsync` first or it could break implementations that rely on it.
> 
> I suspected that this method only existed because of a flawed implementation. I'm OK to remove it.

Then #1827 should be merged now and we can remove it for 4.0.

sungam3r:
So nothing here to worry about, right?

PR #980, opened 2019-01-15
424 When an object's non-null property resolver returns null (or exception), the 'null' isn't bubbled up to the parent 

KoditkarVedant:
Enhancement #424 and #994 

When an object's non-null property resolver returns null (or exception), the 'null' isn't bubbled up to the parent 

Fixed the test cases


joemcbride:
As was mentioned in the other PR for this, we can't just remove the schema results from `IntrospectionResult`.  The tests that use that should be updated so it doesn't return null (if that is the issue) so the test is actually still testing the introspection response.

I'm also weary of doing exception-based control flow.  This other PR seemed to be able to impliment this feature without using exceptions.  #534

KoditkarVedant:
@joemcbride if I understand you correctly, do you mean even if `NonNullable` graph type returns the `null` it should test the complete `IntrospectionResult` (as currently present?) or do you want to add new/update the test case test case with no `NonNullable` graph type in the schema?

joemcbride:
> update the test case test

This.

You _should_ still get a introspection result even with nullable types/fields.  I suspect the problem with the test is that it basically has no query/mutation/subscription.

https://github.com/graphql-dotnet/graphql-dotnet/blob/6ecd4f0980e058eba3f086b717978747fffc4325/src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs#L17

If you still get a top level null after a Query is added then there is a bug in the implementation.

KoditkarVedant:
> 
> 
> > update the test case test
> 
> This.
> 
> You _should_ still get a introspection result even with nullable types/fields. I suspect the problem with the test is that it basically has no query/mutation/subscription.
> 

I have removed the `NonNullableGraphType` from from `queryType` field. I get results as expected. There is an additional test case was failing in `SchemaPrinterTests` which I fixed.

> [graphql-dotnet/src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs](https://github.com/graphql-dotnet/graphql-dotnet/blob/6ecd4f0980e058eba3f086b717978747fffc4325/src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs#L17)
> 
> Line 17 in [6ecd4f0](/graphql-dotnet/graphql-dotnet/commit/6ecd4f0980e058eba3f086b717978747fffc4325)
>  _.Schema = new Schema(); 
> 
> If you still get a top level null after a Query is added then there is a bug in the implementation.

I will quickly check if I get null after adding the query.

joemcbride:
> I have removed the NonNullableGraphType from from queryType field

That needs to be added back as it should be NonNull, according to the GraphQL-JS implementation.

KoditkarVedant:
> 
> 
> > I have removed the NonNullableGraphType from from queryType field
> 
> That needs to be added back as it should be NonNull, according to the GraphQL-JS implementation.

I have reverted that change & reset the `queryType` as NonNull. I have added the a `TestQuery` so we can still get a introspection result.

KoditkarVedant:
@joemcbride I will add some more test cases for the change.

KoditkarVedant:
@joemcbride FYI, I have completed the changes required and as suggested added the failing test cases for not working `Non Null ListGraphType`

KoditkarVedant:
@joemcbride this PR is now ready for review. I am not sure why commits from @SimonCropp are being displayed in this PR?

joemcbride:
@KoditkarVedant I don't know either.  You will need to get the branch fixed before it can be accepted into master.

KoditkarVedant:
@joemcbride I have fixed it.

KoditkarVedant:
@joemcbride friendly ping!

KoditkarVedant:
@joemcbride we shall close this issue?

PR #981, opened 2019-01-16
Fix schema printing of multiple interfaces

freakingawesome:
The [GraphQL Schema Documentation](https://facebook.github.io/graphql/draft/#sec-Interfaces) calls for separating multiple interfaces by an ampersand rather than a comma.

This allows for Apollo code generation based on the schema, which currently fails because interfaces are separated by commas.

### DIFF
--- a/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
@@ -390,7 +390,7 @@ interface Baaz {
   int: Int
 }
 
-type Bar implements Foo, Baaz {
+type Bar implements Foo & Baaz {
   str: String
 }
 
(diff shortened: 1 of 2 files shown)


joemcbride:
Hey, thanks for this.  One thing I want to make sure is that the GraphQL Parser can also handle that ampersand.  Otherwise this printer will be producing something that the parser can’t parse.

freakingawesome:
Good call. I took a quick stab at [updating the parser here](https://github.com/freakingawesome/parser/commit/0549ee03043cee3e9fee4affc0b75adda4a3aaa0), and while it works I wonder if it is too naive, since it just treats `&` as a comma-equivalent separator everywhere (if I'm reading this right). If you're ok with that update I'll open another PR for the parse.

joemcbride:
I'm not sure how it happened, but there were two PRs for this same thing on the same day (do you two work together??).  I like this other implementation since it has a flag for the old syntax.

https://github.com/graphql-dotnet/graphql-dotnet/pull/983

freakingawesome:
Ha! No, we don't work together. Just a coincidence!

Looks like that PR also addresses the other PR I had opened (also on the same day) for fixing the documentation descriptions (#982)

I still think we'll need to update the parser to allow reading of `&`-separated interfaces to make sure graphql-dotnet can read in the same format it generates.

joemcbride:
Closing this in favor of #983.

PR #982, opened 2019-01-16
Fix schema printing of field annotations

freakingawesome:
When including descriptions and deprecation reasons on fields in SchemaPrinter, the current code reprints the annotations on the parent type, rather than the annotations from the fields themselves. This fixes the problem by printing the annotations on the fields.


freakingawesome:
While this fixes the printing of field annotations specifically, there are further inconsistencies with how printing annotations is done which I've started addressing in [another branch](https://github.com/graphql-dotnet/graphql-dotnet/compare/master...freakingawesome:patch-4?expand=1), but that will involve a lot of changes to the utilities tests and I haven't yet had time to fix all of them.

joemcbride:
Closing this in favor of #983.

PR #989, opened 2019-01-23
Fix subscribing to non object graph types

rose-a:
This is an alternative PR to #973 which fixes issue #770 without a breaking change.

Furthermore this also allows returning scalar values from subscriptions (which would still not work with #973)

Tests from #849 and a new test for the scalar return value are included.


src/GraphQL/Execution/ParallelExecutionStrategy.cs
```diff
@@ -6,7 +6,10 @@ namespace GraphQL.Execution
 {
     public class ParallelExecutionStrategy : ExecutionStrategy
     {
-        protected override async Task ExecuteNodeTreeAsync(ExecutionContext context, ObjectExecutionNode rootNode)
+        protected override Task ExecuteNodeTreeAsync(ExecutionContext context, ObjectExecutionNode rootNode)
```
Patrick-Spiegel: Is there a reason why this Async method is not awaitable?

rose-a: It sure is awaitable. It just forwards to the implementation below which accepts an `ExecutionNode` instead of an `ObjectExecutionNode`.

Since there is no `await` used in forwarding to the other method, you don`t need to use the `async` keyword. You just return the async Task created by the forwarded method.

gregbty:
Any time table for this to be released (as well as the server package being updated with it)? We are using the server package and we are running into this issue and there is no workaround.

joemcbride:
This fix is published in the core library in GraphQL 3.0.0-preview-1107.  I will work on getting the server project updated.

ShmulikBoros:
> This fix is published in the core library in GraphQL 3.0.0-preview-1107. I will work on getting the server project updated.

Any update regarding the incoming fix ?

PR #993, opened 2019-01-24
stricter date parsing

SimonCropp:
Not sure on this one. DateGraphType, will always return  "yyyy-MM-dd" but it will (previously) accept any date format  as show by the failing tests. eg `2015-12-01T14:15:07.123Z`. So it seems weird to me that someone can use the same graph type and pushing a string up will come back as a diff format and with less resolution (data loss)


SimonCropp:
i should also point out that the current date parsing seems to be susceptible to ambiguous dates.  eg `09-10-2015")`. which i didnt think should be allowed under ISO_8601?

SimonCropp:
@joemcbride thoughts on this one. If we r going to change it we should do it as part of the major

SimonCropp:
@joemcbride gentle bump

SimonCropp:
@sungam3r thoughts?

sungam3r:
This is one of those issues that I was particularly interested in but still can not find the time to look. I will try to read in the near future.

src/GraphQL/Types/DateGraphType.cs
```diff
@@ -17,13 +18,27 @@ public override object Serialize(object value)
...
 
             return null;
         }
 
-        public override object ParseValue(object value) => ValueConverter.ConvertTo(value, typeof(DateTime));
+        public override object ParseValue(object value)
+        {
+            if (value is DateTime dateTime)
+            {
+                return dateTime;
```
sungam3r: Wouldn’t it be right to take only the part responsible for the date? Or it’s generally worth generating an error if the format of the input data does not match the required format for `Date`. I believe it is wrong to allow the use of a date with a time component for a date scalar

src/GraphQL/Types/DateGraphType.cs
```diff
@@ -17,13 +18,27 @@ public override object Serialize(object value)
...
+            }
+
+            var valueAsString = (string)value;
```
sungam3r: consider to change this to `if (value is string valueAsString)`

src/GraphQL/Types/DateGraphType.cs
SimonCropp: fixed to throw an exception

src/GraphQL/Types/DateGraphType.cs
SimonCropp: yep. agreed. done

src/GraphQL/Types/DateGraphType.cs
```diff
@@ -17,13 +18,34 @@ public override object Serialize(object value)
...
+            if (value is string valueAsString)
+            {
+                if (DateTime.TryParseExact(valueAsString, "yyyy-MM-dd", DateTimeFormatInfo.InvariantInfo, DateTimeStyles.None, out var date))
+                {
+                    return date;
+                }
+                throw new FormatException($"Could not parse date. Expected yyyy-MM-dd. Value: {valueAsString}");
+            }
+
+            throw new FormatException($"Could not parse date. Expected either a string or a DateTime. Value: {value}");
```
sungam3r:
```suggestion
            throw new FormatException($"Could not parse date. Expected either a string or a DateTime without time component. Value: {value}");
```

sungam3r:
So you still need to fix the tests that work with the date.

src/GraphQL.Tests/Types/DateGraphTypeTests.cs
```diff
@@ -20,52 +20,12 @@ public void serialize_string_to_date()
             });
         }
 
-        [Fact]
-        public void serialize_local_date_returns_date_only()
```
sungam3r: Don't you think that we should to keep these tests, just now expect an exception from them about time component?

sungam3r:
@SimonCropp Will you continue to work on this PR? I would like these changes to be merged.

sungam3r:
@SimonCropp friendly bump

SimonCropp:
@sungam3r thanks for the reminder. this should be good now. but it will need a squash on merge

sungam3r:
@Shane32 This will need to be noted in the documentation as breaking change for Data scalar type.

Shane32:
Sure, I'll add it to the upgrade notes.  Correct me if I'm wrong, but looks like now `yyyy-MM-dd` is the only valid format when deserializing, and when serializing, the `DateTime` must not have a time component.

sungam3r:
Right.

SimonCropp:
yep. at my current employer we have been running with a custom date type for over a year now. will be good to remove it

nbarbettini:
> Sure, I'll add it to the upgrade notes. Correct me if I'm wrong, but looks like now `yyyy-MM-dd` is the only valid format when deserializing, and when serializing, the `DateTime` must not have a time component.

@Shane32 Friendly reminder to include this in your upgrade PR: https://github.com/graphql-dotnet/graphql-dotnet/pull/1798/
It caused a bit of head scratching for me during a 2.4->3.0 upgrade.

Shane32:
Thanks for the reminder @nbarbettini !

sungam3r:
> It caused a bit of head scratching for me during a 2.4->3.0 upgrade.

I also had to explain possible problems to our teams. Problems did arise.

PR #997, opened 2019-01-26
994 bubble up null in list graph type

### LINKED ISSUE #994: List Graph Type and Non Nulls
I was going through `graphql` spec for `List and NonNull` and noticed that as of now we do not handle it in our library.
```
myField: [String!]
```
This means that the list itself can be null, but it can't have any null members. For example, in JSON:
```
myField: null // valid
myField: [] // valid
myField: ['a', 'b'] // valid
myField: ['a', null, 'b'] // error
```
Now, let's say we defined a Non-Null List of Strings:
```
myField: [String]!
```
This means that the list itself cannot be null, but it can contain null values:
```
myField: null // error
myField: [] // valid
myField: ['a', 'b'] // valid
myField: ['a', null, 'b'] // valid
```
[ref](https://graphql.github.io/learn/schema/#lists-and-non-null)

I have already added the failing test cases while working on #424 enhancement.

KoditkarVedant:
Fix #994   ---  Dependent on PR #980 
-----------
- [x] Handle NonNull with List
- [x] Add test cases
- [x] Fix failing test cases added in #980 PR


KoditkarVedant:
Merged into #980 PR itself.

PR #1010, opened 2019-02-11
Fix derived type lookup in GraphTypesLookup

JoshuaDickson:
Previously, `this[Type type]` in GraphTypesLookup only got a resolver if the type given was exactly that type. This doesn't work if it's looking for a base type where you have a derived type registered.

When a schema is discovering abstract types, ordinarily it'll just fail to activate them. But with a resolver, it will retrieve the type, but fail to see the type is in the GraphTypesLookup, and attempt to add it again. Without a cycle, this isn't an issue - but with a cycle, this will recurse until your program crashes with a StackOverflowException.

Now, `this[Type type]` checks to see if any type can be assigned to the type. To avoid ambiguity, it uses SingleOrDefault now, rather than FirstOrDefault.

### DIFF
--- a/src/GraphQL/Types/GraphTypesLookup.cs
+++ b/src/GraphQL/Types/GraphTypesLookup.cs
@@ -137,7 +137,7 @@ namespace GraphQL.Types
             {
                 lock (_lock)
                 {
-                    var result = _types.FirstOrDefault(x => type.IsAssignableFrom(x.Value.GetType()));
+                    var result = _types.SingleOrDefault(x => type.IsAssignableFrom(x.Value.GetType()));
                     return result.Value;
                 }
             }
--- a/src/GraphQL/Types/GraphTypesLookup.cs
+++ b/src/GraphQL/Types/GraphTypesLookup.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using GraphQL.Conversion;
 using GraphQL.Introspection;
 
@@ -136,7 +137,7 @@ namespace GraphQL.Types
             {
                 lock (_lock)
                 {
-                    var result = _types.FirstOrDefault(x => x.Value.GetType() == type);
+                    var result = _types.FirstOrDefault(x => type.IsAssignableFrom(x.Value.GetType()));
                     return result.Value;
                 }
             }
--- a/src/GraphQL.Tests/Types/GraphTypesLookupTests.cs
+++ b/src/GraphQL.Tests/Types/GraphTypesLookupTests.cs
@@ -9,6 +9,9 @@ namespace GraphQL.Tests.Types
     {
         private class TestType : GraphType { }
 
+        private class BaseTestType : GraphType { }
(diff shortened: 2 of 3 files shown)


joemcbride:
Looking at this, I don't believe that this would be behavior that we want to support.  Can you provide a better example of how/why you would want this behavior?

JoshuaDickson:
That's fair enough, it is a bit niche behavior.

In my project, I have a code generator that takes a GraphQL schema, and emits graphql-dotnet scaffolding in the form of abstract base classes. This allows me to ensure the ASP.NET Core server and the declarative schema are always in sync. It catches most, but not all errors at compile time, and then a handful more at schema creation in runtime. It's been very useful for me in developing, as I've reduced the boilerplate needed to mimic a schema down to just writing the resolver function, with full type safety (for queries at least - mutations need a bit of work).

I'll likely be open sourcing the generator once I get it in a bit better shape.

sungam3r:
This PR can lead to unpredictable consequences. In essence, this is a backwards incompatible change that alters `GraphTypesLookup` contract. Is there a need to change exactly the existing method? Can the desired goal be achieved with a new (different) method?

sungam3r:
@JoshuaDickson I finally managed to find the time to sort out your case. Honestly, the ability to create a schema with cyclic dependencies that instead of one type generate another type in the resolver is a sort of bug. But we cannot prohibit the creation of such schemas. In fact, cyclic dependencies themselves work fine. In your case, the peculiarity is the use of an abstract type, which you had to make you a resolver that returns an instance of another type. This really leads to `StackOverflowException`.

I reproduced the error and thought for a long time how to get around it. In the end, I decided that the best solution would be to simply determine the occurrence of the loop and generate a specific error. In most cases, this situation is evidence of an incorrect schema configuration.

The approximate error text is
> 'A loop has been detected while registering schema types.
There was an attempt to re-register 'GraphQL.Tests.Types.AbstractGraphType' with instance of 'GraphQL.Tests.Types.ConcreteGraphType'.
Make sure that your ServiceProvider is configured correctly.'

I understand that you are interested in fixing the error in the sense that the schema can be created and used. But as you yourself say "it is a bit niche behavior". So I would rather fix `StackOverflowException`, which I think will be useful to everyone.

PR will be soon.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -136,7 +137,7 @@ public IEnumerable<IGraphType> All()
             {
                 lock (_lock)
                 {
-                    var result = _types.FirstOrDefault(x => x.Value.GetType() == type);
+                    var result = _types.SingleOrDefault(x => type.IsAssignableFrom(x.Value.GetType()));
```
sungam3r: But then the question arises - what to do if there are several child types? All this seems to be an attempt to correct a design flaw in the wrong place.

sungam3r:
Closed in favor of #1421 . @JoshuaDickson I made several more attempts to understand the usefulness of your proposal. Nevertheless, it seems to me that it will bring more complexity than good. I added this problem to the list of problems that require documentation - #1326 .

PR #1026, opened 2019-03-01
Added Execution option, Added logic to overwrite Execution Strategy, …

JasonKitson:
Added Execution option, Added logic to overwrite Execution Strategy, Added unit tests.

This is to overcome a threading issue.

### DIFF
--- a/src/GraphQL.Tests/Execution/EnableSerialExecutionOnQuery.cs
+++ b/src/GraphQL.Tests/Execution/EnableSerialExecutionOnQuery.cs
@@ -1,0 +1,66 @@
+using System.Threading;
+using GraphQL.Conversion;
+using GraphQL.Http;
+using GraphQL.Tests.StarWars;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace GraphQL.Tests.Execution
+{
+    public class EnableSerialExecutionOnQuery : StarWarsTestBase
+    {
+        private readonly ExecutionOptions _executionOptions;
+
+        public EnableSerialExecutionOnQuery()
+        {
+            _executionOptions = new ExecutionOptions()
+            {
+                SetFieldMiddleware = false,
+                EnableMetrics = false,
+                Schema = Schema,
+                Query = @"
+               {
+                  human(id: ""1"") {
+                    name
+                    friends {
+                      name
+                      appearsIn
+                    }
+                  }
+               }
+            ",
+                Root = null,
+                Inputs = null,
+                UserContext = null,
+                CancellationToken = default(CancellationToken),
+                ValidationRules = null,
+                FieldNameConverter = new CamelCaseFieldNameConverter(),
+                EnableSerialExecutionOnQuery = true
+            };
+        }
+
+        [Fact]
+        public async void EnableSerialExecutionOnQuery_NoError()
+        {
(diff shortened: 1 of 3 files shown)


JasonKitson:
Hi Joe,
Thanks for the response.
I am aware that the IExecutionStrategy SelectExecutionStrategy() method is virtual and can be overridden.
However, that requires that users of this product have a strong understanding of the codebase. 
The reason behind this change was due to a threading issue. We are working with a stack that includes GQL.NET and Entity Framework. In one of out Graphs there are 2 properties that both require DatabaseContext (EF). These are competing in the parallel and results in this exception:

Message: System.InvalidOperationException : A second operation started on this context before a previous operation completed. This is usually caused by different threads using the same instance of DbContext, however instance members are not guaranteed to be thread safe. This could also be caused by a nested query being evaluated on the client, if this is the case rewrite the query avoiding nested invocations.
Understanding how this relates back to an Execution Strategy isn’t trivial, this is very challenging to unpack.
I was given guidance by a peer of what the likely cause of the threading issue and how it could be solved by making changes to GraphQL-dotnet. This is a luxury that many developers do not have. 

In my opinion it's much easier to set an options boolean than implement an override solution. As the boolean solution hides the complexity.

Solutions like this -     

public class myDocumentExecutor : DocumentExecuter
    {
        {
            if (context.Operation.OperationType == OperationType.Query)
            {
                return new SerialExecutionStrategy();
            }
            return base.SelectExecutionStrategy(context);
        }
    }

...

   		var docExec = new myDocumentExecuter();
            	var result = await docExec.ExecuteAsync(_executionOptions).ConfigureAwait(false);

Versus this - 


   public EnableSerialExecutionOnQuery()
        {
            _executionOptions = new ExecutionOptions()
            {
		...
                EnableSerialExecutionOnQuery = true
            };
        }

...

   		var docExec = new DocumentExecuter();
            	var result = await docExec.ExecuteAsync(_executionOptions).ConfigureAwait(false);
  

There are quite a few developers who have raised issues around this problem (in the issue register) and they may well appreciate a simpler work around than what is presently available.
I appreciate your concerns about not wanting to make ExecutionOptions anymore complex. There's certainly a lot going on within the ExecuteAsync method.
However, these are execution options and they're doing exactly what they're supposed to, I don't see that this adds any great complexity to the status quo.

joemcbride:
This is trying to fix the problem you’re having improperly.  Instead of changing the execution strategy to work with your code that is not thread safe, fix your code so that it is thread safe.  That can easily be done with a factory method.  This is trying to provide a work-around, not fixing the core problem.

I admit that the re-use of GraphTypes for the lifetime of the Schema is causing some people headaches.  This is not the right fix for that problem.  Not everyone uses EF.  This will instead create another api surface that I have to support with differing logic that can already be changed through inheritance.  What if someone wants to change the behavior of mutations?  Do we add another flag?  What about subscriptions?  It sets a presidence on the design of the api that I don’t want to support.

RaySheikh:
public class EfDocumentExecuter : DocumentExecuter
{
    {
        if (context.Operation.OperationType == OperationType.Query)
        {
            return new SerialExecutionStrategy();
        }
        return base.SelectExecutionStrategy(context);
    }
}

And calling it in my controller
Var docExec

PR #1030, opened 2019-03-02
fix typos InvalidNumberOfRootFieldMessaage

SimonCropp:
@joemcbride this is a public method. so good to fix for v3. do u want me to re-add the old method with an obsolete attribute?


joemcbride:
This was just merged today, so fine with just renaming it.

PR #1033, opened 2019-03-02
remove obsolete types

SimonCropp:
@joemcbride give we r going v3. should we take the opportunity to clean up some obsoletes?

let me know if u want a diff pr with obsolete attributes as error, and implementations stubbed out


joemcbride:
I think this is fine.

PR #1037, opened 2019-03-04
add context dictionary

### LINKED ISSUE #637: How to provide more custom property values to resolvers?
Current solution is to set UserContext but you can only have one.

I propose adding a property bag into the options and pass it along until it reaches the resolvers.  Provide simple Set(key, value), Get<T>(key) for manipulation. This requirement is coming from the server side as currently I'm hijacking the UserContext to pass the MessageHandlingContext to resolvers so I can access it there.

SimonCropp:
when writing extensions for for graphql, it is really helpful to have a dictionary that can be passed through all contexts

fixes #637

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -153,7 +153,8 @@ namespace GraphQL
                     options.CancellationToken,
                     metrics,
                     options.Listeners,
-                    options.ThrowOnUnhandledException);
+                    options.ThrowOnUnhandledException,
+                    options.ContextBag);
 
                 if (context.Errors.Any())
                 {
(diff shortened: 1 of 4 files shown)


SimonCropp:
i would prefer to remove usercontext and have only the dictionary, but didnt want to introduce a breaking change

SimonCropp:
@joemcbride any objections to merging this one?

joemcbride:
Not I'm in love with the name of the field, but this is certainly needed.  `UserContext` was added to be in line with what the JS reference implementation uses.

Looks like another merge created a conflict.  Mind resolving that?  Then I'll get this merged.

joemcbride:
Fixes #637

SimonCropp:
Yep. Happy to rebase. But re the name, one alternative would be to change usercontext to a dictionary

joemcbride:
I'm actually liking that idea quite a bit ... that is closer to the JS idea, since in JS you can just use it like a dictionary.

joemcbride:
Especially since with adding another property, they would both be doing the same thing.

SimonCropp:
Ok. I will do a new pr

SimonCropp:
@joemcbride do u want it as a ConcurrentDictionary, a Dictionary, or an IDictionary?

Issue #1047, opened 2019-03-08
Bad exception text on wrong return type in field resolver


pavel-usachev:
## Description

When resolve field function returns unexpected type I got not really straightforward exception message.

### Steps to reproduce

Using following schema configuration:

```C#
public sealed class BugSchema : Schema
{
    public BugSchema()
    {
        var query = new ObjectGraphType();
        query.Field<BugType>("query", resolve: _ => new TypeB()); // Here's a misuse
        Query = query;
    }
}

public sealed class TypeA { }

public sealed class TypeB{ }

public sealed class BugType : ObjectGraphType<TypeA>
{
    public BugType()
    {
        Field<StringGraphType>("test", resolve: _ => null);
    }
}
```
Send query: `{ query { test } }`

### Expected result

I expect to see following exception message:

`Expected value of type \"GraphQL.Tests.Bugs.WrongExceptionText+TypeA\" for \"BugType\" but got: GraphQL.Tests.Bugs.WrongExceptionTextTest+TypeB.`

### Actual result

I got exception like this:

`Expected value of type \"BugType\" for \"BugType\" but got: GraphQL.Tests.Bugs.WrongExceptionTextTest+TypeB.`

joemcbride:
To expand on this a bit, what type is "allowed" is determined by `IsTypeOf` or `ResolveType`.

https://graphql-dotnet.github.io/docs/getting-started/interfaces#istypeof
https://graphql-dotnet.github.io/docs/getting-started/interfaces#resolvetype

Either of those functions could be checking multiple Domain types.  So the intent is not to be checking a specific Domain type, but to determine which GraphType to use.  So the messaging should probably be updated to reference that vs. trying to guess at what types are valid.

pavel-usachev:
Thank you! That's way better.

PR #1048, opened 2019-03-08
#1047 exception message fix on bad return type

d3vi1h3aRt:
#1047 Issue fix


src/GraphQL/Types/GraphQLTypeReference.cs
```diff
@@ -11,6 +11,8 @@ public GraphQLTypeReference(string typeName)
             TypeName = typeName;
         }
 
+        public Type SourceType => throw new InvalidOperationException("This is just a reference.  Resolve the real type first.");
```
sungam3r: double whitespace

src/GraphQL/Types/ObjectGraphType.cs
```diff
@@ -7,6 +7,7 @@ namespace GraphQL.Types
 {
     public interface IObjectGraphType : IComplexGraphType, IImplementInterfaces
     {
+        Type SourceType { get; }
```
joemcbride: Not every `IObjectGraphType` is going to have a `SourceType` that is helpful so I'm not sure that this is the direction that we should take.

PR #1051, opened 2019-03-08
Add link to article on cursor based pagination and connections to links

corstian:
I just finished writing an article on cursor based pagination and connections using this library.

Figured it might be useful :)


corstian:
For convenience; this is the link: http://corstianboerman.com/2019-03-08/implementing-pagination-with-graphql-net-and-relay.html

joemcbride:
Thanks!

PR #1062, opened 2019-03-14
System.ComponentModel attributes support

### LINKED ISSUE #1061: Support for Description, DeprecationReason and DefaultValue from System.ComponentModel attributes
## Summary

It would be nice to be able to automatically fill in these properties of the schema fields from System.ComponentModel attributes (if any applied).

```c#
public class Data
{
    [Description("Count of items")]  // set to FieldType.Description
    [Obsolete("Use something else")] // set to FieldType.DeprecationReason
    [DefaultValue(100)]              // set to FieldType.DefaultValue
    public string Count { get; set; }
}
```

### Motivation

Easier schemа. You do not need to litter it with settings that can naturally be set through attributes.

### Restrictions

This only applies to the way the field is specified via an property expression. Explicit setting of these properties will occur (if any) after initial initialization with the values from the attributes and overwrite them.

sungam3r:
fixes #1061

### DIFF
--- a/src/GraphQL/Builders/FieldBuilder.cs
+++ b/src/GraphQL/Builders/FieldBuilder.cs
@@ -85,6 +85,12 @@ namespace GraphQL.Builders
             return this;
         }
 
+        internal FieldBuilder<TSourceType, TReturnType> DefaultValue(object defaultValue)
+        {
+            _fieldType.DefaultValue = defaultValue;
+            return this;
+        }
+
         public FieldBuilder<TSourceType, TReturnType> Resolve(IFieldResolver resolver)
         {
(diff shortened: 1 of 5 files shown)


src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -85,6 +85,12 @@ private FieldBuilder(EventStreamFieldType fieldType)
             return this;
         }
 
+        internal FieldBuilder<TSourceType, TReturnType> DefaultValue(object defaultValue)
```
sungam3r: Just for internal use in fluent style.

PR #1069, opened 2019-03-17
GraphQLTypeReference for union possible types

### LINKED ISSUE #1068: ExecutionError if union type goes ahead its components
## Description

**GraphQL.ExecutionError : Must provide types for Union Pet.** in `GraphTypesLookup` as a result of the `Schema.AllTypes` call after reading schema with `Schema.For`.

### Steps to reproduce

Take this schema:
```graphql
type Query {
    animal: Pet
}

type Cat {
  name: String!
}

type Dog {
  age: Int!
}

union Pet = Cat | Dog
```

The `Schema.AllTypes` call on it will work fine. But let's change the order of the types:
```graphql
union Pet = Cat | Dog

type Query {
    animal: Pet
}

type Cat {
  name: String!
}

type Dog {
  age: Int!
}
```
Now we have **GraphQL.ExecutionError : Must provide types for Union Pet.**

### Expected result

The `Schema.AllTypes` method should work without errors.

### Actual result

**GraphQL.ExecutionError : Must provide types for Union Pet.**

### Environment

Windows 10 x64

sungam3r:
fixes #1068

### DIFF
--- a/src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
@@ -1,10 +1,12 @@
-using System;
-using System.Linq;
 using GraphQL.Http;
 using GraphQL.Utilities;
 using GraphQLParser.Exceptions;
 using Newtonsoft.Json.Linq;
 using Shouldly;
+using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
 
 namespace GraphQL.Tests.Utilities
 {
@@ -68,6 +70,11 @@ namespace GraphQL.Tests.Utilities
                 expected = JObject.Parse(result);
             }
             return new ExecutionResult { Data = expected };
+        }
+
+        protected string ReadSchema(string fileName)
+        {
+            return File.ReadAllText(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "Files", fileName));
         }
     }
 
--- a/src/GraphQL/StringExtensions.cs
+++ b/src/GraphQL/StringExtensions.cs
@@ -119,6 +119,8 @@ namespace GraphQL
         /// <summary>
         /// Returns a camel case version of the string.
         /// </summary>
+        /// <param name="s">The source string.</param>
+        /// <returns>System.String.</returns>
         public static string ToCamelCase(this string s)
         {
             if (string.IsNullOrWhiteSpace(s))
@@ -126,13 +128,17 @@ namespace GraphQL
                 return string.Empty;
             }
 
-            return $"{char.ToLowerInvariant(s[0])}{s.Substring(1)}";
+            var newFirstLetter = char.ToLowerInvariant(s[0]);
+            if (newFirstLetter == s[0])
+                return s;
+
+            return newFirstLetter + s.Substring(1);
         }
 
         /// <summary>
         /// Returns a pascal case version of the string.
         /// </summary>
-        /// <param name="s">The s.</param>
+        /// <param name="s">The source string.</param>
         /// <returns>System.String.</returns>
         public static string ToPascalCase(this string s)
         {
@@ -141,7 +147,11 @@ namespace GraphQL
                 return string.Empty;
             }
 
-            return $"{char.ToUpperInvariant(s[0])}{s.Substring(1)}";
+            var newFirstLetter = char.ToUpperInvariant(s[0]);
+            if (newFirstLetter == s[0])
+                return s;
+
+            return newFirstLetter + s.Substring(1);
         }
 
         /// <summary>
--- a/src/GraphQL/Utilities/SchemaBuilder.cs
+++ b/src/GraphQL/Utilities/SchemaBuilder.cs
@@ -350,7 +350,7 @@ namespace GraphQL.Utilities
 
             CopyMetadata(type, typeConfig);
 
-            var possibleTypes = unionDef.Types.Select(x => GetType(x.Name.Value));
+            var possibleTypes = unionDef.Types.Select(x => GetType(x.Name.Value) ?? new GraphQLTypeReference(x.Name.Value));
             possibleTypes.Apply(x => type.AddPossibleType(x as IObjectGraphType));
             return type;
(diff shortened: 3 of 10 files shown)


src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -350,7 +350,7 @@ protected virtual UnionGraphType ToUnionType(GraphQLUnionTypeDefinition unionDef
 
             CopyMetadata(type, typeConfig);
 
-            var possibleTypes = unionDef.Types.Select(x => GetType(x.Name.Value));
+            var possibleTypes = unionDef.Types.Select(x => GetType(x.Name.Value) ?? new GraphQLTypeReference(x.Name.Value));
```
sungam3r: I was guided by [this behavior](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/SchemaBuilder.cs#L466).

src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
```diff
@@ -69,6 +71,11 @@ public ExecutionResult CreateQueryResult(string result)
             }
             return new ExecutionResult { Data = expected };
         }
+
+        protected string ReadSchema(string fileName)
+        {
+            return File.ReadAllText(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "Files", fileName));
```
d3vi1h3aRt: Consider using `AppDomain.CurrentDomain.BaseDirectory` instead of `Assembly.GetExecutingAssembly().Location`.

src/GraphQL/StringExtensions.cs
```diff
@@ -119,20 +119,26 @@ public static Inputs ToInputs(this JObject obj)
...
             {
                 return string.Empty;
             }
 
-            return $"{char.ToLowerInvariant(s[0])}{s.Substring(1)}";
+            var newFirstLetter = char.ToLowerInvariant(s[0]);
+            if (newFirstLetter == s[0])
+                return s;
+
+            return newFirstLetter + s.Substring(1);
```
joemcbride: I would like to pull in this PR but I'm concerned that adding these additional checks will slow down these methods.  Was there a reason for adding these additional checks?  Did you do perf tests and see that this was faster?

https://www.dotnetperls.com/uppercase-first-letter

The above suggests to switch to the following for higher performance:

```csharp
char[] a = s.ToCharArray();
a[0] = char.ToUpper(a[0]);
return new string(a);
```

sungam3r: Oops, I accidentally clicked cancel review.

sungam3r: OK, for the interest I have conducted the tests.
Demo:
```c#
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;

namespace ConsoleApp6
{
    [MemoryDiagnoser]
    public class Program
    {
        [Params("someName", "SomeName")]
        public string s;

        [Benchmark(OperationsPerInvoke = 16)]
        public string ToCamelCaseOld()
        {
            if (string.IsNullOrWhiteSpace(s)) return string.Empty;

            return $"{char.ToLowerInvariant(s[0])}{s.Substring(1)}";
        }

        [Benchmark(OperationsPerInvoke = 16)]
        public string ToCamelCaseNew()
        {
            if (string.IsNullOrWhiteSpace(s)) return string.Empty;

            var newFirstLetter = char.ToLowerInvariant(s[0]);
            if (newFirstLetter == s[0])
                return s;

            return newFirstLetter + s.Substring(1);
        }

        [Benchmark(OperationsPerInvoke = 16)]
        public string ToCamelCaseWithArray()
        {
            if (string.IsNullOrWhiteSpace(s)) return string.Empty;

            char[] a = s.ToCharArray();
            a[0] = char.ToLowerInvariant(a[0]);
            return new string(a);
        }

        {
            var summary = BenchmarkRunner.Run<Program>();
        }
    }
}
```

Results:

``` ini

BenchmarkDotNet=v0.11.4, OS=Windows 10.0.17134.648 (1803/April2018Update/Redstone4)
Frequency=2343750 Hz, Resolution=426.6667 ns, Timer=TSC
.NET Core SDK=2.2.104
  [Host]     : .NET Core 2.1.8 (CoreCLR 4.6.27317.03, CoreFX 4.6.27317.03), 32bit RyuJIT
  DefaultJob : .NET Core 2.1.8 (CoreCLR 4.6.27317.03, CoreFX 4.6.27317.03), 32bit RyuJIT


```
|               Method |        s |      Mean |     Error |    StdDev |    Median | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|--------------------- |--------- |----------:|----------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|       **ToCamelCaseOld** | **SomeName** | **12.975 ns** | **0.5710 ns** | **1.6474 ns** | **12.642 ns** |      **0.0035** |           **-** |           **-** |                 **6 B** |
|       ToCamelCaseNew | SomeName |  5.013 ns | 0.2041 ns | 0.5791 ns |  4.943 ns |      0.0030 |           - |           - |                 5 B |
| ToCamelCaseWithArray | SomeName |  3.533 ns | 0.1134 ns | 0.3254 ns |  3.504 ns |      0.0024 |           - |           - |                 4 B |
|       **ToCamelCaseOld** | **someName** | **12.883 ns** | **0.3666 ns** | **1.0517 ns** | **12.943 ns** |      **0.0035** |           **-** |           **-** |                 **6 B** |
|       ToCamelCaseNew | someName |  1.038 ns | 0.0311 ns | 0.0906 ns |  1.025 ns |           - |           - |           - |                   - |
| ToCamelCaseWithArray | someName |  3.357 ns | 0.1235 ns | 0.3562 ns |  3.153 ns |      0.0024 |           - |           - |                 4 B |

So, in any case, the proposed method impl is much faster than the current one. And if the word begins with a lowercase letter, then there is no allocation at all. Additional character checking is nothing compared to the time spent on allocating objects :).

Issue #1073, opened 2019-03-18
Why there are no [Guid,Short,UInt,ULong,UShort]GraphType in GraphTypesLookup ctor?


sungam3r:
Do I understand correctly that all known types of scalars should be initialized there?
Alse why `GuidGraphType`, `ShortGraphType`, `UIntGraphType`, `ULongGraphType` types do not have name set in the constructor? All other scalars have their name set in the constructor.

joemcbride:
See this PR:  
https://github.com/graphql-dotnet/graphql-dotnet/pull/917

It has some merge issues that need to be resolved.

sungam3r:
Thanks! How can I get a package with these changes now? These errors are blocking our work now. I see on nuget.org only package with version 3.0.0-preview-1107 builded 17 days ago.

joemcbride:
I pushed 3.0.0-preview-1141 to nuget.org.  For immediate builds you can connect to the myget.org feed.

https://www.myget.org/feed/Packages/graphql-dotnet

WhippetsAintDogs:
@joemcbride I've tried the 3.0.0-preview-1141, but the using a field builder with an expression parameter returning a property with the `uint` type causes this error inside an `InputObjectGraphType<>`: 
```
The GraphQL type for Field: FIELD_NAME on parent type: PARENT_GRAPH_TYPE could not be derived implicitly.
``` 
If I try to specify the type to be `typeof(UIntGraphType)` in the `type` parameter of the field builder, I got this error instead with a numeric or string value (e.g. `10` or `"10"`) passed as an UInt argument: 
```
Argument ARGUMENT_NAME has invalid value {FIELD_NAME: 10 or \"10\"}.\nIn field FIELD_NAME: Expected type \"UInt\", found 10 or \"10\".
```

sungam3r:
Can you give an example of your schema configuration?

WhippetsAintDogs:
Yes, there it's is :)

![image](https://user-images.githubusercontent.com/9271409/55160388-99fd2280-5139-11e9-98bf-04c5dc6676d1.png)

![image](https://user-images.githubusercontent.com/9271409/55160415-ada88900-5139-11e9-822e-f75cb3cec530.png)

Code:

![image](https://user-images.githubusercontent.com/9271409/55160496-d761b000-5139-11e9-8555-4290fe6f41d3.png)

[...]

PR #1074, opened 2019-03-18
Fix scalar types added in #812

sungam3r:
This PR is a duplicate of #917 with resolved merge issues. It seems that the original PR is abandoned.


sungam3r:
Fixes #1073

PR #1080, opened 2019-03-19
AutoRegisteringObjectGraphType improvement

### LINKED ISSUE #1079: AutoRegisteringObjectGraphType Improvement
## Summary

It would be nice to have the support of the following features in `AutoRegisteringObjectGraphType<T>`:
1. Enum properties support.
2. Support for property types registered in `GraphTypeTypeRegistry`.
3. Support for `RequiredAttribute` to tune nullability of reference typed properties.

### Motivation

Simplification of the schema.

sungam3r:
Fixes #1079

### DIFF
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -1,11 +1,14 @@
 using GraphQL.StarWars.Types;
 using GraphQL.Types;
-using System;
-using System.Collections.Generic;
-using System.Linq;
+using GraphQL.Utilities;
 using Shouldly;
-using Xunit;
+using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using Xunit;
 
 namespace GraphQL.Tests.Types
 {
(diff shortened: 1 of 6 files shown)


src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -91,7 +138,7 @@ public void infers_from_list_types()
 
             type.Field(d => d.someList, nullable: true);
 
-            type.Fields.Last().Type.ShouldBe(typeof(ListGraphType<IntGraphType>));
+            type.Fields.Last().Type.ShouldBe(typeof(ListGraphType<NonNullGraphType<IntGraphType>>));
```
sungam3r: It seems to me that the behavior regarding this property was wrong:
```c#
 public List<int> someList { get; set; }
```
Elements of this list can never have null value, so it should be coerced to `ListGraphType<NonNullGraphType<IntGraphType>>`.

I added tests for various combinations of types.

src/GraphQL/GraphQL.csproj
```diff
@@ -29,6 +29,7 @@
     <PackageReference Include="GraphQL-Parser" Version="3.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Include="System.Buffers" Version="4.5.0" />
+    <PackageReference Include="System.ComponentModel.Annotations" Version="4.5.0" />
```
sungam3r: Is it permissible to add this dependency?

src/GraphQL/TypeExtensions.cs
```diff
@@ -129,20 +129,23 @@ public static Type GetGraphTypeFromType(this Type type, bool isNullable = false)
...
-            if (IsAnIEnumerable(type))
+            else if (IsAnIEnumerable(type))
+            {
+                var clrElementType = GetEnumerableElementType(type);
+                var elementType = GetGraphTypeFromType(clrElementType, clrElementType.IsNullable()); // isNullable from elementType, not from parent container
+                graphType = typeof(ListGraphType<>).MakeGenericType(elementType);
+            }
+            else
             {
-                var elementType = GetGraphTypeFromType(type.GenericTypeArguments.First(), isNullable);
```
sungam3r: There was an error in the case of `IEnumerable` property (not `IEnumerable<T>`) - `GenericTypeArguments` returns empty array.

src/GraphQL/Types/AutoRegisteringObjectGraphType.cs
```diff
@@ -48,7 +61,45 @@ protected virtual IEnumerable<PropertyInfo> GetRegisteredProperties()
...
                 .GetProperties(BindingFlags.Public | BindingFlags.Instance)
-                .Where(p => p.PropertyType.IsValueType || p.PropertyType == typeof(string));
+                .Where(p => IsEnabledForRegister(p.PropertyType, true));
+        }
+
+        private static bool IsEnabledForRegister(Type propertyType, bool firstCall)
+        {
+            if (propertyType == typeof(string)) return true;
+
+            if (propertyType.IsValueType) return true; // TODO: requires discussion: Nullable<T>, enums, any struct
```
sungam3r: This behavior seems to me to be wrong. You can’t just take and allow to register all structures/enums.
While left as it was before.

src/GraphQL/Utilities/GraphTypeTypeRegistry.cs
```diff
@@ -22,33 +22,28 @@ static GraphTypeTypeRegistry()
                 [typeof(DateTime)] = typeof(DateGraphType),
                 [typeof(DateTimeOffset)] = typeof(DateTimeOffsetGraphType),
                 [typeof(TimeSpan)] = typeof(TimeSpanSecondsGraphType),
-                [typeof(Guid)] = typeof(IdGraphType)
+                [typeof(Guid)] = typeof(IdGraphType),
+                [typeof(short)] = typeof(ShortGraphType),
```
sungam3r: And again forgotten scalars ...

src/GraphQL/GraphQL.csproj
joemcbride: Admittedly it makes me cringe a bit but that is probably the easiest solution.

sungam3r: Well, this is essentially standardized well-known library with common meta-information. If not this, then what? I do not really want to reinvent the wheel.

Issue #925, opened 2018-11-27
ResolveFieldContext.GetArgument won't try to convert argument name to camelCase


branhill:
## Description

QueryArgument will convert Name to camelCase automatically, but ResolveFieldContext.GetArgument won't do that. It only get argument by original name string.

### Steps to reproduce

Use a PascalCase string as the QueryArgument.Name, then use same string in ResolveFieldContext.GetArgument().

### Expected result

The value of the argument.

### Actual result

null

### Environment

GraphQL 2.4.0

sungam3r:
Fixed in #1081

PR #1081, opened 2019-03-20
#925 Convert argument name to camelCase

d3vi1h3aRt:
Not sure if it needs to support field resolve not in schema context.


src/GraphQL/Types/ResolveFieldContext.cs
```diff
@@ -46,6 +47,8 @@ public class ResolveFieldContext<TSource>
 
         public IEnumerable<string> Path { get; set; }
 
+        public IFieldNameConverter FieldNameConverter { get; set; } = new CamelCaseFieldNameConverter();
```
joemcbride: The Schema should always be populated so this should not be needed.

PR #1091, opened 2019-03-28
UInt/UShort/Short scalars fix

### LINKED ISSUE #1073: Why there are no [Guid,Short,UInt,ULong,UShort]GraphType in GraphTypesLookup ctor?

sungam3r:
fixes #1073
add two scalars - Byte and SByte

### DIFF
--- a/src/GraphQL/Types/UIntGraphType.cs
+++ b/src/GraphQL/Types/UIntGraphType.cs
@@ -18,6 +18,11 @@ namespace GraphQL.Types
                         return (uint)intValue.Value;
                     return null;
 
+                case LongValue longValue:
+                    if (uint.MinValue <= longValue.Value && longValue.Value <= uint.MaxValue)
+                        return (uint)longValue.Value;
+                    return null;
+
                 default:
                     return null;
             }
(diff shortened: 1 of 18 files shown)


src/GraphQL/Types/UIntGraphType.cs
```diff
@@ -4,19 +4,31 @@ namespace GraphQL.Types
...
+            {
+                case UIntValue uintValue:
+                    return uintValue.Value;
+
+                case IntValue intValue:
+                    if (intValue.Value >= 0)
+                        return (uint)intValue.Value;
+                    return null;
+
+                case LongValue longValue:
```
sungam3r: Value can come in two containers - `IntValue` and `LongValue` (if overflow Int)

src/GraphQL/Types/ULongGraphType.cs
```diff
@@ -4,19 +4,31 @@ namespace GraphQL.Types
...
+            {
+                case ULongValue ulongValue:
+                    return ulongValue.Value;
+
+                case IntValue intValue:
+                    if (intValue.Value >= 0)
+                        return (ulong)intValue.Value;
+                    return null;
+
+                case LongValue longValue:
```
sungam3r: Value can come in two containers - `IntValue` and `LongValue` (if overflow Int)

Issue #1093, opened 2019-03-29
Support Connection fields of TConnection, TEdge, TNode


benmccallum:
## Summary, example, motivation

Currently the ConnectionBuilder assumes we only ever want to use the framework's `Connection<TNode>` and `Edge<TNode>` classes.

In reality it's common to want to have custom Edge types with additional domain-defined properties like `FriendedAt` if describing the Edge between a User and another User.

Or even a custom Connection type with some aggregation properties that summarise the entire connection or the page subset, like `avg`, `sum`, etc.

I'm happy to fork and work on this, shouldn't be too hard, just creating the issue now in anticipation of any feedback.

Shane32:
As of the time of writing, `ConnectionType`, `EdgeType`, and `PageInfoType` are not tied to `Connection<TNode>` or `Edge<Node>` and may be used with any class that has the necessary properties.  I am hoping to change this for v8, so it is tied to those classes (or perhaps interfaces), for faster execution and AoT compilation support.  Nevertheless, those graph types contain almost no code and it is easy to rewrite them as-needed in your own codebase.  You can even use `AutoRegisteringObjectGraphType<T>` to build graph types from connection types.  It does not seem to be necessary to add additional support into the GraphQL.NET base library to allow for extending these types.

Note that it is true you cannot use the existing connection builders; you will need to add the connection field manually.

As this topic is 4 years old, I'm closing this issue now.  Please respond if you have additional comments.

PR #1066, opened 2019-03-17
remove TGraphType from ConnectionBuilder

SimonCropp:
`TGraphType` on `ConnectionBuilder` doesnt seem to add any value? it is not actually used .

PR #1094, opened 2019-03-29
feat: support user-defined Edge and Connection types

benmccallum:
Adds support for user-defined Edge and Connection types so custom properties can be added to them as needed.

Addresses #1093 
Builds on #1066

### DIFF
--- a/src/GraphQL.Tests/Builders/ConnectionBuilderTests.cs
+++ b/src/GraphQL.Tests/Builders/ConnectionBuilderTests.cs
@@ -143,7 +143,7 @@ namespace GraphQL.Tests.Builders
 
             var field = type.Fields.Single();
             field.Name.ShouldBe("testConnection");
-            field.Type.ShouldBe(typeof(ConnectionType<ObjectGraphType>));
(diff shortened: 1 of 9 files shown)


benmccallum:
FYI, there's a commit in there from @SimonCropp which I'm not sure is about, I just saw it in an active branch and didn't want to be working on stuff in the same place if he had WIP. 

With/without his changes we can make this work.

Includes unit tests, will probably need to update docs too. I'll hold off on that until you give this a review and thumbs up @joemcbride

benmccallum:
I've got one test failing and I'm not sure why. Will take a look.

benmccallum:
Fixed. Over to you @SimonCropp fo rreview!

benmccallum:
Something I noticed yesterday is that the `ConnectionBuilder` doesn't provide any type checking for the resolver. It just requires a response of `object`. Could we add overloads for one that is? I'd like that extra sense of security.

jwx:
Is there an ETA for this?

benmccallum:
@SimonCropp has approved it so I guess if @joemcbride can merge it'd then be a matter of falling into the next release whenever that is.

benmccallum:
Sorry guys I made this a little confusing by targeting @SimonCropp's branch. Will create another that targets master.

PR #1105, opened 2019-04-08
Change Executer to Executor

grounzero:
Was really bugging me so I created a PR.


sungam3r:
It is a breaking change to public api. Although now the major version is already incremented.

grounzero:
I understand.  Do we know roughly when the major version will be live?

sungam3r:
I do not know, I myself am waiting for this.

PR #1107, opened 2019-04-08
Mapping from field name to property expression name

### LINKED ISSUE #850: Mutation type with alias field will not map.
## Description

When having a mutation that uses an alias, `ObjectExtensions.ToObject` will be unable to populate the values.

### Steps to reproduce

See [unit test](https://github.com/graphql-dotnet/graphql-dotnet/pull/851).

### Expected result

Properties should be mapped, `Field("properties", x => x.PropertyValues,
                type: typeof(ListGraphType<ProgramStepPropertyValueInputType>)).Name("properties");` will not be mapped by [ObjectExtension](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/ObjectExtensions.cs#L34).

`Assert.True(Program.Steps.Any(step => step.PropertyValues.Count > 0));`

### Actual result

Empty List of properties.

### Environment

2.0.0, Windows full framework

### LINKED ISSUE #1287: Mutations: How to map alternative name of a Field back to its original property
Scenario: a property on a node was renamed from "FirstName" to "FName". To not break the interface I want to keep the old name for a certain time of periode. Because of that I apply an alternative name "FirstName" on the InputType, see sample below:

```
public class HumanInputType : InputObjectGraphType<Human>
{
        public HumanInputType()
        {
            Name = "HumanInput";
            Field(x => x.Name);
            Field(x => x.FName).Name("FirstName");
        }
}
```

From the client perspective it is working as expected. For above sample I get "Name" and "FirstName" as parameter.

From the resolver perspective I would have expected that "FirstName" gets mapped to "FName" on type "Human". But it is not. Further in the context.Arguments of "human" I see keys for "Name" and "FirstName" (same here). 

```
resolve: context =>
{
            var human = context.GetArgument<Human>("human");
            return data.AddHuman(human);
});
```

How can I map from "FirstName" to "FName" in a generic way? I thought this is done by the field configuration on the HumandInputType automatically.

### LINKED ISSUE #1358: Input properties with a custom name are never mapped
## Summary
The properties of my `InputObjectGraphType` are not mapped when I give them a name.
In the example bellow `MyProp` will always be null because of the name `myCustomName`.
Did I missed something ?

## Relevant information
**ApiQuery**
```C#
 this.Field<ListGraphType<MyType>>(
        "myEndpoint",
        new QueryArgument<MyClassType> { Name = "args" }
        ),
        resolve: context => {
               MyClass = context.GetArgument<MyClass>("args")
        });
```

**MyClass**
```C#
    public class MyClass: IMyClass
    {
        public decimal? MyProp { get; set; }
    }
```

**MyClassType**
```C#
    public class MyClassType: InputObjectGraphType<MyClass>
    {
        public MyClassType()
        {
            this.Name = "ClassName";

            this.Field(x => x.MyProp, true, type: typeof(IdGraphType)).Name("myCustomName");
        }
    }
```
**Query**
```C#
query {
  myEndpoint(args: {myCustomName: 1}) {
}
```

### Environment (if relevant)
DotNetCore 2.2, GraphQL .NET 2.4, Windows 7.

sungam3r:
fixes #850
fixes #1287 
fixes #1358 

I tried to solve this issue with minimal changes in public api. However, the default parameters were added. In addition, I don’t really like the fact that they appeared exactly in that place, but this is a necessary measure, because without this change information about property matching is lost.


src/GraphQL.Tests/Bugs/BugNestedCollection.cs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    public class ProgramStepInputType : InputObjectGraphType<ProgramStepConfig>
+    {
+        public ProgramStepInputType()
+        {
+            Name = "ProgramStepInput";
+            Field(x => x.ProgramStepDefinitionId);
+            Field(x => x.SequenceOrder);
+            Field("properties", x => x.PropertyValues,
```
sungam3r: Here is a field with a name different from the name of the class property

sungam3r:
@joemcbride @Shane32 I returned to this PR after a long break. I minimized the number of modified files to make the review easier.

PR #1119, opened 2019-04-29
remove global.json from sln

SimonCropp:
since it doenst exist


sungam3r:
Wonderful. I suspected that it was not needed, but I was afraid to delete it.

PR #1120, opened 2019-04-29
Cleanup

SimonCropp:


src/GraphQL/Validation/Rules/OverlappingFieldsCanBeMerged.cs
```diff
@@ -14,7 +14,7 @@ public class OverlappingFieldsCanBeMerged : IValidationRule
 
         public static string ReasonMessage(Message reasonMessage)
         {
-            if (reasonMessage.Msgs != null && reasonMessage.Msgs.Count() > 0)
+            if (reasonMessage.Msgs != null && reasonMessage.Msgs.Count > 0)
```
sungam3r: It can be changed to 
```c#
if (reasonMessage.Msgs?.Count > 0)
```

SimonCropp: @sungam3r true. fixed

PR #1121, opened 2019-04-30
use object initializers

SimonCropp:
@joemcbride this is a preference of mine. i think it makes the code more readable. thoughts?


sungam3r:
Indeed, there are recommendations to use initializers instead of imperative assignments.
If we are talking about maintaining a single style of code, then it is worth considering the option of using static analysis tools, for example editorconfig, StyleCop.Analyzers, GCop etc. Only then we can talk about the automatically supported uniform style of the entire solution.

Issue #1136, opened 2019-05-11
Define complex args (might be bug?)


MakaBuka:
Hi guys,

Got very wierd bug, maybe you can give me a hint or might be really bug

Anyway, I've defined complex field that calls orderBy (which gonna use to order resultset in resolver):

	public class GraphqlSort {
		public string FieldName { get; set; }
		public string Direction { get; set; }
	}

	public class GraphqlSortType : InputObjectGraphType<IEnumerable<GraphqlSort>> {

		public GraphqlSortType(SortOptions settings) {
			Name = settings.Key;
			Field<StringGraphType>("field");
			Field<StringGraphType>("direction");
		}
	}

then adding it as arg to defined field (part of code):
			var objType = new GraphqlSortType();
			var listType = new ListGraphType(objType);
			var arg = new QueryArgument(listType);
                       arg.Name = "sortBy";
TableArgs = new QueryArguments(
					arg
				);

GraphQL query, as an example:

query {

clients: customers(sortBy:{field:"id", direction: "ASC"}) {
  id
}
}

but got next an error:

[...]

PR #1139, opened 2019-05-11
avoid NRE for #1136

sungam3r:
avoid NRE for #1136

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1136.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1136.cs
@@ -1,0 +1,21 @@
+using GraphQL.Language.AST;
+using GraphQL.Types;
+using Shouldly;
+using System.Linq;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    public class Bug1136
+    {
+        [Fact]
+        public void IsValidLiteralValue_Should_Not_Throw_NRE()
+        {
+            var type = new DateGraphType() { Name = null };
+            var value = new BooleanValue(true);
+            var result = type.IsValidLiteralValue(value, null).ToArray();
+            result.Length.ShouldBe(1);
+            result[0].ShouldBe("Expected type \"null\", found true.");
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


SimonCropp:
@sungam3r is it possible to add a unit test that verifies this change?

sungam3r:
> is it possible to add a unit test that verifies this change?

Done.

sungam3r:
This change will not get any worse. The original problem at least does not hide behind NRE.

src/GraphQL.Tests/Bugs/Bug1136.cs
```diff
@@ -0,0 +1,21 @@
...
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    public class Bug1136
+    {
+        [Fact]
+        public void IsValidLiteralValue_Should_Not_Throw_NRE()
+        {
+            var type = new DateGraphType() { Name = null };
```
SimonCropp: @joemcbride thoughts on this? should a graphtype ever be allowed to have a null name?

joemcbride: No it shouldn’t.  Perhaps a better solution would be to throw an error when the schema is built if the name is null.

sungam3r: > Perhaps a better solution would be to throw an error when the schema is built if the name is null.

Of course it will be! But this is a completely different story and not only the name should be checked. For example see #1156. Now there are many opportunities for carelessness to create such a schema that will be invalid. I'm sure this is the subject of a **separate task**. In general, now API of schema classes (GraphTypes) allows you to set any properties you like, and the problems will be revealed already at the stage of request execution.

Also note that the test does not use a schema at all. That is, protection when building a schema is required, but this is not a generally sufficient condition.

sungam3r:
@joemcbride , @SimonCropp  Any thoughts?

SimonCropp:
i will admit to not be overly familiar with API boundary validation in this project but i would have assumed it would be better to place a guard on that property. eg https://github.com/graphql-dotnet/graphql-dotnet/pull/1257

sungam3r:
The guard will not save from a mistake. See my comment on your PR.

sungam3r:
Closed in favor of #1284

PR #1158, opened 2019-05-25
Expose all inner exceptions from ExecutionError

### LINKED ISSUE #1157: ExecutionError uses not the most nested exception
## Description

If I do this in my resolver
```c#
throw new NullReferenceException("aaa", new InvalidCastException("bbb", new InvalidOperationException("ccc")));
```
then I get `INVALID_CAST` error code in extensions/code in json result

### Steps to reproduce

Use
```c#
throw new NullReferenceException("aaa", new InvalidCastException("bbb", new InvalidOperationException("ccc")));
```
in any resolver

### Expected result

`INVALID_OPERATION` - This is the most nested exception that describes the error.

### Actual result

`INVALID_CAST` - this is the InnerException of outermost `NullReferenceException`

sungam3r:
fixes #1157


sungam3r:
Are there any obstacles not to merge this?

src/GraphQL/Execution/ExecutionError.cs
```diff
@@ -26,15 +29,36 @@ public ExecutionError(string message, IDictionary data)
         public ExecutionError(string message, Exception exception)
             : base(message, exception)
         {
-            var ex = exception?.InnerException ?? exception;
```
sungam3r: I believe that it would be more correct to take the code from the exception and not try to look further

SimonCropp: agreed

Issue #648, opened 2018-05-02
Execution scoped services


pekkah:
This issue keeps coming up. Would be good to have a way of resolving dependencies per execution while keeping the schema as singleton. This would require a way of injecting the services in a away that their lifetime would be one execution. For example creating entity framework context per execution.

PascalSenn:
I know to ways to workaround this problem:

**User Context**
You can inject the ``IDependencyResolver`` into the user context. As far as the user context and the dependency resolver are scoped services. You can resolve dependencies in every field resolver. Not really a nice way, but it does the job

**IContextProvider (IHttpContextAccessor)**
There is a way to request scoped services on demand with the `IHttpContextAccessor`. 
Sample: 
```csharp

    public interface IContextProvider<T>
    {
        T Get();
    }

    public class ContextProvider<T> : IContextProvider<T>
    {
        IHttpContextAccessor contextAccessor;
        public ContextProvider(IHttpContextAccessor contextAccessor)
        {
            this.contextAccessor = contextAccessor; 
        }

        public T Get()
        {
            if(contextAccessor?.HttpContext?.RequestServices == null) {
                return default(T);
            }
            return contextAccessor.HttpContext.RequestServices.GetService<T>();
        }
    }
    public class FooType : ObjectGraphType<Foo>
    {
        private readonly DependencyToResolve _scopedService;
        public FooType (IContextProvider<DependencyToResolve> context)
        {
            _scopedService = context.Get()
        }
    }
```

pekkah:
Hijacking UserContext for this seems just wrong, but that's an another topic (#637)

In your example _scopedService would still be a singleton as FooType would be part of the Schema and schema would be singleton. You would need to resolve it inside the resolver which you probably meant.

PascalSenn:
@pekkah Totally agree with the user context.

  sure. What I meant to do:
```csharp
    public class FooType : ObjectGraphType<Foo>
    {
        private readonly IContextProvider<DependencyToResolve> _scopedService;
        public FooType (IContextProvider<DependencyToResolve> context)
        {
            _scopedService = context;
            
            Field<BarType>("bar", resolve: x => _scopedService.Get().Magic(x.Source.Bar));
        }
    }
```

What could be nicer way of resolving dependencies in the schema is to have a provider bound to the HttpContext on `ResolvedFieldContext`. 
Something like: 
```csharp
    public class FooType : ObjectGraphType<Foo>
    {
        public FooType()
        {
[...]

Issue #1151, opened 2019-05-20
Add support for execution services


johnrutherford:
I originally posted this suggestion on issue #648. I think this would be a good addition to v3.0. It's not necessary, but I think it would be good to implement #1144 first.

> Allow for a dependency resolver instance to be passed in through the `ExecutionOptions`. This would allow you to wrap `HttpContext.RequestServices` in ASP.NET Core and pass that in. If you're not using ASP.NET Core, this would still allow you to create a custom scope for execution and pass that in. If this is not set at all, it should probably default to the dependency resolver on the schema.
> 
> Expose this as an `ExecutionServices` property on the `ResolveFieldContext` class.
> Then implement some field resolvers with 1-3 additional type parameters that handle the dependency resolution. Like this:
> 
> ``` csharp
> public class FuncFieldResolver<TServiceType, TReturnType> : IFieldResolver<TReturnType>
> {
>     private readonly Func<ResolveFieldContext, TServiceType, TReturnType> _resolver;
> 
>     public FuncFieldResolver(Func<ResolveFieldContext, TServiceType, TReturnType> resolver)
>     {
>         _resolver = resolver;
>     }
> 
>     public TReturnType Resolve(ResolveFieldContext context)
>     {
>         var service1 = context.ExecutionServices.Resolve<TServiceType>();
>         return _resolver(context, service1);
>     }
> 
>     object IFieldResolver.Resolve(ResolveFieldContext context) => Resolve(context);
> }
> ```
> 
> And then set up extension methods so fields could be created like this:
> 
> ``` csharp
> Field<FooType, Foo>()
>     .Name("foo")
>     .ResolveAsync<IMyService>((context, myService) =>
>     {
>         return myService.GetFooAsync();
>     });
> ```
> 
> So this would be equivalent to injecting parameters into a controller action in MVC.

sungam3r:
It is ServiceLocator antipattern.
> Then implement some field resolvers with 1-3 additional type parameters that handle the dependency resolution.

This is a leaky abstraction, the DI container should deal with dependency resolution when introducing them into the class constructor (of query type). Also what if resolver use 4, 5, 10 dependencies? There will be too much noise in public API of resolvers and extension methods.

benmccallum:
I'm all for this approach, we've discussed a bit on the linked PR. Definitely would like see this in 3.0!

sungam3r:
[...]

PR #1160, opened 2019-05-26
Draft for scoped dependencies [WIP]

sungam3r:
Add support to use scoped dependencies from singletons (#648).


sungam3r:
DO NOT MERGE THIS. DEMO ONLY.

sungam3r:
Requires discussion.

corstian:
Could you show a little code snippet of how it should be used? Looking at the changes it goes over my head right now.

sungam3r:
>  I do not think using `Func<T>` does justice to the fact a certian type is registered as being scoped

`Func<T>` is only a generalized way to defer receiving/creating an object. In this case, a specific method of creating a dependency remains inside the DI-container. It may be singleton, scoped, transient - whatever. The main thing is that the dependency consumer (constructor and target class code) is entirely free from the need to independently manipulate calls to the DI-container (sort of `Get<T>`).

Rather correct to say that `Func<T>` is not about scope, it is about deferring object materialization.

Also note that `AddFunc` extension works not only for scoped dependencies.

pekkah:
Isn't Func<T> supported out of the box with Microsoft.Extension dependency container.

sungam3r:
Unfortunately not.

KreativJos:
For me, this issue is resolved by passing the HttpContext IServiceProvider (RequestServices) into the UserContext (which is named quite awfully: it is not a user context, but the GraphQL request context => it is generated with each GQL-request).
But that adds a layer of complexity for the developers. I still think it should be included in the UserContext, but the main framework should be able to resolve scoped services automatically.

sungam3r:
@KreativJos And again RequestServices - it is all about ServiceLocator antipattern.

Cooke:
Great job. I agree with this solution. However I would prefer the "Accessor-pattern" style used for IHttpContextAccessor. Could look something like this:
register: .AddServiceAccessor<T> 
dependency: MyGqlType(IServiceAccessor<T> accessor)
usage: accessor.Value

Its more verbose though.

KreativJos:
> @KreativJos And again RequestServices - it is all about ServiceLocator antipattern.

But the whole graphql-dotnet project is currently an anti-pattern... Schema's should be able to be created in even transient scopes. Middleware should always be Scoped at least.

Cooke:
After trying out this solution in our production code I withdraw my comment and support the suggestion with resolver dependency injection. With this solution the resolver logic becomes cluttered with resolving the scoped dependencies and also the constructor of the object type with "accessor" service arguments becomes cluttered. 

What I think is needed is a way to move out resolver logic which depends on different services (and scoped stuff) into something that can be instantiated per request. Field resolver injection would make that easier:

(context, [InjectedService] myResolverType) => myResolver.Resolve(context)

First class support for this could also be an alternative, similar to how hot chocolate have done with their resolver types (instantiated per request).

sungam3r:
> With this solution the resolver logic becomes cluttered with resolving the scoped dependencies and also the constructor of the object type with "accessor" service arguments becomes cluttered.

@Cooke can you please give an example to take a look. It may be that the problem is not so much in the proposed approach to resolving dependencies, but in the fact that your type is just so big that it itself has so many dependencies (god object) that you have to specify all of them in one way or another. It may be worth decomposing the type into separate types with their own responsibilities.

Field resolver injection (on the one hand a handy thing, really) in this regard is in fact only a convenient tool **to hide** this fact from the eyes. The class will be successfully created (since the dependencies are removed from the constructor), but if you do not add the dependency to the container, the error will pop up only during the execution of a specific resolver. These are very nasty runtime errors. Theoretically, you can get rid of such errors, but then you have to make your own system of field resolver injection parameters validation that may turn out to be much more complicated than to use an already working constructor injection validation services (in Microsoft.Extensions.DependencyInjection or others).

sungam3r:
> But the whole graphql-dotnet project is currently an anti-pattern... Schema's should be able to be created in even transient scopes. Middleware should always be Scoped at least.

@KreativJos Well, you really shouldn't get upset. As far as I know, the middlewares problem is known and there is a desire to solve it. About creating a schema in transient scope - is it impossible to do? But in any case, why? The schema is meta information common to all requests. Why on each request to initialize the whole bunch of types and relationships between them, if it can be done once? In our production backend a few dozen services working through graphql-dotnet and I have not seen a single case where the schema would be required in transient scope.

KreativJos:
> > But the whole graphql-dotnet project is currently an anti-pattern... Schema's should be able to be created in even transient scopes. Middleware should always be Scoped at least.
> 
> @KreativJos Well, you really shouldn't get upset. As far as I know, the middlewares problem is known and there is a desire to solve it. About creating a schema in transient scope - is it impossible to do? But in any case, why? The schema is meta information common to all requests. Why on each request to initialize the whole bunch of types and relationships between them, if it can be done once? In our production backend a few dozen services working through graphql-dotnet and I have not seen a single case where the schema would be required in transient scope.

There is nothing to be upset about. What I meant is that the pull-request here is a workaround for the real issue (I should've said that, it would have been more explicit).
Should the schema be common in each request? It could be different based on auth (headers / cookie) for example. It is the way it is because the project is based on a parser, not because of a design choice.
Personally, I currently do not see a good way of designing the project in a better way apart from rewriting most of it from scratch (which is way to much work and not a good idea most of the time).

A transient schema would be interesting indeed, but I can't think of a use case for it. Scoped is the one that is wished for, because of request service resolving (mainly database / ORM stuff) and logging.

sungam3r:
>  It could be different based on auth (headers / cookie) for example. 

I did not even think about it. In general, I see myself the schema of the same (constant) for the entire application, regardless of the sender of the request. Another thing is that the individual pieces of the schema may carry **directives** indicating specific access rights requirements. Returning different schemas depending on the properties of the query (context) is an obfuscation of behavior, which is fraught with errors. Directives allow you to add meta information about rights, but this is a completely different story.

benmccallum:
Interesting! Nice one. How does this compare to the suggestion by @johnrutherford to have a pattern similar to asp.net core middleware where the dependency to be injected is an arg of the method (e.g. Run) or in our case the resolver, and graphql-dotnet takes care of noticing that and injecting one for you when it calls the resolver? I'd be keen to understand the pros/cons of each. 

In terms of the discussion above, fundamentally it just makes sense that the schema be singleton; it shouldn't change once setup and it should be as optimal as possible. If you wanted to do have it scoped, theoretically you could do that now, right? Register your schema as scoped and measure the hit. I guess one reason to investigate a schema-per-request would be serverless where you're potentially having the schema created regularly, so should we be optimizing it as much as possible to make that feasible? For now, I'm with keeping it singleton and using a solution like this, well documented in the docs, to get singleton + DI.

sungam3r:
> How does this compare to the suggestion by @johnrutherford to have a pattern similar to asp.net core middleware where the dependency to be injected is an arg of the method (e.g. Run) or in our case the resolver, and graphql-dotnet takes care of noticing that and injecting one for you when it calls the resolver? Link to his comment. I'd be keen to understand the pros/cons of each.

It seems to me that this suggestion is not similar to asp.net core middleware at all. Middleware implicitly allows you to extend the signature by substituting objects from the container in place of the parameters. @johnrutherford gives an example of an **explicit** resolver with a fixed signature (with 1-3 additional type parameters).

The method of injecting dependencies into the `Invoke` method of a middleware is somewhat more complicated than the usual constructor injection. I'm not saying that it is bad, no. Simply, it is rather specific to asp.net core runtime. If you look at the source code that implements dependency injection in `Invoke` (`UseMiddlewareExtensions.UseMiddleware` and deeper), you will see that it performs quite a lot of work on finding out which parameter to take from. In principle, you can write a similar code in this project, but you need to weigh the relative benefits of this complication.

> it shouldn't change once setup and it should be as optimal as possible

Exactly.

_____
Probably, I should clearly explain my intention in this PR. I wanted to implement such a mechanism that would allow the use of **constructor injection** in the most ordinary way.

By the way, now I'm not sure that this code should be merged at all. In fact, this code does not apply to graphql-dotnet. It can be implemented as an external library. That's what happens in our projects. There is a shared library with extensions for DI. Developers use `AddFunc` extension method - that's all.

I mean, it would be bad if the `IScopeProvider` interface (and others) becomes part of graphql-dotnet. This will create an **unnecessary dependency** in our components on the graphql-dotnet or it will force us to write adapters from one interface to another. Therefore, I find it difficult to say how to proceed. For now, this PR from is most likely an example (demo) of how "scoped problem" can be solved.

corstian:
@KreativJos, to elaborate on the 'antipattern'; the resolvers themselves represent a way to access data. These functions (or just the field definitions in general) are considered immutable in order to guarantee consistency of the graph, and therefore (by convention) these are defined in the constructor. From this point of view it does not make sense to declare these services as scoped, unless you are doing something which mutates the schema or graph types themselves. In this way I think that defining the schema and related types as singleton makes perfect sense.

When thinking about these resolvers as the actual materialization of the query it would make sense (in my humble opinion) to inject the properties directly in the resolver function, whenever executed. For this the `IServiceProvider` could be used to have support for all dependency providers which implement this specific interface, just the way @johnrutherford has proposed before.

Given the generic `Func` type only supports up to 16 generic arguments, it would make sense also to provide up to 16 generic overloads for the resolver, to be used in the way already proposed before (see https://github.com/graphql-dotnet/graphql-dotnet/issues/1151).

I consider this way to be a better approach than directly injecting the `IServiceProvider` into the resolver as it makes it pretty clear what dependencies the resolver has, where those are used, and what the scope of said dependency is. I feel like [the reasons which make the service locator pattern an anti-pattern, as given in this answer on SO](https://stackoverflow.com/a/22795888/1720761), can all be invalidated for the current use-case.

In case you require more than 15 injected services (and one `ResolveFieldContext`), it is always possible to fall back to a custom type which derives from the `IFieldResolver`, which can use constructor injection.

The only downside I can see to this approach, from a practical standpoint, is that when looking at the usage for a specific type within Visual Studio, the list will become significantly longer. I expect to see hundreds of usages for my data context when following this approach.

benmccallum:
I think the reason I like John's approach is that is so analogous to what the core team did you solve it for middleware Invoke. A resolver function is, in a similar sense, called by the framework and handed all it needs to do it's job. 

The hate for service locator, and thought of an anti-pattern as always needs to be balanced with what you're trying to achieve. I'm fine with a framework using it under the covers to provide DI in situations like this, as I'm sure aspnet core team were. Great work on finding that code too, we can def profit from the work they've already done. 

I think this PR is a great idea though and like you mentioned I think you deliver it as an alternative as a separate package for folks that like the ctor approach.

sungam3r:
Actually the best option for me if this (resolving `Func<T>`) worked out of the box in Microsoft.DependencyInjection package.

KreativJos:
@corstian If the schema is a singleton, then it should not be changed while running the application. Why shouldn't we be able to change the schema at runtime? What if I want to:
1. add a field
1. remove a deprecated field
1. fix a bug in a specific field
1. have a schema based on an external service, which can change and thereby automatically would update your schema

So far I do not see why the schema should be immutable, except for the very expensive building part.

@sungam3r It would be nice if you'd make this idea available based on a separate package. It would be a good option to have indeed.

corstian:
@KreativJos there are several reasons:

1. **You would be changing a (public facing) API dynamically, during runtime...**
2. While live patching of programs is something which can be done, there are easier, and more predictable ways to do so by the use of load-balancers and running multiple instances of your api which can be updated side by side.
3. For the last use case you could consider replacing the singleton type registered with your DI container whenever an external change is registered. It would solve the expense which comes with rebuilding the scheme every request.

Nothing prevents you from registering the schema and related types as scoped, but it directly goes against the technical core which makes the dynamic nature of this project possible with quite decent performance. For further discussion about registering the schema with a transient or scoped scope I would recommend you to create a new issue about that.

sungam3r:
@corstian I completely agree.

duvet86:
Sorry to barge in.
If it can help, the issue disappeared when I replaced .net core DI with autofac.
```
        public IServiceProvider ConfigureServices(IServiceCollection services)
        {
            services.AddSingleton<IDependencyResolver>(s => new FuncDependencyResolver(s.GetRequiredService));

            services.AddSingleton<IDocumentExecuter, DocumentExecuter>();
            services.AddSingleton<IDocumentWriter, DocumentWriter>();

            services.AddSingleton<UserItemGraphType>();
            services.AddSingleton<UserItemQuery>();
            services.AddSingleton<RootQuery>();
            services.AddSingleton<ISchema, UserItemSchema>();

            services.AddApplicationInsightsTelemetry(Configuration);
            services.AddCors();
            services.AddMvc()
                .SetCompatibilityVersion(CompatibilityVersion.Version_2_2);

            services
                .AddDbContext<UserItemsDbContext>(options =>
                {
                    options.UseSqlServer(
                        Configuration.GetConnectionString("DefaultConnection"),
                        sqlServerOptionsAction: sqlOptions =>
                        {
                            sqlOptions.MigrationsAssembly(typeof(Startup).GetTypeInfo().Assembly.GetName().Name);
                            //Configuring Connection Resiliency: https://docs.microsoft.com/en-us/ef/core/miscellaneous/connection-resiliency 
                            sqlOptions.EnableRetryOnFailure(maxRetryCount: 10, maxRetryDelay: TimeSpan.FromSeconds(30), errorNumbersToAdd: null);
                        });
                });

            var container = new ContainerBuilder();
            container.Populate(services);

            return new AutofacServiceProvider(container.Build());
        }
```

sungam3r:
Ok. But this requires to build `IServiceProvider` manually via additional dependency and change method signature (coupling to asp.net core infrastructure).

benmccallum:
Honestly wonder how many ppl are creating a graphql-dotnet project with .NET 4.x... I know we should care, but can we drop it for v3.0+?

joemcbride:
3.0 targets `netstandard2.0` and uses `IServiceProvider`.

benmccallum:
Ah, awesome!

Shane32:
My thoughts on various comments above:

> What I think is needed is a way to move out resolver logic which depends on different services (and scoped stuff) into something that can be instantiated per request. Field resolver injection would make that easier:
>
> (context, [InjectedService] myResolverType) => myResolver.Resolve(context)

Way cool.  Not sure about how practical it would be to implement.  Is that what DelegateFieldModelBinderResolver does?

> Field resolver injection (on the one hand a handy thing, really) in this regard is in fact only a convenient tool to hide this fact from the eyes. The class will be successfully created (since the dependencies are removed from the constructor), but if you do not add the dependency to the container, the error will pop up only during the execution of a specific resolver. These are very nasty runtime errors.

No, they are not 'very nasty runtime errors'.  This is the way dependency injection is designed -- to create classes when they are needed, and dispose of them when they are not.

> Theoretically, you can get rid of such errors, but then you have to make your own system of field resolver injection parameters validation that may turn out to be much more complicated than to use an already working constructor injection validation services (in Microsoft.Extensions.DependencyInjection or others).

Microsoft's dependency injection has no more of an idea if your scoped class can be instated successfully than any other code.  For example, in an Asp.Net Core web app, there are no compile-time errors thrown if you have a invalid parameter for a controller constructor, or if one of the scoped services cannot be created.  Only when you try to visit a page on the controller do errors occur.

> There is nothing to be upset about. What I meant is that the pull-request here is a workaround for the real issue (I should've said that, it would have been more explicit).

Correct.  This pull request is a poor workaround for vitally missing functionality.

> In terms of the discussion above, fundamentally it just makes sense that the schema be singleton; it shouldn't change once setup and it should be as optimal as possible. If you wanted to do have it scoped, theoretically you could do that now, right? Register your schema as scoped and measure the hit.

Agreed.

> By the way, now I'm not sure that this code should be merged at all. In fact, this code does not apply to graphql-dotnet. It can be implemented as an external library. That's what happens in our projects. There is a shared library with extensions for DI. Developers use AddFunc extension method - that's all.

Completely agree - this code should never be merged in.  The AddFunc extension method creates a scope for each dependency.  This would completely break database software which rely on different objects to connect to a single dbcontext.  Further, the scope is not disposed of!  Even if you were to dispose of the object you instated, there's no guarantee that it didn't rely upon other services via DI which also need to be disposed of.  This again is very common in a database framework.

> I mean, it would be bad if the IScopeProvider interface (and others) becomes part of graphql-dotnet. This will create an unnecessary dependency in our components on the graphql-dotnet or it will force us to write adapters from one interface to another. Therefore, I find it difficult to say how to proceed. For now, this PR from is most likely an example (demo) of how "scoped problem" can be solved.

I disagree.  Dependency Injection is a core part of .Net Standard 2.0 with standardized interfaces that can be implemented by any modern dependency injection framework.  Although IServiceProvider is the only interface directly included in .Net Standard 2.0, IServiceScope is a part of the Microsoft.Extensions.DependencyInjection library, which should also be supported by modern DI frameworks.  Even if it was not, a few lines of 'adapter' code could easily be written for any DI framework, which would be standardized on Microsoft's recommended interface, rather than a graphql-dotnet specific one.

> to elaborate on the 'antipattern'; the resolvers themselves represent a way to access data. These functions (or just the field definitions in general) are considered immutable in order to guarantee consistency of the graph, and therefore (by convention) these are defined in the constructor. From this point of view it does not make sense to declare these services as scoped, unless you are doing something which mutates the schema or graph types themselves. In this way I think that defining the schema and related types as singleton makes perfect sense.

Agreed, "defining the schema and related types as singleton makes perfect sense".  It certainly allows for fastest execution, but if you lose the ability to put scoped services in the constructor, then the choice is more in debate.

> When thinking about these resolvers as the actual materialization of the query it would make sense (in my humble opinion) to inject the properties directly in the resolver function, whenever executed. For this the IServiceProvider could be used to have support for all dependency providers which implement this specific interface, just the way @johnrutherford has proposed before.

Agreed, that makes sense

> I consider [the Func sample code] to be a better approach than directly injecting the IServiceProvider into the resolver as it makes it pretty clear what dependencies the resolver has, where those are used, and what the scope of said dependency is. I feel like the reasons which make the service locator pattern an anti-pattern, as given in this answer on SO, can all be invalidated for the current use-case.

Strongly disagree with this code sample, for reasons stated above (scope not disposed, cannot create multiple objects in a scope, etc).  Additional clarity would be nice as to which dependencies the resolver has.  But the way I see it, the only other option would be to make the graphtype a scoped service itself.  This would truly be the best answer in terms of code clarity, as it would then lay out like an Asp.Net Core controller, with dependencies in the constructor.  One could also change all the field definitions into methods with attributes, making it even cleaner, and the singleton schema could inspect the graphtype without creating an instance of it, and even ready MethodInfo references to the methods for fastest execution.

> The hate for service locator, and thought of an anti-pattern as always needs to be balanced with what you're trying to achieve. I'm fine with a framework using it under the covers to provide DI in situations like this, as I'm sure aspnet core team were.

Agreed

sungam3r:
>  The AddFunc extension method creates a scope for each dependency. 

You misunderstood something. `AddFunc` **does not** create any scope. Take a closer look at the code. AddFunc only creates a singleton-proxy, the implementation of which delegates the call back to DI-container. So if you use `AddFunc<T>` than `T` **may be of any scope** - transient, scoped, singleton, custom - as it was configured itself in DI.

All of your further reasoning is based on this incorrect judgment.


> Dependency Injection is a core part of .Net Standard 2.0

Absolutely not. Well, you mix different things. Dependency injection is not included in netstandard2.0. `Microsoft.Extensions.DependencyInjection` **targeted** to the netstandard2.0. There are many packages that target netstandard2.0, but that **does not make them part of it**. By the way, let's take GraphQL as an example :).

Nevertheless, I am glad that you have read the discussions in this issue. I hope they help you understand our concerns and motives.

Shane32:
Correct; I did misunderstand AddFunc.  Correct me if I'm wrong, but if used in the sample manner provided (with AspNetCoreHttpScopeProvider), it would access IScopeProvider which accesses the IHttpContextAccessor and obtains the existing scope from there.  Note that IHttpContextAccessor relies on AsyncLocal to store the scope.  This still prevents me from using ParallelExecutionStrategy, as all requests would always retrieve the same scope, and that in turn prevents me from using the DataLoader.

If a custom IScopeProvider was written to always return a new scope, then the scope would never be disposed.

So as far as I can tell, to use a parallel execution strategy, I would have to write a custom MyScopeProvider which creates an AsyncLocal variable to store the current scope, for use by the IScopeProvider singleton.  A custom ExecutionStrategy would need to be written to tell MyScopeProvider when to create another scope.

Is there anything I'm missing (to accomplish my goal)?  I understand other people may not use EF, and might not care about being locked to a SerialExecutionStrategy, but that is not me.  I do understand your view about AddFunc, putting the DI in the constructor rather than in the resolver, even if that's not what I would do.

sungam3r:
> it would access IScopeProvider which accesses the IHttpContextAccessor and obtains the existing scope from there

Yes, that's what it's intended.

> This still prevents me from using ParallelExecutionStrategy, as all requests would always retrieve the same scope, and that in turn prevents me from using the DataLoader.

Why do you think so? Does this interfere with the use of dataloader?

> If a custom IScopeProvider was written to always return a new scope, then the scope would never be disposed.

If the developer writes his own scope provider, then he needs it and he understands what he is doing.

> Is there anything I'm missing (to accomplish my goal)?

I am sorry but what is your goal? I get the feeling that you want to create your own scopes for each field. Keep in mind that then you actually get 2 scopes - one "external"/"original" scope from hosting framework (ASP.NET Core or some other) and the other "internal" - your scope. I wouldn’t do that. #1160 needs to be considered as a way of **transferring** an external scope to the library code - so we have one single scope but it is implicit and we have not to manage it manually.

Maybe you really don’t need an internal scope? Perhaps it is needed for a specific object. In this case, you can make a separate factory for it and use that scope in that factory.

Shane32:
My primary goal is to use the CollectionBatchDataLoader to retrieve multiple pieces of information from my underlying database in a single call.  I had understood the dataloader to be triggered by the DocumentListener, which listens for BeforeExecutionStepAwaitedAsync.  For the SerialExecutionStrategy, this is executed once for every node, as it should.  Therefore, the dataloader would never load a batch of items with the SerialExecutionStrategy.  So I must use ParallelExecutionStrategy.  My pre-existing EF database code uses many classes and relies heavily upon dependency injection.  So at the moment it would seem that I must create a scope within the resolve function, access the database, and dispose of it through there, just as one would create a HttpClient for a REST call.  Obviously I can just create a scope within the resolve function using CreateScope() on an IServiceProvider passed into the constructor of the graphtype.  So my secondary goal is to find a more elegant solution to this problem, eliminating much redundant code and letting infrastructure handle scope rather than each resolve function.

sungam3r:
Why not use one scope for entire graphql request?

> So I must use ParallelExecutionStrategy

Must? `ParallelExecutionStrategy` is selected automatically for queries, `SerialExecutionStrategy` for mutations.

Shane32:
If I use one scope for the entire graphql request, then EF will give the well-known exceptions when resolving queries on two different threads.

sungam3r:
To solve this problem, there is a pool - https://stackoverflow.com/questions/48443567/adddbcontext-or-adddbcontextpool 

I have a feeling that most of your problems will naturally disappear thanks to the pool.

Shane32:
I really don't understand what this has to do with the problem.  This only solves one tiny use case -- where you are directly accessing a dbcontext from the resolve function.  My codebase has hundreds of interdependent classes, all relying on dependency injection to operate.  It would seem that you suggest that I remove DI from all my classes, create a dbcontext within the resolve function, and pass it to any class that needs it.  I don't understand ... why this is so hard to understand.  My services rely on a DI scope.  Either I have a single scope per query, or a single scope per resolve function.  Either way works.  But if I have a single scope per query (which absolutely requires a serial execution strategy), DataLoaders do not function.  Maybe this entire conversation should be under  #1191

sungam3r:
>  This only solves one tiny use case -- where you are directly accessing a dbcontext from the resolve function. 

Really?

> My codebase has hundreds of interdependent classes, all relying on dependency injection to operate

If this is true, then you should have no problem using the context pool.

>  It would seem that you suggest that I remove DI from all my classes, create a dbcontext within the resolve function, and pass it to any class that needs it.

**Completely not.**

>  But if I have a single scope per query (which absolutely requires a serial execution strategy) DataLoaders do not function

One completely does not follow from the other, no need to equate scoping with multithreading support. You may want to use scoped services which designed to work in parallel on many threads in one scope of handling user request. Regarding dataloader - it is another issue and and I'm not ready to talk about it now.

Maybe someone else will be able to help you to find answers.

sungam3r:
>If I use one scope for the entire graphql request, then EF will give the well-known exceptions when resolving queries on two different threads.

> To solve this problem, there is a pool 

@Shane32  I was wrong, details in #1310

sungam3r:
> It would be nice if you'd make this idea available based on a separate package. It would be a good option to have indeed.

@KreativJos @corstian @pekkah @Cooke @benmccallum @Shane32 @joemcbride Done in https://github.com/sungam3r/SteroidsDI . I updated the example PR. I also want to pay attention to `Defer<T>` as a replacement for `Func<T>`. One can call `AddDefer()` once, no need to call `AddFunc<T>` for each T (but this feature remains).

sungam3r:
Tests failed because `SimpleContainer` is not intended to be used with open generic as service type. PR will be closed soon in favor of #1326 . @Shane32 As I said earlier I think that we should support resolver level injection as well and mention about different DI ways/patterns in docs.

sungam3r:
I updated readme https://github.com/sungam3r/SteroidsDI/blob/master/README.md with the necessary information. Thank you all for the discussion! I close this PR and I hope that the given solution will help someone. I like the fact that GraphQL (like any other project) **does not require any changes** to support it. Nonetheless, we have to do resolver-level DI but this is another story.

Feel free to ask further questions if any.

zaneclaes:
@sungam3r I'm struggling to parse apart these issues as someone who has not followed the entire process. I landed here due to this issue: https://github.com/graphql-dotnet/graphql-dotnet/issues/1198 and am struggling to get `GraphQL-dotnet` to work with `EFCore` and `AddDbContextPool` based upon the provided examples. This seems like a very common use-case, but I'm struggling to find a clear example of how to gain access to my `DbContext` — which, as a scoped DI, is unable to be accessed from a Singleton (as the examples imply should be used for the Schemas)`.

In other words, this fails:

```
services.AddDbContextPool<MyContext>(options => { ... });

services.AddSingleton<MySchema>()
             .addGrahpQL( ... )
```

Where:
```
  public class MySchema : Schema {
    public MySchema(MyContext data, IServiceProvider provider) : base(provider) {
      Query = new MyQuery(data);
    }
  }
```

Shane32:
@zaneclaes The current recommendations are listed here:

https://graphql-dotnet.github.io/docs/getting-started/dependency-injection/#scoped-services-with-a-singleton-schema-lifetime

I'd probably create an extension method like this:

```csharp
public static class MyExtensions
{
    public static MyDbContext GetDbContext(this IResolveFieldContext context)
        => context.RequestServices.GetRequiredService<MyDbContext>();
}
```

Then your graphs can contain code like this:
```cs
Field<ListGraphType<MachineProfileType>>("getMachineProfiles", resolve: ctxt => ctxt.GetDbContext().MachineProfiles.Take(100));
```

Shane32:
However, this only works with EF Core if you are using a serial execution strategy.  Using a parallel execution strategy (the default) with EF Core requires you to create a scope for the resolver, or register the DbContext as a transient.  See here:

https://graphql-dotnet.github.io/docs/getting-started/dependency-injection/#thread-safety-with-scoped-services

Shane32:
@zaneclaes I'd recommend reading the entire page on dependency injection, as it was recently updated, after the majority of this issue's comments were posted.

PR #1164, opened 2019-05-28
Exception delegate (for logs etc.)

### LINKED ISSUE #1152: Logging handler for exceptions
## Summary

A way to configure a logging delegate for exceptions

### Basic example

Naming here is probably terrible, but just to give an idea:

```
_schema.Execute(configure =>
            {
                configure.Query = request.Query;

                configure.ExceptionEvent(exception => _logger.Log(LogLevel.Error, exception));
            });
```

### Motivation

Currently I only have a couple of options for how to deal with exceptions:

- normal behaviour swallows stack traces
- expose stack traces via API
- `ThrowOnUnhandledExceptions` to bubble them up to the calling code

Currently i've got `ThrowOnUnhandledExceptions` set to `true` so that I can log the errors, and have to pass back a 5xx response. Ideally I would have the normal behaviour from the API, but be able to trigger a logging function when there's an issue so that I can keep on top of errors in production but also present a proper GraphQL endpoint.

Side note (/ bug?): It also appears that `ThrowOnUnhandledExceptions` is throwing all exceptions including `ExecutionError` ones? Perhaps due to https://github.com/graphql-dotnet/graphql-dotnet/blob/38a85611aabcabc2344b91d92760ca3dbb00fddb/src/GraphQL/Execution/DocumentExecuter.cs#L206 ?

Happy to put up a PR for this feature if it's thought to be a good idea / there's not a better alternative.

sungam3r:
fixes #1152

Also adds the ability to override an exception. This can be useful for hiding error messages that reveal server implementation details.


sungam3r:
@joemcbride Can it be merged?

src/GraphQL/Execution/ExecutionResultJsonConverter.cs
```diff
@@ -86,7 +85,7 @@ private void WriteErrors(ExecutionErrors errors, JsonWriter writer, JsonSerializ
 
         private void WriteErrorExtensions(ExecutionError error, JsonWriter writer, JsonSerializer serializer)
         {
-            if (string.IsNullOrWhiteSpace(error.Code) && error.Data?.Count == 0)
+            if (string.IsNullOrWhiteSpace(error.Code) && (error.Data == null || error.Data.Count == 0))
```
sungam3r: Condition `error.Data?.Count == 0` gives `false` if Data is null because null <> 0, so NRE may occur later

src/GraphQL/Execution/ExecutionResultJsonConverter.cs
```diff
@@ -131,9 +130,8 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             throw new NotImplementedException();
         }
 
-        public override bool CanConvert(Type objectType)
-        {
-            return objectType == typeof(ExecutionResult);
-        }
+        public override bool CanRead => false;
```
sungam3r: `ReadJson` throws `NotImplementedException` so it will be better explicitly return `false` instead of `true` in base class.

sungam3r:
Now exception delegate accepts additional context parameter which can be used to augment exception with appropriate information.

StefH:
@sungam3r Can you fix merge conflicts on this PR?

sungam3r:
@StefH Done.

sungam3r:
Fixed merge error.

mikelneo:
@sungam3r may be you can add some unit tests for this feature?

sungam3r:
@mikelneo Done.

PR #1172, opened 2019-06-07
Support custom directives in Schema First

joemcbride:
```csharp
var schema = Schema.For(@"
    type Query {
        hello: String @upper
    }
",
_ =>
{
    _.RegisterDirectiveVisitor<UppercaseDirectiveVisitor>("upper");
});

{
      _.Query = "{ hello }";
      _.Root = new { Hello = "Hello World!" };
});

public class UppercaseDirectiveVisitor : SchemaDirectiveVisitor
{
    public override void VisitField(FieldType field)
    {
        var inner = field.Resolver ?? new NameFieldResolver();
        field.Resolver = new FuncFieldResolver<object>(context =>
        {
            var result = inner.Resolve(context);

            if (result is string str)
            {
                return str.ToUpperInvariant();
            }

            return result;
        });
    }
}
```

### DIFF
--- a/src/GraphQL.Tests/ServiceProviderExtensions.cs
+++ b/src/GraphQL.Tests/ServiceProviderExtensions.cs
@@ -3,19 +3,11 @@ using GraphQL.StarWars.IoC;
 
 namespace GraphQL.Tests
 {
-    public static class ServiceProviderExtensions
-    {
-        public static T Resolve<T>(this IServiceProvider services)
-        {
-            return (T)services.GetService(typeof(T));
-        }
-    }
-
-    public class SimpleContainerAdapater : IServiceProvider
+    public class SimpleContainerAdapter : IServiceProvider
     {
         private ISimpleContainer _container;
 
-        public SimpleContainerAdapater(ISimpleContainer container)
+        public SimpleContainerAdapter(ISimpleContainer container)
         {
             _container = container;
         }
--- a/src/GraphQL.Tests/ServiceProviderExtensions.cs
+++ b/src/GraphQL.Tests/ServiceProviderExtensions.cs
@@ -1,0 +1,28 @@
+using System;
+using GraphQL.StarWars.IoC;
+
+namespace GraphQL.Tests
+{
+    public static class ServiceProviderExtensions
+    {
+        public static T Resolve<T>(this IServiceProvider services)
+        {
+            return (T)services.GetService(typeof(T));
+        }
+    }
+
+    public class SimpleContainerAdapater : IServiceProvider
+    {
+        private ISimpleContainer _container;
+
+        public SimpleContainerAdapater(ISimpleContainer container)
+        {
+            _container = container;
+        }
+
+        public object GetService(Type serviceType)
+        {
+            return _container.Get(serviceType);
+        }
+    }
+}
--- a/src/GraphQL/IDependencyResolver.cs
+++ b/src/GraphQL/IDependencyResolver.cs
@@ -5,6 +5,7 @@ namespace GraphQL
     /// <summary>
     /// Basic DependencyResolver
     /// </summary>
+    [Obsolete("Use IServiceProvider instead")]
     public interface IDependencyResolver
     {
         /// <summary>
@@ -23,43 +24,48 @@ namespace GraphQL
     /// <summary>
     /// Dependency resolver based on Activator.CreateInstance
     /// </summary>
-    /// <seealso cref="GraphQL.IDependencyResolver" />
-    public class DefaultDependencyResolver : IDependencyResolver
+    /// <seealso cref="System.IServiceProvider" />
+    public sealed class DefaultServiceProvider : IServiceProvider
     {
         /// <summary>
-        /// Resolves the specified type.
+        /// Gets the specified service type.
         /// </summary>
-        /// <typeparam name="T">Desired type</typeparam>
-        /// <returns>T.</returns>
-        public T Resolve<T>()
-        {
-            return (T)Resolve(typeof(T));
-        }
-
-        /// <summary>
-        /// Resolves the specified type.
-        /// </summary>
-        /// <param name="type">Desired type</param>
-        /// <returns>An instance of <paramref name="type"/>.</returns>
-        public object Resolve(Type type)
+        /// <param name="serviceType">Desired type</param>
+        /// <returns>An instance of <paramref name="serviceType"/>.</returns>
+        public object GetService(Type serviceType)
         {
             try
             {
-                return Activator.CreateInstance(type);
+                return Activator.CreateInstance(serviceType);
             }
             catch (Exception exception)
             {
-                throw new Exception($"Failed to call Activator.CreateInstance. Type: {type.FullName}", exception);
+                throw new Exception($"Failed to call Activator.CreateInstance. Type: {serviceType.FullName}", exception);
             }
+        }
+    }
+
+    public sealed class DependencyResolverToServiceProviderAdapter : IServiceProvider
+    {
+        private readonly IDependencyResolver _resolver;
+
+        public DependencyResolverToServiceProviderAdapter(IDependencyResolver resolver)
+        {
+            _resolver = resolver;
+        }
+
+        public object GetService(Type serviceType)
+        {
+            return _resolver.Resolve(serviceType);
         }
     }
 
     /// <summary>
     /// Func based dependency resolver.
     /// </summary>
-    /// <seealso cref="GraphQL.IDependencyResolver" />
+    /// <seealso cref="System.IServiceProvider" />
     /// <remarks>This is mainly used as an adapter for other dependency resolvers such as DI frameworks.</remarks>
-    public class FuncDependencyResolver : IDependencyResolver
+    public sealed class FuncDependencyResolver : IServiceProvider
     {
         private readonly Func<Type, object> _resolver;
 
@@ -75,33 +81,10 @@ namespace GraphQL
         /// <summary>
         /// Resolves the specified type.
         /// </summary>
-        /// <typeparam name="T">Desired type</typeparam>
-        public T Resolve<T>()
-        {
-            return (T)Resolve(typeof(T));
-        }
-
-        /// <summary>
-        /// Resolves the specified type.
-        /// </summary>
         /// <param name="type">Desired type</param>
-        public object Resolve(Type type)
+        public object GetService(Type type)
         {
             return _resolver(type);
         }
-    }
-
-    internal sealed class ServiceProviderAdapter : IDependencyResolver
-    {
-        private readonly IServiceProvider _services;
-
-        public ServiceProviderAdapter(IServiceProvider services)
-        {
-            _services = services;
-        }
-
-        public T Resolve<T>() => (T)_services.GetService(typeof(T));
-
-        public object Resolve(Type type) => _services.GetService(type);
     }
 }
--- a/src/GraphQL/IDependencyResolver.cs
+++ b/src/GraphQL/IDependencyResolver.cs
@@ -45,7 +45,7 @@ namespace GraphQL
         }
     }
 
-    public sealed class DependencyResolverToServiceProviderAdapter : IServiceProvider
+    internal sealed class DependencyResolverToServiceProviderAdapter : IServiceProvider
     {
         private readonly IDependencyResolver _resolver;
 
@@ -61,13 +61,14 @@ namespace GraphQL
     }
 
     /// <summary>
-    /// Func based dependency resolver.
+    /// Func based service provider.
     /// </summary>
     /// <seealso cref="System.IServiceProvider" />
     /// <remarks>This is mainly used as an adapter for other dependency resolvers such as DI frameworks.</remarks>
+    [Obsolete("Use GraphQL.FuncServiceProvider instead.")]
     public sealed class FuncDependencyResolver : IServiceProvider
     {
-        private readonly Func<Type, object> _resolver;
+        private readonly FuncServiceProvider _provider;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="FuncDependencyResolver"/> class.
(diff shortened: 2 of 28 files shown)


src/GraphQL.Tests/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static class ServiceProviderExtensions
+    {
+        public static T Resolve<T>(this IServiceProvider services)
+        {
+            return (T)services.GetService(typeof(T));
+        }
+    }
+
+    public class SimpleContainerAdapater : IServiceProvider
```
sungam3r: Adapater

src/GraphQL/IDependencyResolver.cs
```diff
@@ -23,43 +24,48 @@ public interface IDependencyResolver
...
+    }
+
     /// <summary>
     /// Func based dependency resolver.
     /// </summary>
-    /// <seealso cref="GraphQL.IDependencyResolver" />
+    /// <seealso cref="System.IServiceProvider" />
     /// <remarks>This is mainly used as an adapter for other dependency resolvers such as DI frameworks.</remarks>
-    public class FuncDependencyResolver : IDependencyResolver
+    public sealed class FuncDependencyResolver : IServiceProvider
```
sungam3r: May it be better to call it `FuncServiceProvider` ?

src/GraphQL.Tests/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using GraphQL.StarWars.IoC;
+
+namespace GraphQL.Tests
+{
+    public static class ServiceProviderExtensions
+    {
+        public static T Resolve<T>(this IServiceProvider services)
```
sungam3r: Changing `Resolve<T>` to `GetService<T>` will be a closer term to `IServiceProvider`

src/GraphQL/Reflection/AccessorFieldResolver.cs
```diff
@@ -7,9 +7,9 @@ namespace GraphQL.Reflection
     internal class AccessorFieldResolver : IFieldResolver
     {
         private readonly IAccessor _accessor;
-        private readonly IDependencyResolver _dependencyResolver;
+        private readonly IServiceProvider _dependencyResolver;
 
-        public AccessorFieldResolver(IAccessor accessor, IDependencyResolver dependencyResolver)
+        public AccessorFieldResolver(IAccessor accessor, IServiceProvider dependencyResolver)
```
sungam3r: I think it is worth renaming all the arguments with `dependencyResolver` name to `serviceProvider` for the purpose of not mentioning now obsolete interface `IDependencyResolver` and thereby not confusing users

src/GraphQL/Types/EnumerationGraphType.cs
```diff
@@ -112,11 +113,21 @@ public void Add(EnumValueDefinition value)
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
     }
 
-    public class EnumValueDefinition
+    public class EnumValueDefinition : IProvideMetadata
```
sungam3r: Is it required to support directives?

src/GraphQL/Types/Schema.cs
```diff
@@ -8,6 +8,8 @@ namespace GraphQL.Types
 {
     public interface ISchema : IDisposable
     {
+        IServiceProvider Services { get; }
```
sungam3r: It seems to me that you should not give access to this property through the interface. This encourages users to use the ServiceLocator anti-pattern in user-code (mostly in resolvers). I understand that the `IServiceProvider` is already in the scheme, but it is used for the internal needs of the scheme itself. User-code can resolve dependencies without a ServiceLocator using standard dependency injection in constructor - #1160.

src/GraphQL/Types/Schema.cs
```diff
@@ -265,7 +259,7 @@ private GraphTypesLookup CreateTypesLookup()
             return GraphTypesLookup.Create(
                 types,
                 _directives,
-                type => DependencyResolver.Resolve(type) as IGraphType,
+                type => Services.GetService(type) as IGraphType,
```
sungam3r: Consider using `GetRequiredService` extension method where necessary instead of `GetService`. `GetService` [**can**](https://docs.microsoft.com/en-us/dotnet/api/system.iserviceprovider.getservice?view=netcore-2.1#System_IServiceProvider_GetService_System_Type_) return null.

```c#
    public static object GetRequiredService(this IServiceProvider provider, Type serviceType)
    {
      if (provider == null)
        throw new ArgumentNullException(nameof (provider));
      if (serviceType == (Type) null)
        throw new ArgumentNullException(nameof (serviceType));
      object service = provider.GetService(serviceType);
      if (service != null)
        return service;
      throw new InvalidOperationException($"Required service for type {serviceType} not found"));
    }
```

This will make it possible to find an error at an early stage, before a possible `null` transmission to the underlying code.

src/GraphQL/Utilities/DeprecatedDirectiveVisistor.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace GraphQL.Utilities
+{
+    public class DeprecatedDirectiveVisistor : SchemaDirectiveVisitor
+    {
+        protected static readonly string DeprecatedDefaultValue = DirectiveGraphType.Deprecated.Arguments.Find("reason").DefaultValue.ToString();
+
+        public override void VisitField(FieldType field)
+        {
+            // if a value has already been set, prefer that
+            if(!string.IsNullOrWhiteSpace(field.DeprecationReason)) return;
```
sungam3r: whitespace before `(` there and all around

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -14,10 +14,17 @@ public class SchemaBuilder
...
+        public IServiceProvider ServiceProvider { get; set; } = new DefaultServiceProvider();
 
         public TypeSettings Types { get; } = new TypeSettings();
 
+        public IDictionary<string, Type> Directives { get; } = new Dictionary<string, Type>
+        {
+            {"deprecated", typeof(DeprecatedDirectiveVisistor)}
+        };
+
+        private List<IVisitorSelector> Visitors { get; } = new List<IVisitorSelector>();
```
sungam3r: `VisitorSelectors` or just `Selectors` property name suits better. I was a little confused when I saw that visitors has other visitors inside.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -14,10 +14,17 @@ public class SchemaBuilder
...
         private readonly IDictionary<string, IGraphType> _types = new Dictionary<string, IGraphType>();
 
-        public IDependencyResolver DependencyResolver { get; set; } = new DefaultDependencyResolver();
+        public IServiceProvider ServiceProvider { get; set; } = new DefaultServiceProvider();
 
         public TypeSettings Types { get; } = new TypeSettings();
 
+        public IDictionary<string, Type> Directives { get; } = new Dictionary<string, Type>
+        {
+            {"deprecated", typeof(DeprecatedDirectiveVisistor)}
```
sungam3r: whitespaces

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -49,7 +56,12 @@ private static GraphQLDocument Parse(string document)
 
         private ISchema BuildSchemaFrom(GraphQLDocument document)
         {
-            var schema = new Schema(DependencyResolver);
+            if (Directives.Any())
+            {
+                Visitors.Add(new DirectiveVisitorSelector(Directives, t => (SchemaDirectiveVisitor)ServiceProvider.GetService(t)));
```
sungam3r: That is, there will now appear a general mechanism for expanding behavior via visitors/selectors, which, if desired, can be applied not only to directives? Cool.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -151,7 +163,7 @@ private ISchema BuildSchemaFrom(GraphQLDocument document)
 
             var typeList = _types.Values.ToArray();
             typeList.Apply(schema.RegisterType);
-            schema.RegisterDirectives(directives);
+            schema.RegisterDirectives(directives.ToArray());
```
sungam3r: Why? I specifically added overload to avoid the need to create an intermediate array.

src/GraphQL/Types/EnumerationGraphType.cs
sungam3r: I see, `SetAstType` requires this.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -225,16 +239,19 @@ protected virtual FieldType ToFieldType(string parentTypeName, GraphQLFieldDefin
 
             var args = fieldDef.Arguments.Select(ToArguments);
             field.Arguments = new QueryArguments(args);
+            field.DeprecationReason = fieldConfig.DeprecationReason;
```
sungam3r: `ToSubscriptionFieldType` also should have this line. Is not it so?

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -457,7 +444,36 @@ private IGraphType ToGraphType(GraphQLType astType)
...
 
-        private object ToValue(GraphQLValue source)
+        protected virtual void CopyMetadata(IProvideMetadata target, IProvideMetadata source)
+        {
+            source.Metadata.Apply(kv => target.Metadata[kv.Key] = kv.Value);
+        }
+
+        protected virtual void VisitNode(object node, Action<ISchemaNodeVisitor> action)
+        {
+            foreach(var selector in Visitors)
```
sungam3r: `foreach (var selector in Selectors)`

src/GraphQL/Utilities/SchemaBuilderExtensions.cs
```diff
@@ -0,0 +1,25 @@
+using GraphQL.Types;
+using GraphQLParser.AST;
+
+namespace GraphQL.Utilities
+{
+    internal static class SchemaBuilderExtensions
+    {
+        public const string __AST_MetaField__ = "__AST_MetaField__";
```
sungam3r: private const?

src/GraphQL/Utilities/DirectiveVisitorSelector.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public DirectiveVisitorSelector(
+            IDictionary<string, Type> directiveVisitors,
+            Func<Type, SchemaDirectiveVisitor> typeResolver)
+        {
+            _directiveVisitors = directiveVisitors;
+            _typeResolver = typeResolver;
+        }
+
+        public IEnumerable<ISchemaNodeVisitor> Select(object node)
```
sungam3r: These 3 (may be more in future) branches of the code can be reduced to just one, if you create an interface like
```c#
public interface IHasDirectivesDefinition
{
  public IEnumerable<GraphQLDirective> Directives { get; set; }
}
```
in GraphQLParser project and apply it to `GraphQLObjectTypeDefinition`, `GraphQLFieldDefinition`, `GraphQLEnumValueDefinition` classes:

```c#
 public IEnumerable<ISchemaNodeVisitor> Select(object node)
 {
          
                var ast = obj.GetAstType<IHasDirectivesDefinition>();
                if (ast == null) yield break;
                foreach (var visitor in BuildVisitors(ast.Directives))
                {
                    yield return visitor;
                }
}    
```

It will be easier and more flexible.

src/GraphQL/Utilities/SchemaBuilder.cs
sungam3r: This is also place where `GetRequiredService` should be better than `GetService`.

src/GraphQL/IDependencyResolver.cs
joemcbride: Agreed ... I'm thinking to keep this class for backwards compatibility but mark it as obsolete, with a second class called `FuncServiceProvider `.

src/GraphQL/Types/Schema.cs
joemcbride: Agreed - I didn't want to include that extra dependency but I can easily add that as an internal extension method.

sungam3r: Yes, no extra dependency, just a little helper/extension method.

src/GraphQL/Types/Schema.cs
joemcbride: Yeah that was a stop-gap.  I like that pattern of the factory.  I'll remove this.  I'm still thinking on how to fix it without the user needing to do any special tricks.  As a note, this Schema First does not have the same issue as the GraphType First approach.  The resolver classes get resolved on each request.

src/GraphQL/Utilities/SchemaBuilder.cs
joemcbride: IDE was showing me an error, but that may just be a problem with VSCode.

src/GraphQL/Utilities/SchemaBuilder.cs
joemcbride: It should, thanks!

src/GraphQL/Utilities/DirectiveVisitorSelector.cs
joemcbride: Yes, agreed.  Something like that is needed in the Parser.

src/GraphQL/Types/Schema.cs
sungam3r: The factory creates something definite and is used by certain classes. If an object creates anything and is accessible from anywhere, then this is ServiceLocator.

What issue?

src/GraphQL/Utilities/DirectiveVisitorSelector.cs
sungam3r: I can do it. Ok?

joemcbride: That would be great, thanks!

sungam3r: https://github.com/graphql-dotnet/parser/pull/34

There are more changes than are required for this task, but they are fairly simple.

src/GraphQL/Types/Schema.cs
joemcbride: The dependency scoping issue.

sungam3r: I did not quite understand, is there any difference in the way a schema was created? The result is the same. In any case, it’s all about not introducing client code into the temptation to use ServiceLocator. I carefully read #648 and saw in given examples all those rakes that I myself went to. All these attempts at any cost simply drag the magic object into any point of the code in order to create anything through it. This puts the whole point o DI upside down - letting the container itself create dependencies and pass them on to the constructors.

By the way, we are now successfully applying the approach proposed in #1160 in production - no manual resolving, no ServiceLocator, no tight coupling with `HttpContext/IHttpContextAccessor` + also simultaneous support of multiple scope types in one application.

sungam3r:
LGTM

sungam3r:
\+ `IHasDirectivesNode` to merge those three branches of code.

src/GraphQL/Types/Schema.cs
joemcbride: > is there any difference in the way a schema was created

Yes - today, if you use the GraphType First approach then your GraphType's will automatically take upon the lifetime of the Schema.  That won't happen with Schema First because the field resolver type is always fetched from the container each time a field is executed.  This is because with Schema First your schema definition and your resolvers are two different classes.

> By the way, we are now successfully applying the approach

Awesome!

joemcbride:
This may need more work in the future but going to merge and see how it goes.

cluedtke:
Is this feature available in a 3.0 preview build? I pulled `3.0.0-preview-1175` but am seeing;
> The type or namespace name 'SchemaDirectiveVisitor' could not be found

sungam3r:
Yes, this feature is in master and available in latest nugets.

PR #1058, opened 2019-03-13
change user context to a dictionary

SimonCropp:
change `ExecutionOptions.UserContext ` to be a `IDictionary<string,object>`

replaces https://github.com/graphql-dotnet/graphql-dotnet/pull/1037

PR #1173, opened 2019-06-07
Change the type of the UserContext-property in some places to IDictionary<string, object>

danielcweber:
The UserContext that is built by IUserContextBuilder.BuildUserContext is an IDictionary<string, object>. That should be reflected in some classes to make code more type safe and avoid tedious casting. Note: It changes an interface, IDocumentValidator, but I think it's ok given the recent change to IUserContextBuilder and the project being in some kind of preview state.

### DIFF
--- a/src/GraphQL/Types/ResolveFieldContext.cs
+++ b/src/GraphQL/Types/ResolveFieldContext.cs
@@ -24,7 +24,7 @@ namespace GraphQL.Types
 
         public object RootValue { get; set; }
 
-        public object UserContext { get; set; }
+        public IDictionary<string, object> UserContext { get; set; }
 
         public TSource Source { get; set; }
 
--- a/src/GraphQL/Validation/DocumentValidator.cs
+++ b/src/GraphQL/Validation/DocumentValidator.cs
@@ -13,7 +13,7 @@ namespace GraphQL.Validation
             ISchema schema,
             Document document,
             IEnumerable<IValidationRule> rules = null,
-            object userContext = null,
+            IDictionary<string, object> userContext = null,
             Inputs inputs = null);
     }
 
@@ -24,7 +24,7 @@ namespace GraphQL.Validation
             ISchema schema,
             Document document,
             IEnumerable<IValidationRule> rules = null,
-            object userContext = null,
+            IDictionary<string, object> userContext = null,
             Inputs inputs = null)
         {
             if (!schema.Initialized)
--- a/src/GraphQL/Validation/ValidationContext.cs
+++ b/src/GraphQL/Validation/ValidationContext.cs
@@ -25,7 +25,7 @@ namespace GraphQL.Validation
 
         public TypeInfo TypeInfo { get; set; }
 
-        public object UserContext { get; set; }
+        public IDictionary<string, object> UserContext { get; set; }
 
         public IEnumerable<ValidationError> Errors => _errors;
 


joemcbride:
Yes we missed a few places for this - thanks!

sungam3r:
Oops, indeed.

andrejohansson:
I can´t find any information why the change to IUserContextbuilder have been made and this change breaks for example authorization which expects a root UserContext object. See [this line](https://github.com/graphql-dotnet/authorization/blob/d500861ae71308d4f69a3dfbe4c58a2517dbda41/src/GraphQL.Authorization/AuthorizationValidationRule.cs#L20) which will always be null after the IDictionary change.

Where is the information regarding the change to a dictionary? How should a dictionary look like in the new preview?

Mine currently looks like this, but its wrong:

```csharp
public class GraphQLUserContextBuilder : IUserContextBuilder
    {
        public Task<IDictionary<string,object>> BuildUserContext(HttpContext httpContext)
        {
            if (httpContext is null)
            {
                throw new ArgumentNullException(nameof(httpContext));
            }

            return Task.FromResult<IDictionary<string,object>>(
                new Dictionary<string, object>()
                {
                   // What should I put here?
                    { "main", new GraphQLUserContext() { User = httpContext.User } }
                }
            );
        }
    }
```

sungam3r:
> See this line which will always be null after the IDictionary change.

Whether null is eveluated on this line or not is in no way associated with the change under discussion. This situation arose on this line even before aforementioned change.

> Where is the information regarding the change to a dictionary? How should a dictionary look like in the new preview? Mine currently looks like this, but its wrong:

https://github.com/graphql-dotnet/authorization/blob/master/src/Harness/GraphQL.cs#L8

andrejohansson:
@sungam3r [the line I linked to looks like this](https://github.com/graphql-dotnet/authorization/blob/d500861ae71308d4f69a3dfbe4c58a2517dbda41/src/GraphQL.Authorization/AuthorizationValidationRule.cs#L20) and is from **AuthorizationValidationRule.cs**:

```csharp           
            var userContext = context.UserContext as IProvideClaimsPrincipal;
```

Since ``context.UserContext`` is not a ``IProvideClaimsPrincipal`` anymore, that line will cause ``userContext`` to become **null** (casting using **as** will lead to null if the cast did not work).

Before ``context.UserContext`` was a ``GraphQLUserContext``, but now it is a ``IDictionary<string,object>`` because of the changes to the interface **IUserContextBuilder**

```csharp
  public interface IUserContextBuilder
  {
    Task<IDictionary<string, object>> BuildUserContext(HttpContext httpContext);
  }
```

With my implementation of ``GraphQLUserContextBuilder`` I can change the line in ``AuthorizationValidationRule.cs`` to this

```csharp           
            var userContext = context.UserContext["main"] as IProvideClaimsPrincipal;
```

And it will work in my case. But what I do not understand is how I should correctly create the usercontext in it´s "dictionary form" in order to make it work with ``GraphQL.Authorization`` (passing along my claims). 

And I do not find any information regarding the refactoring either...and your links does not seem to link to it?

From what I can see there is changes on server side here:
- https://github.com/graphql-dotnet/server/pull/218
- https://github.com/graphql-dotnet/graphql-dotnet/pull/1058
- https://github.com/graphql-dotnet/graphql-dotnet/pull/1037

sungam3r:
> Since context.UserContext is not a IProvideClaimsPrincipal **anymore**

Anymore? Are you talking about interface or implementation? Your context **should implement both** IProvideClaimsPrincipal and IDictionary<string, object>.

PR #1179, opened 2019-06-11
Add the ability to filter an introspection query

joemcbride:
Example usage:

https://gist.github.com/joemcbride/87a5761161748263115f591f4543b6e4

### DIFF
--- a/src/GraphQL/Introspection/ISchemaFilter.cs
+++ b/src/GraphQL/Introspection/ISchemaFilter.cs
@@ -1,0 +1,41 @@
+using GraphQL.Types;
+
+namespace GraphQL.Introspection
+{
+    public interface ISchemaFilter
+    {
+        bool Type(IGraphType type);
+        bool Field(IGraphType parent, IFieldType field);
+        bool Argument(IFieldType field, QueryArgument argument);
+        bool EnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue);
+        bool Directive(DirectiveGraphType directive);
+    }
+
+    public class DefaultSchemaFilter : ISchemaFilter
+    {
+        public bool Type(IGraphType type)
+        {
+            return true;
+        }
+
+        public bool Field(IGraphType parent, IFieldType field)
+        {
+            return true;
+        }
+
+        public bool Argument(IFieldType field, QueryArgument argument)
+        {
+            return true;
+        }
+
+        public bool EnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue)
+        {
+            return true;
+        }
+
+        public bool Directive(DirectiveGraphType directive)
+        {
+            return true;
+        }
+    }
+}
--- a/src/GraphQL/Introspection/ISchemaFilter.cs
+++ b/src/GraphQL/Introspection/ISchemaFilter.cs
@@ -1,41 +1,42 @@
+using System.Threading.Tasks;
 using GraphQL.Types;
 
 namespace GraphQL.Introspection
 {
     public interface ISchemaFilter
     {
-        bool Type(IGraphType type);
-        bool Field(IGraphType parent, IFieldType field);
-        bool Argument(IFieldType field, QueryArgument argument);
-        bool EnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue);
-        bool Directive(DirectiveGraphType directive);
+        Task<bool> AllowType(IGraphType type);
+        Task<bool> AllowField(IGraphType parent, IFieldType field);
+        Task<bool> AllowArgument(IFieldType field, QueryArgument argument);
+        Task<bool> AllowEnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue);
+        Task<bool> AllowDirective(DirectiveGraphType directive);
     }
 
     public class DefaultSchemaFilter : ISchemaFilter
     {
-        public bool Type(IGraphType type)
+        public virtual Task<bool> AllowType(IGraphType type)
         {
-            return true;
+            return Task.FromResult(true);
         }
 
-        public bool Field(IGraphType parent, IFieldType field)
+        public virtual Task<bool> AllowField(IGraphType parent, IFieldType field)
         {
-            return true;
+            return Task.FromResult(true);
         }
 
-        public bool Argument(IFieldType field, QueryArgument argument)
+        public virtual Task<bool> AllowArgument(IFieldType field, QueryArgument argument)
         {
-            return true;
+            return Task.FromResult(true);
         }
 
-        public bool EnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue)
+        public virtual Task<bool> AllowEnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue)
         {
-            return true;
+            return Task.FromResult(true);
         }
 
-        public bool Directive(DirectiveGraphType directive)
+        public virtual Task<bool> AllowDirective(DirectiveGraphType directive)
         {
-            return true;
+            return Task.FromResult(true);
         }
     }
 }
--- a/src/GraphQL/Introspection/SchemaMetaFieldType.cs
+++ b/src/GraphQL/Introspection/SchemaMetaFieldType.cs
@@ -1,4 +1,7 @@
+using System;
+using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using GraphQL.Resolvers;
 using GraphQL.Types;
 
@@ -25,12 +28,12 @@ namespace GraphQL.Introspection
                 "exposes all available types and directives on the server, as well as " +
                 "the entry points for query, mutation, and subscription operations.";
 
-            Field<NonNullGraphType<ListGraphType<NonNullGraphType<__Type>>>>(
+            FieldAsync<NonNullGraphType<ListGraphType<NonNullGraphType<__Type>>>>(
                 "types",
                 "A list of all types supported by this server.",
-                resolve: context =>
+                resolve: async context =>
                 {
-                    return context.Schema.AllTypes.Where(x => context.Schema.Filter.Type(x)).ToList();
+                    return await context.Schema.AllTypes.WhereAsync(x => context.Schema.Filter.AllowType(x));
                 });
 
             Field<NonNullGraphType<__Type>>(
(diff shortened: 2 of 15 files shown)


src/GraphQL/Introspection/SchemaMetaFieldType.cs
```diff
@@ -45,23 +46,31 @@ public __Schema()
                 "If this server supports mutation, the type that mutation operations will be rooted at.",
                 resolve: context =>
                 {
-                    return context.Schema.Mutation;
+                    if(context.Schema.Filter.Type(context.Schema.Mutation))
```
sungam3r: whitespace

src/GraphQL/Introspection/SchemaMetaFieldType.cs
```diff
@@ -45,23 +46,31 @@ public __Schema()
...
+                    return null;
                 });
 
             Field<__Type>(
                 "subscriptionType",
                 "If this server supports subscription, the type that subscription operations will be rooted at.",
                 resolve: context =>
                 {
-                    return context.Schema.Subscription;
+                    if(context.Schema.Filter.Type(context.Schema.Subscription))
```
sungam3r: whitespace

src/GraphQL/Introspection/ISchemaFilter.cs
```diff
@@ -0,0 +1,41 @@
...
+    public interface ISchemaFilter
+    {
+        bool Type(IGraphType type);
+        bool Field(IGraphType parent, IFieldType field);
+        bool Argument(IFieldType field, QueryArgument argument);
+        bool EnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue);
+        bool Directive(DirectiveGraphType directive);
+    }
+
+    public class DefaultSchemaFilter : ISchemaFilter
```
sungam3r: If you intend to use this "hide feature", then you will probably need to define only a part of the behavior, and reuse the rest, so it can be convenient to do all those methods `virtual`.

sungam3r:
It is clear what makes this PR, but is there a real need for this? In principle, I can imagine a situation where it can be useful. There is also the issue of hiding types without hiding fields of this type (by negligence). There is some risk of unobvious behavior.

sungam3r:
And I understand correctly that this filtering does not protect from actually getting hidden parts of the scheme? But the question of protection is most likely not worth it.

GrangerAts:
@sungam3r - I apologize if it's not appropriate for me to jump in here, but I have a significant interest in this functionality. 

TL;DR - When you have to have field/argument level authorization in your graph, the signal-to-noise ratio is inversely proportional to the size/complexity of the schema's query. The primary concern is that the schema needs to be able to filter out things that the current account is not authorized to use.


Since the current "best practice" for GraphQL is to have one endpoint to rule them all, there is a significant problem for those implementing a solution for multi-tenant systems. The more stuff you have in there, the more difficult it is for people to figure out how they can use it, given the permissions they have with any given tenant. And for some custom fields, only a subset of tenants should even know it exists, or you can end up with some "interesting" client-relations issues; think about "employee-only, backend" mutations or query-fields.

Or, in other words, if the schema's query ends up with 500+ top-level GraphTypes, it's a horrible custom experience to tell someone to "read the docs; they'll tell you the authZ" so they can figure out which fields they're allowed to query and which ones are just "noise" to them.

So this new filtering functionality that's being proposed. I really, really want it, and it really needs to be easy to tie it into the authZ paradigm, since that's what'd driving the need.

GrangerAts:
> 
> 
> So this new filtering functionality that's being proposed. I really, really want it, and it really needs to be easy to tie it into the authZ paradigm, since that's what'd driving the need.
> **Edit**: It looks like the direction of the PR's changes is to make something completely independent of the authZ paradigm; if so, that seems like it'll require implementers to maintain their authZ code in 2 different places using different approaches.

OK, so perhaps it can still work. The thought is that since this filtering solution gives you access, ultimately, to the GraphType's Metadata, then you can get to the authZ stuff you setup when defining your GraphType. Then, if you can have reusable code that can work both in the Schema Filter stuff that you're using over in your ValidationRule(s), you should be good-to-go.

sungam3r:
First of all I do not know what is AuthZ. In general, this filtering ability in itself has the right to life. I just wanted to draw attention to the fact that this is pure filtering which is in no way connected with a real ban on the execution of requests to individual parts of the schema. That is, it can rather be used to focus individual clients on those parts of the scheme that are useful to them. Naturally, you still have to write your own filter class, which will somehow differentiate clients.

GrangerAts:
@sungam3r - My apologies; I'm usually good about remembering to define my acronyms. :) "Auth" is ambiguous to me, so I use AuthZ for Authori**z**ation, and AuthC for Authenti**c**ation. 

I don't mind the separation of filtering from authorization, though I agree it would normally be confusing to filter out parts of the schema that the client is allowed to use.

joemcbride:
Good discussion here.  I want to update the filter interface to return `Task<bool>` since that would make it easier to work with the .NET Core Authorization policies.

This would be a potential implementation to filtering types:

https://gist.github.com/joemcbride/87a5761161748263115f591f4543b6e4

aoakeson:
I really like this.  I briefly implemented this as a proof of concept in a project, and it does seem that it will work. As we work with it more, we may come up with more use cases.  For now this is a great start.  Thanks @joemcbride and everyone involved in this.

joemcbride:
Updated the interface a bit to use Tasks.  I also thought that the `AllowXyz` helped with knowing if you should be returning `true` or `false`.

```csharp
public interface ISchemaFilter
{
  Task<bool> AllowType(IGraphType type);
  Task<bool> AllowField(IGraphType parent, IFieldType field);
  Task<bool> AllowArgument(IFieldType field, QueryArgument argument);
  Task<bool> AllowEnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue);
  Task<bool> AllowDirective(DirectiveGraphType directive);
}
```

AndyEdmonds:
I'm just implementing an example of ISchemaFilter using your gist at https://gist.github.com/joemcbride/87a5761161748263115f591f4543b6e4. In order to determine if a user has the right to see a particular part of the schema, I need to know the policies of the logged in user, So I need the GraphQLUserContext. How would this be obtained in a standard implementation?

GrangerAts:
It depends on where you establish your "User" object in your implementation. If you look at tutorials on the web, you'll notice that it's common for people to store the data you need in the `ExecutionOptions.UserContext`. I believe that's recommended because Gql.Net can be hosted in a bunch of different ways.

In my case, I'm using .Net Core's Authentication subsystem (an implementation of `AuthenticationHandler<T>`), so the identity data (authorization/authentication) gets stored in `HttpContext.User`. So there are all sorts of options to get that data to the spot that it's needed. (There's `Controller.User` in .Net Core's Mvc stuff, or you can directly use `IHttpContextAccessor`, or you can make your own "accessor" class to cut down on type-casting and stuff, etc.)

Jonatthu:
Is this already a feature that I can use?

joemcbride:
@Jonatthu This is currently available in the 3.0 preview versions, yes.

Jonatthu:
@joemcbride  once filtered how can I nullify a property from read incase they guessed the field name without throw a 401 https://graphql-dotnet.github.io/docs/getting-started/authorization

Is there a way to remove the field from the query?
Or should I just throw a context error like this is not a known field
```
      _.Match<Field>(fieldAst =>
      {
        var fieldDef = context.TypeInfo.GetFieldDef();
        if (fieldDef.RequiresPermissions() &&
            (!authenticated || !fieldDef.CanAccess(userContext.User.Claims)))
        {
          context.ReportError(new ValidationError(
              context.OriginalQuery,
              "auth-required",
              $"You are not authorized to run this query.",
              fieldAst));
        }
      });
```

joemcbride:
@Jonatthu yes you will have to have both.  The `ISchemaFilter` and some validation rule.  Instead of having a `"You are not authorized"` error, you could instead provide a `Cannot query field "{fieldName}" on type "{type}".` error, similar to the `FieldsOnCorrectType` rule.  Otherwise an attacker would know that is a valid field, they just don't have access.  You may or may not care about that.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Validation/Rules/FieldsOnCorrectType.cs#L26

dorlandode:
I am looking for an example of how to use a custom ISchemaFilter implementation.  Does one exist?  Not sure how to wire it up.

joemcbride:
@dorlandode There is a property named `Filter` on `ISchema`.  So you create the instance of your `ISchemaFilter` implementation and assign that property.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/ISchema.cs#L145

joemcbride:
Maybe a little confusing, but I'll mention that you would typically set this filter in the `ExecutionOptions` vs. on the `Schema` directly.

```
    var schema = new Schema { Query = new StarWarsQuery() };

    var json = await schema.ExecuteAsync(_ =>
    {
      _.Query = "{ hero { id name } }";
      _.SchemaFilter = new MySchemaFilter();
    });

    Console.WriteLine(json);
```

Shane32:
I had suggested we move that in #1589, by the way, but for 3.0 it is required that it be set in the `ExecutionOptions` or it will get overwritten by the `DocumentExecuter`.

Jonatthu:
@joemcbride  is there anyway to apply this filter with user context access?
I need to be able to filter the schema depending of the user role and the http context

GrangerAts:
> 
> 
> @joemcbride is there anyway to apply this filter with user context access?
> I need to be able to filter the schema depending of the user role and the http context

Consider my comment from Aug 19, 2019:
> 
> 
> It depends on where you establish your "User" object in your implementation. If you look at tutorials on the web, you'll notice that it's common for people to store the data you need in the `ExecutionOptions.UserContext`. I believe that's recommended because Gql.Net can be hosted in a bunch of different ways.
> 
> In my case, I'm using .Net Core's Authentication subsystem (an implementation of `AuthenticationHandler<T>`), so the identity data (authorization/authentication) gets stored in `HttpContext.User`. So there are all sorts of options to get that data to the spot that it's needed. (There's `Controller.User` in .Net Core's Mvc stuff, or you can directly use `IHttpContextAccessor`, or you can make your own "accessor" class to cut down on type-casting and stuff, etc.)

There are a number of options.

Jonatthu:
@GrangerAts 
It would be nice to have the option to implement something like:

RequiresAuthValidationRule<TPermissionsEnum> : IValidationRule

and then I can inject the HttpContext from my DI and decide when to show certain field or not

GrangerAts:
@Jonatthu 
That's what I do. The "user" object gets injected in a custom `IValidationRule`'s constructor, and it's used during `public INodeVisitor Validate(ValidationContext context)`.

joemcbride:
@Jonatthu An example of using ASP.NET Core Policies to check access:

```csharp
services.AddAuthorization(options =>
{
     options.AddPolicy("IsAuthenticated", policy =>
        policy.RequireAssertion(context =>
            context.User.IsAuthenticated));
});

{
    public MyGraphType()
    {
        this.Metadata["authPolicy"] = "IsAuthenticated"
    }
}

public class MySchemaFilter : ISchemaFilter
{
    private readonly IAuthorizationService _auth;
    private readonly IHttpContextAccessor _httpContextAccessor;

    public MySchemaFilter(IAuthorizationService auth, IHttpContextAccessor httpContextAccessor)
    {
        _auth = auth;
        _httpContextAccessor = httpContextAccessor;
    }

    public async Task<bool> AllowType(IGraphType type)
    {
        var policy = type.GetMetadata<string>("authPolicy");

         var authorizationResult = await _authorizationService
            .AuthorizeAsync(_httpContextAccessor.HttpContext.User, type, policy);

        return authorizationResult.Succeeded;
    }

    ...
}
```

Jonatthu:
@joemcbride  thanks Joe I will try it out !

dorlandode:
So does the Filter need to be set on the Schema or on the ExecutionOptions?  I am trying it on the Schema, and it is not firing.

Shane32:
Correct for 3.x. In 4.0 it will be set on the Schema.

dorlandode:
This is an example from a Startup.cs file.  This of course results in a warning.  My Filter is attempting to use DI to resolve dependencies.  This feels like an anti-pattern, but I am not sure how else to do what  I am trying to do...  Other than to maybe have the Schema Filter have its own container contained within it.  Thoughts?

```
services.AddGraphQL(o =>
                {
                    o.EnableMetrics = false;
                    o.ExposeExceptions = true;
                    o.SchemaFilter = services.BuildServiceProvider().GetRequiredService<ISchemaFilter>();
                    //o.ComplexityConfiguration.MaxDepth = 3
                })

**Comment by dorlandode** [comment]
I figured it out, I missed that that AddGraphQL() method had a signature with a IServiceProvider in it.  Changed to this and it works as expected now.

```
 services
                .AddGraphQL((o, serviceProvider) =>
                {
                    o.EnableMetrics = false;
                    o.ExposeExceptions = true;
                    o.SchemaFilter = serviceProvider.GetRequiredService<ISchemaFilter>();
                })

sungam3r:
@dorlandode This overload was added just for your case.

PR #1186, opened 2019-06-14
Refactor validation to allow for modification by consumers

AndyBan:
This is largely just an extract refactor to allow for the default validation behaviour to be replaced.  

Where I am using this project I need to return the errors in a different format and also perform null checks differently and this allows me to do that by having pluggable validation.

I have also changed the validation so that when it encounters an error rather than throwing an exception and stopping there it validates all the fields, it then checks if there are any errors before starting execution which allows for all errors to be returned rather than just the first.  

If expose exceptions is turned on then the existing functionality is retained.

### DIFF
--- a/src/GraphQL.Tests/Execution/VariablesTests.cs
+++ b/src/GraphQL.Tests/Execution/VariablesTests.cs
@@ -98,6 +98,8 @@ namespace GraphQL.Tests.Execution
         }
     }
 
+
+
     public class TestType : ObjectGraphType
     {
         public TestType()
@@ -299,8 +301,8 @@ namespace GraphQL.Tests.Execution
 
             var caughtError = result.Errors.Single();
             caughtError.ShouldNotBeNull();
-            caughtError?.InnerException.ShouldNotBeNull();
-            caughtError?.InnerException.Message.ShouldBe("Variable '$input.c' is invalid. Received a null input for a non-null field.");
+            caughtError?.ShouldNotBeNull();
+            caughtError?.Message.ShouldBe("Variable '$input.c' is invalid. Received a null input for a non-null field.");
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


joemcbride:
Can you provide an example of what custom validation you need?  This _should_ be validating based on the GraphQL specification.  Perhaps what you need can actually be accomplished in the validation rules, which are extensible, vs. trying to add additional validation at this level?

joemcbride:
Docs on validation rules:
https://graphql-dotnet.github.io/docs/getting-started/query-validation

Source of validation rules:
https://github.com/graphql-dotnet/graphql-dotnet/tree/master/src/GraphQL/Validation/Rules

AndyBan:
In short I need an alternative error message + some additional details when null validation fails for our webapp. We have a schema generator not unlike the conventions lib which mostly uses linqtodb attributes and our own validation service which already includes null checks.

I can remove the validations as shown above however that doesn't affect AST Validation.  The validation rules only check for null arguments not null properties on object arguments.  Validation will run through to:
https://github.com/graphql-dotnet/graphql-dotnet/blob/febf2b0637628436d8fe5b84d4f51afdfef7a58a/src/GraphQL/Execution/ExecutionHelper.cs#L132

Then promptly stop processing when it encounters the null.  

The pull request allows me to inject a error handler and in there I can customise the error returned, I look for the "Received a null input for a non-null field." message then do something else.  The string comparison isn't great but I don't think enumming the errors is the best solution here.  In my case I just ignore the error above and leave it to be validated later by our validation service.  

My changes also allow for multiple errors to be returned (if you have multiple non-null nulls for example).

The argument could be made that the validation rules should be checking for null properties not the AST validation, I could change my pull request to do that perhaps if you'd rather do it that way although we'd loose the ability to get multiple errors back and customise the errors.

joemcbride:
Unfortunately this is just not a feature that I think I want to try to support (adding the ability to customize validation at that point in the process).  I've double checked and as far as I can tell we are validating based on the GraphQL spec.  It is indeed the case that we're not validating the data provided in the variable, only that the variable is being used correctly within the request.  I've checked the `graphql-js` reference implementation project and it does not appear they are doing that check either.  This should be _relatively_ easy to write a validation rule for though, which would mean you would never encounter the `Received a null input for a non-null field.` error, and you could provide your own error message in that case.

This exact scenario is what the Validation Rules are for.  That code would never even get reached if the validation rule fails.  I would be open to accepting a new validation rule that validated the structure of variables.

sungam3r:
@AndyBan Perhaps this Issue can be solved using a separate validation rule?

AndyBan:
IIRC it validates the null whilst building up the AST which is before any custom rules run, that and the fact it throws an exception to stop further processing didn't work for us as we'd like all errors returned not just the first it comes across.

PR #1196, opened 2019-06-25
Correct the Example to avoid Compiler Error CS0407

bgradyCambridgeFx:


SimonCropp:
@bgradyCambridgeFx thanks for this

PR #1223, opened 2019-07-25
Added else-case for generic Task

irgsmirx:
Method ExecuteAsync on DocumentExecuter failed with Result Property not found on Task.
The following piece of code caused the failure: Task<ILookup<String, Person>> GetPersonsByChefIds(IEnumerable<String> chefIds);

An "is Task<object>" check does not cover all and especially this case.

The proposed change mitigates this error.

### DIFF
--- a/src/GraphQL/TaskExtensions.cs
+++ b/src/GraphQL/TaskExtensions.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading.Tasks;
 
 namespace GraphQL
@@ -22,8 +23,11 @@ namespace GraphQL
             {
                 // Most performant if available
                 return to.Result;
-            }
-            else
+            } else if (task.GetType().IsGenericType)
+            {
+                Type t = task.GetType();
+                return t.GetProperty("Result").GetValue(task, null);
+            }  else
             {
                 // Using dynamic is over 10x faster than reflection
                 return ((dynamic)task).Result;


johnrutherford:
@irgsmirx 

The existing else branch already handles getting the result from a `Task<T>`. And it uses dynamic, which is much faster than using reflection. 

If the error you are getting says "Result Property not found on Task". Then it sounds like somehow a `Task`, not `Task<T>` is getting sent into this method. That shouldn't be happening so you should look into what is causing that.

irgsmirx:
I can assure you that it is not the case that a non-generic Task gets sent into this method.

It is a Task<IEnumerable<SomeClass>>. I also tried an ILookup<String,SomeClass>>.

sungam3r:
@irgsmirx Then perhaps the best solution for you will be to provide a minimal test to reproduce the error, to make sure that this is really the case.

irgsmirx:
Yes, I can do that, but I think the problem just is that Interfaces do not derive from object an thus your if block does not get triggered.

sungam3r:
@irgsmirx I think a error is in something else.

irgsmirx:
Hi, 

I have uploaded an example:

https://github.com/irgsmirx/GraphQL.NETTest

Please check it out and try to reproduce the error I experienced.

sungam3r:
Wow! Well, this example made me scared! I already began to think that 2 + 2 = 5. But then everything fell into place. The case turned out to be specific dynamic behavior. It does not allow you to access a type that you would not already access using static compilation.


Just try this:
```c#
using GraphQL;
using System.Linq;
using System.Threading.Tasks;

namespace GraphQLTest
{
    class Program
    {
        {
            var res = Task.FromResult<ILookup<string, Chef>>(null);
            var a = res.GetResult(); // <----- OOOOOPS RuntimeBinderException
        }
    }

    class Chef { }
}
```

Then try that:
```c#
using GraphQL;
using System.Linq;
using System.Threading.Tasks;

namespace GraphQLTest
{
    class Program
    {
        {
            var res = Task.FromResult<ILookup<string, Chef>>(null);
            var a = res.GetResult(); // <----- OK
        }
    }

    public class Chef { } // <------ now it is public
}
```
or even that
```c#
using GraphQL;
using System.Linq;
using System.Threading.Tasks;

[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("GraphQL")]

namespace GraphQLTest
{
    class Program
    {
        {
            var res = Task.FromResult<ILookup<string, Chef>>(null);
            var a = res.GetResult(); // OK
        }
    }

    class Chef { } // <---- now it is internal again (by default)
}
```

Here we can only grumble at the MS for such a misleading error message. After spending an hour of solving this puzzle, it becomes more or less clear what they wanted to say with this message. But they could also make a normal, more friendly message by hinting about type access qualifiers.

I have updated #1224 to fix this error without losing performance for most cases.

sungam3r:
@joemcbride @johnrutherford I did not expect to ever encounter such problems. Does this fix look normal?

johnrutherford:
@sungam3r Yes, I remember now that this issue came up another time.

sungam3r:
@irgsmirx  #1224 merged in master, you can take the latest preview package (1206) from https://myget.org/F/graphql-dotnet/api/v3/index.json

sungam3r:
Closed in favor to #1224

PR #1233, opened 2019-08-09
Initial work to support federated schemas

### LINKED ISSUE #1169: Support Apollo Federation
## Summary

Apollo released a new feature called federations. It would be awesome if graphql-dotnet supported this out of the box. 

### Basic example

https://blog.apollographql.com/apollo-federation-f260cf525d21

### Specification

https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#federation-schema-specification

### Motivation

This feature would be essentials for services running on a distributed architecture (e.g. Service Fabric). This allows us to comine multiple seperate graph services into a single graph.

joemcbride:
Implements the Apollo Federation specification.

https://www.apollographql.com/docs/apollo-server/federation/federation-spec/

# Syntax

```csharp
var schema = FederatedSchema.For(@"
    extend type Query {
        me: User
    }

    type User @key(fields: ""id"") {
        id: ID!
        username: String!
    }
",
_ =>
{
    _.Types.Include<Query>();
    _.Types.For("User").ResolveReferenceAsync(ctx =>
    {
        var id = (string)ctx.Arguments["id"];
        var user = Query.users.FirstOrDefault(x => x.Id == id);
        return Task.FromResult(user);
    });
});

public class User
{
    public string Id { get; set; }
    public string Username { get; set; }
}

public class Query
{
    public static User[] users = new [] {
        new User { Id = "1", Name = "Ada Lovelace", Username = "@ada" },
        new User { Id = "2", Name = "Alan Turing", Username = "@complete" }
    };

    public User Me()
    {
        return users[0];
    }
}
```

# Other Updates

* Ensures all printed types using `SchemaPrinter` checks the `SchemaPrinterOptions.IncludeDescriptions` for printing descriptions.  Some printers were always printing the description regardless of that setting.  `SchemaPrinterOptions.IncludeDescriptions` defaults to `false`.

closes #1169


sungam3r:
Interesting. I want to check this in more detail at the weekend.

andrewtyped:
Really appreciate the work you're doing on this one. If this were merged it would be huge for our team. We maintain a microservice GraphQL gateway that merges all the sub-schemas together with NuGet packages. Dependency management is painful.

shani117:
Hi,

I'm trying to implement a API that serves up this federated schema that I can then embed into a gateway service, but I'm running into a strange problem. This is how I have defined the schema:


[ChatQuery.zip](https://github.com/graphql-dotnet/graphql-dotnet/files/3971875/ChatQuery.zip)

`public class UserFedSchema : Schema
    {
        public UserFedSchema()
        {
            var schema = FederatedSchema.For(@"
                extend type Query {
                    me: User
                }

                type User @key(fields: ""id"") {
                    id: String!
                    username: String!
                } 
            ",
            _ => 
            {
                _.Types.Include<UsrQuery.Query>();
                _.Types.For("User").ResolveReferenceAsync(ctx =>
                {
                    var id = (string)ctx.Arguments["id"];
                    var user = UsrQuery.Query.users.FirstOrDefault(x => x.Id == id);
                    return Task.FromResult(user);
                });                
            });            
            Query = schema.Query;
            
        }
    }`

And in my container service collection, I'm registering the schema as such:
`public void ConfigureServices(IServiceCollection services)
        {            
            services.Configure<IISServerOptions>(options =>
            {
                options.AllowSynchronousIO = true;
            });
            services                
                .AddSingleton<User>()
                .AddSingleton<Query>()
                .AddSingleton<UserFedSchema>()
                .AddGraphQL(options =>
                {
                    options.EnableMetrics = Environment.IsDevelopment();
                    options.ExposeExceptions = Environment.IsDevelopment();
                    //options.UnhandledExceptionDelegate = ctx =>
                    //{
                    //    Console.WriteLine("error: " + ctx.OriginalException.Message);
                    //};
                })
                .AddGraphTypes(Assembly.GetAssembly(typeof(UserFedSchema)));                
        }
public void Configure(IApplicationBuilder app)
        {
            if (Environment.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }
            app.UseDefaultFiles();
            app.UseStaticFiles();
            app.UseGraphQL<UserFedSchema>();
        }`

When I run this service, I get an error saying:

> "GraphQL.ExecutionError: Unable to resolve reference to type 'User' on 'Query'"

And the strange thing is that if I try to debug the code and spend a few mins in that process, everything works fine and the reference resolution works perfectly. There is something going on with the reference resolution process that is causing this which I'm not able to figure out. Any pointers?

Apologies, if the code snippets don't come through nicely, I've also attached the .cs files in the zip folder.

Thanks,
Shahnaz.

joemcbride:
@shani117 You are double-defining your Schema.  Your code can be simplified to:

```csharp
services.AddSingleton<ISchema>(c => BuildSchema())

public ISchema BuildSchema()
{
    return FederatedSchema.For(@"
        extend type Query {
            me: User
        }

        type User @key(fields: ""id"") {
            id: String!
            username: String!
        } 
    ",
    _ => 
    {
        _.Types.Include<UsrQuery.Query>();
        _.Types.For("User").ResolveReferenceAsync(ctx =>
        {
            var id = (string)ctx.Arguments["id"];
            var user = UsrQuery.Query.users.FirstOrDefault(x => x.Id == id);
            return Task.FromResult(user);
        });
    });
}
```

You will also want to add the following types for federation:

```csharp
// Apollo Federation Types
services.AddSingleton<AnyScalarGraphType>();
services.AddSingleton<ServiceGraphType>();
```

Here is a full example from the Apollo Federation docs, with 3 node services and 1 in .NET:

https://github.com/joemcbride/gql-presentation/tree/master/demos/6%20-%20federation
https://github.com/joemcbride/gql-presentation/blob/master/demos/6%20-%20federation/services/accounts/Startup.cs#L40

shani117:
Awesome. Thanks for the response Joe. Yes, that did it. And thanks a bunch for the other examples project, didn't find that somehow. Much appreciated.

ahammouda:
Is there a way to do this GraphType first rather than Schema first?

If so how would you indicate that a field is a @key in the federated environment?  How would you indicate that you're 'extending' the Query and Mutation type?

joemcbride:
@ahammouda There is not a way to do this at present with GraphType first.  Some of the core types, such as `Any` and `_Entity` could be easily added to the Schema, but there would be work that would need to be done to support the directives such as `@key` and add the Entity specific resolvers.

MatsDahlin:
@joemcbride I've inherited a project that uses GraphType first and need to implement Apollo Federation. Just wondering if there is any plans to add this support in the near future?

ahammouda:
@joemcbride - thanks for the quick response.  At the expense of indicating my inexperience with graphql, C#, and your library - does that mean my federated schema won't be able to implement custom resolvers on federated fields?

e.g:
`                return FederatedSchema.For(@"
                    extend type Query {
                    }

                    type Customer @key(fields: ""id"" {
                          id: ID!
                          quantity: Int
                    }

                    type Order @key(fields: ""id"") {
                        id: ID!
                        customerId: ID
                        customer: Customer
                    }
                ", _ =>

_.Types.For("Order") ... // (resolve reference)
_.Types.For("Customer")...  // (resolve reference)
//Something like this doesn't seem particularly accessible through your existing API 
_.Types.For("Order.customer").ResolveReferenceAsync(ctxt => dataStore.GetCustomerById(ctxt.Arguments["customerId"])`

If I can come to an understanding of your library fast enough, I'll post syntax for a working example - otherwise, any help from the community is highly appreciated.

joemcbride:
@ahammouda You would not do `_.Types.For("Order.customer")`, because that is a field on the `Order` type, not a Graph Type reference that gets resolved through Apollo Federation (the `ResolveReferenceAsync` resolver gets called when trying to fetch and entity through the `_entities` field on `Query` through the `_Entity` type - see the Apollo Federation Spec - https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#federation-schema-specification).  Typically you would have the `Customer` and `Order` types in different independent services, which can be combined together with Apollo Federation.

I suggest to read this overview of Apollo Federation.

https://www.apollographql.com/docs/apollo-server/federation/introduction/

The Apollo Gateway is the magic that glues all the pieces together.

This example shows the example in the Apollo docs with 3 JavaScript GraphQL services and 1 .NET GraphQL service, combined together in a Apollo Gateway:

https://github.com/joemcbride/gql-presentation/tree/master/demos/6%20-%20federation

If you wanted the `Order` and `Customer` types to be in the same service, I would instead just use the normal Schema First syntax, as show here: https://graphql-dotnet.github.io/docs/getting-started/introduction#schema-first-nested-types.  I have not tried this yet but presumably it should work.  Let me know if you do and run into problems.

ahammouda:
@joemcbride - I appreciate your patience with us beginners.  That indeed does the trick - I'll replace this comment shortly with an annotated working snippet in case others come looking.

ahammouda:
Is anyone aware of the syntax for referencing external types from within you dotnet graphql schema (js syntax: https://www.apollographql.com/docs/apollo-server/federation/core-concepts/#referencing-external-types).

It seems like you could simply follow the docs for the schema definition piece, but how would you construct the object representation of the external type to be resolved as is done in the 2nd code snippet of the linked example?

ahammouda:
I think I answered my own question.  I will place my snippets here shortly.  In short this could be handled the same way that you're handling 'extend Query' in your snippet above.

It makes me wonder why apollo's implementation requires the extra object representation on your service resolver (syntax: {__typename: 'ExternalType', fieldKey: localobject.Key }).  Perhaps they're leveraging it for some extra optimizations in their query planner?

ahammouda:
Here's an example of referencing an external type for those looking.

The setup
Our external service has a type called `Bar`.  Here's what it's C# model and and schema look like:
```
// Models.cs
public class Bar {
   public int id { get; set; }
   public string description { get; set; }
}
// Startup.cs in ConfigureServices(IServiceCollection services) 
...
     services.AddTransient<ISchema>(s =>
            {
                var store = s.GetRequiredService<BarStore>();

                return FederatedSchema.For(@"
                                type Bar @key(fields: fields: ""id"") {
                                       id: Int
                                       description: String
                                }", _ =>
                {
                    _.Types.For("Bar").ResolveReferenceAsync(context =>
                    {
                        var id = (int)context.Arguments["id"];
                        return store.GetBarById(id);
                    });
                }
```

Tall-Paul:
> Here's an example of referencing an external type for those looking.
> 
> The setup
> Our external service has a type called `Bar`. Here's what it's C# model and and schema look like:
> 
> ........
> If people find this extra bit of documentation useful, I'll allocate some time and submit a pull request. Hope this helps.

I don't suppose you have a full example, with web request / response etc?

I'm trying to add federation to the Asp.Net middleware server, but not making much progress...if it's already possible to do federation without the middleware I'd love to see a working example.

hwillson:
This is really awesome! We (Apollo) have put together a [Federation compatibility test suite](https://github.com/apollographql/apollo-federation-subgraph-compatibility), and would love to work with someone here to make sure `graphql-dotnet`'s Federation work is highlighted (in our testing repo and in our docs). If anyone is interested in collaborating on this, definitely let me know. Thanks!

sungam3r:
@hwillson I think I can work on it. Please post a new issue to discuss.

Issue #1239, opened 2019-08-13
Restrict the maximum number of parallel tasks in ParallelExecutionStrategy


teo78:
## Summary

We're using the GraphQL dotnet library (version 2.4) in a ASP .NET Core application using MongoDB as the database. We are experiencing extreme performance drops occasionally when running large queries involving a lot of database queries. The issue was found to be due to the MongoDB driver exhausting the connection pool (Default = 100), in which case db queries started to queue up and run extremely slow. As a band-aid solution increasing the connection pool on the MongoDB driver solved the issue.

However, this raises the question of whether the library should have a limit to how many parallel nodes it will execute simultaneously.

### Test using PR: 'Optimize ParallelExecutionStrategy: Mark2'
I did test the same problematic queries using the modifications in the 'Optimize ParallelExecutionStrategy: Mark2' pull request #976. Execution was often even slower, which is not unexpected since it runs even more nodes in parallel.

By modifying the code in this PR to include a maximum number of parallel tasks, the query runs fast. I.e. with the default max connection pool of 100, we do not see the performance drops and do not have to increase the max pool size to alleviate the problem.

The fix tested was by changing the 'pendingNodes' from a List to a Queue and limiting the number of current tasks instead of adding all unrestricted:

```csharp
// Start executing all pending nodes, limiting the max number of current tasks
while (currentTasks.Count() < _taskLimit && pendingNodes.TryDequeue(out var pendingNode))
{
    var task = ExecuteNodeAsync(context, pendingNode);
    currentTasks.AddLast(task);
}
```

### Motivation

Limit the number of maximum parallel nodes executed, keeping it unrestricted can cause issues be starting to many parallel database calls, threads, etc.

joemcbride:
@teo78 Thanks for adding this issue!  Would you mind providing an initial PR with this support?

teo78:
I'll have a go at it. Regarding setting of the maximum parallel limit, my thinking is this:

1. Add a 'MaxParallelExecutionLimit' (Or Similar) int property to the 'ExecutionOptions' class.
    - Default value: int.MaxValue (to keep it unrestricted).
2. In 'ExecuteAsync' method of the 'DocumentExecuter' class, pass this setting to the 'BuildeExecutionContext'
   - This means also adding 'MaxParallelExecutionLimit' property to the 'ExectionContext' class
[...]

PR #1251, opened 2019-08-17
Feature/limitparallelexecution

teo78:
Add 'MaxParallelExecutionCount' setting to ExecutionOptions, which when set is used by ParallelExecutionStrategy to limit the maximum number of parallel executed nodes.


src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -39,6 +39,11 @@ public class ExecutionOptions
 
         public bool ThrowOnUnhandledException { get; set; } = false;
 
+        /// <summary>
+        /// If set, limits the maximum number of nodes executed in parallel
+        /// </summary>
+        public int? MaxParallelExecutionCount { get; set; } = null;
```
sungam3r: =null  is excessively

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -154,7 +154,8 @@ public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
                     options.CancellationToken,
                     metrics,
                     options.Listeners,
-                    options.ThrowOnUnhandledException);
+                    options.ThrowOnUnhandledException,
```
sungam3r: I always wondered if `BuildExecutionContext` method was appropriate. It just pumps a bunch of parameters and makes a new call. Maybe inline it?

src/GraphQL/Execution/ParallelExecutionStrategy.cs
```diff
@@ -11,37 +11,36 @@ protected override Task ExecuteNodeTreeAsync(ExecutionContext context, ObjectExe
...
 
+            var currentTasks = new List<Task<ExecutionNode>>();
             while (pendingNodes.Count > 0)
             {
-                var currentTasks = new Task<ExecutionNode>[pendingNodes.Count];
-
-                // Start executing all pending nodes
-                for (int i = 0; i < pendingNodes.Count; i++)
+                // Start executing pending nodes, while limiting the maximum number of parallel executed nodes to the set limit
+                while ((context.MaxParallelExecutionCount == null || currentTasks.Count < context.MaxParallelExecutionCount)
```
sungam3r: it can now be argued that the behavior of `SerialExecutionStrategy` is equivalent to that of `ParallelExecutionStrategy` with `MaxParallelExecutionCount = 1`

sungam3r:
LGTM

sungam3r:
Fixes #1239

src/GraphQL/Execution/ParallelExecutionStrategy.cs
```diff
@@ -11,37 +11,36 @@ protected override Task ExecuteNodeTreeAsync(ExecutionContext context, ObjectExe
...
+                    && pendingNodes.Count > 0)
                 {
                     context.CancellationToken.ThrowIfCancellationRequested();
-                    currentTasks[i] = ExecuteNodeAsync(context, pendingNodes[i]);
+                    var pendingNode = pendingNodes.Dequeue();
+                    currentTasks.Add(ExecuteNodeAsync(context, pendingNode));
```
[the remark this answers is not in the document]
sungam3r: Good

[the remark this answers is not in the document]
teo78: Do you agree I should add this change to the PR?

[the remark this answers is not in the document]
sungam3r: Yes. This change makes sense. It should only get better.

teo78:
Any chance for this PR to get merged? If there are any issues, let me know and I'll try to adapt it if necessary :)

sungam3r:
> Not sure if this breaks anything regarding the OnBeforeExecutionStepAwaitedAsync, but since execution for the field is already completed it means it does not wait on any dataloader operations and this change is safe?

I just got confused by this comment and decided to wait. Otherwise everything looks good. I'll look again the other day and merge.

teo78:
Great. Yeah I wanted clarification if there were any issues with not adding already completed nodes to the list of tasks being awaited (after the OnBeforeExecutionStepAwaitedAsync). The completed field is still awaited (before the OnBeforeExecutionStepAwaitedAsync), but since it is already completed the await will finish synchronously, and if no errors just get the result. Otherwise it will propagate any exceptions thrown. So my thought was that it would be ok, but looking forward to a review :)

src/GraphQL/Execution/ParallelExecutionStrategy.cs
```diff
@@ -11,37 +11,55 @@ protected override Task ExecuteNodeTreeAsync(ExecutionContext context, ObjectExe
...
+                        {
+                            foreach (var childNode in parentExecutionNode.GetChildNodes())
+                            {
+                                pendingNodes.Enqueue(childNode);
+                            }
+                        }
+                    }
+                    else
+                    {
+                        // Node is actually asynchronous, so add it the list of current tasks being executed in parallel
```
sungam3r: the list
->
to the list

sungam3r:
@teo78 Some time ago I added API Approval tests in solution to verify public api changes. Now any change to the public api requires modification of the .approved file. This also happened with the `MaxParallelExecutionCount`. After successfully completing the tests, I will merge your PR. Well, I still doubt regarding `BeforeExecutionStepAwaitedAsync`, but let's hope this is in vain. I have not yet understood Dataloader intently.

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -52,6 +52,11 @@ public class ExecutionOptions
         /// </summary>
         public Func<Execution.ExecutionContext, Exception, Exception> UnhandledExceptionDelegate { get; set; } = (ctx, ex) => ex;
 
+        /// <summary>
+        /// If set, limits the maximum number of nodes executed in parallel
+        /// </summary>
+        public int? MaxParallelExecutionCount { get; set; } = null;
```
sungam3r: =null is excessively

teo78: Removed =null.

src/GraphQL/Execution/ParallelExecutionStrategy.cs
teo78: Fixed typo.

teo78:
@sungam3r Just to let you know that I applied and pushed the requested changes over the weekend.

sungam3r:
Yes, I saw and was waiting for the tests to be fixed.

> @teo78 Some time ago I added API Approval tests in solution to verify public api changes. Now any change to the public api requires modification of the .approved file. This also happened with the MaxParallelExecutionCount. After successfully completing the tests, I will merge your PR. Well, I still doubt regarding BeforeExecutionStepAwaitedAsync, but let's hope this is in vain. I have not yet understood Dataloader intently.

sungam3r:
@teo78 I will be glad to merge this after you resolve conflicts.

pekkah:
Shouldn't this be separate strategy class?

sungam3r:
I think not. The new setting fits well with the current class.

teo78:
@sungam3r I just pushed changes that resolve the conflicts. However I'm not able to run the tests on my machine anymore. And it fails on the API tests here, something that needs to be updated since I've now added the MaxParallelExecutionCount to the API?

sungam3r:
Just modify txt file with approved api. The purpose of this is to explicitly see changes in public api.

sungam3r:
@teo78 Please remove new line at the end of approval file. It’s better to do it through a regular text editor rather than VS since VS inserts newline. I could not do it through the web interface. Also you can tune this via .editorconfig file - just add rule for txt files:
```
[*.txt]
insert_final_newline = false
```

joemcbride:
@sungam3r I added that a while back. Not sure why it wouldn’t be kicking in.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/.editorconfig#L29

joemcbride:
Could be that this just would need a rebase to pick that up.

sungam3r:
Nope :) Just another "root" file - https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/.editorconfig vs https://github.com/graphql-dotnet/graphql-dotnet/blob/master/.editorconfig

I think first should be deleted.

joemcbride:
That would do it.

> I think first should be deleted.

Agreed.

WellspringCS:
Apologies if the answer is obvious... but I have this error...

```
"GraphQL.ExecutionError: Error trying to resolve roles.\n ---> 
System.InvalidOperationException: A second operation started on this context
before a previous operation completed. This is usually caused by different threads
using the same instance of DbContext.
```
The code running (hoping the below is all you need to see...)

that produces this error is:
```
			
FieldAsync<ListGraphType<UserOrganizationType>>(
  "organizations",
  resolve: async context =>
  {
    var loader = dataLoaderContextAccessor.Context.GetOrAddCollectionBatchLoader<string, VwUserOrganization>(
      "organizationLoader", organizationRepository.GetForUsers
    );
    return await loader.LoadAsync(context.Source.UserId);
  }
);

FieldAsync<ListGraphType<VwUserRoleAllType>>(
  "roles",
  resolve: async context =>
  {
    var loader = dataLoaderContextAccessor.Context.GetOrAddCollectionBatchLoader<string, VwUserRoleAll>(
      "rolesLoader", userRoleRepository.GetForUsers
    );
    return await loader.LoadAsync(context.Source.UserId);
  }
);
```

And from my late-night scouring, I am wondering if setting MaxParallelExecutionCount to 1 will resolve my problem.

A) Am I (obviously) mistaken? Or am I barking up the right tree?
B) If I'm on the right track, how do I set MaxParallelExecutionCount in ASP.NET Core 3 with GraphQL.Net 3.0.0-preview-1352? I've looked everywhere but cannot see examples.

sungam3r:
You're right. See the long discussion in #1310 . `ExecutionOptions.MaxParallelExecutionCount` is what you want.

WellspringCS:
Thanks for the feedback, @sungam3r but two things:
1) I didn't notice your reply and updated my question (see top section)... and
2) I have not yet figured out where to set ExecutionOptions.MaxParallelExecutionCount

Perhaps I'll have better luck with the search on #2 now that I am working with 3.0.0-preview-1490 ?

sungam3r:
>  Is that a fluke?

It is a race conditions. `DbContext` is designed not to be thread safe. Sometimes multiple resolvers operate over it in parallel.

WellspringCS:
> It is a race conditions
Thanks so much for confirming - that's what I was concerned about.

Some burning questions and thanks so much in advance for your thoughts...

1) I read the link you shared again (turns out I had read before) and I think I see how it all works. Are my comments below accurate?
```
public class EfDocumentExecuter : DocumentExecuter
  {
    {
      if (context.Operation.OperationType == OperationType.Query)
      {
        // context.MaxParallelExecutionCount = 1; // this works, but the below is a more std way to accomplish the same thing?
        return new SerialExecutionStrategy();
      }
      return base.SelectExecutionStrategy(context);
    }
}
```
2) If *all* my data retrieval is EF *and* 95% of my GraphQL queries are of this flavor...
```
{
  root {
    datapoint
    datapoint
    nested1 {
      datapoint
      datapoint
      datapoint
      nested2 {
          datapoint
      }
  }
}
```
...then wouldn't those queries be executed in series *anyway*? At which point for the few cases where I have two nested nodes at the same level (producing the race condition, right?) that's where my forcing execution to series will actually produce a (desired, for me) behavior change?

3) I saw someone's comment about forcing series (over parallel) mucking up the N+1 and I don't understand why that would be so. Were they confused?

Thanks again

sungam3r:
1. ParallelExecutionStrategy with MaxParallelExecutionCount = 1 is effective the same as SerialExecutionStrategy.

2. **All fields at any level** will be executed sequentially - https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Execution/SerialExecutionStrategy.cs

3. Because of DataLoader internals.

WellspringCS:
> All fields at any level will be executed sequentially


I think you misunderstood my question. What I meant was...

I suspect nothing here *can* be run in parallel:
```
{
  root {
    datapoint
    datapoint
    nested1 {
      datapoint
      datapoint
      datapoint
      nested2 {
          datapoint
      }
  }
}
```
but here that is possible, and here is where race conditions rear their ugly heads...
```
{
  root {
    datapoint
    datapoint
    nested1 {
        datapoint
    }
    nested2 {
        datapoint
    }
  }
}

Is the above correct? And if most of my queries are of the first sort, I won't see much performance degradation, except in those cases of the latter sort.

Or am I not envisioning the logic of data access correctly, here?

WellspringCS:
Doh! I now see in action, in my code, the N+1 problem with DataLoader and execution in series. Not cool.

So in summary:

I agree, @sungam3r with what you've said (here/elsewhere) that making GraphQL run everything in series is too big a hammer (*especially* since this screws up DataLoader).

I also trust those opinions that shy away from making dbContext transient.

The solution I am going to try is to split out a separate DbContext for slices of the graph and (hopefully!) capture the benefits of parallel for all cases, with the small cost of splitting up a few Contexts. Since this issue (for me) applies only to queries, and not mutations, my hunch is that there is little to no loss here, for me. I'm a bit of a noob where UnitOfWork is concerned, but I can't picture it being relevant as long as your queries run in milliseconds and are just queries.

WellspringCS:
Doh! I now see in action, in my code, the N+1 problem with DataLoader and execution in series. Not cool.

This problem is a real bummer.

Does anyone know if Dapper faces the same challenges?

I'm probably going to give that route a try, just to see.

PR #1256, opened 2019-08-20
fix parsingStat incorrectly used for validationStat

SimonCropp:


SimonCropp:
@joemcbride pretty simple bug fix and a test to verify

PR #1257, opened 2019-08-20
throw for empty graph name

SimonCropp:


sungam3r:
I expected that sooner or later this code would appear. The problem is that this kind of verification is not enough. It will only work if the property is explicitly set (with null or empty string). Often, they simply forget to install this property (do not touch it).

Well, this kind of check will have to be done everywhere, not just for the name. I would prefer to have something like validation, but not for a client request, but a server side schema. It can be run on demand like automapper's `AssertConfigurationIsValid()` .

SimonCropp:
good point :(

PR #1261, opened 2019-08-20
Fix DataLoader deadlock

### LINKED ISSUE #945: DataLoader deadlock with multiple awaits
## Description

When a field resolver awaits a data loader after another await (on anything) the resolver id deadlocked.

### Steps to reproduce

Query the following field:
```c#
         Field<BooleanGraphType, bool>()
                .Name("True")
                .ResolveAsync(async ctx =>
                {
                    await Task.Delay(1);

                    var loader = accessor.Context.GetOrAddLoader("GetTrue",
                        () => Task.FromResult(true));

                    return await loader.LoadAsync();
                });
```

### Expected result

No deadlock

### Actual result

Deadlock

teo78:
This PR is an attempt to fix #945. It adds a task `DispatchNeeded` on the `DataLoaderBase`, this task is completed when the dataloader is waiting for a dispatch. This property is propagated up thru the `DataLoaderContext` and `DataLoaderDocumentListener`, and used in `ParallelExecutionStrategy` to detect that a dataloader needs a dispatch to avoid the deadlock.

The main change in `ParallelExecutionStrategy` is this:
```cs
                // Create a task that completes when all currentTasks are complete
                var commonTask = Task.WhenAll(currentTasks);
                
                // Await tasks for this execution step, use loop to keep dispatching dataloaders
                // in case resolvers uses multiple dataloaders, or the dataloader is not the first async method awaited.
                while (!commonTask.IsCompleted)
                {
                    // Dispatches any waiting dataloaders
                    await OnBeforeExecutionStepAwaitedAsync(context)
                        .ConfigureAwait(false);

                    // Wait for either all currentTask to complete, or any dataloader(s) in need of dispatching
                    // which will continue the loop and dispatch the waiting dataloader(s)
                    await Task.WhenAny(commonTask, DataLoaderDispatchNeeded(context))
                        .ConfigureAwait(false);
                }
                var completedNodes = await commonTask
                    .ConfigureAwait(false);
```

In the scenario that fields do not use dataloaders or the dataloaders are the first async/await in the fields the execution is the same (i.e. the while loop above is only run once/first time).
However if any field use multiple dataloaders or awaits a dataloader after another await this will be detected and cause the loop to continue and dispatch the dataloader(s).

I also added a test for #945, which deadlocked before and now passes.

A potential issue I see is if the dataloaders are awaited after another await, the operations might not be batched as good since the dispatch is triggered when the first field reaches the 'LoadAsync' on a dataloader. However since we don't know how many fields use the dataloader it is impossible to know how long to wait. But atleast the deadlock is avoided.

### DIFF
--- a/src/GraphQL.DataLoader.Tests/DataLoaderDeadlockTest.cs
+++ b/src/GraphQL.DataLoader.Tests/DataLoaderDeadlockTest.cs
@@ -1,0 +1,50 @@
+using GraphQL.Types;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace GraphQL.DataLoader.Tests
+{
+    public class DataLoaderDeadlockTest : QueryTestBase
+    {
+        protected override void ConfigureServices(ServiceCollection services)
+        {
+            base.ConfigureServices(services);
+            services.AddSingleton<DataLoaderSchema>();
+        }
+        [Fact]
+        public void Await_before_LoadAsync_should_not_deadlock()
+        {
+            // Tests Github issue: DataLoader deadlock with multiple awaits #945
+            AssertQuerySuccess<DataLoaderSchema>(
+                query: "{ true }",
+                expected: @"
+                { true: true }
+                ");
+        }
+    }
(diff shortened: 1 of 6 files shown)


Shane32:
Deadlock issue fixed via #1511

PR #1274, opened 2019-08-30
Allocation optimizations for validation stage

sungam3r:
Since validation rules do not contain state, they can be reused for all requests. The `ValidationResult` without errors can also be reused.


src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -19,6 +19,35 @@ public interface IDocumentValidator
 
     public class DocumentValidator : IDocumentValidator
     {
+        public static readonly IEnumerable<IValidationRule> CoreRules = new List<IValidationRule>
```
sungam3r: So now it is not method and will require small change in server project.

PR #1247, opened 2019-08-15
Adds API Approval tests

sungam3r:
Adds a mechanism for controlled change of public api:
 
https://github.com/JakeGinnivan/ApiApprover

PR #1279, opened 2019-09-01
fieldnameconverter no longer allocates memory

sungam3r:
Converters are stateless and can be reused.

### DIFF
--- a/src/GraphQL/Conversion/PascalCaseFieldNameConverter.cs
+++ b/src/GraphQL/Conversion/PascalCaseFieldNameConverter.cs
@@ -1,0 +1,17 @@
+using GraphQL.Introspection;
+using System;
+using System.Linq;
+
+namespace GraphQL.Conversion
+{
+    public class PascalCaseFieldNameConverter : IFieldNameConverter
+    {
+        private static readonly Type[] IntrospectionTypes = { typeof(SchemaIntrospection) };
+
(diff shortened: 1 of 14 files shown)


src/GraphQL/Conversion/PascalCaseFieldNameConverter.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+using System.Linq;
+
+namespace GraphQL.Conversion
+{
+    public class PascalCaseFieldNameConverter : IFieldNameConverter
+    {
+        private static readonly Type[] IntrospectionTypes = { typeof(SchemaIntrospection) };
+
+        public static readonly PascalCaseFieldNameConverter Instance = new PascalCaseFieldNameConverter();
```
SimonCropp: should `instance` be public?

sungam3r: Well this is a common approach - make such static readonly instance. Otherwise, the meaning of reuse by the calling code is lost.

SimonCropp: yeah, just my default approach when any api is added to the public surface area.

sungam3r: Ok. Note that for the verification of the public api changes we have #1247 .

Issue #1153, opened 2019-05-23
How to load many items with DataLoader?


acelot:
I have the following scheme:
```gql
type Query {
  articles: [Article!]!
}

type Article {
  id: ID!,
  title: String!,
  authors: [Author!]!
}

type Author {
  id: ID!,
  name: String!
}
```

Typical N+1 problem when you need to load authors for each article. Authors are stored in `articles` table as an array of integers in `authors` column in PostgreSQL.

Articles table example:

|id|title|authors|
|--|----|---------|
|`1`|Article 1|`[1,2,3]`|
|`2`|Article 2|`[1,2]`|
|`3`|Article 3|`[3,1]`|

What I need to get from graphql:

```json
{
  "data": {
    "articles": [
      {
        "id": 1,
        "title": "Article 1",
        "authors": [
          {
            "id": 1,
            "name": "John Snow"
          },
          {
            "id": 2,
            "name": "Daenerys Targaryen"
          },
          {
            "id": 3,
            "name": "Tyrion Lannister"
          }
        ]
      },
      {
        "id": 2,
        "title": "Article 2",
        "authors": [
          {
            "id": 1,
            "name": "John Snow"
          },
          {
            "id": 2,
            "name": "Daenerys Targaryen"
          }
        ]
      },
      {
        "id": 3,
        "title": "Article 3",
        "authors": [
          {
            "id": 3,
            "name": "Tyrion Lannister"
          },
          {
            "id": 1,
            "name": "John Snow"
          }
        ]
      }
    ]
  }
}
```

It is expected that a single request must be executed to get all the authors.
What I am trying to do:
```cs
public class ArticleType : ObjectGraphType<Article>
{
    private NpgsqlConnection conn;

    public ArticleType(IDataLoaderContextAccessor accessor)
    {
        Field(a => a.Id).Name("id");

        Field(a => a.Title).Name("title");

        Field<ListGraphType<AuthorType>>().Name("authors")
            .ResolveAsync(async (ctx) =>
            {
                var authorsLoader = accessor.Context.GetOrAddCollectionBatchLoader<int, Author>("getAuthorsByIds", GetAuthorsByIds);
                int[] authorsIds = ctx.Source.AuthorsIds;
                var authors = await authorsLoader.LoadAsync(authorsIds);
                ...
            });
    }

    public Task<ILookup<int, Author>> GetAuthorsByIds(IEnumerable<int> ids)
    {
        ...
    }
}
```

[...]

PR #1282, opened 2019-09-02
feat: Convenience BatchDataLoader LoadAsync with multiple keys

benmccallum:
Addresses #1153

### DIFF
--- a/src/GraphQL/DataLoader/DataLoaderExtensions.cs
+++ b/src/GraphQL/DataLoader/DataLoaderExtensions.cs
@@ -17,7 +17,7 @@ namespace GraphQL.DataLoader
         /// </returns>
         public static Task<T[]> LoadAsync<TKey, T>(this IDataLoader<TKey, T> dataLoader, IEnumerable<TKey> keys)
         {
-            var tasks = new List<Task<T>>(keys.Count());
+            var tasks = new List<Task<T>>();
 
             foreach (var key in keys)
             {
--- a/src/GraphQL/DataLoader/DataLoaderExtensions.cs
+++ b/src/GraphQL/DataLoader/DataLoaderExtensions.cs
@@ -1,0 +1,30 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace GraphQL.DataLoader
+{
+    public static class DataLoaderExtensions
+    {
+        /// <summary>
+        /// Asynchronously load data for the provided given keys
+        /// </summary>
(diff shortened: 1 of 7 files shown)


benmccallum:
Toyed with the idea of making it `LoadManyAsync` rather than overloading `LoadAsync`, but just went with what you'd suggested @johnrutherford

benmccallum:
Ready for another review @sungam3r  and @johnrutherford

sungam3r:
Please add also overload with params so one can call `(1, 2, 3)` instead of `new int[] { 1, 2, 3 }`.

benmccallum:
Good idea. Done.

src/GraphQL/DataLoader/DataLoaderExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Asynchronously load data for the provided given keys
+        /// </summary>
+        /// <param name="dataLoader">The dataloader to use</param>
+        /// <param name="keys">Keys to use for loading data</param>
+        /// <returns>
+        /// A task that will complete when the DataLoader has been dispatched,
+        /// or a completed task if the result is already cached.
+        /// </returns>
+        public static async Task<T[]> LoadAsync<TKey, T>(this IDataLoader<TKey, T> dataLoader, IEnumerable<TKey> keys)
```
sungam3r: Also you can remove async/await from both methods.

src/GraphQL/DataLoader/DataLoaderExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        /// <param name="dataLoader">The dataloader to use</param>
+        /// <param name="keys">Keys to use for loading data</param>
+        /// <returns>
+        /// A task that will complete when the DataLoader has been dispatched,
+        /// or a completed task if the result is already cached.
+        /// </returns>
+        public static Task<T[]> LoadAsync<TKey, T>(this IDataLoader<TKey, T> dataLoader, IEnumerable<TKey> keys)
+        {
+            var tasks = new List<Task<T>>(keys.Count());
```
sungam3r: I think that `keys.Count()` call is redundant and can cause the key set to be recalculated if the caller somehow calculates it.  `var tasks = new List<Task<T>>();` will be quite enough.

I'm sure this is the last my remark before merge )

sungam3r:
fixes ##1153

benmccallum:
Resolved.

PR #1284, opened 2019-09-02
Schema fixes

### LINKED ISSUE #1156: No name collision safeguards for graph types
## Description

It's possible to define multiple `IGraphType` implementations with the same value returned for the `Name` property. When this occurs, one of them "wins" causing queries against the "loser" types to misbehave.

### Steps to reproduce

Here is a sample program that demonstrates the behavior.
```csharp
class Program
{
    {
        var schema = new Schema { Query = new QueryType() };

        {
            _.Query = @"
{
    type1 {
        field1A
        field1B
    }

    type2 {
        field2A
        field2B
    }
}"
;
        });

        Console.WriteLine(json);
    }
}

public sealed class Type1 : ObjectGraphType
{
    public Type1()
    {
        Name = "MyType";

        Field<StringGraphType>("Field1A", resolve: _ => "Field1A Value");
        Field<StringGraphType>("Field1B", resolve: _ => "Field1B Value");
    }
}

public sealed class Type2 : ObjectGraphType
{
    public Type2()
    {
        Name = "MyType";

        Field<StringGraphType>("Field2A", resolve: _ => "Field2A Value");
        Field<StringGraphType>("Field2B", resolve: _ => "Field1B Value");
    }
}

public sealed class QueryType : ObjectGraphType
{
    public QueryType()
    {
        Field<Type1>().Name("type1").Resolve(x => new {});
        Field<Type2>().Name("type2").Resolve(x => new {});
    }
}

```

### Expected result

Results have an error indicating that there are two different graph types that have the same name. Perhaps queries involving any of the conflicting types would be considered invalid?

### Actual result

```json
{
  "errors": [
    {
      "message": "Cannot query field \"field2A\" on type \"MyType\". Did you mean \"field1A\" or \"field1B\"?",
      "locations": [
        {
          "line": 9,
          "column": 9
        }
      ],
      "extensions": {
        "code": "5.2.1"
      }
    },
    {
      "message": "Cannot query field \"field2B\" on type \"MyType\". Did you mean \"field1B\" or \"field1A\"?",
      "locations": [
        {
          "line": 10,
          "column": 9
        }
      ],
      "extensions": {
        "code": "5.2.1"
      }
    }
  ]
}
```

### Workaround

I've put this in place in my project.
```csharp
public sealed class CustomGraphqlExecutor<TSchema> : GraphQL.Server.Internal.DefaultGraphQLExecuter<TSchema>
    where TSchema : ISchema
{
    public CustomGraphqlExecutor(TSchema schema, IDocumentExecuter documentExecuter, IOptions<GraphQLOptions> options, IEnumerable<IDocumentExecutionListener> listeners, IEnumerable<IValidationRule> validationRules) : base(schema, documentExecuter, options, listeners, validationRules)
    {
    }


    protected override ExecutionOptions GetOptions(
        string operationName,
        string query,
        Inputs variables,
        object context,
        CancellationToken cancellationToken)
    {
        var options = base.GetOptions(operationName, query, variables, context, cancellationToken);

        options.FieldMiddleware.Use<SchemaTypeValidationMiddleware>();

        return options;
    }
}

public sealed class SchemaTypeValidationMiddleware
{
    private static ConcurrentDictionary<string, System.Type> _knownTypes = new ConcurrentDictionary<string, System.Type>();
    public Task<object> Resolve(
        ResolveFieldContext context,
        FieldMiddlewareDelegate next)
    {
        CheckTypes(context);

        return next(context);
    }

    private void CheckTypes(ResolveFieldContext ctx)
    {
        var graphType = ctx.ReturnType.GetNamedType();

        _knownTypes.AddOrUpdate(graphType.Name, graphType.GetType(), (key, existingValue) => {
            if(existingValue != graphType.GetType()) {
                ctx.Errors.Add(new ExecutionError($"Found two GraphQL types with name \"{key}\"; types are {existingValue.FullName} and {graphType.GetType().FullName}"));
            }

            return existingValue;

        });
    }


}
```
I then register the custom executer via the following line when I configure the `IServicesCollection`:

```csharp
    services.AddTransient(typeof(IGraphQLExecuter<>), typeof(CustomGraphqlExecutor<>));
```

sungam3r:
fixes #1156
add checks in schema and lookup
fix test with duplicated Foo name


src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
```diff
@@ -1052,7 +1060,7 @@ public class FooInterfaceType : InterfaceGraphType
         {
             public FooInterfaceType()
             {
-                Name = "Foo";
+                Name = "IFoo";
```
sungam3r: Oops, the existing test contained an error.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -440,5 +467,36 @@ private IGraphType ConvertTypeReference(INamedType parentType, IGraphType type)
...
+                    // nothing to do
+                }
+                else if (existingGraphType.GetType() == type.GetType())
+                {
+                    _types[typeName] = type; // this case worked before overwriting the old value
+                }
+                else
+                {
+                    throw new InvalidOperationException(
+                        $@"You are trying to register a second GraphType '{type.GetType().FullName}' with the name '{typeName}'.
```
sungam3r: Quite a long message. Correct me if something is wrong. I am not a native speaker.

src/GraphQL/Types/Schema.cs
```diff
@@ -128,37 +84,42 @@ public IEnumerable<DirectiveGraphType> Directives
             get => _directives;
             set
             {
-                if (value == null)
```
sungam3r: Note that I changed the behavior. It seemed strange to me that passing `null` does not clear the collection.

src/GraphQL/Types/GraphTypesLookup.cs
floyd-may: I like the verbosity of the message, but the part of speech may be a bit awkward. Perhaps something like:

> Unable to register GraphType '{type.GetType().FullName}' with the name '{typeName}'; the name '{typeName}' is already registered to '{existingGraphType.GetType().FullName}'.

PR #1291, opened 2019-09-05
Set AutoRegisteringObjectGraphType.Name

### LINKED ISSUE #939: Name of types registered using AutoRegisteringObjectGraphType are wrong
## Description

All types registered with `AutoRegisteringObjectGraphType` have name ``AutoRegisteringObjectGraphType`1``

### Steps to reproduce

Register any type using `AutoRegisteringObjectGraphType`

### Expected result

Name of registered type should be the same as name of the source type passed to `AutoRegisteringObjectGraphType`

### Actual result

Type name is always ``AutoRegisteringObjectGraphType`1``

It is possible to set type name using `Name = typeof(TSourceType).GraphQLName()`. I can create pull request if you are fine with it.

### Environment

Environment

Version: 2.4.0
Windows: 10
Architecture: x64

sungam3r:
fixes #939

### DIFF
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -83,6 +83,7 @@ namespace GraphQL.Tests.Types
             GraphTypeTypeRegistry.Register<Money, AutoRegisteringObjectGraphType<Money>>();
 
             var type = new AutoRegisteringObjectGraphType<TestObject>(o => o.valuePair, o => o.someEnumerable);
+            type.Name.ShouldBe(nameof(TestObject));
             type.Description.ShouldBe("Object for test");
             type.DeprecationReason.ShouldBe("Obsolete for test");
             type.Fields.Count().ShouldBe(18);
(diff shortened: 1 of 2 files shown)


zak123:
I cannot for the life of me figure out how to actually use this -- does the setting of the name happen in the Model?
Would love an example of a model file where these things are set, last piece I need to deploy this into my project is changing the type name. 

What a useful feature btw, thank you for making this!

sungam3r:
Now **default** name for all graph types is set in ctor: https://github.com/graphql-dotnet/graphql-dotnet/blob/f0e192f24cf6d63151eead452f5f7bea90e09642/src/GraphQL/Types/GraphType.cs#L19

You can then **overwrite** this name. `AutoRegisteringObjectGraphType` and `AutoRegisteringInputObjectGraphType` overwrite it:
https://github.com/graphql-dotnet/graphql-dotnet/blob/f0e192f24cf6d63151eead452f5f7bea90e09642/src/GraphQL/Types/AutoRegisteringObjectGraphType.cs#L68

You can overwrite name further - after creating type:
```c#
var type = new AutoRegisteringObjectGraphType<MyModel>();
type.Name = "BlaBla";
```
or by making subclass:
```c#
public class MySuperClass : AutoRegisteringObjectGraphType<MyModel>
{
   public MySuperClass() => Name = "BlaBla";
}
```

zak123:
Thanks for the info! Incredibly helpful.

PR #1315, opened 2019-09-25
LongGraphType and BigIntGraphType

### LINKED ISSUE #1022: Int Scalar Type does not conform to GraphQL spec
## Description

[According to the documentation](https://graphql-dotnet.github.io/docs/getting-started/schema-types), The `IntGraphType` which represents an `Int` GraphQL type can be used for `int` (32-bit) and `long` (64-bit) c# types.
[According to the latest GraphQL specs](https://facebook.github.io/graphql/June2018/#sec-Int), the `Int` scalar type represents a 32-bit integer value. "The Int scalar type represents a signed 32‐bit numeric non‐fractional value. [...] If the integer internal value represents a value less than -2^31 or greater than or equal to 2^31, a field error should be raised."

### Steps to reproduce

Create a `IntGraphType` field with a backing `long` value.
Return a value from the field, that is >2^31.
Query the field.

### Expected result

The query should return an error.
The documentation should reflect, that `IntGraphType` only represents `int` values.
There could be an additional `LongGraphType` scalar type instead.

### Actual result

It returns the result, that is >2^31.

### Why is this a problem

Not conforming to the spec can lead to discrepancies when trying to work with different implementations of the GraphQL specs.

Personal reason:
This issue came up, because we use a development server, written in Javascript with apollo, that stitches together different graphql endpoints.
Requesting a `long` Field via that proxy server fails, because apollo conforms to the spec. Requesting from the endpoint written with graphql-dotnet directly returns a result.

sungam3r:
fixes #1022

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
@@ -1,6 +1,7 @@
 using GraphQL.Types;
 using Shouldly;
 using System;
+using System.Numerics;
 using Xunit;
 
 namespace GraphQL.Tests.Bugs
@@ -29,9 +30,10 @@ namespace GraphQL.Tests.Bugs
                     {
                         Code = "5.3.3.1"
                     }
-                    .AddLocation(1, 16)
                 }
             };
+            expected.Errors[0].AddLocation(1, 16);
+
             AssertQueryIgnoreErrors(query, expected, renderErrors: true, expectedErrorCount: 1);
         }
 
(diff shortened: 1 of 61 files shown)


src/GraphQL/Types/GraphType.cs
```diff
@@ -6,6 +6,13 @@ namespace GraphQL.Types
 {
     public abstract class GraphType : IGraphType
     {
+        protected GraphType()
```
sungam3r: I think it will be better to have default name instead of null

src/GraphQL/Execution/ExecutionError.cs
```diff
@@ -41,14 +41,16 @@ public ExecutionError(string message, Exception exception)
 
         public override IDictionary Data => DataAsDictionary;
 
-        public void AddLocation(int line, int column)
+        public ExecutionError AddLocation(int line, int column)
```
SimonCropp: not a fan of making this a fluent API.

src/GraphQL/Types/DecimalGraphType.cs
```diff
@@ -4,8 +4,6 @@ namespace GraphQL.Types
 {
     public class DecimalGraphType : ScalarGraphType
     {
-        public DecimalGraphType() => Name = "Decimal";
```
SimonCropp: seems these changes are unrelated to the long change? perhaps better as another PR?

src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
```diff
@@ -1,17 +1,56 @@
...
+        public void Very_Long_Number_Should_Return_Error_For_Int()
         {
-            var query = "{ big }";
+            var query = "{ int }";
+            var expected = new ExecutionResult { Errors = new ExecutionErrors { new ExecutionError("Value was either too large or too small for an Int32.", new OverflowException()) } };
+            AssertQueryIgnoreErrors(query, expected, renderErrors: true, expectedErrorCount: 1);
+        }
+
+        [Fact]
+        public void Very_Long_Number_In_Input_Should_Return_Error_For_Int()
```
SimonCropp: should the "bigger than a long scenario" also be tested?

src/GraphQL/Types/DecimalGraphType.cs
sungam3r: These changes are quite simple and obvious. Even last time when new scalars were added I wanted to do that. The appearance of Long was a good occasion.

SimonCropp:
But overall i think the changes look good

src/GraphQL/Execution/ExecutionError.cs
sungam3r: Ok

src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
sungam3r: I agree. I also thought that we really lack the `BigInteger` scalar. With its appearance we will have a complete picture of scalars for integer types. I have added tests.

src/GraphQL/Types/DecimalGraphType.cs
joemcbride: I'll also note that I'm not a fan of this change myself and may just reverse it.  I would rather it be very explicit what the name is and not use conventions.  Conventions like this is often where people start to get confused and do the wrong thing (not even realizing that they are _required_ provide a name), then try to force the framework to do something it was not intended to do.

sungam3r: @joemcbride  First of all, the default constructor’s goal was to give the type name a value other than `null`. Due to the absence of restrictions / checks on this and other properties of the schema, a number of errors could occur. See #1284 #1156 and related issues. The type must have a name. But it’s very easy to forget to set it. Also related to #1257.

> not even realizing that they are _required_ provide a name

Well, even before this change, they did not realize it. They **could** not set a name and subsequently face various kinds of incomprehensible behavior. So the situation has not gotten worse. Don't you think so?

joemcbride: > Due to the absence of restrictions / checks on this

So let's fix that real issue instead of making work-arounds for it.

> So the situation has not gotten worse.

It's worse in my book since now they won't be getting an error.  I prefer that they get an error over trying to teach and support a convention.  I've had to work with frameworks that use conventions heavily and it's a huge learning curve for developers and they forget half the time.  I would rather it just be explicit.

sungam3r: Let's say that this is a weighty argument and I also partially dislike conventions. But what about [this](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/GraphType.cs#L36) then? These are the same conventions. When I figured out why the above errors occurred, I just discovered this place where the type name is calculated. Apparently, the type name [was set](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/GraphTypesLookup.cs#L234) later than the code that caused the error. I thought so - since here the name is already assigned (by conventions from GetType().Name) then why not make it at the earliest moment - in ctor?

PR #1317, opened 2019-09-30
Made WriteAsync use async serilisation

MikeyFriedChicken:
This was done to stop the invalid operation exception as synchronous operations are disallowed


MikeyFriedChicken:

PR #1318, opened 2019-10-01
Call FlushInternalAsync when disposing

MikeyFriedChicken:
This prevents the invalid operation exception

image


MikeyFriedChicken:

PR #1319, opened 2019-10-01
Changed dispose method to async

MikeyFriedChicken:
This avoids the invalid operations exception regarding sync operations:

image

### DIFF
--- a/src/GraphQL/Http/HttpResponseStreamWriter.cs
+++ b/src/GraphQL/Http/HttpResponseStreamWriter.cs
@@ -239,14 +239,14 @@ namespace GraphQL.Http
             return FlushInternalAsync(flushEncoder: true);
         }
 
-        protected override void Dispose(bool disposing)
+        protected override async void Dispose(bool disposing)
         {
             if (disposing && !_disposed)
             {
                 _disposed = true;
                 try
                 {
-                    FlushInternal(flushEncoder: true);
+                    await FlushInternalAsync(flushEncoder: true);
                 }
                 finally
                 {


src/GraphQL/Http/HttpResponseStreamWriter.cs
```diff
@@ -239,14 +239,14 @@ public override Task FlushAsync()
             return FlushInternalAsync(flushEncoder: true);
         }
 
-        protected override void Dispose(bool disposing)
+        protected override async void Dispose(bool disposing)
```
pekkah: You never should use async voids..

src/GraphQL/Http/HttpResponseStreamWriter.cs
```diff
@@ -239,14 +239,14 @@ public override Task FlushAsync()
             return FlushInternalAsync(flushEncoder: true);
         }
 
-        protected override void Dispose(bool disposing)
```
pekkah: Dispose is part of the framework and shouldn't be changed to async. NET standard 2.1 has async disposable if this behavior is desirable.

src/GraphQL/Http/HttpResponseStreamWriter.cs
MikeyFriedChicken: Ahh yeah, I should have removed this pull request as i created the other one which seems fine.

src/GraphQL/Http/HttpResponseStreamWriter.cs
MikeyFriedChicken: Yep not needed now.  Thanks.

MikeyFriedChicken:

PR #1325, opened 2019-10-04
Synchronous operations are disallowed on ASP.NET Core 3

edsondewes:
To use the lib on an ASPNET Core 3 project, you must manually allow syncronous operations.

```csharp
// Startup.ConfigureServices
services.Configure<IISServerOptions>(options =>
{
  options.AllowSynchronousIO = true;
});
```

This pull request is to make it work without changing the default config.
"Serialize" was making syncronous writes to the Body stream, causing exceptions.

I dont know if is the best way to do this, but I couldn't find better async methods in Newtonsoft.Json.

Related to: graphql-dotnet/server#271

### DIFF
--- a/src/GraphQL/Http/DocumentWriter.cs
+++ b/src/GraphQL/Http/DocumentWriter.cs
@@ -1,9 +1,9 @@
-using System;
 using System.Buffers;
 using System.IO;
 using System.Text;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace GraphQL.Http
 {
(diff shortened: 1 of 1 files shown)


sungam3r:
@joemcbride you beat me with your answer in just 10 seconds

edsondewes:
Sorry, I didn't see the other pull request.

sungam3r:
@edsondewes Using asynchronous api should not be an end in itself. Especially when its use leads to degradation of performance for one reason or another. Firstly, the problem can be solved at the config level. There is nothing to worry about to allow sync IO. Secondly, with a high degree of probability in the future, the Newtonsoft.Json serializer will give way to a new api from the netstandard2.1 - System.Text.Json.

edsondewes:
For sure moving to the new json lib is the way to go, but it requires a bit of work to refactor the lib.
The only problem I see with this exception, is for people who are trying the lib for the first time with the new .net core.
Causes a first impression that it doesn't work.
But I understand your points. Thanks for the answer!

PR #1327, opened 2019-10-05
make GraphTypeTypeRegistry thread safe

SimonCropp:


sungam3r:
Are you really faced with thread safety issues? Or is this a change for the future?

SimonCropp:
yeah i had it happen a few times. i have different tests running in parallel that can add diff types to GraphTypeTypeRegistry.

sungam3r:
LGTM. This change will really fix this error, but keep in mind that the whole schema is not thread safe. It is not designed for thread safety.

SimonCropp:
the schema i can instantiate in each test (or group of tests) so that ok

sungam3r:
Also since `GraphTypeTypeRegistry` is a static type all its methods should be thread safe by design. It was an initial mistake not to use a thread safe dictionary.

PR #1329, opened 2019-10-07
[WIP] Add support of schema directives in GraphType first, add directives to introspection

### LINKED ISSUE #1002: How to set a custom directive on field?
## Summary

I want to build this schema programmatically:

```graphql
directive @isAuthenticated on FIELD_DEFINITION
directive @hasScope(scope: [String]) on FIELD_DEFINITION
type Product {
  id: ID!
  sku: String
  qty: Int
  price: Int
  parrot: String
  rating: Int @hasScope(scope: ["read:rating"])
}
type Query {
  product: Product @isAuthenticated
}
```

I know about `RegisterDirective` method on Schema but I don't undestand how I can set this directive to a **specific field**:

```csharp
public class ProductType : ObjectGraphType<Product>
    {
        public ProductType()
        {
            Name = "Product";

            Field(p => p.Id);
            Field(p => p.Qty, nullable: true);
            Field(p => p.Price, nullable: true);
            .....
            Field(p => p.Rating).Description("blablabla");
            // what I want to specify
            // Field(p => p.Rating).Description("blablabla").AddDirective(somedirective);
        }
    }
``` 

How can I add a directive to a field when building a schema?

**[UPDATE]** 
I tried to build a scheme not programmatically, but reading its contents from a SDL file using `Schema.For`. I see that the directives are read from the file and added to the `Schema._directives` field, but I don’t see that they somehow associated with the fields.

Also I checked if the directives are read by the downstream parser using this code:
```csharp
 var lexer = new GraphQLParser.Lexer();
 var parser = new GraphQLParser.Parser(lexer);
 var ast = parser.Parse(new GraphQLParser.Source(File.ReadAllText(PATH_TO_FILE)));
```
Inside the AST, those directives are associated with the relevant fields and their parameters are filled. This information simply does not fall into the `Schema` object.

Do I understand correctly that the object `FieldType` does not yet have support for the directives placed on it (also I checked `FieldType.Metadata` property and it is empty)?

### LINKED ISSUE #1328: Schema Directives in GraphType first
## Support of Schema Directives in GraphType first

Currently there is support of this directives in Schema First syntax (See this PR: #1172). But there is no way to set a directive on field definition using GraphType first syntax (https://github.com/graphql-dotnet/graphql-dotnet/issues/1002#issuecomment-537559320).

### Basic example

It can be implemented in a way:
```C#
public class SomeType : ObjectGraphType<Some>
{
	public SomeType()
	{
		Name = "Some";

		Field(p => p.SomeField).Description("Test").ApplySchemaDirective(...);
	}
}
```

mikelneo:
* Add `AddDirective` method to `ObjectGraphType`, `EnumerationGraphType`, etc. to support schema directives in GraphType first.
* Add directives applayed to Field to Introspection query. relates to https://github.com/graphql-dotnet/graphql-dotnet/issues/1332

fixes #1002 fixes #1328


mikelneo:
Close for a while to add support of all `TypeSystemDirectiveLocation` (https://graphql.github.io/graphql-spec/June2018/#sec-Type-System.Directives)

sungam3r:
This is a good change. It requires alignment of behavior with schema first approach.

mikelneo:
> You only need a single Visitor per directive

If directive contains args I still will need separate visitors for different arg values. But I got your idea, will try to do so.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -381,6 +392,8 @@ protected virtual EnumerationGraphType ToEnumerationType(GraphQLEnumTypeDefiniti
                 Name = enumDef.Name.Value,
                 Description = typeConfig.Description ?? enumDef.Comment?.Text
             };
+            type.SetAstType(enumDef);
```
sungam3r: So `SetAstType` has become much used I suggest to change
```c#
public static void SetAstType<T>(this IProvideMetadata type, T node) where T : ASTNode
{
    type.Metadata[__AST_MetaField__] = node;
}
```
to
```c#
public static TMetadataProvider SetAstType<TMetadataProvider>(this TMetadataProvider type, ASTNode node)
    where TMetadataProvider : IProvideMetadata => type.WithMetadata(__AST_MetaField__, node);
```
to be able to do this:
```c#
var type = new EnumerationGraphType
{
    Name = enumDef.Name.Value,
    Description = typeConfig.Description ?? enumDef.Comment?.Text
}.SetAstType(enumDef);
```

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -483,5 +485,91 @@ public static IValue AstFromValue(this object value, ISchema schema, IGraphType
...
+            {
+                directives = metadataProvider.GetMetadata(SchemaDirectivesMetadataKey, new Dictionary<string, SchemaDirectiveVisitor>());
+            }
+            else
+            {
+                directives = new Dictionary<string, SchemaDirectiveVisitor>();
+                metadataProvider.Metadata.Add(SchemaDirectivesMetadataKey, directives);
+            }
+
+            if (directives.ContainsKey(name))
```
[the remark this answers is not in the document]
sungam3r: I am talking about removing `if` and `throw` and changing `directives.Add(name, directive);` to `directives[name] = directive`.

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -483,5 +485,91 @@ public static IValue AstFromValue(this object value, ISchema schema, IGraphType
...
+        /// </summary>
+        /// <param name="metadataProvider"> Metadata. </param>
+        /// <param name="name"> Directive name. </param>
+        /// <param name="directive"> Schema directive. </param>
+        public static void SetDirective(this IProvideMetadata metadataProvider, string name, SchemaDirectiveVisitor directive)
+        {
+            Dictionary<string, SchemaDirectiveVisitor> directives;
+            if (metadataProvider.HasMetadata(SchemaDirectivesMetadataKey))
+            {
+                directives = metadataProvider.GetMetadata(SchemaDirectivesMetadataKey, new Dictionary<string, SchemaDirectiveVisitor>());
```
sungam3r: So if we are here than there is no need to allocate default value `new Dictionary<string, SchemaDirectiveVisitor>()` since metadata is not empty.

sungam3r: I looked more closely and realized that the second parameter is convenient for inferring the type of result. But this requires an instance to be created, which creates unnecessary memory allocation if the necessary metadata is available. So I suggest to add `TType IProvideMetadata.GetMetadata<TType>(string key, Func<TType> defaultValueFactory);` with that implementation:
```c#
public TType GetMetadata<TType>(string key, Func<TType> defaultValueFactory)
        {
            var local = Metadata;
            return local != null && local.TryGetValue(key, out var item) ? (TType)item : defaultValueFactory();
        }
```
and the caller will look like this
```c#
directives = metadataProvider.GetMetadata(SchemaDirectivesMetadataKey, () => new Dictionary<string, SchemaDirectiveVisitor>());
```

Just add `()=>` and no more unnecessary allocations :)

src/GraphQL.Tests/Utilities/SchemaVisitorTests.cs
```diff
@@ -91,5 +94,234 @@ public override void VisitField(FieldType field)
...
+            var schema = new Schema { Query = objectType };
+
+            AssertQuery(_ =>
+            {
+                _.Schema = schema;
+                _.Query = "{ hello }";
+            }, queryResult);
+        }
+
+        public class RegisterTypeDirectiveVisitor : SchemaDirectiveVisitor
```
sungam3r: Consider to move visitors to separate files.

src/GraphQL.Tests/Utilities/SchemaVisitorTests.cs
```diff
@@ -91,5 +94,234 @@ public override void VisitField(FieldType field)
...
+            var directivesMetadata = schema.GetDirectives();
+            directivesMetadata.ShouldNotBeNull();
+            directivesMetadata.Count.ShouldBe(1, "Only 1 directive should be added");
+            directivesMetadata.ContainsKey("test").ShouldBeTrue();
+
+            schema.FindType("TestType").ShouldNotBeNull();
+        }
+
+        [Fact]
+        public void can_create_custom_directive_for_all_locations()
```
sungam3r: Please add directives and tests for those locations too:

SCHEMA
SCALAR
INPUT_OBJECT
INPUT_FIELD_DEFINITION

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -13,6 +13,8 @@ namespace GraphQL
 {
     public static class GraphQLExtensions
     {
+        public const string SchemaDirectivesMetadataKey = "schema_directives";
```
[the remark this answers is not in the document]
mikelneo: Done.

src/GraphQL/Utilities/ISchemaNodeVisitor.cs
```diff
@@ -5,9 +5,18 @@ namespace GraphQL.Utilities
 {
     public interface ISchemaNodeVisitor
     {
+        void VisitSchema(Schema schema);
+        void VisitScalar(ScalarGraphType scalar);
         void VisitObjectGraphType(ObjectGraphType type);
```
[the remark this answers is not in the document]
mikelneo: Done.

src/GraphQL/GraphQLExtensions.cs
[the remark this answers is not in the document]
mikelneo: Done.

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -154,5 +154,12 @@ private FieldBuilder(EventStreamFieldType fieldType)
             FieldType.AsyncSubscriber = new AsyncEventStreamResolver<TSourceType, TReturnType>(subscribeAsync);
             return this;
         }
+
+        public FieldBuilder<TSourceType, TReturnType> Directive(string name, SchemaDirectiveVisitor directive)
+        {
+            FieldType.AddDirective(name, directive);
+
```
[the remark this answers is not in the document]
mikelneo: Done.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -482,6 +499,12 @@ protected virtual void VisitNode(object node, Action<ISchemaNodeVisitor> action)
                 foreach (var visitor in selector.Select(node))
                 {
                     action(visitor);
+
+                    // save directive to metadata
+                    if (node is IProvideMetadata metadata && visitor is SchemaDirectiveVisitor schemaDirective)
```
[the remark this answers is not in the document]
mikelneo: Done this way. But now if we don't call `base` method in inherited class, directive will not be saved to metadata.

src/GraphQL/Utilities/SchemaBuilder.cs
mikelneo: Done

src/GraphQL/Utilities/SchemaBuilder.cs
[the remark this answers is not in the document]
sungam3r: Yes, there is always such an opportunity.

src/GraphQL/Utilities/SchemaBuilder.cs
mikelneo: Done

src/GraphQL/GraphQLExtensions.cs
mikelneo: Replace with `metadataProvider.GetMetadata<Dictionary<string, SchemaDirectiveVisitor>>(SchemaDirectivesMetadataKey)`

src/GraphQL.Tests/Utilities/SchemaVisitorTests.cs
mikelneo: Moved

src/GraphQL.Tests/Utilities/SchemaVisitorTests.cs
mikelneo: Add test for this locations.
Note that currently there is no support of `ScalarTypeDefinition` in schema first approach. So test only for graph-type first.

src/GraphQL/Utilities/ISchemaNodeVisitor.cs
```diff
@@ -5,9 +5,18 @@ namespace GraphQL.Utilities
 {
     public interface ISchemaNodeVisitor
     {
-        void VisitObjectGraphType(ObjectGraphType type);
+        void VisitSchema(Schema schema);
+        void VisitScalar(ScalarGraphType scalar);
+        void VisitObject(ObjectGraphType type);
+        void VisitObject(IObjectGraphType type);
```
joemcbride: ObjectGraphType and IObjectGraphTpe are the same thing. You do not need both.

joemcbride:
@sungam3r Do NOT merge this until I approve it.  My previous comment still stands.

joemcbride:
As stated in my first review, this is not how the visitors were designed to be used.  They should not be used as multiple instances.  I won’t be approving this PR with that major change.

sungam3r:
@joemcbride OK. My comment about arguments still stands too. What to do with different arguments? What to do with visitor's dependencies?

I spend a lot of time reviewing this code and carefully approach it. I am not going to take hasty steps. Moreover, I see that we are simultaneously correcting other errors.

joemcbride:
> What to do with different arguments?

They should be pulled from the AST model.  Perhaps need a new context class that includes the AST and can use that class to help with getting the arguments.

> What to do with visitor's dependencies?

What dependencies?  Dependencies from a DI container?

sungam3r:
>  They should not be used as multiple instances. 

We already understood your concern. But is there **really** any great value in this - to have only one instance?

And here is my concern:
> Current design of setting `SchemaDirectiveVisitor.Arguments` may lead to errors if visitor configured as singleton. In that case we have global visitor and reset its Arguments property on each iteration in BuildVisitors method.

Now visitor stores the last set of arguments applied in its state. Don't you find this a somewhat design flaw?

> What dependencies? Dependencies from a DI container?

Yes.

src/GraphQL/Utilities/ISchemaNodeVisitor.cs
sungam3r: We had some doubts. Which one is better to leave and which to remove?

mikelneo: `VisitObject(ObjectGraphType type)` is not enough because both `ObjectGraphType` and `ObjectGraphType<TSource>` can have directives and `ObjectGraphType : ObjectGraphType<object>`.
So remove `void VisitObject(ObjectGraphType type)`.

sungam3r: I am for removing ObjectGraphType at all #1331

src/GraphQL/GraphQLExtensions.cs
mikelneo: Extended `IProvideMetadata` and inherited all possible classes from default implementation `MetadataProvider` to remove identical implementation

sungam3r: I think such a change with new base class should go as different PR. This will allow you to break down the issue into small separate problems and quickly merge it into master

mikelneo:
@joemcbride 
> As stated in my first review, this is not how the visitors were designed to be used. They should not be used as multiple instances.

As I can see current implementation of `SchemaDirectiveVisitor` in schema first approach it creates directives via `ServiceProvider`:
```c#
_visitorSelectors.Add(new DirectiveVisitorSelector(Directives, t => (SchemaDirectiveVisitor)ServiceProvider.GetRequiredService(t)));
``` 

So this does not mean that there is single instance of SchemaDirectiveVisitor for a type. For exampe, if we ues `GraphQL.DefaultServiceProvider` it always create new instance by `return Activator.CreateInstance(serviceType);`. Also if we use ASP.Net Core ServiceProvider and define `SchemaDirectiveVisitor ` as `Transient` it also will create new instance on every `GetRequiredService` call.

I want to understand what we actually want to do in GraphType first approach?
1. Do we also want to delegate `SchemaDirectiveVisitor` creation to `ServiceProvider`?
or
2. Do we really want every `SchemaDirectiveVisitor` be a singleton? This way we need to change visitor creation in schema first approach also.

sungam3r:
@mikelneo Nice catch about ServiceProvider. I do not understand why I had not expressed this idea so clearly in my comments before.

> They should not be used as multiple instances.

Now I don’t understand how this corresponds to reality :).

src/GraphQL/GraphQLExtensions.cs
mikelneo: Ok, create separate PR - https://github.com/graphql-dotnet/graphql-dotnet/pull/1349

joemcbride:
Please just leave this feature to me to implement.  The direction this has gone is far from the intended design.

sungam3r:
OK. In the meantime, we can use the modified package from AppVeyor build. But we would like to understand the intended design. We asked a number of questions and have not yet understood how this should work.

mikelneo:
In this PR also added additional field with directives used on field to Introspection query `__Field`.  Only directives allowed by `ISchemaFilter` will be added to introspection result.

sungam3r:
@joemcbride  I with @mikelneo  position this PR as a complex application of directives for the schema - both schema first and graphtype first. Directives can be obtained by introspection request. I studied this question quite a bit and re-read a whole bunch of issues on the github, especially graphql/graphql-spec#300. In fact, there is a big request from the community to receive directives through introspection. As correctly noted (and as I myself expected), the **specification does not say anything that the scheme for introspection is fixed**. I see that implementations for JS, Java, Ruby, etc. all provide the ability to set directives. Another thing is that because of the impossibility of providing information about directives out through introspection, a whole layer of functionality is lost - auth / roles / relations / defer / field cost / etc. Someone must be the first to implement exporting directives.

This PR still is under development, there are some things that need attention and correction. Directives will be added for all relevant elements. I would like to get a community opinion about this PR. We understand that this is a rather big (and in some way security critical) change. **And I would like to emphasize once again that it does not violate the specification**. It is regrettable that the development of the specification hinders the development of the implementation.

I will be grateful for your review.

sungam3r:
By the way, such an organization of directives is an alternative to using field middlewares. Directives, unlike middlewares, apply to specific schema elements. In this regard, they are more convenient and efficient.

joemcbride:
Closing this as it should not be pulled in.

sungam3r:
So, should we expect in the foreseeable future an alternative implementation from you, or at least a explanation of how this should be implemented? If the proposed solution does not at all correspond to what you originally conceived, then we would work on another.

mac2000:
Visitor idea seems to be really cool

BTW
please take into account that there is a chance that we gonna need to traverse up and down in schema while working with field directive, e.g. we are adding `@required(fields: "id")` will be nice if we can throw an exception here if there is no `id` field (yes it will be kind of runtime but still better than nothing)

joemcbride:
Visitor idea/patten works great with Schema first, but doesn’t make sense in GraphType first (you’re already at the definition - don’t need to traverse the Schema).  Doing schema traversal, at least at definition time, wouldn’t be possible with GraphType first (which is what this PR is).

mac2000:
Got it, seems like you are right here, definitely if we are going GraphType (aka code first) there is no need in traversal, I'm just wondering whether it will be possible to do something like:

```
class Acme {
  ctor() {
    Field('id')
    Field('foo').AddDirective(Required(fields: ?))
  }
}
````

to make it able to check if everything is ok at compile time

sungam3r:
> Visitor idea/patten works great with Schema first, but doesn’t make sense in GraphType first 

I understand this and simply began to consider the visitor as a dual-use tool. On the one hand, we can traverse the Schema. On the other hand, we can apply it locally on field. Consider this a special case. The CodeFirst approach inherently knows exactly where the directive needs to be attached, it **has no need** in traversing.

Therefore, I am convinced that visitors make sense in CodeFirst. In my opinion, this is only a problem of perception. You can come up with a different name so as not to feel discomfort.

sungam3r:
About expose directives via introspection and conformity to specification - https://github.com/graphql/graphql-spec/blob/master/CONTRIBUTING.md#contributing-to-graphql-libraries:

> To allow a library to remain spec compliant while also implementing proposals and drafts, it may request that these features are built so they are disabled by default with opt-in option flags or it may simply wait to merge a well-tested pull request until the spec proposal is accepted.

PR #1330, opened 2019-10-07
NameFieldResolver - CreatePropertyDelegate fix for proxies

cime:
This pull request fixes CreatePropertyDelegate method when used on runtime created proxies (for example when using NHibernate).

When a proxy is passed as target argument to CreatePropertyDelegate the following exception is thrown:
```Cannot bind to the target method because its signature is not compatible with that of the delegate type.```

The side effect of this change might be a little faster excecution of the delegate - https://www.palmmedia.de/Blog/2012/2/4/reflection-vs-compiled-expressions-vs-delegates-performance-comparison


sungam3r:
@cime Thanks!

PR #1339, opened 2019-10-11
Remove ConcurrentBag<T> usage im Metrics in favor of List<T>

### LINKED ISSUE #1324: Memory leak if EnableMetrics = true
Yesterday I noticed a large memory leak of my web-applcation. I dumped it through dotMemory and found a lot of System.Strings. In the dominators tab I found this:
image
PerfRecord is using only in Metrics class.
After disabling metrics memory leak stopped.

sungam3r:
fixes #1324

I myself faced the problem of excessive memory consumption on one of my projects. Profiling and investigation led to the real culprit - `ConcurrentBag<T>`:

https://stackoverflow.com/questions/5353164/possible-memoryleak-in-concurrentbag
https://stackoverflow.com/questions/9256735/is-concurrentbag-cause-of-memory-leak
https://stackoverflow.com/questions/10850443/is-there-a-memory-leak-in-the-concurrentbagt-implementation

So just changing it to ordinary `List<T>` with simple locking. I think that there will be problems with such blocking, because adding to the list is almost instantaneous, and the lock design is hybrid and first makes the spinlock instead of putting the thread in the wait state.

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
@@ -709,8 +709,7 @@ namespace GraphQL.Instrumentation
     public class Metrics
     {
         public Metrics(bool enabled = True) { }
-        public System.Collections.Generic.IEnumerable<GraphQL.Instrumentation.PerfRecord> AllRecords { get; }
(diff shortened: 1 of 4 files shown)


sungam3r:
I would like to merge it now. If there are comments, then I will create a new PR. An experiment on one of our services showed that the problem with memory consumption was gone.

Issue #118, opened 2016-07-15
Input 


baluubas:
I'm having some trouble with input variables that can be null and InputGraphObject. I have a mutation similar to this:

```
new QueryArguments(new[] {
      new QueryArgument<MyInput> { Name = "input" }
}) 

...

public class MyInput : InputObjectGraphType
{
	public MyInput()
	{
		Name = "MyInput ";
		Field<NonNullGraphType<StringGraphType>>("id");
		Field<StringGraphType>("foo");
		Field<StringGraphType>("bar");
	}
}
```

If I pass a null value for foo or bar I get: 
`Variable '$input_0' expected value of type 'MyInput'`

From what I can tell from the source code, this doesn't seem to be supported. I'm I thinking all wrong here or is this something that should be possible?

joemcbride:
This should be supported - can you also post the GraphQL mutation you are using?  The query language portion of it.

baluubas:
From the schema?

joemcbride:
The actual GraphQL mutation text that is being sent from the application to the server.

ex:

```
mutation M {
  first: immediatelyChangeTheNumber(newNumber: 1) {
    theNumber
  }
}
```

baluubas:
Ok, this the actual code I'm using and not the example code (react-relay is building these). The structure is the same though.

```
mutation UpdateOrganizationWebPresenceMutation($input_0:UpdateOrganizationWebPresenceInput!) {
     updateOrganizationWebPresence(input:$input_0) {
          clientMutationId,
          ...F1  
     }
}

fragment F0 on Organization {
  id,
  webPresence {
    website,
    careerPage, 
    facebook,
    instagram,
    twitter
  }
}

fragment F1 on UpdateOrganizationPayload {
  organization {
    ...F0  
  }
}

```

And variables:

```
{
  "input_0":{
    "id":"15a29266-fb7e-4c85-a8c0-3190e017e893",
    "__dataID__":"client:-18344258881",
    "website":"google.com",
    "careerPage":"google.com",
    "facebook":"adf",
    "instagram":null,
    "twitter":"asdf",
    "clientMutationId":"0"
  }
}
```

Note, the `instagram` null value which is causing the error..

joemcbride:
Thanks, you may have indeed found a bug.  I believe I've been able to duplicate this, will see if it is a quick fix.

joemcbride:
[...]

PR #1340, opened 2019-10-12
Compliance with spec in type naming rules

### LINKED ISSUE #922: Validate GraphType.Name
Shouldn't `GraphType.Name` also be validated with `NameValidator` to be more consistent with the [graphql specification](https://facebook.github.io/graphql/June2018/#sec-Names)?:

> GraphQL Documents are full of named things: operations, fields, arguments, **types**, directives, fragments, and variables. All names must follow the same grammatical form.

The way it works now is that I can give an invalid name (according to the specification) for a type (e.g. `Name = "My Type"`) and graphql-dotnet will not give any errors. Playground and graphiql complains (but not Voyager).

I would like to catch this error earlier, in my tests (i.e. `GraphQLResponse.Errors` should have this error).

sungam3r:
Type name checks are tightened to meet spec.

fixes #922
relates to #1257


src/GraphQL.Tests/Bugs/Bug118SpacesInTypeNameTests.cs
```diff
@@ -51,7 +51,7 @@ public class MyInput : InputObjectGraphType
     {
         public MyInput()
         {
-            Name = "MyInput ";
+            Name = "MyInput"; // changed from "MyInput "
```
sungam3r: This whitespace wasn't the root cause of #118 so should now delete it to pass the test.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -19,7 +19,7 @@ public class ComplexGraphTypeTests
         internal class ComplexType<T> : ComplexGraphType<T> {
             public ComplexType()
             {
-                Name = typeof(T).GetFriendlyName();
+                Name = typeof(T).GetFriendlyName().Replace("<", "Of").Replace(">", "");
```
sungam3r: `GetFriendlyName` returns forbidden characters for graph type name so just replace it.

src/GraphQL/Introspection/__Directive.cs
```diff
@@ -48,7 +48,7 @@ public class __DirectiveLocation : EnumerationGraphType<DirectiveLocation>
     {
         public __DirectiveLocation()
         {
-            Name = nameof(__DirectiveLocation);
+            SetName(nameof(__DirectiveLocation), validate: false);
```
sungam3r: `EnumerationGraphType` has already set `Name` in ctor to `DirectiveLocation` so need to set it again to `__DirectiveLocation` without validation. The same for `__TypeKind`.

src/GraphQL/Types/GraphQLTypeReference.cs
```diff
@@ -9,7 +9,6 @@ public class GraphQLTypeReference : InterfaceGraphType, IObjectGraphType
     {
         public GraphQLTypeReference(string typeName)
         {
-            Name = "__GraphQLTypeReference";
```
sungam3r: This name does not meet the specification.  `__` is reserved for fixed introspection type list so changed it to be `GraphQLTypeReference`.

sungam3r:
@SimonCropp, @joemcbride I think that now the validation for the type name is the most complete and consistent with the specification.

PR #1345, opened 2019-10-14
Include service provider in ResolveFieldContext

Shane32:
This pull request adds first class support for dependency injection in field resolvers, without breaking any existing applications, and without making the schema a scoped service.  It adds Services as a property to ExecutionOptions, which is passed through to a new Services property on the ResolveFieldContext class when resolving fields.

In addition, the default implementation of the ParallelExecutionStrategy will provide a scoped service provider for each execution (whereas the SerialExecutionStrategy will not).  For service providers that do not support scopes, such as the included DefaultServiceProvider, a scope will not be created and GetService requests will pass through to the root service provider.

To enhance performance, the service scope is only created if a service is requested during the field resolver function (see the ServiceProviderProxy class).  Performance impact appears to be about 1-2ms for 100,000 fields resolved (when not using the service provider within the field resolver), about 10x the speed of creating a service scope for every field resolved.

Notes:
- project requires reference to Microsoft.Extensions.DependencyInjection for the IServiceScope and IServiceScopeFactory interfaces; for maximum compatibility, v2.0.0 is referenced via nuget
- should probably have test code added
- confirm name of ExecutionOptions.Services property
- confirm name of ResolveFieldContext.Services property
- confirm name of ServiceProviderProxy class
- need to confirm which folder and namespace to put ServiceProviderProxy class
- need to update documentation to show that the service provider should be passed as a parameter to ExecutionOptions

Alternative algorithms: (which I decided against)
- when ExecutionOptions.Services is null, pull from schema (attempt cast from ISchema to Schema) and create a scope as schema will have a root scope (not needed for backwards compatibility, as ResolveFieldContext.Services did not previously exist)
- add a boolean property to the ParallelExecutionStrategy constructor to disable creating a scope for each resolved field, and instead just pass in the ExecutionOptions.Services property


sungam3r:
@Shane32 Please see #1160 and related issues. There was quite a bit of discussion about this problem.

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -39,5 +39,7 @@ public class ExecutionContext
         /// This can be useful for hiding error messages that reveal server implementation details.
         /// </summary>
         public Action<UnhandledExceptionContext> UnhandledExceptionDelegate { get; set; }
+
+        public IServiceProvider Services { get; set; }
```
sungam3r: You motivate client code to use ServiceLocator antipatttern. See #1160 and discussion around it.

src/GraphQL/GraphQL.csproj
```diff
@@ -27,6 +27,7 @@
   <ItemGroup>
     <PackageReference Include="GraphQL-Parser" Version="4.1.1" />
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.0.0" />
```
sungam3r: It is very unfortunate to see such a dependency.

src/GraphQL/Execution/ExecutionContext.cs
Shane32: From my point of view, not really.  I mean, the point of graphql is to access a back-end database.  Not static information like the StarWars sample.  So while the resolver function is immutable (which the schema and function definition still is), the database is mutable.  The usage implies that scope is _necessary_.  This is why ASP.Net creates a scope for each request - it just makes sense, regardless of the theory behind it.

src/GraphQL/GraphQL.csproj
Shane32: You could remove it, and allow for a custom ExecutionStrategy to be passed in which creates a scope per request.  The essential bit, having the services property passed into the ResolveFieldContext from the execution, would still remain.  However, database access is always going to be scoped, so functionally I feel that it's necessary.

src/GraphQL/Execution/ExecutionContext.cs
sungam3r: GraphQL is not about accessing databases. 

You expose `Services`. What will the client do with it? Call `GetService` method of course :)

>This is why ASP.Net creates a scope for each request - it just makes sense, regardless of the theory behind it.

Creating the scope and somehow accessing this scope are different things. The problem of resolving dependencies can be successfully solved without resorting to a ServiceLocator. The main thing is to allow the DI container to do the work for which it was created. I advise you to read the discussion in #1160.

src/GraphQL/GraphQL.csproj
sungam3r: > However, database access is always going to be scoped

One can argue with that. I would not say so.

Shane32:
I mean, I suppose you could have a singleton dbcontext for the entire app.  It would be pretty stupid, forcing all code to run on a single thread, with synclock statements for each db access.  But it is possible.

sungam3r:
@Shane32 In your reasoning you are **very attached** to a specific use case. GraphQL can and is successfully used not only for access to databases. For example, it can be an adapter to one or more REST services. In addition, even in a database access scenario, you are again tied to a specific technology - Entity Framework. In the meantime, you can communicate with the databases through ADO.NET, through some special provider/service or just HTTP.

Shane32:
Well, I have not seen non-database use cases for graphql.

ADO.Net (which is a database) relies on a SqlConnection, which is essentially a scoped service.  It does not have to be, but Microsoft documentation recommends that a connection be created for each request, so it can internally use connection pooling.  This especially true if transactions are in use, as the transaction is tied to the connection.

And your other example, REST APIs, are typically databases themselves.  Sure you might not need context for a REST API, but at that point graphql is just acting as an adapter.  Even then, some APIs absolutely do need context -- for instance, they may need to authenticate before the request is made.

Shane32:
One reason I really feel the IServiceProvider should be part of the ResolveFieldContext is because of the poor performance of the two alternative approaches:

Approach A:

1. Service provider is stored into an AsyncLocal (via a custom class)
2. Custom ParallelExecutionStrategy which modifies the AsyncLocal for each field, which in turn duplicates the dictionary held internally during the async call
3. Additional code in the resolvers to pull the service provider from the AsyncLocal

Approach B:

1. Store the IServiceProvider on the UserContext field, causing a string lookup every time the field is read/set.  
2. Custom ParallelExecutionStrategy which modifies the UserContext for each field
3. Additional code in the resolvers to pull the service provider from the UserContext

The sample in #1160 is not a viable option, as it does not dispose of the service scope, or allow for multiple services to pull from the same service scope.

Being that .Net Core seems to be designed with dependency injection in mind, I feel that this graphql library should also support scoped services via dependency injection.  As it is now, it does not.  Even scoping the schema would not help when executing in parallel.  And (as far as I can tell) the DataLoader requires the ParallelExecutionStrategy to be effective, and support of the DataLoader will become critical in my code.

So, I feel that this decision to exclude dependency injection from field resolvers should be reviewed.  At a very minimum, all that needs to happen is for the IServiceProvider to be passed through the ResolveFieldContext class during execution.  Then people can use the SerialExecutionStrategy easily for access to scoped services, or have the choice of writing a custom ExecutionStrategy to create a scope for each resolver (as my pull request did).

Thanks for listening and the good work on the project thus far!  The amount of effort that has gone into it is incredible.

sungam3r:
> The sample in #1160 is not a viable option, as it does not dispose of the service scope

I already answered in a related issue that you are wrong and are based on a wrong judgment.

sungam3r:
@Shane32 By the way, I've been using #1160 for more than half a year now on all of our microservice projects (several dozen pieces) and with it there have been no problems with any incorrect destruction of objects or memory leaks or something else. Please see a little more carefully the proposed solution. It seems to me that you are missing its essence, which fundamentally distorts your perception of advantages and disadvantages.

sungam3r:
I am for "Approach C":
1. Do not use `IServiceProvider` at all and let DI container to do all the work.

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -62,5 +62,11 @@ public class ExecutionOptions
         /// Provides the ability to filter the schema upon introspection to hide types.
         /// </summary>
         public ISchemaFilter SchemaFilter { get; set; } = new DefaultSchemaFilter();
+
+        /// <summary>
+        /// Sets the service scope to be used for this request.  If not set, execution will default
```
sungam3r: > If not set..

Do you assume that this is normal when this property is not set? And what should the client code in the resolver do in this case?

src/GraphQL/Execution/ServiceProviderProxy.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            _serviceScope?.Dispose();
+        }
+
+        /// <summary>
+        /// Gets the service object of the specified type, creating a scope if none has yet been created
+        /// </summary>
+        /// <param name="serviceType">An object that specifies the type of service object to get.</param>
+        /// <returns>A service object of type serviceType. -or- null if there is no service object of type serviceType.</returns>
+        public object GetService(Type serviceType)
```
sungam3r: This method is not thread safe.

src/GraphQL/Execution/ExecutionOptions.cs
Shane32: For maximum compatibility, I designed it such that if the developer left it null, then all the ResolveFieldContext.Services fields would be null.  If the developer wished to used ResolveFieldContext.Services, they would set ExecutionOptions.Services.  It could grab the service provider from the schema, but that is a root service provider, and would not be scoped, so that is not a good option.  It could grab the service provider from the httpcontext, but that would rely on or assume that Asp.Net Core was in use

src/GraphQL/Execution/ParallelExecutionStrategy.cs
```diff
@@ -44,5 +44,19 @@ await OnBeforeExecutionStepAwaitedAsync(context)
...
         }
+
+        protected override async Task<ExecutionNode> ExecuteNodeAsync(ExecutionContext context, ExecutionNode node)
+        {
+            if (context.Services == null)
+            {
+                return await ExecuteNodeAsync(context, node, null)
+                    .ConfigureAwait(false);
+            }
+            using (var serviceProvider = new ServiceProviderProxy(context.Services))
```
sungam3r: So you create a new scope on each execute iteration. This may not be what the client code expects. If you take an ASP.NET Core, then a scope is created on each HTTP request and dependencies are resolved in it. Here you create an indefinite number of nested scopes in which dependencies will be resolved in an independent manner. DbContext (as an example) will be created many times - depending on the number of fields that the client defined in the request.

Did I understand your intentions correctly?

src/GraphQL/Execution/ServiceProviderProxy.cs
Shane32: Fixed (using `lock(this)`)

src/GraphQL/Execution/ParallelExecutionStrategy.cs
Shane32: Yes, those are my exact intentions.  For a use case sample, take a look at the open source GraphQL.EntityFramework library by Simon Cropp.  The "root" resolve function gets all of the data, and all the child resolve functions do not need access to the dbcontext.  Hence why the ServiceProviderProxy exists - because most resolve functions do not need access to the service provider.  (Understandably, this use case is somewhat specific to those that use EF.)  However, other queries can and should execute in parallel for optimal speed - which requires a separate dbcontext.

Shane32: Since almost no classes support multithreading, any code running under a ParallelExecutionStrategy would have to have another scope for each iteration.  As you say, this may be unexpected, in which case SerialExecutionStrategy should really be the default (when using scoped services).  But that's just a preference, I suppose, as REST adapters would probably want a parallel strategy by default.

src/GraphQL/GraphQL.csproj
Shane32: The reference has been changed to Microsoft.Extensions.DependencyInjection.Abstractions (which has no dependencies)

src/GraphQL/Execution/ExecutionOptions.cs
sungam3r: You go through the same series of questions that I went through. Too many assumptions, too much manual work.

src/GraphQL/Execution/ParallelExecutionStrategy.cs
sungam3r: Well, at least now the context and purpose are clear.

src/GraphQL/GraphQL.csproj
```diff
@@ -27,7 +27,7 @@
   <ItemGroup>
     <PackageReference Include="GraphQL-Parser" Version="4.1.1" />
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.0.0" />
```
sungam3r: In fact, move to `.Abstractions` does not change much in comparison with general remarks in this PR.

src/GraphQL/GraphQL.csproj
sungam3r: This is the expected step. But the mere presence of this dependency harms.

benmccallum:
@davidfowl, remember when you tweeted about OS projects that you could potentially help? graphql-dotnet is still trying to figure out the best way to support DI in resolvers that are within a singleton schema. It's been over a year I'd say of proposals, PRs from some, but no real forward momentum. Would be awesome if you had some time to weigh in.

davidfowl:
I’ll find some time to look

sungam3r:
Maybe after a while I will return to viewing this PR. This does not mean yet that it will be merged. I should once again consider the pros and cons.

Shane32:
No problem :)

Shane32:
@sungam3r  I've been thinking about this some more, especially as it relates to #1537 .

Perhaps we should do without the whole `ServiceProviderProxy` class.  To recap, the concept is that you can run a query with a ParallelExecutionStrategy, but yet anything that relies on a scoped service creates a scope for that resolver.  So when using with EF, for example, it would support two different resolvers pulling separate queries from EF via DI simultaneously, and without the need for Transient service registration.

That's great, but then let's consider the costs: first, the scope is never explicitly disposed, and it just relies on GC to dispose of the scope and all the objects that the scope contains.  Not preferable.  Second, the ServiceProviderProxy class is created for every resolver.  Better than creating a scope, but not as fast as a more transparent implementation.

If I removed that class from this PR, then the ParallelExecutionStrategy simply runs resolvers in parallel, ignoring the consequences of simultaneous access to services.  Not necessarily bad, as the name ("parallel") implies this behavior.  Anyone wanting to run resolvers in parallel that relies on scoped services that don't support parallel operation would simply create their own scope within the resolver.

In fact, a IFieldResolver implementation could create a scope to allow parallel operation.  For example you might write `builder.ResolveScoped(context => context.Services.GetService<MyDbContext>().Set<Posts>().ToListAsync())`.  The ScopedFieldResolver would create a scope and dispose of it when the field resolver finishes executing.  This would allow parallel execution of scoped services [which do not support multithreaded access], without affecting the speed of any other resolver.

Adding the ScopedFieldResolver still requires a nuget dependency on `Microsoft.Extensions.DependencyInjection.Abstractions`.  I don't think this is an issue, but as before, this feature could be eliminated to eliminate the dependency.  Users could still write  `builder.Resolve(context => context.Services.CreateScope().ServiceProvider.GetService<MyDbContext>().Set<Posts>().ToListAsync())` for the same functionality (although they should have a using in there to close the scope).

I'm going to change this PR as described above, as it would provide the same features without the performance hit, albeit in a slightly different fashion.  Let me know if you have any comments.

sungam3r:
>  Anyone wanting to run resolvers in parallel that relies on scoped services that don't support parallel operation would simply create their own scope within the resolver.

I would change _would_ to _must_. And in general, you yourself said the thought that I have - if the client has a special behavior about scopes, **then let him realize it himself** instead of cluttering up the common code.

sungam3r:
Regarding  dependency on `Microsoft.Extensions.DependencyInjection.Abstractions` and exposing `context.Services` - I would like to avoid these things in the **core** package to keep it clean, but I am OK to make such a feature in another package. We already have 2 packages for integration with various serializers so I don’t see anything wrong to add one more package with  MS DI integration. Also see #1160.

benmccallum:
+1 if we can have a separate package for the convenience access, kinda like how MediatR has a MediatR.Microsoft.Extensions.DependencyInjection package.

Shane32:
@benmccallum A very simple patch could be implemented in user applications with a few lines of code, by the way:
``` C#
    public static class ContextExtensions
    {
        public static T GetService<T>(this ResolveFieldContext context) {
            return ((Schema)context.Schema).Services.GetRequiredService<IHttpContextAccessor>().HttpContext.RequestServices.GetRequiredService<T>();
        }
    }
```
This snippet is applicable for preview 1352, not the current commit.  This of course is a specific fix when using Asp.Net Core and serial execution.

@sungam3r That was the original goal of #1376 -- a new approach to DI as a separate project to be used along with the core library.  I'm still advocating for rudimentary access to DI (specifically scoped DI) within the core project, which is what this PR will demonstrate (in my opinion).

sungam3r:
@Shane32 your example is good. And the good thing is that in fact, with **one line** of code, it demonstrates that we do not need to do anything in the core package. We **may** make some other package with advanced integration cases for MS DI, ОК. But in general I feel that It will be enough to give just your code example (of course in a more detailed form) in README/DOCS to point reader what to do with scoped dependencies.

>  This of course is a specific fix when using Asp.Net Core and serial execution.

I agree that this is specific to ASP.NET Core, but not to serial execution. We just give a way to obtain scoped dependency. Whether the obtained object itself supports parallel access or not, these are the problems of this object, not ours. And frankly, I’m not much mistaken if I say that 99% consumers of this package indeed are ASP.NET Core apps.

Shane32:
True, and frankly I’m not much mistaken if I say that 99% of scoped dependencies do not allow parallel access.

Shane32:
What I do not know is how many projects rely on scoped dependencies.  All of mine do.  But I can see other projects, such as you previously described (wrappers to REST APIs, and the like), not requiring scoped dependencies.

pekkah:
I'm actually using similar pattern on Tanka GraphQL https://pekkah.github.io/tanka-graphql/0-getting-started/05dependencyinjection.html. It's used by the code generator to allow controller style implementation of resolvers https://pekkah.github.io/tanka-graphql/3-code-generator/6-usage.html. Benefit of my approach is that the services can be registered as singleton, scoped (per graphql execution) or transient (per resolver execution).

sungam3r:
> 99% of scoped dependencies do not allow parallel access.

OK, then let the client code synchronize access to these dependencies. GraphQL itself has nothing to do with it.

@pekkah Thanks for example.

@Shane32 @pekkah @joemcbride @benmccallum I want to understand whether we are close to consensus on the issue raised here after our discussions. So for me the question are:
1. Is it worth adding any code to this project to support scoped deps by referencing MS DI?
2. Is it worth adding any code to this project to support scoped deps?
3. Is it worth adding new project for MS DI integration on top of graphql-dotnet?
4. Is it worth adding code snippet from @Shane32 for requesting scoped deps to the README/DOCS?

For myself, I would answer like:
1. Completely not.
2. Well... maybe, but but I have not seen such option yet.
3. I'm OK with that, but as I said earlier I don’t see much sense in this if the same effect can be achieved with one line of code on client side.
4. Yes.

If we reach a consensus, this will allow us to close a number of existing issues and focus on other tasks.

Shane32:
Personally:

1. I would, but I can see not doing so.  Keep in mind the MS DI reference works with Autofac also.
2. Absolutely
3. Depends on the number of additional features.  For a few lines of code, no.
4. Until scoped dependency support is built-in, yes.

I see scoped dependency support as an essential part of the design of this project.

sungam3r:
@Shane32 Some conflicts here.

Shane32:
Yup I have not gone through this and redone it yet.

sungam3r:
Now `Schema` explicitly implements `IServiceProvider` and we removed `Services` property from `Schema`, so....

GrangerAts:
I read through and followed perhaps half of it, but I'm also looking to solve a similar problem. What's the "Gql.Net way" to solve it? I'd be fine with having a scoped service provider on `IResolveFieldContext` (@Shane32 's #1571 PR), but what's critical is that people **know about it**.

Here's a simplified chronology of dev work in our project:
1. EFC's DbContext doesn't like parallel execution, so that's Scoped.
1. We're unaware of any other way to get DI services into a Field resolver, so the EFC DbContext goes into our GqlType's constructor and the resolver captures it in the closure/lambda. That means that the Schema has to be Scoped too.
1. Someone makes a field who's resolver uses the DbContext from the constructor.
1. Someone adds a second field in the GqlType that uses that same DbContext from the constructor. Execution fails at runtime when both fields are requested ("[InvalidOperationException] A second operation started on this context before a previous operation completed...").

You can't fix that by switching DbContexts to Transient, nor by switching the GqlType to Transient. 

Coming into this discussion, the only options I thought I had were:

1. Complicate every call to a DbContext by wrapping them in a critical section. (Lame)
1. Switch our project to use the [fuget ](https://www.fuget.org/packages/GraphQL) repo so we can set the new `ExecutionOptions.MaxParallelExecutionCount` to 1 and force serial execution. (Lame)

But now, thanks to @Shane32, I am now aware that I should be able to use a different hack based on the HttpContextAccessor. It sure would be better if it didn't feel like I'm working _around_ Gql.Net.

sungam3r:
@GrangerAts did you see #1160 ?

GrangerAts:
@sungam3r - Yes. Sorry, I was going to look at that too, but forgot to mention it.

Shane32:
@GrangerAts Keep in mind that non-async field resolvers always run in series.  So if your only async field resolvers are calling a single EF context, then basically either:

1. You must create a new DI scope for each resolver that relies on scoped dependencies (for example, using the `ScopedFieldResolver` class from #1571 , which with some minor edits you can copy into your own project) so they can run in parallel, or
2. You might as well use a SerialExecutionStrategy

That being said, the way I see it, there are a couple primary ways access scoped dependencies within a singleton graph:

1. See https://github.com/graphql-dotnet/graphql-dotnet/pull/1345#issuecomment-587107092

2. See #1160 

You could also look at PR #1376 , which adds a much more complex layer of DI support to graphql as a separate project.  It also fixes the dataloader.  You can do things like tag certain fields to run in parallel, with or without a dedicated DI scope.  Fork it if you want to use it, of course, or additional changes might break your code.  So far, I'm keeping the branch up to date with the latest published nuget preview release, not the latest updates in the `master` branch.  I do use it in my own projects.

GrangerAts:
@Shane32 Thanks; I'll look into those too.  Thanks for the reminder about the async; we're async pretty much everywhere at this point, even though DbContext doesn't like it. :)

sungam3r:
> UPDATE: Umm, sorry, but I would not want to use that approach. It's a lot of extra stuff to implement all over an existing codebase; the maintenance burden for users of that solution is really high.

Really? In fact, the example shows that everything has already been done in the accompanying libraries. User code declares dependencies and consumes them. That is all. What do you mean when speaking about "maintenance burden" ?

> The problem of resolving dependencies can be successfully solved without resorting to a ServiceLocator. The main thing is **to allow the DI container to do the work for which it was created**.

sungam3r:
> 1. See https://github.com/graphql-dotnet/graphql-dotnet/pull/1345#issuecomment-587107092

It must be noted in the docs. It's an easy way to get what you want without code complication inside graphql-dotnet repo. And I think that DI related documentation will be rather huge.

Shane32:
> > 1. See [#1345 (comment)](https://github.com/graphql-dotnet/graphql-dotnet/pull/1345#issuecomment-587107092)
> 
> It must be noted in the docs. It's an easy way to get what you want without code complication inside graphql-dotnet repo. And I think that DI related documentation will be rather huge.

I've already got it in the upgrade notes.  I should check and see where else it should be mentioned.

GrangerAts:
> 
> > UPDATE: Umm, sorry, but I would not want to use that approach. It's a lot of extra stuff to implement all over an existing codebase; the maintenance burden for users of that solution is really high.
> 
> Really? In fact, the example shows that everything has already been done in the accompanying libraries. User code declares dependencies and consumes them. That is all. What do you mean when speaking about "maintenance burden" ?

What I'm getting at is that your solution requires additional changes to Startup to add more services that wrap the services that already exist in the DI container. Someone attempting to use your solution can't just "fix their code" in the field resolver or even in the GqlType. They have to modify the DI container to get it to work.

> 
> > The problem of resolving dependencies can be successfully solved without resorting to a ServiceLocator. The main thing is **to allow the DI container to do the work for which it was created**.

You solution is a ServiceLocator, then: The constructor is now injecting a service that waits to resolve/locate the actual service until later.


The reality of Gql.Net is that the "field resolver" paradigm introduces this situation where people need to switch from constructor-based DI to method-based DI.  This is the exact same thing that's supported in Startup.ConfigureServices(), Startup.Configure(), and Controller classes: method-based DI.

sungam3r:
> You solution is a ServiceLocator

Disagree. ServiceLocator gives the ability to resolve **anything**. You do not know what exactly. Func<T> gives ability to resolve particular service, nothing more. All this has been said many times in linked issues so I allow myself not to repeat. We suggested a number of ways, each has advantages and disadvantages, you are free to choose any.

GrangerAts:
> 
> 
> > You solution is a ServiceLocator
> 
> Disagree. ServiceLocator gives the ability to resolve **anything**. You do not know what exactly. Func gives ability to resolve particular service, nothing more. All this has been said many times in linked issues so I allow myself not to repeat. We suggested a number of ways, each has advantages and disadvantages, you are free to choose any.

Yes, I can see that; you're right that it's not ServiceLocator. I still disagree with the approach, but whatev's. 

IMO, the last bit of the initial comment on the #1571 PR, and various comments on #1160, is the best direction: Gql.Net should support method-based DI (for the resolvers).  And it would be wise to have an additional version of this support that creates and manages a DI scope that lasts for the execution of the resolver. 

And on that note, since Gql.Net doesn't yet support "method-based DI" for the field resolvers, I'll have to DIY some lesser solution for now.

GrangerAts:
@sungam3r, I think I see the light in the general direction you were going. 

I think the solution is with an Abstract Factory that creates typed/generic Service Factories that simply work with the current DI container. 

E.g. the exact same idea of how .Net core has `ILoggerFactory` and people normally inject an `ILogger<T>`. Except in this case, instead of doing something like `logger.LogError()`, you'd do something like inject an `IServiceFactory<T> sf` in a constructor, then use something like `sf.Create()` to instantiate a `T` from DI services. 

...or even something like `using (T instance = sf.CreateScoped();` if you need to instantiate the object in a new scope. But I'm off-the-cuff brainstorming here.

sungam3r:
This library can use both approaches - by-resolver DI or mine. They complement each other.

GrangerAts:
@sungam3r  and @Shane32  

FYI, I have a solution. No need to wait for Gql.Net to implement more fancy field resolvers, though that would still be really nice. @sungam3r, you may be pleased to note that it's similar to the direction you went.

Essentially, it's changing from the use of `IServiceScopeFactory` (Service Locator pattern) to an `IServiceScopeFactory<T>`. And when I was trying to figure out if it was OK to name it that, I found someone else had attempted the same thing. [My version is here in this response on StackExchange.](https://codereview.stackexchange.com/a/238506/219835)  

You inject this into the constructor instead of your DbContext, then lazy-resolve it when you actually need it. Since you have to make and use a new scope, the DbContext has no problems when Gql.Net runs in parallel.

Shane32:
Basically very similar to #1160 except it provides a scope also. 

This code can only be used to retrieve a single object from the created scope. Of course dependencies are fine. And you have to clean up after the scope. 

I don’t use #1160 but if I did, I’d probably use it along with the ScopedFieldResolver so that there’s less code in each resolver rather than having to write a using statement for each resolver

Shane32:
Note that by using a combination of #1160 and `ScopedFieldResolver`, you can (if you want to) obtain two objects from the same DI scope.  You'd of course have to write your own internal service provider context via `AsyncLocal`, but that would be easy.  The context would be initialized by a DocumentListener (or otherwise), and the ScopedFieldResolver edits the value during the execution of the resolver.  The resolver would use the non-service-locator-anti-pattern `Func<T>` method to pull the service from the service provider context.

That's very similar to how I structured method-parameter-based DI injection in #1376 , but there you don't need to use `Func<T>` as it examines the method parameters to inject the proper services, with a scope if requested.

sungam3r:
I think back to work on #1160 . Over the past time, we have extensively tested this approach on our projects and added new features. I think we should move forward with both #1160 and resolver-level DI approaches.

sungam3r:
>  You'd of course have to write your own internal service provider context via AsyncLocal, but that would be easy.

No more, by the way :)

Shane32:
If you have a new or updated idea @sungam3r I’d be happy to see it.

sungam3r:
Yes, I have. There are new ideas and old ones too, since I initially did not use all in the example. I will return to this PR when repo/package for AddFunc will be ready

fenomeno83:
> @GrangerAts Keep in mind that non-async field resolvers always run in series. So if your only async field resolvers are calling a single EF context, then basically either:
> 
> 1. You must create a new DI scope for each resolver that relies on scoped dependencies (for example, using the `ScopedFieldResolver` class from #1571 , which with some minor edits you can copy into your own project) so they can run in parallel, or
> 2. You might as well use a SerialExecutionStrategy
> 
> That being said, the way I see it, there are a couple primary ways access scoped dependencies within a singleton graph:
> 
> 1. See [#1345 (comment)](https://github.com/graphql-dotnet/graphql-dotnet/pull/1345#issuecomment-587107092)
> 2. See #1160
> 3. Put the `IServiceProvider` into the `UserContext` when starting the execution, and pull services from the service provider inside the `UserContext` property of the `IResolveFieldContext`.  (Also could be done via extension method.)
> 
> You could also look at PR #1376 , which adds a much more complex layer of DI support to graphql as a separate project. It also fixes the dataloader. You can do things like tag certain fields to run in parallel, with or without a dedicated DI scope. Fork it if you want to use it, of course, or additional changes might break your code. So far, I'm keeping the branch up to date with the latest published nuget preview release, not the latest updates in the `master` branch. I do use it in my own projects.

Hi.
I've solved problem to have scoped dbcontext into singleton graphql middleware using factory.

Here my full example, where is included also an automatic graphqltype generator by dto

Shane32:
I propose we close this PR.  While I still like this concept - where the RequestServices property, if used, will create a scope - it's probably not the best design for standard behavior of this project.  Offering a scoped field resolver such as #1571 suggests is a better concept, and can be applied as an extension rather than having to be integrated into the execution strategy.

sungam3r:
@Shane32 Either close this or resolve conflicts and let's go on. We can say that #1730 rebooted the whole set of DI PRs.

PR #1347, opened 2019-10-15
Added IHasUserContext abstraction

J-kit:
Makes it easier to write custom extension methods for the usercontext

### DIFF
--- a/src/GraphQL/Abstractions/IHasUserContext.cs
+++ b/src/GraphQL/Abstractions/IHasUserContext.cs
@@ -1,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace GraphQL.Abstractions
+{
+    public interface IHasUserContext
+    {
+        IDictionary<string, object> UserContext { get; }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/GraphQL/Abstractions/IHasUserContext.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace GraphQL.Abstractions
```
sungam3r: Do we need new namespace? Better move to `GraphQL.Execution`

src/GraphQL/Abstractions/IHasUserContext.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    public interface IHasUserContext
```
sungam3r: There is already `IProvideMetadata` interface so this one can be also named `IProvideUserContext`

J-kit: Hmm IHasX seems a bit more obvious, but if that's the convention im ok with that.

J-kit:
> I will be glad to merge this after you fix the tests. `ApiApprovalTests.PublicApi.approved.txt` needs to be modified with new apis.

Ahh ok thanks, i didn't know this :P
Done

PR #1349, opened 2019-10-15
Extend IProvideMetadata with default value factory

mikelneo:
* Added `TType GetMetadata<TType>(string key, Func<TType> defaultValueFactory)` to IProvideMetadata to prevent unnecessary memory allocation
* Inherited all possible classes from default implementation MetadataProvider to remove identical implementation.


sungam3r:
LGTM

PR #1351, opened 2019-10-17
Increased extensibility of Fieldbuilder

J-kit:
The fieldbuilder forgets about the original fieldname. I want to track the original fieldname in my project, so overriding it and setting the metadata would be the easiest solution for me. This requires this Fieldbuilder method to be virtual.

Code in my project : 

>       public override FieldBuilder<TSourceType, TProperty> Field<TProperty>
>         (
>             string name, 
>             Expression<Func<TSourceType, TProperty>> expression, 
>             bool nullable = false, 
>             Type type = null
>         )
>         {
>             var result =  base.Field(name, expression, nullable, type);
>             result.FieldType.Metadata["OriginalPropertyName"] = expression.NameOf();
>             return result;
>         }

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
@@ -1516,7 +1516,7 @@ namespace GraphQL.Types
             where TGraphType : GraphQL.Types.IGraphType { }
         public GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Field<TGraphType, TReturnType>(string name = "default") { }
         public GraphQL.Builders.FieldBuilder<TSourceType, object> Field<TGraphType>() { }
-        public GraphQL.Builders.FieldBuilder<TSourceType, TProperty> Field<TProperty>(string name, System.Linq.Expressions.Expression<System.Func<TSourceType, TProperty>> expression, bool nullable = False, System.Type type = null) { }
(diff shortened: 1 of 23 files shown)


sungam3r:
> The fieldbuilder forgets about the original fieldname.

Do you mean original _PropertyName_ ? `name` parameter was preserved in `result.FieldType.Name`.

sungam3r:
Well, you can write extension method like this one
```c#
public static FieldBuilder<TSourceType, TProperty> FieldWithMeta<TSourceType, TProperty>(
  this ComplexGraphType<TSourceType> type, 
  string name,
  Expression<Func<TSourceType, TProperty>> expression,
  bool nullable = false,
  Type type = null)
{
  var result = type.Field<TProperty>(name, expression, nullable, type);
  result.FieldType.Metadata["OriginalPropertyName"] = expression.NameOf();
  return result;
}
```
to do the same things without changing graphql source code. `ComplexGraphType<TSourceType>` has many methods and none of them is virtual. It will be strange to make virtual just this one. Either it is necessary to change everything or nothing. I prefer not to change anything.

J-kit:
This would work for me:

```csharp
public static FieldBuilder<TSourceType, TProperty> FieldWithMeta<TSourceType, TProperty>(
	this ComplexGraphType<TSourceType> graphType,
	Expression<Func<TSourceType, TProperty>> expression,
	bool nullable = false,
	Type type = null)
{
	var name = expression.NameOf();
	return graphType.Field<TProperty>(name, expression, nullable, type).WithOriginalName(name);
}
```


But i'd rather keep the existing code intact by just intercepting a central function. 
In my opinion, this would bea much cleaner solution.

sungam3r:
@joemcbride What are you think about it?

J-kit:
Any news? @sungam3r  @joemcbride

PR #1360, opened 2019-10-22
Fix for ShouldIncludeNode

sungam3r:
From spec:
> Neither @skip nor @include has precedence over the other. In the case that both the @skip and @include         directives are provided on the same field or fragment, it must be queried only if the @skip condition         is false and the @include condition is true. Stated conversely, the field or fragment must not be queried         if either the @skip condition is true or the @include condition is false.

### DIFF
--- a/docs2/site/docs/getting-started/arguments.md
+++ b/docs2/site/docs/getting-started/arguments.md
@@ -1,6 +1,6 @@
 # Arguments
 
-You can provide arguments to a field. You can use `GetArgument` on `ResolveFieldContext` to retrieve argument values. `GetArgument` will attempt to coerce the argument values to the generic type it is given, including primitive values, objects, and enumerations. You can gain access to the value directly through the `Arguments` dictionary on `ResolveFieldContext`.
+You can provide arguments to a field.  You can use `GetArgument` on `ResolveFieldContext` to retrieve argument values.  `GetArgument` will attempt to coerce the argument values to the generic type it is given, including primitive values, objects, and enumerations.  You can gain access to the value directly through the `Arguments` dictionary on `ResolveFieldContext`.
 
 ```graphql
 query {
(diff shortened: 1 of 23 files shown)


joemcbride:
There is no reason to be making all of those changes in the docs.  Lets stop doing changes like that in PRs that are 1) a _preference_ 2) in a change set totally unrelated to the specific PR.

joemcbride:
Just so I'm clear, that also applied to code formatting preferences.  The "I resharpered your code for you" PRs are getting tiresome.

sungam3r:
Code formatting preferences should be consistent throughout the whole repository. No matter what specific preferences are used - the main thing is that they are **consistent**. If I find such things working on current issue, then I just make separate commit. Yes, I agree, it is always better to make a separate PR so as not to mix tasks. I myself do not like to make a review of such changes. But sooner or later it will have to be done.

By the way, does the use of double spaces in the documentation make any sense? Does this somehow affect the resulting markup? Or is it the setting of the editor used? Just wondering.

joemcbride:
> Code formatting preferences should be consistent throughout the whole repository.

I agree, and they were until several people started switching them to their own preferences vs. what was already there.  So quit switching them to YOUR preference vs. what was already in the repository.  At this point, please avoid any big code formatting changes.  In the big picture it doesn't matter and just makes it that much harder to actually do code reviews.

> By the way, does the use of double spaces in the documentation make any sense? Does this somehow affect the resulting markup?

No it has no effect on the rendering.  The point is, _it_ _does_ _not_ _matter_.  Let's stop making sweeping changes that don't matter.

sungam3r:
No problem. How do you feel about the practice of using .editorconfig?

joemcbride:
https://github.com/graphql-dotnet/graphql-dotnet/blob/master/.editorconfig

joemcbride:
It's great.  If this was a JS project I would have put in eslint.  I would be fine with a solution for code formatting that isn't StyleCop / Resharper and works with VS + VSCode + can fail the build with invalid formatting.

sungam3r:
Ok. I explicitly described this intention in #1361 .

PR #1366, opened 2019-10-23
Made EnumValueDefinition and QueryArgument more extendable

J-kit:
Problem: I have a query, similar to this:
```csharp
query smorder{
  categories(orderBy:CategoryId){
    id
  }
}
```

I personally work with the GraphiQl webeditor and want to get autocompletition based on the supported input. 
The idea now is to take an EnumerationGraphtype for this. Fine. But im a very lazy person, so i want Graphql to automatically gather the required information by itself, so i wrote the following code to support the guys who have to use my api:

```csharp
   public class FieldAwareQueryArgument<TType> : QueryArgument
    {
        private IGraphType resolvedType;

        public FieldAwareQueryArgument() : base(typeof(TypeFieldEnumeration<TType>))
        {
        }

        public override IGraphType ResolvedType
        {
            get => this.resolvedType;
            set
            {
                if (value is EnumerationGraphType egt)
                {
                    foreach (var fieldDefinition in egt.Values.OfType<FieldAwareEnumValueDefinition>())
                    {
                        if (typeof(ListGraphType).IsAssignableFrom(fieldDefinition.FieldType.Type))
                        {
                            fieldDefinition.Visible = false;
                        }
                    }
                }
                this.resolvedType = value;
            }
        }
    }

    public class TypeFieldEnumeration<TType> : EnumerationGraphType, IAutoResolve
    {
        public TypeFieldEnumeration(IServiceProvider serviceProvider)
        {
            this.Name = $"{typeof(TType).Name}_Enumeration";
            if (typeof(IComplexGraphType).IsAssignableFrom(typeof(TType)))
            {
                var graphType = serviceProvider.GetService<TType>() as IComplexGraphType;
                var fields = graphType.Fields;

                foreach (var field in fields)
                {
                    base.AddValue(new FieldAwareEnumValueDefinition
                    {
                        Name = field.Name,
                        Description = field.Description,
                        Value = field.GetOriginalName(),
                        DeprecationReason = null,
                        FieldType = field
                    });
                }
            }
            else
            {
                var fields = typeof(TType).GetProperties();
                foreach (var field in fields)
                {
                    base.AddValue(field.Name, field.Name, field.Name);
                }
            }
        }
    }

    public class FieldAwareEnumValueDefinition : EnumValueDefinition
    {
        public FieldType FieldType { get; set; }
    }
```

Now, i can use this in my code: 

```csharp
  this.FieldAsync<ListGraphType<CategoryType>>(
                "categories",
                arguments: 
                    new QueryArgument<IntGraphType> { Name = "skip", Description = "filter categories by menuid", DefaultValue = null },
                    new QueryArgument<IntGraphType> { Name = "take", Description = "filter categories by menuid", DefaultValue = null },
                    new FieldAwareQueryArgument<CategoryType> { Name = "orderBy", Description = "order categories by this", DefaultValue = null }
                ),
                resolve: async context =>
                {
                    var skip = context.GetArgument<int>("skip", 0);
                    var take = context.GetArgument<int>("take", int.MaxValue);
                    var sortOrder = context.GetArgument<string>("orderBy", null);

                   ...
                })
```

Without this changes, this wouldnt be possible. If you have any suggestions how i could implement such a feature without changin anything in graphql or a better way of archiving this, i would be more than happy for feedback!

You have my permission to use this or the idea to implement in a future version of graphql-dotnet

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
@@ -1622,6 +1622,7 @@ namespace GraphQL.Types
         public string Description { get; set; }
         public string Name { get; set; }
         public object Value { get; set; }
+        public bool Visible { get; set; }
     }
     public class EnumValues : System.Collections.Generic.IEnumerable<GraphQL.Types.EnumValueDefinition>, System.Collections.IEnumerable
     {
@@ -1906,7 +1907,7 @@ namespace GraphQL.Types
         public object DefaultValue { get; set; }
         public string Description { get; set; }
         public string Name { get; set; }
-        public GraphQL.Types.IGraphType ResolvedType { get; set; }
+        public virtual GraphQL.Types.IGraphType ResolvedType { get; set; }
         public System.Type Type { get; }
     }
     public class QueryArgument<TType> : GraphQL.Types.QueryArgument
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
@@ -1622,7 +1622,6 @@ namespace GraphQL.Types
         public string Description { get; set; }
         public string Name { get; set; }
         public object Value { get; set; }
-        public bool Visible { get; set; }
     }
     public class EnumValues : System.Collections.Generic.IEnumerable<GraphQL.Types.EnumValueDefinition>, System.Collections.IEnumerable
     {
@@ -1907,7 +1906,7 @@ namespace GraphQL.Types
         public object DefaultValue { get; set; }
         public string Description { get; set; }
         public string Name { get; set; }
-        public virtual GraphQL.Types.IGraphType ResolvedType { get; set; }
+        public GraphQL.Types.IGraphType ResolvedType { get; set; }
         public System.Type Type { get; }
     }
     public class QueryArgument<TType> : GraphQL.Types.QueryArgument
--- a/src/GraphQL/Types/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/EnumerationGraphType.cs
@@ -119,15 +119,10 @@ namespace GraphQL.Types
 
     public class EnumValueDefinition : MetadataProvider
     {
-        public EnumValueDefinition()
-        {
-            Visible = true;
-        }
-
         public string Name { get; set; }
         public string Description { get; set; }
         public string DeprecationReason { get; set; }
         public object Value { get; set; }
-        public bool Visible { get; set; }
+        public bool Visible { get; set; } = true;
     }
 }
--- a/src/GraphQL/Types/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/EnumerationGraphType.cs
@@ -112,7 +112,7 @@ namespace GraphQL.Types
 
         public void Add(EnumValueDefinition value) => _values.Add(value ?? throw new ArgumentNullException(nameof(value)));
 
-        public IEnumerator<EnumValueDefinition> GetEnumerator() => _values.Where(x => x.Visible).GetEnumerator();
+        public IEnumerator<EnumValueDefinition> GetEnumerator() => _values.GetEnumerator();
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
     }
@@ -123,6 +123,5 @@ namespace GraphQL.Types
         public string Description { get; set; }
         public string DeprecationReason { get; set; }
         public object Value { get; set; }
-        public bool Visible { get; set; } = true;
     }
 }
--- a/src/GraphQL/Types/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/EnumerationGraphType.cs
@@ -1,8 +1,8 @@
 using GraphQL.Language.AST;
 using GraphQL.Utilities;
+
 using System;
 using System.Collections;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -112,16 +112,22 @@ namespace GraphQL.Types
 
         public void Add(EnumValueDefinition value) => _values.Add(value ?? throw new ArgumentNullException(nameof(value)));
 
-        public IEnumerator<EnumValueDefinition> GetEnumerator() => _values.GetEnumerator();
+        public IEnumerator<EnumValueDefinition> GetEnumerator() => _values.Where(x => x.Visible).GetEnumerator();
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
     }
 
     public class EnumValueDefinition : MetadataProvider
     {
+        public EnumValueDefinition()
+        {
+            Visible = true;
+        }
+
         public string Name { get; set; }
         public string Description { get; set; }
         public string DeprecationReason { get; set; }
         public object Value { get; set; }
+        public bool Visible { get; set; }
     }
 }
--- a/src/GraphQL/Types/QueryArgument.cs
+++ b/src/GraphQL/Types/QueryArgument.cs
@@ -39,7 +39,7 @@ namespace GraphQL.Types
 
         public object DefaultValue { get; set; }
 
-        public virtual IGraphType ResolvedType { get; set; }
+        public IGraphType ResolvedType { get; set; }
 
         public Type Type { get; private set; }
     }
--- a/src/GraphQL/Types/QueryArgument.cs
+++ b/src/GraphQL/Types/QueryArgument.cs
@@ -39,7 +39,7 @@ namespace GraphQL.Types
 
         public object DefaultValue { get; set; }
 
-        public IGraphType ResolvedType { get; set; }
+        public virtual IGraphType ResolvedType { get; set; }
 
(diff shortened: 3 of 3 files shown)


src/GraphQL/Types/EnumerationGraphType.cs
```diff
@@ -112,16 +112,22 @@ public class EnumValues : IEnumerable<EnumValueDefinition>
...
+        public EnumValueDefinition()
+        {
+            Visible = true;
+        }
+
         public string Name { get; set; }
         public string Description { get; set; }
         public string DeprecationReason { get; set; }
         public object Value { get; set; }
+        public bool Visible { get; set; }
```
sungam3r: No need for ctor, just
```c#
public bool Visible { get; set; } = true;
```

J-kit: Wasn't sure 'bout the convention, thanks :3

sungam3r:
I understood your idea, interesting. I don’t understand why this piece of code is needed:
```c#
if (value is EnumerationGraphType egt)
                {
                    foreach (var fieldDefinition in egt.Values.OfType<FieldAwareEnumValueDefinition>())
                    {
                        if (typeof(ListGraphType).IsAssignableFrom(fieldDefinition.FieldType.Type))
                        {
                            fieldDefinition.Visible = false;
                        }
                    }
                }
```

You do not want to see list fields in the values of the enumeration? Could this filtering be done in `TypeFieldEnumeration` ctor? Then `Visible` property not needed - just do not call `base.AddValue`.

And I do not understand the purpose of else case in `TypeFieldEnumeration`.

J-kit:
> I understood your idea, interesting. I don’t understand why this piece of code is needed:
> 
> ```cs
> if (value is EnumerationGraphType egt)
>                 {
>                     foreach (var fieldDefinition in egt.Values.OfType<FieldAwareEnumValueDefinition>())
>                     {
>                         if (typeof(ListGraphType).IsAssignableFrom(fieldDefinition.FieldType.Type))
>                         {
>                             fieldDefinition.Visible = false;
>                         }
>                     }
>                 }
> ```
> 
> You do not want to see list fields in the values of the enumeration? Could this filtering be done in `TypeFieldEnumeration` ctor? Then `Visible` property not needed - just do not call `base.AddValue`.
> 
> And I do not understand the purpose of else case in `TypeFieldEnumeration`.

TypeFieldEnumeration is a general purpose enumeration type which converts a GraphType or System.Type to a EnumerationGraphType. It doesnt differenciate the Fields/Properties in it. Thats not it's responsibility.

Whereas FieldAwareQueryArgument is a concrete implementation of a QueryArgument. This type 'knows' what it's for: ordering results. We wanted to exclude list entries, because we can't order by them, we can only order by primitive types, like string or int.

> Could this filtering be done in `TypeFieldEnumeration` ctor?
No, because TypeFieldEnumeration is not instanciated by my code, thats the frameworks/graphql-dotnet's responsibility

sungam3r:
> TypeFieldEnumeration is a general purpose enumeration type 
> Whereas FieldAwareQueryArgument is a concrete implementation of a QueryArgument.

So your general purpose class uses concrete implementation class and in this way also become concrete.

sungam3r:
Sorry, I mixed up two classes

J-kit:
Is the question still to be answered or was it just the confusion?

sungam3r:
I am confused by the modification of `Visible` property in `ResolvedType` setter. This is some kind of side effect. Moreover, there can be many arguments with this type of enumeration. If you set this property, then this should happen only once and in a more appropriate place.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
```diff
@@ -1622,6 +1622,7 @@ namespace GraphQL.Types
         public string Description { get; set; }
         public string Name { get; set; }
         public object Value { get; set; }
+        public bool Visible { get; set; }
```
sungam3r: No need in such custom property `Visible`. You can use `MetadataProvider`'s api to read and write any data you want.

joemcbride:
> If you set this property, then this should happen only once and in a more appropriate place.

I agree with this.  It seems like the behavior is happening in the wrong place.

This is an interesting idea, but is not part of the GraphQL spec and seems like would be confusing to try to maintain in the project.

sungam3r:
> TypeFieldEnumeration is a general purpose enumeration type which converts a GraphType or System.Type to a EnumerationGraphType. It doesnt differenciate the Fields/Properties in it. Thats not it's responsibility.

And yet it seems to me that this is the responsibility of the data type (enumeration) itself, not the argument of this type. Then no additional properties will be needed. You can make `TypeFieldEnumeration` abstract and add new type like `TypeFieldEnumerationWithoutLists<T> : TypeFieldEnumeration<T>`

sungam3r:
>  but is not part of the GraphQL spec 

Well, this is just nothing more than another internal part of the executing mechanism. This is not to say that it is complex, just an enumeration is assembled from a set of fields of the specified type.

sungam3r:
@J-kit So you can solve this issue without changes in graphq-dotnet - just move filtering code inside enumeration type.

J-kit:
Hmm yea you're right, now that i think about this, it seems to be out of place....

If i want to set the `ResolvedType` in `FieldAwareQueryArgument` to a new instance of TypeFieldEnumeration, this would require a serviceprovider. 
This is not possible as the `arguments` of  `ComplexGraphType`.`FieldAsync` require objects, not types

So how could i solve this?

J-kit:
> And yet it seems to me that this is the responsibility of the data type (enumeration) itself, not the argument of this type. Then no additional properties will be needed. You can make TypeFieldEnumeration abstract and add new type like TypeFieldEnumerationWithoutLists<T> : TypeFieldEnumeration<T>

I'll try that, thanks

sungam3r:
> But im a very lazy person, so i want Graphql to automatically gather the required information by itself

The idea itself is great 😎

J-kit:
> > But im a very lazy person, so i want Graphql to automatically gather the required information by itself
> 
> The idea itself is great 😎

Aww thanks <3
Thanks for so much input, i would have overseen the side effect^^

Turns out that there is no change to be made to get the behaviour i wanted \o/

What i changed / added: 
````csharp
public class TypeFieldEnumerationWithoutLists<TType> : TypeFieldEnumeration<TType>
    {
        public TypeFieldEnumerationWithoutLists(IServiceProvider serviceProvider) : base(serviceProvider)
        {
        }

        public override IEnumerable<EnumValueDefinition> GetEnumValueDefinitions()
        {
            return base.GetEnumValueDefinitions().Where(x => ValidateEnumValueDefinition(x));
        }

        private bool ValidateEnumValueDefinition(EnumValueDefinition valueDefinition)
        {
            if (!(valueDefinition is FieldAwareEnumValueDefinition fieldDefinition))
            {
                return true;
            }
            if (typeof(ListGraphType).IsAssignableFrom(fieldDefinition.FieldType.Type))
            {
                return false;
            }
            else
            {
                var originalName = fieldDefinition.FieldType.GetOriginalName();
                var sourceType = fieldDefinition.FieldType.GetSourceType();
                if (sourceType == null)
                {
                    return true; //TODO or false, not sure, need more brain to evaluate
                }

                var requiredProperty = sourceType.GetProperty(originalName, comparisonType: StringComparison.OrdinalIgnoreCase);
                if (requiredProperty == null)
                {
                    return false;
                }
            }

            return true;
        }
    }

    public class TypeFieldEnumeration<TType> : EnumerationGraphType, IAutoResolve
    {
        private readonly IServiceProvider serviceProvider;

        public TypeFieldEnumeration(IServiceProvider serviceProvider)
        {
            this.serviceProvider = serviceProvider;
            this.Name = $"{typeof(TType).Name}_Enumeration";

            // ReSharper disable once VirtualMemberCallInConstructor
            foreach (var enumValueDefinition in this.GetEnumValueDefinitions())
            {
                base.AddValue(enumValueDefinition);
            }
        }

        public virtual IEnumerable<EnumValueDefinition> GetEnumValueDefinitions()
        {
            if (typeof(IComplexGraphType).IsAssignableFrom(typeof(TType)))
            {
                var graphType = this.serviceProvider.GetService<TType>() as IComplexGraphType;
                var fields = graphType.Fields;

                foreach (var field in fields)
                {
                    yield return new FieldAwareEnumValueDefinition
                    {
                        Name = field.Name,
                        Description = field.Description,
                        Value = field.GetOriginalName(),
                        DeprecationReason = null,
                        FieldType = field
                    };
                }
            }
            else
            {
                var fields = typeof(TType).GetProperties();
                foreach (var field in fields)
                {
                    yield return new EnumValueDefinition
                    {
                        Name = field.Name,
                        Value = field.Name,
                        Description = field.Name
                    };
                }
            }
        }
    }

    public class FieldAwareQueryArgument<TType> : QueryArgument
    {
        public FieldAwareQueryArgument() : base(typeof(TypeFieldEnumerationWithoutLists<TType>))
        {
        }
    }
````
> @J-kit So you can solve this issue without changes in graphq-dotnet - just move filtering code inside enumeration type.

AYE!


> This is an interesting idea, but is not part of the GraphQL spec and seems like would be confusing to try to maintain in the project.

What do you mean by confusing? Its just an enumeration type generated at scheme definition



Do you want me to do a PR with such a feature for the core framework?

J-kit:
Continue with issue: #1367

PR #1368, opened 2019-10-24
Add ability use byte size scalables as graphql variables

KnightVision90:
Added tests and conversions from and to byte data types to be used with variables. Currently, it fails to convert a number given as a variable saying it can't convert from int32 to byte.


src/GraphQL.Tests/Execution/AstFromValueTests.cs
```diff
@@ -96,7 +96,7 @@ public void converts_byte_to_byte_value()
             byte value = 12;
             var result = value.AstFromValue(schema, new ByteGraphType());
             result.ShouldNotBeNull();
-            result.ShouldBeOfType<ByteValue>();
+            result.ShouldBeOfType<GraphQL.Language.AST.ByteValue>();
```
sungam3r: Is it definitely necessary?

KnightVision90: The other option would be to remove ByteValue and ByteGraphType from this file. This test is currently failing when running for me due to it expecting the GraphQL.Tests.Execution.ByteValue but is getting GraphQL.Language.AST.ByteValue.

sungam3r: I think I get it. You added the built-in conversion and now the control does not fall into the test converter. The fact is that these tests were written before adding many primitive types (byte, sbyte, short, ushort, etc.) to the list of standard scalar types. Therefore, you do not need to change the namespace, but simply delete all unnecessary types from tests: `ByteValueConverter`, `ByteValue`, `ByteGraphType`.

KnightVision90: I kept `ByteValueConverter` due to it being used in `registers_ast_from_value_converters` so that it can still be tested that the `schema.RegisterValueConverter` method is still covered.

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -443,9 +443,19 @@ public static IValue AstFromValue(this object value, ISchema schema, IGraphType
...
-                return new ShortValue(int16);
+                return new SByteValue(@sbyte);
+            }
+
+            if (serialized is byte @byte)
+            {
+                return new ByteValue(@byte);
+            }
+
+            if(serialized is short @short)
```
sungam3r: @KnightVision90 I will be happy to merge this after this small fix:
```c#
if(serialized is short @short)
```
->
```c#
if (serialized is short @short)
```

KnightVision90: Oops, that should be fixed now.

PR #1371, opened 2019-10-27
Async validation

### LINKED ISSUE #1140: Update validation rules to support Task<T>
Today validation rules `Validate` method do not return `Task<INodeVisitor>`.  Update this interface so that you can return a `Task` to support async validation.

```csharp
public interface IValidationRule
{
    Task<INodeVisitor> ValidateAsync(ValidationContext context);
}
```

sungam3r:
fixes #1140

### DIFF
--- a/src/GraphQL.Tests/Validation/ValidationTestBase.cs
+++ b/src/GraphQL.Tests/Validation/ValidationTestBase.cs
@@ -91,7 +91,7 @@ namespace GraphQL.Tests.Validation
             var documentBuilder = new GraphQLDocumentBuilder();
             var document = documentBuilder.Build(query);
             var validator = new DocumentValidator();
-            return validator.Validate(query, schema, document, rules);
+            return validator.ValidateAsync(query, schema, document, rules).Result;
         }
     }
 }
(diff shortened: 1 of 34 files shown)


src/GraphQL.Tests/Validation/ValidationTestBase.cs
```diff
@@ -91,7 +91,7 @@ private IValidationResult Validate(string query, ISchema schema, IEnumerable<IVa
             var documentBuilder = new GraphQLDocumentBuilder();
             var document = documentBuilder.Build(query);
             var validator = new DocumentValidator();
-            return validator.Validate(query, schema, document, rules);
+            return validator.ValidateAsync(query, schema, document, rules).Result;
```
SimonCropp: shouldn't this be made async instead of `.Result`?

sungam3r: Can be left as is. `asynk` will not give a significant positive effect in tests, more importantly this:
изображение

Then you have to make a lot of changes and make the tests asynchronous, it’s not worth it.

Issue #1378, opened 2019-10-30
NameFieldResolver class is not public


Shane32:
I'm trying to create a custom IExecutionStrategy, but the default implementation contains the following line of code within the protected virtual ExecuteNodeAsync method, which cannot run in a separate library due to NameFieldResolver being an internal class:
``` C#
var resolver = node.FieldDefinition.Resolver ?? new NameFieldResolver();
```
I suggest that NameFieldResolver be a public class.  This allows for custom implementations of IExecutionStrategy to maintain 100% compatibility with other code without rewriting/copying substantial amounts of graphql code.  I can submit a pull request if desired.

sungam3r:
Ok.

sungam3r:
Closed by #1379

PR #1379, opened 2019-10-30
Change NameFieldResolver to public

Shane32:
See #1378 

Note that the ApiApprovalTests need to be updated.  I assume this is something you'd manage.


sungam3r:
Updating approval tests is your responsibility. They were added just to make all public api changes explicitly visible.

Shane32:
👍  done

Issue #1321, opened 2019-10-01
Fresh Release


msheldon83:
Are there any plans to publish a new release of the library?  There is at least one bug fix that I'm looking for that was fixed back in February.

joemcbride:
There will not be any more 2.x releases, only 3.x.  There have been several 3.x releases.

andrewtyped:
I would like to understand better the release model for graphql-dotnet. I am not usually comfortable with adding pre-release dependencies to my production libraries, but this project goes through long phases of alpha dev (2.0 had ~1 year of alpha releases, 3.0 is at 7 months). It also seems alpha releases are no longer tagged or branched. What are the blockers preventing a stable release? I see a long tail of work under the 3.0 milestone, but also a long tail of open PR's not affiliated with any release. Do all these need closing before stability?

joemcbride:
> I would like to understand better the release model for graphql-dotnet.

There is no fixed release schedule.

> I am not usually comfortable with adding pre-release dependencies to my production libraries

Many people do not have that same issue and 3.x has been in production for several people for quite some time.

> What are the blockers preventing a stable release?

General documentation (both the text in the documentation itself, as well as the documentation site itself) and an upgrade guide.  Because there are breaking changes, those should be documented.

> I see a long tail of work under the 3.0 milestone, but also a long tail of open PR's not affiliated with any release. Do all these need closing before stability?

No, not all of these need closing.  There are some more breaking changes that I would _prefer_ to get into 3.0, though those could be put off.

Since this is an OSS project, people work on what they want.  That doesn't always align with what would be a targeted release.  If the docs were ready, I would be just fine releasing 3.0 today.

msheldon83:
I’d be happy to try to contribute to the docs or remaining enhancements.   Can you point me to some issues that I can help resolve?

I missed the preview releases as that’s typically not my standard either but I’ll switch over as I’m not in prod yet and in anticipation of the stable release.

Thanks

sungam3r:
@msheldon83  https://github.com/graphql-dotnet/graphql-dotnet/milestone/4

@joemcbride Regarding documentation. I have already made changes to the documentation pages that are located on the site (docs2 folder). Do I understand correctly that this site still needs to be somehow manually updated? Now there is documentation for version 2.4.0. Wouldn't it be better to switch to the built-in wiki infrastructure of the github? I think that wiki will be much more convenient.

joemcbride:
@msheldon83

[...]

Issue #1146, opened 2019-05-16
Changes to IFieldResolver


johnrutherford:
I've had some thoughts on changes to `IFieldResolver` and `ResolveFieldContext`.

Currently `IFieldResolver.Resolve()` returns an `object`, but in many cases this may be a `Task<T>`, so we need to check that.

I think it would make more sense to have it be async first and return a `Task` or `Task<T>`. The problem with `Task<T>` is that we still have to do tricks to get the value from the task after it has been awaited. So what if we instead added a `Result` property on the `ResolveFieldContext`? The `ResolveAsync` implementation must set the `Result` property before returning.

``` csharp
public interface IFieldResolver
{
    Task ResolveAsync(ResolveFieldContext context);
}
```

So, for example, the `ExpressionFieldResolver` would look like this:

``` csharp
public class ExpressionFieldResolver<TSourceType, TProperty> : IFieldResolver
{
    private readonly Func<TSourceType, TProperty> _property;

    public ExpressionFieldResolver(Expression<Func<TSourceType, TProperty>> property)
    {
        _property = property.Compile();
    }

    public Task ResolveAsync(ResolveFieldContext context)
    {
        context.Result = _property((TSourceType)context.Source);
        // We don't need to do Task.FromResult(), so there are fewer allocations
        return Task.CompletedTask;
    }
}
```

This would make it simpler and more performant to get the result from field resolution.

The FieldBuilder stuff would stay the same so you could still just return the value from your resolve function.  The `AsyncFieldResolver` would look like this:

``` csharp
public class AsyncFieldResolver<TSourceType, TReturnType> : IFieldResolver
{
    private readonly Func<ResolveFieldContext<TSourceType>, Task<TReturnType>> _resolver;

    public AsyncFieldResolver(Func<ResolveFieldContext<TSourceType>, Task<TReturnType>> resolver)
    {
        _resolver = resolver ?? throw new ArgumentNullException(nameof(resolver), "A resolver function must be specified");
    }

    public async Task ResolveAsync(ResolveFieldContext context)
    {
        context.Result = await _resolver(context.As<TSourceType>());
    }
}
```

sungam3r:
> Currently IFieldResolver.Resolve() returns an object, but in many cases this may be a Task<T>, so we need to check that.

[...]

PR #1380, opened 2019-10-31
Rename IFieldResolver.Resolve to ResolveAsync and return Task<object>

Shane32:
Based on the comment by @joemcbride that IValidationRule.Validate should become ValidateAsync and return a `Task<INodeVisitor>`, I suggest that IFieldResolver.Resolve should become ResolveAsync and return a `Task<object>`.  `IFieldResolver<T>.Resolve` returns a `Task<T>`.  This pull request represents the necessary changes.

Of note, field builders already have Resolve and ResolveAsync methods, so they shouldn't change.  However, many of the asynchronous samples in the tests and examples would build the asynchronous field with Field() rather than FieldAsync().  Using FieldAsync is now required for asynchronous fields.  This would be the biggest breaking change.

This cleans up some of the mess in ExecutionStrategy.ExecuteNodeAsync, which was relying on returned type inspection to determine if it needed to await on the return value or not, and then having to make a call to `((dynamic)task).Result` (within TaskExtensions.GetResult) to retrieve the value.

The internal SyncToAsyncResolverAdapter has been removed, as it serves no purpose.

The DelegateFieldModelBinderResolver still inspects the return type and awaits on it if it is a Task.  Personally, I would remove this class from the base GraphQL library, as it promotes non-type-safe coding, and any third party can reproduce this functionality extremely easily.  Perhaps if I knew how it was being used, I would have a different opinion, but in any case, it's still there.

Finally, the AccessorFieldResolver also still inspects the return type and awaits on it if it is a Task.  This seemed to make sense given the usage.

References: #1321 and #1140


Shane32:
One other note: these changes should have no effect on anyone using synchronous field resolvers using the basic Field() method or FieldBuilder.Resolve() methods, or on anyone using asynchronous field resolvers using the basic FieldAsync() method or FieldBuilder.ResolveAsync() methods.  It's only when (a) defining a custom IFieldResolver, or (b) returning an asynchronous result using the Field() or FieldBuilder.Resolve() methods.

sungam3r:
I looked at the changes, not in great detail, but the overall picture is clear. I can’t say that the code has become simpler. Also, I can’t say that it brings only positive changes. With these changes, you force all consumers (explicitly or implicitly) to use Task to wrap values. That is, **each field** will now be forced to be wrapped in the Task when the value is received from resolver. This is a definite hit on performance.

> Based on the comment by @joemcbride that IValidationRule.Validate should become ValidateAsync and return a Task, I suggest that IFieldResolver.Resolve should become ResolveAsync and return a Task. 

There is a big difference between the two examples. Field resolver is not required to return Task with some object. It can return both it and just some object (synchronous return). `ExecutionStrategy` checks it and do approprite await if needed:
```c#
 if (result is Task task)
                {
                    await task.ConfigureAwait(false);
                    result = task.GetResult();
                }
```

However, I agree that in general, you can get lost in the variety of possible signatures for determining the resolver for the field.

Relates to #1146 

Also I remember that I saw another issue about it would be nice to refactor 6 different ways of setting field, but I can not find it.

sungam3r:
> One other note: these changes should have no effect on anyone using synchronous field resolvers using the basic Field() method or FieldBuilder.Resolve() methods

I don't understand. How will this not affect them if another type of return is now required?

Shane32:
> I don't understand. How will this not affect them if another type of return is now required?

To recap, using Field() or FieldBuilder.Resolve() accepts a synchronous Func<> and wraps the resolve method parameter in a FuncFieldResolver, which handles wrapping the result into a Task<> via Task.FromResult.  (Implicit use of Task.FromResult.)  When using FieldAsync(), the resolve parameter is wrapped in an AsyncFieldResolver which passes the Task through.

So I guess I didn't mean to say that these changes wouldn't affect performance for synchronous fields, but rather that it would not typically be a breaking change in the way the fields are defined, nor in the ultimate execution of the query.

> This is a definite hit on performance.

I agree, there is a hit on performance.  However, .Net Core is optimized for this behavior.  Based on this website: https://github.com/dotnet/corefx/issues/30655 , when using ConfigureAwait(false), it is about 10 nanoseconds.  My own testing showed it to be about 16 nanoseconds, as opposed to 5 nanoseconds when using the prior code design (the 'is' operator is quite slow).  This means that for 10,000 executions of the resolve function, it's about 0.11ms slower.  Not really enough to quibble about.

Personally, I don't think it's a big deal either way.  If I was picking, although I normally go for every performance enhancement possible, in this case I would probably choose tightly bound interfaces over using 'is' and dynamically invoking members, as the performance impact is negligible.  But I only posted it because of the goals @joemcbride had for 3.0 in relation to IValidationRule, and felt this would be a change in the same vein.  If the change is desired, it should probably make it into 3.0, as it is a breaking change and would not make sense to introduce at a later time.

sungam3r:
Any way it needs to be rebased after #1441

sungam3r:
Is this still actual? To be honest, I don't see the benefit of renaming.

Shane32:
I would on principle, and I also thought it would match the change in the IValidationRule.  But I really don't think it matters too much, and we can just close it if you think it's better to leave as-is.

sungam3r:
My intuition says it's worth leaving everything as it is.

Shane32:
Ok with me.  It's merged if you want to review the diff again.

Shane32:
As all synchronous field resolvers are required to use Task.FromResult to return the value, there are additional objects created and disposed of, consisting of a slight performance decrease, as shown here:

```BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18362
.NET Core SDK=3.1.101

*** BEFORE ***

|        Method |        Mean |    Error |   StdDev |    Gen 0 |   Gen 1 | Gen 2 |  Allocated |
|-------------- |------------:|---------:|---------:|---------:|--------:|------:|-----------:|
| Introspection | 1,516.36 us | 9.253 us | 8.202 us | 201.1719 | 83.9844 |     - | 1050.88 KB |
|          Hero |    31.56 us | 0.266 us | 0.249 us |   6.2256 |       - |     - |   25.55 KB |

*** AFTER ***

|        Method |        Mean |     Error |    StdDev |    Gen 0 |    Gen 1 | Gen 2 |  Allocated |
|-------------- |------------:|----------:|----------:|---------:|---------:|------:|-----------:|
| Introspection | 1,635.75 us | 13.142 us | 12.293 us | 222.6563 | 109.3750 |     - | 1134.81 KB |
|          Hero |    32.10 us |  0.266 us |  0.249 us |   6.2866 |   0.0610 |     - |   25.77 KB |```

sungam3r:
8% memory growth is quite noticeable and I think it will be all the more noticeable the more leaf nodes in the schema for which the results are wrapped into a `Task`.

sungam3r:
Also #1146 should show the opposite effect - memory desrease. Then we can compare two extremes - when everything wrapped into a `Task` and when nothing wrapped.

Issue #1383, opened 2019-11-04
Type comparison appears to be wrong in InputObjectGraphType


Shane32:
InputObjectGraphType is defined as:

``` C#
    public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType>, IInputObjectGraphType
    {
        public override FieldType AddField(FieldType fieldType)
        {
            if (fieldType.Type == typeof(ObjectGraphType))
            {
                throw new ArgumentException("InputObjectGraphType cannot have fields containing a ObjectGraphType.", nameof(fieldType.Type));
            }

            return base.AddField(fieldType);
        }
    }
```

It would appear that this line of code:

``` C#
if (fieldType.Type == typeof(ObjectGraphType))
```

should have been written as:
``` C#
if ((fieldType.Type != null && typeof(ObjectGraphType).IsAssignableFrom(fieldType.Type)) || ((fieldType.ResolvedType as ObjectGraphType) != null))
```

Is that correct?  If so, I can submit a pull request

Shane32:
(I'm not exactly familiar with how input graphs deserialize input data into models.)

sungam3r:
The direction of thoughts is right. I have long wanted to add an implementation of [this algorithm](https://graphql.github.io/graphql-spec/June2018/#sec-Input-and-Output-Types) from the specification to check in/out types.

sungam3r:
Also note what `ObjectGraphType` is :
```c#
 public class ObjectGraphType : ObjectGraphType<object>
 {
 }
```
so `ObjectGraphType<T>` is more suitable for check against.

Shane32:
Good reference.  I was just thinking that my code would not perform sufficient checks for NonNullGraphType<>

joemcbride:
There is the following:

https://github.com/graphql-dotnet/graphql-dotnet/blob/fe0368f15dcea1dd578e83c521b985798e1a2214/src/GraphQL/GraphQLExtensions.cs#L44-L48

Though it only works with objects that have a `ResolvedType`.  It could be updated to also check the `Type`, if there is no `ResolvedType`.

I would suggest to use the interface `IObjectGraphType` vs. the specific type.

joemcbride:
There is a `GetNamedType`, which handled the `NonNullGraphType`, and `ListGraphType` which works with both `Type` and `ResolvedType` cases.

[...]

PR #1384, opened 2019-11-04
Update and disable field validation on input object graph types progress

Shane32:
Please see issue #1383 

Validation code added and disabled (along with associated tests) until the issue with the SchemaBuilder has been fixed.  I'm not sure what needs to be done from this point.

### DIFF
--- a/src/GraphQL/Types/InputObjectGraphType.cs
+++ b/src/GraphQL/Types/InputObjectGraphType.cs
@@ -16,10 +16,7 @@ namespace GraphQL.Types
         {
             if (!ValidateFieldType(fieldType))
             {
-                // todo: must fix errors in SchemaBuilder before uncommenting
-                // test named "should_throw_an_exception_if_input_object_graph_type_contains_object_graph_type_field" is disabled until this bug is fixed
-
-                // throw new ArgumentException("InputObjectGraphType cannot have fields containing a ObjectGraphType.", nameof(fieldType.Type));
+                throw new ArgumentException("InputObjectGraphType cannot have fields containing a ObjectGraphType.", nameof(fieldType.Type));
             }
 
             return base.AddField(fieldType);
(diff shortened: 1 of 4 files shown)


src/GraphQL/Types/InputObjectGraphType.cs
```diff
@@ -14,13 +14,23 @@ public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType>,
...
+
+                // throw new ArgumentException("InputObjectGraphType cannot have fields containing a ObjectGraphType.", nameof(fieldType.Type));
             }
 
             return base.AddField(fieldType);
         }
+
+        private static bool ValidateFieldType(FieldType fieldType)
+        {
+            if (fieldType.ResolvedType != null && fieldType.ResolvedType.IsInputType()) return true;
```
sungam3r:
```suggestion
            if (fieldType.ResolvedType?.IsInputType() == true) return true;
```

src/GraphQL/Types/InputObjectGraphType.cs
```diff
@@ -14,13 +14,23 @@ public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType>,
...
+            if (fieldType.Type != null && fieldType.Type.IsInputType()) return true;
```
sungam3r:
```suggestion
            if (fieldType.Type?.IsInputType() == true) return true;
```

Shane32:
Updated pursuant to pending comments in #1383

src/GraphQL/Types/InputObjectGraphType.cs
```diff
@@ -14,13 +14,40 @@ public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType>,
     {
         public override FieldType AddField(FieldType fieldType)
         {
-            if (fieldType.Type == typeof(ObjectGraphType))
+            if (!ValidateFieldType(fieldType))
             {
                 throw new ArgumentException("InputObjectGraphType cannot have fields containing a ObjectGraphType.", nameof(fieldType.Type));
```
sungam3r:
```suggestion
                 throw new ArgumentException("InputObjectGraphType should contain only fields of 'input types' - enumerations, scalars and other InputObjectGraphTypes", nameof(fieldType.Type));
```

This error message is more informative.

sungam3r:
Well, since we are adding such check, then let's make it complete - a similar check is necessary for the mirror case - when input field is added to the object type :).

sungam3r:
Closed in favor of #1436

PR #1385, opened 2019-11-05
Remove dead link

Toudahl:


Toudahl:
uhm, sorry - this was a mistake.

This: https://graphql-dotnet.github.io/docs/getting-started/databases
has a link to http://fiyazhasan.me/tag/graphql-dotnet/ which is dead.

When i clicked edit, i didnt notice that i was not taken to the same version.

PR #1388, opened 2019-11-06
increase Operation_Can_Be_Cancelled Delay

SimonCropp:
it was intermittently failing for me


sungam3r:
Well, tests with delays can be unpredictable in CI environment.

PR #1409, opened 2019-11-10
some minor tests added

mac2000:
Hello, while trying to figure out how everything works did tried to implement few tests based on coverage results

### DIFF
--- a/src/GraphQL.Tests/NameFieldResolverTests.cs
+++ b/src/GraphQL.Tests/NameFieldResolverTests.cs
@@ -16,6 +16,7 @@ namespace GraphQL.Tests
         [InlineData("naMe", "Anyone")]
         [InlineData("FullInfo", "Anyone 20")]
         [InlineData("fullInfo", "Anyone 20")]
+        [InlineData(null, null)]
         [InlineData("unknown", "", true)]
         [InlineData("FullInfoWithParam", "", true)]
         public void resolve_should_work_with_properties_and_methods(string name, object expected, bool throws = false)
--- a/src/GraphQL.Tests/TypeExtensionsTests.cs
+++ b/src/GraphQL.Tests/TypeExtensionsTests.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using Shouldly;
 using Xunit;
 
@@ -30,6 +31,12 @@ namespace GraphQL.Tests
         public void is_nullable_using_nullable_types(Type type)
         {
             TypeExtensions.IsNullable(type).ShouldBeTrue();
+        }
+
+        [Fact]
+        public void will_throw_on_unknown_list()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => TypeExtensions.GetGraphTypeFromType(typeof(List<>)));
         }
 
         private enum TestEnum { }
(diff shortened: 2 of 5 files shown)


SimonCropp:
@sungam3r LGTM

src/GraphQL.Tests/NameFieldResolverTests.cs
```diff
@@ -16,6 +16,7 @@ public class NameFieldResolverTests
         [InlineData("naMe", "Anyone")]
         [InlineData("FullInfo", "Anyone 20")]
         [InlineData("fullInfo", "Anyone 20")]
+        [InlineData(null, null)]
```
joemcbride: This looks like a case that should never happen.  You should never actually have a FieldAst with a null name.

mac2000: Understood, my intent was only to cover edge case, and to prove that it wont blowup with null ref exception, if you wish we can remove this

sungam3r: This test is OK.

src/GraphQL.Tests/TypeExtensionsTests.cs
```diff
@@ -32,6 +33,12 @@ public void is_nullable_using_nullable_types(Type type)
             TypeExtensions.IsNullable(type).ShouldBeTrue();
         }
 
+        [Fact]
+        public void will_throw_on_unknown_list()
```
sungam3r:
```suggestion
        public void will_throw_on_open_generic()
```

src/GraphQL.Tests/TypeExtensionsTests.cs
```diff
@@ -32,6 +33,12 @@ public void is_nullable_using_nullable_types(Type type)
             TypeExtensions.IsNullable(type).ShouldBeTrue();
         }
 
+        [Fact]
+        public void will_throw_on_open_generic()
```
sungam3r: And one more - please move this test to src\GraphQL.Tests\Utilities\TypeExtensionsTests.cs

sungam3r: You also mind to rename it to GetGraphTypeFromTypeTests.cs to avoid ambiguity.

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -120,5 +121,34 @@ public void entity_query()
                 _.ExpectedResult = expected;
             });
         }
+
+        [Fact]
+        public void input_types_and_types_without_key_directive_are_not_added_to_entities_union()
```
sungam3r: @joemcbride I haven’t had time yet to sort out federated schemas. Does this test seem right to you?

src/GraphQL.Tests/TypeExtensionsTests.cs
mac2000: I did move test and rename file, the only thing I did not touch all other tests, this file has half tests named in snake case and half in something between pascal and snake case

sungam3r: OK

src/GraphQL.Tests/TypeExtensionsTests.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
```
sungam3r:
```suggestion
```

mac2000: Oops, sorry for that, my bad

PR #1416, opened 2019-11-14
Obsolete SchemaExtensions.Execute

SimonCropp:
removes an unnecessarily `.GetAwaiter().GetResult();`

### DIFF
--- a/src/GraphQL/SchemaExtensions.cs
+++ b/src/GraphQL/SchemaExtensions.cs
@@ -7,6 +7,7 @@ namespace GraphQL
 {
     public static class SchemaExtensions
     {
+        [Obsolete(@"This action is async. Use ExecuteAsync. Will be removed in v4.")]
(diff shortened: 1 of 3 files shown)


src/GraphQL/SchemaExtensions.cs
```diff
@@ -7,6 +7,7 @@ namespace GraphQL
 {
     public static class SchemaExtensions
     {
+        [Obsolete(@"This action is async. Use ExecuteAsync. Will be removed in v4.")]
```
sungam3r: Let`s remove this method. Is there any point in trying to maintain  such "local" backward compatibility if backward compatibility has already been lost for a long time for version 3.0.0?

SimonCropp: done

joemcbride:
A note on this.  The official graphql.org docs uses this method, as well as our own docs.

https://github.com/graphql/graphql.github.io/pull/555

PR #1417, opened 2019-11-15
delete redundant docs dir

SimonCropp:


sungam3r:
@joemcbride Is it safe to remove this stuff?

PR #1423, opened 2019-11-17
Remove unused scalars from SDL

sungam3r:
According to the specification, it is not necessary to print unused types (in particular scalars).

Relates to graphql/graphql-spec#648


src/GraphQL/GraphQLExtensions.cs
```diff
@@ -491,5 +491,63 @@ public static IValue AstFromValue(this object value, ISchema schema, IGraphType
 
             throw new ExecutionError($"Cannot convert value to AST: {serialized}");
         }
+
+
+        public static bool References(this ISchema schema, string typeName)
```
sungam3r: I'm not sure about the name. This method works not only for scalars, but for all types in general. Theoretically, it is possible to make such a schema in which there will be a "independent" type.

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -13,27 +13,28 @@ public class SchemaPrinter : IDisposable
...
+        };
 
         public SchemaPrinter(
             ISchema schema,
             SchemaPrinterOptions options = null)
         {
             Schema = schema;
             Options = options ?? new SchemaPrinterOptions();
-
-            if (Options.CustomScalars?.Count > 0)
```
sungam3r: The name of this property did not match its use.

joemcbride: This option allows the user to add a custom scalar that will be treated as if it was a built-in scalar and never be printed.  Removing this removes that ability.

sungam3r: I guessed about it. The question is **why** is such a _feature_ needed at all? Custom scalar is the same part of the schema along with all its other parts. What is the point of not printing this scalar in SDL? This scalar is used somewhere in the schema - in field arguments, in field types, etc. The resulting schema will simply be an incorrect document.

joemcbride: The same reason why `String` is not printed.  If the type is considered part of your domain, there is no reason to print it.  That does not make it incorrect.

I could make the same argument you are about excluding scalar types from printing when they are not used.  They are defined in the Schema, you are just excluding them from printing because you don't want to see them.  Just because they are not used does not mean they don't exist.

Both of things are just features of the printer.

sungam3r: > The same reason why String is not printed.

This is a different situation. The `String` is a built-in scalar. The specification clearly defines the behavior when displaying this scalar in SDL:
> When representing a GraphQL schema using the type system definition language, the built‐in scalar types **should be** omitted for brevity.

Regarding unused scalars. Do I understand correctly that you have already read my questions in the spec repo - graphql/graphql-spec#648 ?

>  Just because they are not used does not mean they don't exist.

Nope. This just means that they do not exist. The specification says this, but not as explicitly as we would like. In fact, for a long time, just no one paid attention to these scalars in SDL. If we add support for new scalar types, it doesn’t mean that these scalars should be printed in SDL.

In the end, the application developer does not even know about these scalars. This is just some payload from us as library developers to help the application developer. We provide an **initial set** of supported (native for .NET) scalars, but these should by no means **be equated** with the use of these scalars. These are two completely different things. In the future, we can add 100 new scalars, but this should not change the application schema.

joemcbride: > This is a different situation.

No, "should be omitted for brevity" - is the exact reason why that feature is there.  So that is the exact same situation.

> Nope. This just means that they do not exist.

They do exist in the code, and the Schema configuration, they would just not being printed with this class.  They would still show up in introspection.

> Do I understand correctly that you have already read my questions in the spec repo

Yes.  Though adding this feature does not mean you have to remove a completely unrelated feature.

There are a ton of features that this codebase has that are not related to the spec at all.  This is one of them.  There is no reason to remove it as it is an optional feature and has its uses.

joemcbride: > They would still show up in introspection.

So in this regard, if that is really the end goal here, I would say this filtering should actually be in the Schema itself and not just this printer.

sungam3r: Indeed, introspection needs to be fixed.

sungam3r: By the way, I remembered that just in case I made the `SchemaPrinterOptions.IncludeAllTypes` property. But to be honest, I still don’t understand when it can be useful to form an SDL in which there will be custom scalars that are not actually used in the SDL.

joemcbride: > I still don’t understand when it can be useful to form an SDL in which there will be custom scalars that are not actually used in the SDL.

When me, as a human, wants to view the printed schema without all the extra scalars.

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -13,27 +13,28 @@ public class SchemaPrinter : IDisposable
...
+        private readonly List<string> _builtInScalars = new List<string>
+        {
+            "String",
+            "Boolean",
+            "Int",
+            "Float",
+            "ID"
+        };
+
+        private readonly List<string> _buildInDirectives = new List<string>
```
joemcbride:
```diff
- _buildInDirectives
+ _builtInDirectives
```

src/GraphQL/Utilities/SchemaPrinter.cs
joemcbride: Looking at this class as a whole again, it does appear there are enough hooks in place to accomplish not printing the scalars without this feature in the Options.  I'm okay with removing it.

I _do_ think that the filtering should actually be done on the Schema itself though and not in the printer.  That would give us some hooks to also support filtering the Schema based on Authentication/Authorization rules.

src/GraphQL.Tests/Introspection/IntrospectionResult.cs
```diff
@@ -32,196 +32,6 @@ public class IntrospectionResult
           ""enumValues"": null,
           ""possibleTypes"": null
         },
-        {
```
joemcbride: I think we need to keep a test (this one) that includes all scalars, and a new test that includes the filtered ones.

I'm also concerned about removing the built-in scalars from introspection before this is an official part of the spec.  Does the `graphql-js` project do that today?  Does GraphiQL and GraphQL Playground still work with these changes?

src/GraphQL/Utilities/SchemaPrinter.cs
sungam3r: Done for introspection.

> I do think that the filtering should actually be done on the Schema itself though and not in the printer.

Something like `ISchema.AllUsedTypes` ? I think it will look strange.

sungam3r: > When me, as a human, wants to view the printed schema without all the extra scalars.

So... It seemed to me that I was doing just that.

joemcbride: This feature will exclude _some_ scalars, but not _all_ scalars.

src/GraphQL.Tests/Introspection/IntrospectionResult.cs
sungam3r: https://github.com/graphql/graphql-js/issues/2005

src/GraphQL/Utilities/SchemaPrinter.cs
joemcbride: I would probably do the filtering in this property.

https://github.com/graphql-dotnet/graphql-dotnet/blob/424f59696fa82396ea0b74013804b4793aff8d56/src/GraphQL/Types/Schema.cs#L90

... and provide a way to control the filtering.  Right now there would be no way to turn off the reference check, and it is hard-coded in both the printer and introspection.  You can instead move the filtering to the Schema and not have to make changes to the printer nor the introspection types.

I added this `Filter` feature, but it is only used in Introspection today.

https://github.com/graphql-dotnet/graphql-dotnet/blob/424f59696fa82396ea0b74013804b4793aff8d56/src/GraphQL/Types/Schema.cs#L74

This idea could be applied more broadly to all of the types in the Schema.

src/GraphQL.Tests/Introspection/IntrospectionResult.cs
joemcbride: @sungam3r Can you summarize please?  I don't have time to do the whole investigation.

sungam3r: So far I have returned the builtin scalars to introspection. I will be back with an answer when I check graphq-js and GraphiQL.

sungam3r: > I'm also concerned about removing the built-in scalars from introspection before this is an official part of the spec. 

Now it is.

It looks like the spec has become clearer with respect to embedded scalars - graphql/graphql-spec#597 :

> When returning the set of types from the `__Schema` introspection type, all
referenced built-in scalars must be included. If a built-in scalar type is not
referenced anywhere in a schema (there is no field, argument, or input field of
that type) then it must not be included.

\+

> When representing a GraphQL schema using the type system definition language,
all built-in scalars must be omitted for brevity.

src/GraphQL/GraphQLExtensions.cs
sungam3r: Renamed to `Supports` due to spec: 
> A schema is defined in terms of the types and directives it supports as well as the root operation types for each kind of operation

src/GraphQL/Utilities/SchemaPrinter.cs
sungam3r: If I do filtering in `AllTypes` property, it will most likely lead to errors, because the property is used everywhere to get the **full set** of types. This property is for internal GraphQL server consumption.

sungam3r: One more thought about filtering. I understand that the user can filter parts of the schema. I do not consider this an unnecessary featurey. But I believe that it should be limited as much as possible so that as a result of this filtering a valid schema is obtained. If the specification says that the introspection request MUST return only used scalars, then only they should fall into the user filter, and not all scalars.

sungam3r:
PR is ready for review.

Summary changes after information from graphql/graphql-spec#597 :

1. Introspection response does not return ALL unused types (which essentially means `not supported` types): built-in scalars, custom scalars, graph types.
2. `SchemaPrinter` does the same as (1.)  + **also** does not print all built-in scalars at all.
3. `SchemaPrinter.IsDefinedType` can be overridden to allow printing more types. Example - `AllTypesSchemaPrinter` from tests.

These changes are in accordance with the specification. @joemcbride, can you do a review, please?

sungam3r:
This PR is no longer needed after rework of `GraphTypesLookup` in #1887 . However, there will be one small change, maybe later in separate PR:

> 2. SchemaPrinter does the same as (1.) + also does not print all built-in scalars at all.

sungam3r:
OK. As I expected, all scalars are already printed correctly thanks to our edits in other PRs.

sungam3r:
@Shane32 PR is ready for review.

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -16,10 +16,29 @@ namespace GraphQL
...
+            "deprecated"
+        };
+
+        internal static bool IsIntrospectionType(this IGraphType type) => type?.Name?.StartsWith("__", StringComparison.InvariantCulture) ?? false;
+
+        internal static bool IsIntrospectionType(this string typeName) => typeName.StartsWith("__", StringComparison.InvariantCulture);
+
+        internal static bool IsBuiltInScalar(this string typeName) => _builtInScalars.Contains(typeName);
+
+        internal static bool IsBuiltInDirective(this string directiveName) => _builtInDirectives.Contains(directiveName);
```
Shane32, line 41: We should not be adding extension methods to strings that relate to type names, even for internal use.  It just doesn't make sense.  For example, this statement makes no sense:

```cs
var ret = fieldType.Description.IsIntrospectionType();
```

Let's put those in a proper class (maybe this same one) so it needs to be fully qualified such as this:

```cs
var ret = GraphQL.IsIntrospectionType(typeName);
```

You can decide where/how, and if you really just want to leave it here, that's okay -- at least it's internal.

src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
```diff
@@ -98,7 +97,7 @@ public override string PrintFields(IComplexGraphType type)
         public string PrintFederatedSchema()
         {
             return PrintFilteredSchema(
-                directiveName => !IsSpecDirective(directiveName) && !IsFederatedDirective(directiveName),
+                directiveName => !directiveName.IsBuiltInDirective() && !IsFederatedDirective(directiveName),
```
Shane32, line 100: For example:

```suggestion
                directiveName => !IsBuiltInDirective(directiveName) && !IsFederatedDirective(directiveName),
```

src/GraphQL/GraphQLExtensions.cs
Shane32: If these are only used by the schema printer, I might just leave them there.

Shane32 (commented):
I only have one minor comment.  Please read and then merge at your discretion @sungam3r

src/GraphQL/GraphQLExtensions.cs
sungam3r: I doubted where it would be better to place these methods. As a result, I added them to `GraphQLExtensions`.

sungam3r: Done

Shane32 (approved):

PR #1427, opened 2019-11-21
Updated ConnectionBuilder to allow bidirectional slicing

kenewbie:
According to the relay specification "before" and "after" are allowed to be specified simultaneously when querying a connection. I need this in a project so I have included support for it by adding an optional parameter to .Bidirectional() which defaults to false to preserve backwards compatibility.

### DIFF
--- a/src/GraphQL/Builders/ConnectionBuilder.cs
+++ b/src/GraphQL/Builders/ConnectionBuilder.cs
@@ -36,6 +36,8 @@ namespace GraphQL.Builders
 
         private bool _isBidirectional;
 
+        private bool _allowBidirectionalSlicing;
+
         private int? _pageSize;
 
         public FieldType FieldType { get; protected set; }
@@ -48,6 +50,7 @@ namespace GraphQL.Builders
         {
             _isUnidirectional = isUnidirectional;
             _isBidirectional = isBidirectional;
+            _allowBidirectionalSlicing = false;
             _pageSize = pageSize;
             FieldType = fieldType;
         }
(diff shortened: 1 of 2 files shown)


RehanSaeed:
Since it's in the spec to allow both, I'd prefer a breaking change but that's just my 2 cents.

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -194,7 +199,7 @@ public void ResolveAsync(Func<ResolveConnectionContext<TSourceType>, Task<object
 
         private void CheckForErrors(ResolveConnectionContext<TSourceType> args)
         {
-            if (args.First.HasValue && args.Last.HasValue)
+            if (args.First.HasValue && args.Last.HasValue && !_allowBidirectionalSlicing)
```
SimonCropp, line 202: @sungam3r seems the spec allows this? so we should just remove this arg check?

SimonCropp:
@kenewbie can u add a test that verifies this change?

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: I must admit that I still can’t find the time to read the specification and check this PR. I will try to do this ASAP.

Shane32: https://relay.dev/graphql/connections.htm#sec-Pagination-algorithm

Shane32: > **Pagination Algorithm:**
>
> To determine what edges to return, the connection evaluates the before and after cursors to filter the edges, then evaluates first to slice the edges, then last to slice the edges.
>
> _Note: Including a value for both first and last is strongly discouraged, as it is likely to lead to confusing queries and results. The PageInfo section goes into more detail here._

Shane32: The actual implementation isn't within the GraphQL library.  If we remove the arg check (which really should have been an `ExecutionError`) it will be up to the users to write code that accurately returns results.  See `GraphQL.StarWars.Extensions.ResolveFieldContextExtensions.GetPagedResults(...)`.

The original suggestion here might be best: to only allow "bidirectional slicing" when configured to do so.

PR #1436, opened 2019-11-25
Prevent use ObjectGraphType as property of InputObjectGraphType and vice versa

### LINKED ISSUE #1434: Prevent use ObjectGraphType as property of InputObjectGraphType
Relates to #587

sungam3r:
fixes #1434

### DIFF
--- a/src/GraphQL/Types/InputObjectGraphType.cs
+++ b/src/GraphQL/Types/InputObjectGraphType.cs
@@ -1,5 +1,3 @@
-using System;
-
 namespace GraphQL.Types
 {
     public interface IInputObjectGraphType : IComplexGraphType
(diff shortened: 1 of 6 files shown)


src/GraphQL/Types/InputObjectGraphType.cs
```diff
@@ -12,15 +10,6 @@ public class InputObjectGraphType : InputObjectGraphType<object>
 
     public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType>, IInputObjectGraphType
     {
-        public override FieldType AddField(FieldType fieldType)
```
sungam3r: Since it was necessary to simultaneously check for input and output types, it was more convenient to transfer them to the base class.

sungam3r:
This test fails:
```c#
[Fact]
public void should_not_throw_an_exception_if_input_object_graph_type_doesnt_contains_object_graph_type_field()
{
    var type = new InputObjectGraphType();
    var exception = type.Field<ComplexGraphType<object>>().Name("test");
}
```

@joemcbride I doubt the correctness of this test. An abstract class is specified as the field type (not input, not output). Does that make sense?

src/GraphQL/Types/InputObjectGraphType.cs
sungam3r: In addition, this check did not take into account all cases.

joemcbride:
@sungam3r No that test as is doesn't make sense.  Could change that to `StringGraphType` or something like that.

PR #1438, opened 2019-11-27
[WIP] add editorconfig rules

### LINKED ISSUE #1361: Code Style Rules
There is already .editorconfig file in this repo. This issue is about defining a clear set of formatting rules (and there are a lot of available rules) and stick to it.

 As an example (starting point), we can take [this](https://github.com/DbUp/DbUp/blob/master/src/.editorconfig).

sungam3r:
fixes #1361

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,21 @@
 # EditorConfig is awesome: http://EditorConfig.org
+
+# Create portable, custom editor settings with EditorConfig
 # https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options
-# https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
+
+# .NET coding convention settings for EditorConfig
+# https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2019
+
+# Language conventions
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -17,3 +20,157 @@ indent_size = 4
...
+dotnet_code_quality_unused_parameters = all:warning
+
+# CSharp code style settings:
+[*.cs]
+# Prefer "var" everywhere
+csharp_style_var_for_built_in_types = false:suggestion
+csharp_style_var_when_type_is_apparent = true:error
+csharp_style_var_elsewhere = true:suggestion
+ 
+# Expression-bodied members
```
sungam3r: @joemcbride I think you can be against it

sungam3r:
Now I have 1083 warnings and 1033 suggestions. Many rules require discussion. This is a preview of the rules.

sungam3r:
I would really appreciate it if anyone took the time to discuss these rules. I looked through all the documentation and carefully approached each rule. Nevertheless, there are still controversial points and preferences. Of course, after applying this file, I saw hundreds of warnings and suggestions about changing the code. Something I like more, something less. But a common denominator formalized as automated rules is better than nothing.

sungam3r:
OK. I think that it's time to merge this PR. After that, there will probably be PRs aimed at maintaining these rules. This is the first approximation. We can decide that the rules are worth changing.

PR #1440, opened 2019-12-01
Swap inheritance of ResolveFieldContext and ResolveFieldContext<T>

### LINKED ISSUE #1437: Add IResolveFieldContext<T> and IResolveFieldContext classes
## Summary, Motivation & Benefits

Currently, `ResolveFieldContext<TSource>` does not inherit from `ResolveFieldContext`, but rather the reverse.  So custom methods cannot be configured to accept a `ResolveFieldContext` parameter, but rather must rely on generics in order to allow for a context with unknown source type to be passed to it.  I would like to suggest that `IResolveFieldContext` and `IResolveFieldContext<TSource>` interfaces be added, so that code that does not care about the typing of TSource can be written without requiring generic typing on the method.

An alternative route would be to change the inheritance tree, but C# does not allow hidden members to be overridden.  So I believe it would be best just to add interfaces so that methods parameters can be typed `IResolveFieldContext` or `IResolveFieldContext<TSource>` as desired.

Once interfaces are added, `GetArgument` and `HasArgument` can be written as extension functions, so that they can operate on any class that supports `IResolveFieldContext`, irrespective of TSource.

Another benefit is that `ResolveFieldContext.As` could be written to simply type-check the source property and then return an class that implements `IResolveFieldContext<T>` where the members reference the original `ResolveFieldContext`, with no need to copy all of the properties.  It would probably be faster overall, by some immeasurable amount.

I can post a pull request if desired.

### Basic example

Here is what the interfaces would look like:
``` C#
    {
        public string FieldName { get; }
        public Field FieldAst { get; }
        public IGraphType ReturnType { get; }
        public IObjectGraphType ParentType { get; }
        public Dictionary<string, object> Arguments { get; }
        public object RootValue { get; }
        public IDictionary<string, object> UserContext { get; }
        public object Source { get; }
        public ISchema Schema { get; }
        public Document Document { get; }
        public Operation Operation { get; }
        public Fragments Fragments { get; }
        public Variables Variables { get; }
        public CancellationToken CancellationToken { get; }
        public Metrics Metrics { get; }
        public ExecutionErrors Errors { get; }
        public IEnumerable<string> Path { get; }
        public IDictionary<string, Field> SubFields { get; }
    }

    public interface IResolveFieldContext<TSource> : IResolveFieldContext
    {
        public new TSource Source { get; }
    }
```

Here is a sample of how the As function could be written: (or it could be left as-is)

``` C#
    internal class ResolveFieldContextProxy<TSource> : IResolveFieldContext<TSource>
    {
        private readonly IResolveFieldContext _context;

        public ResolveFieldContextProxy(IResolveFieldContext context)
        {
            if (context == null) throw new ArgumentNullException(nameof(context));
            if (context.Source != null || !(context.Source is TSource))
                throw new InvalidOperationException("Source property is wrong type");
            _context = context;
        }

        public TSource Source => (TSource)_context.Source;
        public string FieldName => _context.FieldName;
        public Field FieldAst => _context.FieldAst;
        public FieldType FieldDefinition => _context.FieldDefinition;
        public IGraphType ReturnType => _context.ReturnType;
        public IObjectGraphType ParentType => _context.ParentType;
        public Dictionary<string, object> Arguments => _context.Arguments;
        public object RootValue => _context.RootValue;
        public IDictionary<string, object> UserContext => _context.UserContext;
        public ISchema Schema => _context.Schema;
        public Document Document => _context.Document;
        public Operation Operation => _context.Operation;
        public Fragments Fragments => _context.Fragments;
        public Variables Variables => _context.Variables;
        public CancellationToken CancellationToken => _context.CancellationToken;
        public Metrics Metrics => _context.Metrics;
        public ExecutionErrors Errors => _context.Errors;
        public IEnumerable<string> Path => _context.Path;
        public IDictionary<string, Field> SubFields => _context.SubFields;
        object IResolveFieldContext.Source => _context.Source;
    }
    public static class ResolveFieldContextExtensions
    {
        public static IResolveFieldContext<TSource> As<TSource>(this IResolveFieldContext context)
        {
            if (context is IResolveFieldContext<TSource> ret) return ret;
            return new ResolveFieldContextProxy<TSource>(context);
        }
    }
```

sungam3r:
fixes #1437

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
@@ -1932,12 +1932,7 @@ namespace GraphQL.Types
         public GraphQL.Types.QueryArgument Find(string name) { }
         public System.Collections.Generic.IEnumerator<GraphQL.Types.QueryArgument> GetEnumerator() { }
     }
-    public class ResolveFieldContext : GraphQL.Types.ResolveFieldContext<object>
-    {
-        public ResolveFieldContext() { }
-        public ResolveFieldContext(GraphQL.Execution.ExecutionContext context, GraphQL.Language.AST.Field field, GraphQL.Types.FieldType type, object source, GraphQL.Types.IObjectGraphType parentType, System.Collections.Generic.Dictionary<string, object> arguments, System.Collections.Generic.IEnumerable<string> path) { }
-    }
-    public class ResolveFieldContext<TSource> : GraphQL.Execution.IProvideUserContext
+    public class ResolveFieldContext : GraphQL.Execution.IProvideUserContext
     {
         public ResolveFieldContext() { }
         public ResolveFieldContext(GraphQL.Types.ResolveFieldContext context) { }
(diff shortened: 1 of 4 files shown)


sungam3r:
@Shane32 see if this PR solves your problem.

@joemcbride Do you mind if I remove this constructor?
https://github.com/graphql-dotnet/graphql-dotnet/blob/aed5298a435012d41d42dd0a9accebdb3cff1de3/src/GraphQL/Types/ResolveFieldContext.cs#L156
It is not used in the project.

sungam3r:
Closed in favor of #1441

PR #1441, opened 2019-12-01
Add IResolveFieldContext

### LINKED ISSUE #1437: Add IResolveFieldContext<T> and IResolveFieldContext classes

Shane32:
Fixes #1437 
See discussion at #1437 

This pull request includes #1440 

Note: removed these fields that were not used in the codebase:
- ResolveConnectionContext.NumberOfSkippedEntries
- ResolveConnectionContext.TotalCount
- ResolveConnectionContext.IsPartial

These can be restored by reverting commit 3a1b276 

Breaking changes are only that strongly typed resolver functions or subscription functions now need to be a `Func<IResolveFieldContext...` rather than `Func<ResolveFieldContext...`.  Functions for backwards compatibility can be added if desired.


src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -202,10 +202,6 @@ private void CheckForErrors(ResolveConnectionContext<TSourceType> args)
             {
                 throw new ArgumentException("Cannot use `last` with unidirectional connections.");
             }
-            if (args.IsPartial && args.NumberOfSkippedEntries.HasValue)
-            {
-                throw new ArgumentException("Cannot specify `numberOfSkippedEntries` with partial connection resolvers.");
-            }
```
sungam3r: Why?

src/GraphQL/Types/IResolveFieldContext.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public IEnumerable<string> Path { get; }
+
+        public IDictionary<string, Field> SubFields { get; }
+
+    }
+
+    public interface IResolveFieldContext<out TSource> : IResolveFieldContext
+    {
+        new public TSource Source { get; }
```
sungam3r: `public` access modifier in interface??

src/GraphQL/Types/ResolveFieldContext.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace GraphQL.Types
 {
-    public class ResolveFieldContext<TSource> : IProvideUserContext
+    public class ResolveFieldContext : IProvideUserContext, IResolveFieldContext
```
sungam3r:
```suggestion
    public class ResolveFieldContext : IResolveFieldContext, IProvideUserContext
```

src/GraphQL/Types/ResolveFieldContext.cs
```diff
@@ -79,42 +79,22 @@ public ResolveFieldContext(ResolveFieldContext context)
 
         public TType GetArgument<TType>(string name, TType defaultValue = default)
         {
-            return (TType)GetArgument(typeof(TType), name, defaultValue);
+            return ResolveFieldContextExtensions.GetArgument(this, name, defaultValue);
```
sungam3r: If three methods have become extension methods and backward compatibility has been already broken, then it makes no sense to leave these three methods here. Just remove.

Shane32: The methods are there to increase backwards compatibility.  Additional changes can ensure 100% compatibility.  However, I would agree to remove them, as 3.0 has not been released yet.

src/GraphQL/Types/ResolveFieldContext.cs
```diff
@@ -79,42 +79,22 @@ public ResolveFieldContext(ResolveFieldContext context)
...
-
-            return result == null && argumentType.IsValueType ? defaultValue : result;
+            return ResolveFieldContextExtensions.GetArgument(this, argumentType, name, defaultValue);
         }
 
-        public bool HasArgument(string argumentName) => Arguments?.ContainsKey(argumentName) ?? false;
+        public bool HasArgument(string argumentName) => ResolveFieldContextExtensions.HasArgument(this, argumentName);
 
-        public Task<object> TryAsyncResolve(Func<ResolveFieldContext<TSource>, Task<object>> resolve, Func<ExecutionErrors, Task<object>> error = null)
+        public Task<object> TryAsyncResolve(Func<ResolveFieldContext, Task<object>> resolve, Func<ExecutionErrors, Task<object>> error = null)
```
sungam3r: `TryAsyncResolve` and `TryAsyncResolve<TResult>` are used only in tests. Let's make them extension methods too. Then the `ResolveFieldContext` will become a pure state container.

sungam3r: And I think that `Func<ResolveFieldContext, Task<object>> resolve` should be changed to `Func<IResolveFieldContext, Task<object>> resolve`

src/GraphQL/Types/ResolveFieldContext.cs
```diff
@@ -139,21 +119,25 @@ public async Task<TResult> TryAsyncResolve<TResult>(Func<ResolveFieldContext<TSo
...
         {
-            if (this is ResolveFieldContext<TSourceType> typedContext)
-                return typedContext;
+        }
 
-            return new ResolveFieldContext<TSourceType>(this);
+        public ResolveFieldContext(IResolveFieldContext context) : base(context)
+        {
+            if (context.Source != null && !(context.Source is TSource))
+                throw new ArgumentException($"ResolveFieldContext.Source must be an instance of type '{typeof(TSource).Name}'", nameof(context));
```
sungam3r:
```suggestion
                throw new ArgumentException($"IResolveFieldContext.Source must be an instance of type '{typeof(TSource).Name}'", nameof(context));
```

src/GraphQL/Types/ResolveFieldContext.cs
sungam3r: > The methods are there to increase backwards compatibility. 

Prior to that, there were dozens of changes that completely killed backward compatibility.

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: It will always be false, as IsPartial and NumberOfSkippedEntries is never initialized before this code is executed.  Therefore, the code has no purpose.

Shane32:
I will push with all your suggested changes and undo the removal of the extraneous properties.  I can post a separate pull request for that.

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: Let's consider this change in another PR.

sungam3r:
Why separate?

Shane32:
I'm sorry; I thought you asked that we discuss removal of those properties separately from this PR.

Shane32:
I will leave it in this PR then.

sungam3r:
>  I can post a separate pull request for **that**.

For some reason I thought that we were talking about a completely separate PR for all your changes :).

Shane32:
The removal of those properties is still in the PR as it is now.  Let me know if you or @joemcbride would like me to revert the removal of those properties (commit 3a1b276 ) and post the removal of those properties as a separate PR.

src/GraphQL/Types/IResolveFieldContext.cs
Shane32: Removed unnecessary modifier.

sungam3r:
Once again to be exact - removal of the extraneous properties should (if any) go in separate PR, all fixes regarding comments - in that PR. Sorry for misleading.

src/GraphQL/Types/IResolveFieldContext.cs
sungam3r: Did it compile?

Shane32:
Ok not sure which empty line, but I cleaned up the file in any case.  Everything compiles and all tests pass on my machine.

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: I'll post a PR regarding removal of these properties after this one has been accepted, as it will need to be removed from the new interfaces also.

src/GraphQL/Types/ResolveFieldContext.cs
Shane32: Done

src/GraphQL/Types/ResolveFieldContext.cs
Shane32: Done

src/GraphQL/Types/ResolveFieldContext.cs
Shane32: Done

src/GraphQL/Types/ResolveFieldContext.cs
Shane32: Done

Shane32:
Let me know if there is anything else you notice in the code - thanks!

src/GraphQL/Subscription/IResolveEventStreamContext.cs
```diff
@@ -0,0 +1,17 @@
+using GraphQL.Types;
+using System;
+using System.Collections.Generic;
+using System.Text;
```
sungam3r: Unused usings

src/GraphQL/Types/IResolveFieldContext.cs
```diff
@@ -0,0 +1,54 @@
...
+        CancellationToken CancellationToken { get; }
+
+        Metrics Metrics { get; }
+
+        ExecutionErrors Errors { get; }
+
+        IEnumerable<string> Path { get; }
+
+        IDictionary<string, Field> SubFields { get; }
+
```
sungam3r: remove empty line please

src/GraphQL/Types/IResolveFieldContext.cs
```diff
@@ -0,0 +1,54 @@
+using GraphQL.Execution;
+using GraphQL.Instrumentation;
+using GraphQL.Language.AST;
+using System.Collections.Generic;
+using System.Threading;
+using Field = GraphQL.Language.AST.Field;
+
+namespace GraphQL.Types
+{
+    public interface IResolveFieldContext : IProvideUserContext
```
sungam3r: I like that now all properties are read-only. Of course, someone could set them and his code will stop compiling, but this is not scary, since one can cast to class.

src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
```diff
@@ -35,9 +35,9 @@ public static IFieldMiddlewareBuilder Use(this IFieldMiddlewareBuilder builder,
                 }
 
                 var parameters = methodInfo.GetParameters();
-                if (parameters.Length != 2 || parameters[0].ParameterType != typeof(ResolveFieldContext) || parameters[1].ParameterType != typeof(FieldMiddlewareDelegate))
+                if (parameters.Length != 2 || parameters[0].ParameterType != typeof(IResolveFieldContext) || parameters[1].ParameterType != typeof(FieldMiddlewareDelegate))
```
sungam3r: Most of the time I spent thinking about whether it is right to put the inequality operator here - see changed condition in `ReflectionHelper.BuildArguments`. In principle, it’s correct, but some doubts remain.

Shane32: I'd leave it as-is for this PR.  It could be "enhanced" to match on a typed IResolveFieldContext and then only run the middleware if the context type matches.  I probably wouldn't do that, though.

src/GraphQL/Types/IResolveFieldContext.cs
Shane32: Agreed

src/GraphQL/Types/IResolveFieldContext.cs
Shane32: Will do

src/GraphQL/Subscription/IResolveEventStreamContext.cs
Shane32: Removed.

Shane32:
@joemcbride Where are we at on this?  Is there any questions I can answer?  Any suggestions?

sungam3r:
@Shane32 I looked through all the changes. From my point of view, there are no obvious flaws so waiting for comments from @joemcbride .

sungam3r:
@Shane32 One more change - we should check docs and change `ResolveFieldContext` to `IResolveFieldContext`. For example, here: https://graphql-dotnet.github.io/docs/getting-started/field-middleware or there https://graphql-dotnet.github.io/docs/getting-started/introduction#schema-first-nested-types

Shane32:
Ok I’ll change that tomorrow

sungam3r:
I think @joemcbride has no objections about merging this PR. As a rule, if something goes against the design of the project, then he gives his comments about it.

sungam3r:
So do I merge it? We believe that everything is ready and checked several times?

Shane32:
I just looked through it again.  All tests pass.  I believe we're all good.

benmccallum:
Hey @Shane32, just going through my codebase and adapting to this change. In general it's pretty sweet and cleaning things up. One question I have though. Any reason why the `IResolveFieldContext` interface doesn't support setters on most of the properties? I was doing a set of `ReturnType` when resolving an interface type, and now I can't do that unless I cast it to a `ResolveFieldContext`.

benmccallum:
Though perhaps I should just use `ObjectGraphType.ResolveType` and I was doing it wrong anyway :P

Shane32:
You’re correct in that there are no setters. This change is part of what gives this design such a significant memory decrease and speed increase. You can call `new ResolveFieldContext(context)` for a mutable copy. As you’ve probably realized by now, the underlying type is ReadonlyResolveFieldContext or ResolveFieldContextAdapter depending on the field resolver.

benmccallum:
@Shane32, I think I was probably doing what I was trying to achieve wrongly anyway, so I've adjusted and will be testing soon. Thanks for your response. Good to know there's a good reason why :)

sungam3r:
Besides the optimization argument, the interface without setters seemed to me cleaner and more suitable for client code.

PR #1442, opened 2019-12-02
Change middleware Use to type-check and compile ahead of time

Shane32:
Simple performance enhancement to existing code.

### DIFF
--- a/src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
+++ b/src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
@@ -11,7 +11,7 @@ namespace GraphQL.Instrumentation
     {
         private const string InvokeMethodName = "Resolve";
 
-        public static IFieldMiddlewareBuilder Use<T>(this IFieldMiddlewareBuilder builder) => Use(builder, typeof(T));
+        public static IFieldMiddlewareBuilder Use<T>(this IFieldMiddlewareBuilder builder) where T : new() => Use(builder, typeof(T));
 
         public static IFieldMiddlewareBuilder Use(this IFieldMiddlewareBuilder builder, System.Type middleware)
         {
--- a/src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
+++ b/src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
@@ -1,6 +1,7 @@
 using GraphQL.Types;
 using System;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


Shane32:
I would suggest replacing this code with an IFieldMiddleware interface rather than dynamic binding.  See #1443

sungam3r:
Just for information - do you know that this code is usually executed only once when the `Schema` is initialized?

Shane32:
I figured as much, but wasn’t sure. I did think that this way, any compilation error would be raised when Use was called rather than later on — e.g. when the schema is initialized.

sungam3r:
> I did think that this way, any compilation error would be raised when Use was called rather than later on — e.g. when the schema is initialized.

In principle, this is better.

src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
```diff
@@ -10,39 +11,50 @@ public static class FieldResolverBuilderExtensions
     {
         private const string InvokeMethodName = "Resolve";
 
-        public static IFieldMiddlewareBuilder Use<T>(this IFieldMiddlewareBuilder builder) => Use(builder, typeof(T));
+        public static IFieldMiddlewareBuilder Use<T>(this IFieldMiddlewareBuilder builder) where T : new() => Use(builder, typeof(T));
```
sungam3r: > where T : new()

nice catch

sungam3r:
@Shane32 We also can think about what to do if middleware implements `IDisposable`. Since we ourselves create middleware, we must also destroy it, checking for `IDisposable` support, but this is certainly not the concern of this PR.

sungam3r:
> Simple performance enhancement to existing code.

And strictly speaking, these are not performance enhancements. If you specifically conduct a test, I’m sure that it will show a slowdown, since the method is actually called once, and the cost of dynamic compilation is high. Dynamic compilation gives a performance gain only when the additional cost of it is negligible compared to the cost of repeatedly calling the method through reflection.

But still it seems to me that this is a step in the right direction due to the argument already mentioned above in favor of early detection of errors.

Shane32:
I just realized something: Assuming there the builder code only ran once as you say (during schema initialization), then there is a significant difference between the new and old code -- the new code creates a new instance of the type every time the middleware executes, whereas the old code creates an instance once and calls the Resolve function repeatedly every time the middleware executes.

I didn't realize this when I wrote the code.  Should the code be modified to create the instance once, when the schema is initialized?

sungam3r:
Wow. You messed up something. Check the code again. Both options create a middleware once, just the second one activates it delayed, at the execution stage, and not the installation stage. As you said before.

Shane32:
The compiled expression is equivalent to:
``` C#
//func = (context, next) => (new <middleware>()).Resolve(context, next);
```
So within this:
``` C#
            return builder.Use(next =>
            {
                return context => func(context, next);
            });
```
We can see that every time the lambda is executed, an instance of the middleware will be created.

The behavior, if desired to be fixed, is fixed in #1463

sungam3r:
I'm surprised that I could be so wrong

PR #1443, opened 2019-12-02
Change middleware to use new IFieldMiddleware interface

Shane32:
Rather than dynamically binding middleware, this PR adds an interface which is used for binding middleware.

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
@@ -697,8 +697,9 @@ namespace GraphQL.Instrumentation
     public delegate System.Threading.Tasks.Task<object> FieldMiddlewareDelegate(GraphQL.Types.ResolveFieldContext context);
     public static class FieldResolverBuilderExtensions
     {
-        public static GraphQL.Instrumentation.IFieldMiddlewareBuilder Use<T>(this GraphQL.Instrumentation.IFieldMiddlewareBuilder builder) { }
-        public static GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(this GraphQL.Instrumentation.IFieldMiddlewareBuilder builder, System.Type middleware) { }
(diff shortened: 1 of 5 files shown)


sungam3r:
@Shane32 Please rebase this PR on master since #1442 was merged.

sungam3r:
Also this is a breaking change for existing middlewares forcing them to implement `IFieldMiddleware`. I understand that we have already done many backward incompatible changes, but this is not the situation. I see here a complete analogy with ASP.NET Core middleware api. See this https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Abstractions/src/Extensions/UseMiddlewareExtensions.cs  Pay attention to how `IMiddleware` is used. Your suggestion about the new interface are legal, but it should be an additional option that does not break the existing api.

sungam3r:
Yes, and as you can see, MS middleware code supports integration with a DI container, which is not supported now in this project. This is another area for improvements.

Shane32:
Just a thought: If the middleware code supported integration with a DI container, then the DI container would be responsible for disposing of the object.

Shane32:
I read up on the ASP.NET Core middleware that you mentioned.  It seems that the only reason for the middleware to not be an interface is so that per-request scoped DI resources can be injected into the Invoke method of the middleware during execution (a good reason).

sungam3r:
You can see the pros and cons of both approaches. The main thing to know is that there are such approaches.

sungam3r:
Did you decide to completely stop the work on this PR?

Shane32:
Yes.  If you feel there's still benefit to this branch, I can reopen it.  Seemed like the consensus was that the middleware builder could use an overhaul, but I really don't have the time to dig down into it.  I expect that once scoped DI support is added to the project, it will be a good time to take a look at what improvements can be made here.

Shane32:
Or if you suggest I make the IFieldMiddleware interface an additional option to the existing builder (rather than replacing existing code), I can do that easily.

sungam3r:
In general, I think we should not touch this place. This will not bring tangible benefits. We can focus on other things.

PR #1446, opened 2019-12-04
Check for redefinitions in SDL

sungam3r:
Relates to graphql/graphql-spec#653

`SchemaBuilder` should fail to build incorrectly typed SDL. Otherwise, one could came across the problem of non-deterministic behavior. When constructing the schema, a second instance of the type (some input type in my case) was created and replaced the reference to this type already added in the type dictionary. This led to the fact that, depending on the type arrangement in the SDL, the `GraphTypesLookup.ApplyTypeReferences` method did not completely replace `GraphQLTypeReference`.  And this, in turn, led to runtime errors:
```
System.InvalidCastException: 'Unable to cast object of type 'GraphQL.Types.GraphQLTypeReference' to type 'GraphQL.Types.ScalarGraphType'.'
```
in `GraphQLExtensions.IsValidLiteralValue`.


sungam3r:
@joemcbride Could you please also look at graphql-dotnet/parser#43 ? I need these changes to simplify check.

sungam3r:
Thanks.

Issue #1148, opened 2019-05-16
Use System.Text.Json for serialization/deserialization


johnrutherford:
https://www.nuget.org/packages/System.Text.Json

Remove JSON.NET dependency. Make JSON.NET support a separate, optional package.

This is the direction that ASP.NET Core is going for performance so I thought I'd throw it out there. Though I don't know when System.Text.Json will be ready for the 3.0 time frame.

sungam3r:
Yes, the thought is good. Now there is a broad tendency to take away this package.

DaniilSokolyuk:
Blocked by https://github.com/dotnet/corefx/issues/39808
Dictionary null values sterilized as empty objects in preview7

MerickOWA:
This appears to be fixed in core 3.1, any chance this can move forward now?

sungam3r:
@joemcbride I propose to consider the option of taking serialization to abstraction, something like `interface ISerializer`. The graphql-dotnet package will no longer depend on Newtonsoft.Json. We will have to make 2 additional packages - like `GraphQL.Serialization.NewtonsoftJson` and `GraphQL.Serialization.SystemTextJson`. And the developer will have to install one or another package. This is a common practice of transferring things that are not specific to the main package to external packages. In ASP.NET Core this is done in a similar way - https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.NewtonsoftJson

So GraphQL will get rid of the only external dependency and all the problems associated with managing its versions. It will be possible to update specific serialization packages independently.

Generally speaking, the options are not limited to the two serializers mentioned above. For example, there is a [Jil](https://github.com/kevin-montrose/Jil).

sungam3r:
Also, move of the serializer in a separate package will allow us not to resort to multi-targeting in GraphQL-dotnet.

joemcbride:
@sungam3r Agreed, I think that is a good approach.  It would be good to see if we could somehow make it generic enough that we could still have the main logic in the core project but just offload the serialization part.  That could get too far into the weeds, but could save some duplication.

MerickOWA:
I opened a pull request to try and help move this along.  I extracted out the Newtonsoft specific APIs into the GraphML.Serialization.Newtonsoft.  I was able to get all the unit tests to pass once I added the reference to the new library and added a hack for the missing newtonsoft attributes on the ExecutionResult class.

MerickOWA:
[...]

Issue #1322, opened 2019-10-01
HttpResponseStreamWriter presence


sungam3r:
This comes from #1318 

@joemcbride I see that `HttpResponseStreamWriter` class was copied from [here](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webutilities.httpresponsestreamwriter?view=aspnetcore-2.2). I do not know if there are any changes in the copied version. It seems to me that no. But in general it seems strange to see the mention of HTTP in this library, which is independent of any transport. I understand that the class has an auxiliary role, in fact, there is nothing specific for HTTP but only naming. Maybe rename it? I understand that it would be foolish to make dependency from one of ASP.NET Core assemblies. Copying a single small auxiliary class is certainly better.

I also discovered `namespace GraphQL.Http`:
\src\GraphQL\Http\DocumentWriter.cs
\src\GraphQL\Http\HttpResponseStreamWriter.cs
\src\GraphQL\Http\JsonArrayPool.cs

I suggest changing this namespace to `GraphQL` so as not to mislead anyone. The listed classes do not have any specific HTTP logic.

joemcbride:
I agree with the `Http` verbiage, that could be removed.

sungam3r:
**[reminder]** : close this issue after merging #1448

sungam3r:
Resolved in #1448

PR #1448, opened 2019-12-05
Extraction of Newtonsoft dependent APIs into separate library

MerickOWA:
As part of trying to help with Issue #1148, I moved everything having to do with Newtonsoft into a separate library "GraphQL.Serialization.Newtonsoft"

I tried to leave all the APIs in the same namespace as they were originally, just in the new library.

Of note however, is that internal HasCodes and Code properties on an ExecutionError got moved into the Newtonsoft library since it was only used inside the ExecutionResultJsonConverter

I created the NewtonsoftContractResolver to try and allow users passing in a newtonsoft contract resolver to use the user's resolver, but to set the converter to use to be the "default" ExecutionResultJsonConverter if no default is set.  I wasn't sure of a better way to do this since we can't put Newtonsoft attributes on the GraphQL classes now.

Still TODO, ApolloTrace has lost its Newtonsoft settings for using camelcase, and I haven't figured out a way yet to replace it.  Possibly just need more hacks inside the NewtonsoftContractResolver.


MerickOWA:
Should I just fix the PublicApi test change for this to be merged? or should the changes be reviewed first?

sungam3r:
> Should I just fix the PublicApi test change for this to be merged? or should the changes be reviewed first?

Yes. You can be sure that your changes will be closely monitored before merge.

>  ApolloTrace has lost its Newtonsoft settings for using camelcase, and I haven't figured out a way yet to replace it. Possibly just need more hacks inside the NewtonsoftContractResolver.

Yes, just a little more configuration inside NewtonsoftContractResolver.

sungam3r:
@MerickOWA Thank you for taking the initiative. As always, there is simply not enough time for all tasks.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
```diff
@@ -0,0 +1,49 @@
+using System.Buffers;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+
+namespace GraphQL.Http
+{
+    public class DocumentWriter : IDocumentWriter
```
sungam3r: Please look at #1322

src/GraphQL.sln
```diff
@@ -64,6 +64,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Docs", "Docs", "{EB92DFCF-A
 		..\docs2\site\docs\getting-started\variables.md = ..\docs2\site\docs\getting-started\variables.md
 	EndProjectSection
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GraphQL.Serialization.Newtonsoft", "GraphQL.Serialization.Newtonsoft\GraphQL.Serialization.Newtonsoft.csproj", "{A8905FD0-EE1C-4847-AFA6-B9EB590FEE79}"
```
sungam3r: In the future this project should be moved to a separate repository.

joemcbride: I think it should just stay in this one.  I've been thinking about also moving the server project into the main repo so there isn't as much churn between the two.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
joemcbride: Adding the comment here too, I like the removal of `Http` since it isn't really Http related.

src/GraphQL.sln
sungam3r: The advantages of this approach are simplicity. The drawbacks are in spurious changes, when new versions of packages will be released in which nothing was actually changed. In fact, this is already happening with Playground/Graphiql/Voyager in the server project now.

sungam3r:
@MerickOWA Your PR does not contain the most important element of the puzzle - serialization abstraction.

src/GraphQL.Serialization.Newtonsoft/ExecutionErrorExtensions.cs
```diff
@@ -0,0 +1,24 @@
+using System.Collections.Generic;
+
+namespace GraphQL
+{
+    internal static class ExecutionErrorExtensions
```
sungam3r: What is the reason for moving this code from `ExecutionError`?

sungam3r: Ok. I got it. Just make these properties public in original class and then ignore them in the created contract resolver.

src/GraphQL.Serialization.Newtonsoft/ExecutionResultJsonConverter.cs
```diff
@@ -99,11 +99,11 @@ private void WriteErrorExtensions(ExecutionError error, JsonWriter writer, JsonS
                 serializer.Serialize(writer, error.Code);
             }
 
-            if (error.HasCodes)
+            if (error.HasCodes())
```
sungam3r: No need to change - just make these properties public in original class and then ignore them in the created contract resolver.

src/GraphQL.Serialization.Newtonsoft/GraphQL.Serialization.Newtonsoft.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <RootNamespace>GraphQL</RootNamespace>
```
sungam3r: I think that it will be better to use different namespace - GraphQL.Serialization.Newtonsoft.

src/GraphQL.sln
joemcbride: Understood.  I think there has currently been not enough "compatibility" between this core project and the server project.  Things that don't need to change together, such as the Playground etc. make sense to be independent.  If we can get proper abstractions for the serialization, I could see those being in different repos.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
```diff
@@ -0,0 +1,49 @@
...
+                indent ? Formatting.Indented : Formatting.None,
+                new JsonSerializerSettings())
+        {
+        }
+
+        public DocumentWriter(Formatting formatting, JsonSerializerSettings settings)
+        {
+            _serializer = JsonSerializer.CreateDefault(settings);
+            _serializer.Formatting = formatting;
+            _serializer.ContractResolver = new NewtonsoftContractResolver(_serializer.ContractResolver);
```
sungam3r:
```c#
_serializer = JsonSerializer.CreateDefault(settings);
_serializer.Formatting = formatting;
_serializer.ContractResolver = new NewtonsoftContractResolver(_serializer.ContractResolver);
```
->
```c#
if (_settings.ContractResolver == null)
   _settings.ContractResolver =  new NewtonsoftContractResolver()
_serializer = JsonSerializer.CreateDefault(settings);
 _serializer.Formatting = formatting;
```

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public DocumentWriter(bool indent)
+            : this(
+                indent ? Formatting.Indented : Formatting.None,
+                new JsonSerializerSettings())
+        {
+        }
+
+        public DocumentWriter(Formatting formatting, JsonSerializerSettings settings)
```
sungam3r: Also `Formatting formatting` parameter is not needed, just set `JsonSerializerSettings.Formatting` property in the second ctor.

src/GraphQL.Serialization.Newtonsoft/Http/NewtonsoftContractResolver.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace GraphQL.Http
+{
+    internal class NewtonsoftContractResolver : DefaultContractResolver
+    {
+        private readonly IContractResolver _contractResolver;
```
sungam3r: No need in inner resolver, see other comments.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
sungam3r: By the way, this way we allow the client to use their own serialization format.

src/GraphQL/Execution/ExecutionError.cs
```diff
@@ -98,7 +78,7 @@ private void SetData(IDictionary dict)
             }
         }
 
-        private static string GetErrorCode(Exception exception) => _exceptionErrorCodes.GetOrAdd(exception.GetType(), NormalizeErrorCode);
+        public static string GetErrorCode(Exception exception) => _exceptionErrorCodes.GetOrAdd(exception.GetType(), NormalizeErrorCode);
```
sungam3r: No need to change this line, only `HasCodes` / `Codes` access modifiers to public.

src/GraphQL/Http/DocumentWriter.cs
```diff
@@ -11,48 +10,10 @@ public interface IDocumentWriter
         Task WriteAsync<T>(Stream stream, T value);
     }
 
-    public class DocumentWriter : IDocumentWriter
+    public static class DocumentWriterExtensions
```
sungam3r: Please also change file name and namespace.

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -1,11 +1,8 @@
 using System;
 using System.Collections.Generic;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Serialization;
 
 namespace GraphQL.Instrumentation
 {
-    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
```
sungam3r: Contract for `ApolloTrace` should be configured in the custom contract resolver along with contract for `ExecutionResult`.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            _serializer = JsonSerializer.CreateDefault(settings);
+            _serializer.Formatting = formatting;
+            _serializer.ContractResolver = new NewtonsoftContractResolver(_serializer.ContractResolver);
+        }
+
+        public async Task WriteAsync<T>(Stream stream, T value)
+        {
+            using (var writer = new HttpResponseStreamWriter(stream, Utf8Encoding))
```
joemcbride: I think both the `JsonArrayPool` and `HttpResponseStreamWriter` classes need to also be moved to this library.

src/GraphQL.sln
sungam3r: Of course, let it stay in this project for now. I think that by the time of version 3 release we can also reorganize the structure of the repositories, not earlier.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
sungam3r: Agree

src/GraphQL/Execution/ExecutionError.cs
MerickOWA: If we change those to public, it would change the Serialization of the class.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
MerickOWA: I see your point.  I didn't want to modify the incoming settings object as it could be a static/shared object used multiple times in various places, and we shouldn't modify it.

src/GraphQL.Serialization.Newtonsoft/GraphQL.Serialization.Newtonsoft.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <RootNamespace>GraphQL.Serialization.Newtonsoft</RootNamespace>
```
sungam3r: @joemcbride  GraphQL.Serialization.Newtonsoft or GraphQL.Serialization.NewtonsoftJson ?

MerickOWA:
> @MerickOWA Your PR does not contain the most important element of the puzzle - serialization abstraction.

Currently most of the abstraction is already in place with IDocumentWriter.  The only reading is done in the extension method StringExtensions.ToDictionary which i've moved out into GraphML.Serialization.Newtonsoft.

I assume you'd want something like IDocumentWriter to be changed to ISerializer with the following interface:
```
    public interface ISerializer
    {
        Task WriteAsync<T>(Stream stream, T value);
        Task<Dictionary<string, object>> ReadInputsAsync<T>(Stream stream);
    }
```
?

src/GraphQL.Serialization.Newtonsoft/SchemaExtensions.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Threading.Tasks;
+using GraphQL.Types;
+
+namespace GraphQL.Serialization.Newtonsoft
+{
+    public static class SchemaExtensions
+    {
+        public static Task<string> ExecuteAsync(this ISchema schema, Action<ExecutionOptions> configure)
```
sungam3r: I do not think this method will have consumers.

src/GraphQL/Execution/ExecutionError.cs
sungam3r: Ignore these properties in your custom contract resolver. Use `ShouldSerialize`.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
sungam3r: Your comment makes sense. Then we just need to remove the `formatting` parameter.

src/GraphQL.Serialization.Newtonsoft/SchemaExtensions.cs
MerickOWA: Its used in unit tests and is part of the public API currently.  If users upgrade to 3.0 and add a reference to GraphQL.Serialization.Newtonsoft, their code can at least keep working as before.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
MerickOWA: This should now be moved.

src/GraphQL.Serialization.Newtonsoft/GraphQL.Serialization.Newtonsoft.csproj
MerickOWA: Done.

sungam3r:
> @sungam3r Currently most of the abstraction is already in place with IDocumentWriter. The only reading is done in the extension method StringExtensions.ToDictionary which i've moved out into GraphML.Serialization.Newtonsoft.

I was wrong.

> I assume you'd want something like IDocumentWriter to be changed to ISerializer with the following interface:

No, nothing required.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
MerickOWA: The formatting parameter was there before, its used when the indent option is passed in, and could be used by other users.  If we modify the settings passed in, it could inadvertently affect users of the API if their settings are shared in other places of their application.

src/GraphQL.Serialization.Newtonsoft/Http/NewtonsoftContractResolver.cs
MerickOWA: Removed

src/GraphQL.Serialization.Newtonsoft/SchemaExtensions.cs
joemcbride: This method is specifically for new users "getting started" with the library, so they don't have to do a ton of setup just to run a query and see the results.

joemcbride: This method is specifically for new users "getting started" with the library, so they don't have to do a ton of setup just to run a query and see the results.

sungam3r: This method has little value for the calling code. It can be moved to test project.
Yes, it is a part of oublic API, but we have already changed it a lot working on 3.0.0 version. Also, the reduction of public api itself can be considered as a positive feature.

sungam3r: > This method is specifically for new users "getting started"

If so, let it stay.

joemcbride: @sungam3r It has a ton of value and should absolutely stay.  There needs to be an easy "let me try this out" without needing to new up several classes.  That was a major point of feedback for when this library was first introduced.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
sungam3r: Yes, that’s it. But this parameter is redundant. It is already contained in the second parameter. It only confuses.

> it could inadvertently affect users of the API 

Remember that you are already making backward incompatible changes. It makes no sense to try to maintain backward compatibility at a single point when everything around is already redone. Better to fix the old design flaw.

src/GraphQL.Serialization.Newtonsoft/GraphQL.Serialization.Newtonsoft.csproj
joemcbride: `GraphQL.Serialization.NewtonsoftJson` is better I think

joemcbride: Even `GraphQL.NewtonsoftJson` would work for me.

src/GraphQL.Serialization.Newtonsoft/SchemaExtensions.cs
sungam3r: >  That was a major point of feedback for when this library was first introduced.

Well, I don’t know much.

src/GraphQL.Serialization.Newtonsoft/GraphQL.Serialization.Newtonsoft.csproj
sungam3r: OK. So `GraphQL.NewtonsoftJson` and `GraphQL.SystemTextJson` for symmetry?

src/GraphQL/Instrumentation/ApolloTrace.cs
MerickOWA: Yep this should be done now.

src/GraphQL.Serialization.Newtonsoft/SchemaExtensions.cs
joemcbride: np - there was more than one person that mentioned it, but that feedback also came directly from Lee Byron.

src/GraphQL.Serialization.Newtonsoft/GraphQL.Serialization.Newtonsoft.csproj
joemcbride: 👍 I like the shorter names.

src/GraphQL/Execution/ExecutionError.cs
MerickOWA: Actually I was wrong, the ExecutionError already uses a special json converter, so making these public didn't affect serialization.  This should be changed now.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
MerickOWA: Moved.

src/GraphQL/Execution/ExecutionError.cs
sungam3r: Indeed, I myself forgot about it.

src/GraphQL/SchemaExtensions.cs
```diff
@@ -1,13 +1,12 @@
...
 using System.Threading.Tasks;
-using GraphQL.Http;
 using GraphQL.Types;
 
 namespace GraphQL
 {
     public static class SchemaExtensions
     {
-        public static async Task<string> ExecuteAsync(this ISchema schema, Action<ExecutionOptions> configure)
+        public static async Task<string> ExecuteAsync(this ISchema schema, IDocumentWriter documentWriter, Action<ExecutionOptions> configure)
```
joemcbride: I'm feeling like adding the `IDocumentWriter` parameter here pretty much defeats the purpose of the method.  I don't like it, but is seems like this method would have to move to the `NewtonsoftJson` lib.

That's making me think that the abstraction is wrong here.  Seems like we're moving too much to the other library.

src/GraphQL.Serialization.Newtonsoft/NewtonsoftContractResolver.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Reflection;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace GraphQL.Serialization.Newtonsoft
+{
+    internal class NewtonsoftContractResolver : DefaultContractResolver
```
sungam3r: Consider to make this public in case when someone want to tune this resolver.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
```diff
@@ -210,6 +210,10 @@ namespace GraphQL
         protected virtual GraphQL.Language.AST.Operation GetOperation(string operationName, GraphQL.Language.AST.Document document) { }
         protected virtual GraphQL.Execution.IExecutionStrategy SelectExecutionStrategy(GraphQL.Execution.ExecutionContext context) { }
     }
+    public static class DocumentWriterExtensions
```
sungam3r: By the way, add new package to api approval test :)
```c#
  [InlineData(typeof(DocumentWriter))]
```
in ApiApprovalTests.cs

src/GraphQL/SchemaExtensions.cs
MerickOWA: There is a second SchemaExtensions class in GraphQL.NewtonsoftJson.

    {
```
        public static Task<string> ExecuteAsync(this ISchema schema, Action<ExecutionOptions> configure)
        {
            return schema.ExecuteAsync(new DocumentWriter(indent: true), configure);
        }
    }
```

This allows a user to still have the simpler extension if they've chosen to use NewtonJson.

The only way to keep this entirely in GraphQL would be to make the result be a `Task<ExecutionResult>` instead of a `Task<String>`

src/GraphQL.Serialization.Newtonsoft/GraphQL.Serialization.Newtonsoft.csproj
MerickOWA: Library name and namespace has been changed

src/GraphQL/SchemaExtensions.cs
sungam3r: No, do not worry. We transferred as much as needed. In such situations, boundary places always appear, where additional configuration is required. Previously, they were simply invisible, since this entire configuration was hard-coded inside.

joemcbride: 👍 Missed the other half of that, thanks.  Yeah I'm thinking that return result from the executor should be revisited at some point.

src/GraphQL.Serialization.Newtonsoft/NewtonsoftContractResolver.cs
MerickOWA: Done

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
MerickOWA: I created it as a separate unit test since adding it to the existing PublicApi test caused it to fail as the approval text file can only match only one of the assemblies.

src/GraphQL.Serialization.Newtonsoft/ExecutionResultJsonConverter.cs
MerickOWA: This is reverted

src/GraphQL/Http/DocumentWriter.cs
MerickOWA: Done

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
sungam3r: OK. You can use overload with an additional parameter.

src/GraphQL.Serialization.Newtonsoft/Http/DocumentWriter.cs
MerickOWA: Its removed

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
MerickOWA: Merged the two unit tests together now, using the assembly name as the discriminator

sungam3r:
@MerickOWA Any progress on this? I don’t remember what we stopped at. It seems that then we actively discussed the comments and there were literally a couple of corrections left.

MerickOWA:
I believe all the corrections are made, unless I missed something. Currently this just separates the newtonsoft library serialization from GraphQL.  Adding support for System.Text.Json should be easy enough, but my work on that got pauses and is untested.

sungam3r:
@MerickOWA then could you resolve the conflicts please? I will check your PR again.

MerickOWA:
@sungam3r updated the pull request and squashed my commits down

sungam3r:
Need more time on this.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.NewtonsoftJson.approved.txt
```diff
@@ -0,0 +1,51 @@
+namespace GraphQL.NewtonsoftJson
```
StefanSchoof: Is GraphQL.Json.Newtonsoft a better namespace? Then we can have a other like GraphQL.Json.SystemText. I think this is more clear, that the can use for the same purpose.

sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/pull/1448#discussion_r354559306

StefanSchoof: Thanks, overlooked it.

MerickOWA:
Resolved the conflicts and rebased my changes on top of master

sungam3r:
@MerickOWA This PR is on the list of my obligatory tasks for the New Year holidays.

benmccallum:
@MerickOWA, I'm gonna get started on #1449 , but just wanted to check in to make sure we're not doubling effort. 
1. You haven't started anything to introduce System.Text.Json support yet have you? Your focus here was solely on separating out Newtonsoft.Json support into a separate project yea? 
2. Should I just follow along your project structure but making that uses the System.Text.Json serializer?  Is there enough interfaces in place now for me to just add this as an alternate serializer? Or do I need to introduce some more interfaces?

Cheers!

sungam3r:
Hey guys, does anyone know why I have 10 files marked as viewed, if this is actually not the case?
изображение

Is this a bug of the github interface after the rebase and force-push?

src/GraphQL.NewtonsoftJson/GraphQL.NewtonsoftJson.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <RootNamespace>GraphQL.NewtonsoftJson</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
```
sungam3r: Maybe update to 12.0.3 ?

sungam3r: @joemcbride I also remembered that I wanted to ask you what needs to be done so that a new project begins to fall into [artifacts](https://ci.appveyor.com/project/graphql-dotnet-ci/graphql-dotnet/builds/29906112/artifacts) and will be published to myget. I see that there are some JS scripts, but I understand almost nothing in them.

src/GraphQL.NewtonsoftJson/HttpResponseStreamWriter.cs
```diff
@@ -9,7 +9,7 @@
 using System.Text;
 using System.Threading.Tasks;
 
-namespace GraphQL.Http
+namespace GraphQL.NewtonsoftJson
```
sungam3r: `HttpResponseStreamWriter` can be made `internal` like `JsonArrayPool`.

src/GraphQL.NewtonsoftJson/NewtonsoftContractResolver.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Reflection;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace GraphQL.NewtonsoftJson
+{
+    public class NewtonsoftContractResolver : DefaultContractResolver
```
sungam3r: The name `NewtonsoftContractResolver` seems strange given that the namespace is already `GraphQL.NewtonsoftJson`. I propose to change it to `ExecutionResultContractResolver` to make the purpose clear.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+                new JsonSerializerSettings
+                {
+                    DateFormatHandling = DateFormatHandling.IsoDateFormat,
+                    DateParseHandling = DateParseHandling.None
+                });
+            return GetValue(values) as Dictionary<string, object>;
+        }
+
+        /// <summary>
+        /// Gets the value contained in a JObject, JValue, JProperty, or JArray.
```
sungam3r:
```suggestion
        /// Gets the value contained in a JObject, JValue, JProperty or JArray.
```

src/GraphQL/DocumentWriter.cs
```diff
@@ -0,0 +1,32 @@
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GraphQL
+{
+    public interface IDocumentWriter
```
sungam3r: I propose to split this file into two and rename.

src/GraphQL.NewtonsoftJson/DocumentWriter.cs
```diff
@@ -4,13 +4,8 @@
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 
-namespace GraphQL.Http
+namespace GraphQL.NewtonsoftJson
```
sungam3r: `Utf8Encoding` can be `private` now

src/GraphQL/DocumentWriter.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        Task WriteAsync<T>(Stream stream, T value);
+    }
+
+    public static class DocumentWriterExtensions
+    {
+        internal static readonly Encoding Utf8Encoding = new UTF8Encoding(false);
```
sungam3r: `Utf8Encoding` can be `private` now after separation.

sungam3r:
@MerickOWA Once again I carefully looked at your PR. Great job! I left my comments. The most important question is what needs to be done to start publishing a new package. The rest is just cosmetic things that you can fix in a couple of minutes.

src/GraphQL.NewtonsoftJson/GraphQL.NewtonsoftJson.csproj
MerickOWA: Should be upgraded

src/GraphQL.NewtonsoftJson/HttpResponseStreamWriter.cs
MerickOWA: It should be listed as internal now and the public api json has been updated to indicate that its gone.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
MerickOWA: Removed the comma

src/GraphQL.NewtonsoftJson/DocumentWriter.cs
MerickOWA: Should be private now

src/GraphQL/DocumentWriter.cs
MerickOWA: Should be private now

src/GraphQL/DocumentWriter.cs
MerickOWA: It should be split into two files now.

src/GraphQL.NewtonsoftJson/NewtonsoftContractResolver.cs
MerickOWA: Renamed

sungam3r:
@joemcbride PR is ready. Only question from me - what to do (if any) to publish new package?

joemcbride:
@sungam3r This is where the packing is currently being done:

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/tools/tasks/dotnetPack.js#L29

We will probably have to add another exec line there to add the new package.  It uses JavaScript promises, which are like .NET Tasks.  So would need to do a `Promise.all[exec(..), exec(...)]` to have it wait till both are built - or check to see if there is a way to pass multiple projects to `dotnet pack`.

sungam3r:
#1481 should fix it

sungam3r:
@joemcbride  Now I have a different question. Should we move `VersionPrefix` and `Authors` into `Directory.Build.props` to share these settings among all published projects? Or will the new package have a different author? Or is it worth expanding the list of authors?

@MerickOWA Also `PackageTags` and `Description` should be set inside `GraphQL.NewtonsoftJson` csproj to give a little info about new published package on Nuget.org download page.

sungam3r:
Great, packages are created as they should
изображение

sungam3r:
@joemcbride , @MerickOWA Any feedback on this [comment](https://github.com/graphql-dotnet/graphql-dotnet/pull/1448#issuecomment-571772602) ?

MerickOWA:
@sungam3r No I'm not really sure what the text of `PackageTags` and `Description` should be.  If you have a suggestion or need me to update it to something just let me know.

sungam3r:
```xml
<PackageTags>GraphQL;json</PackageTags>
<Description>JSON.NET serializer for GraphQL.NET</Description>
```

MerickOWA:
@sungam3r Done.  I also resolved the latest merge conflicts, however, I had to add a reference to logo.64x64.png in the GraphQL.Benchmarks project that was recently added in order to successfully build.

PR #1451, opened 2019-12-08
GraphType first schema directives

sungam3r:


sungam3r:
It is an experimental PR, not for merging.

sungam3r:
@Shane32 What do you think about it? We have been using the builds from this PR in production for a very long time but we ran into a problem when the UI (Playground/GraphiQL) stops working because it sees unknown elements in introspection. In theory, the new elements in introspection simply should not have been used by UI and caused any problems. But it was not so. I have to rebuild this branch after merging with up to date master from time to time to update our internal libraries.

Shane32:
Let me take a look.  Supporting Playground/GraphiQL is probably fairly important, but perhaps we can add a flag to enable 'experimental features', or some such thing.  Or perhaps the extra introspection data can be returned by an optional piece of middleware or something like that.  I'll have to read up on directives and the introspection fields so I can better understand what this feature does, and why Playground/GraphiQL have a problem with it.

Shane32:
So this PR lets an introspection query return the schema directives applied to a field -- for example, `@deprecated` (ignoring for the moment that introspection already allows for querying if a field is deprecated)?  Is that correct?  Not for query directives, which control execution.

And there is two notable changes:
* the addition of two new introspection types that start with "__" which violate the graphql rule that says that type names cannot start with "__"; and
* a new "directives" field on some of the introspection types

Question: does adding the new "directives" field to the introspection type also violate graphql spec?  It would seem so.

Separate question: why doesn't the SkipDirective or IncludeDirective get returned within a standard introspection query (regardless of this PR)?  Is it not a valid directive type as defined by the spec?  I understand no specific type name was specified in the spec, but it would seem that a query for `__schema { types { ... } }` should return directive types as well as user defined types.

I'm still thinking about how and if this PR could be redesigned to be external to the main graphql library.  Basically, allow the schema to add fields to the introspection types.  For example, you could add a "roles" field to `__Field` to return a list of strings indicating which permission roles this field is available to.  Then this PR could be packaged, either internally or externally, as an available option to utilize.  Of course this PR itself could be used for the same purpose as my example via a schema directive.

sungam3r:
> So this PR lets an introspection query return the schema directives applied to a field -- for example, @deprecated (ignoring for the moment that introspection already allows for querying if a field is deprecated)? Is that correct?

Yes. This PR is for custom server directives.

> Question: does adding the new "directives" field to the introspection type also violate graphql spec? It would seem so.

I cannot give a definite answer to this question, because I believe that the specification is incomplete in relation to introspection. It does not explicitly say that introspection cannot be extensible. Introspection query itself is GraphQL query, so... just read https://github.com/graphql/graphql-spec/issues/300

> why doesn't the SkipDirective or IncludeDirective get returned within a standard introspection query 

They do. See `IntrospectionResult.cs`:
```
""directives"": [
        {
          ""name"": ""include"",
          ""description"": ""Directs the executor to include this field or fragment only when the 'if' argument is true."",
          ""locations"": [
            ""FIELD"",
            ""FRAGMENT_SPREAD"",
            ""INLINE_FRAGMENT""
          ],
          ""args"": [
            {
              ""name"": ""if"",
              ""description"": ""Included when true."",
              ""type"": {
                ""kind"": ""NON_NULL"",
                ""name"": null,
                ""ofType"": {
                  ""kind"": ""SCALAR"",
                  ""name"": ""Boolean"",
                  ""ofType"": null
                }
              },
              ""defaultValue"": null
            }
          ]
        },
        {
          ""name"": ""skip"",
          ""description"": ""Directs the executor to skip this field or fragment when the 'if' argument is true."",
          ""locations"": [
            ""FIELD"",
            ""FRAGMENT_SPREAD"",
            ""INLINE_FRAGMENT""
          ],
          ""args"": [
            {
              ""name"": ""if"",
              ""description"": ""Skipped when true."",
              ""type"": {
                ""kind"": ""NON_NULL"",
                ""name"": null,
                ""ofType"": {
                  ""kind"": ""SCALAR"",
                  ""name"": ""Boolean"",
                  ""ofType"": null
                }
              },
              ""defaultValue"": null
            }
          ]
        },
        {
          ""name"": ""deprecated"",
          ""description"": ""Marks an element of a GraphQL schema as no longer supported."",
          ""locations"": [
            ""FIELD_DEFINITION"",
            ""ENUM_VALUE""
          ],
          ""args"": [
            {
              ""name"": ""reason"",
              ""description"": ""Explains why this element was deprecated, usually also including a suggestion for how to access supported similar data. Formatted in [Markdown](https://daringfireball.net/projects/markdown/)."",
              ""type"": {
                ""kind"": ""SCALAR"",
                ""name"": ""String"",
                ""ofType"": null
              },
              ""defaultValue"": ""\""No longer supported\""""
            }
          ]
        }
      ]
```

>  I understand no specific type name was specified in the spec, but it would seem that a query for __schema { types { ... } } should return directive types as well as user defined types.

`Directive` is not a `Type`. IT is a separate language and runtime element. See http://spec.graphql.org/June2018/#sec-Introspection.

> For example, you could add a "roles" field to __Field to return a list of strings indicating which permission roles this field is available to.

This is only a particular example of the directive's application. And then each time you need to change the structure of the introspection types. The ability to export directives makes it possible to attach arbitrary meta information to any element of the schema and give it to the client. This is a one-time change.

Shane32:
Well I like the concept.  Just not the fact that the implementation violates the current spec and causes errors in the common Playground/GraphiQL libraries.  It would seem that if the new types were not named with names starting with `__` then it would be unlikely that it would cause any problems with Playground/GraphiQL (or any other library).  I did test this with GraphiQL and found that its "intellisense" did not pick up on the new field defined for `__Field`, but it didn't kick any errors.  But of course having an introspection type name makes the most sense, so it won't conflict with any user defined types.

I'd probably investigate a way to include this as an optional feature or an extension to the main project.  Perhaps we target this feature for 3.1 which will give me a little time to come up with a more defined suggestion?  Any other thoughts?

sungam3r:
PR needs some refactoring in schema visitors and passing/storing state. I did not rush and decided to wait for the result of the work about directives in the GraphQL specification. I have no explicit intention of including this in the 3.1 milestone.

sungam3r:
> I have no explicit intention of including this in the 3.1 milestone.

And even in the 4.0 milestone.

>  Just not the fact that the implementation violates the current spec and causes errors in the common Playground/GraphiQL libraries.

The violation is not so obvious. You **can** interpret this as either a violation or a gap in the specification.

@Shane32 I resolved conflicts. You can take another look. Unfortunately no work will be carried out in https://github.com/graphql/graphql-spec/issues/300 : https://github.com/graphql/graphql-spec/issues/300#issuecomment-715211203

src/GraphQL/Types/IProvideMetadata.cs
```diff
@@ -10,10 +10,15 @@ namespace GraphQL.Types
     /// </summary>
     public interface IProvideMetadata
     {
+        /// <summary>
+        /// Provides all directives applied to this provider.
+        /// </summary>
+        AppliedDirectives AppliedDirectives { get; }
```
sungam3r, line 16: Directives can be applied to any element of the schema (even to directives 😄  ). Generally speaking, this is meta information. We can get directives from `Metadata` property indeed. I hesitated for a long time whether to do so. So far I've made a separate property.

src/GraphQL/Utilities/DirectiveVisitorSelector.cs
```diff
@@ -41,9 +41,13 @@ private IEnumerable<ISchemaNodeVisitor> BuildVisitors(List<GraphQLDirective> dir
                 if (_directiveVisitors.TryGetValue(name, out var type))
                 {
                     var visitor = _typeResolver(type);
-                    visitor.Name = name;
```
sungam3r, line 44: This line surprised me a lot.

sungam3r:
@joemcbride @Shane32 All the problems I described above have been resolved. By default, the server does not return any introspection types/fields that are not described in the specification. The default filter is used for this. Also, I completely separated the concepts of applied directives from `SchemaDirectiveVisitor` concept. I must admit that at that time I did not yet understand that these are completely different things. PR is ready for review.

Shane32:
Nice! I’ll review as soon as I’m able; probably tonight or tomorrow night.

src/GraphQL/Utilities/MetadataProvider.cs
```diff
@@ -11,24 +10,67 @@ namespace GraphQL.Utilities
...
             return local != null && local.TryGetValue(key, out object item) ? (TType)item : defaultValue;
         }
 
         /// <inheritdoc />
         public TType GetMetadata<TType>(string key, Func<TType> defaultValueFactory)
         {
-            var local = Metadata;
+            var local = _metadata;
             return local != null && local.TryGetValue(key, out object item) ? (TType)item : defaultValueFactory();
         }
```
Shane32, line 68: It would be nice to change these methods to a single `bool TryGetMetadata(string key, out TType item)` method, perhaps keeping these temporarily for v4.x and marking them as deprecated.

src/GraphQL/Utilities/MetadataProvider.cs
```diff
@@ -11,24 +10,67 @@ namespace GraphQL.Utilities
...
+                        if (_appliedDirectives == null)
+                        {
+                            _appliedDirectives = new AppliedDirectives();
+                        }
+                    }
+                }
+
+                return _appliedDirectives;
+            }
+        }
```
Shane32, line 34:
```suggestion
        public AppliedDirectives AppliedDirectives => _appliedDirectives ??= new AppliedDirectives();
```

Shane32 (changes requested):
- I looked through everything, but I don't use directives personally.  Seems fine now that the default schema filter prevents problems with GraphiQL and so on.

- I'd be interested to see "detailed" benchmarks (so we can see schema build times and memory use), but separately from the changes to metadata... I can probably run those after the code is all finalized though.

src/GraphQL/Utilities/MetadataProvider.cs
```diff
@@ -11,24 +10,67 @@ namespace GraphQL.Utilities
...
+                        if (_metadata == null)
+                        {
+                            _metadata = new Dictionary<string, object>();
+                        }
+                    }
+                }
+
+                return _metadata;
+            }
+        }
```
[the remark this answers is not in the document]
Shane32: The only way I could see a problem is if someone tries to lock on `obj.Metadata` to safely write information within a field resolver.  They could still lock on `obj`.  But if they don't write information, the worst thing that could happen is that an extra, empty dictionary gets created.  Thoughts?

src/GraphQL/Utilities/MetadataProvider.cs
Shane32: See above comments about concurrency with Metadata.

Shane32:
Just so I have a better understanding of directives, can you explain how you use schema directives and/or query directives?

src/GraphQL/Introspection/__Field.cs
```diff
@@ -79,6 +80,32 @@ public __Field()
             Field<NonNullGraphType<BooleanGraphType>>("isDeprecated", resolve: context => (!string.IsNullOrWhiteSpace(context.Source.DeprecationReason)).Boxed());
 
             Field(f => f.DeprecationReason, nullable: true).Description(null);
+
+            FieldAsync<NonNullGraphType<ListGraphType<NonNullGraphType<__AppliedDirective>>>>(
```
mikelneo, line 84: Is it already approved by graphql specification? https://github.com/graphql/graphql-spec/issues/300 is still open.

As long as I remember additional fields/types in introspection query ruins some playgrounds such as altair.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -412,6 +421,9 @@ namespace GraphQL.Builders
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> DefaultValue(TReturnType defaultValue = default) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> DeprecationReason(string deprecationReason) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Description(string description) { }
+        public GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Directive(string name) { }
+        public GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Directive(string name, System.Action<GraphQL.Types.AppliedDirective> configure) { }
+        public GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Directive(string name, string argumentName, object argumentValue) { }
```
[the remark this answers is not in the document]
sungam3r: I forgot. Done.

src/GraphQL/Types/DirectiveGraphType.cs
```diff
@@ -114,6 +115,13 @@ public DirectiveGraphType(string name, IEnumerable<DirectiveLocation> locations)
         /// </summary>
         public string Description { get; set; }
 
+        /// <summary>
+        /// Indicates whether the directive and its usages for schema elements should return in response
+        /// to an introspection request. By default (null) if the directive has all its locations of
+        /// type ExecutableDirectiveLocation, only then it will be present in the introspection response.
+        /// </summary>
+        public virtual bool? Introspectable { get; }
```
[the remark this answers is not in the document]
sungam3r: Correct. Done.

src/GraphQL/Types/IProvideMetadata.cs
```diff
@@ -10,10 +10,15 @@ namespace GraphQL.Types
...
+
         /// <summary>
         /// Provides all meta information as a key-value dictionary.
         /// </summary>
-        IDictionary<string, object> Metadata { get; }
+        Dictionary<string, object> Metadata { get; }
```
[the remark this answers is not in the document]
sungam3r: I found it pointless to provide the interface instead of a class as it is. We gain nothing with the interface.

src/GraphQL/Types/Schema.cs
```diff
@@ -350,6 +350,81 @@ private void CreateSchemaTypes()
...
             _allTypes.ApplyMiddleware(FieldMiddleware);
+
+            Validate();
+        }
+
+        protected virtual void Validate()
+        {
+            //TODO: add different validations, also see SchemaBuilder.Validate
+            //TODO: checks for parsed SDL may be expanded in the future, see https://github.com/graphql/graphql-spec/issues/653
+            ValidateAppliedDirectives();
```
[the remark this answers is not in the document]
sungam3r: I already thought about it yesterday. I will rewrite the code so that any validation happens in the visitor.

src/GraphQL/Utilities/MetadataProvider.cs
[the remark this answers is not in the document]
sungam3r: > Why are we trying to provide a lock here? As a general rule, no schema initialization code is thread-safe, and nor is it multi-threaded.

Agree. Yesterday I didn't have time to do that, it was too late.

src/GraphQL/Utilities/MetadataProvider.cs
sungam3r: `GetMetadata` actually already works as `TryGetMetadata`

src/GraphQL/Introspection/__Field.cs
sungam3r: @mikelneo  See `ISchemaFilter.cs` file above and `DefaultSchemaFilter` inside it:
```c#
public virtual Task<bool> AllowType(IGraphType type) => type is __AppliedDirective || type is __DirectiveArgument ? _forbidden : _allowed;
public virtual Task<bool> AllowField(IGraphType parent, IFieldType field) => parent is __Field && field.Name == "directives" ? _forbidden : _allowed;
```

src/GraphQL/Utilities/MetadataProvider.cs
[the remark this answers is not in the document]
sungam3r: Done.

src/GraphQL/Utilities/MetadataProvider.cs
sungam3r: Done.

sungam3r:
@Shane32 This example should help.

```graphql
type Account @authorized(role: "Admin")
{
  ....
}

type Query
{
  somenormalinfo: String! @authorized
  vipinfo: String! @authorized(role: "Advanced")
}

directive @authorized(role: String = "Reader") on FIELD | TYPE
```

src/GraphQL/Utilities/MetadataProvider.cs
Shane32: Only if you can provide a default value that isn’t valid for the dictionary item. That’s why a dictionary provides TryGetValue, so it can conclusively be determined that a value has never been set. And we don’t need an overload with a factory.

src/GraphQL/Types/IProvideMetadata.cs
sungam3r: Done.

src/GraphQL/Types/Schema.cs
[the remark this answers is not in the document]
sungam3r: Done.

src/GraphQL/Utilities/MetadataProvider.cs
sungam3r: Well, maybe, but not now.

src/GraphQL.Tests/Bugs/Bug1715.cs
```diff
@@ -1,29 +0,0 @@
...
-using GraphQL.Types;
-using Shouldly;
-using Xunit;
-
-namespace GraphQL.Tests.Bugs
-{
-    public class Bug1715
-    {
-        [Fact]
-        public void Register_Should_Throw_If_Schema_Initialized()
```
sungam3r, line 12: We have another test for this. Duplicate.

docs2/site/docs/migrations/migration4.md
```diff
@@ -9,6 +9,7 @@
 * Document caching supported via `IDocumentCache` and a default implementation within `DefaultDocumentCache`.
   Within the `GraphQL.Caching` nuget package, a memory-backed implementation is available which is backed by `Microsoft.Extensions.Caching.Memory.IMemoryCache`.
 * `ExecutionOptions.EnableMetrics` is disabled by default
+* Ability to expose user-defined meta-information via introspection. See https://github.com/graphql/graphql-spec/issues/300
```
Shane32, line 12: I suggest adding a note that MetaData is not thread safe

Shane32: Also probably need a few more notes about the changes made in this PR...?

src/GraphQL/Types/Collections/SchemaDirectives.cs
```diff
@@ -11,23 +11,59 @@ public class SchemaDirectives : IEnumerable<DirectiveGraphType>
...
+        /// <br/><br/>
+        /// Directives are used by the GraphQL runtime as a way of modifying execution
+        /// behavior. Type system creators do not usually create them directly.
+        /// </summary>
+        public void Register(DirectiveGraphType directive)
         {
             if (directive == null)
                 throw new ArgumentNullException(nameof(directive));
 
+            Schema?.CheckDisposed();
```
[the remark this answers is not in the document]
Shane32: Sounds fine to me

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -10,7 +10,7 @@ namespace GraphQL.Types
     /// Represents a field of a graph type.
     /// </summary>
     [DebuggerDisplay("{Name,nq}: {ResolvedType,nq}")]
-    public class FieldType : MetadataProvider, IFieldType, IProvideResolvedType
+    public class FieldType : MetadataProvider, IFieldType
```
sungam3r, line 13: `IFieldType` inherits from `IProvideResolvedType`

src/GraphQL/Types/Collections/SchemaDirectives.cs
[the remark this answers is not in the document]
sungam3r: And make `CheckInitialized` and `CheckDisposed` as `private` back.

docs2/site/docs/migrations/migration4.md
sungam3r: Done. I write notes after I commit, because it is convenient to watch changes in the GitHub UI.

src/GraphQL/Types/Collections/SchemaDirectives.cs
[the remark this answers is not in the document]
sungam3r: Done.

sungam3r:
I'm done here. I did everything I wanted. If there are no suggestions that require corrections, then I do not plan to make any more changes.

Shane32:
@sungam3r I think we need to review the performance impact of this PR.  build time is 20% slower and memory usage is 26% higher

```
## develop


|        Method |           Stage |             Mean |          Error |         StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|-------------- |---------------- |-----------------:|---------------:|---------------:|---------:|---------:|------:|----------:|
| Introspection |           Build | 5,461,493.917 ns | 37,105.8312 ns | 32,893.3529 ns | 414.0625 | 132.8125 |     - | 2630973 B |
| Introspection |           Parse |    17,843.112 ns |     25.4572 ns |     21.2579 ns |   1.8921 |   0.0610 |     - |   11920 B |
| Introspection |         Convert |     9,528.126 ns |     14.3479 ns |     12.7190 ns |   2.6245 |   0.1373 |     - |   16552 B |
| Introspection |        Validate |   180,451.329 ns |    343.0972 ns |    304.1467 ns |   1.7090 |        - |     - |   11104 B |
| Introspection | DeserializeVars |         1.304 ns |      0.0035 ns |      0.0033 ns |        - |        - |     - |         - |
| Introspection |  ParseVariables |         2.929 ns |      0.0081 ns |      0.0076 ns |        - |        - |     - |         - |
| Introspection |         Execute |   668,659.310 ns |  1,825.3314 ns |  1,707.4162 ns |  50.7813 |  16.6016 |     - |  319130 B |
| Introspection |       Serialize |   114,013.001 ns |    197.8588 ns |    185.0772 ns |   4.3945 |   0.3662 |     - |   27920 B |
|               |                 |                  |                |                |          |          |       |           |
|          Hero |           Build | 5,491,381.458 ns | 23,440.8919 ns | 21,926.6248 ns | 414.0625 | 132.8125 |     - | 2630973 B |
|          Hero |           Parse |       816.150 ns |      1.4354 ns |      1.2725 ns |   0.1249 |        - |     - |     784 B |
|          Hero |         Convert |       481.836 ns |      1.1387 ns |      1.0651 ns |   0.1678 |        - |     - |    1056 B |
|          Hero |        Validate |    11,031.324 ns |     25.9113 ns |     24.2375 ns |   0.3815 |        - |     - |    2464 B |
|          Hero | DeserializeVars |         1.307 ns |      0.0066 ns |      0.0058 ns |        - |        - |     - |         - |
|          Hero |  ParseVariables |         2.942 ns |      0.0125 ns |      0.0117 ns |        - |        - |     - |         - |
|          Hero |         Execute |     2,139.920 ns |      5.9347 ns |      5.5513 ns |   0.3586 |        - |     - |    2264 B |
|          Hero |       Serialize |       537.271 ns |      0.8017 ns |      0.7499 ns |   0.0858 |        - |     - |     544 B |
|               |                 |                  |                |                |          |          |       |           |
|     Variables |           Build | 5,126,564.342 ns | 26,925.8329 ns | 23,869.0495 ns | 398.4375 | 101.5625 |     - | 2522196 B |
|     Variables |           Parse |     1,577.414 ns |      2.6310 ns |      2.3323 ns |   0.1602 |        - |     - |    1008 B |
|     Variables |         Convert |       532.449 ns |      1.1819 ns |      1.0477 ns |   0.1917 |        - |     - |    1208 B |
|     Variables |        Validate |    19,936.600 ns |     37.9512 ns |     33.6428 ns |   1.5564 |        - |     - |    9865 B |
|     Variables | DeserializeVars |     9,535.063 ns |     16.6044 ns |     15.5317 ns |   1.3123 |   0.0305 |     - |    8288 B |
|     Variables |  ParseVariables |    34,277.234 ns |    162.5118 ns |    152.0136 ns |   2.3193 |        - |     - |   14879 B |
|     Variables |         Execute |     2,914.619 ns |     22.8662 ns |     21.3891 ns |   0.3204 |        - |     - |    2024 B |
|     Variables |       Serialize |       391.652 ns |      6.9701 ns |      6.5199 ns |   0.0863 |        - |     - |     544 B |
|               |                 |                  |                |                |          |          |       |           |
|       Literal |           Build | 5,370,508.594 ns | 38,135.9052 ns | 35,672.3494 ns | 398.4375 | 101.5625 |     - | 2522199 B |
|       Literal |           Parse |    18,958.147 ns |     23.6639 ns |     19.7604 ns |   1.4343 |   0.0305 |     - |    9016 B |
|       Literal |         Convert |     7,766.882 ns |     22.6809 ns |     21.2157 ns |   2.0294 |   0.0610 |     - |   12768 B |
|       Literal |        Validate |   135,637.817 ns |    260.6810 ns |    243.8412 ns |   4.1504 |        - |     - |   27200 B |
|       Literal | DeserializeVars |         1.341 ns |      0.0008 ns |      0.0006 ns |        - |        - |     - |         - |
|       Literal |  ParseVariables |         2.438 ns |      0.0039 ns |      0.0037 ns |        - |        - |     - |         - |
|       Literal |         Execute |    32,194.915 ns |    146.4560 ns |    136.9951 ns |   2.2583 |        - |     - |   14304 B |
|       Literal |       Serialize |       389.348 ns |      0.6584 ns |      0.5498 ns |   0.0863 |        - |     - |     544 B |

## changes excluding metadata changes



|        Method |           Stage |             Mean |          Error |         StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|-------------- |---------------- |-----------------:|---------------:|---------------:|---------:|---------:|------:|----------:|
| Introspection |           Build | 6,694,760.573 ns | 26,923.0597 ns | 25,183.8467 ns | 531.2500 | 171.8750 |     - | 3377338 B |
| Introspection |           Parse |    17,824.618 ns |     39.7846 ns |     37.2145 ns |   1.8921 |   0.0610 |     - |   11920 B |
| Introspection |         Convert |     9,382.351 ns |     12.5784 ns |     10.5035 ns |   2.6245 |   0.1373 |     - |   16552 B |
| Introspection |        Validate |   178,195.525 ns |    565.5679 ns |    501.3612 ns |   1.7090 |        - |     - |   11104 B |
| Introspection | DeserializeVars |         1.741 ns |      0.0037 ns |      0.0033 ns |        - |        - |     - |         - |
| Introspection |  ParseVariables |         2.471 ns |      0.0043 ns |      0.0038 ns |        - |        - |     - |         - |
| Introspection |         Execute |   708,382.819 ns |  4,037.9590 ns |  3,777.1093 ns |  51.7578 |  16.6016 |     - |  328204 B |
| Introspection |       Serialize |   115,112.977 ns |    156.8885 ns |    146.7536 ns |   4.3945 |   0.3662 |     - |   28256 B |
|               |                 |                  |                |                |          |          |       |           |
|          Hero |           Build | 6,638,965.792 ns | 20,967.6074 ns | 18,587.2378 ns | 531.2500 | 171.8750 |     - | 3377347 B |
|          Hero |           Parse |       826.463 ns |      1.8337 ns |      1.7152 ns |   0.1249 |        - |     - |     784 B |
|          Hero |         Convert |       471.280 ns |      1.1507 ns |      0.9609 ns |   0.1683 |   0.0005 |     - |    1056 B |
|          Hero |        Validate |    11,346.315 ns |     44.5526 ns |     39.4948 ns |   0.3815 |        - |     - |    2464 B |
|          Hero | DeserializeVars |         1.722 ns |      0.0029 ns |      0.0027 ns |        - |        - |     - |         - |
|          Hero |  ParseVariables |         2.622 ns |      0.0073 ns |      0.0069 ns |        - |        - |     - |         - |
|          Hero |         Execute |     2,160.693 ns |      4.7885 ns |      4.4792 ns |   0.3586 |        - |     - |    2264 B |
|          Hero |       Serialize |       524.044 ns |      0.9428 ns |      0.8819 ns |   0.0858 |        - |     - |     544 B |
|               |                 |                  |                |                |          |          |       |           |
|     Variables |           Build | 6,499,235.677 ns | 25,824.5490 ns | 24,156.2991 ns | 515.6250 | 164.0625 |     - | 3271139 B |
|     Variables |           Parse |     1,571.796 ns |      0.7827 ns |      0.6939 ns |   0.1602 |        - |     - |    1008 B |
|     Variables |         Convert |       531.763 ns |      2.5626 ns |      2.3971 ns |   0.1917 |        - |     - |    1208 B |
|     Variables |        Validate |    20,093.443 ns |     65.7022 ns |     61.4579 ns |   1.5564 |        - |     - |    9865 B |
|     Variables | DeserializeVars |     9,590.070 ns |     24.3039 ns |     22.7339 ns |   1.3123 |   0.0305 |     - |    8288 B |
|     Variables |  ParseVariables |    34,223.042 ns |    137.8664 ns |    128.9603 ns |   2.3193 |        - |     - |   14879 B |
|     Variables |         Execute |     2,811.322 ns |     14.3659 ns |     13.4379 ns |   0.3204 |        - |     - |    2024 B |
|     Variables |       Serialize |       379.179 ns |      0.7881 ns |      0.6987 ns |   0.0863 |        - |     - |     544 B |
|               |                 |                  |                |                |          |          |       |           |
|       Literal |           Build | 6,371,884.688 ns | 24,330.3680 ns | 22,758.6413 ns | 515.6250 | 164.0625 |     - | 3270796 B |
|       Literal |           Parse |    18,926.753 ns |     29.6839 ns |     26.3140 ns |   1.4343 |   0.0305 |     - |    9016 B |
|       Literal |         Convert |     7,655.748 ns |     19.3371 ns |     18.0879 ns |   2.0294 |   0.0610 |     - |   12768 B |
|       Literal |        Validate |   138,112.439 ns |    118.3820 ns |     92.4249 ns |   4.1504 |        - |     - |   27200 B |
|       Literal | DeserializeVars |         1.722 ns |      0.0055 ns |      0.0049 ns |        - |        - |     - |         - |
|       Literal |  ParseVariables |         2.834 ns |      0.0088 ns |      0.0083 ns |        - |        - |     - |         - |
|       Literal |         Execute |    32,214.735 ns |    158.1931 ns |    147.9739 ns |   2.2583 |        - |     - |   14304 B |
|       Literal |       Serialize |       391.276 ns |      0.7417 ns |      0.6575 ns |   0.0863 |        - |     - |     544 B |
```

sungam3r:
Wow. Could you please merge it with new diff tool?

sungam3r:
изображение

sungam3r:
Somewhere there is clearly a mistake. I'll deal.

Shane32:
> Wow. Could you please merge it with new diff tool?

Sure.  I really have very little time today, and have not looked at the diff tool yet.  Maybe I can tonight.

I think we should merge in the metadata changes (lazy load dictionary, use dictionary instead of concurrent dictionary) in a separate PR first.

Shane32:
|        Method | Stage |     Mean |     Error |    StdDev |   Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |------ |---------:|----------:|----------:|--------:|--------:|------:|----------:|
| Before | Build | 1.409 ms | 0.0113 ms | 0.0105 ms | 27.3438 | 13.6719 |     - | 172.16 KB |
| After | Build | 1.387 ms | 0.0078 ms | 0.0073 ms | 29.2969 | 13.6719 |     - | 180.41 KB |

sungam3r:
My results:
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.14393.3986 (1607/AnniversaryUpdate/Redstone1)
Frequency=3515611 Hz, Resolution=284.4456 ns, Timer=TSC
.NET Core SDK=5.0.102
  [Host]     : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT
  DefaultJob : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT


```
|Method|Stage|Mean [base]|Mean [current]|Mean [% of base]|Allocated [base]|Allocated [current]|Allocated [% of base]|
|---|---|--:|--:|--:|--:|--:|--:|
|Introspection|Build|1,326,711.911 ns|1,290,096.026 ns|<span style="color:green">97</span>|176967 B|185315 B|<span style="color:red">104</span>|
|Introspection|Parse|20,957.060 ns|20,117.013 ns|<span style="color:green">95</span>|11920 B|11920 B|100|
|Introspection|Convert|11,257.845 ns|10,513.416 ns|<span style="color:green">93</span>|16552 B|16552 B|100|
|Introspection|Validate|211,778.636 ns|205,800.621 ns|<span style="color:green">97</span>|11104 B|11106 B|100|
|Introspection|DeserializeVars|2.111 ns|1.752 ns|<span style="color:green">82</span>|-|-|-|
|Introspection|ParseVariables|4.039 ns|3.232 ns|<span style="color:green">80</span>|-|-|-|
|Introspection|Execute|820,194.525 ns|819,756.662 ns|<span style="color:green">99</span>|319176 B|328262 B|<span style="color:red">102</span>|
|Introspection|Serialize|133,181.629 ns|125,985.829 ns|<span style="color:green">94</span>|27920 B|28256 B|<span style="color:red">101</span>|
|||||||||
|Hero|Build|1,329,491.007 ns|1,269,003.149 ns|<span style="color:green">95</span>|176967 B|185271 B|<span style="color:red">104</span>|
|Hero|Parse|983.707 ns|926.734 ns|<span style="color:green">94</span>|784 B|784 B|100|
|Hero|Convert|570.095 ns|531.154 ns|<span style="color:green">93</span>|1056 B|1056 B|100|
|Hero|Validate|13,055.161 ns|12,620.991 ns|<span style="color:green">96</span>|2464 B|2464 B|100|
|Hero|DeserializeVars|2.164 ns|1.903 ns|<span style="color:green">87</span>|-|-|-|
|Hero|ParseVariables|3.863 ns|3.105 ns|<span style="color:green">80</span>|-|-|-|
|Hero|Execute|2,426.455 ns|2,572.477 ns|<span style="color:red">106</span>|2264 B|2264 B|100|
|Hero|Serialize|610.046 ns|632.858 ns|<span style="color:red">103</span>|544 B|544 B|100|
|||||||||
|Variables|Build|1,010,974.234 ns|1,108,171.282 ns|<span style="color:red">109</span>|135239 B|144242 B|<span style="color:red">106</span>|
|Variables|Parse|1,805.206 ns|1,872.593 ns|<span style="color:red">103</span>|1008 B|1008 B|100|
|Variables|Convert|596.432 ns|669.682 ns|<span style="color:red">112</span>|1208 B|1208 B|100|
|Variables|Validate|22,925.215 ns|24,641.164 ns|<span style="color:red">107</span>|7801 B|7801 B|100|
|Variables|DeserializeVars|11,012.407 ns|11,549.144 ns|<span style="color:red">104</span>|8288 B|8288 B|100|
|Variables|ParseVariables|36,646.056 ns|37,498.492 ns|<span style="color:red">102</span>|14197 B|14197 B|100|
|Variables|Execute|3,174.930 ns|3,544.834 ns|<span style="color:red">111</span>|2025 B|2025 B|100|
|Variables|Serialize|464.782 ns|487.943 ns|<span style="color:red">104</span>|544 B|544 B|100|
|||||||||
|Literal|Build|992,736.572 ns|1,110,617.606 ns|<span style="color:red">111</span>|135254 B|144257 B|<span style="color:red">106</span>|
|Literal|Parse|21,514.666 ns|22,221.844 ns|<span style="color:red">103</span>|9016 B|9016 B|100|
|Literal|Convert|8,643.107 ns|8,916.594 ns|<span style="color:red">103</span>|12768 B|12768 B|100|
|Literal|Validate|155,186.451 ns|149,563.961 ns|<span style="color:green">96</span>|27200 B|27200 B|100|
|Literal|DeserializeVars|2.029 ns|1.651 ns|<span style="color:green">81</span>|-|-|-|
|Literal|ParseVariables|3.308 ns|2.983 ns|<span style="color:green">90</span>|-|-|-|
|Literal|Execute|35,534.045 ns|35,632.568 ns|100|14312 B|14312 B|100|
|Literal|Serialize|435.801 ns|463.086 ns|<span style="color:red">106</span>|544 B|544 B|100|

Shane32 (approved):
Seems fine to me.

sungam3r:
I will continue investigation. Serialize result 101 for introspection confuses me. It seems that somehow I return more data than before.

src/GraphQL/Introspection/ISchemaFilter.cs
```diff
@@ -44,25 +44,48 @@ public interface ISchemaFilter
...
     /// </summary>
     public class DefaultSchemaFilter : ISchemaFilter
     {
-        private static readonly Task<bool> _completed = Task.FromResult(true);
+        private static readonly Task<bool> _allowed = Task.FromResult(true);
+        private static readonly Task<bool> _forbidden = Task.FromResult(false);
 
         /// <inheritdoc/>
-        public virtual Task<bool> AllowType(IGraphType type) => _completed;
+        public virtual Task<bool> AllowType(IGraphType type) => type is __AppliedDirective || type is __DirectiveArgument ? _forbidden : _allowed;
```
mikelneo, line 57: I think that hardcode the values ​​in this place is not the best way to hide the directive introspection feature. 
For someone unfamiliar with the implementation details, it will be unclear how to enable the feature. 
It seems to me that it would be better to move the enable / disable of this feature to the Schema level.

Something like 
```c#
var shcema = new Schema(); 
shcema.UseDirectiveIntrospection() 
``` 
would be great

Shane32: We certainly should well-document it. Maybe we could at least provide two schema filters even if the default hides the out-of-spec fields.

sungam3r: Done:
```c#
 public static TSchema EnableExperimentalIntrospectionFeatures<TSchema>(this TSchema schema)
            where TSchema : Schema
        {
            schema.Filter = new ExperimentalFeaturesSchemaFilter();
            return schema;
        }
```

Shane32: Nice

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -1002,13 +1019,21 @@ namespace GraphQL.Introspection
     }
     public class DefaultSchemaFilter : GraphQL.Introspection.ISchemaFilter
     {
+        protected static readonly System.Threading.Tasks.Task<bool> _allowed;
+        protected static readonly System.Threading.Tasks.Task<bool> _forbidden;
```
Shane32, line 1023: Improper case for protected members.  Should be Allowed and Forbidden.  Perhaps something more descriptive would be appropriate also, but it's probably fine.

Shane32 (changes requested):

src/GraphQL/Introspection/ISchemaFilter.cs
```diff
@@ -44,25 +44,64 @@ public interface ISchemaFilter
...
+    /// The default schema filter. By default nothing is hidden. Please note
+    /// that some features that are not in the official specification may be
+    /// hidden by default. These features can be unlocked using special
+    /// <see cref="ExperimentalFeaturesSchemaFilter"/> filter.
     /// </summary>
     public class DefaultSchemaFilter : ISchemaFilter
     {
-        private static readonly Task<bool> _completed = Task.FromResult(true);
+        protected static readonly Task<bool> _allowed = Task.FromResult(true);
+        protected static readonly Task<bool> _forbidden = Task.FromResult(false);
```
Shane32, line 55: Missing xml comments

Shane32 (changes requested):

src/GraphQL/Introspection/ISchemaFilter.cs
```diff
@@ -44,25 +44,64 @@ public interface ISchemaFilter
...
 
         /// <inheritdoc/>
-        public virtual Task<bool> AllowArgument(IFieldType field, QueryArgument argument) => _completed;
+        public virtual Task<bool> AllowArgument(IFieldType field, QueryArgument argument) => _allowed;
 
         /// <inheritdoc/>
-        public virtual Task<bool> AllowEnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue) => _completed;
+        public virtual Task<bool> AllowEnumValue(EnumerationGraphType parent, EnumValueDefinition enumValue) => _allowed;
+
+        public virtual Task<bool> AllowDirective(DirectiveGraphType directive)
```
Shane32, line 69:
```suggestion
        /// <inheritdoc/>
        public virtual Task<bool> AllowDirective(DirectiveGraphType directive)
```

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: Done

src/GraphQL/Introspection/ISchemaFilter.cs
sungam3r: done

src/GraphQL/Introspection/ISchemaFilter.cs
sungam3r: done

src/GraphQL/Types/Schema.cs
```diff
@@ -345,11 +297,23 @@ private void CreateSchemaTypes()
...
+        }
+
+        /// <summary>
+        /// Validates correctness of the created schema. This method is called only once - during schema initialization.
+        /// </summary>
+        protected virtual void Validate()
+        {
+            //TODO: add different validations, also see SchemaBuilder.Validate
+            //TODO: checks for parsed SDL may be expanded in the future, see https://github.com/graphql/graphql-spec/issues/653
+            this.Run(new AppliedDirectivesVisitor(this));
```
sungam3r, line 316: This line causes additional build time (and allocations?).

Shane32 (commented):
I'm not sure what the goal is of all these changes.  I'm guessing that you are trying to reduce memory consumption for people that do not want applied directives?  I would revert all these changes if that is the case.  If it is just additional checks to be sure someone does not apply an applied directive to a field unless that's enabled, then again, I would not implement these changes.  So either way I would not implement these changes.

I don't know enough about applied directives, but isn't it possible that someone will want to use applied directives while still having the filter enabled to be spec compliant?  Maybe not; I still don't quite understand these directives enough.

src/GraphQL/Types/Schema.cs
Shane32: Your last benchmark only shows 4% memory increase, and faster build time.  It's not a big deal for those very few people that have scoped schemas.  If they were concerned about speed or memory they would not use a scoped schema.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -132,14 +153,14 @@ private void CheckSealed()
...
+        /// <param name="schema">A schema for which this instance is created.</param>
         public static SchemaTypes Create(
             IEnumerable<IGraphType> types,
             IEnumerable<DirectiveGraphType> directives,
             Func<Type, IGraphType> resolveType,
-            INameConverter nameConverter)
+            ISchema schema)
         {
-            var lookup = new SchemaTypes(nameConverter ?? CamelCaseNameConverter.Instance);
+            var lookup = new SchemaTypes(schema.NameConverter ?? CamelCaseNameConverter.Instance, schema.Filter is ExperimentalFeaturesSchemaFilter);
```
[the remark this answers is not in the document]
Shane32: If necessary, you could add a property to `ISchemaFilter` - e.g. `AllowAppliedDirectives`.

src/GraphQL/Introspection/__EnumValue.cs
```diff
@@ -11,6 +11,11 @@ public class __EnumValue : ObjectGraphType<EnumValueDefinition>
         /// Initializes a new instance of the <c>__EnumValue</c> introspection type.
         /// </summary>
         public __EnumValue()
+            : this(false)
+        {
+        }
+
+        internal __EnumValue(bool allowAppliedDirectives)
```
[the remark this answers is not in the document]
Shane32: Basically, I think this new feature should be part of the main codebase, not some "secret switch" that if we enable it works only one way.  So if you want all these types to have this boolean, then add it.  Perhaps `public __EnumValue(bool allowAppliedDirectives = false)`.  But don't have all this stuff internal so only when you use a certain filter it enables a whole bunch of code that is otherwise inaccessible.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -1016,16 +1016,16 @@ namespace GraphQL.Introspection
...
+        protected static readonly System.Threading.Tasks.Task<bool> Forbidden;
         public DefaultSchemaFilter() { }
         public virtual System.Threading.Tasks.Task<bool> AllowArgument(GraphQL.Types.IFieldType field, GraphQL.Types.QueryArgument argument) { }
         public virtual System.Threading.Tasks.Task<bool> AllowDirective(GraphQL.Types.DirectiveGraphType directive) { }
         public virtual System.Threading.Tasks.Task<bool> AllowEnumValue(GraphQL.Types.EnumerationGraphType parent, GraphQL.Types.EnumValueDefinition enumValue) { }
         public virtual System.Threading.Tasks.Task<bool> AllowField(GraphQL.Types.IGraphType parent, GraphQL.Types.IFieldType field) { }
         public virtual System.Threading.Tasks.Task<bool> AllowType(GraphQL.Types.IGraphType type) { }
     }
-    public class ExperimentalFeaturesSchemaFilter : GraphQL.Introspection.DefaultSchemaFilter
+    public sealed class ExperimentalFeaturesSchemaFilter : GraphQL.Introspection.DefaultSchemaFilter
```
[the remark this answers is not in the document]
sungam3r: Done

src/GraphQL/Introspection/__EnumValue.cs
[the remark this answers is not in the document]
sungam3r: Good. It's even easier for me.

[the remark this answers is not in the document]
sungam3r: Done

src/GraphQL/Types/Collections/SchemaTypes.cs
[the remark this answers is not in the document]
sungam3r: Already done. The situation is a little more complicated. See comments to the `Mode` property. Condition against `ExperimentalFeaturesSchemaFilter` will remain.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -132,14 +153,14 @@ private void CheckSealed()
...
         /// <param name="directives">A list of directives to register.</param>
         /// <param name="resolveType">A delegate which returns an instance of a graph type from its .NET type.</param>
-        /// <param name="nameConverter">A name converter to use for the specified graph types.</param>
+        /// <param name="schema">A schema for which this instance is created.</param>
         public static SchemaTypes Create(
             IEnumerable<IGraphType> types,
             IEnumerable<DirectiveGraphType> directives,
             Func<Type, IGraphType> resolveType,
-            INameConverter nameConverter)
+            ISchema schema)
```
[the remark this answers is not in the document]
sungam3r: I was expecting this comment. I intentionally did not pass the additional bool parameter. I think the schema is like a bunch of settings. Now, to create `SchemaTypes `, I took a second bool setting besides `nameConverter`. Then I may need another one tomorrow. Are we going to change the signature of this public method every time? It is better to pass a reference to the schema itself and use what is needed from it inside.

src/GraphQL/Types/Schema.cs
sungam3r: Nevertheless, I have already taken this line under the condition.

Shane32 (changes requested):
I read all of the comments on the Mode property.

There is no need for the Mode property of the `ExperimentalFeaturesSchemaFilter`.  It is a simple filter that exposes all types.  If you want it anyway, okay, but there is no need for it.

Code elsewhere should not have a check against the above filter.  It is a schema filter.  It is there solely to filter outbound introspection requests.  It is not an options class.  Please remove the check.  There are many alternatives.  You could add an option to ISchema called `EnableExperimentalFeatures`.  Although a still poorer design, you could add a property to `ISchemaFilter`.  Do not have a hard coded check for the schema filter to configure the types.  They are completely unrelated bits of code.

There is no need for it and it is a poor design to have a mandatory check against a certain type for a property that is designed to simply .

src/GraphQL/Types/Collections/SchemaTypes.cs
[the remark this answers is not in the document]
Shane32: > Are we going to change the signature of this public method every time?

I would say yes, or add overloads, but if I'm okay if we want to make the above change, I suppose.

src/GraphQL/SchemaExtensions.cs
```diff
@@ -9,6 +11,21 @@ namespace GraphQL
...
+        /// </summary>
+        /// <param name="schema">The schema for which the features are enabled.</param>
+        /// <param name="mode">Experimental features mode.</param>
+        /// <returns>Reference to the provided <paramref name="schema"/>Experimental features mode.</returns>
+        public static TSchema EnableExperimentalIntrospectionFeatures<TSchema>(this TSchema schema, ExperimentalIntrospectionFeaturesMode mode = ExperimentalIntrospectionFeaturesMode.ExecutionOnly)
+            where TSchema : Schema
+        {
+            schema.Filter = new ExperimentalFeaturesSchemaFilter { Mode = mode };
+            return schema;
+        }
```
Shane32, line 27: Something like this:
```suggestion
        public static TSchema EnableExperimentalIntrospectionFeatures<TSchema>(this TSchema schema, ExperimentalIntrospectionFeaturesMode mode = ExperimentalIntrospectionFeaturesMode.ExecutionOnly)
            where TSchema : Schema
        {
            schema.EnableAppliedDirectives = true;
            if (mode == Modes.ExecutionAndIntrospection)
                schema.Filter = new ExperimentalFeaturesSchemaFilter();
            return schema;
        }
```

Shane32: Or maybe:
```cs
schema.ExperimentalFeatures.AppliedDirectives = true;
```

sungam3r:
> There is no need for the Mode property of the ExperimentalFeaturesSchemaFilter. It is a simple filter that exposes all types. If you want it anyway, okay, but there is no need for it.
Code elsewhere should not have a check against the above filter. It is a schema filter. It is there solely to filter outbound introspection requests. It is not an options class. Please remove the check. There are many alternatives. You could add an option to ISchema called EnableExperimentalFeatures. Although a still poorer design, you could add a property to ISchemaFilter. Do not have a hard coded check for the schema filter to configure the types. They are completely unrelated bits of code.
There is no need for it and it is a very bad design to have a mandatory check against a certain type for a property that is designed to simply filter outbound introspection requests.

I understand all this, I just thought to use an already existing property. Now it's obvious to me that I needed to make a new property right away. I will redo it.

src/GraphQL/SchemaExtensions.cs
sungam3r: Maybe it is worth to move all properties from ISchema into one settings type? Then there will be fewer backward incompatible changes in the future (if we want to keep `ISchema` interface at all).

Shane32: Maybe. What would be on it? Filter and Comparer? Looking through the property list I don’t see much else you’d want to move.

sungam3r: I'll put it off for a while. I would like to redesign and remove some properties from the schema.

src/GraphQL/Types/Collections/SchemaTypes.cs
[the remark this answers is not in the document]
sungam3r: I added new property to `Schema`.

src/GraphQL/SchemaExtensions.cs
sungam3r: I made changes with a filter. They were spinning in my head and now everything seemed to be in place.

Shane32 (approved):

Shane32:
Looking good @sungam3r thanks! :)

sungam3r:
So... the question is - why 101 % on serialization stage?

Shane32:
> So... the question is - why 101 % on serialization stage?

The tests you changed demonstrated that some of the types are returned in a different order than previously.  Maybe that changes how much memory System.Text.Json uses for some reason.

Btw did that get changed back?  I thought I saw a commit that changed the order again.  If so, maybe another set of memory tests would prove/disprove this theory.

Regardless, I wouldn't worry about it.

sungam3r:
I disabled applied directives so order of types became as it was before all changes.

Shane32:
> I disabled applied directives so order of types became as it was before all changes.

Maybe serialization memory will go back to 100%

src/GraphQL/Introspection/__Type.cs
```diff
@@ -49,8 +50,9 @@ public __Type()
                     }),
                 async context =>
                 {
-                    if (context.Source is IComplexGraphType type)
```
sungam3r, line 52: @Shane32 I found bug. It was in that line. `InputObjectGraphType` also inherits from `IComplexGraphType` but it should not be here. So I really returned more fields.

sungam3r:
Now it is too late to make benchmarks. I'll do it tomorrow.

src/GraphQL/Introspection/__Type.cs
Shane32: I suggest you add a test for this potential issue.

sungam3r: #2233

sungam3r:
Finally! I did it
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.14393.3986 (1607/AnniversaryUpdate/Redstone1)
Frequency=3515611 Hz, Resolution=284.4456 ns, Timer=TSC
.NET Core SDK=5.0.102
  [Host]     : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT
  DefaultJob : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT


```
|Method|Stage|Mean [base]|Mean [current]|Mean [% of base]|Allocated [base]|Allocated [current]|Allocated [% of base]|
|---|---|--:|--:|--:|--:|--:|--:|
|Introspection|Build|355,887.570 ns|337,232.999 ns|<span style="color:green">94</span>|97314 B|96537 B|<span style="color:green">99</span>|
|Introspection|Parse|21,299.654 ns|20,686.241 ns|<span style="color:green">97</span>|11920 B|11920 B|100|
|Introspection|Convert|11,608.027 ns|11,329.900 ns|<span style="color:green">97</span>|16552 B|16552 B|100|
|Introspection|Validate|197,548.426 ns|204,554.089 ns|<span style="color:red">103</span>|11104 B|11104 B|100|
|Introspection|DeserializeVars|2.052 ns|1.864 ns|<span style="color:green">90</span>|-|-|-|
|Introspection|ParseVariables|3.879 ns|3.492 ns|<span style="color:green">90</span>|-|-|-|
|Introspection|Execute|786,298.085 ns|816,292.325 ns|<span style="color:red">103</span>|319180 B|322360 B|100|
|Introspection|Serialize|121,396.062 ns|125,617.871 ns|<span style="color:red">103</span>|27920 B|27880 B|<span style="color:green">99</span>|
|||||||||
|Hero|Build|332,107.765 ns|329,336.219 ns|<span style="color:green">99</span>|97243 B|96478 B|<span style="color:green">99</span>|
|Hero|Parse|945.108 ns|983.678 ns|<span style="color:red">104</span>|784 B|784 B|100|
|Hero|Convert|557.313 ns|573.219 ns|<span style="color:red">102</span>|1056 B|1056 B|100|
|Hero|Validate|12,599.999 ns|12,885.186 ns|<span style="color:red">102</span>|2464 B|2464 B|100|
|Hero|DeserializeVars|1.522 ns|1.854 ns|<span style="color:red">121</span>|-|-|-|
|Hero|ParseVariables|2.830 ns|3.392 ns|<span style="color:red">119</span>|-|-|-|
|Hero|Execute|2,578.424 ns|2,583.273 ns|100|2264 B|2264 B|100|
|Hero|Serialize|611.260 ns|627.634 ns|<span style="color:red">102</span>|544 B|504 B|<span style="color:green">92</span>|
|||||||||
|Variables|Build|227,071.075 ns|232,067.626 ns|<span style="color:red">102</span>|65929 B|65489 B|<span style="color:green">99</span>|
|Variables|Parse|1,791.690 ns|1,827.924 ns|<span style="color:red">102</span>|1008 B|1008 B|100|
|Variables|Convert|618.683 ns|638.078 ns|<span style="color:red">103</span>|1208 B|1208 B|100|
|Variables|Validate|23,291.088 ns|23,736.276 ns|<span style="color:red">101</span>|7801 B|7705 B|<span style="color:green">98</span>|
|Variables|DeserializeVars|11,039.643 ns|12,122.854 ns|<span style="color:red">109</span>|8288 B|8288 B|100|
|Variables|ParseVariables|35,858.479 ns|38,505.299 ns|<span style="color:red">107</span>|13020 B|13020 B|100|
|Variables|Execute|3,241.102 ns|3,674.601 ns|<span style="color:red">113</span>|2025 B|2025 B|100|
|Variables|Serialize|441.904 ns|473.287 ns|<span style="color:red">107</span>|544 B|504 B|<span style="color:green">92</span>|
|||||||||
|Literal|Build|240,331.347 ns|242,996.873 ns|<span style="color:red">101</span>|65930 B|65489 B|<span style="color:green">99</span>|
|Literal|Parse|21,177.198 ns|23,324.517 ns|<span style="color:red">110</span>|9016 B|9016 B|100|
|Literal|Convert|8,919.389 ns|9,631.910 ns|<span style="color:red">107</span>|12768 B|12768 B|100|
|Literal|Validate|158,196.648 ns|170,517.856 ns|<span style="color:red">107</span>|27200 B|27200 B|100|
|Literal|DeserializeVars|1.584 ns|1.912 ns|<span style="color:red">120</span>|-|-|-|
|Literal|ParseVariables|2.966 ns|3.550 ns|<span style="color:red">119</span>|-|-|-|
|Literal|Execute|36,431.067 ns|37,462.084 ns|<span style="color:red">102</span>|13134 B|13134 B|100|
|Literal|Serialize|467.804 ns|494.283 ns|<span style="color:red">105</span>|544 B|504 B|<span style="color:green">92</span>|

PR #1452, opened 2019-12-08
Add more directive locations for visitors

sungam3r:
Relates to #1451 
@joemcbride Do you mind these changes? I am still working on how to best implement the application of directives in GraphType first approach in #1451. However, it is obvious that there is a part that we will have to implement in any case. This PR adds directive support for all necessary elements of the schema. After merge, #1451 will be more focused on further changes.

### DIFF
--- a/src/GraphQL/Types/ComplexGraphType.cs
+++ b/src/GraphQL/Types/ComplexGraphType.cs
@@ -32,15 +32,7 @@ namespace GraphQL.Types
             DeprecationReason ??= typeof(TSourceType).ObsoleteMessage();
         }
 
-        public IEnumerable<FieldType> Fields
-        {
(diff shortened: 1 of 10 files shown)


src/GraphQL/Types/ComplexGraphType.cs
```diff
@@ -32,15 +32,7 @@ protected ComplexGraphType()
             DeprecationReason ??= typeof(TSourceType).ObsoleteMessage();
         }
 
-        public IEnumerable<FieldType> Fields
-        {
-            get => _fields;
-            private set
```
sungam3r: It has not been used anywhere.

sungam3r:
@joemcbride any objections?

Issue #1147, opened 2019-05-16
Use a yield iterator for ResolveFieldContext.Path


johnrutherford:
Currently during the document execution, string arrays are created to set the `Path` property on ResolveFieldContexts. This results in a lot of arrays being allocated for a property that is probably very rarely used. 

Since the property is `IEnumerable<string>`, we should use a yield iterator to generate the enumerable when it's actually needed. This should be pretty simple to do by walking up the execution node ancestors and then doing a `Reverse()`.

Shane32:
We could also just change AppendPath as follows:
``` C#
        public static IEnumerable<string> AppendPath(IEnumerable<string> path, string pathSegment)
        {
            return path.Append(pathSegment);
        }
```
The Enumerable.Append code does not actually clone the array, but rather as it is enumerated, the proper values are returned.

sungam3r:
But Enumerable.Append yet creates new object - iterator instance:
```c#
 public static IEnumerable<TSource> Append<TSource>(this IEnumerable<TSource> source, TSource element)
    {
      if (source == null)
        throw Error.ArgumentNull(nameof (source));
      Enumerable.AppendPrependIterator<TSource> appendPrependIterator = source as Enumerable.AppendPrependIterator<TSource>;
      if (appendPrependIterator != null)
        return (IEnumerable<TSource>) appendPrependIterator.Append(element);
      return (IEnumerable<TSource>) new Enumerable.AppendPrepend1Iterator<TSource>(source, element, true);
    }
```

Shane32:
True, it's not lazy loaded per the suggestion, which would probably be best.  I just thought this solution was better than what's there now, and very simple to implement.

sungam3r:
Honestly, I don’t think it will be easy. Intuition suggests that in order to optimize the memory allocation we will have to rethink a much larger amount of code, not only `Path`. This object is passed on to other objects, and they are passed on to others.

Earthmark:
I believe list in the core framework gets around this by returning a struct instead of an interface, as long as you don't cast it to IEnumerable it's free. Doing foreach on it ends up using the struct directly, all it cares about is that the iterator calls line up.

sungam3r:
[...]

PR #1456, opened 2019-12-14
Heap allocation optimizations

sungam3r:
Introspection requests are good material for experimenting with memory allocation. This is what my test for requesting introspection shows:

``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.14393.3243 (1607/AnniversaryUpdate/Redstone1)
Frequency=3515629 Hz, Resolution=284.4441 ns, Timer=TSC
.NET Core SDK=3.1.100
  [Host]     : .NET Core 3.1.0 (CoreCLR 4.700.19.56402, CoreFX 4.700.19.56404), X64 RyuJIT  [AttachedDebugger]


```
|        Method |     Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated | commit
|-------------- |---------:|----------:|----------:|---------:|--------:|------:|----------:|-------:|
| Introspection | 2.108 ms | 0.0406 ms | 0.0543 ms | 269.5313 | 93.7500 |     - |   1.25 MB | f0e192f [master] |
| Introspection | 1.910 ms | 0.0386 ms | 0.0814 ms | 230.4688 | 82.0313 |     - |   1.07 MB |  |
| Introspection | 1.661 ms | 0.0097 ms | 0.0086 ms | 222.6563 | 74.2188 |     - |   1.02 MB | d75a119 |
| Introspection | 1.567 ms | 0.0143 ms | 0.0134 ms | 222.6563 | 68.3594 |     - |   1.01 MB | c2e6b3c |
| Introspection | 1.568 ms | 0.0341 ms | 0.0615 ms | 246.0938 |       - |     - | 1020.74 KB | 51d31f3 |
| Introspection | 1.463 ms | 0.0132 ms | 0.0117 ms | 244.1406 | 7.8125 |     - |   1019 KB | b07256e |

This is just what I managed to fix, without really getting into code refactoring.

**[UPDATE]**
After #1500 :

|        Method |     Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated | commit
|-------------- |---------:|----------:|----------:|---------:|--------:|------:|----------:|-------:|
| Introspection | 1.470 ms | 0.0135 ms | 0.0126 ms | 179.6875 | 35.1563 |     - | 806.81 KB |


src/GraphQL/Types/ResolveFieldContext.cs
```diff
@@ -79,32 +79,47 @@ public ResolveFieldContext(ResolveFieldContext context)
 
         public TType GetArgument<TType>(string name, TType defaultValue = default)
```
sungam3r: previous implementation of this method resulted in `defaultValue` parameter boxing

sungam3r:
Relates to #1147:
image

src/GraphQL/BoolBox.cs
```diff
@@ -0,0 +1,27 @@
+namespace GraphQL
+{
+    public static class BoolBox
```
sungam3r: In principle, this class can be made `internal`. @joemcbride What do you think?

sungam3r:
In all places indicated by this comment
`// DO NOT USE LINQ ON HOT PATH`
delegates actively allocated on the heap due to closures.

sungam3r:
Fix for #1147 will be in a separate PR as it will require a design change.

sungam3r:
@Shane32 I have some thoughts on `ResolveFieldContext` and your changes in #1441 
image

`ResolveFieldContext` is the topmost

Shane32:
@sungam3r Excellent; let's discuss.  I have also have some significant performance improvements to ResolveFieldContext I'd like to suggest, but they depend on the IFieldResolver interfaces being implemented.  I actually have the PR ready to go, which I could post now, but it depends on #1441 .  This new PR will cut the number of allocations by ExecutionStrategy per resolver run by as much as 80%, depending on the resolver, not to mention reduced memory requirements and increased execution speed.

I only glanced at this PR, but I see that one of the changes is to convert ResolveFieldContext to a struct.  If interfaces are used such as in #1441 , all performance enhancements (if any) are lost, as structs are boxed when cast to an interface.

sungam3r:
We need to merge #1441 first.

sungam3r:
Then after merge this PR you can bench your changes with the new benchmarks project.

src/GraphQL/BoolBox.cs
sungam3r: Now it is internal.

sungam3r:
@joemcbride @SimonCropp PR is ready for review. I resolved conflicts and double-checked each file. Many files have been changed, but all the changes come down to fairly simple operations that do not change the whole design. After merging this ASAP I would like to update parser dependency graphql-dotnet/parser#53 to see further effect.

@joemcbride I would appreciate your feedback. I would not want merging such PR without your approve.

src/GraphQL/Utilities/EmptyEnumerator.cs
```diff
@@ -0,0 +1,22 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace GraphQL.Utilities
+{
+    internal sealed class EmptyEnumerator<T> : IEnumerator<T>
```
Shane32: Doesn't Enumerable.Empty<> work the same?  See https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq/src/System/Linq/Enumerable.SpeedOpt.cs

Shane32: Finally found the EmptyPartition class: https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq/src/System/Linq/Partition.SpeedOpt.cs

sungam3r: ok

Shane32:
Looks good to me.  I think the boxing and short-circuit optimizations are making the real difference here.  LINQ optimizes calls to `Any()` and `Count()` (by testing if they can be cast before enumerating), and my benchmarks show that replacing IEnumerable with List has little effect.  However, every little bit helps, and the required changes to the API are minimal and seem fine to me.  Excellent work @sungam3r  !

sungam3r:
The only thing I don't like is a lot of changes in one PR.

Shane32:
** If ** you wanted to break it down .... you could put the boxing optimizations in one PR, the short-circuit optimizations in another PR, and the removal of LINQ in a third PR.  Something like that.  Then it would be easier to review the code and/or run benchmarks.

PR #1457, opened 2019-12-14
[WIP] Astprinter fix and forgotten nodes

sungam3r:
- support for some `ValueNodes`: `SByteValue`, `UIntValue` etc; **DateTime, Guid, TimeSpan nodes still require support, clarification about formatting required**
- turn `Matches` from gettable/settable property to virtual/override only gettable property
- move `PrintAst` from settable property to the required ctor argument - actually all `Configure` calls for different nodes was required to call `Print` otherwise there would be a NRE on
```c#
return config.PrintAst(vals);
```
It seems to me that the resulting api is cleaner.

@joemcbride  While adding printing support for the _built-in custom scalars_ (forgotten nodes) in the library, I wondered, but what about user-defined scalars? Now for them there is no support and the introspection request will return `null` as the default value for `__InputValue` even if it is specified as not null.

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.approved.txt
@@ -2151,20 +2151,21 @@ namespace GraphQL.Types.Relay.DataObjects
 }
 namespace GraphQL.Utilities
 {
-    public class AstPrintConfig
+    public abstract class AstPrintConfig
     {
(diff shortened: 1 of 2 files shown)


SimonCropp:
@sungam3r LGTM. want me to have a go at the merge conflicts?

sungam3r:
It was a long time ago but I remember this PR. In the process of solving my problem, I had questions that I posed in summary. I did not want to merge the code of the half-solved problem. Let's figure it out if you want to move on.

PR #1461, opened 2019-12-20
Fix docs for async only extension

joemcbride:


sungam3r:
Thanks for correction. I specifically kept your [remark](https://github.com/graphql-dotnet/graphql-dotnet/pull/1416#issuecomment-562337354) open in the browser as a reminder that the documentation needed to be fixed. Thank you for taking this job.

PR #1462, opened 2019-12-23
Prevent use ObjectGraphType in QueryArgument

sungam3r:
Relates to #1436 

By the way, two tests were written incorrectly. Fixed.

### DIFF
--- a/src/GraphQL/Types/QueryArgument.cs
+++ b/src/GraphQL/Types/QueryArgument.cs
@@ -16,6 +16,9 @@ namespace GraphQL.Types
     [DebuggerDisplay("{Name,nq}: {ResolvedType,nq}")]
     public class QueryArgument : MetadataProvider, IHaveDefaultValue
     {
+        private Type _type;
+        private IGraphType _resolvedType;
+
         public QueryArgument(IGraphType type)
         {
             ResolvedType = type ?? throw new ArgumentOutOfRangeException(nameof(type), "QueryArgument type is required");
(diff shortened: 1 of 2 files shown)


sungam3r:
@joemcbride If you don't mind I would like to merge these changes ASAP.

src/GraphQL/Types/QueryArgument.cs
```diff
@@ -37,8 +40,35 @@ public QueryArgument(Type type)
...
+        {
+            if (type?.IsInputType() == false)
+                throw Create(nameof(Type), type);
+
+            return type;
+        }
+
+        private IGraphType CheckResolvedType(IGraphType type)
+        {
+            if (!(type.GetNamedType() is GraphQLTypeReference) && type?.IsInputType() == false)
```
joemcbride: Just noticed, I think missing a `type?` here on the `type.GetNamedType()`.

Maybe add a test with null for the `ResolvedType` property?

sungam3r: I already raised a finger over the merge button :)

```c#
  public static IGraphType GetNamedType(this IGraphType type)
        {
            return type switch
            {
                NonNullGraphType nonNull => GetNamedType(nonNull.ResolvedType),
                ListGraphType list => GetNamedType(list.ResolvedType),
                _ => type
            };
        }
```

sungam3r: I would say that on the contrary we can remove `?` when used before `IsInputType` or `IsOutputType` (`IsOutputType` requires small fix).

sungam3r: image

Nope, nothing to fix.

joemcbride: I would still like to see a test that sets `ResolvedType` to null to confirm the behavior.

sungam3r: Ok. I was wrong about `IsInputType`, `bool?` vs `bool`.

sungam3r: Done

PR #1463, opened 2019-12-24
Middleware regression fix - create middleware once

Shane32:
In #1442 middleware functionality was changed to create an instance of the middleware every time the middleware's resolve function was executed.  This PR fixes that behavior.


Shane32:
Note that unlike #1442 this code does not validate the constructor during the installation stage.  The generic version of `Use<T>` will prevent classes without default constructors from being passed in, but the non-generic version will throw an exception during the build stage if the type cannot be created.

PR #1469, opened 2020-01-02
Add interface match check

sungam3r:
Also fixes `friendsConnection` field type and some tests. Regression in #995.

This PR fixes the problem of mismatching field types in interface implementations. I came across this problem when I wanted to add GraphQL Playground. It was a long time ago and I did not understand why Playground refuses to work. Now I managed to take a closer look at the problem. It turned out that the server returned an invalid introspection response, and the SPA application validated it and displayed an error in the console. At the same time, neither the schema nor the docs were displayed as if the server was not responding to the request at all.

### DIFF
--- a/src/GraphQL.StarWars/Types/CharacterInterface.cs
+++ b/src/GraphQL.StarWars/Types/CharacterInterface.cs
@@ -1,4 +1,5 @@
 using GraphQL.Types;
+using GraphQL.Types.Relay;
 
 namespace GraphQL.StarWars.Types
 {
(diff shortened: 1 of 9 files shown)


src/GraphQL.StarWars/Types/CharacterInterface.cs
```diff
@@ -12,7 +13,7 @@ public CharacterInterface()
             Field(d => d.Name, nullable: true).Description("The name of the character.");
 
             Field<ListGraphType<CharacterInterface>>("friends");
-            Field<ListGraphType<CharacterInterface>>("friendsConnection");
+            Field<ConnectionType<CharacterInterface, EdgeType<CharacterInterface>>>("friendsConnection");
```
sungam3r: The source of problems with Playground.

sungam3r:
@joemcbride I would like to merge this ASAP to continue my work to add Playground support.

sungam3r:
Example of error message:
```
System.ArgumentException: 'Type 'Human' (HumanType) does not implement 'Character' interface. Field 'Human.friendsConnection' must be of type 'ListGraphType<CharacterInterface>', but in fact it is of type 'ConnectionType<CharacterInterface,EdgeType<CharacterInterface>>'.'
```

sungam3r:
I will merge this PR. If there are any comments, then I will do another PR to fix them.

PR #1470, opened 2020-01-02
Field description can be taken from the interface field

### LINKED ISSUE #1004: Implementations of a InterfaceGraphType should inherit that interface's field descriptions
## Summary

If an `InterfaceGraphType` is defined as described in the [docs](https://graphql-dotnet.github.io/docs/getting-started/interfaces), then field descriptions from the interface could be inherited by the fields on any `ObjectGraphType` that implements that interface. Furthermore, a field description defined on the type implementing the interface should then override the inherited description.

### Motivation

This would allow interface field descriptions to show up in GraphiQL documentation on any field in the hierarchy -- unless they are overridden by a child type in that hierarchy -- thereby simplifying the documentation effort.

sungam3r:
fixes #1004

### DIFF
--- a/src/GraphQL.Tests/Bugs/Issue1004.cs
+++ b/src/GraphQL.Tests/Bugs/Issue1004.cs
@@ -1,0 +1,66 @@
+using GraphQL.Types;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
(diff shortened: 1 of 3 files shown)


SimonCropp:
should there be a tests to verify this?

sungam3r:
Done

PR #1471, opened 2020-01-02
Harness reworking

sungam3r:
- remove not working GraphiQL, add Playground
- fix HTTP status code to always return 200 OK
- use IOptions pattern and values from config in middleware


src/GraphQL.Harness/GraphQLSettings.cs
```diff
@@ -6,7 +6,7 @@ namespace Example
 {
     public class GraphQLSettings
     {
-        public PathString Path { get; set; } = "/api/graphql";
+        public PathString GraphQLPath { get; set; }
```
sungam3r: Renamed as conflicted with embedded PATH property.

joemcbride:
The GraphiQL _does_ work, last I checked, you just need to run the webpack to build the JavaScript bundle. So if we’re removing tat GraphiQL setup we should also remove that webpack config file.  I don’t necessarily mind using the Playground (it is nicer).

sungam3r:
I did not know about webpack. Do you mean webpack.config.js ? Any other files to delete?
I agree that Playground is nicer. In addition, it requires less config/code files to work which may be useful as an example.

sungam3r:
I removed webpack config file and also added GraphiQL support.

PR #1474, opened 2020-01-06
support type hierachy of usercontext type

### LINKED ISSUE #731: UserContext can not be injected as interface
Using the Handler/Endpoints approach, lets assume I have defined the following method on my Schema class :

```csharp
public Test Test(IMyUserContext context){
   //Do some stuff
}
```
Additionally, I have registered an UserContext that implements the interface.

This will not work, because the [code](https://github.com/graphql-dotnet/graphql-dotnet/blob/cc5d15fe68563b206e713d2144e6af9217860630/src/GraphQL/Resolvers/FieldResolverBase.cs#L36) that checks the parameter type and compares it with the registered UserContext type only takes the concrete type into account.

SimonCropp:
fixes #731

### DIFF
--- a/src/GraphQL/Reflection/ReflectionHelper.cs
+++ b/src/GraphQL/Reflection/ReflectionHelper.cs
@@ -98,11 +98,13 @@ namespace GraphQL.Reflection
             }
 
             if (parameters.Length > index
-                && context.UserContext != null
-                && context.UserContext?.GetType() == parameters[index].ParameterType)
+                && context.UserContext != null)
             {
-                arguments[index] = context.UserContext;
-                index++;
+                if (parameters[index].ParameterType.IsInstanceOfType(context.UserContext))
+                {
+                    arguments[index] = context.UserContext;
+                    index++;
+                }
             }
 
             foreach (var parameter in parameters.Skip(index))
--- a/src/GraphQL/Reflection/ReflectionHelper.cs
+++ b/src/GraphQL/Reflection/ReflectionHelper.cs
@@ -98,13 +98,11 @@ namespace GraphQL.Reflection
             }
 
             if (parameters.Length > index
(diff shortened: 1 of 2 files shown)


src/GraphQL/Reflection/ReflectionHelper.cs
```diff
@@ -98,11 +98,13 @@ public static PropertyInfo PropertyForField(this Type type, string field)
             }
 
             if (parameters.Length > index
-                && context.UserContext != null
-                && context.UserContext?.GetType() == parameters[index].ParameterType)
+                && context.UserContext != null)
             {
-                arguments[index] = context.UserContext;
-                index++;
+                if (parameters[index].ParameterType.IsInstanceOfType(context.UserContext))
```
sungam3r: Sometimes I wonder at the names of the methods. It would have to be called `IsTypeOfInstance`.

src/GraphQL/Reflection/ReflectionHelper.cs
```diff
@@ -98,11 +98,13 @@ public static PropertyInfo PropertyForField(this Type type, string field)
             }
 
             if (parameters.Length > index
-                && context.UserContext != null
-                && context.UserContext?.GetType() == parameters[index].ParameterType)
```
sungam3r: Do you mind to move `parameters[index].ParameterType.IsInstanceOfType(context.UserContext)` condition right into this line of code to minimize changes?

SimonCropp: sorry i dont really groc your comment. can you make the change?

sungam3r: done

SimonCropp: thanks

Issue #1323, opened 2019-10-03
Debug symbols for 2.4.0


maksims-ahadovs:
Where is it possible to download .pdb files for 2.4.0 release? Previous releases had separate NuGet package with them. But for this release I cannot find them.

sungam3r:
Relates to #931

sungam3r:
Closed in favor of #1475

PR #1475, opened 2020-01-06
add SourceLink and move to embedded pdb

SimonCropp:
info on why embedded pdbs https://github.com/dotnet/sdk/issues/2679


src/Directory.Build.props
```diff
@@ -1,5 +1,7 @@
 <Project>
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
+    <DebugType>embedded</DebugType>
+    <DebugSymbols>true</DebugSymbols>
```
sungam3r: Why not in GraphQL.csproj ?

src/GraphQL/GraphQL.csproj
```diff
@@ -18,13 +18,15 @@
...
     <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
   </PropertyGroup>
 
   <ItemGroup>
     <Compile Include="..\CommonAssemblyInfo.cs" />
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" Condition="$(Configuration) == 'Release'" />
```
sungam3r: Is this `Condition` necessary?

src/Directory.Build.props
sungam3r: Also it seems that `DebugSymbols` is `true` by default.

SimonCropp: i think it is only true in debug? and i will move this into GraphQL.csproj

src/GraphQL/GraphQL.csproj
SimonCropp: i try to avoid impacting build perf. and since this interrogates the git files system i would rather only pay that cost in release mode.

Happy to remove if u think it is not worth it

src/Directory.Build.props
sungam3r: You are right, only in debug. But this page https://github.com/dotnet/sourcelink says nothing about `DebugSymbols`. It is necessary?

src/GraphQL/GraphQL.csproj
sungam3r: It makes sense. What about building in Release configuration not on the CI server? Will it slow down dev machine?

src/Directory.Build.props
SimonCropp: https://github.com/dotnet/sourcelink#using-source-link-in-net-projects

> If you distribute the library via a package published to NuGet.org, it is recommended to build a symbol package and publish it to NuGet.org as well. This will make the symbols available on NuGet.org symbol server, where the debugger can download it from when needed. Alternatively, you can include the symbols in the main package. However, doing so is not recommended as it increases the size of the package and thus restore time for projects that consume your package.

Although i think the " However, doing so is not recommended as it increases the size of the package and thus restore time for projects that consume your package." is FUD

Using embedded symbols means the package goes from 171KB => 242KB. So yes it is bigger, but it means we always get line numbers for exceptions, even in production environments. Using a snupkg (symbols package) keeps the nuget smaller, but in my experience no one sets those up correctly to work in production environments. Many dont even have them working in dev environs, since many orgs white list packages. So i go with embedded for all my nugets, it is just simpler

src/GraphQL/GraphQL.csproj
SimonCropp: > What about building in Release configuration not on the CI server

yep they will pay the price

TBH i never measured this one specifically. I just dont do unnecessary stuff in Debug.

src/Directory.Build.props
sungam3r: My question is not about the `DebugType` property, but about `DebugSymbols`. I understand what `DebugType` is for. I experimented a bit with building packages with different settings. I found this https://github.com/Microsoft/msbuild/issues/2169 and I see that you also left a comment. It seems that `DebugType` takes precedence over `DebugSymbols`.

I understand that all this may not be important, I just don’t want to litter the project files with insignificant settings that will cause unnecessary questions.

src/GraphQL/GraphQL.csproj
sungam3r: I used to run benchmarks in release configuration. This is a rather painstaking job with many edits and rebuilds. I worked on graphql-dotnet/parser#53 recently and after merging this I am going to work on this repo. So these changes may affect me. Could you measure the delay build time?

src/Directory.Build.props
SimonCropp: yep fair enough. will remove DebugSymbols

Shane32:
I'd endorse this change, as I find it particularly difficult to track down errors that occur within GraphQL, whether it's during development or production.  I typically log errors to a database server-side, including full stack trace, and having a more detailed stack trace will help debug problems.  The extra 71kb dll size doesn't bother me a bit.

SimonCropp:
seems the timing is in the sub 20ms range. so i remove the "if release"

sungam3r:
We have new packages for serializers, maybe we should transfer the added properties to `Directory.Build.props`?

SimonCropp:
@sungam3r done

SimonCropp:
note this should be squashed on merge since the commits are a bit ad-hoc

sungam3r:
I always squash. But what about `PublishRepositoryUrl` and packagereference to SourceLink?

SimonCropp:
good catch on the PublishRepositoryUrl.

re the packagereference. i dont like to use solution level package references yet, since the nuget tooling does not yet properly support them. i can move it but i would suspect it would mean some friction and possible manually updating that package until the proper sln level nuget support comes out.

SimonCropp:
sorry now realized u prob meant adding the PackageReference  to the json project. this is now done

sungam3r:
Well, this is not such a big trouble, especially since it is unlikely that you will often have to update the package often. And please add `Condition="'$(IsPackable)' == 'true'"` to PackageReference.

sungam3r:
I mean to add reference to `Directory.Build.props`.

SimonCropp:
done

sungam3r:
relates to #1323

PR #1479, opened 2020-01-06
remove  || namedType is EnumerationGraphType

SimonCropp:
EnumerationGraphType inherits from ScalarGraphType so the second "is" is redundant

### DIFF
--- a/src/GraphQL/GraphQLExtensions.cs
+++ b/src/GraphQL/GraphQLExtensions.cs
@@ -56,7 +56,7 @@ namespace GraphQL
             return typeof(ScalarGraphType).IsAssignableFrom(namedType) ||
                    typeof(IObjectGraphType).IsAssignableFrom(namedType) ||
                    typeof(IInterfaceGraphType).IsAssignableFrom(namedType) ||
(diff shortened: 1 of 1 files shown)


src/GraphQL/GraphQLExtensions.cs
```diff
@@ -30,7 +30,7 @@ public static bool IsCompositeType(this IGraphType type)
         public static bool IsLeafType(this IGraphType type)
         {
             var namedType = type.GetNamedType();
-            return namedType is ScalarGraphType || namedType is EnumerationGraphType;
```
sungam3r: I think it is here for clarity. See comments below.

SimonCropp: i am not sure it adds any clarity. anyone can inherit from ScalarGraphType and we dont include them. perhaps @joemcbride can be tie breaker?

sungam3r: I am not against removal. But then you need to remove all such uses from this file (or even solution).

SimonCropp: done

PR #1480, opened 2020-01-06
first param of ArgumentOutOfRangeException is param name

SimonCropp:


src/GraphQL/Types/GraphType.cs
```diff
@@ -49,7 +49,7 @@ internal void SetName(string name, bool validate)
                     NameValidator.ValidateName(name, "type");
 
                     if (IsTypeModifier)
-                        throw new ArgumentOutOfRangeException("A type modifier (List, NonNull) name must be null");
+                        throw new ArgumentOutOfRangeException(nameof(name), "A type modifier (List, NonNull) name must be null");
```
sungam3r: Gotcha! Thanks.

PR #1483, opened 2020-01-08
Block body for ctors

sungam3r:
I started applying the rules from .editorconfig . I hope that gradually we will come to a common opinion about which set of rules should be used.


SimonCropp:
LGTM

PR #1492, opened 2020-01-09
Add GraphiQL docs

### LINKED ISSUE #499: Expand documentation to include tutorial on setting up GraphiQL UI.
Hello!

Thanks for this awesome library. I'm working to build a case for using this at my workplace, and having GraphiQL set up is a huge part of that. Unfortunately, I'm having trouble getting the UI set up without existing documentation, and an example project that I can't seem to run (also no docs for how to run). 

This seems like an easy win since the code is already written. I would love to see a tutorial for setting up the UI. This would be, as far as I can tell, the only tool that has a working UI for a non-Core version of ASP.NET. 

If I'm missing something, or there are existing resources I could use in the meantime, I'd really appreciate it. 

Thanks!

sungam3r:
fixes #499


sungam3r:
@joemcbride Thanks for early feedback. I have not yet been able to generate the documentation site myself. When installing the necessary components, something fails. Could you please update the docs site?

joemcbride:
@sungam3r I deployed the docs.  It looks like we need to fix the link to the interfaces documentation in the Union docs.

sungam3r:
#1493

PR #1499, opened 2020-01-13
Remove unused properties from ResolveConnectionContext

Shane32:
These properties of the ResolveConnectionContext class and IResolveConnectionContext interface are not used by the project:
``` c#
int? NumberOfSkippedEntries { get; }
int? TotalCount { get; }
bool IsPartial { get; }
```
These properties are only `{ get; }` properties on the interface, and therefor will only ever return default values.  I recommend removing the properties.


sungam3r:
@joemcbride can you give your opinion? I am not familiar with the Relay yet.

SimonCropp:
lets merge this. want me to handle the conflicts?

Shane32:
I wouldn't have minded, but since I'm at my PC, I'll do it.

sungam3r:
If you understand what you are doing, then do it. It seems that those properties are really not used. I'm out of the game so far with Relay.

PR #1500, opened 2020-01-13
ResolveFieldContext performance enhancement

Shane32:
Attached is a multi-part draft performance enhancement contained within a new class `ResolveFieldContextSlim`, a public class for use by `ExecutionStrategy`.  Also included is an internal class 'ResolveFieldContextAs', explained below.

First, the code always attempts to create an instance of `ResolveFieldContextSlim` which most closely matches the type of the returned object.  Since the `IResolveFieldContext` interface is covariant, it can be directly cast to any compatible type.  This greatly reduces the number of times that `As<>` needs to be performed, as the only time the type cannot be inferred is when `Source` is `null`.

Secondly, the interface only holds references to two values (the `ExecutionNode` and the `ExecutionContext`).  If and when a field resolver wants to retrieve one of the interface properties (e.g. `Schema`), it is pulled from the execution node/context directly.

Third, the `Arguments` and `SubFields` properties are lazy loaded via `Lazy<T>`.  Due to the infrequent use of these properties and amount of code required to initialize these variables, there should be a notable reduction in both allocated objects and execution time.  Also, the `Lazy<T>` code could be integrated to the class directly for a minuscule gain (since `Lazy<T>` is a class).  I have verified that the ExecutionStrategy does not rely on these properties for proper execution.

Since `As<>` would break the lazy-loaded nature of the Arguments and SubFields properties, a new class has been created called `ResolveFieldContextAs` which handles the mapping without breaking lazy-loaded properties.  The class is marked internal since the `As<>` method is marked internal.

The `ResolveFieldContextSlim` class is specific to the included `ExecutionNode` and `ExecutionContext` classes.  It has been marked as public to allow third party `IExecutionStrategy` implementations to benefit from this code.

Thoughts?

### DIFF
--- a/src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
+++ b/src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
@@ -199,6 +199,25 @@ namespace GraphQL.Tests.Execution
             result.ShouldBe("Test Result");
         }
 
+        [Fact]
+        public void resolveFieldContextAdapter_throws_error_when_null()
+        {
+            Should.Throw<NullReferenceException>(() =>
+            {
+                var adapter = new ResolveFieldContextAdapter<object>(null);
+            });
+        }
+
+        [Fact]
+        public void resolveFieldContextAdapter_throws_error_if_invalid_type()
+        {
+            var context = new ResolveFieldContext() { Source = "test" };
+            Should.Throw<InvalidCastException>(() =>
+            {
+                var adapter = new ResolveFieldContextAdapter<int>(context);
+            });
+        }
+
         enum SomeEnum
         {
             One,
--- a/src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
+++ b/src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
@@ -218,6 +218,14 @@ namespace GraphQL.Tests.Execution
             });
         }
 
+        [Fact]
+        public void resolveFieldContextAdapter_accepts_null_sources()
+        {
+            var context = new ResolveFieldContext();
+            var adapter = new ResolveFieldContextAdapter<int>(context);
+            adapter.Source.ShouldBe(0);
+        }
+
         enum SomeEnum
         {
             One,
--- a/src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
+++ b/src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
@@ -219,11 +219,29 @@ namespace GraphQL.Tests.Execution
         }
 
         [Fact]
-        public void resolveFieldContextAdapter_accepts_null_sources()
+        public void resolveFieldContextAdapter_accepts_null_sources_ref()
         {
             var context = new ResolveFieldContext();
-            var adapter = new ResolveFieldContextAdapter<int>(context);
-            adapter.Source.ShouldBe(0);
+            var adapter = new ResolveFieldContextAdapter<string>(context);
+            adapter.Source.ShouldBe(null);
+        }
+
+        [Fact]
+        public void resolveFieldContextAdapter_accepts_null_sources_nullable()
+        {
+            var context = new ResolveFieldContext();
+            var adapter = new ResolveFieldContextAdapter<int?>(context);
+            adapter.Source.ShouldBe(null);
+        }
+
+        [Fact]
+        public void resolveFieldContextAdapter_throws_error_for_null_values()
+        {
+            var context = new ResolveFieldContext();
+            Should.Throw<ArgumentException>(() =>
+            {
+                var adapter = new ResolveFieldContextAdapter<int>(context);
+            });
         }
 
         enum SomeEnum
--- a/src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
+++ b/src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
@@ -202,7 +202,7 @@ namespace GraphQL.Tests.Execution
         [Fact]
         public void resolveFieldContextAdapter_throws_error_when_null()
         {
-            Should.Throw<NullReferenceException>(() =>
+            Should.Throw<ArgumentNullException>(() =>
             {
                 var adapter = new ResolveFieldContextAdapter<object>(null);
             });
@@ -212,7 +212,7 @@ namespace GraphQL.Tests.Execution
         public void resolveFieldContextAdapter_throws_error_if_invalid_type()
         {
             var context = new ResolveFieldContext() { Source = "test" };
-            Should.Throw<InvalidCastException>(() =>
+            Should.Throw<ArgumentException>(() =>
             {
                 var adapter = new ResolveFieldContextAdapter<int>(context);
             });
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -182,7 +182,7 @@ namespace GraphQL.Execution
 
             try
             {
-                resolveContext = ResolveFieldContextSlim.Create(node, context);
+                resolveContext = new ReadonlyResolveFieldContext(node, context);
 
                 var resolver = node.FieldDefinition.Resolver ?? NameFieldResolver.Instance;
                 var result = resolver.Resolve(resolveContext);
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -178,35 +178,11 @@ namespace GraphQL.Execution
             if (node.IsResultSet)
                 return node;
 
-            ResolveFieldContext resolveContext = null;
+            IResolveFieldContext resolveContext = null;
 
             try
             {
-                var arguments = GetArgumentValues(context.Schema, node.FieldDefinition.Arguments, node.Field.Arguments, context.Variables);
-                var subFields = SubFieldsFor(context, node.FieldDefinition.ResolvedType, node.Field);
-
-                resolveContext = new ResolveFieldContext
-                {
-                    FieldName = node.Field.Name,
-                    FieldAst = node.Field,
-                    FieldDefinition = node.FieldDefinition,
-                    ReturnType = node.FieldDefinition.ResolvedType,
-                    ParentType = node.GetParentType(context.Schema),
-                    Arguments = arguments,
-                    Source = node.Source,
-                    Schema = context.Schema,
-                    Document = context.Document,
-                    Fragments = context.Fragments,
-                    RootValue = context.RootValue,
-                    UserContext = context.UserContext,
-                    Operation = context.Operation,
-                    Variables = context.Variables,
-                    CancellationToken = context.CancellationToken,
-                    Metrics = context.Metrics,
-                    Errors = context.Errors,
-                    Path = node.Path,
-                    SubFields = subFields
-                };
+                resolveContext = ResolveFieldContextSlim.Create(node, context);
 
                 var resolver = node.FieldDefinition.Resolver ?? NameFieldResolver.Instance;
                 var result = resolver.Resolve(resolveContext);
--- a/src/GraphQL/Execution/ResolveFieldContextSlim.cs
+++ b/src/GraphQL/Execution/ResolveFieldContextSlim.cs
@@ -1,12 +1,11 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Threading;
 using GraphQL.Instrumentation;
 using GraphQL.Language.AST;
 using GraphQL.Types;
-using System;
-using System.Collections.Generic;
-using System.Collections.Concurrent;
-using System.Linq.Expressions;
-using System.Text;
-using System.Threading;
 using Field = GraphQL.Language.AST.Field;
 
 namespace GraphQL.Execution
@@ -55,23 +54,21 @@ namespace GraphQL.Execution
     {
         private readonly ExecutionNode _executionNode;
         private readonly ExecutionContext _executionContext;
-        private readonly Lazy<Dictionary<string, object>> _arguments;
-        private readonly Lazy<IDictionary<string, Field>> _subFields;
+        private Dictionary<string, object> _arguments;
+        private IDictionary<string, Field> _subFields;
 
         public ResolveFieldContextSlim(ExecutionNode node, ExecutionContext context)
         {
-            _executionNode = node;
-            _executionContext = context;
-            _arguments = new Lazy<Dictionary<string, object>>(LazyArgumentInitializer, LazyThreadSafetyMode.PublicationOnly);
-            _subFields = new Lazy<IDictionary<string, Field>>(LazySubfieldInitializer, LazyThreadSafetyMode.PublicationOnly);
+            _executionNode = node ?? throw new ArgumentNullException(nameof(node));
+            _executionContext = context ?? throw new ArgumentNullException(nameof(context));
         }
 
-        private IDictionary<string, Field> LazySubfieldInitializer()
+        private IDictionary<string, Field> GetSubFields()
         {
             return ExecutionHelper.SubFieldsFor(_executionContext, _executionNode.FieldDefinition.ResolvedType, _executionNode.Field);
         }
 
-        private Dictionary<string, object> LazyArgumentInitializer()
+        private Dictionary<string, object> GetArguments()
         {
             return ExecutionHelper.GetArgumentValues(_executionContext.Schema, _executionNode.FieldDefinition.Arguments, _executionNode.Field.Arguments, _executionContext.Variables);
         }
@@ -88,7 +85,7 @@ namespace GraphQL.Execution
 
         public IObjectGraphType ParentType => _executionNode.GetParentType(_executionContext.Schema);
 
-        public Dictionary<string, object> Arguments => _arguments.Value;
+        public Dictionary<string, object> Arguments => _arguments ?? (_arguments = GetArguments());
 
         public object RootValue => _executionContext.RootValue;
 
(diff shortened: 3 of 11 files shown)


Shane32:
@sungam3r let me know what you think.  If we are really optimizing this, I think the `Lazy<T>` class I used should be replaced with integrated code that functions the same, to eliminate two more heap allocations per ResolveFieldContextSlim instance.

Also note that if anyone wants to modify the ResolveFieldContext class in a custom implementation, although they can't cast the IResolveFieldContext directly to ResolveFieldContext anymore, they can simply execute `new ResolveFieldContext(context)` which performs essentially the same, dissecting the ExecutionNode and ExecutionContext into discrete mutable fields.

Shane32:
Another thought is that with the `As<>` enhancement, the cost of performing the dictionary lookup to retrieve the ResolveFieldContextSlim constructor function may be almost the same as simply creating a `ResolveFieldContextSlim<object>` and letting the resolver call `As<>` when needed (even if this typically results in two allocations per resolver).  Ignoring the one-time cost of compiling the function of course.

sungam3r:
1. Please remove all not related to this PR commits. I see 23 commits now since 28 November.
2. Add bencmarks test to verify if changes really good.

Shane32:
That's correct -- I started this PR in Nov.  I can consolidate into a single commit if you like.

Can you point me to an example of how the benchmarks work, or another PR that has benchmark testing for me to review?  I've not done that before.

sungam3r:
This https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.Benchmarks/ExecutionBenchmark.cs can be an example. Add your class. Then change to your class in https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.Benchmarks/Program.cs#L14  Note that you need to run bench before and after your changes in release mode. Your class will have only one method marked with  Benchmark attribute.  Well I think you even can to use existing bench.

sungam3r:
This file https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.Benchmarks/GraphQL.Benchmarks.Reference.md is used to track progress

sungam3r:
Officialcdocs https://benchmarkdotnet.org

Shane32:
Changes have all been consolidated into a single commit.  Thanks for the benchmark pointers.  I probably won't have time during the week to look at it, but perhaps this weekend.

src/GraphQL/Execution/ResolveFieldContextSlim.cs
```diff
@@ -0,0 +1,119 @@
...
+            //compile the expression into a function and return it
+            return lambdaExpression.Compile();
+        }
+    }
+
+    public class ResolveFieldContextSlim<T> : IResolveFieldContext<T>
+    {
+        private readonly ExecutionNode _executionNode;
+        private readonly ExecutionContext _executionContext;
+        private readonly Lazy<Dictionary<string, object>> _arguments;
```
sungam3r: Do not use Lazy. Less allocations is better. There are no race conditions since `ResolveFieldContext` is used only by one resolver.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -178,35 +178,11 @@ protected virtual async Task<ExecutionNode> ExecuteNodeAsync(ExecutionContext co
             if (node.IsResultSet)
                 return node;
 
-            ResolveFieldContext resolveContext = null;
+            IResolveFieldContext resolveContext = null;
```
sungam3r: Please remove `using static GraphQL.Execution.ExecutionHelper;`

src/GraphQL/Execution/ResolveFieldContextSlim.cs
```diff
@@ -0,0 +1,119 @@
...
+{
+    public static class ResolveFieldContextSlim
+    {
+        private static readonly ConcurrentDictionary<System.Type, Func<ExecutionNode, ExecutionContext, IResolveFieldContext>> dic =
+            new ConcurrentDictionary<System.Type, Func<ExecutionNode, ExecutionContext, IResolveFieldContext>>();
+
+        public static IResolveFieldContext Create(ExecutionNode node, ExecutionContext context)
+        {
+            //get the type of the source object, and just use typeof(object) when the source is null
+            var sourceType = node.Source?.GetType() ?? typeof(object);
```
sungam3r: You can try to get type from resolver

sungam3r:
> If we are really optimizing this, I think the Lazy<T> class I used should be replaced with integrated code that functions the same, to eliminate two more heap allocations per ResolveFieldContextSlim instance.

Yes.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: It is still in used by 7 other calls in this file.  I did remove excess usings in the two new files.

src/GraphQL/Execution/ResolveFieldContextSlim.cs
Shane32: Done

Shane32:
Ok the performance shows substantial improvement lazy-loading the subfields and arguments fields, but better off using all `ResolveFieldContextSlim<object>` classes (utilizing `As<>` when necessary) compared to inspecting the type of the source to create a properly typed ResolveFieldContextSlim class.  This might (or might not) be due to heavy usage of NameFieldResolver, which does not rely on a typed IFieldResolver.  Of course, it is likely that the NameFieldResolver is the most-used field resolver in actual workload conditions anyway.

```
.NET Core SDK=3.1.100
  [Host]     : .NET Core 3.1.0 (CoreCLR 4.700.19.56402, CoreFX 4.700.19.56404), X64 RyuJIT

==== code as shown in pull request ====

|        Method |        Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 |  Allocated |
|-------------- |------------:|----------:|----------:|---------:|--------:|------:|-----------:|
| Introspection | 2,017.60 us | 14.375 us | 12.743 us | 281.2500 | 93.7500 |     - | 1393.32 KB |
|          Hero |    45.51 us |  0.545 us |  0.483 us |   7.5684 |       - |     - |   30.95 KB |

==== all ResolveFieldContextSlim<object> ====

|        Method |        Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 |  Allocated |
|-------------- |------------:|----------:|----------:|---------:|--------:|------:|-----------:|
| Introspection | 1,877.63 us | 15.444 us | 14.447 us | 255.8594 | 99.6094 |     - | 1318.82 KB |
|          Hero |    44.39 us |  0.185 us |  0.173 us |   7.5073 |       - |     - |   30.76 KB |

==== master branch ====

|        Method |        Mean |     Error |    StdDev |    Gen 0 |    Gen 1 | Gen 2 |  Allocated |
|-------------- |------------:|----------:|----------:|---------:|---------:|------:|-----------:|
| Introspection | 2,347.60 us | 15.996 us | 14.963 us | 363.2813 | 101.5625 |     - | 1847.51 KB |
|          Hero |    45.78 us |  0.261 us |  0.244 us |   7.7515 |        - |     - |   31.82 KB |
```

For the second test, ResolveFieldContextSlim.Create was changed as follows:
``` C#
        public static IResolveFieldContext Create(ExecutionNode node, ExecutionContext context)
        {
            //get the type of the source object, and just use typeof(object) when the source is null
            //var sourceType = node.Source?.GetType() ?? typeof(object);

            //retrieve the constructor
            //var func = dic.GetOrAdd(sourceType, LazyCreate);

            //execute the constructor and return the new ResolveFieldContextSlim<T> (cast as IResolveFieldContext)
            //return func(node, context);
            return new ResolveFieldContextSlim<object>(node, context);
        }
```

Tests were run without debugging, in Release build, with no changes to the Benchmarks project.  Did I do that right?

I'll update the PR based on the findings tomorrow.

Shane32:
Any thoughts before I make changes to the PR?  @sungam3r

sungam3r:
1. Does `ResolveFieldContextSlim` still need to be generic?
2. Maybe rename `ResolveFieldContextSlim` to `ReadonlyResolveFieldContext`.

Shane32:
Not anymore.  That name sounds fine to me.

Shane32:
If there ever is a reason to have a generic version of ReadonlyResolveFieldContext, we can add it at that time.

I've updated the code and verified that the performance benchmarks did not change.  All tests pass.

src/GraphQL/Execution/ResolveFieldContextSlim.cs
Shane32: Not applicable due to performance findings

sungam3r:
`ReadonlyResolveFieldContext` more accurately reflects the nature of the new class. `Slim` in this regard raises more questions.

src/GraphQL/Types/ResolveFieldContextAs.cs
```diff
@@ -0,0 +1,58 @@
+using System.Collections.Generic;
+using System.Threading;
+using GraphQL.Instrumentation;
+using GraphQL.Language.AST;
+
+namespace GraphQL.Types
+{
+    internal class ResolveFieldContextAs<T> : IResolveFieldContext<T>
```
sungam3r: I think that `ReadonlyResolveFieldContext`, `ResolveFieldContextAs` and `ResolveFieldContext` all should be in one namespace.

sungam3r: Also consider to rename `ResolveFieldContextAs` to `ResolveFieldContextAdapter`.

src/GraphQL/Types/ResolveFieldContextAs.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private readonly IResolveFieldContext _baseContext;
+
+        public ResolveFieldContextAs(IResolveFieldContext baseContext)
+        {
+            Source = (T)baseContext.Source; //will throw NullReferenceException or InvalidCastException if there's a problem
```
sungam3r: Could you add a test to verify this?

src/GraphQL/Execution/ResolveFieldContextSlim.cs
sungam3r: Well, it turned out easier.

src/GraphQL/Types/ResolveFieldContextAs.cs
Shane32: Sure I can move it, but realize that the ReadonlyResolveFieldContext is designed exclusively for use with the ExecutionNode and ExecutionContext classes.  The other three are in the GraphQL.Types namespace, so I'll move it there.

Shane32: `ResolveFieldContextAdapter` does sound better

src/GraphQL/Types/ResolveFieldContextAs.cs
Shane32: Done

src/GraphQL/Types/ResolveFieldContextAs.cs
Shane32: Done - moved ReadonlyResolveFieldContext, and renamed ResolveFieldContextAs to ResolveFieldContextAdapter

Shane32:
The ResolveFieldContextAdapter constructor could return a ArgumentNullException (rather than NullReferenceException) when the context is null, however it's an internal class at present, and the null check would be redundant.  I would add the check if ResolveFieldContextAdapter becomes public.  InvalidCastException seems appropriate given the usage.

src/GraphQL/Types/ResolveFieldContextAdapter.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace GraphQL.Types
+{
+    internal class ResolveFieldContextAdapter<T> : IResolveFieldContext<T>
+    {
+        private readonly IResolveFieldContext _baseContext;
+
+        public ResolveFieldContextAdapter(IResolveFieldContext baseContext)
+        {
+            Source = (T)baseContext.Source; //will throw NullReferenceException or InvalidCastException if there's a problem
```
sungam3r: OK, now I think that we should provide more specific error text to describe situation to user, not just NRE/InvalidCast. It needs to be done so that he understands what happened and what he needs to fix.

sungam3r: Maybe `InvalidOperationException` or `ArgumentException`.

sungam3r: After that final change I will be glad to merge this PR. Good work.

Shane32: Makes sense - I didn't think of the fact that the invalid cast exception could be triggered by 'user code'.  The constructor is a bit more wordy now but should produce an explanatory error message.

Shane32:
I added a test for casting a null source via the adapter, as the C# "is" operator returns false in these cases.

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
```diff
@@ -199,6 +199,33 @@ public async Task try_resolve_async_properly_resolves_result()
...
+            });
+        }
+
+        [Fact]
+        public void resolveFieldContextAdapter_accepts_null_sources()
+        {
+            var context = new ResolveFieldContext();
+            var adapter = new ResolveFieldContextAdapter<int>(context);
+            adapter.Source.ShouldBe(0);
+        }
```
sungam3r: I think that behavior should be different for value and ref+nullable types.

```c#
 [Fact]
        public void resolveFieldContextAdapter_accepts_null_sources_for_reference_types()
        {
            var context = new ResolveFieldContext();
            var adapter = new ResolveFieldContextAdapter<string>(context);
            adapter.Source.ShouldBeNull();
        }
 [Fact]
        public void resolveFieldContextAdapter_accepts_null_sources_for_nullable_types()
        {
            var context = new ResolveFieldContext();
            var adapter = new ResolveFieldContextAdapter<int?>(context);
            adapter.Source.ShouldBeNull();
        }
 [Fact]
        public void resolveFieldContextAdapter_throw_on_null_sources_for_value_types()
        {
            var context = new ResolveFieldContext();
            Should.Throw<ArgumentException>(() => new ResolveFieldContextAdapter<int>(context));
        }
```

src/GraphQL/Types/ResolveFieldContextAdapter.cs
```diff
@@ -0,0 +1,65 @@
...
+    internal class ResolveFieldContextAdapter<T> : IResolveFieldContext<T>
+    {
+        private readonly IResolveFieldContext _baseContext;
+
+        public ResolveFieldContextAdapter(IResolveFieldContext baseContext)
+        {
+            _baseContext = baseContext ?? throw new ArgumentNullException(nameof(baseContext));
+
+            if (baseContext.Source is null)
+                Source = default(T);
```
sungam3r: For value types this code will be misleading, converting null to default(valuetype). See my another comment about tests.

Shane32: True, but I'm not sure what you're suggesting.  Shall I add a comment?  Writing `if (typeof(T).IsValueType)` would have a significant performance impact when using typed IFieldResolvers.

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
Shane32: I'm open to suggestions as to how best to accomplish this.

Shane32:
To throw an exception for value types (but not nullable value types), you'd probably need code like this:

``` C#
        public ResolveFieldContextAdapter(IResolveFieldContext baseContext)
        {
            _baseContext = baseContext ?? throw new ArgumentNullException(nameof(baseContext));

            if (baseContext.Source is null)
            {
                var typeofT = typeof(T);
                if (typeofT.IsValueType)
                {
                    if (typeofT.IsNullable())
                    {
                        Source = default(T);
                    }
                    else
                    {
                        throw new ArgumentException("baseContext.Source is not of type " + typeof(T).Name, nameof(baseContext));
                    }
                }
                else
                {
                    Source = default(T);
                }
            }
            else if (baseContext.Source is T source)
                Source = source;
            else
                throw new ArgumentException("baseContext.Source is not of type " + typeof(T).Name, nameof(baseContext));
        }
```

Seems like it's going a bit too far for the purpose of this class.

Shane32:
Note that IsNullable is a GraphQL extension function defined as follows:
``` C#
        public static bool IsNullable(this Type type)
        {
            return type == typeof(string) || (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>));
        }
```
It could be shortened a little of course and inlined.

src/GraphQL/Types/ResolveFieldContextAdapter.cs
sungam3r: I am talking about runtime check not about comment.

> significant performance impact

Why do you think so? You can check this with benchmarks. I generally think that the JIT should expand this construction into a constant and delete the entire block of code or execute it unconditionally.

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
sungam3r: With runtime check inside ResolveFieldContextAdapter.cs

sungam3r:
> Seems like it's going a bit too far for the purpose of this class.

Why? The adapter must adapt, but not allow accidental conversion from `null` to `0`.

Shane32:
I’ll take another look at it. Maybe there’s another way to achieve the desired results without the performance hit.

Shane32:
I put the native cast in a try..catch block.  Benchmarks shows that to be 6 times faster than the above code with same results.  Added tests to verify operation of reference, value, and nullable value types.

sungam3r:
:) Yeah, this is really the best solution, it is simpler and more efficient. I myself resort to the strategy of "letting potentially dangerous code execute and just catch the error," but in this case, for some reason, I forgot about this possibility.

Issue #1310, opened 2019-09-20
ParallelExecution in queries


sergeyshaykhullin:
## Problem
I am using EFCore + DataLoaders and trying to query
```gql
{
  root {
    nested1 {
      nomatter
    }
    nested2 {
      nomatter 
    }
  }
}
```
Each nested node using DataLoader with access to DbContext from UserContext from HttpContext scope. Executing both in parallel will fail with DbContext ```A second operation started on this context before a previous operation completed```. 
I found https://github.com/graphql-dotnet/graphql-dotnet/issues/889 and https://github.com/graphql-dotnet/graphql-dotnet/pull/891 about parallel execution in same nested nodes and don't get why it is. DbContext is not working with parallel access =(

## Summary
I found 2 workarounds
- Use DbContextPool and get pooled DbContext only for query nodes
- Write custom DocumentExecuter

But they both is poor. Why i should write custom document executor or fall down in performance just to integrate EFCore? =/

Are there any better solutions?

## Ideas
Maybe we need to move configuration of execution strategies to Options?
```csharp
options.ExecutionStrategies.Query = new ...();
options.ExecutionStrategies.Mutation= new ...();
options.ExecutionStrategies.Subscription = new ...();
```
With default implementation to get from options or override SelectExecutionStrategy in derive Executer

sungam3r:
This question has already been raised several times. The problem is that you use `DbContext` in a way that it is not intended to be used. You cannot use the **same** `DbContext` for different resolvers. Use pooling. Alse see #648 - it discusses general dependency injection issues.

sungam3r:
> Why i should write custom document executor

You don’t have to do this at all. The root of the problem is not this, but the choice of the wrong lifetime of the used `DbContext`.

floyd-may:
I ran into the same issue. In my particular use case, having the `DbContext` lifetime set to `Scoped` is the right choice for broader reasons outside just GraphQL. Given that parallel execution is primarily a performance optimization, would it be better to consider parallel execution as something to opt-in to rather than opt-out? It seems that serial is the safer default.

sungam3r:
Parallel execution of query fields is the part of GraphQL [spec](https://graphql.github.io/graphql-spec/June2018/#sec-Normal-and-Serial-Execution).

floyd-may:
[...]

Issue #1191, opened 2019-06-17
DataLoader and SerialExecutionStrategy not bulking


kYann:
## Description

When using DataLoader and SerialExecutionStrategy, the DispatchAllAsync is executed after each nodes, loosing the benefit of dataloader on list.

### Steps to reproduce

Set SerialExecutionStrategy for Query.
Return List of item where one of the Field is using the BatchLoader
Put a breakpoint in your fetch function and see only one id in the list.

### Expected result

The fetch function should be called after all the nodes have been executed

### Actual result

Fetch function is called after each node have been executed

### Environment

Windows, GraphQL 2.3.0

Shane32:
This is an important feature, due to the fact that there is no Dependency Injection scope management when running the ParallelExecutionStrategy, preventing the parallel strategy from being used with class libraries which rely upon scoped services.  Hence, the SerialExecutionStrategy is often a requirement, preventing the use of DataLoaders at present.

The DataLoader mechanism should be redesigned to support proper operation within a SerialExecutionStrategy.  I suggest something like this (when run with the SerialExecutionStrategy):

1. All queued nodes that are not data loaders should be run
2. All data loader nodes should be prepped
3. All data loaders should be run
4. All data loader nodes should be resolved
5. Loop above steps

This is not possible with the current DataLoader design, and might unfortunately require that the DataLoader functionality be more tightly integrated to graphql.

sebastienros:
Also blocking for OrchardCore. We need to use Serial Execution (for all the bad things that can happen with the parallel one), and then data loaders can't be used, which is a huge limitation in terms of perf.

Issue #1042, opened 2019-03-05
How to handle multiple data loader calls?


magnusottosson:
Hi,

I have some code like this:

```
Field<ProcductGraphType, Product>()
                .Name("accessories")
                .ResolveAsync(async ctx =>
                {

                    var relationShipLoader = accessor.Context.GetOrAddLoader("RelationShipLoader", GetAccessoriesRelationShip);

                    var relationShips = await relationShipLoader.LoadAsync("product1");
                    var productIds = relationShips.Select(x => x.Id);
               
                    var productsLoader = accessor.Context.GetOrAddLoader("ProductsLoader", GetProducts);

                    var products = await relationShipLoader.LoadAsync(productIds);

                    return products;
                });
```

I have read that you have to create all task before calling await but how do I handle scenarious like above where I need the result from the first loader as input for the second loader?

In the above case the second await will never execute.

joemcbride:
Admittedly I'm no expert on the DataLoader feature, though I'm fairly certain this scenario is not currently supported by it.

magnusottosson:
I guess that the dataloader implementation in GraphQL does not support this because the task are only triggered once in this scope. Previously I have been working with facebooks javascript implementation and it supports using multiple dataloaders and you can even chain dataloaders so one loader uses another loader inside itself.

The use case above seems like a common scenario that should be supportet? Maybe it's just me that dont understand how to solve this?

My solution was to create another dataloader that performs all the calls it seend inside itself. This works but the loader will be very specific to this use case and wont combine requests with others where only the relationship is needed etc...

jpabeem:
I am currently facing the same problem, only difference is that I am encountering it when querying two fields which both use DataLoaders within the same request. 

@magnusottosson care to elaborate on your solution? Thanks!

johnrutherford:
@magnusottosson  Yeah, it's not supported. Async stuff works a lot differently in node.js than in .NET. 

I've spent a lot of time working on this problem. It would be possible to do this with a separate thread that is constantly checking for more work queued for a data loader. The problem is that there would be no way to know when no more requests will be added. So it wouldn't be able to batch optimally.

Maybe we could have two implementations to choose from...? That might be more confusing, though.

@jpabeem, could you elaborate and share some code? What problem are you having? There's no reason the DataLoader shouldn't work with multiple fields.

PR #1508, opened 2020-01-18
Add failing DataLoader tests

Shane32:
This adds two failing tests to address the following bugs:
- DataLoaders do not batch load data when run with a SerialExecutionStrategy
- Field resolve functions that await before returning the DataLoaders's LoadAsync function result in deadlock.

The tests are set to skip execution so they will not run at present, so this PR can be merged if desired.

Relevant issues are:
#1191 DataLoader and SerialExecutionStrategy not bulking
#945 DataLoader deadlock with multiple awaits
#837 Should the DataLoader work inside a Mutation
Also reference:
#1310 ParallelExecution in queries

I suggest we define the tests; then we can work on solutions.

As #945 is tagged for 3.0 release, I will provide a separate PR suggesting a solution.

**UPDATE**

Added test for supporting chained data loaders within a resolver:
#1042 How to handle multiple data loader calls?

### DIFF
--- a/src/GraphQL.DataLoader.Tests/DataLoaderQueryTests.cs
+++ b/src/GraphQL.DataLoader.Tests/DataLoaderQueryTests.cs
@@ -289,7 +289,6 @@ mutation {
             ordersMock.Verify(x => x.GetOrdersByUserIdAsync(new[] { 1, 2 }, default), Times.Once);
             ordersMock.Verify(x => x.GetItemsByOrderIdAsync(new[] { 1, 2, 3, 4 }), Times.Once);
             ordersMock.VerifyNoOtherCalls();
-
         }
 
         [Fact]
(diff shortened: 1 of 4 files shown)


src/GraphQL.DataLoader.Tests/DataLoaderQueryTests.cs
```diff
@@ -139,6 +167,131 @@ public void TwoLevel_MultipleResults_OperationsAreBatched()
...
+    }]
+}
+");
+            usersMock.Verify(x => x.GetAllUsersAsync(default), Times.Once);
+            usersMock.VerifyNoOtherCalls();
+
+            ordersMock.Verify(x => x.GetOrdersByUserIdAsync(new[] { 1, 2 }, default), Times.Once);
+            ordersMock.Verify(x => x.GetItemsByOrderIdAsync(new[] { 1, 2, 3, 4 }), Times.Once);
+            ordersMock.VerifyNoOtherCalls();
+
```
sungam3r:
```suggestion
```

Shane32:
Note that these tests need to be modified due to the change of the JSON serializer in use.  The required changes have already been made in #1511 but not here yet.  (If #1511 is pulled, then this issue can be closed, as #1511 includes the relevant tests.)

Issue #1449, opened 2019-12-05
Add GraphQL.SystemTextJson project


sungam3r:
Relates to #1148, #1448

sungam3r:
https://aloiskraus.wordpress.com/2019/09/29/net-serialization-benchmark-2019-roundup/

benmccallum:
Let me know if you need help, or if we need to do something in the server project. Happy to split the workload.

benmccallum:
Perhaps we should create an issue over on that repo and lay out the plan?

sungam3r:
I will not be engaged in this task in near future. There are several other bugs on my todo list that need to be worked on. I also need to open source csharp graphql generator. All this will take quite a while. You can work on PR if you wish.

benmccallum:
@sungam3r, I'll do this in the next day or so. Please assign the issue to me. Cheers

sungam3r:
Fixed by #1512 , thanks to @benmccallum .

PR #1510, opened 2020-01-19
bump GraphQL-Parser to 5.0.1

sungam3r:
The new version of the GraphQL-Parser allocates fewer objects and requires additional null checks.

Before:

|        Method |        Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |------------:|----------:|----------:|---------:|--------:|------:|----------:|
| Introspection | 2,665.72 us | 15.886 us | 12.403 us | 203.1250 | 58.5938 |     - | 605.74 KB |
|          Hero |    62.84 us |  0.462 us |  0.410 us |   9.5215 |       - |     - |  14.61 KB |

After:

|        Method |        Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |------------:|----------:|----------:|---------:|--------:|------:|----------:|
| Introspection | 2,517.18 us | 34.878 us | 32.625 us | 183.5938 | 58.5938 |     - | 594.49 KB |
|          Hero |    54.87 us |  0.696 us |  0.581 us |   8.9722 |       - |     - |  13.84 KB |

PR #1512, opened 2020-01-21
Add GraphQL.SystemTextJson

benmccallum:
Addresses #1449.
* Introduces GraphQL.SystemTextJson project with its own implementation of an `IDocumentWriter`.
* Introduces an `ObjectDictionaryConverter` (thanks Pekkah) and associated extension method/s off `string` to facilitate getting a real dictionary of real values from a JSON string.
* Introduces a single `ToExecutionResult` ext method, rather than doing `JObject.Parse` in many unit test base classes to form that result.
* Side-effect of now using System.Text.Json, all our JSON strings (for expected values/inputs) need to be valid JSON (which IMO is a good thing. See [here](https://github.com/dotnet/runtime/issues/31005)):
  * Double quotes for property names and string values
  *  No trailing commas (we could support this if we wanted, as it's configurable, but I don't think we should)
  * Seems to encode special characters more strictly into their \u1234 format.
* Converts all usage of Newtonsoft.Json in tests and test harness projects over to System.Text.Json. The only remaining usage of Newtonsoft.Json is now in that specific serializer project (GraphQL.NewtonsoftJson).

**Breaking changes**
* Since System.Text.Json is very strict to JSON standard, people who were getting away with sending bad JSON for variables will be affected.

Since that'd be rare and this is a new package though, seems fair.


benmccallum:
Issues so far:
* `ExecutionResultJsonConverter`, according to docs, you can't pass the serializer options down recursively, so potentially that could cause issues if the defaults aren't acceptable. 
* There's a lot of extension methods in StringExtensions for ToInputs, ToDictionary, etc. - most of them could be marked `private` in the Newtonsoft.Json project, and hopefully `JsonSerializer.Deserialize<Inputs>(json)` will work, but we'll find out. It certainly worked in the server project.
* The Harness project, I guess I'll need to multi-target it and update it like I did in the server project so that we get test coverage of legacy and new serializers.

src/Directory.Build.props
```diff
@@ -5,4 +5,8 @@
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageIcon>logo.64x64.png</PackageIcon>
   </PropertyGroup>
+
+  <PropertyGroup>
+    <IsNetCore3OnwardsTarget>$(TargetFramework.StartsWith('netcoreapp3.'))</IsNetCore3OnwardsTarget>
```
sungam3r: Does it work? I think that properties from Directory.Build.props are imported before TargetFramework defined in csproj.

benmccallum: I think that's what I was doing over in the server sln, but definitely worth checking. It seemed to pull in the right references in the Tests project I used it in.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -31,11 +31,11 @@ public static Inputs ToInputs(this JObject obj)
         }
 
         /// <summary>
-        /// Converts a JSON formatted string into a the dictionary.
+        /// Converts a JSON formatted string into a dictionary.
         /// </summary>
         /// <param name="json">The json.</param>
         /// <returns>Returns a <c>null</c> if the object cannot be converted into a dictionary.</returns>
-        public static Dictionary<string, object> ToDictionary(this string json)
+        private static Dictionary<string, object> ToDictionary(this string json)
```
sungam3r: aren't these methods used by the server project?

benmccallum: I don't think so anymore. Both deserializers just deserialize straight into `Inputs` which seemed to work, but I think I need to add some more tests around that.

You're right though that if we've exposed them as `public` then potentially we need to keep that going here and in the SystemTextJson project.

src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
+    <RootNamespace>GraphQL.SystemTextJson</RootNamespace>
+    <Description>System.Text.Json serializer for GraphQL.NET</Description>
+    <PackageTags>GraphQL;json</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="..\..\assets\logo.64x64.png" Pack="true" PackagePath="\"/>
```
sungam3r: Move to `Directory.Build.props`?

src/GraphQL.SystemTextJson/SchemaExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Threading.Tasks;
+using GraphQL.Types;
+
+namespace GraphQL.SystemTextJson
+{
+    public static class SchemaExtensions
+    {
+        public static Task<string> ExecuteAsync(this ISchema schema, Action<ExecutionOptions> configure)
+        {
+            var documentWriter = new DocumentWriter(new JsonSerializerOptions() { WriteIndented = true });
```
sungam3r:
```suggestion
            var documentWriter = new DocumentWriter(new JsonSerializerOptions { WriteIndented = true });
```

src/GraphQL.Tests/GraphQL.Tests.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="../Tests.props" />
   
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>netcoreapp3.1;netstandard2.0</TargetFrameworks>
```
sungam3r: But can tests work on netstandard?

benmccallum: Will be undoing this and taking a diff approach

src/GraphQL.SystemTextJson/SchemaExtensions.cs
benmccallum: This will be deleted when I commit next.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
benmccallum: Have made just the `GetValue` method private for now, will review this again later.

benmccallum:
@sungam3r @joemcbride , one issue I'm struggling with at the moment is how to get good test coverage of this new `IDocumentWriter` across the existing tests. There's a number of TestBase classes that `new` up a `NewtonsoftJson.DocumentWriter`, but obviously we want to now test both.

I started by removing all the `new`ing up of these across all base classes and then having the utility methods accept a parameter for the `IDocumentWriter`, therefore all the specific referencing of one goes away. And then am currently working my way through all the `Fact` methods, converting them to `Theory` with test data for each document writer, but it seems weird to have methods completely unrelated to IDocumentWriter becoming theories with that as test data, so I'm having second thoughts.

The alternatives are:
1. Make the base class utility methods loop over a configured list of `IDocumentWriter` instances, effectively doing their assertion `n` times. Downside being, it'll just be in one test in the Test Explorer, but upside being we don't need to put Theory everywhere.
2. Only use the SystemTextJson one for most tests and then have specific IDocumentWriter tests that cover each implementation and validate them. Downside being, we'd potentially lose some coverage on the old serializer.

It's an interesting problem. What do you guys think?

sungam3r:
https://stackoverflow.com/questions/51852183/xunit-theory-like-data-at-the-level-of-class-not-method   maybe this helps

benmccallum:
Yea I essentially need a mixture of that (a context/data provider at the test base class levels, to inject the writers in) + the "iteration" that `[Theory, XyzData]` provides.

joemcbride:
I think that the simpler approach in `2` is fine for a first iteration, and do `1` if it becomes a problem.

benmccallum:
No worries. Couldn't get on to this end of this week due to SameSite cookie fun, and I'm on hols next week until Thursday, then I'll hopefully get back on this.

I think 2, although a little risky for the old serializer in future changes, is fwd looking. The old serializer is essentially proven, so unless features are added that require weird serialization, we should be ok. Then worst case we do 1 in some places like you said.

src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
benmccallum: I've just copied what NewtonsoftJson project did here.

sungam3r: Ok

sungam3r:
I look forward to the moment when I can make a review of the final version. This feature blocks me from migrating to the netcore3.1 :)

benmccallum:
Yea, it'll be great to get this done. I think we're in a good place with the tests now, though I'll prob write a few more at the end. Now I need to work through all the tests that are failing. 

If anyone has any ideas for how to implement `GraphQL.SystemTextJson.StringExtensions.GetValue(this JsonElement jsonElement)`, I'd appreciate an assist. That should make most of the tests pass.

benmccallum:
Compared to NewtonsoftJson implementation, whose `JValue.Value` property was `object?`, I'm not sure how to extract the real type out of the `JElement`, specifically enums and different "Number" sub types.

pekkah:
@benmccallum you could check this https://github.com/pekkah/tanka-graphql/blob/master/src/graphql.server.links/DTOs/ObjectDictionaryConverter.cs

It works with dictionaries of objects where object can be array, value type or another object. It's dirty but until s.t.json provides better alternative I couldn't think of another way.

benmccallum:
Thanks @pekkah ! Looks like that will indeed be useful. Glad to see you couldn't find a better way than a series of `TryParseXyz`. 

The other fun part is that a bunch of our tests use `JObject.Parse` and set that on a fresh `ExecutionResult.Data` to be used for the expected test result, which works when it's written out by the Newonsoft.Json serializer, but not the System.Text.Json one which doesn't recognise a `JObject` and `JElement`. Essentially we need the ability to parse to/form JSON and `object` for both serializers, but without the serializer's specific implementation types under the hood. 

Not sure if I wanna solve that in this PR though or not. I guess it's just calling a variation of the `ToInputs` extension method we have off string though in each serializer.

sungam3r:
About parsing numbers - https://github.com/dotnet/corefx/issues/39314

StefanSchoof:
Here is how the cosmos sdk solves this problem: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/73c6df78bb732a0ab26516c24bc8aea8c97f4011/Microsoft.Azure.Cosmos/azuredata/Serializer/TextJsonObjectToPrimitiveConverter.cs#L89

benmccallum:
Hey @pekkah, one issue I had with your `ObjectDictionaryConverter` is that the `ReadArray` returning itself a `ReadArray` or `ReadObject` result wouldn't immediately evaluate those results, so I'd find later in usage of the dictionary (e.g. to write it back to a string), that it'd fail as the `JsonDocument` was disposed.

I just `.ToArray()` now and it works, but wanted to mention it in case it could cause you issues or there's a downside to what I'm doing.
* `yield return ReadDictionary(item, options).ToArray();`
* `yield return ReadArray(item, options).ToArray();`

benmccallum:
Getting damn close. Just a couple of failing tests around big ints and then I'll start making sure the server project has all it needs in terms of those extension methods. Might be able to get this ready for review by today.

sungam3r:
I'm already mentally preparing to spend a few hours reviewing this.

benmccallum:
Hey @sungam3r, do you mind having a look at this last failing test? I updated the inputs parsing in SystemTextJson.StringExtensions to try parse a BigInteger too, which fixed the previous failing test, but for some reason the output is different between the writers. Looks like newtonsoft wrote BigInteger as `1.2345678901234568E+18` but SysTextJson writes it as the full number `1234567890123456789`. Not even sure if that's an issue tbh.

benmccallum:
Seems like it might be an issue as JS console tells me they are different

sungam3r:
I think the new behavior 1234567890123456789 is right.

benmccallum:
I think I've actually got a mistake in the expected value. It's my bad. Almost sorted it.

sungam3r:
I see failed tests.

src/GraphQL.ApiTests/ApiApprovalTests.cs
```diff
@@ -12,7 +11,7 @@ public class ApiApprovalTests
     {
         [Theory]
         [InlineData(typeof(IGraphType))]
-        [InlineData(typeof(DocumentWriter))]
+        [InlineData(typeof(IDocumentWriter))]
```
sungam3r: Why? This is a type from the same assembly. There should be two types from two serializers assemblies.

src/GraphQL.SystemTextJson/StringExtensions.cs
```diff
@@ -0,0 +1,48 @@
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace GraphQL.SystemTextJson
+{
+    public static class StringExtensions
+    {
+        private static readonly JsonSerializerOptions _jsonOptions = new JsonSerializerOptions()
```
sungam3r:
```suggestion
        private static readonly JsonSerializerOptions _jsonOptions = new JsonSerializerOptions
```

src/GraphQL.SystemTextJson/StringExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="json">The json.</param>
+        /// <returns>Dictionary.</returns>
+        public static ExecutionResult ToExecutionResult(this string json, ExecutionErrors errors = null)
+        {
+            object expected = null;
+            if (!string.IsNullOrWhiteSpace(json))
+            {
+                expected = json.ToDictionary();
+            }
+            return new ExecutionResult { Data = expected, Errors = errors };
```
sungam3r: `expected` - copy/paste artifact

```c#
 return new ExecutionResult { Data = string.IsNullOrWhiteSpace(json) ? null : json.ToDictionary(), Errors = errors };
```

src/GraphQL.Tests/BasicQueryTestBase.cs
```diff
@@ -27,7 +26,7 @@ public class BasicQueryTestBase
             IEnumerable<IValidationRule> rules = null)
         {
             var queryResult = CreateQueryResult(expected);
-            return AssertQuery(schema, query, queryResult, inputs, root, userContext, cancellationToken, rules);
+            return AssertQuery(schema, query, queryResult, inputs, root,  userContext, cancellationToken, rules);
```
sungam3r:
```suggestion
            return AssertQuery(schema, query, queryResult, inputs, root, userContext, cancellationToken, rules);
```

src/GraphQL.Harness.Tests/DocumentWritersTestData.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace GraphQL.Harness.Tests
+{
+    public class DocumentWritersTestData : IEnumerable<object[]>
+    {
+
```
sungam3r:
```suggestion
```

src/GraphQL.Tests/Introspection/IntrospectionResult.cs
```diff
@@ -2,7 +2,1152 @@ namespace GraphQL.Tests.Introspection
 {
     public class IntrospectionResult
     {
-        public static readonly string Data =
+        public static readonly string DataWhenSystemTextJson =
```
sungam3r: What is the difference between `DataWhenSystemTextJson` and `DataWhenNewtonsoftJson` ?

src/GraphQL.Harness.Tests/SuccessResultAssertion.cs
```diff
@@ -1,40 +1,46 @@
+using System.Linq;
 using Alba;
-using GraphQL.NewtonsoftJson;
+using GraphQL.SystemTextJson;
```
sungam3r:
```suggestion
```

src/GraphQL.SystemTextJson/DocumentWriter.cs
```diff
@@ -0,0 +1,35 @@
...
+            _options = options;
+
+            if (!_options.Converters.Any(c => c.CanConvert(typeof(ExecutionResult))))
+            {
+                _options.Converters.Add(new ExecutionResultJsonConverter());
+            }
+        }
+
+        public async Task WriteAsync<T>(Stream stream, T value)
+            => await JsonSerializer.SerializeAsync(stream, value, _options).ConfigureAwait(false);
```
sungam3r:
```c#  
public Task WriteAsync<T>(Stream stream, T value) => JsonSerializer.SerializeAsync(stream, value, _options);
```

src/GraphQL.Tests/DocumentWritersTestData.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace GraphQL.Tests
+{
+    public class DocumentWritersTestData : IEnumerable<object[]>
+    {
+
```
sungam3r:
```suggestion
```

src/GraphQL.Tests/Bugs/Bug271DateTimeInStringGraphTypeTests.cs
```diff
@@ -14,7 +14,7 @@ public void supports_date_inputs_inside_string_fields()
                             }";
 
             var expected = "{ \"fieldWithObjectInput\": \"{\\\"a\\\":\\\"2017-01-27T15:19:53.000Z\\\",\\\"b\\\":[\\\"bar\\\"],\\\"c\\\":\\\"baz\\\"}\" }";
-            var inputs = "{'input': {'a':'2017-01-27T15:19:53.000Z', 'b':['bar'], 'c': 'baz'} }".ToInputs();
+            var inputs = "{ \"input\": { \"a\": \"2017-01-27T15:19:53.000Z\", \"b\": [\"bar\"], \"c\": \"baz\"} }".ToInputs();
```
sungam3r:
```suggestion
            var inputs = @"{ ""input"": { ""a"": ""2017-01-27T15:19:53.000Z"", ""b"": [""bar""], ""c"": ""baz""} }".ToInputs();
```

src/GraphQL.Tests/Serialization/SerializationTestBase.cs
```diff
@@ -0,0 +1,8 @@
+namespace GraphQL.Tests.Serialization
+{
+    public abstract class SerializationTestBase
```
sungam3r: Remove?

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -2,6 +2,7 @@
 using GraphQL.Utilities.Federation;
 using Xunit;
 using Newtonsoft.Json.Linq;
```
sungam3r: remove?

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -70,8 +71,8 @@ scalar Uri
   username: String!
 }
 ";
-
-            var expected = $@"{{ '_service': {{ 'sdl' : '{sdl}' }}}}";
+
```
sungam3r:
```suggestion
```

src/GraphQL.Harness.Tests/GraphQL.Harness.Tests.csproj
```diff
@@ -11,6 +11,8 @@
 
   <ItemGroup>
     <ProjectReference Include="..\GraphQL.Harness\GraphQL.Harness.csproj" />
```
sungam3r: GraphQL.Harness contains `<ProjectReference Include="..\GraphQL.NewtonsoftJson\GraphQL.NewtonsoftJson.csproj" />`. Need to be removed either from here or from there.

sungam3r:
I will continue the review tomorrow.

src/GraphQL.ApiTests/ApiApprovalTests.cs
benmccallum: I guess I'm confused about what this test does. My thinking was it was for the public api we care about, but I guess it's any public methods in any packages. Will add both. Pretty sure this is the failing test too, but I don't understand the reason.

src/GraphQL.SystemTextJson/StringExtensions.cs
benmccallum: Lol, I'll never prefer the short hand, I want to see the ctor call! But I'll conform :P

src/GraphQL.Tests/Introspection/IntrospectionResult.cs
benmccallum: S.T.Json turns ` into \u0060 (from memory) and ' into another \u1234 and so on. It's a painful difference, but wasn't sure what else to do, as this is a pretty nice, comprehensive test for the serializers that I wanted to do a Theory for to cover both.

src/GraphQL.Harness.Tests/SuccessResultAssertion.cs
benmccallum: This is needed for the `actualResultJson.ToDictionary();` as that is an extension method. It's not a great name for that method though, but I think that was legacy and I'd keep it as is. Certainly a bit confusing.

src/GraphQL.ApiTests/ApiApprovalTests.cs
sungam3r: This test helps you see changes in the public api (no changes/minor/breaking) and follow SemVer.

>  I guess it's any public methods in any packages

Yes, it should show public APIs of our three packages.

src/GraphQL.Tests/Serialization/SerializationTestBase.cs
benmccallum: Do you think it's worth implementing any tests here? I'm not sure whether to bother or just wait for any potential issues to come up as per convo with @joemcbride earlier.

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
benmccallum: Looks like there's a JObject.Parse and then use of `SelectToken` in here that I'd have to rewrite. There's no SelectToken equiv in S.T.Json, but I might be able to just read it outta the strongly-typed `ExecutionResult` instead. Let me see.

src/GraphQL.Tests/Serialization/SerializationTestBase.cs
sungam3r: There is simply no point in this blank stub.

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
benmccallum: It's not pretty, but I've fixed this up. I've also removed all other uses of newtonsoft.json in other tests, the only last remaining usage is in the test harness middleware, which I'll convert over now since it's a 3.1 project.

src/GraphQL.NewtonsoftJson/SchemaExtensions.cs
```diff
@@ -1,14 +0,0 @@
-using System;
-using System.Threading.Tasks;
-using GraphQL.Types;
-
-namespace GraphQL.NewtonsoftJson
-{
-    public static class SchemaExtensions
-    {
-        public static Task<string> ExecuteAsync(this ISchema schema, Action<ExecutionOptions> configure)
```
sungam3r: Was this method deleted or moved? I remember that @joemcbride insisted that this method should remain.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.received.txt
```diff
@@ -0,0 +1,2895 @@
+namespace GraphQL.Builders
```
sungam3r: This file should not be pushed to git.

src/GraphQL.NewtonsoftJson/SchemaExtensions.cs
benmccallum: Removed. Was being used as a convenience in tests, but after refactoring off, didn't need it and didn't want to make a S.T.Json one. Do we really need it @joemcbride ?

src/GraphQL.Harness.Tests/GraphQL.Harness.Tests.csproj
benmccallum: Removed these refs. STJson one is now inherited from harness project and is the only one tested there now.

src/GraphQL.NewtonsoftJson/SchemaExtensions.cs
sungam3r: See this thread https://github.com/graphql-dotnet/graphql-dotnet/pull/1448#discussion_r354562378

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.received.txt
benmccallum: Ugh. It needs to be ignored - remembered to not push it 100 times except this one. Can I add it to .gitignore?

src/GraphQL.ApiTests/ApiApprovalTests.cs
benmccallum: It's still failing, but it's some kind of weird exception where it can't even exec the test. Will look after lunch but might need some help if you have a sec

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.received.txt
benmccallum: Ignored it and updated rule, though not sure why previous one wasn't ignoring it for me. Seemed right.

sungam3r: When the api test is successful, this file is deleted automatically.

src/GraphQL.SystemTextJson/DocumentWriter.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            _options = options;
+
+            if (!_options.Converters.Any(c => c.CanConvert(typeof(ExecutionResult))))
+            {
+                _options.Converters.Add(new ExecutionResultJsonConverter());
+            }
+        }
+
+        public Task WriteAsync<T>(Stream stream, T value) => JsonSerializer.SerializeAsync(stream, value, _options);
```
sungam3r: Much simpler implementation compared to Newtonsoft.Json

sungam3r:
49 of 65 changes files checked, take a break and continue

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            var data = result.Data;
+
+            if (result.Errors?.Count > 0 && data == null)
+            {
+                return;
+            }
+
+            writer.WritePropertyName("data");
+            JsonSerializer.Serialize(writer, data);
```
sungam3r: Have a test when `data` is `null` and `result.Errors.Count == 0`?

src/GraphQL.Tests/Serialization/SerializationTestBase.cs
benmccallum: Yea it was a todo I had to write some tests. There's a few custom things happening in the serialization projects with custom converters, etc. Anyway, I've renamed this to `DocumentWriterTests` and done a super basic test and if needed we can add to that foundation in the future where we are lacking coverage.

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
```diff
@@ -0,0 +1,138 @@
+using System;
+using System.Linq;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace GraphQL.SystemTextJson
+{
+    public class ExecutionResultJsonConverter : JsonConverter<ExecutionResult>
```
sungam3r: I am pleasantly surprised how this implementation looks like existing one.

src/GraphQL.Tests/Serialization/SerializationTestBase.cs
sungam3r: Good

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+{
+    /// <summary>
+    /// A custom JsonConverter for converting into a dictionary of objects of their real underlying type.
+    /// </summary>
+    /// <remarks>
+    /// With thanks to @pekkah who shared this from tanka-graphql.
+    /// </remarks>
+    public class ObjectDictionaryConverter : JsonConverter<Dictionary<string, object>>
+    {
+        private readonly JsonSerializerOptions _internalOptions = new JsonSerializerOptions()
```
sungam3r:
```suggestion
        private readonly JsonSerializerOptions _internalOptions = new JsonSerializerOptions
```

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+                        resultValue = ReadArray(value).ToList();
+                        break;
+                    default:
+                        throw new InvalidOperationException($"Unexpected value kind: {value.ValueKind}");
+                }
+
+                result[key] = resultValue;
+            }
+
+            if (!result.Any())
```
sungam3r: I propose to move `var result = new Dictionary<string, object>();` allocation from the beginning of this method:
```c#
 if (result == null)
  result = new Dictionary<string, object>();
result[key] = resultValue;
```

and remove `if (!result.Any()) return null;`

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+
+        private object ReadNumber(JsonElement value)
+        {
+            if (value.TryGetInt32(out var i))
+                return i;
+            else if (value.TryGetInt64(out var l))
+                return l;
+            //else if (value.TryGetUInt64(out var ui))
```
sungam3r: ?

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+                        yield return item.GetString();
+                        break;
+                    case JsonValueKind.Null:
+                        yield return null;
+                        break;
+                    case JsonValueKind.Array:
+                        yield return ReadArray(item).ToList();
+                        break;
+                    default:
+                        throw new InvalidOperationException($"Unexpected value kind: {item.ValueKind}");
```
sungam3r: Copy/paste from `ReadDictionary`. Consider to move in method like `ReadValue`

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+            JsonSerializerOptions options)
+        {
+            using var doc = JsonDocument.ParseValue(ref reader);
+            return ReadDictionary(doc.RootElement);
+        }
+
+        public override void Write(Utf8JsonWriter writer, Dictionary<string, object> value,
+            JsonSerializerOptions options)
+        {
+            JsonSerializer.Serialize(writer, value, _internalOptions);
```
sungam3r: Why not `WriteDictionary(writer, value, _internalOptions);` ?

I don’t see private write methods being called at all.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+            {
+                var value = entry.Value;
+
+                if (value == null)
+                {
+                    writer.WriteNull(entry.Key);
+                    continue;
+                }
+
+                WriteProperty(writer, options, value, entry.Key);
```
sungam3r:
```
 if (value == null)
                    writer.WriteNull(entry.Key);
                else
                    WriteProperty(writer, options, value, entry.Key);
```

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
benmccallum: Added. See `Writes_Correct_Execution_Result_With_Null_Data_And_Some_Errors`. 

I'm not 100% if graphql-dotnet is 100% to spec on all this. The writers might have a hard time distinguishing between errors occurring before or after execution started, as per spec [here](https://github.com/graphql/graphql-spec/blob/master/spec/Section%207%20--%20Response.md#response-format), but not really tempted to dig too deep into all that. Just solidified some tests around what currently is the behaviour.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+        {
+            switch (value)
+            {
+                case int intValue:
+                    writer.WriteNumber(key, intValue);
+                    break;
+                case double doubleValue:
+                    writer.WriteNumber(key, doubleValue);
+                    break;
+                case decimal decimalValue:
```
sungam3r: float, ulong, uint, etc cases ?

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
benmccallum: Yea it's pretty lucky. The only thing I was worried about was that I'm not currently passing the `options` recursively down, so if someone wanted PascalCase all the way down, there could be a gap there. The thing is, you're not supposed to pass it down, so you have to `new` up a new options for each recursion which is just nasty. 

Since all tests currently pass, until a test is written that breaks it, or someone complains, kinda tempted to just leave as is as an MVP.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+                case bool boolValue:
+                    writer.WriteBooleanValue(boolValue);
+                    break;
+                case Dictionary<string, object> subDictionary:
+                    WriteDictionary(writer, subDictionary, options);
+                    break;
+                case IEnumerable list:
+                    WriteArray(writer, list, options);
+                    break;
+            }
```
sungam3r: Copy/paste from `WriteProperty`. Consider to move inside method like `WriteValue` + there is no default case :(

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: If we can guarantee that `JsonElement` is a `ValueKind.Object`, should it always return a empty dictionary rather than `null`?

I can check upfront in the `Read` method that it is non-null and and a Object kind then remove the following:
```
            if (!result.Any())
                return null;
```

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
sungam3r: I also thought about it when I was engaged in a graphql client. We will see.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: The write methods aren't even being used, since I only use this to read out the `variables`. Can just drop all these write methods and make them throw. How does that sound?

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
sungam3r: ok

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
sungam3r: `ReadDictionary` is called not from only `Read`.

sungam3r: > should it always return a empty dictionary rather than null?

I don't know. Maybe in some cases there is a difference.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: Done - will push in a sec.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: Will remove.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: Will remove write methods entirely.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: Done - will push shortly.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
sungam3r: Either fix this to use private methods (wit fix copy-paste parts) or completely remove ones. At your discretion.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,223 @@
...
+                return bi;
+            else if (value.TryGetDouble(out var d))
+                return d;
+            else if (value.TryGetDecimal(out var dd))
+                return dd;
+
+            throw new NotImplementedException($"Unexpected Number value. Raw text was: {value.GetRawText()}");
+        }
+
+        private void WriteDictionary(Utf8JsonWriter writer, Dictionary<string, object> dictionary,
```
pekkah: I think you don't actually need these write methods.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
sungam3r: If you prefer to delete them, then I would like to leave in the comments or in the error message a mention of the reason why this is done.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: Yep, removed. Just pushed without.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: Done. Throw has an error msg.

sungam3r: good

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            foreach (var item in value.EnumerateArray())
+            {
+                yield return ReadValue(item);
+            }
+        }
+
+        private object ReadValue(JsonElement value)
+        {
+            switch (value.ValueKind)
```
sungam3r: Maybe convert to switch expression?

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: > ReadDictionary is called not from only Read.

Yea but in the other case (when looping over internal levels) I'd imagine if it's reporting a valuekind that isn't `ValueKind.Null` and is `ValueKind.Object`, it'd have to be an object. A null check in the Read method may even be redundant as the `Utf8JsonReader` throws very quickly if it doesn't recognize anything it recognizes. 

> I don't know. Maybe in some cases there is a difference.

I think since the `.ValueKind` property is telling us it's either `Null` or an `Object` that we should match that in our internal representation.

src/GraphQL.SystemTextJson/StringExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Converts a JSON-formatted string into a dictionary.
+        /// </summary>
+        /// <param name="json">A JSON formatted string.</param>
+        /// <returns>Inputs.</returns>
+        public static Inputs ToInputs(this string json)
+            => json?.ToDictionary().ToInputs();
+
+        /// <summary>
+        /// Converts a JSON formatted string into a dictionary of objects of their actual type.
```
sungam3r:
```suggestion
        /// Converts a JSON-formatted string into a dictionary of objects of their actual type.
```

src/GraphQL.SystemTextJson/StringExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        /// <param name="json">The json.</param>
+        /// <returns>Dictionary.</returns>
+        public static Dictionary<string, object> ToDictionary(this string json)
+            => JsonSerializer.Deserialize<Dictionary<string, object>>(json, _jsonOptions);
+
+        /// <summary>
+        /// Converts a JSON formatted string into a dictionary of objects of their actual type.
```
sungam3r: Copy-paste xmldoc from `ToDictionary`.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
sungam3r: I think that if desired, then we can figure this issue out, but now it’s not worth it.

src/GraphQL.SystemTextJson/StringExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+            => JsonSerializer.Deserialize<Dictionary<string, object>>(json, _jsonOptions);
+
+        /// <summary>
+        /// Creates an <see cref="ExecutionResult"/> with it's <see cref="ExecutionResult.Data" />
+        /// property set to the strongly-typed representation of <paramref name="json"/>.
+        /// </summary>
+        /// <param name="json">A json representation of the <see cref="ExecutionResult.Data"/> to be set.</param>
+        /// <param name="errors">Any errors.</param>
+        /// <returns>ExecutionResult.</returns>
+        public static ExecutionResult ToExecutionResult(this string json, ExecutionErrors errors = null)
```
sungam3r: But is this method generally needed somewhere other than tests? Will it be used by the server project?

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: Write methods were removed. This is no longer relevant.

src/GraphQL.SystemTextJson/StringExtensions.cs
benmccallum: Nope. I can move it somewhere else if we have a common place between Harness.Tests and .Tests projects.

src/GraphQL.Tests/AssertionExtensions.cs
```diff
@@ -1,21 +1,23 @@
...
-        public static void ShouldBeCrossPlat(this string a, string b)
+        public static void ShouldBeCrossPlat(this string actual, string expected)
+            => Normalize(actual).ShouldBe(Normalize(expected));
+
+        public static void ShouldBeCrossPlatJson(this string actualJson, string expectedJson)
         {
-            var aa = a?.Replace("\r\n", "\n").Replace("\\r\\n", "\\n");
-            var bb = b?.Replace("\r\n", "\n").Replace("\\r\\n", "\\n");
-            aa.ShouldBe(bb);
+            using var actualJsonDoc = JsonDocument.Parse(Normalize(actualJson));
```
sungam3r: By parse/serilize you wanted to make sure if it is a valid json inside?

src/GraphQL.SystemTextJson/StringExtensions.cs
benmccallum: Should I create a GraphQL.Tests.Core project? Essentially there was a tonne of methods before called `CreateQueryResult` that all used JObject directly, so I was trying to consolidate that.

src/GraphQL.Tests/DocumentWriterTests.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// Tests for <see cref="IDocumentWriter"/> implementations and the custom converters
+    /// that are used in the process of serializing an <see cref="ExecutionResult"/> to JSON.
+    /// </summary>
+    public class DocumentWriterTests
+    {
+        [Theory]
+        [ClassData(typeof(DocumentWritersTestData))]
+        public async void Can_Write_Execution_Result(IDocumentWriter writer)
+        {
+            var executionResult = new ExecutionResult()
```
sungam3r:
```suggestion
            var executionResult = new ExecutionResult
```

src/GraphQL.Tests/DocumentWriterTests.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    /// <summary>
+    /// Tests for <see cref="IDocumentWriter"/> implementations and the custom converters
+    /// that are used in the process of serializing an <see cref="ExecutionResult"/> to JSON.
+    /// </summary>
+    public class DocumentWriterTests
+    {
+        [Theory]
+        [ClassData(typeof(DocumentWritersTestData))]
+        public async void Can_Write_Execution_Result(IDocumentWriter writer)
```
sungam3r: Thank you for these tests. 🥇

src/GraphQL.Tests/AssertionExtensions.cs
benmccallum: This is just to ignore any differences in whitespace/formatting. 

Pretty much all the test bases do a similar thing. They have an `Assert` method that has an argument for the json expected. It turns that into an ExecutionResult then back into a string (removing any formatting), then runs the graph query and writes the actual result to a string too. Then compares them.

src/GraphQL.SystemTextJson/StringExtensions.cs
sungam3r: Let's use shared source code and not make new test projects.

src/GraphQL.Tests/DocumentWriterTests.cs
benmccallum: There's probably more that can be written, I didn't nitpick the serializers, for instance I think there's something custom for apollo trace, like it always need to be camelcase, but... it's a start.

sungam3r:
54 of 66 files done, I will continue after a few hours.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: This was removed when I removed the Write methods. No longer relevant.

src/GraphQL.NewtonsoftJson/SchemaExtensions.cs
benmccallum: No worries. Added it back, with a full comment, and one for S.T.Json as well.

src/GraphQL.SystemTextJson/ObjectDictionaryConverter.cs
benmccallum: I've made this change and I think it's the correct one. Going to resolve this; I'm happy with where it is and we've got bigger fish to fry.

src/GraphQL.SystemTextJson/StringExtensions.cs
benmccallum: There's no common projects between those two other than i guess GraphQL.csproj, but it doesn't make sense in here since it's purely for unit tests. Anywhere else you think it should go? Or should I just copy paste it into both those projects and move on?

benmccallum:
Tests should all pass after latest commit. 

I was getting confused as if the `publicApi.ShouldMatchApproved(` call finds a mismatch as it was trying to launch a diff tool window which I guess I don't have configured correctly on my machine. Added a quick comment in case anyone else loses time on this.

sungam3r:
Yes, indeed there are some problems with diff tool in vs 2019.

.gitignore
```diff
@@ -31,4 +31,4 @@ _NCrunch_*/
 bundle.js
 style.css
 
-ApiApprovalTests.PublicApi.received.txt
+*.received.txt
```
benmccallum: Ah yea, the old ignore was wrong now that it includes the assembly name. My fix makes sense now :P Getting tired!

src/GraphQL.SystemTextJson/StringExtensions.cs
sungam3r: I mean that shared code for tests stay in Test project (`shared` folder) and other projects use it as link.
```
  <ItemGroup>
    <Compile Include="..\GraphQL.Tests\Shared\*.cs" Link="Shared\*.cs" />
  </ItemGroup>
```

This solves copy-paste problem without creating new projects.

README.md
```diff
@@ -25,15 +25,29 @@ This project uses a [lexer/parser](http://github.com/graphql-dotnet/parser) orig
 > If errors occur, it is recommended that you first check the behavior on the latest available preview version before
 > reporting a issue.
 
-You can install the latest version via [NuGet](https://www.nuget.org/packages/GraphQL/).
+You can install the latest version via NuGet.
```
sungam3r:
```suggestion
You can install the latest stable version via [NuGet](https://www.nuget.org/packages/GraphQL).
```

README.md
```diff
@@ -25,15 +25,29 @@ This project uses a [lexer/parser](http://github.com/graphql-dotnet/parser) orig
...
 
+If you're targeting .NET Core 3+, you should use [GraphQL.SystemTextJson](https://www.nuget.org/packages/GraphQL.SystemTextJson/).
```
sungam3r:
```suggestion
If you're targeting .NET Core 3+, you are advised to use [GraphQL.SystemTextJson](https://www.nuget.org/packages/GraphQL.SystemTextJson/).
```

README.md
```diff
@@ -25,15 +25,29 @@ This project uses a [lexer/parser](http://github.com/graphql-dotnet/parser) orig
...
 ```
-> dotnet add package GraphQL
-> dotnet add package GraphQL -v 3.0.0-preview-1352
+> dotnet add package GraphQL.SystemTextJson
 ```
 
-Or you can get the latest pre-release packages from the [MyGet feed](https://www.myget.org/F/graphql-dotnet/api/v3/index.json).
+Else, you should use the legacy [GraphQL.NewtonsoftJson](https://www.nuget.org/packages/GraphQL.NewtonsoftJson/)
```
sungam3r:
```suggestion
Else, you should use the legacy and time-tested [GraphQL.NewtonsoftJson](https://www.nuget.org/packages/GraphQL.NewtonsoftJson/)
```

src/GraphQL.Tests/Execution/VariablesTests.cs
```diff
@@ -58,32 +58,21 @@ public TestJsonScalarReturningObject()
...
-        }
+        public override object Serialize(object value) => value;
 
         public override object ParseValue(object value)
-        {
-            if (!(value is string stringValue))
-                return null;
-            return JsonConvert.DeserializeObject<TestJsonScalarObject>(stringValue);
-        }
+            => !(value is string stringValue) ? null : JsonSerializer.Deserialize<TestJsonScalarObject>(stringValue);
```
sungam3r: Please invert this condition and one more below.

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
```diff
@@ -22,9 +22,13 @@ public async Task validate_core_schema()
                 _.Query = SchemaIntrospection.IntrospectionQuery;
             });
 
-            var json = await new DocumentWriter(true).WriteToStringAsync(executionResult);
+            var json = await documentWriter.WriteToStringAsync(executionResult);
 
-            ShouldBe(json, IntrospectionResult.Data);
+            var expected = documentWriter is SystemTextJson.DocumentWriter
```
sungam3r: Send through ClassData?

src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
```diff
@@ -30,14 +29,16 @@ public ExecutionResult AssertQuery(Action<ExecuteConfig> configure)
...
-                _.Inputs = config.Variables.ToInputs();
-                _.Root = config.Root;
-                _.ThrowOnUnhandledException = config.ThrowOnUnhandledException;
-            }, queryResult);
+            return AssertQuery(
+                _ =>
+                {
+                    _.Schema = schema;
+                    _.Query = config.Query;
+                    _.Inputs = SystemTextJson.StringExtensions.ToInputs(config.Variables);
```
sungam3r:
```suggestion
                    _.Inputs = config.Variables.ToInputs();
```

src/GraphQL/Execution/InputsExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace GraphQL
+{
+    public static class InputsExtensions
+    {
+        /// <summary>
+        /// Converts a dictionary into an <see cref="Inputs"/>.
+        /// </summary>
+        /// <param name="json">A dictionary.</param>
+        /// <returns>Inputs.</returns>
+        public static Inputs ToInputs(this Dictionary<string, object> dictionary)
```
sungam3r: I think that this method has a little value. Nevertheless, we can change it a little. I propose to make `Inputs` immutable:

```c#
 public class Inputs : ReadOnlyDictionary<string, object>
    {
        public static readonly Inputs Empty = new Inputs(new Dictionary<string, object>());

        public Inputs(IDictionary<string, object> dictionary)
            : base(dictionary)
        {
        }
    }
``` 

No default ctor + public Empty. Then in `InputExtensions`:
```c#
 public static Inputs ToInputs(this Dictionary<string, object> dictionary)
            => dictionary == null ? Inputs.Empty : new Inputs(dictionary);
```

Then only one edit in the `StarWarsBasicQueryTests` is required:
```c#
  var inputs = new Inputs(new Dictionary<string, object> {{"id", "1"}});
```

This will allow not to allocate a new `Dictionary` for variables to each request when there are no variables.

src/GraphQL.Tests/Execution/InputConversionTestsBase.cs
```diff
@@ -63,27 +62,27 @@ public class Child
         [Fact]
         public void converts_small_numbers_to_int()
         {
-            var json = @"{'a': 1}";
-            var inputs = json.ToInputs();
+            var json = @"{""a"": 1}";
+            var inputs = VariablesToInputs(json);
```
sungam3r: Good trick with testing both serializers with inheritance. Please add one more test where json is `null`, now all cases pass not `null` json.

src/GraphQL.SystemTextJson/SchemaExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// </summary>
+        /// <param name="schema">A schema to use.</param>
+        /// <param name="configure">An action that configures something to execute.</param>
+        /// <returns>The JSON result as a string.</returns>
+        /// <remarks>
+        /// Useful for quickly executing something and "getting started".
+        /// Part of the public API and should not be removed even if it has no references.
+        /// </remarks>
+        public static Task<string> ExecuteAsync(this ISchema schema, Action<ExecutionOptions> configure)
+            => schema.ExecuteAsync(new DocumentWriter(indent: true), configure);
```
sungam3r: Since all calls to `public static async Task<string> ExecuteAsync(this ISchema schema, IDocumentWriter documentWriter, Action<ExecutionOptions> configure)` pass not null delegate (at least `Query` property must be set) consider to add 
```c#
 if (configure == null)
                throw new ArgumentNullException(nameof(configure));
```
in the beginning of ` public static async Task<string> ExecuteAsync(this ISchema schema, IDocumentWriter documentWriter, Action<ExecutionOptions> configure)`

sungam3r:
How can I select all currently unresolved comments from PR to see what need to be done?

sungam3r:
Also do not forget to pull changed from master - #1510 has breaking changes that can indirectly affect your branch.

README.md
benmccallum: I'd removed this link as the GraphQL package itself is not really a valid entry point for most folks now. I'm basically now saying, "You can install from NuGet" then "Either one of these". The stable word is useful though.

README.md
benmccallum: I'd say there's no point even suggesting otherwise, and it's highly risky like the guys from MS have said. I'd prefer to not even suggest it.

README.md
benmccallum: Similar to above, I don't think we should be promoting this at all.

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
benmccallum: I thought about this, but was using the same ClassData type for a few different Theory tests. I'll rework this in a way that it's provided by a class data but we still only need to maintain one list of `IDocumentWriter`s somewhere.

src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
benmccallum: Ah yea, was getting a method not found at one stage and trying to figure out why. Eventually it disappeared once I deleted all my bin and obj folders. VS is fun sometimes.

src/GraphQL.Tests/Execution/InputConversionTestsBase.cs
benmccallum: Yea it seemed like the best solution haha. Bit of a weird situation we find ourselves in with extension methods per serializer implementation.

README.md
sungam3r: I understand what it is about. But in fact, I always had in mind the idea of taking one more step forward and putting `IDocumentWriter` (and in future some other interfaces) in a separate package (GraphQL.Abstractions maybe) and make both serializers packages dependent on this package. The goal is to separate common abstractions from graphql engine implementation stuff. Ideally, this package should not be changed (often) and may be even located in a separate repository. So `GraphQL` package continues to act as _entry point_ and the user will install into app both packages - GraphQL + GraphQL.SomeSerializer (both dependant from GraphQL.Abstractions). @joemcbride @benmccallum How do you like that?

sungam3r: And I approve of your change pointing the user to myget instead of nuget since often people don’t know where to get the latest versions. The 3.0.0 version of the package is born for a very long time and intermediate previews of the version appear on the nuget with a delay (manually published). That is why I focused on the word **stable**, although there is also a bunch of preview versions. So it’s better to immediately send the reader to myget for latest versions as you point later in readme.

src/GraphQL.Tests/Execution/InputConversionTestsBase.cs
benmccallum: Added test, found bug, win win.

sungam3r: I believe that we did everything as conveniently as possible.

src/GraphQL.SystemTextJson/SchemaExtensions.cs
benmccallum: Yep, makes sense. Not 100% foolproof as you elude to, but better than nothing.

src/GraphQL/Execution/InputsExtensions.cs
benmccallum: Great idea. Done.

README.md
benmccallum: What if we change the wording to something like:

> You can install the latest stable version via [NuGet](...).
> `> dotnet add package GraphQL`
> 
> For serialized outputs, you'll need an `IDocumentWriter` implementation. We support [GraphQL.SystemTextJson](...) for .NET Core 3+ else there's a legacy [GraphQL.NewtonsoftJson](...) (formerly included in [GraphQL](...),) or you can bring your own.
> `> dotnet add package GraphQL.SystemTextJson`
> `> dotnet add package GraphQL.NewtonsoftJson`
> 
> You can get the latest pre-release packages from the [MyGet feed](...), where you may want to explicitly pull a certain version using `-v`.
> `> dotnet add package GraphQL.SystemTextJson -v 3.0.0-preview-1448`

README.md
sungam3r: As you wish

README.md
sungam3r: ok

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -13,7 +13,10 @@ public static class StringExtensions
         /// <param name="json">A JSON-formatted string.</param>
         /// <returns>Inputs.</returns>
         public static Inputs ToInputs(this string json)
-            => json?.ToDictionary().ToInputs();
```
sungam3r: As I remember `?.` operator itself first checks left value for `null` so no need to change this code.

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
benmccallum: Done. Also thought of a better way of dealing with the encoding differences.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
benmccallum: I thought that too, hence why I refactored to that from the original. But it seems to short-circuit the entire call chain, so you get `null` back rather than the intended `Inputs.Empty`. That's what the unit test helped pick up.

README.md
sungam3r: ok

src/GraphQL.Tests/DocumentWritersTestData.cs
```diff
@@ -5,13 +5,19 @@ namespace GraphQL.Tests
...
+            new SystemTextJson.DocumentWriter(indent: true),
+            new NewtonsoftJson.DocumentWriter(indent: true)
         };
 
-        public IEnumerator<object[]> GetEnumerator() => _data.GetEnumerator();
+        public IEnumerator<object[]> GetEnumerator()
+        {
+            foreach (var writer in AllWriters)
+            {
+                yield return new object[] { new SystemTextJson.DocumentWriter(indent: true) };
```
sungam3r:
```suggestion
                yield return new object[] { writer };
```
?

src/GraphQL.NewtonsoftJson/StringExtensions.cs
sungam3r: > But it seems to short-circuit the entire call chain

Exactly, I didn’t notice the most important thing. So maybe change to
```c#
  public static Inputs ToInputs(this string json) => json?.ToDictionary().ToInputs() ?? Inputs.Empty;
```
to keep code more compact?

sungam3r: not so important anyway

README.md
benmccallum: Tweaked it again slightly. I thought about it and the async write is only an issue in an aspnet context, so just put in a warning for that situation.

benmccallum:
Have refreshed from upstream master, had no conflicts and I confirmed all tests still passing. #homestretch

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.SystemTextJson.approved.txt
```diff
@@ -0,0 +1,32 @@
...
+        public override void Write(System.Text.Json.Utf8JsonWriter writer, System.Collections.Generic.Dictionary<string, object> value, System.Text.Json.JsonSerializerOptions options) { }
+    }
+    public static class SchemaExtensions
+    {
+        public static System.Threading.Tasks.Task<string> ExecuteAsync(this GraphQL.Types.ISchema schema, System.Action<GraphQL.ExecutionOptions> configure) { }
+    }
+    public static class StringExtensions
+    {
+        public static System.Collections.Generic.Dictionary<string, object> ToDictionary(this string json) { }
+        public static GraphQL.ExecutionResult ToExecutionResult(this string json, GraphQL.ExecutionErrors errors = null) { }
```
benmccallum: @sungam3r did you decide what I should do with this? Dupe into the two test projs that use it or leave it here?

sungam3r: I prefer to move it to shared test code: https://github.com/graphql-dotnet/graphql-dotnet/pull/1512#discussion_r374824711

benmccallum: Ah sorry, missed that response. Done.

README.md
```diff
@@ -28,15 +28,29 @@ by [Marek Magdziak](https://github.com/mkmarek) and released with a MIT license.
...
+We support [GraphQL.SystemTextJson](https://www.nuget.org/packages/GraphQL.SystemTextJson/) for .NET Core 3+,
+[GraphQL.NewtonsoftJson](https://www.nuget.org/packages/GraphQL.NewtonsoftJson/)
+(formerly included in [GraphQL](https://www.nuget.org/packages/GraphQL/)), or you can bring your own.
+```
+> dotnet add package GraphQL.SystemTextJson
+> dotnet add package GraphQL.NewtonsoftJson
+```
+> *Note: You can use `GraphQL.NewtonsoftJson` with .NET Core 3+, just be aware it lacks async writing 
+> capabilities so writing to an ASP.NET Core 3.0 `HttpResponse.Body` will require you to set 
+> `AllowSynchronousIO` to `false` as per [this announcement](https://github.com/aspnet/Announcements/issues/342);
```
sungam3r:
```suggestion
> `AllowSynchronousIO` to `true` as per [this announcement](https://github.com/aspnet/Announcements/issues/342);
```

benmccallum: 🤦‍♂

src/GraphQL/Execution/InputsExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+            => dictionary == null ? Inputs.Empty : new Inputs(dictionary);
```
sungam3r:
```suggestion
            => dictionary == null || dictionary.Count == 0 ? Inputs.Empty : new Inputs(dictionary);
```

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -26,12 +24,12 @@ public static Inputs ToInputs(this string json)
         public static Inputs ToInputs(this JObject obj)
         {
             var variables = obj?.GetValue() as Dictionary<string, object>
-                            ?? new Dictionary<string, object>();
+                ?? new Dictionary<string, object>();
             return new Inputs(variables);
```
sungam3r:
```c#
var variables = obj?.GetValue() as Dictionary<string, object>;
return variables.ToInputs();
```

src/GraphQL/Execution/InputsExtensions.cs
BenjaBobs: Could also be
```suggestion
            => dictionary?.Count > 1 ? new Inputs(dictionary) : Inputs.Empty;
```

sungam3r: Yes, it could.

sungam3r:
@benmccallum I have no permissions to push into your repo so you should to make the last two edits yourself. After that I would be glad to merge this PR. It was a very long story.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
benmccallum: One less allocation. Nice. I've also added `remarks` to let anyone know this method is used by the server project. It could potentially be removed from here entirely, since we don't use it anymore, but I guess some folks could be using it in their own projects.

benmccallum:
@sungam3r , just checked the docs to see if there's anything worth adding in there and updated the installation.md since it was a bit bare. Almost tempted to move the readme.md stuff in there and link to the docs, but can do that another time.

LGTM!

sungam3r:
One test failed.

benmccallum:
I think this test has a race condition... It failed on my local once too and then second run it worked.

src/GraphQL.DataLoader.Tests/SimpleDataLoaderTests.cs
```diff
@@ -59,7 +59,7 @@ public async Task Operation_Can_Be_Cancelled()
             mock.Setup(store => store.GetAllUsersAsync(cts.Token))
                 .Returns(async (CancellationToken ct) =>
                 {
-                    await Task.Delay(100);
+                    await Task.Delay(200);
                     ct.ThrowIfCancellationRequested();
```
benmccallum: If you look further down, the cancellation happens after 5ms, and the delay here was 100ms, so really not sure why that test sometimes works and sometimes doesn't. There could be a more sinister bug here. Should we raise it as another issue to be looked at?

benmccallum: Maybe the cancellation fires before work even runs, so the delay in `CancelAfter` should be a bit higher?

sungam3r: I have been watching the history of this test for a long time. Something is really wrong with it. I assume that CI has strict restrictions on parallelism and because of this, the usual order of things is violated. Or an error somewhere else. In any case, you just need to repeat the build until the test is completed.

sungam3r: > Maybe the cancellation fires before work even runs, so the delay in CancelAfter should be a bit higher?

Maybe, but now we need to finish this PR.

sungam3r:
So... 78 of 78 files (double) checked and ready for merge. Is there anything else to note/fix?

benmccallum:
Not that I can think of.

Do I need to do anything to regen the docs? I changed that one file.

sungam3r:
Readme for generating docs https://github.com/graphql-dotnet/graphql-dotnet/tree/master/docs2

I tried but I couldn’t install the necessary tools. The documentation has already been changed and requires regeneration. This can be done later.

sungam3r:
Merged 😃

sungam3r:
Now the adventure will continue in the server project.

Vouzamo:
I've tried to upgrade to latest pre-release versions of packages on MyGet and can build but get the following exception when running despite registering a DocumentWriter for System.Text.Json:

InvalidOperationException: Unable to resolve service for type 'GraphQL.Server.Transports.AspNetCore.Common.IGraphQLRequestDeserializer' while attempting to activate 'GraphQL.Server.Transports.AspNetCore.GraphQLHttpMiddleware`1[Vouzamo.ERM.Api.Graph.MySchema]'.

Tried referencing the latest Harness but can't seem to find what I am missing. Any ideas?

sungam3r:
Did you upgrade the server packages as well? Install `Transports.AspNetCore.SystemTextJson` and call ` AddSystemTextJson` on `IGraphQLBuilder`. Example [here](https://github.com/graphql-dotnet/server).

Vouzamo:
I was missing GraphQL.Server.Transports.AspNetCore.SystemTextJson.

Thank you for the quick response.

PR #1523, opened 2020-02-08
Change ExecuteNodeAsync to return Task

Shane32:
This small change reduces an allocation per _synchronous_ resolver.  Note that some of the performance benefits may be lost with the updated DataLoader code, unfortunately.

```
// * Summary *

.NET Core SDK=3.1.101

*** BEFORE ***

| Introspection | 1,509.92 us | 9.103 us | 8.070 us | 207.0313 | 76.1719 |     - | 1050.9 KB |
|          Hero |    31.69 us | 0.201 us | 0.178 us |   6.2256 |       - |     - |  25.55 KB |

*** AFTER ***

|        Method |        Mean |     Error |   StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |------------:|----------:|---------:|---------:|--------:|------:|----------:|
| Introspection | 1,478.62 us | 10.099 us | 9.446 us | 187.5000 | 76.1719 |     - | 943.65 KB |
|          Hero |    31.95 us |  0.320 us | 0.299 us |   6.1646 |       - |     - |   25.3 KB |
```

### DIFF
--- a/src/GraphQL/Execution/ParallelExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ParallelExecutionStrategy.cs
@@ -14,7 +14,8 @@ namespace GraphQL.Execution
             var pendingNodes = new Queue<ExecutionNode>();
             pendingNodes.Enqueue(rootNode);
 
-            var currentTasks = new List<(Task, ExecutionNode)>();
+            var currentTasks = new List<Task>();
+            var currentNodes = new List<ExecutionNode>();
             while (pendingNodes.Count > 0)
             {
                 // Start executing pending nodes, while limiting the maximum number of parallel executed nodes to the set limit
@@ -26,7 +27,7 @@ namespace GraphQL.Execution
                     var pendingNodeTask = ExecuteNodeAsync(context, pendingNode);
                     if (pendingNodeTask.IsCompleted)
                     {
-                        //throw any caught exceptions
+                        // Throw any caught exceptions
                         await pendingNodeTask;
 
                         // Node completed synchronously, so no need to add it to the list of currently executing nodes
@@ -42,7 +43,8 @@ namespace GraphQL.Execution
                     else
                     {
                         // Node is actually asynchronous, so add it to the list of current tasks being executed in parallel
-                        currentTasks.Add((pendingNodeTask, pendingNode));
+                        currentTasks.Add(pendingNodeTask);
+                        currentNodes.Add(pendingNode);
                     }
 
                 }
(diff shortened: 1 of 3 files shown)


src/GraphQL/Execution/ParallelExecutionStrategy.cs
```diff
@@ -26,10 +26,12 @@ protected async Task ExecuteNodeTreeAsync(ExecutionContext context, ExecutionNod
                     var pendingNodeTask = ExecuteNodeAsync(context, pendingNode);
                     if (pendingNodeTask.IsCompleted)
                     {
+                        //throw any caught exceptions
```
sungam3r:
```suggestion
                        // throw any caught exceptions
```

src/GraphQL/Execution/ParallelExecutionStrategy.cs
```diff
@@ -14,7 +14,7 @@ protected async Task ExecuteNodeTreeAsync(ExecutionContext context, ExecutionNod
             var pendingNodes = new Queue<ExecutionNode>();
             pendingNodes.Enqueue(rootNode);
 
-            var currentTasks = new List<Task<ExecutionNode>>();
+            var currentTasks = new List<(Task, ExecutionNode)>();
```
sungam3r: I would prefer to split this data into two lists. It allows you to eliminate `Select` calls later.

Shane32: Yeah I thought about that. Just causes two object allocations when expanding the array. I can change it

sungam3r: `Select`s will also allocate, so...

sungam3r: We will see the results, it will be interesting to compare 😄

Shane32: The code in question never runs since none of the async resolvers calls Task.Yield.

Shane32: (During benchmarks)

src/GraphQL/Execution/ParallelExecutionStrategy.cs
Shane32: Done

src/GraphQL/Execution/ParallelExecutionStrategy.cs
Shane32: Done.  Benchmarks did not change, of course.

PR #1524, opened 2020-02-08
Add cancellationToken to IDocumentWriter.WriteAsync

sungam3r:
@benmccallum We completely forgot about canceling async IO.


sungam3r:
I added cancellation support here specifically for your PR in server project.

PR #1526, opened 2020-02-09
Simplify tests

sungam3r:


sungam3r:
@benmccallum I returned single `IntrospectionResult.Data` and removed all \uXXXX from tests.

PR #1527, opened 2020-02-09
[Discussion] Change IFieldResolver to return Task

Shane32:
Here's a first draft of the changes required to implement #1146 as discussed

Benchmarks show practically no change at present, mainly due to there being no truly async resolvers (need to add Task.Yield to have a fair comparison).  On the other hand, if we assume that the majority of field resolvers are synchronous, then the benchmarks are still relevant.

```
// * Summary *

.NET Core SDK=3.1.101

*** BEFORE ***

| Introspection | 1,509.92 us | 9.103 us | 8.070 us | 207.0313 | 76.1719 |     - | 1050.9 KB |
|          Hero |    31.69 us | 0.201 us | 0.178 us |   6.2256 |       - |     - |  25.55 KB |

*** AFTER ***

|        Method |        Mean |    Error |   StdDev |    Gen 0 |   Gen 1 | Gen 2 |  Allocated |
|-------------- |------------:|---------:|---------:|---------:|--------:|------:|-----------:|
| Introspection | 1,495.77 us | 6.912 us | 6.466 us | 201.1719 | 82.0313 |     - | 1050.88 KB |
|          Hero |    32.31 us | 0.308 us | 0.288 us |   6.2256 |  0.0610 |     - |   25.63 KB |
```

I added IFieldResolverInternal with SetResultAsync for the new functionality, and converted all or almost all of the internal field resolvers to use that.  Of course we can rename the interfaces and stuff if we actually want to proceed with this code change.  Only two tests fail; not sure exactly why yet.

One thing of note is that you cannot return a `Task<T>` from a FuncFieldResolver, as you could before.  A lot of field builders in the tests used Field rather than FieldAsync and returned a `Task<T>`, which doesn't work anymore.  I patched the tests to use FieldAsync.

I'd probably assume that the majority of the field resolvers are synchronous and call it quits at this point, considering the benchmarks.  Any truly asynchronous field resolver is presumably doing enough work that any speed gains would be dwarfed by whatever code is executing in the resolver.

So all that remains (from my point of view) is to determine if this code pattern is desired.  Assuming that most consumer's code will use FuncFieldResolver or AsyncFieldResolver, which does not change, it's not so bad, and (as it is now), IFieldResolver is still supported, which I like for backwards compatibility.  Using IFieldResolver should be almost identical performance as the prior code - they won't be worse off - but IFieldResolverInternal will be faster.

Again, all these optimizations only pertain to asynchronous resolvers and only save a single allocation per resolver ... so is it really worth the change?

That's it for tonight I think.  Maybe you want to pull and play around with it @sungam3r


Shane32:
A few comments about the implementation:
- The field middleware signatures changed to match the new design.  I tried to leave that alone but it didn't work out.  I've never used middleware but it passes all the tests
- FuncFieldResolver and other resolvers cannot implement both IFieldResolver and IFieldResolverInternal or else the compiler cannot determine the correct override when using the FieldBuilder.Resolve method
- FieldAsync with it's current overrides cannot auto detect the return type of `Task<T>` so pretty much you always have to use an async lambda and await the result to cast the result to an object.   Take a look at the signatures and you'll see what I mean.  Not ideal.  This is probably why FieldAsync wasn't used in the tests.

src/GraphQL/Resolvers/IFieldResolverInternal.cs
```diff
@@ -0,0 +1,10 @@
+using System.Threading.Tasks;
+using GraphQL.Types;
+
+namespace GraphQL.Resolvers
+{
+    public interface IFieldResolverInternal
```
sungam3r: The already overloaded API of resolvers becomes even more complicated.

sungam3r:
A lot of changes (as expected) and an almost imperceptible effect at the cost of complication. @johnrutherford what do you think? Should we try to solve #1146 further?

Compare for example with #1523 where just a few lines of code have been changed with 10% memory allocation gain. As I said before, there are enough other places where optimization will give a more noticeable effect at the cost of much smaller changes.

sungam3r:
@Shane32 Thank you very much for your time spent for these experiments 🥇

sungam3r:
Closed due to lack of benefit.

PR #1505, opened 2020-01-16
ResolveFieldContext.Path optimization

Shane32:
Slight performance enhancement

```
BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18362
Intel Core i7-7740X CPU 4.30GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.100
  [Host]     : .NET Core 3.1.0 (CoreCLR 4.700.19.56402, CoreFX 4.700.19.56404), X64 RyuJIT
  DefaultJob : .NET Core 3.1.0 (CoreCLR 4.700.19.56402, CoreFX 4.700.19.56404), X64 RyuJIT

** before **

|        Method |        Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 |  Allocated |
|-------------- |------------:|----------:|----------:|---------:|--------:|------:|-----------:|
| Introspection | 1,882.08 us | 23.988 us | 21.265 us | 255.8594 | 99.6094 |     - | 1318.82 KB |
|          Hero |    44.97 us |  0.443 us |  0.414 us |   7.4463 |       - |     - |   30.76 KB |

** after **

|        Method |        Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 |  Allocated |
|-------------- |------------:|----------:|----------:|---------:|--------:|------:|-----------:|
| Introspection | 1,818.70 us | 14.744 us | 13.070 us | 259.7656 | 99.6094 |     - | 1319.12 KB |
|          Hero |    44.59 us |  0.383 us |  0.359 us |   7.5073 |       - |     - |   30.86 KB |
```

PR #1533, opened 2020-02-13
ExecutionNode.Path optimization v2

Shane32:
A proposed update pertaining to #1147 and replacing #1505 .

Benchmarks:
```
// * Summary *

.NET Core SDK=3.1.101

*** BEFORE ***

| Introspection | 1,483.73 us | 12.154 us | 10.774 us | 191.4063 | 72.2656 |     - | 943.57 KB |
|          Hero |    31.96 us |  0.247 us |  0.219 us |   6.1035 |       - |     - |  25.21 KB |

*** AFTER ***

| Introspection | 1,354.75 us | 6.971 us | 6.180 us | 189.4531 | 62.5000 |     - | 835.55 KB |
|          Hero |    31.53 us | 0.348 us | 0.326 us |   6.1035 |       - |     - |   25.1 KB |
```

In order to validate new code design, I first added the new code without removing the old code, and temporarily used this code within the constructor of ExecuteNode to validate the new design:
``` C#
if (!Enumerable.SequenceEqual(this.Path, this.Path2))
{
    throw new Exception("Invalid path");
}
```
Everything worked, so the old code was removed.  All tests pass, of course.


src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -444,16 +444,6 @@ public static bool DoesFragmentConditionMatch(ExecutionContext context, string f
...
-            var newPath = new string[path.Length + 1];
-
-            path.CopyTo(newPath, 0);
-
-            newPath[path.Length] = pathSegment;
-
-            return newPath;
-        }
     }
+
```
sungam3r:
```suggestion
```

Shane32:
Also note that it could be programmed to use a yield iterator, but since the results need to be reversed, the list will need to be enumerated before it can be returned anyway.  See sample below:

``` C#
        public IEnumerable<string> Path => ReversePath().Reverse();

        private IEnumerable<string> ReversePath()
        {
            var node = this;
            while (!(node is RootExecutionNode))
            {
                if (node.PathIndex.HasValue)
                    yield return GetStringIndex(node.PathIndex.Value);
                else
                    yield return node.Field.Name;
                node = node.Parent;
            }
        }
```

You could walk the nodes' parents first and reverse them before yielding the names, but then you're creating the list ahead of time also.  So I just compiled the list, reversed and returned it.

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -59,6 +59,40 @@ public IObjectGraphType GetParentType(ISchema schema)
...
+            0 => "0",
+            1 => "1",
+            2 => "2",
+            3 => "3",
+            4 => "4",
+            5 => "5",
+            6 => "6",
+            7 => "7",
+            8 => "8",
+            9 => "9",
```
sungam3r: Let's extend the list to 16 cases as a tribute to the binary system :)

Shane32: haha sure

Shane32: it shouldn't make much difference anymore, unless someone is using this for extensive logging...

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -132,12 +131,12 @@ public static void SetArrayItemNodes(ExecutionContext context, ArrayExecutionNod
                             + $" Field: {parent.Name}, Type: {parent.FieldDefinition.ResolvedType}.");
 
                         error.AddLocation(parent.Field, context.Document);
-                        error.Path = path;
+                        error.Path = parent.Path.Append(index.ToString());
```
sungam3r: GetStringIndex ?

Shane32: GetStringIndex is a private member in ExecutionNode.cs  -- and as this is just exception handling, I didn't think it mattered much.

sungam3r: Yes, I understand this, just wanted to make sure.

sungam3r:
@Shane32 I really like this PR. In fact when I told you in https://github.com/graphql-dotnet/graphql-dotnet/pull/1505#issuecomment-575250542 that I want to change this place in a complete different way I wanted to do just that. But as often happens, I put off this work because I had to do something else.

sungam3r:
Also I think that we need the better name for `int? pathIndex` to point that it is index of **current** node inside parent node **if** parent node is array, otherwise `null`.

Shane32:
I'm glad you like it.  Not trying to step on your toes, but just trying to help get the 3.0 milestones done.  Honestly I'd like to see a non-preview 3.0 release, even if that means the very next version is 4.0 preview 1.  Also hoping my dataloader PR gets pulled in, as I really think it solves a number of bugs and works much better for the community.  (And I use my custom dataloader code all the time, so of course I think it's great!)

Shane32:
Give me a suggestion and I'll rename it, no big deal.  It was just the first thing that made sense to me.  I was going to have it protected, but then it couldn't be accessed by the subscription code that duplicated the node.  Besides, probably not a bad idea to have it accessible.

sungam3r:
1. I do not have permissions to publish this package to nuget.
2. The problem with the dataloder is that, basically, I can’t afford to merge code that I don’t understand completely. I have been conducting a review for quite long time because of a detailed analysis of the code. Dataloder and subscriptions are two things that I don’t have time to deal with, so many PRs are still open. But I do not give up hope to change this trend.

Shane32:
No I understand both points completely.  I think joemcbride wants to look at it himself, based on his comment, actually.  He's probably a busy person.  And my suggested changes to the dataloader are not insignificant.

sungam3r:
I propose `IndexInParentArrayExecutionNode`. It seems rather long name though.

Shane32:
Yeah ..... how about simplifying it just a bit -- `IndexInParent` or `IndexInParentArray` or `IndexInParentNode` ?  Maybe reverse the name and remove 'in' e.g. `ParentNodeIndex` ?

We can do the long name, but it's ... quite long ... even for descriptive variable names.  What do you think?

Shane32:
`ParentNodeIndex` doesn't seem clear enough to me.  I'd favor `IndexInParentArray` or possibly `IndexInParentNode`

sungam3r:
IndexInParentNode

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -59,6 +59,46 @@ public IObjectGraphType GetParentType(ISchema schema)
 
             return null;
         }
+
+        public IEnumerable<string> Path
+        {
+            get
+            {
+                var pathList = new List<string>();
```
sungam3r: Have you tried looking at the performance numbers of this method, if you determine the size of the list in advance and create it with the desired capacity? It requires do `while` loop twice, one time to count desired size, second time to fill the list, but it can save some allocations, so...

Shane32: No but I could.  Should be pretty easy to try

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
```diff
@@ -534,7 +534,7 @@ namespace GraphQL.Execution
         protected abstract System.Threading.Tasks.Task ExecuteNodeTreeAsync(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ObjectExecutionNode rootNode);
         protected virtual System.Threading.Tasks.Task OnBeforeExecutionStepAwaitedAsync(GraphQL.Execution.ExecutionContext context) { }
         protected virtual void ValidateNodeResult(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
-        public static GraphQL.Execution.ExecutionNode BuildExecutionNode(GraphQL.Execution.ExecutionNode parent, GraphQL.Types.IGraphType graphType, GraphQL.Language.AST.Field field, GraphQL.Types.FieldType fieldDefinition, string[] path = null) { }
+        public static GraphQL.Execution.ExecutionNode BuildExecutionNode(GraphQL.Execution.ExecutionNode parent, GraphQL.Types.IGraphType graphType, GraphQL.Language.AST.Field field, GraphQL.Types.FieldType fieldDefinition, int? pathIndex = default) { }
```
sungam3r: indexInParentNode

src/GraphQL/Execution/ExecutionNode.cs
sungam3r: It will be interesting to compare the results.

Shane32: Uhh, well, it's easy to write the code anyway.  But Path isn't used by the standard benchmarks, so none of this code typically runs.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -148,34 +147,17 @@ public static void SetArrayItemNodes(ExecutionContext context, ArrayExecutionNod
...
-            5 => "5",
-            6 => "6",
-            7 => "7",
-            8 => "8",
-            9 => "9",
-            _ => index.ToString()
-        };
-
-        public static ExecutionNode BuildExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, string[] path = null)
+        public static ExecutionNode BuildExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, int? pathIndex = null)
```
sungam3r: indexInParentNode

src/GraphQL/Execution/ExecutionNode.cs
Shane32: It also eliminates the call to Reverse

sungam3r:
```c#
  private ExecutionResult ExecuteQuery(ISchema schema, string query)
        {
            return _executer.ExecuteAsync(_ =>
            {
                _.Schema = schema;
                _.Query = query;
                _.EnableMetrics = true;       USE THIS TO FORCE ACCESS TO PATH
            }).GetAwaiter().GetResult();
        }
```

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Fixed

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
Shane32: Fixed

src/GraphQL/Execution/ExecutionNode.cs
Shane32: Take a look.  I didn't run benchmarks, but it's pretty obvious it should run faster, when there's no List object to allocate, and no Reverse() call -- just the raw array

sungam3r: By the way, I think that it makes no sense to make an iterator inside `Path`, because if there is a call to this property, then the consumer will read all the elements. By default, no one needs this property. It is used in advanced scenarios.

Shane32:
It could, but it's a very quick calculation, and considering how infrequent Path (should) be used, I'd save the memory in ExecutionNode.  I can add it if you think that would be better.

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -59,6 +59,55 @@ public IObjectGraphType GetParentType(ISchema schema)
...
+            {
+                var node = this;
+                var count = 0;
+                while (!(node is RootExecutionNode))
+                {
+                    node = node.Parent;
+                    ++count;
+                }
+
+                var pathList = new string[count];
```
sungam3r: By the way, in the future you can use ArrayPool here. But there will be some difficulties in choosing where to return the array to the pool.

sungam3r:
> It could, but it's a very quick calculation, and considering how infrequent Path (should) be used, I'd save the memory in ExecutionNode. I can add it if you think that would be better.

No, not worth it. I understand that we will get memory increase for each `ExecutionNode`.

sungam3r:
So, anything else to fix before merge?

Shane32:
Looks good to me

sungam3r:
Thanks for contribution.

sungam3r:
Results of our work. No more `String[]` in top of memory traffic:
image

PR #1537, opened 2020-02-15
Add DI support for Field Middleware, IFieldMiddleware interface

### LINKED ISSUE #1020: Field Middleware Dependencies
Field middleware constructor parameters do not seems to get resolved by the Schema dependency resolver. Is there way to inject dependencies into field middleware?

sungam3r:
Fixes #1020 

Requires changes in server project: `InstrumentFieldsMiddleware` must be registered in DI now.

@Shane32 You might be interested. You had a PR #1443 that was not merged. It was an attempt to improve Field Middlewares though without all DI stuff. It so happened that on our projects there was a need to use middleware with dependencies, so I returned to this issue. As it turned out, adding this feature was not so difficult.

@joemcbride Please review this and update the site with the documentation.

### DIFF
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -84,6 +84,7 @@ If you use `IFieldMiddlewareBuilder.Use` overloads which accept type parameter (
 those that do not accept a `IFieldMiddleware` instance or a middleware delegate) then your
 middleware creation will be delegated to DI-container. Thus, you can pass any dependencies to
 the Field Middleware constructor, provided that you have registered them correctly in DI.
+Note that it is required to have a schema that inherits from `Schema` or implements `IServiceProvider`.
 
 Also, the middleware itself should be registered in DI:
 
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -102,7 +102,34 @@ General recommendations for lifetimes are:
 | singleton | singleton  | singleton  | the safest and the most performant option recommended by default |
 | scoped    | scoped     | singleton  | much less performant option |
 | scoped    | scoped     | scoped     | the least performant option |
-| scoped    | singleton  | scoped     | DO NOT DO THAT! |
+| scoped    | singleton  | scoped     | DO NOT DO THAT! Explanation above. |
+| singleton | singleton  | scoped     | DO NOT DO THAT! InvalidOperationException: Cannot resolve scoped service from root provider |
+
+If your Field Middleware has scoped dependencies but your Schema and Graph Types are singletons
+(which is recommended for them) you can make Field Middleware singleton too and obtain the necessary
+dependencies right in the `Resolve` method. Here is an example of such an approach:
+
+```csharp
+public class MyFieldMiddleware : IFieldMiddleware
+{
+  private readonly IHttpContextAccessor _accessor;
+  private readonly IMySingletonService _service;
+
+  public MyFieldMiddleware(IHttpContextAccessor accessor, IMySingletonService service)
+  {
+    _accessor = accessor;
+    _service = service;
+  }
+
+  public Task<object> Resolve(IResolveFieldContext context, FieldMiddlewareDelegate next)
+  {
+    var scopedDependency1 = accessor.HttpContext.RequestServices.GetRequiredService<IMyService1>();
+    var scopedDependency2 = accessor.HttpContext.RequestServices.GetRequiredService<IMyService2>();
+    ...
+    return next(context);
+  }
+}
+```
 
 Options are also possible using transient lifetime, but are not given here (not recommended).
 
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -90,7 +90,7 @@ services.AddSingleton<InstrumentFieldsMiddleware>();
 ## Known issues
 
 Perhaps the most important thing with Field Middlewares that you should be aware of is that the
-default `DocumentExecuter` applies middlewares to the schema only once before the schema is
+default `DocumentExecuter` applies middlewares to the schema only once while the schema is being
 initialized. After this, calling any `IFieldMiddlewareBuilder.Use` methods has no effect.
 
 Field Middleware, when applying the schema, **modifies** the resolver of each field. Therefore,
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -81,6 +81,12 @@ those that do not accept a `IFieldMiddleware` instance or a middleware delegate)
 middleware creation will be delegated to DI-container. Thus, you can pass any dependencies to
 the Field Middleware constructor, provided that you have registered them correctly in DI.
 
+Also, the middleware itself should be registered in DI:
+
+```csharp
+services.AddSingleton<InstrumentFieldsMiddleware>();
+```
+
 ## Known issues
 
 Perhaps the most important thing with Field Middlewares that you should be aware of is that the
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -48,7 +48,10 @@ _.FieldMiddleware.Use(next =>
 {
   return context =>
   {
-    return next(context);
+    // your code here
+    var result = next(context);
+    // your code here
+    return result;
   };
 });
 ```
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -62,7 +62,7 @@ public interface IFieldMiddleware
 }
 ```
 
-It doesn’t have to be implemented on your middleware. Then a search will be made for such a method
+It doesn’t have to be implemented on your middleware. If you do not, a search will be made for a method named `Resolve`
 with a suitable signature.
 
 Nevertheless, **to improve performance, it is recommended to implement this interface.**
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -5,7 +5,7 @@ calculating the field value. You can write middleware for fields to provide addi
 during field resolution. After connecting the middleware to the schema, it is applied to all
 fields of all schema types. You can connect several middlewares to the schema. In this case,
 they will be called sequentially along the chain where the previous middleware decides to call
-the next one. This process is very similar to how middlewares work in ASP.NET Core HTTP request
+the next one. This process is very similar to how middlewares work in the ASP.NET Core HTTP request
 pipeline.
 
 The following example is how Metrics are captured. You register Field Middleware in the `ExecutionOptions`.
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -1,6 +1,6 @@
 # Field Middleware
 
-Field Middleware is a component connected to the schema, which is embedded in the process of
+Field Middleware is a component connected to the schema, which is embedded into the process of
 calculating the field value. You can write middleware for fields to provide additional behaviors
 during field resolution. After connecting the middleware to the schema, it is applied to all
 fields of all schema types. You can connect several middlewares to the schema. In this case,
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -8,7 +8,7 @@ they will be called sequentially along the chain where the previous middleware d
 the next one. This process is very similar to how middlewares work in the ASP.NET Core HTTP request
 pipeline.
 
-The following example is how Metrics are captured. You register Field Middleware in the `ExecutionOptions`.
+The following example is how Metrics are captured. You register Field Middleware in `ExecutionOptions`.
 
 ```csharp
 await schema.ExecuteAsync(_ =>
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -1,6 +1,14 @@
 # Field Middleware
 
-You can write middleware for fields to provide additional behaviors during field resolution.  The following example is how Metrics are captured.  You register Field Middleware in the `ExecutionOptions`.
+Field Middleware is a component connected to the schema, which is embedded in the process of
+calculating the field value. You can write middleware for fields to provide additional behaviors
+during field resolution. After connecting the middleware to the schema, it is applied to all
+fields of all schema types. You can connect several middlewares to the schema. In this case,
+they will be called sequentially along the chain where the previous middleware decides to call
+the next one. This process is very similar to how middlewares work in ASP.NET Core HTTP request
+pipeline.
+
+The following example is how Metrics are captured. You register Field Middleware in the `ExecutionOptions`.
 
 ```csharp
 await schema.ExecuteAsync(_ =>
@@ -10,7 +18,7 @@ await schema.ExecuteAsync(_ =>
 });
 ```
 
-You can write a class that has a `Resolve` method or you can register a middleware delegate directly.
+You can write a class that has a `Resolve` method:
 
 ```csharp
 public class InstrumentFieldsMiddleware
@@ -33,11 +41,7 @@ public class InstrumentFieldsMiddleware
 }
 ```
 
-The middleware delegate is defined as:
-
-``` csharp
-public delegate Task<object> FieldMiddlewareDelegate(IResolveFieldContext context);
-```
+Or you can register a middleware delegate directly:
 
 ```csharp
 _.FieldMiddleware.Use(next =>
@@ -48,3 +52,63 @@ _.FieldMiddleware.Use(next =>
   };
 });
 ```
+
+Also you can implement `IFieldMiddleware` interface in your Field Middleware:
+
+```csharp
+public interface IFieldMiddleware
+{
+  Task<object> Resolve(IResolveFieldContext context, FieldMiddlewareDelegate next);
+}
+```
+
+It doesn’t have to be implemented on your middleware. Then a search will be made for such a method
+with a suitable signature. Nevertheless, to improve performance, it is recommended to implement
+this interface.
+
+The middleware delegate is defined as:
+
+```csharp
+public delegate Task<object> FieldMiddlewareDelegate(IResolveFieldContext context);
+```
+
+## Field Middleware and Dependency Injection
+
+First, you are advised to read the article about [Dependency Injection](Dependency-Injection).
+
+If you use `IFieldMiddlewareBuilder.Use` overloads which accept type parameter (that is,
+those that do not accept a `IFieldMiddleware` instance or a middleware delegate) then your
+middleware creation will be delegated to DI-container. Thus, you can pass any dependencies to
+the Field Middleware constructor, provided that you have registered them correctly in DI.
+
+## Known issues
+
+Perhaps the most important thing with Field Middlewares that you should be aware of is that the
+default `DocumentExecuter` applies middlewares to the schema only once before the schema is
+initialized. After this, calling any `IFieldMiddlewareBuilder.Use` methods has no effect.
+
+Field Middleware, when applying the schema, **modifies** the resolver of each field. Therefore,
+you should be careful when using different lifetimes (singleton, scoped, transient) for your
+GraphTypes, Schema and Field Middleware. You **can** use any of lifetime, but for example in
+case of using singleton lifetime for some GraphType and scoped lifetime for Field Middleware
+and Schema this will cause the middleware to be applied to the same fields multiple times.
+In the case of ASP.NET Core app the resolvers of these fields will be wrapped again on each
+HTTP request to the server.
+
+General recommendations for lifetimes are:
+
+| Schema    | Graph Type | Middleware | Rating | 
+|-----------|------------|------------|--------|
+| singleton | singleton  | singleton  | the safest and the most performant option recommended by default |
+| scoped    | scoped     | singleton  | much less performant option |
+| scoped    | scoped     | scoped     | the least performant option |
+| scoped    | singleton  | scoped     | DO NOT DO THAT! |
+
+Options are also possible using transient lifetime, but are not given here (not recommended).
+
+## Field Middleware vs Directive
+
+If we consider Field Middleware as a way to globally affect the method of calculating all fields
+of all types in the Schema, then the directive can be considered as a way to locally affect only
+specific fields. The mechanism of their work is similar. For more information about directives
+see [Directives](Directives).
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -63,8 +63,9 @@ public interface IFieldMiddleware
 ```
 
 It doesn’t have to be implemented on your middleware. Then a search will be made for such a method
-with a suitable signature. Nevertheless, to improve performance, it is recommended to implement
-this interface.
+with a suitable signature.
+
+Nevertheless, **to improve performance, it is recommended to implement this interface.**
 
 The middleware delegate is defined as:
 
@@ -93,7 +94,7 @@ Perhaps the most important thing with Field Middlewares that you should be aware
 default `DocumentExecuter` applies middlewares to the schema only once while the schema is being
 initialized. After this, calling any `IFieldMiddlewareBuilder.Use` methods has no effect.
 
-Field Middleware, when applying the schema, **modifies** the resolver of each field. Therefore,
+Field Middleware, when applying to the schema, **modifies** the resolver of each field. Therefore,
 you should be careful when using different lifetimes (singleton, scoped, transient) for your
 GraphTypes, Schema and Field Middleware. You **can** use any of lifetime, but for example in
 case of using singleton lifetime for some GraphType and scoped lifetime for Field Middleware
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -53,7 +53,7 @@ _.FieldMiddleware.Use(next =>
 });
 ```
 
-Also you can implement `IFieldMiddleware` interface in your Field Middleware:
+Also, you can implement the `IFieldMiddleware` interface in your Field Middleware class:
 
 ```csharp
 public interface IFieldMiddleware
--- a/src/GraphQL.Harness/CountFieldMiddleware.cs
+++ b/src/GraphQL.Harness/CountFieldMiddleware.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using GraphQL.Instrumentation;
+using GraphQL.StarWars;
+using GraphQL.Types;
+using Microsoft.AspNetCore.Http;
+
+namespace Example
+{
+    /// <summary>
+    /// Example of Field Middleware with dependencies. When configured as singleton in DI
+    /// counts fields execution for the entire application lifetime. When configured as
+    /// scoped counts how many fields were executed when processing the single request.
+    /// </summary>
+    public sealed class CountFieldMiddleware : IFieldMiddleware, IDisposable
+    {
+        private int _count;
+
+        public CountFieldMiddleware(IHttpContextAccessor accessor, StarWarsData data)
+        {
+            // these dependencies are not needed here and are used only for demonstration purposes
+            Debug.Assert(accessor != null);
+            Debug.Assert(data != null);
+        }
+
+        public Task<object> Resolve(IResolveFieldContext context, FieldMiddlewareDelegate next)
+        {
+            Interlocked.Increment(ref _count);
+
+            return next(context);
+        }
+
+        public void Dispose()
+        {
+            Console.WriteLine($"{_count} fields were executed");
+        }
+    }
+}
--- a/src/GraphQL.Harness/CountFieldMiddleware.cs
+++ b/src/GraphQL.Harness/CountFieldMiddleware.cs
@@ -12,7 +12,7 @@ namespace Example
     /// <summary>
     /// Example of Field Middleware with dependencies. When configured as singleton in DI
     /// counts fields execution for the entire application lifetime. When configured as
-    /// scoped counts how many fields were executed when processing the single request.
+    /// scoped (with a scoped schema), counts how many fields were executed when processing the single request.
     /// </summary>
     public sealed class CountFieldMiddleware : IFieldMiddleware, IDisposable
     {
--- a/src/GraphQL.Harness/Startup.cs
+++ b/src/GraphQL.Harness/Startup.cs
@@ -1,4 +1,5 @@
 using Example;
+using GraphQL.Instrumentation;
 using GraphQL.StarWars;
 using GraphQL.StarWars.Types;
 using GraphQL.SystemTextJson;
@@ -24,10 +25,14 @@ namespace GraphQL.Harness
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+            // add execution components
             services.AddSingleton<IDocumentExecuter, DocumentExecuter>();
             services.AddSingleton<IDocumentWriter, DocumentWriter>();
 
+            // add something like repository
             services.AddSingleton<StarWarsData>();
+
+            // add graph types
             services.AddSingleton<StarWarsQuery>();
             services.AddSingleton<StarWarsMutation>();
             services.AddSingleton<HumanType>();
@@ -35,13 +40,21 @@ namespace GraphQL.Harness
             services.AddSingleton<DroidType>();
             services.AddSingleton<CharacterInterface>();
             services.AddSingleton<EpisodeEnum>();
+
+            // add schema
             services.AddSingleton<ISchema, StarWarsSchema>();
 
+            // add infrastructure stuff
             services.AddHttpContextAccessor();
             services.AddLogging(builder => builder.AddConsole());
 
+            // add options configuration
             services.Configure<GraphQLSettings>(Configuration);
             services.Configure<GraphQLSettings>(settings => settings.BuildUserContext = ctx => new GraphQLUserContext { User = ctx.User });
+
+            // add Field Middlewares
+            services.AddSingleton<CountFieldMiddleware>();
+            services.AddSingleton<InstrumentFieldsMiddleware>();
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
--- a/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
+++ b/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
@@ -24,6 +24,7 @@ namespace GraphQL.Tests.Instrumentation
                 FieldAst = new Field(null, new NameNode("Name")),
                 Source = new Person { Name = "Quinn" },
                 Errors = new ExecutionErrors(),
+                Schema = new Schema(),
                 Metrics = new Metrics().Start(null)
             };
         }
--- a/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
+++ b/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
@@ -86,7 +86,7 @@ namespace GraphQL.Tests.Instrumentation
         {
             _builder.Use<SimpleMiddleware>();
 
-            var result = _builder.Build().Invoke(_context).Result;
+            var result = _builder.Build(_context.Schema).Invoke(_context).Result;
             result.ShouldBe("Quinn");
 
             var record = _context.Metrics.Finish().Skip(1).Single();
--- a/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
+++ b/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
@@ -86,7 +86,7 @@ namespace GraphQL.Tests.Instrumentation
         {
             _builder.Use<SimpleMiddleware>();
 
-            var result = _builder.Build(schema: _context.Schema).Invoke(_context).Result;
+            var result = _builder.Build(start: null, schema: _context.Schema).Invoke(_context).Result;
             result.ShouldBe("Quinn");
 
             var record = _context.Metrics.Finish().Skip(1).Single();
@@ -156,5 +156,10 @@ namespace GraphQL.Tests.Instrumentation
                 }
             }
         }
+    }
+
+    internal static class TestExtensions
+    {
+        public static FieldMiddlewareDelegate Build(this FieldMiddlewareBuilder builder) => builder.Build(null, null);
     }
 }
--- a/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
+++ b/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
@@ -86,7 +86,7 @@ namespace GraphQL.Tests.Instrumentation
         {
             _builder.Use<SimpleMiddleware>();
 
-            var result = _builder.Build(_context.Schema).Invoke(_context).Result;
+            var result = _builder.Build(schema: _context.Schema).Invoke(_context).Result;
             result.ShouldBe("Quinn");
 
             var record = _context.Metrics.Finish().Skip(1).Single();
--- a/src/GraphQL/Execution/ExecutionOptions.cs
+++ b/src/GraphQL/Execution/ExecutionOptions.cs
@@ -17,7 +17,7 @@ namespace GraphQL
     {
         /// <summary>
         /// Schema of graph to use; required
-        /// <br/>
+        /// <br/><br/>
         /// Schema will be initialized if it has not yet been initialized.
         /// </summary>
         public ISchema Schema { get; set; }
--- a/src/GraphQL/Execution/IDocumentExecuter.cs
+++ b/src/GraphQL/Execution/IDocumentExecuter.cs
@@ -6,7 +6,7 @@ namespace GraphQL
     /// <summary>
     /// Processes an entire GraphQL request, given an input GraphQL request string. This is intended to
     /// be called by user code to process a query.
-    /// <br/>
+    /// <br/><br/>
     /// Typical implementation starts metrics if enabled (see <see cref="Instrumentation.Metrics">Metrics</see>),
     /// relies on <see cref="Execution.IDocumentBuilder">IDocumentBuilder</see> to parse the query,
     /// <see cref="Validation.IDocumentValidator">IDocumentValidator</see> to validate it, and
--- a/src/GraphQL/Execution/IProvideUserContext.cs
+++ b/src/GraphQL/Execution/IProvideUserContext.cs
@@ -9,7 +9,7 @@ namespace GraphQL.Execution
     {
         /// <summary>
         /// Mutable user-defined context to be passed to and shared by all field resolvers.
-        /// <br/>
+        /// <br/><br/>
         /// A custom implementation of <see cref="IDictionary{TKey, TValue}">IDictionary</see> may be
         /// used in place of the default <see cref="Dictionary{TKey, TValue}">Dictionary</see>.
         /// </summary>
--- a/src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
+++ b/src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
@@ -9,17 +9,17 @@ namespace GraphQL.Instrumentation
 {
     public interface IFieldMiddlewareBuilder
     {
-        IFieldMiddlewareBuilder Use(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware);
+        IFieldMiddlewareBuilder Use(Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware);
 
         void ApplyTo(ISchema schema);
     }
 
     public class FieldMiddlewareBuilder : IFieldMiddlewareBuilder
     {
-        private Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> _singleComponent;
-        private IList<Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate>> _components;
+        private Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate> _singleComponent;
+        private IList<Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate>> _components;
 
-        public IFieldMiddlewareBuilder Use(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware)
+        public IFieldMiddlewareBuilder Use(Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware)
         {
             middleware = middleware ?? throw new ArgumentNullException(nameof(middleware));
 
@@ -30,7 +30,7 @@ namespace GraphQL.Instrumentation
             }
             else if (_components == null)
             {
-                _components = new List<Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate>>
+                _components = new List<Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate>>
                 {
                     _singleComponent,
                     middleware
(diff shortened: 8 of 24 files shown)


sungam3r:
I also ask you to review my changes in the documentation since I am not a native speaker.

src/GraphQL.Harness/Startup.cs
```diff
@@ -24,24 +25,36 @@ public Startup(IConfiguration configuration)
...
             services.AddHttpContextAccessor();
             services.AddLogging(builder => builder.AddConsole());
 
+            // add options configuration
             services.Configure<GraphQLSettings>(Configuration);
             services.Configure<GraphQLSettings>(settings => settings.BuildUserContext = ctx => new GraphQLUserContext { User = ctx.User });
+
+            // add Field Middlewares
+            services.AddSingleton<CountFieldMiddleware>();
+            services.AddSingleton<InstrumentFieldsMiddleware>();
```
sungam3r: Just a note - this line of code should be done in the server project.

src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
```diff
@@ -24,6 +24,7 @@ public FieldMiddlewareBuilderTests()
                 FieldAst = new Field(null, new NameNode("Name")),
                 Source = new Person { Name = "Quinn" },
                 Errors = new ExecutionErrors(),
+                Schema = new Schema(),
```
sungam3r: Without this, resolving middlewares through DI will not work.

sungam3r:
@Shane32 Also I think that it no longer makes sense to optimize the call through reflection, since it is much easier (for middleware author) to implement the `IFieldMiddleware` interface in order to get a gain in the performance of the method call.

src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
```diff
@@ -8,42 +8,72 @@ namespace GraphQL.Instrumentation
 {
     public static class FieldResolverBuilderExtensions
     {
-        private const string InvokeMethodName = "Resolve";
+        private const string INVOKE_METHOD_NAME = "Resolve";
 
-        public static IFieldMiddlewareBuilder Use<T>(this IFieldMiddlewareBuilder builder) where T : new() => Use(builder, typeof(T));
+        public static IFieldMiddlewareBuilder Use(this IFieldMiddlewareBuilder builder, IFieldMiddleware middleware)
+            => builder.Use(next => context => middleware.Resolve(context, next));
```
Shane32: I like the added support for creating middleware that is "statically linked", and does not rely on DI at runtime.  I did notice that the default DI provider included with GraphQL performs identical to previously (as it pertains to middleware), with Activator.CreateInstance

docs2/site/docs/getting-started/field-middleware.md
```diff
@@ -48,3 +52,63 @@ _.FieldMiddleware.Use(next =>
...
+HTTP request to the server.
+
+General recommendations for lifetimes are:
+
+| Schema    | Graph Type | Middleware | Rating | 
+|-----------|------------|------------|--------|
+| singleton | singleton  | singleton  | the safest and the most performant option recommended by default |
+| scoped    | scoped     | singleton  | much less performant option |
+| scoped    | scoped     | scoped     | the least performant option |
+| scoped    | singleton  | scoped     | DO NOT DO THAT! |
```
Shane32: It would seem that `| singleton | singleton | scoped |` is worth mentioning here.  Based on code review (not testing) it would seem to be that the middleware would pull from the root service provider, in essence operating the same as if the middleware was registered as a singleton.  Note that typically the `InvalidOperationException: Cannot resolve scoped service from root provider` exception is thrown, although setting `options.ValidateScopes = false` ignores the error (but does not change the outcome).

If it were supported -- that is, a singleton schema with a scoped middleware -- that tied the middleware correctly to the current scoped service provider, it would seem to have a significant performance increase over having the schema, graph type, and middleware all scoped.  Perhaps I'm straying off-topic, but this could easily be done if the IResolveFieldContext had a Services property such as in PR #1345 -- the middleware would simply use the `context.Services` property instead of `context.Schema.Services`.  Something to think about anyway.

src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
```diff
@@ -8,42 +8,72 @@ namespace GraphQL.Instrumentation
...
-            if (invokeMethods.Length > 1)
+            // if IFieldMiddleware interface is supported, then just call its Resolve method
+            if (typeof(IFieldMiddleware).IsAssignableFrom(middleware))
             {
-                throw new InvalidOperationException($"There should be only a single method named {InvokeMethodName}. Middleware actually has {invokeMethods.Length} methods.");
+                return builder.Use(next =>
+                {
+                    return context =>
+                    {
+                        return context.Schema is Schema schema
```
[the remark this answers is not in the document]
sungam3r: I know. This is the only place in PR that has been haunting me for some time. I understand that it would be possible to add the `ISchema.Services ` property, but I have always been against it. This motivates the use of a ServiceLocator pattern in client code. In addition, I do not think that such custom implementations exist at all. It’s much easier to inherit from `Schema`, not `ISchema`. Therefore, I use an explicit cast here.

src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
```diff
@@ -8,42 +8,72 @@ namespace GraphQL.Instrumentation
...
+                            ? CheckNotNull((IFieldMiddleware)schema.Services.GetService(middleware), middleware).Resolve(context, next)
+                            : throw new NotSupportedException();
+                    };
+                });
             }
-
-            if (invokeMethods.Length == 0)
+            // otherwise, we first find this method on the type and then call it through reflection
+            else
```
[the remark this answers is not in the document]
sungam3r: Maybe. I just did not break what already works. The presence of a second code branch does not cause problems.

docs2/site/docs/getting-started/field-middleware.md
sungam3r: Right, ( singleton | singleton | scoped ) will cause `InvalidOperationException`, therefore, I did not mention this option.

> that tied the middleware correctly to the current scoped service provider, it would seem to have a significant performance increase over having the schema, graph type, and middleware all scoped.

Not really. I do not see the point in such a combination, because the middleware is still applied to the schema once so it is equivalent to ( singleton | singleton | singleton ) . Performance increase is primarily determined by the ability of the schema and its types to be singletons.

If the middleware needs scoped dependencies, they can be obtained indirectly through `IHttpContextAccessor` ctor argument or something like that, maybe `Func<T>`. We discussed this in other issues. But I admit that something else can be improved for middlewares.

sungam3r:
Side note - `InstrumentFieldsMiddleware` is created only once now, not on each field.

docs2/site/docs/getting-started/field-middleware.md
sungam3r: I added a description of your case in the readme.

Shane32: Looks good to me, given the present options

src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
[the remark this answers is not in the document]
Shane32: You're replacing one "anti-pattern" with another anti-pattern.  And it's not like exposing the Service property is going to help anyone -- it only provides access to the root service provider, and not the executing context's service provider.  Any singletons someone wants to use can be placed in the constructor of the graphtype.  Any scoped services still need to be accessed via an accessor.

Take a look at the Schema class.  Services doesn't even belong as a public member at all, unless to be exposed through ISchema, based on the 'anti-pattern' philosophy.  Just needs one change: `_lookup = new Lazy<GraphTypesLookup>(() => CreateTypesLookup(services));` and Schema.Services could be removed from the entire project.

I really don't think this PR should be pulled with this cast still in place.  I do think Services should be added to ISchema.  But if it isn't, then another solution should be found.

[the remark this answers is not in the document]
sungam3r: > You're replacing one "anti-pattern" with another anti-pattern.

Where?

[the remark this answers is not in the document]
Shane32: `context.Schema is Schema schema`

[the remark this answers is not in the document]
sungam3r: > Take a look at the Schema class. Services doesn't even belong as a public member at all, unless to be exposed through ISchema, based on the 'anti-pattern' philosophy. Just needs one change: _lookup = new Lazy<GraphTypesLookup>(() => CreateTypesLookup(services)); and Schema.Services could be removed from the entire project.

Good observation. It may be worth doing in order to be consistent until we find the best option.

> I really don't think this PR should be pulled with this cast still in place. I do think Services should be added to ISchema. But if it isn't, then another solution should be found.

let's consider an option _another solution should be found_.

Anyway i see it like this:

1. `Services` exists as an entity and the schema works with it to integrate with outer world.
2. So... in fact schema uses ServiceLocator pattern since schema is an infrastructure code not user code. In the end, some code MUST call `IServiceProvider.GetService()` and "get hands dirty".
3. User code should by all means avoid explicit work with `IServiceProvider.GetService()`. And we should not motivate user to call `IServiceProvider.GetService()` exposing `Schema.Services` or something else.

As a result: library code one way or another can work with `IServiceProvider`, but the user code is not.

Yes, and keep in mind that despite the fact that I do my best to protect the user code from the appearance of `Schema.Services`, sometimes I still think about giving it up and let it be.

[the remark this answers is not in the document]
sungam3r: But I would start with the question "why do we need `ISchema` interface at all"? Is this really necessary?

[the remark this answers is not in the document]
Shane32: Good question.  I do not have the answer.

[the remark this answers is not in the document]
sungam3r: I too. @joemcbride we need your help.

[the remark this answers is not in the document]
sungam3r: How do you like this option?
```c#
public class Schema : ISchema, IServiceProvider
{
  object IServiceProvider.GetService(Type type) => _services.GetService(type);
}
```
\+Also remove `public IServiceProvider Services` from Schema and do this change:
```c#
  return context.Schema is IServiceProvider provider
 ? CheckNotNull((IFieldMiddleware)provider.GetService(middleware), middleware).Resolve(context, next)
                            : throw new NotSupportedException("Schema should implement IServiceProvoder interface");
```

So at least there may be custom `ISchema` implementations which could implement `IServiceProvider`.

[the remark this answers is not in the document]
Shane32: Well, it's a better option than before.  It is still poor code design; there's still a cast with a possibility of failure, when interfaces (which already exist) should protect against this.  In this example, ISchema should inherit from IServiceProvider.  (This of course then breaks the goal of hiding Services from the user.)  I still wouldn't do it.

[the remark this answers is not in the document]
sungam3r: > It is still poor code design

In fact, I do not see the insoluble problems of this design. The only drawback of the inability to use custom (although I doubt that there are any) schemes we decided. It is enough for the client to implement the interface, a message about which he will see. And we (at least me) still not expose `Schema.Services`.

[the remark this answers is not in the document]
sungam3r: > ISchema should inherit from IServiceProvider

I do not think so. Some interfaces are infrastructural in nature and are not intended for access by clients. At least clients should make some efforts and obtain access explicitly, that is, understand what they are doing.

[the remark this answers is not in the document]
sungam3r: @Shane32 I followed your suggestion and removed `Schema.Services` property. In addition, I think it was dangerous to allow changing it after creating the schema.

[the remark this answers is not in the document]
Shane32: I believe the actual removal of the Schema.Services property should be in a separate PR.  While it does not affect any of my code, others may have differing points of view, and it would be good to allow time for comments before the removal.  It should not affect your implementation of this PR.

I would not object to the change, since if anyone wanted access to the root IServiceProvider, they could simply add a reference to it in the constructor of the graphtype.  Further, they could expose the property in their `MySchema` class if desired.  However, I certainly have used `((Schema)context.Schema).Services` in the past, and others may have similar code in their codebase.

> I think it was dangerous to allow changing it after creating the schema.

I see nothing wrong with making it a readonly property.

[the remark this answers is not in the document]
sungam3r: ((Schema)context.Schema).Services -> (IServiceProvider)context.Schema

I also thought about readonly property, but the fix is so simple that I decided to remove it right away.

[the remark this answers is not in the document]
Shane32: Makes sense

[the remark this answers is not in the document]
sungam3r: > others may have differing points of view, and it would be good to allow time for comments before the removal

I'm in no hurry.

Shane32:
@sungam3r FYI there are now merge conflicts to be fixed.

Have you thought any more about removing the old pattern-based approach to middleware, as it cleans up the code and implementers only need to add `: IFieldMiddleware` to make custom implementations work again?  I can't think of any reason to keep the old approach.

sungam3r:
I want to leave this piece of code. It is not so complicated and burdensome.

>  I can't think of any reason to keep the old approach.

Because the authors of middlewares might not have the desire or the ability to change anything. For example, middleware can be implemented in an external library consumed by another client. 

I would delete, but in this particular case I see no problem leaving both code branches.

Shane32:
Perhaps just mark the old pattern `[Obsolete]` ?

sungam3r:
Where and how?

Shane32:
Hmmm... good question.  Apparently C# doesn't support an overload that differs only in a generic constraint.

Shane32:
With the new changes, it looks like the middleware is pulled from the schema's service provider for each call to a field resolver.  Before, the Activator.CreateInstance was only created while the middleware was applied to the schema (e.g. during schema initialization).  This would result in a significant performance hit.  I've carefully examined the DI impact, which I believe is none, unless the middleware was registered as transient.  Take a look at this suggestion:

``` C#
// for IMiddleware
                return builder.Use(next =>
                {
                    // Not an ideal solution, but at least it allows to work with custom schemas which are not inherited from Schema type
                    var middlewareInstance = context.Schema is IServiceProvider provider
                        ? CheckNotNull((IFieldMiddleware)provider.GetService(middleware), middleware)
                        : throw NotSupported(context.Schema);

                    return context =>
                    {
                        return middlewareInstance.Resolve(context, next);
                    };
                });

// for pattern-based middleware
                return builder.Use(next =>
                {
                    // Not an ideal solution, but at least it allows to work with custom schemas which are not inherited from Schema type
                    var middlewareInstance = context.Schema is IServiceProvider provider
                        ? CheckNotNull((IFieldMiddleware)provider.GetService(middleware), middleware)
                        : throw NotSupported(context.Schema);

                    return context =>
                    {
                        return (Task<object>)methodInfo.Invoke(middlewareInstance, new object[] { context, next });
                    };
                });
```

src/GraphQL/Resolvers/AsyncFieldResolver.cs
```diff
@@ -33,7 +33,7 @@ object IFieldResolver.Resolve(IResolveFieldContext context)
 
     /// <summary>
     /// When resolving a field, this implementation calls a predefined <see cref="Func{T, TResult}"/> and returns the result.
-    /// The returned value must be of an <see cref="Task{TResult}"/> type.<br/>
```
Shane32: All these `<br/>` are necessary to create the blank line between paragraphs. Please do not remove them.

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -15,7 +15,11 @@ namespace GraphQL
     /// <summary>Configuration options to be passed to <see cref="IDocumentExecuter"/> to execute a query</summary>
     public class ExecutionOptions : IProvideUserContext
     {
-        /// <summary>Schema of graph to use; required<br/><br/>Schema will be initialized if has not yet been initialized.</summary>
+        /// <summary>
+        /// Schema of graph to use; required
```
Shane32:
```suggestion
        /// Schema of graph to use; required<br/>
```
There must be two `<br/>` between sentences to create a blank line between paragraphs.

src/GraphQL/Resolvers/AsyncFieldResolver.cs
sungam3r: The text was not so big to break it into paragraphs. Is it important for you?

Shane32: I believe it looked very clean, for the same reason it looks clean in the code.  Visual Studio only puts like half a line of space between paragraphs anyway. Any text that is not separated into paragraphs should not have any `<br/>` so that it is a single paragraph.

src/GraphQL/Execution/IDocumentExecuter.cs
```diff
@@ -5,7 +5,7 @@ namespace GraphQL
 {
     /// <summary>
     /// Processes an entire GraphQL request, given an input GraphQL request string. This is intended to
-    /// be called by user code to process a query.<br/>
+    /// be called by user code to process a query.
```
Shane32:
```suggestion
    /// be called by user code to process a query.<br/>
```

src/GraphQL/Execution/IProvideUserContext.cs
```diff
@@ -8,7 +8,7 @@ namespace GraphQL.Execution
     public interface IProvideUserContext
     {
         /// <summary>
-        /// Mutable user-defined context to be passed to and shared by all field resolvers.<br/>
+        /// Mutable user-defined context to be passed to and shared by all field resolvers.
```
Shane32:
```suggestion
        /// Mutable user-defined context to be passed to and shared by all field resolvers.<br/>
```

sungam3r:
> With the new changes, it looks like the middleware is pulled from the schema's service provider for each call to a field resolver.

Right. As designed.

> Before, the Activator.CreateInstance was only created while the middleware was applied to the schema (e.g. during schema initialization)

Right, but pulling from the schema's service provider is not the same as `Activator.CreateInstance`. In the case of a singleton, it's just getting the already created (once) object.

>  This would result in a significant performance hit.

Are you sure? And I see that your suggestion does not compile - `context` variable available only inside inner delegate.

src/GraphQL/Resolvers/AsyncFieldResolver.cs
```diff
@@ -33,7 +33,7 @@ object IFieldResolver.Resolve(IResolveFieldContext context)
...
+    /// The returned value must be of an <see cref="Task{TResult}"/> type.
```
Shane32:
```suggestion
    /// The returned value must be of an <see cref="Task{TResult}"/> type.<br/>
```

src/GraphQL/Resolvers/FuncFieldResolver.cs
```diff
@@ -27,7 +27,7 @@ object IFieldResolver.Resolve(IResolveFieldContext context)
     }
 
     /// <summary>
-    /// When resolving a field, this implementation calls a predefined <see cref="Func{T, TResult}"/> and returns the result.<br/>
+    /// When resolving a field, this implementation calls a predefined <see cref="Func{T, TResult}"/> and returns the result.
```
Shane32:
```suggestion
    /// When resolving a field, this implementation calls a predefined <see cref="Func{T, TResult}"/> and returns the result.<br/>
```

src/GraphQL/Resolvers/IFieldResolver.cs
```diff
@@ -4,19 +4,19 @@
 namespace GraphQL.Resolvers
 {
     /// <summary>
-    /// A field resolver returns an object for a given field within a graph.<br/>
+    /// A field resolver returns an object for a given field within a graph.
```
Shane32:
```suggestion
    /// A field resolver returns an object for a given field within a graph.<br/>
```

src/GraphQL/Resolvers/IFieldResolver.cs
```diff
@@ -4,19 +4,19 @@
...
     /// <br/>
-    /// The <see cref="FieldType.Resolver"/> property defines the field resolver to be used for the field.<br/>
+    /// The <see cref="FieldType.Resolver"/> property defines the field resolver to be used for the field.
```
Shane32:
```suggestion
    /// The <see cref="FieldType.Resolver"/> property defines the field resolver to be used for the field.<br/>
```

src/GraphQL/Resolvers/IFieldResolver.cs
```diff
@@ -4,19 +4,19 @@
...
     /// <br/>
     /// Typically an instance of <see cref="FuncFieldResolver{TSourceType, TReturnType}">FuncFieldResolver</see> or
     /// <see cref="AsyncFieldResolver{TSourceType, TReturnType}">AsyncFieldResolver</see> is created when code needs
     /// to execute within the field resolver - typically by calling
     /// <see cref="Builders.FieldBuilder{TSourceType, TReturnType}">FieldBuilder</see>.<see cref="Builders.FieldBuilder{TSourceType, TReturnType}.Resolve(System.Func{IResolveFieldContext{TSourceType}, TReturnType})">Resolve</see>
-    /// or <see cref="Builders.FieldBuilder{TSourceType, TReturnType}">FieldBuilder</see>.<see cref="Builders.FieldBuilder{TSourceType, TReturnType}.ResolveAsync(System.Func{IResolveFieldContext{TSourceType}, System.Threading.Tasks.Task{TReturnType}})">ResolveAsync</see>.<br/>
+    /// or <see cref="Builders.FieldBuilder{TSourceType, TReturnType}">FieldBuilder</see>.<see cref="Builders.FieldBuilder{TSourceType, TReturnType}.ResolveAsync(System.Func{IResolveFieldContext{TSourceType}, System.Threading.Tasks.Task{TReturnType}})">ResolveAsync</see>.
```
Shane32:
```suggestion
    /// or <see cref="Builders.FieldBuilder{TSourceType, TReturnType}">FieldBuilder</see>.<see cref="Builders.FieldBuilder{TSourceType, TReturnType}.ResolveAsync(System.Func{IResolveFieldContext{TSourceType}, System.Threading.Tasks.Task{TReturnType}})">ResolveAsync</see>.<br/>
```

src/GraphQL/Resolvers/IFieldResolver.cs
```diff
@@ -4,19 +4,19 @@
...
     /// <br/>
     /// When mapping fields to source object properties via
     /// <see cref="ComplexGraphType{TSourceType}.Field{TProperty}(System.Linq.Expressions.Expression{System.Func{TSourceType, TProperty}}, bool, System.Type)">Field(x => x.Name)</see>,
-    /// <see cref="ExpressionFieldResolver{TSourceType, TProperty}">ExpressionFieldResolver</see> is used.<br/>
+    /// <see cref="ExpressionFieldResolver{TSourceType, TProperty}">ExpressionFieldResolver</see> is used.
```
Shane32:
```suggestion
    /// <see cref="ExpressionFieldResolver{TSourceType, TProperty}">ExpressionFieldResolver</see> is used.<br/>
```

src/GraphQL/Resolvers/NameFieldResolver.cs
```diff
@@ -7,7 +7,7 @@ namespace GraphQL.Resolvers
 {
     /// <summary>
     /// Attempts to return a value for a field from the graph's source object, matching the name of
-    /// the field to a property or a method with the same name on the source object.<br/>
+    /// the field to a property or a method with the same name on the source object.
```
Shane32:
```suggestion
    /// the field to a property or a method with the same name on the source object.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -6,9 +6,9 @@
 namespace GraphQL.Types
 {
     /// <summary>
-    /// The schema for the GraphQL request. Contains references to the 'query', 'mutation', and 'subscription' base graph types.<br/>
+    /// The schema for the GraphQL request. Contains references to the 'query', 'mutation', and 'subscription' base graph types.
```
Shane32:
```suggestion
    /// The schema for the GraphQL request. Contains references to the 'query', 'mutation', and 'subscription' base graph types.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -6,9 +6,9 @@
...
     /// <br/>
-    /// Also allows for adding custom directives, additional graph types, and custom value converters.<br/>
+    /// Also allows for adding custom directives, additional graph types, and custom value converters.
```
Shane32:
```suggestion
    /// Also allows for adding custom directives, additional graph types, and custom value converters.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -20,7 +20,7 @@ public interface ISchema
         bool Initialized { get; }
 
         /// <summary>
-        /// Initializes the schema. Called by <see cref="IDocumentExecuter"/> before validating or executing the request.<br/>
+        /// Initializes the schema. Called by <see cref="IDocumentExecuter"/> before validating or executing the request.
```
Shane32:
```suggestion
        /// Initializes the schema. Called by <see cref="IDocumentExecuter"/> before validating or executing the request.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -47,12 +47,12 @@ public interface ISchema
         IObjectGraphType Subscription { get; set; }
 
         /// <summary>
-        /// Returns a list of directives supported by the schema.<br/>
+        /// Returns a list of directives supported by the schema.
```
Shane32:
```suggestion
        /// Returns a list of directives supported by the schema.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -47,12 +47,12 @@ public interface ISchema
...
         /// <br/>
         /// Directives are used by the GraphQL runtime as a way of modifying execution
-        /// behavior. Type system creators do not usually create them directly.<br/>
+        /// behavior. Type system creators do not usually create them directly.
```
Shane32:
```suggestion
        /// behavior. Type system creators do not usually create them directly.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -47,12 +47,12 @@ public interface ISchema
...
         /// <br/>
-        /// <see cref="Schema"/> intializes the list to include <see cref="DirectiveGraphType.Include"/>, <see cref="DirectiveGraphType.Skip"/> and <see cref="DirectiveGraphType.Deprecated"/> by default.
+        /// <see cref="Schema"/> initializes the list to include <see cref="DirectiveGraphType.Include"/>, <see cref="DirectiveGraphType.Skip"/> and <see cref="DirectiveGraphType.Deprecated"/> by default.
```
Shane32:
```suggestion
        /// <see cref="Schema"/> initializes the list to include <see cref="DirectiveGraphType.Include"/>, <see cref="DirectiveGraphType.Skip"/> and <see cref="DirectiveGraphType.Deprecated"/> by default.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -77,47 +77,47 @@ public interface ISchema
         IEnumerable<Type> AdditionalTypes { get; }
 
         /// <summary>
-        /// Add a specific instance of an <see cref="IGraphType"/> to the schema.<br/>
+        /// Add a specific instance of an <see cref="IGraphType"/> to the schema.
```
Shane32:
```suggestion
        /// Add a specific instance of an <see cref="IGraphType"/> to the schema.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -77,47 +77,47 @@ public interface ISchema
...
         /// <br/>
         /// Not typically required as schema initialization will scan the <see cref="Query"/>, <see cref="Mutation"/> and <see cref="Subscription"/> graphs,
         /// creating instances of <see cref="IGraphType"/>s referenced therein as necessary.
         /// </summary>
         void RegisterType(IGraphType type);
 
         /// <summary>
-        /// Add specific instances of <see cref="IGraphType"/>s to the schema.<br/>
+        /// Add specific instances of <see cref="IGraphType"/>s to the schema.
```
Shane32:
```suggestion
        /// Add specific instances of <see cref="IGraphType"/>s to the schema.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -77,47 +77,47 @@ public interface ISchema
...
         /// <br/>
         /// Not typically required as schema initialization will scan the <see cref="Query"/>, <see cref="Mutation"/> and <see cref="Subscription"/> graphs,
         /// creating instances of <see cref="IGraphType"/>s referenced therein as necessary.
         /// </summary>
         void RegisterTypes(params IGraphType[] types);
 
         /// <summary>
-        /// Add specific graph types to the schema. Each type must implement <see cref="IGraphType"/>.<br/>
+        /// Add specific graph types to the schema. Each type must implement <see cref="IGraphType"/>.
```
Shane32:
```suggestion
        /// Add specific graph types to the schema. Each type must implement <see cref="IGraphType"/>.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -77,47 +77,47 @@ public interface ISchema
...
         /// <br/>
         /// Not typically required as schema initialization will scan the <see cref="Query"/>, <see cref="Mutation"/> and <see cref="Subscription"/> graphs,
         /// creating instances of <see cref="IGraphType"/>s referenced therein as necessary.
         /// </summary>
         void RegisterTypes(params Type[] types);
 
         /// <summary>
-        /// Add a specific graph type to the schema.<br/>
+        /// Add a specific graph type to the schema.
```
Shane32:
```suggestion
        /// Add a specific graph type to the schema.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -77,47 +77,47 @@ public interface ISchema
...
         /// <br/>
         /// Not typically required as schema initialization will scan the <see cref="Query"/>, <see cref="Mutation"/> and <see cref="Subscription"/> graphs,
         /// creating instances of <see cref="IGraphType"/>s referenced therein as necessary.
         /// </summary>
         void RegisterType<T>() where T : IGraphType;
 
         /// <summary>
-        /// Add a specific directive to the schema.<br/>
+        /// Add a specific directive to the schema.
```
Shane32:
```suggestion
        /// Add a specific directive to the schema.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -77,47 +77,47 @@ public interface ISchema
...
         /// <br/>
         /// Directives are used by the GraphQL runtime as a way of modifying execution
         /// behavior. Type system creators do not usually create them directly.
         /// </summary>
         void RegisterDirective(DirectiveGraphType directive);
 
         /// <summary>
-        /// Add specific directives to the schema.<br/>
+        /// Add specific directives to the schema.
```
Shane32:
```suggestion
        /// Add specific directives to the schema.<br/>
```

src/GraphQL/Types/Schema.cs
```diff
@@ -76,9 +77,30 @@ public void Initialize()
         public IObjectGraphType Subscription { get; set; }
 
         /// <summary>
-        /// The service provider used to create objects, such as graph types, requested by the schema.<br/>
+        /// Gets the service object of the specified type. Schema itself acts as a service provider used to
+        /// create objects, such as graph types, requested by the schema.
```
Shane32:
```suggestion
        /// create objects, such as graph types, requested by the schema.<br/>
```

src/GraphQL/Types/Schema.cs
```diff
@@ -76,9 +77,30 @@ public void Initialize()
...
         /// <br/>
-        /// Note that most objects are created during schema initialization, which then have the same lifetime as the schema's lifetime.<br/>
+        /// Note that most objects are created during schema initialization, which then have the same lifetime
+        /// as the schema's lifetime.
```
Shane32:
```suggestion
        /// as the schema's lifetime.<br/>
```

src/GraphQL/Types/Schema.cs
```diff
@@ -76,9 +77,30 @@ public void Initialize()
...
+        /// <br/>
+        /// Other types created by the service provider may include directive visitors, middlewares, validation
+        /// rules, and name converters, among others.
```
Shane32:
```suggestion
        /// rules, and name converters, among others.<br/>
```

src/GraphQL/Types/Schema.cs
```diff
@@ -76,9 +77,30 @@ public void Initialize()
...
+        /// </summary>
+        /// <param name="serviceType">An object that specifies the type of service object to get.</param>
+        /// <returns>
+        /// A service object of type <paramref name="serviceType"/> or <c>null</c> if there is no service
+        /// object of type serviceType.
+        /// </returns>
+        object IServiceProvider.GetService(Type serviceType) => _services.GetService(serviceType);
+
+        /// <summary>
+        /// The service provider used to create objects, such as graph types, requested by the schema.
```
Shane32:
```suggestion
        /// The service provider used to create objects, such as graph types, requested by the schema.<br/>
```

src/GraphQL/Types/Schema.cs
```diff
@@ -76,9 +77,30 @@ public void Initialize()
...
+        /// <br/>
+        /// Note that most objects are created during schema initialization, which then have the same lifetime as the schema's lifetime.
```
Shane32:
```suggestion
        /// Note that most objects are created during schema initialization, which then have the same lifetime as the schema's lifetime.<br/>
```

src/GraphQL/Resolvers/AsyncFieldResolver.cs
sungam3r: done with doubled br on separate line
```
<br/><br/>
```

Shane32:
> Are you sure? And I see that your suggestion does not compile - `context` variable available only inside inner delegate.

Yes, I'm sure.  At best it's a dictionary lookup by the service provider for every field resolved.  You certainly could compile the call for old middleware if desired.

Shane32:
At least, considering how much time we've spent optimizing speed around here, this would be pretty significant.  In the big scheme of things, it's unlikely that any of these performance improvements are worth it.

sungam3r:
I probably don’t understand something. I cannot follow your example because it does not compile.

Shane32:
Ok I'll fix it hold on

Shane32:
Ok I see your issue.  The middleware builder does not have access to the schema.

sungam3r:
Bingo!

sungam3r:
Frankly I do not share your concerns about performance. Did you make benchmarks? Obtaining a middleware (or other object) from a service provider is not a long operation, especially in comparison with all other overheads.

Shane32:
I can, but frankly I think you should provide benchmarks for your PR.

Shane32:
`ApplyTo` has access to the `ISchema`.  Perhaps something can be done there.  I have to go now; if I think of something I'll let you know.

sungam3r:
I will try to approach the problem from the other side. The current `provider.GetService(middleware)` call supports all middleware lifetimes - singleton, scoped and even transient. Even assuming that your example would be compiled or we would find another way to get a `ServiceProvider` and make a single `GetService` call **before** inner delegate, then this would work correctly only for singleton scope.

I understand what you want to achieve. But you reason from the previous implementation. There really was an early creation of middleware  by `Activator.CreateInstance()` **before** inner delegate. But this code, in principle, could not work with scoped and transient middlewares. Now it can because it resolves the object **in the right place** - inside inner delegate.

> `ApplyTo` has access to the `ISchema`. Perhaps something can be done there.

When I wrote this PR, I asked **exactly the same questions**. The fact is that `ApplyTo` cannot help us for the reason described above - it is called at the wrong time, too early.

Shane32:
By your logic, we should create all graph types when they are needed, so they can be registered as scoped or transient.  I would agree with this philosophy, but it does not match the current project design.

Shane32:
The current project design is: Create the entire graph, resolvers -- everything -- when the schema is initialized.  Middleware should match the rest of the design, for better or worse.

sungam3r:
My logic cannot be extended to graph types, because their lifetime is tied to the lifetime of the schema, the schema caches them.

Shane32:
Exactly.  The schema should cache middleware.  There is no reason why we should apply one set of logic to graph types and another to middleware.  Especially since applying the middleware modifies the graphs.

sungam3r:
Agree

Shane32:
One final thought: since ISchema implements IServiceProvider, it's impossible to pull scoped middleware from a singleton schema.  And if the schema was scoped, we can initialize all the scoped middleware when the schema is initialized.  So, I think we just need to figure out the 'how' part, but there's no reason with the current design to have middleware pull from the service provider within the resolver (in concept)

Shane32:
How about this idea:
- `FieldMiddlewareBuilder` stores references to a list of `Func<ISchema, IFieldMiddleware>` or `Func<IServiceProvider, IFieldMiddleware>`
- `FieldMiddlewareBuilder.Use(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware)` wraps the parameter into an IFieldMiddleware for compatibility purposes and adds it to a list, for example `(schema) => new FuncFieldMiddleware(delegate)`
- `FieldMiddlewareBuilder.Use<T>() where T : IFieldMiddleware` creates a `Func<ISchema, IFieldMiddleware>` which pulls from the service provider to create the middleware, adding the func to the list - e.g. `(schema) => ((IServiceProvider)schema).GetRequiredService<T>()`
- `FieldMiddlewareBuilder.Use<T>()` where is T is pattern-based middleware support can be added
- `FieldMiddlewareBuilder.Use(IFieldMiddleware middleware)` can add `(schema) => middleware` to the list
- When `FieldMiddlewareBuilder.ApplyTo` is executed, the list of functions are called to build the middleware, and then the middleware is applied to each resolver

sungam3r:
I'm already doing something like this.

Shane32:
Cool

sungam3r:
1. Are you satisfied with the current state of <br> tags ?
2. I made middleware resolving one time when creating the schema, precisely speaking each time `ApplyTo` called.

I will spend some time thinking about this, perhaps in some places I will change the documentation.

src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
```diff
@@ -85,7 +86,7 @@ public void can_use_class()
         {
             _builder.Use<SimpleMiddleware>();
 
-            var result = _builder.Build().Invoke(_context).Result;
+            var result = _builder.Build(_context.Schema).Invoke(_context).Result;
```
sungam3r: `InvalidOperationException("Schema is null. Schema required for resolving middlewares from DI container.")` if not set

src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
```diff
@@ -9,17 +9,17 @@ namespace GraphQL.Instrumentation
 {
     public interface IFieldMiddlewareBuilder
     {
-        IFieldMiddlewareBuilder Use(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware);
+        IFieldMiddlewareBuilder Use(Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware);
```
sungam3r: @Shane32 in order to achieve the desired behavior, we will have to change the interface to pass schema down. There is an extension method that has the old signature so clients need only to recompile their code:
```c#
public static IFieldMiddlewareBuilder Use(this IFieldMiddlewareBuilder builder, Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware)
            => builder.Use((_, next) => middleware(next));
```

docs2/site/docs/getting-started/field-middleware.md
```diff
@@ -48,3 +52,96 @@ _.FieldMiddleware.Use(next =>
...
+Also, the middleware itself should be registered in DI:
+
+```csharp
+services.AddSingleton<InstrumentFieldsMiddleware>();
+```
+
+## Known issues
+
+Perhaps the most important thing with Field Middlewares that you should be aware of is that the
+default `DocumentExecuter` applies middlewares to the schema only once before the schema is
```
Shane32:
```suggestion
default `DocumentExecuter` applies middlewares to the schema only once while the schema is being
```
Just a suggestion; ok either way.  Technically it is before the `Schema.Initialize()` function runs -- but it's part of the initialization process by the document executer, and the only valid time to run it is once, before initialization.

src/GraphQL.Harness/CountFieldMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+using GraphQL.StarWars;
+using GraphQL.Types;
+using Microsoft.AspNetCore.Http;
+
+namespace Example
+{
+    /// <summary>
+    /// Example of Field Middleware with dependencies. When configured as singleton in DI
+    /// counts fields execution for the entire application lifetime. When configured as
+    /// scoped counts how many fields were executed when processing the single request.
```
Shane32:
```suggestion
    /// scoped (with a scoped schema), counts how many fields were executed when processing the single request.
```
Or something similar

Shane32:
One last comment: I'd recommend changing the documentation to show how to implement IFieldMiddleware, with a very brief paragraph on the backward-compatibility aspect of the code. I think we should encourage using the interface.

Shane32:
Other than my very brief comments on the documentation above, I think this looks good!

Shane32:
Oh, perhaps it might be in order to add a test that uses the old middleware pattern that we are trying to support.

src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
```diff
@@ -8,42 +8,84 @@ namespace GraphQL.Instrumentation
...
+                }
+
+                return builder.Use((schema, next) =>
+                {
+                    // Not an ideal solution, but at least it allows to work with custom schemas which are not inherited from Schema type
+                    object instance = CheckSchemaNotNull(schema) is IServiceProvider provider
+                        ? CheckNotNull(provider.GetService(middleware), middleware)
+                        : throw NotSupported(schema);
+
+                    return context => (Task<object>)methodInfo.Invoke(instance, new object[] { context, next });
```
Shane32:
```suggestion
                    var resolver = (Func<IResolveFieldContext, FieldMiddlewareDelegate, Task<object>>)methodInfo.CreateDelegate(typeof(Func<IResolveFieldContext, FieldMiddlewareDelegate, Task<object>>), instance);
                    return context => resolver(context, next);
```
It is possible, for a minute gain, to compile the resolver with an additional line of code.  I did test it to be sure it works.

Shane32: Doesn't really matter to me.  People should use `IFieldMiddleware` !

sungam3r:
> Oh, perhaps it might be in order to add a test that uses the old middleware pattern that we are trying to support.

There were already tests that fell on that branch of code:
```c#
public void can_use_class()
```

src/GraphQL/Instrumentation/FieldResolverBuilderExtensions.cs
sungam3r: Right. If one want to optimize, then this is the best option (and easier).

Shane32:
Oh sorry!  For some reason I thought that that class was modified with `: IFieldMiddleware`.

sungam3r:
@Shane32 Please look again to make sure everything suits you.

src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
```diff
@@ -7,63 +7,59 @@
...
             {
-                _components.Add(middleware);
+                _middlewares.Add(middleware);
             }
 
             return this;
         }
 
-        public FieldMiddlewareDelegate Build(FieldMiddlewareDelegate start = null)
+        public FieldMiddlewareDelegate Build(FieldMiddlewareDelegate start = null, ISchema schema = null)
```
Shane32: Just curious, why is this function public at all?  Certainly if it is private, schema can be required, as it will always receive one from `ApplyTo`.

docs2/site/docs/getting-started/field-middleware.md
```diff
@@ -48,3 +52,97 @@ _.FieldMiddleware.Use(next =>
   };
```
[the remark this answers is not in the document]
sungam3r: I thought about it and forgot. Done.

src/GraphQL/Instrumentation/FieldMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <param name="builder"></param>
+        /// <param name="middleware">Middleware instance.</param>
+        /// <returns>Reference to the same <see cref="IFieldMiddlewareBuilder"/>.</returns>
+        public static IFieldMiddlewareBuilder Use(this IFieldMiddlewareBuilder builder, IFieldMiddleware middleware)
+            => builder.Use(next => context => middleware.Resolve(context, next));
+
+        /// <summary>
+        /// Adds the specified delegate to the list of delegates that will be applied to the schema when invoking <see cref="ApplyTo(ISchema)"/>.
+        /// <br/><br/>
+        /// This is a compatibility shim when compiling delegates without schema specified. https://github.com/graphql-dotnet/graphql-dotnet/pull/1537#issuecomment-589798669
```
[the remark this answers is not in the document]
sungam3r: Then the reader will not understand the purpose of this overload and may begin to use it.

src/GraphQL/Instrumentation/FieldMiddlewareBuilderExtensions.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <returns>Reference to the same <see cref="IFieldMiddlewareBuilder"/>.</returns>
+        public static IFieldMiddlewareBuilder Use(this IFieldMiddlewareBuilder builder, System.Type middleware)
+        {
+            static Exception NotSupported(ISchema schema) => new NotSupportedException($"'{schema.GetType().FullName}' should implement 'IServiceProvider' interface for resolving middlewares.");
+
+            static T CheckNotNull<T>(T instance, System.Type type)
+            {
+                if (instance == null)
+                    throw new InvalidOperationException(@$"Field middleware of type '{type.FullName}' must be registered in DI container.
+If you called 'FieldMiddlewareBuilder.Build()' manually set the 'schema' parameter.");
```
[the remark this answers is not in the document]
sungam3r: Because I made a mistake with the method.

src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
sungam3r: I am sure that in this project there are a huge number of APIs that can and should be made private. I even saw a corresponding issue for this. This simplifies understanding of important APIs, and not important ones go into implementation details. Each case must be considered separately. I am not familiar with the history of the project so much as to understand 100% which APIs can be removed and which not. @joemcbride can we make this method `internal`?

src/GraphQL/Instrumentation/FieldMiddlewareBuilderExtensions.cs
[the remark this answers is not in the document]
sungam3r: Removed link

sungam3r:
@Shane32 Thanks for the wording. So the only question remains - `internal Build()`.

Shane32:
Just the question about `internal Build()` remains.

Shane32:
If we can make it `internal`, I'd remove `= null` from the `schema` parameter, and then delete the latter half of that one error message -- you know what I mean.

sungam3r:
I think we can make it `internal` safely.

Shane32:
> I also ask you to review my changes in the documentation since I am not a native speaker.

I must say, you have an excellent command of the English language 👍 👍

sungam3r:
Glory to google translator!

src/GraphQL.Harness/Startup.cs
sungam3r: @Shane32 Strange, but without this registration in the server project, the middleware still works. I don’t understand why this is happening.

sungam3r: I figured it out - `DefaultServiceProvider`

Issue #369, opened 2017-07-10
Add XML documentation comments for VS IntelliSense and "Go To Definition"


antmdvs:
We should ship [XML Documentation Comments](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/xml-documentation-comments). This would improve the discoverability and understandability of the APIs. For example, there are currently 6+ `Field` registration methods, which could be confusing at best/overwhelming at worst to someone trying to navigate the API for the first time.

drobertson123:
Just my 2 cents. I would certainly love more documentation, but in my experience, it seems like XML Documentation has a tendency to overwhelm the code and make it harder to follow.

In many cases, the XML docs just state what is already there and little effort is put into adding the knowledge and context that needed to understand what is actually going on. I can figure out a method signature easily enough. What I would like to get from any documentation is the context for using the code. What is it used for, is there anything special in the way it works, are there side effects from interacting with the objects/code, etc...

I am not saying that XML docs are horrible or that in this case they would be a bad idea, but there is a tradeoff when using them. Code files explode in size with less code per screen space vs the comments in the file.

Aside from disagreeing (generally) with the idea of XML docs I think more documentation and information would be a wonderful thing. I am happy to pitch in to help some if that is useful.

Personally, I like the idea of more of a Living Documentation approach to the documentation. I believe Marten uses this. The basic concept is combining markdown files with a tool that renders them into a website containing the documentation (github pages). The markdown has syntax for linking directly to code examples that are rendered into the HTML documents as you go. The focus would be on writing good documentation of the software concepts and using descriptive Tests as the embedded code to show how things are done.

If the tests get out of sync with the markdown due to code changes it becomes fairly obvious and the markdown can be revised. Knowing that the tests are likely to be visible in the documentation encourages people to make well written and expressive tests (hopefully).

[Storyteller](http://storyteller.github.io/documentation/docs/) seems to have a good system for this. Jeremy Miller has an article on doing this. [How I’m Documenting OSS Projects](https://jeremydmiller.com/2016/01/05/how-im-documenting-oss-projects/)

joemcbride:
@drobertson123  I have started some markdown doc files for basically that purpose.  I just haven't gotten around to generating a site for them yet.  I do like what the st-docs is doing.

https://github.com/graphql-dotnet/graphql-dotnet/tree/master/docs

drobertson123:
[...]

PR #1538, opened 2020-02-16
Add xml comments to IDocumentExecuter

Shane32:
See 'files changed' for suggested change

### DIFF
--- a/src/GraphQL/Execution/IDocumentExecuter.cs
+++ b/src/GraphQL/Execution/IDocumentExecuter.cs
@@ -5,8 +5,16 @@ namespace GraphQL
 {
     public interface IDocumentExecuter
     {
+        /// <summary>
+        /// Executes a GraphQL request and returns the result
+        /// </summary>
+        /// <param name="options">The options of the execution</param>
         Task<ExecutionResult> ExecuteAsync(ExecutionOptions options);
 
+        /// <summary>
+        /// Executes a GraphQL request and returns the result
+        /// </summary>
+        /// <param name="configure">A delegate which configures the execution options</param>
         Task<ExecutionResult> ExecuteAsync(Action<ExecutionOptions> configure);
     }
 }
--- a/src/GraphQL/Execution/IDocumentExecuter.cs
+++ b/src/GraphQL/Execution/IDocumentExecuter.cs
@@ -9,7 +9,7 @@ namespace GraphQL
     /// <br/>
     /// Typical implementation starts metrics if enabled, relies on IDocumentBuilder to parse the query,
     /// executes document listeners if attached, selects an execution strategy, and executes the query
-    /// via IExecutionStrategy.  Unhandled exceptions are handled as appropriate for the selected options.
+    /// via IExecutionStrategy. Unhandled exceptions are handled as appropriate for the selected options.
     /// </summary>
     public interface IDocumentExecuter
     {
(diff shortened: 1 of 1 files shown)


sungam3r:
What is the purpose of this PR? I remember that @joemcbride was negative about excessive xml commenting. See #369:

> In many cases, the XML docs just state what is already there and little effort is put into adding the knowledge and context that needed to understand what is actually going on. I can figure out a method signature easily enough.

Shane32:
I am requesting limited XML comments to the public interfaces (and data-only classes where prudent) so that their purpose can be clarified.  This does not clutter any code, per se, and has a benefit that any classes that inherit these methods will inherit their XML comments as well.  It should be extremely useful.  I do not think this goes against anything @joemcbride stated in #369 including his statement (taken out of context):

> I think XML docs could help though won't necessarily solve the problem of knowing which one to use and why, without perhaps adding tons of docs in the XML comments which I think I would be opposed to.

We can asses the changes on a case-by-case basis, as I will post separate PRs for separate files.  Builders perhaps are not such a good candidate, as they may be well documented on the site.  But it will really help people understand the "internal workings" of graphql-dotnet if the interfaces required for its core operation (ISchema, IDocumentExecuter, etc) have xml comments.  For example, what's the difference between a IDocumentExecuter and a IExecutionStrategy?  They both have a single method ExecuteAsync, and both return a ExecutionResult.  There is no documentation for this.  You and I may know what they do, but a few xml comments will sufficiently document this for the project to help others understand.

sungam3r:
> For example, what's the difference between a IDocumentExecuter and a IExecutionStrategy? They both have a single method ExecuteAsync, and both return a ExecutionResult

Ok. I read the comments on the methods of both interfaces, but still did not understand the difference. if you add documentation, it should be meaningful in order to be useful.

Shane32:
Agreed

sungam3r:
Here it would be possible to describe the areas of their responsibility and the relationship between them (`IDocumentExecuter` calls `IExecutionStrategy` as one of steps, etc)

Shane32:
@sungam3r Please provide comments to the suggested changes.  I'd be happy to work with you on suggested text, wording, amount of detail, etc.  I do feel some documentation in this area is valuable to the project.

src/GraphQL/Execution/IDocumentExecuter.cs
```diff
@@ -3,10 +3,26 @@
 
 namespace GraphQL
 {
+    /// <summary>
+    /// Processes an entire GraphQL request, given an input GraphQL request string.  This is intended to
```
sungam3r:
```suggestion
    /// Processes an entire GraphQL request, given an input GraphQL request string. This is intended to
```

src/GraphQL/Execution/IDocumentExecuter.cs
```diff
@@ -3,10 +3,26 @@
...
+    /// be called by user code to process a query.<br/>
+    /// <br/>
+    /// Typical implementation starts metrics if enabled, relies on IDocumentBuilder to parse the query,
+    /// executes document listeners if attached, selects an execution strategy, and executes the query
+    /// via IExecutionStrategy.  Unhandled exceptions are handled as appropriate for the selected options.
```
sungam3r:
```suggestion
    /// via IExecutionStrategy. Unhandled exceptions are handled as appropriate for the selected options.
```

sungam3r:
Now your comment makes much more sense. Now it’s clear not only what’s happening from it, but also that this interface is intended to be called by user code. If you intend to continue to add documentation (this PR or others), then I recommend using `<see cref="XXX"/>` notation instead of just `XXX`.

PR #1541, opened 2020-02-16
Add serialization benchmarks

sungam3r:
@benmccallum This may interest you. Perhaps I made a mistake somewhere or we didn’t take into account something. The results are disappointing.

``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.17134.1304 (1803/April2018Update/Redstone4)
Frequency=2343753 Hz, Resolution=426.6661 ns, Timer=TSC
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT


```
|                 Method |     Mean |    Error |   StdDev |   Median | Ratio | RatioSD |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------- |---------:|---------:|---------:|---------:|------:|--------:|--------:|------:|------:|----------:|
|         NewtonsoftJson | 599.5 us | 45.61 us | 90.03 us | 557.6 us |  1.00 |    0.00 |  7.8125 |     - |     - |  13.36 KB |
| NewtonsoftJsonIndented | 708.6 us |  5.21 us |  4.35 us | 708.3 us |  1.00 |    0.16 |  8.7891 |     - |     - |  13.48 KB |
|         SystemTextJson | 734.8 us |  7.77 us |  7.27 us | 734.9 us |  1.08 |    0.18 | 33.2031 |     - |     - |  52.08 KB |
| SystemTextJsonIndented | 819.0 us |  7.96 us |  7.45 us | 820.7 us |  1.20 |    0.20 | 33.2031 |     - |     - |  52.08 KB |


benmccallum:
Ahhhh. That's weird. Hey @davidfowl / @JamesNK, who could help us dig deeper into this?

benmccallum:
Essentially we've got two serialization implementations of an `IDocumentWriter` that are very similar, but we're not seeing the expected perf benefits. 

The Newtonsoft.Json implementation seems to be using an array pool, perhaps that could be the difference?

sungam3r:
> Would be surprised if the S.T.Json implementation doesn't do something similar under the hood though.

S.T.Json is heavily optimized with a lot of pooling. I looked into the source code. I think the available options are:
1. We made a fundamental mistake somewhere in our code.
2. I wrote a test case incorrectly.
3. Large serialization object degrades performance in S.T.Json when serializing to stream (least likely case).

src/GraphQL.Benchmarks/NullStream.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.IO;
+
+namespace GraphQL.Benchmarks
+{
+    internal class NullStream : Stream
```
davidfowl: Stream.Null?

sungam3r: When I wrote this, I was in a hurry and did not think about Stream.Null.

sungam3r: Ok. With Stream.Null nothing changed.

src/GraphQL.Benchmarks/Benchmarks/SerializationBenchmark.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            return _executer.ExecuteAsync(options =>
+            {
+                options.Schema = schema;
+                options.Query = query;
+            }).GetAwaiter().GetResult();
+        }
+
+        [Benchmark(Baseline = true)]
+        public void NewtonsoftJson() => _nsjWriter.WriteAsync(_stream, _result).Wait();
```
pekkah: Nit: benchmarkdotnet does support tasks. Example https://github.com/pekkah/graphql-benchmarks/blob/master/src/benchmark/QueryBenchmarks.cs

sungam3r:
Results for `{ hero { id name } }`
``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.17134.1304 (1803/April2018Update/Redstone4)
Frequency=2343753 Hz, Resolution=426.6661 ns, Timer=TSC
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT


```
|                 Method |     Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------- |---------:|----------:|----------:|------:|--------:|-------:|------:|------:|----------:|
|         NewtonsoftJson | 3.929 us | 0.0833 us | 0.1738 us |  1.00 |    0.00 | 0.4044 |     - |     - |     640 B |
| NewtonsoftJsonIndented | 3.799 us | 0.0745 us | 0.1224 us |  0.96 |    0.06 | 0.4807 |     - |     - |     764 B |
|         SystemTextJson | 2.242 us | 0.0463 us | 0.1194 us |  0.57 |    0.04 | 0.2403 |     - |     - |     380 B |
| SystemTextJsonIndented | 2.220 us | 0.0375 us | 0.0351 us |  0.56 |    0.03 | 0.2403 |     - |     - |     380 B |

sungam3r:
Results for
```graphql
{
  hero
  {
    name
    id
    friends
    {
      id
      name
      friends
      {
        id
        name
        friends
        {
          id
          name
        }
      }
    }
  }
}
```

``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.17134.1304 (1803/April2018Update/Redstone4)
Frequency=2343753 Hz, Resolution=426.6661 ns, Timer=TSC
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT


```
|                 Method |     Mean |    Error |   StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------- |---------:|---------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|         NewtonsoftJson | 10.94 us | 0.167 us | 0.139 us |  1.00 |    0.00 | 0.7629 |     - |     - |   1.18 KB |
| NewtonsoftJsonIndented | 13.57 us | 0.176 us | 0.165 us |  1.24 |    0.02 | 0.8392 |     - |     - |    1.3 KB |
|         SystemTextJson | 10.28 us | 0.120 us | 0.112 us |  0.94 |    0.02 | 0.7629 |     - |     - |   1.17 KB |
| SystemTextJsonIndented | 11.63 us | 0.166 us | 0.138 us |  1.06 |    0.02 | 0.7629 |     - |     - |   1.17 KB |

The trend is obvious.

benmccallum:
Obviously disappointing. :P

benmccallum:
Do they have benchmarks that they've shared so we can see if we are getting similar results?

sungam3r:
They?

benmccallum:
Microsoft and the team that wrote S.T.Json.

sungam3r:
Top allocations:
image

sungam3r:
Getting closer
```c#
   private static JsonEncodedText EncodeHelper(ReadOnlySpan<byte> utf8Value, JavaScriptEncoder encoder)
        {
            int idx = JsonWriterHelper.NeedsEscaping(utf8Value, encoder);

            if (idx != -1)
            {
                return new JsonEncodedText(GetEscapedString(utf8Value, idx, encoder));
            }
            else
            {
                return new JsonEncodedText(utf8Value.ToArray());
            }
        }
```

sungam3r:
But... we are on the `else` branch, so nothing needs to be escaped. Why `ToArray` ??? @davidfowl I really don’t understand something.

sungam3r:
S.T.J version : 0f7f38c4fd323b26da10cce95f857f77f0f09b48

sungam3r:
And also conversion to a string:
image

sungam3r:
image

sungam3r:
I’m very upset because I don’t understand anything.

sungam3r:
Also this place:
image

sungam3r:
My preliminary conclusion is that S.T.J is basically not designed for efficient work with dictionaries. All our response structures in `ExecutionResult.Data` are dictionaries.

I **really hope** that I'm wrong.

sungam3r:
I wrote a test for nested dictionaries, which showed that S.T.J is very fast. I think the problem is `ExecutionResultJsonConverter`.

sungam3r:
Yep, `JsonSerializer.Serialize(writer, data)` is the cause of problems. @benmccallum see updated PR. Now results for IntrospectionQuery are:
``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.14393.3384 (1607/AnniversaryUpdate/Redstone1)
Frequency=3515629 Hz, Resolution=284.4441 ns, Timer=TSC
.NET Core SDK=3.1.100
  [Host]     : .NET Core 3.1.0 (CoreCLR 4.700.19.56402, CoreFX 4.700.19.56404), X64 RyuJIT


```
|                 Method |     Mean |   Error |   StdDev |   Median | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------- |---------:|--------:|---------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|         NewtonsoftJson | 303.0 us | 6.04 us | 14.24 us | 296.4 us |  1.00 |    0.00 | 5.3711 |     - |     - |   23184 B |
| NewtonsoftJsonIndented | 379.9 us | 7.56 us | 15.44 us | 372.0 us |  1.25 |    0.07 | 5.3711 |     - |     - |   23344 B |
|         SystemTextJson | 127.9 us | 1.71 us |  1.52 us | 127.2 us |  0.40 |    0.02 |      - |     - |     - |     200 B |
| SystemTextJsonIndented | 155.2 us | 4.17 us |  4.10 us | 153.9 us |  0.49 |    0.02 |      - |     - |     - |     200 B |

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
```diff
@@ -29,8 +30,64 @@ private void WriteData(Utf8JsonWriter writer, ExecutionResult result)
...
+                        writer.WriteStringValue(s);
+                    else
+                        WriteDataRecursively(writer, null, item);
+                }
+
+                writer.WriteEndArray();
+            }
+            else
+            {
+                // Need to avoid this call by all means! The question remains open - why this API so expensive?
```
sungam3r: @davidfowl could you explain please?  Another question that interests me is whether the serialization options for ` writer.WriteString(name, s);` and other low level `WriteXXX` APIs apply? For example `PropertyNamingPolicy`.

benmccallum: Yea the docs are a bit funky saying you shouldn't pass down the options as it could cause a stackoverflow or something. So you have to recreate them at every level it seemed to say. I avoided that by just not adding any tests that test for it :P

Pretty sure the docs say to use the static JsonSerializer.Serialize in examples, so it's kinda weird this makes such a difference.

benmccallum:
Can we avoid using a custom converter entirely and put the "logic" to match the spec in a non-serialization concerned method? The spec is quite hard to follow, but when I read it last it seemed like some of the required logic would've been better in the execution anyway, as there's cases where depending on where the error happened, data prop is written or not, etc.

Might be hard to distinguish between "data is null but plz write it out" vs "data is null but don't write it out".

benmccallum:
This is also why I don't necessarily agree with exposing the whole json options out to users as we need to limit what they can tweak to what's going to still conform to spec. E.g. The null handling setting shouldn't be exposed to them IMO.

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
sungam3r: > Another question that interests me is whether the serialization options for writer.WriteString(name, s); and other low level WriteXXX APIs apply? For example PropertyNamingPolicy.

It looks like the answer is **no**. `PropertyNamingPolicy` is called by `JsonPropertyInfo.DeterminePropertyName` when building `JsonClassInfo` from `Type`.

Therefore the adventure continues.

sungam3r:
You can get rid of the converter only if the structure of the classes (ExecutionResult, Errors, etc) exactly matches the final structure of the response. Also there is a lot of properties which should be ignored - `Document`, `Operation`, `Perf` . It would be possible to solve this problem with attributes, but you yourself understand that this is not an option. We followed the exact same path in the server project. `ExecutionResult` class is rather large and solves common problems; it is not intended ONLY for serialization. Therefore, the appearance of a special converter that knows what it needs to be serialized is quite natural in my opinion.

sungam3r:
> This is also why I don't necessarily agree with exposing the whole json options out to users as we need to limit what they can tweak to what's going to still conform to spec. E.g. The null handling setting shouldn't be exposed to them IMO.

I understand what you are talking about. But I do not see any particular danger in providing a complete set of options for the configuration. If you wish, you can always shoot yourself in the foot.

benmccallum:
> We followed the exact same path in the server project. ExecutionResult class is rather large and solves common problems; it is not intended ONLY for serialization.

What if we had a final "type" before it was serialized, that was in fact just a `Dictionary<string, object>`, then we could add into the dictionary just the keys we need (whether they be null or actual values) and then the writer is always configured to write null property values out? 

Here's the spec on the response: http://spec.graphql.org/June2018/#sec-Response

The tricky bits are:
> If an error was encountered before execution begins, the data entry should not be present in the result.
> If an error was encountered during the execution that prevented a valid response, the data entry in the response should be null.

(And the corresponding rules around population of the errors property)

sungam3r:
Perhaps something would be simplified. But something got complicated. We would have to change the design. Now everything works as it should, so I don’t see a win.

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
sungam3r: In fact, some of the options apply, for example `WriteIndented`.

sungam3r:
@benmccallum Results after final edits. It is clearly seen that the allocated memory size does not depend on the size of the sezialized result. I believe that despite the remaining questions, we can safely merge this. I would not like to keep for a long time in the public domain a version of the "optimized" serializer, which works an order of magnitude worse than the old one :)

``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.17134.1304 (1803/April2018Update/Redstone4)
Frequency=2343753 Hz, Resolution=426.6661 ns, Timer=TSC
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT


```
|                 Method |          Code |         Mean |       Error |      StdDev | Ratio |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------- |-------------- |-------------:|------------:|------------:|------:|-------:|------:|------:|----------:|
|         **NewtonsoftJson** | **Introspection** | **491,476.4 ns** | **3,208.01 ns** | **3,000.78 ns** |  **1.00** | **7.8125** |     **-** |     **-** |   **13441 B** |
| NewtonsoftJsonIndented | Introspection | 643,490.6 ns | 5,408.66 ns | 4,794.64 ns |  1.31 | 7.8125 |     - |     - |   13565 B |
|         SystemTextJson | Introspection | 191,588.1 ns |   811.19 ns |   719.10 ns |  0.39 |      - |     - |     - |     116 B |
| SystemTextJsonIndented | Introspection | 249,350.6 ns | 1,404.95 ns | 1,314.19 ns |  0.51 |      - |     - |     - |     116 B |
|                        |               |              |             |             |       |        |       |       |           |
|         **NewtonsoftJson** |        **Middle** |  **11,468.6 ns** |    **50.97 ns** |    **45.18 ns** |  **1.00** | **0.7629** |     **-** |     **-** |    **1212 B** |
| NewtonsoftJsonIndented |        Middle |  14,164.3 ns |    45.52 ns |    40.35 ns |  1.24 | 0.8392 |     - |     - |    1336 B |
|         SystemTextJson |        Middle |   2,893.4 ns |    12.10 ns |    11.32 ns |  0.25 | 0.0725 |     - |     - |     116 B |
| SystemTextJsonIndented |        Middle |   4,067.7 ns |    28.89 ns |    22.56 ns |  0.35 | 0.0687 |     - |     - |     116 B |
|                        |               |              |             |             |       |        |       |       |           |
|         **NewtonsoftJson** |         **Small** |   **2,394.8 ns** |    **18.50 ns** |    **17.30 ns** |  **1.00** | **0.4044** |     **-** |     **-** |     **640 B** |
| NewtonsoftJsonIndented |         Small |   2,838.7 ns |    14.11 ns |    13.20 ns |  1.19 | 0.4845 |     - |     - |     764 B |
|         SystemTextJson |         Small |     792.6 ns |     4.50 ns |     3.51 ns |  0.33 | 0.0734 |     - |     - |     116 B |
| SystemTextJsonIndented |         Small |   1,066.6 ns |    21.49 ns |    43.40 ns |  0.44 | 0.0734 |     - |     - |     116 B |

PR #1543, opened 2020-02-17
Add xml comments to ExecutionOptions

Shane32:
Also fixed note in IDocumentExexuter

### DIFF
--- a/src/GraphQL/Execution/ExecutionOptions.cs
+++ b/src/GraphQL/Execution/ExecutionOptions.cs
@@ -43,7 +43,7 @@ namespace GraphQL
         public IEnumerable<IValidationRule> ValidationRules { get; set; }
 
         /// <summary>
-        /// User-defined context to be passed to the field resolvers.<br/>
+        /// Mutable user-defined context to be passed and shared by all field resolvers.<br/>
         /// <br/>
         /// A custom implementation of <see cref="IDictionary{TKey, TValue}">IDictionary</see> may be
         /// used in place of the default <see cref="Dictionary{TKey, TValue}">Dictionary</see>.
--- a/src/GraphQL/Execution/ExecutionOptions.cs
+++ b/src/GraphQL/Execution/ExecutionOptions.cs
@@ -63,7 +63,7 @@ namespace GraphQL
         /// <summary>Field and argument names are sanitized by the provided <see cref="IFieldNameConverter"/>; defaults to <see cref="CamelCaseFieldNameConverter"/></summary>
         public IFieldNameConverter FieldNameConverter { get; set; } = CamelCaseFieldNameConverter.Instance;
 
-        /// <summary>Allows unhandled <see cref="Exception"/> stack traces to be serialized into GraphQL query result json; defaults to only <see cref="Exception.Message"/></summary>
+        /// <summary>Allows unhandled <see cref="Exception"/> stack traces to be serialized into GraphQL query result json along with exception messages; defaults to only <see cref="Exception.Message"/></summary>
         public bool ExposeExceptions { get; set; }
 
         /// <summary>This setting essentially allows Apollo Tracing. Disabling will increase performance.</summary>
--- a/src/GraphQL/Execution/ExecutionOptions.cs
+++ b/src/GraphQL/Execution/ExecutionOptions.cs
@@ -24,7 +24,7 @@ namespace GraphQL
         /// <summary>GraphQL query to parse and execute; required</summary>
         public string Query { get; set; }
 
-        /// <summary>GraphQL query operation name; defaults to first operation defined in query</summary>
+        /// <summary>GraphQL query operation name; optional, defaults to first (if any) operation defined in query</summary>
         public string OperationName { get; set; }
 
         /// <summary>Parsed GraphQL request; will parse <see cref="Query"/> if not set</summary>
--- a/src/GraphQL/Execution/ExecutionOptions.cs
+++ b/src/GraphQL/Execution/ExecutionOptions.cs
@@ -27,11 +27,7 @@ namespace GraphQL
         /// <summary>GraphQL query operation name; optional, defaults to first (if any) operation defined in query</summary>
         public string OperationName { get; set; }
 
-        /// <summary>
-        /// Parsed GraphQL request; will parse <see cref="Query"/> if not set.<br/>
-        /// <br/>
-        /// Can be used when implementing a cache of parsed GraphQL requests (a <see cref="Language.AST.Document"/>)
-        /// </summary>
+        /// <summary>Parsed GraphQL request; can be used to increase performance when implementing a cache of parsed GraphQL requests (a <see cref="Language.AST.Document"/>). If not set, it will be parsed from <see cref="Query"/></summary>
         public Document Document { get; set; }
 
         /// <summary>Input variables to GraphQL request</summary>
--- a/src/GraphQL/Execution/ExecutionOptions.cs
+++ b/src/GraphQL/Execution/ExecutionOptions.cs
@@ -12,6 +12,7 @@ using GraphQL.Validation.Complexity;
 
 namespace GraphQL
 {
+    /// <summary>Configuration options to be passed to <see cref="IDocumentExecuter"/> to execute a query</summary>
     public class ExecutionOptions : IProvideUserContext
     {
         /// <summary>Schema of graph to use; required<br/><br/>Schema will be initialized if has not yet been initialized.</summary>
--- a/src/GraphQL/Execution/ExecutionOptions.cs
+++ b/src/GraphQL/Execution/ExecutionOptions.cs
@@ -27,7 +27,10 @@ namespace GraphQL
         /// <summary>GraphQL query operation name; optional, defaults to first (if any) operation defined in query</summary>
         public string OperationName { get; set; }
 
-        /// <summary>Parsed GraphQL request; will parse <see cref="Query"/> if not set</summary>
+        /// <summary>
+        /// Parsed GraphQL request; will parse <see cref="Query"/> if not set.<br/>
+        /// <br/>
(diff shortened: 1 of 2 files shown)


src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -12,22 +12,39 @@
...
+        /// <summary>Schema of graph to use; required<br/><br/>Schema will be initialized if has not yet been initialized.</summary>
         public ISchema Schema { get; set; }
+
+        /// <summary>Object to pass to the <see cref="IResolveFieldContext.Source"/> property of first-level resolvers</summary>
         public object Root { get; set; }
+
+        /// <summary>GraphQL query to parse and execute; required</summary>
         public string Query { get; set; }
+
+        /// <summary>GraphQL query operation name; defaults to first operation defined in query</summary>
```
sungam3r:
```suggestion
        /// <summary>GraphQL query operation name; optional, defaults to first (if any) operation defined in query</summary>
```

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -12,22 +12,39 @@
...
         public string OperationName { get; set; }
+
+        /// <summary>Parsed GraphQL request; will parse <see cref="Query"/> if not set</summary>
```
sungam3r: Here it is worth saying why both objects Document and Query are presented in options

Shane32: Good question ... why are they?  I noticed that if Query is null it will throw an exception, regardless of whether Document is set.  Not sure exactly how I should change this comment.

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -12,22 +12,39 @@
...
         public Document Document { get; set; }
+
+        /// <summary>Input variables to GraphQL request</summary>
         public Inputs Inputs { get; set; }
-        
+
+        /// <summary><see cref="System.Threading.CancellationToken">CancellationToken</see> of request; optional</summary>
```
sungam3r:
```suggestion
        /// <summary><see cref="System.Threading.CancellationToken">CancellationToken</see> to cancel the request at any stage of its execution; defaults to <see cref="System.Threading.CancellationToken.None"/></summary>
```

Shane32: If you think this is a good change, I don't mind changing it.  CancellationToken.None is the same as leaving it undefined, as CancellationToken is a struct.  So I think of it as an optional parameter.

I like this part for sure: `to cancel the request at any stage of its execution`

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -12,22 +12,39 @@
...
         public CancellationToken CancellationToken { get; set; }
 
-        /// <summary>
-        /// Note if not set then standard list of validation rules will be used.
-        /// </summary>
+        /// <summary>Validation rules to be used by the <see cref="IDocumentValidator"/>; defaults to standard list of of validation rules - see <see cref="DocumentValidator.CoreRules"/></summary>
         public IEnumerable<IValidationRule> ValidationRules { get; set; }
 
+        /// <summary>
+        /// User-defined variables to be passed to the field resolvers.<br/>
```
sungam3r:
```suggestion
        /// User-defined context to be passed to the field resolvers.<br/>
```
The word "variables" can be confusing to reader as there is `Inputs` property for variables.

Shane32:
Simply writing these comments helped me understand the execution process better.  I feel that these comments, few as they are, are an important asset to the project.

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -37,35 +54,35 @@ public class ExecutionOptions : IProvideUserContext
...
+        /// <summary>Complexity constraints for <see cref="IComplexityAnalyzer"/> to use to validate maximum query complexity</summary>
         public ComplexityConfiguration ComplexityConfiguration { get; set; }
 
+        /// <summary>A list of <see cref="IDocumentExecutionListener"/>s, enabling code to be executed at various points during the processing of the GraphQL query</summary>
         public List<IDocumentExecutionListener> Listeners { get; } = new List<IDocumentExecutionListener>();
 
+        /// <summary>Field and argument names are sanitized by the provided <see cref="IFieldNameConverter"/>; defaults to <see cref="CamelCaseFieldNameConverter"/></summary>
         public IFieldNameConverter FieldNameConverter { get; set; } = CamelCaseFieldNameConverter.Instance;
 
+        /// <summary>Allows unhandled <see cref="Exception"/> stack traces to be serialized into GraphQL query result json; defaults to only <see cref="Exception.Message"/></summary>
```
sungam3r:
```suggestion
        /// <summary>Allows unhandled <see cref="Exception"/> stack traces to be serialized into GraphQL query result json along with exception messages; defaults to only <see cref="Exception.Message"/></summary>
```

src/GraphQL/Execution/ExecutionOptions.cs
Shane32: Seems fine anyway however...

src/GraphQL/Execution/ExecutionOptions.cs
sungam3r: I think the ability to set `Document` in options is to optimize performance. If there is some caching on the caller side, then it can set the already created and cached document so there will be no need in `_documentBuilder.Build(options.Query);`. But these are my assumptions. @joemcbride It is right?

src/GraphQL/Execution/ExecutionOptions.cs
sungam3r: I know that it is struct but despite this we could set this property to some value other than None. For example, we could by default set cancellation after 60 seconds. Therefore, it’s better to clearly say that it is equal to None.

sungam3r:
> Simply writing these comments helped me understand the execution process better. I feel that these comments, few as they are, are an important asset to the project.

The process of writing comments is always more interesting and longer than reading them (if you approach it with due attention and patience).

src/GraphQL/Execution/ExecutionOptions.cs
Shane32: It is certainly a valid example.  I can add that into the comment.

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -12,22 +12,39 @@
...
         public CancellationToken CancellationToken { get; set; }
 
-        /// <summary>
-        /// Note if not set then standard list of validation rules will be used.
-        /// </summary>
+        /// <summary>Validation rules to be used by the <see cref="IDocumentValidator"/>; defaults to standard list of of validation rules - see <see cref="DocumentValidator.CoreRules"/></summary>
         public IEnumerable<IValidationRule> ValidationRules { get; set; }
 
+        /// <summary>
+        /// User-defined context to be passed to the field resolvers.<br/>
```
sungam3r:
```suggestion
        /// Mutable user-defined context to be passed and shared by all field resolvers.<br/>
```

src/GraphQL/Execution/ExecutionOptions.cs
sungam3r: OK

Shane32: (Although in that example, they should probably provide a custom IDocumentBuilder to perform the caching.)

sungam3r: Caching can be implemented anywhere, including `IDocumentBuilder`.

Shane32: How does the updated comment look now?

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -12,22 +12,42 @@
...
+        /// <summary>GraphQL query to parse and execute; required</summary>
         public string Query { get; set; }
+
+        /// <summary>GraphQL query operation name; optional, defaults to first (if any) operation defined in query</summary>
         public string OperationName { get; set; }
+
+        /// <summary>
+        /// Parsed GraphQL request; will parse <see cref="Query"/> if not set.<br/>
+        /// <br/>
+        /// Can be used when implementing a cache of parsed GraphQL requests (a <see cref="Language.AST.Document"/>)</summary>
```
sungam3r:
```suggestion
        /// Can be used to increase performance when implementing a cache of parsed GraphQL requests (a <see cref="Language.AST.Document"/>). If not set, it will be parsed from <see cref="Query"/></summary>
```

src/GraphQL/Execution/ExecutionOptions.cs
sungam3r: I posted new suggestion.

src/GraphQL/Execution/ExecutionOptions.cs
Shane32: Just to clarify, are you suggesting I remove the first line also?

So:
```
/// <summary>Can be used to increase performance when implementing a cache of parsed GraphQL requests (a <see cref="Language.AST.Document"/>). If not set, it will be parsed from <see cref="Query"/></summary>
```

sungam3r:
```
/// <summary>Parsed GraphQL request; can be used to increase performance when implementing a cache of parsed GraphQL requests (a <see cref="Language.AST.Document"/>). If not set, it will be parsed from <see cref="Query"/></summary>
```

Shane32: Looks good. Updated as suggested.

sungam3r:
Thanks for your contribution. I hope that I have not tired you too much with my nitpicking.

Shane32:
Lol no I don't mind and your comments are good.  I hope you don't mind my differing points of view occasionally ;-)

Issue #1545, opened 2020-02-17
DefaultFieldNameConverter question


Shane32:
@sungam3r I'm trying to understand the `DefaultFieldNameConverter` and `PascalCaseFieldNameConverter`.  It appears to only apply camel case to "instrospection types".  However, this is searching a private fixed array, always containing a single static type `SchemaIntrospection`.  As a private array, it can't be changed.  And static types can't be created.  Am I missing something or does this make no sense at all?

sungam3r:
You have the same question that I once had. In fact, everything is simple - the **parent** type is checked. And the converter is applied to **child** fields. Also look into `GraphTypesLookup`:
```c#
 var introspectionType = typeof(SchemaIntrospection);

            lookup.HandleField(introspectionType, SchemaIntrospection.SchemaMeta, ctx);
            lookup.HandleField(introspectionType, SchemaIntrospection.TypeMeta, ctx);
            lookup.HandleField(introspectionType, SchemaIntrospection.TypeNameMeta, ctx);
```

```c#
 private void HandleField(Type parentType, FieldType field, TypeCollectionContext context)
```

Shane32:
Thanks. I felt it must have made sense somehow but couldn’t figure it out!

Shane32:
But ... the introspection types start with `__` -- so this code still actually does nothing, right???

sungam3r:
Why nothing?

Shane32:
The `Name` property of `SchemaMetaFieldType` is set to `"__schema"`.  `HandleField` passes the parent type and the `Name` property of the field to the `IFieldConverter`.  If this then ran this through `StringExtensions.ToPascalCase` (it doesn't currently), it would grab the first character and try to capitalize it.  The first character is `'_'`.  So it would return the original string `"__schema"`.  Same with `TypeMetaFieldType` and `TypeNameMetaFieldType`, the only three types this code applies to.  Right?  So it would seem that you could run any field name through `ToPascalCase` without the name of those introspection type names being affected.  And they certainly don't need to run through `ToCamelCase` as they are predefined with the correct case.

Shane32:
Maybe there's more I'm not seeing.  I'll run some tests.

sungam3r:
I think I understood your question.

>  it would grab the first character and try to capitalize it. The first character is '_'

[...]

PR #1547, opened 2020-02-18
Bug fix: introspection includeDeprecated field argument does not work with PascalCaseFieldNameConverter

Shane32:
`GetArgument` converts argument names via the `NameFieldConverter`.  As introspection field names and arguments are not subject to the schema's `NameFieldConverter`, but rather always have camel case naming conventions, `GetArgument` cannot be used within field resolvers of introspection types.

### DIFF
--- a/src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
+++ b/src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
@@ -1,4 +1,5 @@
 using System.Threading.Tasks;
+using GraphQL.Conversion;
 using GraphQL.Introspection;
 using GraphQL.Types;
 using Shouldly;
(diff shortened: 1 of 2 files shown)


Shane32:
Note that the test failed before the change in `__Type.cs`

Shane32:
See #1545

sungam3r:
Ok, I understand this case. The converter should not affect the names of introspection fields and their arguments. But you corrected only the symptom and not the main problem. And the problem is that `TryGetArgument` passes `null` as `parentType`. And this happens because the field name converter is used to convert argument names, which is incorrect by itself. I see two places in code where `parentType` is `null`. I suggest investigate this issue in more detail. Perhaps we should add something like `IArgumentNameConverter`.

Shane32:
I’ll be posting a separate PR to clean up this code. But I’m running into a test failing when run at the same time as another test, but successfully when separate so it’s not ready yet.

Shane32:
I wasn’t sure if we should change the design of the IFieldNameConverter or fix the issues without changing the interface. Take a look at my next PR and we can review together.

Shane32:
> And this happens because the field name converter is used to convert argument names, which is incorrect by itself. I see two places in code where parentType is null.

Also, even if `parentType` was set correctly, the parent field type of the `includeDeprecated` argument is `__Type`, not `SchemaIntrospection`, so the conditions built into `PascalCaseFieldNameConverter` would not work, as they only look for the `SchemaIntrospection` type.  This can be demonstrated by changing the default `FieldNameConverter` in `GraphTypesLookup` to `PascalCaseFieldNameConverter`.  Now all the introspection fields get processed by that converter, which changes the case on all the fields instead of converting them to camel case due to them being introspection fields.

sungam3r:
> Also, even if parentType was set correctly, the parent field type of the includeDeprecated argument is __Type, not SchemaIntrospection, so the conditions built into PascalCaseFieldNameConverter would not work, as they only look for the SchemaIntrospection type. 

I know. I wanted to add necessary args to `NameFor` about the field itself, and not only about parent i.e. expand the method signature with additional context.

sungam3r:
Closed in favor of #1549 .

PR #1549, opened 2020-02-18
Changes to IFieldNameConverter and related code

Shane32:
This PR changes the `FieldNameConverter` to become independent from caring what the parent type is.  Now the `GraphTypesLookup` ensures the proper `FieldNameConverter` is applied for introspection types (`CamelCaseFieldNameConverter` is required).

To do:
- ~~Right now, various tests will fail after the new `validate_core_schema_abstractfieldnameconverter` runs. Must be some type of shared memory access, possibly in the validation code.  Every test executes successfully when run independently.~~ Fixed
- Add some comments

Questions:
- Should we change the signature of `IFieldNameConverter.NameFor` ?  Perhaps remove the `parentType` ?
- Should `HandleField` pass the parent type for arguments?
- Should we have `NameForField` and `NameForArgument` ?


Shane32:
Includes #1547

Shane32:
Since all of the introspection types are part of the internal graphql library, I think it makes sense to remove the "is this an introspection type?" question from the realm of the `IFieldNameConverter` altogether, and put the burden on the graphql library to call the correct field name converter for introspection fields.  This makes it much simpler to write a custom `IFieldNameConverter`.  For instance, an `UppercaseFieldNameConverter` could be written with just a couple lines of code.  It really seemed to be very easy to implement, and the only catch is to be sure not to use the `GetArgument()` extension function within introspection fields.  (Which wasn't working anyway as shown in #1547 .)

This PR is a representation of that change.

Shane32:
Also, I'm still looking for comments on the 3 questions posed above.  I think a decent answer is to simply remove the `parentType` parameter from `NameFor`, as it is not needed anymore.  This would result in 'reduced functionality' for implementers, so we could leave it as-is, another decent option.

Or, we could expand upon the functionality by changing `NameFor` to `NameForField` and adding `NameForArgument`, and passing the proper type in when calling `NameForArgument`.  This raises the question: should `NameForArgument` have a third parameter to indicate which field?  This surely would provide the most flexibility for implementers.

I guess the only thing I would not do is to leave the `NameFor` signature the same, but pass the proper field type in for arguments.  This would prevent you from being able to discern a field name from a field argument (at present, a simple `parentType == null` check can determine this).  Of course, you can't tell which graph type the argument applies to.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -329,7 +339,10 @@ private void HandleField(Type parentType, FieldType field, TypeCollectionContext
 
             foreach (var arg in field.Arguments)
             {
-                arg.Name = FieldNameConverter.NameFor(arg.Name, null);
+                if (parentType != typeof(SchemaIntrospection))
+                {
+                    arg.Name = FieldNameConverter.NameFor(arg.Name, null);
```
sungam3r: I want to once again focus your attention on this call (and a similar call in one more place). This method for fields is used for the argument. **An argument is not a field**. I believe that these 2 places need to be changed anyway.

sungam3r: We need to think about whether argument names should be converted at all. Possible not. Related to #1081

sungam3r: @pavel-usachev What guided you when you thought that argument names should also be converted? I see that you specifically added tests where you call `GetArgument` with different arg name. Why it was impossible for you to get the value of the argument by its correct name?

Shane32: In the referenced PR #1081 , the change makes it so that GetArgument converts argument names so they match the already-converted argument name (by `HandleField`) and the value can be returned.  I experienced this issue myself with an older release.

As an example, if you define an argument `MyArg`, then `context.GetArgument("MyArg")` should return the proper value, whether or not the argument name had been converted to camel case (i.e. `myArg`).  This would fail prior to #1081 .

Personally, I have no issue with argument names not being converted.  Then we would remove the applicable code from `HandleField` and revert #1081 .  In the prior example, `MyArg` would simply stay `MyArg`.

Be aware that changing this functionality would cause significant breaking changes in existing user codebases, if the user had used pascal case argument names in their code but are expecting camel case argument names exposed through graphql.

sungam3r: OK. I see now. You just adjusted to the argument names obtained when constructing the schema. So...
@Shane32 I see two three options:
1. Make new converter interface or method for converting argument names, with or without additional check for introspection args. I think that **with**.
2. Remove two argument conversion calls and not provide ability to change argument names anyway.
3. Your option with `if` check only, `NameFor` does not change.

Honestly, I would prefer 1 or 2 because I don't like that:
```c#
 // do not use context.GetArgument<> as it will translate the argument name via the selected schema's IFieldNameConverter
 var includeDeprecated = (bool)context.Arguments["includeDeprecated"];
```

Shane32: I like option 2 the best, but am concerned about it breaking existing user code. Thoughts ?

sungam3r: After discussion with @pavel-usachev  we came to the conclusion that following option 1 would be better - there will be no hidden breaking changes. And besides, there is nothing wrong with enabling the conversion of argument names, we just need the appropriate API for this.

sungam3r: > Since all of the introspection types are part of the internal graphql library, I think it makes sense to remove the "is this an introspection type?" question from the realm of the IFieldNameConverter altogether, and put the burden on the graphql library to call the correct field name converter for introspection fields.

Agree.

> Be aware that changing this functionality would cause significant breaking changes in existing user codebases, if the user had used pascal case argument names in their code but are expecting camel case argument names exposed through graphql.

Also agree. So option 1 with check.

> Or, we could expand upon the functionality by changing NameFor to NameForField and adding NameForArgument, and passing the proper type in when calling NameForArgument. This raises the question: should NameForArgument have a third parameter to indicate which field? This surely would provide the most flexibility for implementers.

I would like to see two methods, one for fields and one for args, within one interface (`IFieldNameConverter` -> `INameConverter`). It is a breaking change for custom converters (which no one uses I think) but explicit and simple to implement.

Shane32: I like that answer and will update the PR

Shane32:
@sungam3r Take a look when you have a chance.  Of note, the `NameForField` definition now has an `IComplexGraphType` argument instead of `Type` for the parent graph type.  This is to match the `NameForArgument` method, which uses the same for speed (eliminating a call to `GetType` for every call to `GetArgument`).  Anyone can check the type by calling `graphType.GetType()` of course, so no functionality is reduced.

Shane32:
We could also include another `INameConverter` implementation which converts field names to pascal case and argument names to camel case.  But how many predefined converters do we need?  Anyone can write their own with a few lines of code.

Shane32:
So presently the two three name converters should function identical to previously -- converting both field and argument names to the selected case.

sungam3r:
> We could also include another INameConverter implementation which converts field names to pascal case and argument names to camel case. 

Do not.

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
```diff
@@ -27,6 +29,53 @@ public async Task validate_core_schema(IDocumentWriter documentWriter)
...
+                _.NameConverter = new TestNameConverter();
+                _.Query = SchemaIntrospection.IntrospectionQuery;
+            });
+
+            var json = await documentWriter.WriteToStringAsync(executionResult);
+
+            ShouldBe(json, IntrospectionResult.Data);
+        }
+
+        public class TestNameConverter : INameConverter
```
sungam3r: Methods of this converter are never called?

src/GraphQL/Conversion/CamelCaseNameConverter.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using GraphQL.Types;
+
+namespace GraphQL.Conversion
+{
+    /// <summary>
+    /// Camel case name converter; set as the default <see cref="INameConverter"/> within <see cref="ExecutionOptions.NameConverter"/>.
+    /// Always used by all introspection fields regardless of the selected <see cref="INameConverter"/>.
```
sungam3r: Generally speaking, it would be more correct to say that introspection fields do not use any converter. And so it should be in code.

src/GraphQL/Conversion/CamelCaseNameConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    public class CamelCaseNameConverter : INameConverter
+    {
+        /// <summary>
+        /// Static instance of <see cref="CamelCaseNameConverter"/>
```
sungam3r:
```suggestion
        /// Static instance of <see cref="CamelCaseNameConverter"/> that can be reused instead of creating new.
```

src/GraphQL/Conversion/CamelCaseNameConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        public static readonly CamelCaseNameConverter Instance = new CamelCaseNameConverter();
+
+        /// <summary>
+        /// Returns the field name converted to camel case
```
sungam3r:
```suggestion
        /// Returns the field name converted to camelCase.
```

src/GraphQL/Conversion/CamelCaseNameConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        public string NameForField(string fieldName, IComplexGraphType graphType) => fieldName.ToCamelCase();
+
+        /// <summary>
+        /// Returns the argument name converted to camel case
```
sungam3r:
```suggestion
        /// Returns the argument name converted to camelCase.
```

src/GraphQL/Conversion/DefaultNameConverter.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using GraphQL.Types;
+
+namespace GraphQL.Conversion
+{
+    /// <summary>
+    /// A name converter which does not modify the names passed to it.
+    /// Not used for introspection fields.
```
sungam3r:
```suggestion
```

I don't think that this comment makes sense.

src/GraphQL/Conversion/DefaultNameConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    public class DefaultNameConverter : INameConverter
+    {
+        /// <summary>
+        /// Static instance of <see cref="DefaultNameConverter"/>
```
sungam3r:
```suggestion
        /// Static instance of <see cref="DefaultNameConverter"/> that can be reused instead of creating new.
```

src/GraphQL/Conversion/CamelCaseNameConverter.cs
Shane32: That is presently incorrect.  Due to some of the introspection fields being defined as `Field(x => x.Name)`, they do rely on a camel case converter for proper operation (verified).  I can change this if you want.

src/GraphQL/Conversion/INameConverter.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using GraphQL.Types;
+
+namespace GraphQL.Conversion
+{
+    /// <summary>
+    /// Sanitizes graph field and argument names to a particular case convention, such as camel case or pascal case.<br/>
```
sungam3r:
```suggestion
    /// Sanitizes graph field and argument names to a particular case convention, such as camelСase or PascalCase.<br/>
```

src/GraphQL/Conversion/INameConverter.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public interface INameConverter
+    {
+        /// <summary>
+        /// Santizes a field name for a specified parent graph type; returns the updated field name
+        /// </summary>
+        string NameForField(string fieldName, IComplexGraphType graphType);
+
+        /// <summary>
+        /// Santizes a argument name for a specified parent graph type and field definition; returns the updated field name
```
sungam3r:
```suggestion
        /// Santizes an argument name for a specified parent graph type and field definition; returns the updated field name
```

src/GraphQL/Conversion/PascalCaseNameConverter.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using GraphQL.Types;
+
+namespace GraphQL.Conversion
+{
+    /// <summary>
+    /// Pascal case name converter. Not used for introspection fields.
```
sungam3r:
```suggestion
    /// PascalCase name converter.
```

src/GraphQL/Conversion/CamelCaseNameConverter.cs
sungam3r: Also this comment can be written on the `INameConverter` interface.

src/GraphQL/Conversion/PascalCaseNameConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    public class PascalCaseNameConverter : INameConverter
+    {
+        /// <summary>
+        /// Static instance of <see cref="PascalCaseNameConverter"/>
```
sungam3r:
```suggestion
        /// Static instance of <see cref="PascalCaseNameConverter"/> that can be reused instead of creating new.
```

src/GraphQL/Conversion/PascalCaseNameConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// </summary>
+        public static readonly PascalCaseNameConverter Instance = new PascalCaseNameConverter();
+
+        /// <summary>
+        /// Returns the field name converted to pascal case
```
sungam3r:
```suggestion
        /// Returns the field name converted to PascalCase.
```

src/GraphQL/Conversion/PascalCaseNameConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// </summary>
+        public string NameForField(string fieldName, IComplexGraphType graphType) => fieldName.ToPascalCase();
+
+        /// <summary>
+        /// Returns the argument name converted to pascal case
```
sungam3r:
```suggestion
        /// Returns the argument name converted to PascalCase.
```

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
Shane32: Not explicitly.  The class is passed to `ExecutionOptions` and a `DocumentExecuter` executes the query - see test `validate_core_schema_abstractnameconverter` which is located immediately above this class definition.  The purpose of the test is to verify that given a very wacky name converter, all introspection fields still return with the proper case / spelling / etc.

src/GraphQL/ResolveFieldContextExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static object GetArgument(this IResolveFieldContext context, System.Type
 
         private static bool TryGetArgument(this IResolveFieldContext context, System.Type argumentType, string name, out object result)
         {
-            var argumentName = context.Schema?.FieldNameConverter.NameFor(name, null) ?? name;
+            var argumentName = context.Schema?.NameConverter.NameForArgument(name, context.ParentType, context.FieldDefinition) ?? name;
```
sungam3r:
```suggestion
        var argumentName = name;
        if (!introspection_related_argument_condition)
            argumentName = context.Schema?.NameConverter.NameForArgument(name, context.ParentType, context.FieldDefinition) ?? name;
```
Then there is no need in this warning
```
  // do not use context.GetArgument<> as it will translate the argument name via the selected schema's INameConverter
```
and you can revert introspection resolvers to use `GetArgument`. I believe that this will be more correct.

src/GraphQL/Introspection/__Type.cs
```diff
@@ -78,7 +79,8 @@ public __Type()
                 {
                     if (context.Source is EnumerationGraphType type)
                     {
-                        var includeDeprecated = context.GetArgument<bool>("includeDeprecated");
+                        // do not use context.GetArgument<> as it will translate the argument name via the selected schema's INameConverter
+                        var includeDeprecated = (bool)context.Arguments["includeDeprecated"];
```
sungam3r: This file can stay unchanged.

src/GraphQL/Introspection/TypeMetaFieldType.cs
```diff
@@ -15,6 +15,7 @@ public TypeMetaFieldType()
                 {
                     Name = "name"
                 });
+            // do not use context.GetArgument<> as it will translate the argument name via the selected schema's INameConverter
```
sungam3r:
```suggestion
```

Shane32: Done

src/GraphQL/Introspection/__Type.cs
Shane32: Done

src/GraphQL/ResolveFieldContextExtensions.cs
Shane32: I really don't like conditions inside `GetArgument` intended to determine if the field is an introspection type or not, as this would probably involve using `GetType()` to determine the parent field type, and then comparing the type against a predetermined list.

However, I've found another solution which both lets introspection fields use `GetArgument`, has no performance impact on other fields, does not call `GetType()`, and does not require a list of introspection types within `GetArgument`.  I simply added field extensions scoped to GraphQL.Introspection that will be called instead of the standard ones.  Now any introspection type can use the same `GetArgument` syntax just like any other field resolver, and so long as it is in the `GraphQL.Introspection` namespace, the name will not be passed through the `INameConverter`.  Please review.

Shane32:
I've also updated `HasArgument` to convert field names according to the selected `IFieldNameConverter`.

src/GraphQL/Conversion/CamelCaseNameConverter.cs
Shane32: Added; please review.

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: The PR has been updated with INameConverter and two methods NameForField and NameForArgument

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
sungam3r: But there are no fields other than introspection fields. Right? So these methods are never called. So I propose to throw exception from rhem to explicitly show that
> given a very wacky name converter, all introspection fields still return with the proper case / spelling / etc.

src/GraphQL/ResolveFieldContextExtensions.cs
sungam3r: Code became more complicated. Single condition inside GetArgument would be simpler but I will review.

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
Shane32: Good idea!  Done.

Shane32: I didn't understand your question at first.  Nope they are never called, and updating the test confirms it.

src/GraphQL/Conversion/CamelCaseNameConverter.cs
sungam3r: > That is presently incorrect. Due to some of the introspection fields being defined as Field(x => x.Name), they do rely on a camel case converter for proper operation 

You're right.

src/GraphQL/ResolveFieldContextExtensions.cs
Shane32: If you like this idea better, I could add a dummy type (e.g. `IIntrospectionGraphType`) and then `GetArgument` could check against the dummy type for the `introspection_related_argument_condition`.  Each introspection type would need to implement this type.

sungam3r: This option has a drawback. Anyone can implement this interface and "become" an introspection type. Therefore, it is better to check the type name for double underscores. See `GraphType` ctor and `NameValidator`. Also there this `SchemaPrinter.IsIntrospectionType`. You can consolidate all of this into extension method in `TypeExtensions`.

Shane32: The type need not be public.  It certainly is faster code.  And the current sample is faster yet.  But your suggestion with `StartsWith("__")` makes sense - it's just the slowest option.  Well, calling `GetType()` would be even slower.

Shane32: I updated it.  I still don't like repeated string comparisons for the introspection check for every call to GetArgument.

sungam3r: > The type need not be public

Are you sure?

> It certainly is faster code. And the current sample is faster yet. But your suggestion with StartsWith("__") makes sense - it's just the slowest option. Well, calling GetType() would be even slower.

It is not a place worth to optimizing. You do not notice any performance boost in real cases. We get only code complication. If you want to make sure you can provide benchmarks.

Shane32: True; at least it isn't hard-coded to a list of types.

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -76,6 +76,16 @@ public static bool IsInputObjectType(this IGraphType type)
...
 
+        /// <summary>
+        /// Determines if this graph type is an introspection type
+        /// </summary>
+        public static bool IsIntrospectionType(this IGraphType graphType) => graphType?.Name?.StartsWith("__") ?? false;
+
+        /// <summary>
+        /// Determines if this field is an introspection type
+        /// </summary>
+        public static bool IsIntrospectionField(this FieldType fieldType) => fieldType?.Name?.StartsWith("__") ?? false;
```
sungam3r: Well, is `types` field or any other child field in `__schema` introspection field or not?

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -239,7 +245,7 @@ public void AddType(IGraphType type, TypeCollectionContext context)
             {
                 foreach (var field in complexType.Fields)
                 {
-                    HandleField(type.GetType(), field, context);
+                    HandleField(complexType, field, context, false);
```
sungam3r: I noticed that control gets here for the nested fields of introspection, not just `__XXX`. `introspectionType` may confuse a bit. The converter should not affect not only `__XXX` fields, but also all nested fields and their arguments - the entire introspection graph. If I understand correctly, now the converter **works** for all the nested introspection fields, but since it is equal to `CamelCaseNameConverter.Instance`, it does not produce any effect. Right?

src/GraphQL/GraphQLExtensions.cs
sungam3r: It is possible that to solve the problem, the time has come to do what I have long wanted - to make a reference from `FieldType` to `IGraphType`, `Parent` property. I needed this property to solve other problems. If it were, then the question of whether the field belongs to introspection would become trivial. In addition, the conditions inside `TryGetArgument`/`HasArgument` will be simplified. I believe that now these conditions are not correct because they do not take into account nested introspection fields. The converter should not be applied to the entire introspection tree.

I believe that the task of adding `Parent` property to `FieldType` is more important to solve in separate PR first. Then we can return to this PR and do things simpler. What do you think?

Shane32: I'm not sure exactly how that is implemented in the code, but I did verify that both user field names and user argument names must not start with `__`.  So the check I've implemented, pursuant to your previous suggestion, should be a valid one.

Because I think this is a valid way to handle this issue, I'd pull it in, which will fix two existing bugs in the code base.  (a) introspection arguments not working with pascal case, and (b) HasArgument not converting names via the selected name converter.  It will also implement the new interface design we have come up with.

Once `FieldType.Parent` is added, we can update this code to use a more elegant solution, as it should not change the work already accomplished.

As it relates to your idea, this method is checking for an introspection query field "on the `Query` type", rather than a field as a child of an introspection type.  Would `FieldType.Parent` point to the Query graph, or a constant, or null?  Is this `Parent` field an `IGraphType` or a `Type`?  I'm already checking `context.ParentType` for `null` to determine if this might be an introspection query field (see `GetArgument`), so depending on the decisions to the above answers, it may or may not help this code at all.

In general, I like the idea of adding `Parent` to `FieldType`.  I just feel that there's some unanswered questions about how this will play out, and whatever the solution, it should not affect the changes to `IFieldNameConverter` that constitutes the majority of this PR - but rather a simple optimization (if any) of the `IsIntrospectionField` or `IsIntrospectionType` methods.  We could make those two new methods `internal` if we think the definitions may change soon -- there's no real reason they need to be public at this time.

Shane32: > Well, is types field or any other child field in __schema introspection field or not?

I'm sorry, I don't understand your question.  I hope I clarified what the two new methods do in my previous comment.  Perhaps I should update the xml comments to more clearly define what the methods do, as I admit it's not that clear.

sungam3r:
```graphql
__schema (1)
{
   types   (2)
   {
      name  (3)
      kind   (4)
   }
}
```

IsIntrospectionField(1) ?
IsIntrospectionField(2) ?
IsIntrospectionField(3) ?
IsIntrospectionField(4) ?

Following the logic of the `IsIntrospectionField` method, will the names of the fields be converted in each of the four cases? Should a conversion take place?

sungam3r: > I'm not sure exactly how that is implemented in the code, but I did verify that both user field names and user argument names must not start with __.

This is a **necessary** condition, but **not sufficient**.

Shane32: At present, and by design, only 1 returns true, as it is the only introspection field [on a non-introspection type].

2, 3, and 4 are covered by IsIntrospectionType, checking the parent type of the field.

GetArgument only executes either IsIntrospectionType or IsIntrospectionField, not both, depending on if parent is null.

Shane32: Conversions upon schema initialization is handled as before, where they are all converted by the camel case converter, no matter what converter was selected for the schema.  This is necessary as some of the fields are defined in pascal case.

sungam3r: Damn, I forgot about it again. 🤦‍♂

sungam3r: I again thought about the unnecessary side effects of using a camel case converter on the introspection fields. And again I forgot that these effects are actually useful. I will continue the review tomorrow.

Shane32: Ok no problem.

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: I think I understand what you're saying.  Yes the converter works and is operational for all nested introspection fields.  The property you referred to, `introspectionType` is like an override, specifically for the three root introspection query fields -- not fields of an introspection type.  The parameter is necessary because at the specific time that HandleField is called for these "query fields", the `NameFieldConverter` property has already been set to the schema's selected name converter.  When this boolean property is true, HandleField will ignore the `NameFieldConverter` altogether, leaving the field names alone.  This only applies to the three root introspection query fields -- `__schema`, `__type`, and `__typename`.  It does not apply to any of the introspection types or their fields.

I've renamed the field to `applyNameConverter`, inversed the logic, and added a boatload of comments to help clarify this for others in the future.

By the way, this design isn't mine; this is the way it was when I found it.  I just changed it from passing a `Type` and then comparing `type == typeof(SchemaIntrospection)` to a simple boolean, since the parameter served no other purpose.

docs2/site/docs/guides/migration.md
```diff
@@ -69,7 +69,7 @@ The `DateGraphType` has been split into multiple types.  [See the GitHub issue](
 
 Fields, enumerations, and arguments all now have their names validated according to the GraphQL spec, which is `/[_A-Za-z][_0-9A-Za-z]*/`.
 
-`QueryArgument` names are now run through the `IFieldNameConverter` set on the `Schema`.
+`QueryArgument` names are now run through the `INameConverter` set on the `Schema`.
```
joemcbride: This should not be changed since this is the upgrade guide from 0.17 to 2.x.  This should be put in an upgrade guide from 2.x to 3.x.

Shane32:
```suggestion
`QueryArgument` names are now run through the `IFieldNameConverter` set on the `Schema`.
```

Shane32: Done.  Sorry about that!

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -15,8 +15,16 @@ public class GraphTypesLookup
...
-        public GraphTypesLookup()
+        public GraphTypesLookup() : this(CamelCaseNameConverter.Instance) { }
+
+        public GraphTypesLookup(INameConverter nameConverter)
         {
+            if (nameConverter == null)
+                throw new ArgumentNullException(nameof(nameConverter));
+
+            // introspection fields use camelcase
+            NameConverter = CamelCaseNameConverter.Instance;
```
joemcbride: Assigning the `NameConverter` twice is a little confusing when first looking at it.  Is this needed now with the `applyNameConverter` check?  If so, can you add a comment here why it is being assigned twice and maybe add it closer to the introspection types - `__EnumValue`, `__InputValue`, etc.

Shane32: It always did that before, but it was hidden:  When the constructor executed, the NameConverter was set to the property default (CamelCaseNameConverter), and then the `Create` function would set the NameConverter after the fact.  I simply changed it so that it was obvious what was happening.

As it is now, the NameConverter must be CamelCaseNameConverter while adding the introspection types, as some places `Field(x => x.Name)` is run, so the field name is pascal case.

The `applyNameConverter` check is only set to `false` when adding the three introspection fields (as opposed to types) -- `__schema`, `__type`, and `__typename` -- on the base query type.  There's a whole boatload of comments in `HandleField` to explain this.

As I mentioned before, for better or worse, this is an old design.  I have no problem updating it, but I really didn't change the logic from before.  It was pretty hard to understand what was going on before, so hopefully this is at least an improvement.  Let me know if you think I should make further changes.  I did move the two NameConverter assignments immediately around the introspection types, as requested.  Let me know if you think that's clear.

Shane32: With the last change, `NameConverter` is null while adding the scalar types in the `GraphTypeLookup` constructor.  This does not matter as they do not have fields.  If there was a call to `HandleField` at that time, an `NullReferenceException` would occur.  This is probably for the best.

Shane32: Just FYI, the old code passed `typeof(SchemaIntrospection)` as the `parentType` parameter to `HandleField`, which would then check the `parentType` field against `typeof(SchemaIntrospection)` to determine if it should ignore the `NameConverter`.  I simply changed it to an appropriately named boolean, to make it more clear the purpose of the code.

Shane32: Well, I added the boolean, but that's what it did before.  I'm also using `IGraphType` instead of a graph `Type` so that we don't incur a `GetType()` lookup in `GetArgument` and `HasArgument`.

Shane32: I had the boolean named `introspectionField` but with comments from @sungam3r I ended up changing it to more precisely indicate its function.  So now it is called `applyNameConverter`.

Shane32:
@sungam3r @joemcbride I would like to point out that the three introspection field definitions are have static instances:
``` C#
        public static readonly SchemaMetaFieldType SchemaMeta = new SchemaMetaFieldType();
        public static readonly TypeMetaFieldType TypeMeta = new TypeMetaFieldType();
        public static readonly TypeNameMetaFieldType TypeNameMeta = new TypeNameMetaFieldType();
```
Field types are mutable.  Particularly, they contain references to other `IGraphType` elements within the schema.  I feel this is actually a flaw in the programming -- for example, imagine there were two different schemas in play.  Each would initialize the same instances of the above field types.  So one of the two schema's introspection field type's `ResolvedType` property would point to an instance 'owned' by the other schema.

I noticed this when I was testing some code to force one of these types to have the incorrect capitalization of the field name.  Even though only one test made this change, it broke other running tests.

Anyway, in the real world this may never matter, as even if it were to occur, the introspection types are probably the same anyway.  So I can solve this "issue" if you like.  Or I can leave it alone.

Shane32:
@sungam3r @joemcbride I appreciate both of your times spent reviewing this PR with me.  I had not intended to make such drastic changes, but simply to solve the bug in #1547 .  However I do feel it's a good change for the project.  We can always go back to implementing the 2-line fix in #1547 if not.  (This PR does fix another separate bug also, namely that HasArgument does not convert field names)

joemcbride:
> Particularly, they contain references to other IGraphType elements within the schema.

Can you point to where?  As far as I can tell they just use what is passed to them in the Resolver with the `ResolveFieldContext`, so seems like that shouldn't be an issue ...

Shane32:
`SchemaMetaFieldType` inherits from `FieldType`.  `FieldType.ResolvedType` is set in the first few lines of `HandleField` to an instance (not type) of a graph type.  For the `types` field, the `Type` property is set to `NonNullGraphType<ListGraphType<NonNullGraphType<__Type>>>`.  `HandleField` will resolve this to the schema's instance of `__Type`, which is created in the constructor of `GraphTypesLookup`.

docs2/site/docs/guides/migration.md
sungam3r: My apologies, I also did not notice.

Shane32:
A second schema created simultaneously would overwrite the static instance of `SchemaMetaFieldType` 's `ResolvedType` property to its own separate instance of `__Type`

Shane32:
If we want to change this, I can post a separate PR.  There's been enough changes in this PR!!

sungam3r:
@Shane32 You are right talking about the problem of static instances. I somehow did not pay attention to it. Of course, various unpleasant effects are possible due to work with a shared state. For example, meta-information inside a shared introspection field can be overwritten. I think it’s worth fixing and of course in a separate PR.

Also, I would like to spend some more time reviewing this PR before merge because I still have some questions.

sungam3r:
> Can you point to where?

@joemcbride for example, middleware that modifies each field of each type of schema, including introspection. By the way, I now had the thought that because of this there will be a memory leak if someone uses a sсoped schema with middleware, since middleware will modify static fields for each request by adding its wrapping resolver over and over again. @Shane32 could you check this scenario? I think you understand what I mean.

Shane32:
@sungam3r Good example, but no it doesn't.  The three introspection fields in question are not attached to any specific graph type, and so are not returned by `GraphTypesLookup.All`, and so are not modified by `ApplyTo`.

Shane32:
Tested by adding `if (field == Introspection.SchemaIntrospection.SchemaMeta) throw new Exception();` into `ApplyTo`

sungam3r:
Well, OK. But so far I’m not sure that the dangerous scenarios are limited to this. Probably in the documentation on middleware a separate section is needed that they do not affect introspection fields.

Shane32:
Middleware certainly does affect all other introspection type's fields.  Just not those three fields, since they are not attached to a type.  So given a request of `{ __schema { types { .. } } }`, middleware will affect `types` and child fields, but not `__schema`.

I'll look into making those fields non-static (just like every other scalar, graph and introspection type) in a separate PR.

Shane32:
Documentation changes for middleware should probably go in your PR #1537 , as you are changing a bunch of middleware documentation.  Or I can post a separate PR.

Shane32:
We could also program `ApplyTo` to modify the three introspection fields.  Not yet, of course, or it will create the memory leak you described above.

sungam3r:
I will document it in #1537 . I have ahead a couple of free days to work well.

Shane32:
@sungam3r Did you say you wanted to review this some more?

sungam3r:
Yes. I have one more day left to do all this.

Shane32:
@sungam3r Are we ready to pull this in?

sungam3r:
Not yet, I have 2 free days ahead to deal with this and other PRs.

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: OK.

src/GraphQL/Conversion/INameConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Santizes a field name for a specified parent graph type; returns the updated field name
+        /// </summary>
+        string NameForField(string fieldName, IComplexGraphType graphType);
+
+        /// <summary>
+        /// Santizes an argument name for a specified parent graph type and field definition; returns the updated field name
+        /// </summary>
+        string NameForArgument(string argumentName, IComplexGraphType graphType, FieldType field);
```
sungam3r: `IComplexGraphType graphType` -> `IComplexGraphType parentGraphType` ? Comment says _parent graph type_. The same suggestion for `NameForField` method.

src/GraphQL/GraphQLExtensions.cs
sungam3r: So, I review your changes again and I think that both `IsIntrospectionType` and `IsIntrospectionField` methods should be `internal`. In fact, as discussed above, they work only for some introspection fields and their names can be misleading. It will be even better to move these methods inside `ResolveFieldContextExtensions` and make them `private`.

src/GraphQL/Conversion/INameConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <br/>
+    /// Set <see cref="ExecutionOptions.NameConverter"/> to an instance of a derived class to select a converter to use.
+    /// The default converter is <see cref="CamelCaseNameConverter"/>.<br/>
+    /// <br/>
+    /// Introspection fields always use <see cref="CamelCaseNameConverter"/> regardless of the selected <see cref="INameConverter"/>.
+    /// </summary>
+    public interface INameConverter
+    {
+        /// <summary>
+        /// Santizes a field name for a specified parent graph type; returns the updated field name
```
sungam3r:
```suggestion
        /// Sanitizes a field name for a specified parent graph type; returns the updated field name
```

src/GraphQL/Conversion/INameConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// </summary>
+        string NameForField(string fieldName, IComplexGraphType graphType);
+
+        /// <summary>
+        /// Santizes an argument name for a specified parent graph type and field definition; returns the updated field name
```
sungam3r:
```suggestion
        /// Sanitizes an argument name for a specified parent graph type and field definition; returns the updated field name
```

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -42,14 +47,19 @@ public GraphTypesLookup()
             AddType<ByteGraphType>();
             AddType<SByteGraphType>();
 
-            // introspection types
+            // Add introspection types.  Note that introspection types rely on the
+            //   CamelCaseNameConverter, as some fields are defined in pascal case - e.g. Field(x => x.Name)
```
sungam3r: 👍

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -62,13 +72,10 @@ private void CheckSealed()
...
             Func<Type, IGraphType> resolveType,
-            IFieldNameConverter fieldNameConverter,
+            INameConverter nameConverter,
             bool seal = false)
         {
-            var lookup = new GraphTypesLookup
-            {
-                FieldNameConverter = fieldNameConverter ?? CamelCaseFieldNameConverter.Instance
-            };
+            var lookup = nameConverter != null ? new GraphTypesLookup(nameConverter) : new GraphTypesLookup();
```
sungam3r: Please invert this condition.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -25,9 +25,10 @@ public static object GetArgument(this IResolveFieldContext context, System.Type
                 : defaultValue;
         }
 
-        private static bool TryGetArgument(this IResolveFieldContext context, System.Type argumentType, string name, out object result)
+        internal static bool TryGetArgument(this IResolveFieldContext context, System.Type argumentType, string name, out object result)
```
sungam3r:
```suggestion
        private static bool TryGetArgument(this IResolveFieldContext context, System.Type argumentType, string name, out object result)
```

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -55,7 +56,12 @@ private static bool TryGetArgument(this IResolveFieldContext context, System.Typ
         }
 
         /// <summary>Determines if the specified field argument has been provided in the GraphQL query request</summary>
-        public static bool HasArgument(this IResolveFieldContext context, string argumentName) => context.Arguments?.ContainsKey(argumentName) ?? false;
+        public static bool HasArgument(this IResolveFieldContext context, string argumentName)
+        {
+            var isIntrospection = context.ParentType == null ? context.FieldDefinition.IsIntrospectionField() : context.ParentType.IsIntrospectionType();
+            var argumentNameConverted = isIntrospection ? argumentName : (context.Schema?.NameConverter.NameForArgument(argumentName, context.ParentType, context.FieldDefinition) ?? argumentName);
+            return context.Arguments?.ContainsKey(argumentNameConverted) ?? false;
+        }
```
sungam3r:
```suggestion
        public static bool HasArgument(this IResolveFieldContext context, string name)
        {
            var isIntrospection = context.ParentType == null ? context.FieldDefinition.IsIntrospectionField() : context.ParentType.IsIntrospectionType();
            var argumentName = isIntrospection ? name : (context.Schema?.NameConverter.NameForArgument(name, context.ParentType, context.FieldDefinition) ?? name);
            return context.Arguments?.ContainsKey(argumentName) ?? false;
        }
```

To be consistent with other APIs in this file..

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: Changed to the following:
```suggestion
            var lookup = nameConverter == null ? new GraphTypesLookup() : new GraphTypesLookup(nameConverter);
```
If that's not what you meant, lmk.

src/GraphQL/Conversion/INameConverter.cs
Shane32: Sounds good.  Done.  Changed all implementations also.

Shane32:
@sungam3r Implemented suggestions; updated apiapprovals; merged; passes tests.

Please be sure this is what you wanted:
https://github.com/graphql-dotnet/graphql-dotnet/pull/1549/commits/ec441baa38fcc4b5bdb992f5d5e7cacce4a4fe0c

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: This is it.

sungam3r:
One suggestion left https://github.com/graphql-dotnet/graphql-dotnet/pull/1549/files#r386109911

src/GraphQL/GraphQLExtensions.cs
Shane32: > So, I review your changes again and I think that both `IsIntrospectionType` and `IsIntrospectionField` methods should be `internal`. In fact, as discussed above, they work only for some introspection fields and their names can be misleading. It will be even better to move these methods inside `ResolveFieldContextExtensions` and make them `private`.

Done; with updated xml comment

Shane32:
Sorry I missed that.  Done now.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -55,7 +57,23 @@ private static bool TryGetArgument(this IResolveFieldContext context, System.Typ
...
         /// <summary>Determines if the specified field argument has been provided in the GraphQL query request</summary>
-        public static bool HasArgument(this IResolveFieldContext context, string argumentName) => context.Arguments?.ContainsKey(argumentName) ?? false;
+        public static bool HasArgument(this IResolveFieldContext context, string name)
+        {
+            var isIntrospection = context.ParentType == null ? context.FieldDefinition.IsIntrospectionField() : context.ParentType.IsIntrospectionType();
+            var argumentName = isIntrospection ? name : (context.Schema?.NameConverter.NameForArgument(name, context.ParentType, context.FieldDefinition) ?? name);
+            return context.Arguments?.ContainsKey(argumentName) ?? false;
+        }
+
+
```
sungam3r:
```suggestion
```

PR #1550, opened 2020-02-19
Move IDisposable from ISchema to Schema

Shane32:
Shouldn't `IDisposable` be implemented on `Schema`, rather than `ISchema`?  It's not like `ISchema` is expected to hold any unmanaged references, so implementing `IDisposable` at all seems questionable.

Ok to merge or close.


sungam3r:
> t's not like ISchema is expected to hold any unmanaged references, so implementing IDisposable at all seems questionable

To implement `IDisposable` it is not necessary to have any unmanaged references.

sungam3r:
In general, I see this change as correct.

PR #1552, opened 2020-02-20
Add xml comments to ISchema

Shane32:
@sungam3r Please review. In particular, the comments for `RegisterValueConverter` and `FindValueConverter`, as I do not know how that would be used.  Perhaps a short example would be fitting.  Would this be for a custom date/time formatter?  But then wouldn't you simply implement a custom `ScalarFieldType` for bi-directional conversion?

Note that #1537 and #1549 will both slightly affect these comments, but this PR is based off the current `master` branch.  If we pull this PR first, I'll update #1549 - or vice versa.

The comment from `Schema.Filter` was removed so that it inherits the comments from `ISchema.Filter`.  Also, the `Schema` class was set to clone the comment from the `ISchema` interface.

### DIFF
--- a/src/GraphQL/Types/ISchema.cs
+++ b/src/GraphQL/Types/ISchema.cs
@@ -112,7 +112,7 @@ namespace GraphQL.Types
         /// Add a specific directive to the schema.<br/>
         /// <br/>
         /// Directives are used by the GraphQL runtime as a way of modifying execution
-        /// behavior. Type system creators will usually not create these directly.
+        /// behavior. Type system creators do not usually create them directly.
         /// </summary>
         void RegisterDirective(DirectiveGraphType directive);
 
--- a/src/GraphQL/Types/ISchema.cs
+++ b/src/GraphQL/Types/ISchema.cs
@@ -72,7 +72,7 @@ namespace GraphQL.Types
         DirectiveGraphType FindDirective(string name);
 
         /// <summary>
-        /// A list of additional graph types manually added to the schema
+        /// A list of additional graph types manually added to the schema by RegisterType call.
         /// </summary>
         IEnumerable<Type> AdditionalTypes { get; }
 
--- a/src/GraphQL/Types/ISchema.cs
+++ b/src/GraphQL/Types/ISchema.cs
@@ -137,6 +137,8 @@ namespace GraphQL.Types
         /// <summary>
         /// Provides the ability to filter the schema upon introspection to hide types. This is set by <see cref="IDocumentExecuter"/>
         /// to the filter passed to it within <see cref="ExecutionOptions.SchemaFilter"/>. By default, no types are hidden.
+        /// Note that this filter in fact does not prohibit the execution of queries that contain hidden types. To limit
+        /// access to the particular fields, you should use some authorization logic.
         /// </summary>
         ISchemaFilter Filter { get; set; }
     }
--- a/src/GraphQL/Types/ISchema.cs
+++ b/src/GraphQL/Types/ISchema.cs
@@ -5,48 +5,138 @@ using System.Collections.Generic;
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The schema for the GraphQL request. Contains references to the 'query', 'mutation', and 'subscription' base graph types.<br/>
+    /// <br/>
+    /// Also allows for adding custom directives, additional graph types, and custom value converters.<br/>
+    /// <br/>
+    /// <see cref="Schema"/> only requires the <see cref="Schema.Query">Query</see> property to be set; although commonly the <see cref="Schema.Mutation">Mutation</see> and/or <see cref="Schema.Subscription">Subscription</see> properties are also set.
+    /// </summary>
     public interface ISchema
     {
+        /// <summary>
+        /// Returns true once the schema has been initialized
+        /// </summary>
         bool Initialized { get; }
 
+        /// <summary>
+        /// Initializes the schema. Called by <see cref="IDocumentExecuter"/> before validating or executing the request.<br/>
+        /// <br/>
+        /// Note that middleware cannot be applied once the schema has been initialized. See <see cref="ExecutionOptions.FieldMiddleware"/>.
+        /// </summary>
         void Initialize();
 
+        /// <summary>
+        /// The <see cref="IFieldNameConverter"/> used by the schema. This is set by <see cref="IDocumentExecuter"/> to the converter passed to it within <see cref="ExecutionOptions.FieldNameConverter"/>.
+        /// </summary>
         IFieldNameConverter FieldNameConverter { get; set; }
 
+        /// <summary>
+        /// The 'query' base graph type; required
+        /// </summary>
         IObjectGraphType Query { get; set; }
 
+        /// <summary>
+        /// The 'mutation' base graph type; optional
+        /// </summary>
         IObjectGraphType Mutation { get; set; }
 
+        /// <summary>
+        /// The 'subscription' base graph type; optional
+        /// </summary>
         IObjectGraphType Subscription { get; set; }
 
+        /// <summary>
+        /// Returns a list of directives supported by the schema.<br/>
+        /// <br/>
+        /// Directives are used by the GraphQL runtime as a way of modifying execution
+        /// behavior. Type system creators will usually not create these directly.<br/>
+        /// <br/>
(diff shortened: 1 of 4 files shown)


src/GraphQL/Types/Schema.cs
```diff
@@ -66,11 +75,11 @@ public void Initialize()
 
         public IObjectGraphType Subscription { get; set; }
 
-        public IServiceProvider Services { get; set; }
-
         /// <summary>
-        /// Provides the ability to filter the schema upon introspection to hide types.
+        /// The service provider used to create instances of graph types during initialization of the schema
```
sungam3r: Not only graph types. See all usages - directive visitors, middlewares, etc.

Shane32:
I'll appreciate all the comments you have on this @sungam3r - thanks!

src/GraphQL/Types/Schema.cs
Shane32: Updated to just say 'create objects during initialization' and similar.

sungam3r: Well, not only during initialization. It is important.

src/GraphQL/Types/Schema.cs
```diff
@@ -66,11 +75,11 @@ public void Initialize()
 
         public IObjectGraphType Subscription { get; set; }
 
-        public IServiceProvider Services { get; set; }
-
         /// <summary>
-        /// Provides the ability to filter the schema upon introspection to hide types.
+        /// The service provider used to create objects during initialization of the schema
```
Shane32:
```suggestion
        /// The service provider used to create objects, such as graph types, requested by the schema.
        /// Note that most objects are created once, during schema initialization.
```
How about that? I'm just trying to help point out that graph types are (mostly) not created on an as-needed basis, like would be done for Asp Core.Net, but rather are typically created once, during schema initialization.

Shane32: I mentioned also graph types as that is what most users will relate to easiest, rather than directives, middleware, and other "back-end stuff" they may not know or care about.

sungam3r: > but rather are typically created once, during schema initialization

If the schema (and types) was configured as scoped, then they will be created for each HTTP request. So wording "once" may confuse.

> most users will relate to easiest, rather than directives, middleware, and other "back-end stuff" they may not know or care about.

Users certainly may not work with these concepts, but nevertheless I am sure that these concepts should be mentioned in the comments to give the reader a good picture.

Shane32: I updated the suggestion above; please review.  If you'd like to suggest a change, feel free to do so.

Shane32:
I also added a comment to `SchemaExtensions.ExecuteAsync` as this extension method will appear as a member of `ISchema`

src/GraphQL/Types/Schema.cs
```diff
@@ -66,11 +75,15 @@ public void Initialize()
...
 
-        public IServiceProvider Services { get; set; }
-
         /// <summary>
-        /// Provides the ability to filter the schema upon introspection to hide types.
+        /// The service provider used to create objects, such as graph types, requested by the schema.<br/>
+        /// <br/>
+        /// Note that most objects are created during schema initialization, which then have the same lifetime as the schema's lifetime.<br/>
+        /// <br/>
+        /// Other types created by the service provider include directives, middleware, validation rules, and name converters, among others.
```
sungam3r:
```suggestion
        /// Other types created by the service provider may include directives, middleware, validation rules, and name converters, among others.
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -5,48 +5,138 @@
...
+        /// <summary>
+        /// The 'subscription' base graph type; optional
+        /// </summary>
         IObjectGraphType Subscription { get; set; }
 
+        /// <summary>
+        /// Returns a list of directives supported by the schema.<br/>
+        /// <br/>
+        /// Directives are used by the GraphQL runtime as a way of modifying execution
+        /// behavior. Type system creators will usually not create these directly.<br/>
```
sungam3r:
```suggestion
        /// behavior. Type system creators do not usually create them directly.<br/>
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -5,48 +5,138 @@
...
+        /// <br/>
+        /// Directives are used by the GraphQL runtime as a way of modifying execution
+        /// behavior. Type system creators do not usually create them directly.<br/>
+        /// <br/>
+        /// <see cref="Schema"/> intializes the list to include <see cref="DirectiveGraphType.Include"/>, <see cref="DirectiveGraphType.Skip"/> and <see cref="DirectiveGraphType.Deprecated"/> by default.
+        /// </summary>
         IEnumerable<DirectiveGraphType> Directives { get; set; }
 
+        /// <summary>
+        /// Returns a list of all the graph types utilized by this schema
```
sungam3r: In fact, the schema contains many unnecessary scalars. I have PL to remove them. Let the comment remain so.

src/GraphQL/Types/ISchema.cs
```diff
@@ -5,48 +5,138 @@
...
+        /// </summary>
         IGraphType FindType(string name);
 
+        /// <summary>
+        /// Returns a <see cref="DirectiveGraphType"/> for a given name
+        /// </summary>
         DirectiveGraphType FindDirective(string name);
 
+        /// <summary>
+        /// A list of additional graph types manually added to the schema
```
sungam3r:
```suggestion
        /// A list of additional graph types manually added to the schema by RegisterType call.
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -5,48 +5,138 @@
...
+        /// Directives are used by the GraphQL runtime as a way of modifying execution
+        /// behavior. Type system creators will usually not create these directly.
+        /// </summary>
         void RegisterDirective(DirectiveGraphType directive);
 
+        /// <summary>
+        /// Add specific directives to the schema.<br/>
+        /// <br/>
+        /// Directives are used by the GraphQL runtime as a way of modifying execution
+        /// behavior. Type system creators will usually not create these directly.
```
sungam3r:
```suggestion
        /// behavior. Type system creators do not usually create them directly.
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -5,48 +5,138 @@
...
+        /// Not typically required as schema initialization will scan the <see cref="Query"/>, <see cref="Mutation"/> and <see cref="Subscription"/> graphs,
+        /// creating instances of <see cref="IGraphType"/>s referenced therein as necessary.
+        /// </summary>
         void RegisterType<T>() where T : IGraphType;
 
+        /// <summary>
+        /// Add a specific directive to the schema.<br/>
+        /// <br/>
+        /// Directives are used by the GraphQL runtime as a way of modifying execution
+        /// behavior. Type system creators will usually not create these directly.
```
sungam3r:
```suggestion
        /// behavior. Type system creators do not usually create them directly.
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -5,48 +5,138 @@
...
 
+        /// <summary>
+        /// Search the schema for a <see cref="IAstFromValueConverter"/> that matches the provided object and graph type, and return the converter.
+        /// </summary>
         IAstFromValueConverter FindValueConverter(object value, IGraphType type);
 
         /// <summary>
-        /// Provides the ability to filter the schema upon introspection to hide types.
+        /// Provides the ability to filter the schema upon introspection to hide types. This is set by <see cref="IDocumentExecuter"/>
+        /// to the filter passed to it within <see cref="ExecutionOptions.SchemaFilter"/>. By default, no types are hidden.
```
sungam3r:
```suggestion
        /// to the filter passed to it within <see cref="ExecutionOptions.SchemaFilter"/>. By default, no types are hidden.
        /// Note that this filter in fact does not prohibit the execution of queries that contain hidden types. To limit access
        /// to particular fields, you should use some authorization logic.
```

sungam3r:
> In particular, the comments for RegisterValueConverter and FindValueConverter, as I do not know how that would be used. 

I did not use this API. I can roughly imagine why it is needed. For now, leave the comments as they are.

PR #1553, opened 2020-02-20
Add basic xml comments to IDocumentExecutionListener

Shane32:
Added very basic comments for this interface. See changed files.

I might suggest that `IDocumentExecutionListener<T>` inherit from `IDocumentExecutionListener`.  Since the infrastructure only supports the latter interface, it is basically required for any implementation.  If you agree I can make that change also, or do so in a separate PR.  For now I just let it alone.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecutionListener.cs
+++ b/src/GraphQL/Execution/DocumentExecutionListener.cs
@@ -12,7 +12,6 @@ namespace GraphQL.Execution
     {
         /// <summary>Executes after document validation is complete. Can be used to log validation failures.</summary>
         Task AfterValidationAsync(object userContext, IValidationResult validationResult, CancellationToken token);
-
         /// <summary>Executes after document validation passes, before calling <see cref="IExecutionStrategy.ExecuteAsync(ExecutionContext)"/>.</summary>
         Task BeforeExecutionAsync(object userContext, CancellationToken token);
 
--- a/src/GraphQL/Execution/DocumentExecutionListener.cs
+++ b/src/GraphQL/Execution/DocumentExecutionListener.cs
@@ -22,7 +22,7 @@ namespace GraphQL.Execution
         /// <summary>Executes after the <see cref="IExecutionStrategy"/> has completed executing the request</summary>
         Task AfterExecutionAsync(object userContext, CancellationToken token);
 
-        /// <summary>Executes before each time the <see cref="IExecutionStrategy"/> awaits the <see cref="Task{TResult}"/> returned by <see cref="IFieldResolver.Resolve"/></summary>
+        /// <summary>Executes before each time the <see cref="IExecutionStrategy"/> awaits the <see cref="Task{TResult}"/> returned by <see cref="IFieldResolver.Resolve"/>. For parallel resolvers, this may execute a single time prior to awaiting multiple tasks.</summary>
         Task BeforeExecutionStepAwaitedAsync(object userContext, CancellationToken token);
     }
 
--- a/src/GraphQL/Execution/DocumentExecutionListener.cs
+++ b/src/GraphQL/Execution/DocumentExecutionListener.cs
@@ -1,77 +1,63 @@
 using System.Threading;
 using System.Threading.Tasks;
 using GraphQL.Validation;
+using GraphQL.Resolvers;
 
 namespace GraphQL.Execution
 {
+    /// <summary>
+    /// Provides the ability to log query validation failures and monitor progress of a GraphQL request's execution.
+    /// </summary>
     public interface IDocumentExecutionListener
     {
+        /// <summary>Executes after document validation is complete. Can be used to log validation failures.</summary>
         Task AfterValidationAsync(object userContext, IValidationResult validationResult, CancellationToken token);
+        /// <summary>Executes after document validation passes, before calling <see cref="IExecutionStrategy.ExecuteAsync(ExecutionContext)"/>.</summary>
         Task BeforeExecutionAsync(object userContext, CancellationToken token);
+        /// <summary>Executes before the <see cref="IDocumentExecuter"/> awaits the <see cref="Task"/> returned by <see cref="IExecutionStrategy.ExecuteAsync(ExecutionContext)"/></summary>
         Task BeforeExecutionAwaitedAsync(object userContext, CancellationToken token);
+        /// <summary>Executes after the <see cref="IExecutionStrategy"/> has completed executing the request</summary>
         Task AfterExecutionAsync(object userContext, CancellationToken token);
+        /// <summary>Executes before each time the <see cref="IExecutionStrategy"/> awaits the <see cref="Task{TResult}"/> returned by <see cref="IFieldResolver.Resolve"/></summary>
         Task BeforeExecutionStepAwaitedAsync(object userContext, CancellationToken token);
     }
 
+    /// <inheritdoc cref="IDocumentExecutionListener"/>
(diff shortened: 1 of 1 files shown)


src/GraphQL/Execution/DocumentExecutionListener.cs
```diff
@@ -1,77 +1,63 @@
...
+using GraphQL.Resolvers;
 
 namespace GraphQL.Execution
 {
+    /// <summary>
+    /// Provides the ability to log query validation failures and monitor progress of a GraphQL request's execution.
+    /// </summary>
     public interface IDocumentExecutionListener
     {
+        /// <summary>Executes after document validation is complete. Can be used to log validation failures.</summary>
```
sungam3r: Please separate all methods with empty lines.

Shane32: Done

Shane32:
Haha appveyor failed again.  We really need to pull in #1554 and see if it fixes the issue

sungam3r:
@Shane32  I think that we can remove `IDocumentExecutionListener<T>` and `DocumentExecutionListenerBase<T>` without negative consequences. This code was written before `userContext` became `IDictionary<string,object>`. It was just `object` before that. So now it doesn't make much sense to keep generic version. @joemcbride do you agree?

src/GraphQL/Execution/DocumentExecutionListener.cs
```diff
@@ -1,77 +1,71 @@
...
+        /// <summary>Executes after document validation passes, before calling <see cref="IExecutionStrategy.ExecuteAsync(ExecutionContext)"/>.</summary>
         Task BeforeExecutionAsync(object userContext, CancellationToken token);
+
+        /// <summary>Executes before the <see cref="IDocumentExecuter"/> awaits the <see cref="Task"/> returned by <see cref="IExecutionStrategy.ExecuteAsync(ExecutionContext)"/></summary>
         Task BeforeExecutionAwaitedAsync(object userContext, CancellationToken token);
+
+        /// <summary>Executes after the <see cref="IExecutionStrategy"/> has completed executing the request</summary>
         Task AfterExecutionAsync(object userContext, CancellationToken token);
+
+        /// <summary>Executes before each time the <see cref="IExecutionStrategy"/> awaits the <see cref="Task{TResult}"/> returned by <see cref="IFieldResolver.Resolve"/></summary>
```
sungam3r: Here we need to think about a more appropriate wording. `ParallelExecutionStrategy` can await multiple results. Besides `SubscriptionExecutionStrategy` does not call this method.

Shane32: I'll think about.  Note that SubscriptionExecutionStrategy inherits from ParallelExecutionStrategy, and the former calls the latter's ExecuteNodeTreeAsync method, which contains the call in question.  So it does get executed.

Shane32:
```suggestion
        /// <summary>Executes before each time the <see cref="IExecutionStrategy"/> awaits the <see cref="Task{TResult}"/> returned by <see cref="IFieldResolver.Resolve"/>. For parallel resolvers, this may execute a single time prior to awaiting multiple tasks.</summary>
```
How about that?

Shane32:
> I think that we can remove `IDocumentExecutionListener<T>`

We may be able to.  However, I would venture to guess that most people will just change their custom UserContext class to inherit from `Dictionary<string, object>` so they don't need to change any of their existing code.  Even if I wrote new code, it would always use a typed `UserContext` rather than a dictionary.  So, while we may be able to remove that class, I wouldn't on the basis that `UserContext` is now a dictionary, but rather that the interface should never have existed in the first place.  (It's not like there are any typed resolver classes for the usercontext field.)

So, I would leave it there, albeit mostly for backwards compatibility purposes.

sungam3r:
Well, look how much code has been written just to make `(T)` cast and delegate calls to other typed virtual methods. I honestly think this is redundant (even keeping in mind the consideration of backward compatibility).

sungam3r:
By the way, I forgot to explicitly say that I propose changing the signature of all `userContext` params from `object` to `IDictionary<string, object>`.

> but rather that the interface should never have existed in the first place

So now this interface is `IDictionary<string, object>`  :)

Shane32:
lol - well, I can't stand untyped code.  So I try to never use `ViewBag` / `ViewData`, always use typescript over javascript (and complain if anyone uses `as any` in their code), and so of course here I wouldn't store props in the dictionary if I can help it.

We could add a blub to the docs about how to do this btw -- an extension method for `IResolveFieldContext` can be added for `.MyContext` which casts `UserContext` to their own type.  I never thought of this before, but I'm going to add it to my code right now.  Better than `((MyContext)context.UserContext).MyProp`

sungam3r:
In the case of UserContext, we have an arbitrary set of some properties, so `IDictionary<string, object>` is suitable for this role. All the same, we will not be able to make UserContext typed. This will require the introduction of generics. It will be difficult to maintain. Moreover, typed UserContext is not always needed. For example, I don’t use it.

Shane32:
> Well, look how much code has been written just to make `(T)` cast and delegate calls to other typed virtual methods. I honestly think this is redundant (even keeping in mind the consideration of backward compatibility).

50/50 to me

> In the case of UserContext, we have an arbitrary set of some properties, so `IDictionary<string, object>` is suitable for this role. All the same, we will not be able to make UserContext typed. This will require the introduction of generics. It will be difficult to maintain. Moreover, typed UserContext is not always needed. For example, I don’t use it.

Oh, of course; I agree.  I already tried to see if it could be added, but it seemed that it would be very difficult to do so.

sungam3r:
So I propose not to solve problems that do not need to be solved (and remove IDocumentExecutionListener<T> and ListenerBase).

Shane32:
Personally, this whole interface seems pretty dumb.  I mean, it can record times, log validation failures, and assist a totally-broken dataloader.  I'm not sure what else it can be used for.  It has no access to anything during the execution.  It can't see the fields being resolved.  It can't change the execution result.  It can't modify execution exceptions.  It's like a bad version of middleware.

Shane32:
In my opinion, the middleware and document listener functionality should be merged, so field types's resolver properties are not overwritten, and the 'listener' functionality is more versatile.  However, I certainly have not thought this through entirely.  Just a concept.

sungam3r:
The idea is understandable, but still they have different goals and points of application.

sungam3r:
FieldMiddleware works for **fields**, so no fields - no deal. On the other hand `IDocumentExecutionListener` is just a way to callback to user code at any stage of execution graphql request.

sungam3r:
Also some listener's methods could take one `ExecutionContext` parameter instead two `userContext` and `cancellationToken`.

sungam3r:
Probably it was originally thought that the same method signature is good.

Shane32:
> Also some listener's methods could take one `ExecutionContext` parameter instead two `userContext` and `cancellationToken`.

Yes, see then the listener would actually have some data to work with.  If, for example, it was a listener for metrics, it would then be able to examine the field names being resolved.  (Although I'd probably pass `IEnumerable<IResolveFieldContext>` to the field, so it is immutable and can handle multiple fields.)

Shane32:
All that being said, this PR was just for documentation, so I think we should pull it if the docs are good and start another issue or PR if we want to change functionality.  I don't care if we delete the docs I just added on the next PR - lol.

PR #1554, opened 2020-02-20
Increase appveyor testing reliability

Shane32:
Occassionally appveyor fails testing due to SimpleDataLoaderTests.Operation_Can_Be_Cancelled

See: https://ci.appveyor.com/project/graphql-dotnet-ci/graphql-dotnet/builds/30927760

This is due to the wrong order of execution of the threads of testing, probably due to server load.  This increases the time limit for a successful test from 200ms to 5 sec.  It also configures `Task.Delay` to immediately return if the task is canceled (which it should be canceled after 5ms), so it actually runs faster in most instances (since it would wait a fixed 200ms previously).

### DIFF
--- a/src/GraphQL.DataLoader.Tests/SimpleDataLoaderTests.cs
+++ b/src/GraphQL.DataLoader.Tests/SimpleDataLoaderTests.cs
@@ -59,7 +59,7 @@ namespace GraphQL.DataLoader.Tests
             mock.Setup(store => store.GetAllUsersAsync(cts.Token))
                 .Returns(async (CancellationToken ct) =>
                 {
-                    await Task.Delay(200);
+                    await Task.Delay(5000, ct);
                     ct.ThrowIfCancellationRequested();
 
                     return users;
--- a/src/GraphQL.DataLoader.Tests/SimpleDataLoaderTests.cs
+++ b/src/GraphQL.DataLoader.Tests/SimpleDataLoaderTests.cs
@@ -59,7 +59,7 @@ namespace GraphQL.DataLoader.Tests
             mock.Setup(store => store.GetAllUsersAsync(cts.Token))
                 .Returns(async (CancellationToken ct) =>
                 {
-                    await Task.Delay(5000, ct);
+                    await Task.Delay(60000, ct);
                     ct.ThrowIfCancellationRequested();
 
                     return users;


Shane32:
Another example of a failed appveyor build due to the same test:
https://ci.appveyor.com/project/graphql-dotnet-ci/graphql-dotnet/builds/30840826

Shane32:
Yet another example: 
https://ci.appveyor.com/project/graphql-dotnet-ci/graphql-dotnet/builds/30930449

sungam3r:
Yeah, there were problems with this test for a long time.

src/GraphQL.DataLoader.Tests/SimpleDataLoaderTests.cs
```diff
@@ -59,7 +59,7 @@ public async Task Operation_Can_Be_Cancelled()
             mock.Setup(store => store.GetAllUsersAsync(cts.Token))
                 .Returns(async (CancellationToken ct) =>
                 {
-                    await Task.Delay(200);
+                    await Task.Delay(5000, ct);
                     ct.ThrowIfCancellationRequested();
```
sungam3r: If I understand this test correctly then it can be rewritten as 
```c#
            mock.Setup(store => store.GetAllUsersAsync(cts.Token))
                .Returns(async (CancellationToken ct) =>
                {
                    await Task.Delay(Timeout.Infinite, ct); // wait endlessly until the operation is canceled
                    return users;
                });
```

so no need to specify timeout - 200, 5000 or anything else.

Shane32: You probably could, but I wouldn’t unless we have continuing problems, or else upon failure it would hang forever

Shane32: We could go longer like a minute 60000. There should be no point to waiting longer than that.

Shane32: Updated to 60000ms

sungam3r: I understand that setting an endless timeout is dangerous. I just wanted to say that the algorithm in fact should always end with the cancellation of the operation, so you can wait endlessly inside the method. I'm OK with 5000 or 60000.

Shane32: Let's leave it longer.  I don't know how long appveyor might hang for, but in the debug environment, you can always just stop execution.  So a longer timeout should help tests pass.  Hopefully it will never hang that long, but a minute isn't a problem for appveyor.

PR #1556, opened 2020-02-20
Add xml comments to field resolvers

Shane32:
See changed files

### DIFF
--- a/src/GraphQL/Resolvers/AsyncFieldResolver.cs
+++ b/src/GraphQL/Resolvers/AsyncFieldResolver.cs
@@ -4,6 +4,10 @@ using GraphQL.Types;
 
 namespace GraphQL.Resolvers
 {
+    /// <summary>
+    /// When resolving a field, this implementation calls a predefined <see cref="Func{T, TResult}"/> and returns the result.
+    /// The returned value must be of an <see cref="Task{TResult}"/> type.
+    /// </summary>
     public class AsyncFieldResolver<TReturnType> : IFieldResolver<Task<TReturnType>>
     {
         private readonly Func<IResolveFieldContext, Task<TReturnType>> _resolver;
@@ -24,6 +28,7 @@ namespace GraphQL.Resolvers
         }
     }
 
+    /// <inheritdoc cref="AsyncFieldResolver{TReturnType}"/>
     public class AsyncFieldResolver<TSourceType, TReturnType> : IFieldResolver<Task<TReturnType>>
     {
         private readonly Func<IResolveFieldContext<TSourceType>, Task<TReturnType>> _resolver;
--- a/src/GraphQL/Resolvers/AsyncFieldResolver.cs
+++ b/src/GraphQL/Resolvers/AsyncFieldResolver.cs
@@ -17,6 +17,9 @@ namespace GraphQL.Resolvers
             _resolver = resolver;
         }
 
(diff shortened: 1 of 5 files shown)


src/GraphQL/Resolvers/AsyncFieldResolver.cs
```diff
@@ -24,6 +28,7 @@ object IFieldResolver.Resolve(IResolveFieldContext context)
         }
     }
 
+    /// <inheritdoc cref="AsyncFieldResolver{TReturnType}"/>
```
sungam3r: Would you like to say a few words about how this resolver differs from the previous one?

Shane32: Sure

src/GraphQL/Resolvers/FuncFieldResolver.cs
```diff
@@ -23,6 +26,7 @@ object IFieldResolver.Resolve(IResolveFieldContext context)
         }
     }
 
+    /// <inheritdoc cref="FuncFieldResolver{TReturnType}"/>
```
sungam3r: Here the same.

Shane32:
By the way, I think I’m about done with xml comments. I think the FieldType could really use some xml comments, probably some of the IGraphType and related interfaces, perhaps IValidationRule, and personally I’d do all the field builder methods. But I’m running out of ideas and have some other things I’d like to spend time on. If you want to take a stab at any of the above I’ll be happy to work with you on it.

src/GraphQL/Resolvers/NameFieldResolver.cs
```diff
@@ -5,6 +5,9 @@
 
 namespace GraphQL.Resolvers
 {
+    /// <summary>
+    /// Attempts to return a value for a field from the graph's source object, matching the name of the field to a property on the source object
```
sungam3r:
```suggestion
    /// Attempts to return a value for a field from the graph's source object, matching the name of the field to a property or a method with the same name on the source object
```

src/GraphQL/Resolvers/IFieldResolver.cs
```diff
@@ -1,14 +1,40 @@
...
+    /// <see cref="ComplexGraphType{TSourceType}.Field{TProperty}(System.Linq.Expressions.Expression{System.Func{TSourceType, TProperty}}, bool, System.Type)">Field(x => x.Name)</see>,
+    /// <see cref="ExpressionFieldResolver{TSourceType, TProperty}">ExpressionFieldResolver</see> is used.<br/>
+    /// <br/>
+    /// When a field resolver is not defined, such as with <see cref="ComplexGraphType{TSourceType}.Field{TGraphType, TReturnType}(string)">Field("Name")</see>,
+    /// the static instance of <see cref="NameFieldResolver"/> is used.
+    /// </summary>
     public interface IFieldResolver
     {
+        /// <summary>
+        /// Returns an object, null, or a <see cref="Task{TResult}"/> for the specified field
```
sungam3r:
```suggestion
        /// Returns an object, null, or a <see cref="Task{TResult}"/> for the specified field. If the task is returned, then this task will be awaited to obtain the actual object.
```

sungam3r:
>  and have some other things I’d like to spend time on

So do I.

PR #1559, opened 2020-02-21
Make ResolveFieldContextExtensions.As public

Shane32:
I'd like to make `ResolveFieldContextExtensions.As` public, as I have a need for it in one of my projects.  I can of course duplicate the entire `ResolveFieldContextAdapter` into my own project, but if it's okay, I'd like to simply make the method public.

### DIFF
--- a/src/GraphQL/ResolveFieldContextExtensions.cs
+++ b/src/GraphQL/ResolveFieldContextExtensions.cs
@@ -59,6 +59,7 @@ namespace GraphQL
         public static bool HasArgument(this IResolveFieldContext context, string argumentName) => context.Arguments?.ContainsKey(argumentName) ?? false;
 
         /// <summary>Returns the <see cref="IResolveFieldContext"/> typed as an <see cref="IResolveFieldContext{TSource}"/></summary>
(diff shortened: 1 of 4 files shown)


Shane32:
Making this method public should often be useful when making custom implementations of `IFieldResolver`

src/GraphQL/ResolveFieldContextExtensions.cs
```diff
@@ -58,15 +58,17 @@ private static bool TryGetArgument(this IResolveFieldContext context, System.Typ
         /// <summary>Determines if the specified field argument has been provided in the GraphQL query request</summary>
         public static bool HasArgument(this IResolveFieldContext context, string argumentName) => context.Arguments?.ContainsKey(argumentName) ?? false;
 
-        internal static IResolveFieldContext<TSourceType> As<TSourceType>(this IResolveFieldContext context)
+        /// <summary>Returns the <see cref="IResolveFieldContext"/> typed as an <see cref="IResolveFieldContext{TSource}"/></summary>
```
sungam3r: It would be good to mention throw conditions (inside ResolveFieldContextAdapter).

Shane32: Done; added to `As`, `ResolveFieldContextAdapter`, and the constructor of `ResolveFieldContext<TSource>`

PR #1561, opened 2020-02-23
Remove generic listeners, change userContext to IDictionary<string, object>

sungam3r:
See discussion in #1553

@joemcbride @Shane32 I propose changing the signature of all userContext params from object to `IDictionary<string, object>` and remove generic version of IDocumentExecutionListener.


sungam3r:
Also found #792 - this is exactly what we were talking about, @Shane32

sungam3r:
Need to consider #792 first.

BenjaBobs:
With #792 you could access the `UserContext` through `ExecutionContext` and it would be typed as `IDictionary<string, object>`, right?

sungam3r:
Right. I just want to make sure that it will be normal to change methods signatures with `ExecutionContext` - object that carries much more implementation details to these methods. Would you like to rebase #792 so that I can review it?

BenjaBobs:
I can do that, but I probably won't get have time for it right away since I'm only following this project on a hobby basis and don't have much time. I'll see if I can get some time this week, but no promises.

sungam3r:
We are not in a hurry.

sungam3r:
Closed in favor of #792

PR #1564, opened 2020-02-23
Change static introspection field types to instances

Shane32:
@sungam3r Take a look.  The `GraphTypesLookup` class creates the instances when it is constructed when it initializes its properties.  I then exposed the introspection types through `ISchema` because they are now inaccessible through the static instance, and also due to the fact that it is necessary to be accessible from `ExecutionHelper` (used by `ExecutionStrategy`).

Of note, now the `SchemaIntrospection` class only has a single member, `IntrospectionQuery`, which is only used for tests.


Shane32:
See comments in #1549 regarding this issue:

https://github.com/graphql-dotnet/graphql-dotnet/pull/1549#issuecomment-589872592

sungam3r:
I am very not sure that these properties need to be made public. This is the part of the library that user code does not need to see.

sungam3r:
`SchemaIntrospection.IntrospectionQuery` may be used by external client code. But I'm OK with moving it to test to cut public API surface.

sungam3r:
And I think that PR should be merged before #1549

Shane32:
> I am very not sure that these properties need to be made public

I'm open to suggestions, but if it's not part of `ISchema`, then it's not available to `ExecutionHelper.GetFieldDefinition`, which is used by `ExecutionStrategy`.  Interface members cannot be made `internal`.  I implement my own `IExecutionStrategy`.  So if it was not public then my implementation would not have access to it -- at least, not through `ISchema`.  Do you have another suggestion?

We certainly could make the implementations of them on `Schema` hidden so that they are not exposed through `Schema`.  But I'm not sure now it's so much different than the many other methods available on the `ISchema` to retrieve types.  For example, `ISchema.AllTypes` returns references to all the the introspection types among the returned values.

If you simply don't care for the fact that we are adding three more fields to `ISchema`, I'd agree, and suggest that instead we add a single property like this:

```C#
public interface ISchema
{
    ...
    IntrospectionFieldTypes IntrospectionFields { get; };
}
public class IntrospectionFieldTypes
{
    public FieldType SchemaMeta { get; } = new SchemaMetaFieldType();
    public FieldType TypeMeta { get; } = new TypeMetaFieldType();
    public FieldType TypeNameMeta { get; } = new TypeNameMetaFieldType();
}
{
    public static FieldType GetFieldDefinition(Document document, ISchema schema, IObjectGraphType parentType, Field field)
    {
        if (field.Name == schema.IntrospectionFields.SchemaMeta.Name && schema.Query == parentType)
        {
            return schema.IntrospectionFields.SchemaMeta;
        }
        ...
    }
}
```
Personally, I like the idea, but it does add another class to the exposed public API.

> And I think that PR should be merged before #1549

If you say so.  I don't see how one PR substantially affects the other.  They are dealing with completely different problems.

sungam3r:
>  But I'm not sure now it's so much different than the many other methods available on the ISchema to retrieve types. For example, ISchema.AllTypes returns references to all the the introspection **types** among the returned values.

Types, not fields.

sungam3r:
> So if it was not public then my implementation would not have access to it -- at least, not through ISchema. Do you have another suggestion?

Shane32:
> What if make `ISchema : IProvideMetadata` and then put these props inside metadata.

I wouldn't.  I see no reason to hide them; especially to bury access to them inside 'metadata', when they are required by the core of the project.

Besides, this is the **schema**.  It's supposed to hold references to all types required.  It has methods to access all the graph types, and all the introspection types, and through them, all the fields of those types (including fields of introspection graph types).  There's no reason whatsoever to hide, arbitrarily, the three introspection fields, just because they are not associated with a particular graph type.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
```diff
@@ -802,9 +802,6 @@ namespace GraphQL.Introspection
     public static class SchemaIntrospection
     {
         public static readonly string IntrospectionQuery;
```
sungam3r: Let's move this to test to cut public API surface.

Shane32: Agreed. Lmk where

Shane32: I guess anywhere will do if it’s tests. Not at a PC just now.

sungam3r:
>  I see no reason to hide them; especially to bury access to them inside 'metadata', when they are required by the core of the project.

The reason is not to give public access to things for which it is not needed. The project already has a lot of open **internal** APIs.

> when they are required by the core of the project.

Of course, **core project logic** requires access to these fields. But this is our problem, not the problem of our clients. You expose these internal properties to public API only because one internal module communicates with another internal module through the public interface. The problem is precisely the mixing of areas of responsibility. In such a situation, it might seem natural to expand the interface with new entities, but this is a mistake. You are expanding the interface due to implementation needs. This is a classic case of leaky abstractions.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
sungam3r: Anywhere will be OK :)

Shane32:
I, respectfully, disagree.  The design of the project is for each piece to be replaceable -- open architecture.  If it were not so, there would be no reason to have IExecutionStrategy public, or many other interfaces throughout the project.

We are also not discussing hiding members of a class where one can replace the entire class with their own implementation.  For example, the `HandleField` method of the `GraphTypesLookup` class. 
 We are discussing one interface's necessary ability to communicate with another interface, and specifically where the project was designed to allow you to choose different execution strategies as necessary.  The interfaces should expose all the necessary properties so that if one piece is replaced, it has all of the methods and properties available to it to allow it to perform its work.

Perhaps we have too many interfaces in some situations.  For instance, I do not know the purpose of the `IFieldType` interface, when the `IComplexGraphType` interface requires the `FieldType` class passed to it.  But nevertheless, the `FieldType` class is still public.

I honestly don't understand why this is not clear.  If the project was closed-architecture, this would not be a point of discussion.  All of the execution strategies, the GraphTypesLookup class, the introspection types, and so on would be hidden.  The properties on the Schema object would be a relatively few properties designed to initially configure it.  Classes like scalar graph types would still be exposed because they would be necessary.  Again, this is not the design architecture of this project.

It is possible to delete some interfaces, and require inheritance to implement custom code.  For example, if `ISchema` and `IExecutionStrategy` were removed, then protected members of `ExecutionStrategy` could interact with `Schema` internally to provide access to members as necessary.  A good example is `FieldType` - there is likely no reason for `IFieldType`.  Also, someone could inherit from `Schema` or `ExecutionStrategy` and expose protected members in their own code easily when required.  This would be a significant change to the project (as it could probably apply in a number of areas), and although I would not necessarily favor it, I would not be opposed to it either.  If we do this, I would want to ensure that `virtual` and `protected` are used liberally throughout the project.  As it is now, very few members are `virtual` or `protected` (for example, the `Schema` class has no `virtual` or `protected` members except `Dispose`), but as the project is mostly based on interfaces, it does not matter much.

Without that change, I do not feel that we should attempt to hide "core project logic" by burying it in metadata, for information which is required to be passed between two interfaces.  There is no similar code pattern in the existing code base.  I'm open to other alternatives, such as refactoring the code so the dependency is not required -- if such other alternatives exist.  I also would not mind if the members are not exposed on the `Schema` class.  But until such time as then entire inheritance/interface strategy is reconsidered for the entire project, they should be exposed on the `ISchema` interface in some fashion (or refactored to eliminate the need).

As a side note; this is why I was distinctly against removing `ISchema.Services`.  It's an anti-pattern in itself.  If we want to hide `Services`, then `ISchema` should not exist and all schemas should inherit from `Schema`, which should expose `Services` as `protected internal`, so the middleware can access it -- not by casting `ISchema` to `IServiceProvider`.  (And I do not think `Services` should be removed at all from `ISchema`, but we've already discussed that in depth.)

Shane32:
I also don’t see the logic in hiding three introspection fields while leaving all the introspection types and their fields exposed. It should be all or none, subject to my other comments, above.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
Shane32: I cannot move it until #1549 is pulled in, as the name converters rely on `nameof(SchemaIntrospection)`.  After #1549 then I can make this change.

Shane32:
This PR is on hold until #1549 is pulled in

sungam3r:
> It should be all or none

In the long run, I have nothing against it. In the short term, I doubt that we can bite off such a large piece at a time.

> The design of the project is **for each** piece to be replaceable

Even assuming that this **was** so at the initial stage of development, this approach has drawbacks.

It will be useful for all us to hear other opinions. @joemcbride @benmccallum @pekkah @SimonCropp  @RehanSaeed I ask you to join the discussion.

benmccallum:
I don't really have the time or head-space to think this through to the level you guys have already, but I do (I think) in general agree with the concept of balancing the desire to limit our public api but not be so afraid that we lock things down so tightly that others can't extend/modify implementations as they desire. 

One of the trickiest parts of building on this fx is that we don't have consistent, stable releases and ownership to the point where this is possible. It's no-one's fault mind you, it's just the nature of the beast and the time we all have. But, as such, extension points are necessary for us to all have sanity, e.g. be able to extend until we can get something added for real into the core project. 

But then on the flip-side, we don't want to promise these extension points too much if we continue to iterate and make breaking changes. 

It's a tough one :P Maybe it depends on how confident we are per feature/PR on how stable that will be in the long run.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
Shane32: Done

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
```diff
@@ -1882,7 +1878,10 @@ namespace GraphQL.Types
         GraphQL.Types.IObjectGraphType Mutation { get; set; }
         GraphQL.Conversion.INameConverter NameConverter { get; set; }
         GraphQL.Types.IObjectGraphType Query { get; set; }
+        GraphQL.Types.FieldType SchemaMetaFieldType { get; }
```
sungam3r: By the way this is a breaking change for custom schemas (if any).

Shane32: If it gets pulled, I'll put it in the upgrade notes

Shane32:
We could change the three additional fields to a single method `GetIntrospectionField<T>()` which compares `typeof(T)` to `typeof(SchemaMetaFieldType)` and so on.  Still a breaking change but lowers number of exposed methods.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
sungam3r: Of course it will be pulled.

src/GraphQL/Types/ISchema.cs
```diff
@@ -141,5 +141,20 @@ public interface ISchema
...
+        /// </summary>
+        FieldType SchemaMetaFieldType { get; }
+
+        /// <summary>
+        /// Returns a reference to the __type introspection field available on the query graph type
+        /// </summary>
+        FieldType TypeMetaFieldType { get; }
+
+        /// <summary>
+        /// Returns a reference to the __typename introspection field available on any graph type
```
sungam3r: _on any_ - no

http://spec.graphql.org/June2018/#sec-Type-Name-Introspection :
> GraphQL supports type name introspection at any point within a query by the meta‐field __typename: String! **when querying against any Object, Interface, or Union**.

sungam3r:
Looks good. Just one fix in comments before merge.

sungam3r:
> We could change the three additional fields to a single method GetIntrospectionField<T>() which compares typeof(T) to typeof(SchemaMetaFieldType) and so on. Still a breaking change but lowers number of exposed methods.

Not worth it.

src/GraphQL/Types/ISchema.cs
Shane32:
```suggestion
        /// Returns a reference to the __typename introspection field available on object, interface, or union graph types
```
How about that?

sungam3r: OK

sungam3r: on **any** object....  ?

Shane32: it's valid now, but that also works; need to change "types" to "type"

src/GraphQL/Types/ISchema.cs
```diff
@@ -153,7 +153,7 @@ public interface ISchema
         FieldType TypeMetaFieldType { get; }
 
         /// <summary>
-        /// Returns a reference to the __typename introspection field available on any graph type
+        /// Returns a reference to the __typename introspection field available on object, interface, or union graph types
```
Shane32:
```suggestion
        /// Returns a reference to the __typename introspection field available on any object, interface, or union graph type
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -141,5 +141,20 @@ public interface ISchema
...
+        /// </summary>
+        FieldType SchemaMetaFieldType { get; }
+
+        /// <summary>
+        /// Returns a reference to the __type introspection field available on the query graph type
+        /// </summary>
+        FieldType TypeMetaFieldType { get; }
+
+        /// <summary>
+        /// Returns a reference to the __typename introspection field available on object, interface, or union graph types
```
Shane32:
```suggestion
        /// Returns a reference to the __typename introspection field available on any object, interface, or union graph type
```

src/GraphQL/Types/ISchema.cs
sungam3r: OK

benmccallum:
> SchemaIntrospection.IntrospectionQuery may be used by external client code. But I'm OK with moving it to tests to cut public API surface.

Was using this in Benchmarks over in the Server project ;P. Will just copy paste I guess.

sungam3r:
I know, I know. This is just one "standard" string literal that could be copied.

PR #1567, opened 2020-02-24
Move all ResolveFieldContext classes and interfaces from GraphQL.Types namespace into GraphQL namespace

sungam3r:
Also a few code style rules applied.


sungam3r:
These are breaking changes but I'm sure it’s worth making them in order to group APIs correctly. In addition, with high probability, customers will only need to recompile the code, and in the worst case, add some usings.

Shane32:
These changes will be convenient for implementers.  Better to get it in now before 3.0 drops and add it to the upgrade guide.

sungam3r:
I'm thinking of making red `BREAKING` label. What do you think?

Shane32:
If you're asking me, that's a good idea.  It will be a lot easier to keep track of breaking changes post-3.0.

sungam3r:
done

PR #1571, opened 2020-02-24
Add MS-specific dependency injection extensions

Shane32:
This is a follow-up to the discussion in #1345 

This PR adds `Services` back to `ISchema` and adds `RequestServices` to `IResolveFieldContext` and `ExecutionOptions`.  This allows implementers to access the currently executing request's scoped service provider.  No additional dependencies or functionality is added to the base GraphQL project.

The naming of the `IResolveFieldContext.RequestServices` property matches Asp.Net Core, where their `HttpContext` has a `RequestServices` property to access the service provider of the executing request.

I also added the recently-removed `Services` property back into the `ISchema` interface, so code can access the service provider used to create the graph type and other schema-specific classes if desired.

By adding `RequestServices` into the GraphQL pipeline, resolvers that are based on scoped services are free from framework-specific dependencies, such as `IHttpContextAccessor.HttpContext.RequestServices`, or needing to pass the service provider or services through the `IResolveFieldContext.UserContext` property.  It also alleviates the need to create a custom service accessor singleton.

Anyone that wishes to keep their dependent services listed in the constructor can simply register the schema as a scoped service, as they have always done.  For those of us that want to keep the performance of the schema as a singleton, 

I've also added a GraphQL.Extensions.DI.Microsoft project, which provides an easy way to have parallel graphql resolvers execute scoped services not designed for multithreaded operation (as most scoped services are not, such as Entity Framework).  Simply call `Field().ResolveScoped` or `Field().ResolveAsyncScoped` with the same field resolver, and the extension will create a DI scope during the execution of the resolver, disposing of it appropriately when complete.  This project works with any DI container compatible with the necessary interface - notably, autofac is also compatible.  I feel that these additional extension methods would be convenient, but not necessary as they can always use the `SerialExecutionStrategy` and ignore this issue.  I would include this functionality as notes in the documentation, but it seems like a bit too much code for that.  Perhaps it can be simply a link in the documentation to a project with these extra classes, or perhaps it should be released as another nuget package.

I had one additional idea, not in this PR, to reduce the calls to `RequestServices`:  We could add additional `Resolve` overloads for resolvers that need services.  For example, one could write:
```C#
Field<ListGraphType<UserType>, IEnumerable<User>>("Users")
    .ResolveAsync<MyDbContext>((context, dbContext) => dbContext.Users.ToListAsync());
```
The resolver would simply pull the requested services from the context's RequestServices property prior to calling the lambda function.

Unfortunately it cannot be implemented as an extension method very well, as the call to `ResolveAsync` would need to include the `TSourceType` and `TReturnType` types also.  So it doesn't lend itself well to adding these methods to a third party library.  If we add this feature, it would have to be directly on the `FieldBuilder<TSourceType, TReturnType>` class.


Shane32:
@sungam3r We've already added the `RequestServices` property to `IResolveFieldContext`.  The rest of this PR contains some proposed extensions for adding scoped field resolvers.  Are you interested in:

1. Keeping this PR open for now, just for reference
2. Replace with a new PR containing the suggestions that are not yet merged
3. Close for now

I'm okay either way.  Now that we have the `RequestServices` property, I can throw these extensions directly into my projects, so it's no big deal.  I might even consider posting a separate nuget package with some of these extensions if you don't want them as part of the base graphql project.  Just lmk.

sungam3r:
I'm absolutely sure that after #1730, we have to go ahead and include these changes, so option 2 or resolve conflicts/rebase here.

Shane32:
I merged changes here, and removed `Services` from `ISchema`.

sungam3r:
Only additions now, no deletions. Good. Will review soon.

sungam3r:
@Shane32 
```sh
C:\Program Files\dotnet\sdk\3.1.302\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(187,6): warning : This project cannot be packaged because packaging has been disabled. Add <IsPackable>true</IsPackable> to the project file to enable producing a package from this project. [C:\projects\graphql-dotnet\src\GraphQL.Harness\GraphQL.Harness.csproj]
C:\Program Files\dotnet\sdk\3.1.302\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(198,5): error NU5046: The icon file 'logo.64x64.png' does not exist in the package. [C:\projects\graphql-dotnet\src\GraphQL.Extensions.DI.Microsoft\GraphQL.Extensions.DI.Microsoft.csproj]
```

Shane32:
I've added a rough draft of a 'resolver injection' technique to this PR.  It requires review.  Since it's impossible with extension methods, I added builder classes to facilitate this.  Perhaps you have a better idea @sungam3r but I'm not sure how else to do it.  Anyway, with this sample, you could write code like this:
```csharp
Field<StringGraphType, string>("test")
    .Argument(...)
    .Resolve()
        .WithScope()
        .WithService<MyService>()
        .WithService<MyOtherService>()
        .ResolveAsync(async (context, service, service2) => await service.GetAnswer(context.Source))
    .Description("hello");
```
Of course, the `WithScope` would only be needed if you specifically wanted to create a scope for this field resolver.  Probably when using the parallel execution method with EF or similar.  Whereas with the serial execution method, you would not need `WithScope` and it would pull the services directly from the service provider.

This resolve builder is within the `GraphQL.Extensions.DI.Microsoft` project.  However, the builder is not really specific to MS at all - only creating the scope relies on Microsoft DI's `CreateScope` extension method, which simply relies on an implementation of `IServiceScopeFactory` to be registered with the base service provider, and then retrieves the `IServiceProvider` from the provided `IServiceScope` interface returned from the factory.

Personally I think we should combine the two projects together.  Someone can always register their own implementation of an `IServiceScopeFactory` if they want to create scopes with an alternate service provider.  We could of course create our own interface for creating service scopes - but there's no reason to rewrite good code that already exists, especially as an abstraction.  It's certainly an option, however, if we want to publish this project with no dependencies whatsoever.

Finally, I would not let this hold up 3.0.  Nothing here is going to be a breaking change, so there's no reason it can't make it into 3.1 if we have not the time to review it fully.  I would like to keep this on the top of the pile, however, and if we can get this done quickly, that's great!

Shane32:
@sungam3r If you like the idea, let me know, and I can add tests and so on.  I just threw this together real quick.

src/GraphQL.sln
```diff
@@ -83,6 +83,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GraphQL.SystemTextJson", "G
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Serialization", "Serialization", "{7BA2B19D-2839-4C7D-9BB0-4F3ADE691ADD}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GraphQL.Extensions.DI.Microsoft", "GraphQL.Extensions.DI.Microsoft\GraphQL.Extensions.DI.Microsoft.csproj", "{806E7A14-668C-46A9-9C68-E98A49400E08}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "DependencyInjection", "DependencyInjection", "{2EA5354E-F4EF-4469-86AE-F1D36C7F8977}"
```
sungam3r, line 88: Is this a solution directory?

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace GraphQL.Extensions.DI.Microsoft
+{
+    public class ScopedFieldResolver<TReturnType> : FuncFieldResolver<TReturnType>
+    {
+        public ScopedFieldResolver(Func<IResolveFieldContext, TReturnType> resolver):base(GetScopedResolver(resolver)) { }
+
+        private static Func<IResolveFieldContext, TReturnType> GetScopedResolver(Func<IResolveFieldContext, TReturnType> resolver)
+        {
+            return (context) =>
```
sungam3r, line 13:
```suggestion
            return context =>
```

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                using (var scope = context.RequestServices.CreateScope())
```
sungam3r, line 15:
```suggestion
                using (var scope = (context.RequestServices ?? throw new InvalidOperationException("YOUR_MESSAGE")).CreateScope())
```

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }
+
+    public class ScopedFieldResolver<TSourceType, TReturnType> : FuncFieldResolver<TSourceType, TReturnType>
+    {
+        public ScopedFieldResolver(Func<IResolveFieldContext<TSourceType>, TReturnType> resolver) : base(GetScopedResolver(resolver)) { }
+
+        private static Func<IResolveFieldContext<TSourceType>, TReturnType> GetScopedResolver(Func<IResolveFieldContext<TSourceType>, TReturnType> resolver)
+        {
+            return (context) =>
```
sungam3r, line 29:
```suggestion
            return context =>
```

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                using (var scope = context.RequestServices.CreateScope())
```
sungam3r, line 31:
```suggestion
                using (var scope = (context.RequestServices ?? throw new InvalidOperationException("YOUR_MESSAGE")).CreateScope())
```

src/GraphQL.sln
Shane32: I believe so

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
Shane32:
```suggestion
                using (var scope = (context.RequestServices ?? throw new InvalidOperationException("No service provider specified. Please set the value of the ExecutionOptions.RequestServices to a valid service provider.  Typically, this would be a scoped service provider from your dependency injection framework.")).CreateScope())
```

sungam3r: ok

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
Shane32:
```suggestion
                using (var scope = (context.RequestServices ?? throw new InvalidOperationException("No service provider specified. Please set the value of the ExecutionOptions.RequestServices to a valid service provider. Typically, this would be a scoped service provider from your dependency injection framework.")).CreateScope())
```

sungam3r: ok

Shane32:
These extension functions rely on a Nuget dependency: `Microsoft.Extensions.DependencyInjection.Abstractions`.  Autofac already supports these abstractions, as will any dependency injection framework that supports ASP.Net Core.  So these extensions can be considered generic extensions, rather than MS-specific.

sungam3r:
Well, it is microsoft namespace :)

sungam3r:
By the way, do you know whether autofac supports replacing registrations?

Shane32:
I am not sure.  I do use autofac's 'decorator' support whereby you can intercept calls to an already-registered service.  It's almost like deriving a class, except you don't have access to protected members.  It works very nicely when the class you're trying to 'decorate' is in another library.

Shane32:
@sungam3r We had planned to release this for 3.1, but never did.  Can we release this with 4.0?  We need to review:
* Name of project/nuget package
* Resolver builders -- probably at the same time as we review the argument builders

I should add some tests and a few xml comments.

sungam3r:
Give me 1-2 days. And side note - https://github.com/dotnet/runtime/issues/47113. I found this by accident. Perhaps this is the reason for the falling test.

src/GraphQL.sln
```diff
@@ -84,6 +84,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GraphQL.SystemTextJson", "G
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Serialization", "Serialization", "{7BA2B19D-2839-4C7D-9BB0-4F3ADE691ADD}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GraphQL.Extensions.DI.Microsoft", "GraphQL.Extensions.DI.Microsoft\GraphQL.Extensions.DI.Microsoft.csproj", "{806E7A14-668C-46A9-9C68-E98A49400E08}"
```
sungam3r, line 87: I would shorten this name to `GraphQL.MicrosoftDI` as it is done for other projects.

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using GraphQL.Resolvers;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GraphQL.Extensions.DI.Microsoft
+{
+    public class ScopedFieldResolver<TReturnType> : FuncFieldResolver<TReturnType>
+    {
+        public ScopedFieldResolver(Func<IResolveFieldContext, TReturnType> resolver):base(GetScopedResolver(resolver)) { }
```
sungam3r, line 9:
```suggestion
        public ScopedFieldResolver(Func<IResolveFieldContext, TReturnType> resolver) : base(GetScopedResolver(resolver)) { }
```

src/GraphQL.Extensions.DI.Microsoft/GraphQL.Extensions.DI.Microsoft.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <RootNamespace>GraphQL.Extensions.DI.Microsoft</RootNamespace>
```
sungam3r, line 5:
```suggestion
```

src/GraphQL.Extensions.DI.Microsoft/ScopedResolveFieldContextAdapter.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+            }
+        }
+
+        public new TSource Source { get; }
+
+        TSource IResolveFieldContext<TSource>.Source => Source;
+    }
+
+    internal class ScopedResolveFieldContextAdapter : IResolveFieldContext, IResolveFieldContext<object>
```
sungam3r, line 39:
```suggestion
    internal sealed class ScopedResolveFieldContextAdapter : IResolveFieldContext, IResolveFieldContext<object>
```

src/GraphQL.Extensions.DI.Microsoft/ScopedResolveFieldContextAdapter.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using GraphQL.Execution;
+using GraphQL.Instrumentation;
+using GraphQL.Language.AST;
+using GraphQL.Types;
+
+namespace GraphQL.Extensions.DI.Microsoft
+{
+    internal class ScopedResolveFieldContextAdapter<TSource> : ScopedResolveFieldContextAdapter, IResolveFieldContext<TSource>
```
sungam3r, line 11:
```suggestion
    internal sealed class ScopedResolveFieldContextAdapter<TSource> : ScopedResolveFieldContextAdapter, IResolveFieldContext<TSource>
```

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using GraphQL.Resolvers;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GraphQL.Extensions.DI.Microsoft
+{
+    public class ScopedFieldResolver<TReturnType> : FuncFieldResolver<TReturnType>
```
sungam3r, line 7: Maybe it is worth making `ScopedFieldResolver`s `internal` ?

sungam3r: The same for both `ScopedAsyncFieldResolver`s.

src/GraphQL.Extensions.DI.Microsoft/ScopedAsyncFieldResolver.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    public class ScopedAsyncFieldResolver<TSourceType, TReturnType> : AsyncFieldResolver<TReturnType>
+    {
+        public ScopedAsyncFieldResolver(Func<IResolveFieldContext<TSourceType>, Task<TReturnType>> resolver) : base(GetScopedResolver(resolver)) { }
+
+        private static Func<IResolveFieldContext, Task<TReturnType>> GetScopedResolver(Func<IResolveFieldContext<TSourceType>, Task<TReturnType>> resolver)
+        {
+            return async (context) =>
+            {
+                using (var scope = context.RequestServices.CreateScope())
```
sungam3r, line 32: (context.RequestServices ?? throw new InvalidOperationException("No....

src/GraphQL.Extensions.DI.Microsoft/ScopedAsyncFieldResolver.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public class ScopedAsyncFieldResolver<TReturnType> : AsyncFieldResolver<TReturnType>
+    {
+        public ScopedAsyncFieldResolver(Func<IResolveFieldContext, Task<TReturnType>> resolver) : base(GetScopedResolver(resolver)) { }
+
+        private static Func<IResolveFieldContext, Task<TReturnType>> GetScopedResolver(Func<IResolveFieldContext, Task<TReturnType>> resolver)
+        {
+            return async (context) =>
+            {
+                using (var scope = context.RequestServices.CreateScope())
```
sungam3r, line 16: (context.RequestServices ?? throw new InvalidOperationException("No

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
sungam3r, line 7: Is `ScopedFieldResolver<TReturnType>` class used anywhere (`ScopedAsyncFieldResolver<TReturnType>` as well) ?

src/GraphQL.sln
sungam3r: The same suggestion for `GraphQL.Extensions.DI.Microsoft` namespace.

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
Shane32: I would definitely use them in my implementation.

Shane32: Yes - ScopedFieldBuilderExtensions.cs

Shane32:
Good comments.  I'll work on it.

src/GraphQL.Extensions.DI.Microsoft/ScopedFieldResolver.cs
sungam3r: How? As base classes?

sungam3r: I talked about `ScopedFieldResolver<TReturnType>` not `ScopedFieldResolver<TSourceType, TReturnType>`

Shane32: Here is one example:

https://github.com/Shane32/GraphQL.DI/blob/1779914c4ebc52305df26497dec0ffc05cbfa391/src/DIObjectGraphType.cs#L228

Shane32: Oh. They might not be.

sungam3r:
I probably won't be working on other tasks until v4. Let's wrap up this one.

Shane32:
Replaced with #2261

PR #1572, opened 2020-02-26
Change IFieldType and IProvideMetadata to read-only interfaces

Shane32:
Another conceptual draft to fix #1564 

Changes:
* `IFieldType` has changed from a read/write type to a read-only type
* `IProvideMetadata` is now a read-only type
* `ISchema` has a single additional method to access the introspection fields: `GetIntrospectionField<T>()` which returns an `IFieldType`
* `FieldType`, `MetadataProvider`, `GraphType` and other classes still provide read/write access to the underlying fields and metadata dictionary
* Various changes throughout the code base were changed from `FieldType` to `IFieldType` to support this change
* Due to the above changes, `IGraphType` (which inherits from `IProvideMetadata`) now has read-only access to metadata, although all classes that inherit from `GraphType` still have read/write access to the metadata

Thoughts:
* This concept could be applied to graph types as well.  `IGraphType` would change to read-only. 
 Many references to `IGraphType` would change to `GraphType`, as they would require read/write access (to set `ResolvedType` or apply middleware, for example).
* With these changes in place, marking certain fields `internal` is possible where it is not possible now (since `IGraphType` is used extensively, and interface members cannot be marked `internal`).
* I would endorse these changes across the entire project

### DIFF
--- a/src/GraphQL/Types/Schema.cs
+++ b/src/GraphQL/Types/Schema.cs
@@ -121,11 +121,20 @@ namespace GraphQL.Types
 
         public IEnumerable<Type> AdditionalTypes => _additionalTypes;
 
-        public FieldType SchemaMetaFieldType => _lookup?.Value.SchemaMetaFieldType;
+        /// <summary>
+        /// Returns a reference to the __schema introspection field available on the query graph type
+        /// </summary>
+        public IFieldType SchemaMetaFieldType => _lookup?.Value.SchemaMetaFieldType;
 
-        public FieldType TypeMetaFieldType => _lookup?.Value.TypeMetaFieldType;
+        /// <summary>
+        /// Returns a reference to the __type introspection field available on the query graph type
+        /// </summary>
+        public IFieldType TypeMetaFieldType => _lookup?.Value.TypeMetaFieldType;
 
-        public FieldType TypeNameMetaFieldType => _lookup?.Value.TypeNameMetaFieldType;
+        /// <summary>
+        /// Returns a reference to the __typename introspection field available on any graph type
+        /// </summary>
+        public IFieldType TypeNameMetaFieldType => _lookup?.Value.TypeNameMetaFieldType;
 
         public void RegisterType(IGraphType type)
         {
--- a/src/GraphQL/Types/Schema.cs
+++ b/src/GraphQL/Types/Schema.cs
@@ -121,19 +121,10 @@ namespace GraphQL.Types
 
         public IEnumerable<Type> AdditionalTypes => _additionalTypes;
 
-        /// <summary>
-        /// Returns a reference to the __schema introspection field available on the query graph type
-        /// </summary>
         public IFieldType SchemaMetaFieldType => _lookup?.Value.SchemaMetaFieldType;
 
-        /// <summary>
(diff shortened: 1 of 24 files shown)


src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -308,6 +308,7 @@ public void AddType(IGraphType type, TypeCollectionContext context)
                     union.AddPossibleType(objType);
                 }
             }
+
```
sungam3r:
```suggestion
```

src/GraphQL/Types/Schema.cs
```diff
@@ -208,6 +223,18 @@ public IGraphType FindType(string name)
             return _lookup?.Value[name];
         }
 
+        public IFieldType GetIntrospectionField<T>()
```
sungam3r: Maybe switch expression?

sungam3r:
I have some concerns that fewer APIs will become available to clients. But this is obvious and I hope that nothing bad will happen, at least in the vast majority of scenarios. It seems to me that such a read/write divergence of APIs is due to the fact that the schema is used in two scenarios - constructing a schema and executing user code using this schema. We need approval from @joemcbride for this PR. It’s also obvious that this is only the first step in changing the design, so an overall strategy needs to be approved.

sungam3r:
@Shane32 Is this still actual?

src/GraphQL/Types/Schema.cs
Shane32: Unfortunately impossible due to `typeof(T)` is not a constant

Shane32: However, we could exchange this method for three separate fields, as was done for #1564 which was pulled in earlier today.

Shane32:
This PR is "phase 1" of #1578 - I left it here for discussion purposes.  #1578 is more of a WIP that simply takes this PR and changes tons of more code to continue with the same design philosophy of read-write classes to define the schema, and read-only interfaces to read from the schema.

Shane32:
This PR does change IFieldType to be read-only, so it's a complete and independent step from changing IGraphType to be read-only.  If we want to pull it, we can.  I just would need to tweak it a bit and merge any other changes from the project.  Then you could review it prior to it being pulled.

Shane32:
@sungam3r This PR is now focused on changing IFieldType to read-only, as we have already pulled in the changes to make the introspection fields instances on the schema.  If this looks good to everyone, we can pull it in and review changing IGraphType to read-only (PR #1578 ) at an appropriate time.  I suggest you review again.

Shane32:
Note that a few more separate changes are required before `IFieldType` is 100% read-only:
- `QueryArgument` needs to implement a read-only `IQueryArgument` See #1594 
- `IFieldType.Arguments` needs to be of type `IEnumerable<IQueryArgument>` or a similar.  See #1593 which contains a supporting change.

sungam3r:
I began to doubt whether all types should be done "readonly". I have not done a review yet.

Shane32:
I'm fine with closing this PR and leaving the class/interface structure alone

sungam3r:
OK.

PR #1574, opened 2020-02-26
cut APIs: remove middlewareresolver and synctoasyncresolveradapter in favor of extension method

### LINKED ISSUE #1573: MiddlewareResolver and SyncToAsyncResolverAdapter are nearly identical
`MiddlewareResolver` and `SyncToAsyncResolverAdapter` are nearly identical.  Should one be deleted?

* `SyncToAsyncResolverAdapter` was added since 2.4.0, so deleting it would not be a breaking change from 2.4.0
* It functions almost identical to MiddlewareResolver, except it does not change a null inner `IFieldResolver` to the default `NameFieldResolver.Instance`
* `BaseSchemaNodeVisitor.WrapResolver` is the **only** function to use `SyncToAsyncResolverAdapter`, and does change a null inner `IFieldResolver` to the default `NameFieldResolver.Instance`, thereby **making the function of both resolvers the same**, when used in this function
* Currently, `WrapResolver` returns a `SyncToAsyncResolverAdapter` class, which could be changed to an `IFieldResolver<Task<object>>` signature.  However, the former has a `ResolveAsync` method, where the latter would just be called `Resolve`.  So there's a slight signature difference.

Proposals:
1. Delete `SyncToAsyncResolverAdapter` and make `WrapResolver` return an `IFieldResolver<Task<object>>`, which is a `MiddlewareResolver` (breaking change since the `ResolveAsync` method is now called `Resolve`, although not breaking compared to 2.4.0)
2. Delete `MiddlewareResolver` and change references to it to point to `SyncToAsyncResolverAdapter`. Middleware changed plenty since 2.4.0 anyway, so if this is breaking, it doesn't matter much.
3. Have one of the two classes inherit from the other.  Nothing breaking from current branch.
4. Leave it alone.

sungam3r:
fixes #1573


src/GraphQL/Resolvers/FieldResolverExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace GraphQL.Resolvers
+{
+    public static class FieldResolverExtensions
+    {
+        public static async Task<object> ResolveAsync(this IFieldResolver resolver, IResolveFieldContext context)
```
sungam3r: We might think of a better name for this method.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -181,16 +181,7 @@ protected virtual async Task ExecuteNodeAsync(ExecutionContext context, Executio
...
-                var result = resolver.Resolve(resolveContext);
-
-                if (result is Task task)
-                {
-                    await task.ConfigureAwait(false);
-                    result = task.GetResult();
-                }
-
-                node.Result = result;
+                node.Result = await (node.FieldDefinition.Resolver ?? NameFieldResolver.Instance).ResolveAsync(resolveContext);
```
Shane32: Nice for cleanup, but this change will create an object of type Task<object> for every resolved field. I'd advise not to change this section, or permanently change `IFieldResolver.Resolve` to `ResolveAsync`.  See #1527

src/GraphQL/Resolvers/FieldResolverExtensions.cs
Shane32: `ResolveAsync` seems fine to me.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Of note, whether this change is pulled or not, any schema with middleware will be affected by the performance decrease documented in #1380 at the very least

sungam3r: Agree about `Task<object>` allocations. I thought about it when I wrote that extension. I decided to postpone the decision until the benchmarks run. Before and after:

| Introspection | 2,669.90 us | 22.349 us | 19.812 us | 199.2188 | 54.6875 |     - | 486.73 KB |
|          Hero |    59.27 us |  0.296 us |  0.277 us |   8.8501 |       - |     - |  13.61 KB |


|        Method |        Mean |     Error |     StdDev |      Median |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |------------:|----------:|-----------:|------------:|---------:|--------:|------:|----------:|
| Introspection | 2,899.30 us | 59.505 us | 169.771 us | 2,849.28 us | 214.8438 | 50.7813 |     - | 520.96 KB |
|          Hero |    58.96 us |  0.420 us |   0.328 us |    58.95 us |   8.9111 |       - |     - |  13.73 KB |

I will roll back changes to this file. In two other cases, the use of this method will not get worse, since there have already been `Task<object>` allocations. Agree?

sungam3r:
Getting `StackOverflowException` on AppVeyor, but OK locally.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Agreed

Shane32:
So the foreach variable was changing before the lambda executed probably?

sungam3r:
`StackOverflowException` was due to lambda closure in `FieldMiddlewareBuilder.ApplyTo`. Fixed!

Shane32:
Interesting that there was no warning in VS with C#.  I know that in VB.Net it's a very common warning to get: "iteration variable used within lambda".

sungam3r:
In two other places I always have the correct code. Moreover, I wrote this on purpose, knowing about such behavior. And in the case of `FieldMiddlewareBuilder` apparently just did not notice! 🤦‍♂

sungam3r:
> Interesting that there was no warning in VS with C#. I know that in VB.Net it's a very common warning to get: "iteration variable used within lambda".

Of greater interest to me is the fact that there are no errors when running locally.

sungam3r:
I think that the adventure has come to an end and it’s time to merge.

Shane32:
I'm reading multiple notes on this, but all from unofficial sources.  One says this:

> Avoid capturing the loop variable of foreach statements in lambda or delegate functions because the value captured varies depending on the version of C# used. In C# 5.0 and above the value captured is the original value but in earlier versions of C# the value captured is the current value. If the function is used outside of the loop, then the function will behave differently in different versions of C#.

In any case, it seems clear that different runtimes' lambda expressions either capture or don't capture the variable within foreach.  Also it can be different between for and foreach.

sungam3r:
I know. Sometimes you know where the puddle is, but still you step into it :)

PR #1577, opened 2020-02-26
More conversions

### LINKED ISSUE #956: Ability to short circuit ienum conversions to be able to handle b64 => byte[]
## Summary

Brief explanation of the feature.

I would like to have a simple way to have a conversion of type string => byte[]

currently this will decompose to an attempt to convert char => byte.

### Motivation

Why are we doing this? What use cases does it support? What is the expected outcome?

Allow for passing binary information as arguments

sungam3r:
fixes #956

### DIFF
--- a/src/GraphQL/Conversion/ValueConverter.cs
+++ b/src/GraphQL/Conversion/ValueConverter.cs
@@ -114,6 +114,11 @@ namespace GraphQL
             Register(typeof(double), typeof(decimal), value => Convert.ToDecimal(value, NumberFormatInfo.InvariantInfo));
 
             Register(typeof(string), typeof(Uri), value => value is string s ? new Uri(s) : (Uri)value);
+            // registering such a default conversion for string->byte[] seems useful
+            Register(typeof(string), typeof(byte[]), value => Convert.FromBase64String((string)value));
+
+            Register(typeof(char), typeof(byte), value => Convert.ToByte((char)value));
+            Register(typeof(char), typeof(int), value => Convert.ToInt32((char)value));
         }
 
         public static object ConvertTo(object value, Type targetType)
@@ -140,15 +145,22 @@ namespace GraphQL
             throw new InvalidOperationException($"Could not find conversion from '{valueType.FullName}' to '{targetType.FullName}'");
         }
 
+        /// <summary>
+        /// Allows you to register your own conversion method from one type to another.
+        /// If the conversion from valueType to targetType is already registered, then it will be overwritten.
+        /// </summary>
+        /// <param name="valueType">Type of original value.</param>
+        /// <param name="targetType">Converted value type. </param>
+        /// <param name="conversion">Conversion delegate; <c>null</c> for unregister already registered conversion.</param>
         public static void Register(Type valueType, Type targetType, Func<object, object> conversion)
         {
-            if (conversion == null)
-                throw new ArgumentNullException(nameof(conversion));
-
             if (!_valueConversions.TryGetValue(valueType, out var conversions))
                 _valueConversions.Add(valueType, conversions = new Dictionary<Type, Func<object, object>>());
 
-            conversions[targetType] = conversion;
+            if (conversion == null)
+                conversions.Remove(targetType);
+            else
+                conversions[targetType] = conversion;
         }
     }
 }
--- a/src/GraphQL/Conversion/ValueConverter.cs
+++ b/src/GraphQL/Conversion/ValueConverter.cs
@@ -121,28 +121,61 @@ namespace GraphQL
             Register(typeof(char), typeof(int), value => Convert.ToInt32((char)value));
         }
 
-        public static object ConvertTo(object value, Type targetType)
-        {
-            if (value == null || targetType.IsInstanceOfType(value))
-                return value;
-
-            var conversion = GetConversion(value.GetType(), targetType);
-            return conversion(value);
-        }
-
         public static T ConvertTo<T>(object value)
         {
-            var v = ConvertTo(value, typeof(T));
+            object v = ConvertTo(value, typeof(T));
 
             return v == null ? default : (T)v;
         }
 
+        public static object ConvertTo(object value, Type targetType)
+        {
+            if (!TryConvertTo(value, targetType, out object result))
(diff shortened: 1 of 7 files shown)


src/GraphQL/ObjectExtensions.cs
```diff
@@ -103,6 +103,17 @@ public static object ToObject(this IDictionary<string, object> source, Type type
         /// <remarks>There is special handling for strings, IEnumerable&lt;T&gt;, Nullable&lt;T&gt;, and Enum.</remarks>
         public static object GetPropertyValue(this object propertyValue, Type fieldType, IGraphType mappedType = null)
         {
+            bool SkipEnumerable()
+            {
+                if (fieldType == typeof(string))
+                    return true;
+
+                if (fieldType == typeof(byte[]) && propertyValue is string)
```
sungam3r: I think this can be safely considered a special case.

src/GraphQL/Conversion/ValueConverter.cs
```diff
@@ -114,6 +114,11 @@ static ValueConverter()
             Register(typeof(double), typeof(decimal), value => Convert.ToDecimal(value, NumberFormatInfo.InvariantInfo));
 
             Register(typeof(string), typeof(Uri), value => value is string s ? new Uri(s) : (Uri)value);
+            // registering such a default conversion for string->byte[] seems useful
+            Register(typeof(string), typeof(byte[]), value => Convert.FromBase64String((string)value));
```
sungam3r: @Shane32 This is worth noting in the documentation.

src/GraphQL/ObjectExtensions.cs
sungam3r: Good idea.

sungam3r: If we go further, we can make the assertion that `IEnumerable` has nothing to do with it, we can simply move `ValueConverter.ConvertTo(value, fieldType);`/lookup to the beginning of the method.

sungam3r: Done.

src/GraphQL/Conversion/ValueConverter.cs
```diff
@@ -1,14 +1,14 @@
...
 using System.Globalization;
 using System.Numerics;
 
 namespace GraphQL
 {
     public static class ValueConverter
     {
-        private static readonly Dictionary<Type, Dictionary<Type, Func<object, object>>> _valueConversions
-            = new Dictionary<Type, Dictionary<Type, Func<object, object>>>();
+        private static readonly ConcurrentDictionary<Type, ConcurrentDictionary<Type, Func<object, object>>> _valueConversions
```
sungam3r: Accidentally noticed that this static field was still not thread safe. All other static dictionaries are `ConcurrentDictionary<>`.

src/GraphQL/Conversion/ValueConverter.cs
Shane32: Probably should be noted in `schema-types.md` as well as in the upgrade notes; would you agree?

sungam3r: yes

sungam3r: There will be another change with value conversion soon.

Shane32:
I didn't read all the code in depth, but can you help me with a very quick overview of this here and be sure I understand this correctly?  It will help me write better documentation.

- When a input variable comes in via the variables, it gets converted by the proper ScalarGraphType to a .Net type, and then gets converted by ValueConverter to the type requested by GetArgument

- When an parameter comes in via the query (as a literal), it gets converted by the proper ScalarGraphType to a .Net type, and then gets converted by ValueConverter to the type requested by GetArgument

- When serializing response data, the .Net type gets passed directly to the proper ScalarGraphType, ignoring the ValueConverter, which handles the serialization?

- Or, when serializing response data, the ValueConverter always makes a first pass, before feeding the ScalarGraphType?  If this is true how does the ValueConverter need to know what graph type to prep for?

Shane32:
And
- When GetArgument does not request a particular type, it skips the ValueConverter (I assume)

Shane32:
If you can give me some quick pointers, it will probably save me a while of digging through the code to figure out how it all fits together - thanks @sungam3r

sungam3r:
There will be no quick answer. I made these changes mostly for `GetArgument<T>`. I understand that `ValueConverter` is used not only for this, but has not yet verified all usages. I will come back with answers later.

Shane32:
Ok thanks; I don't have time to dig into it now.  If you post what you learn, it will help when I do have time.  Thanks!

sungam3r:
Relates to #1616

sungam3r:
@Shane32 Regarding your questions.

1. When a input variable comes in via the variables, it gets converted by the proper `ScalarGraphType` to a .Net type, and then gets converted by `ValueConverter` to the type requested by `GetArgument`.

Yes.

2. When an parameter comes in via the query (as a literal), it gets converted by the proper `ScalarGraphType` to a .Net type, and then gets converted by `ValueConverter` to the type requested by `GetArgument`.

Yes.

3. When serializing response data, the .Net type gets passed directly to the proper `ScalarGraphType`, ignoring the `ValueConverter`, which handles the serialization?

No. Most `Serialize` methods delegate call to `ParseValue` which calls `ValueConverter.ConvertTo`. See https://github.com/graphql-dotnet/graphql-dotnet/blob/master/docs2/site/docs/getting-started/custom-scalars.md

4. Or, when serializing response data, the `ValueConverter` always makes a first pass, before feeding the `ScalarGraphType`? If this is true, does the `ValueConverter` need to know what graph type to prep for?

No. `Serialize` is used only in `ValueExecutionNode.ToValue()`:
```c#
  public override object ToValue()
        {
            if (Result == null)
                return null;

            var scalarType = GraphType as ScalarGraphType;
            return scalarType?.Serialize(Result);
        }
```
No. `ValueConverter` knows nothing about GraphType, it works with registered .NET -> .NET conversions.

5. When `GetArgument` does not request a particular type, it skips the `ValueConverter` (I assume).

Yes.
```c#
 if (argumentType == typeof(object))
                {
                    result = arg;
                    return true;
                }
```

PR #1578, opened 2020-02-27
[Discussion] Change interfaces to read-only

Shane32:
See #1572 

This is testing the changes required to have the entire schema represented both in read/write classes started with `Schema` (and `GraphType` and `FieldType`, etc), and read-only interfaces via `ISchema` (and `IGraphType`, `IFieldType`, etc).  Execution strategies and similar classes should only need access to the interfaces.  `Schema` initialization support code, such as middleware building and `GraphTypesLookup` would all be based on the classes.

Changing `FieldType` to `IFieldType` was easy.  Changing `IGraphType` to `GraphType` is proving to be quite a challenge.

The `QueryArguments` was rewritten to inherit from `List<QueryArgument>` and thereby be able to be cast to `IEnumerable<IQueryArgument>`.  Since the class also implements `ICollection<QueryArgument>`, the .Net `Enumerable.Count()` extension method will use short-circuit evaluation to return the number of elements in the collection.  So we can safely use `.Count()` at will with no ill effects.  I'll have to check the MS source, but `.Any()` might even be a better choice, just in case somehow it isn't a `List<T>`.

`ExecutionOptions.SchemaFilter` and `ExecutionOptions.FieldNameConverter` have been moved to the schema.  They belong either in `Schema`, or in both `ExecutionOptions` and `ExecutionContext` -- but the `ExecutionOptions` should not modify the schema properties.

With each commit, I've verified that all tests pass.  So you can check the code changes and API diff at each step.

I'll add additional notes here as I play with this code.

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
@@ -79,12 +79,10 @@ namespace GraphQL
         public GraphQL.Inputs Inputs { get; set; }
         public System.Collections.Generic.List<GraphQL.Execution.IDocumentExecutionListener> Listeners { get; }
         public int? MaxParallelExecutionCount { get; set; }
-        public GraphQL.Conversion.INameConverter NameConverter { get; set; }
         public string OperationName { get; set; }
         public string Query { get; set; }
         public object Root { get; set; }
         public GraphQL.Types.ISchema Schema { get; set; }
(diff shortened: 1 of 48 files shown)


Shane32:
@sungam3r should I keep working on this in my spare time?  (if any, lol)

sungam3r:
On the one hand I want to answer _yes_. On the other hand, I would prefer to work with any other PR which solves one of 146 open issues. These changes are quite global and I feel that they are worth doing only when we are 100% sure of their result. Now I'm not even 50% sure.

Shane32:
@sungam3r  Sounds like you're feeling we should close this topic for now.  Fine with me.

sungam3r:
OK

PR #1582, opened 2020-02-27
Add object converter for GetArgument

### LINKED ISSUE #669: Handling Map Types (looking for a better solution than Serialisation)
Greetings.

I am looking for a way to handle an input map types myself as I know that they are not supported out of the box in the GraphQL as I need to send a collection of Key Value pairs over for an arbitrary set. However, I running into some trouble with defining my `InputObjectGraphType` to handle the conversion.

Here's what I have (names simplified and bits cut out for brevity):

The query I am using to send over:
```
{
  "create": {
    "name": "name",
    "set": "DOMAIN",
    "stats": [
      {"key": "hello", "value": "1"}
    ]
  }
}
```
And the Class I want it to be in when all is said and done:
```
public class CreateCharacter
{
    public string Set { get; set; }
    public IEnumerable<KeyValuePair<string, int>> Stats { get; set; }
    public IEnumerable<KeyValuePair<string, string>> MetaData { get; set; }
}
```
I know that these two ultimately work because I can do either of the following in my mutation (where I desire this input) and get the `context.Arguments` onto my object (ignoring the fact it doesn't loop for the sake of this example)
```
var step1 = context.Arguments["create"];
var step2 = (step1 as Dictionary<string, object>)["stats"];
var step3 = (step2 as object[])[0];
var step4 = (step3 as Dictionary<string, object>);
var keyValues = new List<KeyValuePair<string, int>>();
keyValues.Add(new KeyValuePair<string, int>(step4["key"].ToString(), int.Parse(step4["value"].ToString())));
var create = new Create { Stats =  keyValues };
```
or
```
var createCharacter = JsonConvert.DeserializeObject<Create>(JsonConvert.SerializeObject(context.Arguments["create"]));
```
The GraphiQL style ui also lets me enter this without issue so the schema definition works and  I am reasonable confident I am sending over data that matches what I want.
Where if's falling down is in trying to set up my argument type so I can leverage `context.GetArgument<CreateType>("create")` so I don't have to keep writing long conversion chains or serialise and unserialise in each query or mutation.

Currently, my input object graph look like this:
```
CreateType : InputObjectGraphType<Create>...
public CreaterType()
{
    Name = "Character";
    Field(x => x.Name);
    Field(x => x.Set, type: typeof(EnumerationGraphType<Set>));
    Field<ListGraphType<InputMapType>>(nameof(Create.Stats), resolve: context => {
        var step1 = context.Arguments["createCharacter"];
        var step2 = (step1 as Dictionary<string, object>)["stats"];
        var step3 = (step2 as object[])[0];
        var step4 = (step3 as Dictionary<string, object>);
        var keyValues = new List<KeyValuePair<string, int>>();
        keyValues.Add(new KeyValuePair<string, int>(step4["key"].ToString(), int.Parse(step4["value"].ToString())));
        new CreateCharacter { Stats = keyValues };
    });
}

public class InputMapType : InputObjectGraphType
{
    public InputMapType()
    {
        Name = "InputMap";
        Field<StringGraphType>("key");
        Field<StringGraphType>("value");
    }
}
```
At the moment: I never hit this code so debugging it isn't a ready option. If I just try and call  `context.GetArgument<Create>("create")` I get a nice fat Argumentation that is typically vague (though included below anyway for reference). Any idea what is going on here and how I can define the input type to do this conversion for me as I don't really know what it's trying to do when it fails so begin trying to fix it.

Thank you.

### LINKED ISSUE #684: Map input fields.
Hi, 

I have the following use case:

My objects are nested dictionaries (JObjects). But the keys are not always valid for GraphQL,e g. the contain dashes or underscores.

But of course I can easily map them:

	string key; // From internal schema definition.
	
	var resolver = new FuncFieldResolver<object>(c =>
	{
		if (((JObject)c.Source).TryGetValue(key, out var value))
		{
			return fieldInfo.Resolver(value, c);
		}
		else
		{
			return fieldInfo;
		}
	});

	AddField(new FieldType
	{
		Name = key.ToCamelCase(),
		Resolver = resolver,
		ResolvedType = fieldInfo.ResolveType,
	});

But I cannot do something similar for the input types, because I just get a nested IDictionary<string, object>. Is there a solution I don't see, except than converting the final dictionary? It would be great to have something like a reverse resolver so that I can also skip the generation of the IDictionary and create the objects manually or set the properties.

### LINKED ISSUE #864: How to map input object to complex object
## Summary
I would like to know how to map the following mutation (PersonInput) to the models Person having an Account, using GraphQL.NET 2.0

```
mutation createPerson ($person: PersonInput!){
  createPerson(person: $person)
  {
	id
  }
}
```

```
{
 "person": {
	"firstName": "Sjaak",
	"lastName": "Trekhaak",
	"email": "jan@test.com",
	"password": "sjanie"
  }
}
```

```
public class PersonInputObject : InputObjectGraphType
{
	public RentePersonInputObject InputObject()
	{
		this.Name = "PersonInput";
		
		this.Field(typeof(StringGraphType), "firstName");
		
		this.Field(typeof(StringGraphType), "lastName");
		
		this.Field(typeof(StringGraphType), "password");
		
		this.Field(typeof(StringGraphType), "email");
	}
}

public class Person
{
	public int Id { get; set; }
	public string FirstName { get; set; }
	public string LastName { get; set; }

	public Account Account { get; set; }
}

public class Account
{
	public string Email { get; set; }
	public string Password { get; set; }
}
```

This is my mutation object.

```
public MutationObject()
{
	this.Name = "Mutation";
	this.Description = "The mutation type, represents all updates we can make to our data.";

        

	this.FieldAsync<PersonObject, Person>(
		"createPerson",
		"Create a person.",
			new QueryArgument<NonNullGraphType<PersonInputObject>>()
			{
				Name = "person",
				Description = "The person you want to create.",
			}),
		resolve: context =>
		{
			var person= context.GetArgument<Person>("person");
// dostuff
			
			return person;
		}
	);
}
```

Or should I resolve everything by hand in the mutation's resolver?

Any examples available?

sungam3r:
fixes #669
fixes #684 
fixes #864


benmccallum:
This argument grabbing stuff confuses me. Isn't this already supported? I'm certainly doing it, albeit via `GetValidatedArgumentAsync<TObject>` from GraphQL.Validation. I even added tests in GraphQL.Validation to prove it works. 

What functionality is this PR adding? I'm interested!

**Mutation field def**
 ```
FieldAsync<CreateSupplierHolidaysOutput, CreateSupplierHolidaysOutputDto>(
    "createSupplierHolidays",
    "Creates holidays for a supplier.",
        new QueryArgument<NonNullGraphType<CreateSupplierHolidaysInput>>
        {
            Name = ArgumentNames.Input,
            Description = "The new supplier holidays desired."
        }
    ),
    createSupplierHolidaysResolver.ResolveAsync)
    .AuthorizeWith(PolicyNames.RequireSupplierDataAccess);
```

**Input graph types (an object which itself has a list of another object)**
```
    public class CreateSupplierHolidaysInput : InputObjectGraphType<CreateSupplierHolidaysInputDto>
    {
        public CreateSupplierHolidaysInput()
        {
            Field<NonNullGraphType<IdGraphType>>(
                nameof(CreateSupplierHolidaysInputDto.SupplierId));

            Field<NonNullGraphType<ListGraphType<NonNullGraphType<NewSupplierHolidayInput>>>>(
                nameof(CreateSupplierHolidaysInputDto.NewHolidays));
        }
    }

    public class NewSupplierHolidayInput : InputObjectGraphType<NewSupplierHolidayInputDto>
    {
        public NewSupplierHolidayInput()
        {
            Field<NonNullGraphType<DateGraphType>>(nameof(NewSupplierHolidayInputDto.StartsAt).ToCamelCase());
            Field<NonNullGraphType<DateGraphType>>(nameof(NewSupplierHolidayInputDto.EndsAt).ToCamelCase());
        }
    }
```

**Input POCO types**
```
    public class CreateSupplierHolidaysInputDto
    {
        public IntGlobalId SupplierId { get; set; }

        public List<NewSupplierHolidayInputDto> NewHolidays { get; set; } = new List<NewSupplierHolidayInputDto>();
    }

    public class NewSupplierHolidayInputDto
    {
        public DateTime StartsAt { get; set; }

        public DateTime EndsAt { get; set; }
    }
```

**Resolver grabbing arg**
```
var input = await context.GetValidatedArgumentAsync<CreateSupplierHolidaysInputDto>(ArgumentNames.Input);
```

**Comment by sungam3r** [comment]
It allows you to register your own delegate for creating DTO from dictionary. PR is still in progress.

**Comment by benmccallum** [comment]
Cool; would be keen to see a use case. Add me as a reviewer when you get some tests up.

**Comment by sungam3r** [comment]
See linked issues https://github.com/graphql-dotnet/graphql-dotnet/issues/684#issuecomment-413537001

**Comment by sungam3r** [comment]
@benmccallum done

File: src/GraphQL/ObjectExtensions.cs [hunk 1]
```diff
@@ -36,6 +34,13 @@ public static T ToObject<T>(this IDictionary<string, object> source)
         /// </param>
         public static object ToObject(this IDictionary<string, object> source, Type type, IGraphType mappedType = null)
         {
+            if (source == null)
```
sungam3r: I have doubts about the need for this check.

src/GraphQL/ObjectExtensions.cs
```diff
@@ -16,10 +16,8 @@ public static class ObjectExtensions
         /// <param name="source">The source of values.</param>
         /// <returns>T.</returns>
         public static T ToObject<T>(this IDictionary<string, object> source)
-            where T : class, new()
```
sungam3r: `new()` is not needed anymore since `ToObject` supports parameterized ctors.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
```diff
@@ -338,8 +338,9 @@ namespace GraphQL
         public static object ConvertTo(object value, System.Type targetType) { }
         public static T ConvertTo<T>(object value) { }
         public static void Register(System.Type valueType, System.Type targetType, System.Func<object, object> conversion) { }
-        public static bool TryConvertTo(object value, System.Type targetType, out object result) { }
-        public static bool TryConvertTo<T>(object value, out T result) { }
```
sungam3r: No need for these APIs to be public (yet). Not big problem to remove because these APIs were introduced in previous commit.

sungam3r:
@Shane32 I added some docs to value conversion classes.

Shane32:
Just took an in-depth look.  In general, I have no specific comments about this particular PR.  Just a few thoughts about the `ValueConverter` and `ObjectExtensions.ToObject` in general:

`ToObject` scan constructors for parameters, but it doesn't seem to translate names based on the original expression name.  Also, for consistency, it does not ignore case.  I'd consider this a bug.

The `ValueConverter` uses a `ConcurrentDictionary<Type, ConcurrentDictionary<Type, Func<object, object>>>`.  This should definitely operate faster as a `ConcurrentDictionary<(Type, Type), Func<object, object>>`, and would clean up some of the code.  It's also on the 'hot path' as it is executed by most scalar graph types when serializing --- but, generally is short-circuited by `if (value == null || targetType.IsInstanceOfType(value))`.  So my gut feeling is that benchmarks won't show much difference.

Preregistering more conversions should not negatively affect user code.

Allowing custom conversions to be registered with the `ValueConverter` is fine, although in theory different schemas could use different value converters.  I still wouldn't change the design, however, even if we make `ISchema` read-only (along with `IGraphType` and so on).

sungam3r:
Thanks. I agree about original expression name and ignore case. I also don’t like that `ValueConverter` is static. I wanted to change this, but I won’t be able to do this without an significant design changes.

sungam3r:
@Shane32 `ToObject` is ready. The code has become more complicated, but in 99% of cases there will be no performance loss. I doubt anyone is using parameterized constructors for arguments widely.

Shane32:
Looks good

src/GraphQL/ObjectExtensions.cs
sungam3r: In any case, it won’t get any worse, since before this check it would be NRE later

PR #1535, opened 2020-02-13
Remove ReadonlyResolveFieldContext

Shane32:
This PR makes ExecutionNode implement IResolveFieldContext, so an instance of ReadonlyResolveFieldContext is not required, saving an object allocation per resolver.  Doesn't really enhance performance, although the number of object allocations decrease notably as can be seen in the GC stats below.  Memory doesn't change much due to the previous enhancement of the ReadonlyResolveFieldContext over the prior ResolveFieldContext class.

Just a thought - ok to merge or close.

```
// * Summary *

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18362
Intel Core i7-7740X CPU 4.30GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT


*** BEFORE ***

|        Method |        Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |------------:|----------:|----------:|---------:|--------:|------:|----------:|
| Introspection | 1,503.10 us | 19.721 us | 17.483 us | 189.4531 | 60.5469 |     - | 835.56 KB |
|          Hero |    34.32 us |  0.639 us |  0.566 us |   6.1035 |  0.0610 |     - |   25.1 KB |

*** AFTER ***

|        Method |        Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |------------:|----------:|----------:|---------:|--------:|------:|----------:|
| Introspection | 1,496.35 us | 19.890 us | 18.605 us | 171.8750 | 48.8281 |     - | 807.63 KB |
|          Hero |    34.44 us |  0.597 us |  0.558 us |   6.1035 |       - |     - |  25.05 KB |
```

PR #1511, opened 2020-01-20
DataLoader rework for serial execution and fixing deadlocks

Shane32:
This PR changes the DataLoader to return an `IDataLoaderResult<T>` rather than a `Task<T>` in the field resolver.  The `ExecutionStrategy` then monitors for `IDataLoaderResult` values and queues them to execute at the appropriate time.  Executing `IDataLoaderResult<T>.GetResultAsync()` retrieves a `Task<T>` which contains the resolved value once the loader has completed its execution.

This overcomes the limitations of the DataLoader as outlined in #1508 as follows:
- DataLoaders do not batch load data when run with a SerialExecutionStrategy
- Field resolve functions that await before returning the DataLoaders's LoadAsync function result in deadlock.
- DataLoaders cannot be chained together

Relevant issues listed here:
- #1191 DataLoader and SerialExecutionStrategy not bulking
- #945 DataLoader deadlock with multiple awaits (tagged to be fixed in GraphQL 3.0)
- #837 Should the DataLoader work inside a Mutation
- #1042 How to handle multiple data loader calls?

This revised code is also immune to deadlocks when used with federation resolvers -- see here:
- #1565 Problem with Data loaders in federation resolvers

This revised code does not suffer from "exponential" delays when return large sets of data -- see here:
- #1710 DataLoader gets exponentially slower when number of items increase

It also adds support for MaxBatchSize as requested here:
- #954 Add MaxBatchSize to BatchDataLoader/CollectionDataLoader

I've been using the guts of this code for a while now as it's buried in #1376 along with a lot of other stuff.  However, I've reworked it to keep the naming and so on as close as possible to the current DataLoader.  It should be pretty rock solid, but there certainly are optimizations possible.   There's comments in it, but I didn't make a final pass through it.

Some things to like:
- DataLoaders work with serial execution strategy (including mutations)
- DataLoaders can work in conjunction with async/await operations before or after the data is loaded
- DataLoaders can be chained together
- All existing DataLoaderContext infrastructure has not changed at all
- Impossible to deadlock, as attempting to resolve a `IDataLoaderResult<T>` will trigger the `DispatchAsync` function
- Minimal connection between ExecutionStrategy and DataLoader -- only a single interface `IDataLoaderResult` containing a single method `GetResultAsync()`

Some things not to like:
- Chaining DataLoaders is not intuitive
- Chained DataLoaders return `IDataLoaderResult<IDataLoaderResult<T>>` types
- `Then` pattern for chaining DataLoaders
[...]

PR #1584, opened 2020-03-02
[WIP]  Add pool for ReadonlyResolveFieldContext

sungam3r:
Slight optimization over `ReadonlyResolveFieldContext`. No breaking changes, no public api changes, only internals. GraphQL package gets new dependency from `Microsoft.Extensions.ObjectPool` (netstandard2.0). I find it acceptable for internal things. One could do the cache himself, but why if there is a ready-made solution?

Before
``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.17134.1304 (1803/April2018Update/Redstone4)
Frequency=2343753 Hz, Resolution=426.6661 ns, Timer=TSC
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT


```
| Introspection | 2,607.83 us | 26.774 us | 23.735 us | 195.3125 | 54.6875 |     - | 486.71 KB |
|          Hero |    57.52 us |  0.308 us |  0.257 us |   8.8501 |       - |     - |  13.61 KB |

After
``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.17134.1304 (1803/April2018Update/Redstone4)
Frequency=2343753 Hz, Resolution=426.6661 ns, Timer=TSC
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X86 RyuJIT


```
| Introspection | 2,643.91 us | 14.938 us | 13.973 us | 171.8750 | 50.7813 |     - | 454.89 KB |
|          Hero |    58.63 us |  0.341 us |  0.319 us |   8.7891 |       - |     - |  13.55 KB |

~7% memory decrease on introspection test. There is some slowdown but it is not so significant.

@Shane32 Perhaps you will find the use of the pool in other places.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -9,14 +9,16 @@ using GraphQL.Language.AST;
 using GraphQL.Types;
 using GraphQL.Validation;
 using GraphQL.Validation.Complexity;
+using Microsoft.Extensions.ObjectPool;
 using static GraphQL.Execution.ExecutionHelper;
 using ExecutionContext = GraphQL.Execution.ExecutionContext;
 
 namespace GraphQL
 {
-
     public class DocumentExecuter : IDocumentExecuter
     {
+        private static readonly DefaultObjectPool<ReadonlyResolveFieldContext> _pool = new DefaultObjectPool<ReadonlyResolveFieldContext>(new ReadonlyResolveFieldContextPolicy());
+
         private readonly IDocumentBuilder _documentBuilder;
         private readonly IDocumentValidator _documentValidator;
         private readonly IComplexityAnalyzer _complexityAnalyzer;
@@ -247,6 +249,7 @@ namespace GraphQL
         {
             var context = new ExecutionContext
             {
+                Pool = _pool,
                 Document = document,
                 Schema = schema,
                 RootValue = root,
--- a/src/GraphQL/Execution/ExecutionContext.cs
+++ b/src/GraphQL/Execution/ExecutionContext.cs
@@ -1,15 +1,17 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
 using GraphQL.Instrumentation;
 using GraphQL.Language.AST;
 using GraphQL.Types;
+using Microsoft.Extensions.ObjectPool;
 
 namespace GraphQL.Execution
 {
     public class ExecutionContext : IProvideUserContext
     {
+        internal ObjectPool<ReadonlyResolveFieldContext> Pool { get; set; }
+
         public Document Document { get; set; }
 
         public ISchema Schema { get; set; }
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -175,11 +175,12 @@ namespace GraphQL.Execution
             if (node.IsResultSet)
                 return;
 
-            IResolveFieldContext resolveContext = null;
+            ReadonlyResolveFieldContext resolveContext = null;
+            var pool = context.Pool;
 
             try
             {
-                resolveContext = new ReadonlyResolveFieldContext(node, context);
+                resolveContext = pool == null ? new ReadonlyResolveFieldContext(node, context) : pool.Get().Initialize(node, context);
 
                 var resolver = node.FieldDefinition.Resolver ?? NameFieldResolver.Instance;
                 var result = resolver.Resolve(resolveContext);
@@ -234,6 +235,11 @@ namespace GraphQL.Execution
                 context.Errors.Add(error);
 
                 node.Result = null;
+            }
+            finally
+            {
+                if (pool != null && resolveContext != null)
+                    pool.Return(resolveContext);
             }
         }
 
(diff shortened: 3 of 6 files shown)


Shane32:
> No breaking changes, no public api changes, only internals.

The context is exposed outside of the internals of the project.  The execution strategy can only assume it has been discarded, so it would be false to say that it was not a breaking change.  For instance, if someone utilized an UnhandledExceptionDelegate to store references to the context of failed resolvers, saving values from them in the database via a DocumentListener after execution was complete, the contexts would contain incorrect values.

Interestingly enough, some of my own code in production would break (in two different spots) if the ObjectPool was implemented in this manner.

> I find it acceptable for internal things. 

I would agree, and suggest that we only utilize the ObjectPool where the object is not exposed, directly or indirectly, to the user.  This is clearly not one of those scenarios.

> GraphQL package gets new dependency

Personally, I do not have a problem with a dependency, so long as it does not cause incompatibility with prior versions of .Net.  Note that this package is not part of .Net Standard 2.0, but rather is a part of Asp.Net Core.

According to Microsoft: 
> Microsoft.Extensions.ObjectPool is part of the ASP.NET Core infrastructure
> https://docs.microsoft.com/en-us/aspnet/core/performance/objectpool?view=aspnetcore-3.1

However, it only has a single dependency, and that is .Net Standard 2.0.  We should at least verify that applications that rely on .Net Core 2.2 (like mine) should be able to safely upgrade that one nuget package without breaking compatibility with the rest of their application.  I could probably test this if necessary.

Personally, I'd import the ObjectPool version 2.0 nuget package and then allow compatibility with >= 2.0 so it works without changing a user's dependency tree, at least as far back as Asp.Net Core 2.0.  If a person wanted the latest 3.1.1 version, they can always import that (if they do not already have it imported automatically when they import Asp.Net Core 3.1).

I'm also confused why you don't seem to mind this dependency, whereas you've called other dependencies "leaky", even when they are only abstractions and contain no code.  Again, I do not mind dependencies so long as they work properly with older versions of Asp.Net Core.  (It's a very big problem when one nuget package requires the very latest dependencies of everything for no good reason.)

> ~7% memory decrease on introspection test. There is some slowdown but it is not so significant.

For the sake of discussion, let's compare this to #1535 .  With the latter, there is no performance drop, and has the same GC drop.  Memory usage drops 4% instead of 7%.  There is no negative side effects.  It would even be tough to call #1535 "breaking" if it wasn't for the fact that ReadonlyResolveFieldContext is public, added in the last preview build.  By simply not removing the class, it wouldn't even be breaking.

I could not justify adding a dependency and adding known negative side effects for a minor memory gain (and performance loss), when a second alternative has a similar effect without the negative consequences.  Even if there was no alternative, I would not pull it in because of its unique requirements on user code (namely, that the context is only valid during the execution of the resolver, middleware, error handlers, etc).

Also: my data loader PR #1511 will not work with the ObjectPool at all.  I would need to rewrite ExecuteNodeTreeAsync and ExecuteNodeAsync to implement the ObjectPool in that PR, if the ObjectPool was deemed to be a good choice for the project -- if it could be utilized at all in an effective manner.

Let's look for another use for the ObjectPool.

sungam3r:
>  For instance, if someone utilized an UnhandledExceptionDelegate to store references to the context of failed resolvers

I would say that this should not be done regardless of my PR. Anyway inside the context there may be references to already destroyed objects. I am sure we should note it in the documentation. 

> Note that this package is not part of .Net Standard 2.0

So what? Netstandard2.0 targeted.

> applications that rely on .Net Core 2.2 (like mine) should be able to safely upgrade that one nuget package without breaking compatibility with the rest of their application.

https://www.fuget.org/packages/Microsoft.Extensions.ObjectPool/3.1.1/lib/netstandard2.0/diff/2.2.0/
No breaking changes, also I can downgrade to 2.2 or even lower. It doesn’t matter to me.

> I'm also confused why you don't seem to mind this dependency, whereas you've called other dependencies "leaky"

Because this dependency is entirely internal, it by definition does not leak anywhere in user code - no interfaces to implement, no classes to use.

> let's compare this to #1535

As an alternative, I agree. Waiting for detailed review.

> the context is only valid during the execution of the resolver, middleware, error handlers, etc

Still, I would mention this in the documentation. The context (in a broad sense) contains some information for the duration of some piece of code (algorithm section). The correctness of the context outside this section is not guaranteed, although sometimes, by a fluke, the context is not destroyed. I can give `HttpContext` as an example. What will happen if you save it and use it somehow further? Probably nothing good.

About DataLoader - sorry, so far this topic is beyond my knowledge.

BenjaBobs:
> Anyway inside the context there may be references to already destroyed objects. I am sure we should note it in the documentation.
Doesn't hurt, and I think it would add value. I imagine this is one of the cases where if someone tries this it could take them a while to figure out what's happening.

Shane32:
@sungam3r do you plan to revisit this PR?  I suggest closing it due to (1) complications with #1511, (2) additional requirements on user code, and (3) minimal "benefit".  #1535 does not carry the added requirements or complications, and we have already dismissed that PR as unnecessary.

sungam3r:
I would like to keep it for now. I'll take another look and close myself.

sungam3r:
This PR doesn't make sense after #2188

PR #1589, opened 2020-03-05
Remove SchemaFilter and NameConverter from ExecutionOptions

Shane32:
As a step towards making `ISchema` read-only, this PR removes `SchemaFilter` and `NameConverter` from `ExecutionOptions`, which used to simply set the corresponding properties on the schema.  Now the schema's constructor should typically be used to set these properties.

Alternatively, we could move `Filter` and `NameConverter` from `ISchema` to `ExecutionContext`, although this may prove to be difficult to implement.


sungam3r:
How to set these properties to the calling code now?

https://github.com/graphql-dotnet/server/blob/develop/src/Core/Internal/DefaultGraphQLExecuter.cs#L69
https://github.com/graphql-dotnet/server/blob/develop/src/Core/Internal/DefaultGraphQLExecuter.cs#L71

Shane32:
It should be set in the schema constructor.  I'll update the docs in this PR.  You could of course pull `Schema` from DI and set the properties manually before execution, but the point of this PR is that it's actually improper to do so, as modifying those properties could (in theory) change the execution of another schema executing in parallel.

Shane32:
Actually the docs say nothing about either field, except within the 1.x to 2.x migration guide, which already indicates that the `FieldNameConverter` should be set on the schema.

Of course these changes would be added to the 3.0 migration guide.

Shane32:
Just for reference, the equivalent code (but equally as "wrong" as before) would be:
```
            var schema = Schema;
            schema.SchemaFilter = _options.SchemaFilter ?? new DefaultSchemaFilter();
            schema.NameConverter = _options.NameConverter ?? CamelCaseNameConverter.Instance;
            {
                Schema = schema,
                OperationName = operationName,
                Query = query,
                Inputs = variables,
                UserContext = context,
                CancellationToken = cancellationToken,
                ComplexityConfiguration = _options.ComplexityConfiguration,
                EnableMetrics = _options.EnableMetrics,
                ExposeExceptions = _options.ExposeExceptions,
                UnhandledExceptionDelegate = _options.UnhandledExceptionDelegate,
            };
```

sungam3r:
`ExecutionOptions` contains many other properties. Why did you choose only these?

Shane32:
They are the only ones that modify the schema.  The other options are stored within `ExecutionContext`.

Shane32:
Well, the FieldMiddleware also modifies the schema.  But it only does so once ever, and then the schema is locked.....  Thoughts?

Shane32:
Come to think of it, I don't think there are any locks while initializing the middleware or schema.  It is entirely possible that the middleware could be applied twice if two DocumentExecuters started executing simultaneously on an uninitialized schema.

Shane32:
If we can get to a point where ISchema is read-only, it would help prevent these types of issues from unintentionally getting into the codebase.

Shane32:
@sungam3r I really feel that these particular options belong on the schema, as they change and define the schema versus the execution.  Middleware falls in the same boat.  However, it's not going to make any practical different to users if you'd rather not make these changes.  Feel free to close this PR if you wish.

sungam3r:
I have now looked again at the code. I don't like the fact that every time we execute a request, we create a new ExecutionOptions instance, set the schema to it, and then inside ExecuteAsync we fill the schema properties with the properties from the options. This in itself causes some confusion. Second, there are potential problems with the singleton schema. Perhaps more serious refactoring is needed, not just the elimination of a couple of properties. I don't know how to do it yet.

Shane32:
Well, this would be a start....

sungam3r:
I would also like to point out that any get/set property on the `Schema` potentially leads to a problem when using the schema as a singleton if we one way or another expose schema via `ExecutionOptions`, where caller may set different values each request. It is possible that it simply does not make sense to protect yourself from such problems, this is a fight against windmills.

Shane32:
I agree, which is why a read-only schema really should be best.  It just changes a whole lot of code.  In this particular instance, I can set `ISchema.NameConverter` and `ISchema.Filter` to be read-only; but there's still a bunch of other read/write properties on the schema.

joemcbride (approved):

Shane32:
Let’s merge this into a new develop branch for 4.0 changes, as this would be a breaking change for 3.x

sungam3r:
I will review this again next days.

sungam3r (approved):
Let's see how it turns out.

PR #1590, opened 2020-03-05
Change QueryArguments to IList and IEnumerable

Shane32:
In preparation for changing interfaces to read-only, this PR changes `QueryArguments` to inherit from `List<QueryArgument>` and changes the `FieldType.Arguments` property to an `IList<QueryArgument>`.  This provides a few benefits:
1. `Field()` methods and similar can accept an array directly (yay!)
2. The `QueryArguments` can be cast directly to `IEnumerable<QueryArgument>`, so when #1572 is pulled, `IFieldType.Arguments` will return the list as an `IEnumerable<QueryArgument>`
3. `Enumerable` extension methods are all optimized with short-circuit evaluation, so `.Any()`, `.Count()` and so on do not execute an enumerator.
4. Since the `List` class is not necessary, there is one less object initialization when adding a value.

Also note that `List` already contains code to leave the underlying array null until at least one value is added.

I also added `.None()` as an internal optimization -- see the code for notes.

As an alternative, `IReadOnlyCollection<QueryArgument>` could be used in place of `IEnumerable<QueryArgument>`, which provides `.Count` as a property.  However, that interface is so infrequently used that I just stuck to `IEnumerable<QueryArgument>` with some short-circuit evaluation where needed.

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
@@ -597,7 +597,7 @@ namespace GraphQL.Execution
         public static object CoerceValue(GraphQL.Types.ISchema schema, GraphQL.Types.IGraphType type, GraphQL.Language.AST.IValue input, GraphQL.Language.AST.Variables variables = null) { }
         public static System.Collections.Generic.Dictionary<string, GraphQL.Language.AST.Field> CollectFields(GraphQL.Execution.ExecutionContext context, GraphQL.Types.IGraphType specificType, GraphQL.Language.AST.SelectionSet selectionSet) { }
         public static bool DoesFragmentConditionMatch(GraphQL.Execution.ExecutionContext context, string fragmentName, GraphQL.Types.IGraphType type) { }
-        public static System.Collections.Generic.Dictionary<string, object> GetArgumentValues(GraphQL.Types.ISchema schema, GraphQL.Types.QueryArguments definitionArguments, GraphQL.Language.AST.Arguments astArguments, GraphQL.Language.AST.Variables variables) { }
+        public static System.Collections.Generic.Dictionary<string, object> GetArgumentValues(GraphQL.Types.ISchema schema, System.Collections.Generic.IEnumerable<GraphQL.Types.QueryArgument> definitionArguments, GraphQL.Language.AST.Arguments astArguments, GraphQL.Language.AST.Variables variables) { }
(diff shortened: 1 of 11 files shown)


src/GraphQL/Types/QueryArguments.cs
```diff
@@ -5,79 +5,99 @@
...
-                    throw new IndexOutOfRangeException();
+                NameValidator.ValidateName(value.Name, "argument");
 
-                ArgumentsList[index] = value;
+                base[index] = value;
             }
         }
 
-        internal List<QueryArgument> ArgumentsList { get; private set; }
+        public new int Count { get; }
```
sungam3r: ?

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
```diff
@@ -2004,15 +2004,21 @@ namespace GraphQL.Types
...
         public QueryArguments(System.Collections.Generic.IEnumerable<GraphQL.Types.QueryArgument> list) { }
-        public int Count { get; }
+        public new int Count { get; }
         public GraphQL.Types.QueryArgument this[int index] { get; set; }
         public void Add(GraphQL.Types.QueryArgument argument) { }
+        public void AddRange(System.Collections.Generic.IEnumerable<GraphQL.Types.QueryArgument> arguments) { }
         public GraphQL.Types.QueryArgument Find(string name) { }
-        public System.Collections.Generic.IEnumerator<GraphQL.Types.QueryArgument> GetEnumerator() { }
+        public void Insert(int index, GraphQL.Types.QueryArgument argument) { }
+        public void InsertRange(int index, System.Collections.Generic.IEnumerable<GraphQL.Types.QueryArgument> arguments) { }
```
sungam3r: Why these methods? Are they needed?

src/GraphQL/Types/QueryArguments.cs
Shane32: That’s a mistake. Whoops!

Shane32:
I’m thinking this whole PR probably needs to be scrapped. If an IList is accepted by FirldType, it would skip validation by QueryArguments.Add

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
Shane32: Natural consequence of inheriting from List

Shane32:
So maybe I’ll just do a simple PR that changes the ExecutionHelper to accept an IEnumerable and similar. Not at a pc now but I’ll think about it some more.

Shane32:
Or if validation was moved into the QueryArgument constructor, then the entire QueryArguments class could be removed (or simply map to `List<QueryArgument>` but I'm not sure that’s possible or desirable

PR #1593, opened 2020-03-05
Change QueryArguments to implement IReadOnlyCollection and ICollection

Shane32:
This facilitates changing `IFieldType` to a readonly interface in #1572 .

Implementing `IReadOnlyCollection` does not require any other code changes, as it only adds `Count` which already exists on the class.

Implementing `ICollection` facilitates short-circuit evaluation by `Enumerable.Count()`

There are no breaking changes, as `GetArgumentValues` is a static method and cannot be overriden.

### DIFF
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -213,7 +213,7 @@ namespace GraphQL.Execution
             return scalar.ParseValue(input);
         }
 
-        public static Dictionary<string, object> GetArgumentValues(ISchema schema, QueryArguments definitionArguments, Arguments astArguments, Variables variables)
+        public static Dictionary<string, object> GetArgumentValues(ISchema schema, IReadOnlyCollection<QueryArgument> definitionArguments, Arguments astArguments, Variables variables)
         {
             if (definitionArguments == null || definitionArguments.Count == 0)
             {
@@ -222,7 +222,7 @@ namespace GraphQL.Execution
 
             var values = new Dictionary<string, object>(definitionArguments.Count);
 
-            foreach (var arg in definitionArguments.ArgumentsList)
+            foreach (var arg in definitionArguments)
             {
                 var value = astArguments?.ValueFor(arg.Name);
                 var type = arg.ResolvedType;
--- a/src/GraphQL/Types/QueryArguments.cs
+++ b/src/GraphQL/Types/QueryArguments.cs
@@ -5,7 +5,7 @@ using GraphQL.Utilities;
 
 namespace GraphQL.Types
 {
-    public class QueryArguments : IEnumerable<QueryArgument>
+    public class QueryArguments : IReadOnlyCollection<QueryArgument>, ICollection
     {
         public QueryArguments(params QueryArgument[] args)
         {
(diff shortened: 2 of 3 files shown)


src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -222,7 +222,7 @@ private static object ValueFromScalar(ScalarGraphType scalar, object input)
 
             var values = new Dictionary<string, object>(definitionArguments.Count);
 
-            foreach (var arg in definitionArguments.ArgumentsList)
+            foreach (var arg in definitionArguments)
```
sungam3r: `ArgumentsList` was used to not allocate iterator object in the heap. `List<T>.Iterator` is struct.

Shane32: It's only used when there are arguments, but still, I add an optimization

Shane32: It's not worth making an optimization here (or retaining the existing one).  I've ran benchmarks over and over with different techniques.  I can't do better than 0.1% less memory allocated (0.5 KB) with no measurable speed improvement or GC effect on the introspection test.  For the hero test, it shows about 1.5% speed improvement and no GC or memory difference at all.

The code just isn't on the hot path, as it only executes specifically when a field resolver calls `GetArgument<>`, and even then, only once per resolver.  Sometimes the simplest answer is the best.

See the `ReadonlyResolveFieldContext.Arguments` property, which lazy loads the argument values

If you want me to optimize it anyway, I can unroll the foreach loop into two separate copies, one when `definitionArguments is QueryArguments` and one otherwise, and after that, I can optionally consolidate the inner code with a local function or private class function.  The compiler can prevent heap allocations when using local functions rather than lambda functions (which always require a heap allocation).

See: https://docs.microsoft.com/en-us/dotnet/csharp/local-functions-vs-lambdas

sungam3r: can be left unchanged

src/GraphQL/Types/QueryArguments.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace GraphQL.Types
 {
-    public class QueryArguments : IEnumerable<QueryArgument>
+    public class QueryArguments : IReadOnlyCollection<QueryArgument>, ICollection
```
SimonCropp, line 8: with this change doe we still need `QueryArguments` ?

sungam3r: I asked this question from time to time even before these changes.

Shane32: It's a read-only collection and supports backwards compatibility.  If we are not going to pursue changing the design to utilize read-only interfaces, I'd rather use the `new[] { ... }` syntax (currently not supported due to field builders requiring a `QueryArguments` class).  An array isn't read-only, however.

sungam3r:
@Shane32 do you still find it worthwhile? I would like to shorten the list of PRs with minor changes, the positive effect of which is not obvious.

Shane32:
No, I'm fine with closing most of these PRs which deal with restructuring the project

PR #1594, opened 2020-03-05
Add IQueryArgument interface

Shane32:
In preparation for changing interfaces to read-only, this PR adds an `IQueryArgument` interface which will eventually be a read-only interface once the underlying interfaces are changed to read-only.

Since `ISchemaFilter.AllowArgument` now has an `IQueryArgument` parameter, this is a breaking change.

Once combined with the changes in #1572 and #1593 this will allow `IFieldType.Arguments` to be of type `IEnumerable<IQueryArguments>` or a similar read-only type, facilitating the transition of `IFieldType` to be a completely read-only interface, including its arguments.


sungam3r:
I propose to close this PR. I don't think this is the right direction. It complicates the code. We've already talked about a plethora of interfaces that should probably be reduced.

Shane32:
I'm fine if we don't want to change the class/interface structure of the project.

sungam3r:
It *may* be changed in the future more dramatically fixing design flaw with resolvers on input graph types.

PR #1611, opened 2020-03-13
Expose Extensions

### LINKED ISSUE #1606: Expose Extensions in IResolveFieldContext and IExecutionContext
## Summary

Currently if you're writing a query resolver, you can handle errors by accessing the `IResolveFieldContext.Errors` and adding the error.
Alternative you can throw the error.

As far as I can tell there is currently no way to do this with `Extensions`.
What I propose is to expose `Extensions` which a resolver can then fill with relevant data.

### Basic example

```c#
public interface IExecutionContext : IProvideUserContext
{
	//...

	public Dictionary<string, object> Extensions { get; }
}

{
	//...

	public Dictionary<string, object> Extensions { get; }
}
```

```c#
public class StarWarsQuery : ObjectGraphType
{
  public StarWarsQuery()
  {
    Field<DroidType>("hero", resolve: context => {
		var result = DataBase.GetDroid("1");

		if (!context.Extensions.TryGetValue("databaseStats", out var statsObj) && statsObj is Dictionary<string, object> stats) {
			stats = new Dictionary<string, object>();
			context.Extensions["databaseStats"] = stats;
		}

		stats["Cost"] = result.Cost;
		stats["TimesWaitingForThrottle"] = result.ThrottleTimes;

		return result.Data;
	  }
    );
  }
}
```

### Motivation

This is all still possible, but as far as I can tell requires one to create a per-execution context object to keep track of all the extensions you want to add, and then wrap the `IDocumentExecutor` to add the extensions gathered in the context after executing the document.

This change would simply make adding extension data easier.

sungam3r:
fixes #1606 

@BenjaBobs take a look. Are you comfortable with such an API?

### DIFF
--- a/src/GraphQL/Execution/ExecutionContext.cs
+++ b/src/GraphQL/Execution/ExecutionContext.cs
@@ -41,5 +41,7 @@ namespace GraphQL.Execution
         public Action<UnhandledExceptionContext> UnhandledExceptionDelegate { get; set; }
 
         public int? MaxParallelExecutionCount { get; set; }
+
+        public Dictionary<string, object> Extensions { get; set; } = new Dictionary<string, object>();
     }
 }
--- a/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
+++ b/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
@@ -120,5 +120,41 @@ namespace GraphQL
                 }
             }
         }
+
+        private static readonly object lockExtensions = new object();
+        private static readonly char[] _separators = new char[] { '.' };
+
+        public static object GetExtension(this IResolveFieldContext context, string path)
+        {
+            lock (lockExtensions)
+            {
+             
+            }
+        }
+
+        public static void SetExtension(this IResolveFieldContext context, string path, object value)
+        {
+            lock (lockExtensions)
+            {
+                string[] keys = path.Split(_separators);
+                var values = context.Extensions;
+
+                for (int i = 0; i < keys.Length - 1; ++i)
+                {
+                    if (values.TryGetValue(keys[i], out object v) && v is IDictionary<string, object> d)
+                    {
+                        values = d;
+                    }
+                    else
+                    {
+                        var temp = new Dictionary<string, object>();
+                        values[keys[i]] = temp; // override value if any
+                        values = temp;
+                    }
+                }
+
+                values[keys[keys.Length - 1]] = value;
+            }
+        }
     }
 }
--- a/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
+++ b/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
@@ -121,7 +121,6 @@ namespace GraphQL
             }
         }
 
-        private static readonly object _lockExtensions = new object();
         private static readonly char[] _separators = new char[] { '.' };
 
         /// <summary>
@@ -138,7 +137,7 @@ namespace GraphQL
             if (context.Extensions == null || context.Extensions.Count == 0)
                 return null;
 
-            lock (_lockExtensions)
+            lock (context.Extensions)
             {
                 var values = context.Extensions;
 
@@ -178,7 +177,7 @@ namespace GraphQL
             if (context.Extensions == null)
                 throw new ArgumentException("Extensions property is null", nameof(context));
 
-            lock (_lockExtensions)
+            lock (context.Extensions)
             {
                 var values = context.Extensions;
 
--- a/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
+++ b/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
@@ -121,39 +121,91 @@ namespace GraphQL
             }
         }
 
-        private static readonly object lockExtensions = new object();
+        private static readonly object _lockExtensions = new object();
         private static readonly char[] _separators = new char[] { '.' };
 
+        /// <summary>
+        /// Thread safe method to get value by path (key1.key2.keyN) from extensions dictionary.
+        /// </summary>
+        /// <param name="context">Context with extensions response map.</param>
+        /// <param name="path">Path to value in key1.key2.keyN format.</param>
+        /// <returns>Value, if any exists on the specified path, otherwise <c>null</c>.</returns>
         public static object GetExtension(this IResolveFieldContext context, string path)
         {
-            lock (lockExtensions)
+            if (context == null)
+                throw new ArgumentNullException(nameof(context));
+
+            if (context.Extensions == null || context.Extensions.Count == 0)
+                return null;
+
+            lock (_lockExtensions)
             {
-             
+                var values = context.Extensions;
+
+                if (path.IndexOf('.') != -1)
+                {
+                    string[] keys = path.Split(_separators);
+
+                    for (int i = 0; i < keys.Length - 1; ++i)
+                    {
+                        if (values.TryGetValue(keys[i], out object v) && v is IDictionary<string, object> d)
+                            values = d;
+                        else
+                            return null;
+                    }
+
+                    return values.TryGetValue(keys[keys.Length - 1], out object result) ? result : null;
+                }
+                else
+                {
+                    return values.TryGetValue(path, out object result) ? result : null;
+                }
             }
         }
 
+        /// <summary>
+        /// Thread safe method to set value by path (key1.key2.keyN) to extensions dictionary.
+        /// if the given path or its part contains values, then they will be overwritten.
+        /// </summary>
+        /// <param name="context">Context with extensions response map.</param>
+        /// <param name="path">Path to value in key1.key2.keyN format.</param>
+        /// <param name="value">Value to set.</param>
         public static void SetExtension(this IResolveFieldContext context, string path, object value)
         {
-            lock (lockExtensions)
+            if (context == null)
+                throw new ArgumentNullException(nameof(context));
+
+            if (context.Extensions == null)
+                throw new ArgumentException("Extensions property is null", nameof(context));
+
+            lock (_lockExtensions)
             {
-                string[] keys = path.Split(_separators);
                 var values = context.Extensions;
 
-                for (int i = 0; i < keys.Length - 1; ++i)
+                if (path.IndexOf('.') != -1)
                 {
-                    if (values.TryGetValue(keys[i], out object v) && v is IDictionary<string, object> d)
-                    {
-                        values = d;
-                    }
-                    else
-                    {
-                        var temp = new Dictionary<string, object>();
-                        values[keys[i]] = temp; // override value if any
-                        values = temp;
-                    }
-                }
+                    string[] keys = path.Split(_separators);
 
-                values[keys[keys.Length - 1]] = value;
+                    for (int i = 0; i < keys.Length - 1; ++i)
+                    {
+                        if (values.TryGetValue(keys[i], out object v) && v is IDictionary<string, object> d)
+                        {
+                            values = d;
+                        }
+                        else
+                        {
+                            var temp = new Dictionary<string, object>();
+                            values[keys[i]] = temp; // overwrite value if any
+                            values = temp;
+                        }
+                    }
+
+                    values[keys[keys.Length - 1]] = value;
+                }
+                else
+                {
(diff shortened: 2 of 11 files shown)


src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -41,5 +41,7 @@ public class ExecutionContext : IExecutionContext
         public Action<UnhandledExceptionContext> UnhandledExceptionDelegate { get; set; }
 
         public int? MaxParallelExecutionCount { get; set; }
+
+        public Dictionary<string, object> Extensions { get; set; } = new Dictionary<string, object>();
```
sungam3r: +1 allocation on each request

sungam3r:
@Shane32 Please review.

Shane32:
I'd like to better understand the purpose here.  Is this so that when writing third-party extensions, they can use a dedicated dictionary to store information?  Because functionally this `Extensions` property is nearly identical to the `UserContext` property.

Shane32:
**Assuming this functionality is being added for users (rather than third-party libraries)**, it seems identical to UserContext, with some sugar on top in the `GetExtension` and `SetExtension` method.  I'm not even sure the `GetExtension` and `SetExtension` methods should be part of GraphQL.  It seems to be an implementation detail.

In the provided example, it seems the need is to return values back from field resolvers to `ExecutionResult`. The new field (`Extensions`) isn't even thread safe (perhaps with good reason), which is necessary when using the default parallel execution strategy.

Perhaps we should just expose `UserContext` (via the `IProvideUserContext` interface) on `ExecutionResult`.  This simple code could do everything shown in @BenjaBobs sample:

1. Initialize `UserContext["databaseStats"] = new ConcurrentDictionary<string, object>();` while starting the request, or in a document listener while starting the request.

2. Add this extension function to their project:
``` csharp
public ConcurrentDictionary<string, object> GetDatabaseStats(this IProvideUserContext context)
  => (ConcurrentDictionary<string, object>)(context.UserContext["databaseStats"]);
```

Now resolvers can write this code to save values:
``` csharp
var stats = context.GetDatabaseStats();
stats["cost"] = result.Cost;
stats["TimesWaitingForThrottle"] = result.ThrottleTimes;
```

By adding the `UserContext` to the `ExecutionResult`, the values can be retrieved in a similar fashion:
``` csharp
var stats = executionResult.GetDatabaseStats();
var cost = (decimal)stats["cost"];
var throttleTimes = (string)stats["TimesWaitingForThrottle"];
```

By using a `ConcurrentDictionary`, it's thread safe without extra code.  And if someone wanted to save/return a list of values, rather than a dictionary, they could modify the above to use `ConcurrentList`.

**Again, this assumes we are not talking about special support for third-party libraries.**

Shane32:
If we are talking about third-party library support, `GetExtension` and `SetExtension` seem a bit overly complex.  I also think built-in thread safety should be provided through the `Extensions` property.  What about if `Extensions` was a `ConcurrentDictionary<Guid, ConcurrentDictionary<string, object>>`?  Or something along those lines, where extensions provide a guid or other means of uniquely identifying itself, which can be used to retrieve a dedicated dictionary for their own means.  Perhaps instead of directly exposing the interface, we expose `IExtensions` like this:

```csharp
    public interface IExtensions
    {
        ConcurrentDictionary<string, object> this[Guid id] { get; }
    }
    public class Extensions : ConcurrentDictionary<Guid, ConcurrentDictionary<string, object>>, IExtensions
    {
        public ConcurrentDictionary<string, object> this[Guid id]
            => GetOrAdd(id, _ => new ConcurrentDictionary<string, object>());
    }
```
I would also add `Extensions` to `IProvideUserContext` and have `ExecutionResult` inherit `IProvideUserContext`
``` csharp
    public interface IProvideUserContext
    {
        IDictionary<string, object> UserContext { get; }
        IExtensions Extensions { get; }
    }
```
Of course a string could be used to identify an extension rather than a Guid

sungam3r:
Thanks for feedback.

>  Because functionally this `Extensions` property is nearly identical to the `UserContext` property.

1. These are completely different things serving different purposes. `Extensions` property serves to return arbitrary data in the response - it is _output_ thing. `UserContext` is any shared information used by the server during request exection, this info is not intended to publish outside For example, authentication data can be stored there. How then to distinguish one data from another?

> The new field (Extensions) isn't even thread safe (perhaps with good reason), which is necessary when using the default parallel execution strategy.

2. Yes, isn't even thread safe intentionally. There are two extension methods which thread safe. By the way, `UserContext` is not thread safe too.

> Perhaps we should just expose `UserContext` (via the `IProvideUserContext` interface) on `ExecutionResult`.

3. 100% not. See point 1.

> What about if Extensions was a `ConcurrentDictionary<Guid, ConcurrentDictionary<string, object>>`? Or something along those lines, where extensions provide a guid or other means of uniquely identifying itself, which can be used to retrieve a dedicated dictionary for their own means. 

4. Why only two dictionary levels? Extensions can have an arbitrary level of complexity. Well, the key of course should be a string, not Guid.

5. `GetExtension` and `SetExtension` seem a bit overly complex. 

Maybe just a little. These are rather short methods. In fact, in 99% of cases only the first level of extensions will be used.

sungam3r:
> If we are talking about third-party library support

When I wrote this, I didn’t think of anything like this.

Shane32:
> `Extensions` property serves to return arbitrary data in the response

Then thread safety should be provided, as by default it is expected to run in a parallel execution strategy.  As it is now, `GetExtension` and `SetExtension` provide thread safety, but cannot be mixed with any other functions in a thread-safe manner, as the locking object is a private member.  At very minimum, I'd lock on the dictionary's `SyncRoot` property instead of a private member.

Also, I don't understand why the property is called `Extensions`.  Perhaps `Outputs` or something, but why `Extensions`?  To me, `Extensions` brings "third party extensions" to my memory.

Since the use case is not for third-party extensions, but simply a way to return information back to the result, I'd question that this is all necessary; a user can utilize UserContext for these needs with the few lines of code shown in my first sample, without any changes at all.

> By the way, `UserContext` is not thread safe too.

True if writing was involved, but the common use case is to only read from `UserContext` during the resolver's execution, which is thread safe.

> Why only two dictionary levels

Why any more?  The entire use case is rare in itself, and upon the even-rarer case that additional levels are required, the user could write an extension method for their own needs.

That all being said, I don't really mind however you choose to implement this feature, if at all.  It just seems like overkill to me.  Considering the use case, the most I'd do is add a simple `ConcurrentDictionary<string, object>` (not `IDictionary`) named `Outputs` to `ExecutionContext`, `IResolveFieldContext` and `ExecutionResult`, and call it a day.  No extension methods.  No special locking code.  No support for multi-level data.  Full support for concurrent writing to the dictionary, as that is expected.

BenjaBobs:
It would certainly solve my use case. I was on purpose not suggesting a specific implementation because `Errors`, for which the data structure is a simple dictionary, is not implemented exactly as such.
I was thinking you might want to do something similar with `Extensions` for the same reasons as was done with `Errors` (I am not aware of the reason why, but I assume there is one).

In the end I just want to be able to add extension data from a resolver, so I can see it in my frontend, and this definitely solves that. 😄

BenjaBobs:
> Also, I don't understand why the property is called Extensions. Perhaps Outputs or something, but why Extensions? To me, Extensions brings "third party extensions" to my memory.

See [GraphQL Spec - Response Format](https://spec.graphql.org/June2018/#sec-Response-Format)

Shane32:
If the goal is to match it up to the graphql spec, and if the PR enables this, that's great.  I would at least lock on `SyncRoot` instead of a private member, however.

sungam3r:
> Also, I don't understand why the property is called Extensions. Perhaps Outputs or something, but why Extensions? To me, Extensions brings "third party extensions" to my memory.

Because it is such a term that is used in the specification.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -120,5 +120,93 @@ public static async Task<TResult> TryAsyncResolve<TResult>(this IResolveFieldCon
                 }
             }
         }
+
+        private static readonly object _lockExtensions = new object();
```
Shane32:
```suggestion
```

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -120,5 +120,93 @@ public static async Task<TResult> TryAsyncResolve<TResult>(this IResolveFieldCon
...
+        /// <returns>Value, if any exists on the specified path, otherwise <c>null</c>.</returns>
+        public static object GetExtension(this IResolveFieldContext context, string path)
+        {
+            if (context == null)
+                throw new ArgumentNullException(nameof(context));
+
+            if (context.Extensions == null || context.Extensions.Count == 0)
+                return null;
+
+            lock (_lockExtensions)
```
Shane32:
```suggestion
            lock (context.Extensions)
```
Apparently `IDictionary<TKey, TValue>` doesn't inherit from `ICollection` and does not provide `SyncRoot`

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -120,5 +120,93 @@ public static async Task<TResult> TryAsyncResolve<TResult>(this IResolveFieldCon
...
+        /// <param name="value">Value to set.</param>
+        public static void SetExtension(this IResolveFieldContext context, string path, object value)
+        {
+            if (context == null)
+                throw new ArgumentNullException(nameof(context));
+
+            if (context.Extensions == null)
+                throw new ArgumentException("Extensions property is null", nameof(context));
+
+            lock (_lockExtensions)
```
Shane32:
```suggestion
            lock (context.Extensions)
```

sungam3r:
> If the goal is to match it up to the graphql spec, and if the PR enables this, that's great.

Yes. this PR closes the gap.

> I would at least lock on SyncRoot instead of a private member, however.

OK.

Shane32:
👍 

Didn't realize we were just trying to fulfill a part of the graphql spec !

sungam3r:
No one has any suggestions left?

BenjaBobs:
Takkng a last look.

BenjaBobs:
I might be wrong on this, but isn't it generally better to `await` on a a `SemaphoreSlim` than to `lock` on an object?

Otherwise looks good to me.

Shane32:
In this case I’d use lock, so other implementations can simply and effectively utilize the same lock. Using SemaphoreSlim would require another object on the context or a specialized interface exposed to the context.

sungam3r:
> I might be wrong on this, but isn't it generally better to `await` on a a `SemaphoreSlim` than to lock on an object?

This is the first time I hear this. I can say with confidence that for such a simple code as this, the simplest synchronization design will be enough.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -120,5 +120,92 @@ public static async Task<TResult> TryAsyncResolve<TResult>(this IResolveFieldCon
...
+        /// <param name="path">Path to value in key1.key2.keyN format.</param>
+        /// <returns>Value, if any exists on the specified path, otherwise <c>null</c>.</returns>
+        public static object GetExtension(this IResolveFieldContext context, string path)
+        {
+            if (context == null)
+                throw new ArgumentNullException(nameof(context));
+
+            if (context.Extensions == null || context.Extensions.Count == 0)
+                return null;
+
```
SimonCropp: why not make extensions a concurrent dictionary?

Shane32:
Each level of the nested dictionaries would need to be a concurrent dictionary. Then the explicit lock could be deleted

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
sungam3r: Because then it would be required to do concurrent dictionary at each level of nesting which is impossible. It would be strange to provide concurrent dictionary only at the first level. In fact, the user can set his own dictionary as value (not concurrent) and then the thread safety behavior will be different from level to level. So I decided to use regular dictionaries and 2 thread safe extension methods.

sungam3r:
> Each level of the nested dictionaries would need to be a concurrent dictionary. 

This is impossible to achieve (and not necessary).

PR #1614, opened 2020-03-15
Remove extra xml comment in ExecutionContext.cs

Shane32:
The applicable comment already exists in both IExecutionContext and ExecutionOptions


sungam3r:
OK, but `ExecutionContext` comment is better, more detailed than `IExecutionContext`.

Shane32:
The comment in `ExecutionOptions` contains the extra commentary, as it is designed to be used when people are initializing the field.  When used in `ExecutionContext`, it's mainly for read-only purposes, so the extra commentary is not required.  If you think it should be changed, let me know.

PR #1616, opened 2020-03-16
Scalars docs

sungam3r:


sungam3r:
@Shane32 Please review.

Shane32:
@sungam3r  Give me a couple days - I'm a bit swamped at the moment - thanks!!

src/GraphQL/Types/Scalars/ScalarGraphType.cs
```diff
@@ -2,12 +2,52 @@
...
+        /// responsible for preparing the scalar for transport to the client. It is only responsible
+        /// for generating an object which can eventually be serialized by some transport-focused API.
+        /// </summary>
+        /// <param name="value"> Resolved value. </param>
+        /// <returns>
+        /// The returned value of a the result coercion is part of the overall execution result.
+        /// Normally this value is a primitive value like String or Integer to make it easy for
+        /// the serialization layer. For complex types like a Date or Money Scalar this involves
+        /// formatting the value.
+        /// </returns>
```
Shane32: The text in `<returns>` doesn't show up with Visual Studio's intellisense, as far as I know, so I shy away from using it.  I'm okay if you want to leave it this way; but it simply means that it's only useful as a code comment, rather than as an intellisense helper.

Personally I'd put this text in another paragraph of the `<summary>` tag

sungam3r: I know. But I would not want this consideration to force me to transfer the entire description to summary. Why then all the other elements? Let's hope that in the next VS version intellisense will learn even better to show comments from all xml tags.

Shane32: Yup.  Most of the other elements, like types and params, work fine.  The only reason I didn't include types/params in most of the my comments is because (a) it seemed obvious/redundant (b) so the code doesn't have "excessive" xml comments without purpose.

Issue #1603, opened 2020-03-11
graphql-parser comments are lost when I call CoreToVanillaConverter.Convert(string body, GraphQLDocument source)


vincent-duret:
I am using graphql-dotnet v3-preview1490 and graphql-parser v5.1.0.
I did notice that the comments are well-parsed and present in the GraphQLDocument source parameter, but I have no way to get access to them through the Document returned by the CoreToVanillaConverter.Convert(string body, GraphQLDocument source) method.
Is it a well-known bug/limitation? Any clue to workaround this?

By the way, what is the justification to maintain two different ASTs? One inside graphql-dotnet and the other inside graphql-parser. It forces to stay in sync when the parser evolves for instance.

Thanks by advance
Vincent

sungam3r:
> By the way, what is the justification to maintain two different ASTs? One inside graphql-dotnet and the other inside graphql-parser. It forces to stay in sync when the parser evolves for instance.

@joemcbride ?

sungam3r:
Do you mean input request document?

vincent-duret:
Not sure to get your question but if it might help, here is the document I am trying to parse:
#id: id of the employee
query _($id: ID = "00000000-0000-0000-0000-000000000000") {
  DummyEmp(_where: {idDummyEmp: {_eq: $id}}) {
    #comment for id
    id: idDummyEmp
    name: strName
    email: Email
    phone: Phone
    address: Address
    city: City
  }
}

vincent-duret:
I did have a look to the source of the CoreToVanillaConverter class.
For instance, in the method Field Field(GraphQLFieldSelection source), the Comment property from the source parameter is not copied in the field variable returned by that function.

sungam3r:
Yes, it is input query document. I see that comments are not used anywhere in the `CoreToVanillaConverter`. @joemcbride is this done on purpose?

vincent-duret:
@sungam3r @joemcbride maybe you should have a way to access the underlying type from the parser. E.g. Field.Source

sungam3r:
What underlying type from the parser?

vincent-duret:
For instance, in the method `public Field Field(GraphQLFieldSelection source)` you build a field from the source parameter.
In the method, we could have:
`field.Source = source` where Source is of type GraphQLFieldSelection

It would allow to access all the properties coming from the graphql-parser node.

PR #1617, opened 2020-03-17
add support for comments

vincent-duret:
@sungam3r in regards to issue #1603 I've created a PR to add support for comments in graphql-dotnet.
Please note that I had to comment 3 tests that fail due to graphql-parser that does return a null instead of the proper comment.

### DIFF
--- a/src/GraphQL.Tests/Language/CommentTests.cs
+++ b/src/GraphQL.Tests/Language/CommentTests.cs
@@ -45,35 +45,6 @@ query _ {
         }
 
         [Fact]
-        public void query_comment_should_be_null()
-        {
-            const string query = @"
-query _ {
-    person {
-        name
-    }
-}";
-
-            var document = CoreToVanillaConverter.Convert(query, _parser.Parse(new Source(query)));
-            document.Operations.First().SelectionSet.Selections.OfType<Field>().First().Comment.ShouldBeNull();
-        }
-
-        [Fact]
-        public void query_comment_should_not_be_null()
-        {
-            const string query = @"
-query _ {
-    #comment
-    person {
-        name
-    }
-}";
-
-            var document = CoreToVanillaConverter.Convert(query, _parser.Parse(new Source(query)));
-            document.Operations.First().SelectionSet.Selections.OfType<Field>().First().Comment.ShouldBe("comment");
-        }
-
-        [Fact]
         public void field_comment_should_be_null()
         {
             const string query = @"
@@ -85,6 +56,8 @@ query _ {
 
             var document = CoreToVanillaConverter.Convert(query, _parser.Parse(new Source(query)));
             document.Operations.First()
+                .SelectionSet.Selections.OfType<Field>().First().Comment.ShouldBeNull();
+            document.Operations.First()
                 .SelectionSet.Selections.OfType<Field>().First()
                 .SelectionSet.Selections.OfType<Field>().First().Comment.ShouldBeNull();
         }
@@ -94,16 +67,19 @@ query _ {
         {
             const string query = @"
 query _ {
+    #comment1
     person {
-        #comment
+        #comment2
         name
     }
 }";
 
             var document = CoreToVanillaConverter.Convert(query, _parser.Parse(new Source(query)));
             document.Operations.First()
+                .SelectionSet.Selections.OfType<Field>().First().Comment.ShouldBe("comment1");
+            document.Operations.First()
                 .SelectionSet.Selections.OfType<Field>().First()
-                .SelectionSet.Selections.OfType<Field>().First().Comment.ShouldBe("comment");
+                .SelectionSet.Selections.OfType<Field>().First().Comment.ShouldBe("comment2");
         }
 
         [Fact]
--- a/src/GraphQL.Tests/Language/CommentTests.cs
+++ b/src/GraphQL.Tests/Language/CommentTests.cs
@@ -1,0 +1,203 @@
+using System.Linq;
+using GraphQL.Language;
+using GraphQL.Language.AST;
+using GraphQLParser;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Language
+{
+    public class CommentTests
+    {
+        private readonly Parser _parser;
+
+        public CommentTests()
+        {
+            _parser = new Parser(new Lexer());
+        }
+
+        [Fact]
+        public void operation_comment_should_be_null()
+        {
+            const string query = @"
+query _ {
+    person {
+        name
+    }
+}";
+            
+            var document = CoreToVanillaConverter.Convert(query, _parser.Parse(new Source(query)));
+            document.Operations.First().Comment.ShouldBeNull();
+        }
+
+        [Fact]
+        public void operation_comment_should_not_be_null()
+        {
+            const string query = @"#comment
+query _ {
+    person {
+        name
+    }
+}";
+
+            var document = CoreToVanillaConverter.Convert(query, _parser.Parse(new Source(query)));
+            document.Operations.First().Comment.ShouldBe("comment");
+        }
+
+        [Fact]
+        public void query_comment_should_be_null()
+        {
+            const string query = @"
+query _ {
+    person {
+        name
+    }
+}";
+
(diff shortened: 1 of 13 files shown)


sungam3r:
Fixes #1603

src/GraphQL.Tests/Language/CommentTests.cs
```diff
@@ -0,0 +1,203 @@
...
+#comment
+fragment human on person {
+        name
+}";
+
+            var document = CoreToVanillaConverter.Convert(query, _parser.Parse(new Source(query)));
+            document.Fragments.First().Comment.ShouldBe("comment");
+        }
+
+        //[Fact]
```
sungam3r: Can be uncommented now.

src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -297,6 +305,14 @@ public NameNode Name(GraphQLName name)
             return new NameNode(name.Value).WithLocation(name, _body);
         }
 
+        public CommentNode Comment(GraphQLComment comment)
+        {
+            if (comment == null)
+                throw new ArgumentNullException(nameof(comment));
+
+            return new CommentNode(comment.Text).WithLocation(comment, _body);
+        }
```
sungam3r:
```suggestion
        private CommentNode Comment(GraphQLComment comment)
        {
            if (comment == null)
                return null;

            return new CommentNode(comment.Text).WithLocation(comment, _body);
        }
```

And no need in
```c#
frag.CommentNode = source.Comment != null ? Comment(source.Comment) : null;
```
->
```c#
frag.CommentNode = Comment(source.Comment);
```

src/GraphQL.Tests/Language/CommentTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            var document = CoreToVanillaConverter.Convert(query, _parser.Parse(new Source(query)));
+            document.Operations.First().SelectionSet.Selections.OfType<Field>().First().Comment.ShouldBeNull();
+        }
+
+        [Fact]
+        public void query_comment_should_not_be_null()
+        {
+            const string query = @"
+query _ {
+    #comment
```
sungam3r: It is not _query_ comment, it is field comment.

src/GraphQL/Language/AST/Argument.cs
```diff
@@ -16,6 +16,10 @@ public Argument(NameNode name)
 
         public string Name => NamedNode?.Name;
         public NameNode NamedNode { get; }
```
sungam3r:
```suggestion
        public NameNode NameNode { get; }
```

To align this property name in other classes.

src/GraphQL/Language/AST/Argument.cs
```diff
@@ -16,6 +16,10 @@ public Argument(NameNode name)
...
+
+        public string Comment => CommentNode?.Comment;
+        public CommentNode CommentNode { get; set; }
```
sungam3r: http://spec.graphql.org/June2018/#sec-Source-Text.Ignored-Tokens

> Before and after **every** lexical token may be any amount of ignored tokens including WhiteSpace and Comment.

So I propose to move these properties
```c#
        public string Comment => CommentNode?.Comment;
        public CommentNode CommentNode { get; set; }
```

in `AbstractNode`.  So there will be less public api changes.

src/GraphQL/Language/AST/CommentNode.cs
```diff
@@ -0,0 +1,19 @@
...
+        public CommentNode(string comment)
+        {
+            Comment = comment;
+        }
+
+        public string Comment { get; }
+
+        public override bool IsEqualTo(INode node)
+        {
+            throw new NotImplementedException();
```
sungam3r: Why? Every node implements this except `NameNode`.

src/GraphQL/Language/AST/Argument.cs
vincent-duret: Ok

src/GraphQL.Tests/Language/CommentTests.cs
vincent-duret: ok

src/GraphQL/Language/CoreToVanillaConverter.cs
vincent-duret: Ok

src/GraphQL/Language/AST/Argument.cs
vincent-duret: CommentNode inherits from AbstractNode, so we cannot move this in AbstractNode

src/GraphQL/Language/AST/CommentNode.cs
vincent-duret: Done

src/GraphQL.Tests/Language/CommentTests.cs
vincent-duret: No, the commented tests will fail until a new version of the parser is available.
https://github.com/graphql-dotnet/parser/pull/66

src/GraphQL/Language/AST/Argument.cs
sungam3r: Why not?

src/GraphQL.Tests/Language/CommentTests.cs
sungam3r: изображение

vincent-duret: Does it mean that the appveyor tests will use that version when running the tests even if the solution still references the 5.0.1 version from NuGet?
image

sungam3r: You should update package of course.

sungam3r: >  still references the 5.0.1 version **from** NuGet

Project does not reference package **from** Nuget. Project uses some (one or more) feeds to resolve (download) references. Nuget feed has no 5.1.2 yet, but myget feed has.

sungam3r: Also we can ask @joemcbride to publish 5.1.2 to Nuget.

vincent-duret: @sungam3r what it is the feed url for the MyGet package ?
Thx

src/GraphQL/Language/AST/Argument.cs
vincent-duret: Done

src/GraphQL.Tests/Language/CommentTests.cs
vincent-duret: Done

src/GraphQL/Language/AST/CommentNode.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+
+namespace GraphQL.Language.AST
+{
+    public class CommentNode : AbstractNode
+    {
+        protected string _comment;
```
sungam3r: Why protected and not private?

src/GraphQL/Language/AST/CommentNode.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public CommentNode(string comment)
+        {
+            _comment = comment;
+        }
+
+        public new string Comment => _comment;
```
sungam3r: Is it required? With this property one may get the impression that you can set a comment on the comment. I guess that here you wanted to express this property as `Value` but I doubt that anyone needs this explicit property, since all work will go through `AbstractNode.Comment`.

vincent-duret: I did rename this property as Value but we need it because we need a way to get the actual text value of a comment through AbstractNode.Comment :
`public string Comment => CommentNode?.Value;`
If I remove it entirely, some tests are failing.
But maybe I get you wrong.

sungam3r: OK

vincent-duret:
@sungam3r Thx to have made my PR merged
Do you know when @joemcbride will publish new pakages for both graphql-dotnet and parser?
I will need those to make progress on my project.

sungam3r:
I don't know but you can always just add myget feed to your nuget config.

PR #1621, opened 2020-03-20
Fix validation rule for required arguments

sungam3r:
http://spec.graphql.org/June2018/#sec-Required-Arguments :
> Arguments can be required. An argument is required if the argument type is non‐null **and does not have a default value**. Otherwise, the argument is optional.

Given such a schema

```graphql
type Query
{
  a: A
}

type A
{
  field(p: Int! = 10): SomeType
}
```

and  query
```graphql
{
  a {
    field { <---- no args
    ...
  }
}
}
```

`ValidationError` occures: `$"Expected \"{ofType.Name}!\", found null."`

### DIFF
--- a/src/GraphQL.Tests/Bugs/ScalarsInputTest.cs
+++ b/src/GraphQL.Tests/Bugs/ScalarsInputTest.cs
@@ -34,10 +34,49 @@ mutation {
     sbyte
     sbyteArray
   }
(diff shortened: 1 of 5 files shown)


sungam3r:
The numbering of validation rules is currently out of specification. This will be fixed in a separate PR.

PR #1623, opened 2020-03-25
Support for serialize custom ExecutionResult

sungam3r:
I ran into an error when serializing `class MyExecutionResult : ExecutionResult { ... }` led to `OutOfMemoryException` since custom converter for `ExecutionResult` was not used because of strict condition inside `ExecutionResultContractResolver`. Then serializer started to serialize all properties from `ExecutionResult`. I wanted to use the copy constructor, but I could not, because it is `protected`. So it is `public` now too.

### DIFF
--- a/src/GraphQL.NewtonsoftJson/ExecutionResultContractResolver.cs
+++ b/src/GraphQL.NewtonsoftJson/ExecutionResultContractResolver.cs
@@ -10,7 +10,7 @@ namespace GraphQL.NewtonsoftJson
         private readonly CamelCaseNamingStrategy _camelCase = new CamelCaseNamingStrategy();
 
         protected override JsonConverter ResolveContractConverter(Type objectType) =>
-            objectType == typeof(ExecutionResult)
+            objectType == typeof(ExecutionResult) || objectType.IsSubclassOf(typeof(ExecutionResult))
                 ? new ExecutionResultJsonConverter()
                 : base.ResolveContractConverter(objectType);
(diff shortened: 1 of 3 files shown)


src/GraphQL.NewtonsoftJson/ExecutionResultContractResolver.cs
```diff
@@ -10,7 +10,7 @@ public class ExecutionResultContractResolver : DefaultContractResolver
         private readonly CamelCaseNamingStrategy _camelCase = new CamelCaseNamingStrategy();
 
         protected override JsonConverter ResolveContractConverter(Type objectType) =>
-            objectType == typeof(ExecutionResult)
+            objectType == typeof(ExecutionResult) || objectType.IsSubclassOf(typeof(ExecutionResult))
```
BenjaBobs: I think this is easier to understand:
```suggestion
            typeof(ExecutionResult).IsAssignableFrom(objectType)
```

Or maybe even
```suggestion
            objectType.IsAssignableTo(typeof(ExecutionResult))
```
with
```c#
internal static class TypeExtensions
{
	internal static bool IsAssignableTo(this Type A, Type b) => b.IsAssignableFrom(a);
}
```

sungam3r: OK, `typeof(ExecutionResult).IsAssignableFrom(objectType)`

PR #1632, opened 2020-04-03
Firecamp link added in readme

Nishchit14:


sungam3r:
Thanks for contribution. Please tell me, can Firecamp provide the same experience as [Playground](https://github.com/graphql-dotnet/server/tree/develop/src/Ui.Playground) / [Altair](https://github.com/graphql-dotnet/server/tree/develop/src/Ui.Altair) / [GraphiQL](https://github.com/graphql-dotnet/server/tree/develop/src/Ui.GraphiQL) SPA apps? These all are just [embedded html](https://github.com/graphql-dotnet/server/blob/develop/src/Ui.GraphiQL/Internal/graphiql.cshtml) with some replacement tokens inside. If so could you please also make a similar package in the server project to support Firecamp ?

Nishchit14:
@sungam3r Yes, It's a rich IDE kind of playground. The cloud-sync and team-collaboration like features make it different than others. 
Sure, we're experimenting on how we can make it server integrated package too. But for that purpose, many features will be removed as they are not supported by browser environment.

The web experiment is here as [FireQL.dev](https://fireql.dev/?url=https://api.spacex.land/graphql)

PR #1636, opened 2020-04-07
bugfix: Expected type "Decimal", found 39614081257132168796771975168.

sungam3r:
Now `decimal` takes priority over `BigInteger`. Also `DecimalGraphType` now can parse from `BigIntValue` literal.


Shane32:
Let me know if there is, or will be, a breaking change in this regard.  I don't believe I have any notes so far to this effect.

sungam3r:
I think no. No public API changes and negative side effects are unlikely.

PR #1638, opened 2020-04-08
Serialize any errors first

### LINKED ISSUE #1637: Sort `errors` first in serialized result
## Summary

Spec recommends errors (if present) should come first to help notice them. I just had a situation where I didn't notice errors were happening as they were showing after the data.
http://spec.graphql.org/June2018/#sec-Response-Format

### Motivation

Improve DX. 

### Implementation

Would need to take a look, but I image the `IDocumentWriter` impls would need to do this and we'd need to implement it for the Newtonsoft and S.T.Json impls.

benmccallum:
Fixes #1637

### DIFF
--- a/src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
+++ b/src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
@@ -12,8 +12,8 @@ namespace GraphQL.NewtonsoftJson
             {
                 writer.WriteStartObject();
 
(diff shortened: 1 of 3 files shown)


benmccallum:
@sungam3r, this is ready for review. Pretty simple change. If you wanted it controlled by a setting people could set on startup, we could do that, but surely people aren't indexing into responses :P

sungam3r:
If someone needs to set the serialization order, then we will do it later.

Issue #1641, opened 2020-04-11
Option to skip serialization of exception data to prevent leaking sensitive information


jsantha:
## Summary

Similary to `ExposeExceptions` allow to disable serialization of [Exception.Data](https://docs.microsoft.com/en-gb/dotnet/api/system.exception.data) that can contain sensitive data.

### Basic example

```
services.AddGraphQL(options =>
{
    options.ExposeExceptions = false;
    options.ExposeExceptionExtensions = false; // Something like this would be great
})
```

### Motivation

Certain exceptions can possible leak sensitive information through [Exception.Data](https://docs.microsoft.com/en-gb/dotnet/api/system.exception.data).

In the example response below, we are leaking confidential data about our database server through `Microsoft.Data.SqlClient.SqlException` and thre is no simple way to prevent this from happening globally (according to [ExecutionResultJsonConverter](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs))

```
{
    "data": ...,
    "errors": [
        {
            "message": "Error trying to resolve...",
            "locations": [],
            "path": [],
            "extensions": {
                "code": "SQL",
                "data": {
                    "HelpLink.ProdName": "Microsoft SQL Server",
                    "HelpLink.ProdVer": "12.00.2000",
                    "HelpLink.EvtSrc": "MSSQLServer",
                    "HelpLink.EvtID": "208",
                    "HelpLink.BaseHelpUrl": "http://go.microsoft.com/fwlink",
                    "HelpLink.LinkId": "20476",
                    "OriginalClientConnectionId": "<CENSORED>",
                    "RoutingDestination": "<CENSORED>"
                }
            }
        }
    ]
}
```

sungam3r:
I understand what you want, but let's first clarify one thing. `ExecutionResultJsonConverter` (in fact both of STJ and NSJ) fills `data` property only for "top-most" `ExecutionError`.  Please check source code. `Microsoft.Data.SqlClient.SqlException` can not be such top-most `ExecutionError`, it can only be `InnerException` inside `ExecutionError`. But the converter does not serialize data from inner exceptions. There is another option when the `Data` property is manually filled (copied) from another source. Can you explain your case?

jsantha:
I understand but that doesn't seem like my case.

In my case:

[...]

PR #1656, opened 2020-04-28
Separate error parsing from serialization

Shane32:
See #1641

### DIFF
--- a/src/GraphQL.Harness/GraphQLMiddleware.cs
+++ b/src/GraphQL.Harness/GraphQLMiddleware.cs
@@ -63,7 +63,7 @@ namespace Example
                 options.Inputs = request.Variables.ToInputs();
                 options.UserContext = _settings.BuildUserContext?.Invoke(context);
                 options.EnableMetrics = _settings.EnableMetrics;
-                options.ExposeExceptions = _settings.ExposeExceptions;
+                //options.ExposeExceptions = _settings.ExposeExceptions;
                 if (_settings.EnableMetrics)
                 {
                     options.FieldMiddleware
--- a/src/GraphQL.Harness/Startup.cs
+++ b/src/GraphQL.Harness/Startup.cs
@@ -1,4 +1,5 @@
 using Example;
+using GraphQL.Execution;
 using GraphQL.Instrumentation;
 using GraphQL.StarWars;
 using GraphQL.StarWars.Types;
@@ -10,6 +11,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace GraphQL.Harness
 {
@@ -28,6 +30,11 @@ namespace GraphQL.Harness
             // add execution components
             services.AddSingleton<IDocumentExecuter, DocumentExecuter>();
             services.AddSingleton<IDocumentWriter, DocumentWriter>();
+            services.AddSingleton<IErrorParser>(services =>
+            {
+                var settings = services.GetRequiredService<IOptions<GraphQLSettings>>();
+                return new ErrorParser(options => options.ExposeExceptions = settings.Value.ExposeExceptions);
+            });
 
             // add something like repository
             services.AddSingleton<StarWarsData>();
--- a/src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
+++ b/src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
@@ -1,18 +1,26 @@
 using System;
 using System.Linq;
+using GraphQL.Execution;
 using Newtonsoft.Json;
 
 namespace GraphQL.NewtonsoftJson
 {
     public class ExecutionResultJsonConverter : JsonConverter
     {
+        private readonly IErrorParser _errorParser;
+
+        public ExecutionResultJsonConverter(IErrorParser errorParser)
+        {
+            _errorParser = errorParser;
+        }
+
         public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
         {
             if (value is ExecutionResult result)
             {
                 writer.WriteStartObject();
 
-                WriteErrors(result.Errors, writer, serializer, result.ExposeExceptions);
+                WriteErrors(result.Errors, writer, serializer);
                 WriteData(result, writer, serializer);
                 WriteExtensions(result, writer, serializer);
 
@@ -33,7 +41,7 @@ namespace GraphQL.NewtonsoftJson
             serializer.Serialize(writer, data);
         }
 
-        private void WriteErrors(ExecutionErrors errors, JsonWriter writer, JsonSerializer serializer, bool exposeExceptions)
+        private void WriteErrors(ExecutionErrors errors, JsonWriter writer, JsonSerializer serializer)
         {
             if (errors == null || errors.Count == 0)
             {
@@ -44,14 +52,13 @@ namespace GraphQL.NewtonsoftJson
 
             writer.WriteStartArray();
 
-            errors.Apply(error =>
+            errors.Select(error => _errorParser.Parse(error)).Apply(error =>
             {
                 writer.WriteStartObject();
 
                 writer.WritePropertyName("message");
 
-                // check if return StackTrace, including all inner exceptions
-                serializer.Serialize(writer, exposeExceptions ? error.ToString() : error.Message);
+                serializer.Serialize(writer, error.Message);
 
                 if (error.Locations != null)
                 {
@@ -75,51 +82,16 @@ namespace GraphQL.NewtonsoftJson
                     serializer.Serialize(writer, error.Path);
                 }
 
-                WriteErrorExtensions(error, writer, serializer);
+                if (error.Extensions?.Count > 0)
+                {
+                    writer.WritePropertyName("extensions");
+                    serializer.Serialize(writer, error.Extensions);
+                } 
 
                 writer.WriteEndObject();
             });
 
             writer.WriteEndArray();
-        }
-
-        private void WriteErrorExtensions(ExecutionError error, JsonWriter writer, JsonSerializer serializer)
-        {
-            if (string.IsNullOrWhiteSpace(error.Code) && (error.Data == null || error.Data.Count == 0))
-            {
-                return;
-            }
-
-            writer.WritePropertyName("extensions");
-            writer.WriteStartObject();
-
-            if (!string.IsNullOrWhiteSpace(error.Code))
-            {
-                writer.WritePropertyName("code");
-                serializer.Serialize(writer, error.Code);
-            }
-
-            if (error.HasCodes)
-            {
-                writer.WritePropertyName("codes");
-                writer.WriteStartArray();
-                error.Codes.Apply(code => serializer.Serialize(writer, code));
-                writer.WriteEndArray();
-            }
-
-            if (error.Data?.Count > 0)
-            {
-                writer.WritePropertyName("data");
-                writer.WriteStartObject();
-                error.Data.Apply((key, value) =>
-                {
-                    writer.WritePropertyName(key.ToString());
-                    serializer.Serialize(writer, value);
-                });
-                writer.WriteEndObject();
-            }
-
-            writer.WriteEndObject();
         }
 
         private void WriteExtensions(ExecutionResult result, JsonWriter writer, JsonSerializer serializer)
--- a/src/GraphQL.SystemTextJson/DocumentWriter.cs
+++ b/src/GraphQL.SystemTextJson/DocumentWriter.cs
@@ -4,12 +4,14 @@ using System.Linq;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
+using GraphQL.Execution;
 
 namespace GraphQL.SystemTextJson
 {
     public class DocumentWriter : IDocumentWriter
     {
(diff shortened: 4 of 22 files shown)


src/GraphQL/Execution/ErrorParserOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// Provides options to be used with <see cref="ErrorParser"/>
+    /// </summary>
+    public class ErrorParserOptions
+    {
+        /// <summary>
+        /// Specifies whether stack traces should be serialized
+        /// </summary>
+        public bool ExposeExceptions { get; set; }
+        // public bool ExposeExtensions { get; set; }
```
Shane32, line 16: Prep for adding additional options to the exception parsing engine

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
```diff
@@ -75,53 +82,18 @@ private void WriteErrors(ExecutionErrors errors, JsonWriter writer, JsonSerializ
                     serializer.Serialize(writer, error.Path);
                 }
 
-                WriteErrorExtensions(error, writer, serializer);
+                if (error.Extensions?.Count > 0)
+                {
+                    writer.WritePropertyName("extensions");
+                    serializer.Serialize(writer, error.Extensions);
+                }
```
Shane32, line 89: This code within both of the serializers gets cleaned up a whole lot, just writing whatever is in the extensions property.

src/GraphQL.SystemTextJson/DocumentWriter.cs
```diff
@@ -21,26 +23,42 @@ public DocumentWriter(bool indent)
...
         }
 
+        public DocumentWriter(bool indent, IErrorParser errorParser)
+            : this(GetDefaultSerializerOptions(indent), errorParser ?? throw new ArgumentNullException(nameof(errorParser)))
+        {
+            _errorParser = errorParser;
+        }
+
+        public DocumentWriter(IErrorParser errorParser)
+            : this(false, errorParser)
```
Shane32, line 33: DI will automatically call this constructor if an `IErrorParser` is registered

src/GraphQL.Tests/Errors/ErrorExtensionsTests.cs
```diff
@@ -9,17 +9,11 @@ namespace GraphQL.Tests.Errors
...
+        public void should_add_extension_object_when_exception_is_thrown_with_error_code()
         {
             string query = "{ firstSync }";
             string code = "FIRST";
 
-            var result = await Executer.ExecuteAsync(_ =>
-            {
-                _.Schema = Schema;
-                _.Query = query;
-            });
```
Shane32, line 21: `result` was not being used so I deleted it.

src/GraphQL.Tests/Errors/ErrorExtensionsTests.cs
```diff
@@ -32,16 +26,10 @@ public async Task should_add_extension_object_when_exception_is_thrown_with_erro
...
-        public async Task should_not_add_extension_object_when_exception_is_thrown_without_error_code()
+        public void should_not_add_extension_object_when_exception_is_thrown_without_error_code()
         {
             string query = "{ uncodedSync }";
 
-            var result = await Executer.ExecuteAsync(_ =>
-            {
-                _.Schema = Schema;
-                _.Query = query;
-            });
```
Shane32, line 43: `result` was not being used so I deleted it

src/GraphQL/Execution/ErrorParser.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Linq;
+using System.Text;
+
+namespace GraphQL.Execution
+{
+    public class ErrorParser : IErrorParser
+    {
+        private readonly ErrorParserOptions _options;
+
+        public ErrorParser() : this(new ErrorParserOptions()) { }
```
Shane32, line 12: We could add another constructor with a `bool ExposeExceptions` parameter

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -61,9 +61,6 @@ public class ExecutionOptions : IProvideUserContext
         /// <summary>Field and argument names are sanitized by the provided <see cref="INameConverter"/>; defaults to <see cref="CamelCaseNameConverter"/></summary>
         public INameConverter NameConverter { get; set; } = CamelCaseNameConverter.Instance;
 
-        /// <summary>Allows unhandled <see cref="Exception"/> stack traces to be serialized into GraphQL query result json along with exception messages; defaults to only <see cref="Exception.Message"/></summary>
-        public bool ExposeExceptions { get; set; }
```
Shane32, line 65: It would be difficult to support backwards compatibility, since the `DocumentExecuter` does not call the serializer, and `ErrorParser` does not receive a reference to `ExecutionResult`.

src/GraphQL.Harness/Startup.cs
```diff
@@ -28,6 +30,11 @@ public void ConfigureServices(IServiceCollection services)
             // add execution components
             services.AddSingleton<IDocumentExecuter, DocumentExecuter>();
             services.AddSingleton<IDocumentWriter, DocumentWriter>();
+            services.AddSingleton<IErrorParser>(services =>
+            {
+                var settings = services.GetRequiredService<IOptions<GraphQLSettings>>();
+                return new ErrorParser(options => options.ExposeExceptions = settings.Value.ExposeExceptions);
+            });
```
Shane32, line 37: This is a sample of the code necessary to be added to a project that allows configuration of the `ExposeExceptions` property from within a settings object.  Simpler implementations could use something like this:
```csharp
services.AddSingleton<IErrorParser>(_ => new ErrorParser(o => o.ExposeExceptions = true));
```

src/GraphQL.Harness/GraphQLMiddleware.cs
```diff
@@ -63,7 +63,6 @@ private async Task ExecuteAsync(HttpContext context, ISchema schema)
                 options.Inputs = request.Variables.ToInputs();
                 options.UserContext = _settings.BuildUserContext?.Invoke(context);
                 options.EnableMetrics = _settings.EnableMetrics;
-                options.ExposeExceptions = _settings.ExposeExceptions;
```
Shane32, line 66: The breaking change is, mainly, to remove this line of code and instead register an `IErrorParser` within the DI framework

src/GraphQL/Execution/ParsedError.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// starting from 1 which describe the beginning of an associated syntax element
+        /// </summary>
+        public IEnumerable<ErrorLocation> Locations { get; set; }
+
+        /// <summary>
+        /// A list of path segments starting at the root of the response and ending with the field associated with the error.
+        /// Path segments that represent fields should be strings, and path segments that represent list indices should be 0‐indexed integers.
+        /// If the error happens in an aliased field, the path to the error should use the aliased name, since it represents a path in the response, not in the query.
+        /// </summary>
+        public IEnumerable<object> Path { get; set; }
```
Shane32, line 32: Note that the current implementation of the `Path` property does not meet spec, as it returns strings for indices rather than numbers as shown in the spec description and reference sample code.  This code is prepped to rectify that error, but will require further changes.

src/GraphQL/Execution/ErrorParser.cs
```diff
@@ -0,0 +1,61 @@
...
+            _options = new ErrorParserOptions();
+            optionsBuilder(_options);
+        }
+
+        public ParsedError Parse(ExecutionError executionError)
+        {
+            IDictionary<string, object> extensions = null;
+            if (!string.IsNullOrWhiteSpace(executionError.Code) ||
+                // skip next check to match existing functionality
+                // (executionError.HasCodes) ||
```
Shane32, line 40: Currently, a new `ExecutionError` with an `InnerException` of `System.Exception` creates an `ExecutionError` with the `Code` property set to `String.Empty`.  This should be reviewed, as I would change it.  In any case, I was required to comment out the above line of code due to the semantics of the `HasCodes` property to pass all tests.

src/GraphQL.Harness/Startup.cs
```diff
@@ -28,6 +30,11 @@ public void ConfigureServices(IServiceCollection services)
             // add execution components
             services.AddSingleton<IDocumentExecuter, DocumentExecuter>();
             services.AddSingleton<IDocumentWriter, DocumentWriter>();
+            services.AddSingleton<IErrorParser>(services =>
+            {
+                var settings = services.GetRequiredService<IOptions<GraphQLSettings>>();
+                return new ErrorParser(settings.Value.ExposeExceptions);
+            });
```
Shane32, line 37: This is a sample of the code necessary to be added to a project that allows configuration of the `ExposeExceptions` property from within a settings object. Simpler implementations could use something like this:
```csharp
services.AddSingleton<IErrorParser>(_ => new ErrorParser(o => o.ExposeExceptions = true));
```

sungam3r:
I will not be able to review PR right away. Need some time.

src/GraphQL/Execution/ParsedError.cs
sungam3r: Let's merge #1657 first.

Shane32: Fine with me.  FYI, there are no conflicts between the two PRs (besides apiapprovals).  Either can be merged first with no changes to the other PR.

src/GraphQL.Tests/Errors/ErrorParserTests.cs
```diff
@@ -0,0 +1,243 @@
...
+            parsed.Extensions.ShouldBeNull();
+        }
+
+        [Fact]
+        public void message_and_path()
+        {
+            var error = new ExecutionError(null)
+            {
+                //Path = new object[] { "root", 23, "child" },
+                Path = new string[] { "root", "23", "child" },
```
Shane32, line 141:
```suggestion
                Path = new object[] { "root", 23, "child" },
```
Once #1657 is pulled in, we can update this test to appropriately reflect the sample.

src/GraphQL.Tests/Errors/ErrorParserTests.cs
```diff
@@ -0,0 +1,243 @@
...
+            };
+
+            var parsed = new ErrorParser().Parse(error);
+            parsed.Message.ShouldBe(error.Message);
+            parsed.Locations.ShouldBeNull();
+            //parsed.Path.ShouldBe(new object[] { "root", 23, "child" });
+            parsed.Path.ShouldBe(new string[] { "root", "23", "child" });
```
Shane32, line 148:
```suggestion
            parsed.Path.ShouldBe(new object[] { "root", 23, "child" });
```
Here also.

src/GraphQL.Tests/Errors/ErrorParserTests.cs
```diff
@@ -0,0 +1,290 @@
...
+            error.Code.ShouldBe("");
+            error.HasCodes.ShouldBeTrue();
+            error.Codes.ShouldBe(new[] { "" });
+
+            var parsed = new ErrorParser().Parse(error);
+            parsed.Extensions.ShouldBeNull();
+        }
+
+        [Fact]
+        public void codes_with_blank_code_has_undefined_behavior()
```
Shane32, line 269: This behavior should be reviewed.  Presently when the `InnerException` property is of type `Exception`, it prevents `code` and `codes` from serializing -- unless there is items in the `Data` collection.

Shane32:
I added the tests you requested @sungam3r

sungam3r:
It is different PR.

Shane32:
Yup, well, won't hurt here.

src/GraphQL.Tests/Errors/ErrorParserTests.cs
```diff
@@ -0,0 +1,288 @@
...
+            parsed.Path.ShouldNotBeNull();
+            parsed.Path.ShouldBe(error.Path);
+        }
+
+        [Fact]
+        public void innerException_with_path()
+        {
+            var error = new ExecutionError("Test error message", new ArgumentNullException(null, new ArgumentOutOfRangeException()))
+            {
+                Path = new string[] { "path1", "path2" },
```
Shane32, line 183:
```suggestion
                Path = new object[] { "path1", 22, "path2" },
```

sungam3r:
@Shane32  I haven’t done a review yet but when I started thinking about how to provide a convenient configuration of all the settings/options, I came to the conclusion that it lacked access to `IServiceProvider`. The lack of this feature limited me in refactoring options. See https://github.com/graphql-dotnet/server/pull/360. You may have some thoughts.

Shane32:
Just to be clear: what doesn’t have access to IServiceProvider? IErrorParser is constructed from the service provider, as shown in the updated Harness sample of the PR. Assuming of course that IDocumentWriter is pulling from DI.

sungam3r:
I did not speak about your code. I mean hypothetical refactoring. I no longer remember what was exactly on my mind. You just can ignore it.

Shane32:
@sungam3r Any thoughts about this?  I'm going to be heavy into error handling for the next few weeks, and I think this, along with something like #1658 , would really be a benefit.  I plan to do a thorough review of #1658 and/or propose similar changes depending on what I see and what makes sense in production.

sungam3r:
I will re-read this thread and review it in the coming days.

Shane32:
I have not worked on it in the last three months, so I'll review it again also.

Shane32:
I reviewed this again; I think it's ready to go.

PR #1657, opened 2020-04-28
Fix errors.path property to return numbers for list indexes

Shane32:
Currently the `errors[].path` property returned in the graphql response returns strings to indicate positions within an array.  According to the spec it should return numbers.  This fixes that error.

See https://spec.graphql.org/June2018/#sec-Errors

> This field should be a list of path segments starting at the root of the response and ending with the field associated with the error. Path segments that represent fields should be strings, and path segments that represent list indices should be 0‐indexed integers. If the error happens in an aliased field, the path to the error should use the aliased name, since it represents a path in the response, not in the query.
>
> For example, if fetching one of the friends’ names fails in the following query:
> ```
> {
>   hero(episode: $episode) {
>     name
>     heroFriends: friends {
>       id
>       name
>     }
>   }
> }
> ```
> The response might look like:
> ```
> {
>   "errors": [
>     {
>       "message": "Name for character with ID 1002 could not be fetched.",
>       "locations": [ { "line": 6, "column": 7 } ],
>       "path": [ "hero", "heroFriends", 1, "name" ]
>     }
>   ],
>   "data": {
>     "hero": {
>       "name": "R2-D2",
>       "heroFriends": [
>         {
>           "id": "1000",
>           "name": "Luke Skywalker"
>         },
>         {
>           "id": "1002",
>           "name": null
>         },
>         {
>           "id": "1003",
>           "name": "Leia Organa"
>         }
>       ]
>     }
>   }
> }
> ```

### DIFF
--- a/src/GraphQL/Execution/ExecutionNode.cs
+++ b/src/GraphQL/Execution/ExecutionNode.cs
@@ -88,24 +88,40 @@ namespace GraphQL.Execution
             }
         }
 
+        private static readonly object _num0 = 0;
+        private static readonly object _num1 = 1;
+        private static readonly object _num2 = 2;
+        private static readonly object _num3 = 3;
+        private static readonly object _num4 = 4;
+        private static readonly object _num5 = 5;
+        private static readonly object _num6 = 6;
+        private static readonly object _num7 = 7;
+        private static readonly object _num8 = 8;
+        private static readonly object _num9 = 9;
+        private static readonly object _num10 = 10;
+        private static readonly object _num11 = 11;
+        private static readonly object _num12 = 12;
+        private static readonly object _num13 = 13;
+        private static readonly object _num14 = 14;
+        private static readonly object _num15 = 15;
         private static object GetObjectIndex(int index) => index switch
         {
-            0 => 0,
-            1 => 1,
-            2 => 2,
-            3 => 3,
-            4 => 4,
-            5 => 5,
-            6 => 6,
-            7 => 7,
-            8 => 8,
-            9 => 9,
-            10 => 10,
-            11 => 11,
-            12 => 12,
-            13 => 13,
-            14 => 14,
-            15 => 15,
+            0 => _num0,
+            1 => _num1,
+            2 => _num2,
+            3 => _num3,
+            4 => _num4,
+            5 => _num5,
+            6 => _num6,
+            7 => _num7,
+            8 => _num8,
+            9 => _num9,
+            10 => _num10,
+            11 => _num11,
+            12 => _num12,
+            13 => _num13,
+            14 => _num14,
+            15 => _num15,
             _ => index
         };
     }
--- a/src/GraphQL/Execution/ExecutionNode.cs
+++ b/src/GraphQL/Execution/ExecutionNode.cs
@@ -60,7 +60,7 @@ namespace GraphQL.Execution
             return null;
         }
 
-        public IEnumerable<string> Path
+        public IEnumerable<object> Path
         {
             get
             {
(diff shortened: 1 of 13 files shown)


src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -88,25 +88,25 @@ public IEnumerable<string> Path
...
-            8 => "8",
-            9 => "9",
-            10 => "10",
-            11 => "11",
-            12 => "12",
-            13 => "13",
-            14 => "14",
-            15 => "15",
-            _ => index.ToString()
+        private static object GetObjectIndex(int index) => index switch
```
sungam3r, line 91: Well :) This method is no longer needed.

sungam3r: No. I was wrong. We need it to eliminate boxing. You need to change implementation to return cached boxed integers instead of boxing them every call.

src/GraphQL/Instrumentation/ApolloTracingExtensions.cs
```diff
@@ -65,20 +64,5 @@ public static void EnrichWithApolloTracing(this ExecutionResult result, DateTime
...
             return trace;
         }
-
-        private static IEnumerable<object> ConvertPath(IEnumerable<string> stringPath)
-        {
-            foreach (var step in stringPath)
-            {
-                if (int.TryParse(step, out var arrayIndex))
-                {
-                    yield return arrayIndex;
```
sungam3r, line 75: It's funny, but in this place, the behavior matches the specification, converting the string back to a number.

sungam3r:
@rose-a Is graphql-client ready for these changes?

rose-a:
> 
> 
> @rose-a Is graphql-client ready for these changes?

`Path` is implemented as an `object[]`, so it should be fine... wouldn't do any harm to add a corresponding test though...

src/GraphQL/Execution/ExecutionNode.cs
Shane32: I know the goal but wasn't sure what was necessary to accomplish it.  I had `1 => (object)1`, but C# reported that the cast to object was unnecessary, which is why it is currently `1 => 1`.  I thought of using an array, but wasn't sure if that would be slower or faster than a switch expression.  I'm not going to spend time benchmarking this (it never executes anyway unless you're logging), but if you have a suggestion I'll code it.  Should I create 16 readonly fields?  e.g. `private static readonly object num0 = 0;`

Shane32:
> Could you please add tests for path serialization results?

Is there something inadequate with the existing `should_include_path_with_list_index` test?  I don't mind adding a test... but what do you want to test?  I can duplicate that test and name the new one `path_list_index_returns_number` if you like.

src/GraphQL/Execution/ExecutionNode.cs
sungam3r: Right.

Shane32: Done

sungam3r:
should_include_path_with_list_index tests object properties, not serialized results.

Shane32:
Sorry I was thinking about the other PR

Shane32:
Yes I can add some tests

Shane32:
Ok no, I was confused again, and added tests to the other PR.  I still don't understand what test you want added to this PR @sungam3r

Shane32:
I really didn't change any code other than having list indexes return integers instead of strings.  The below test already covers this issue.  It was simply coded wrong.

```csharp
        [Fact]
        public async Task should_include_path_with_list_index()
        {
            var result = await Executer.ExecuteAsync(_ =>
            {
                _.Schema = Schema;
                _.Query = @"{ testSubList { one two } }";
            });

            result.Errors.Count.ShouldBe(1);
            var error = result.Errors.First();
            //error.Path.ShouldBe(new[] { "testSubList", "0", "two" }); //<-- old, incorrect
            error.Path.ShouldBe(new object[] { "testSubList", 0, "two" }); //<-- new, correct
        }
```

Shane32:
If you like, I can duplicate it to a new test that performs identically, with a new name `should_return_integers_for_list_indexes`. I don't know what else to do.

sungam3r:
I am talking about **serialization**. You changed type from `IEnumerable<string>` to `IEnumerable<object>`. `System.Text.Json` has special requirements and rules about serialized types. This bothers me. We need to serialize results and see if serializer(s) does it well.

Shane32:
Ahh.  Thanks for explaining!  I think I have it covered now; please review.

Issue #1650, opened 2020-04-17
ExecutionError: Differentiate between client and server error.


GrangerAts:
## Summary

As the author of a GraphQL.Net API, it would be very helpful to be able to differentiate between an ExecutionError where the **Api-author screwed up** as opposed to when the **Api-client screwed up**. Presently, one needs to hard-code checks to exception messages found in the GraphQL.Net source code, which is a brittle approach.

### Basic example

Implement a `FieldAsync<ListGraphType<...>>` resolver and directly return your data from an `async` method; make the mistake of omitting the `await` keyword. `ExecutionStrategy.SetArrayItemNodes()` will raise an `ExecutionError` that says "User error: expected an IEnumerable list though did not find one."

### Motivation

The above error is obviously the Api-author's fault, not the Api-client's. As an Api-author, I'd like to have my Api clearly behave as if it's my fault, not the client's fault. I want to log that; I don't want to log `ExecutionErrors` that are not my fault.

I feel like there are 2 options:

#### 1. Use a non-ExecutionError
For Api-author errors, don't use `ExecutionError` at all. Instead use a built-in Exception (like `InvalidOperationException` or `ApplicationException`). This way when the Api-author screws up, they handle it just like any other "unhandled" exception that comes out of GraphQL.Net.

Searching on `ExecutionError` in GraphQL.Net, there's a number of other spots that raise "Api-author" level errors besides the above "Basic example". E.g. "Only add root types.", "Schema is not configured for mutations", "A schema is required.", "Must provide an enter or leave function.", "Unhandled document definition", etc.

#### 2. Sub-class ExecutionError
If the `ExecutionError` was instead a new subclass that could be used to globally determine "It's the Api-author's fault", then the GraphQL service has the opportunity to behave better.   

But it feels like `ExecutionError` is not supposed to be used for things that aren't problems caused by the Api-author. For example, there is already `InvalidValueException`, and that's (appropriately) used when it's an Api-client error (validation).

sungam3r:
First, I can not find `InvalidValueError` in the current codebase. Second, I think (generally) there are only **Api-author errors** at the execution stage and only **Api-client errors** at the validation stage. It seems to me that you are trying to solve the problem by inappropriate means.

> Presently, one needs to hard-code checks to exception messages found in the GraphQL.Net source code, which is a brittle approach.

See this https://github.com/graphql/graphql-spec/issues/698

GrangerAts:
[...]

PR #1658, opened 2020-04-29
[DRAFT] Change server-side errors to exceptions

GrangerAts:
This is in regards to #1650, and is a starting-point. I don't know the code well enough to feel like I got it right, but it should be enough to clarify my intent.

Were this sort of change to be made, it should make no significant difference to a developer using `ExceptionOptions.ThrowOnUnhandledException = false`. GraphQL.Net would wrap them into `ExecutionError` objects like all the other existing non-`ExecutionError` exceptions that are possibly thrown. 

But for a developer using `ExceptionOptions.ThrowOnUnhandledException = true`, they would have the ability to manually handle those server-side errors in whatever way they felt necessary. For example, they could use custom FieldMiddleware and change the exception into an `ExecutionError` that they were more comfortable with.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -41,12 +41,12 @@ namespace GraphQL
         {
             if (options.Schema == null)
             {
-                throw new ExecutionError("A schema is required.");
+                throw new InvalidOperationException("A schema is required.");
             }
 
             if (string.IsNullOrWhiteSpace(options.Query))
             {
-                throw new ExecutionError("A query is required.");
+                throw new InvalidOperationException("A query is required.");
             }
         }
 
@@ -92,7 +92,7 @@ namespace GraphQL
 
                 if (operation == null)
                 {
-                    throw new ExecutionError("Unable to determine operation from query.");
+                    throw new InvalidOperationException("Unable to determine operation from query.");
                 }
 
                 IValidationResult validationResult;
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -95,8 +95,7 @@ namespace GraphQL.Execution
 
             if (!(parent.Result is IEnumerable data))
             {
-                var error = new ExecutionError("User error: expected an IEnumerable list though did not find one.");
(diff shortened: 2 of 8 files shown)


GrangerAts:
Additionally, here's the comment where I put the method that I _presently_ intend to use to identify `ExecutionErrors` that are the result of an error on the part of the Api-author/developer: https://github.com/graphql-dotnet/graphql-dotnet/issues/1650#issuecomment-621502247

sungam3r:
I see many failed tests.

GrangerAts:
Yes. :)
When I built locally (VS2019), there were zero errors, but tons of warnings and messages (none of which seemed to apply to my changes). I had expected that the "tests" ran as part of the build process, but I was wrong---I noticed after the CI process ran after I submitted the PR. 

I looked and it appears they're all just unit tests that expected `ExecutionError` for situations where I'd turned them into different exceptions, so the fix would be to change those tests. I wasn't sure at that point if it was going to matter, since I was considering this PR as more of a quick way for a you or another contributor to see the direction I was going and to note any obvious mistakes/assumptions I'd made. At that point I could do more work to move forward, or something else could be done.

How do you wish to proceed?

sungam3r:
I will review some time later.

GrangerAts:
Also, _if_ you want me to move forward with this (and update the unit tests, of course), I think it would be wise to additionally change `DocumentExecutor.ExecuteAsync()` to add a `catch (ExecutionError error)` block similar to what's being done in `ExecutionStrategy.ExecuteNodeAsync()`.

E.g.
```
catch (ExecutionError ex)
{
    result = new ExecutionResult
    {
        Errors = new ExecutionErrors
        {
            new ExecutionError(ex.Message, ex)
        }
    };
}
```

Shane32:
I split this PR into PRs #1773, #1735, #1777, #1778, #1779, and #1769.  Once they are all individually approved and merged, we can close this PR.

GrangerAts:
Thanks! Agreed.
I didn't realize I'd gotten side-tracked for **all summer**; this rough-draft is definitely stale and should be scrapped in favor of the other work you're doing. I'll just close this now.

Shane32:
Thanks for your lead on this @GrangerAts - it will make a definite improvement for this project.

PR #1659, opened 2020-04-30
Update schema-generation.md

sungam3r:
delete link


BenjaBobs:
I think this is the right decision until https://github.com/nav-malik/Dotnet-GraphQL.Code.Generator gets some content.

Issue #1565, opened 2020-02-24
Problem with Data loaders in federation resolvers


ShadyNawara:
## Summary
I have been trying to use dataloaders with the apollo federation but something doesnt seem to be working right.
Here is my implementation so for

[GraphQLController]
```
    [Route("[controller]")]
    public class GraphQLController : Controller
    {
        private readonly IDocumentExecuter _documentExecuter;
        private readonly IDocumentWriter _documentWriter;
        private readonly IDocumentExecutionListener _documentListener;

        public GraphQLController(IDocumentExecuter executer, IDocumentWriter writer, DataLoaderDocumentListener documentListener)
        {
            _documentExecuter = executer;
            _documentWriter = writer;
            _documentListener = documentListener;
        }

        [HttpPost]
        public async Task<IActionResult> Post([FromServices]ISchema schema)
        {
            // serialization some times fail because of special characters in the request body so manual deserialization is better
            StreamReader requestBody = new StreamReader(Request.Body, encoding: System.Text.Encoding.UTF8, detectEncodingFromByteOrderMarks: false);
            GraphQLRequest request = Newtonsoft.Json.JsonConvert.DeserializeObject<GraphQLRequest>(await requestBody.ReadToEndAsync());
            requestBody.Dispose();

            if (request == null) return Content("{}", "application/json");

            var options = new ExecutionOptions
            {
                Schema = schema,
                OperationName = request.OperationName,
                Query = request.Query,
                Inputs = request.Variables.ToInputs(),
                ExposeExceptions = true,
            };
            options.Listeners.Add(_documentListener);

            var result = await _documentExecuter.ExecuteAsync(options).ConfigureAwait(false);

            if (result.Errors?.Count > 0)
            {
                return BadRequest(result);
            }

            return Content(await _documentWriter.WriteToStringAsync(result), "application/json");
        }
    }
```
[startup.cs]

```
// Dataloaders
services.AddSingleton<IDataLoaderContextAccessor, DataLoaderContextAccessor>();
services.AddSingleton<DataLoaderDocumentListener>();

services.AddTransient<ISchema>(s =>
            {
                var store = s.GetRequiredService<BazStore>();
                var accessor = s.GetRequiredService<IDataLoaderContextAccessor>();

                return FederatedSchema.For(@"
                           extend type Query {
                                  getBazzes: [Baz]
                           }

                           extend type Bar @key(fields: ""id"") {
                                  id: Int @external
                           }

[...]

PR #1660, opened 2020-05-03
Fixing and issue #1565 - data loader in federated schema

mac2000:
Hello, at moment I were able to run code first federation with gateway, metrics and apollo graph manager, except this one

Thanks to @ShadyNawara and his #1565 issue - this one needs to be fixed first before everything else, without working data loader federation makes no sense.

After taking a closer look it seems that problem is caused by `await resolver.Resolve(resolveContext)` in resolve reference, and indeed why should we await every and each reference instead of collecting all tasks and wait for them all?

I have added an test which to prevent regression on this with corresponding changes

BR
Alex

### DIFF
--- a/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
+++ b/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
@@ -207,14 +207,13 @@ type User @key(fields: ""id"") {
 
             var expected = @"{ ""_entities"": [{ ""__typename"": ""User"", ""id"" : ""1"", ""username"": ""One"" }, { ""__typename"": ""User"", ""id"" : ""2"", ""username"": ""Two"" }] }";
 
-            var executionResult = Executer.ExecuteAsync(_ =>
+            AssertQuery(_ =>
             {
-                _.Schema = Builder.Build(definitions);
+                _.Definitions = definitions;
                 _.Query = query;
+                _.ExpectedResult = expected;
                 _.Listeners.Add(listener);
-            }).GetAwaiter().GetResult();
-
-            Assert.Equal(CreateQueryResult(expected).Data, executionResult.Data);
+            });
         }
     }
 }
--- a/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
+++ b/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
@@ -1,7 +1,9 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Json;
 using System.Threading.Tasks;
+using GraphQL.DataLoader;
 using GraphQL.Utilities.Federation;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -156,5 +158,63 @@ public void input_types_and_types_without_key_directive_are_not_added_to_entitie
...
+                        ... on User {
+                            id
+                            username
+                        }
+                    }
+                }";
+
+            var expected = @"{ ""_entities"": [{ ""__typename"": ""User"", ""id"" : ""1"", ""username"": ""One"" }, { ""__typename"": ""User"", ""id"" : ""2"", ""username"": ""Two"" }] }";
+
+            var executionResult = Executer.ExecuteAsync(_ =>
```
joemcbride, line 210: I think this test should be able to be simplified by using `AssertQuery` vs. the `Executor` directly?

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -156,5 +158,63 @@ public void input_types_and_types_without_key_directive_are_not_added_to_entitie
...
+            {
+                _.Schema = Builder.Build(definitions);
+                _.Query = query;
+                _.Listeners.Add(listener);
+            }).GetAwaiter().GetResult();
+
+            Assert.Equal(CreateQueryResult(expected).Data, executionResult.Data);
```
joemcbride, line 217: Instead of using XUnit `Assert`, this project instead tries to use `Shouldly` assertions where it can.

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
mac2000: Did tried this in first place, but because of `AssertQuery` lack of `Listeners` it was not possible, in a next commit I have added required `Listeners` to `ExecuteConfig` which solves both comments

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
mac2000: Did removed it in favor to `AssertQuery`

ShadyNawara:
just tried this PR, works great. thanks @mac2000

Shane32:
I do not believe this patch / bug fix is necessary with the enhanced dataloaders in PR #1511 .  I've added the test here into PR #1511 and it runs flawlessly, albeit with a necessary change to the test due to the revised dataloader design.  However, I've never used federation, so I'm not sure exactly what we are testing.  My guess is that the `IDataLoaderResult<T>` gets passed through the federation code back to the execution strategy, which loads the data loader at the appropriate time, which is why no additional changes were necessary to fix this bug with #1511 .

As I am unfamiliar with federation, perhaps the underlying bug in this PR would be good to fix regardless of whether dataloaders within #1511 require this patch.

mac2000:
This patch is not about dataloader itself actually all we do in it is not awaiting for each concrete entity e.g. incoming requests are something like this:

```
{
  _entities(representations:[{__typename:"Account",id:"1"},{__typename:"Account",id:"2"}]) {
    ... on Account {
      id
      name
    }
  }
}
```

as you can see we are trying to get id and name of accounts with id 1 and 2, original code did await each concrete account

hope that helps

Shane32:
One other question: does this modified behavior need to change based on whether it's in a serial execution strategy or parallel execution strategy?  Sorry if this is a dumb question; I don't use federation in my graphs.

mac2000:
@Shane32 shouldnt be, it will be executed as usual query (not mutation) so should be executed as any other query, e.g.:

```gql
{
  getAccount(id:1) { id, name }
  getAccount(id:2) { id, name }
}
```

is equivalent to

```gql
{
  _entities(representations:[{__typename:"Account",id:"1"},{__typename:"Account",id:"2"}]) {
    ... on Account {
      id
      name
    }
  }
}
```

`Query._entities` is just kind of "generic" resolver where input is `T {__typename, id}`

Shane32:
The test in this PR has been added within #1511 and now passes successfully with the revised data loaders.

PR #1663, opened 2020-05-04
Add netstandard2.0 target for GraphQL.SystemTextJson

### LINKED ISSUE #1661: Support netstandard in the GraphQL.SystemTextJson package
## Summary

Currently, the `GraphQL.SystemTextJson` package can only be used in dotnetcore3.0/3.1. However the `System.Text.Json` nuget package (https://www.nuget.org/packages/System.Text.Json/) supports netstandard as well. Would it be possible to allow netstandard (and framework) as target frameworks of the `GraphQL.SystemTextJson` package?

sungam3r:
fixes #1661 
@benmccallum I also removed netcoreapp3.1 target since it is not required.

### DIFF
--- a/src/GraphQL.SystemTextJson/JsonConverterBigInteger.cs
+++ b/src/GraphQL.SystemTextJson/JsonConverterBigInteger.cs
@@ -16,6 +16,14 @@ namespace GraphQL.SystemTextJson
         public override BigInteger Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
             => TryGetBigInteger(ref reader, out var bi) ? bi : throw new JsonException();
 
+#if NETSTANDARD2_0
+        public static bool TryGetBigInteger(ref Utf8JsonReader reader, out BigInteger bi)
+        {
+            var byteArray = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan.ToArray();
+            var str = Encoding.UTF8.GetString(byteArray);
+            return BigInteger.TryParse(str, out bi);
+        }
+#else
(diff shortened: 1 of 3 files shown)


src/GraphQL.SystemTextJson/JsonConverterBigInteger.cs
```diff
@@ -16,13 +16,22 @@ public sealed class JsonConverterBigInteger : JsonConverter<BigInteger>
         public override BigInteger Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
             => TryGetBigInteger(ref reader, out var bi) ? bi : throw new JsonException();
 
+#if NETSTANDARD2_0
+        public static bool TryGetBigInteger(ref Utf8JsonReader reader, out BigInteger bi)
+        {
+            var byteArray = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan.ToArray();
+            var str = Encoding.UTF8.GetString(byteArray);
+            return BigInteger.TryParse(str, out bi);
```
sungam3r, line 24: Unfortunately, in the case of netstandard, we have to use methods that allocate memory.

benmccallum:
Just thought, I guess this could also be done in the server project potentially? Would also need an update to the readme. 

Might not be worth the effort as much though. Until someone asks for instance.

sungam3r:
@benmccallum Yes, it could but we have a small problem in `DeserializeFromJsonBodyAsync` for netstandard2.0.
изображение

If you recall the name of the project (AspNetCore) and other problems with .NET Framework support (WebSockets) then this idea seems pointless.

PR #1669, opened 2020-05-07
[proposal] code first federation

mac2000:
Have prepared small video (i do belive that it is better see once rather than read many times)

[demo](https://www.youtube.com/watch?v=Cdzo7HU_r-c)

# Code First Federation

- [ ] directives
- [ ] _Service
- [ ] _Entity
- [ ] Query._service
- [ ] Query._entities
- [ ] resolve reference
- [ ] tests

We already have an schema first federation implementation and going to reuse as much as possible of its code in code first aproach

## directives

1. We do not have real directives at moment in graphql-dotnet
2. We do not need real directives for federation to work
3. We do need print directives in Query._service for gateway to be able to understand our graph (already have this code in `FederatedSchemaPrinter`)

**How does it work in schema first implementation:** `FederatedSchemaPrinter` expecting fields to have `AST_METAFIELD` and types to have `EXTENSION_AST_METAFIELD` with `GraphQLObjectTypeDefinition` inside, it does use this metadata to print directives in `_serivce { sdl }`

**Proposal:** instead of trying to provide real directives which in either case not required for federation to work (e.g. they are nor runtime nor generation) lets provide the same metadata

For this to work we will have to provide few `FieldBuilder` extensions methods which underneath will construct and set required metadata, e..g. in my code it looks like:

```csharp
public class AccountType: ObjectGraphType<Account> {
    public AccountType() {
        Field(x => x.Id).External();
    }
}
```

The same approach might be done for types, e.g. we can:

- make ObjectGraphType extensions to add key and extend directives
- make FederatedObjectGraphType which will hold this methods

# _Service & Query._service

There is already `ServiceType` defined which can be totally reused, underneath while resolving `sdl` it does creates `FederatedSchemaPrinter` and prints schema from context

All we have to do is to `RegisterType<ServiceGraphType>();` in our schema and inside our `Field<NonNullGraphType<ServiceGraphType>>().Name("_service").Resolve(context => new { });` inside our Query and everything will start to work

 # _Entity

In schema first aproach entities are generated on the fly, I did not found a way to do it dynamically and not really sure whether it should be implicit vs explicit

In my case I do have:

```csharp
public class EntityType : UnionGraphType
{
    public EntityType()
    {
        Name = "_Entity";
        Type<ArticleType>();
        Type<AccountType>();
    }
}
```

which is registered as type in my schema

# Query._entities

This one can be request from `FederatedSchemaBuilder` as is

# resolve references

In my case I have created

```csharp
public class FederatedObjectGraphType<TSourceType> : ObjectGraphType<TSourceType> {
  public void Key(string fields) => this.BuildAstMeta("key", fields);
  public void ExtendByKey(string fields) {
    this.BuildExtensionAstMeta("key", fields);
    Key(fields);
  }

  public void ResolveReferenceAsync(Func<FederatedResolveContext, Task<TSourceType>> resolver) => ResolveReferenceAsync(new FuncFederatedResolver<TSourceType>(resolver));

  public void ResolveReferenceAsync(IFederatedResolver resolver) {
    // Metadata[FederatedSchemaBuilder.RESOLVER_METADATA_FIELD] = resolver;
    Metadata["__FedResolver__"] = resolver;
  }
}
```

Which allows me to have something like this:

```csharp
public class AccountType: FederatedObjectGraphType<Account>
{
    public AccountType(InMemoryStorage storage)
    {
        ExtendByKey("id");
        Name = nameof(Account);
        
        Field(x => x.Id, type: typeof(NonNullGraphType<IdGraphType>)).External();

        Field<NonNullGraphType<ListGraphType<ArticleType>>>().Name("articles").Resolve(context => storage.articles.Where(article => article.AccountId == context.Source.Id));

        ResolveReferenceAsync(context => {
            var id = context.Arguments["id"].ToString();
            var account = new Account { Id = id };
            return Task.FromResult(account);
        });
    }
}
```

I have tested on following sample, where we have

**Accounts Service**

```graphql
type Account key(fields:"id") {
  id: ID!
  name: String!
  username: String!
}

type Query {
  account(id: ID!): Account
  accounts: [Account]!
}
```

**Articles Service**

```graphql
type Article key(fields:"id") {
  id: ID!
  title: String!
  account: Account!
}

extend type Account @key(fields: "id") {
  id: ID! @external
  articles: [Article]!
}
```

which are compatible with apollo gateway and apollo graph manager, e.g. can be connected to gateway and run queries like:

```
{
  article(id:"1") {
    title
    account {
      username
      articles {
        title
      }
    }
  }
  
  account(id:"1") {
    username
    articles {
      title
    }
  }
}
```

# Questions

- **Whether it is ok to reuse same metadata instead of trying to implement real directives - if so - can be first step**
- For me as a user will be nice to have something like `FederatedQuery` which will hide both `Query._service` and `Query._entities` implementations, also `FederatedSchema` which will hide  registration of required types
- Not sure which way to enforce `EntityType`
- Not sure how to enforce existance of `ResolveReference` in graph types

### DIFF
--- a/docs2/site/docs/guides/federation.md
+++ b/docs2/site/docs/guides/federation.md
@@ -1,0 +1,6 @@
+# Apollo Federation
+
+GraphQL .NET includes an implementation of [Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/).
+
+Federation works best when you have two or more teams working on two or more services
+
--- a/docs2/site/docs/guides/federation.md
+++ b/docs2/site/docs/guides/federation.md
@@ -1,6 +1,5 @@
 # Apollo Federation
 
-GraphQL .NET includes an implementation of [Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/).
+GraphQL.NET includes an implementation of [Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/).
 
 Federation works best when you have two or more teams working on two or more services
-
--- a/docs2/site/docs/guides/federation.md
+++ b/docs2/site/docs/guides/federation.md
@@ -2,4 +2,4 @@
 
 GraphQL.NET includes an implementation of [Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/).
 
-Federation works best when you have two or more teams working on two or more services
+Federation works best when you have two or more teams working on two or more services.
--- a/src/GraphQL/Utilities/Federation/Proposal.cs
+++ b/src/GraphQL/Utilities/Federation/Proposal.cs
@@ -1,0 +1,243 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using GraphQL.Builders;
+using GraphQL.Types;
+using GraphQLParser.AST;
+
+namespace GraphQL.Utilities.Federation
+{
+    public static class MySchemaBuilderExtensions
+    {
+        /// <summary>
+        /// Internal helper to build GraphQLDirective instance which is then used by FederationSchemaPrinter while printing federated schema sdl
+        /// Directives are stored in fields metadata under key "__AST_MetaField__" and does not used anywhere else
+        /// Additional properties like Location are added to avoid null ref exceptions only
+        /// </summary>
+        /// <param name="name">name of the directive, e.g. key, extend, provides, requires</param>
+        /// <param name="value">value for fields argument, e.g. @key(fields: "id")</param>
+        /// <param name="kind">for field directives - ASTNodeKind.StringValue, for type directives - ASTNodeKind.Argument</param>
+        /// <returns></returns>
+        private static GraphQLDirective BuildGraphQLDirective(string name, string value = null, ASTNodeKind kind = ASTNodeKind.StringValue) => new GraphQLDirective
+        {
+            Name = new GraphQLName
+            {
+                Value = name,
+                Location = new GraphQLLocation()
+            },
+            Arguments = string.IsNullOrEmpty(value) ? new List<GraphQLArgument>() : new List<GraphQLArgument>() {
+                        new GraphQLArgument
+                        {
+                            Name = new GraphQLName {
+                                Value = "fields",
+                                Location = new GraphQLLocation()
+                            },
+                            Value = new GraphQLScalarValue(kind) {
+                                Value = value,
+                                Location = new GraphQLLocation()
+                            },
+                            Location = new GraphQLLocation()
+                        }
+                    },
+            Location = new GraphQLLocation()
+        };
+
+        /// <summary>
+        /// The same as above
+        /// </summary>
+        /// <returns></returns>
+        private static GraphQLObjectTypeDefinition BuildGraphQLObjectTypeDefinition() => new GraphQLObjectTypeDefinition
+        {
+            Directives = new List<GraphQLDirective>(),
+            Location = new GraphQLLocation(),
+            Fields = new List<GraphQLFieldDefinition>()
+        };
+
+        private static void AddDirective(GraphQLObjectTypeDefinition definition, GraphQLDirective directive) => ((List<GraphQLDirective>)definition.Directives).Add(directive);
+
+        /// <summary>
+        /// Public helper which will allow us to set "__EXTENSION_AST_MetaField__" on a type which then will be used by FederatedSchemaPrinter
+        /// </summary>
+        /// <param name="type"></param>
+        /// <param name="name"></param>
+        /// <param name="value"></param>
+        public static void BuildExtensionAstMeta(this IProvideMetadata type, string name, string value = null)
+        {
+            var definition = BuildGraphQLObjectTypeDefinition();
+            var directive = BuildGraphQLDirective(name, value, ASTNodeKind.Argument);
+            AddDirective(definition, directive);
+            // type.AddExtensionAstType(definition);
+            type.Metadata["__EXTENSION_AST_MetaField__"] = new List<ASTNode> { definition };
+        }
+
+        /// <summary>
+        /// Public helper which will allow us to set "__AST_MetaField__" on a field which then will be used by FederatedSchemaPrinter
+        /// </summary>
+        /// <param name="type"></param>
+        /// <param name="name"></param>
+        /// <param name="value"></param>
+        public static void BuildAstMeta(this IProvideMetadata type, string name, string value = null)
+        {
+            // var definition = type.GetAstType<GraphQLObjectTypeDefinition>() ?? BuildGraphQLObjectTypeDefinition();
+            var definition = (GraphQLObjectTypeDefinition)type.GetMetadata<ASTNode>("__AST_MetaField__", () => BuildGraphQLObjectTypeDefinition());
+            var directive = BuildGraphQLDirective(name, value);
+            AddDirective(definition, directive);
+            //type.SetAstType(definition);
+            type.Metadata["__AST_MetaField__"] = definition;
+        }
+    }
+
+    /// <summary>
+    /// FieldBuillderExtensions, provide shortcuts for filling field metadata with expected objects, e.g. instead of:
+    /// Field().Name("foo").FieldType.Metadata["__AST_MetaField__"] = new GraphQLObjectTypeDefinition() {...};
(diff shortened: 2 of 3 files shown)


docs2/site/docs/guides/federation.md
```diff
@@ -0,0 +1,6 @@
+# Apollo Federation
+
+GraphQL .NET includes an implementation of [Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/).
```
sungam3r, line 3:
```suggestion
GraphQL.NET includes an implementation of [Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/).
```

docs2/site/docs/guides/federation.md
```diff
@@ -0,0 +1,6 @@
...
+
+Federation works best when you have two or more teams working on two or more services
```
sungam3r, line 5:
```suggestion
Federation works best when you have two or more teams working on two or more services.
```

sungam3r:
I watched your video, thanks. Federation is a big topic that I am not ready to discuss yet. I imagine how it all works, but did not dive deeply. No doubt we need code first federation.

joemcbride:
Thanks @mac2000, this look like a fairly straight forward implementation and is possibly a route we can go.  I would like to see more of the code you have written.

I think a "nice to have" feature would be not requiring the user to build the `_Entity` type themselves.  I think this could be accomplished by building that type when the Schema is initialized, As is done in the [`FederatedSchemaBuilder`](https://github.com/graphql-dotnet/graphql-dotnet/blob/68be167176daab0f8791fc6c6db6821803e83cec/src/GraphQL/Utilities/Federation/FederatedSchemaBuilder.cs#L135-L182) today

mac2000:
Totally agree with both of you guys, that was exactly a reason why I did not provide at once whole code, there is a questions to discuss before, like you have mentioned about `_Entity` to be fair I just did not realize yet how to do it and hope you will help.

My very first attempt was to kind of mimique schema first approach, it was pretty easy and straight forward for resolve reference metadata and not so easy for directives, I did create `MyFederationSchemaPrinter: FederationSchemaPrinter` and inside override `PrintFederatedDirectivesFromAst` which did allow me to do something like this:

```csharp
public class AccountType: ObjectGraphType<Account> {
  public AccountType() {
    Field<String>().Name("name").FieldType.Metadata.Add("provides", "name");
  }
}
```

So inside my schema printer I did rely on my simplified metadata, but after some time I did figure out that actually original approach just rely on given `GraphQLObjectTypeDefinition` which is not used anywhere except printing federated schema sdl and can be created, e.g. I can do something like this:

```csharp
public class AccountType: ObjectGraphType<Account> {
  public AccountType() {
    Field<String>().Name("name").FieldType.Metadata.Add("__AST_MetaField__", new GraphQLObjectTypeDefinition {...});
  }
}
```

and put it to `FederatedSchemaPrinter` and everything will work as expected

Note: that is what I mean when saying that there is no need for real directives for federation to work, all we need is to print federated schema sdl, my idea here is to stick with currently implemented federated schema printer and in future when there will be directives support it will be easier to switch implementation

That is the main concern which I not sure about, everything else really a matter of preference and can be done via extensions or extending existing classes, so if you are ok with that, I might try commit some extensions around that field and we can start moving from that

Or I can put everything at once so you will be able to kind of see whole picture and then we will remove pieces one by one

mac2000:
Actually after looking again, it is just 100 lines of code, more than half of them will be removed (reused from schema first approach), the biggest piece is around extensions to make our live easier for providing metadata

I did commit `/src/GraphQL/Utilities/Federation/Proposal.cs` so you guys can take a look in overall how it looks like

Please remember that this file was written outside code base, so I were forced to copy some stuff like _entities resolvers and hardc ode metadata names

mac2000:
Seems like I have broke approved public api but for now it is probably ok, no matter what i did test not working except me commenting everything

BTW recorded an [video](https://youtu.be/PpGpvB1iK9s) for my teammates about graph manager proving it does work (sorry it is in Russian but I bed that from image you will understand what is going on) it is optional and the same can be achieved without graph manager using tools like [graphql-inspector](https://github.com/kamilkisiela/graphql-inspector)

whymatter:
What is the status of this? I would like to use federation with the code-first approach and now I am wondering why this is on hold for almost 10 months?

Is there anything I can help with?

sungam3r:
Federation is out of scope now for v4 release. Moreover I did not work with federation before. Ping @joemcbride for help.

src/GraphQL/Utilities/Federation/Proposal.cs
```diff
@@ -0,0 +1,243 @@
...
+    /// Can be replaced with ObjectGraphType extensions, the same as above - idea to allow us to define type directives, e.g.:
+    /// public class AcmeType: ObjectGraphType {
+    ///   public AcmeType() {
+    ///     // Metadata["__AST_MetaField__"] = new GraphQLObjectTypeDefinition() {...};
+    ///     Key("id");
+    ///   }
+    /// }
+    ///
+    /// TODO:
+    /// - there is only two cases: `type Account @key(fields:"id")` and `extend type Account @key(fields:"id")`
```
cyberhck, line 120: hey @mac2000 
can you give me a general idea on how I can do the `extend type Account {}` in code first? I tried a bunch of things and couldn't succeed.

src/GraphQL/Utilities/Federation/Proposal.cs
```diff
@@ -0,0 +1,243 @@
...
+            RegisterValueConverter(new AnyValueConverter());
+            RegisterType<AnyScalarGraphType>();
+            RegisterType<ServiceGraphType>();
+            RegisterType<EntityType>();
+        }
+    }
+
+    /// <summary>
+    /// TODO:
+    /// - implicit vs explicit
```
cyberhck, line 175: how could we do this implicit way? would this mean using reflection?

src/GraphQL/Utilities/Federation/Proposal.cs
```diff
@@ -0,0 +1,243 @@
...
+    /// </summary>
+    public class FederatedSchemaProposal : Schema
+    {
+        public FederatedSchemaProposal() : this(new DefaultServiceProvider()) { }
+        public FederatedSchemaProposal(IServiceProvider serviceProvider) : base(serviceProvider)
+        {
+            RegisterValueConverter(new AnyValueConverter());
+            RegisterType<AnyScalarGraphType>();
+            RegisterType<ServiceGraphType>();
+            RegisterType<EntityType>();
```
cyberhck, line 169: do we need a any value converter? I tried without one and it still did work :| I tried that because I couldn't find source for any value converter I also couldn't find it on schema first (same as RegisterValueConverter)

cyberhck: also if we were to put this directly on Schema instead of having federation on a separate class, would that cause any issues?

v12:
Is this functionality planned to be included in the v5 release?

sungam3r:
I can not say for sure. A set of changes for v5 continues to expand. v5 will have such changes that I first did not suspect, for example, new complexity analizer as a new validation rule. Also `SchemaPrinter` (and `FederatedSchemaPrinter` I think) will be fully rewritten. I don't use federation but it seems that I have to understand it because it's required by other changes that I want to do. So the formal answer is - _no, there are no such plans_ ... but it never turns out to say exactly what happens in the end.

Mithras:
@sungam3r 
I've done a lot of refactoring after my initial federation attempt so the git repo I shared above is outdated and have a few bugs. Currently we successfully use federation in production with a few C# microservices and Apollo Gateway in front of them. If/When you have time to work on implementing federation support in graphql-dotnet and/or interested, let me know and I can share pieces of code we use to implement it. I just don't want to create a PR which will be stuck in review for years like this one.

sungam3r:
In front of the v5 release we will definitely look through all the current posted PRs and decide which of them we want to work on or postpone. Then I will give to know.

Shane32:
Federation 2 support is available for schema-first, code-first, and type-first schemas in GraphQL.NET 8.0 (still in beta at the time of writing). See:

- https://github.com/graphql-dotnet/graphql-dotnet/pull/3921

Thanks for everyone's contributions to make this a reality!  And please open an issue if anything does not work or if there are potential avenues for improvement.

PR #1679, opened 2020-05-14
Allow full manipulation of Directives in AST node

kzu:
This was one of the few (only?) exceptions to the AST being
full read-write, and prevented some processing of docs to
sanitize directives, for example, prior to execution.

There was really no apparent reason to introduce this limitation
at the AST level, so we implement `ICollection<Directive>` to
allow full mutation of this collection.

### DIFF
--- a/src/GraphQL/Language/AST/Directives.cs
+++ b/src/GraphQL/Language/AST/Directives.cs
@@ -33,7 +33,7 @@ namespace GraphQL.Language.AST
 
         public bool HasDuplicates => _directives?.Count != _unique.Count;
 
-        public bool IsReadOnly => ((ICollection<Directive>)_directives).IsReadOnly;
+        public bool IsReadOnly => false;
 
         public IEnumerator<Directive> GetEnumerator()
         {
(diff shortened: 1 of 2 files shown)


src/GraphQL/Language/AST/Directives.cs
```diff
@@ -33,6 +33,8 @@ public void Add(Directive directive)
 
         public bool HasDuplicates => _directives?.Count != _unique.Count;
 
+        public bool IsReadOnly => ((ICollection<Directive>)_directives).IsReadOnly;
```
sungam3r, line 36:
```suggestion
        public bool IsReadOnly => false;
```

src/GraphQL/Language/AST/Directives.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace GraphQL.Language.AST
 {
-    public class Directives : AbstractNode, IEnumerable<Directive>
+    public class Directives : AbstractNode, IEnumerable<Directive>, ICollection<Directive>
```
sungam3r, line 7:
```suggestion
    public class Directives : AbstractNode, ICollection<Directive>
```

SimonCropp:
LGTM with @sungam3r suggestions

kzu:
Feel free to squash, of course. I accepted via the web, seemed reasonable :).

PR #1680, opened 2020-05-14
Added IProvideResolvedType

JKamsker:
I think its pretty self-explainatory

### DIFF
--- a/src/GraphQL/Types/Composite/ListGraphType.cs
+++ b/src/GraphQL/Types/Composite/ListGraphType.cs
@@ -2,7 +2,7 @@ using System;
 
 namespace GraphQL.Types
 {
-    public class ListGraphType<T> : ListGraphType
+    public class ListGraphType<T> : ListGraphType, IProvideResolvedType
         where T : IGraphType
     {
         public ListGraphType()
--- a/src/GraphQL/Types/Composite/ListGraphType.cs
+++ b/src/GraphQL/Types/Composite/ListGraphType.cs
@@ -2,7 +2,7 @@ using System;
 
 namespace GraphQL.Types
 {
-    public class ListGraphType<T> : ListGraphType, IProvideResolvedType
+    public class ListGraphType<T> : ListGraphType
         where T : IGraphType
     {
         public ListGraphType()
@@ -11,7 +11,7 @@ namespace GraphQL.Types
         }
     }
 
-    public class ListGraphType : GraphType
+    public class ListGraphType : GraphType, IProvideResolvedType
     {
         public ListGraphType(IGraphType type)
         {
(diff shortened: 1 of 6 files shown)


src/GraphQL/Types/IProvideResolvedType.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
```
sungam3r, line 4:
```suggestion
```

src/GraphQL/Types/Composite/ListGraphType.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace GraphQL.Types
 {
-    public class ListGraphType<T> : ListGraphType
+    public class ListGraphType<T> : ListGraphType, IProvideResolvedType
```
sungam3r, line 5: Why not on `ListGraphType` ?

sungam3r: Same question for `NonNullGraphType`.

sungam3r:
Yes, it is self-explainatory but what is your case/need in this interface?

JKamsker:
Cant tell right now exactly but for some reason i needed to get the root type.

This comment will be updated as ive researched the root cause for my need. @sung3r

JKamsker:
The reason i needed this, is because i want all fieldtypes from an IGraphType:

```csharp
private IEnumerable<FieldType> GetFieldTypes(IGraphType returnType = null)
{
	if (returnType == null)
	{
		returnType = this.source.ReturnType;
	}

	if (returnType is IComplexGraphType cgt)
	{
		return cgt.Fields;
	}

	if ((returnType is ListGraphType || returnType is NonNullGraphType) && returnType is IProvideResolvedType gt)
	{
		return this.GetFieldTypes(gt.ResolvedType);
	}

	throw new NotImplementedException($"Cannot resolve returntype of '{returnType.GetType().FullName}'");
}
```

src/GraphQL/Types/Composite/ListGraphType.cs
sungam3r: @JKamsker  ?

JKamsker:
https://github.com/graphql-dotnet/graphql-dotnet/pull/1680/files#diff-ca5a56c1707ab2a0a55aaf8404e2b59eR1957

src/GraphQL/Types/Composite/ListGraphType.cs
sungam3r: I mean to inherit `ListGraphType` and `NonNullGraphType` from `IProvideResolvedType`, not `ListGraphType<T>` and `NonNullGraphType<T>>`.

sungam3r:
@JKamsker I see that you change PR but for some reason github shows it in the past. It seems to me that I have not encountered this before. All force pushes always displayed.

sungam3r:
AppVeyor build also shows 2 monts ago date. No matter, merging.

PR #1685, opened 2020-05-20
Update schema-types.md

tony-asu:
ByteGraphType is not in the GraphQL.net project.


sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/Scalars/ByteGraphType.cs

Where is it then?

sungam3r:
Honestly, I am surprised at such a PR. Perhaps I misunderstood something.

tony-asu:
i only did this pr due to it being missing from the latest release version of the code, which is what the rest of the documentation is using.

sungam3r:
Current documentation is written for 3.0.0-preview-XXXX versions.

PR #1690, opened 2020-05-22
Add a RelayBatchDataLoader to existing DataLoaders

karrocon:
### PULL REQUEST DESCRIPTION
This is a proposal for a new dataloader based on the Relay specification. The proposed code will be used to ease the development of GraphQL Relay servers by providing the required page information to the resolving (typically infrastructure) layer.

### RELATED TOPIC
In my opinion, using EntityFrameworkCore is the most natural way of resolving GraphQL .NET server requests. However, with the current solution, it is quite painful to fetch data from a DataSet<T> with a Relay implementation. To release the stress, I have coded a ToConnectionDictionary method that is an extension of IQueryable<T> with the following heading:
```
private static async Task<IDictionary<TKey, Connection<T>>> ToConnectionDictionary<T, TKey>(this IQueryable<T> query, IEnumerable<PageRequest<TKey>> pageRequests, Expression<Func<T, TKey>> keySelector, Expression<Func<T, string>> cursorSelector)
```
This method fits perfectly with the new RelayBatchDataLoader and relieves the pain of correctly coding it.

### DIFF
--- a/src/GraphQL/DataLoader/DataLoaderContextExtensions.cs
+++ b/src/GraphQL/DataLoader/DataLoaderContextExtensions.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using GraphQL.Types.Relay.DataObjects;
 
 namespace GraphQL.DataLoader
 {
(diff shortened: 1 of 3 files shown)


Shane32:
I suggest this PR be revised and reviewed after #1511 as the latter PR will slightly change the code in this PR.

Shane32:
I feel that we should consider moving this functionality to the Relay project.  It may not be needed (or be able to be greatly simplified) now that the Relay project supports `IQueryable`.

sungam3r:
Also I would like to move relay graph types into relay projects.

sungam3r:
Everything from here
изображение
\+ connection builder

PR #1693, opened 2020-05-23
Make sure error paths prefer aliases

ificator:
According to [the GraphQL spec](http://spec.graphql.org/June2018/#sec-Errors) the path in an error, when a field is aliased, should include the alias and not the name. Specifically:
```
If the error happens in an aliased field, the path to the error should use the aliased name, since it represents a path in the response, not in the query.
```
My first thought was to simply change `ExecutionNode.Path` to prefer the alias since it is only used to populate error paths, however I quickly decided that could be a breaking change for consumers that are accessing the path in their resolvers and are expecting the "query" path.

So instead I've implemented a new `ResponsePath` property that is now used to populate `ExecutionError.Path`. It's also exposed in the `ResolveFieldContext` for any resolvers that want the path to contain aliases as well.

### DIFF
--- a/src/GraphQL.Tests/Execution/ExecutionNodeTests.cs
+++ b/src/GraphQL.Tests/Execution/ExecutionNodeTests.cs
@@ -1,0 +1,91 @@
+using System.Collections.Generic;
+using System.Linq;
+using GraphQL.Execution;
+using GraphQL.Types;
+using Xunit;
+
+using AST = GraphQL.Language.AST;
+
+namespace GraphQL.Tests.Execution
+{
+    public class ExecutionNodeTests
+    {
+        [Fact]
+        public void Path_Alias()
+        {
+            var objectGraphType = new AliasedFieldTestObject();
+
+            var node = new ValueExecutionNode(
+                new RootExecutionNode(objectGraphType),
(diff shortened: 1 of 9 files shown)


src/GraphQL.Tests/Execution/ExecutionNodeTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            var node = new ValueExecutionNode(
+                new RootExecutionNode(objectGraphType),
+                new StringGraphType(),
+                new AST.Field(new AST.NameNode("alias"), new AST.NameNode("name")),
+                objectGraphType.GetField("value"),
+                indexInParentNode: null);
+
+            var path = node.Path.ToList();
+            string pathSegment = Assert.Single(path) as string;
```
joemcbride, line 26: This project uses `Shouldly` for assertions.  Would you mind updating these tests to use that?

sungam3r: fixed

sungam3r:
@ificator Thank you for fixing this. I resolved conflicts and made a few changes.

ificator:
Thanks folks!

Issue #1326, opened 2019-10-05
Known issues documentation


sungam3r:
Add documentation (maybe like FAQ) about known issues:
1. ~AllowSynchronousIO: #1116, #1299, #1325, graphql-dotnet/server#271, #1317, graphql-dotnet/server#277~
2. ~DI: Scoped services issue: #1345, #1160 and related issues~
3. ~Auto generating classes and other stuff #576~
4. ~Auto registering output and input types #897~
5. Using derived graph types and cyclic dependencies #1010  -   **too narrow to document, cancelled**
6. ~File downloading #511~
7. ~Note about two authorization models - generic and ASP.NET Core graphql-dotnet/authorization#68 graphql-dotnet/authorization#74~
8. ~Note about descriptions inheritance #1004~ 
9. ~Recommendations to always use singleton schema and graph types over scoped ones #1011~
10. ~Many to many with dataLoader #1625~j
11. ~Subscription Authentication #1244~
12. ~Flags enums #1666~
13. ~`null` in scalars #1542~

corstian:
On a related note; is there someone who can trigger a rebuild of the documentation? I have the feeling the currently published docs are several months behind.

sungam3r:
Yes, docs are outdated. I think this is done intentionally so that the documentation matches the latest stable version 2.4.0.

joemcbride:
>  I think this is done intentionally so that the documentation matches the latest stable version 2.4.0.

No, I would be fine deploying the latest.  The gatsby website needs to be updated before it can be deployed.

sgebhardt:
> > I think this is done intentionally so that the documentation matches the latest stable version 2.4.0.
> No, I would be fine deploying the latest. The gatsby website needs to be updated before it can be deployed.

Can you tell what version the Docs are targeting? At least the part about `[dependency-injection](https://graphql-dotnet.github.io/docs/getting-started/dependency-injection#dependency-injection)` seems not to target version 2.4.0.

sungam3r:
Documentation targets `develop` branch, version 3.0.0-preview-xxxx.

SimonCropp:
@sungam3r would a possible (more maintainable) solution to this be a GH label and point to a search query for open issues with that label?

sungam3r:
I thought about it, but chose the option with a separate issue with a small description on each item.

sungam3r:
I will be grateful if someone can contribute. @Shane32 ?

[...]

PR #1694, opened 2020-05-26
Docs - Adding Scoped Services to dependency-injection.md

michal-ciechan:
Relating to #1326, adding some information on using Scoped Services in GraphQL. 

Hopefully will help someone out there without needing to go through GitHub issues.

### DIFF
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -110,3 +110,128 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
     );
 }
 ```
+
+# Scoped Services
+
+To use scoped services (e.g. HttpContext Scoped services in ASP.NET Core) you will either need to
+
+* use [SteroidsDI](https://github.com/sungam3r/SteroidsDI)
+* register a scoped `IServiceProvider` in the `UserContext` before running `IDocumentExecuter` ([anti-pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/))
+* register a scope and provide custom Func<T> or factories (see how [SteroidsDI](https://github.com/sungam3r/SteroidsDI) is implemented)
+
+## SteroidsDI
+
+To use [SteroidsDI](https://github.com/sungam3r/SteroidsDI) with ASP.NET Core, add `Defer<>` and `IScopeProvider` in your `Startup.ConfigureServices`:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    ...
+
+    // Add SteroidsDI Open Generic Defer<> Factory Class
+    services.AddDefer();
+
+    // Add SteroidsDI IScopeProvider to use the AspNetCoreHttpScopeProvider
+    // which internally uses the IHttpContextAccessor.HttpContext.RequestServices;
+    services.AddHttpScope();
+
+    ...
+}
+```
+
+Then in your query graph types you can request services using `Defer<T>` to be injected via DI,
+ which will be evaluated at runtime to their relevant Scoped services, `T` has been registered with a `Scoped` DI lifetime:
+
+```csharp
+public class StarWarsQuery : ObjectGraphType
+{
+  // #1 - Add dependecies using Defer<T>
+  public StarWarsQuery(Defer<IDroidRepo> repoFactory)
+  {
+    Field<DroidType>(
+      "hero",
+
+      // #2 Resolve dependeiceis using current scope provider
+      resolve: context => repoFactory.Value.GetDroid("R2-D2")
+
+    );
+  }
+}
+```
+
+1. Add `Defer<T>` to be injected by the dependency injection container. This is a factory which upon calling `Defer.Value` will resolve the requested service using any currently registered scope provider (e.g. `AspNetCoreHttpScopeProvider`)
+2. Use the `Defer<T>` factory class to resolve the requested dependency using any currently registered scope provider. In our case it will attempt to use the `IHttpContextAccessor.HttpContext.RequestServices` which is the ASP.NET Core Scoped `IServiceProvider` in order to resolve the dependency.
+
+## UserContext Scoped IServiceProvider
+
+**Be Aware**: Exposing the service locator is considered an Anti-Pattern. See [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/)
+
+You can add the relevant scoped `IServiceProvider` when creating the `UserContext` in order to run `IDocumentExecuter`.
+
+E.g. In ASP.NET Core [GraphQl.Harness](https://github.com/graphql-dotnet/graphql-dotnet/tree/master/src/GraphQL.Harness) example, you could add `IServiceProvider` to the [GraphQLUserContext](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.Harness/GraphQLUserContext.cs) class, and then set it to the `HttpContext.RequestServices` when configuring the `GraphQLSettings.BuildUserContext`
+
+
+1. Add ServiceProvider to the UserContext
+
+    ```csharp
+    public class GraphQLUserContext: Dictionary<string, object>
+    {
+        ...
+        IServiceProvider ServiceProvider { get; set; }
+        ...
+    }
+    ```
+
+2. Set it the the HttpContext.RequestServices
+
+    ```csharp
+    public void ConfigureServices(IServiceCollection services)
+    {
+        services.Configure<GraphQLSettings>(settings => settings.BuildUserContext = ctx => new GraphQLUserContext
+        {
+            User = ctx.User,
+            ServiceProvider = ctx.RequestServices
+        });
+    }
+    ```
+
+3. Use it in resolvers
+
+    ```csharp
+    public class StarWarsQuery : ObjectGraphType<object>
+    {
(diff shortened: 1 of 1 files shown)


docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -110,3 +110,128 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
...
+}
+```
+
+Then in your query graph types you can request services using `Defer<T>` to be injected via DI,
+ which will be evaluated at runtime to their relevant Scoped services, `T` has been registered with a `Scoped` DI lifetime:
+
+```csharp
+public class StarWarsQuery : ObjectGraphType
+{
+  // #1 - Add dependecies using Defer<T>
```
sungam3r, line 148:
```suggestion
  // #1 - Add dependencies using Defer<T>
```

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -110,3 +110,128 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
...
+  public StarWarsQuery(Defer<IDroidRepo> repoFactory)
+  {
+    Field<DroidType>(
+      "hero",
+
+      // #2 Resolve dependeiceis using current scope provider
```
sungam3r, line 154:
```suggestion
      // #2 Resolve dependencies using current scope provider
```

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -110,3 +110,128 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
...
+    ```csharp
+    public class StarWarsQuery : ObjectGraphType<object>
+    {
+        public StarWarsQuery()
+        {
+            ...
+            Field<CharacterInterface>("hero",
+                resolve: context =>
+                    ((GraphQLUserContext)context.UserContext)
+                    .GetRequiredService<IDroidRepo>()
```
sungam3r, line 209:
```suggestion
                    .ServiceProvider.GetRequiredService<IDroidRepo>()
```

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -110,3 +110,128 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
...
+        }
+    }
+    ```
+
+ You could always create an extension method:
+
+ ```csharp
+public static class GraphQLExtensions
+{
+    public static T GetRequiredService<T>(this IResolveFieldContext<object> context) =>
```
sungam3r, line 223:
```suggestion
    public static T GetRequiredService<T>(this IResolveFieldContext context) =>
```

sungam3r:
@Shane32 Will you have any comments? It seems that @michal-ciechan described all the suitable ways that we talked about before.

Shane32:
Looking good. It doesn’t contain my most recent suggestion, which is having the extension method access the existing scoped service provider directly. (For ASP.Net, this is via the HttpContextAccessor). It’s the simplest example, meaning that it requires no setup.

sungam3r:
You are right. I remember that there was something else very simple. Сould you find that comment pointing to this extension and example?

michal-ciechan:
> Looking good. It doesn’t contain my most recent suggestion, which is having the extension method access the existing scoped service provider directly. (For ASP.Net, this is via the HttpContextAccessor). It’s the simplest example, meaning that it requires no setup.

Can I ask what was the suggestion?  

I can't see anyway to access `HttpContextAccessor` directly inside a resolver without either:

1. Adding it to a custom `UserContext` during Middleware (# 2. in the section I added)
2. Adding it as a an entry into an existing UserContext dictionary.

I do not see any static members on `HttpContextAccessor`, nor do MS mention anything in their [HttpContext docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-context?view=aspnetcore-3.1)

I am probably missing some static API surface somewhere?

Unless it is related to the advanced FieldMiddleware which I just came across on master, and can be scoped?

Shane32:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/1345#issuecomment-587107092 . An updated version of the snippet is below:
```csharp
public static class ContextExtensions
{
    public static T GetService<T>(this IResolveFieldContext context) {
        return ((IServiceProvider)context.Schema).GetRequiredService<IHttpContextAccessor>().HttpContext.RequestServices.GetRequiredService<T>();
    }
}
```
The code above requires no setup to utilize.  It's based on the fact that in the current version of GraphQL, the `Schema` class implements `IServiceProvider` to allow access to the root service provider.  This design pattern replaced a `Services` property which was exposed on the `Schema` class in prior versions.  Unfortunately `ISchema` does not yet inherit from `IServiceProvider`, so a cast is necessary, but typically the schema inherits from `Schema` anyway, and if the schema didn't implement `IServiceProvider`, middleware would cease to function (throwing a runtime error).  So it's a certainly a version-specific patch, but it works for now, and can easily be updated to retrieve from the `UserContext` if the design pattern were to change in the future.

The snippet also depends on the fact that the current `HttpContext` can be obtained via a singleton, (`IHttpContextAccessor`) and from it, the scoped service provider.  Almost all of the suggestions, other than utilizing the `UserContext`, are based upon this, of course.

I am still hopeful that in a future version of GraphQL, the current (scoped) service provider can be obtained from the `IResolveFieldContext`, much in the same way that the current service provider can be obtained from an `HttpContext`.  This would eliminate the dependency on `IHttpContextAccessor` and not require the service provider to be stashed within `UserContext`.  But this is not yet the case.

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -110,3 +110,128 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
...
+2. Set it the the HttpContext.RequestServices
+
+    ```csharp
+    public void ConfigureServices(IServiceCollection services)
+    {
+        services.Configure<GraphQLSettings>(settings => settings.BuildUserContext = ctx => new GraphQLUserContext
+        {
+            User = ctx.User,
+            ServiceProvider = ctx.RequestServices
+        });
```
Shane32, line 194: Note that the `BuildUserContext` property only exists within the `Harness` sample project, as the entire `GraphQLSettings` class is simply part of the sample.  This may be confusing to people that did not build their code from the sample (but it's pretty easy to discover that).

michal-ciechan: Yeah I noticed that, but I wanted to stay with the samples in this repo, and wasn't sure how to specify the intent without needing to provide a massive code sample. Even the `GraphQLMiddleware` as far as I am aware is only in the harness (I copied mine from there)

michal-ciechan:
> See [#1345 (comment)](https://github.com/graphql-dotnet/graphql-dotnet/pull/1345#issuecomment-587107092) . An updated version of the snippet is below:
> 
> ```cs
> public static class ContextExtensions
> {
>     public static T GetService<T>(this IResolveFieldContext context) {
>         return ((IServiceProvider)context.Schema).GetRequiredService<IHttpContextAccessor>().HttpContext.RequestServices.GetRequiredService<T>();
>     }
> }
> ```

Sorry, yes, this makes complete sense! Didn't think of using the root provider directly to get the `IHttpContextAccessor` which in turn returns the scoped context! Awesome suggestion. Thinking of making this the #1 option as it requires no additional setup? Are you OK with that @sungam3r  and @Shane32?

docs2/site/docs/getting-started/dependency-injection.md
michal-ciechan: I will add a note, specifying this.

Shane32:
I would suggest this as the #1 suggestion, as the point is to get people going and used to GraphQL.  Advanced users will understand the benefits and consequences of pulling directly from the service provider within the code.

docs2/site/docs/getting-started/dependency-injection.md
michal-ciechan: Added, I have split the extension method out, to make it clear to anyone using it, which interfaces/services it uses.

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -110,3 +110,167 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
...
+E.g. You can have the following extension method:
+
+```csharp
+public static class ContextExtensions
+{
+    public static T GetRequiredScopedService<T>(this IResolveFieldContext context)
+    {
+        var schema = (IServiceProvider) context.Schema;
+
+        var contextAccessor = schema.GetRequiredService<IHttpContextAccessor>();
```
Shane32, line 135:
```suggestion
        var rootServiceProvider = (IServiceProvider)context.Schema;

        var contextAccessor = rootServiceProvider.GetRequiredService<IHttpContextAccessor>();
```

Shane32: As you are casting the schema to `IServiceProvider`, I suggest that you name the variable `rootServiceProvider`, to more accurately represent the object interface.

michal-ciechan: Makes sense! Thanks! Initially was doing the case on the context accessor line, forgot to look at the variable name after moving it up.

sungam3r:
> Thinking of making this the # 1 option as it requires no additional setup? 

Yes.

docs2/site/docs/getting-started/dependency-injection.md
sungam3r: No generic argument required.

PR #1701, opened 2020-06-05
Raise errors upon failed coercion to meet graphql specifications

### LINKED ISSUE #1699: Invalid enum values do not raise an exception
When serializing an enumeration value that does not correspond to a enumeration member, no exception is raised, and the value gets serialized as null.  If it is a non-null graph, the parent object gets set to null without an exception getting raised.  Is there a way to incorporate validation into the serialization such that an exception is raised rather than incorrect data being returned?  Perhaps within `ValidateNodeResult`?  Note that if `null` was returned, an exception would be thrown, rather than the `null` value "bubbling up" to a nullable field.

See here where the enumeration serializer returns null for invalid members:

https://github.com/graphql-dotnet/graphql-dotnet/blob/1e2154d4885cd3a5415a46aca6be7e082c807d71/src/GraphQL/Types/Scalars/EnumerationGraphType.cs#L56

Note that this does not occur until the graph is serialized via this code, after all the resolvers have executed:

https://github.com/graphql-dotnet/graphql-dotnet/blob/1e2154d4885cd3a5415a46aca6be7e082c807d71/src/GraphQL/Execution/ExecutionStrategy.cs#L24

And here is the code that causes parent objects to return null if a non-null child (in this case a scalar) returns null:

https://github.com/graphql-dotnet/graphql-dotnet/blob/1e2154d4885cd3a5415a46aca6be7e082c807d71/src/GraphQL/Execution/ExecutionNode.cs#L166

Here is where the exception occurs if you attempt to return a null value from a resolver to a non-null graph:

https://github.com/graphql-dotnet/graphql-dotnet/blob/1e2154d4885cd3a5415a46aca6be7e082c807d71/src/GraphQL/Execution/ExecutionStrategy.cs#L251

I tried to alter `ExecutionStrategy` to throw an exception if a non-null child object returns null, but this breaks error situations where nulls are supposed to bubble up.  See this test:

https://github.com/graphql-dotnet/graphql-dotnet/blob/1e2154d4885cd3a5415a46aca6be7e082c807d71/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs#L189

Changing the enumeration serializer causes similar issues.  Any suggestions?  This is a problem for my code.

Shane32:
Fixes #1699 

Fixes both input and output serialization coercion to meet these graphql requirements:

Scalars:
> 3.5 Result Coercion: A GraphQL server, when preparing a field of a given scalar type, must uphold the contract the scalar type describes, either by coercing the value or producing a field error if a value cannot be coerced or if coercion may result in data loss.

> 3.5 Input Coercion: If a GraphQL server expects a scalar type as input to an argument, coercion is observable and the rules must be well defined. If an input value does not match a coercion rule, a query error must be raised.

Enums:

> 3.9 Result Coercion: GraphQL servers must return one of the defined set of possible values. If a reasonable coercion is not possible they must raise a field error.

> 3.9 Input Coercion: GraphQL has a constant literal to represent enum input values. GraphQL string literals must not be accepted as an enum input and instead raise a query error.

Errors:
> 6.6.4 If an error is thrown while resolving a field, it should be treated as though the field returned null, and an error must be added to the "errors" list in the response.

Notes:
* The `when_argument_provided_cannot_be_parsed` test explicitly tested for incorrect behavior.  This has been fixed.
* The `Very_Long_Number_Should_Return_Error_For_Int` and `Very_Long_Number_In_Input_Should_Work_For_Long` test violated rule 6.6.4 as they did not return null to the field and add the error to the result.  Rather, they did not return any data, and only the one error.  This has been fixed.
* Although there are no enum-specific changes in this PR, I have added a number of tests for enums to test coercion properties.
* No breaking changes for applications which meet the graphql specification (passing valid values as arguments to queries, and returning valid values from field resolvers).  Can potentially break applications that rely on the incorrect behavior demonstrated within `when_argument_provided_cannot_be_parsed`, or applications that return invalid values from field resolvers, as these will now cause a query error.

Limitations of this PR:
* When returning a list of scalars, and one of the scalars causes a coercion error, an error is reported for the entire list, rather than the single member that caused the coercion error.  Nulls are properly bubbled up (starting with the entire list, rather than the member that caused the error).  Tests are in place to demonstrate this functionality, along with an appropriate comment.  This can possibly be addressed in a future PR, but would require notable refactoring or redesign, due to the design of the exception processing in place.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -35,7 +35,7 @@ namespace GraphQL.Tests.Bugs
         public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1' to \u0027Bug1699Enum\u0027", 1, 3, "happy");
 
         [Fact]
-        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50'", 1, 3, "invalidEnum");
+        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnum");
 
         // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
         [Fact]
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -16,7 +16,7 @@ namespace GraphQL.Tests.Bugs
         private void AssertQueryWithError(string query, string result, string message, int line, int column, object[] path, Exception exception = null, string code = null, string inputs = null)
         {
             var error = exception == null ? new ExecutionError(message) : new ExecutionError(message, exception);
-            error.AddLocation(line, column);
+            if (line != 0) error.AddLocation(line, column);
             error.Path = path;
             if (code != null)
                 error.Code = code;
@@ -51,10 +51,19 @@ namespace GraphQL.Tests.Bugs
         public void Input_Enum_Valid2() => AssertQuerySuccess("{ input(arg: SLEEPY) }", @"{ ""input"": ""Sleepy"" }");
 
         [Fact]
+        public void Input_Enum_InvalidEnum() => AssertQueryWithError("{ input(arg: DOPEY) }", null, "Argument \u0022arg\u0022 has invalid value DOPEY.\nExpected type \u0022Bug1699Enum\u0022, found DOPEY.", 1, 9, (object[])null, code: "5.3.3.1");
+
+        [Fact]
         public void Input_Enum_InvalidString() => AssertQueryWithError(@"{ input(arg: ""SLEEPY"") }", null, "Argument \u0022arg\u0022 has invalid value \u0022SLEEPY\u0022.\nExpected type \u0022Bug1699Enum\u0022, found \u0022SLEEPY\u0022.", 1, 9, (object[])null, code: "5.3.3.1");
 
         [Fact]
         public void Input_Enum_InvalidInt() => AssertQueryWithError(@"{ input(arg: 2) }", null, "Argument \u0022arg\u0022 has invalid value 2.\nExpected type \u0022Bug1699Enum\u0022, found 2.", 1, 9, (object[])null, code: "5.3.3.1");
+
+        [Fact]
+        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to Bug1699Enum", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
+
+        [Fact]
+        public void Input_Enum_InvalidInt_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u00272\u0027 to Bug1699Enum", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":2}");
 
         [Fact]
         public void Input_Enum_UndefinedDefault() => AssertQuerySuccess("{ input }", @"{ ""input"": ""Grumpy"" }");
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -60,6 +60,9 @@ namespace GraphQL.Tests.Bugs
         public void Input_Enum_InvalidInt() => AssertQueryWithError(@"{ input(arg: 2) }", null, "Argument \u0022arg\u0022 has invalid value 2.\nExpected type \u0022Bug1699Enum\u0022, found 2.", 1, 9, (object[])null, code: "5.3.3.1");
 
         [Fact]
+        public void Input_Enum_Valid_Variable() => AssertQuerySuccess("query($arg: Bug1699Enum!) { input(arg: $arg) }", @"{ ""input"": ""Grumpy"" }", "{\"arg\":\"GRUMPY\"}".ToInputs());
+
+        [Fact]
         public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to Bug1699Enum", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
 
         [Fact]
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -63,7 +63,7 @@ namespace GraphQL.Tests.Bugs
         public void Input_Enum_Valid_Variable() => AssertQuerySuccess("query($arg: Bug1699Enum!) { input(arg: $arg) }", @"{ ""input"": ""Grumpy"" }", "{\"arg\":\"GRUMPY\"}".ToInputs());
 
         [Fact]
-        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to Bug1699Enum", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
+        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to \u0027Bug1699Enum\u0027", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
 
         [Fact]
         public void Input_Enum_InvalidInt_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u00272\u0027 to Bug1699Enum", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":2}");
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -32,7 +32,7 @@ namespace GraphQL.Tests.Bugs
 
         // within C#, (int)Bug1699Enum.Happy does not equal Bug1699.Happy
         [Fact]
-        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1'", 1, 3, "happy");
+        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1' to \u0027Bug1699Enum\u0027", 1, 3, "happy");
 
         [Fact]
         public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50'", 1, 3, "invalidEnum");
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -78,10 +78,10 @@ namespace GraphQL.Tests.Bugs
         public void Input_Enum_OverrideDefault() => AssertQuerySuccess("{ inputWithDefault(arg: SLEEPY) }", @"{ ""inputWithDefault"": ""Sleepy"" }");
 
         [Fact]
-        public void Input_Enum_MissingRequired() => AssertQueryWithError(@"{ inputRequired }", null, "Field \u0022inputRequired\u0022 argument \u0022arg\u0022 of type \u0022Bug1699Enum!\u0022 is required but not provided.", 1, 3, (object[])null, code: "5.3.3.2");
+        public void Input_Enum_MissingRequired() => AssertQueryWithError(@"{ inputRequired }", null, "Argument \u0022arg\u0022 of type \u0022Bug1699Enum!\u0022 is required for field \u0022inputRequired\u0022 but not provided.", 1, 3, (object[])null, code: "5.3.3.2");
 
         [Fact]
-        public void Input_Enum_RequiredWithDefault() => AssertQueryWithError(@"{ inputRequiredWithDefault }", null, "Field \u0022inputRequiredWithDefault\u0022 argument \u0022arg\u0022 of type \u0022Bug1699Enum!\u0022 is required but not provided.", 1, 3, (object[])null, code: "5.3.3.2");
+        public void Input_Enum_RequiredWithDefault() => AssertQueryWithError(@"{ inputRequiredWithDefault }", null, "Argument \u0022arg\u0022 of type \u0022Bug1699Enum!\u0022 is required for field \u0022inputRequiredWithDefault\u0022 but not provided.", 1, 3, (object[])null, code: "5.3.3.2");
     }
 
     public class Bug1699InvalidEnumSchema : Schema
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -39,7 +39,7 @@ namespace GraphQL.Tests.Bugs
 
         // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
         [Fact]
-        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to Bug1699Enum", 1, 3, "invalidEnumWithinList");
+        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnumWithinList");
 
         [Fact]
         public void Invalid_Enum_Within_NonNullList() => AssertQueryWithError("{ invalidEnumWithinNonNullList }", @"{ ""invalidEnumWithinNonNullList"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnumWithinNonNullList");
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -39,10 +39,10 @@ namespace GraphQL.Tests.Bugs
 
         // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
         [Fact]
-        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50'", 1, 3, "invalidEnumWithinList");
+        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to Bug1699Enum", 1, 3, "invalidEnumWithinList");
 
         [Fact]
-        public void Invalid_Enum_Within_NonNullList() => AssertQueryWithError("{ invalidEnumWithinNonNullList }", @"{ ""invalidEnumWithinNonNullList"": null }", "Unable to serialize '50'", 1, 3, "invalidEnumWithinNonNullList");
+        public void Invalid_Enum_Within_NonNullList() => AssertQueryWithError("{ invalidEnumWithinNonNullList }", @"{ ""invalidEnumWithinNonNullList"": null }", "Unable to serialize '50' to Bug1699Enum", 1, 3, "invalidEnumWithinNonNullList");
 
         [Fact]
         public void Input_Enum_Valid() => AssertQuerySuccess("{ inputEnum(arg: SLEEPY) }", @"{ ""inputEnum"": ""SLEEPY"" }");
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -37,7 +37,7 @@ namespace GraphQL.Tests.Bugs
         [Fact]
         public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50'", 1, 3, "invalidEnum");
 
-        // does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
+        // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
         [Fact]
         public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50'", 1, 3, "invalidEnumWithinList");
 
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -42,7 +42,7 @@ namespace GraphQL.Tests.Bugs
         public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to Bug1699Enum", 1, 3, "invalidEnumWithinList");
 
         [Fact]
-        public void Invalid_Enum_Within_NonNullList() => AssertQueryWithError("{ invalidEnumWithinNonNullList }", @"{ ""invalidEnumWithinNonNullList"": null }", "Unable to serialize '50' to Bug1699Enum", 1, 3, "invalidEnumWithinNonNullList");
+        public void Invalid_Enum_Within_NonNullList() => AssertQueryWithError("{ invalidEnumWithinNonNullList }", @"{ ""invalidEnumWithinNonNullList"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnumWithinNonNullList");
 
         [Fact]
         public void Input_Enum_Valid() => AssertQuerySuccess("{ inputEnum(arg: SLEEPY) }", @"{ ""inputEnum"": ""SLEEPY"" }");
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -1,0 +1,134 @@
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using System;
+using System.Linq;
+using System.Numerics;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/1699
+    public class Bug1699InvalidEnum : QueryTestBase<Bug1699InvalidEnumSchema>
+    {
+        private void AssertQueryWithError(string query, string result, string message, int line, int column, string path, Exception exception = null, string code = null, string inputs = null)
+            => AssertQueryWithError(query, result, message, line, column, new object[] { path }, exception, code, inputs);
+
+        private void AssertQueryWithError(string query, string result, string message, int line, int column, object[] path, Exception exception = null, string code = null, string inputs = null)
+        {
+            var error = exception == null ? new ExecutionError(message) : new ExecutionError(message, exception);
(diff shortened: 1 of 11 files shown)


Shane32:
@sungam3r can we review this?  It caused about 4 man-hours of downtime for my team again today because no error was returned, as it should have pursuant to graphql specifications.  There's no reason not to merge this PR, unless someone is counting on out-of-spec behavior of their graphs.  And if that's the case, I'd say they can stick with an older version of the library.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -207,10 +208,10 @@ private static object ValueFromScalar(ScalarGraphType scalar, object input)
         {
             if (input is IValue value)
             {
-                return scalar.ParseLiteral(value);
+                return scalar.ParseLiteral(value) ?? throw new ArgumentException($"Unable to convert '{value.Value}'");
```
sungam3r, line 211: Look at `AssertValidValue`:
```c#
 if (ValueFromScalar(scalar, input) == null)
                    throw new InvalidValueException(fieldName, "Invalid Scalar value for input field.");  <-- this line is no longer needed
```
Lets inline `ValueFromScalar` method body inside `AssertValidValue` and remove `ValueFromScalar` method. And change
```c#
throw new ArgumentException($"Unable to convert '{value.Value}'");
```
to
```c#
throw new ArgumentException($"Invalid Scalar value for input field {fieldName}. Unable to convert '{value.Value}' to {type.Name}");
```

I think everyone wants to see exactly this kind of detailed error message.

sungam3r:
@Shane32 You say everything right, good job. Indeed, this is a bug in the library behavior. I've already started looking at your edits. In recent months, I have not been able to devote enough time to supporting this and other projects. I just do not have enough time because of my family, so many PRs and issues remain unanswered for a long time. Only recently I have had a little time, but it is still not enough to work thoughtfully on the review.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I updated it as you suggested.  This code only runs when an input variable is unable to be converted, so I have it throw an `InvalidValueException` as this inherits from `ExecutionError` and returns the field name and location information back to the caller.

Note that the location information is blocked currently due to another bug; fixed in PR #1735 which can be reviewed after this PR is merged.

Shane32:
I really appreciate you spending the time to review this.  If we can grunt through a few more issues together, I think the code will be in great shape for a 3.0 release.  I'm swamped too, working most evenings (when I should be playing with my kids!), but I really need to spend the time to get these few outstanding issues merged:

* This issue/PR
* Add service provider to IResolveFieldContext - PR #1730
* Fix data loaders - PR #1511
* Properly separate "client" vs "server" errors - see #1650 (needs a lot work, which I can do, but I'll need collaboration from you)
* Upgrade documentation (which I've got mostly done)

We can close all my other PRs about refactoring `IGraphType` and stuff.  It really doesn't matter.

Anyway, if you can commit some time to helping me get through these, even if slowly, that would be great.  Otherwise I might have to fork the project, and nobody wants that.  Certainly not me.  I have a large project I'm working on and I just really need this stuff done, and some of this has been outstanding for nearly a year.

Let's work together, given the small amount of time we each have, and get through it.  Thanks again!!

Also, I just want to say, @joemcbride , if you want an additional team member on board, I'm open to helping this project in any way I can.

sungam3r:
> even if slowly

Most likely it will.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -143,8 +144,18 @@ public static void AssertValidValue(ISchema schema, IGraphType type, object inpu
 
             if (type is ScalarGraphType scalar)
             {
-                if (ValueFromScalar(scalar, input) == null)
-                    throw new InvalidValueException(fieldName, "Invalid Scalar value for input field.");
+                // validate value can be parsed correctly
```
sungam3r, line 147: I do not understand this comment.

Shane32: hmmm.... well, what it does is attempt a conversion, testing to see if it converts properly, and discards the result

Shane32: how about `// verify value can be converted successfully` ?

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -143,8 +144,18 @@ public static void AssertValidValue(ISchema schema, IGraphType type, object inpu
...
+
+                if (input is IValue value)
+                {
+                    if (scalar.ParseLiteral(value) == null)
+                        throw new InvalidValueException(fieldName, $"Unable to convert '{value.Value}' to {type.Name}");
```
sungam3r, line 152: Lets rename `fieldName` to `variableName` in ` public static void AssertValidValue(ISchema schema, IGraphType type, object input, string fieldName)` and also in `InvalidValueException` ctor. It's not really a variable name, because the check can happen on array elements and on nested properties, but nevertheless it's a better name than `fieldName`, because this method is made for input values. So I even think it would be better to rename `AssertValidValue` to `AssertValidVariableValue`.

Shane32: no problem; hang on a sec

Shane32: yeah me too.  only during debugging did I realize it was only for variables

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -299,7 +300,7 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
 
             if (type is ScalarGraphType scalarType)
             {
-                return scalarType.ParseLiteral(input);
+                return scalarType.ParseLiteral(input) ?? throw new ArgumentException($"Unable to convert '{input}'");
```
sungam3r, line 303:
```suggestion
                return scalarType.ParseLiteral(input) ?? throw new InvalidValueException($"Unable to convert '{input}' to {type.Name}");
```

Shane32:
both suggestions done

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: updated to the comment written above

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: OK

Shane32:
CoerceValue doesn't have the variable name so it can't throw an InvalidValueException.  Let me check it a little more.  I don't think it's possible to hit that error since AssertValidVariableValue applies first.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: Do you mean "verif**iable** value can be converted successfully" ?

Shane32:
The `AssertValidVariableValue` validates input variables before `CoerceValue` is hit, and there's an `IValidationRule` that normally protects against invalid input arguments.  Without with validation rule, `CoerceValue` throws an exception if `ParseLiteral` fails.  However, this would only occur within a field resolver's call to `GetArgument`.  So I think we can safely stick with an `ArgumentException`.

sungam3r:
Looking through code I found ` public static IEnumerable<string> IsValidLiteralValue(this IGraphType type, IValue valueAst, ISchema schema)` with 
```c#
var parseResult = scalar.ParseLiteral(valueAst);

            if (parseResult == null)
            {
                return new[] { $"Expected type \"{type.Name}\", found {AstPrinter.Print(valueAst)}." };
            }
```

inside. How does this check correlate with you PR? I see that `IsValidLiteralValue` is called only during validation phase. Is there a duplication of logic between literal and variable checks?

Shane32:
Fix tests

Shane32:
> Looking through code I found ` public static IEnumerable<string> IsValidLiteralValue(this IGraphType type, IValue valueAst, ISchema schema)` with
> 
> ```cs
> var parseResult = scalar.ParseLiteral(valueAst);
> 
>             if (parseResult == null)
>             {
>                 return new[] { $"Expected type \"{type.Name}\", found {AstPrinter.Print(valueAst)}." };
>             }
> ```
> 
> I see that `IsValidLiteralValue` is called only during validation phase. Is there a duplication of logic between literal and variable checks?

Probably.  `AssertValidVariableValue` appears to validate the actual object (e.g. a string), throwing an error upon failure, whereas `IsValidLiteralValue` works on an `IValue` and returns a string message upon failure.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: no.  `validate` is a verb; verifiable is an adjective.  More correctly, I could write "This code validates that the value can be parsed".

Shane32: Actually I wrote `verify value can be converted successfully`.   Which could be written as "This code verifies that the value can be converted successfully."

sungam3r:
>  I can create an issue if you want to track refactoring this method.

OK.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: Now I see: verify **that** value can be ...
OK

Shane32:
See #1736

sungam3r:
@Shane32 I remember that I have to review this PR. I have not forgotten. Every day there is urgent work, so things are not fast.

sungam3r:
@Shane32 I will review it in the next couple of days.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -105,12 +106,12 @@ public static object GetVariableValue(Document document, ISchema schema, Variabl
...
             try
             {
-                AssertValidValue(schema, type, input, variable.Name);
+                AssertValidVariableValue(schema, type, input, variable.Name);
             }
             catch (InvalidValueException error)
             {
                 error.AddLocation(variable, document);
-                throw;
+                throw error;
```
sungam3r, line 114: Why is this done?

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -143,8 +145,18 @@ public static void AssertValidValue(ISchema schema, IGraphType type, object inpu
...
             if (type is ScalarGraphType scalar)
             {
-                if (ValueFromScalar(scalar, input) == null)
-                    throw new InvalidValueException(fieldName, "Invalid Scalar value for input field.");
+                // verify value can be converted successfully
+
+                if (input is IValue value)
+                {
+                    if (scalar.ParseLiteral(value) == null)
+                        throw new InvalidValueException(variableName, $"Unable to convert '{value.Value}' to {type.Name}");
```
sungam3r, line 153: Below I can see messages
` $"Unable to parse input as a '{type.Name}' type. Did you provide a List or Scalar value accidentally?");`
and
`   $"Unrecognized input fields {string.Join(", ", unknownFields.Select(k => $"'{k}'"))} for type '{type.Name}'.");`

with quoted `type.Name`. Let's do here and below also.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -299,7 +301,7 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
 
             if (type is ScalarGraphType scalarType)
             {
-                return scalarType.ParseLiteral(input);
+                return scalarType.ParseLiteral(input) ?? throw new ArgumentException($"Unable to convert '{input}' to {type.Name}");
```
sungam3r, line 304: quoted `type.Name` also

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+        public void String_Enum() => AssertQuerySuccess("{ sleepy }", @"{ ""sleepy"": ""SLEEPY"" }");
+
+        // within C#, (int)Bug1699Enum.Happy does not equal Bug1699.Happy
+        [Fact]
+        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1'", 1, 3, "happy");
+
+        [Fact]
+        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50'", 1, 3, "invalidEnum");
+
+        // does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
```
sungam3r, line 40:
```suggestion
        // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
```

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+        public void Input_Enum_UndefinedDefault() => AssertQuerySuccess("{ input }", @"{ ""input"": ""Grumpy"" }");
+
+        [Fact]
+        public void Input_Enum_SetDefault() => AssertQuerySuccess("{ inputWithDefault }", @"{ ""inputWithDefault"": ""Happy"" }");
+
+        [Fact]
+        public void Input_Enum_OverrideDefault() => AssertQuerySuccess("{ inputWithDefault(arg: SLEEPY) }", @"{ ""inputWithDefault"": ""Sleepy"" }");
+
+        [Fact]
+        public void Input_Enum_MissingRequired() => AssertQueryWithError(@"{ inputRequired }", null, "Field \u0022inputRequired\u0022 argument \u0022arg\u0022 of type \u0022Bug1699Enum!\u0022 is required but not provided.", 1, 3, (object[])null, code: "5.3.3.2");
```
sungam3r, line 81: Somewhat awkward error message from `ProvidedNonNullArguments`. Let's fix it like
`$"Argument \"{argName}\" of type \"{type}\" is required for field \"{fieldName}\" but not provided."`

The same fix for `MissingDirectiveArgMessage`. OK?

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        [Fact]
+        public void Input_Enum_RequiredWithDefault() => AssertQueryWithError(@"{ inputRequiredWithDefault }", null, "Field \u0022inputRequiredWithDefault\u0022 argument \u0022arg\u0022 of type \u0022Bug1699Enum!\u0022 is required but not provided.", 1, 3, (object[])null, code: "5.3.3.2");
```
sungam3r, line 84: It seems wrong to me. See http://spec.graphql.org/June2018/#sec-Required-Arguments . The numbering of the validation rules is a little off, but it doesn't matter.

> Arguments can be required. An argument is required if the argument type is non‐null **and does not have a default value**. Otherwise, the argument is optional.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I don't know.  I just verified that the location is added regardless of `throw` or `throw error`.  I'll fix it.

Shane32:
```suggestion
                throw;
```

src/GraphQL/Execution/ExecutionHelper.cs
Shane32:
```suggestion
                        throw new InvalidValueException(variableName, $"Unable to convert '{value.Value}' to '{type.Name}'");
```

src/GraphQL/Execution/ExecutionHelper.cs
Shane32:
```suggestion
                return scalarType.ParseLiteral(input) ?? throw new ArgumentException($"Unable to convert '{input}' to '{type.Name}'");
```

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
Shane32: Makes sense.  I'll need to review the code to see where to implement the fix.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -119,6 +119,11 @@ public static void SetArrayItemNodes(ExecutionContext context, ArrayExecutionNod
                     {
                         SetArrayItemNodes(context, arrayNode);
                     }
+                    else if (node is ValueExecutionNode valueNode)
+                    {
+                        node.Result = ((ScalarGraphType)itemType).Serialize(d)
```
sungam3r, line 124: You check if `node` is `ValueExecutionNode` but then cast different variable `itemType` to `ScalarGraphType`. I propose to change ValueExecutionNode ctor second parameter from `IGraphType` to `ScalarGraphType` and then use here `((ScalarGraphType)valueNode.GraphType).Serialize(d)`.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
Shane32: Done.  See commit below

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -119,6 +119,11 @@ public static void SetArrayItemNodes(ExecutionContext context, ArrayExecutionNod
...
+                            ?? throw new ExecutionError($"Unable to serialize '{d}'");
```
sungam3r, line 125:
```suggestion
                            ?? throw new ExecutionError($"Unable to serialize '{d}' to {valueNode.GraphType.Name}");
```

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: The same suggestions to chenged lines inside `ExecuteNodeAsync`

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Done (see commit below)

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Well, if the resolved value is null, `SetArrayItemNodes` creates a `ValueExecutionNode` with a result of null, no matter if it's a scalar graph type or not.  Would you want to create a `NullExecutionNode` to handle this scenario?

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
Shane32: Done -- fixed validation rule for this.  See commit below.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: `throw error` loses the stack

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Done; added NullExecutionNode as described above.  Please review commit below.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
sungam3r: It seems that I fixed it too in one of my draft PR ). Otherwise I would not pay attention to this part. Will check my commits later to validate changes. Probably we will close one more issue.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Ok

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -143,8 +145,18 @@ public static void AssertValidValue(ISchema schema, IGraphType type, object inpu
...
+
+                if (input is IValue value)
+                {
+                    if (scalar.ParseLiteral(value) == null)
+                        throw new InvalidValueException(variableName, $"Unable to convert '{value.Value}' to '{type.Name}'");
+                }
+                else
+                {
+                    if (scalar.ParseValue(input) == null)
+                        throw new InvalidValueException(variableName, $"Unable to convert '{input}' to {type.Name}");
```
sungam3r, line 158:
```suggestion
                        throw new InvalidValueException(variableName, $"Unable to convert '{input}' to '{type.Name}'");
```

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -205,6 +207,11 @@ protected virtual async Task ExecuteNodeAsync(ExecutionContext context, Executio
                     {
                         SetArrayItemNodes(context, arrayNode);
                     }
+                    else if (node is ValueExecutionNode valueNode)
+                    {
+                        node.Result = valueNode.GraphType.Serialize(node.Result)
+                            ?? throw new ExecutionError($"Unable to serialize '{node.Result}'");
```
sungam3r, line 213:
```suggestion
                            ?? throw new ExecutionError($"Unable to serialize '{node.Result}' to '{valueNode.GraphType.Name}'");
```

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -119,6 +119,11 @@ public static void SetArrayItemNodes(ExecutionContext context, ArrayExecutionNod
                     {
                         SetArrayItemNodes(context, arrayNode);
                     }
+                    else if (node is ValueExecutionNode valueNode)
+                    {
+                        node.Result = valueNode.GraphType.Serialize(d)
+                            ?? throw new ExecutionError($"Unable to serialize '{d}' to {valueNode.GraphType.Name}");
```
sungam3r, line 125:
```suggestion
                            ?? throw new ExecutionError($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}'");
```

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
sungam3r: No, I fixed another rule. The problem is similar, the default value for the argument is ignored.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+        public void Input_Enum_InvalidString() => AssertQueryWithError(@"{ input(arg: ""SLEEPY"") }", null, "Argument \u0022arg\u0022 has invalid value \u0022SLEEPY\u0022.\nExpected type \u0022Bug1699Enum\u0022, found \u0022SLEEPY\u0022.", 1, 9, (object[])null, code: "5.3.3.1");
+
+        [Fact]
+        public void Input_Enum_InvalidInt() => AssertQueryWithError(@"{ input(arg: 2) }", null, "Argument \u0022arg\u0022 has invalid value 2.\nExpected type \u0022Bug1699Enum\u0022, found 2.", 1, 9, (object[])null, code: "5.3.3.1");
+
+        [Fact]
+        public void Input_Enum_Valid_Variable() => AssertQuerySuccess("query($arg: Bug1699Enum!) { input(arg: $arg) }", @"{ ""input"": ""Grumpy"" }", "{\"arg\":\"GRUMPY\"}".ToInputs());
+
+        [Fact]
+        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to Bug1699Enum", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
```
sungam3r, line 66:
```suggestion
        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to \u0027Bug1699Enum\u0027", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
```

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+        public void Input_Enum_InvalidInt() => AssertQueryWithError(@"{ input(arg: 2) }", null, "Argument \u0022arg\u0022 has invalid value 2.\nExpected type \u0022Bug1699Enum\u0022, found 2.", 1, 9, (object[])null, code: "5.3.3.1");
+
+        [Fact]
+        public void Input_Enum_Valid_Variable() => AssertQuerySuccess("query($arg: Bug1699Enum!) { input(arg: $arg) }", @"{ ""input"": ""Grumpy"" }", "{\"arg\":\"GRUMPY\"}".ToInputs());
+
+        [Fact]
+        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to \u0027Bug1699Enum\u0027", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
+
+        [Fact]
+        public void Input_Enum_InvalidInt_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u00272\u0027 to Bug1699Enum", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":2}");
```
sungam3r, line 69:
```suggestion
        public void Input_Enum_InvalidInt_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u00272\u0027 to \u0027Bug1699Enum\u0027", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":2}");
```

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        [Fact]
+        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50'", 1, 3, "invalidEnum");
+
+        // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
+        [Fact]
+        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to Bug1699Enum", 1, 3, "invalidEnumWithinList");
+
+        [Fact]
+        public void Invalid_Enum_Within_NonNullList() => AssertQueryWithError("{ invalidEnumWithinNonNullList }", @"{ ""invalidEnumWithinNonNullList"": null }", "Unable to serialize '50' to Bug1699Enum", 1, 3, "invalidEnumWithinNonNullList");
```
sungam3r, line 45:
```suggestion
        public void Invalid_Enum_Within_NonNullList() => AssertQueryWithError("{ invalidEnumWithinNonNullList }", @"{ ""invalidEnumWithinNonNullList"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnumWithinNonNullList");
```

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+        // within C#, (int)Bug1699Enum.Happy does not equal Bug1699.Happy
+        [Fact]
+        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1'", 1, 3, "happy");
+
+        [Fact]
+        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50'", 1, 3, "invalidEnum");
+
+        // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
+        [Fact]
+        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to Bug1699Enum", 1, 3, "invalidEnumWithinList");
```
sungam3r, line 42:
```suggestion
        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnumWithinList");
```

sungam3r:
@Shane32 I think I looked carefully at all your changes. At least now I don't see things that I would like to change. Do you have any other edits?

Shane32:
Nope; I'm good if you are.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        [Fact]
+        public void Simple_Enum() => AssertQuerySuccess("{ grumpy }", @"{ ""grumpy"": ""GRUMPY"" }");
+
+        [Fact]
+        public void String_Enum() => AssertQuerySuccess("{ sleepy }", @"{ ""sleepy"": ""SLEEPY"" }");
+
+        // within C#, (int)Bug1699Enum.Happy does not equal Bug1699.Happy
+        [Fact]
+        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1'", 1, 3, "happy");
```
sungam3r, line 35:
```suggestion
        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1' to \u0027Bug1699Enum\u0027", 1, 3, "happy");
```

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        [Fact]
+        public void String_Enum() => AssertQuerySuccess("{ sleepy }", @"{ ""sleepy"": ""SLEEPY"" }");
+
+        // within C#, (int)Bug1699Enum.Happy does not equal Bug1699.Happy
+        [Fact]
+        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1' to \u0027Bug1699Enum\u0027", 1, 3, "happy");
+
+        [Fact]
+        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50'", 1, 3, "invalidEnum");
```
sungam3r, line 38:
```suggestion
        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnum");
```

sungam3r:
@Shane32 merged! I know I promised you to review your other PRs. What would you like to be next?

Shane32:
The ServiceProvider should be a very simple PR to review; not much actual code just some structure changes.... https://github.com/graphql-dotnet/graphql-dotnet/pull/1730

The data loader we can do after that, which is what everyone really needs, but it will take the most time for you to review I would guess.

Shane32:
And thanks @sungam3r  :)

PR #1720, opened 2020-07-08
Stop escaping enum literals when used as default values

isaksky:
Co-Authored-By: Maria Miller <mariajmiller@users.noreply.github.com>

Enum values like `RED` were getting schema-printed as `"RED"` when used as default values, which is wrong. The fix is to just move the check of whether it is an enum first when printing.


sungam3r:
Thanks.

PR #1725, opened 2020-07-14
Decimal precision fix

### LINKED ISSUE #1639: decimal precission issue in graphql
I am using GraphQL nuget package in one of my project. The version is 2.4.0.

When I am passing the decimal value as 12.10 the graphql is somehow transforming that into 12.1
If i send something like 12.11 then its coming as 12.11 but not when the last digit is zero. Is there something that i need to change in my code to get it working?

GraphQL decleration

`            Field(x => x.percentage, nullable: true).Name("percentage");`

Decleration

`public decimal? percentage{ get; set; }`

sungam3r:
fixes #1639

### DIFF
--- a/src/GraphQL/Language/CoreToVanillaConverter.cs
+++ b/src/GraphQL/Language/CoreToVanillaConverter.cs
@@ -231,7 +231,18 @@ namespace GraphQL.Language
                 case ASTNodeKind.FloatValue:
                 {
                     var str = (GraphQLScalarValue)source;
-                    return new FloatValue(ValueConverter.ConvertTo<double>(str.Value)).WithLocation(str, _body);
+
+                    // the idea is to see if there is a loss of accuracy of value
+                    // for example, 12.1 or 12.11 is double but 12.10 is decimal
+                    double dbl = ValueConverter.ConvertTo<double>(str.Value);
+                    decimal dec = ValueConverter.ConvertTo<decimal>(str.Value);
+                    // TODO: make more efficient, current solution allocates memory
+                    int[] decBits = decimal.GetBits(dec);
+                    int[] dblAsDecBits = decimal.GetBits(new decimal(dbl));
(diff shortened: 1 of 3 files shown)


src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -231,7 +231,18 @@ public IValue Value(GraphQLValue source)
                 case ASTNodeKind.FloatValue:
                 {
                     var str = (GraphQLScalarValue)source;
-                    return new FloatValue(ValueConverter.ConvertTo<double>(str.Value)).WithLocation(str, _body);
+
+                    // the idea is to see if there is a loss of accuracy of value
+                    // for example, 12.1 or 12.11 is double but 12.10 is decimal
+                    double dbl = ValueConverter.ConvertTo<double>(str.Value);
+                    decimal dec = ValueConverter.ConvertTo<decimal>(str.Value);
+                    // TODO: make more efficient, current solution allocates memory
```
sungam3r, line 239: I will be glad if someone suggests a more efficient way.

src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -231,7 +231,18 @@ public IValue Value(GraphQLValue source)
...
+                    int[] decBits = decimal.GetBits(dec);
+                    int[] dblAsDecBits = decimal.GetBits(new decimal(dbl));
+                    if (decBits[0] != dblAsDecBits[0] || decBits[1] != dblAsDecBits[1] || decBits[2] != dblAsDecBits[2] || decBits[3] != dblAsDecBits[3])
+                        return new DecimalValue(dec);
```
SimonCropp, line 243: should there be a withlocation here?

src/GraphQL/Language/CoreToVanillaConverter.cs
SimonCropp: happy to go with this for now

src/GraphQL/Language/CoreToVanillaConverter.cs
sungam3r: 🤦 Of course, fixed.

PR #1730, opened 2020-07-19
Add service provider to IResolveFieldContext - basic

Shane32:
This PR is an implementation of the proposal by @johnrutherford in #1151 - simply adding a `RequestServices` property to the `ExecutionOptions` and `IResolveFieldContext` property, and a `Services` property to `Schema`.  This is similar to the `RequestServices` property in Asp.Net Core's `HttpContext`.  This is a non-breaking change, and in addition to allowing field resolvers to easily pull a scoped service, allows for custom execution strategies to do things like create individual scopes for field resolvers.

This PR is identical to #1571 except it does not include the extra project to assist creating scopes within field resolvers.

Note that this PR (and #1571) also reverts the recent change to `ISchema` which removed the `Services` property and fixes the middleware layer to pull services from `schema.Services` rather than `(IServiceProvider)schema`.  This could be considered a breaking change, although it would be more accurate to say that it restores functionality that was removed in a prior beta version.

Please see #1571 and #1345 for discussion regarding this feature.  Can we merge this?


sungam3r:
Well, I see the old question is back :), it's been a long time since then. In fact, I want to point out first that I am no longer so categorical in my beliefs **not to add** this property. I understand that we will gain more than we lose because we can move the project forward. Nevertheless, I would like to draw your attention to the following issues that concern me.

1. Naming. `RequestServices` does not seem to me a name that is 100% suitable for this property.
2. Scope.  _Request_ Services implies that the created objects are in the request scope. In fact, this is not the case because `CreateScope()` method is not called anywhere. Often the schema is created as a singleton. This is the recommended way to use it. Thus, the root provider will be used. This means that scoped objects cannot be created through it.
3. It is necessary to clearly point in the [documentation](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/docs2/site/docs/getting-started/dependency-injection.md) the role of this property, its place in the system with its advantages and disadvantages.

If these problems are solved, I do not mind to merge PR. I will be happy to deal with other issues/PRs about scoped dependencies.

Shane32:
Thank you @sungam3r !  Let's discuss the naming and implementation details, and then I will be happy to write documentation, tests, etc.

I would like to clarify that in this PR, the service provider is simply passed from `ExecutionOptions`, and defaults to null.  If the user does not set the new `ExecutionOptions.RequestServices` property, the new `IResolveFieldContext.RequestServices` property will remain null.  It is designed this way so that people cannot accidentally misuse the property as you anticipated.  As such, it is up to the user to decide how to utilize the `RequestServices` property.

I suggest that rather than returning null, a descriptive error message is thrown if someone tries to access `IResolveFieldContext.RequestServices` when they have not initialized `ExecutionOptions.RequestServices`.

For Asp.Net Core, the user should set `ExecutionOptions.RequestServices` equal that of the `HttpContext.RequestServices` property.  In this case, it would be the request scope; `CreateScope()` is not required.  I use this with Asp.Net, sometimes with middleware as shown in the Harness sample, and sometimes within a MVC controller.  Either way, the `HttpContext.RequestServices` property provides a scoped service provider.  If you can let me know how other common implementations of this project are set up, I would be better able to write documentation.

If the user has another form of dependency injection in use, they may set it to the DI provider of their choice.  Or they may choose to set it to the root service provider, if that works best for their needs.  I can certainly provide a sample in the documentation of someone using `CreateScope()` to create a scope during document execution.

Note that the `Services` property on `ISchema` would normally be the root service provider (or `DefaultServiceProvider`), and is required by `Schema`.  So middleware and field resolvers would also have access to the root service provider.

As to the name of the property, `RequestServices`, the goal of this property is to provide a scoped service provider.  If someone wants to access the root service provider, they could do so through `ISchema.Services`.  Further, the name matches the name of `HttpContext.RequestServices`, so it should be familiar to Asp.Net users.  Do you have another naming suggestion for this property?

I can clarify all of the above in the documentation, if you agree with the design.

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -38,5 +38,7 @@ public class ExecutionContext : IExecutionContext
         public int? MaxParallelExecutionCount { get; set; }
 
         public Dictionary<string, object> Extensions { get; set; } = new Dictionary<string, object>();
+
+        public IServiceProvider RequestServices { get; set; }
```
Shane32, line 42:
```suggestion
        private IServiceProvider _requestServices;
        public IServiceProvider RequestServices
        {
            get => _requestServices ?? throw new InvalidOperationException("No scoped service provider specified.  Please set the value of ExecutionOptions.RequestServices to a valid scoped service provider.");
            set => _requestServices = value;
        }
```

Shane32: In the example above, the error message says "scoped service provider", which of course is not strictly necessary.  The error message could read something like this:

> No service provider specified.  Please set the value of the ExecutionOptions.RequestServices to a valid service provider.  Typically, this would be a scoped service provider from your dependency injection framework.

It could even explicitly mention Asp.Net...

sungam3r:
> I would like to clarify that in this PR, the service provider is simply passed from ExecutionOptions, and defaults to null.

That is, we will have two independent. `IServiceProvider`s. One in `Schema` ctor, second in `ExecutionOptions`, `ExecutionContext`, `IFieldResolver`. Not good.

> If the user has another form of dependency injection in use, they may set it to the DI provider of their choice. Or they may choose to set it to the root service provider, if that works best for their needs. 

I understand your intentions, they are clear. My concerns are caused by the fact that the two service providers are independent of each other and it is easy to get confused, because both can be used in the resolver. The schema can be either singleton (recommended) or scoped (yes, there are people who do this), so things become even complicated since `IServiceProvider` in schema ctor may be rooted or scoped as well.

> Note that the Services property on ISchema would normally be the root service provider (or DefaultServiceProvider), and is required by Schema. So middleware and field resolvers would also have access to the root service provider.

_normally_ but not always as I wrote above

>  Although this would typically be a scoped service provider, it may not be so. Also, if someone wants to access the root service provider, they could do so through ISchema.Services.

So I think that you also understand what I'm talking about.

It seems to me that the problem is not with the property you are adding. The problem is in the already existing `IServiceProvider` in schema ctor. Its echoes are visible in the way the middleware is created. I have a feeling that having got rid of the setting of `IServiceProvider` to the schema ctor in one way or another, we will untie our hands. Breaking changes so better think twice. Anyway I believe that having two service providers is a design oversight. @joemcbride ?

src/GraphQL/Execution/ExecutionContext.cs
sungam3r: OK.

Shane32:
In Asp.Net, there is a separate hidden DI container for use by the Asp.Net internals.  Perhaps the schema's use of the service provider falls into a similar category - where `ISchema.Services` property should only ever be used while the graphs are being set up.  As such, perhaps we leave the `ISchema.Services` property hidden as it is now.

In theory, as long as the `RequestServices` property is available for the resolvers' execution (be it the root or a scoped service provider), the resolvers should not need access to the schema's service provider.

My only issue with this solution is the fact that middleware builders need to cast `ISchema` to `IServiceProvider`.  This seems like a design flaw.  Perhaps we add a method to `ISchema` called `GetMiddleware<T>`, for example:

```csharp
public virtual T GetMiddleware<T>() where T : IFieldMiddleware
    => _services.GetService(typeof(T));
```

Another option perhaps would be to **_require_** the schema to be of type `Schema`, for example by changing `ExecutionOptions.Schema` to type `Schema`.  I don't care for that option.

src/GraphQL/Execution/ExecutionContext.cs
Shane32:
```suggestion
        private IServiceProvider _requestServices;
        public IServiceProvider RequestServices
        {
            get => _requestServices ?? throw new InvalidOperationException("No service provider specified.  Please set the value of the ExecutionOptions.RequestServices to a valid service provider.  Typically, this would be a scoped service provider from your dependency injection framework.");
            set => _requestServices = value;
        }
```

sungam3r:
>  ISchema.Services property should only ever be used while the graphs are being set up.

Yes. I want to say that this is a rather narrow use and one could (I believe) get rid of it. For example via some schema initialization refactoring setting all needed graphtypes through ctor ienumerable parameter (di approach). So we will have both schema object and graphtypes objects build out of the box from di. No need to call GetService later inside GraphTypesLookup.

>  the resolvers should not need access to the schema's service provider .

Yes. 

> My only issue with this solution is the fact that middleware builders need to cast ISchema to IServiceProvider. This seems like a design flaw.

I agree that middlewares should be slightly redesigned.

Shane32:
@sungam3r Causing `IResolveFieldContext.RequestServices` to throw an exception when `ExecutionOptions.RequestServices` is null causes a problem whenever code attempts to duplicate all the properties of `ResolveFieldContext`.  Typically this does not happen due to the use of `ReadonlyResolveFieldContext` and `ResolveFieldContextAdapter`.  I've created similar classes for `ResolveEventStreamContext` and `ResolveConnectionContext` so they will not throw an exception.

Since the some of the constructors have changed for `ResolveFieldContext`, `ResolveEventStreamContext`, and `ResolveConnectionContext`, this is a breaking change.  Also, the `ResolveFieldContext(IResolveFieldContext context)` constructor can now throw an exception if `RequestServices` is null -- another breaking change.  These classes are now unused by the project.  We have a few options:

1. remove the unused classes
2. remove the unused classes and expose ResolveFieldContextAdapter as a public class (as it is often useful as a base class where ResolveFieldContext was used before)
3. replace the unused classes with code similar to ResolveFieldContextAdapter, so it can serve as a base class for any custom implementation, probably removing the internal `ResolveFieldContextAdapter` classes.
4. adding constructors & other code to eliminate these changes from being breaking changes.
5. leave the breaking changes as-is
6. do not throw an exception when accessing `IResolveFieldContext.RequestServices` while `ExecutionOptions.RequestServices` is null

Thoughts?

Shane32:
For now, I have removed the `Services` property from `ISchema` and left the middleware as-is, where it casts `ISchema` to `IServiceProvider` in order to create the middleware.

Shane32:
I'll fix the merge conflicts tonight.

sungam3r:
>Thoughts?

I'm for option 6. I think it's not worth it to complicate things and throw an exception now. We will probably rework these classes again. I propose to delete unused `ResolveFieldContext` ctor with many parameters. With this in mind, are new adapter classes needed? I would not want their hierarchy to grow.

Shane32:
> I'm for option 6. 

I agree.  I'll revert that change.

> I propose to delete unused `ResolveFieldContext` ctor with many parameters. 

I agree. I'll delete the constructor you mentioned.  There are other constructors available for use.

> With this in mind, are new adapter classes needed? I would not want their hierarchy to grow.

The new adapter classes are better for future changes to the project - for example, if another property is added to `IResolveFieldContext`, the adapter can include a mapping for the new property.  So less breaking changes down the road.  Let me know if you think I should revert the newly added adapter classes.  I'm okay either way.

So do you think the new adapter classes should stay or should I revert those also?  I'm okay with whatever you suggest, as my custom code can always just implement the interface and so it really doesn't matter what classes are exposed anymore.

Shane32:
Also, keeping the new adapters also increases execution speed and memory compared to concrete classes, and reduces the amount of changes internally required to the project if `IResolveFieldContext` was to change.  We can mark them internal, but I don't see any reason to hide them.

sungam3r:
> So do you think the new adapter classes should stay or should I revert those also?

Revert.

> Also, keeping the new adapters also increases execution speed and memory compared to concrete classes

Let's do such changes in the separate PR, because benchmark is always required.

Shane32:
Done

src/GraphQL/Execution/IExecutionContext.cs
```diff
@@ -81,5 +81,11 @@ public interface IExecutionContext : IProvideUserContext
         /// protocol however they see fit, and hence there are no additional restrictions on its contents.
         /// </summary>
         Dictionary<string, object> Extensions { get; }
+
+        /// <summary>
+        /// The service provider for the executing request.  Typically this is a scoped service provider
```
sungam3r, line 86:
```suggestion
        /// The service provider for the executing request. Typically this is a scoped service provider
```

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -234,7 +235,8 @@ await listener.AfterExecutionAsync(context)
             List<IDocumentExecutionListener> listeners,
             bool throwOnUnhandledException,
             Action<UnhandledExceptionContext> unhandledExceptionDelegate,
-            int? maxParallelExecutionCount)
+            int? maxParallelExecutionCount,
+            IServiceProvider services)
```
sungam3r, line 239:
```suggestion
            IServiceProvider requestServices)
```
?

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -252,7 +254,8 @@ await listener.AfterExecutionAsync(context)
                 Listeners = listeners,
                 ThrowOnUnhandledException = throwOnUnhandledException,
                 UnhandledExceptionDelegate = unhandledExceptionDelegate,
-                MaxParallelExecutionCount = maxParallelExecutionCount
+                MaxParallelExecutionCount = maxParallelExecutionCount,
+                RequestServices = services
```
sungam3r, line 258:
```suggestion
                RequestServices = requestServices
```

?

Shane32:
Sounds good.  I've created a new issue and will likely post a PR for that alongside (or prior to) the upgrade documentation.

Shane32:
Probably I'll post it relatively soon.  I think I have a bunch of material that I wrote for the upgrade docs that would be useful.

PR #1735, opened 2020-07-22
Allow DocumentExecuter to return ExceptionErrors while parsing variables

Shane32:
This PR is based on #1701 - please pull #1701 before reviewing this PR

Right now, variable parsing errors occur within a DocumentExecuter but not within a resolver try..catch block and throw an `ExecutionError` exception.  However, they are treated as an unhandled exception.  They are not unhandled; they are properly handled and should be returned to the caller as an query error.  This PR fixes that.  This change also allows for the error location information to be returned back to the caller for those malformed queries.


Shane32:
@sungam3r Please take a look when you get a chance.  Thanks!!

sungam3r:
> variable parsing errors occur within a DocumentExecuter 

What method are you talking about exactly?

Shane32:
> > variable parsing errors occur within a DocumentExecuter
> 
> What method are you talking about exactly?

https://github.com/graphql-dotnet/graphql-dotnet/blob/e6fc914072dd5b3cddb4b18d8e6abcfa18fe83c2/src/GraphQL/Execution/DocumentExecuter.cs#L247

That line of code throws an InvalidValueException if a variable is invalid.

Shane32:
The most specific example of how this would change the project is when `ThrowOnUnhandledException` is true:

Suppose someone queries `{ missingField }`.  The document executer will return a graphql result with a validation error that the field is missing.

Now suppose someone queries `{ badquery`.  The document executer will throw the `ExecutionError` back to the caller rather than returning a graphql response containing an error.

The difference here is that the exception that occurs isn't unhandled.  It's properly handled by the infrastructure and is returning a proper response.  Unhandled exceptions should be reserved for non-ExecutionError errors, for instance, when an ObjectDisposedException occurs within a document listener.

This is already handled in this manner for field resolvers - why not for the entire document as well?

Shane32:
Note: this feature was requested in #1658 also

Shane32:
@sungam3r I've added tests to demonstrate and verify how this change affects the operation of the project.  For anyone using `ThrowOnUnhandledException`, this is a very important fix.

sungam3r:
Will review this today.

src/GraphQL.Tests/Bugs/Bug1735.cs
```diff
@@ -0,0 +1,67 @@
+using System.Threading.Tasks;
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1735
+    public class Bug1735 : QueryTestBase<Bug1735Schema>
```
sungam3r, line 10: There is no reported bug 1735.

sungam3r: I mean the Bugs folder is for tests that verify reported bugs i.e. issues.

Shane32: Moved to `GraphQL.Tests.Execution.DocumentExecuterExceptions`

PR #1749, opened 2020-07-29
update refs

SimonCropp:


sungam3r:
There might be people who have problems after raising the minimum required version of packages.  I myself am not against it.

SimonCropp:
> There might be people who have problems after raising the minimum required version of packages

i never understood that logic. i think packages should release with the current stable of dependencies at every release. for comparison, when adding a dependency, you always pick the latest stable, so if u dont make concessions for the first release with a dependency, i dont understand why you would on the subsequent releases

SimonCropp:
and i would prefer to optimise to better support for people who actively keep their software up-to-date, rather than optimise for people who dont

sungam3r:
The logic is to allow users to use wider range of compatible versions. They may have incompatibility issues with other parts of their software. Loosening requirements for minimum versions of dependencies may help them to choose - just keep older version or upgrade explicitly adding the same dependency but with higher version.

SimonCropp:
but they can also explicitly use an older version of a dependency

sungam3r:
No. Only specified in nuspec or higher. Nuget doesn't allow to use lower version.

SimonCropp:
i just checked, and `<NoWarn>NU1605;NU1603</NoWarn>` still works to force a lower version

sungam3r:
Wow, ок.

PR #1752, opened 2020-07-31
make SuccessfullyValidatedResult public

SimonCropp:
At the moment in consuming code I have to have

```
var result = new ValidationResult(Enumerable.Empty<ValidationError>());
```


sungam3r:
OK. Approved API should be updated.

Issue #1189, opened 2019-06-17
throw new ExecutionError dose't work in Schema first


jianghongfei:
## Description

Throw an ExecutionError in Schema first example don't work as expected

### Steps to reproduce

Throw an ExecutionError in `Friend` method in **[Schema First Nested Types](https://graphql-dotnet.github.io/docs/getting-started/introduction#schema-first-nested-types)** example
```
public class Droid
{
  public string Id { get; set; }
  public string Name { get; set; }
}

public class Character
{
  public string Name { get; set; }
}

public class Query
{
  [GraphQLMetadata("hero")]
  public Droid GetHero()
  {
    return new Droid { Id = "1", Name = "R2-D2" };
  }
}

[GraphQLMetadata("Droid", IsTypeOf=typeof(Droid))]
public class DroidType
{
  public string Id(Droid droid) => droid.Id;
  public string Name(Droid droid) => droid.Name;

  // these two parameters are optional
  // ResolveFieldContext provides contextual information about the field
  public Character Friend(ResolveFieldContext context, Droid source)
  {
    throw new ExecutionError("Error Message")
  }
}

var schema = Schema.For(@"
  type Droid {
    id: String!
    name: String!
    friend: Character
  }

  type Character {
    name: String!
  }

  type Query {
    hero: Droid
  }
", _ => {
    _.Types.Include<DroidType>();
    _.Types.Include<Query>();
});

var json = schema.Execute(_ =>
{
  _.Query = "{ hero { id name friend { name } } }";
});
```

### Expected result
```
{
  "errors": [
    {
      "message": "Error Message",
    }
  ]
}
```
### Actual result

```
{
  "errors": [
    {
      "message": "Error trying to resolve friend.",
    }
  ]
}
```

### Environment
Windows 10, x64

sungam3r:
See #1158

sungam3r:
Also try set `ExecutionOptions.ExposeExceptions = true`

sungam3r:
Fixed in #1753

PR #1753, opened 2020-07-31
fix issue #1189

jiaguilera:
When using schema-first approach, resolvers are dynamically dispatched. This causes any exception raised inside resolver logic to be wrapped in a TargetInvocationException. As a result ExecutionError mechanism doesn't work properly (as described in #1189)


sungam3r:
Fixes #1189

sungam3r:
Thanks!

Issue #1736, opened 2020-07-22
Combine/refactor input variable validation code


Shane32:
These two methods both contain code to validate that input variables are valid:
* GraphQLExtensions.IsValidLiteralValue
* ExecutionHelper.AssertValidVariableValue

These could be combined to eliminate duplicate code and eliminate discrepancies between the logic of the two.

Shane32:
Merging these methods should cause the validation rule `ArgumentsOfCorrectType` to validate input variables, rather than assuming they are of the correct type.

PR #1763, opened 2020-08-07
Prevent unnecessary wrapping of ExecutionErrors in DocumentExecuter

Shane32:
Right now all unhandled `Exception`s are wrapped inside an `ExecutionError` within `DocumentExecuter` (unless `ThrowOnUnhandledException` is set).  This prevents additional data, such as location or codes, from being returned either from library code or from the `UnhandledExceptionDelegate`.  This can be seen in the `Input_Enum_InvalidEnum_Variable` test, as it can only return the location information with this change.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -63,10 +63,10 @@ namespace GraphQL.Tests.Bugs
         public void Input_Enum_Valid_Variable() => AssertQuerySuccess("query($arg: Bug1699Enum!) { input(arg: $arg) }", @"{ ""input"": ""Grumpy"" }", "{\"arg\":\"GRUMPY\"}".ToInputs());
 
         [Fact]
-        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to \u0027Bug1699Enum\u0027", 1, 7, (object[])null, inputs: "{\"arg\":\"DOPEY\"}");
+        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to \u0027Bug1699Enum\u0027", 1, 7, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
 
         [Fact]
-        public void Input_Enum_InvalidInt_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u00272\u0027 to \u0027Bug1699Enum\u0027", 1, 7, (object[])null, inputs: "{\"arg\":2}");
+        public void Input_Enum_InvalidInt_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u00272\u0027 to \u0027Bug1699Enum\u0027", 1, 7, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":2}");
 
         [Fact]
         public void Input_Enum_UndefinedDefault() => AssertQuerySuccess("{ input }", @"{ ""input"": ""Grumpy"" }");
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -63,10 +63,10 @@ namespace GraphQL.Tests.Bugs
         public void Input_Enum_Valid_Variable() => AssertQuerySuccess("query($arg: Bug1699Enum!) { input(arg: $arg) }", @"{ ""input"": ""Grumpy"" }", "{\"arg\":\"GRUMPY\"}".ToInputs());
 
         [Fact]
-        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to \u0027Bug1699Enum\u0027", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
+        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to \u0027Bug1699Enum\u0027", 1, 7, (object[])null, inputs: "{\"arg\":\"DOPEY\"}");
 
         [Fact]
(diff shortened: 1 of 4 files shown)


sungam3r:
Is this the only problematic place? I see many uses for this constructor.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -63,10 +63,10 @@ private void AssertQueryWithError(string query, string result, string message, i
         public void Input_Enum_Valid_Variable() => AssertQuerySuccess("query($arg: Bug1699Enum!) { input(arg: $arg) }", @"{ ""input"": ""Grumpy"" }", "{\"arg\":\"GRUMPY\"}".ToInputs());
 
         [Fact]
-        public void Input_Enum_InvalidEnum_Variable() => AssertQueryWithError(@"query($arg: Bug1699Enum!) { input(arg: $arg) }", null, "Variable \u0027$arg\u0027 is invalid. Unable to convert \u0027DOPEY\u0027 to \u0027Bug1699Enum\u0027", 0, 0, (object[])null, code: "INVALID_VALUE", inputs: "{\"arg\":\"DOPEY\"}");
```
sungam3r, line 66: I have a question why we are losing the error code. I can see that we are taking it from a inner exception. But what if there is no inner exception (as in your fixed version)? Is it worth setting the error code anyway, and what value will it have? And after that I thought about why we even do such checks at the execution stage. This must be validated by the validation rules.

Then I checked `ArgumentsOfCorrectType` and see this inside `IsValidLiteralValue`:
изображение

I think it's worth working on it.

Shane32:
So, it looks like we need to update the validation rules to validate the variable values, rather than relying on `GetVariableValues` to do the job, correct?  I agree and I can make a separate PR for that.

I plan to provide a number of small pull requests that will individually cleanup exceptions throughout the project.  For example, there are no tests for graphql parsing, and when a parsing error occurs, it gets wrapped and loses the location information.  So that is why I posted this PR first.

I was thinking that maybe we would have a page in the documentation with all the execution errors and error codes, so we can review them all at once and have it documented.  It would list any error that could be a result of a client issuing an invalid query.

Shane32:
> So, it looks like we need to update the validation rules to validate the variable values, rather than relying on `GetVariableValues` to do the job, correct?  I agree and I can make a separate PR for that.

Probably a duplicate of #1736 actually...

Shane32:
> Is this the only problematic place? I see many uses for this constructor.

I plan to fix other issues once this PR is merged.  There are inadequate tests to cover the applicable scenarios.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
Shane32: Technically the code is all performed correctly right now, even if not clear.  Variables are all validated within `GetVariableValues` prior to execution, returning an ExecutionError with the proper location information -- just not within a validation rule.  (This PR patches the exception handling code so the location is returned as intended.)  The above code is also correct because `GetVariableValues` will always validate the value types, so there is no need for redundant code to run within `IsValidLiteralValue`.

Perhaps it would be better if there was another validation rule such as `ValidateVariableValues`.  But this already occurs within `GetVariableValues`, and any additional validation rule would be redundant (since `GetVariableValues` still needs to deserialize the values prior to exeuction).

sungam3r:
> I have a question why we are losing the error code. I can see that we are taking it from a inner exception. But what if there is no inner exception (as in your fixed version)? Is it worth setting the error code anyway, and what value will it have? 

What about this?

sungam3r:
> Perhaps it would be better if there was another validation rule such as ValidateVariableValues

Each validation rule should conform one from spec with corresponding error code.

Shane32:
> > I have a question why we are losing the error code. I can see that we are taking it from a inner exception. But what if there is no inner exception (as in your fixed version)? Is it worth setting the error code anyway, and what value will it have?
> 
> What about this?

The code is meaningless ("EXECUTION_ERROR") if you create an ExecutionError from an inner ExecutionError.  It would be much better to return the code from the inner execution error, which this patch would do.

Shane32:
Remember that this code just puts the UnhandledExecutionDelegate on par with any field resolver, which are able to return an error with or without a code as the user desires.

Yes, we will need to add codes for ExecutionErrors generated by graphql.  For example, when a empty query is submitted.

Shane32:
> > Perhaps it would be better if there was another validation rule such as ValidateVariableValues
> 
> Each validation rule should conform one from spec with corresponding error code.

Fine with me.  This PR does not conflict with that.

sungam3r:
> But this already occurs within GetVariableValues, and any additional validation rule would be redundant (since GetVariableValues still needs to deserialize the values prior to exeuction).

Validate variables in rule is better than within execution stage. Even if we will have some redundant checks.

sungam3r:
> So, it looks like we need to update the validation rules to validate the variable values, rather than relying on GetVariableValues to do the job, correct? I agree and I can make a separate PR for that.

I think so.

sungam3r:
> The code is meaningless ("EXECUTION_ERROR") if you create an ExecutionError from an inner ExecutionError. It would be much better to return the code from the inner execution error, which this patch would do.

I know. I talk about different thing. I talk about returning code from first level exception (not always maybe) if inner exception was not supplied. In our case code equals INVALID_VALUE. In your PR this info was lost.

Shane32:
We should fix the library.  If `INVALID_VALUE` was intended to be returned, we should set it.  We have already discussed that it should not even have executed, and that it should be a validation rule with a proper code set.  I can make a PR for that.  In the meantime, I will set the code for this exception.

Again, this PR simply makes unhandled exceptions function the same as any field resolver, which if you did not set a code, it would not return one.

Shane32:
> Again, this PR simply makes unhandled exceptions function the same as any field resolver, which if you did not set a code, it would not return one.

Unhandled exceptions of any other type, like ArgumentNullException, still get wrapped and have the ARGUMENT_NULL code within them.

Shane32:
If you think that every error should return a code, then we could write code in ExecutionError, where if Code was null, then it will examine the type name of itself and generate a code.  Then it would take place across the entire library.  I did not think this was the intent.  I'm simply trying to make the code uniform across the codebase.

Shane32:
To summarize, there are four main points in the code where errors are generated:
1. Within a field resolver (e.g. a timeout connecting to a SQL server)
2. Within the graphql framework prior to the execution of the document (e.g. query is an empty string)
3. Within attached middleware or document executers
4. Rethrown from UnhandledExceptionDelegate

I think we should make ExecutionErrors uniform, where in all 4 cases they get passed directly back to the ExecutionResult, without any wrapping.  Currently it happens three different ways:
1. unwrapped
2. wrapped with no message prefix (lose location/code/data)
3. wrapped with no message prefix (lose location/code/data)
4. from field resolver: wrapped with message prefix (lose location/code/data) -- from document execution: without message prefix (lose location/code/data)

sungam3r:
> Yes, we will need to add codes for ExecutionErrors generated by graphql. For example, when a empty query is submitted.

See https://github.com/graphql/graphql-spec/issues/708

Shane32:
We could put Code in the `ExecutionError (string message)` constructor, requiring a code when not using an `InnerException`.  (But I would guess you'd want to hold off until it is part of the official spec.)

Shane32:
That way it applies to the entire project - field resolvers and etc.

sungam3r:
>  I can make a PR for that. In the meantime, I will set the code for this exception. Again, this PR simply makes unhandled exceptions function the same as any field resolver, which if you did not set a code, it would not return one.

sungam3r:
>  I can make a PR for that. In the meantime, I will set the code for this exception. Again, this PR simply makes unhandled exceptions function the same as any field resolver, which if you did not set a code, it would not return one.

Shane32:
> I want to say that we should not lose error code for INVALID_VALUE in GraphQL response before adding/modifying validation rule.

Understood.

PR #1764, opened 2020-08-07
Prevent unnecessary wrapping of ExecutionErrors in ExecutionStrategy

Shane32:
This allows `UnhandledExceptionDelegate`s to return custom messages, just like throwing an `ExecutionError` directly from a resolver can already do.

### DIFF
--- a/src/GraphQL.Tests/Errors/UnhandledExceptionTests.cs
+++ b/src/GraphQL.Tests/Errors/UnhandledExceptionTests.cs
@@ -29,11 +29,46 @@ namespace GraphQL.Tests.Errors
             });
         }
 
+        [Fact]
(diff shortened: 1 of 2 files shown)


src/GraphQL.Tests/Errors/UnhandledExceptionTests.cs
```diff
@@ -29,12 +29,47 @@ public void rethrows_unhandled_exception()
             });
         }
 
+        [Fact]
+        public void unhandled_error_delegate_can_rethrow_custom_exception()
```
sungam3r, line 33:
```suggestion
        public void unhandled_exception_delegate_can_rethrow_custom_exception()
```

PR #1765, opened 2020-08-07
Add location information to query parsing errors

Shane32:
Right now graphql syntax errors are passed through directly from the graphql parser library.  The error messages are multi-line and the location information is embedded inside the message string rather than attached to the graphql error response.  This PR parses the exception message from the graphql-parser library into the proper fields for the graphql response.

~~Note that this PR requires #1763 in order to work properly, as currently ExecutionErrors are wrapped, dropping location information.~~ Merged

The returned error code will not change with this PR (currently SYNTAX_ERROR).  In a separate PR, I will look at updating all of the error codes, including this one, that should be returned.

### DIFF
--- a/src/GraphQL/Execution/GraphQLDocumentBuilder.cs
+++ b/src/GraphQL/Execution/GraphQLDocumentBuilder.cs
@@ -1,7 +1,8 @@
-using System;
 using GraphQL.Language;
 using GraphQL.Language.AST;
 using GraphQLParser;
+using GraphQLParser.AST;
+using GraphQLParser.Exceptions;
 
 namespace GraphQL.Execution
 {
@@ -18,12 +19,12 @@ namespace GraphQL.Execution
         public Document Build(string body)
         {
             var source = new Source(body);
-            GraphQLParser.AST.GraphQLDocument result;
+            GraphQLDocument result;
             try
             {
                 result = _parser.Parse(source);
             }
-            catch (GraphQLParser.Exceptions.GraphQLSyntaxErrorException ex)
+            catch (GraphQLSyntaxErrorException ex)
             {
                 var e = new ExecutionError(ex.Description, ex);
                 e.AddLocation(ex.Line, ex.Column);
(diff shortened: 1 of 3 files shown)


Shane32:
@sungam3r What do you think of this?  I'm open to suggestions.  Just seemed like we could make an improvement here.

sungam3r:
I think that it will be **more** easy to just expose location info from [GraphQLSyntaxErrorException](https://github.com/graphql-dotnet/parser/blob/master/src/GraphQLParser/Exceptions/GraphQLSyntaxErrorException.cs#L18) rather then parse two ints back from rendered error message.

Shane32:
True. I could add a PR to that project...

sungam3r:
Pushed - https://www.nuget.org/packages/GraphQL-Parser/5.2.0

Shane32:
Let me know if any other changes have to be made for the graphql nuget package to reference the 5.2.0 version of the parser.

sungam3r:
Maybe it is worth to use different exception type? Well, parsing is not actual execution. Or we can use just `Validation error`.

Shane32:
That was going to be one of my next PRs - separating exception types throughout the project (only where appropriate of course). I didn’t want to make a suggestion until I thought it through fully. So for now it just mimics old behavior with the added location information. 

I’ll provide a chart that we can review together

src/GraphQL/Execution/GraphQLDocumentBuilder.cs
```diff
@@ -17,7 +19,17 @@ public GraphQLDocumentBuilder()
...
             var source = new Source(body);
-            var result = _parser.Parse(source);
+            GraphQLDocument result;
+            try
+            {
+                result = _parser.Parse(source);
+            }
+            catch (GraphQLSyntaxErrorException ex)
+            {
+                var e = new ExecutionError(ex.Description, ex);
```
sungam3r, line 29:
```suggestion
                var e = new ExecutionError("Error parsing query: " + ex.Description, ex);
```

PR #1767, opened 2020-08-09
Wrap exception from ScalarGraphType.ParseXXX into InvalidValueException (prepare to fix validation rule for parsing variables)

Shane32:
Right now, while parsing variables, if the value is out of range, such as 300 for a `byte`, an `OverflowException` occurs during the cast.  The exception is not caught and is passed back to the `DocumentExecuter`, which wraps it in a `ExecutionError`.

This PR catches any exception raised while parsing the value and throws an `InvalidValueException`, just the same as if `null` (indicating a failed conversion) were returned from `ParseValue`.

This PR adds a test `Input_Byte_Invalid_Variable` to test for this scenario.  Note that we have yet to assign a proper code to `InvalidValueException` (or replace it with `ValidationError`).

I would like to merge this PR before writing a validation rule that validates the variable values, so all the proper tests are in place first.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
@@ -8,7 +8,7 @@ using Xunit;
 namespace GraphQL.Tests.Bugs
 {
     // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1766
-    public class Bug1766InvalidByte : QueryTestBase<Bug1766Schema>
+    public class Bug1767InvalidByte : QueryTestBase<Bug1767Schema>
     {
         private void AssertQueryWithError(string query, string result, string message, int line, int column, object[] path, Exception exception = null, string code = null, string inputs = null)
         {
@@ -34,17 +34,17 @@ namespace GraphQL.Tests.Bugs
         public void Input_Byte_Invalid_Argument() => AssertQueryWithError("query { input(arg: 300) }", null, "Argument \"arg\" has invalid value 300.\nExpected type \"Byte\", found 300.", 1, 15, null, code: "5.3.3.1");
     }
 
-    public class Bug1766Schema : Schema
+    public class Bug1767Schema : Schema
     {
-        public Bug1766Schema()
+        public Bug1767Schema()
         {
-            Query = new Bug1766Query();
+            Query = new Bug1767Query();
         }
     }
 
-    public class Bug1766Query : ObjectGraphType
+    public class Bug1767Query : ObjectGraphType
     {
-        public Bug1766Query()
+        public Bug1767Query()
         {
             Field<StringGraphType>(
                 "input",
--- a/src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
@@ -7,7 +7,7 @@ using Xunit;
 
 namespace GraphQL.Tests.Bugs
 {
-    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1766
+    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1767
     public class Bug1767InvalidByte : QueryTestBase<Bug1767Schema>
     {
         private void AssertQueryWithError(string query, string result, string message, int line, int column, object[] path, Exception exception = null, string code = null, string inputs = null)
--- a/src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
@@ -28,7 +28,7 @@ namespace GraphQL.Tests.Bugs
         public void Input_Byte_Valid_Argument() => AssertQuerySuccess("query { input(arg: 2) }", @"{ ""input"": ""2"" }", "{\"arg\":2}".ToInputs());
 
         [Fact]
-        public void Input_Byte_Invalid_Variable() => AssertQueryWithError("query($arg: Byte!) { input(arg: $arg) }", null, "Variable '$arg' is invalid. Unable to convert '300' to 'Byte'", 1, 7, null, null, "INVALID_VALUE", "{\"arg\":300}");
+        public void Input_Byte_Invalid_Variable() => AssertQueryWithError("query($arg: Byte!) { input(arg: $arg) }", null, "Variable '$arg' is invalid. Unable to convert '300' to 'Byte'", 1, 7, null, new OverflowException(), "INVALID_VALUE", "{\"arg\":300}");
 
         [Fact]
         public void Input_Byte_Invalid_Argument() => AssertQueryWithError("query { input(arg: 300) }", null, "Argument \"arg\" has invalid value 300.\nExpected type \"Byte\", found 300.", 1, 15, null, code: "5.3.3.1");
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -148,29 +148,27 @@ namespace GraphQL.Execution
 
                 if (input is IValue value)
                 {
-                    object convertedValue = null;
                     try
                     {
-                        convertedValue = scalar.ParseLiteral(value);
+                        if (scalar.ParseLiteral(value) == null)
+                            throw new InvalidValueException(variableName, $"Unable to convert '{value.Value}' to '{type.Name}'");
                     }
-                    catch
+                    catch (Exception ex)
                     {
+                        throw new InvalidValueException(variableName, $"Unable to convert '{value.Value}' to '{type.Name}'", ex);
                     }
-                    if (convertedValue == null)
-                        throw new InvalidValueException(variableName, $"Unable to convert '{value.Value}' to '{type.Name}'");
                 }
                 else
                 {
-                    object convertedValue = null;
                     try
                     {
-                        convertedValue = scalar.ParseValue(input);
+                        if (scalar.ParseValue(input) == null)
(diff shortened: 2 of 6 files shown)


src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
```diff
@@ -0,0 +1,55 @@
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using System;
+using System.Linq;
+using System.Numerics;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1766
```
sungam3r, line 10:
```suggestion
    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1767
```

src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
```diff
@@ -0,0 +1,55 @@
...
+        public void Input_Byte_Valid_Variable() => AssertQuerySuccess("query($arg: Byte!) { input(arg: $arg) }", @"{ ""input"": ""2"" }", "{\"arg\":2}".ToInputs());
+
+        [Fact]
+        public void Input_Byte_Valid_Argument() => AssertQuerySuccess("query { input(arg: 2) }", @"{ ""input"": ""2"" }", "{\"arg\":2}".ToInputs());
+
+        [Fact]
+        public void Input_Byte_Invalid_Variable() => AssertQueryWithError("query($arg: Byte!) { input(arg: $arg) }", null, "Variable '$arg' is invalid. Unable to convert '300' to 'Byte'", 1, 7, null, null, "INVALID_VALUE", "{\"arg\":300}");
+
+        [Fact]
+        public void Input_Byte_Invalid_Argument() => AssertQueryWithError("query { input(arg: 300) }", null, "Argument \"arg\" has invalid value 300.\nExpected type \"Byte\", found 300.", 1, 15, null, code: "5.3.3.1");
```
sungam3r, line 34: I think we should unify '' and "" inside messages.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -148,12 +148,28 @@ public static void AssertValidVariableValue(ISchema schema, IGraphType type, obj
...
                 }
                 else
                 {
-                    if (scalar.ParseValue(input) == null)
+                    object convertedValue = null;
+                    try
+                    {
+                        convertedValue = scalar.ParseValue(input);
+                    }
+                    catch
```
sungam3r, line 169: Let's use thrown exception as inner exception for InvalidValueException.

src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
Shane32: I agree, but there are many uses of single quotes and double quotes throughout the project.  It would need to be a separate PR.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Ok but I want to be sure that `Data` from an inner exception does not get exposed outside the project, especially here where it is a query error (bad data), and not an unhandled exception.  Also, ExceptionError does not properly allow for setting the code to override the default.  I will make a separate PR to fix these issues, and then come back to this PR.

Shane32: Actually I'll just skip the code "issue", and throw the new constructor in this PR.

src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
sungam3r: Yes.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: If you are going to fix something in other PRs, temporary bypassing review remarks leaving such places unchanged, then I would really like to have the corresponding `//TODO: ` comments. I do not mind promoting the master branch, but only in such a way as to keep pending tasks in front of my eyes. Ok?

src/GraphQL.Tests/Bugs/Bug1767InvalidByte.cs
Shane32: > Is it really necessary? Do we need `copyData` parameter?

No it is not strictly necessary.  I can execute `Data.Clear();` inside the `InvalidValueException` constructor instead.

> And why is it equal to `false` here?

This error is a syntax error, only occurring due to an invalid request.  Any information within the `Data` property could potentially be sensitive data and should not be returned to the client.  I'm not sure that the additional codes (e.g. `OVERFLOW`) should be returned either, but if you feel this is useful, that's fine with me.  Note that there are no other places where an ExecutionError is generated from infrastructure with an inner exception (besides unhandled exceptions), so this is the only spot where this constructor would be is used.

See also: https://github.com/graphql-dotnet/graphql-dotnet/issues/1641

I personally cannot see how at any time one would want the Data property to be copied.  So if it was me, I'd remove the `SetData` line of code from the `ExecutionError` constructor.  But at least this can be cleared within the `UnhandledExceptionDelegate` for unhandled exceptions.  (This particular error is a handled exception, not an unhandled exception.)

Let me know what you prefer - add the constructor, or use `Data.Clear();`.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I'm happy to oblige but want to make sure I understand you correctly.  You are saying that if you make a comment, and I suggest we fix in another PR, then I should add a comment to the code that says `//TODO:` -- for example, where you suggested that we unify the single and double quotation marks?  I'm not sure what you mean about promoting the master branch.

Shane32: (Usually I try to create a new issue in these cases, but am not always diligent in that.)

Shane32:
> Is it really necessary? Do we need `copyData` parameter?

No it is not strictly necessary.  I can execute `Data.Clear();` inside the `InvalidValueException` constructor instead.

> And why is it equal to `false` here?

This error is a syntax error, only occurring due to an invalid request.  Any information within the `Data` property could potentially be sensitive data and should not be returned to the client.  I'm not sure that the additional codes (e.g. `OVERFLOW`) should be returned either, but if you feel this is useful, that's fine with me.  Note that there are no other places where an ExecutionError is generated from infrastructure with an inner exception (besides unhandled exceptions), so this is the only spot where this constructor would be is used.

See also: https://github.com/graphql-dotnet/graphql-dotnet/issues/1641

I personally cannot see how at any time one would want the Data property to be copied.  So if it was me, I'd remove the `SetData` line of code from the `ExecutionError` constructor.  But at least this can be cleared within the `UnhandledExceptionDelegate` for unhandled exceptions.  (This particular error is a handled exception, not an unhandled exception.)

Let me know what you prefer - add the constructor, or use `Data.Clear();`.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: I am not a native speaker. I mean I am not agains pushing in master PRs with some pending little issues not directly related to PR. Just marking these places in code with `// TODO: ` please, so as not to forget. Or making a new issue, yes. In case of `InvalidValueException` some data was lost. I always pay attention to such changes. My suggestion about unifying quotations is a different thing. This was just an abstract remark.

Shane32: I understand - thanks for the clarification!

sungam3r:
As you rightly noted, nested exceptions can occur in other cases as well. All filtering of sensitive data can be successfully done at the response serialization level, #1641 is exactly about that IIRC.

> I personally cannot see how at any time one would want the Data property to be copied. 

We use this in our projects. If the authentication fails, additional data is sent there. It could have been done differently, but nevertheless now it works like this.

> Let me know what you prefer - add the constructor, or use `Data.Clear();`.

I don't think any of this is needed. Let the developer to decide (#1641?).

Shane32:
Well I wouldn't as it's an infrastructure error, not user error, but okay.

sungam3r:
Perhaps, but then it is necessary to solve the problem in the general case, and this is the subject of another PR. Better to solve it in #1641.

Shane32:
@sungam3r Any other questions on this PR?

src/GraphQL/Execution/InvalidValueException.cs
```diff
@@ -10,5 +10,12 @@ public class InvalidValueException : ExecutionError
         {
             Code = "INVALID_VALUE";
         }
+
+        public InvalidValueException(string variableName, string message, Exception innerException) :
+            base($"Variable '${variableName}' is invalid. {message}", innerException)
+        {
+            Code = "INVALID_VALUE";
+            // note: Codes will also return codes of inner exceptions, plus the Data in the inner exception
```
sungam3r, line 18: Codes will also return Data in the inner exception ? 😕

sungam3r:
I'm ready to merge just one small question about comment above.

src/GraphQL/Execution/InvalidValueException.cs
Shane32: lol - bad choice of words there - I'll fix it!

Shane32:
```suggestion
            // note: this ExecutionError will also return Codes of inner exceptions, plus the Data in the inner exception
```

Shane32: Is that clear?

sungam3r: Yes

PR #1769, opened 2020-08-10
Add checks for null DocumentExecuter parameters

Shane32:
With these changes, if `options.Query` or `options.Schema` is `null`, an exception is raised back to the caller, rather than a graphql result being returned.  This also validates that the constructor parameters are valid.

If `options.Query` is an empty string or only consists of whitespace, a ~~`SYNTAX_ERROR`~~ `NO_OPERATION` graphql error result is returned to the caller.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1769.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1769.cs
@@ -62,7 +62,10 @@ namespace GraphQL.Tests.Bugs
         }
 
         [Fact]
-        public void query_is_empty() => AssertQueryWithError("", null, "Unable to determine operation from query.", 0, 0, (object[])null);
+        public void query_is_empty() => AssertQueryWithError("", null, "A query is required.", 0, 0, (object[])null);
+
+        [Fact]
+        public void query_is_whitespace() => AssertQueryWithError("\t \t \r\n", null, "A query is required.", 0, 0, (object[])null);
 
         [Fact]
         public void DocumentExecuter_cannot_have_null_constructor_parameters()
--- a/src/GraphQL.Tests/Bugs/Bug1769.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1769.cs
@@ -29,7 +29,7 @@ namespace GraphQL.Tests.Bugs
         public async Task DocumentExecuter_has_valid_options()
         {
             var de = new DocumentExecuter();
-            var valid = await de.ExecuteAsync(new ExecutionOptions()
+            var valid = await de.ExecuteAsync(new ExecutionOptions
             {
                 Query = "{test}",
                 Schema = Schema,
--- a/src/GraphQL.Tests/Bugs/Bug1769.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1769.cs
@@ -66,7 +66,8 @@ namespace GraphQL.Tests.Bugs
         [Fact]
         public void DocumentExecuter_cannot_have_null_constructor_parameters()
         {
-            var valid = new DocumentExecuter(new GraphQLDocumentBuilder(), new DocumentValidator(), new ComplexityAnalyzer());
+            var valid1 = new DocumentExecuter();
+            var valid2 = new DocumentExecuter(new GraphQLDocumentBuilder(), new DocumentValidator(), new ComplexityAnalyzer());
             Assert.Throws<ArgumentNullException>(() => new DocumentExecuter(null, new DocumentValidator(), new ComplexityAnalyzer()));
             Assert.Throws<ArgumentNullException>(() => new DocumentExecuter(new GraphQLDocumentBuilder(), null, new ComplexityAnalyzer()));
             Assert.Throws<ArgumentNullException>(() => new DocumentExecuter(new GraphQLDocumentBuilder(), new DocumentValidator(), null));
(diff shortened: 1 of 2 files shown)


sungam3r:
I think it requires to review test cases in the server project.

src/GraphQL.Tests/Bugs/Bug1769.cs
```diff
@@ -0,0 +1,91 @@
...
+            error.Path = path;
+            if (code != null)
+                error.Code = code;
+            var expected = CreateQueryResult(result, new ExecutionErrors { error });
+            AssertQueryIgnoreErrors(query, expected, inputs?.ToInputs(), renderErrors: true, expectedErrorCount: 1);
+        }
+
+        [Fact]
+        public async Task DocumentExecuter_has_valid_options() {
+            var de = new DocumentExecuter();
```
sungam3r, line 30:
```suggestion
        public async Task DocumentExecuter_has_valid_options()
        {
            var de = new DocumentExecuter();
```

src/GraphQL.Tests/Bugs/Bug1769.cs
```diff
@@ -0,0 +1,91 @@
...
+            var valid = await de.ExecuteAsync(new ExecutionOptions()
```
sungam3r, line 31:
```suggestion
            var valid = await de.ExecuteAsync(new ExecutionOptions
```

src/GraphQL.Tests/Bugs/Bug1769.cs
```diff
@@ -0,0 +1,91 @@
...
+            });
+        }
+
+        [Fact]
+        public void query_is_empty() => AssertQueryWithError("", null, "Unable to determine operation from query.", 0, 0, (object[])null);
+
+        [Fact]
+        public void DocumentExecuter_cannot_have_null_constructor_parameters()
+        {
+            var valid = new DocumentExecuter(new GraphQLDocumentBuilder(), new DocumentValidator(), new ComplexityAnalyzer());
```
sungam3r, line 69:
```suggestion
            var valid1 = new DocumentExecuter();
            var valid2 = new DocumentExecuter(new GraphQLDocumentBuilder(), new DocumentValidator(), new ComplexityAnalyzer());
```

src/GraphQL.Tests/Bugs/Bug1769.cs
```diff
@@ -0,0 +1,91 @@
...
+                {
+                    Query = "{test}",
+                    Schema = Schema,
+                    FieldMiddleware = null,
+                });
+            });
+        }
+
+        [Fact]
+        public void query_is_empty() => AssertQueryWithError("", null, "Unable to determine operation from query.", 0, 0, (object[])null);
```
sungam3r, line 64: Please add test for `"\t    \t \r\n"` query.

Shane32: Done

Shane32:
Apparently the document parser does not throw an exception if the query is blank -- but rather there are no operations.  I added code to test for this condition, with the original error message of `A query is required.` as it was before.  As described in the notes above, it will now return a SYNTAX_ERROR code just the same as if the document parser was unable to parse the query. 
 Perhaps we could revise the error message, but I'm not sure what to write.

Shane32:
> I think it requires to review test cases in the server project.

Is there an easy way to do this?  If you have it all set up, can you run the tests real quick and see if we need to make changes there?

sungam3r:
> Apparently the document parser does not throw an exception if the query is blank -- but rather there are no operations. I added code to test for this condition, with the original error message of A query is required. as it was before. As described in the notes above, it will now return a SYNTAX_ERROR code just the same as if the document parser was unable to parse the query.
Perhaps we could revise the error message, but I'm not sure what to write.

Right. Empty document creates empty AST. Formally it is not invalid syntax tree, it is just empty tree. In order not to confuse the client, we should make another error message/code.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -87,6 +76,14 @@ public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
                     }
                 }
 
+                if (document.Operations.Count == 0)
+                {
+                    throw new ExecutionError("A query is required.")
```
sungam3r, line 81:
```suggestion
                    throw new ExecutionError("Cannot execute query if no operation is specified.")
```

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -87,6 +76,14 @@ public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
...
+                    {
+                        Code = "SYNTAX_ERROR"
```
sungam3r, line 83:
```suggestion
                        Code = "NO_OPERATION"
```
?

sungam3r:
> Is there an easy way to do this? If you have it all set up, can you run the tests real quick and see if we need to make changes there?

In principle, you can specify a nuget package from the AppVeyor PR build, or we can simply check the result after merging this PR. Most likely, there won't be anything long to fix or add.

Shane32:
> Right. Empty document creates empty AST. Formally it is not invalid syntax tree, it is just empty tree. In order not to confuse the client, we should make another error message/code.

Interesting.  Makes sense.

Shane32:
This PR contains changes also within #1658

PR #1772, opened 2020-08-11
Change DocumentExecuter to throw for invalid operation name

Shane32:
Right now, if `options.OperationName` is invalid, the `DocumentExecuter` treats this as a query error, whereas this should be an `InvalidOperationException` thrown back to the caller, since it is an invalid option passed to the `DocumentExecuter`.

However, the document needs to be parsed before this option can be validated, and document parsing takes place within the `try { }` block.  I propose we select one of the following options:

1. Throw `InvalidOperationException`, and let this get caught by the unhandled exception processing - for instance, throwing immedaitely when `ThrowOnUnhandledException` is true, or wrapping it when it is false.
2. Create a small amount of additional logic to throw this exception directly to the caller.

Another option (option 3) would be to leave it as a "query error" (rather than an invalid operation), but at least revise the error message to be more appropriate for the situation.

Old error message | New error message
-|-
Unable to determine operation from query. | Query does not contain operation 'firstOperation'

This PR currently is written for option 1.  Comments?

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1772.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1772.cs
@@ -1,13 +1,7 @@
-using GraphQL.Execution;
-using GraphQL.SystemTextJson;
-using GraphQL.Types;
-using GraphQL.Validation;
-using GraphQL.Validation.Complexity;
-using Shouldly;
 using System;
-using System.Linq;
-using System.Numerics;
 using System.Threading.Tasks;
+using GraphQL.Types;
+using Shouldly;
 using Xunit;
 
 namespace GraphQL.Tests.Bugs
--- a/src/GraphQL.Tests/Bugs/Bug1772.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1772.cs
@@ -12,8 +12,8 @@ using Xunit;
 
 namespace GraphQL.Tests.Bugs
 {
-    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1770
-    public class Bug1770 : QueryTestBase<Bug1770Schema>
+    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1772
+    public class Bug1772 : QueryTestBase<Bug1772Schema>
     {
         [Theory]
         [InlineData("")]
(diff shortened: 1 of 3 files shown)


src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -89,7 +89,7 @@ public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
 
                 if (operation == null)
                 {
-                    throw new ExecutionError("Unable to determine operation from query.");
+                    throw new InvalidOperationException($"Query does not contain operation '{options.OperationName}'.");
```
sungam3r, line 92: I want to draw your attention to a problem in terminology. This issue was discussed in the GraphQL spec repository. Formally, the request text cannot be identified with the query. We must not forget that there are mutations and subscriptions. This error message can be confusing in case of mutation or subscriptions. One could say `GraphQL request does not contain...` or something similar. In any case, this remark refers rather to the general approach of composing message texts, and not to a specific piece of code. I am ok with the text that is now. If you want you can create a new issue about such wording.

Shane32: 👍 This can be reviewed at the same time as the quotes vs double quotes

src/GraphQL.Tests/Bugs/Bug1772.cs
```diff
@@ -0,0 +1,75 @@
...
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using GraphQL.Validation;
+using GraphQL.Validation.Complexity;
+using Shouldly;
+using System;
+using System.Linq;
+using System.Numerics;
+using System.Threading.Tasks;
+using Xunit;
```
sungam3r, line 11: It seems to be unused usings.

sungam3r:
I'm for option 1; throwing this exception directly to the caller is a breaking change and I'm not sure at all if we should throw an exception directly.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: See #1775

src/GraphQL.Tests/Bugs/Bug1772.cs
Shane32: Wasn't sure if you cared since it was in a test.  No problem, I will keep them clean in tests from now on.

Shane32:
> I'm for option 1; throwing this exception directly to the caller is a breaking change and I'm not sure at all if we should throw an exception directly.

Sounds good; that's how I coded this PR so it should be all set.

src/GraphQL.Tests/Bugs/Bug1772.cs
sungam3r: I care about it.

Issue #1768, opened 2020-08-09
Exception types and codes need to be updated throughout project


Shane32:
I would like to propose that the exception types and codes be reviewed throughout the project.  Below is a table of what revisions are proposed.

## Exceptions occurring during schema construction

These will not change, and consist of exceptions such as `ArgumentNullException`, `ArgumentOutOfRangeException`, and so forth.  An exception of type `ExecutionError` will not be thrown.

## Exceptions occurring prior to document execution

These exceptions are not wrapped into an `ExecutionResult`, but are thrown directly from `DocumentExecuter.ExecuteAsync` to its caller.

Description | Exception
-|-
options == null | ArgumentNullException
options.Schema == null | InvalidOperationException
options.FieldMiddleware == null | InvalidOperationException
options.Query == null | InvalidOperationException

## Validation errors

All validation errors return an exception of type ValidationError.

Rule | Code now | New code
-|-|-
ArgumentsOfCorrectType | 5.3.3.1 | 5.6.1
DefaultValuesOfCorrectType | 5.7.2 | 5.6.1
FieldsOnCorrectType | 5.2.1 | 5.3.1
FragmentsOnCompositeTypes | 5.4.1.3 | 5.5.1.3
KnownArgumentNames | 5.3.1 | 5.4.1
KnownDirectives | 5.6.1 | 5.7.1
KnownFragmentNames | 5.4.2.1 | 5.5.1.4
KnownTypeNames | 5.4.1.2 | 5.5.1.2
LoneAnonymousOperation | 5.1.2.1 | 5.2.2.1
NoFragmentCycles | 5.4 | 5.5.2.2
NoUndefinedVariables | 5.7.4 | 5.8.3
NoUnusedFragments |  5.4.1.4 | 5.5.1.4
NoUnusedVariables | 5.7.5 | 5.8.4
OverlappingFieldsCanBeMerged | 5.3.2 | 5.3.2
PossibleFragmentSpreads | 5.4.2.3 | 5.5.2.3
ProvidedNonNullArguments | 5.3.3.2 | 5.4.2.1
ScalarLeafs | 5.2.3 | 5.3.3
SingleRootFieldSubscriptions | 5.2.3.1 | 5.2.3.1
UniqueArgumentNames | 5.3.2 | 5.4.2
UniqueDirectivesPerLocation | 5.6.3 | 5.7.3
UniqueFragmentNames | 5.4.1.1 | 5.5.1.1
[...]

PR #1773, opened 2020-08-11
Clarify and unify conversion exceptions from resolver returned values

Shane32:
Right now, after a resolver returns a value, the infrastructure attempts to convert it to the correct graph type.  For example, if the resolver returns a string of "1" for an IntGraphType, the infrastructure attempts to convert it to the number 1.  When this fails, the conversion throws an exception of FormatException which can be handled and logged by the UnhandledExceptionDelegate.  It is then wrapped in a generic "failure resolving {fieldname}" message, which is accurate - this is a server side code issue.  The ExecutionError also properly includes a code of "FORMAT" in this case.

However, if the resolver returns a string of "test" for an enumeration graph type (where "test" is invalid), the infrastructure throws a ExecutionError.  This cannot be caught or logged by the UnhandledExceptionDelegate, and the error inappropriately gets returned directly to the graphql client.  It does not have a code at all.

This PR fixes that so that any inappropriate returned value from a field resolver will:
* be able to be caught by the UnhandledExceptionDelegate
* wrapped and exact error message hidden from clients by default
* include a proper code

Note that, as was documented in #1699, the graphql infrastructure does not yet fully meet spec, where it will return members of lists that are able to be serialized, with nulls and individual errors for unserializable values.  This is not addressed in this PR.

Thanks to @GrangerAts for his work on #1658 which this is partially based off of.  Also see #1768

I will keep #1768 up to date, and in another PR we might create different exception classes rather than using InvalidOperationException for all of these cases, possibly wrapping serialization exceptions, similar to #1767.  This PR just contains the basic infrastructure and tests needed to continue this cleanup.

### DIFF
--- a/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
+++ b/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
@@ -106,9 +106,9 @@ namespace GraphQL.Tests.Bugs
             var data = new Data {ListOfStrings = new List<string> {"text", null, null}};
             var errors = new[]
             {
-                new ExecutionError("Cannot return null for non-null type. Field: listOfNonNullable, Type: [String!].")
+                new ExecutionError("Error trying to resolve listOfNonNullable.")
                 {
-                    Path = new object[] {"nonNullableDataGraph", "listOfNonNullable", 1}
+                    Path = new object[] {"nonNullableDataGraph", "listOfNonNullable"}
                 }
             };
 
@@ -158,9 +158,9 @@ namespace GraphQL.Tests.Bugs
             var errors = new[]
             {
                 new ExecutionError(
-                    "Cannot return null for non-null type. Field: nonNullableListOfNonNullable, Type: [String!]!.")
+                    "Error trying to resolve nonNullableListOfNonNullable.")
                 {
-                    Path = new object[] {"nullableDataGraph", "nonNullableListOfNonNullable", 1}
+                    Path = new object[] {"nullableDataGraph", "nonNullableListOfNonNullable"}
                 }
             };
 
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -32,17 +32,17 @@ namespace GraphQL.Tests.Bugs
 
         // within C#, (int)Bug1699Enum.Happy does not equal Bug1699.Happy
         [Fact]
-        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Unable to serialize '1' to \u0027Bug1699Enum\u0027", 1, 3, "happy");
+        public void Int_Enum() => AssertQueryWithError("{ happy }", @"{ ""happy"": null }", "Error trying to resolve happy.", 1, 3, "happy", exception: new InvalidOperationException());
 
         [Fact]
-        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnum");
+        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Error trying to resolve invalidEnum.", 1, 3, "invalidEnum", exception: new InvalidOperationException());
 
         // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
         [Fact]
-        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnumWithinList");
+        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Error trying to resolve invalidEnumWithinList.", 1, 3, "invalidEnumWithinList", exception: new InvalidOperationException());
 
         [Fact]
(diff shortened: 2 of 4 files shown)


src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
```diff
@@ -106,9 +106,9 @@ public void ListOfNonNull_containing_null_should_bubble_up_the_null()
             var data = new Data {ListOfStrings = new List<string> {"text", null, null}};
             var errors = new[]
             {
-                new ExecutionError("Cannot return null for non-null type. Field: listOfNonNullable, Type: [String!].")
+                new ExecutionError("Error trying to resolve listOfNonNullable.")
                 {
-                    Path = new object[] {"nonNullableDataGraph", "listOfNonNullable", 1}
+                    Path = new object[] {"nonNullableDataGraph", "listOfNonNullable"}
```
Shane32, line 111: The list index is lost from the Path property in this particular case.  However, that matches the updated message, as in this case the list is a list of non-nullable objects, and the entire object was unable to be resolved.  I have added the index to the message property of the inner exception.  See test `list_throws_for_invalid_type_when_conversion_returns_null`.  Note that if the list was nullable, then the error should be returned for the particular item in the list, and that item set to null.  But this is not yet supported by the graphql infrastructure, as noted in the PR.

Shane32:
Perhaps `InvalidCastException` would be more appropriate than `InvalidOperationException`

Shane32:
@sungam3r Any comments before I merge this?

sungam3r:
I would like to review this today or tomorrow.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -95,8 +95,7 @@ public static void SetArrayItemNodes(ExecutionContext context, ArrayExecutionNod
 
             if (!(parent.Result is IEnumerable data))
             {
-                var error = new ExecutionError("User error: expected an IEnumerable list though did not find one.");
-                throw error;
+                throw new InvalidOperationException("Expected an IEnumerable list though did not find one.");
```
sungam3r, line 98:
```suggestion
                throw new InvalidOperationException($"Expected an IEnumerable list though did not find one. Found: {parent.Result?.GetType().Name}");
```

Shane32:
@sungam3r Good here?

sungam3r:
I will review this tonight. Probably also 1571.

src/GraphQL.Tests/Bugs/Bug1773.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    public class Bug1773Query : ObjectGraphType
+    {
+        public Bug1773Query()
+        {
+            Field<ListGraphType<IntGraphType>>("testListValid", resolve: context => new object[] { 123 });
+            Field<ListGraphType<IntGraphType>>("testListInvalid", resolve: context => 123);
+            Field<ListGraphType<IntGraphType>>("testListInvalidType", resolve: context => new object[] { "test" });
+            Field<ListGraphType<EnumerationGraphType<Bug1773Enum>>>("testListInvalidType2", resolve: context => new object[] { "test" });
+            Field<ListGraphType<NonNullGraphType<IntGraphType>>>("testListNullValid", resolve: context => new object[] { 123 });
```
sungam3r, line 103:
```suggestion
            Field<ListGraphType<NonNullGraphType<IntGraphType>>>("testListNonNullValid", resolve: context => new object[] { 123 });
```

and so on for other non null fields, otherwise the meaning is inverted

src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
sungam3r: > Note that if the list was nullable

Do you mean _if the list **elements** were nullable_ ?

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -32,17 +32,17 @@ private void AssertQueryWithError(string query, string result, string message, i
...
 
         [Fact]
-        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnum");
+        public void Invalid_Enum() => AssertQueryWithError("{ invalidEnum }", @"{ ""invalidEnum"": null }", "Error trying to resolve invalidEnum.", 1, 3, "invalidEnum", exception: new InvalidOperationException());
 
         // TODO: does not yet fully meet spec (does not return members of the enum that are able to be serialized, with nulls and individual errors for unserializable values)
         [Fact]
-        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Unable to serialize '50' to \u0027Bug1699Enum\u0027", 1, 3, "invalidEnumWithinList");
+        public void Invalid_Enum_Within_List() => AssertQueryWithError("{ invalidEnumWithinList }", @"{ ""invalidEnumWithinList"": null }", "Error trying to resolve invalidEnumWithinList.", 1, 3, "invalidEnumWithinList", exception: new InvalidOperationException());
```
sungam3r, line 42: On the one hand, I do not like that the error message for the client will become less detailed. On the other hand, I understand that it was more of an error message to the server author because it described an internal execution error. I think you are correct in doing so.

src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
sungam3r: OK

src/GraphQL.Tests/Bugs/Bug1773.cs
sungam3r: Nevermind, I'm OK with current naming.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
Shane32: Just FYI, the inner error message still contains the same error message, so server logging won't suffer.

sungam3r: I know. It's OK.

PR #1777, opened 2020-08-11
Cleanup uses of Invariant.Check

Shane32:
These exceptions should not be throwing an ExecutionError, but rather are invalid for the state of the procedure.  Therefor I've change them to ArgumentNullException, ArgumentOutOfRangeException, and InvalidOperationException.  Most likely they are impossible to trigger from normal use of the library.

If these exceptions were to occur, they would be caught by the unhandled exception delegate and wrapped appropriately with a code attached.

These are the only 3 uses of the Invariant.Check method.  ~~We can discuss in a follow-up PR if the method should be removed or marked as Deprecated~~  The Invariant class has been removed.


Shane32:
Also see #1658

Shane32:
The unused Invariant class has been removed.

PR #1778, opened 2020-08-11
Cleanup AST converter exceptions

Shane32:
All of the errors changed below should never occur during normal operation, but would represent a coding error.  They have been changed to a native exception type.  If it is possible to produce these exceptions, they would be wrapped into an ExecutionError inside the DocumentExecuter with a code that represents the inner error.

There is one exception within CoreToAstConverter that has not been changed - where a literal number is being parsed.  This should have a proper code attached to the ExecutionError and a proper test added.  A note has been added to the codebase; this will be changed in a separate PR.


Shane32:
Also see #1658

src/GraphQL/Language/AST/Document.cs
```diff
@@ -36,7 +36,7 @@ public void AddDefinition(IDefinition definition)
             }
             else
             {
-                throw new ExecutionError("Unhandled document definition");
+                throw new ArgumentOutOfRangeException(nameof(definition), "Unhandled document definition");
```
sungam3r, line 39:
```suggestion
                throw new ArgumentOutOfRangeException(nameof(definition), $"Unhandled document definition '{definition.GetType().Name}'");
```

src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -225,6 +226,7 @@ public IValue Value(GraphQLValue source)
                         return new BigIntValue(bigIntegerResult).WithLocation(str, _body);
                     }
 
+                    // TODO: add code and test
```
sungam3r, line 229: 👍

PR #1779, opened 2020-08-11
Cleanup GraphTypesLookup exceptions

Shane32:
All of the errors changed below should never occur during normal operation, but would represent a schema error. They can only occur while the `GraphTypesLookup` class is being built (during `Schema.Initialize`).  Any information contained within these errors could potentially represent privileged information.

These exceptions have been changed to a native exception type. If one of these exceptions were to occur, they would be wrapped into an ExecutionError inside the DocumentExecuter with a code that represents the inner error (since the schema is initialized during document execution, if it has not yet been initialized).

### DIFF
--- a/src/GraphQL.Tests/Types/SchemaTests.cs
+++ b/src/GraphQL.Tests/Types/SchemaTests.cs
@@ -83,7 +83,7 @@ namespace GraphQL.Tests.Types
         public void throw_error_on_missing_istypeof()
         {
             var schema = new InvalidUnionSchema();
-            //note: The exception occurs during CreateTypesLookup, not during FindType
+            //note: The exception occurs during Schema.CreateTypesLookup(), not during Schema.FindType()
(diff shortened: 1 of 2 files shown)


Shane32:
Also see #1658

src/GraphQL.Tests/Types/SchemaTests.cs
```diff
@@ -83,7 +83,8 @@ public void registers_union_types()
         public void throw_error_on_missing_istypeof()
         {
             var schema = new InvalidUnionSchema();
-            Should.Throw<ExecutionError>(() => schema.FindType("a"));
+            //note: The exception occurs during CreateTypesLookup, not during FindType
```
sungam3r, line 86:
```suggestion
            //note: The exception occurs during Schema.CreateTypesLookup(), not during Schema.FindType()
```

PR #1781, opened 2020-08-14
CoreToVanillaConverter cleanup

Shane32:
As indicated in #1778 I went back to review the last `ExecutionError` within `CoreToVanillaConverter`.  I determined that it is impossible for an `ASTNodeKind.IntValue` to produce a value that cannot be cast to a `BigInteger`.  So I changed that exception to an `InvalidOperationException`, as if it was triggered, it would indicate a parsing problem within graphql-parser.

Then I reviewed `ASTNodeKind.FloatValue` and determined that for values beyond the range of a `decimal`, an overflow/parsing exception can occur.  This I fixed by using `decimal.TryParse`.  I also determined that values beyond the range of a `double` are simply interpreted as `Infinity` and `-Infinity` so there is no possibility of a failed conversion.  I added a test (which would have previously failed) to verify this change.

Finally, I noticed that `ValueConverter` was used to transliterate these strings to certain types of values.  I don't feel that's an appropriate use of the `ValueConverter` in this instance.  Note that the `ValueConverter` was not used for `ASTNodeKind.IntValue`.  So I changed the parsing code to use `double.Parse`, `decimal.TryParse` and `bool.Parse` in the appropriate locations.

Due to a well-designed parser, I was unable to trigger any other exceptions within this code.

Note that if we do want to use `ValueConverter`, it should be wrapped in a `try..catch` block and it should throw an `ExecutionError` upon a failed conversion.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1781.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1781.cs
@@ -1,0 +1,59 @@
+using System.Threading.Tasks;
+using GraphQL.Types;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1781
+    public class Bug1781 : QueryTestBase<Bug1781Schema>
+    {
+        [Fact]
+        public async Task DocumentExecuter_really_big_double_valid()
+        {
+            var de = new DocumentExecuter();
+            var valid = await de.ExecuteAsync(new ExecutionOptions
+            {
+                Query = $"{{ test(arg:{double.MaxValue.ToString("0")}0.0) }}",
+                Schema = Schema,
+            });
+            valid.ShouldNotBeNull();
+            valid.Data.ShouldNotBeNull();
+            valid.Errors.ShouldBeNull();
+        }
+
+        [Fact]
+        public async Task DocumentExecuter_really_small_double_valid()
+        {
+            var de = new DocumentExecuter();
+            var valid = await de.ExecuteAsync(new ExecutionOptions
+            {
+                Query = $"{{ test(arg:{double.MinValue.ToString("0")}0.0) }}",
+                Schema = Schema,
+            });
+            valid.ShouldNotBeNull();
+            valid.Data.ShouldNotBeNull();
+            valid.Errors.ShouldBeNull();
+        }
+    }
+
+    public class Bug1781Schema : Schema
+    {
+        public Bug1781Schema()
+        {
+            Query = new Bug1781Query();
+        }
+    }
+
+    public class Bug1781Query : ObjectGraphType
+    {
+        public Bug1781Query()
+        {
+            Field<StringGraphType>("Test",
+                resolve: context => "ok",
+                arguments: new QueryArguments(
+                    new QueryArgument(typeof(FloatGraphType)) { Name = "arg" }
+                ));
+        }
+    }
+}
--- a/src/GraphQL.Tests/Bugs/Bug1781.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1781.cs
@@ -1,5 +1,8 @@
+using System.Numerics;
 using System.Threading.Tasks;
+using GraphQL.Language;
 using GraphQL.Types;
+using GraphQLParser.AST;
 using Shouldly;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


Shane32:
Hmmm.... `double.Parse` doesn't work...

src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -203,52 +204,63 @@ public IValue Value(GraphQLValue source)
...
+                            return new DecimalValue(dec).WithLocation(str, _body);
+                    }
 
                     return new FloatValue(dbl).WithLocation(str, _body);
                 }
                 case ASTNodeKind.BooleanValue:
                 {
                     var str = (GraphQLScalarValue)source;
-                    return new BooleanValue(ValueConverter.ConvertTo<bool>(str.Value)).WithLocation(str, _body);
+                    return new BooleanValue(bool.Parse(str.Value)).WithLocation(str, _body);
```
Shane32, line 263: `bool.Parse` does allow for whitespace and is case insensitive, and has no culture specific overloads.  This should not matter; even simply `str.Value == "true"` would work correctly.

joemcbride:
@Shane32 I sent you an invite to the Organization.  Thank you for all your contributions!

src/GraphQL/Language/CoreToVanillaConverter.cs
sungam3r: You're right. Actually GraphQL parser allows only `true` and `false` literals for `ASTNodeKind.BooleanValue`.

src/GraphQL.Tests/Bugs/Bug1781.cs
```diff
@@ -0,0 +1,95 @@
...
+using GraphQL.Language;
+using GraphQL.Types;
+using GraphQLParser.AST;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1781
+    public class Bug1781 : QueryTestBase<Bug1781Schema>
```
sungam3r, line 12: This is the way to name the tests that correspond to the bugs that have been fixed. Actually there is no 1781 issue reporting some bug.

src/GraphQL.Tests/Bugs/Bug1781.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        [Fact]
+        public async Task DocumentExecuter_really_big_double_valid()
+        {
+            var de = new DocumentExecuter();
+            var valid = await de.ExecuteAsync(new ExecutionOptions
+            {
+                Query = $"{{ test(arg:{double.MaxValue.ToString("0")}0.0) }}",
```
sungam3r, line 20: What is the reason to use `0.0`?

sungam3r: Also consider to use `:0` instead of `.ToString("0")`

src/GraphQL.Tests/Bugs/Bug1781.cs
sungam3r: I propose to rename this file to `CoreToVanillaConverterTests.cs` and move into GraphQL.Tests root folder.

Shane32: 👍

src/GraphQL.Tests/Bugs/Bug1781.cs
Shane32: * `double.MaxValue.ToString("0")` would be intrepreted as an integer; this test is attempting to test floating point parsing
* This test is attempting to test floating point parsing where the result would be larger than `double.MaxValue`
Essentially adding `0.0` multiplies the value by 10 and then forces the parser to interpret it as a floating point value

Shane32: > Also consider to use `:0` instead of `.ToString("0")`

👍

src/GraphQL.Tests/Bugs/Bug1781.cs
Shane32: I will rely on your guidance for how/where to name tests.  So please comment along these lines as you see fit.

src/GraphQL.Tests/Bugs/Bug1781.cs
sungam3r: > Essentially adding 0.0 multiplies the value by 10 and then forces the parser to interpret it as a floating point value

It is worth to add comments about this.

src/GraphQL.Tests/CoreToVanillaConverterTests.cs
```diff
@@ -0,0 +1,94 @@
...
+    // https://github.com/graphql-dotnet/graphql-dotnet/pulls/1781
+    public class CoreToVanillaConverterTests : QueryTestBase<PR1781Schema>
+    {
+        [Fact]
+        public async Task DocumentExecuter_really_big_double_valid()
+        {
+            var de = new DocumentExecuter();
+            var valid = await de.ExecuteAsync(new ExecutionOptions
+            {
+                Query = $"{{ test(arg:{double.MaxValue:0}0.0) }}",
```
Shane32, line 19:
```suggestion
                // create a floating-point value that is larger than double.MaxValue
                // in the expression "{double.MaxValue:0}0.0" below, the 0.0 effectively
                //   multiplies double.MaxValue by 10 and the .0 forces the parser to
                //   assume it is a floating point value rather than a large integer
                Query = $"{{ test(arg:{double.MaxValue:0}0.0) }}",
```

src/GraphQL.Tests/Bugs/Bug1781.cs
Shane32: Done

src/GraphQL.Tests/CoreToVanillaConverterTests.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        [Fact]
+        public async Task DocumentExecuter_really_big_double_valid()
+        {
+            var de = new DocumentExecuter();
+            var valid = await de.ExecuteAsync(new ExecutionOptions
+            {
+                // create a floating-point value that is larger than double.MaxValue
+                // in the expression "{double.MaxValue:0}0.0" below, the 0.0 effectively
+                //   multiplies double.MaxValue by 10 and the .0 forces the parser to
```
sungam3r, line 21:
```suggestion
                // multiplies double.MaxValue by 10 and the .0 forces the parser to
```

src/GraphQL.Tests/CoreToVanillaConverterTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        [Fact]
+        public async Task DocumentExecuter_really_big_double_valid()
+        {
+            var de = new DocumentExecuter();
+            var valid = await de.ExecuteAsync(new ExecutionOptions
+            {
+                // create a floating-point value that is larger than double.MaxValue
+                // in the expression "{double.MaxValue:0}0.0" below, the 0.0 effectively
+                // multiplies double.MaxValue by 10 and the .0 forces the parser to
+                //   assume it is a floating point value rather than a large integer
```
sungam3r, line 22:
```suggestion
                // assume it is a floating point value rather than a large integer
```

PR #1785, opened 2020-08-18
Update ValidateNodeResult to throw InvalidOperationException

Shane32:
In #1773 somehow I changed the exceptions thrown during `ExecutionStrategy.SetArrayItemNodes` but not `ExecutionStrategy.ValidateNodeResult`.  These represent the same basic changes - invalid results from the resolver (e.g. null for a non-null field) throw an `InvalidOperationException` rather than an `ExecutionError`.

### DIFF
--- a/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
+++ b/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
@@ -243,6 +243,7 @@ namespace GraphQL.Tests.Bugs
                     else
                     {
                         actualError.InnerException.ShouldNotBeNull();
+                        actualError.InnerException.ShouldBeOfType(expectedError.InnerException.GetType());
                         actualError.InnerException.Message.ShouldBe(expectedError.InnerException.Message);
                     }
                 }
--- a/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
+++ b/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
@@ -38,7 +38,8 @@ namespace GraphQL.Tests.Bugs
             var data = new Data {NonNullable = null};
             var errors = new[]
             {
-                new ExecutionError("Cannot return null for non-null type. Field: nonNullable, Type: String!.")
+                new ExecutionError("Error trying to resolve nonNullable.", new InvalidOperationException(
+                    "Cannot return null for non-null type. Field: nonNullable, Type: String!."))
                 {
                     Path = new[] {"nullableDataGraph", "nonNullable"}
                 }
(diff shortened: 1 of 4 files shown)


src/GraphQL.Tests/Types/NonNullGraphTypeTests.cs
```diff
@@ -44,9 +44,12 @@ public void nonnullable_fields_without_values_do_complain()
                 expectedErrorCount: 3);
 
             var errors = result.Errors.ToArray();
-            errors[0].Message.ShouldBe("Cannot return null for non-null type. Field: a, Type: Int!.");
-            errors[1].Message.ShouldBe("Cannot return null for non-null type. Field: b, Type: Boolean!.");
-            errors[2].Message.ShouldBe("Cannot return null for non-null type. Field: c, Type: String!.");
+            errors[0].Message.ShouldBe("Error trying to resolve a.");
```
sungam3r, line 47: Maybe it is worth to change base error message from `$"Error trying to resolve {node.Name}."` to `$"Error trying to resolve field '{node.Name}'."` in `SubscriptionExecutionStrategy` and `ExecutionStrategy` ?

src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
```diff
@@ -106,7 +110,8 @@ public void ListOfNonNull_containing_null_should_bubble_up_the_null()
             var data = new Data {ListOfStrings = new List<string> {"text", null, null}};
             var errors = new[]
             {
-                new ExecutionError("Error trying to resolve listOfNonNullable.")
+                new ExecutionError("Error trying to resolve listOfNonNullable.", new InvalidOperationException(
+                    "Cannot return a null member within a non-null list for list index 1."))
```
sungam3r, line 114: There was no this message here before, but tests passed. This means that the test only checks the type of the inner exception or, in other words, the error code (INVALID_OPERATION, FORMAT, etc.). Let's add message validation for inner exceptions.

sungam3r: 😄 I read to the end of the file and saw that you have already done so. I would also add a check for matching exception types.

src/GraphQL.Tests/Types/NonNullGraphTypeTests.cs
Shane32: I agree.  It will probably change a million tests, so I'll put in a separate PR.

src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
Shane32: No problem 👍 

Done.

src/GraphQL.Tests/Types/NonNullGraphTypeTests.cs
sungam3r: OK

PR #1786, opened 2020-08-18
Update exception types

Shane32:
* Add abstract `DocumentError`, which inherits `ExecutionError` and has these implementations:
  * `NoOperationError` (code `NO_OPERATION`)
  * `InvalidOperationError` (code `INVALID_OPERATION`)
  * `SyntaxError` (code `SYNTAX_ERROR`)
  * `InvalidVariableError` (renamed from `InvalidValueException`; unchanged code `INVALID_VALUE`)
  * `ValidationError` (code depends on which validation rule failed)
* Add `UnhandledError`, which inherits from `ExecutionError`, which is created when an unhandled error occurs.  No additional code is set for this error type; only the inner exception's codes are returned.

By convention, errors that inherit from `ExecutionError` now all end with the word `Error`, and regular exceptions end with `Exception`.

See #1768 

I also changed this error message: "Cannot execute query if no operation is specified." to "Document does not contain any operations.", which applies when there is an empty document (for the `NoOperationError`)

### DIFF
--- a/src/GraphQL/Execution/DocumentError.cs
+++ b/src/GraphQL/Execution/DocumentError.cs
@@ -1,0 +1,10 @@
+using System;
+
+namespace GraphQL.Execution
+{
+    public abstract class DocumentError : ExecutionError
+    {
+        public DocumentError(string message) : base(message) { }
+        public DocumentError(string message, Exception innerException) : base(message, innerException) { }
+    }
+}
--- a/src/GraphQL/Execution/DocumentError.cs
+++ b/src/GraphQL/Execution/DocumentError.cs
@@ -5,6 +5,7 @@ namespace GraphQL.Execution
     public abstract class DocumentError : ExecutionError
     {
         public DocumentError(string message) : base(message) { }
+        
         public DocumentError(string message, Exception innerException) : base(message, innerException) { }
     }
 }
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -25,7 +25,7 @@ namespace GraphQL.Execution
                     type = schema.Mutation;
                     if (type == null)
                     {
-                        error = new ExecutionError("Schema is not configured for mutations");
+                        error = new InvalidOperationError("Schema is not configured for mutations");
                         error.AddLocation(operation, document);
                         throw error;
                     }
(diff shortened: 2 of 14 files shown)


Shane32:
@sungam3r If this looks good in concept, you might want to consider which namespace/folder to add these exceptions into.

src/GraphQL/Execution/SyntaxError.cs
```diff
@@ -0,0 +1,24 @@
...
+            AddLocation(ex.Line, ex.Column);
+        }
+
+        // available for use with third-party parsing engines
+        public SyntaxError(string message, Exception innerException)
+            : base("Error parsing query: " + message, innerException)
+        {
+            // the inner exception is of an unknown type, so set the code
+            Code = "SYNTAX_ERROR";
+        }
```
sungam3r, line 22: Maybe we won't add API ahead of time that no one needs yet?

src/GraphQL/Execution/InvalidOperationError.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
```
sungam3r, line 3:
```suggestion
```

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -35,16 +35,14 @@ public static IObjectGraphType GetOperationRootType(Document document, ISchema s
...
                         error.AddLocation(operation, document);
                         throw error;
                     }
                     break;
 
                 default:
-                    error = new ExecutionError("Can only execute queries, mutations and subscriptions.");
-                    error.AddLocation(operation, document);
-                    throw error;
+                    throw new ArgumentOutOfRangeException(nameof(operation), "Can only execute queries, mutations and subscriptions.");
```
sungam3r, line 45: That is, we consider this situation extremely non-standard and now we throw an exception instead of `ExecutionError` with filled location?

src/GraphQL/Execution/DocumentError.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace GraphQL.Execution
+{
+    public abstract class DocumentError : ExecutionError
+    {
+        public DocumentError(string message) : base(message) { }
+        public DocumentError(string message, Exception innerException) : base(message, innerException) { }
```
sungam3r, line 8:
```suggestion
        public DocumentError(string message) : base(message) { }
        
        public DocumentError(string message, Exception innerException) : base(message, innerException) { }
```

sungam3r:
> you might want to consider which namespace/folder to add these exceptions into.

`GraphQL.Execution` is OK, `ValidationError` may stay in `GraphQL.Validation` namespace.

src/GraphQL/Execution/SyntaxError.cs
Shane32: Sounds fine to me

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: The enum only contains the three members, and this is fully checked by the parser. If somehow this was called with another enum then it's an invalid argument to the function.

Shane32:
I'll make the suggested changes when I'm at a pc later.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: OK

src/GraphQL/Execution/SyntaxError.cs
Shane32:
```suggestion
```

sungam3r (approved):

PR #1787, opened 2020-08-18
DocumentExecuter does not properly hide unhandled exception error messages

Shane32:
The `DocumentExecuter` does not properly hide unhandled exception error messages, as unhandled errors that occur within a field resolver do.  This fixes that.  Inner exception codes are still returned as always.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1772.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1772.cs
@@ -45,9 +45,10 @@ namespace GraphQL.Tests.Bugs
             result.Data.ShouldBeNull();
             result.Errors.ShouldNotBeNull();
             result.Errors.Count.ShouldBe(1);
-            result.Errors[0].Message.ShouldBe($"Query does not contain operation '{operationName}'.");
+            result.Errors[0].Message.ShouldBe("Error executing document.");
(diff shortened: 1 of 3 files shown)


src/GraphQL.Tests/Bugs/Bug1772.cs
```diff
@@ -45,9 +45,10 @@ public async Task DocumentExecuter_throws_for_invalid_operation(string operation
             result.Data.ShouldBeNull();
             result.Errors.ShouldNotBeNull();
             result.Errors.Count.ShouldBe(1);
-            result.Errors[0].Message.ShouldBe($"Query does not contain operation '{operationName}'.");
+            result.Errors[0].Message.ShouldBe("Error executing document.");
             result.Errors[0].InnerException.ShouldNotBeNull();
             result.Errors[0].InnerException.ShouldBeOfType<InvalidOperationException>();
+            result.Errors[0].InnerException.Message.ShouldBe($"Query does not contain operation '{operationName}'.");
```
Shane32, line 51: Note that this particular test produces an `InvalidOperationException` because an invalid operation name was passed to `ExecuteAsync` within `ExecutionOptions`.  It is not invalid data within the query or variables; it is an incorrect parameter passed to the execution engine.

PR #1791, opened 2020-08-18
Change default field resolver message

Shane32:
* Changes `Error trying to resolve {node.Name}.` to `Error trying to resolve field '{node.Name}'.`, and same for subscriptions
* Changes `Subscriber not set for field {node.Field.Name}` to `Subscriber not set for field '{node.Field.Name}'.`
* Changes `Could not subscribe to field '{node.Field.Name}' in query '{context.Document.OriginalQuery}'` to include period at end

### DIFF
--- a/src/GraphQL.Tests/Errors/ErrorLocationTests.cs
+++ b/src/GraphQL.Tests/Errors/ErrorLocationTests.cs
@@ -80,7 +80,7 @@ namespace GraphQL.Tests.Errors
         [Fact]
         public void async_field_with_errors()
         {
-            var error = new ExecutionError("Error trying to resolve testasync.");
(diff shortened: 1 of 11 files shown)


src/GraphQL.Tests/Errors/ErrorLocationTests.cs
```diff
@@ -80,7 +80,7 @@ public async Task should_include_path_with_list_index()
         [Fact]
         public void async_field_with_errors()
         {
-            var error = new ExecutionError("Error trying to resolve testasync.");
+            var error = new ExecutionError("Error trying to resolve field 'testasync'.");
```
sungam3r, line 83: so it became much clearer

PR #1793, opened 2020-08-18
Add checks for Schema.RegisterXXX methods

### LINKED ISSUE #1715: schema.FindType("ID") causes other types to disappear from displayed schema
## Description

I am using GraphQL.Net 3.0.0-preview-1648. I am attempting to programmatically generate a schema, but get weird side effects related to looking up the `ID` type for use.

### Steps to reproduce

1. Create a new project: `dotnet new console -o App`
2. `cd App`
3. Add GraphQL dependency: `dotnet add package GraphQL -v 3.0.0-preview-1648`
4. Modify Program.cs to:

```
using System;

namespace App
{
    class Program
    {
        {
            var schema = new GraphQL.Types.Schema();
            var query = new GraphQL.Types.ObjectGraphType { Name = "Query" };
            schema.RegisterType(query);
            schema.Query = query;

            Console.WriteLine($"id = {schema.FindType("ID")}");

            var jedi = new GraphQL.Types.ObjectGraphType { Name = "Jedi" };
            schema.RegisterType(jedi);

            var name = new GraphQL.Types.FieldType { Name = "name", ResolvedType = schema.FindType("String") };
            jedi.AddField(name);

            var side = new GraphQL.Types.FieldType { Name = "side", ResolvedType = schema.FindType("String") };
            jedi.AddField(side);

            var jediById = new GraphQL.Types.FieldType { Name = "jedi", ResolvedType = jedi };
            var id = new GraphQL.Types.QueryArgument(schema.FindType("ID")) { Name = "id" };
            jediById.Arguments = new GraphQL.Types.QueryArguments(id);
            query.AddField(jediById);

            var options = new GraphQL.Utilities.SchemaPrinterOptions {
                IncludeDescriptions = true,
                IncludeDeprecationReasons = true,
            };
            var printer = new GraphQL.Utilities.SchemaPrinter(schema, options);
            Console.WriteLine(printer.Print());
        }
    }
}
```

5. Run the program: `dotnet run`
6. Notice in the output no type `Jedi` is defined, though `Query` references it:

```
type Query {
  jedi(id: ID): Jedi
}
```

7. Comment line 14: `// Console.WriteLine($"id = {schema.FindType("ID")}");`
8. Rerun the program: `dotnet run`
9. Notice in the output that a type `Jedi` is defined:

```
type Jedi {
  name: String
  side: String
}
```

In my other tests, it appears that the schema is defined properly, in the sense that I can query `jedi(id: ID)` and get back a `Jedi` even though it doesn't show up in the output of `SchemaPrinter`. 

### Expected result

I expect that calling `schema.FindType("ID")` wouldn't change the schema.

### Actual result

But it appears to or at least to change what the schema displays as.

### Environment

Windows 10.
`dotnet --version` 3.1.301

sungam3r:
fixes #1715

### DIFF
--- a/src/GraphQL/Types/Schema.cs
+++ b/src/GraphQL/Types/Schema.cs
@@ -67,8 +67,7 @@ namespace GraphQL.Types
         private void CheckInitialized()
         {
             if (Initialized)
-                throw new InvalidOperationException(@"Schema is already initialized and sealed for modifications.
-There was an attempt to modify schema after it was initialized. You should call RegisterXXX methods only when Schema.Initialized = false.");
+                throw new InvalidOperationException("Schema is already initialized and sealed for modifications. You should call RegisterXXX methods only when Schema.Initialized = false.");
         }
 
         public void Initialize()
--- a/src/GraphQL/Types/Schema.cs
+++ b/src/GraphQL/Types/Schema.cs
@@ -63,6 +63,14 @@ namespace GraphQL.Types
 
         public bool Initialized => _lookup?.IsValueCreated == true;
 
+        // TODO: It would be worthwhile to think at all about how to redo the design so that such a situation does not arise.
+        private void CheckInitialized()
+        {
+            if (Initialized)
+                throw new InvalidOperationException(@"Schema is already initialized and sealed for modifications.
+There was an attempt to modify schema after it was initialized. You should call RegisterXXX methods only when Schema.Initialized = false.");
+        }
+
         public void Initialize()
         {
             CheckDisposed();
@@ -107,6 +115,7 @@ namespace GraphQL.Types
             set
             {
                 CheckDisposed();
+                CheckInitialized();
 
                 _directives.Clear();
 
@@ -132,6 +141,7 @@ namespace GraphQL.Types
         public void RegisterType(IGraphType type)
         {
             CheckDisposed();
+            CheckInitialized();
 
             _additionalInstances.Add(type ?? throw new ArgumentNullException(nameof(type)));
         }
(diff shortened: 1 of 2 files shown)


src/GraphQL/Types/Schema.cs
```diff
@@ -63,6 +63,14 @@ public static ISchema For(string typeDefinitions, Action<SchemaBuilder> configur
 
         public bool Initialized => _lookup?.IsValueCreated == true;
 
+        // TODO: It would be worthwhile to think at all about how to redo the design so that such a situation does not arise.
```
sungam3r, line 66: I'm sure that it will be a breaking change and I'm not ready to do it yet.

src/GraphQL/Types/Schema.cs
```diff
@@ -63,6 +63,14 @@ public static ISchema For(string typeDefinitions, Action<SchemaBuilder> configur
...
+        private void CheckInitialized()
+        {
+            if (Initialized)
+                throw new InvalidOperationException(@"Schema is already initialized and sealed for modifications.
+There was an attempt to modify schema after it was initialized. You should call RegisterXXX methods only when Schema.Initialized = false.");
```
Shane32, line 71: I suggest we do not include line breaks within messages.

```suggestion
                throw new InvalidOperationException("Schema is already initialized and sealed for modifications. You should call RegisterXXX methods only when Schema.Initialized = false.");
```

Shane32: The message could be simplified even further, but whatever you think.

```suggestion
                throw new InvalidOperationException("Schema is already initialized and sealed for modifications.");
```

sungam3r: If possible, I always write in the error message not only that something bad happened, but also how to prevent it. It's more friendly.

PR #1794, opened 2020-08-19
remove Operation from ApolloTrace

SimonCropp:
apollo trace doesnt use operation. https://github.com/apollographql/apollo-tracing  so remove a redundant parameter


sungam3r (approved):
I didn't even notice that the parameter is not used.

PR #1796, opened 2020-08-20
Cleanup some misc ExecutionErrors

Shane32:
I will note each change with a comment

### DIFF
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -48,7 +48,7 @@ namespace GraphQL.Execution
             return type;
         }
 
-        public static FieldType GetFieldDefinition(Document document, ISchema schema, IObjectGraphType parentType, Field field)
+        public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType parentType, Field field)
         {
             if (field.Name == schema.SchemaMetaFieldType.Name && schema.Query == parentType)
             {
@@ -65,9 +65,7 @@ namespace GraphQL.Execution
 
             if (parentType == null)
             {
-                var error = new ExecutionError($"Schema is not configured correctly to fetch {field.Name}.  Are you missing a root type?");
-                error.AddLocation(field, document);
-                throw error;
+                throw new ArgumentNullException(nameof(parentType), $"Schema is not configured correctly to fetch {field.Name}.  Are you missing a root type?");
             }
 
             return parentType.GetField(field.Name);
--- a/src/GraphQL/Introspection/__Type.cs
+++ b/src/GraphQL/Introspection/__Type.cs
@@ -25,7 +25,7 @@ namespace GraphQL.Introspection
                     return KindForInstance(type);
                 }
 
-                throw new ExecutionError($"Unknown kind of type: {context.Source}");
+                throw new InvalidOperationException($"Unknown kind of type: {context.Source}");
             });
             Field<StringGraphType>("name", resolve: context => ((IGraphType)context.Source).Name);
             Field<StringGraphType>("description");
@@ -122,7 +122,7 @@ namespace GraphQL.Introspection
             IInputObjectGraphType _ => TypeKindBoxed.INPUT_OBJECT,
             ListGraphType _ => TypeKindBoxed.LIST,
             NonNullGraphType _ => TypeKindBoxed.NON_NULL,
-            _ => throw new ExecutionError("Unknown kind of type: {0}".ToFormat(type))
+            _ => throw new ArgumentOutOfRangeException(nameof(type), $"Unknown kind of type: {type}")
         };
     }
 }
--- a/src/GraphQL/ObjectExtensions.cs
+++ b/src/GraphQL/ObjectExtensions.cs
@@ -247,7 +247,7 @@ namespace GraphQL
 
                 if (!IsDefinedEnumValue(fieldType, value))
                 {
-                    throw new ExecutionError($"Unknown value '{value}' for enum '{fieldType.Name}'.");
+                    throw new InvalidOperationException($"Unknown value '{value}' for enum '{fieldType.Name}'.");
                 }
 
                 string str = value.ToString();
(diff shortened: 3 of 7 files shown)


src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -65,9 +65,7 @@ public static FieldType GetFieldDefinition(Document document, ISchema schema, IO
 
             if (parentType == null)
             {
-                var error = new ExecutionError($"Schema is not configured correctly to fetch {field.Name}.  Are you missing a root type?");
-                error.AddLocation(field, document);
-                throw error;
+                throw new ArgumentNullException(nameof(parentType), $"Schema is not configured correctly to fetch {field.Name}.  Are you missing a root type?");
```
Shane32, line 68: `parentType` should never be null unless you're querying for a introspection type.  After this change, `document` is not needed and was removed from the method signature

src/GraphQL/Introspection/__Type.cs
```diff
@@ -25,7 +25,7 @@ public __Type()
                     return KindForInstance(type);
                 }
 
-                throw new ExecutionError($"Unknown kind of type: {context.Source}");
+                throw new InvalidOperationException($"Unknown kind of type: {context.Source}");
```
Shane32, line 28: `Source` will always be of the correct type -- or else it's an internal error

src/GraphQL/Introspection/__Type.cs
```diff
@@ -122,7 +122,7 @@ public __Type()
             IInputObjectGraphType _ => TypeKindBoxed.INPUT_OBJECT,
             ListGraphType _ => TypeKindBoxed.LIST,
             NonNullGraphType _ => TypeKindBoxed.NON_NULL,
-            _ => throw new ExecutionError("Unknown kind of type: {0}".ToFormat(type))
+            _ => throw new ArgumentOutOfRangeException(nameof(type), $"Unknown kind of type: {type}")
```
Shane32, line 125: There shouldn't be any other kind of graph types

src/GraphQL/ObjectExtensions.cs
```diff
@@ -247,7 +247,7 @@ public static object GetPropertyValue(this object propertyValue, Type fieldType,
 
                 if (!IsDefinedEnumValue(fieldType, value))
                 {
-                    throw new ExecutionError($"Unknown value '{value}' for enum '{fieldType.Name}'.");
+                    throw new InvalidOperationException($"Unknown value '{value}' for enum '{fieldType.Name}'.");
```
Shane32, line 250: This function only executes when GetArgument is called, and even then, for properly validated schema it shouldn't fail this check.  Plus, a bug within `IsDefinedEnumValue` prevented this line of code from ever running.  I did fix the bug, however, in this PR.

src/GraphQL/ObjectExtensions.cs
```diff
@@ -296,25 +296,31 @@ public static T GetPropertyValue<T>(this object value)
...
-            var converted = Convert.ChangeType(value, underlyingType);
+                var underlyingType = Enum.GetUnderlyingType(type);
+                var converted = Convert.ChangeType(value, underlyingType);
 
-            var values = Enum.GetValues(type);
+                var values = Enum.GetValues(type);
 
-            foreach (var val in values)
-            {
-                var convertedVal = Convert.ChangeType(val, underlyingType);
```
Shane32, line 312: This line of code, which is questionable if it belongs here or not, throws an exception when called with an invalid enum string.

src/GraphQL/ObjectExtensions.cs
```diff
@@ -296,25 +296,31 @@ public static T GetPropertyValue<T>(this object value)
...
+                    var convertedVal = Convert.ChangeType(val, underlyingType);
+                    if (convertedVal.Equals(converted))
+                    {
+                        return true;
+                    }
                 }
             }
+            catch
+            {
+            }
```
Shane32, line 323: I added a try..catch block so that this function returns false when an exception is thrown.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -441,7 +441,7 @@ private DirectiveLocation ToDirectiveLocation(string name)
                 return (DirectiveLocation)result;
             }
 
-            throw new ExecutionError($"{name} is an unknown directive location");
+            throw new ArgumentOutOfRangeException(nameof(name), $"{name} is an unknown directive location");
```
Shane32, line 444: ExecutionErrors should not be thrown while the schema is being built

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -584,7 +584,7 @@ public static object ToValue(this GraphQLValue source)
                         return bigIntegerResult;
                     }
 
-                    throw new ExecutionError($"Invalid number {str.Value}");
+                    throw new InvalidOperationException($"Invalid number {str.Value}");
```
Shane32, line 587: ExecutionErrors should not be thrown while the schema is being built

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -626,7 +626,7 @@ public static object ToValue(this GraphQLValue source)
                     return values;
                 }
                 default:
-                    throw new ExecutionError($"Unsupported value type {source.Kind}");
+                    throw new InvalidOperationException($"Unsupported value type {source.Kind}");
```
Shane32, line 629: ExecutionErrors should not be thrown while the schema is being built

src/GraphQL/Utilities/SchemaBuilder.cs
Shane32: Note that this entire `ToValue` function is a duplicate, more or less, of similar code within the CoreToAstConverter.  It should probably be deduplicated, but I have not investigated it enough to know for sure.

src/GraphQL/Validation/Rules/KnownDirectives.cs
```diff
@@ -74,7 +75,7 @@ private DirectiveLocation getDirectiveLocationForAstPath(INode[] ancestors, Vali
             if (appliedTo is InlineFragment) return DirectiveLocation.InlineFragment;
             if (appliedTo is FragmentDefinition) return DirectiveLocation.FragmentDefinition;
 
-            throw new ExecutionError($"Unable to determine directive location for \"{context.Print(appliedTo)}\".");
+            throw new InvalidOperationException($"Unable to determine directive location for \"{context.Print(appliedTo)}\".");
```
Shane32, line 78: An error here would indicate an invalid schema - correct??

src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
```diff
@@ -166,7 +166,7 @@ await listener.AfterExecutionAsync(context)
             IEnumerable<object> path,
             Exception ex = null)
         {
-            var error = new ExecutionError(message, ex);
+            var error = new UnhandledError(message, ex);
```
Shane32, line 169: Unhandled exceptions within `SubscriptionExecutionStrategy` should throw an `UnhandledException`

Shane32:
Once this PR and #1797 are merged, there will only be one remaining call to `new ExecutionError`:

https://github.com/graphql-dotnet/graphql-dotnet/blob/c89cb5d71ec768fbf975c96ea2c0bb586543f17d/src/GraphQL/GraphQLExtensions.cs#L467

I'm rather stumped by that entire method.  For example, it calls `inputType.Serialize` on a variable value rather than `ParseValue`, which seems to make no sense.  Maybe ParseValue should return an `IValue`; I don't know.  So I left it alone.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r, line 68:
```suggestion
                throw new ArgumentNullException(nameof(parentType), $"Schema is not configured correctly to fetch {field.Name}. Are you missing a root type?");
```

PR #1797, opened 2020-08-20
Remove ResolveFieldContextExtensions.TryAsyncResolve

Shane32:
I propose to delete `ResolveFieldContextExtensions.TryAsyncResolve` from the main GraphQL library.  It is only used by `ResolveFieldContextTests.cs`, and seems a bit useless otherwise.  Also, the logic is designed for tests: if an error occurs and an exception handler was specified, it will completely ignore the exception (not adding it to the list of document errors), and call the exception handler with the list of errors from the context, returning a `TResult` from the exception handler.  This logic is obviously useful only to a test.


Shane32:
Also, the only tests using `TryAsyncResolve` were only there to test the functionality of `TryAsyncResolve`!  So I removed those tests, at which point nothing referenced `TryAsyncResolve`, so I removed the function completely.

sungam3r:
Probably it was left from some refactoring.

sungam3r (approved):

PR #1799, opened 2020-08-20
Support UnhandledExceptionContext.ErrorMessage within DocumentExecuter

Shane32:
When setting `UnhandledExceptionContext.ErrorMessage` during an error caught in the `DocumentExecuter`, the error message is now overridden as it should be.

Added tests to verify this both for `DocumentExecuter` and `ExecutionStrategy`.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -199,6 +199,9 @@ namespace GraphQL
             }
             catch (Exception ex)
             {
+                if (options.ThrowOnUnhandledException)
+                    throw;
+
                 UnhandledExceptionContext exceptionContext = null;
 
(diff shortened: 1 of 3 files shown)


src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -202,9 +202,11 @@ await listener.AfterExecutionAsync(context)
                 if (options.ThrowOnUnhandledException)
                     throw;
 
+                UnhandledExceptionContext exceptionContext = null;
+
```
sungam3r, line 206:
```suggestion
```

To align with code in `ExecutionStrategy.cs`

sungam3r:
Yes, I really forgot about `ErrorMessage` for `DocumentExecuter`.

sungam3r (approved):

PR #1811, opened 2020-08-25
Change ExposeExceptionStackTrace to only expose stack traces of UnhandledErrors

Shane32:
Regression fix: Originally the `ExposeExceptions` only exposed stack traces of unhandled exceptions.  Now it exposes stack traces of all errors, such as validation errors.  This PR fixes that.

Should the property name be changed to `ExposeUnhandledExceptionStackTrace`?  It seems rather long.

### DIFF
--- a/src/GraphQL.Tests/Errors/ErrorInfoProviderTests.cs
+++ b/src/GraphQL.Tests/Errors/ErrorInfoProviderTests.cs
@@ -131,7 +131,7 @@ namespace GraphQL.Tests.Errors
         public void exposeExceptions()
         {
             var innerException = new ArgumentNullException(null, new ArgumentOutOfRangeException());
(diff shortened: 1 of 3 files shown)


sungam3r:
> Originally the ExposeExceptions only exposed stack traces of unhandled exceptions. 

Really? It seemed to me like it worked for all exceptions.

Shane32:
You're right!  I should have checked before posting this.

sungam3r:
So how was it really?

Shane32:
Like you said, it applied to all `ExecutionError` instances, regardless of whether it was caused by an unhandled exception or not.

https://github.com/graphql-dotnet/graphql-dotnet/blob/d982d20d52b6874c0b01b8da76e638ebb057013c/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs#L145

PR #1814, opened 2020-08-26
Validation error code rework

Shane32:
This is a implementation of the proposal here:

https://github.com/graphql-dotnet/graphql-dotnet/issues/1768#issuecomment-680220818

~I will of course update the rest of the exceptions, add tests, and so on.  Please confirm if this design looks good.~ Done


Shane32:
With this change, the `ValidationError` class will simply be the base type for all other validation error classes, such as in this case `ArgumentsOfCorrectTypeError`.  The `Code` property will be auto named to match the class name, excluding the "Error" suffix.  In this example it will be `ARGUMENTS_OF_CORRECT_TYPE`.  Error classes will match the validation rule class name (which in this case is `ArgumentsOfCorrectType`).  The new `Number` property was added to `ValidationError` and will be set only in the constructor of the error class, so it is easy to find and update the error codes when they need to be revised.  The constructors of each error class can be tailored for the needs of that validation rule, if applicable.  Finally, the new `ErrorInfoProvider` will include the `number` for any validation error where it would also include the `code`.

Shane32:
@sungam3r I'll probably start finishing this PR now with all the validation rules and tests, so if you think I'm going in the wrong direction, please let me know asap so I'm not wasting time.

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -36,7 +36,10 @@ public ValidationError(string originalQuery, string errorCode, string message, p
             params INode[] nodes)
             : base(message, innerException)
         {
-            Code = errorCode;
+            Code = ErrorInfoProvider.GetErrorCode(GetType());
+            if (Code != "VALIDATION_ERROR" && Code.EndsWith("_ERROR"))
+                Code = Code.Substring(0, Code.Length - 6);
+            Number = errorCode;
```
sungam3r, line 42: Please change `errorCode` argument to `number`.

Shane32: Of course

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -53,5 +56,10 @@ public ValidationError(string originalQuery, string errorCode, string message, p
         /// Returns a list of AST nodes that this error applies to.
         /// </summary>
         public IEnumerable<INode> Nodes => _nodes;
+
+        /// <summary>
+        /// Gets or sets the rule number of this validation error.
```
sungam3r, line 61:
```suggestion
        /// Gets or sets the rule number of this validation error corresponding to the paragraph number from the official specification.
```

sungam3r:
I'm OK with this PR.

Shane32:
@sungam3r I added a const to each class for the number of the code - called `PARAGRAPH` - primarily used for tests, so they do not fail based on updated code numbers.  I was thinking this should perhaps be an `internal` const.  I could rename it to `NUMBER` as C# is case sensitive.  Let me know what you think.

Also was thinking to perhaps write a set of tests that simply checks all the numbers to see if they changed.

Shane32:
All of the message generation functions are `internal` as they are only used for tests.

Shane32:
All of the numbers now match the codes shown on #1768 .

sungam3r:
I'm OK with internal const.

>  I could rename it to NUMBER

OK.

sungam3r:
Will review tomorrow.

Shane32:
I added the list of codes to the error documentation as an appendix.  Now all codes generated by GraphQL.NET should be listed in the documentation on that one page.

docs2/site/docs/getting-started/errors.md
```diff
@@ -65,7 +65,7 @@ Empty query document | NoOperationError | NO_OPERATION
 Query parsing error | SyntaxError | SYNTAX_ERROR
 Attempting a mutation or subscription when none are defined | InvalidOperationError | INVALID_OPERATION
 Invalid variable values | InvalidVariableError | INVALID_VALUE
-Validation errors | ValidationError | (varies)
+Validation errors | dervied from ValidationError | [(varies; see list below)](#ValidationErrors)
```
sungam3r, line 68:
```suggestion
Validation errors | derived from ValidationError | [(varies; see list below)](#ValidationErrors)
```

docs2/site/docs/getting-started/errors.md
```diff
@@ -200,3 +200,36 @@ For example, to show the stack traces for unhandled errors during development, y
 You can also write your own implementation of `IErrorInfoProvider`. For instance, you might want to override
 the numerical codes provided by GraphQL.NET for validation errors, or reveal stack traces
 only to logged-in administrators.
+
+## <a name="ValidationErrors"></a>Validation error reference list
+
+Here is a full list of validation errors produced by GraphQL.NET:
```
sungam3r, line 206: Let's order this list by error number, not by rule name/code.

src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
```diff
@@ -32,9 +34,9 @@ public void Very_Long_Number_In_Input_Should_Return_Error_For_Int()
             {
                 Errors = new ExecutionErrors
                 {
-                    new ExecutionError("Argument \"in\" has invalid value 636474637870330463.\nExpected type \"Int\", found 636474637870330463.")
+                    new ValidationError(null, ArgumentsOfCorrectTypeError.NUMBER, "Argument \"in\" has invalid value 636474637870330463.\nExpected type \"Int\", found 636474637870330463.")
```
sungam3r, line 37: Why don't you use `new ArgumentsOfCorrectTypeError()`?

src/GraphQL.Tests/Validation/DefaultValuesOfCorrectTypeTests.cs
```diff
@@ -97,8 +98,14 @@ public void list_variables_with_invalid_item()
                     query InvalidItem($a: [String] = [""one"", 2]) {
                       dog { name }
                     }";
-                _.Error(Rule.BadValueForDefaultArgMessage("a", "[String]", "[\"one\", 2]", new[] { "In element #1: Expected type \"String\", found 2." }), 2, 54);
+                _.Error(BadValueForDefaultArgMessage("a", "[String]", "[\"one\", 2]", new[] { "In element #1: Expected type \"String\", found 2." }), 2, 54);
             });
         }
+
+        private string BadValueForDefaultArgMessage(string varName, string type, string value, string[] verboseErrors)
```
sungam3r, line 105:
```suggestion
        private static string BadValueForDefaultArgMessage(string varName, string type, string value, string[] verboseErrors)
```

src/GraphQL.Tests/Validation/DefaultValuesOfCorrectTypeTests.cs
```diff
@@ -97,8 +98,14 @@ public void list_variables_with_invalid_item()
...
+            => DefaultValuesOfCorrectTypeError.BadValueForDefaultArgMessage(varName, type, value, verboseErrors);
+
+        private string BadValueForNonNullArgMessage(string varName, string type, string guessType)
```
sungam3r, line 108:
```suggestion
        private static string BadValueForNonNullArgMessage(string varName, string type, string guessType)
```

src/GraphQL.Tests/Validation/NoFragmentCyclesTests.cs
```diff
@@ -259,23 +260,26 @@ public void no_spreading_itself_deeply_and_immediately()
...
                 {
-                    e.Message = Rule.CycleErrorMessage("fragB", new[] { "fragC" });
+                    e.Message = CycleErrorMessage("fragB", new[] { "fragC" });
                     e.Loc(3, 53);
                     e.Loc(4, 53);
                 });
             });
         }
+
+        private string CycleErrorMessage(string fragName, string[] spreadNames)
```
sungam3r, line 282:
```suggestion
        private static string CycleErrorMessage(string fragName, string[] spreadNames)
```

src/GraphQL.Tests/Validation/ScalarLeafTests.cs
```diff
@@ -153,10 +154,16 @@ public void scalar_selection_not_allowed_with_directives_and_args()
...
                 _.Query = query;
                 _.Error(
-                    message: Rule.NoSubselectionAllowedMessage("doesKnowCommand", "Boolean"),
+                    message: NoSubselectionAllowedMessage("doesKnowCommand", "Boolean"),
                     line: 3,
                     column: 71);
             });
         }
+
+        private string NoSubselectionAllowedMessage(string field, string type)
```
sungam3r, line 163:
```suggestion
        private static string NoSubselectionAllowedMessage(string field, string type)
```

src/GraphQL.Tests/Validation/ScalarLeafTests.cs
```diff
@@ -153,10 +154,16 @@ public void scalar_selection_not_allowed_with_directives_and_args()
...
+            => ScalarLeafsError.NoSubselectionAllowedMessage(field, type);
+
+        private string RequiredSubselectionMessage(string field, string type)
```
sungam3r, line 166:
```suggestion
        private static string RequiredSubselectionMessage(string field, string type)
```

src/GraphQL.Tests/Validation/VariablesInAllowedPositionTests.cs
```diff
@@ -350,11 +351,14 @@ public void string_to_nonnull_boolean_in_directive()
...
                 {
-                    err.Message = Rule.BadVarPosMessage("stringVar", "String", "Boolean!");
+                    err.Message = BadVarPosMessage("stringVar", "String", "Boolean!");
                     err.Loc(2, 27);
                     err.Loc(3, 34);
                 });
             });
         }
+
+        private string BadVarPosMessage(string varName, string varType, string expectedType)
```
sungam3r, line 361:
```suggestion
        private static string BadVarPosMessage(string varName, string varType, string expectedType)
```

src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
Shane32: I'd have to make additional constructors that don't require `ValidationContext`, etc

docs2/site/docs/getting-started/errors.md
Shane32: Done

src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
sungam3r: OK

docs2/site/docs/getting-started/errors.md
```diff
@@ -200,3 +200,36 @@ For example, to show the stack traces for unhandled errors during development, y
...
+OverlappingFieldsCanBeMerged | OVERLAPPING_FIELDS_CAN_BE_MERGED | 5.3.2
+ScalarLeafs                  | SCALAR_LEAFS                     | 5.3.3
+KnownArgumentNames           | KNOWN_ARGUMENT_NAMES             | 5.4.1
+UniqueArgumentNames          | UNIQUE_ARGUMENT_NAMES            | 5.4.2
+ProvidedNonNullArguments     | PROVIDED_NON_NULL_ARGUMENTS      | 5.4.2.1
+UniqueFragmentNames          | UNIQUE_FRAGMENT_NAMES            | 5.5.1.1
+KnownTypeNames               | KNOWN_TYPE_NAMES                 | 5.5.1.2
+FragmentsOnCompositeTypes    | FRAGMENTS_ON_COMPOSITE_TYPES     | 5.5.1.3
+KnownFragmentNames           | KNOWN_FRAGMENT_NAMES             | 5.5.1.4
+NoUnusedFragments            | NO_UNUSED_FRAGMENTS              | 5.5.1.4
```
sungam3r, line 223: изображение

Shane32: I'd appreciate it if you'd verify the numbers that I listed.

Shane32: Maybe known fragment names should be 5.5.1.1 ?

Shane32: (because of "fragments must be a set of one.")

docs2/site/docs/getting-started/errors.md
```diff
@@ -200,3 +200,36 @@ For example, to show the stack traces for unhandled errors during development, y
...
+NoFragmentCycles             | NO_FRAGMENT_CYCLES               | 5.5.2.2
+PossibleFragmentSpreads      | POSSIBLE_FRAGMENT_SPREADS        | 5.5.2.3
+ArgumentsOfCorrectType       | ARGUMENTS_OF_CORRECT_TYPE        | 5.6.1
```
sungam3r, line 226: ARGUMENTS_OF_CORRECT_TYPE and DEFAULT_VALUES_OF_CORRECT_TYPE both have 5.6.1

Shane32: Let me know what numbers to use; I'll make the appropriate changes.

docs2/site/docs/getting-started/errors.md
sungam3r: UniqueFragmentNames is 5.5.1.1

Shane32: Yes, I know.  I don't know what you want me to do.

sungam3r: I think KnownFragmentNames is 5.5.1.2 and KnownTypeNames is some sort of aggregated check which is not directly listed in the specification. KnownTypeNames ia about all types defined in document - variable definitions and fragment conditions. All 5.5 paragraph is about fragments. Variable definition is handled by 5.8.2 - `IsInputType` to some extent implies that the variable type must be defined.  Fragment type existence is handled by 5.5.1.2, so....  Maybe KnownTypeNames rule has become redundant and needs to be removed so as not to get confused?

Shane32: If we remove that check, then `use_undefined_fragment` incorrectly returns data.

> 5.5.1.1
> For each fragment definition fragment in the document
> Let fragmentName be the name of fragment.
> Let fragments be all fragment definitions in the document named fragmentName.
> **fragments must be a set of one.**

Fragments would be a set of zero, instead of one.  Test does not meet 5.5.1.1

Shane32: For reference, `use_undefined_fragment` is this:
```gql
query someDroids {
  r2d2: droid(id: ""3"") {
    ...unknown_fragment
    name
  }
}
```

Shane32: KnownFragmentNames has been updated to 5.5.1.1

Shane32: Actually that's wrong...... 5.5.1.1 is checking for duplicate fragments......

Shane32: Ok it needs to be 5.5.1.2 - `use_undefined_fragment` fails because it's not defined in the schema

docs2/site/docs/getting-started/errors.md
sungam3r: Looks like 5.6.1 really covers 2 cases for value literals - argument literal and default values for variables.

docs2/site/docs/getting-started/errors.md
Shane32: Oh I misread what you said.  let me check KnownTypeNames

docs2/site/docs/getting-started/errors.md
sungam3r: By the way, could you check `DefaultValuesOfCorrectType` rule: 
```c#
 if (inputType is NonNullGraphType nonNullType && defaultValue != null)  <--- I don't undestand this condition
                    {
                        context.ReportError(new ValidationError(
                            context.OriginalQuery,
                            "5.7.2",
                            BadValueForNonNullArgMessage(
                                name,
                                context.Print(nonNullType),
                                context.Print(nonNullType.ResolvedType)),
                            defaultValue));
                    }
```

docs2/site/docs/getting-started/errors.md
Shane32: Ok I did it wrong again.  5.5.1.2 is "fragment spread TYPE EXISTENCE" and 5.5.2.1 is "Fragment spread TARGET DEFINED"

So I think KnownFragmentNames is 5.5.2.1 and KnownTypeNames is 5.5.1.2

docs2/site/docs/getting-started/errors.md
Shane32: The code you pointed out covers the test `no_required_variables_with_default_values`:

```gql
query UnreachableDefaultValues($a: Int! = 3, $b: String! = ""default"") {
  dog { name }
}
```

Where the failure is that a non-null type (`Int!`) requires the argument to be provided and so should not specify a default value.

Shane32:
Please let me know if you see any other errors in the numbering

docs2/site/docs/getting-started/errors.md
sungam3r: It seems to me that all this numbering will still need to be checked again someday. We will need to add the missing rules, then it will be possible to find out the exact numbering. I'm OK with any current numbering.

docs2/site/docs/getting-started/errors.md
sungam3r: I think it's wrong. I've already seen a similar situation for argument default values. And I told you about it. I think the situation is similar with variables, although I cannot easily find words about this in the specification. I talk about that:
> Arguments can be required. An argument is required if the argument type is non‐null and does not have a default value. Otherwise, the argument is optional.

Let's make a separate issue for this, so as not to keep this PR opened.

Shane32: I agree it's probably wrong.  I'm happy to set it however you suggest.  It seems to me like this is more of a warning than an error that violates spec.  Someone coded it, so perhaps it was explicitly an error in a prior version of the spec; I don't know.  Anyway, I'll open an issue for it.

sungam3r (approved):

docs2/site/docs/getting-started/errors.md
sungam3r: I don't know too.

Issue #781, opened 2018-08-13
Lost task exception in ParallelExecutionStrategy


SapphireBrand:
## Description

I am seeing tasks whose exception is not observed, which later causes an exception when the task is garbage collected

### Steps to reproduce
In ParallelExecutionStrategy.cs there is the following code around line 21:

                // Start executing all pending nodes
                for (int i = 0; i < pendingNodes.Count; i++)
                {
                    context.CancellationToken.ThrowIfCancellationRequested();
                    currentTasks[i] = ExecuteNodeAsync(context, pendingNodes[i]);
                }

ThrowIfCancellationRequested will throw an exception if the current request has been cancelled. However, existing tasks have been queued, so they might throw exceptions. Since the Task.WhenAll is after the loop, those exceptions will not be awaited.

Therefore, to trigger the bug, you should have tasks that will fail but not be cancelled. For example, a synchronous node can throw an exception.

### Expected result

All tasks should be cancelled.

(If the node is async then it should use the cancellation token directly. But I think that synchronous fields could have exceptions that are not observed.)

### Actual result

I am seeing lost exceptions on synchronous fields.

### Environment

I don't think the environment matters, but it was GraphQl 2.0.0-alpha-978, .Net 4.6.2 on 64-bit Windows 7, using Any CPU platform target.

joemcbride:
Think you could write a test to demonstrate this behavior?

sungam3r:
I did some research and am not sure that this is a bug. First, it must be said that the described behavior will be only in the case of setting `ExecutionOptions.ThrowOnUnhandledException = true`. Otherwise, all errors are [intercepted here](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Execution/ExecutionStrategy.cs#L252) and the created tasks **always** (except of exotic exceptions like `StackOverflowException` or `ThreadAbortException` which I do not consider) go to the `RanToCompletion` state with some execution error in context and null result.

If `ExecutionOptions.ThrowOnUnhandledException = true` then such a situation with unobserved exception does arise, but not for every exception.

Let's take an example:
```c#
using GraphQL.Types;
using System;
using System.Threading;
using System.Threading.Tasks;
using Xunit;

namespace GraphQL.Tests.Bugs
{
    public class Bug781UnobservedFaultedTasks
    {
        [Fact]
        public void cancel_execution_should_not_create_zombie_tasks()
        {
[...]

PR #1824, opened 2020-08-28
Add test for #781

sungam3r:
See #781

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug781UnobservedTasks.cs
+++ b/src/GraphQL.Tests/Bugs/Bug781UnobservedTasks.cs
@@ -1,0 +1,96 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    public class Bug781UnobservedTasks
+    {
+        private bool _unobserved;
+
+        [Theory]
+        [InlineData("{ do(throwCanceled: false) cancellation never }", true)]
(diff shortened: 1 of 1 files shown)


Shane32:
As previously stated, I don't like tests that are dependent on a time factor.  I've found them to be unreliable, as if the azure VM is running slow, then it breaks the test.  Plus, in this particular test, there's nothing to say that the opposite behavior is undesired.

Merge if you like; but I wouldn't.

src/GraphQL.Tests/Bugs/Bug781UnobservedTasks.cs
```diff
@@ -0,0 +1,96 @@
...
+                    options.ThrowOnUnhandledException = true; // required
+                });
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine(ex);
+            }
+
+            GC.Collect(); // GC causes UnobservedTaskException event
+            await Task.Delay(1000); // Wait some time for GC to complete
```
Shane32, line 40:
```suggestion
            GC.Collect(); // call finalizer of Task
            await Task.Delay(1000); // Wait some time for GC to complete
            GC.Collect(); // Task is collected and GC causes UnobservedTaskException event
            await Task.Delay(1000); // Wait some time for GC to complete
```

Shane32: This fixes the test, at least on my machine.  Why?  I don't know; I'd just be guessing.  Presumably you had the code working before.  Another example of the test being unreliable.

Shane32:
`[Fact(Skip)]` 👍

sungam3r:
Indeed the results were unstable.

PR #1826, opened 2020-08-28
Try to skip build on changes in docs

sungam3r:
I think it's time to suppress the build when we change the documentation only.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,10 @@ pull_requests:
   do_not_increment_build_number: true
 skip_tags: true
 skip_branch_with_pr: false
+skip_commits:
+  files:
+    - docs2/*
+    - README.md
 test: off
 
 os: Visual Studio 2019


appveyor.yml
```diff
@@ -3,6 +3,10 @@ pull_requests:
   do_not_increment_build_number: true
 skip_tags: true
 skip_branch_with_pr: false
```
sungam3r, line 5: Also it seems that it should be `skip_branch_with_pr: true` . @joemcbride  ?

Shane32:
Sounds great, but I'm no expert on yaml files.  I use GitHub Actions for my own projects, struggling through the CD scripts.  I have noticed that when we make PRs, it always runs two checks, whereas if you make a PR from a separate repo, it only does one check.  We can probably delete the duplicate check.

PR from same repo:
image

PR from different repo:
image

I think we should configure it to also run against .net framework, since we target .Net Standard 2.0.  Right now there's a few tests that fail (minor issues).

sungam3r:
Setting up CI is a separate big topic to work on. I don't do this in my main job. Basically, I've wanted to try it for a long time. Maybe next month I'll do it.

Shane32:
If we use GitHub Actions, I can probably do it, since I have most of the scripts set up for my own projects.  We can probably use them both in tandem and see if we like it or if we should just stick with appveyor.  I'm fine either way; up to you -- but I may not be much help with appveyor.

appveyor.yml
joemcbride: Seems like there was a reason for having that false, but I can double check.

joemcbride:
There are some things I like about AppVeyor, but I'm definitely for trying GitHub actions and using GitHub's package hosting vs. AppVeyor + MyGet.  That would make it much easier to maintain.

joemcbride (approved):

Shane32:
@joemcbride Ok then I'll check into it.

PR #1827, opened 2020-08-28
Mark DocumentListener.BeforeExecutionStepAwaited obsolete

Shane32:
@sungam3r what do you think?  I believe these methods were added purely to support the old data loader, and it is due to this design that caused all the deadlock issues people were having.

More specifically, people would use this event to:
1. Create a task that never completed
2. Wait for this "before awaited" execution step to occur
3. Complete the task

So if one task was waiting for another, the "before awaited" step would only run once and break the system.

Another example is if someone throws an exception when `ThrowOnUnhandledException` is true.  It would never run the "before awaited" step for the remaining tasks and never complete the abandoned task.

So it seems to me that the use case of this event is flawed from its outset.

Can you think of any other reasons to have this event handler?  It does not run per-field resolver, such as with the parallel execution strategy.  So it can't be used for places that middleware could be.

I feel that we should mark these events as obsolete, but not delete them, as there may be people with custom dataloader-like code that is still relying on its behavior.  Then there is plenty of time to have people let us know if they need the functionality.


sungam3r:
I haven't had time to understand/review dataloaders. I'm going to do that. I can only say that my intuition tells me that you are right. Maybe @joemcbride will give more information.

joemcbride:
It's a little hazy, but I think at least some of these methods were made available to do things such a query batching in some data libraries such as Marten.  It has been a while since I used that though, so I would need to look at it further.

sungam3r (approved):

Shane32:
Those should not be alerts.  There are code comments specifically to ignore that warning (which was already detected by C#).

sungam3r:
False positives maybe.

Issue #1832, opened 2020-09-02
EnumerationGraphType no longer supports mapping Enums of different names 3.0.0-preview-1719


asgerl:
Hey. 

I have encountered a change in behavior from 3.0.0-preview-1648 to 3.0.0-preview-1719 which I am not sure is intentional or whether it is a bug.

I could not find a similar issue - likely because 3.0.0-preview-1719 was released recently. 

## Summary

**3.0.0-preview-1648 behavior:**

FYI: This is not a real life example, it is just an example which boils down what the change in behavior is.

In 3.0.0-preview-1648 if I create an enum.

```
public enum IsSet
{
      NotSet = 0,
      Set = 1
}
```

I could then define a EnumerationGraphType like so.

```
public class AssignmentStateEnumType  : EnumerationGraphType {
    public AssignmentStateEnumType()
    {
      Name = "AssignmentState";
      Description = "State of assignment";

      AddValue(name: "NotAssigned", description: "There is no assignment", value: 0);
      AddValue(name: "Assigned", description: "There is an assignment", value: 1);
    }
}
```

Finally I could define the GraphQL field as such.

```
Field<AssignmentStateEnumType>("assignmentState", "State of assignment field", resolve: context => context.Source.IsSet);
```

The above assume that context.Source.IsSet is an instance of the IsSet enum. This field would resolve successfully, e.g. if the instance of IsSet has the value NotSet then my GraphQL API would return NotAssigned - since NotSet and NotAssigned both have the value 0.

To clarify, it was possible to map a C# enum to a EnumerationGraphType even if its values has different names.

**3.0.0-preview-1719 behavior:**

The above no longer resolves successfully, instead I get the following exception: _GraphQL.ExecutionError: Unable to serialize 'NotSet' to 'AssignmentState'_

I can work around this error by renaming 'NotSet' on the enum to 'NotAssigned' or vise versa. However I use GraphQL as a public API while my Enums are internal part of my server code - so for abstraction purposes I liked having the option to output GraphQL enums with different names.

**Question:**
Is this an intentional change? If so what was the reason behind changing the behavior?

## Relevant information

**Own investigation:**

I found [this pull request](https://github.com/graphql-dotnet/graphql-dotnet/pull/1701) for 3.0.0-preview-1719 which changes the behavior of ScalarGraphType from which EnumerationGraphType inherits from. Maybe these changes to ScalarGraphType was not meant to affect EnumerationGraphType like it did?

[...]

PR #1835, opened 2020-09-03
Add upgrade notes for custom enum types

Shane32:
See issue #1832

### DIFF
--- a/docs2/site/docs/guides/migration3.md
+++ b/docs2/site/docs/guides/migration3.md
@@ -274,7 +274,7 @@ that are required to be implemented.
 ### Enumeration Graph Types
 
 Version 3.0 has stricter type checking when serializing or deserializing between types, including
-enumeration types. Since C# does not equate a enumeration member to its value (e.g. the number `1`
+enumeration types. Since C# does not equate an enumeration member to its value (e.g. the number `1`
 does not equal `MyEnum.Value1`), you may need to make changes to your custom enumeration graph
 types as shown in the following example:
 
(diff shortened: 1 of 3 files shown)


Shane32:
Also added some tests to the many enumeration tests within `Bug1699InvalidEnum.cs`.  No functional changes.

docs2/site/docs/guides/migration3.md
```diff
@@ -271,6 +271,31 @@ the `ResolveFieldContext` class has changed. Custom implementations will need to
 new properties. Please see the `IResolveFieldContext` interface for a complete list of properties
 that are required to be implemented.
 
+### Enumeration Graph Types
+
+Version 3.0 has stricter type checking when serializing or deserializing between types, including
+enumeration types. Since C# does not equate a enumeration member to its value (e.g. the number `1`
```
sungam3r, line 277:
```suggestion
enumeration types. Since C# does not equate an enumeration member to its value (e.g. the number `1`
```
?

sungam3r (approved):

Shane32:
FYI, I believe that there was no breaking change from 2.4.0 apart from throwing an exception upon a failed conversion of an enum (as it should).

See https://github.com/graphql-dotnet/graphql-dotnet/pull/1836#issuecomment-688484028

Shane32:
Since there was no breaking changes from 2.4, should I remove the added notes to the upgrade guide?

sungam3r:
> FYI, I believe that there was no breaking change from 2.4.0 apart from throwing an exception upon a failed conversion of an enum (as it should).

throwing an exception **IS** a breaking change

sungam3r:
Well, after #1836 I think we actually do not need upgrade notes about enums as they just work. At least you need to change the wording because the examples will work.

Shane32:
> Well, after #1836 I think we actually do not need upgrade notes about enums as they just work. At least you need to change the wording because the examples will work.

I agree -- if we merge #1836 then I don't think this PR is necessary.

PR #1836, opened 2020-09-04
Enhancement: allows enumeration members to be specified by underlying value

Shane32:
See #1832 and #1835 

Given:
```csharp
public enum IsSet
{
    NotSet = 0,
    Set = 1
}
```

Since C# does not consider `IsSet.NotSet` to equal `0`, an `EnumerationGraphType` must add values based on the enumeration member.  For example:

```csharp
public class AssignmentStateEnumType  : EnumerationGraphType {
    public AssignmentStateEnumType()
    {
        //this does not work
        AddValue(name: "NotAssigned", description: "There is no assignment", value: 0);
        AddValue(name: "Assigned", description: "There is an assignment", value: 1);
        //this is correct
        AddValue(name: "NotAssigned", description: "There is no assignment", value: IsSet.NotSet);
        AddValue(name: "Assigned", description: "There is an assignment", value: IsSet.Set);
    }
}
```

~~This did work in prior versions of `GraphQL.NET`.~~  This PR ~~reverts~~ **enhances** that behavior by converting the enum value to the underlying type when adding to the `EnumValues` collection, and then performing enum comparisons based on the underlying type.

There are a couple points worth of consideration here:

* It is strictly against GraphQL spec to allow enum arguments to be specified by number.  This PR does not change that.  So it's only a "server-side" change; the graphs do not function any different.
* Care was taken to prevent this from being a breaking change in any form besides the specified change.

### DIFF
--- a/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
@@ -135,7 +135,7 @@ namespace GraphQL.Types
 
             // DO NOT USE LINQ ON HOT PATH
             foreach (var def in _values)
-                if (def.Value.Equals(value))
+                if (def.UnderlyingValue.Equals(value))
                     return def;
 
             return null;
@@ -156,13 +156,15 @@ namespace GraphQL.Types
         {
             get => _value;
             set {
+                _value = value;
                 if (value != null)
                 {
                     if (value is Enum)
                         value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
                 }
-                _value = value;
+                UnderlyingValue = value;
             }
         }
+        internal object UnderlyingValue { get; set; }
     }
 }
--- a/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
@@ -114,14 +114,7 @@ namespace GraphQL.Types
 
         public EnumValueDefinition this[string name] => FindByName(name);
 
-        public void Add(EnumValueDefinition value)
-        {
-            if (value == null)
-                throw new ArgumentNullException(nameof(value));
-            if (value.Value is Enum enumValue)
-                value.Value = Convert.ChangeType(enumValue, Enum.GetUnderlyingType(enumValue.GetType()));
-            _values.Add(value);
-        }
+        public void Add(EnumValueDefinition value) => _values.Add(value ?? throw new ArgumentNullException(nameof(value)));
 
         public EnumValueDefinition FindByName(string name, StringComparison comparison = StringComparison.OrdinalIgnoreCase)
         {
@@ -158,6 +151,18 @@ namespace GraphQL.Types
         public string Name { get; set; }
         public string Description { get; set; }
         public string DeprecationReason { get; set; }
-        public object Value { get; set; }
+        private object _value;
+        public object Value
+        {
+            get => _value;
+            set {
+                if (value != null)
+                {
+                    if (value is Enum)
+                        value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
+                }
+                _value = value;
+            }
+        }
     }
 }
--- a/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
@@ -163,6 +163,9 @@ namespace GraphQL.Types
                 UnderlyingValue = value;
             }
         }
+        /// <summary>
+        /// For enums, contains the underlying enumeration value; otherwise contains <see cref="Value" />.
+        /// </summary>
         internal object UnderlyingValue { get; set; }
     }
 }
--- a/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
@@ -114,7 +114,14 @@ namespace GraphQL.Types
 
         public EnumValueDefinition this[string name] => FindByName(name);
 
-        public void Add(EnumValueDefinition value) => _values.Add(value ?? throw new ArgumentNullException(nameof(value)));
+        public void Add(EnumValueDefinition value)
+        {
+            if (value == null)
+                throw new ArgumentNullException(nameof(value));
+            if (value.Value is Enum enumValue)
+                value.Value = Convert.ChangeType(enumValue, Enum.GetUnderlyingType(enumValue.GetType()));
+            _values.Add(value);
+        }
 
         public EnumValueDefinition FindByName(string name, StringComparison comparison = StringComparison.OrdinalIgnoreCase)
         {
@@ -128,6 +135,11 @@ namespace GraphQL.Types
 
         public EnumValueDefinition FindByValue(object value)
         {
+            if (value is Enum)
+            {
+                value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
+            }
+
             // DO NOT USE LINQ ON HOT PATH
             foreach (var def in _values)
                 if (def.Value.Equals(value))
--- a/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
@@ -155,7 +155,8 @@ namespace GraphQL.Types
         public object Value
         {
             get => _value;
-            set {
+            set
+            {
                 _value = value;
                 if (value != null)
                 {
--- a/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
@@ -158,11 +158,8 @@ namespace GraphQL.Types
             set
             {
                 _value = value;
-                if (value != null)
-                {
-                    if (value is Enum)
-                        value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
-                }
+                if (value is Enum)
+                    value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
                 UnderlyingValue = value;
             }
         }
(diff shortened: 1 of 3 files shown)


Shane32:
Personally I'm rather indifferent.  If I chose, I probably would not merge this PR, since it's C# that's preventing the comparison, and by modifying this behavior, you could say that we are changing the default C# mechanics.  However, since GraphQL.NET 2.4.0 allowed specifying by underlying value, it would make sense to allow the same functionality in 3.0.0.  Either way is fine with me.

joemcbride:
Thanks for taking a look at this.  From what I recall, I had this working elsewhere in the process.  I think really the only place that needs to do the conversion is if you want the `context.Source`, or `context.GetArgument` converted to the .NET type.  So that would happen in the `ToObject` extension method if I'm remembering correctly.

Shane32:
I believe that's correct: `GetArgument<int>` still works for enums, and `GetArgument<MyEnum>` works for `int`s.  I don't quite understand what you're trying to say, though -- did you want me to change or look at something else?  I can add tests for `GetArgument<>` as described above...  ???

Shane32:
I have not investigated why it worked in prior versions.

Shane32:
I've reviewed prior behavior.  Enums were matched by name, not by number.  So if `MyEnum.Set` equals 1, and if `MyEnum.Set` was returned from a resolver, it would get serialized to "Set" and then looked up in the list of strings in the `EnumerationGraphType`.  Only if `Set` was defined in the list would it get returned.  For instance, if "IS_SET" was defined in the enumeration as equaling 1, it would not match.  Without a match, a null value is silently returned, rather than an exception being thrown.

In practical terms, that means that as long as the enums were returned from the resolvers, it would not make any difference what the enum graph type value was, or even if it was correct, since it would typically be matched by name.  It's just when customizing enums to have a different name does this issue come into play.

That prior behavior is still retained even without this PR as demonstrated by the `Bug1699InvalidEnum.String_Enum` test.  As such, this PR represents an enhancement rather than a regression fix, as it allows for matching by numerical value.  (All of the new tests in this PR failed with 2.4.0.)

Finally, it is important to know that with or without this patch, currently enums match on name over number.  I think this is incorrect behavior, and that matching by value should take priority.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -146,6 +151,20 @@ public class EnumValueDefinition : MetadataProvider
         public string Name { get; set; }
         public string Description { get; set; }
         public string DeprecationReason { get; set; }
-        public object Value { get; set; }
+        private object _value;
+        public object Value
+        {
+            get => _value;
+            set {
```
sungam3r, line 158:
```suggestion
            set
            {
```

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -146,6 +151,21 @@ public class EnumValueDefinition : MetadataProvider
...
+                _value = value;
+                if (value != null)
+                {
+                    if (value is Enum)
+                        value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
+                }
+                UnderlyingValue = value;
+            }
+        }
+        internal object UnderlyingValue { get; set; }
```
sungam3r, line 169:
```suggestion
        internal object EnumUnderlyingValue { get; set; }
```

To focus that is is only for enum special cases.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -146,6 +151,21 @@ public class EnumValueDefinition : MetadataProvider
...
+        {
+            get => _value;
+            set
+            {
+                _value = value;
+                if (value != null)
+                {
+                    if (value is Enum)
+                        value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
+                }
```
sungam3r, line 165:
```suggestion
                if (value is Enum)
                    value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
```

sungam3r:
> Finally, it is important to know that with or without this patch, currently enums match on name over number. I think this is incorrect behavior, and that matching by value should take priority.

Probably yes. I'm sure that most of the time resolvers return enumeration values, not names. Also it would allow us to eliminate allocation from `  var valueString = value.ToString();`.

sungam3r:
> As such, this PR represents an enhancement rather than a regression fix, as it allows for matching by numerical value.

To be honest, I've already forgotten the history of changes for enums. Let's just add the ability to compare enum by its numeric value.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -128,9 +128,14 @@ public EnumValueDefinition FindByName(string name, StringComparison comparison =
...
         {
+            if (value is Enum)
+            {
+                value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
+            }
+
             // DO NOT USE LINQ ON HOT PATH
             foreach (var def in _values)
-                if (def.Value.Equals(value))
+                if (def.UnderlyingValue.Equals(value))
```
sungam3r, line 138: By the way, think about such situation:
```c#
enum ParityEnum
{
  Zero = 0,
  One = 1,
  Two = 0,
  Three = 1,
  Four = 0
}
```

CLR enum itself can be thought of as a pair of name and arbitrary (not necessarily unique) value. You probably guess what strange errors can be in this case, if you start comparing by `UnderlyingValue` instead of `Value`? The results will vary depending on the order in which the enumeration items are registered.

Maybe we can eleminate these cases by that code?
```c#
 public EnumValueDefinition FindByValue(object value)
        {
            // DO NOT USE LINQ ON HOT PATH
            foreach (var def in _values)
                if (def.Value.Equals(value))
                    return def;

            if (value is Enum)
            {
                value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));

                // DO NOT USE LINQ ON HOT PATH
                foreach (var def in _values)
                    if (def.UnderlyingValue.Equals(value))
                        return def;
            }

            return null;
        }
```

Shane32: Good idea, but doesn't work in practice.  C# equates enumeration members with the same value.  So `ParityEnum.Zero == PairtyEnum.Two` returns `true`.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
Shane32: Well, it's used when comparing for all types.  It's only modified to the enum raw value for `Enum` types.  I don't mind changing it, but I'm not sure if that's any more appropriate.  Lmk if you still think we should change it; doesn't matter to me, and it's only an `internal` property anyway.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
Shane32: It is also noteworthy that `Console.WriteLine(ParityEnum.Zero)` prints the text `Four`, so name based matching would not work either.

Shane32:
> > As such, this PR represents an enhancement rather than a regression fix, as it allows for matching by numerical value.
> 
> To be honest, I've already forgotten the history of changes for enums. Let's just add the ability to compare enum by its numeric value.

Sounds fine to me.  We can merge this whenever you think it's ready, and close PR #1835 without merging it.

Shane32:
> > Finally, it is important to know that with or without this patch, currently enums match on name over number. I think this is incorrect behavior, and that matching by value should take priority.
> 
> Probably yes. I'm sure that most of the time resolvers return enumeration values, not names. Also it would allow us to eliminate allocation from ` var valueString = value.ToString();`.

After this PR is merged, I will propose a breaking change for 4.0 that removes or demotes the match-by-name functionality.  We can then discuss whether to proceed with that change for 4.0 or not.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
sungam3r: Yeah. I can guess why. Enumeration of values goes from the end. There are always some problems with enums.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
sungam3r: > Well, it's used when comparing for **all** types.

I missed it. I'm OK to keep `UnderlyingValue` though it looks strange. What is _underlying_ for string or bool ?

sungam3r:
In what cases do we ever need match-by-name functionality?

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
Shane32: Agreed; I'll leave it alone for now but lmk if you think of anything more appropriate

sungam3r (approved):

Shane32:
> In what cases do we ever need match-by-name functionality?

I cannot think of any.  The naming of the values in the schema should not have any effect on the rest of the code.  I would remove the match-by-name capability from 4.0.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
sungam3r: Let's just write a short comment and it will be enough. We have already spent too much time with enums.

Shane32:
Before I merge this: are we sure we want to add this feature to the codebase?  I mean, it's not like any prior version had this capability, and it's not like someone can't just use the enum member as the value when defining their `EnumGraphType` (as they probably have been doing).  I personally don't know that it really provides any great enhancement to the project and would probably not merge it.  But either way is fine with me.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
Shane32:
```suggestion
        /// <summary>
        /// For enums, contains the underlying enumeration value; otherwise contains <see cref="Value" />.
        /// </summary>
        internal object UnderlyingValue { get; set; }
```

sungam3r:
One can register values for enumeration graph type with `int` (in one assembly/codebase) and other can use it in `GetArgument<T>` with appropriate enum from another assembly (codebase). So int <-> enum conversion out of the box will not hurt I think.

Shane32:
OK.  I added an xml comment; ok to merge?

sungam3r:
Yes.

PR #1841, opened 2020-09-06
Fixes tests for .Net Core 2.1 LTS and .Net Framework 4.8

Shane32:
Fixes tests with .Net Core 2.1 LTS and .Net Framework 4.8 due to the following:
- System.Text.Json does not properly deserialize floating point numbers in prior versions of .Net Core - it has rounding errors, not sure why.  The broken test was actually testing serialization, which works just fine in prior versions.
- Some native exception messages have changed over the different framework versions
- Parsing overflowed floating point values changed in .Net Core 3.1 (bug fix for prior .Net Core versions)

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
+++ b/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
@@ -5,7 +5,11 @@ namespace GraphQL.Tests.Bugs
 {
     public class Bug138DecimalPrecisionTests : QueryTestBase<DecimalSchema>
     {
+#if NETCOREAPP3_1
         [Fact]
+#else
+        [Fact(Skip = "Deserialization error with .NET Core 2.1")]
+#endif
         public void double_to_decimal_does_not_lose_precision()
         {
             var query = @"
--- a/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
+++ b/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
@@ -8,7 +8,7 @@ namespace GraphQL.Tests.Bugs
 #if NETCOREAPP3_1
         [Fact]
 #else
-        [Fact(Skip = "Deserialization error with .NET Core 2.1")]
+        [Fact(Skip = "Deserialization error with .NET Core < 3.1")]
 #endif
         public void double_to_decimal_does_not_lose_precision()
         {
--- a/src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
@@ -68,7 +68,7 @@ namespace GraphQL.Tests.Utilities
         public ExecutionResult CreateQueryResult(string result) => result.ToExecutionResult();
 
         protected string ReadSchema(string fileName)
-            => File.ReadAllText(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "Files", fileName));
+            => File.ReadAllText(Path.Combine(Directory.GetCurrentDirectory(), "Files", fileName));
     }
 
     public class ExecuteConfig
--- a/src/GraphQL/Language/CoreToVanillaConverter.cs
+++ b/src/GraphQL/Language/CoreToVanillaConverter.cs
@@ -242,7 +242,7 @@ namespace GraphQL.Language
                         CultureInfo.InvariantCulture,
                         out var dbl) == false)
                     {
-                        dbl = str.Value.StartsWith("-") ? double.NegativeInfinity : double.PositiveInfinity;
+                        dbl = str.Value[0] == '-' ? double.NegativeInfinity : double.PositiveInfinity;
                     }
 
                     //it is possible for a FloatValue to overflow a decimal; however, with a double, it just returns Infinity or -Infinity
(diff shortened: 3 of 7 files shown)


src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -236,10 +236,13 @@ public IValue Value(GraphQLValue source)
...
                     // for example, 12.1 or 12.11 is double but 12.10 is decimal
-                    double dbl = double.Parse(
+                    if (double.TryParse(
                         str.Value,
                         NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent,
-                        CultureInfo.InvariantCulture);
+                        CultureInfo.InvariantCulture,
+                        out var dbl) == false) {
+                        dbl = str.Value.StartsWith("-") ? double.NegativeInfinity : double.PositiveInfinity;
+                    }
```
Shane32, line 245: Without this change, floating point numbers that overflow the maximum/minimum value will throw a format exception rather than some type of exception that inherits from `DocumentError`.  This change makes prior versions of .Net Core act the same as .Net Core 3.1, where overflowed values return negative or positive infinity.

src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -236,10 +236,13 @@ public IValue Value(GraphQLValue source)
...
                     // the idea is to see if there is a loss of accuracy of value
                     // for example, 12.1 or 12.11 is double but 12.10 is decimal
-                    double dbl = double.Parse(
+                    if (double.TryParse(
                         str.Value,
                         NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent,
-                        CultureInfo.InvariantCulture);
+                        CultureInfo.InvariantCulture,
+                        out var dbl) == false) {
+                        dbl = str.Value.StartsWith("-") ? double.NegativeInfinity : double.PositiveInfinity;
```
sungam3r, line 244:
```suggestion
                        dbl = str.Value[0] == '-' ? double.NegativeInfinity : double.PositiveInfinity;
```

src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -236,10 +236,13 @@ public IValue Value(GraphQLValue source)
 
                     // the idea is to see if there is a loss of accuracy of value
                     // for example, 12.1 or 12.11 is double but 12.10 is decimal
-                    double dbl = double.Parse(
+                    if (double.TryParse(
                         str.Value,
                         NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent,
-                        CultureInfo.InvariantCulture);
+                        CultureInfo.InvariantCulture,
+                        out var dbl) == false) {
```
sungam3r, line 243:
```suggestion
                        out var dbl) == false)
                    {
```

src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
```diff
@@ -68,7 +68,7 @@ public ExecutionResult AssertQuery(Action<ExecutionOptions> options, ExecutionRe
         public ExecutionResult CreateQueryResult(string result) => result.ToExecutionResult();
 
         protected string ReadSchema(string fileName)
-            => File.ReadAllText(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "Files", fileName));
+            => File.ReadAllText(Path.Combine(Directory.GetCurrentDirectory(), "Files", fileName));
```
sungam3r, line 71: Why? Current dir may be changed (`SetCurrentDirectory`), it is a mutable concept. I used `Assembly.GetExecutingAssembly().Location` specifically to ensure correct paths.

src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
```diff
@@ -5,7 +5,11 @@ namespace GraphQL.Tests.Bugs
 {
     public class Bug138DecimalPrecisionTests : QueryTestBase<DecimalSchema>
     {
+#if NETCOREAPP3_1
         [Fact]
+#else
+        [Fact(Skip = "Deserialization error with .NET Core 2.1")]
```
sungam3r, line 11:
```suggestion
        [Fact(Skip = "Deserialization error with .NET Core < 3.1")]
```

src/GraphQL/Language/CoreToVanillaConverter.cs
sungam3r: OK

src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
Shane32: I don't know.  `Assembly.GetExecutingAssembly().Location` gives some oddball path when running tests with .NET Framework 4.8.  Probably something peculiar about Visual Studio's test runner.  In any case, using `Directory.GetCurrentDirectory()` works properly for both .NET Framework 4.8 and .NET Core 2.1 / 3.1.  I can take the data and embed it as a resource into the assembly, or I can just copy the text into a string constant.  But `Assembly.GetExecutingAssembly().Location` doesn't work.

sungam3r: It is strange. Well, let's use getcurrentdirectory, ok.

sungam3r (approved):

PR #1846, opened 2020-09-07
Setup GitHub Actions for develop branch

Shane32:
These GitHub Actions perform the following for the `develop` branch:
* For pull requests to the `develop` branch with changes to the `docs2` folder, build the documentation
* For pull requests to the `develop` branch with changes to the `src` folder,
  * Ubuntu
    1. Build the solution in release config
    2. Build the solution in debug config
    3. Test debug config against .NET Core 2.1 LTS & .NET Core 3.1 LTS
  * Windows
    1. Build the solution in debug config
    2. Test against .NET Framework 4.8, .NET Core 2.1 LTS and .NET Core 3.1 LTS
* For commits to the `develop` branch (e.g. after a pull request has been merged)
  1. Build the solution on ubunto in release config
  2. Pack the project into a nuget package with a revision number corresponding to a unique GitHub identifier
  3. TODO: Upload to myget

These tests will fail until #1841 is merged


src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
sungam3r, line 1: encoding?

src/NuGet.config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sungam3r, line 1: encoding?

src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
Shane32: I guess ... ???

.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Pack solution (release config)
+        working-directory: src
+        run: dotnet pack --no-restore -c Release -p:VersionSuffix=${{ github.run_id }} -o out
+      - name: Publish artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: Nuget packages
+          path: |
+            src/out/*.nupkg
+#      - name: Publish Nuget packages to GitHub registry
```
sungam3r, line 34: GitHub registry or MyGet?

sungam3r:
Are these actions run in a specific order or in an arbitrary order?

Shane32:
The jobs run concurrently and are in an arbitrary order. The steps of each job are in a specific order.

.github/workflows/build-artifacts.yml
Shane32: After much testing and investigation, I’ve determined that the GitHub registry is only accessible to authenticated users, so probably continue using MyGet. We should update the comment if/when we start using this in place of appveyor.

Note that I was unable to get it to upload to the GitHub registry with the GITHUB_TOKEN as shown in the documentation and multiple websites. It would only work with a personal access token. I was probably doing something wrong but I couldn’t figure it out. 

Second, the unique identifier is 8 digits or so instead of the nice short 4 digit ones created by appveyor.

However, the integration with GitHub for artifacts, in progress jobs and build errors is rather nice. It also runs notably faster than appveyor.

Shane32: I should probably get it working to the GitHub registry ... just to try it out. Maybe GitHub will allow public access at some point. We can always upload to both myget and GitHub packages at once rather easily. 

I just need to create a GitHub account with the proper privileges.

PR #1852, opened 2020-09-10
Rethrow canceled operations properly; do not call unhandled exception delegate

Shane32:
Right now, if a request is canceled, pending async operations get caught as an `OperationCanceledException` and passed to the `UnhandledExceptionDelegate`.

Then when the next node is attempting to be resolved, the cancellation token is checked and thrown back to the `DocumentExecuter`, which then runs the `UnhandledExceptionDelegate` a second time and serializes a response.

This code change makes it so that when an operation is cancelled, the `OperationCanceledException` is immediately rethrown back to the caller of the `DocumentExecuter` -- no unhandled exception delegates are called, and no `ExecutionResult` is returned.

Without this change, the `UnhandledExceptionDelegate` often runs twice when a pending document is canceled.  With typical database logging code in the delegate, this would result in unnecessary database logs when something as simple as a browser is closed while loading.  (Of course the delegate could be coded to ignore these exceptions.)

In any case, I would expect this (canceled operations rethrow to the caller) to be the expected operation of any `Async` method, including `DocumentExecuter.ExecuteAsync`, and would be ideal for typical web server applications.

### DIFF
--- a/src/GraphQL.Tests/Execution/Cancellation/CancellationTests.cs
+++ b/src/GraphQL.Tests/Execution/Cancellation/CancellationTests.cs
@@ -2,6 +2,7 @@ using System;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using GraphQL.Execution;
 using GraphQL.Types;
 using Shouldly;
 using Xunit;
@@ -24,6 +25,7 @@ namespace GraphQL.Tests.Execution.Cancellation
 
             Field<StringGraphType>("one", resolve: GetOneAsync);
             Field<StringGraphType>("two", resolve: GetTwoAsync);
+            Field<StringGraphType>("three", resolve: GetThreeAsync);
         }
 
         public Task<string> GetOneAsync(IResolveFieldContext<object> context)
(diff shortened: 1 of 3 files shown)


joemcbride (approved):

sungam3r:
Nice catch. In fact, this is one of the troubles I ran into and wanted to fix. Everything is written correctly - when the browser is closing (or something else), the token is canceled, but the result is still sent back (200 OK in case of server project). Because of this, we saw parasitic errors in our logs. Thanks for the fix. However, this is actually a backward incompatible change, previously the exception was handled, now it is not. So this commit should be published in version 4, not 3.x.

sungam3r:
By the way, the situation is also interesting in that when trying to write the `ExecutionResult` to the output stream, another exception occurred due to the fact that the browser (and http connection) was closed.

sungam3r:
We need to decide if we want to revert this commit on master and apply it to the develop branch.

Shane32:
I understand the behavior changes, but I think we should keep it in 3.1. The goal of 3.0 was to have the unhandled exception handler only handle “processing errors”. From that viewpoint, this is a bug fix, not a breaking change.

Secondly, from a practical viewpoint, it doesn’t negatively affect any typical code, and does not change any code that processes without the task being canceled. Basically, it’s an enhancement. 

With either of the above viewpoints, there’s no reason to push it back to 4.0.

Shane32:
Fortunately it is easy to patch in 3.0 - just check for an OperationCanceledException within the UnhandledExceptionDelegate and skip the database logging code

sungam3r:
I am OK with any option.

PR #1856, opened 2020-09-13
Fix all build warnings except CS1591 warnings

Shane32:
Mostly everything was one of the following:
* Incorrect xml comment syntax (missing member)
* Missing some, but not all, xml comments for parameters
* Shortcut syntax for methods and properties

### DIFF
--- a/src/GraphQL/EnumerableExtensions.cs
+++ b/src/GraphQL/EnumerableExtensions.cs
@@ -8,6 +8,7 @@ namespace GraphQL
         /// <summary>
         /// Performs the indicated action on each item.
         /// </summary>
+        /// <param name="items">The list of items to act on.</param>
         /// <param name="action">The action to be performed.</param>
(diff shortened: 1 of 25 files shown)


src/GraphQL/EnumerableExtensions.cs
```diff
@@ -8,6 +8,7 @@ public static class EnumerableExtensions
         /// <summary>
         /// Performs the indicated action on each item.
         /// </summary>
+        /// <param name="items">The list of items to act on.</param>
```
sungam3r, line 11: Not a list.

sungam3r (approved):

src/GraphQL/EnumerableExtensions.cs
Shane32: Huh?  It's not a `List<T>` I know, but it's a list......  

Please suggest a change

sungam3r: I'm OK with it.

PR #1859, opened 2020-09-14
Update xml comments for NameConverter

Shane32:


sungam3r (approved):

Shane32:
@sungam3r I set GitHub to require the new tests to pass before allowing merging.  It appears that if the test doesn't run (in this case because the `docs2` folder had not changed), it blocks merging.  The same would happen if only docs were changed and not code; the code build check wouldn't run and GitHub would block merging.

We should decide if we want to remove the merge condition, or just force both code and docs build checks to run every time the code has been updated.

For now I'm just going to remove the requirement the merge condition.  Lmk what you think.

sungam3r:
What is merge condition? Where is it?

Shane32:
GitHub -> `graphql-dotnet/graphql-dotnet` -> Settings tab -> Branches -> Branch Protection Rules

Click edit next to `master` to view how it's set up for the `master` branch.  I had it set up the similarly for the `develop` branch (with the new tests, not the appveyor tests), but removed it as noted above.

sungam3r:
Should we make develop branch as default?

Shane32:
I wouldn’t until we decide not to make any more 3.x versions.

PR #1864, opened 2020-09-16
Adds Join Monster .NET to the list of tools to use with GraphQL.NET

sitereactor:
Join Monster for .NET is a query planner and batch data fetcher between GraphQL and SQL built for GraphQL for NET.

Adding it to the list as it could be valuable to people using GraphQL for NET.


sitereactor:
@sungam3r yes, we'll do that in the near future. At the time of implementation that preview was the best option :)

I have created an issue for it here: https://github.com/umbraco/join-monster-dotnet/issues/2

sitereactor:
The library has now been updated https://github.com/umbraco/join-monster-dotnet/pull/3

sungam3r (approved):

PR #1865, opened 2020-09-16
Develop: Publish nuget artifacts to GitHub Packages

Shane32:
I figured out how to prevent the unique build number from being 8 digits long.  Now it will start at ~~1000~~ 1 and increment.

I've also figured out how to publish to GitHub Packages.  I'd like to merge this and see how usable GitHub Packages is compared to myget.


.github/workflows/build-artifacts.yml
```diff
@@ -16,21 +18,24 @@ jobs:
         uses: actions/setup-dotnet@v1
         with:
           dotnet-version: '3.1.x'
+          source-url: https://nuget.pkg.github.com/graphql-dotnet/index.json
+        env:
+          NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
sungam3r, line 23: Where does it come from?

.github/workflows/build-artifacts.yml
```diff
@@ -16,21 +18,24 @@ jobs:
...
       - name: Install dependencies
         working-directory: src
         run: dotnet restore
       - name: Build solution (release config)
         working-directory: src
-        run: dotnet build --no-restore -c Release -p:NoWarn=CS1591 -p:VersionSuffix=${{ github.run_id }}
+        run: dotnet build --no-restore -c Release -p:NoWarn=CS1591 -p:VersionSuffix=$(( 1000 + $GITHUB_RUN_NUMBER ))
```
sungam3r, line 29: Why 1000+ ?

sungam3r:
I need to understand where we are going. Do we end up wanting to automatically publish packages from the master branch (only from master?) to nuget? Or will we manually publish them to nuget from github registry?

sungam3r:
I made a special CI label. I have a feeling that we will have a lot of PRs until we set everything up.

.github/workflows/build-artifacts.yml
Shane32: The GitHub token has predefined permissions and is a built in feature. One of the permissions is publishing a package to GitHub packages for its own registry. It has an advantage of being invalid once the workflow ends.

.github/workflows/build-artifacts.yml
Shane32: Just to make it a 4 digit number. I don’t know enough about bash scripts to format a number with 4 digits.  We can change it.

Shane32:
I think the goal should be:
- when a change is made within master or develop, (usually by merging a pull request), automatically push new packages to GitHub Packages. Maybe also myget if we find that GitHub Packages is a problem in some way. 
And: when issuing a release on any branch, automatically publish a package to nuget with the specified version number. 

This PR is just testing the first part, and only on develop, and doesn’t interfere with appveyor code that pushes to myget (if any). 

Based on my work in the server project I’m pretty confident all these goals can be easily attained (without appveyor).

.github/workflows/build-artifacts.yml
sungam3r: Do I need to configure/enable this token from GitHub UI ?

.github/workflows/build-artifacts.yml
sungam3r: Is GITHUB_RUN_NUMBER a monotonically increasing counter? Can it go beyond the value 9999 (8999) and thus result will no longer be 4 digits? Why aim for a 4-digit number in version suffix at all?

Shane32: Yes, it is basically a run counter.

Yes, if we trigger the workflow > 9999 times, by making 10,000+ commits to master/develop.

I have no opinion - I simply chose that because appveyor created 4 digit numbers.  We don't have to.  The version number suffix would start at 1 (it's already at around 20 actually) and keep increasing.  So you'd get version numbers like 4.0.0.25, and eventually 4.0.0.103.  That's fine with me also.

.github/workflows/build-artifacts.yml
Shane32: No it's always enabled

Shane32: https://docs.github.com/en/actions/reference/authentication-in-a-workflow

https://github.com/actions/setup-dotnet

sungam3r:
> when issuing a release on any branch

How will it work? Placing a tag x.y.z ?

sungam3r:
Also could one set up github registry feed in IDE to get the latest packages quickly?

Shane32:
> > when issuing a release on any branch
> 
> How will it work? Placing a tag x.y.z ?

Yes.  When issuing the release, you just set the tag to the correct version number for that release.  It overrides anything in the project settings.

Shane32:
> Also could one set up github registry feed in IDE to get the latest packages quickly?

I believe so.  I suggest we merge this PR and try it out and see.  If authentication is a problem, we may still need to use myget.

.github/workflows/build-artifacts.yml
```diff
@@ -16,21 +18,24 @@ jobs:
...
       - name: Pack solution (release config)
         working-directory: src
-        run: dotnet pack --no-restore -c Release -p:VersionSuffix=${{ github.run_id }} -o out
+        run: dotnet pack --no-restore -c Release -p:VersionSuffix=$GITHUB_RUN_NUMBER -o out
       - name: Publish artifacts
         uses: actions/upload-artifact@v2
         with:
           name: Nuget packages
           path: |
             src/out/*.nupkg
```
Shane32, line 38:
```suggestion
            src/out/*
```
In case we publish symbols separately (currently embedded, I think)

.github/workflows/build-artifacts.yml
```diff
@@ -16,21 +18,24 @@ jobs:
...
-#      - name: Publish Nuget packages to GitHub registry
-#        working-directory: src
-#        run: dotnet nuget push out/*.nupkg -s "myget-graphql-dotnet" -k ${{ secrets.GITHUB_TOKEN }}
+      - name: Publish Nuget packages to GitHub registry
+        working-directory: src
+        run: dotnet nuget push out/*.nupkg
```
Shane32, line 41:
```suggestion
        run: dotnet nuget push out/*
```

Shane32:
This GitHub Action workflow will now create the documentation package and upload it to the workflow as an artifact.

sungam3r (approved):

.github/workflows/build-artifacts.yml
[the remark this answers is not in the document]
Shane32: yes

Shane32:
See this file for an example of how the tag would override the version from project properties:

https://github.com/graphql-dotnet/server/blob/github-packages/.github/workflows/publish.yml

There is no publish workflow in this PR at this time.  It would probably be a nearly identical copy of the above file.

Shane32:
Just to be clear, _this_ workflow auto builds when a PR is merged to the develop branch with changes to the src branch.  For the version number, it uses the prefix defined in the source, plus an auto-generated suffix.  The version number is not based on a tag.  The version prefix in the source is "4.0.0-preview" so an auto-generated version number might be "4.0.0-preview.34".  After building the project, it packs up the nuget packages and uploads them to GitHub Packages.  It also saves the built nuget packages and a copy of the current documentation as artifacts for this workflow.

See: https://github.com/graphql-dotnet/graphql-dotnet/blob/develop/src/Directory.Build.props

Any other questions before I merge this @sungam3r ?

Shane32:
Actually, I think I'd like to split this workflow into one for code and one for docs, based on the last discussion about docs auto publishing.

Shane32:
I just renamed the file; should be good for now.

sungam3r:
> Any other questions before I merge

All CI process with all scripts/properties/steps/variables/dependencies/etc is one big question for me 😄. Let's merge and go on.

.github/workflows/build-artifacts.yml
joemcbride: > I simply chose that because appveyor created 4 digit numbers

Appveyor started a 1, there have just been that many builds.  😉

sungam3r: OK

PR #1871, opened 2020-09-22
small fixes

sungam3r:


sungam3r:
@Shane32 Please look at the failed actions checks.

Shane32:
GitHub issued a broken release yesterday of their setup-dotnet workflow. Although it was reported almost immediately, apparently they didn’t think it important to roll back the changes (?!?) so I’ll need to update the workflow until they fix their code. Good news is that the new version, when fixed, will officially support targeting multiple frameworks. 

https://github.com/actions/setup-dotnet/issues/128

Shane32 (changes requested):
These changes are only correct while the file is in the repo.  Once published, it will not work with the `.md` extension.  So it is proper that it not contain the `.md` extension.  Unfortunately this means the links from within the repo.  Click links to test:

With md:
- https://graphql-dotnet.github.io/docs/getting-started/dependency-injection.md

Without:
- https://graphql-dotnet.github.io/docs/getting-started/dependency-injection

Shane32:
I re ran the tests and they now pass

sungam3r:
Can we try to change it and see what happens? At least it will be no worse (the link works in at least one case).

Shane32 (approved):

Shane32:
Sure. We just need to check the links when we release 4.0

sungam3r:
Will the documentation site be built automatically after the merge?

Shane32:
Now, no it won’t. 

I would like to add that functionality. Although probably not in the develop branch unless we add a version selector to the doc site. I was going to suggest that also. I have not had time to work on it.

Shane32:
Oh, sorry, yes it will be built but not saved. I could probably adjust the workflow to make a zip of the docs and add it as an artifact.

sungam3r:
The artifact itself is not that valuable. I would like to generate the site right away so that its state matches the source code. It may not be necessary for the develop branch, but it is required for the master. So merging now into develop.

Shane32:
I agree

PR #1872, opened 2020-09-22
Added maxBatchSize to DataLoaderContextExtensions

Rensvind:
Since maxBatchSize has been added to DataLoaderBase I added it to DataLoaderContextExtensions to ease usage of it.


Shane32:
Looks good.  You need to update the apiapprovals file however.  Run the tests; one will fail.  Delete the applicable `approved` file and rename the `received` file to `approved`.  Run the tests again and it should pass.  Then commit the change

sungam3r (changes requested):
It is still binary breaking changes. I'm OK with such changes but it should target develop branch.

Rensvind:
I hope I got it right this time.

Shane32:
@sungam3r Are you sure this is a breaking change?  They are static methods that cannot be inherited.

Shane32:
@Rensvind Can you merge from develop again?

sungam3r:
@Shane32 Imagine I wrote a library that uses these estension methods. Then some application uses my library (compiled code, nuget package, not source code). This application then updates the version of the dependency (GraphQL.NET) because it sees a minor/patch version change. As a result, an `MissingMethodException` exception will be thrown. Adding default parameter is backward compatible only for source code, but not to compiled one.

Shane32:
Ok; that's probably why Microsoft rarely uses optional parameters in their method signatures.

sungam3r:
> I hope I got it right this time.

Your PR contains extra commits from other contributors. Please rebase your branch onto develop and then force-push. Then you get the only one commit - your.

Shane32:
I merged the prior commits from master into develop, so if he merges from develop again, it should merge in the merge commit which will clean up the history.

Shane32 (approved):

Rensvind:
Thanks for being so patient with me, my git skills are non-existent 🙂

PR #1873, opened 2020-09-22
Patch GitHub workflows

Shane32:
See https://github.com/actions/setup-dotnet/issues/128


Shane32:
Based on the issue notes, as of a few minutes ago, the tag has been reverted to the old version, so this PR is not necessary.

PR #1875, opened 2020-09-23
Fix package upload workflow

Shane32:
It seems the key must be added to the `dotnet nuget push` command.  (Tested in my fork.)  One of my other projects did not require this here as it was configured with the `actions/setup-dotnet` command as described in the documentation.  No idea why it is required here.  Perhaps the `Directory.Build.props` file's `<clear/>` is messing with it.

### DIFF
--- a/.github/workflows/build-artifacts-code.yml
+++ b/.github/workflows/build-artifacts-code.yml
@@ -59,4 +59,4 @@ jobs:
             src/out/*
       - name: Publish Nuget packages to GitHub registry
         working-directory: src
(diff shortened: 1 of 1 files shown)


sungam3r:
> Perhaps the Directory.Build.props file's <clear/> is messing with it.

I can not see `clear` tag. What is it?

Shane32:
Sorry, in `NuGet.config`:

https://github.com/graphql-dotnet/graphql-dotnet/blob/develop/src/NuGet.config

PR #1878, opened 2020-09-27
Add support for Apollo Studio trace logging

mattjhosking:
This will require that NullTraceLogger is registered for those looking to ignore logs, or they pull in the GraphQL.ApolloStudio NuGet and use the ApolloReportingTraceLogger. Incredibly valuable for anyone looking to monitor and improve their GraphQL performance and reliability.

An example .NET Core BackgroundService that uses this:

```
    public class GraphQLTraceService : BackgroundService
    {
        private readonly ITraceLogger _traceLogger;
        private readonly ILogger<GraphQLTraceService> _logger;

        public GraphQLTraceService(ITraceLogger  traceLogger, ILogger<GraphQLTraceService> logger)
        {
            _traceLogger = traceLogger;
            _logger = logger;
        }

        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
        {
            _logger.LogDebug("GraphQLTraceService is starting.");

            while (!stoppingToken.IsCancellationRequested)
            {
                // Send every 20 seconds or when forced due to size threshold reached
                var nextExecution = DateTime.Now.AddSeconds(20);
                await Task.WhenAny(_traceLogger.ForceSendTrigger.WaitAsync(),  Task.Delay(Math.Max((int)(nextExecution - DateTime.Now).TotalMilliseconds, 0), stoppingToken));

                _logger.LogDebug("Sending queued traces...");
                await _traceLogger.Send();
            }

            _logger.LogDebug("GraphQLTraceService is stopping.");
        }
    }
```


Shane32:
Should this be part of the base library?  It seems that this functionality could be implemented as a wrapper just as easily, like this:

```csharp
public class ApolloStudioDocumentExecuter: IDocumentExecuter
{
    private readonly IDocumentExecuter _executer;
    private readonly ITraceLogger _traceLogger;
    public ApolloStudioDocumentExecuter (IDocumentExecuter baseExecuter, ITraceLogger traceLogger) {
        _executer = baseExecuter ?? throw new ArgumentNullException(nameof(baseExecuter));
        _traceLogger = traceLogger ?? throw new ArgumentNullException(nameof(traceLogger));
    }
    public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options) {
        DateTime start = DateTime.UtcNow;
        var result = await _executer.ExecuteAsync(options).ConfigureAwait(false);
        try {
            _traceLogger.LogTrace(start, options.OperationName, options.Query, result);
        }
        catch {
            // We don't want log tracing to break it - but we should notify the user of this
        }
        return result;
    }
}
```

And if the user is using Autofac as their dependency injection provider, they can register this class as a "decorator" without any changes to the rest of their project.  (With MS DI they would need to change the existing registration of the `IDocumentExecuter` interface.)

https://autofaccn.readthedocs.io/en/latest/advanced/adapters-decorators.html

```csharp
//MS DI (replacing normal DocumentExecuter registration)
services.AddSingleton<IDocumentExecuter>(services => {
    var traceLogger = services.GetRequiredService<ITraceLogger>();
    var executer = new DocumentExecuter();
    return new ApolloStudioDocumentExecuter(executer, traceLogger);
});

//Autofac (in addition to normal DocumentExecuter registration)
builder.RegisterDecorator<ApolloStudioDocumentExecuter, IDocumentExecuter>();
```

It just seems like there's no real negative to creating a new project for this functionality, rather than embedding it directly into GraphQL.NET.

mattjhosking:
Nice! Seems like a pretty clean way to handle it without any breaking changes. I'll have a look into migrating it into its own package.

PR #1881, opened 2020-09-28
Add test to demonstrate bug with aliased fields in input types used in a mutation argument that is a list

chrisg2:
See bug description at #1882


sungam3r:
Thanks for the repro. Thanks to repro, I managed to make a fix in just 5 minutes. It's good that I still remember the details. See #1893 , just one line of code to pass additional context.

PR #1884, opened 2020-10-01
AutoRegisteringObjectGraphType can register methods

dnsfour:
Hello. I did not understand why registration of methods was not implemented for AutoRegistraionObjectGraphType and decided to make my own implementation. If this is ok, then I propose to consider introducing a new class Optional<T> for the parameters of such methods, which will indicate whether values have been passed.

Example:
```
UpdatePerson(int id, Optional<string> name, Optional<int?> age){
   var person = getPerson(id);
   if(name != null) person.Name = name;
   if(age != null) person.Age = age;
   UpdatePerson(person);
}
```


Shane32:
Just FYI, I've written a fairly complex version of this functionality within PR #1376 in a separate but complementary library.  For instance, it can handle a method signature like this:

```csharp
[Description("Updates the specified person's name and optionally their age")]
public static Task<Person> Update([FromSource] int personId, [FromServices] MyDbContext dbContext, [Required] string name, int? age)
{
    var user = await dbContext.Persons.Find(personId);
    user.Name = name;
    if (age.HasValue) user.Age = age;
    await dbContext.SaveChanges();
    return user;
}
```

You can add scoped services in the constructor, or within the method signature as seen above.  You can specify which methods can run asynchronously, and which ones must run synchronously.

I would like to infer if the argument is optional via nullable reference types, but have not worked on that yet.

As you'll see, I have not polished it, but I do use it internally and hope to sometime add tests & documentation, and publish it as a nuget package.

There's also the GraphQL conventions project which also contains similar functionality - see https://github.com/graphql-dotnet/conventions - but I have not used it.

dnsfour:
Let me explain my motivation:
1) My colleagues are complaining about a lot of code that goes with graphql.
2) With optional parameters, there is no need to get the whole entity to update one field, and I think this is important for point 1)
3) Source objects become adapters to entities, which must be wrapped each time in the resolve method manually, while services and data can be manually or ActivatorUtils.CreateInstance() injected. So you can do without additional attributes.


```cs
    public class PersonData
    {
        public int Id { get; set; }
        public int Name { get; set; }
    }

    /// <summary>Query root object</summary>
    public class Query
    {
        private readonly PersonStore personStore;
        private readonly IServiceProvider sp;

        public Query(PersonStore personStore, IServiceProvider sp)
        {
            this.personStore = personStore;
            this.sp = sp;
        }

        public async Task<Person> Person(int id)
        {
            var personData = await personStore.Read(id);
            return ActivatorUtilities.CreateInstance<Person>(sp, personData);
        }
    }

    /// <summary>Source for ObjectGraphType<Person></summary>
    public class Person
    {
        private readonly PersonData person;
        private readonly IDataLoaderContext dataLoader;
        private readonly PersonStore personStore;

        public Person(PersonData data, IDataLoaderContext dataLoader, PersonStore personStore)
        {
            this.person = data;
            this.dataLoader = dataLoader;
            this.personStore = personStore;
        }

        public int Id { get; set; }

        public Task<Person> Relativies(int skip, int take) => dataLoader.LoadBatch(...).LoadKey(data.Id);
        public Task<Person> Parents(IResolveFieldContext ctx) => dataLoader.LoadBatch(...).LoadKey(data.Id);
    }
```

Shane32:
Interesting; I had not thought of that.  PR #1376 does not work this way, although it probably could be adapted to do so quite easily.  I not sure I follow how this design applies to changes to the database.  In other words, typically changes only occur in mutations, so wouldn't a separate object be required for mutations vs queries?  Perhaps you could share some more samples of how you anticipate this wires up.

dnsfour:
Right now i guess only to do parsing start of query
Example with another way to pass services and data into source:
```cs
	                //arrange
			var de = new DocumentExecuter();
			var spMock = new Mock<IServiceProvider>();
			var sp = spMock.Object;
			var schema = new Schema {
				Query = new AutoRegisteringObjectGraphType<Query>(true),
				Mutation = new AutoRegisteringObjectGraphType<Mutation>(true)
			};
			var query = "mutation { ... }";
			//act
			var result = await de.ExecuteAsync(_ => {
				_.Schema = schema;
				_.Query = query;
				_.Root = query.TrimStart().StartsWith("mutation") ? sp.GetRequiredService(typeof(Mutation))
					: query.TrimStart().StartsWith("query") ? sp.GetRequiredService(typeof(Query))
					: null;
			});
		}


		public class ServicesContext {
			private readonly IServiceProvider sp;
			public ServicesContext(IServiceProvider sp) => this.sp = sp;

			private PersonStore personStore;
			public PersonStore PersonStore => personStore ??= sp.GetRequiredService<PersonStore>();
                        //...
		}

		// source of ObjectGraphType<Mutation>
		public class Mutation {
			private readonly ServicesContext services;

			public Mutation(ServicesContext services) {
				this.services = services;
			}

			public async Task<Person> UpdatePerson(int id, Optional<string> name) {
				var personData = await services.PersonStore.Read(id);
				if (name != null) personData.Name = name;
                                await services.PersonStore.Write(personData);
				return new Person(services, personData);
			}
		}

		// source of ObjectGraphType<Person>
		public class Person {
			private readonly ServicesContext services;
			private readonly PersonData data;

			public Person(ServicesContext services, PersonData data) {
				this.services = services;
				this.data = data;
			}

			//...
		}
```

dnsfour:
And I also think that the new c#9 Record type should help in passing services and data:
```cs
return person with { Data = personData }
...
public record Person {
 public PersonData Data { get; init; }
 constructor(PersonStore store) {...}
}
```

joemcbride:
Just noting here that schema-first syntax allows you to write your GraphQL types exactly as described.  It uses the [AccessorFieldResolver](https://github.com/graphql-dotnet/graphql-dotnet/blob/a35b22cc0c9ca4c8387af5daaf71635b56008ee4/src/GraphQL/Reflection/AccessorFieldResolver.cs#L6-L33) internally to accomplish this.

https://graphql-dotnet.github.io/docs/getting-started/introduction#schema-first-nested-types

https://github.com/graphql-dotnet/graphql-dotnet/blob/a35b22cc0c9ca4c8387af5daaf71635b56008ee4/src/GraphQL/Utilities/TypeConfig.cs#L43-L55

https://github.com/graphql-dotnet/graphql-dotnet/blob/a35b22cc0c9ca4c8387af5daaf71635b56008ee4/src/GraphQL/Reflection/AccessorFieldResolver.cs#L6-L33

joemcbride:
Slightly updated for how schema-first would handle it:

```gql
type Person {
    Id: ID!
    name: String!
    relativies(skip: Int!, take: Int!): [Person!]!
    parents: [Person!]!
}

type Query {
    person(id: ID!): Person
}
```

```csharp
public class PersonData
{
    public int Id { get; set; }
    public int Name { get; set; }
}

/// <summary>Query root object</summary>
public class Query
{
    private readonly PersonStore personStore;

    public Query(PersonStore personStore)
    {
        this.personStore = personStore;
    }

    public async Task<Person> Person(int id)
    {
        return await personStore.Read(id);
    }
}

/// <summary>Source for ObjectGraphType<Person></summary>
public class Person
{
    private readonly IDataLoaderContext dataLoader;
    private readonly PersonStore personStore;

    public Person(IDataLoaderContext dataLoader, PersonStore personStore)
    {
        this.data = data;
        this.dataLoader = dataLoader;
        this.personStore = personStore;
    }

    // these are not required
    // public int Id(PersonData data) => data.Id;
    // public string Name(PersonData data) => data.Name;

    public Task<Person[]> Relativies(int skip, int take) => dataLoader.LoadBatch(...).LoadKey(data.Id);
    public Task<Person[]> Parents(IResolveFieldContext ctx) => dataLoader.LoadBatch(...).LoadKey(data.Id);
}
```

```csharp
var schema = Schema.For("..." _ =>
{
    _.Types.Include<Query>();
    _.Types.Include<Person>();
});

var json = await schema.ExecuteAsync(_ =>
{
    _.Query = "{ person(id: ""123"") { id name parents { name } } }";
    _. RequestServices = ...
});

Console.WriteLine(json);
```

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt.orig
```diff
@@ -0,0 +1,3130 @@
+namespace GraphQL
```
sungam3r, line 1: Please remove this file.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -82,11 +90,16 @@ public void auto_register_object_graph_type()
         {
             GraphTypeTypeRegistry.Register<Money, AutoRegisteringObjectGraphType<Money>>();
 
-            var type = new AutoRegisteringObjectGraphType<TestObject>(o => o.valuePair, o => o.someEnumerable);
+            var type = new AutoRegisteringObjectGraphType<TestObject>(
+                true,
+                o => o.valuePair,
+                o => o.someEnumerable,
+                o => o.someIntMethod2(default, default, default)
```
sungam3r, line 97: it looks a little strange

src/GraphQL/Resolvers/MethodResolver.cs
```diff
@@ -0,0 +1,85 @@
...
+
+namespace GraphQL.Resolvers
+{
+    /// <summary>
+    /// Experimental. Probably should be a part of <see cref="NameFieldResolver"/>
+    /// It has <see cref="CreateQueryArguments"/> for field registration
+    /// </summary>
+    internal class MethodResolver : IFieldResolver
+    {
+        public static readonly ConcurrentDictionary<MethodInfo, Func<object, object[], object>> cache = new ConcurrentDictionary<MethodInfo, Func<object, object[], object>>();
```
sungam3r, line 16: public?

src/GraphQL/TypeExtensions.cs
```diff
@@ -94,6 +103,8 @@ public static string GraphQLName(this Type type)
                 : typeName;
         }
 
+
+
```
sungam3r, line 107:
```suggestion
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -18,18 +20,34 @@ namespace GraphQL.Types
...
         /// </summary>
-        public AutoRegisteringObjectGraphType() : this(null) { }
+        public AutoRegisteringObjectGraphType() : this(false, null) { }
 
         /// <summary>
         /// Creates a GraphQL type from <typeparamref name="TSourceType"/> by specifying fields to exclude from registration.
         /// </summary>
-        /// <param name="excludedProperties"> Expressions for excluding fields, for example 'o => o.Age'. </param>
-        public AutoRegisteringObjectGraphType(params Expression<Func<TSourceType, object>>[] excludedProperties)
+        /// <param name="excludesMembers"> Expressions for excluding fields, for example 'o => o.Age' or 'o => o.Age(defaul, ...). </param>
```
sungam3r, line 35:
```suggestion
        /// <param name="excludesMembers"> Expressions for excluding fields, for example 'o => o.Age' or 'o => o.Age(default, ...). </param>
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -18,18 +20,34 @@ namespace GraphQL.Types
...
+        public AutoRegisteringObjectGraphType(params Expression<Func<TSourceType, object>>[] excludesMembers): this(false, excludesMembers)
         {
-            AutoRegisteringHelper.SetFields(this, GetRegisteredProperties(), excludedProperties);
+        }
+
+        /// <summary>
+        /// Creates a GraphQL type from <typeparamref name="TSourceType"/> by specifying fields to exclude from registration.
+        /// </summary>
+        /// <param name="registerMethods"> Register methods of <see cref="GetRegisteredMethods"/> </param>
+        /// <param name="excludesMembers"> Expressions for excluding fields, for example 'o => o.Age' or 'o => o.Age(defaul, ...). </param>
```
sungam3r, line 44:
```suggestion
        /// <param name="excludesMembers"> Expressions for excluding fields, for example 'o => o.Age' or 'o => o.Age(default, ...). </param>
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -38,6 +56,13 @@ protected virtual IEnumerable<PropertyInfo> GetRegisteredProperties()
                 .GetProperties(BindingFlags.Public | BindingFlags.Instance)
                 .Where(p => AutoRegisteringHelper.IsEnabledForRegister(p.PropertyType, true));
         }
+
+        protected virtual IEnumerable<MethodInfo> GetRegisteredMethods()
+        {
+            return typeof(TSourceType)
+                .GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly)
```
sungam3r, line 63: Why DeclaredOnly?

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -75,78 +101,102 @@ internal static class AutoRegisteringHelper
...
         }
 
-        internal static bool IsEnabledForRegister(Type propertyType, bool firstCall)
+        internal static bool IsEnabledForRegister(Type returnType, bool firstCall)
         {
-            if (propertyType == typeof(string)) return true;
+            if (returnType == typeof(string))
+                return true;
 
-            if (propertyType.IsValueType) return true; // TODO: requires discussion: Nullable<T>, enums, any struct
```
sungam3r, line 118: Please revert all code formatting changes. These changes are not relevant and only complicate the review.

sungam3r (changes requested):
> I did not understand why registration of methods was not implemented for AutoRegistraionObjectGraphType

Most likely because method registration is not supported for regular GraphTypes. If you are adding this function for AutoRegisteringObjectGraphType, then the first thing to ask yourself is whether it is worth adding it to the regular GraphType APIs.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
dnsfour: Agree, but powerful to exclude overloaded methods

src/GraphQL/Resolvers/MethodResolver.cs
dnsfour: Okay. I was not sure about the name and whether it should be separated from the NameFieldResolver.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
dnsfour: The graphql spec does not support inheritance and so there will be better schema match

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
dnsfour: done

dnsfour:
> first thing to ask yourself is whether it is worth adding it to the regular GraphType APIs.

Maybe I'm wrong, but I thought that AutoRegisteringObjectGraphType is not regular GraphType

dnsfour:
> Just noting here that schema-first syntax allows you to write your GraphQL types exactly as described. It uses the AccessorFieldResolver internally to accomplish this.

I really missed it.
1) Reflection is used to call methods. This is slower than my implementation, I can improve
2) It seems that the schema-first approach increases the level of entry into graphql-dotnet for line developers, since you need to know the graphql syntax well.
3)  Schema-first means more code.

src/GraphQL/Resolvers/MethodResolver.cs
dnsfour: > Okay. I was not sure about the name and whether it should be separated from the NameFieldResolver.

Or is it about cache? Fixed public #cache.

dnsfour:
Sorry for wasting time.
I realized this was a bad idea and it would be better if I do a couple of PRs:

- Improve NameFieldResolver to optinal pass source, ResolveFieldContext and arguments to methodsS
- Add new Optional<T> class so that you can distinguish between updatePerson (id: 1 name: null) and updatePerson (id: 1 dateOfBirth: null)
- Do breaking change for AutoRegisteringObjectGraphType: rename GetRegisteredProperties to GetRegisterMembers, add to constructor arguments registerProperties registerMethods and depricate existing constructors.

If I still fit into the concept of regular graphtypes.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: It is all about implementation details here, not about spec.

sungam3r:
> Maybe I'm wrong, but I thought that AutoRegisteringObjectGraphType is not regular GraphType

Yes but I talked about regular graph types and graphtype first syntax.

PR #1886, opened 2020-10-08
Publish merged PRs as nuget packages to GitHub Packages

Shane32:
This configures merged PRs to automatically publish nuget packages to GitHub Packages

This would be effective for both master and develop branches.  Version numbers will be different because `Directory.Build.props` contains a different version prefix for each.

- master branch: `3.0.0-preview.{build}`
- develop branch `4.0.0-preview.{build}`

Where `{build}` is incremented automatically with every execution of this GitHub Action.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
-    <VersionPrefix>3.0.0</VersionPrefix>
+    <VersionPrefix>3.0.0-preview</VersionPrefix>
     <PackageProjectUrl>https://github.com/graphql-dotnet/graphql-dotnet</PackageProjectUrl>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageIcon>logo.64x64.png</PackageIcon>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
-    <VersionPrefix>3.0.0-preview</VersionPrefix>
+    <VersionPrefix>3.1.0-preview</VersionPrefix>
     <PackageProjectUrl>https://github.com/graphql-dotnet/graphql-dotnet</PackageProjectUrl>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageIcon>logo.64x64.png</PackageIcon>
--- a/.github/workflows/build-artifacts-code.yml
+++ b/.github/workflows/build-artifacts-code.yml
@@ -1,0 +1,63 @@
+name: Build artifacts
+
+# ==== NOTE: do not rename this yml file or the run_number will be reset ====
+
+on:
+  push:
+    branches:
+      - master
+      - develop
+    paths:
+      - src/**
+
+env:
+  NODE_VERSION: '10.x'   # Node 10 LTS
+
+jobs:
+  pack:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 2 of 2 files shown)


sungam3r:
Do I understand correctly that you need to commit the same file to `develop` branch?

src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
-    <VersionPrefix>3.0.0</VersionPrefix>
+    <VersionPrefix>3.0.0-preview</VersionPrefix>
```
sungam3r, line 4: Why? We have already published 3.0.0 release and go on with next version, so either 3.0.1-preview or 3.1.0-preview .

Shane32: Sounds good

Shane32:
```suggestion
    <VersionPrefix>3.1.0-preview</VersionPrefix>
```

Shane32:
> Do I understand correctly that you need to commit the same file to `develop` branch?

Yes (although I've already created the file in that branch).

sungam3r:
> although I've already created the file in that branch

I thought so.

sungam3r (approved):
Let's try

Shane32:
Apparently the `VersionSuffix` tag appends a dash and the suffix so `4.0.0-preview` as the prefix and `2` as the suffix becomes `4.0.0-preview-2`.  I can probably fix this if desired.  Should I attempt to do so or is it fine?  Seems fine to me.  To change it, I think I would need to mess with the MSBuild process.

As it is now, the nuget package beta versions get generated with this format version number:

- master branch: 3.1.0-preview-{build}
- develop branch: 4.0.0-preview-{build}

Otherwise it seems to work perfectly.  See: https://github.com/orgs/graphql-dotnet/packages?repo_name=graphql-dotnet

sungam3r:
It is fine. Excellent. What is the feed to consume these packages?

Shane32:
The feed is `https://nuget.pkg.github.com/graphql-dotnet/index.json`

It is important to note that even though this is a public repository, authentication is required, in this form:
- Username: your GitHub username
- Password: one of your Personal Access Tokens (PAT), with the "read packages" permission

I executed this at a command prompt in order to link up my PC to GitHub Packages:

- `dotnet nuget add source https://nuget.pkg.github.com/graphql-dotnet/index.json -n GitHub -u (my GitHub Username) -p (my GitHub PAT)` (no parentheses)

It did not change any files in the repo, so maybe it stores in the registry; I'm not sure.  Then I added the package to my project:

- `dotnet add package GraphQL --version 3.1.0-preview-2`

See:
- https://docs.github.com/en/free-pro-team@latest/packages/using-github-packages-with-your-projects-ecosystem/configuring-dotnet-cli-for-use-with-github-packages
- https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token
- https://github.com/graphql-dotnet/graphql-dotnet/packages/433360

We can easily configure any GitHub Action workflows to authenticate to GitHub using the GITHUB_SECRET token provided by the workflow, so if they reference preview builds, it can still restore the dependencies and build the project.  We can also be sure that when building releases, it does not authenticate to GitHub, so we can be sure the released package is not dependent on a preview build.

If needed, we could create a GitHub account with no special privileges, and create a PAT under that account with only the "read package" permission if we needed to put the PAT directly into the repo.

sungam3r:
Thanks for the info. I will try to configure feed.

> so maybe it stores in the registry; I'm not sure.

It is under your user directory.
https://github.com/NuGet/docs.microsoft.com-nuget/blob/master/docs/consume-packages/configuring-nuget-behavior.md#config-file-locations-and-uses

Issue #1885, opened 2020-10-07
Allow to override built-in scalars


randyridge:
Upgrading from 2.x to 3.x our schema had a Uri type...  Is there a handy way to unregister the built in type?

sungam3r:
I expected that sooner or later someone would make this request. I have long disliked that there is a conflict with custom scalars. I'll see what i can do.

sungam3r:
@Shane32 What do you think about `GraphTypesLookup` and `GraphTypeTypeRegistry`? `GraphTypeTypeRegistry` alllows to override type, but I can't say that for `GraphTypesLookup`.

Shane32:
Without looking at the code, going strictly from memory, perhaps we can provide a workaround in the form of removing the `UriGraphType` entry from the `GraphTypesLookup` before the schema gets initialized.  If there's no method to do so, perhaps we can add one for version 3.1.  Long term answer, maybe `GraphTypesLookup` should not add ANY ".NET custom scalars" automatically, but only add them in as needed just like any other graph type that is referenced by another graph type.  We can still leave the `GraphTypeTypeRegistry` links in place, but those can be overridden if needed already.

Side note: I don't see a registration within the `GraphTypeTypeRegistry` that maps `Uri` to `UriGraphType`.....

randyridge:
I spent about an hour trying to make it happen, I tried everything I could think of, modifying the private types dictionary through a couple levels of reflection, and finally just decided to change our Uri type name...  yolo

Shane32:
In order to do it now you will need to:
- Duplicate the static `GraphTypesLookup.Create` method
- After the first line of code that creates the class (`var lookup =`), before the rest of the code,
  - Use reflection to obtain the `_types` dictionary
  - Remove the entry corresponding to the `UriGraphType`
- Duplicate `Schema.CreateTypesLookup` in your own schema class
  - Have it call the new method you created above instead of the standard one
- In your schema class, either:
  1. Use reflection to change the private `_lookup` to a new instance of `Lazy<GraphTypesLookup>` which points to the new `CreateTypesLookup` function you created in the prior step; or
  2.  Create a new `_lookup` variable and override the following properties so they point to the new `_lookup` variable:
      - `Initialized`
      - `SchemaMetaFieldType`
      - `TypeMetaFieldType`
      - `TypeNameMetaFieldType`
      - `FindType`
      - `Dispose` (should call `base.Dispose` also)

[...]

PR #1039, opened 2019-03-04
remove the requiremnent for connection types to be in the container

SimonCropp:
fixes #451 and possibly #352

PR #1887, opened 2020-10-08
Do not register unused graph types

Shane32:
This PR prevents `GraphTypesLookup` from registering custom .Net scalar graph types (e.g. `DateTimeOffsetGraphType`) that are not used by the schema.  See #1885 

Also allows custom `Schema` implementations necessary hooks to edit the list of built-in scalar graph types during schema initialization.  Not actually necessary to solve #1885 since the `UriGraphType` would never have been registered if it was not used, so there would be no conflict.

This currently targets version 3.1.  Should it target 4.0?  The binary is backwards compatible, but generated schemas will not include unused types, so introspection results will differ (even though the missing types are unused by the schema).


randyridge:
To be explicit, what I hit was that going from 2.x to 3.x -- I had my own UriType, the base library added it's own UriType (both having name Uri).  Scanning finds both with the same name and fires exception.

sungam3r:
Relates to #1423 .

These changes should be carefully reviewed prior to merging.

sungam3r:
I skimmed through the edits. Yet it seems to me that this issue should proceed differently. I'll be back with an answer later.

Shane32:
Also see https://github.com/graphql/graphql-spec/issues/222 and https://github.com/graphql/graphql-spec/issues/648

Shane32:
> Faced with a similar problem in my graphql client generator, I came to the conclusion that all custom scalars should be connected externally. The base library should only provide 5 built-in scalars.

_Originally posted by @sungam3r in https://github.com/graphql-dotnet/graphql-dotnet/issue_comments/705484550_

That's EXACTLY what this PR does.  Only the 5 basic scalars are registered with a simple schema.  And just like any graph type, additional scalars referenced by the graph are automatically added to the schema.  The only difference here is that we have special code to prevent the extra provided scalars (like `UriGraphType`) from having to be registered with the service provider.

There might be another way to go about it, but I think this is the right answer.  We should not include these extra scalars in the schema by default.  Nor should we remove them from the library.  And there's no reason to force users to register them with their DI provider.  Finally, it's not a breaking change -- in the sense that all existing code will work perfectly with this update.

sungam3r:
> We should not include these extra scalars in the schema by default. Nor should we remove them from the library.

Well, actually, I think that moving all custom scalars into a separate package is the right way.

> And there's no reason to force users to register them with their DI provider.

The reason is to make the library free of implementation assumptions and not create conflicts with custom scalars. Custom scalars are a good example of when you can move a piece of code into a separate package without negative consequences.

It will be breaking change for clients, yes. I think this is a long-awaited technical debt. The emergence of individual packages is a normal evolutionary process.

Shane32:
I'm going to simplify this PR a bit, and get rid of the other changes...

Shane32:
Let's assume for the moment that in 4.0 we have custom scalars in a separate nuget package.  Is there any reason not to include this change in 3.1?  I think we should.

I have removed the API changes, so please re-review.

sungam3r:
Give me 1-2 days.

sungam3r:
It will take me longer. I had to deal with other bugfixes.

sungam3r:
Nice workaround. This is an understandable little change. However, I would like to simplify a few things. I'll be back with another PR.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -184,6 +189,27 @@ public IEnumerable<IGraphType> All()
             }
         }
 
+        private void AddType(Type builtInGraphType)
```
joemcbride, line 192: I would suggest perhaps a more distinct name like `AddBuiltInType`?  Too closely named to the other `AddType` methods I think.

I like this direction.  👍🏼

Shane32: Good idea! - done. 👍 

Let me know if you have any other comments.  It's a pretty simple enhancement, and will, I believe, work great all the way 'round.

Shane32:
> Nice workaround. This is an understandable little change. However, I would like to simplify a few things. I'll be back with another PR.

I copied from the `AddType<T>()` and `AddType<T>(context)` methods to create `AddBuiltInType()`.  If you think I should eliminate duplicated code between the functions, I can add a `CreateTypeCollectionContext()` method.

Since these new methods and fields are all private, we can refactor at will later also.

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: I already intervened in this PR and changed a lot. This method will no longer exist. I will update the PR and describe the changes in the next hour.

sungam3r: Resolving conflicts 😞

sungam3r:
Will update PR soon.

src/GraphQL/Types/GraphType.cs
```diff
@@ -99,7 +99,7 @@ public override bool Equals(object obj)
     }
 
     /// <summary>
-    /// This sucks, find a better way
+    /// TODO: This sucks, find a better way
```
sungam3r, line 102: I understood this remark, but I don’t know how best to proceed. Maybe later.

sungam3r:
What was done:
1. Move introspection result into json file. I'm terribly tired of editing it in the source without the ability to highlight syntax and collapse blocks.
2. Further removal of unused types, **even built-in ones**. See discussions from #1423 and https://github.com/graphql/graphql-spec/issues/648
3. Rework of GraphTypesLookup:
  a) Move built-in graph types and introspetion types into static.
  b) Eliminate allocations of graph types in `Activator.CreateInstance()`.
  c) Eliminate allocations of `TypeCollectionContext`.


I think that #1423 is no longer needed.

sungam3r:
One consideration against static dictionaries. Types can contain metadata. Then a conflict can arise if the user adds different metadata into the built-in types.

sungam3r:
Switched to instance fields. This will return some allocations, but still there will be fewer of them than before this PR.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -8,10 +8,60 @@
...
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// Provides lookup for all schema types and has algorithms for discovering them.
+    /// </summary>
     public class GraphTypesLookup
     {
-        private readonly IDictionary<string, IGraphType> _types = new Dictionary<string, IGraphType>();
+        // Introspection types http://spec.graphql.org/June2018/#sec-Schema-Introspection
+        private readonly Dictionary<Type, IGraphType> _introspectionTypes = new IGraphType[]
```
sungam3r, line 17: All these types are always added to the schema. Each type is created only **once** within schema instance. Before these changes types were created several times. It is also true for built-in scalars.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -22,41 +72,24 @@ public GraphTypesLookup(INameConverter nameConverter)
...
-            AddType<UIntGraphType>();
-            AddType<LongGraphType>();
-            AddType<BigIntGraphType>();
-            AddType<ULongGraphType>();
-            AddType<ByteGraphType>();
-            AddType<SByteGraphType>();
-
-            // Add introspection types.  Note that introspection types rely on the
-            //   CamelCaseNameConverter, as some fields are defined in pascal case - e.g. Field(x => x.Name)
+            _context = new TypeCollectionContext(
```
sungam3r, line 75: `TypeCollectionContext` is just like a proxy to resolve/add methods pair. It can be shared across all introspection types whilst `AddType` calls.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -393,18 +423,19 @@ private void AddTypeIfNotRegistered(Type type, TypeCollectionContext context)
                 if (namedType == typeof(PageInfoType))
                 {
                     AddType(new PageInfoType(), context);
-                    return;
                 }
-                if (namedType.IsGenericType)
+                else if (namedType.IsGenericType && (namedType.ImplementsGenericType(typeof(EdgeType<>)) || namedType.ImplementsGenericType(typeof(ConnectionType<,>))))
```
sungam3r, line 427: Perhaps these types should be taken out of here. I've been thinking about this for a long time. Not in thin PR though.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -510,14 +541,23 @@ private IGraphType ConvertTypeReference(INamedType parentType, IGraphType type)
...
             var reference = type as GraphQLTypeReference;
-            var result = reference == null ? type : this[reference.TypeName];
+            if (reference != null)
+            {
+                type = this[reference.TypeName];
+                if (type == null)
+                {
+                    type = _builtInScalars.Values.FirstOrDefault(t => t.Name == reference.TypeName);
+                    if (type != null)
+                        this[type.Name] = type;
```
sungam3r, line 551: Example - discovering `ID` built-in type during schema first for 
```graphql
type Post {
                    id: ID!
                    title: String!
                }

                type Query {
                    post(id: ID!): Post
                }
```

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: It would be better if upon finding a generic type (e.g. `EdgeType<MyModel>`), it uses reflection to retrieve the generic version of the type (e.g. `EdgeType<>`) and then checks `_builtInCustomScalars` for the generic type.  If matched, it will use `Activator.CreateInstance` to create an instance.  Fine as-is though.

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: I don't use schema-first so I'll trust you here :-)

joemcbride: Wouldn't this also need to check the `_builtInCustomScalars`?  Seems like that would be the case everywhere, it would need to check both wouldn't it?

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: Better to make new issue and link it to #1039

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: You are right. Fixed.

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
```diff
@@ -430,7 +430,7 @@ public void can_execute_complex_schema()
                 }
                 type Blog {
                     title: String!
-                    post(id: ID!): Post
+                    post(id: ID!, unused: Long!): Post
```
sungam3r, line 433: One of custom built-in scalars.

sungam3r (approved):
In fact, I approve my own changes 😄 .

sungam3r:
This is the last issue in 3.1 milestone. I am very glad that this work has been done. I've wanted to get rid of unused scalars for a very long time, but I've been putting it off all the time. Thanks for the review. Merging.

PR #1891, opened 2020-10-09
fix check for interface covariance

### LINKED ISSUE #1889: The interface match check has break covariant interface returns
## Description

While migrating to v3, I've noticed a regression with the `IsValidInterfaceFor` method in the file `src/GraphQL/GraphQLExtensions.cs` (tracked to this specific commit: https://github.com/graphql-dotnet/graphql-dotnet/commit/9a7410e199de887d8ec436892ce642ae0910f19d#diff-4586aaacee7063d6089049fcfdae01a3R111).

When you have a GraphType that implement an interface field by returning a type that inherit from the base type of this field, the check in `if (found.Type != field.Type)` is too strict and throws an error.

### Steps to reproduce

Here is a dummy exemple:

```cs
class R {}
class SpecializedR : R {}

class A {
  public R methodUsedToGetRValue() {
    // returns a R object
  }
}
class SpecializedA : A {
  public SpecializedR methodUsedToGetSpecializedRValue() {
    // returns a SpecializedR object
  }
}

public class RGraphInterface : InterfaceGraphType<R> {
  public RGraphInterface() {
    Name = "RInterface";
  }
}

public class SpecializedRGraphType : ObjectGraphType<SpecializedR> {
  public SpecializedRGraphType() {
    Name = "SpecializedRGraphType";
    Interface<RGraphInterface>();
  }
}

public class AGraphInterface : InterfaceGraphType<A> {
  public AGraphInterface() {
    Name = "AInterface";

    Field<NonNullGraphType<RGraphInterface>>(
        "R",
        resolve: ctx => ctx.Source.methodUsedToGetRValue()
    );
  }
}

public class SpecializedAGraphInterface : ObjectGraphType<SpecializedA> {
  public SpecializedAGraphInterface() {
    Name = "SpecializedAGraphInterface";
    Interface<AGraphInterface>();

    Field<NonNullGraphType<SpecializedRGraphType>>(
        "R",
        resolve: ctx => ctx.Source.methodUsedToGetSpecializedRValue()
    );
  }
}
```

### Expected result

The contract specified by the interface is fulfilled since we do return a `R` object (just more specialized, but that should not be an issue as specified in the graphql spec here : https://github.com/graphql/graphql-spec/pull/123).

### Potential fix

The current implementation that just check the types with `==` is not enough, interfaces should be checked to ensure we're not returning a more specified type (thus, making the `==` fails while the contract being respected). Maybe something like `if(!field.Type.IsAssignableFrom(found.Type))` ?

EDIT: After some tests, the potential fix suggested is still not enough, if we use Composite types like `NonNullGraphType` the `IsAssignableFrom` would still fail.

sungam3r:
fixes #1889

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1889.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1889.cs
@@ -1,0 +1,95 @@
+using GraphQL.Types;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    public class Bug1889WithCovariant : QueryTestBase<CovariantSchema>
+    {
+        [Fact]
+        public void supports_covariant_schemas()
+        {
+            string query = @"query { a { r { value } } }";
+            string expected = @"{ ""a"": { ""r"": { ""value"": ""spec"" } } }";
+
+            AssertQuerySuccess(query, expected);
+        }
+    }
+
+    public class CovariantSchema : Schema
+    {
+        public CovariantSchema()
+        {
+            Query = new CovariantQuery();
+        }
+    }
+
+    public class CovariantQuery : ObjectGraphType
+    {
+        public CovariantQuery()
+        {
+            Name = "CovariantQuery";
+
+            Field<NonNullGraphType<SpecializedAGraphType>>("A", resolve: ctx => new SpecializedA());
+        }
+    }
+
+    public class R
+    {
+        public string Value = "base";
+    }
+
+    public class SpecializedR : R
+    {
+        public new string Value = "spec";
+    }
+
+    public class A
+    {
+        public R methodUsedToGetRValue() => new R();
+    }
+
+    public class SpecializedA : A
+    {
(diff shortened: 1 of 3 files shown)


Elyhaka:
I'm going to try this against my project asap :)

Elyhaka:
I can confirm this PR fixes the whole #1889 issue, including the union part that I couldn't reproduce in a test yet!

sungam3r:
@Shane32 I targeted this PR to `master` since there is no breaking changes. Right?

sungam3r:
By the way, as I talked earlier, it should be 3.0.1 (no new APIs), but I'm OK with 3.1.0 since sooner or later they will appear.

Shane32 (approved):

sungam3r:
@Shane32 I see only one published package in GitHub UI. It seems that preview-2 was "ovewritten".

Shane32:
It's impossible to remove packages from GitHub Packages for public repositories.  It just seems that the UI automatically hides old versions:

image

sungam3r:
OK

PR #1892, opened 2020-10-10
Bump deps

sungam3r:


src/GraphQL/GraphQL.csproj
```diff
@@ -13,7 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="GraphQL-Parser" Version="5.3.0" />
+    <PackageReference Include="GraphQL-Parser" Version="5.3.3" />
```
sungam3r, line 16: The main motivation was updating the parser, which began to allocate much less memory.

Shane32: With a "side effect" of fixing a bug where syntax error messages show inaccurate locations

sungam3r: Of course

PR #1893, opened 2020-10-10
Fix for aliased list arguments

### LINKED ISSUE #1882: Aliased field in mutation input type does not get set when argument is a list
## Description

Field values do not get set from an mutation's input argument when the field has an alias and the input argument is a _list_ of input objects.

This appears to be in a similar vein to bugs #850 and #1358, but this behavior remains despite both those bugs being fixed. As far as I have observed, field aliases work if the mutation argument is a single value (not a list) - I have only observed this problem affecting arguments that are a list.

### Steps to reproduce

See unit test at PR #1881.

Configure a field with an alias in `MyEntityInputType`:
```C#
public class MyEntityInputType : InputObjectGraphType<MyEntity>
{
	public MyEntityInputType() { Field("fieldAlias", x => x.BField); }
}

public class MyEntity
{
	public int BField { get; set; }
}
```

Configure a mutation like:
```C#
Field<BooleanGraphType>()
	.Name("mutateMyEntity")
	.Argument<ListGraphType<MyEntityInputType>>("multipleEntities")
	.Resolve(context => {
		var es = context.GetArgument<List<MyEntity>>("multipleEntities");
		es[0].BField.ShouldBe(4);
		return true;
	});
```

Execute a mutation query like the following:
```
mutation {
	mutateMyEntity( multipleEntities: [{ fieldAlias: 4 }] )
}
```

### Expected result

The `es[0].BField.ShouldBe(4)` assertion succeeds. 

### Actual result

The `BField` value that comes through is 0 and the `...ShouldBe(4)` assertion fails.

### Environment

graphql-dotnet version 3.0.0

sungam3r:
fixes #1882

### DIFF
--- a/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
+++ b/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
@@ -52,7 +52,7 @@ namespace GraphQL
                 return true;
             }
 
-            result = arg.GetPropertyValue(argumentType);
+            result = arg.GetPropertyValue(argumentType, context.FieldDefinition?.Arguments?.Find(argumentName)?.ResolvedType);
             return true;
         }
 
--- a/src/GraphQL.Tests/Bugs/BugAliasedInputFieldInList.cs
+++ b/src/GraphQL.Tests/Bugs/BugAliasedInputFieldInList.cs
@@ -1,0 +1,70 @@
+using GraphQL.Types;
+using System.Collections.Generic;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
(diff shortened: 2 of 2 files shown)


src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -52,7 +52,7 @@ private static bool TryGetArgument(this IResolveFieldContext context, Type argum
                 return true;
             }
 
-            result = arg.GetPropertyValue(argumentType);
+            result = arg.GetPropertyValue(argumentType, context.FieldDefinition?.Arguments?.Find(argumentName)?.ResolvedType);
```
sungam3r, line 55: It was pretty straightforward.

Shane32 (approved):

Issue #1831, opened 2020-08-31
Is it possible to use ListGraphType<CustomType> inside of InputObjectGraphTypes?


sellingsolutions:
I've spen the last 3 hours trying to mimic https://gist.github.com/DanielRobinsonSoftware/9ef4aa2834f2c912b5a3d4ef63ba6188 as well as various other workarounds on stackoverflow. 

Afaik, the documentation doesn't mention how to use input arrays, I've so far had no issues with ListGraphType<StringGraphType> but I can't seem to understand how to use ListGraphType<MyOwnGraphType> inside of InputObjectGraphType.

This is all I want to do, to be able to pass in arrays with custom objects. What am I missing?
```
public class Input
{
        [JsonProperty("id")]
        public string Id { get; set; }
        
        [JsonProperty("rows")]
        public IEnumerable<Row> Rows { get; set; }
}
public class InputType : InputObjectGraphType<Input>
{
      public InputType() 
      {
          Field(o => o.Id, true, typeof(StringGraphType));
          Field(o => o.Rows, true, typeof(ListGraphType<RowType>));
      }
}
public class RowType : InputObjectGraphType<Row>
{
      public RowType() 
      {
          Field(o => o.Name, true, typeof(StringGraphType));
      }
}
```

**The exception that's being thrown**
```
[...]

PR #1896, opened 2020-10-11
Prevent variable parsing errors from masking validation errors

Shane32:
Patches #1831

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -124,7 +124,7 @@ namespace GraphQL
                         options.MaxParallelExecutionCount,
                         options.RequestServices);
                 }
-                catch
+                catch (InvalidVariableError)
                 {
                     // error parsing variables
                     // attempt to run AfterValidationAsync with null for the 'ExecutionContext.Variables' property
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -107,20 +107,70 @@ namespace GraphQL
                         _complexityAnalyzer.Validate(document, options.ComplexityConfiguration);
                 }
 
-                context = BuildExecutionContext(
-                    options.Schema,
-                    options.Root,
-                    document,
-                    operation,
-                    options.Inputs,
-                    options.UserContext,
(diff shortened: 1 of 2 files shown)


Shane32:
With this change, `BuildExecutionContext` sets `Variables` to `null` when `inputs` is `null`.  Previously, when `inputs` was `null`, `Variables` would be initialized as usual, assuming `null` for each individual variable.  To retain this behavior in that scenario, `Inputs.Empty` will be passed to `BuildExecutionContext` rather than `null`.

Shane32:
For anyone that does not utilize `AfterValidationAsync`, or for people that do not access `ExecutionContext.Variables` from within `AfterValidationAsync`, this PR will entirely fix the bug described in #1831 .

However, if a user attempts to access `ExecutionContext.Variables` from within `AfterValidationAsync` when there is a problem parsing the passed variables, behavior will revert to the former behavior with the only side effect that some code within `AfterValidationAsync` has already executed (up to the point of the `NullReferenceException`).

Shane32:
We can issue a more comprehensive change for 4.0, such as using `ValidationContext` instead of `ExecutionContext` within `AfterValidationAsync`.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -107,20 +107,70 @@ public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
...
+                        options.UserContext,
+                        options.CancellationToken,
+                        metrics,
+                        options.Listeners,
+                        options.ThrowOnUnhandledException,
+                        options.UnhandledExceptionDelegate,
+                        options.MaxParallelExecutionCount,
+                        options.RequestServices);
+                }
+                catch
```
sungam3r, line 127: Catch should work with more specific exception.

sungam3r:
> with the only side effect that some code within AfterValidationAsync has already executed (up to the point of the NullReferenceException).

Yes, there is some potential for side effects for "specially prepared" queries.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: Should be fine.

```suggestion
                catch (InvalidVariableError)
```

sungam3r (approved):

PR #1899, opened 2020-10-11
Create codeql-analysis.yml

sungam3r:
I noticed this function in [GitHub UI](https://github.com/graphql-dotnet/graphql-dotnet/security/code-scanning) and decided to try it. This is the file that was created by default. @Shane32 what do you think?

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '0 15 * * 2'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Override automatic language detection by changing the below list
+        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
+        language: ['csharp', 'javascript']
+        # Learn more...
+        # https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#overriding-automatic-language-detection
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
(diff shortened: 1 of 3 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Override automatic language detection by changing the below list
+        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
+        language: ['csharp', 'javascript']
```
sungam3r, line 27: I would remove `javascript`.

sungam3r:
Well, nothing found https://github.com/graphql-dotnet/graphql-dotnet/security/code-scanning , nothing interest 😞 .

Shane32:
Similar to dependabot, I don’t think it’s going to be very useful for this .Net Core project, as it seems to be focused on security alerts, which are more applicable to node.js and other languages. It probably doesn’t hurt and if you want to try it, we can certainly do so.

Shane32:
It’s possible we need to merge it to see any data because it doesn’t have results from the base branch. See here:

https://github.com/graphql-dotnet/graphql-dotnet/pull/1899/checks?check_run_id=1237906472

sungam3r:
I will not merge it yet and, if possible, will read the documentation a little more.

README.md
```diff
@@ -2,8 +2,9 @@
 
 [![Join the chat at https://gitter.im/graphql-dotnet/graphql-dotnet](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/graphql-dotnet/graphql-dotnet?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-[![Build status](https://github.com/graphql-dotnet/graphql-dotnet/workflows/Build%20artifacts/badge.svg)](https://github.com/graphql-dotnet/graphql-dotnet/actions)
-[![Build status](https://github.com/graphql-dotnet/graphql-dotnet/workflows/Publish%20code/badge.svg)](https://github.com/graphql-dotnet/graphql-dotnet/actions)
+[![Build status](https://github.com/graphql-dotnet/graphql-dotnet/workflows/Build%20artifacts/badge.svg)](https://github.com/graphql-dotnet/graphql-dotnet/actions?query=workflow%3A%22%22Build+artifacts%22%22)
```
sungam3r, line 5: @Shane32 Do you remember we failed to make the url because of the quotes? I accidentally discovered that if I double-quote (%22%22) everything works.

Shane32: Nice

sungam3r: Alas. It turned out that this only worked when clicked from the vs/source code. If you click on the link from GitHub readme, it still doesn't work.

sungam3r:
> It’s possible we need to merge it to see any data because it doesn’t have results from the base branch.

Let's merge and see what happens. I will remove this action if it is useless.

PR #1900, opened 2020-10-11
Add dependabot

sungam3r:
https://dependabot.com/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+- package-ecosystem: nuget
+  directory: "/"
+  schedule:
+    interval: daily
+    time: '02:00'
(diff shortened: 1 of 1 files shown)


Shane32:
I know this is not a popular philosophy, but to maximize compatibility, I think library dependencies should be updated on an as-required basis (e.g. oldest version available that is binary-compatible with the newest available version), rather than an as-available basis (always newest compatible version).  So I wouldn't use dependabot, or update nuget package versions as new ones become available.  That's just me...

(Applications & library test code, on the other hand, I would upgraded to the newest available dependencies.)

sungam3r:
That's right, only dependabot won't automatically update anything. It just makes the PR, and the decision is made by us (merge/decline). This automates manual labor. Remember our parser fixes and bumping dependencies.

PR #1903, opened 2020-10-11
Add GitHub Action build checks

Shane32:


.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,60 @@
...
+    paths: 
+      - src/**
+
+jobs:
+  ubuntu:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+      - name: Use .NET Core 2.1 LTS SDK
```
sungam3r, line 17: Why do you need a separate 2.1 SDK installation step if all tests are targeted at 3.1?

.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Use .NET Core 3.1 LTS SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '3.1.x'
+      - name: Install dependencies
+        working-directory: src
+        run: dotnet restore
+      - name: Build solution (release config)
+        working-directory: src
+        run: dotnet build --no-restore -c Release -p:NoWarn=CS1591
```
sungam3r, line 30: Why is the packages restore taken as a separate step? So is it preferable to do instead of ` run: dotnet build -c Release -p:NoWarn=CS1591`?

sungam3r: Ah! Because of second build step with Debug configuration I think.

.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Build solution (debug config)
+        working-directory: src
+        run: dotnet build --no-restore -p:NoWarn=CS1591
```
sungam3r, line 33:
```suggestion
        run: dotnet build --no-restore -c Debug -p:NoWarn=CS1591
```

Let's point that explicitly.

.github/workflows/pr-tests-code.yml
sungam3r: And dotnet test also ).

.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Test solution (debug config)
+        working-directory: src
+        run: dotnet test --no-restore
+      
+  windows:
+    runs-on: windows-2019
```
sungam3r, line 39: why not windows-latest ?

.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+      - name: Use .NET Core 2.1 LTS SDK
```
sungam3r, line 44: the same question

.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,60 @@
...
+
+on:
+  pull_request:
+    branches: 
+      - master
+      - develop
+    paths: 
+      - src/**
+
+jobs:
```
sungam3r, line 11: It seems to me that both steps are equivalent and can be combined into one with `matrix` tag like in this [example](https://github.com/adamralph/minver/blob/master/.github/workflows/ci.yml#L15)

.github/workflows/pr-tests-docs.yml
```diff
@@ -0,0 +1,33 @@
+name: Run documentation tests
```
sungam3r, line 1: I still don't know how the documentation is generated. I'm just wondering what is meant by the word `tests` here? Just checking that the specified commands fail/success? At the same time, the documentation is not published anywhere?

.github/workflows/pr-tests-code.yml
Shane32: I forgot; I already fixed up 4.0 so it passes .Net Core 2.1 tests, but not 3.x.  I should take 2.1 testing out of this PR.

.github/workflows/pr-tests-code.yml
Shane32: We could just write `dotnet test` and ignore all these prior steps.  But it's easier to review a workflow run when the individual tasks are separated.  I also added an extra step to build it twice - once for release and once for debug.  Debug is used for tests.  It would really only matter if there was compile-time conditionals in the code.  But as we are not billed for workflow run time (and it really doesn't take long to build), I included that step.

.github/workflows/pr-tests-code.yml
Shane32: For test reliability.  Now that `actions/setup-dotnet` is compatible with prepping multiple versions of the .NET SDK (so I removed the hack that was in here), I'd feel better about using windows-latest.

Shane32:
```suggestion
    runs-on: windows-latest
```

.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,60 @@
...
+  ubuntu:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+      - name: Use .NET Core 2.1 LTS SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '2.1.x'
```
Shane32, line 20:
```suggestion
```

.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,56 @@
...
+  windows:
+    runs-on: windows-latest
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+      - name: Use .NET Core 2.1 LTS SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '2.1.x'
```
Shane32, line 43:
```suggestion
```

.github/workflows/pr-tests-docs.yml
Shane32: Right now, it just checks that the documentation is capable of being built.  There are no 'tests', per se.  You are correct that the documentation does not get published anywhere in this workflow.

.github/workflows/pr-tests-code.yml
Shane32: I think there was a difference previously when `actions/setup-dotnet` would not load multiple versions of the .NET SDK simultaneously, and I had a hack in here to fix that.  The script could probably be consolidated with the matrix tag now.

Shane32: There was one difference: I didn't bother having the windows run build the release version.  I'm sure if it builds on ubuntu, it will build on windows.  And windows is so much slower to run workflows.  So I just added a conditional.

.github/workflows/pr-tests-code.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v2
+      - name: Use .NET Core 3.1 LTS SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '3.1.x'
+      - name: Install dependencies
+        working-directory: src
+        run: dotnet restore
+      - name: Build solution (release config)
+        if: ${{ matrix.os == 'ubuntu-latest' }}
```
Shane32, line 29:
```suggestion
        if: ${{ startsWith(matrix.os, 'ubuntu') }}
```

.github/workflows/pr-tests-docs.yml
sungam3r: Ok

.github/workflows/pr-tests-code.yml
sungam3r: How much slower?

Shane32: Try it and see. It’s like 2x-3x slower

sungam3r (approved):

PR #1906, opened 2020-10-12
Add ISchema.Comparer

### LINKED ISSUE #1905: Field ordering in introspection queries
Hi, I see that as of #594 the fields of a type in a schema are always ordered alphabetically. This is somewhat confusing with a schema-first approach to a server as the fields are all out of order from their originally shared schema. Could we please make this behaviour optional? It would make sense to me to control ordering by providing the user a setter for an `IComparer<FieldType>` which defaults to the original ordering and having a "ByName" implementation provided OOTB to smooth the common case.

Or at the very least can the behaviour be opt-out if not opt-in?

sungam3r:
fixes #1905

Also adds a bunch of xml comments.

@Shane32 can we afford to make this change in 3.1 or should we wait for 4.0.0? Formally, there is one backward incompatible change here. I have extended the `ISchema` interface by adding a new `Comparer` property.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -52,6 +52,7 @@ namespace GraphQL
 
             options.Schema.NameConverter = options.NameConverter;
             options.Schema.Filter = options.SchemaFilter;
+            options.Schema.Comparer = options.SchemaComparer;
 
             ExecutionResult result = null;
             ExecutionContext context = null;
--- a/src/GraphQL/Introspection/__Type.cs
+++ b/src/GraphQL/Introspection/__Type.cs
@@ -43,13 +43,15 @@ namespace GraphQL.Introspection
                         var includeDeprecated = context.GetArgument<bool>("includeDeprecated");
                         var type = context.Source as IComplexGraphType;
                         var fields = !includeDeprecated
-                            ? type?.Fields.Where(f => string.IsNullOrWhiteSpace(f.DeprecationReason))
-                            : type?.Fields;
+                            ? type.Fields.Where(f => string.IsNullOrWhiteSpace(f.DeprecationReason))
+                            : type.Fields;
 
                         fields ??= Enumerable.Empty<FieldType>();
-                        fields = await fields.WhereAsync(f => context.Schema.Filter.AllowField(context.Source as IGraphType, f)).ConfigureAwait(false);
-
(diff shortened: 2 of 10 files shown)


src/GraphQL/Introspection/__Type.cs
```diff
@@ -43,13 +43,15 @@ public __Type()
                         var includeDeprecated = context.GetArgument<bool>("includeDeprecated");
                         var type = context.Source as IComplexGraphType;
                         var fields = !includeDeprecated
-                            ? type?.Fields.Where(f => string.IsNullOrWhiteSpace(f.DeprecationReason))
```
sungam3r, line 46: `type` always non null here because both `IObjectGraphType` and `IInterfaceGraphType` inherit `IComplexGraphType`

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -52,6 +52,7 @@ public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
 
             options.Schema.NameConverter = options.NameConverter;
             options.Schema.Filter = options.SchemaFilter;
+            options.Schema.Comparer = options.SchemaComparer;
```
Shane32, line 55: This code does not belong here or in `ExecutionOptions`, but rather should be set by the Schema constructor.  See #1589 .

Shane32 (changes requested):

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: I remember that well. This change is made on purpose to maintain the current style. In version 4, we can change everything.

Shane32:
> @Shane32 can we afford to make this change in 3.1 or should we wait for 4.0.0? Formally, there is one backward incompatible change here. I have extended the `ISchema` interface by adding a new `Comparer` property.

I thought the goal for 3.x was to maintain binary compatibility.  This is neither binary compatible, nor is it code-compatible (e.g. if someone implemented their own `ISchema` it would break).  I'm okay with a different goal if it's stated and we follow those same guidelines for other PRs.

In this case, this change needs to be added to the migration guide that this property must be implemented for `ISchema`.

sungam3r:
So, should I change the base branch to develop ?

sungam3r:
I myself am more inclined to this.

sungam3r:
Closed in favor of #1907

PR #1907, opened 2020-10-12
Add ISchema.Comparer

### LINKED ISSUE #1905: Field ordering in introspection queries

sungam3r:
fixes #1905 

Also adds a bunch of xml comments.


Shane32:
I like the PR....except I don't understand how this fulfills #1905 - there is nothing provided to preserve ordering for schema-first schemas.  So it doesn't seem to have any practical usefulness as of yet.

Before merging this, I think we should finish the code required to preserve ordering for schema-first schemas, and then tests should be added.  I would not merge it quite yet, unless you plan to follow up with another PR very shortly.

sungam3r:
```c#
 public class MySchemaBuilder : SchemaBuilder
    {
        protected override void PreConfigure(Schema schema)
        {
            base.PreConfigure(schema);
            schema.Comparer = new MyComparer();
        }
    }
...
 var b = new MySchemaBuilder();
 b.Build("");
```

sungam3r:
`Schema.For` is just a shortcut for the default builder
```c#
  public static ISchema For(string typeDefinitions, Action<SchemaBuilder> configure = null)
        {
            var builder = new SchemaBuilder();
            configure?.Invoke(builder);
            return builder.Build(typeDefinitions);
        }
```

Shane32:
If a default comparer (with all null properties) will preserve ordering, then we should write a test case for that.

sungam3r:
Not all null, field comparer is not null and compares fields by their names. I will add tests later.

joemcbride (approved):
Additional "default" test would be good, but looks good.

A thought.  Should this also be used in the `SchemaPrinter`?  🤔

sungam3r:
The same concept _may be used_ in `SchemaPrinter`.

docs2/site/docs/guides/migration4.md
```diff
@@ -5,4 +5,7 @@
 ## Breaking Changes
 
 * `NameConverter` and `SchemaFilter` have been removed from `ExecutionOptions` and are now properties on the `Schema`.
+
 * `GraphQL.Utilities.ServiceProviderExtensions` has been made internal. This affects usages of it's extension method `GetRequiredService`. Instead reference the `Microsoft.Extensions.DependencyInjection.Abstractions` NuGet package and use extension method from `Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions` class.
```
Shane32, line 9:
```suggestion
* `GraphQL.Utilities.ServiceProviderExtensions` has been made internal. This affects usages of its extension method `GetRequiredService`. Instead, reference the `Microsoft.Extensions.DependencyInjection.Abstractions` NuGet package and use the extension method provided by the `Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions` class.
```

src/GraphQL/Introspection/ISchemaComparer.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// <summary>
+        /// Returns a comparer for GraphQL directives.
+        /// </summary>
+        IComparer<DirectiveGraphType> DirectiveComparer { get; }
+    }
+
+    /// <summary>
+    /// Default schema comparer. By default all elements are returned as is, no sorting is applied.
+    /// </summary>
+    public class DefaultSchemaComparer : ISchemaComparer
```
sungam3r, line 43: @Shane32 we have already discussed this. I am in favor of removing the default sorting in v4. `LegacyV3SchemaComparer` can be used to switch to the old behavior.

src/GraphQL/Introspection/ISchemaComparer.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        /// <inheritdoc/>
+        public virtual IComparer<IFieldType> FieldComparer(IGraphType parent) => null;
+    }
+
+    /// <summary>
+    /// Default schema comparer for GraphQL.NET v3.x.x.
+    /// By default only fields are ordered by their names within enclosing type.
+    /// </summary>
+    public sealed class LegacyV3SchemaComparer : DefaultSchemaComparer
```
sungam3r, line 70: Do we need this class?

src/GraphQL/Introspection/ISchemaComparer.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        /// <inheritdoc/>
+        public override IComparer<IFieldType> FieldComparer(IGraphType parent) => _instance;
+    }
+
+    /// <summary>
+    /// All elements are sorted in alphabetical order of their names.
+    /// </summary>
+    public class AlphabeticallySchemaComparer : ISchemaComparer
```
sungam3r, line 86: I'm not sure about naming - _Alphabetically_.

sungam3r:
I added new label for PRs with new APIs with non breaking changes.

sungam3r:
Both new alerts from lgtm bot are false positives.

src/GraphQL/Introspection/ISchemaComparer.cs
Shane32: Sounds fine to me.

src/GraphQL/Introspection/ISchemaComparer.cs
Shane32: Can we just include the source code for this class in the migration notes in case someone needs it?

sungam3r: done

Shane32 (approved):

src/GraphQL/Introspection/ISchemaComparer.cs
Shane32: I agree.  Don't have any better suggestions.  Perhaps `AlphabeticalSchemaComparer` ?

sungam3r: Done

PR #1910, opened 2020-10-13
Sync changes with master

sungam3r:


.github/workflows/build-artifacts-code.yml
```diff
@@ -5,6 +5,7 @@ name: Build artifacts
 on:
   push:
     branches:
+      - master
```
sungam3r, line 8: @Shane32 It is right?

.github/workflows/pr-tests-code.yml
```diff
@@ -3,20 +3,21 @@ name: Run code tests
 on:
   pull_request:
     branches: 
+      - master
```
sungam3r, line 6: @Shane32 It is right?

.github/workflows/pr-tests-docs.yml
```diff
@@ -3,6 +3,7 @@ name: Run documentation tests
 on:
   pull_request:
     branches: 
+      - master
```
sungam3r, line 6: @Shane32 It is right?

sungam3r:
`
Testhost process exited with error: It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '2.1.0' was not found.
`

Should we keep 2.1 target or just remove it?

Shane32:
Yes it should have `- master` in the workflows.

Yes we should keep the 2.1 target.

I've made the necessary changes

Shane32 (approved):

Shane32:
I re-did the merge to eliminate the extra commits in the history.

PR #1916, opened 2020-10-14
add solution folders

sungam3r:
To make it easier to work from the VS.


Shane32:
I wouldn't add the GitHub Action workflows within the solution.  I have not been able to push commits to GitHub if they contain changes to a workflow, and changes to the workflows shouldn't be represented within the solution file.  Doesn't matter too much, so whatever you want.

sungam3r:
> changes to the workflows shouldn't be represented within the solution file

Why? Workflow is just a yml text file. VS gives formatting, syntax highlighting, IntelliSense, folding etc. It is simply more convenient to work with these files from VS.

Shane32:
I don't know why.  If I make a change and commit it and try to push the commit, GitHub refuses the push.  Any other file, sure, but not GitHub Action workflow files.  It only lets me make changes to workflows from their website or from GitHub Desktop.  Maybe there's a configuration issue with my PC.

Shane32:
See https://github.com/github/VisualStudio/issues/2455

sungam3r:
Interesting. I use SourceTree to interact with git/GitHub and SSH key. I'll keep in mind and see what happens.

PR #1919, opened 2020-10-15
Add test for variable coercion

sungam3r:
Small test to verify behavior for variable coercion and error for wrong literal.


sungam3r:
At first glance, it seems strange why one request works, and the second does not. This question came up on one of our development teams.

sungam3r:
Well, in fact, everything is a little more interesting. See https://github.com/graphql/graphql-js/issues/1324
@Shane32 I remember you wondered at some places in code when serializing / passing scalars. It seems this is exactly the case. Good candidate for 3.2 milestone.

sungam3r:
If you want, then I can merge this test for 3.1 and then in 3.2 we will "fix" it. We can also defer this PR to 3.2 and make corrections right here.

Shane32:
> If you want, then I can merge this test for 3.1 and then in 3.2 we will "fix" it. We can also defer this PR to 3.2 and make corrections right here.

I’d probably at least add a todo note that this is unintended or incorrect behavior. 

The way I’m planning to fix this will probably be breaking and need to be in 4.0. I can explain later. Maybe a patch could be implemented for 3.2. Good to have this test though.

Shane32:
@sungam3r I suggest we merge this test.  It may be useful when I start reworking the input code.

Shane32 (approved):

PR #1922, opened 2020-10-18
DOTNET_SKIP_FIRST_TIME_EXPERIENCE

sungam3r:
To get rid of
изображение

### DIFF
--- a/.github/workflows/wipcheck.yml
+++ b/.github/workflows/wipcheck.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     types: 
       - opened        # when PR is opened
-      - edited        # when PR title is edited
+      - edited        # when PR is edited
       - synchronize   # when code is added
       - reopened      # when a closed PR is reopened
 
--- a/src/Tests.props
+++ b/src/Tests.props
@@ -6,7 +6,7 @@
   
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
-    <PackageReference Include="Shouldly" Version="4.0.0-beta0002" />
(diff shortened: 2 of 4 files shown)


.github/workflows/wipcheck.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
     types: 
       - opened        # when PR is opened
-      - edited        # when PR title is edited
+      - edited        # when PR is edited
```
sungam3r, line 7: Not only title. I noticed that editing my comment also triggers wipcheck.

sungam3r:
Also see this https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet#environment-variables

.github/workflows/wipcheck.yml
sungam3r: Well, it is true only for the first comment. Maybe it is treated in some special way.

src/Tests.props
```diff
@@ -6,7 +6,7 @@
   
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
-    <PackageReference Include="Shouldly" Version="4.0.0-beta0002" />
+    <PackageReference Include="Shouldly" Version="4.0.0-beta0004" />
```
sungam3r, line 9: I changed this to trigger actions on `src`

Shane32 (approved):
I really don’t understand why you’d want to reduce the log output of a workflow. GitHub collapses the information anyway so you don’t see it unless you expand the details for that action.  So I wouldn’t but it doesn’t really matter so ok.....

sungam3r:
Yes, you are right, there is not much difference, I just wanted to try it for a long time.

PR #1925, opened 2020-10-20
Add publishing workflow

Shane32:


.github/workflows/publish-code.yml
```diff
@@ -0,0 +1,74 @@
...
+        if: ${{ !startsWith(github.ref, 'refs/tags/') }}
+        run: |
+          echo Error! github.ref does not start with 'refs/tags'
+          echo github.ref: ${{ github.ref }}
+          exit 1
+      - name: Set version number environment variable
+        env:
+          github_ref: ${{ github.ref }}
+        run: |
+          ref="${github_ref:10}"
```
sungam3r, line 26: What is `ref`? Action variable? Why do you need to first create an environment variable, put the tag value into it, and then transfer it to the `ref`? Also what is `:10` ?

.github/workflows/publish-code.yml
```diff
@@ -0,0 +1,74 @@
...
+        with:
+          dotnet-version: '3.1.x'
+          source-url: https://api.nuget.org/v3/index.json
+        env:
+          NUGET_AUTH_TOKEN: ${{secrets.NUGET_AUTH_TOKEN}}
+      - name: Install dependencies
+        working-directory: src
+        run: dotnet restore
+        env:
+          DOTNET_NOLOGO: true
```
sungam3r, line 58: 👍

.github/workflows/publish-code.yml
```diff
@@ -0,0 +1,74 @@
...
+          DOTNET_CLI_TELEMETRY_OPTOUT: true
+      - name: Build solution (release config)
+        working-directory: src
+        run: dotnet build --no-restore -c Release -p:NoWarn=CS1591 -p:Version=$ref
```
sungam3r, line 62: Now I seem to understand that you were talking about automatic version overwriting when creating a release. In fact, it becomes automatic only after we automate it 😄 . Thus, when creating a release, we are now obliged to give it only a tag in the semver format, right?

.github/workflows/publish-code.yml
```diff
@@ -0,0 +1,74 @@
...
+          name: Documentation
+          path: docs2/public/**
+          if-no-files-found: error
+      - name: Use .NET Core 3.1 SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '3.1.x'
+          source-url: https://api.nuget.org/v3/index.json
+        env:
+          NUGET_AUTH_TOKEN: ${{secrets.NUGET_AUTH_TOKEN}}
```
sungam3r, line 53: I see `  run: dotnet nuget push out/* -k ${{ secrets.NUGET_AUTH_TOKEN }}` in the last line of this script. So why to bother with env variable for token?

.github/workflows/publish-code.yml
Shane32: the C# equivalent would be:
```
var ref = github_ref.Substring(10);
```
basically, strip off the first 10 characters, and store in "ref"

.github/workflows/publish-code.yml
Shane32: yup

.github/workflows/publish-code.yml
Shane32: In my tests for the server project, the `-k` option was not needed in the last line of the script.  For some reason, it is required with this project.

We certainly need to set the source-url here, so I set the auth token in both places to be sure it works correctly.  You will see that this file is pretty much a copy of the build-artifacts workflow, which first prompted this change.

.github/workflows/publish-code.yml
Shane32: You'll also notice the following line which is required to store the temporary environment variable as permanent for the rest of the workflow:
```
echo "::set-env name=ref::$ref"
```
See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions

.github/workflows/publish-code.yml
sungam3r: I don't think I understand. My question was about why declare an `NUGET_AUTH_TOKEN` environment variable and not use it later. I see that the last line uses NUGET_AUTH_TOKEN **from secrets**.

Shane32: That environment variable setting only applies to that step of the workflow.

.github/workflows/publish-code.yml
Shane32: It's possible that something like this could work:
```
echo "::set-env name=ref::${github_ref:10}"
```
I have not tried that.

.github/workflows/publish-code.yml
sungam3r: OK but then how does `actions/setup-dotnet@v1` use `NUGET_AUTH_TOKEN` ? I indeed see its usage in docs https://github.com/actions/setup-dotnet but don't understand what it is for.

Shane32: `setup-dotnet` basically just runs `dotnet nuget add source`, as far as I know.

https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-add-source

sungam3r: Let's say this is so, but what does the token have to do with it? Does the script stop working without this variable?

Shane32: You know as much as I do. It’s the recommended pattern here according to the GitHub docs. The real question is why the -k option is necessary, because it shouldn’t be. I know the -k wasn’t needed when I tested the workflows for the server project.

Shane32:
@sungam3r Any other comments before I merge this?  I was going to work on this tonight.

.github/workflows/publish-code.yml
sungam3r: OK. Let's assume that despite our misunderstanding some parts of this process, we are still moving in the right direction. We'll figure it out over time.

sungam3r:
Of course let's move on. Good job.

sungam3r (approved):

sungam3r:
Well, I stll have one suggestion though. Let's try to name action file name and action name in this file similarly.

sungam3r:
изображение

sungam3r:
изображение

Shane32:
How about this @sungam3r  ?  See last commit.  Also, if you want, I'll try deleting `NuGet.config` and then see if it work without the `-k` option.  I think perhaps since the `NuGet.config` file contains `<clear />` it is messing up the key stored by `setup-dotnet`.  But I'm just guessing; I need a little trial and error to see.

Shane32:
I also don't think we need the `NuGet.config` file anymore anyway.

sungam3r:
Sounds reasonable. Nuget.config is used to easily consume preview and release packages from myget. In principle, I also think that now you can get rid of it. At least try.

Regarding names: 

file name | action name
----|-----
publish-documentation.yml  |  Publish documentation
publish-code.yml                   | Publish code
test-documentation.yml    |  Test documentation
test-code.yml | Test code

So the names are completely symmetrical. It seems to me that the both names need to say **what** the action does, and not **when** (PR).

.github/workflows/build-artifacts-code.yml
```diff
@@ -62,5 +62,4 @@ jobs:
           path: |
             src/out/*
       - name: Publish Nuget packages to GitHub registry
-        working-directory: src
```
Shane32, line 65: It is possible this was causing the problem also -- if `setup-dotnet` added a `NuGet.config` file to the root, but the working directory was `src`, then it may not have been able to find the `NuGet.config` file with the password... just another guess of course.

PR #1928, opened 2020-10-20
remove ServiceProviderExtensions

SimonCropp:
The majority of things consuming will already reference `Microsoft.Extensions.DependencyInjection.Abstractions`. Which means ServiceProviderExtensions regularly causes type conflicts when using as an extension method

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0121	The call is ambiguous between the following methods or properties: 
'GraphQL.Utilities.ServiceProviderExtensions.GetRequiredService<T>(System.IServiceProvider)' and 
'Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService<T>(System.IServiceProvider)'	
```

So this PR removes ServiceProviderExtensions and adds a ref to Microsoft.Extensions.DependencyInjection.Abstractions


joemcbride:
I think my vote would be to just make that class internal vs. adding the additional dependency for 2 methods.

PR #1932, opened 2020-10-21
Use empty contructor if possible for graphtype resolution

SimonCropp:
This changes the resolve behavior to fall back to trying an empty constructor if the graph type cannot be found in DI.

For me the specific scenario (as show in the test) is the failure to resolve `EnumerationGraphType<MyEnum>` unless it is registered in the container. While it is possible to use reflection to register an instance EnumerationGraphType for each enums using reflection. but IMO it seems this should be the OOTB behavior.

thoughts?

### DIFF
--- a/src/GraphQL/Types/GraphTypesLookup.cs
+++ b/src/GraphQL/Types/GraphTypesLookup.cs
@@ -126,7 +126,7 @@ namespace GraphQL.Types
                         return graphType;
                     }
 
-                    var constructor = t.GetConstructor(new Type[]{});
+                    var constructor = t.GetConstructor(Type.EmptyTypes);
                     if (constructor != null)
                     {
                         return (IGraphType) constructor.Invoke(null);
--- a/src/GraphQL/Types/GraphTypesLookup.cs
+++ b/src/GraphQL/Types/GraphTypesLookup.cs
@@ -129,7 +129,7 @@ namespace GraphQL.Types
                     var constructor = t.GetConstructor(Type.EmptyTypes);
                     if (constructor != null)
                     {
-                        return (IGraphType) constructor.Invoke(null);
+                        return (IGraphType)constructor.Invoke(null);
                     }
 
                     throw new InvalidOperationException($"Could not resolve a GraphType for '{t.FullName}'. Tried resolving from service provider and searched for a public empty constructor.");
--- a/src/GraphQL/Types/GraphTypesLookup.cs
+++ b/src/GraphQL/Types/GraphTypesLookup.cs
@@ -132,7 +132,7 @@ namespace GraphQL.Types
                         return (IGraphType)constructor.Invoke(null);
                     }
 
-                    throw new InvalidOperationException($"Could not resolve a GraphType for '{t.FullName}'. Tried resolving from service provider and searched for a public empty constructor.");
+                    throw new InvalidOperationException($"Could not resolve a GraphType for '{t.FullName}'. Failed resolving from service provider or searching for a public empty constructor.");
                 },
                 addType: (name, graphType, context) =>
                 {
--- a/src/GraphQL/Types/GraphTypesLookup.cs
+++ b/src/GraphQL/Types/GraphTypesLookup.cs
@@ -83,7 +83,7 @@ namespace GraphQL.Types
                    }
                    ctx.AddType(trimmed, type, null);
                });
-            
+
             // Add introspection types. Note that introspection types rely on the
             // CamelCaseNameConverter, as some fields are defined in pascal case - e.g. Field(x => x.Name)
             NameConverter = CamelCaseNameConverter.Instance;
@@ -112,13 +112,36 @@ namespace GraphQL.Types
         {
             var lookup = nameConverter == null ? new GraphTypesLookup() : new GraphTypesLookup(nameConverter);
 
-            var ctx = new TypeCollectionContext(t => _builtInScalars.TryGetValue(t, out var graphType) ? graphType : resolveType(t), (name, graphType, context) =>
-            {
-                if (lookup[name] == null)
+            var ctx = new TypeCollectionContext(
+                resolver: t =>
                 {
-                    lookup.AddType(graphType, context);
-                }
-            });
+                    if (_builtInScalars.TryGetValue(t, out var graphType))
+                    {
+                        return graphType;
+                    }
+
+                    graphType = resolveType(t);
+                    if (graphType != null)
+                    {
+                        return graphType;
+                    }
+
+                    if (t.IsGenericType &&
+                        t.GetGenericTypeDefinition() == typeof(EnumerationGraphType<>))
+                    {
+                        var constructor = t.GetConstructor(new Type[]{});
+                        return (IGraphType) constructor.Invoke(null);
+                    }
+
+                    throw new InvalidOperationException($"Could not resolve a GraphType for '{t.FullName}'.");
+                },
+                addType: (name, graphType, context) =>
+                {
+                    if (lookup[name] == null)
+                    {
+                        lookup.AddType(graphType, context);
+                    }
+                });
 
             foreach (var type in types)
             {
@@ -170,7 +193,7 @@ namespace GraphQL.Types
         }
 
         /// <summary>
-        /// Removes all discovered types from lookup. 
+        /// Removes all discovered types from lookup.
         /// </summary>
         public void Clear() => Clear(false);
 
--- a/src/GraphQL/Types/GraphTypesLookup.cs
+++ b/src/GraphQL/Types/GraphTypesLookup.cs
@@ -126,14 +126,13 @@ namespace GraphQL.Types
                         return graphType;
                     }
 
-                    if (t.IsGenericType &&
-                        t.GetGenericTypeDefinition() == typeof(EnumerationGraphType<>))
+                    var constructor = t.GetConstructor(new Type[]{});
+                    if (constructor != null)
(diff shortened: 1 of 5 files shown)


Shane32:
Makes sense.  There's some special code to handle the graphql-relay generic types; that might be able to be removed with this change.

SimonCropp:
i think also bring us inline with how other frameworks work. eg in aspnet u dont need to register all controllers in the container

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -112,13 +112,35 @@ private void CheckSealed()
...
+                        return graphType;
+                    }
+
+                    graphType = resolveType(t);
+                    if (graphType != null)
+                    {
+                        return graphType;
+                    }
+
+                    var constructor = t.GetConstructor(new Type[]{});
```
sungam3r, line 129:
```suggestion
                    var constructor = t.GetConstructor(Type.EmptyTypes);
```

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -112,13 +112,35 @@ private void CheckSealed()
...
+                    if (constructor != null)
+                    {
+                        return (IGraphType) constructor.Invoke(null);
```
sungam3r, line 132:
```suggestion
                        return (IGraphType)constructor.Invoke(null);
```

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -112,13 +112,35 @@ private void CheckSealed()
...
+                    }
+
+                    throw new InvalidOperationException($"Could not resolve a GraphType for '{t.FullName}'. Tried resolving from service provider and searched for a public empty constructor.");
```
sungam3r, line 135:
```suggestion
                    throw new InvalidOperationException($"Could not resolve a GraphType for '{t.FullName}'. Failed resolving from service provider or searching for a public empty constructor.");
```
?

sungam3r:
OK in general. My thougths are:
1. Now we can and **should** get rid of
```c#
 if (namedType == typeof(PageInfoType))
                {
                    AddType(new PageInfoType(), context);
                }
                else if (namedType.IsGenericType && (namedType.ImplementsGenericType(typeof(EdgeType<>)) || namedType.ImplementsGenericType(typeof(ConnectionType<,>))))
                {
                    AddType((IGraphType)Activator.CreateInstance(namedType), context);
                }
```
introduced in #1039. It will be the first step to get rid of any relay stuff from this repo in favor of https://github.com/graphql-dotnet/relay (if we decide so in the future). At least the code will be cleaner.

2. GraphType can implement `IDisposable`, unlikely but real. All types instances created by us (that is, not with `services.Get[Required]Service()` call) **should** be tracked and disposed further when schema is disposed. 

3. Also I have some concern about `Activator.CreateInstance` in `GraphTypesLookup` ctor and second call to `TypeCollectionContext` ctor after creating `GraphTypesLookup` instance but this is a matter of another PR.

So I would like you to fix at least point 1. Maybe 2.

Shane32:
I agree.  Ideally, I think we should call `ActivatorUtilities.CreateInstance` rather than `Activator.CreateInstance`, so the constructor 
is called with DI-injected services, just like a MVC controller.  This would require a dependency on `Microsoft.Extensions.DependencyInjection.Abstractions`.  We still want to use `GetService` first, so we know if we need to track disposing of the service or not.

It really makes the most sense, because of the fact that the graph types are locked to the schema registration upon the schema initialization.  Which is why it's impossible to have a singleton schema with a scoped graph type -- the graph type will end up functioning as if it was a singleton also.  And since the service lifetime of the graph type now doesn't actually matter (they always function with the same lifetime as the schema), why register them in DI?  The user can still register the schema as scoped or singleton as desired.

So I'd suggest we take this a step further and use `ActivatorUtilities.CreateInstance` rather than `Activator.CreateInstance`.

As for the fact that it requires a dependency on `Microsoft.Extensions.DependencyInjection.Abstractions`, I'm not opposed to it.  I'm don't even think it's a "breaking" change.  Further, we _could_ do what Microsoft does for their own projects: duplicate the entire class into the project so that the dependency isn't required.  (I wouldn't.)

Finally, `ActivatorUtilities.CreateInstance` relies an `IServiceProvider`, which is a .NET Standard type.  So it has nothing to do with the Microsoft DI versus any other DI provider.

The source code can be seen here:

https://github.com/dotnet/runtime/blob/master/src/libraries/Common/src/Extensions/ActivatorUtilities/ActivatorUtilities.cs

sungam3r:
What I definitely would not want to do:

1. Add `Microsoft.Extensions.DependencyInjection.Abstractions` dependency.
2. Mix responsibility for creating and disposing services. With `ActivatorUtilities.CreateInstance` we don't know if we should dispose the resulting object ourselves. I did not see a place in that code where the created object would be registered in the container. It is simply created, and each time it is different.

By the way I have another suggestion - pass service provider in `GraphTypesLookup` ctor.

Shane32:
> Mix responsibility for creating and disposing services. With ActivatorUtilities.CreateInstance we don't know if we should dispose the resulting object ourselves. I did not see a place in that code where the created object would be registered in the container. It is simply created, and each time it is different.

Of course. Which is why I said:

> We still want to use GetService first, so we know if we need to track disposing of the service or not.

Shane32:
> I definitely would not want to add Microsoft.Extensions.DependencyInjection.Abstractions dependency.

Why "definitely"?  It certainly merits a discussion.  Almost everyone using this project will use that dependency, directly or indirectly.

Shane32:
If the behavior is not included, we should expose the proper members (e.g. as protected) to make this behavior "opt-in".

sungam3r:
> Why "definitely"? 

Because it is an external entity. This dependency is not part of the scope of this project. The fewer external dependencies, the better. The more freedom in decision-making both on our part and on the part of the consumer. Each new dependency carries versioning problems etc.

> Almost everyone using this project will use that dependency, directly or indirectly.

**Almost** but not **absolutely**.  e once discussed with @joemcbride the issue of consuming this dependency. Then we agreed that we could well solve issues without this. So far it has been.

> However, I guess a custom IServiceProvider implementation could serve to do the trick.

Absolutely. This is the point. We already have well-defined interfaces to offload this work outside the core GraphQL.NET. This is a question of small adapter (IServiceProvider) class. You can even just give an example of this class in the dependencies documentation.

SimonCropp:
should i do another pr that only targets enum type in the short term?

sungam3r:
I don't think we should somehow split the logic for creating enums or other types like relay stuff. This will only complicate the situation. In fact, you can only do point 1 - get rid of creating relay types stuff. It is easy. We will not lose much if we do not track the created instances yet. I have little idea who can create **disposable** enumerations 😮 . You can just create a separate issue and link it to this one.

Shane32:
> Why "definitely"? Because it is an external entity

I disagree with this philosophy -- specifically the "definitely" part.  Sometimes "the needs of the many outweigh those of the few".  We should not have a blanket statement "no third party dependencies" without weighing the gains and losses open-mindedly.

> This dependency is not part of the scope of this project.

This project relies on dependency injection, in various forms.  Dependency injection is not something that can be tossed aside as someone else's problem.  We can discuss how best to integrate dependency injection in the project, or if it should be a separate nuget package.  But it's fairly critical to anyone with serious thoughts of using this project, as can be seen from the numerous questions, issues and pull-requests relating to the topic.

As such, if there are features that relate to dependency injection and how it interacts with this project, they should be discussed and weighed against other design goals.  This happens to be one of them.

We may find in the future that other third party libraries greatly benefit the project.  Purely for example, let's say we want to use a `MemoryCache` from `Microsoft.Extensions.Caching.Memory`.  I would think we should weigh the benefits and costs of integrating the third party library into the project with an open mind, considering potential versioning conflicts, known issues, complexity of the library, potential for updates, how common the library is, or even how uncommon it is.  We should not immediately rule out having a dependency.

This is how open-source projects thrive - building on others' work.  Ruling out third party packages without consideration is not a good policy.

As for this particular idea, using `ActivatorUtilities.CreateInstance`, there are multiple ways to integrate this idea.  Perhaps we should consider this along with #1571 and see how it might fit into the project.

Shane32:
> should i do another pr that only targets enum type in the short term?

As far as I'm concerned, go ahead and finish the PR like you had planned.  `IDisposable` support is probably wise.  We can always add support for DI-created instances via `ActivatorUtilities.CreateInstance` in a subsequent PR without breaking any existing code.  Only question I'm now thinking: should we limit the ability to auto-construct objects to those with ONLY a single constructor without parameters?  (vs multiple constructors)

Shane32:
> I don't think we should somehow split the logic for creating enums or other types like relay stuff. This will only complicate the situation. In fact, you can only do point 1 - get rid of creating relay types stuff. It is easy. We will not lose much if we do not track the created instances yet. I have little idea who can create **disposable** enumerations 😮 . You can just create a separate issue and link it to this one.

Agree

sungam3r:
I could have given the wrong impression. I am not against new dependencies as such.

joemcbride (approved):
I like this direction and think it would simplify things.  I agree that the Relay types could maybe be cleaned up, but this is a fairly small change for a good gain.  👍🏼

sungam3r:
@SimonCropp Now the problem solved by `GraphQL.MicrosoftDI.SelfActivatingServiceProvider`.

PR #1933, opened 2020-10-21
Create NuGet.config

Shane32:
See:
- https://github.com/actions/setup-dotnet/issues/75#issuecomment-613118478
- https://docs.myget.org/docs/how-to/nuget-configuration-inheritance


Shane32:
Unable to get anything to work like the docs says.  Probably could use the nuget CLI (if it's configured in the docker) to create/modify the NuGet.config file.  So just reverting to what worked before.

PR #1938, opened 2020-10-23
Add common status check for build checks

Shane32:
With this change, both code and documentation build checks (which are in different workflows) will create a common "buildcheck" status which can be set in the branch protection options.

I imagine we want to turn off the old branch protection option (set for appveyor) and change to the 'buildcheck' status, which is triggered by either the `test-code.yml` or `test-documentation.yml` workflows.

Current configuration:
image

### DIFF
--- a/.github/workflows/test-code.yml
+++ b/.github/workflows/test-code.yml
@@ -5,8 +5,6 @@ on:
     branches: 
       - master
       - develop
-    paths: 
-      - src/**
 
 jobs:
   test:
@@ -38,3 +36,7 @@ jobs:
       - name: Test solution (debug config)
         working-directory: src
         run: dotnet test --no-restore
+  buildcheck:
+    needs:
+      - test
+    runs-on: ubuntu-latest
--- a/.github/workflows/test-code.yml
+++ b/.github/workflows/test-code.yml
@@ -5,6 +5,8 @@ on:
     branches: 
       - master
       - develop
+    paths:
+      - src/**
 
 jobs:
   test:
--- a/.github/workflows/test-code.yml
+++ b/.github/workflows/test-code.yml
@@ -5,7 +5,7 @@ on:
     branches: 
       - master
       - develop
-    paths:
+    paths: 
       - src/**
 
 jobs:
--- a/.github/workflows/test-code.yml
+++ b/.github/workflows/test-code.yml
@@ -40,3 +40,6 @@ jobs:
     needs:
       - test
     runs-on: ubuntu-latest
+    steps:
+      - name: Pass build check
+        run: exit 0
--- a/.github/workflows/test-documentation.yml
+++ b/.github/workflows/test-documentation.yml
@@ -5,10 +5,6 @@ on:
     branches: 
       - master
       - develop
-    paths: 
-      - docs2/**
-      - package.json
-      - yarn.lock
 
 env:
   NODE_VERSION: '10.x'   # Node 10 LTS
--- a/.github/workflows/test-documentation.yml
+++ b/.github/workflows/test-documentation.yml
@@ -14,7 +14,7 @@ env:
   NODE_VERSION: '10.x'   # Node 10 LTS
 
 jobs:
-  build-docs:
+  buildcheck:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source
--- a/.github/workflows/test-documentation.yml
+++ b/.github/workflows/test-documentation.yml
@@ -5,6 +5,10 @@ on:
     branches: 
       - master
       - develop
+    paths: 
+      - docs2/**
+      - package.json
+      - yarn.lock
 
 env:
   NODE_VERSION: '10.x'   # Node 10 LTS


.github/workflows/test-code.yml
```diff
@@ -38,3 +38,10 @@ jobs:
       - name: Test solution (debug config)
         working-directory: src
         run: dotnet test --no-restore
+  buildcheck:
+    needs:
+      - test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pass build check
```
sungam3r, line 46: That is, is it a edge case when all the work has already been done in the dependent step `test`? Also why do you use `runs-on: ubuntu-latest` ? Does this narrow the check?

sungam3r: Ah, buildcheck is just a name for ordinary action.

Shane32: What this job does is wait for the prior job (which functions as two separate jobs) to fully complete, and if both pass, then pass a "buildcheck" status back to the PR.  Without this, the two separate jobs report status checks separately for each OS.

`runs-on` is a required field for a job, as is `steps`.  As you can see, this job doesn't actually "do" anything.  It immediately passes ... so long as the prior steps both pass.

sungam3r:
OK. I approximately understood and agree to turn off the old branch protection option. Maybe also turn on `Require branches to be up to date before merging ` option? Or it will be too restrictive ?

sungam3r:
The same question about `Require linear history`.

.github/workflows/test-code.yml
sungam3r: Just wondering. Is it essential that the name of the job be unique among all actions? Now the same name `buildcheck` is used inside two actions, but the actions themselves never fire together because they are responsible for different parts of the project. 

Could there be two entries with the same name?
изображение

sungam3r: And one more question. Don't get me wrong, I'm just learning from your examples. Until now, any action could exit with a code of 1 (error) and would it still allow us to merge? Only after the buildcheck is turned on, a strict check could be introduced to forbid merging?

sungam3r: > Without this, the two separate jobs report status checks separately for each OS.

That is, did I understand correctly that the successful status will be set on PR only when the tests pass successfully on all operating systems? If that's true, then I'm just a little confused by the `runs-on: ubuntu-latest` line. This gives the impression that this will only check the success status on one OS.

Shane32: > Just wondering. Is it essential that the name of the job be unique among all actions? Now the same name buildcheck is used inside two actions, but the actions themselves never fire together because they are responsible for different parts of the project.

Well, I tested both workflows and it didn't add two "buildcheck" items in the list.  So it would seem to match based on name.

I think the safest way to go would to be to merge the two test workflow files together.  But, we cannot have path filtering on different jobs in the same workflow.  So, it would always run both code and documentation checks.  (Which considering that workflow runtime is free for public repos is not a big deal.)

I do think this PR will work as-is, but we may find otherwise.  For instance, what happens if both code and docs are changed in the same PR?  Does it fail if any fail?  Does it fail only if the last to finish fails?  I don't know for sure.  I'm just hoping it works like we want it to.

> Until now, any action could exit with a code of 1 (error) and would it still allow us to merge?

Not exactly.  Right now it is requiring "continuous-integration/appveyor/pr" to pass (unless you just changed it).  But as an 'administrator' on this repo, you and I can merge regardless of status checks.

> That is, did I understand correctly that the successful status will be set on PR only when the tests pass successfully on all operating systems? If that's true, then I'm just a little confused by the runs-on: ubuntu-latest line. This gives the impression that this will only check the success status on one OS.

That line with the `runs-on` indicates what platform that that "buildcheck" job runs on.  The "buildcheck" job requires the "test" job to fully pass (including both matrix steps) before it will run at all.  So if windows or ubuntu fails in the first job, then the "buildcheck" job will never run, regardless of what platform that the "buildcheck" job is programmed to run on.

Shane32:
> Maybe also turn on `Require branches to be up to date before merging ` option? Or it will be too restrictive ?

Either way is fine, but I'd probably leave it off for a combination of three reasons:

1. When the PR runs its build checks, it will run against a merge of the branch with the base branch.  So problems can only arise when (1) a PR is approved, and (2) a different conflicting PR is merged, and (3) the first PR is merged -- and only in that order.  
2. It's often just a pain because you want to merge a couple PRs, and now you need to squash & merge one to master, then merge the second PR with master, wait for build checks to complete, and then squash & merge the second.
3. Normally if the PRs have conflicts, they will be reflected in the fact that they have modified the same files, and it won't be possible to merge the PR anyway.

Again, as administrators, you and I can bypass this restriction.

Shane32:
> The same question about `Require linear history`.

I think it's more important to change the merge button functionality so someone doesn't accidentally do a regular merge. I suggest as follows:

image

The only issue is that when we need to merge master into develop, we can't do it from the browser.  Ok with me.  Too bad there's no admin-override for this option.

As for requiring linear history, this will not prevent a regular merge (if it's from a fresh/rebased PR) or a "rebase & merge" into master.  The branch is already protected from changes directly into master by non-admins by having this option set:

image

And as administrators, we are not affected by the restriction.  So enabling the option doesn't change anything for anyone, as far as I can tell -- assuming we only allow the "squash and merge" button, that is.  So.... I guess we can enable it???

.github/workflows/test-code.yml
sungam3r: >  For instance, what happens if both code and docs are changed in the same PR? Does it fail if any fail? Does it fail only if the last to finish fails?

Interesting question. I don't know too.

> Right now it is requiring "continuous-integration/appveyor/pr" to pass (unless you just changed it)

I khow abou it. I meant a situation where we already turn it off. It's just that before that I thought that any failed action prevents merge. I didn't know about picking special action (or actions) as a build status indicator.

>  The "buildcheck" job requires the "test" job to fully pass (including both matrix steps) before it will run at all. So if windows or ubuntu fails in the first job, then the "buildcheck" job will never run, regardless of what platform that the "buildcheck" job is programmed to run on.

OK. Exactly what is needed.

sungam3r:
Agree. So lets keep `Require branches to be up to date before merging` and `Require linear history` disabled but allow only squashed merges.

> When the PR runs its build checks, it will run against a merge of the branch with the base branch.

Just to clarify - does `on:
  pull_request:` work on **every commit** into created PR and not only to "first time created" PR. The question may be stupid, but still I wanted to ask it.

> The only issue is that when we need to merge master into develop, we can't do it from the browser. 

Also when merging develop into master, I think, although this will happen less often.

sungam3r (approved):

Shane32:
> Just to clarify - does `on: pull_request:` work on **every commit** into created PR and not only to "first time created" PR. The question may be stupid, but still I wanted to ask it.

Umm, I'm not sure I understand, but I think so?  As far as I know, when the workflow runs, it takes the current branch of the PR and merges it with the current base branch (at the time of the execution of the workflow), and then executes the workflow based on the resulting merge.  So in other words, it executes the same as if someone had just merged the PR.  It does not automatically run again if the base branch is updated.  That's just my understanding and it could be wrong.

Here's what GitHub's docs says:

> Last merge commit on the GITHUB_REF branch

(whatever that means)

We can also see an option in the `actions/checkout` workflow:

> [How to] Checkout pull request HEAD commit instead of merge commit

Shane32:
I'll merge this PR and make the suggested changes to the options, and we'll see how it goes.

sungam3r:
> It does not automatically run again if the base branch is updated. 

But it DOES automatically run again if the CURRENT branch is updated. Right?

sungam3r:
изображение

sungam3r:
Two identical checks:
https://github.com/graphql-dotnet/graphql-dotnet/runs/1296732979
https://github.com/graphql-dotnet/graphql-dotnet/runs/1296749013

sungam3r:
изображение

sungam3r:
Ah! I understood why there are no other checks, because only workflow files are changed. But the question still remains, why are there two?

Shane32:
Either the `edited` or the `synchronize` event must have triggered the workflow to run again.  The documentation doesn't really seem to explain what exactly triggers those events.  But let's say someone edited the PR description.  The workflow would run again with the same files and same commit -- it would appear to be an identical run from the logs.  I think that's what you're seeing.  For all I know, maybe the act of merging the PR triggers the event, creating the duplicate status check...

Shane32:
Btw, we could alter the workflows so that the code and doc test workflows both run when the yml files change.

Shane32:
> But it DOES automatically run again if the CURRENT branch is updated. Right?

Yes

sungam3r:
> Btw, we could alter the workflows so that the code and doc test workflows both run when the yml files change.

This makes sense and will allow to debug actions easily.

Shane32:
Seems to be working great!  Notice that both checks are listed separately and both get marked as required.

image

Shane32:
Umm... wait maybe not.... if windows-latest fails, then the bottom buildcheck might not show up.  So it might not actually work quite perfectly.

Shane32:
Yeah this is what I mean -- it will never start the code buildcheck if the windows or ubuntu checks fail.  Notice it's not preventing a merge in this case.  Of course if the docs were not changed then it would be fine because the docs wouldn't have reported success.

image

sungam3r:
Maybe just rename `test` to `buildcheck` how is it done in test-documentation.yml and remove 
```
 buildcheck:
    needs:
      - test
    runs-on: ubuntu-latest
    steps:
      - name: Pass build check
        run: exit 0
```
?

sungam3r:
Also I doubt the need of ` if: ${{ startsWith(matrix.os, 'ubuntu') }}` . As you said we do not pay for processing time. Let's do full checks.

Shane32:
> Maybe just rename `test` to `buildcheck` how is it done in test-documentation.yml and remove
> 
> ```
>  buildcheck:
>     needs:
>       - test
>     runs-on: ubuntu-latest
>     steps:
>       - name: Pass build check
>         run: exit 0
> ```
> 
> ?

Already tried.  I can unroll the matrix and then it will work as you described.

So either we can do that and retain the existing conditional behavior (so code checks only run for code changes, etc), or we can combine both the test workflows in a single file (so code and doc checks run for any PR).  What do you think?

sungam3r:
I don't know which is better. On the one hand, I don't want to run tests if I only changed the readme. What do you think should be done?

Shane32:
I found a way to fix it properly - see #1947

PR #1941, opened 2020-10-23
Run code & documentation tests when workflows are modified

Shane32:


.github/workflows/test-code.yml
```diff
@@ -7,6 +7,7 @@ on:
       - develop
     paths: 
       - src/**
+      - .github/**
```
sungam3r, line 10:
```suggestion
      - .github/workflows/**
```

.github/workflows/test-documentation.yml
```diff
@@ -9,6 +9,7 @@ on:
       - docs2/**
       - package.json
       - yarn.lock
+      - .github/**
```
sungam3r, line 12:
```suggestion
      - .github/workflows/**
```

sungam3r (approved):
изображение
I think it is not worth to run all tests when we change funding or issue templates.

PR #1942, opened 2020-10-23
Remove set-env from workflow

Shane32:
- https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
- https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files


Shane32:
I didn't actually test this change, but it seems to be a pretty straightforward change.

sungam3r (approved):

PR #1946, opened 2020-10-25
Add labeler

sungam3r:
Partially automates the labeling process.


.github/labeler.yml
```diff
@@ -0,0 +1,19 @@
+"test":
```
sungam3r, line 1: I set the labels that were most obvious.

Shane32 (approved):

Issue #1952, opened 2020-10-27
[Allow to replace built-in scalars] Different  behavior in data conversion and serialization between versions 2.4 and 3.1


rmeshksar:
## Summary
We have an old big solution based on GraphQL version 0.14.6 and trying to migrate to version 3.1.

Unfortunately we cannot fix the schema definition of the solution and need to make sure it is backward compatible.
Almost all issues have been resolved expect the one that I am going to describe here.

When returning the response, versions 0.14.6 and 2.4 just serialize the data based on the returned object type, however in version 3.1, the serlization is based on the used graph type of the query. Now if there is mismatch between POCO object and GraphType, the end result is different.

I know they should match, but as I mentioned since the code set that we have is very big, there is no way to manually check this, unless there is a recommendation on how to do that programmatically.

The following code shows the issue the way that the schema is defined and used in our code.
As you can see in the TestReponse class IsAvailable is boolean but in TestReponseType the field is StringGraphType (we have a lot of this bad setting in the code). 
Now version 2.4 returns isAvailable true (boolean) and version 3.1 returns "True" (string).

```

void Main()
{
	var schema = new Schema { Query = new TestQuery() };

	var exec = new DocumentExecuter();

	var result = exec.ExecuteAsync(options =>
			{
				options.Schema = schema;
				options.Root = null;
				options.Query = "{ test { name isAvaialble  } }";
				options.OperationName = "";
			}
	).Result;

	var json = JsonConvert.SerializeObject(result.Data);  
	/*
		version 0.14.6 and 2.4  --> {"test":{"name":"myName","isAvaialble":true}}
		version 3.1		--> {"test":{"name":"myName","isAvaialble":"True"}}
	*/
        

	

}

public class TestQuery : ObjectGraphType
{
	public TestQuery()
	{
		Name = "testQuery";
		Description = "Test description";

		Field<TestResponseType>(
			name: "test",
			description: "Test",
			resolve: context =>
			{
				return new TestResponse();
			}
		);
	}
}

public class TestResponseType : ObjectGraphType
{
	public TestResponseType()
	{
		Name = "TestReponse";
		Description = "Test Response Description";

		Field<StringGraphType>("isAvaialble");
		Field<StringGraphType>("name");

		IsTypeOf = value => value is TestResponse;
	}
}

public class TestResponse
{
	public bool IsAvaialble { get; set; }
	public string Name { get; set; }
	public TestResponse()
	{
		IsAvaialble = true;
[...]

PR #1953, opened 2020-10-27
Expose Schema.CreateTypesLookup and related members as protected

Shane32:
Possible fix for #1952 

Purpose:
- Allow changing behavior of built-in scalars such as `StringGraphType`

Challenges:
- Built-in scalar behaviors should probably be retained for introspection types

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
@@ -1901,6 +1901,9 @@ namespace GraphQL.Types
     {
         public GraphTypesLookup() { }
         public GraphTypesLookup(GraphQL.Conversion.INameConverter nameConverter) { }
+        protected virtual System.Collections.Generic.IReadOnlyDictionary<System.Type, GraphQL.Types.IGraphType> BuiltInCustomScalars { get; }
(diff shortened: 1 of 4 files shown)


Shane32:
The problem with this sample is that changing the built-in scalars affects the built-in introspection types, which is not the intent of the sample.  Further, the test would not work correctly at first because the introspection types were `static`, so the `GraphTypesLookup` class would 'lock onto' the `StringGraphType` previously used in the static introspection types, ignoring the `BuiltInScalars` protected member.  This wouldn't matter if you didn't use more than one schema in your codebase, but it's noteworthy.

I do like where this is heading, but it would seem to me that this requires some further review before any change is made.

sungam3r:
Wow. Wait a minute. Why static??? I remember that I changed them to be instance fields. https://github.com/graphql-dotnet/graphql-dotnet/pull/1887#issuecomment-708261218 I don't understand yet what happened. They are really static.

sungam3r:
Some kind of nonsense has happened. Either I lost the last commit, or I am hallucinating.

sungam3r:
#1954

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/releases/tag/3.1.1

Shane32:
`GraphTypesLookup` is renamed `SchemaTypes` and is mostly private as of #2183 so closing this PR.  We can revisit this if there is a need for the class to become more extensible.

PR #1954, opened 2020-10-27
Make types in GraphTypesLookup non static

sungam3r:


Shane32 (approved):

sungam3r:
Because of this, someone may have serious problems with the previous release.

sungam3r:
I think it is time to release 3.1.1 with this bugfix.

Shane32:
Ok want me to do so?  Or you can.

Shane32:
You already did 👍

sungam3r:
Will we continue to name each release simply as 'Version X.Y.Z'?

изображение

Also will we drop published nugets into each release (3 nupkgs)?

sungam3r:
I updated 3.1.0 release with 3 packages. No action triggered. Good.

Shane32:
> Will we continue to name each release simply as 'Version X.Y.Z'?

Seems fine to me.

Shane32:
Note: bug introduced in #1887 / version 3.1.0

Fixed in version 3.1.1

PR #1956, opened 2020-10-28
Add docs and packages to release artifacts

Shane32:


.github/workflows/publish-code.yml
```diff
@@ -72,3 +72,26 @@ jobs:
       - name: Publish Nuget packages to Nuget registry
         working-directory: src
         run: dotnet nuget push "out/*" -k ${{secrets.NUGET_AUTH_TOKEN}}
+      - name: Zip documentation
+        working-directory: docs2/public
+        run: zip -r ../../src/out/Documentation.zip "./*"
```
Shane32, line 77: Is `zip -r ../../src/out/Documentation.zip "./*"` correct?  I added the quotes so we don't have problems with bash again.

Shane32: See https://stackoverflow.com/questions/26646616/how-to-create-a-zip-file-using-shell-script

.github/workflows/publish-code.yml
```diff
@@ -72,3 +72,26 @@ jobs:
...
+      - name: Publish release artifacts
+        uses: actions/github-script@v2
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            console.log('environment', process.versions);
+            const fs = require('fs').promises;
+            const { repo: { owner, repo }, sha } = context;
+
+            for (let file of await fs.readdir('src/out')) {
```
Shane32, line 87: Does `file` contain just the filename or also the directory?

.github/workflows/publish-code.yml
```diff
@@ -72,3 +72,26 @@ jobs:
...
+              console.log('uploading', file);
+
+              await github.repos.uploadReleaseAsset({
+                owner,
+                repo,
+                release_id: ${{ github.event.release.id }},
```
Shane32, line 93: Is `github.event.release.id` correct?

Shane32: It would appear so, like how it is used here: https://github.com/marketplace/actions/release-feedback

.github/workflows/publish-code.yml
sungam3r: OK

.github/workflows/publish-code.yml
```diff
@@ -72,3 +72,26 @@ jobs:
       - name: Publish Nuget packages to Nuget registry
         working-directory: src
         run: dotnet nuget push "out/*" -k ${{secrets.NUGET_AUTH_TOKEN}}
+      - name: Zip documentation
```
sungam3r, line 75: I'm not sure if anyone needs it.

.github/workflows/publish-code.yml
sungam3r: I don't know

.github/workflows/publish-code.yml
```diff
@@ -72,3 +72,26 @@ jobs:
...
+        working-directory: docs2/public
+        run: zip -r ../../src/out/Documentation.zip "./*"
+      - name: Publish release artifacts
+        uses: actions/github-script@v2
```
sungam3r, line 79: It's a shame that we have to resort to a lower-level tool instead of single call to `upload-release-asset` action.

.github/workflows/publish-code.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       - name: Pack solution (release config)
         working-directory: src
         run: dotnet pack --no-restore -c Release -p:Version=$ref -o out
-      - name: Publish artifacts
+      - name: Publish code artifacts
```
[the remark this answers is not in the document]
sungam3r: Maybe add `working-directory: src` to this step and change `src/out/*` to ` out/*` to align with other steps?

sungam3r, line 66:
```suggestion
      - name: Upload nuget packages
```

.github/workflows/publish-code.yml
```diff
@@ -72,3 +72,26 @@ jobs:
       - name: Publish Nuget packages to Nuget registry
         working-directory: src
         run: dotnet nuget push "out/*" -k ${{secrets.NUGET_AUTH_TOKEN}}
+      - name: Zip documentation
+        working-directory: docs2/public
+        run: zip -r ../../src/out/Documentation.zip "./*"
+      - name: Publish release artifacts
```
sungam3r, line 78:
```suggestion
      - name: Upload nuget packages to release
```

.github/workflows/publish-code.yml
sungam3r: Maybe add `working-directory: src ` ?

Shane32: Tried but that argument is only valid when using runs and not uses

sungam3r: Ah, exactly.

.github/workflows/publish-code.yml
Shane32:
```suggestion
      - name: Upload nuget packages as workflow artifacts
```

.github/workflows/publish-code.yml
Shane32:
```suggestion
      - name: Upload nuget packages as release artifacts
```

Shane32:
All of the changes occur after the existing steps in the workflow.  So even if there's a problem, it would occur after the nuget packages have successfully been published to NuGet.

We can either wait and try these changes during the next release (e.g. 3.2.0), or one of us can do additional testing to confirm that it works properly.

.github/workflows/publish-code.yml
sungam3r: Finally, the clearest title! 👍

sungam3r:
I have a good candidate to test it - #1962 for 3.1.2 release

sungam3r:
By the way in your links I see action which creates release.  Perhaps in the future it will be possible to automate the creation of releases. But I like the way it is happening now.

sungam3r (approved):

Shane32:
> By the way in your links I see action which creates release. Perhaps in the future it will be possible to automate the creation of releases. But I like the way it is happening now.

Yes I much prefer that we create the release -- with the proper description and so on -- and the action occurs due to that.  Maybe for private repos it works better to have a release created automatically when a proper tag is detected, as then (perhaps) you can issue a release directly from Visual Studio.

Shane32:
> I have a good candidate to test it - #1962 for 3.1.2 release

Excellent.

.github/workflows/publish-code.yml
Shane32: lol 👍

Shane32:
It didn't work 👎   - but as anticipated the NuGet packages deployed to NuGet.org successfully

When I get some time, I'll mess around with this some more.

sungam3r:
Maybe because docs are already zipped?
изображение

PR #1961, opened 2020-10-28
Move ValidateName calls after NameConverter is applied

### LINKED ISSUE #1840: Validate names after they are mapped with INameConverter
## Summary

At the moment the field name is validated when you add a field to a complex types, but with the `INameConverter` you can modify the name afterwards.

The current approach has two issues:

1. You cannot write a `INameConverter` that fixes invalidate graphql names.
2. You can write a buggy `INameConverter` that creates invalidate graphql names.

If you remove the name validation from the Complex Type and do it just after `INameConverter` is called you could fix both issues.

Shane32:
Closes #1840 

The purpose of this PR is the following:
1. Allow invalid field and argument graphql names prior to schema initialization (not currently allowed)
2. Apply NameConverter to field and argument names during schema initialization (already exists)
3. Validate that the updated names are valid during schema initialization (currently does not happen)

The purpose of this PR is NOT:
1. Code cleanup / consolidation
2. Allow invalid field / argument names in an intialized schema

### DIFF
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -357,6 +357,24 @@ namespace GraphQL.Tests.Types
             schema.Initialize();
         }
 
+        [Fact]
+        public void throws_with_bad_namevalidator()
+        {
+            var exception = Should.Throw<ArgumentOutOfRangeException>(() =>
+            {
+                var type = new ObjectGraphType();
+                type.Field<StringGraphType>("hello");
+                var schema = new Schema
+                {
+                    Query = type,
+                    NameConverter = new TestNameConverter("hello", "hello$")
+                };
+                schema.Initialize();
+            });
+
+            exception.Message.ShouldStartWith($"A field name must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but hello$ does not.");
+        }
+
         private class TestNameConverter : INameConverter
         {
             private readonly string _from;
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using GraphQL.Conversion;
 using GraphQL.StarWars.Types;
 using GraphQL.Types;
 using GraphQL.Utilities;
@@ -302,13 +303,27 @@ namespace GraphQL.Tests.Types
             exception.Message.ShouldStartWith("The declared field 'genericname' on 'ListOfDroid' requires a field 'Type' when no 'ResolvedType' is provided.");
         }
 
+        private Exception test_field_name(string fieldName)
+        {
+            // test failure
+            return Should.Throw<ArgumentOutOfRangeException>(() =>
+            {
+                var type = new ObjectGraphType();
+                type.Field<StringGraphType>(fieldName);
+                var schema = new Schema
+                {
+                    Query = type
+                };
+                schema.Initialize();
+            });
+        }
+
         [Theory]
         [InlineData("__id")]
         [InlineData("___id")]
         public void throws_when_field_name_prefix_with_reserved_underscores(string fieldName)
         {
-            var type = new ComplexType<TestObject>();
-            var exception = Should.Throw<ArgumentOutOfRangeException>(() => type.Field<StringGraphType>(fieldName));
+            var exception = test_field_name(fieldName);
 
             exception.Message.ShouldStartWith($"A field name: {fieldName} must not begin with \"__\", which is reserved by GraphQL introspection.");
         }
@@ -319,10 +334,44 @@ namespace GraphQL.Tests.Types
         [InlineData("id$")]
         public void throws_when_field_name_doesnot_follow_spec(string fieldName)
         {
-            var type = new ComplexType<TestObject>();
-            var exception = Should.Throw<ArgumentOutOfRangeException>(() => type.Field<StringGraphType>(fieldName));
+            var exception = test_field_name(fieldName);
 
             exception.Message.ShouldStartWith($"A field name must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but {fieldName} does not.");
+        }
+
+        [Theory]
+        [InlineData("__id")]
+        [InlineData("___id")]
+        [InlineData("i#d")]
+        [InlineData("i$d")]
+        [InlineData("id$")]
+        public void does_not_throw_with_filtering_namevalidator(string fieldName)
+        {
+            var type = new ObjectGraphType();
+            type.Field<StringGraphType>(fieldName);
+            var schema = new Schema
+            {
+                Query = type,
+                NameConverter = new TestNameConverter(fieldName, "pass")
+            };
+            schema.Initialize();
+        }
+
+        private class TestNameConverter : INameConverter
+        {
+            private readonly string _from;
+            private readonly string _to;
+            public TestNameConverter(string from, string to)
+            {
+                _from = from;
+                _to = to;
+            }
+
+            public string NameForArgument(string argumentName, IComplexGraphType parentGraphType, FieldType field)
+                => argumentName == _from ? _to : argumentName;
+
+            public string NameForField(string fieldName, IComplexGraphType parentGraphType)
+                => fieldName == _from ? _to : fieldName;
         }
 
         [Theory]
--- a/src/GraphQL/Builders/ConnectionBuilder.cs
+++ b/src/GraphQL/Builders/ConnectionBuilder.cs
@@ -107,7 +107,8 @@ namespace GraphQL.Builders
 
         public ConnectionBuilder<TSourceType> Name(string name)
         {
-            NameValidator.ValidateName(name);
+            if (string.IsNullOrEmpty(name))
+                NameValidator.ValidateName(name);
 
             FieldType.Name = name;
             return this;
--- a/src/GraphQL/Builders/ConnectionBuilder.cs
+++ b/src/GraphQL/Builders/ConnectionBuilder.cs
@@ -107,9 +107,6 @@ namespace GraphQL.Builders
 
         public ConnectionBuilder<TSourceType> Name(string name)
         {
-            if (string.IsNullOrEmpty(name))
-                NameValidator.ValidateName(name);
-
             FieldType.Name = name;
             return this;
         }
--- a/src/GraphQL/Builders/FieldBuilder.cs
+++ b/src/GraphQL/Builders/FieldBuilder.cs
@@ -55,9 +55,6 @@ namespace GraphQL.Builders
 
         public virtual FieldBuilder<TSourceType, TReturnType> Name(string name)
         {
-            if (string.IsNullOrEmpty(name))
-                NameValidator.ValidateName(name);
-
             FieldType.Name = name;
             return this;
         }
--- a/src/GraphQL/Builders/FieldBuilder.cs
+++ b/src/GraphQL/Builders/FieldBuilder.cs
@@ -55,7 +55,8 @@ namespace GraphQL.Builders
 
         public virtual FieldBuilder<TSourceType, TReturnType> Name(string name)
         {
-            NameValidator.ValidateName(name);
+            if (string.IsNullOrEmpty(name))
+                NameValidator.ValidateName(name);
 
             FieldType.Name = name;
             return this;
--- a/src/GraphQL/Types/Composite/ComplexGraphType.cs
+++ b/src/GraphQL/Types/Composite/ComplexGraphType.cs
@@ -75,7 +75,8 @@ namespace GraphQL.Types
                 }
             }
 
-            if (string.IsNullOrEmpty(fieldType.Name)) NameValidator.ValidateName(fieldType.Name);
+            //check if fieldType.Name has never been set
+            NameValidator.ValidateNameNotNull(fieldType.Name);
 
             if (HasField(fieldType.Name))
             {
--- a/src/GraphQL/Types/Composite/ComplexGraphType.cs
+++ b/src/GraphQL/Types/Composite/ComplexGraphType.cs
@@ -75,7 +75,7 @@ namespace GraphQL.Types
                 }
             }
 
-            NameValidator.ValidateName(fieldType.Name);
+            if (string.IsNullOrEmpty(fieldType.Name)) NameValidator.ValidateName(fieldType.Name);
 
             if (HasField(fieldType.Name))
             {
--- a/src/GraphQL/Types/Fields/FieldType.cs
+++ b/src/GraphQL/Types/Fields/FieldType.cs
@@ -16,7 +16,8 @@ namespace GraphQL.Types
         public string Name
         {
             get => _name;
-            set {
+            set
+            {
                 NameValidator.ValidateNameNotNull(value);
                 _name = value;
             }
--- a/src/GraphQL/Types/Fields/FieldType.cs
+++ b/src/GraphQL/Types/Fields/FieldType.cs
@@ -12,7 +12,15 @@ namespace GraphQL.Types
         private object _defaultValue;
         private IValue _defaultValueAST;
 
-        public string Name { get; set; }
+        private string _name;
+        public string Name
+        {
+            get => _name;
+            set {
+                NameValidator.ValidateNameNotNull(value);
+                _name = value;
+            }
+        }
 
         public string Description { get; set; }
 
--- a/src/GraphQL/Types/GraphTypesLookup.cs
+++ b/src/GraphQL/Types/GraphTypesLookup.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using GraphQL.Conversion;
 using GraphQL.Introspection;
 using GraphQL.Types.Relay;
+using GraphQL.Utilities;
 
 namespace GraphQL.Types
 {
@@ -362,6 +363,7 @@ namespace GraphQL.Types
             if (applyNameConverter)
             {
                 field.Name = NameConverter.NameForField(field.Name, parentType);
+                NameValidator.ValidateName(field.Name);
             }
 
             if (field.ResolvedType == null)
@@ -382,6 +384,7 @@ namespace GraphQL.Types
                 if (applyNameConverter)
                 {
                     arg.Name = NameConverter.NameForArgument(arg.Name, parentType, field);
+                    NameValidator.ValidateName(arg.Name, "argument");
                 }
 
                 if (arg.ResolvedType != null)
--- a/src/GraphQL/Types/QueryArgument.cs
+++ b/src/GraphQL/Types/QueryArgument.cs
@@ -37,7 +37,15 @@ namespace GraphQL.Types
             Type = type;
         }
 
-        public string Name { get; set; }
+        private string _name;
+        public string Name
+        {
+            get => _name;
+            set {
+                NameValidator.ValidateNameNotNull(value, "argument");
+                _name = value;
+            }
+        }
 
         public string Description { get; set; }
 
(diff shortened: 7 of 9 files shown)


Shane32:
Regardless of this PR, I don't like the fact that the schema initialization is within the `try..catch` block of `DocumentExecuter.ExecuteAsync`.  Initialization of the schema has nothing to do with the execution of the document, or at least it shouldn't -- and won't in 4.0 because `NameConverter` and `Filter` has been moved from `ExecutionOptions` to `Schema`.  And it is impossible for the schema initialization to throw an `ExecutionError` of any variety.

Ideally the schema would be initialized during DI construction of the `Schema`.  For example, if the `DocumentExecuter` received a `ISchemaFactory` (e.g. `SchemaFactory<MySchema>`) then the schema could be constructed and initialized there, outside of the execution of the document, and with thread safety against simultaneous initialization.  The `DocumentExecuter` would simply throw an exception if the schema wasn't initialized.

In this case, now errors due to invalid field names will now be wrapped and returned as GraphQL execution errors rather than thrown directly on the stack.

I suggest we move the initialization of the schema prior to the `try..catch` block inside `ExecuteAsync`.  If not now, then certainly for 4.0.

sungam3r:
> then certainly for 4.0

Of course for 4.0, not now.

src/GraphQL/Types/QueryArguments.cs
```diff
@@ -49,7 +49,8 @@ public void Add(QueryArgument argument)
             if (argument == null)
                 throw new ArgumentNullException(nameof(argument));
 
-            NameValidator.ValidateName(argument.Name, "argument");
+            if (string.IsNullOrEmpty(argument.Name))
```
sungam3r, line 52: ?

src/GraphQL/Types/QueryArguments.cs
```diff
@@ -28,7 +28,7 @@ public QueryArguments(IEnumerable<QueryArgument> list)
             get => ArgumentsList != null ? ArgumentsList[index] : throw new IndexOutOfRangeException();
             set
             {
-                if (value != null)
+                if (value != null && string.IsNullOrEmpty(value.Name))
```
sungam3r, line 31: ?

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -362,6 +363,7 @@ private void HandleField(IComplexGraphType parentType, FieldType field, TypeColl
             if (applyNameConverter)
             {
                 field.Name = NameConverter.NameForField(field.Name, parentType);
+                NameValidator.ValidateName(field.Name);
```
sungam3r, line 366: just move this check into setter

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -382,6 +384,7 @@ private void HandleField(IComplexGraphType parentType, FieldType field, TypeColl
                 if (applyNameConverter)
                 {
                     arg.Name = NameConverter.NameForArgument(arg.Name, parentType, field);
+                    NameValidator.ValidateName(arg.Name, "argument");
```
sungam3r, line 387: just move this check into setter

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -55,7 +55,8 @@ private FieldBuilder(EventStreamFieldType fieldType)
 
         public virtual FieldBuilder<TSourceType, TReturnType> Name(string name)
         {
-            NameValidator.ValidateName(name);
+            if (string.IsNullOrEmpty(name))
```
sungam3r, line 58: ?

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -75,7 +75,7 @@ public virtual FieldType AddField(FieldType fieldType)
                 }
             }
 
-            NameValidator.ValidateName(fieldType.Name);
+            if (string.IsNullOrEmpty(fieldType.Name)) NameValidator.ValidateName(fieldType.Name);
```
sungam3r, line 78: ?

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -107,7 +107,8 @@ public ConnectionBuilder<TSourceType> Bidirectional()
 
         public ConnectionBuilder<TSourceType> Name(string name)
         {
-            NameValidator.ValidateName(name);
+            if (string.IsNullOrEmpty(name))
+                NameValidator.ValidateName(name);
```
sungam3r, line 111: No need keep it here, just move into setter.

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: That’s impossible. Or at least it entirely defeats the purpose of this PR. The NameConverter has not been applied when the setter runs, and nor has the field or argument been added to a schema.

src/GraphQL/Types/QueryArguments.cs
Shane32: The only check left here (which I can move to the setter) is a null/empty check. Ditto everywhere else. The name can’t be tested as it hasn’t yet had the NameConverter applied

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: I am talking about `NameValidator.ValidateName` check, not `NameConverter.NameForArgument` or I don't understand. What are we trying to achieve? Guaranteed validation of names? If so it is not needed to call `NameValidator.ValidateName` anywhere else but inside setter.

> The NameConverter has not been applied when the setter runs

?

`arg.Name = NameConverter.NameForArgument(arg.Name, parentType, field);` means apply converter and call setter, does not ?

src/GraphQL/Types/QueryArguments.cs
sungam3r: > The name can’t be tested as it hasn’t yet had the NameConverter applied

Why? I see no reason not to test name. Really.

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: That is not the goal. The goal is to validate after the NameConverter has been applied, **and not before**.

Shane32: If the check is in the setter, it gets checked before and after.

src/GraphQL/Types/QueryArguments.cs
Shane32: See summary of issue #1840 . The goal is to allow invalid names for a field or argument and specify a name converter that patches them up. Validation occurs after, and not before, the name converter has been applied. So if the schema has names like “my-field” they could be globally translated to “myField”

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: >  The goal is to validate after the NameConverter has been applied, and not before.

Not before, let's say, OK. But checking inside setter covers both cases.

sungam3r: What's the point in these checks at all, if you can always just take any field and directly change its name to invalid value at any time?

src/GraphQL/Types/QueryArguments.cs
sungam3r: Ah, I completely forgot about it. OK, but let's do this in another way. I propose to move check in setters but allow to `overwrite` it with user supplied one:
```c#
public static class NameValidator
    {
        private const string RESERVED_PREFIX = "__";
        private const string NAME_RX = @"^[_A-Za-z][_0-9A-Za-z]*$";

        public static Action<string, string> NameValidation = ValidateName;  // <---------------------------- ADDED, use it everywhere across repo instead the call to static ValidateName

        public static void ValidateName(string name, string type = "field")
        {
            if (string.IsNullOrWhiteSpace(name))
            {
                throw new ArgumentOutOfRangeException(nameof(name),
                    $"A {type} name can not be null or empty.");
            }

            if (name.Length > 1 && name.StartsWith(RESERVED_PREFIX, StringComparison.InvariantCulture))
            {
                throw new ArgumentOutOfRangeException(nameof(name),
                    $"A {type} name: {name} must not begin with \"__\", which is reserved by GraphQL introspection.");
            }
            if (!Regex.IsMatch(name, NAME_RX))
            {
                throw new ArgumentOutOfRangeException(nameof(name),
                    $"A {type} name must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but {name} does not.");
            }
        }
    }
```

sungam3r: It will be backward compatible change :)

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: We can write another PR for that issue. But most of the codebase relies on the fact that people don’t make changes to the schema after it has been initialized. And this PR covers that scenario, and even moreso than the current codebase.

Shane32: > > The goal is to validate after the NameConverter has been applied, and not before.
> 
> Not before, let's say, OK. But checking inside setter covers both cases.

Checking inside setter covers before AND after. Not ‘after and not before’, which is what we are after here.

sungam3r: So it might be worth changing the goal ?

> We can write another PR for that issue.

I don't see the point in doing this because the proposed changes are very small. The new PR will remove some of the edits of this one. Why do we need this?

src/GraphQL/Types/QueryArguments.cs
Shane32: The proposed change is already backwards compatible and does not introduce new api. Any code that wouldn’t run before still won’t run at all.  Any code that ran before will still run correctly. 

I think the library should enforce a valid schema. Rewriting the name validation check doesn’t seem to cover that. 

Perhaps if this new property was used by the setters and the built in method was used by the schema initialization (and type name setters, etc) — but that seems confusing.

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: See this https://github.com/graphql-dotnet/graphql-dotnet/pull/1961#discussion_r513394623

src/GraphQL/Types/QueryArguments.cs
Shane32: Or a simple static Boolean to skip pre-initialization name checks.

Shane32: > The proposed change is already backwards compatible and does not introduce new api. Any code that wouldn’t run before still won’t run at all. Any code that ran before will still run correctly

With the exception of a **bug** fix: NameConverters could have been used to generate invalid names inside a schema. Now they cannot.

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: I don’t mind moving checks into the setters, if it fits with the goal of this PR. I would normally keep separate PRs to fix separate issues. Personally I think moving the checks into the setters has little practical value over the existing design, considering the design of the rest of the codebase (nothing is protected against runtime changes).

src/GraphQL/Types/QueryArguments.cs
sungam3r: > I think the library should enforce a valid schema. 

Yes. But now (and even after your edits) this will not be the case. In addition, we have already come to the understanding that we both want to make it possible to use non-standard (actually invalid) names if someone **really** needs it. And since it is so that it should still be possible to bypass the check somewhere and somehow. That is why I propose not to copy this check throughout the code, which is fraught with errors when one more place for setting the property is added, but to place it in one place. Actually not event in one place since there are many types with name (`INamedType` interface). Now I see check for `type`, `argument`, `enum`. That is, it is still possible to set an invalid name for the directive, for example.

I am trying to convey my idea that it is pointless to try to solve only the local problem indicated in the issue. We need to look at it more broadly and do a centralized check for all named types **inside** these types. No need to scatter checks around codebase. It's error prone, it's hard to maintain.

sungam3r: > With the exception of a bug fix: NameConverters could have been used to generate invalid names inside a schema. Now they cannot.

I see no point in it. Why to disallow set invalid names by NameConverters if we still allow to set invalid names from anywhere else ?

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: The current design is far from perfect and will inevitably undergo changes. This is not a reason not to move towards ensuring schema correctness.

src/GraphQL/Types/QueryArguments.cs
Shane32: > We need to look at it more broadly and do a centralized check for all named types inside these types.

I think you need to re-read my code changes again, since this is almost exactly what this PR does.  Now, and unlike previously, the field and argument names are checked in **only one spot** with this PR: `GraphTypesLookup.HandleField`.  It does not matter if you set it in a field builder, or if it's set before or after you've added it to the schema.  It gets checked long after all that's done, when you run your first query against the schema.

I did leave a few null/empty checks in there.  It calls `NameValidatior.ValidateName` only to produce a consistent error message, not because it's actually checking anything.  I can consolidate those checks into the setters.  But they are simply 'bonus checks', and could be completely removed.

sungam3r: I have read it several times and I understand what you are talking about. You are talking about single validation when building the schema. And I'm talking about checking regardless of whether schema was built or not. I also say that moving check into setters will make the code **cleaner**. I really don't like the fact that checks are now scattered throughout the code.

> only to produce a consistent error message

I guessed it, but it was not obvious.

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -12,7 +12,15 @@ public class FieldType : MetadataProvider, IFieldType, IProvideResolvedType
         private object _defaultValue;
         private IValue _defaultValueAST;
 
-        public string Name { get; set; }
+        private string _name;
+        public string Name
+        {
+            get => _name;
+            set {
```
sungam3r, line 19:
```suggestion
            set
            {
```

src/GraphQL/Types/QueryArgument.cs
```diff
@@ -37,7 +37,15 @@ public QueryArgument(Type type)
             Type = type;
         }
 
-        public string Name { get; set; }
+        private string _name;
+        public string Name
+        {
+            get => _name;
+            set {
```
sungam3r, line 44:
```suggestion
            set
            {
```

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -12,7 +12,16 @@ public class FieldType : MetadataProvider, IFieldType, IProvideResolvedType
...
         private IValue _defaultValueAST;
 
-        public string Name { get; set; }
+        private string _name;
+        public string Name
+        {
+            get => _name;
+            set
+            {
+                NameValidator.ValidateNameNotNull(value);
```
sungam3r, line 21: Hmm, why not `ValidateName` ?

src/GraphQL/Types/QueryArguments.cs
Shane32: > I also say that moving check into setters will make the code cleaner. 

I agree, but we must consider that `Name` may never be set

Shane32: > And I'm talking about checking regardless of whether schema was built or not. 

Which is the point of this PR: so that it does not check before the schema is built.

sungam3r: > I agree, but we must consider that Name may never be set

This can be checked when initializing the schema. See `SchemaBuilder.Validate` and comment inside. I've wanted to do something like server-side schema validation for a long time (similar to request validation). It is another issue and placing basic checks inside name setters does not conflict with this.

src/GraphQL/Types/Fields/FieldType.cs
Shane32: It will not fulfill the #1840

src/GraphQL/Types/QueryArguments.cs
Shane32: > In addition, we have already come to the understanding that we both want to make it possible to use non-standard (actually invalid) names if someone **really** needs it.

If we did, that is beyond the scope of this PR.  This PR should **not** allow invalid names just because someone wants to use a custom `NameConverter`.

Shane32:
While attempting to make proposed changes along the lines you've suggested presents another problem: introspection types begin with `__` which is invalid for user code.  We can of course use internal methods to workaround this, but this starts to defeat the purpose of simplifying the codebase.

I am not pushing further changes along those lines to this PR as it is outside the scope of #1840 .  If you'd like to write a separate PR along those lines, we can review it prior to or after this PR.

Shane32:
I updated the PR description to define the scope of this PR.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -319,12 +334,46 @@ public void throws_when_field_name_prefix_with_reserved_underscores(string field
...
             exception.Message.ShouldStartWith($"A field name must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but {fieldName} does not.");
         }
 
+        [Theory]
+        [InlineData("__id")]
+        [InlineData("___id")]
+        [InlineData("i#d")]
+        [InlineData("i$d")]
+        [InlineData("id$")]
+        public void does_not_throw_with_filtering_namevalidator(string fieldName)
```
sungam3r, line 348: namevalidator or nameconverter ?

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -75,7 +75,8 @@ public virtual FieldType AddField(FieldType fieldType)
                 }
             }
 
-            NameValidator.ValidateName(fieldType.Name);
+            //check if fieldType.Name has never been set
+            NameValidator.ValidateNameNotNull(fieldType.Name);
```
sungam3r, line 79: That is, at this point we now postpone the detection of the error?

sungam3r:
> If you'd like to write a separate PR along those lines, we can review it prior to or after this PR.

OK. I will (after). To be honest, I still don't understand why it can't be done right away.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32:
```suggestion
        public void does_not_throw_with_filtering_nameconverter(string fieldName)
```

src/GraphQL/Types/Composite/ComplexGraphType.cs
Shane32: Correct.

sungam3r: Do you think this is good?

Shane32: No, but it gets the job done.  If you'd like to incorporate a better suggestion in your subsequent PR, that would be just fine.  I do think we should enforce a valid schema regardless of whether a NameConverter is used to fix up names.  A separate option to disable this enforcement (breaking spec) would be fine.

See goal 3 of this PR:
> Validate that the updated names are valid during schema initialization (currently does not happen)

Shane32: If you want to push changes to this branch, that's fine too.  I've spent too much time on this already.  But I do want to meet the goals of this PR, regardless of other changes or enhancements.

sungam3r (approved):
OK. Will we take this in 3.1.2? Will you release? I will make another PR on top of 3.1.2

src/GraphQL/Types/Composite/ComplexGraphType.cs
sungam3r: I think that it will be better if I make changes in separate PR after 3.1.2.

Shane32:
I added another test.

> OK. Will we take this in 3.1.2? Will you release? I will make another PR on top of 3.1.2

Whatever you like.  I'm happy to if you want me to.  I'll review the GitHub Action logs and release artifacts either way.

Issue #1949, opened 2020-10-26
Enum name validation ArgumentOutOfRangeException


jhoward321:
## Description

I have an enum (that I don't control) that I need to map to names from an EnumMember attribute on the enum. Which looks like something this.

```
public enum BasicType
  {
    [EnumMember(Value = "file")] File,
    [EnumMember(Value = "folder")] Folder,
    [Obsolete("Kit is no longer valid"), EnumMember(Value = "kit")] Kit,
    [Obsolete("Template is no longer valid"), EnumMember(Value = "template")] Template,
    [EnumMember(Value = "url")] Url,
    [EnumMember(Value = "article")] Article,
    [EnumMember(Value = "email-template-component")] EmailTemplateComponent,
    [EnumMember(Value = "email-template")] EmailTemplate,
  }
```

In order to do this I created a custom enumeration graph type which looks like this. This works for what I want, except when adding the enum value names containing dashes, it fails validation in NameValidator.ValidateName(value.Name, "enum");

The regex supplied does not allow dashes. In order to be compatible with some other apis I need to be able to use the dashes in the name. Is this something that's possible?

```
public class BasicTypeEnum : EnumerationGraphType
    {
        public BasicTypeEnum()
        {
            Name = nameof(BasicType);
            Description = typeof(BasicType).Description();
            DeprecationReason = typeof(BasicType).ObsoleteMessage();
            var enumValues = Enum.GetValues(typeof(BasicType));
            foreach (var value in enumValues)
            {
                var name = Enum.GetName(typeof(BasicType), value);
                var correctedName = typeof(BasicType).GetMember(name).First().GetCustomAttribute<EnumMemberAttribute>()
                    .Value ?? name.ToLowerInvariant();
                var obsoleteReason = typeof(BasicType).GetMember(name).First().GetCustomAttribute<ObsoleteAttribute>()
                    ?.Message;
                AddValue(correctedName, null, value, obsoleteReason);
            }
        }
    }
```

### Steps to reproduce

Create an enumeration graph type for an enum with EnumMember attributes with dashes in name. Code above should reproduce it.

### Expected result

I should be able to have a graphql enum name containing dashes.

### Actual result

ArgumentOutOfRangeException thrown in NameValidator.ValidateName

### Environment

Running graphql v3.0.0.2026

sungam3r:
[...]

PR #1964, opened 2020-10-28
Strict checks for GraphQL names

sungam3r:
See #1840


src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -347,8 +347,10 @@ public void throws_when_field_name_doesnot_follow_spec(string fieldName)
         [InlineData("id$")]
         public void does_not_throw_with_filtering_nameconverter(string fieldName)
         {
+            NameValidator.Validation = (n, t) => { };
             var type = new ObjectGraphType();
             type.Field<StringGraphType>(fieldName);
+            NameValidator.Validation = NameValidator.ValidateDefault;
```
Shane32, line 353: I do not think this solution is a satisfactory answer.  Would the user really set it back after they initialized the graph?  If they did it as shown in the sample here with a scoped schema, multiple threads would be modifying the `NameValidator.Validation` property simultaneously, for sure breaking their code.  If they didn't, the schema would not be properly verified.  And it should be, even when using a custom `NameValidator`.

sungam3r: :)  I guessed that these lines might raise questions. In fact, these calls are only made to save your test, no more. No need from client to set back static property. I called ` NameValidator.Validation = (n, t) => { };` to allow setting _bad_ name in line `  type.Field<StringGraphType>(fieldName);` . Then I set it back to show that your `TestNameConverter` works. It changes field name from _bad_ one to _good_ one. That is all.

Shane32: Well that's why I didn't put it in my PR.

I think:
- The NameConverter needs the ability to convert bad names to good ones
- The schema should be validated after that happens.

These changes invalidate that.

Shane32: I suggest a boolean that disables pre-initialization checks, retaining the checks during initialization.

src/GraphQL/Types/QueryArguments.cs
```diff
@@ -5,6 +5,7 @@
 
 namespace GraphQL.Types
 {
+    // TODO: better to rename QueryArguments and QueryArgument to something like FieldArguments and FieldArgument thought directives have arguments too.
```
joemcbride, line 8: How about `GraphQLArgument`?  It appears that is what the JavaScript project calls it.

Shane32: Regardless of the name, if we do rename these classes in 4.0, I would maintain backwards compatibility, marking the old ones as obsolete.  There are going to be a LOT of references in user code, and it cannot be fixed with just a namespace import.

sungam3r: Maybe. Either way we will be thinking abput it only for 4.0 release, not now.

sungam3r: GraphQLArgument seems to be obvious good candidate.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
sungam3r: > The NameConverter needs the ability to convert bad names to good ones

So it happens.

> The schema should be validated after that happens.

This validation is not enough now until we freeze the whole schema. We now allow to change almost any part of the schema at any time. If we change the design in future, then of course it will be possible to get rid of these checks inside setters.

> I suggest a boolean that disables pre-initialization checks, retaining the checks during initialization.

What does it look like? `NameValidator.Validation = (n, t) => { };` is some sort of boolean already.

sungam3r: In general, now everything works as before, with the difference that there are fewer options to make a error in runtime.

src/GraphQL/Types/QueryArguments.cs
sungam3r: If they remain and are marked as obsolete but are no longer used, they will still be of little use. All the same, the client will not be able to work until he changes all his code. Or I didn't understand your intentions.

Shane32: I meant to leave stubs in there
```csharp
[Obsolete]
public class QueryArgument : GraphQLArgument { }
```
Of course with any necessary constructors that previously existed.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32: The difference is that the boolean doesn't disable the check during schema initialization.

Shane32: An equivalent suggestion would be to leave all the code you wrote as-is, except to change the calls within `HandleField` to call `ValidateDefault` rather than `Validate`.

src/GraphQL/Types/QueryArguments.cs
sungam3r: 👍 👍 👍

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
sungam3r: That is, do you propose to tighten checks?

Shane32: Compared to this current PR, yes.  Compared to current codebase, no.

sungam3r: > except to change the calls within HandleField to call ValidateDefault rather than Validate

I don't understand how to do this yet. `NameValidator` is static 'low-level' class. I saw it as a global switch that users turn on at their own discretion. The library itself doesn't manipulate it.

> I suggest a boolean that disables pre-initialization checks

What can we achieve by disabling pre-initialization checks?

Shane32: > What can we achieve by disabling pre-initialization checks?

Retaining proper validation of the initialized schema, while allowing a NameConverter to fix up broken names.  In my mind, this is half of the goal of #1840 .  Allowing a user to bypass ALL validation doesn't fit with that.

Separately if we have some way of bypassing ALL validation (e.g. for #1451) … ok … but we shouldn't get rid of our normal validation checks just because we want to add that feature.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -363,7 +362,6 @@ private void HandleField(IComplexGraphType parentType, FieldType field, TypeColl
             if (applyNameConverter)
             {
                 field.Name = NameConverter.NameForField(field.Name, parentType);
-                NameValidator.ValidateName(field.Name);
```
Shane32, line 366: This removes validation of names after they have been converted by the NameConverter.  This is the whole point of #1840 -- "Validate names after they are mapped with INameConverter".  So basically this PR reverts #1840

sungam3r: Why? Validation happens inside `field.Name` setter. This is the point - to drill external validation from many places and keep it in one place.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
sungam3r: To be honest I think that NameConverter never used for _fix up broken names_. It is used to follow conventions - PascalCase, camelCase, snake_case and so on.

> Allowing a user to bypass ALL validation doesn't fit with that.

I share this concern. But we have already found out that we want to give the opportunity to do so in exceptional cases. I can specifically write a separate paragraph in the generated documentation.

Regarding #1451 . Two new introspection types bypass validation by means of internal methods. The same methods have long been used for other types of instrospection. I'm talking about `internal void GraphType.SetName(string name, bool validate)`.

> but we shouldn't get rid of our normal validation checks 

It seems that we do not get rid of them actually. What is _normal validation checks_ in your opinion? I honestly don't see anything _abnormal_ in the checks if this PR. Everything (checks, name converters) works by default. If (and only if) someone wants to bypass scrict checks for names, well ... OK, he has the possibility to do so _after being warned in advance_.

Although I have a suspicion that I see only part of the picture, therefore I do not understand your concerns.

src/GraphQL/Types/GraphTypesLookup.cs
Shane32: So that schema has been validated after the specified name converter has been applied.  So that a valid schema is enforced, while allowing a custom name converter.

> to drill external validation from many places and keep it in one place.

No, it is in **only one place now**.  You're moving it from one place (`GraphTypesLookup.HandleField`) and moving it to another place (`Field.Name` / `Argument.Name`).  And this new place does not properly validate the schema if someone uses a custom name converter.  You are creating a bug; the very same bug that #1840 was written to address.

I will grant you that the current place is not ideal for debugging field naming issues.  However what's the worst thing that can happen?  Someone will try to start their program, and it will still tell them what field is incorrectly named.  We can make the error a bit more specific to say which graph type contains the invalid field/argument name.  It simply won't highlight the invalid line of code in the debugger.  But it won't let an invalid schema be generated.

Now if the one check is insufficient for the needs of the program, then we can add another.  But if you can only pick one spot to put the check, then it's in the proper location now.  Otherwise, we are right back to where we started before #1840 was merged -- it's being checked in the wrong spot.  Which is why #1840 was titled "Validate names **after** they are mapped with INameConverter".  #1840 did not say anything about consolidating the number of checks, when in fact, that's what it did do.  It **was** scattered in multiple places in the code, like the field builders and stuff, and **now** it is in only one place: `GraphTypesLookup.HandleField`.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32: I'm not sure what there is to explain.  Most likely, I see my viewpoint as clearly as you see yours - lol.  It seems like we are each focusing on completely different issues here, and likely have some different viewpoints on those issues.

1. You are trying to allow a way to bypass validation for invalid names

I don't have a problem with this.  But I think that bypassing validation has nothing to do with moving the checks from `GraphTypesLookup.HandleField` to `FieldType.Name` and `QueryArgument.Name`.  If you want to write a PR for bypassing validation, then do so.  But if the validation is not bypassed, then the name should be checked after the `INameConverter` is applied, and not before.

2. You are trying to consolidate checks

Conceptually I agree, but they are already consolidated.  Names are only validated within `HandleField.GraphTypesLookup`.  There are basic null checks where it makes sense, of course.  I don't see how you can consolidate something that's already consolidated.

Futher, if we need more than one check to fulfill the criteria here, so be it.  I wouldn't sacrifice code quality for a couple line of code.

3. We would both prefer if the checks were within the property setters

I agree.... if it can be done within the confines of the other criteria here.

4. I'm trying to enforce a valid schema (unless someone specifies otherwise, if we provide that option), regardless of whether they use a `INameConverter` or not.

I think we have real, valid cases of people wanting to use a name converter to fix up names that may be invalid.  See #1840 and https://github.com/graphql-dotnet/graphql-dotnet/issues/1949#issuecomment-717328587 .  And, it just makes sense.  No validation on names after being converted - like really?  The name converter is a main feature of the program, used by almost every schema out there.  True, the default one won't generate invalid names (if not fed invalid names), but this isn't some interface that was haphazardly added to allow some crazy scenario.

It would seem that you don't feel that this is a valuable feature / bugfix.  At least, you would sacrifice this for points 2 and 3 above.

Shane32:
@sungam3r At the end of the day, it probably doesn't matter much one way or the other.  However, if you merge this PR as it is now, you must reopen #1840 since that bug will again exist.  **Field and argument names will not be validated after, and not before, the INameConverter is applied.**  Is that the end of the world?  No.  Do I think it's the right answer?  No.  But it'll work.  Just reopen the bug report.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32: > Regarding #1451 . Two new introspection types bypass validation by means of internal methods. The same methods have long been used for other types of instrospection. I'm talking about `internal void GraphType.SetName(string name, bool validate)`.

Unless it can be implemented in a way that does not break spec, I think we should explore a way to introduce this as a separate library.  This would also be a valuable sample of how third-parties could add similar out-of-spec features to the core library.  So perhaps we should simply expose `SetName` as a public member, so anyone can set a name and bypass validation.  (Would require changes to `HandleField` as it is now, but not with this PR.)

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: > No, it is in only one place now. 

OK. Let's say so, but now I still can do that: `myField.Name = "$-#!"`

> You're moving it from one place (GraphTypesLookup.HandleField) and moving it to another place (Field.Name / Argument.Name).

OK. From one to another. But that new place handles all set cases. It doesn't matter when initializing the schema, before or after it.

> And this new place does not properly validate the schema if someone uses a custom name converter. 

What do you mean '_does not properly validate_' ? An exception from setter occurring at the earliest stages of schema building when doing something like `myField.Name = "$-#!"`? In this case, one can use `NameValidator.Validation` delegate. In general, it is better to fix the reason why such `$-#!` names are set so that the error does not occur.

That is, there are two situations when names are converted - when a 'good' name is converted to some other 'good' name (applying conventions) and when a 'bad' name is converted to a 'good' one (I have not encountered this in practice). I assume that you are only talking about the bug #1840 for the second case. For the first case, everything works as before. Right?

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
sungam3r: Regarding https://github.com/graphql-dotnet/graphql-dotnet/issues/1949#issuecomment-717328587 . `NameConverter` does not help since it works with arg names and field names, not type names and enum value names.

sungam3r: I'm OK with periodically merging master into my #1451 fork. It is difficult to move it into a separate library.

sungam3r:
@Shane32 I authorized in lgtm.com and can see alerts:
изображение

sungam3r:
Well, the single error I can see is false positive :( 
изображение

Shane32:
I think what the message means is that the collection is never read from, so it can be removed.

I didn’t write that file and don’t use it in my own code, so I don’t know what the purpose of that variable originally was. Looks like it can be safely deleted to me. 

How do I see the code analysis report?

sungam3r:
> I think what the message means is that the collection is never read from, so it can be removed.

Right. It is true alert. I was wrong.

> How do I see the code analysis report?

sungam3r:
@Shane32  I suggest to review this PR again.

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: @Shane32 Are you still sure about your statement https://github.com/graphql-dotnet/graphql-dotnet/pull/1964/files#r514275928 ? Validation occurs in setters.

Shane32:
I remember perfectly what we discussed.  If you merge this, you break the issue that was reported in #1840 (_Validate names after they are mapped with INameConverter_) and fixed in #1961 (_Move ValidateName calls after NameConverter is applied_), namely that names are checked **after and not before** they are processed by the name converter.  Your solution requires either validating names both before and after, or never.  There is no option for "after and not before".  I'm okay with "both before and after" as long as there is also a "after and not before" option as well.  I recall suggesting compromises which you did not seem to find satisfactory.  I'm happy to reiterate them if you like.

I do not know why you reopened #1840.  It was fixed in #1961.  It need only be opened again if you break the issue by merging this PR as-is.  If you have identified another issue, let's open a new issue for it.

I do not believe this PR should be merged as-is.  I'm happy to discuss alternatives.

sungam3r:
>  If you merge this, you break the issue that was reported in #1840

Are you sure? Can you please add a test into my branch that will show an error?


> Your solution requires either validating names both before and after, or never.

I understand that there are no more concepts _before_ and _after_ in my solution. Names are checked at the time they are set. However, of course, this check can be turned off by setting a global validation delegate.

>  I'm okay with "both before and after" and "never" as long as there is also a "after and not before" option as well.

I'm OK to add "after and not before" option but how?

Shane32:
> > If you merge this, you break the issue that was reported in #1840
> 
> Are you sure? Can you please add a test into my branch that will show an error?

There already exists a pair of tests that I added/updated in #1961.  You changed one by adding this line of code to only one of the tests:

```cs
NameValidator.Validation = (n, t) => { };
```

So now the pair of tests does not demonstrate what it was designed to anymore because they do not have the same initial setup, and the pair of tests (as a pair) is pointless.

If you like, I can update the codebase and combine the two tests into a single method that requires the functionality described in #1840 - validating a name after and not before the NameConverter is applied.

> I understand that there are no more concepts _before_ and _after_ in my solution. Names are checked at the time they are set. However, of course, this check can be turned off by setting a global validation delegate.
> 
> > I'm okay with "both before and after" and "never" as long as there is also a "after and not before" option as well.
> 
> I'm OK to add "after and not before" option but how?

In some fashion separate the validation when set vs when the schema is validated.  For example, you can use the global validation delegate in the setters, but a separate boolean to disable secondary validation (using the official naming rules) when the schema is validated.  It could be `bool ISchema.EnforceValidGraphQlNames` and default to `true`.  So even with the delegate changed to `(name) => true` names are validated against the official spec when the schema is built unless you explicitly turn that off.

There are other ways to do so; that was merely one example.

sungam3r:
> I'm okay with "both before and after" and "never" as long as there is also a "after and not before" option as well.

I think now we both got what we wanted. Please review again.

src/GraphQL/Utilities/NameValidator.cs
```diff
@@ -3,12 +3,52 @@
...
+        /// 'enum-member'. Only change it when absolutely necessary.
+        /// </summary>
+        public static Action<string, string> ValidationOnSchemaInitialize = ValidateDefault;
+
+        /// <summary>
+        /// Validates a specified name.
+        /// </summary>
+        /// <param name="name">GraphQL name.</param>
+        /// <param name="type">Type of element: field, type, argument, enum.</param>
+        public static void ValidateName(string name, string type = "field") => Validation(name, type);
```
Shane32, line 37: Is it wise to have a default value for `type`?  I think not, but am okay either way.  Ditto for `ValidateNameOnSchemaInitialize`.

src/GraphQL/Utilities/NameValidator.cs
```diff
@@ -3,12 +3,52 @@
...
+        /// <param name="name">GraphQL name.</param>
+        /// <param name="type">Type of element: field, type, argument, enum.</param>
+        public static void ValidateNameOnSchemaInitialize(string name, string type = "field") => ValidationOnSchemaInitialize(name, type);
+
+        /// <summary>
+        /// Validates a specified name according to the GraphQL <see href="http://spec.graphql.org/June2018/#sec-Names">specification</see>.
+        /// </summary>
+        /// <param name="name">GraphQL name.</param>
+        /// <param name="type">Type of element: field, type, argument, enum.</param>
+        public static void ValidateDefault(string name, string type)
```
Shane32, line 51: Perhaps `ValidateDefault` could be named something better -- it seems confusing -- but I don't have a better suggestion, and you have clear xml comments.

Shane32 (approved):
I made a few notes, but it looks good to me.

src/GraphQL/Utilities/NameValidator.cs
sungam3r: I don't like it either but PR is targeted to `master` so no breaking changes. Change it later on `develop` if you see fit.

Shane32:
@sungam3r It would seem that this PR is causing a problem with a test on .NET Core 2.2 (in the develop branch, tests run on both 2.2 and 3.1).  Can you check it out and see if you can replicate the problem?

image

sungam3r:
I think it is because race conditions around static field. Test run in parallel.

Shane32:
We need to fix that. It is causing intermittently failing runs during CI checks.

PR #1968, opened 2020-10-29
Fix release workflow

Shane32:
I did some testing.  Pretty sure this is the only change necessary.  Hope so, anyway!


sungam3r (approved):

sungam3r:
I would like to try this for #1969

sungam3r:
Ah, I forgot that this is for publish release, not for test.

PR #1972, opened 2020-10-29
Move reference to System.Buffers into GraphQL.NewtonsoftJson

Shane32:


sungam3r:
Wow! How did we not notice it before ?!

sungam3r (approved):

Issue #653, opened 2018-05-07
Issue Resolving Dependencies in context.GetArgument<T>


Historyman:
I am running into an error when trying to get an argument from a mutation

```namespace Performance.GraphQl
{
    public class PerformanceMutation :ObjectGraphType<object>
    {
        public PerformanceMutation(PerformanceService service, Core core)
        {
            Field<SessionType>(
                "createSession",
                arguments: new QueryArguments(
                    new QueryArgument<NonNullGraphType<SessionInputType>> { Name = "session" }
                ),              
                resolve: context =>
                {
                    var session = context.GetArgument<PerformanceSession>("session"); //System.MissingMethodException: 'No parameterless constructor defined for this object.'
                    var test = new PerformanceSession(core);//this works fine
                    return service.AddSession(test); //As does this.
                }
            );
        }
    }

 }
```

The PerformanceSession Object by my company's design needs a constructor that takes a Core object
```
namespace Performance.Data
{
    public class PerformanceSession : ModelBase
    {
        public PerformanceSession(Core core) : base(core)
        {
        }

        public override IModelPersistenceBase GetPersistence()
        {
            return new PerformanceSessionPersistence(UlineCore);
        }
///OTHER LOGIC///
    }
```

Is having a DI dependent class in the context.GetArguments<T> method not supported?  

Here is my Schmea in case that is needed to help, but I pretty much copied from the StarWars example.
```
namespace Performance.GraphQl
{
    public class PerformanceSchema : Schema
    {
        public PerformanceSchema(IDependencyResolver resolver): base(resolver)
        {
            Query = resolver.Resolve<PerformanceQuery>();
            Mutation = resolver.Resolve<PerformanceMutation>();
            
        }
    }
}
```

And the Resolver
```namespace Performance
{
    public class UnityResolver : IDependencyResolver
    {
        protected IUnityContainer container;

        public UnityResolver(IUnityContainer container)
        {
            if (container == null)
            {
                throw new ArgumentNullException("container");
            }
            this.container = container;
        }

        public object GetService(Type serviceType)
        {
            try
            {
                return container.Resolve(serviceType);
            }
            catch (ResolutionFailedException)
            {
                return null;
            }
        }

        public IEnumerable<object> GetServices(Type serviceType)
        {
            try
            {
                return container.ResolveAll(serviceType);
            }
            catch (ResolutionFailedException)
            {
                return new List<object>();
            }
        }

        public IDependencyScope BeginScope()
[...]

PR #1973, opened 2020-10-29
Subscriptions moved to separate package

### LINKED ISSUE #1971: Moving subscriptions stuff in the separate package
Just an idea. No everyone uses subscriptions but they create additional burden on the main package and clients when dealing with versioning problems:
изображение
изображение

It is theoretically possible to take out all the things necessary for subscriptions in a separate package. This will require a careful approach. In the process, it may turn out that the subscription mechanism is too closely tied to the core of the project and the game is not worth the trouble.

sungam3r:
~DRAFT and a subject of discussion!~

Fixes #1971 

I was pleasantly surprised at how easy everything turned out.

Note that all 'declaration/configuration' types like `IResolveEventStreamContext`, `EventStreamFieldType`, `IEventStreamResolver` stay in place. They act as a facade on top of `System.IObservable<out T>` from netstandard. The implementation on top of 'Reactive stuff' - `SubscriptionDocumentExecuter/SubscriptionExecutionStrategy` - is moved to an external package.


src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.Subscriptions.approved.txt
```diff
@@ -0,0 +1,24 @@
+namespace GraphQL.Execution
+{
+    public class SubscriptionExecutionStrategy : GraphQL.Execution.ParallelExecutionStrategy
+    {
+        public SubscriptionExecutionStrategy() { }
+        public override System.Threading.Tasks.Task<GraphQL.ExecutionResult> ExecuteAsync(GraphQL.Execution.ExecutionContext context) { }
+        protected virtual System.Threading.Tasks.Task<System.IObservable<GraphQL.ExecutionResult>> ResolveEventStreamAsync(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
+    }
+}
+namespace GraphQL
```
sungam3r, line 10: I decided not to use `GraphQL.Subscriptions` namespace.

src/GraphQL.ApiTests/ApiApprovalTests.PublicApi.GraphQL.approved.txt
```diff
@@ -1645,10 +1639,6 @@ namespace GraphQL.Subscription
 {
     public interface IResolveEventStreamContext : GraphQL.Execution.IProvideUserContext, GraphQL.IResolveFieldContext { }
     public interface IResolveEventStreamContext<out TSource> : GraphQL.Execution.IProvideUserContext, GraphQL.IResolveFieldContext, GraphQL.IResolveFieldContext<TSource>, GraphQL.Subscription.IResolveEventStreamContext { }
-    public interface ISubscriptionExecuter
```
sungam3r, line 1648: This interface was not used anywhere, not even in a server project.

src/GraphQL.DataLoader.Tests/QueryTestBase.cs
```diff
@@ -18,7 +18,7 @@ namespace GraphQL.DataLoader.Tests
 {
     public abstract class QueryTestBase : DataLoaderTestBase
     {
-        private readonly IDocumentExecuter executer = new DocumentExecuter();
+        private readonly IDocumentExecuter executer = new SubscriptionDocumentExecuter();
```
sungam3r, line 21: Note that in most tests it was not necessary to make this replacement.

Shane32:
Shouldn't we base this draft off of the `develop` branch for use in 4.0 ?

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -331,7 +331,7 @@ protected virtual IExecutionStrategy SelectExecutionStrategy(ExecutionContext co
             {
                 OperationType.Query => new ParallelExecutionStrategy(),
                 OperationType.Mutation => new SerialExecutionStrategy(),
-                OperationType.Subscription => new SubscriptionExecutionStrategy(),
+                OperationType.Subscription => throw new NotSupportedException($"DocumentExecuter does not support executing subscriptions. You can use SubscriptionDocumentExecuter from GraphQL.Subscriptions package to execute subscriptions."),
```
sungam3r, line 334: It's all :)

src/GraphQL.DataLoader.Tests/QueryTestBase.cs
Shane32: That executer is a test class, right?  It's not part of GraphQL proper.

sungam3r:
> Shouldn't we base this draft off of the develop branch for use in 4.0 ?

Of course. It's just that I'm not sure yet if it's worth doing at all. If someone is interested in this, then of course I will prepare everything properly. So far, this is the very initial draft. Although it seems to be in pretty good shape already.

sungam3r:
I know for sure that at most only one in ten of our microservices uses subscriptions. Therefore, undoubtedly most of them can reference only GraphQL package.

src/GraphQL.DataLoader.Tests/QueryTestBase.cs
sungam3r: `SubscriptionDocumentExecuter` ? It is 'real` class from GraphQL.Subscriptions project. DataLoader.Tests project calls `var result = await ExecuteSubscribeAsync("subscription OrderAdded { orderAdded { orderId } }");` so we need `SubscriptionDocumentExecuter`.

Shane32: Can you point me to the project?  I am not familiar with it.

src/GraphQL.Subscriptions/GraphQL.Subscriptions.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <Description>Document executer for subscriptions</Description>
```
sungam3r, line 5: link

src/GraphQL.DataLoader.Tests/QueryTestBase.cs
sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/pull/1973/files#r514568778

This is **new** project :)

Shane32: Duh.  Obviously I didn't read the PR.... lol

sungam3r: Everything turned out pretty simple.

Shane32:
General comment here:

Maybe we should consider how much we want to split up the main project for 4.0.  For instance, we could easily split the project into the following pieces that may all commonly be used together:

- Core
- Subscriptions
- Serialization (already done)
- Relay
- Data loaders
- SchemaBuilder (for schema-first syntax)
- SchemaPrinter
- Dependency injection

And this ignores the server, authorization, and other commonly used projects.

There's something to be said for having a single nuget reference contain all the "pieces of the puzzle", as well as arguments for having it all in small pieces.  We can also have one project all the typical pieces (similar to how the `AspNetCore.All` package was used for a period of time).

I'm not sure I have a suggestion here.  Just thinking out loud.  I guess the most valuable thing to me would be to have a consolidated documentation base that spans all of the available packages, including for example the server project.

Shane32:
Splitting up the project for the sole purpose of eliminating some dependencies doesn't really seem _that_ valuable to me.

sungam3r:
Agree. By the way, relay already exists as it's own repo - https://github.com/graphql-dotnet/relay and I would like to cut off all relay copy-pasted code from here. Regarding metapackage - we have already talked about it with @joemcbride . I'm OK with that approach.

sungam3r:
By the way with Source Link it will be more easily to debug projects splitted across different repos.

joemcbride (approved):

joemcbride:
While I don't mind the different projects, I think it would be good to move all of the main projects into a single repo. Would make things a lot easier to keep in sync. That may be an unpopular opinion.  I'm also not opposed to keeping them separate.

The main reason the server project was put in a different repo originally was because I wasn't sure how long a maintainer would stick around for it and I didn't want to take that on at the time.  😉 

> I'm not sure I have a suggestion here. Just thinking out loud. I guess the most valuable thing to me would be to have a consolidated documentation base that spans all of the available packages, including for example the server project.

💯

Shane32:
> While I don't mind the different projects, I think it would be good to move all of the main projects into a single repo. Would make things a lot easier to keep in sync.

Actually, I'd agree.  All tests can run together; version numbers can stay in sync and all packages publish a new version together.  It will enforce having a compatible set of projects published together.

I suggest we reorgainze the graphql-dotnet repo and packages as follows:

Main repo includes the following projects:
- GraphQL - stub that references projects marked with a * below
- \* GraphQL.Core - everything that GraphQL is now, minus what's separated below
- \* GraphQL.Subscriptions - this PR
- GraphQL.Serialization.SystemTextJson - already done
- GraphQL.Serialization.NewtonsoftJson - already done
- \* GraphQL.Relay - import other repo and delete duplicate classes from core project
- \* GraphQL.DataLoader - move everything except the `IDataLoaderResult` interface
- \* GraphQL.SchemaBuilder - stuff for schema-first syntax, not necessary for core operation
- \* GraphQL.SchemaPrinter - I'm not sure what this is for; so maybe it's necessary but I don't think so
- \* GraphQL.DependencyInjection - see PR #1571 
- GraphQL.Server.Core
- GraphQL.Server.Transports.AspNetCore
- GraphQL.Server.Transports.AspNetCore.NewtonsoftJson
- GraphQL.Server.Transports.AspNetCore.SystemTextJson
- GraphQL.Server.Transports.Subscriptions.Abstractions
- GraphQL.Server.Transports.WebSockets
- GraphQL.Server.Ui.Altair
- GraphQL.Server.Ui.Playground
- GraphQL.Server.Ui.GraphiQL
- GraphQL.Server.Ui.Voyager
- GraphQL.Server.Authorization.AspNetCore

Then a person would only need to import these for a full feature set:
- GraphQL
- GraphQL.Server.Transports.AspNetCore.SystemTextJson

At this point I'd keep the authorization and conventions project in their own repos.

sungam3r:
I don't like dumping everything into one repo, because then versioning problems arise. All versions change at the same time even if only one project has changed.

sungam3r:
Also, it will be a real challenge when major versions change.

sungam3r:
Just FYI. Good news!
изображение

sungam3r:
Bad news - links do not work in downloaded documentation.

sungam3r:
Another reason why I do not like single (mono) repos is that they are much easier to break the boundaries of isolation between layers. The distinction between repositories creates an additional barrier to such errors.

Shane32:
> Bad news - links do not work in downloaded documentation.

It should be exactly what was uploaded, for better or worse.  Even if there are no link with the hostname, it probably won't work as a download because of absolute links.  We could work on fixing all links to use relative links, if that's even possible.  Maybe there's some other trick I'm unaware of, like using `<base href>` or something.

sungam3r:
PR is ready for review.

Shane32:
I looked over it again just now.  I'm not sure about it.  I like separating removable pieces from the project.  But it feels like half the subscription code is still embedded in the main project.  `EventStreamFieldResolver` hasn't been moved, nor have the `Subscribe` methods on the field builder.  I'm not saying they can or should be, but that subscriptions are integral to the current library's design.  Further, subscriptions are part of the GraphQL specification.  There are a number of other pieces that are much more easily separated, and to various degrees, not so much a part of the specification.  For instance, the `SchemaPrinter` could easily be very cleanly removed.  Federation is not part of the spec.  Even the schema-first syntax, while debatably part of the spec, could be completely separated with almost no negative side-effects.  There's also relay and data loaders.  (Data loaders do require 'IDataLoaderResult' as part of the execution strategy, but it only has one member, it is not exactly specific to data loaders, and all of the data loader logic can be separated.)  Finally, there's the amount of code we are separating.  There's only 3 files, and 2 of them have < 20 lines of code.

So in summary:
- Subscriptions are integral to the complete implementation of the graphql specification
- The current implementation of field builders rely on subscription methods and field types being part of the main project
- The amount of code that is being split off is a very small amount
- There are much cleaner spots to split the codebase

When you add up all the above reasons, it seems to outweigh benefits.  Am I missing something here?

sungam3r:
> Am I missing something here?

Here is my point of view:

> Subscriptions are integral to the complete implementation of the graphql specification
   
The specification does not describe implementation details. I focus on the choice of specific packages. I would not be tied to them.

> The current implementation of field builders rely on subscription methods and field types being part of the main project.

It's OK since all these methods work with **abstractions** on top of `netstandard`'s `IObservable`.

>  The amount of code that is being split off is a very small amount
   
The same as in the case of two serialization projects. We can only be glad that the adapter project turned out to be small.

> There are much cleaner spots to split the codebase

Well, OK. Something must be the first step.

Shane32:
Is there's a logical reason why this particular implementation of the subscription execution strategy might be only applicable for certain users, and other users might have another execution strategy?

sungam3r:
Such situations are possible.

1. Versioning conflicts and transitive dependencies. I have encountered this many times in practice.
2. Personal developer's preferences.
3. In some cases, testing is simplified.

sungam3r:
Perhaps I will be in time, I have one more hour.

Shane32:
I'm not sure I understand.  Do we have any examples or issues where the subscription execution strategy that is supplied is insufficient?  Any at all?

Of course versioning problems would not be possible (between GraphQL and GraphQL.Subscriptions) if we don't split the codebase, so obviously not splitting would be better on that point.  Unless you were talking about something else.  I'm not sure what you were trying to say.

sungam3r:
> Of course versioning problems would **not be** possible (between GraphQL and GraphQL.Subscriptions) if we **don't** split the codebase.

I was talking about the opposite. Imagine there are two versions of System.Reactive - v4 and v5. We are using the v4 version, someone's library is using the v5 version. This person wants to use both libraries (ours and his) in the same application. He gets a version conflict. However, he may not use subscriptions at all. Either we have to release a new major version of the core package with an increased dependency, or he needs to downgrade its version of the dependency. The situation is aggravated if many packages with different owners come into play (transitive dependencies).

Also look at this:
изображение

1 - 59 KB
2 - 3.75 MB  <--
3 - 59 KB

I just don't see the point of burdening the core library with such dependencies, if they can be taken out without much hassle. And it doesn't break the design because like I said before, the library just keeps on working with `IObservable`-based abstractions from `netstandard`.

Shane32:
Ok so the reason is to remove a dependency on System.Reactive.  That's a decent reason.  Have you considered removing the dependency altogether and just copying the small amount of code needed over?  That is what's done in linq2db so they don't have dependencies.  It should be pretty easy; I can check into it if you like.

sungam3r:
> Have you considered removing the dependency altogether and just copying the small amount of code needed over?

Copying where?

sungam3r:
Oh, I get it. That is, copying the Reactive source code.

sungam3r:
Well, I do that sometimes. But I have big suspicions that this will not work here. We'll have to copy too much. You can check if you're interested. Besides, I would not want to be involved in maintaining the copied code later.

Shane32:
> Well, I do that sometimes. But I have big suspicions that this will not work here. We'll have to copy too much. You can check if you're interested. Besides, I would not want to be involved in maintaining the copied code later.

I checked it out.  It's not worth the effort.  It would be easier to maintain different projects for reactive 4.4.3 and reactive 5.0 if we needed to.  (I wouldn't at this time.)  Let's proceed here.

Shane32:
I'll review one more time now that I understand the goal better.

sungam3r:
Probably I should to name the new project `GraphQL.Reactive` by analogy with `GraphQL.NewtonsoftJson` and `GraphQL.SystemTextJson`. The naming of all of these projects reflects implementation.

Shane32 (commented):
Last comment: I think we need to leave SubscriptionExecutionResult in the main project.  If the goal is to simply separate the execution implementation (because it relies on System.Reactive), it would make sense to leave the result class in the main project, even though it's not actually used there.  Then stuff like serializers and transport code can utilize it directly without relying on project containing the implementation of the execution strategy.

sungam3r:
Agree, done.

Shane32:
Thanks. Let me review.

src/GraphQL/GraphQL.csproj
```diff
@@ -16,8 +16,7 @@
     <PackageReference Include="GraphQL-Parser" Version="5.3.3" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
-    <PackageReference Include="System.Reactive.Core" Version="4.4.1" />
-    <PackageReference Include="System.Reactive.Linq" Version="4.4.1" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.7.1" />
```
Shane32, line 19: We should revert the decimal "enhancement" made recently and remove this reference.

Shane32 (approved):

src/GraphQL/GraphQL.csproj
sungam3r: Why so? I would not. There has already been dependency on `System.Runtime.CompilerServices.Unsafe` from `System.Reactive`, so... we are not losing anything. This is a small package.

Shane32: This has nothing to do with this PR -- go ahead and merge this PR -- we can discuss separately.

fdbeirao:
Hello @sungam3r . Sorry to open such an old issue, but I am currently running into this issue after upgrading my Graphql, and I must be missing something here.

I don't see in the examples/samples any usage of `SubscriptionDocumentExecuter` so I don't really understand if I need to add a magic line to my DI configuration to get Subscriptions to work again. I have already added the package reference to `GraphQL.SystemReactive` version `4.5.0` but it doesn't seem to work "out-the-box" and I missing what do I need to do in my codebase to "use SubscriptionDocumentExecuter from GraphQL.Subscriptions package to execute subscriptions".

Thank you in advance 🙏🏻

Shane32:
@fdbeirao I'd check in your startup if you are registering the DocumentExecuter service; if so, change it to SubscriptionDocumentExecuter.  If it's not registered as an IDocumentExecuter, you might need to also change your middleware to request SubscriptionDocumentExecuter from DI.  If you wrote your own DocumentExecuter (typically to override the SelectExecutionMethod), then simply change it to inherit from SubscriptionDocumentExecuter and that might do the trick.  If you still are at a loss, post a new issue along with your startup.cs file and the list of NuGet packages you're importing.  Thanks!

fdbeirao:
@Shane32 thank you for the quick reply. As far as I understand I have a somewhat "vanilla" situation. As far as I understand, I didn't explicitly register any `DocumentExecutor`. Definitely didn't write my own, that is for sure :)

Where in the [samples](https://github.com/graphql-dotnet/server/blob/master/samples/Samples.Server/Startup.cs) are we registering the document executor?

If my codebase looked like that sample startup, how would I go about getting subscriptions to work then?

I guess if that sample code gets upgraded to the latest version of `Graphql-dotnet` it would also run into the same issue as I am having, right?

fdbeirao:
So, if I add `.AddSingleton<IDocumentExecuter>(sp => new SubscriptionDocumentExecuter())` to my `ConfigureServices` then I get the subscriptions back. I was trying to avoid having to do that "myself".

I am still confused on why the `Server` example would work, I'll clone it and try to fire it up locally. 
(15 minutes later, after locally bumping it to graphql `4.5.0`) I at least see the very same error I was expecting:

image

This is pretty much what I was expecting, since I cannot find (documentation or source code) where the magic of picking `SubscriptionDocumentExecuter` would be happening.  So I guess for now I'll need to stick with that `AddSigleton` that I did.

@sungam3r  I guess somewhere in the migration/upgrade document this explicit addition to the DI should be mentioned (if that is the correct way to go).

Once again, thanks for all the help 🙏🏻

cmadh:
I ran into the same problem as @fdbeirao , yesterday. Service throwing 

> DocumentExecuter does not support executing subscriptions. You can use SubscriptionDocumentExecuter from GraphQL.SystemReactive package to handle subscriptions. occurred

Glad I've found this Issue. Adding the IDocumentExecutor with 

`.AddSingleton<IDocumentExecuter>(sp => new SubscriptionDocumentExecuter())` 

fixes this. Without testing I'm quite sure the examples in current master won't work without it. At least someone should add Documentation on how to add the IDocumentExecutor to handle subscriptions to the official docs.

sungam3r:
@fdbeirao  Fixed in #653. Sorry, I had a long break from OSS this year.

PR #1988, opened 2020-11-08
Retain stack traces and unwrap constructor exceptions

Shane32:
This fixes two bugs:
- When using `SingleMethodAccessor`, exceptions are unwrapped but stack traces are not retained.
- When using the `ToObject` extension method (for instance, within `TryGetArgument`), constructor exceptions are not unwrapped.

### DIFF
--- a/src/GraphQL/ObjectExtensions.cs
+++ b/src/GraphQL/ObjectExtensions.cs
@@ -4,6 +4,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.ExceptionServices;
 using GraphQL.Types;
 
 namespace GraphQL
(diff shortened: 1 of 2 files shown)


src/GraphQL/Reflection/SingleMethodAccessor.cs
```diff
@@ -29,7 +30,8 @@ public object GetValue(object target, object[] arguments)
             }
             catch (TargetInvocationException ex)
             {
-                throw ex.InnerException;
+                ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
+                throw ex.InnerException; //necessary only for intellisense
```
sungam3r, line 34:
```suggestion
                return null; // never executed, necessary only for intellisense
```

src/GraphQL/ObjectExtensions.cs
```diff
@@ -107,7 +108,16 @@ string[] MatchSourceKeys(ParameterInfo[] parameters)
...
-            object obj = targetCtor.Invoke(ctorArguments);
+            object obj;
+            try
+            {
+                obj = targetCtor.Invoke(ctorArguments);
+            }
+            catch (TargetInvocationException ex)
+            {
+                ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
+                throw ex.InnerException; //necessary only for intellisense
```
sungam3r, line 119:
```suggestion
                return null; // never executed, necessary only for intellisense

sungam3r (approved):

Issue #1857, opened 2020-09-13
Add xml comments for all public members


Shane32:
I think there should be a simple `<summary>` tag for each public member (all CS1591 build warnings) by the time 4.0 is released.  This issue will track progress as additional documentation is added to 3.x versions.

sungam3r:
This is too general and a big task. It is just impossible to make a PR fixing that issue because of its size and time spent.

Shane32:
This issue is for tracking within the milestone and progress, much like the issue regarding error handling, which we finished over several pull requests.  I'll probably add checkboxes in the description above for the different namespaces to track progress.  I'll post individual PRs for different classes or sets of classes as I have time.

If you also have large goals, I suggest you create similar issues to track progress and visibility from the milestone.

Shane32:
I actually had it outlined already, but the browser refreshed, and I lost it; half hour's work gone.....

sungam3r:
Maybe use [projects](https://github.com/graphql-dotnet/graphql-dotnet/projects) for this? I haven't tried it yet.

Shane32:
We have use projects within our company. I don’t think it’s a fit here. It’s basically another sorting technique like milestones. So if you had multiple teams working in one repo each team can have their own board and monitor progress. It doesn’t work well enough to fully replace other job management software unfortunately. We ended up just using it to group issues by topic (which could be done with tags just as well).

sungam3r:
I see 75 warnings for CS1591 after disabling it in StarWars sample project and benchmarks (#1979). Not so much but not so little, especially if you write thoughtful and useful comments.

Shane32:
I really do want to work on this.

Shane32:
I’m not so concerned about the comments being so very thought out and detailed, but just any sort of comment is 100% improvement. Of course I don’t want to do a shoddy job either.

sungam3r:
I messed something up in my previous comment about warnings. Actually I see ~1600 CS1591 warnings 😞

Shane32:
It seems that methods that inherit xml documentation from the interface method they implement still generate a CS1591 warning.  I'm not sure we want to add `<inheritdoc/>` tags everywhere when VS already infers the doc inheritance.  But unless we do, it will show many more warnings than really exist.  What do you think @sungam3r  ?

sungam3r:
[...]

PR #2000, opened 2020-11-10
Add some xml docs

Shane32:
See #1857

Mostly putting finishing touches on namespaces/classes already nearly complete.

- src/GraphQL/Conversion/ValueConverter.cs 
- src/GraphQL/DataLoader/DataLoaderContextAccessor.cs 
- src/GraphQL/DataLoader/DataLoaderContextExtensions.cs 
- src/GraphQL/DataLoader/DataLoaderDocumentListener.cs 
- src/GraphQL/DataLoader/DataLoaderExtensions.cs 
- src/GraphQL/Execution/ErrorInfoProvider.cs 
- src/GraphQL/Execution/ExecutionError.cs 
- src/GraphQL/Execution/ExecutionOptions.cs

### DIFF
--- a/src/GraphQL/Conversion/ValueConverter.cs
+++ b/src/GraphQL/Conversion/ValueConverter.cs
@@ -154,7 +154,7 @@ namespace GraphQL
         }
 
         /// <summary>
-        /// Tries to convert an object to the specified type and returns a boolean indicating it the conversion was a success.
+        /// Tries to convert an object to the specified type and returns a boolean indicating whether the conversion was successful.
         /// </summary>
         internal static bool TryConvertTo(object value, Type targetType, out object result, Type sourceType = null)
         {
--- a/src/GraphQL/Conversion/ValueConverter.cs
+++ b/src/GraphQL/Conversion/ValueConverter.cs
@@ -133,7 +133,8 @@ namespace GraphQL
         }
 
         /// <summary>
-        /// Returns an object of the specified type and whose value is equivalent to the specified object.
+        /// <para>Returns an object of the specified type and whose value is equivalent to the specified object.</para>
+        /// <para>Throws a <see cref="InvalidOperationException"/> if there is no conversion registered; conversion functions may throw other exceptions</para>
         /// </summary>
         public static T ConvertTo<T>(object value)
         {
@@ -143,7 +144,8 @@ namespace GraphQL
         }
 
         /// <summary>
-        /// Returns an object of the specified type and whose value is equivalent to the specified object.
+        /// <para>Returns an object of the specified type and whose value is equivalent to the specified object.</para>
+        /// <para>Throws a <see cref="InvalidOperationException"/> if there is no conversion registered; conversion functions may throw other exceptions</para>
         /// </summary>
         public static object ConvertTo(object value, Type targetType)
         {
@@ -154,7 +156,8 @@ namespace GraphQL
         }
 
         /// <summary>
-        /// Tries to convert an object to the specified type and returns a boolean indicating whether the conversion was successful.
+        /// <para>If a conversion delegate was registered, converts an object to the specified type and returns <c>true</c>; returns <c>false</c> if no conversion delegate is registered.</para>
+        /// <para>Conversion delegates may throw exceptions if the conversion was unsuccessful</para>
         /// </summary>
         internal static bool TryConvertTo(object value, Type targetType, out object result, Type sourceType = null)
         {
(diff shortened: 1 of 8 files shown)


src/GraphQL/Conversion/ValueConverter.cs
```diff
@@ -132,13 +132,19 @@ static ValueConverter()
...
+        /// </summary>
         public static T ConvertTo<T>(object value)
         {
             object v = ConvertTo(value, typeof(T));
 
             return v == null ? default : (T)v;
         }
 
+        /// <summary>
+        /// Returns an object of the specified type and whose value is equivalent to the specified object.
```
sungam3r, line 146: Consider to add `  /// <exception cref="InvalidOperationException">blabla</exception>` or extend summary to cover this case. This applies to all methods.

src/GraphQL/Conversion/ValueConverter.cs
```diff
@@ -147,6 +153,9 @@ public static object ConvertTo(object value, Type targetType)
             return result;
         }
 
+        /// <summary>
+        /// Tries to convert an object to the specified type and returns a boolean indicating it the conversion was a success.
```
sungam3r, line 157:
```suggestion
        /// Tries to convert an object to the specified type and returns a Boolean indicating whether the conversion was successful.
```

src/GraphQL/DataLoader/DataLoaderContextExtensions.cs
```diff
@@ -6,6 +6,9 @@
 
 namespace GraphQL.DataLoader
 {
+    /// <summary>
+    /// Provides extension functions for retrieving <see cref="IDataLoader"/> implementations via a <see cref="DataLoaderContext"/>
```
sungam3r, line 10:
```suggestion
    /// Provides extension methods for retrieving <see cref="IDataLoader"/> implementations via a <see cref="DataLoaderContext"/>
```

src/GraphQL/DataLoader/DataLoaderExtensions.cs
```diff
@@ -8,6 +8,9 @@
 
 namespace GraphQL.DataLoader
 {
+    /// <summary>
+    /// Provides extension functions useful for data loaders
```
sungam3r, line 12:
```suggestion
    /// Provides extension methods useful for data loaders
```

src/GraphQL/Execution/ErrorInfoProvider.cs
```diff
@@ -7,22 +7,32 @@
...
+        /// <summary>
+        /// Initializes an <see cref="ErrorInfoProvider"/> with a specified set of <see cref="ErrorInfoProviderOptions"/>.
+        /// </summary>
         public ErrorInfoProvider(ErrorInfoProviderOptions options)
         {
             _options = options ?? throw new ArgumentNullException(nameof(options));
         }
 
+        /// <summary>
+        /// Initializes an <see cref="ErrorInfoProvider"/> with a set of <see cref="ErrorInfoProviderOptions"/> indiciated by the specified builder.
```
sungam3r, line 34:
```suggestion
        /// Initializes an <see cref="ErrorInfoProvider"/> with a set of <see cref="ErrorInfoProviderOptions"/> filled out by the specified delegate.
```

src/GraphQL/Execution/ErrorInfoProvider.cs
```diff
@@ -68,6 +79,13 @@ public virtual ErrorInfo GetInfo(ExecutionError executionError)
             };
         }
 
+        /// <summary>
+        /// <para>Returns a list of error codes derived from a specified <see cref="ExecutionError"/> instance.</para>
+        /// <para>
+        /// By default, this returns the <see cref="ExecutionError.Code"/> value if set, along with
+        /// codes generated from the type of the <see cref="Exception.InnerException"/> and their inner exceptions.
```
sungam3r, line 86:
```suggestion
        /// codes generated from the type of the <see cref="Exception.InnerException"/> and all their inner exceptions.
```

src/GraphQL/Conversion/ValueConverter.cs
Shane32: While I agree with you, my goal within #1857 is to add `<summary>` tags to each member, plus enough notes that users can understand the purpose of each method where it isn't obvious.  I think this is achievable, over time, and important.  If I spend time adding tags for parameters, exception, and return values to each member, the work will never get done.  Most parameters, exceptions, and return values are obvious.

Note that in this particular case, `InvalidOperationException` only occurs when there is no conversion registered; but we do not know what exceptions would be thrown if the conversion fails.  I'll add some notes here to the summary tag as it could be rather important when using these particular methods.

sungam3r: > `InvalidOperationException` occurs when there is no conversion registered.

This will be quite enough.

sungam3r (approved):

PR #2016, opened 2020-11-19
[WIP] 'Precompile' an execution tree to optimize performance on large requests

IdeaHunter:
### A little bit of context
I have an app with ~300 of graphql types and using graphql-dotnet and have a couple of places where there is heavy use of union types. I was looking for overall performance of app and noticed that there is big performance impact created by the library in case where a lot of data is returned. The schema is looking like following:
```graphql
 type Query {
    list(id: String!): List
 }
 type List {
        items(query: String, sortFields: [SortField],  groupByFields: [String], fieldKeys: [String]): [Item]
 }
 type Item {
        props:  [FieldValue]
        rights: Crud
 }
union FieldValue =
        NumberFieldValue
        | IntFieldValue
        | PercentageFieldValue
        | StringFieldValue
       """ 
       and 15+ more types like that 
       """
 
```

 It takes ~300ms to return a big table with 500 rows and 20 columns (~10k objects) whenever time spend in sql and app code that process it takes a 36ms and 30ms respectively, which is huge (5 times) difference compared to graphql-dotnet. The digging of code yielded couple of problems:

1. An internal graphql api `ISchema.FindType` and methods around it called too many times
2. Snowballing GC pressure

this pool requests fully addresses (1) and partially addresses (2) and making it easier to solve in future, but please keep in mind its a POC and names of newly added classes are far from being easy to understand 👼 and feedback on them is welcomed.

## Changes and idea behind them
Major objective of this rework was to make sure we prepare all fields and types before an actual request evaluation. It achieved by evaluating all selections sets for all possible (with regard to type system limitations) `IGraphType`s. 

In essence the code is reworked in a way that its 'evaluate' Document in current context of current Variables and Schema in order to produce data structure that is optimized for future repeated use by a `QueryCompilation` class. Also couple of methods that was used at ExecutionHelper moved to `QueryCompilation` because with exception of unit tests they havent being used anywhere else. The classes `CompiledField` and `CompiledNode` was added. `CompiledField` represents a ready-to-resolve object that have info about its original field. `CompiledNode` represents a `CompiledField` resolved for a specific field value type and have information about subfields. `QueryCompilation` is called right before context creation and creates a `CompiledField` for a root node. 

### CPU
The result of changes is substantial (~2.5x times) improvement in execution times , avoiding calling `ISchema.FindType` and relevant `lock()` and `Regex.Replace` each time we evaluate type of an object.

Here is analysis of running query 200x times under CPU profiler:
#### Before CPU
before-cpu
#### After CPU
after-cpu

### Memory & GC pressure
From a memory perspective you may see a noticeable reduction of live objects and substantial decreases of memory allocation of graphql internal objects. You also may have noticed that JIT_NEW from previous pictures looks like it takes more % of time (about 2 times) but since the query is running ~2.5x times faster it actually tells that allocations are slightly decreased. External code is most likely a GC time, so I'm planning to do another PR to solve issues with GC pressure specifically.

Here is analysis of allocation for running query 5x times (cant run more because running on space on NVME)
#### Before memory
before-mem
#### After memory
after-mem

## Tests
Since no new functionality was added, no new tests or documentation has being added

There are only some side changes to questionable unit tests for federation functionality which expects for federated query to return an `__typename` field which wasn't requested in query.

This PR is POC and me was having hard time giving stuff proper names so any feedback on class/members names is welcomed.

### DIFF
--- a/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
+++ b/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
@@ -90,7 +90,6 @@ type User @key(fields: ""id"") {
         }
 
         [Theory]
-        [InlineData("...on User { id }", false)]
         [InlineData("__typename ...on User { id }", false)]
         [InlineData("...on User { __typename id }", false)]
         [InlineData("...on User { ...TypeAndId }", true)]
--- a/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
+++ b/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
@@ -78,7 +78,7 @@ type User @key(fields: ""id"") {
                 }";
 
             var variables = @"{ ""_representations"": [{ ""__typename"": ""User"", ""id"": ""123"" }] }";
-            var expected = @"{ ""_entities"": [{ ""__typename"": ""User"", ""id"" : ""123"", ""username"": ""Quinn"" }] }";
+            var expected = @"{ ""_entities"": [{ ""id"" : ""123"", ""username"": ""Quinn"" }] }";
 
             AssertQuery(_ =>
             {
@@ -221,7 +221,7 @@ type User @key(fields: ""id"") {
                     }
                 }";
 
-            var expected = @"{ ""_entities"": [{ ""__typename"": ""User"", ""id"" : ""1"", ""username"": ""One"" }, { ""__typename"": ""User"", ""id"" : ""2"", ""username"": ""Two"" }] }";
+            var expected = @"{ ""_entities"": [{ ""id"" : ""1"", ""username"": ""One"" }, { ""id"" : ""2"", ""username"": ""Two"" }] }";
 
             AssertQuery(_ =>
             {
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -438,6 +438,32 @@ namespace GraphQL.Builders
         public int? PageSize { get; }
     }
 }
+namespace GraphQL.Compilation
+{
+    public class CompiledField
+    {
+        public CompiledField(GraphQL.Types.FieldType definition, GraphQL.Language.AST.Field field, System.Func<object, bool, GraphQL.Compilation.CompiledNode> resolve = null) { }
+        public GraphQL.Types.FieldType Definition { get; }
+        public GraphQL.Language.AST.Field Field { get; }
+        public System.Func<object, bool, GraphQL.Compilation.CompiledNode> Resolve { get; }
+    }
+    public class CompiledNode
+    {
+        public CompiledNode(GraphQL.Types.IGraphType graphType, System.Collections.Generic.Dictionary<string, GraphQL.Compilation.CompiledField> fields) { }
+        public System.Collections.Generic.Dictionary<string, GraphQL.Compilation.CompiledField> Fields { get; }
+        public GraphQL.Types.IGraphType GraphType { get; }
+    }
+    public class QueryCompilation
+    {
+        public QueryCompilation() { }
+        public static System.Collections.Generic.Dictionary<string, GraphQL.Language.AST.Field> CollectFields(GraphQL.Types.ISchema schema, GraphQL.Language.AST.Variables variables, GraphQL.Language.AST.Document document, GraphQL.Types.IGraphType specificType, GraphQL.Language.AST.SelectionSet selectionSet) { }
+        public static GraphQL.Compilation.CompiledNode Compile(GraphQL.Types.ISchema schema, GraphQL.Language.AST.Document document, GraphQL.Language.AST.Variables variables, GraphQL.Language.AST.Operation operation) { }
+        public static bool DoesFragmentConditionMatch(GraphQL.Types.ISchema schema, string fragmentName, GraphQL.Types.IGraphType type) { }
+        public static GraphQL.Types.FieldType GetFieldDefinition(GraphQL.Types.ISchema schema, GraphQL.Types.IObjectGraphType parentType, GraphQL.Language.AST.Field field) { }
+        public static GraphQL.Types.IObjectGraphType GetOperationRootType(GraphQL.Language.AST.Document document, GraphQL.Types.ISchema schema, GraphQL.Language.AST.Operation operation) { }
+        public static bool ShouldIncludeNode(GraphQL.Types.ISchema schema, GraphQL.Language.AST.Variables variables, GraphQL.Language.AST.Directives directives) { }
+    }
+}
 namespace GraphQL.Conversion
 {
     public class CamelCaseNameConverter : GraphQL.Conversion.INameConverter
@@ -596,7 +622,7 @@ namespace GraphQL.Execution
 {
     public class ArrayExecutionNode : GraphQL.Execution.ExecutionNode, GraphQL.Execution.IParentExecutionNode
     {
-        public ArrayExecutionNode(GraphQL.Execution.ExecutionNode parent, GraphQL.Types.IGraphType graphType, GraphQL.Language.AST.Field field, GraphQL.Types.FieldType fieldDefinition, int? indexInParentNode) { }
+        public ArrayExecutionNode(GraphQL.Execution.ExecutionNode parent, GraphQL.Types.IGraphType graphType, GraphQL.Compilation.CompiledField compiled, int? indexInParentNode) { }
         public System.Collections.Generic.List<GraphQL.Execution.ExecutionNode> Items { get; set; }
         public override object ToValue() { }
     }
@@ -646,6 +672,7 @@ namespace GraphQL.Execution
     {
         public ExecutionContext() { }
         public System.Threading.CancellationToken CancellationToken { get; set; }
+        public GraphQL.Compilation.CompiledNode CompiledRootNode { get; set; }
         public GraphQL.Language.AST.Document Document { get; set; }
         public GraphQL.ExecutionErrors Errors { get; set; }
         public System.Collections.Generic.Dictionary<string, object> Extensions { get; set; }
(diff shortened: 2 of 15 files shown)


sungam3r:
> and `Regex.Replace` each time we evaluate type of an object.

What `Regex.Replace` are you talking about ?

sungam3r:
Also note that you have changed public API surface. Check GraphQL.ApiTests and change approved file.

Shane32:
This is very exciting; thanks for the effort!!  I'll take a look as soon as I have time.

Assuming that this makes breaking changes to the API, we might need to target the `develop` branch (aka target version 4.0)

IdeaHunter:
> What Regex.Replace are you talking about ?

@sungam3r this one which called a lots of (>200k) times due to how type resolution works

> Also note that you have changed public API surface. Check GraphQL.ApiTests and change approved file.

Are there no other special kind of work needed for this kind of change? Also are we happy with class and methods names (because i assume that changing this file would be the last change)? 😇

IdeaHunter:
I commited changes to public api surface.
@Shane32 by the nature of this change this rather should go for next major release (4.0) that be here even if there wasn't any public api impact.

Shane32:
@IdeaHunter I agree.  So let's retarget this PR to the `develop` branch (you/I can do so by clicking Edit at the top of this PR webpage), and you will probably need to merge the develop branch in here and update the apiapprovals file.  Note that updating the apiapprovals file is as easy as abc: (a) run tests (b) delete `.approved` file (c) rename the `.received` file to `.approved`.  This helps in the review process as it provides a consolidated way to review the changes and naming of public API members.

Shane32:
@IdeaHunter also, if you can add a paragraph to the v4 migration doc with some real quick notes about this change, containing, say, what classes were added/removed or what methods were changed, that would be really great.  I'll probably rewrite the migration guide once we actually upgrade to v4, but this helps us remember what changes were made.

https://github.com/graphql-dotnet/graphql-dotnet/blob/develop/docs2/site/docs/migrations/migration4.md

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -78,7 +78,7 @@ public void entity_query()
                 }";
 
             var variables = @"{ ""_representations"": [{ ""__typename"": ""User"", ""id"": ""123"" }] }";
-            var expected = @"{ ""_entities"": [{ ""__typename"": ""User"", ""id"" : ""123"", ""username"": ""Quinn"" }] }";
```
joemcbride, line 81: This is required for federation, so we can't remove this otherwise federation would quit working.

joemcbride: This is the code that modifies the selection to add `__typename` if it does not exist.

https://github.com/graphql-dotnet/graphql-dotnet/blob/f6ad2eacd9e51b08fad6805e9226ec7326e1b5d7/src/GraphQL/Utilities/Federation/FederatedSchemaBuilder.cs#L72

IdeaHunter: Thanks would add this! I was tried to find explanation why would it is needed( to return a field nobody asking about) in docs for federation, but with no luck whatsoever...
The original query doesnt asks for it https://github.com/graphql-dotnet/graphql-dotnet/blob/acda5595ff8046b13f29323a3c1bcf37d8da8ddb/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs#L211-L219

IdeaHunter: https://www.apollographql.com/docs/federation/federation-spec/
Federation spec doesnt impose this on entities, it only requires it for a representations. 
> Any representation without a __typename: String field is invalid.

IdeaHunter: found somebody's sandbox with nodejs Apollo and it does exactly the same - if nobody asks for typename it doesn't returns it
image
Going to ask folks at Apollo to verify if **returning** `__typename` is a must, because what the point of implementing something that neither standard nor actual nodejs implementation has

Shane32: I've never used federation, but I would like to comment that Apollo GraphQL always adds `__typename` requests on each field type so that it can be used to determine proper caching rules.  So the typename field cannot be stripped off of incoming requests.

Shane32: I really don't know if that applies here; just trying to help.

IdeaHunter: @Shane32 it doesnt get stripped when it got asked to return it
see here
https://github.com/graphql-dotnet/graphql-dotnet/blob/acda5595ff8046b13f29323a3c1bcf37d8da8ddb/src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs#L128
its like whole point of GraphQL - not return what not being asked for

Shane32: sounds good. (but I really don't know anything about federation.)

joemcbride: Thanks for looking more closely at this!

> its like whole point of GraphQL - not return what not being asked for

Agreed.  That has me wondering ... I swear that was required for me to get it working ... 🤔 

Perhaps it is this part of the `_Entity` union, which relies upon the `__typename`.

https://github.com/graphql-dotnet/graphql-dotnet/blob/f6ad2eacd9e51b08fad6805e9226ec7326e1b5d7/src/GraphQL/Utilities/Federation/FederatedSchemaBuilder.cs#L159-L171

I think that would get hit when the type of object returned in the resolver is different than the entity name, so that is probably a work-around.

Maybe we should add a validation rule that verifies that `__typename` is added to the representations.

IdeaHunter: > I think that would get hit when the type of object returned in the resolver is different than the entity name, so that is probably a work-around.

I can add an unit test for this and see how it work/isn't work, because current test is more of positive test and in this case we need a negative one to be sure that it works as intended.

> Maybe we should add a validation rule that verifies that __typename is added to the representations
Should it be part of this PR?

Side note - this specific test that we talking about was the only one that wast actually working out of the box after I fixed all silly mistakes in my implementation. The issue was that type choice has to be delayed because this api returns collection of dataloader objects which type cant be figured out right away.
```cs
var gqlToNode = new Dictionary<IObjectGraphType, CompiledNode>();
foreach (var possilbe in abstractType.PossibleTypes)
{
	var fields = CollectFieldsRecursive(schema, document, variables, field.SelectionSet, possilbe);
	var rootNode = new CompiledNode(possilbe, fields);
	gqlToNode.Add(possilbe, rootNode);
}
var defaultType = new CompiledNode(abstractType, new Dictionary<string, CompiledField>());

return (value, isResolved) =>
{
	var objType = abstractType.GetObjectType(value, schema);
	return objType is { } && gqlToNode.TryGetValue(objType, out var foundType)
		? foundType
		: defaultType;
};
```

As you can see this implementation defaults to type that has no fields if type itself is a different one

sungam3r:
@IdeaHunter I'll try to do a review this weekend. I will just note that your remark about the frequent call to Regex interested me and I did a little optimization. See #2020. Please try to bench this PR. I suspect the numbers should change for the better.

Shane32:
@sungam3r I think we need to review this for 4.0.  Also all regular expressions should probably be removed.

sungam3r:
As I see the plan:
1. Retarget to develop and resolve conflicts.
2. Add notes in migration docs or/and some documentation.
2. Run benchmarks again to prove the benefit of this PR.
3. Make a review if improvements are obvious from benchmarks results **or** if we see overall design improvements.

sungam3r:
#2020 was merged into `develop`

IdeaHunter:
@sungam3r  Agree to the plan, thanks for merging your PR, this making job much easier.

> 3. Run benchmarks again to prove the benefit of this PR.

The benchmarks i used to verify fix or there is some other kinds of benchmarks?

IdeaHunter:
Btw i was tried to rebase but master branch was ahead of develop branch and this caused me lot of problems, lets see if this issue is resolved first

Shane32:
@IdeaHunter I copied your code into a new PR #2177 that is based off the develop branch and ran performance tests in our usual manner. It demonstrates better performance with lists and slower with small queries, probably as anticipated. However, when caching is enabled this PR has the potential to make all queries faster. (There are some obstacles to overcome. )

@sungam3r suggested we shelve this feature for now, as we are running out of time to get it into 4.0. 

If you’d like to continue development, I’d suggest figuring out how to make the new caching service benefit from precompiled queries.

IdeaHunter:
@Shane32 is it correct that there is caching feature in 4.0? if it is im going to make it work and use it myself (i have some oversized queries as well on production)
What obstacles do you have? And what can be done to put in 4.0?

Shane32:
Caching was done in #2173 . This provides a way to cache the parsed document so that parsing and validation does not need to occur for repeated queries (even if variables change)

Shane32:
> What obstacles do you have? And what can be done to put in 4.0?

See #2177 .  I will continue the discussion over there.  You may wish to open a new PR based off of that branch, or update this PR with that code.  (However, it is simply your changes applied to the develop branch; I didn't make any other changes.)

IdeaHunter:
Im abandoning and closing this PR because after [discussion](https://github.com/graphql-dotnet/graphql-dotnet/pull/2177#issuecomment-763126609) I figured out I no longer interested in pushing this further.
The reasons are:
1) Initial PR was created in a way that no new **external** behavior is changed, except one we found out with federation and still under discussion if old behavior is bug or not
2) The proposed improvement was to cache what is already being a cache
3) The suggested place to **cache a compilation** is a put it in a document and use **document cache** 
4) The major obstacle when I was doing this PR was ExecutionContext being following antipattern of god object and because of (3) we going to be again on same track
5) Suggested implementation would require extensive amount of new tests because caching compiled query may expose new category of cache related bugs that are hard to test and hard to catch.

PR #2017, opened 2020-11-19
Add migration docs

### LINKED ISSUE #2014: migration 2.4->3.1.3 issue related to EnumerationGraphType
In 2.4. i wanted to enrich some enums with additional fields and had a graphtype which had an Enum as datasource and the same enum value as EnumerationGraphType<T> as field. This worked in 2.4. but in 3.1.3 it throws an InvalidOperationException.

Is this not supposed to be used like this anymore or is it a bug?

image

image

sungam3r:
fixes #2014

### DIFF
--- a/docs2/site/docs/migrations/migration3.md
+++ b/docs2/site/docs/migrations/migration3.md
@@ -47,6 +47,23 @@ This project now requires the .NET Standard 2.0 and breaks compatibility with ap
 .NET Framework 4.6 and earlier. See https://docs.microsoft.com/en-us/dotnet/standard/net-standard for a list of
 frameworks that support .Net Standard 2.0.
 
+### Naming
+
+By default, the `GraphType.Name` property now is initialized in constructor and is the same as the name of the .NET type
+unless the name of .NET type ends in _GraphType_. In this case, the _GraphType_ suffix is truncated and `GraphType.Name`
+is set to this value.
+
+Examples:
+
+| .NET GraphType Name | Default Name |
+|---------------------|--------------|
+| MoneyType           | MoneyType    |
+| MoneyGraphType      | Money        |
+
+You can always set your own type name either in the constructor or externally. The result of this change is that it is no
+longer possible to have Graph Types with unspecified (`null`) or invalid names. This allows to avoid a number of errors
+and incomprehensible behavior at runtime.
+
 ### Dependency Injection
 
 The previous `IDependencyResolver` interface and `FuncDependencyResolver` class have been replaced by the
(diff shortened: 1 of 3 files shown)


docs2/site/docs/migrations/migration3.md
```diff
@@ -47,6 +47,23 @@ This project now requires the .NET Standard 2.0 and breaks compatibility with ap
...
+Examples:
+
+| .NET GraphType Name | Default Name |
+|---------------------|--------------|
+| MoneyType           | MoneyType    |
+| MoneyGraphType      | Money        |
+
+You can always set your own type name either in the constructor or externally. The result of this change is that it is no
+longer possible to have Graph Types with unspecified (`null`) or invalid names. This allows to avoid a number of errors
+and incomprehensible behavior at runtime.
```
Shane32, line 65:
```suggestion
You can always set your own type name either in the constructor or externally.
```
I deleted some text that seems irrelevant.  Whatever you want is fine, however.

Shane32 (commented):
This isn't a breaking change, is it?  All 2.4.x code, which required the name be set in the graph type constructor, should work properly in 3.x.  So I would think this belongs in the FAQ.

sungam3r:
It seems to be a breaking change. Names in version 2.4.0 were set to different values but I don't remember exactly. @dBug404 ?

docs2/site/docs/migrations/migration3.md
sungam3r: I would like to leave a mention of the motivation behind this decision.

dBug404:
It is a breaking change. After the upgrade I have to modify all mutation parameter names which have a class name with suffix GraphType either on the server or the client side.
The client mutation calls include the names of the graphtype for the parameters:
image

src/GraphQL.sln
```diff
@@ -73,7 +73,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "guides", "guides", "{F279F2
 		..\docs2\site\docs\guides\known-issues.md = ..\docs2\site\docs\guides\known-issues.md
 		..\docs2\site\docs\guides\links.md = ..\docs2\site\docs\guides\links.md
 		..\docs2\site\docs\guides\migration.md = ..\docs2\site\docs\guides\migration.md
```
Shane32, line 75:
```suggestion
```

Shane32 (approved):

src/GraphQL.sln
Shane32: This file doesn't exist.

PR #2020, opened 2020-11-20
CollectTypes rework

sungam3r:
Reworked `GraphType.CollectTypes` method.

1. Removes `CollectTypes` from public API surface.
2. Eliminates some `TrimGraphQLTypes` calls. I think we can remove all calls to this method, but I was in no rush.

Relates to #2016. @IdeaHunter Could you please repeat your performance tests with this branch and evaluate the effect obtained.


src/GraphQL/Types/Composite/ListGraphType.cs
```diff
@@ -27,15 +27,6 @@ protected ListGraphType(Type type)
 
         public IGraphType ResolvedType { get; set; }
 
-        public override string CollectTypes(TypeCollectionContext context)
-        {
-            var innerType = context.ResolveType(Type);
-            ResolvedType = innerType;
-            var name = innerType.CollectTypes(context);
-            context.AddType(name, innerType, context);
-            return "[{0}]".ToFormat(name);
```
sungam3r, line 36: This string modification is not actually required as well as one for `NonNullGraphType`.

src/GraphQL/Types/GraphType.cs
```diff
@@ -69,20 +68,7 @@ public string Name
 
         public string DeprecationReason { get; set; }
 
-        public virtual string CollectTypes(TypeCollectionContext context)
-        {
-            if (string.IsNullOrWhiteSpace(Name))
-            {
-                Name = GetType().Name;
```
sungam3r, line 76: I've always been confused by this line of code - side effect when calling `CollectTypes`, setting `Name` property. I guessed that this line is never executed after we merged PR where each type has been assigned a default name in its constructor. I ran tests to check. And so it happened. In any case, I am against the side effects. If for some reason the name is not specified, then let it return `null` and you can then deal with the error instead of hiding it.

src/GraphQL/Types/GraphType.cs
```diff
@@ -69,20 +68,7 @@ public string Name
...
-            }
-
-            return Name;
-        }
-
-        public override string ToString() =>
-            string.IsNullOrWhiteSpace(Name)
-                ? GetType().Name
```
sungam3r, line 84: Same considerations here. We just return the name.

sungam3r:
@Shane32 @joemcbride I believe that `CollectTypes` is really internal thing.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -195,10 +195,9 @@ public IEnumerable<IGraphType> All()
                 }
 
                 IGraphType type;
-                var name = typeName.TrimGraphQLTypes();
                 lock (_lock)
```
IdeaHunter, line 198: @sungam3r I haven't run my test yet, but I already see you removed it from the right place. In ideal world lock next to it should be removed as well because its a disaster for any possible async code.

Shane32:
@sungam3r I was looking at this code quite a bit during my code commenting earlier.  With my cursory look, it seems to me that this class might be able to be a private class within `GraphTypesLookup`, or more likely the methods moved directly into the `GraphTypesLookup` class.  Also the `CollectTypes` method might be able to be moved out of `IGraphType`.

Shane32:
> @Shane32 @joemcbride I believe that `CollectTypes` is really internal thing.

I agree

sungam3r:
I will update PR today.

sungam3r:
@Shane32 @joemcbride  According to your reaction to my https://github.com/graphql-dotnet/graphql-dotnet/pull/2020#issuecomment-731454933, I realized that I was moving in the right direction. `TypeCollectionContext` is now internal. I didn’t run benchmarks to see the performance gains (I just don’t want to waste time). I think that without benchmarks it is clear that there are fewer memory allocations. @IdeaHunter If you are interested, then you can post your measurements. I think this is necessary to estimate the value of #2016 after merging this PR.

Shane32 (changes requested):
Can you merge with `develop` and move the xml comments over with the rest of the code that was moved to a new file? I'm good with these changes otherwise.

Shane32 (approved):

src/GraphQL/Types/GraphTypesLookup.cs
sungam3r: > In ideal world lock next to it should be removed as well because its a disaster for any possible async code.

If it's not on the hot path with high contention, then there is no problem.

PR #2036, opened 2020-12-06
ISchema inherits IProvideMetadata

sungam3r:
@Shane32 Formally it is breaking change since I added a new requirement to `ISchema`. But I don't know anymore if it's worth worrying about this kind of change and postponing it until the next major version. See https://github.com/graphql-dotnet/server/pull/470#issuecomment-739498710

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,6 +4,7 @@
   - src/GraphQL.Dummy/**/*
   - src/GraphQL.Harness.Tests/**/*
   - src/GraphQL.Tests/**/*
+  - '!src/GraphQL.ApiTests/**/*.txt'
 
 "CI":
   - .github/workflows/**/*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,7 +4,7 @@
   - src/GraphQL.Dummy/**/*
   - src/GraphQL.Harness.Tests/**/*
   - src/GraphQL.Tests/**/*
-  - '!src/GraphQL.ApiTests/**/*.txt'
+  - !src/GraphQL.ApiTests/**/*.txt
 
 "CI":
   - .github/workflows/**/*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,10 +1,5 @@
 "test":
-  - src/GraphQL.ApiTests/**/*
-  - src/GraphQL.DataLoader.Tests/**/*
-  - src/GraphQL.Dummy/**/*
-  - src/GraphQL.Harness.Tests/**/*
-  - src/GraphQL.Tests/**/*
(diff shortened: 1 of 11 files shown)


.github/labeler.yml
```diff
@@ -1,9 +1,5 @@
 "test":
-  - src/GraphQL.ApiTests/**/*
-  - src/GraphQL.DataLoader.Tests/**/*
-  - src/GraphQL.Dummy/**/*
-  - src/GraphQL.Harness.Tests/**/*
-  - src/GraphQL.Tests/**/*
+  - any: ['src/GraphQL.ApiTests/**/*', 'src/GraphQL.DataLoader.Tests/**/*', 'src/GraphQL.Dummy/**/*', 'src/GraphQL.Harness.Tests/**/*', 'src/GraphQL.Tests/**/*', '!src/GraphQL.ApiTests/**/*.txt']
```
sungam3r, line 2: Changes in public API are not considered now as changes in tests.

Shane32:
> Formally it is breaking change since I added a new requirement to ISchema. But I don't know anymore if it's worth worrying about this kind of change and postponing it until the next major version. See graphql-dotnet/server#470 (comment)

Tough call here.  Every implementation will have their own implementation of `Schema` defined (at least, for graphtype-first users).  Granted, most will inherit from `Schema`.  Let's assume for the moment that someone has a compiled application, with a schema that inherits from `Schema`.  Are these changes binary-breaking?

If the answer is yes (app won't work), then it seems that we are throwing away our goal (of providing only binary-compatible updates between major versions) just because we have a feature request (albeit a reasonable one).  If we want to change our goals (e.g. only binary-compatible within minor versions), then let's do so, but not simply make exceptions without a better reason.

If the answer is no (app is fine; it is still binary compatible as long as you're inheriting Schema), then I think this breaking change is reasonable, but we should document the change.  I expect it to be pretty unlikely that someone implements their own `ISchema`, and if they are, this is a minor fix for them.

sungam3r:
>  Let's assume for the moment that someone has a compiled application, with a schema that inherits from Schema. Are these changes binary-breaking?

Of course, not.

sungam3r:
> we should document the change

OK. Where and in what form? New migration guide to v3.2? There will be questions about what kind of migration between minor versions.

sungam3r:
@Shane32 Maybe to just mention this "semi-breaking" change in the release notes?

Shane32:
> @Shane32 Maybe to just mention this "semi-breaking" change in the release notes?

Good idea, and I also think it should also be in the 2.x -> 3.x migration guide.  We can note within the paragraph that the change only applies to >= 3.2.

sungam3r:
Done

Shane32 (approved):

Issue #2048, opened 2020-12-13
Refactor ValueConverter usage from ScalarGraphTypes to avoid boxing


sungam3r:
Example:

```c#
 public class DecimalGraphType : ScalarGraphType
    {
        public override object ParseValue(object value) => ValueConverter.ConvertTo(value, typeof(decimal));

        public override object ParseLiteral(IValue value) => value switch
        {
            DecimalValue decimalValue => decimalValue.Value,
            StringValue stringValue => ParseValue(stringValue.Value),
            IntValue intValue => ParseValue(intValue.Value),  <-------- box int to object
            LongValue longValue => ParseValue(longValue.Value), <-------- box long to object
            FloatValue floatValue => ParseValue(floatValue.Value), <-------- box double to object
            BigIntValue bigIntValue => ParseValue(bigIntValue.Value), <-------- box BigInteger to object
            _ => null
        };
    }
```

ValueConverter uses direct explicit cast from one .NET type to other without boxing:
![изображение](https://user-images.githubusercontent.com/21261007/102024255-128a9380-3da2-11eb-8473-e0c75faf4faa.png)

Why is that? Is it possible to directly use explicit type casts in GraphTypes themselves? Surely yes. Do we still need `ValueConverter`?  Yes. Will using casts be a duplication of logic? I doubt it. I propose to use direct casts. See https://github.com/graphql-dotnet/graphql-dotnet/issues/2047#issuecomment-744063619

Shane32:
I agree direct casts would be better here, and provide a performance benefit (as it would eliminate a dictionary lookup).  However, I fail to see how this has any effect on boxing.  `ParseLiteral` returns an `object`.  So it's going to get boxed sooner or later -- with the current code, within `ValueConverter.ConvertTo`, or with the suggested change, within `DecimalGraphType.ParseLiteral`.

sungam3r:
We have typed `int` value in `IntValue`, then we pass it to `ParseValue` as `object`. I am talking about arguments, not about return values.

sungam3r:
Of cource then this boxed value will be unboxed inside ValueConverter.

Shane32:
I see.  This should be easy to fix.

PR #2049, opened 2020-12-13
Unify scalar conversions

### LINKED ISSUE #2047: FloatGraphType arguments can be parsed as integers
I'm running into an issue where a Float is coming through into the `Arguments` dictionary as an `int`. My code expects it to be a `double` and fails on the downcast. It looks like `FloatGraphType` should have

        public override object ParseLiteral(IValue value) => value switch
        {
            FloatValue floatVal => floatVal.Value,
            IntValue intVal => intVal.Value,
            LongValue longVal => longVal.Value,
            _ => null
        };

changed to

        public override object ParseLiteral(IValue value) => value switch
        {
            FloatValue floatVal => floatVal.Value,
            IntValue intVal => ParseValue(intVal.Value),
            LongValue longVal => ParseValue(longVal.Value),
            _ => null
        };

in line with the other numeric scalar types. I can send a PR if we agree this is a bug.

sungam3r:
Fixes #2047 . Also fixes bug when `FloatGraphType.Value` cannot be casted to `float` even when fractional part was specified.
I decided to remove all `Convert.To` calls and use the `checked` operator explicitly. First, this is how our intentions are more clearly expressed. Secondly, it turned out that we left unchecked context (C# uses unchecked by default) in some places. Examples:
```
 Register<sbyte, byte>(value => (byte)value);
 Register<sbyte, ulong>(value => (ulong)value);
```

I didn't add tests for all cases, because it is very expensive. I became convinced that some existing tests started failing when I changed `Convert.To` calls to explicit cast operator. So I added `checked` in every line with downcast.

### DIFF
--- a/src/GraphQL/Conversion/ValueConverter.cs
+++ b/src/GraphQL/Conversion/ValueConverter.cs
@@ -57,58 +57,58 @@ namespace GraphQL
             Register<DateTimeOffset, DateTime>(value => value.UtcDateTime);
             Register<TimeSpan, long>(value => (long)value.TotalSeconds);
 
-            Register<int, sbyte>(value => { checked { return (sbyte)value; } });
-            Register<int, byte>(value => { checked { return (byte)value; } });
-            Register<int, short>(value => { checked { return (short)value; } });
-            Register<int, ushort>(value => { checked { return (ushort)value; } });
+            Register<int, sbyte>(value => checked((sbyte)value));
+            Register<int, byte>(value => checked((byte)value));
+            Register<int, short>(value => checked((short)value));
+            Register<int, ushort>(value => checked((ushort)value));
             Register(typeof(int), typeof(bool), value => Convert.ToBoolean(value, NumberFormatInfo.InvariantInfo).Boxed());
-            Register<int, uint>(value => { checked { return (uint)value; } });
+            Register<int, uint>(value => checked((uint)value));
             Register<int, long>(value => value);
-            Register<int, ulong>(value => { checked { return (ulong)value; } });
+            Register<int, ulong>(value => checked((ulong)value));
             Register<int, BigInteger>(value => new BigInteger(value));
             Register<int, double>(value => value);
             Register<int, float>(value => value);
             Register<int, decimal>(value => value);
             Register<int, TimeSpan>(value => TimeSpan.FromSeconds(value));
 
-            Register<long, sbyte>(value => { checked { return (sbyte)value; } });
-            Register<long, byte>(value => { checked { return (byte)value; } });
-            Register<long, short>(value => { checked { return (short)value; } });
-            Register<long, ushort>(value => { checked { return (ushort)value; } });
-            Register<long, int>(value => { checked { return (int)value; } });
-            Register<long, uint>(value => { checked { return (uint)value; } });
-            Register<long, ulong>(value => { checked { return (ulong)value; } });
+            Register<long, sbyte>(value => checked((sbyte)value));
+            Register<long, byte>(value => checked((byte)value));
+            Register<long, short>(value => checked((short)value));
+            Register<long, ushort>(value => checked((ushort)value));
+            Register<long, int>(value => checked((int)value));
+            Register<long, uint>(value => checked((uint)value));
+            Register<long, ulong>(value => checked((ulong)value));
             Register<long, BigInteger>(value => new BigInteger(value));
             Register<long, double>(value => value);
             Register<long, float>(value => value);
             Register<long, decimal>(value => value);
             Register<long, TimeSpan>(value => TimeSpan.FromSeconds(value));
 
-            Register<BigInteger, sbyte>(value => { checked { return (sbyte)value; } });
-            Register<BigInteger, byte>(value => { checked { return (byte)value; } });
-            Register<BigInteger, decimal>(value => { checked { return (decimal)value; } });
-            Register<BigInteger, double>(value => { checked { return (double)value; } });
-            Register<BigInteger, short>(value => { checked { return (short)value; } });
-            Register<BigInteger, long>(value => { checked { return (long)value; } });
-            Register<BigInteger, sbyte>(value => { checked { return (sbyte)value; } });
-            Register<BigInteger, ushort>(value => { checked { return (ushort)value; } });
-            Register<BigInteger, uint>(value => { checked { return (uint)value; } });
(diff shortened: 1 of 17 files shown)


sungam3r:
#2048 may be done in this PR too. @Shane32 ?

src/GraphQL/Types/Scalars/LongGraphType.cs
```diff
@@ -7,7 +7,7 @@ public class LongGraphType : ScalarGraphType
         public override object ParseLiteral(IValue value) => value switch
         {
             LongValue longValue => longValue.Value,
-            IntValue intValue => (long)intValue.Value,
+            IntValue intValue => ParseValue(intValue.Value),
```
sungam3r, line 10: Just to unify all scalars.

sungam3r: Could/should be reverted for #2048 .

Shane32:
Regarding #2048, I would suggest that we include those changes in this PR, using value casts rather than performing dictionary lookups within the ValueConverter for these conversions.

src/GraphQL/Conversion/ValueConverter.cs
```diff
@@ -57,79 +57,84 @@ static ValueConverter()
             Register<DateTimeOffset, DateTime>(value => value.UtcDateTime);
             Register<TimeSpan, long>(value => (long)value.TotalSeconds);
 
-            Register<int, sbyte>(value => Convert.ToSByte(value, NumberFormatInfo.InvariantInfo));
-            Register<int, byte>(value => Convert.ToByte(value, NumberFormatInfo.InvariantInfo));
-            Register<int, short>(value => Convert.ToInt16(value, NumberFormatInfo.InvariantInfo));
-            Register<int, ushort>(value => Convert.ToUInt16(value, NumberFormatInfo.InvariantInfo));
+            Register<int, sbyte>(value => { checked { return (sbyte)value; } });
```
Shane32, line 60:
```suggestion
            Register<int, sbyte>(value => checked((sbyte)value));
```
I would suggest using the `checked` keyword rather than the `checked` block.

See: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/checked

src/GraphQL/Types/Scalars/FloatGraphType.cs
```diff
@@ -9,8 +9,10 @@ public class FloatGraphType : ScalarGraphType
         public override object ParseLiteral(IValue value) => value switch
         {
             FloatValue floatVal => floatVal.Value,
-            IntValue intVal => intVal.Value,
-            LongValue longVal => longVal.Value,
+            IntValue intVal => ParseValue(intVal.Value),
```
Shane32, line 12:
```suggestion
            IntValue intVal => (double)intVal.Value,
```
I suggest the above and similar for performance to eliminate a dictionary lookup.  Note that for a widening conversion like this one, `checked` is not necessary, but nor should it create a performance impact as the compiler would know that.  Alternate syntax:
```suggestion
            IntValue intVal => checked((double)intVal.Value),
```

Shane32:
I don't see any tests to confirm that it will convert a literal of type decimal (e.g. "9223372036854775808") to a FloatValue of type double.  Are we not adding that?

src/GraphQL/Conversion/ValueConverter.cs
sungam3r: I read the same page and didn't notice it 🤦

Shane32:
> I didn't add tests for all cases, because it is very expensive. I became convinced that some existing tests started failing when I changed `Convert.To` calls to explicit cast operator. So I added `checked` in every line with downcast.

I'm not really worried about test cases for all the overflow scenarios, etc.  I just feel that when there's an explicit issue/bug brought to our attention, we should make a point to add a test for the fix, to help ensure we never experience that bug again.  In this case, we should add a test to ensure that integer literals interpreted as FloatGraphType values return objects of type `double`, not `int`.  There is no new test for this, as far as I can tell.

sungam3r:
In fact, I noticed that the situation is a little more complicated. See for example how ByteGraphType handles IntValue.
```
  public class ByteGraphType : ScalarGraphType
    {
        public override object ParseLiteral(IValue value) => value switch
        {
            ByteValue byteValue => byteValue.Value,
            IntValue intValue => byte.MinValue <= intValue.Value && intValue.Value <= byte.MaxValue ? (byte?)intValue.Value : null, 
            _ => null
        };

        public override object ParseValue(object value) => ValueConverter.ConvertTo(value, typeof(byte));
    }
```
I remembered that in case of conversion failure, we return null and this serves as an indication of the impossibility of conversion. It was your PR. Now I doubt if we should use an explicit cast under `checked` context. Now I don't understand how best to do it. I will listen to your thoughts. I will come back to this PR later.

sungam3r:
I will add tests after I understand if I'm going in the right direction.

Shane32:
If I remember correctly, we handle conversion exceptions also.  The null check is typically used if the conversion can't even be attempted.  I'll check and see.

sungam3r:
> If I remember correctly, we handle conversion exceptions also.

I know. I began to be confused by the variety of approaches used - with and without throwing an exception, calling checks directly and through `ValueConverter`.

Shane32:
I confirmed that exceptions thrown within `ParseLiteral` will be wrapped appropriately.  See link below.

https://github.com/graphql-dotnet/graphql-dotnet/blob/e76446be4318d83e279d77d22fce9fae0767f48d/src/GraphQL/Execution/ExecutionHelper.cs#L163..L175

src/GraphQL/Conversion/ValueConverter.cs
Shane32: At least you thought of it!  Coming from a VB.Net background, in which arithmetic is checked by default, I forget that arithmetic is unchecked by default in C#.

Shane32 (approved):

Shane32:
Any reason why these changes should go in 3.3.0 vs 3.2.1?  Since valid queries might (properly) return a float instead of an int, it could change operation of existing working code.  On the other hand, it is certainly a bug fix, and should not have returned an int to begin with.

Just double checking @sungam3r before we release 3.2.1

sungam3r:
3.3.0 ? It is 3.2.1. I bump patch number since there were no new or breaking changes in public APIs. 
изображение

Shane32:
Just checking thanks

PR #2051, opened 2020-12-14
XML comments for scalar graph types

Shane32:
Also fixes a bug:
- `ULongGraphType` cannot parse integer literals greater than `long.MaxValue` and less than or equal to `ulong.MaxValue`.

### DIFF
--- a/src/GraphQL/Introspection/__EnumValue.cs
+++ b/src/GraphQL/Introspection/__EnumValue.cs
@@ -12,11 +12,11 @@ namespace GraphQL.Introspection
                 "a placeholder for a string or numeric value. However an Enum value is " +
                 "returned in a JSON response as a string.";
 
-            Field(f => f.Name);
-            Field(f => f.Description, nullable: true);
+            Field(f => f.Name).Description(null);
+            Field(f => f.Description, nullable: true).Description(null);
 
             Field<NonNullGraphType<BooleanGraphType>>("isDeprecated", resolve: context => (!string.IsNullOrWhiteSpace(context.Source?.DeprecationReason)).Boxed());
-            Field(f => f.DeprecationReason, nullable: true);
+            Field(f => f.DeprecationReason, nullable: true).Description(null);
         }
     }
 }
--- a/src/GraphQL/Types/Scalars/DateGraphType.cs
+++ b/src/GraphQL/Types/Scalars/DateGraphType.cs
@@ -4,14 +4,21 @@ using GraphQL.Language.AST;
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The Date scalar graph type represents a year, month and day.
+    /// </summary>
     public class DateGraphType : ScalarGraphType
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DateGraphType"/> class
+        /// </summary>
         public DateGraphType()
         {
             Description = "The `Date` scalar type represents a year, month and day in accordance with the " +
                 "[ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) standard.";
         }
 
+        /// <inheritdoc/>
         public override object Serialize(object value)
         {
             var date = ParseValue(value);
@@ -24,6 +31,7 @@ namespace GraphQL.Types
             return null;
         }
 
+        /// <inheritdoc/>
         public override object ParseValue(object value)
         {
             if (value is DateTime dateTime)
@@ -47,6 +55,7 @@ namespace GraphQL.Types
             throw new FormatException($"Could not parse date. Expected either a string or a DateTime without time component. Value: {value}");
         }
 
+        /// <inheritdoc/>
         public override object ParseLiteral(IValue value)
         {
             if (value is DateTimeValue timeValue)
--- a/src/GraphQL/Types/Scalars/DateGraphType.cs
+++ b/src/GraphQL/Types/Scalars/DateGraphType.cs
@@ -10,7 +10,7 @@ namespace GraphQL.Types
     public class DateGraphType : ScalarGraphType
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="DateGraphType"/> class
+        /// Initializes a new instance of the <see cref="DateGraphType"/> class.
         /// </summary>
         public DateGraphType()
         {
--- a/src/GraphQL/Types/Scalars/DateGraphType.cs
+++ b/src/GraphQL/Types/Scalars/DateGraphType.cs
@@ -5,7 +5,7 @@ using GraphQL.Language.AST;
 namespace GraphQL.Types
 {
     /// <summary>
-    /// The Date scalar graph type represents a year, month and day.
+    /// The Date scalar graph type represents a year, month and day in accordance with the ISO-8601 standard.
     /// </summary>
     public class DateGraphType : ScalarGraphType
     {
--- a/src/GraphQL/Types/Scalars/DateTimeGraphType.cs
+++ b/src/GraphQL/Types/Scalars/DateTimeGraphType.cs
@@ -4,7 +4,7 @@ using GraphQL.Language.AST;
 namespace GraphQL.Types
 {
     /// <summary>
-    /// The DateTime scalar graph type represents a date and time.
+    /// The DateTime scalar graph type represents a date and time in accordance with the ISO-8601 standard.
     /// </summary>
     public class DateTimeGraphType : ScalarGraphType
     {
--- a/src/GraphQL/Types/Scalars/DateTimeGraphType.cs
+++ b/src/GraphQL/Types/Scalars/DateTimeGraphType.cs
@@ -1,10 +1,12 @@
 using System;
 using GraphQL.Language.AST;
+using GraphQL.Utilities;
 
 namespace GraphQL.Types
 {
     /// <summary>
     /// The DateTime scalar graph type represents a date and time in accordance with the ISO-8601 standard.
+    /// By default <see cref="GraphTypeTypeRegistry"/> maps all <see cref="DateTime"/> .NET values to this scalar graph type.
     /// </summary>
     public class DateTimeGraphType : ScalarGraphType
     {
--- a/src/GraphQL/Types/Scalars/DateTimeGraphType.cs
+++ b/src/GraphQL/Types/Scalars/DateTimeGraphType.cs
@@ -3,8 +3,14 @@ using GraphQL.Language.AST;
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The DateTime scalar graph type represents a date and time.
+    /// </summary>
     public class DateTimeGraphType : ScalarGraphType
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DateTimeGraphType"/> class
+        /// </summary>
         public DateTimeGraphType()
         {
             Description =
@@ -12,8 +18,10 @@ namespace GraphQL.Types
                 "to be formatted in accordance with the [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) standard.";
         }
 
+        /// <inheritdoc/>
         public override object ParseValue(object value) => ValueConverter.ConvertTo(value, typeof(DateTime));
 
+        /// <inheritdoc/>
         public override object ParseLiteral(IValue value)
         {
             if (value is DateTimeValue timeValue)
--- a/src/GraphQL/Types/Scalars/DateTimeGraphType.cs
+++ b/src/GraphQL/Types/Scalars/DateTimeGraphType.cs
@@ -9,7 +9,7 @@ namespace GraphQL.Types
     public class DateTimeGraphType : ScalarGraphType
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="DateTimeGraphType"/> class
+        /// Initializes a new instance of the <see cref="DateTimeGraphType"/> class.
         /// </summary>
         public DateTimeGraphType()
         {
--- a/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
@@ -184,7 +184,7 @@ namespace GraphQL.Types
     }
 
     /// <summary>
-    /// A class that represents an enumeratoin definition.
+    /// A class that represents an enumeration definition.
     /// </summary>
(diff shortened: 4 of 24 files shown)


Shane32:
See #1857

src/GraphQL/Introspection/__EnumValue.cs
```diff
@@ -12,11 +12,11 @@ public __EnumValue()
...
                 "returned in a JSON response as a string.";
 
-            Field(f => f.Name);
-            Field(f => f.Description, nullable: true);
+            Field(f => f.Name).Description(null);
+            Field(f => f.Description, nullable: true).Description(null);
 
             Field<NonNullGraphType<BooleanGraphType>>("isDeprecated", resolve: context => (!string.IsNullOrWhiteSpace(context.Source?.DeprecationReason)).Boxed());
-            Field(f => f.DeprecationReason, nullable: true);
+            Field(f => f.DeprecationReason, nullable: true).Description(null);
```
Shane32, line 19: Without `.Description(null)`, the schema pulls the description from the xml comments that were added.

src/GraphQL/Types/Scalars/UriGraphType.cs
```diff
@@ -3,10 +3,15 @@
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The Uri scalar graph type represents an Uri represented as a string value.
```
sungam3r, line 7:
```suggestion
    /// The Uri scalar graph type represents a string Uri specified in RFC 2396, RFC 2732, RFC 3986, and RFC 3987.
```

src/GraphQL/Types/Scalars/ULongGraphType.cs
```diff
@@ -2,16 +2,22 @@
...
+    /// </summary>
     public class ULongGraphType : ScalarGraphType
     {
+        /// <inheritdoc/>
         public override object ParseLiteral(IValue value) => value switch
         {
             ULongValue ulongValue => ulongValue.Value,
             IntValue intValue => intValue.Value >= 0 ? (ulong?)intValue.Value : null,
             LongValue longValue => longValue.Value >= 0 ? (ulong?)longValue.Value : null,
+            BigIntValue bigIntValue => bigIntValue.Value >= 0 && bigIntValue.Value <= ulong.MaxValue ? (ulong?)bigIntValue.Value : null,
```
sungam3r, line 16: I just want to ask one more time and make sure we are on the right track by adding such branches of code to many scalar classes?

src/GraphQL/Types/Scalars/UIntGraphType.cs
```diff
@@ -2,8 +2,12 @@
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The ULong scalar graph type represents an unsigned 32-bit integer value.
```
sungam3r, line 6:
```suggestion
    /// The UInt scalar graph type represents an unsigned 32-bit integer value.
```

src/GraphQL/Types/Scalars/TimeSpanSecondsGraphType.cs
```diff
@@ -3,15 +3,22 @@
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The Seconds scalar graph type represents a period of time represented as a total number of seconds.
```
sungam3r, line 7:
```suggestion
    /// The Seconds scalar graph type represents a period of time represented as an integer value of the total number of seconds.
```

src/GraphQL/Types/Scalars/TimeSpanMillisecondsGraphType.cs
```diff
@@ -3,15 +3,22 @@
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The Milliseconds scalar graph type represents a period of time represented as a total number of milliseconds.
```
sungam3r, line 7:
```suggestion
    /// The Milliseconds scalar graph type represents a period of time represented as an integer value of the total number of milliseconds.
```

src/GraphQL/Types/Scalars/StringGraphType.cs
```diff
@@ -2,10 +2,15 @@
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The String scalar graph type represents a string value.
```
sungam3r, line 6:
```suggestion
    /// The String scalar graph type represents a string value. It is one of the five built-in scalars. 
```

src/GraphQL/Types/Scalars/SByteGraphType.cs
```diff
@@ -2,15 +2,20 @@
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The SByte scalar graph type represents a signed 8-bit integer value.
```
sungam3r, line 6: I think we should add this comment for each scalar graph type here:
> By default `GraphTypeTypeRegistry` maps all `sbyte` .NET values to this scalar graph type. 

and so on for `int`, `long`, etc.

src/GraphQL/Types/Scalars/StringGraphType.cs
sungam3r: Also see summary for IntGraphType:
>  https://graphql.github.io/graphql-spec/June2018/#sec-Int

It is worth linking to the spec section for each of the built-in scalars.

src/GraphQL/Types/Scalars/IdGraphType.cs
```diff
@@ -2,22 +2,32 @@
...
+        /// Initializes a new instance of the <see cref="IdGraphType"/> class.
+        /// </summary>
         public IdGraphType()
         {
             Name = "ID";
             //Description =
             //    "The `ID` scalar type represents a unique identifier, often used to re-fetch an object or " +
             //    "as key for a cache. The `ID` type appears in a JSON response as a `String`; however, it " +
-            //    "is not intended to be human-readable. When expected as an input type, any string (such " +
+            //    "is not intended to be human-readable. When accepted as an input type, any string (such " +
```
sungam3r, line 20: This wording is from spec:
изображение

src/GraphQL/Types/Scalars/DateTimeGraphType.cs
```diff
@@ -3,17 +3,25 @@
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The DateTime scalar graph type represents a date and time.
```
sungam3r, line 7:
```suggestion
    /// The DateTime scalar graph type represents a date and time in accordance with the ISO-8601 standard.
```

src/GraphQL/Types/Scalars/DateGraphType.cs
```diff
@@ -4,14 +4,21 @@
 
 namespace GraphQL.Types
 {
+    /// <summary>
+    /// The Date scalar graph type represents a year, month and day.
```
sungam3r, line 8:
```suggestion
    /// The Date scalar graph type represents a year, month and day in accordance with the ISO-8601 standard.
```

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -17,11 +17,17 @@ namespace GraphQL.Types
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EnumerationGraphType"/> class.
+        /// </summary>
         public EnumerationGraphType()
         {
             Values = new EnumValues();
         }
 
+        /// <summary>
+        /// Adds a value to the allowed set of enumeration values.
```
sungam3r, line 29: The meaning of the arguments needs to be explained here. Especially the difference between `name` (public API surface) and `value` (internal server value).

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -104,17 +122,33 @@ public EnumerationGraphType()
             }
         }
 
+        /// <summary>
+        /// Changes the case of the specified string to constant case (uppercase, using underscores to separate words).
```
sungam3r, line 126:
```suggestion
        /// Changes the case of the specified enum name. By default changes it to constant case (uppercase, using underscores to separate words).
```

src/GraphQL/Introspection/__EnumValue.cs
sungam3r: Funny that we have to hide the description. It is right?

sungam3r (changes requested):
Thanks. I marked all the places that I considered necessary.

src/GraphQL/Introspection/__EnumValue.cs
Shane32: There was no description in the schema before I added the xml comment.  It's default behavior for a `Field` method call with an expression to pull the description from the xml comment, I guess.  So to override this default, I set it to null, so that the schema retains the prior description (none).  I'm open to suggestions, but I had not intended to add a description to the schema in this PR, but only to add xml comments to the code.

sungam3r: The spec says obvious descriptions can be omitted. We _can_ add these descriptions if we _want_.

src/GraphQL/Types/Scalars/IdGraphType.cs
Shane32: Well, it doesn't look right to me.  Perhaps because the context is different (in the code, the context is the field, whereas in the spec, the context is "input coercion").  Perhaps it's fine.  In any case I'll just leave it.  It's not like it matters.

Shane32:
```suggestion
            //    "is not intended to be human-readable. When expected as an input type, any string (such " +
```

src/GraphQL/Introspection/__EnumValue.cs
Shane32: Regardless, I would not auto-import schema descriptions from the xml code comments.

src/GraphQL/Types/Scalars/ULongGraphType.cs
Shane32: It will work just fine either way.  If we use `checked((ulong)bigIntValue.Value)` then overflows will throw an exception, get caught, and the proper exception is then thrown.  If we return null, it will detect that and throw the same type of exception.  I'm indifferent.

src/GraphQL/Introspection/__EnumValue.cs
sungam3r: Let's leave them `null`.

src/GraphQL/Types/Scalars/ULongGraphType.cs
sungam3r: I understand  but sooner or later someone will have a question why in some cases we use one, and in some other.

src/GraphQL/Types/Scalars/SByteGraphType.cs
Shane32: Done

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
Shane32: I'm adding more details to EnumValueDefinition.Name and EnumValueDefinition.Value.  Please review.

Shane32: Oh I see this overload could use some more definitions also... I'll make corresponding changes.....

Shane32: After updating those definitions, I copied the descriptions into the param xml comments; take a look.

src/GraphQL/Types/Scalars/IntGraphType.cs
```diff
@@ -1,13 +1,16 @@
 using GraphQL.Language.AST;
+using GraphQL.Utilities;
 
 namespace GraphQL.Types
 {
     /// <summary>
-    /// The Int scalar type represents a signed 32‐bit numeric non‐fractional value.
+    /// The Int scalar type represents a signed 32‐bit numeric non‐fractional value. It is one of the five built-in scalars.
+    /// By default <see cref="GraphTypeTypeRegistry"/> maps all <see cref="int"/> .NET values to this scalar graph type.
     /// https://graphql.github.io/graphql-spec/June2018/#sec-Int
```
Shane32, line 9:
```suggestion
```
I don't think we should embed links to the spec into the xml comments.  The purpose of the comment is so that people don't need the link.

src/GraphQL/Types/Scalars/ULongGraphType.cs
Shane32: Consistency is good.  In this case I copied the form used in the prior two lines of code.  We can always make another PR to consolidate all of the scalars to use the form in the ValueConverter if desired.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
sungam3r: ok

src/GraphQL/Types/Scalars/IntGraphType.cs
sungam3r: Why so? It is additional context to read about.

src/GraphQL/Types/Scalars/ULongGraphType.cs
sungam3r: ok

src/GraphQL/Types/Scalars/IntGraphType.cs
Shane32: Nobody does that.  I can't remember ever seeing a link in a Microsoft xml comment.  I don't think it's a good practice.  If someone wants to look up the spec, then they can look it up.

sungam3r: ok

sungam3r (approved):

src/GraphQL/Types/Scalars/IntGraphType.cs
Shane32: > ok

So shall I commit the suggestion?

Shane32: I assume you're fine with the change, so I'll commit the suggestion and merge.

sungam3r: I'm ok with any option.

PR #2058, opened 2020-12-20
Extension methods to configure authorization requirements for GraphQL elements: types, fields, schema.

sungam3r:
See https://github.com/graphql-dotnet/server/issues/293 and especially https://github.com/graphql-dotnet/server/issues/293#issuecomment-740469302 .I believe it would be correct to provide a base API for declaration of authorization requirements in the main project. In this case, the implementation will remain in the corresponding separate repositories.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
+++ b/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
@@ -6,7 +6,11 @@ namespace GraphQL.Tests.Bugs
 {
     public class Bug138DecimalPrecisionTests : QueryTestBase<DecimalSchema>
     {
+#if NETCOREAPP3_1
         [Fact]
+#else
+        [Fact(Skip = "24.149999999999999 with .NET Core < 3.1")]
+#endif
         public void double_to_decimal_does_not_lose_precision()
         {
             var query = @"
--- a/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
+++ b/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
@@ -1,4 +1,5 @@
 using GraphQL.Types;
+using Shouldly;
 using Xunit;
 
 namespace GraphQL.Tests.Bugs
@@ -38,6 +39,7 @@ namespace GraphQL.Tests.Bugs
                 resolve: context =>
                 {
                     var val = context.GetArgument<decimal>("request");
+                    val.ShouldBe(24.15m);
                     return val;
                 });
         }
--- a/src/GraphQL/AuthorizationExtensions.cs
+++ b/src/GraphQL/AuthorizationExtensions.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using GraphQL.Builders;
 using GraphQL.Types;
@@ -55,6 +56,9 @@ namespace GraphQL
         public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataProvider provider, string policy)
             where TMetadataProvider : IProvideMetadata
         {
+            if (policy == null)
+                throw new ArgumentNullException(nameof(policy));
+
             var list = GetPolicies(provider) ?? new List<string>();
 
             if (!list.Contains(policy))
--- a/src/GraphQL/AuthorizationExtensions.cs
+++ b/src/GraphQL/AuthorizationExtensions.cs
@@ -1,0 +1,98 @@
+using System.Collections.Generic;
+using GraphQL.Builders;
+using GraphQL.Types;
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Extension methods to configure authorization requirements for GraphQL elements: types, fields, schema.
+    /// </summary>
+    public static class AuthorizationExtensions
+    {
+        /// <summary>
+        /// Metadata key name for storing authorization policy names. Value of this key
+        /// is a simple list of strings.
+        /// </summary>
+        public const string POLICY_KEY = "Authorization__Policies";
+
+        /// <summary>
+        /// Gets a list of authorization policy names for the specified metadata provider.
+        /// </summary>
+        /// <param name="provider">
+        /// Metadata provider. This can be an instance of <see cref="GraphType"/>,
+        /// <see cref="FieldType"/>, <see cref="Schema"/> or others.
+        /// </param>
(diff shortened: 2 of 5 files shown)


src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
```diff
@@ -1,11 +1,16 @@
...
 using Xunit;
 
 namespace GraphQL.Tests.Bugs
 {
     public class Bug138DecimalPrecisionTests : QueryTestBase<DecimalSchema>
     {
+#if NETCOREAPP3_1
         [Fact]
+#else
+        [Fact(Skip = "24.149999999999999 with .NET Core < 3.1")]
```
sungam3r, line 12: @Shane32 Earlier this test was also ignored. Then I deleted #if/#else. I didn't understand how it worked before. Now I saw that the test started to fail and added again #if/#else. Also I see weird messages from windows build:
```
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1101: Unable to find package Microsoft.SourceLink.GitHub. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1102: Unable to find package NETStandard.Library with version (>= 2.0.3) [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1102:   - Found 1 version(s) in Microsoft Visual Studio Offline Packages [ Nearest version: 1.6.0 ] [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1101: Unable to find package GraphQL-Parser. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1102: Unable to find package Microsoft.CSharp with version (>= 4.7.0) [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1102:   - Found 2 version(s) in Microsoft Visual Studio Offline Packages [ Nearest version: 4.0.1 ] [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1102: Unable to find package System.ComponentModel.Annotations with version (>= 4.7.0) [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1102:   - Found 2 version(s) in Microsoft Visual Studio Offline Packages [ Nearest version: 4.1.0 ] [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1101: Unable to find package System.Reactive.Core. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1101: Unable to find package System.Reactive.Linq. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
```

I've seen this before but thought it was a single glitch. Some info I found: https://github.com/actions/virtual-environments/issues/1090#issuecomment-748432535 I don’t have time to go into more detail yet. If you want then go on

Shane32:
@sungam3r At first I was thinking that this makes sense.  And then I was wondering if we shouldn't take it a step further, and include a validation rule that walks the nodes of the document, checking the document metadata to see if any policies are defined.  The actual implementation of the verification is dependent on the authentication method in use, not part of GraphQL.

This validation rule might actually be a bit complex, for example, if it skips validation for nodes depending on directives, for example.  Also, there might be a lot of speed improvements that could be made here.  For example, caching pass/fail results for a given policy (within a specific document), so that the same policy referenced on other nodes would not need to run the authorization check again.  So it might make a lot of sense to include this as part of GraphQL instead of the specific authorization framework in use.

Keep in mind that I have not used either of the authorization frameworks that exist, as I wrote my own authorization code from scratch.

sungam3r:
My goal is to separate the API declaration of requirements from the API of the implementation of these requirements. Now we have two projects for authorization (generic + ASP.NET Core specific) with a common part. I do not want the code that declares the fields /graph types to depend on one or another authorization implementation. I understand that authorization itself has nothing to do with core graphql code in this repo but the added code does not bring new dependencies and fits into the concept of meta information on fields and graph types.

It is possible that in the future I will also try to bring both implementations together to avoid code duplication.

src/GraphQL/AuthorizationExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <see cref="FieldType"/>, <see cref="Schema"/> or others.
+        /// </param>
+        /// <param name="policy"> Authorization policy name. </param>
+        /// <returns> The reference to the specified <paramref name="provider"/>. </returns>
+        public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataProvider provider, string policy)
+            where TMetadataProvider : IProvideMetadata
+        {
+            var list = GetPolicies(provider) ?? new List<string>();
+
+            if (!list.Contains(policy))
```
Shane32, line 60:
```suggestion
            if (!list.Contains(policy ?? throw new ArgumentNullException(nameof(policy))))
```

Shane32 (changes requested):

src/GraphQL/AuthorizationExtensions.cs
sungam3r: Added before `  var list = GetPolicies(provider) ?? new List<string>();`

Shane32 (approved):

PR #2061, opened 2020-12-21
DecimalValue check does not allocate memory

sungam3r:
Inspired by https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/984.
Resolves old TODO with `decimal.GetBits`.

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -14,4 +14,5 @@ jobs:
     steps:
     - uses: actions/labeler@v3
       with:
+        sync-labels: ""
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/src/GraphQL/Language/CoreToVanillaConverter.cs
+++ b/src/GraphQL/Language/CoreToVanillaConverter.cs
@@ -393,5 +393,4 @@ namespace GraphQL.Language
 
         internal bool Equals(ref DecimalData other) => Flags == other.Flags && Hi == other.Hi && Lo == other.Lo && Mid == other.Mid;
     }
-
 }
--- a/src/GraphQL/Language/CoreToVanillaConverter.cs
+++ b/src/GraphQL/Language/CoreToVanillaConverter.cs
@@ -250,12 +250,13 @@ namespace GraphQL.Language
                         str.Value,
                         NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent,
                         CultureInfo.InvariantCulture,
-                        out var dec))
+                        out decimal dec))
                     {
-                        // TODO: make more efficient, current solution allocates memory
-                        int[] decBits = decimal.GetBits(dec);
-                        int[] dblAsDecBits = decimal.GetBits(new decimal(dbl));
-                        if (decBits[0] != dblAsDecBits[0] || decBits[1] != dblAsDecBits[1] || decBits[2] != dblAsDecBits[2] || decBits[3] != dblAsDecBits[3])
+                        // Cast the decimal to our struct to avoid the decimal.GetBits allocations.
+                        var decBits = System.Runtime.CompilerServices.Unsafe.As<decimal, DecimalData>(ref dec);
+                        decimal temp = new decimal(dbl);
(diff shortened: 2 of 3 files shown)


src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -364,4 +365,33 @@ public static T WithLocation<T>(this T node, ASTNode astNode)
...
+        {
+            Flags = flags;
+            Hi = hi;
+            Lo = lo;
+            Mid = mid;
+        }
+
+        internal bool Equals(ref DecimalData other) => Flags == other.Flags && Hi == other.Hi && Lo == other.Lo && Mid == other.Mid;
+    }
+
```
sungam3r, line 396:
```suggestion
```

src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -364,4 +365,32 @@ public static T WithLocation<T>(this T node, ASTNode astNode)
...
+        internal DecimalData(uint flags, uint hi, uint lo, uint mid)
+        {
+            Flags = flags;
+            Hi = hi;
+            Lo = lo;
+            Mid = mid;
+        }
+
+        internal bool Equals(ref DecimalData other) => Flags == other.Flags && Hi == other.Hi && Lo == other.Lo && Mid == other.Mid;
```
sungam3r, line 394:
```suggestion
        internal bool Equals(in DecimalData other) => Flags == other.Flags && Hi == other.Hi && Lo == other.Lo && Mid == other.Mid;
```

Shane32:
I don't understand why the code can't simply compare two `decimal` instances (structs) to each other.

Shane32:
I guess I don't understand what this code is doing.

.github/workflows/label.yml
```diff
@@ -14,4 +14,5 @@ jobs:
     steps:
     - uses: actions/labeler@v3
       with:
+        sync-labels: ""
```
sungam3r, line 17: https://github.com/actions/labeler/issues/104#issuecomment-715911428

sungam3r:
@Shane32 for me it was somewhat strange, but in principle the behavior can be understood
изображение

sungam3r:
> I guess I don't understand what this code is doing.

изображение

Shane32:
Makes sense.  But when we can't even get 1.15 to equal 1.15 (referring to a recent change where we skipped a test related to that), you wonder about the value of preserving trailing zeros.  I'm not even sure how to read the number of trailing zeros without calling `.ToString()` on the `decimal`.

Shane32:
By the way, according to the spec, any values outside the range of `double` precision floating point values should be refused:

> If the integer input value represents a value not representable by IEEE 754, a query error should be raised.

See: https://spec.graphql.org/June2018/#sec-Float

So really we should have no support for `decimal` types at all, since it's pretty clear that the `Float` scalar type is designed to represent IEEE 754 values and nothing else.

sungam3r:
> referring to a recent change where we skipped a test related to that

It works for netcoreapp3.1

> So really we should have no support for decimal types at all, since it's pretty clear that the Float scalar type is designed to represent IEEE 754 values and nothing else.

I admit that it would be possible to follow this path. Then we would have to generate an error while still inside the parser. I haven't read the IEEE 754 spec and I admit that trailing zeros may be not included in it. However, I would like to remain user-friendly and allow the values to be kept accurate. Now I just removed memory allocation without changing the general behavior. We may consider strict compliance with the specification in another PR.

Shane32 (approved):
It's a bit of a hack, but okay.

sungam3r:
We can also ask for advice from the developers of the reference implementation (graphql-js).

Shane32:
I'm not sure what there is to ask.  It's abundantly clear that a Float is intended to represent an IEEE 754 value, which is exactly what a C# `double` is.  And an IEEE 754 value typically contains the following formula: `+/- 1.fffffff * (2^exp)`.  Look up 'normalized mantissa'.  (there are exceptions for very small or very large numbers).

https://www.geeksforgeeks.org/ieee-standard-754-floating-point-numbers/

sungam3r:
OK, I agree with you. If we strictly follow the spec, then we should remove a few features that already work. It will hurt me to do this. Better to do something new than delete the working code.

PR #2064, opened 2020-12-24
Handle IDataLoaderResult properly within list graph types

### LINKED ISSUE #2062: DataLoader does not allow for null to be returned for federated schema
## Description

When using DataLoader in tandem with a federated schema, `null` can no longer be returned and instead an error is thrown. Without DataLoader in place this is not an issue and `null` can be returned for scenarios when the requested data does not exist.

### Steps to reproduce

This can be reproduced with the following code:

**Mock service:**
```
public interface IProductService
{
    Task<Product> GetProductById(int id);

    Task<IDictionary<int, Product>> GetProductsByIds(IEnumerable<int> ids);
}

public class ProductService : IProductService
{
    private readonly IDictionary<int, Product> _productStore = new Dictionary<int, Product>
    {
        {1, new Product {Id = 1, Name = "Product 1"}},
        {2, new Product {Id = 2, Name = "Product 2"}},
        {3, new Product {Id = 3, Name = "Product 3"}}
    };

    public async Task<Product> GetProductById(int id)
    {
        _productStore.TryGetValue(id, out var product);

        return await Task.FromResult(product);
    }

    public async Task<IDictionary<int, Product>> GetProductsByIds(IEnumerable<int> ids)
    {
        var results = new Dictionary<int, Product>();

        foreach (var id in ids)
        {
            _productStore.TryGetValue(id, out var product);

            results.Add(id, product);
        }

        return await Task.FromResult(results);
    }
}
```
**Controller:**
```
[Route("[controller]")]
[ApiController]
public class GraphQlController : Controller
{
    private readonly ISchema _schema;
    private readonly IDocumentExecuter _documentExecuter;
    private readonly IDocumentExecutionListener _documentExecutionListener;

    public GraphQlController(
        ISchema schema,
        IDocumentExecuter documentExecuter,
        IDocumentExecutionListener documentExecutionListener)
    {
        _schema = schema;
        _documentExecuter = documentExecuter;
        _documentExecutionListener = documentExecutionListener;
    }

    [HttpPost]
    public async Task<IActionResult> Post([FromBody] GraphQlQuery query)
    {
        var options = new ExecutionOptions
        {
            Schema = _schema,
            OperationName = query.OperationName,
            Query = query.Query,
            Inputs = query.Variables == string.Empty ? null : query.Variables.ToInputs(),
        };

        options.Listeners.Add(_documentExecutionListener);

        var result = await _documentExecuter.ExecuteAsync(options);

        if (result.Errors?.Count > 0)
        {
            return BadRequest(result.Errors.Select(e => e.Message));
        }

        return Ok(result);
    }
}
```
**Startup.ConfigureServices:**
```
{
    services.AddControllers();

    services.AddSingleton<IProductService, ProductService>();

    services.AddSingleton<AnyScalarGraphType>();
    services.AddSingleton<ServiceGraphType>();
    services.AddSingleton<IDataLoaderContextAccessor, DataLoaderContextAccessor>();
    services.AddSingleton<IDocumentExecutionListener, DataLoaderDocumentListener>();
    services.AddTransient<IDocumentExecuter, DocumentExecuter>();

    services.AddTransient(serviceProvider =>
    {
        var productService = serviceProvider.GetRequiredService<IProductService>();
        var dataLoaderContextAccessor = serviceProvider.GetRequiredService<IDataLoaderContextAccessor>();

        return FederatedSchema.For(
            @"
                type Product @key(fields: ""id"") {
                    Id: Int
                    Name: String
                }
            ", _ =>
            {
                _.ServiceProvider = serviceProvider;
                _.Types.For("Product").ResolveReferenceAsync(async context =>
                {
                    // this code works as expected, allowing null to be returned
                    //
                    // var product = await productService.GetProductById((int) context.Arguments["id"]);
                    // return product;

                    // mock async
                    await Task.CompletedTask;

                    var loader = dataLoaderContextAccessor.Context.GetOrAddBatchLoader<int, Product>(
                        "GetProductsById",
                        productService.GetProductsByIds
                    );

                    // this throws errors when null is returned
                    return loader.LoadAsync((int)context.Arguments["id"]);
                });
            });
    });

    services
        .AddMvcCore(options => { options.SuppressOutputFormatterBuffering = true; })
        .AddJsonOptions(options =>
        {
            options.JsonSerializerOptions.Converters.Add(
                new ExecutionResultJsonConverter(new ErrorInfoProvider())
            );
        });
}
```
**Using the following query:**
```
query GetTheProduct {
  _entities(
    representations: [
      { __typename: "Product", id: 1 },
      { __typename: "Product", id: 2 },
      { __typename: "Product", id: 3 },
      { __typename: "Product", id: 4 },
    ]
  ) {
    ... on Product {
      id
      name
    }
  }
}

```

### Expected result

The expected output of the above code and query would be the following:

```
{
  "data": {
    "_entities": [
      {
        "__typename": "Product",
        "id": 1,
        "name": "Product 1"
      },
      {
        "__typename": "Product",
        "id": 2,
        "name": "Product 2"
      },
      {
        "__typename": "Product",
        "id": 3,
        "name": "Product 3"
      },
      null
    ]
  }
}
```

### Actual result

Instead of the above output, we get a `"Error trying to resolve field '_entities'."` with the inner exception of `"Cannot return null for non-null type. Field: _entities, Type: [_Entity]!."`

### Environment

* Host OS: Windows 10
* GraphQL: v3.2.0
* GraphQL.Server.Ui.Playground: v4.4.0
* GraphQL.SystemTextJson: v3.2.0

Shane32:
Fixes #2062

### DIFF
--- a/src/GraphQL/Execution/ExecutionNode.cs
+++ b/src/GraphQL/Execution/ExecutionNode.cs
@@ -11,9 +11,26 @@ namespace GraphQL.Execution
         public IGraphType GraphType { get; }
         public Field Field { get; }
         public FieldType FieldDefinition { get; }
-        //todo: make public
-        internal IGraphType ResolvedType { get; }
         public int? IndexInParentNode { get; protected set; }
+
+        //todo: make public
+        internal IGraphType ResolvedType
+        {
+            get
+            {
+                if (IndexInParentNode.HasValue)
+                {
+                    var t = FieldDefinition.ResolvedType;
+                    if (t is NonNullGraphType nonNullGraphType)
+                        t = nonNullGraphType.ResolvedType;
+                    return ((ListGraphType)t).ResolvedType;
+                }
+                else
+                {
+                    return FieldDefinition?.ResolvedType;
+                }
+            }
+        }
 
         public string Name => Field?.Alias ?? Field?.Name;
 
@@ -37,24 +54,12 @@ namespace GraphQL.Execution
             set => _source = value;
         }
 
-        //todo: add parameter to constructor for resolvedType
         protected ExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, int? indexInParentNode)
         {
             Parent = parent;
             GraphType = graphType;
             Field = field;
             FieldDefinition = fieldDefinition;
-            if (indexInParentNode.HasValue)
-            {
-                var t = fieldDefinition.ResolvedType;
-                if (t is NonNullGraphType nonNullGraphType)
-                    t = nonNullGraphType.ResolvedType;
-                ResolvedType = ((ListGraphType)t).ResolvedType;
-            }
-            else
-            {
-                ResolvedType = fieldDefinition?.ResolvedType;
-            }
             IndexInParentNode = indexInParentNode;
         }
 
--- a/src/GraphQL/Execution/ExecutionNode.cs
+++ b/src/GraphQL/Execution/ExecutionNode.cs
@@ -11,6 +11,8 @@ namespace GraphQL.Execution
         public IGraphType GraphType { get; }
         public Field Field { get; }
         public FieldType FieldDefinition { get; }
+        //todo: make public
+        internal IGraphType ResolvedType { get; }
         public int? IndexInParentNode { get; protected set; }
 
         public string Name => Field?.Alias ?? Field?.Name;
@@ -35,12 +37,24 @@ namespace GraphQL.Execution
             set => _source = value;
         }
 
+        //todo: add parameter to constructor for resolvedType
         protected ExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, int? indexInParentNode)
         {
             Parent = parent;
             GraphType = graphType;
             Field = field;
             FieldDefinition = fieldDefinition;
+            if (indexInParentNode.HasValue)
+            {
+                var t = fieldDefinition.ResolvedType;
+                if (t is NonNullGraphType nonNullGraphType)
+                    t = nonNullGraphType.ResolvedType;
+                ResolvedType = ((ListGraphType)t).ResolvedType;
+            }
+            else
+            {
+                ResolvedType = fieldDefinition?.ResolvedType;
+            }
             IndexInParentNode = indexInParentNode;
         }
 
--- a/src/GraphQL/Execution/ExecutionNode.cs
+++ b/src/GraphQL/Execution/ExecutionNode.cs
@@ -37,7 +37,11 @@ namespace GraphQL.Execution
         /// </summary>
         public int? IndexInParentNode { get; protected set; }
 
-        //todo: make public
+        /// <summary>
+        /// Returns the underlying graph type of this node, retaining the <see cref="NonNullGraphType"/> wrapping if applicable.
+        /// For child nodes of an array execution node, this property unwraps the <see cref="ListGraphType"/> instance and returns
+        /// the underlying graph type, retaining the <see cref="NonNullGraphType"/> wrapping if applicable.
+        /// </summary>
         internal IGraphType ResolvedType
         {
             get
--- a/src/GraphQL/Execution/ExecutionNode.cs
+++ b/src/GraphQL/Execution/ExecutionNode.cs
@@ -48,10 +48,7 @@ namespace GraphQL.Execution
             {
                 if (IndexInParentNode.HasValue)
                 {
-                    var t = FieldDefinition.ResolvedType;
-                    if (t is NonNullGraphType nonNullGraphType)
-                        t = nonNullGraphType.ResolvedType;
-                    return ((ListGraphType)t).ResolvedType;
+                    return ((ListGraphType)Parent.GraphType).ResolvedType;
                 }
                 else
                 {
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -90,7 +90,7 @@ namespace GraphQL.Execution
         {
             var listType = (ListGraphType)parent.GraphType;
             var itemType = listType.ResolvedType;
- 
+
             if (itemType is NonNullGraphType nonNullGraphType)
                 itemType = nonNullGraphType.ResolvedType;
 
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -90,7 +90,7 @@ namespace GraphQL.Execution
         {
             var listType = (ListGraphType)parent.GraphType;
             var itemType = listType.ResolvedType;
-
+ 
             if (itemType is NonNullGraphType nonNullGraphType)
                 itemType = nonNullGraphType.ResolvedType;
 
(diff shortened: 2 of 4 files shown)


src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -111,18 +111,21 @@ public static void SetArrayItemNodes(ExecutionContext context, ArrayExecutionNod
...
-                    if (node is ObjectExecutionNode objectNode)
-                    {
-                        SetSubFieldNodes(context, objectNode);
-                    }
-                    else if (node is ArrayExecutionNode arrayNode)
-                    {
-                        SetArrayItemNodes(context, arrayNode);
-                    }
-                    else if (node is ValueExecutionNode valueNode)
+                    if (!(d is IDataLoaderResult))
```
Shane32, line 114: SetSubFieldNodes and similar was incorrectly executing twice for `IDataLoaderResult` nodes, even if none of the values were `null`.  However, when a value was `null`, the sub-field nodes had already been set when they should not have been, causing a null reference exception.

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -11,6 +11,8 @@ public abstract class ExecutionNode
         public IGraphType GraphType { get; }
         public Field Field { get; }
         public FieldType FieldDefinition { get; }
+        //todo: make public
+        internal IGraphType ResolvedType { get; }
```
Shane32, line 15: Making this public now is not a breaking change and could be done.

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -35,12 +37,24 @@ public object Source
             set => _source = value;
         }
 
+        //todo: add parameter to constructor for resolvedType
         protected ExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, int? indexInParentNode)
```
Shane32, line 41: Adding another parameter is a breaking change, so I suggest we add the parameter in v4.  Of course we could add another constructor to provide an optimized path when processing array nodes in v3.

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -13,6 +13,25 @@ public abstract class ExecutionNode
         public FieldType FieldDefinition { get; }
         public int? IndexInParentNode { get; protected set; }
 
+        //todo: make public
+        internal IGraphType ResolvedType
```
Shane32, line 17: Making this public now is not a breaking change and could be done.

Shane32:
I'm going to merge this unless you have any other suggestions, @sungam3r .

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -37,6 +37,25 @@ public abstract class ExecutionNode
...
+                if (IndexInParentNode.HasValue)
+                {
+                    var t = FieldDefinition.ResolvedType;
+                    if (t is NonNullGraphType nonNullGraphType)
+                        t = nonNullGraphType.ResolvedType;
+                    return ((ListGraphType)t).ResolvedType;
+                }
+                else
+                {
+                    return FieldDefinition?.ResolvedType;
```
sungam3r, line 54: Why `?.` ?

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -37,6 +37,25 @@ public abstract class ExecutionNode
...
+        internal IGraphType ResolvedType
+        {
+            get
+            {
+                if (IndexInParentNode.HasValue)
+                {
+                    var t = FieldDefinition.ResolvedType;
+                    if (t is NonNullGraphType nonNullGraphType)
+                        t = nonNullGraphType.ResolvedType;
+                    return ((ListGraphType)t).ResolvedType;
```
sungam3r, line 50:
```suggestion
                   return FieldDefinition.ResolvedType.GetNamedType();
```

sungam3r: But why you unwrap `NonNullGraphType` ? I see non null check on line https://github.com/graphql-dotnet/graphql-dotnet/pull/2064/files#diff-df2b6c74045af40593068c2612253adc21462ab45546c45ca3a27bd29fd83ab1R369

Shane32: `FieldDefinition.ResolvedType` is guaranteed to be a list graph type.  I need to unwrap the list only, not the nullability of the inside type.  In order to unwrap the list, I first need to check if it's a non-null list.  This code specifically does not unwrap the nullability of the interior graph type, as that is its entire purpose.  This is also why `GetNamedType()` does not work.

src/GraphQL/Execution/ExecutionNode.cs
Shane32: If the `ExecutionNode` is a root node then `FieldDefinition` will return null.

src/GraphQL/Execution/ExecutionNode.cs
Shane32: Note that the `ExecutionNode.GraphType` property already equals `FieldDefinition?.ResolvedType.GetNamedType()`, I believe.  (It's set from the constructor, but that's basically what it contains.)

sungam3r: > FieldDefinition.ResolvedType is guaranteed to be a list graph type in this context. I need to unwrap the list only, not the nullability of the inside type. In order to unwrap the list, I first need to check if it's a non-null list. This code specifically does not unwrap the nullability of the interior graph type, as that is its entire purpose. This is also why GetNamedType() does not work.

I highly recommend adding comments about this. It wasn't obvious to me.

sungam3r: > Note that the ExecutionNode.GraphType property already equals FieldDefinition?.ResolvedType.GetNamedType(), I believe. (It's set from the constructor, but that's basically what it contains.)

I looked at the code and I had suspicions that we can make mistakes and pass wrapped types sometime, and sometimes not. I suggest to add type checks into `ExecutionNode` and its desccendants ctors to enforce rules and make rules explicit.

Shane32: Done. Please review

Shane32: Added checks for `ExecutionNode` ctors in PR #2069

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: > SetSubFieldNodes and similar was incorrectly executing twice for IDataLoaderResult nodes, even if none of the values were null

What side effects did it have?

sungam3r (approved):

Shane32:
@sungam3r I made additional changes to properly handle lists of lists.  Please review again if you can.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Other than the bug already noted, performance only.  Ran some code twice when it didn't need to.

sungam3r:
Could you add a test to verify changes?

Shane32:
Yes I will. Not sure when.

Shane32:
@sungam3r I added a test.....but it doesn't actually fail with the old (incorrect) code.  I did manually verify that the revised code is correct.  I'm struggling to figure out where to properly add tests.

Shane32:
I think what we really need to test is that `ExecutionNode.GraphType` is set properly.  But this is set within `BuildExecutionNode`, which is called from `SetSubFieldNodes`.  We probably need tests for `SetSubFieldNodes` and `SetArrayItemNodes`.  Since it and `BuildExecutionNode` are both static, we cannot mock the methods and independently test operation of the members.  (Or else I don't know how.)

Once we can verify that `ExecutionNode.GraphType` is set correctly, it's easy to write a few extra lines of code to verify that `ExecutionNode.ResolvedType` also returns the correct value.

PR #2067, opened 2020-12-28
NET5 support

sungam3r:


src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0</TargetFrameworks>
```
sungam3r, line 4: See https://docs.microsoft.com/en-us/dotnet/standard/net-standard#when-to-target-net50-vs-netstandard

sungam3r: Both netcoreapp3.1 and net5 implement netstandard2.1 which contains APIs for spans used here under `#if` preprocessor directives.

Shane32:
Build artifacts and publish workflows should not change. Just the test workflow. Or we can build and publish the project with .Net 5 SDK. Then we can delete the setup for .Net 3.1 sdk in those workflows. The built project will still be compatible with .Net standard

sungam3r:
How can I build projects with <TargetFrameworks>netcoreapp3.1;net5</TargetFrameworks> without installing both SDKs ?

Shane32:
The SDK will allow building any project that targets the same or older framework. You just can’t run a built project if the proper library isn’t installed. So to run tests that target 2.1 then you need 2.1 runtime or SDK installed.

Shane32:
I can test it to be sure but if you look in the development branch, that’s basically how it’s set up. For running tests both 2.1 and 3.1 are installed, but for building and publishing only 3.1 is installed

sungam3r:
I think I have encountered this behavior before, but I forgot.

> Or we can build and publish the project with .Net 5 SDK. Then we can delete the setup for .Net 3.1 sdk in those workflows. 

Assuming that you are right, which is the best choice? Perhaps it's better to use net5 as the newest?

Shane32:
Not sure; either way is fine with me.  3.1 is a LTS version but 5.0 is newest.  I'd probably stick with 3.1 since it's working already ("if it ain't broke don't fix it") and also because it's a LTS version.

If we want to use C# 9.0 syntax or features then we need the 5.0 SDK to compile the project.  But C# 7.3 is the last language version that fully supports .Net Standard 2.0 projects.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

sungam3r:
> If we want to use C# 9.0 syntax or features then we need the 5.0 SDK to compile the project.

Let's go on with NET5. Gradually, the appearance of features of this SDK is inevitable.

sungam3r:
@Shane32 Is everything right now?

Shane32 (approved):

sungam3r:
Well, it works but
```
MSBUILD : error MSB1025: An internal failure occurred while running MSBuild.
[2020-12-30 23:47:17] [build] Microsoft.Build.BackEnd.NodeFailedToLaunchException: No such file or directory
```

sungam3r:
I tried to build #1964 in order to check if it is net5 releated problem.

Shane32:
Looks like it's just a lgtm.com error? I'm not worried about that.  Just disable lgtm.com if needed.

sungam3r:
Analisys for #1964  works.

> Just disable lgtm.com if needed.

The point is that this analyzer actually found places that, in my opinion, required fixing. That is, it is beneficial. Apparently it has some kind of problem when using NET5. If now I merge this PR into master, then lgtm will stop doing analysis for the master branch. It will be a shame. I will not merge this PR for now and will try to get more information about the error. In the meantime, I will do other PRs based on the analysis results.

Shane32:
Just make a decision and go with it. I’m good either way.

sungam3r:
I'm ready to merge it into `develop`.

sungam3r:
@Shane32 OK?

Issue #1604, opened 2020-03-11
Registering nullable value types doesn't allow inferring the Graph Type


ebekker:
## Summary
I tried to register with `GraphTypeTypeRegistry` the nullable variations of the built-in value types to the corresponding `*GraphTypes` (essentially duplicating [this built-in code](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/GraphTypeTypeRegistry.cs) but for the nullable CLR types) in the hopes that it would be able to _infer_ the types of my fields without explicitly specifying the Graph type.

Unfortunately, it only works if the GraphQL field is itself declared nullable.  If the GraphQL field is declared not nullable, but I have a nullable CLR type field that it maps to, I have to explicitly repeat the type mapping in the field declaration.  Is this by design?  Any way to overcome this?

### Environment (if relevant)
* GraphQL 3.0.0-preview-1490
* .NET SDK 3.1.200-preview-014883
* Windows Server 2016

## Examples

Let's say I have an entity class like so:

```csharp
public class Foobar
{
    public int Id { get; set; }
    public string Name { get; set; }
    public bool? Disabled { get; set; }
}
```

And I have a Graph type class that maps to it:

``` csharp
public class FoobarGraph : ObjectGraphType<Foobar>
{
    public FoobarGraph()
    {
        Field(t => t.Id);
        Field(t => t.Name);

        // This generates an exception, unable to infer, unless I also add
        // type: typeof(BooleanGraphType)
        Field(t => t.Disabled,
            nullable: false);
    }
}
```

sungam3r:
>  Is this by design? 

Yes, it is. Another thing is that such a design can hardly be considered good. There is still a backward compatibility issue. Looking through code I see that you can set `nullable: true` to solve it. Using `type: typeof(BooleanGraphType)` is another option to achive the same. Try it. I will leave this issue open as a reminder to change `Field<T>` API.

sungam3r:
Relates to [comment](https://github.com/graphql-dotnet/graphql-dotnet/issues/369#issuecomment-314830399) about API for definining fields.

@joemcbride What are you think about refactor fields API? Optional parameters cause difficulties for many users, since the effect of their use is not obvious. Many overloads complicates the situation. I would try to somehow simplify the API if you do not mind.

HrvojeFER:
I'm having this exact same problem. 

`
            Field<StringGraphType>(
                "id",
[...]

PR #2075, opened 2021-01-01
Add connection and field builder xml comments

Shane32:


src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -105,36 +159,51 @@ public ConnectionBuilder<TSourceType> Bidirectional()
             return this;
         }
 
+        /// <inheritdoc cref="FieldBuilder{TSourceType, TReturnType}.Name(string)"/>
```
sungam3r, line 162: `inheritdoc` here? `ConnectionBuilder<TSourceType>` does not inherit from `FieldBuilder`.

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -105,36 +159,51 @@ public ConnectionBuilder<TSourceType> Bidirectional()
...
 
+        /// <inheritdoc cref="FieldBuilder{TSourceType, TReturnType}.DeprecationReason(string)"/>
         public ConnectionBuilder<TSourceType> DeprecationReason(string deprecationReason)
         {
             FieldType.DeprecationReason = deprecationReason;
             return this;
         }
 
+        /// <summary>
+        /// Sets the default page size
```
sungam3r, line 184: Why default?

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -6,24 +6,51 @@
 
 namespace GraphQL.Builders
 {
+    /// <summary>
+    /// Static methods to create connection builders.
```
sungam3r, line 10:
```suggestion
    /// Static methods to create connection field builders.
```

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -47,30 +71,46 @@ private FieldBuilder(EventStreamFieldType fieldType)
...
+        /// Sets the deprecation reason of the field.
+        /// </summary>
         public virtual FieldBuilder<TSourceType, TReturnType> DeprecationReason(string deprecationReason)
         {
             FieldType.DeprecationReason = deprecationReason;
             return this;
         }
 
+        /// <summary>
+        /// Sets the default value of fields on input object graph types.
```
sungam3r, line 112: I suggest to throw `InvalidOperationException` if `FieldType` is not an input type.

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -83,28 +123,53 @@ private FieldBuilder(EventStreamFieldType fieldType)
             return this;
         }
 
+        /// <summary>
+        /// Sets the resolver for the field.
```
sungam3r, line 127: I suggest to throw `InvalidOperationException` if `FieldType` is not an input type.

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -281,6 +281,12 @@ public virtual FieldType AddField(FieldType fieldType)
             });
         }
 
+        /// <summary>
+        /// Returns a builder for a new field.
```
sungam3r, line 285:
```suggestion
        /// Adds a new field to the complex graph type and returns a builder for this newly added field.
```

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -289,8 +295,20 @@ public virtual FieldType AddField(FieldType fieldType)
             return builder;
         }
 
+        /// <summary>
+        /// Returns a builder for a new field.
```
sungam3r, line 299:
```suggestion
        /// Adds a new field to the complex graph type and returns a builder for this newly added field.
```

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -289,8 +295,20 @@ public virtual FieldType AddField(FieldType fieldType)
...
+        /// </summary>
+        /// <typeparam name="TGraphType">The graph type of the field.</typeparam>
         public virtual FieldBuilder<TSourceType, object> Field<TGraphType>() => Field<TGraphType, object>();
 
+        /// <summary>
+        /// Returns a builder for a new field that is linked to a property of the source object.
```
sungam3r, line 305:
```suggestion
        /// Adds a new field to the complex graph type and returns a builder for this newly added field that is linked to a property of the source object.
```

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -289,8 +295,20 @@ public virtual FieldType AddField(FieldType fieldType)
...
+        /// </summary>
+        /// <typeparam name="TProperty">The return type of the field.</typeparam>
+        /// <param name="name">The name of this field.</param>
+        /// <param name="expression">The property of the source object represented within an expression.</param>
+        /// <param name="nullable">Indicates if this field should be nullable or not.</param>
```
sungam3r, line 310: Please note in comments that this property is only used when the type property is not set (and for type property comment too). I have probably explained this behavior a hundred times already to anyone who encountered the incomprehensible behavior of their types.

sungam3r: See #1604  and related issues.

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -326,6 +344,14 @@ public virtual FieldType AddField(FieldType fieldType)
             return builder;
         }
 
+        /// <summary>
+        /// Returns a builder for a new field that is linked to a property of the source object.
```
sungam3r, line 348:
```suggestion
        /// Adds a new field to the complex graph type and returns a builder for this newly added field that is linked to a property of the source object.
```

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -326,6 +344,14 @@ public virtual FieldType AddField(FieldType fieldType)
...
+        /// The default name of this field is inferred by the property represented within the expression.
+        /// </summary>
+        /// <typeparam name="TProperty">The return type of the field.</typeparam>
+        /// <param name="expression">The property of the source object represented within an expression.</param>
+        /// <param name="nullable">Indicates if this field should be nullable or not.</param>
```
sungam3r, line 353: The same suggestions here.

sungam3r:
Also could you please fix this warning? I think it is because of `internal` access modifier on `ResolvedType`.
изображение

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: Sure why not?  It's just copying the one-line description, and the description is still accurate -- it's the name of the field.

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: That's what it is.  If specified, the optional `first` or `last` field arguments sets the number of records to return, thereby becoming the page size for the request.

src/GraphQL/Builders/FieldBuilder.cs
Shane32: I agree in principle.  I'm not sure exactly how this would work in reality, since it's the graph type that's an input type, not the field, and since the builder doesn't know what graph type the field is connected to, or even if it has been added to a graph type at all.  If you wish to explore further, I suggest you open a bug report for this issue and we can solve it in a separate PR.

src/GraphQL/Builders/FieldBuilder.cs
Shane32: This appears to be a duplicate of the above comment, not intended for this member.  If you have another comment that applies here, please let me know.

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32:
```suggestion
        /// Sets the default page size.
```

Shane32: Maybe we should rename this member to `DefaultPageSize` for v4.0 ?

Shane32: If so, I might also make it accept a `int?` and remove the `ReturnAll` member.

sungam3r: ok

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: ok

src/GraphQL/Builders/FieldBuilder.cs
sungam3r: Let's keep it as is for now.

src/GraphQL/Types/Composite/ComplexGraphType.cs
Shane32: Do you really think we have to mention it twice?  See updated xml comments.

sungam3r: It's not obligatory. At least it got better.

sungam3r (approved):

PR #2078, opened 2021-01-01
Add performance metrics xml comments

Shane32:


src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -3,51 +3,113 @@
...
 {
+    /// <summary>
+    /// Contains Apollo tracing metrics.
+    /// </summary>
     public class ApolloTrace
     {
+        /// <summary>
+        /// Initializes a new instance with the specified parameters.
+        /// </summary>
+        /// <param name="start">The UTC date and time that the GraphQL document began execution.</param>
```
sungam3r, line 14: We have no control over the `DateTime.Kind` here. It may be any kind.

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -3,51 +3,113 @@
...
             Duration = ConvertTime(durationMs);
         }
 
+        /// <summary>
+        /// Returns the Apollo tracing version number.
+        /// </summary>
         public int Version => 1;
 
+        /// <summary>
+        /// Returns the UTC date and time when the document began execution.
```
sungam3r, line 29: I suggest not to mention `UTC`.

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -3,51 +3,113 @@
...
+        /// </summary>
         public DateTime StartTime { get; }
 
+        /// <summary>
+        /// Returns the UTC date and time when the document completed execution.
```
sungam3r, line 34: I suggest not to mention `UTC`.

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -3,51 +3,113 @@
...
 
+        /// <summary>
+        /// Returns the execution metrics.
+        /// </summary>
         public ExecutionTrace Execution { get; } = new ExecutionTrace();
 
+        /// <summary>
+        /// Converts a quantity of milliseconds to nanoseconds.
+        /// </summary>
         public static long ConvertTime(double ms) => (long)(ms * 1000 * 1000);
```
sungam3r, line 61: This method should not be public. May be done in separate PR targeted to `develop`.

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -3,51 +3,113 @@
...
 
+        /// <summary>
+        /// Represents the start offset and duration of an operation.
+        /// </summary>
         public class OperationTrace
         {
+            /// <summary>
+            /// Sets or returns the start offset of the operation, in nanoseconds.
```
sungam3r, line 69:
```suggestion
            /// Gets or sets the start offset of the operation, in nanoseconds.
```

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -3,51 +3,113 @@
...
+            /// </summary>
             public long StartOffset { get; set; }
 
+            /// <summary>
+            /// Sets or returns the duration of the operation, in nanoseconds.
```
sungam3r, line 74:
```suggestion
            /// Gets or sets the duration of the operation, in nanoseconds.
```

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -3,51 +3,113 @@
...
             public List<ResolverTrace> Resolvers { get; } = new List<ResolverTrace>();
         }
 
+        /// <summary>
+        /// Represents metrics pertaining to the execution of a field resolver.
+        /// </summary>
         public class ResolverTrace : OperationTrace
         {
+            /// <summary>
+            /// Sets or returns the path of the field.
```
sungam3r, line 96: The same suggestion `Gets or sets` here and further.

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -3,51 +3,113 @@
...
+            /// </summary>
             public List<object> Path { get; set; } = new List<object>();
```
sungam3r, line 98:
```suggestion
            public List<object> Path { get; set; }
```
See `Path = fieldStat.MetaField<IEnumerable<object>>("path").ToList(),` Initialization is an extra allocation of memory for an empty list.

src/GraphQL/Instrumentation/ApolloTracingExtensions.cs
```diff
@@ -4,8 +4,20 @@
...
+        /// <summary>
+        /// Adds Apollo tracing metrics to an <see cref="ExecutionResult"/> instance,
+        /// stored within <see cref="ExecutionResult.Extensions"/>["tracing"].
+        /// Requires that the GraphQL document was executed with metrics enabled;
+        /// see <see cref="ExecutionOptions.EnableMetrics"/>. With <see cref="InstrumentFieldsMiddleware"/>
+        /// installed, also includes metrics from field resolvers.
+        /// </summary>
+        /// <param name="result">An <see cref="ExecutionResult"/> instance.</param>
+        /// <param name="start">The UTC date and time that the GraphQL document began execution.</param>
         public static void EnrichWithApolloTracing(this ExecutionResult result, DateTime start)
```
sungam3r, line 21: What do you think about such implementation?
```c#
public static void EnrichWithApolloTracing(this ExecutionResult result, DateTime start)
        {
            var perf = result?.Perf;
            if (perf != null)
                (result.Extensions ?? new Dictionary<string, object>())["tracing"] = CreateTrace(perf, start);
        }
```

src/GraphQL/Instrumentation/ApolloTracingExtensions.cs
```diff
@@ -4,8 +4,20 @@
...
     {
+        /// <summary>
+        /// Adds Apollo tracing metrics to an <see cref="ExecutionResult"/> instance,
+        /// stored within <see cref="ExecutionResult.Extensions"/>["tracing"].
+        /// Requires that the GraphQL document was executed with metrics enabled;
+        /// see <see cref="ExecutionOptions.EnableMetrics"/>. With <see cref="InstrumentFieldsMiddleware"/>
+        /// installed, also includes metrics from field resolvers.
+        /// </summary>
+        /// <param name="result">An <see cref="ExecutionResult"/> instance.</param>
+        /// <param name="start">The UTC date and time that the GraphQL document began execution.</param>
```
sungam3r, line 20:
```suggestion
        /// <param name="start">The date and time that the GraphQL document began execution.</param>
```

src/GraphQL/Instrumentation/ApolloTracingExtensions.cs
```diff
@@ -22,6 +34,12 @@ public static void EnrichWithApolloTracing(this ExecutionResult result, DateTime
             result.Extensions["tracing"] = trace;
         }
 
+        /// <summary>
+        /// Initializes an <see cref="ApolloTrace"/> instance and populates it with performance
+        /// metrics gathered during the GraphQL document execution.
+        /// </summary>
+        /// <param name="perf">A list of performance records; typically as returned from <see cref="Metrics.Finish"/>.</param>
+        /// <param name="start">The UTC date and time that the GraphQL document began execution.</param>
```
sungam3r, line 42:
```suggestion
        /// <param name="start">The date and time that the GraphQL document began execution.</param>
```

src/GraphQL/Instrumentation/PerfRecord.cs
```diff
@@ -14,20 +20,45 @@ public PerfRecord(string category, string subject, double start, Dictionary<stri
             Metadata = metadata;
         }
 
+        /// <summary>
+        /// Sets the completion time, represented as an offset in milliseconds from starting the GraphQL operation's execution.
+        /// </summary>
         public void MarkEnd(double end) => End = end;
 
+        /// <summary>
+        /// Sets or returns the category name.
```
sungam3r, line 29: `Gets or sets` here and further.

src/GraphQL/Instrumentation/ApolloTracingExtensions.cs
Shane32: Apart from the fact that it requires `??=` instead of `??`, I'd say it looks good.  It operates exactly the same, so I'll include it in this pull request.

Shane32:
The specification in https://github.com/apollographql/apollo-tracing requires that the timestamp format is RFC 3339.  The serializer will output that format only if the `DateTime.Kind` property is `DateTimeKind.Utc`.  So I am leaving the UTC notes in place.  The middleware in the sample already uses `DateTime.NowUtc`.  I have added additional notes that RFC 3339 is required during serialization.

src/GraphQL/Instrumentation/ApolloTracingExtensions.cs
sungam3r: Of course ??=. 🤦

sungam3r:
Then we should add utc checks.

Shane32:
> Then we should add utc checks.

The reason the xml comments are in this PR is so that people know that to _properly_ implement Apollo tracing, they need to use UTC timestamps. This is not obvious at present. It is certainly possible to use local timestamps, and System.Text.Json will happily serialize them with the time zone offset. It simply doesn't meet spec.

Because it _works_ fine without UTC timestamps, it's a breaking change to add checks or conversions to the code.  Since we plan to release 4.0 soon, I added `.ToUniversalTime()` in the constructor within #2083, and then no other checks are needed.  I will further update the xml comments in that PR once this PR is merged.

src/GraphQL/Instrumentation/ApolloTrace.cs
Shane32: Technically a breaking change. Done in #2083

src/GraphQL/Instrumentation/ApolloTrace.cs
Shane32: Done in #2083

src/GraphQL/Instrumentation/ApolloTrace.cs
Shane32: Discussed below.

sungam3r:
OK

src/GraphQL/Instrumentation/ApolloTrace.cs
sungam3r: OK

sungam3r (approved):

PR #2079, opened 2021-01-01
Remove LightweightCache.GetKey and .First

Shane32:
`GetKey` isn't implemented and doesn't override a member.

`First` is not used and does not seem to fulfill a purpose.

Targeting `develop` of course.


sungam3r:
Maybe add some notes in migration guide?

sungam3r (approved):

Shane32:
> Maybe add some notes in migration guide?

I can't think of a reason, but why not? Done.

sungam3r (approved):

Issue #2076, opened 2021-01-01
Remove StatsReport?


Shane32:
I suggest we remove `StatsReport` or at least mark it deprecated.  It is not used by any other code, there are no samples for it, and there are no tests for it.  It does not appear ever to even have been completed -- for instance, the `Types` property always returns `null` because of this code:

```cs
        public static Type[] TypesFromSchema(ISchema schema)
        {
            return null;
        }
```

It was first written by @joemcbride on Nov 16, 2016 and has never been updated since then.  It's only 130 lines of code if someone needs to copy it into their own project.

@sungam3r @joemcbride  do you know if this is used at all?  Can/should we remove it?

sungam3r:
Agree. I have seen this code before and it seems to me that I even removed it in some of my PRs (possibly canceled).

PR #2080, opened 2021-01-01
Remove StatsReport and associated classes?

Shane32:
See question in #2076 

Targets `develop`

Alternatively, could simply be marked as deprecated.


sungam3r (approved):

Shane32:
I'll wait for additional comments for a few days before merging.

PR #2085, opened 2021-01-01
Add xml comments for GraphQL relay objects

Shane32:


src/GraphQL/Types/Relay/ConnectionType.cs
```diff
@@ -1,9 +1,19 @@
...
+    /// the parent field's resolver) is <see cref="ConnectionType{TNodeType, TEdgeType}"/>
+    /// or <see cref="ConnectionType{TNodeType}"/> or has the same members.
+    /// </summary>
+    /// <typeparam name="TNodeType">The graph type of the result data set's data type.</typeparam>
+    /// <typeparam name="TEdgeType">The edge graph type of node, typically <see cref="EdgeType{TNodeType}"/>.</typeparam>
     public class ConnectionType<TNodeType, TEdgeType> : ObjectGraphType<object>
         where TNodeType : IGraphType
         where TEdgeType : EdgeType<TNodeType>
     {
+        /// <inheritdoc/>
```
sungam3r, line 16: I think that description for `edges` field is wrong, it seems to be copy-pasted from `pageInfo` field description.

PR #2086, opened 2021-01-01
Update description for ConnectionType's edges field

Shane32:
While not exactly a breaking change, this can potentially change the schema, possibly breaking tests for users of the library.

Should this be included in the next major or minor revision?


sungam3r:
I'm OK with minor.

sungam3r (approved):

PR #2088, opened 2021-01-01
Bump dotnet-format

sungam3r:


src/Directory.Build.props
```diff
@@ -15,6 +15,7 @@
     <DebugType>embedded</DebugType>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <NoWarn>$(NoWarn);1591</NoWarn> <!--TODO: temporary solution-->
```
Shane32, line 18: With this change, let's remove ` -p:NoWarn=CS1591` from the github action workflow(s).

Shane32 (changes requested):

src/Directory.Build.props
sungam3r: I expected this suggestion :)

sungam3r: done

.github/workflows/test-code.yml
```diff
@@ -30,22 +30,22 @@ jobs:
...
+      - name: Install dependencies
+        working-directory: src
+        run: dotnet restore
       - name: Check formatting
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         working-directory: src
         run: |
-          dotnet tool install -g dotnet-format
-          dotnet format --check -v diag || (echo "Run 'dotnet format' to fix formatting issues" && exit 1)
-      - name: Install dependencies
```
sungam3r, line 39: Some analyzers require already installed dependencies, so this step has been moved up.

sungam3r:
изображение

src/GraphQL.DataLoader.Tests/Types/UserType.cs
```diff
@@ -43,15 +43,10 @@ public UserType(IDataLoaderContextAccessor accessor, IOrdersStore orders)
...
-                            //var batchResults = await orders.GetItemsByOrderIdAsync(orderResults.Select(o => o.OrderId));
-                            //var allResults = orderResults.Select(o => batchResults[o.OrderId]);
-
-                            //flatten and return the results
-                            return allResults.SelectMany(x => x);
-                        });
+                        //without dataloader, this would be:
+                        //var batchResults = await orders.GetItemsByOrderIdAsync(orderResults.Select(o => o.OrderId));
+                        //var allResults = orderResults.Select(o => batchResults[o.OrderId]);
+                        return itemsLoader.LoadAsync(orderResults.Select(o => o.OrderId)).Then(allResults => allResults.SelectMany(x => x)); //flatten and return the results
```
Shane32, line 49:
```suggestion
                        return itemsLoader.LoadAsync(orderResults.Select(o => o.OrderId))
                            .Then(allResults => allResults.SelectMany(x => x)); //flatten and return the results
```

Shane32 (approved):

Shane32:
I would rather that within Visual Studio we could see the 1591 warnings.

Shane32:
I've got rid of a couple hundred 1591 warnings in the last couple weeks.  Hopefully I can blast through a bunch more here over the weekend.

Shane32:
> I would rather that within Visual Studio we could see the 1591 warnings.

But it's fine for now.  We need to get through the bulk of them first anyway.  And that's just double clicking on every file and adding comments.

sungam3r:
> I would rather that within Visual Studio we could see the 1591 warnings.

Just comment NoWarn in props file.

Shane32:
I think I'll need your help for some of the xml comments.  I'm trying to go through everything I can and hopefully you can tackle the last few methods and classes I'm unfamiliar with.

PR #2093, opened 2021-01-02
Change Connection.TotalCount to nullable int

Shane32:
@sungam3r Any comments? Seems rather obvious to me when considering this existing code:

```cs
            Field<IntGraphType>()
                .Name("totalCount")
                .Description(
                    "A count of the total number of objects in this connection, ignoring pagination. " +
                    "This allows a client to fetch the first five objects by passing \"5\" as the argument " +
                    "to `first`, then fetch the total count so it could display \"5 of 83\", for example. " +
                    "In cases where we employ infinite scrolling or don't have an exact count of entries, " +
                    "this field will return `null`.");
```

### DIFF
--- a/src/GraphQL/Types/Relay/DataObjects/Connection.cs
+++ b/src/GraphQL/Types/Relay/DataObjects/Connection.cs
@@ -12,7 +12,7 @@ namespace GraphQL.Types.Relay.DataObjects
         where TEdge : Edge<TNode>
     {
         /// <summary>
-        /// The total number of records available.
+        /// The total number of records available. Returns <see langword="null"/> if the total number is unknown.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/GraphQL/Types/Relay/DataObjects/Connection.cs
```diff
@@ -14,7 +14,7 @@ public class Connection<TNode, TEdge>
         /// <summary>
         /// The total number of records available.
```
sungam3r, line 15:
```suggestion
        /// The total number of records available. Returns <see langword="null"/> if the total number is unknown.
```

sungam3r (approved):

Shane32:
@sungam3r It doesn't appear that the labeler fix worked in that position

PR #2095, opened 2021-01-02
Prevent failed build checks due to labeler

Shane32:


.github/workflows/label.yml
```diff
@@ -16,3 +16,4 @@ jobs:
       with:
         sync-labels: ""
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
+      continue-on-error: true
```
sungam3r, line 19: Was a labeler error once an obstacle for other actions? See #2094:
изображение

Shane32:
An obstacle? Not that I know of.

Shane32:
This really bugs me:

image

sungam3r:
Now I understand.
изображение
vs
изображение

Maybe switch to the second? In fact, the result is the same, but it seems to me that the intention will be expressed more clearly.

sungam3r (approved):

PR #2099, opened 2021-01-03
Add xml comments to ExecutionHelper

Shane32:
@sungam3r Perhaps you can help with the two methods marked TODO. Or I can remove the xml comment for those methods and we can revisit later.

### DIFF
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -461,7 +461,10 @@ namespace GraphQL.Execution
         }
 
         /// <summary>
-        /// TODO
+        /// This method calculates the criterion for matching fragment definition (spread or inline) to a given graph type.
+        /// This criterion determines the need to fill the resulting selection set with fields from such a fragment.
+        /// <br/><br/>
+        /// See http://spec.graphql.org/June2018/#DoesFragmentTypeApply()
         /// </summary>
         public static bool DoesFragmentConditionMatch(ExecutionContext context, string fragmentName, IGraphType type)
         {
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -281,7 +281,7 @@ namespace GraphQL.Execution
         }
 
         /// <summary>
-        /// Coerces variable values to a compatible .NET type for the variable's graph type.
+        /// Coerces a variable value to a compatible .NET type for the variable's graph type.
         /// </summary>
         public static object CoerceValue(ISchema schema, IGraphType type, IValue input, Variables variables = null)
         {
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -409,6 +409,11 @@ namespace GraphQL.Execution
         }
 
         /// <summary>
+        /// Before execution, the selection set is converted to a grouped field set by calling CollectFields().
+        /// Each entry in the grouped field set is a list of fields that share a response key (the alias if defined,
+        /// otherwise the field name). This ensures all fields with the same response key included via referenced
+        /// fragments are executed at the same time.
+        /// <br/><br/>
         /// See http://spec.graphql.org/June2018/#sec-Field-Collection and http://spec.graphql.org/June2018/#CollectFields()
         /// </summary>
         public static Dictionary<string, Field> CollectFields(
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -7,8 +7,14 @@ using GraphQL.Types;
 
 namespace GraphQL.Execution
 {
+    /// <summary>
+    /// Provides helper methods for document execution.
+    /// </summary>
     public static class ExecutionHelper
     {
+        /// <summary>
+        /// Returns the root graph type for the execution -- for a specified schema and operation type.
+        /// </summary>
         public static IObjectGraphType GetOperationRootType(Document document, ISchema schema, Operation operation)
         {
             IObjectGraphType type;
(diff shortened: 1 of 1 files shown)


src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -257,6 +280,9 @@ public static void AssertValidVariableValue(ISchema schema, IGraphType type, obj
             return values;
         }
 
+        /// <summary>
+        /// Coerces variable values to a compatible .NET type for the variable's graph type.
```
sungam3r, line 284:
```suggestion
        /// Coerces variable value to a compatible .NET type for the variable's graph type.
```

Shane32:
```suggestion
        /// Coerces a variable value to a compatible .NET type for the variable's graph type.
```

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -454,6 +490,9 @@ public static bool DoesFragmentConditionMatch(ExecutionContext context, string f
             return false;
         }
 
+        /// <summary>
+        /// Returns a list of subfields (child nodes) for a result node based on the selection set from the document.
```
sungam3r, line 494: I suggest to inline this method body into `ReadonlyResolveFieldContext.GetSubFields()` and remove it from here.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -427,6 +460,9 @@ public static bool ShouldIncludeNode(ExecutionContext context, Directives direct
             return true;
         }
 
+        /// <summary>
+        /// TODO
```
sungam3r, line 464: see http://spec.graphql.org/June2018/#DoesFragmentTypeApply()

sungam3r: This method calculates the criterion for matching fragment definition (spread or inline) to a given graph type. This criterion determines the need to fill the resulting selection set with fields from such a fragment.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -382,6 +408,9 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
             return fields;
         }
 
+        /// <summary>
+        /// TODO
```
sungam3r, line 412: see http://spec.graphql.org/June2018/#sec-Field-Collection and http://spec.graphql.org/June2018/#CollectFields()

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I wouldn't.  `ReadonlyResolveFieldContext` happens to be the only use, but it shouldn't contain logic code.  And the code contained therein seems to be a good helper method.  If anyone wanted to make their own execution strategy along with their own `ReadonlyResolveFieldContext`, they should not have to re-implement `GetSubFields()`.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32:
```suggestion
        /// This method calculates the criterion for matching fragment definition (spread or inline) to a given graph type.
        /// This criterion determines the need to fill the resulting selection set with fields from such a fragment.
        /// <br/><br/>
        /// See http://spec.graphql.org/June2018/#DoesFragmentTypeApply()
```
Is this what you suggest?  It's still too mind-boggling for me to understand just now!

src/GraphQL/Execution/ExecutionHelper.cs
Shane32:
```suggestion
        /// See http://spec.graphql.org/June2018/#sec-Field-Collection and http://spec.graphql.org/June2018/#CollectFields()
```
Is this what you suggest at this time?

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: All logic is inside `CollectFields`. `SubFieldsFor` is no more then a thin wrapper. And as you see there is only one client for this wrapper - `ReadonlyResolveFieldContext.GetSubFields()` so it seems to me that this method is in vain here.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: It will fit.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: I think it is OK to just point to the algorithm from the spec.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I don't care.  I wouldn't but it doesn't matter much.

Shane32: Of course it has to be done in the `develop` branch.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Sure you can't come up with some type of description?

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: OK. I will do it along with other changes.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: You can copy-paste description from the spec :)

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -382,6 +408,9 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
             return fields;
         }
 
+        /// <summary>
+        /// See http://spec.graphql.org/June2018/#sec-Field-Collection and http://spec.graphql.org/June2018/#CollectFields()
```
Shane32, line 412:
```suggestion
        /// Before execution, the selection set is converted to a grouped field set by calling CollectFields().
        /// Each entry in the grouped field set is a list of fields that share a response key (the alias if defined,
        /// otherwise the field name). This ensures all fields with the same response key included via referenced
        /// fragments are executed at the same time.
        /// <br/><br/>
        /// See http://spec.graphql.org/June2018/#sec-Field-Collection and http://spec.graphql.org/June2018/#CollectFields()
```

sungam3r (approved):

Issue #2102, opened 2021-01-03
Rework TypeExtensions.GraphTypeFromType


sungam3r:
`GraphQL.Types.TypeExtensions.GraphTypeFromType` can be called multiple time each request with relatively cost operations inside: `MakeGenericType` + `Activator.CreateInstance` . Moreover sometimes the result of this method can be unwrapped with `GetNamedType`.

Note: also there is `GraphQL.TypeExtensions` class with "similar" `GetGraphTypeFromType` method.

sungam3r:
Fixed in #2304

PR #2103, opened 2021-01-03
Eliminate GraphTypeFromType call

sungam3r:
See #2102

### DIFF
--- a/src/GraphQL/Validation/Rules/VariablesAreInputTypes.cs
+++ b/src/GraphQL/Validation/Rules/VariablesAreInputTypes.cs
@@ -21,14 +21,22 @@ namespace GraphQL.Validation.Rules
             {
                 _.Match<VariableDefinition>(varDef =>
                 {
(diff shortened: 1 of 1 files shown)


src/GraphQL/Validation/Rules/VariablesAreInputTypes.cs
```diff
@@ -21,14 +21,22 @@ public Task<INodeVisitor> ValidateAsync(ValidationContext context)
             {
                 _.Match<VariableDefinition>(varDef =>
                 {
-                    var type = varDef.Type.GraphTypeFromType(context.Schema);
+                    var type = GetNamedGraphTypeFromType(varDef.Type, context.Schema);
 
                     if (!type.IsInputType())
```
sungam3r, line 26: `IsInputType()` unwraps `type` so there is no reason to call expensive `GraphTypeFromType` here

Shane32 (approved):

PR #2107, opened 2021-01-03
Optimize allocations for ExecutionNode.GeneratePath + add test for RootExecutionNode

sungam3r:


sungam3r:
Apparently labeler v3 still has problems.

Shane32:
> Apparently labeler v3 still has problems.

Didn't like it before. Still don't now.

Shane32 (approved):

PR #2111, opened 2021-01-03
Rework of Locations structs

sungam3r:


Shane32:
Haha you beat me to this one. If we can merge the other PR first it will help prevent merge conflicts when merging master into develop.

Shane32:
This is what I don't like about the labeler:

image

It doesn't know what it's doing, and then it keeps doing it over and over.

src/GraphQL/Language/AST/SourceLocation.cs
```diff
@@ -1,10 +1,14 @@
...
+
 namespace GraphQL.Language.AST
 {
     /// <summary>
     /// Provides information regarding the location of a node within a document's original query text.
     /// </summary>
-    public class SourceLocation
+    public readonly struct SourceLocation : IEquatable<SourceLocation>
     {
+        public static readonly SourceLocation Empty = new SourceLocation(-1, -1, -1, -1);
```
Shane32, line 10: Missing xml comment.

src/GraphQL/Language/AST/SourceLocation.cs
```diff
@@ -42,22 +46,13 @@ public SourceLocation(int line, int column, int start = -1, int end = -1)
...
-                return true;
-            if (obj.GetType() != GetType())
-                return false;
-            return Equals((SourceLocation)obj);
-        }
+        public static bool operator ==(SourceLocation first, SourceLocation second) => first.Equals(second);
+
+        public static bool operator !=(SourceLocation first, SourceLocation second) => !first.Equals(second);
+
+        public override bool Equals(object obj) => obj is SourceLocation loc && Equals(loc);
```
Shane32, line 55: Missing xml comment.

Shane32 (changes requested):

Shane32:
Structs already have comparison operators and stuff based on contents.  So if we can compare based on all fields, we should be able to remove `IEquatable` and both `Equals` and both operators.

Also see this code:

```cs
    public static class NodeExtensions
    {
        public static T WithLocation<T>(this T node, int line, int column, int start = -1, int end = -1)
            where T : AbstractNode
        {
            node.SourceLocation = new SourceLocation(line, column, start, end);
            return node;
        }
    }
```

There is only one reference to that code.  It is here:

```cs
    public static class AstNodeExtensions
    {
        public static T WithLocation<T>(this T node, ASTNode astNode)
            where T : AbstractNode
        {
            return node.WithLocation(0, 0, astNode?.Location.Start ?? -1, astNode?.Location.End ?? -1);
        }
    }
```

There are no other references to the `SourceLocation` constructor.  There are no other references besides the constructor that _write_ to the `Line` property.  It seems like we can remove `Line` and `Column`, unless I missed something.

And if `Line` and `Column` are always 0, then it is unlikely that any code compares two `SourceLocation` instances to each other.

sungam3r:
> It doesn't know what it's doing, and then it keeps doing it over and over.

It does exactly what it was told:
```
"test":
  - any: ['src/GraphQL.ApiTests/**/*', 'src/GraphQL.DataLoader.Tests/**/*', 'src/GraphQL.Dummy/**/*', 'src/GraphQL.Harness.Tests/**/*', 'src/GraphQL.Tests/**/*', '!src/GraphQL.ApiTests/**/*.txt']
```

Let's decide if the changes in `docs2/site/docs/migrations/migration4.md` belong to the `documentation` label.

sungam3r:
> Structs already have comparison operators and stuff based on contents. So if we can compare based on all fields, we should be able to remove IEquatable and both Equals and both operators.

Yes, but then it will be a comparison based on reflection. It is always recommended to define your own comparison methods and operators.

src/GraphQL/Language/AST/SourceLocation.cs
sungam3r: done

src/GraphQL/Language/AST/SourceLocation.cs
sungam3r: done

sungam3r:
> There are no other references to the SourceLocation constructor. There are no other references besides the constructor that write to the Line property. It seems like we can remove Line and Column, unless I missed something. And if Line and Column are always 0, then it is unlikely that any code compares two SourceLocation instances to each other.

@joemcbride Could you please help?

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -42,7 +42,7 @@ public ValidationError(string originalQuery, string number, string message, para
             nodes?.Apply(n =>
             {
                 _nodes.Add(n);
-                if (n.SourceLocation != null)
+                if (n.SourceLocation != SourceLocation.Empty)
```
Shane32, line 45: Are you sure this is correct? An uninitialized `SourceLocation` property will have its values set to 0, not to `SourceLocation.Empty`.

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -42,7 +42,7 @@ public ValidationError(string originalQuery, string number, string message, para
...
                 {
                     var location = new Location(new Source(originalQuery), n.SourceLocation.Start);
```
Shane32, line 47: This line of code calls `GraphQLParser` to retrieve the `Line` and `Column` from the query based on the start position within the document.

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -42,7 +42,7 @@ public ValidationError(string originalQuery, string number, string message, para
...
                     AddLocation(location.Line, location.Column);
```
Shane32, line 48: And here it adds the line/column to the error, without ever referencing the `SourceLocation.Line` or `SourceLocation.Column` property

Shane32:
If you remove the `if (n.SourceLocation != null)` check (which isn't valid as a struct now), and then make these changes:

image

You will find that all tests pass.  None of that is used anywhere in the codebase.

If we want to preserve the `if (n.SourceLocation != null)` check, perhaps we could add a boolean property called `IsSet` or something.

src/GraphQL/Validation/ValidationError.cs
sungam3r: 🤦

Shane32:
> Yes, but then it will be a comparison based on reflection. 

Not always.  https://medium.com/@semuserable/c-journey-into-struct-equality-comparison-deep-dive-9693f74562f1

> It is always recommended to define your own comparison methods and operators.

I agree, but let's not bog down the API with operators that are never used. The default comparison is perfectly adequate for running tests -- if there even were any!

sungam3r:
I realized very interesting things. I'll change everything.

sungam3r:
@Shane32 I want to release https://github.com/graphql-dotnet/parser/pull/97 first and go on. My idea is to use `GraphQLLocation` struct from parser repo directly and remove `SourceLocation`.

Shane32:
@sungam3r The entire `Language.AST` namespace should be removed, or in some fashion combined with the GraphQLParser types.  Re-using a single struct containing a couple integers will not save any CPU cycles, as it has to copy the containing variables anyway.

If the idea is to allow third-party parsers, then I certainly wouldn't use any classes/structs from GraphQLParser, or else the third party parser would require a dependency on GraphQLParser.

I would suggest that we add another project (separate repo) with the AST classes and structs, and then both GraphQL.NET and GraphQLParser can use it via nuget.

sungam3r:
Here the matter is completely different. I remember that @joemcbride told it happened so historically.

Shane32:
> My idea is to use GraphQLLocation struct from parser repo directly and remove SourceLocation.

I do not think that should be done.

sungam3r:
> I do not think that should be done.

Two structs are the same 100% binary and semantically.

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -42,11 +42,8 @@ public ValidationError(string originalQuery, string number, string message, para
             nodes?.Apply(n =>
             {
                 _nodes.Add(n);
-                if (n.SourceLocation != null)
-                {
-                    var location = new Location(new Source(originalQuery), n.SourceLocation.Start);
```
sungam3r, line 47: no more memory allocation for `Source`

src/GraphQL/Language/CoreToVanillaConverter.cs
```diff
@@ -13,18 +12,12 @@ namespace GraphQL.Language
 {
     public class CoreToVanillaConverter
     {
-        private readonly ISource _body;
```
sungam3r, line 16: I was surprised when I saw that this field is not being used. The class is stateless and can generally be made static. I declared a static instance. Maybe make the class static?

src/GraphQL/Execution/ExecutionError.cs
```diff
@@ -139,20 +148,15 @@ public static class ExecutionErrorExtensions
...
+            where TError: ExecutionError
         {
-            if (abstractNode == null)
-                return;
+            if (abstractNode == null || document == null)
+                return error;
 
-            if (document != null)
-            {
-                var location = new Location(new Source(document.OriginalQuery), abstractNode.SourceLocation.Start);
```
sungam3r, line 149: no more memory alocation for `Source`

sungam3r:
> The entire Language.AST namespace should be removed, or in some fashion combined with the GraphQLParser types. Re-using a single struct containing a couple integers will not save any CPU cycles, as it has to copy the containing variables anyway.

I'm not against. This has not been my goal yet.

Shane32:
> > I do not think that should be done.
> 
> Two structs are the same 100% binary and semantically.

I understand.  But now another parser needs a reference to GraphQLParser in order to compile.  Please undo the change and I'll review again.

docs2/site/docs/migrations/migration4.md
```diff
@@ -12,11 +12,16 @@
...
-* `IGraphType.CollectTypes` has been removed.
-* `ExecutionHelper.SubFieldsFor` has been removed.
-* `TypeCollectionContext` is now internal, also all methods with this parameter in `GraphTypesLookup` are private.
-* `GraphQLTypeReference` is now internal, also `GraphTypesLookup.ApplyTypeReferences` is now private.
+* `LightweightCache.First` method has been removed.
+* `IGraphType.CollectTypes` method has been removed.
+* `ExecutionHelper.SubFieldsFor` method has been removed.
+* `SourceLocation`, `NodeExtensions`, `AstNodeExtensions` classes have been removed.
+* `ErrorLocation` struct became `readonly`.
+* `GraphQL.Language.AST.INode.SourceLocation` property now has `GraphQLParser.AST.GraphQLLocation` type.
```
Shane32, line 20: > `GraphQL.Language.AST.INode.SourceLocation` property now has `GraphQLParser.AST.GraphQLLocation` type.

Please undo.

Shane32 (changes requested):

docs2/site/docs/migrations/migration4.md
sungam3r: done

src/GraphQL/Language/CoreToVanillaConverter.cs
sungam3r: done

src/GraphQL/Language/SourceLocation.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        public int End { get; }
+
+        public bool Equals(SourceLocation other) => Start == other.Start && End == other.End;
+
+        public override bool Equals(object obj) => obj is SourceLocation l && Equals(l);
+
+        public override int GetHashCode() => (Start, End).GetHashCode();
+
+        public override string ToString() => $"({Start},{End})";
```
Shane32, line 42:
```suggestion
        /// <includedoc/>
        public override string ToString() => $"({Start},{End})";
```

Shane32 (changes requested):

src/GraphQL/Language/SourceLocation.cs
```diff
@@ -0,0 +1,44 @@
...
+        ///     { field { subfield } }
+        ///                         ^ field.Location.End = 20
+        /// </summary>
+        public int End { get; }
+
+        public bool Equals(SourceLocation other) => Start == other.Start && End == other.End;
+
+        public override bool Equals(object obj) => obj is SourceLocation l && Equals(l);
+
+        public override int GetHashCode() => (Start, End).GetHashCode();
```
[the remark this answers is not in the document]
sungam3r: Can you show a link to the default implementation? I admit that there are optimized versions of the standard Equals and GetHashCode methods, but you shouldn't rely on them, as even a small code change can turn them off.

src/GraphQL/Language/SourceLocation.cs
```diff
@@ -0,0 +1,44 @@
...
+        ///
+        /// For example,
+        ///     { field { subfield } }
+        ///                         ^ field.Location.End = 20
+        /// </summary>
+        public int End { get; }
+
+        public bool Equals(SourceLocation other) => Start == other.Start && End == other.End;
+
+        public override bool Equals(object obj) => obj is SourceLocation l && Equals(l);
```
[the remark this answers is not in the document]
sungam3r: `inheritdoc` maybe?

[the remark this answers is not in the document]
sungam3r: done with `inheritdoc`

src/GraphQL/Language/SourceLocation.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// The index for the character immediately after the node in the source (i.e. it's exclusive).
+        ///
+        /// For example,
+        ///     { field { subfield } }
+        ///                         ^ field.Location.End = 20
+        /// </summary>
+        public int End { get; }
+
+        public bool Equals(SourceLocation other) => Start == other.Start && End == other.End;
```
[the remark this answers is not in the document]
sungam3r: done

src/GraphQL/Language/SourceLocation.cs
```diff
@@ -0,0 +1,44 @@
...
+        public int Start { get; }
+
+        /// <summary>
+        /// The index for the character immediately after the node in the source (i.e. it's exclusive).
+        ///
+        /// For example,
+        ///     { field { subfield } }
+        ///                         ^ field.Location.End = 20
+        /// </summary>
+        public int End { get; }
```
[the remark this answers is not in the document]
sungam3r: Done with formatting:
изображение

src/GraphQL/Language/SourceLocation.cs
sungam3r: done

src/GraphQL/Language/SourceLocation.cs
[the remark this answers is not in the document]
Shane32: What I mean is that SourceLocation is a struct and GetHashCode would return a hash of the members.  The function you defined creates a struct and executes GetHashCode on it.  So it does the exact same thing, regardless of the .NET implementation of GetHashCode for structs.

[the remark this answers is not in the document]
Shane32: Isn't it ValueType.GetHashCode either way?  Maybe I'm all wrong.

src/GraphQL/Language/AST/SourceLocation.cs
```diff
@@ -17,28 +20,38 @@ public SourceLocation(int start, int end)
...
+        /// For example:
+        /// <code>
+        /// { field { subfield } }
+        /// <br/>
+        /// --------------------^ field.Location.End = 20
+        /// </code>
         /// </summary>
         public int End { get; }
 
+        /// <inheritdoc/>
```
Shane32, line 45: This is not inheriting anything.  Should cref the default equality operator.

Shane32 (changes requested):

src/GraphQL/Language/AST/SourceLocation.cs
sungam3r: изображение

and

изображение

Looks pretty good.

Shane32: Ok then if it works.  I don't understand what it's inheriting.  I think it's still wrong.

Shane32 (approved):

src/GraphQL/Language/SourceLocation.cs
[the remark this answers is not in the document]
sungam3r:
```c#
 /// <summary>Calculates the hash code for the current <see cref="T:System.ValueTuple`2" /> instance.</summary>
    /// <returns>The hash code for the current <see cref="T:System.ValueTuple`2" /> instance.</returns>
    public override int GetHashCode()
    {
      return ValueTuple.CombineHashCodes(EqualityComparer<T1>.Default.GetHashCode(this.Item1), EqualityComparer<T2>.Default.GetHashCode(this.Item2));
    }
```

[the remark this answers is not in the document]
sungam3r:
```c#
using System;

namespace ConsoleApp8
{
    class Program
    {
        {
            var a = GC.GetAllocatedBytesForCurrentThread();

            var loc1 = new Location1(10, 20);

            for (int i = 0; i < 1000000; ++i)
                _ = loc1.GetHashCode();

            var b = GC.GetAllocatedBytesForCurrentThread();

            Console.WriteLine(b - a);

            var loc2 = new Location2(10, 20);

            for (int i = 0; i < 1000000; ++i)
                _ = loc2.GetHashCode();

            var c = GC.GetAllocatedBytesForCurrentThread();

            Console.WriteLine(c - b);
            Console.ReadKey();
        }
    }

    public readonly struct Location1
    {
        public Location1(int start, int end)
        {
            Start = start;
            End = end;
        }

        public override int GetHashCode() => (Start, End).GetHashCode();

        public int Start { get; }

        public int End { get; }
    }

    public readonly struct Location2
    {
        public Location2(int start, int end)
        {
            Start = start;
            End = end;
        }

        public int Start { get; }

        public int End { get; }
    }
}
```
изображение

So...

src/GraphQL/Language/AST/SourceLocation.cs
sungam3r: It inherits docs from interface:
изображение

PR #2115, opened 2021-01-04
SourceLocation additional improvements

Shane32:


src/GraphQL/Execution/ExecutionError.cs
```diff
@@ -141,18 +141,11 @@ public static class ExecutionErrorExtensions
...
-                var location = new Location(new Source(document.OriginalQuery), abstractNode.SourceLocation.Start);
-                error.AddLocation(location.Line, location.Column);
-            }
-            else if (abstractNode.SourceLocation.Line > 0 && abstractNode.SourceLocation.Column > 0)
-            {
-                error.AddLocation(abstractNode.SourceLocation.Line, abstractNode.SourceLocation.Column);
-            }
+            var location = new Location(new Source(document.OriginalQuery), abstractNode.SourceLocation.Start);
+            error.AddLocation(location.Line, location.Column);
         }
```
Shane32, line 149: The `document` argument is always passed to this method, so the `else` condition (the only part of the codebase that used `Line` and `Column`) was never executed.

src/GraphQL/Language/AST/SourceLocation.cs
```diff
@@ -3,63 +3,32 @@ namespace GraphQL.Language.AST
...
         /// </summary>
-        public SourceLocation(int line, int column, int start = -1, int end = -1)
+        public SourceLocation(int start = -1, int end = -1)
         {
-            Line = line;
-            Column = column;
-            Start = start;
-            End = end;
+            _start = start + 1;
+            _end = end + 1;
```
Shane32, line 14: This allows for identifying an empty struct.  It should be equally as fast as adding a bool to the struct while using less memory.

src/GraphQL/Language/AST/SourceLocation.cs
```diff
@@ -3,63 +3,32 @@ namespace GraphQL.Language.AST
...
-            if (ReferenceEquals(this, obj))
-                return true;
-            if (obj.GetType() != GetType())
-                return false;
-            return Equals((SourceLocation)obj);
-        }
-
-        /// <inheritdoc/>
-        public override int GetHashCode() => (Line, Column).GetHashCode();
+        public bool IsSet => _start >= 0 && _end >= _start;
```
Shane32, line 32: We can move this check directly into the one line in validationcontext that uses it, or just assume that SourceLocation is always valid (and remove the +1 / -1 stuff).

sungam3r:
I suggest to close this PR. See https://github.com/graphql-dotnet/parser/pull/97 and related comments.

PR #2116, opened 2021-01-04
Remove INode.IsEqualTo

Shane32:
`IsEqualTo` is never used, and the various protected overloaded `Equals` members are only there to assist `IsEqualTo`.  They don't affect the system-provided `Equals` method.

### DIFF
--- a/src/GraphQL/Language/AST/Field.cs
+++ b/src/GraphQL/Language/AST/Field.cs
@@ -84,31 +84,5 @@ namespace GraphQL.Language.AST
 
         /// <inheritdoc />
         public override string ToString() => $"Field{{name='{Name}', alias='{Alias}', arguments={Arguments}, directives={Directives}, selectionSet={SelectionSet}}}";
-
-        /// <summary>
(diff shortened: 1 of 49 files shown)


src/GraphQL/Language/AST/Field.cs
```diff
@@ -88,29 +88,17 @@ public override IEnumerable<INode> Children
...
-            return Equals((Field)obj);
-        }
-
         /// <summary>
         /// Returns a new field selection node with the child field selection nodes merged with another field's child field selection nodes.
         /// </summary>
         public Field MergeSelectionSet(Field other)
         {
-            if (Equals(other))
+            if (Equals(other)) //TODO: is this check necessary?
```
sungam3r, line 101: I suggest to remove this method and unwrap its body inside single consumer: `Fields.Add(Field field)`

sungam3r: Also remove Equals of course

Shane32: I agree

Shane32: Done

sungam3r (approved):

PR #2117, opened 2021-01-04
Change NameNode to a struct

Shane32:
NameNode simply contains a string and a SourceLocation.  Since Name is not accessible via interfaces, we can be sure it is not boxed when the Name is being read.  Most of the classes that use NameNode also provide a property like Name that directly retrieves the string from the NameNode.  There are certainly times when boxing can occur, but it seems to be pretty rare.  Seems like a good candidate for a struct.  I have not yet run any performance tests.

Note that it's even possible that SourceLocation is never read and could be removed from the struct.  This would take more investigation.


Shane32:
If you like where this is heading, let's wait until SourceLocation is a struct before we fully review this PR.

docs2/site/docs/migrations/migration4.md
```diff
@@ -20,6 +20,7 @@
 * `IHaveDefaultValue.Type` has been moved to `IProvideResolvedType.Type`
 * `Connection<TNode, TEdge>.TotalCount` has been changed from an `int` to an `int?`. This allows for returning `null` indicating that the total count is unknown.
 * By default fields returned by introspection query are no longer sorted by their names. `LegacyV3SchemaComparer` can be used to switch to the old behavior.
+* `NameNode` has been changed to a struct.
```
sungam3r, line 23:
```suggestion
* `NameNode` has been changed to a readonly struct.
```

sungam3r:
Let's merge #2116 first.

sungam3r:
Waiting for #2111

Shane32:
I seriously don't know when the labeler is useful.  Every time I make a change, I have to undo stuff like this:

image

Shane32:
We most certainly don't need labels to tell us when the tests change.  They should change with every bugfix.  Documentation labels should label when we are updating documentation, which could be code comments.  The labeler doesn't know when I do that.  It really is pointless.

sungam3r:
Current rules:
```
test:
  - src/GraphQL.DataLoader.Tests/**/*
  - src/GraphQL.Dummy/**/*
  - src/GraphQL.Harness.Tests/**/*
  - src/GraphQL.Tests/**/*

CI:
  - .github/workflows/**/*

code style:
  - .editorconfig

documentation:
  - README.md
  - docs2/**/*

performance:
  - src/GraphQL.Benchmarks/**/*
```

Change rules as you see fit.

sungam3r:
#2111 merged

Shane32:
Another thing we could consider is removing NameNode entirely.

src/GraphQL/Language/AST/NameNode.cs
```diff
@@ -1,21 +1,39 @@
...
             Name = name;
+            SourceLocation = default;
         }
 
         /// <summary>
         /// Returns the contained name.
         /// </summary>
         public string Name { get; }
+
+        IEnumerable<INode> INode.Children => throw new NotImplementedException();
```
sungam3r, line 34:
```suggestion
        IEnumerable<INode> INode.Children => null;
```

To align behavior with other leaf nodes.

sungam3r:
> Another thing we could consider is removing NameNode entirely.

I'm not sure.

src/GraphQL/Language/AST/NameNode.cs
sungam3r: It's still safer.

src/GraphQL/Language/AST/NameNode.cs
```diff
@@ -1,21 +1,39 @@
+using System;
```
sungam3r, line 1:
```suggestion
```

sungam3r (approved):

PR #2121, opened 2021-01-04
Remove NameNode?

Shane32:


Shane32:
There some spots where I had changed null to default; I'll put those back if we are going to proceed.

sungam3r:
Then the information about the location will be lost.

Shane32:
> Then the information about the location will be lost.

Correct, but how/when could this possibly be used?  Within GraphQL.NET an error could be raised for many different nodes, but never a `NameNode`.  The only time an error would relate to the `NameNode` is if a name had invalid characters, and then parser would never return it to begin with, and there is no code within GraphQL.NET to detect this scenario.  So the error is always returned for the parent of the `NameNode`, not the `NameNode` itself.

However, if we want to prepare the library for third-party parsing algorithms, then that's a good reason to leave `NameNode` in there (even though it's not needed by GraphQL.NET).

Shane32:
@sungam3r close this then?

sungam3r:
Let's close for now. I think that in the future it will be possible to get rid of all conversions altogether. Ideally, I would like to get a non-memory allocating parser.

PR #2122, opened 2021-01-04
Change some CoreToVanillaConverter members to be public

Shane32:
Not sure why, but two members were private and the rest public.


sungam3r:
On the contrary, I would make all the methods possible private.

sungam3r:
3:15 a.m. now, I'm done for today.

Shane32:
> On the contrary, I would make all the methods possible private.

Fine with me.

PR #2124, opened 2021-01-05
Change CoreToVanillaConverter members to be private

Shane32:


sungam3r (approved):

Shane32:
I just was thinking that there's one other option: `protected virtual` members (vs `private static`) - so that someone could have their own converter and override the single node they need to.  What do you think?  We can still provide a static `Convert` method for the default implementation, and none of the members are exposed unless the class is overridden.  It seems like this is a spot that it might be a good design, since there are individual methods for each node type, perfect for allowing someone to override a single node.  I'll put in a PR just so we can look at it, anyway.

sungam3r:
I think nobody needs it. If someone needs it, we will do it, but for now it is better to systematically reduce the already bloated area of public api.

Shane32:
Ok

PR #2128, opened 2021-01-05
More xml comments

Shane32:
Some unused methods have been marked obsolete.

### DIFF
--- a/src/GraphQL/DocumentWriterExtensions.cs
+++ b/src/GraphQL/DocumentWriterExtensions.cs
@@ -4,6 +4,9 @@ using System.Threading.Tasks;
 
 namespace GraphQL
 {
+    /// <summary>
+    /// Provides extension functions for document writers.
+    /// </summary>
     public static class DocumentWriterExtensions
     {
         private static readonly Encoding _utf8Encoding = new UTF8Encoding(false);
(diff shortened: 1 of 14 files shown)


src/GraphQL/DocumentWriterExtensions.cs
```diff
@@ -4,6 +4,9 @@
 
 namespace GraphQL
 {
+    /// <summary>
+    /// Provides extension functions for document writers.
```
sungam3r, line 8:
```suggestion
    /// Provides extension methods for document writers.
```

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -289,6 +289,10 @@ IEnumerable<ExecutionNode> IParentExecutionNode.GetChildNodes()
     /// </summary>
     public class RootExecutionNode : ObjectExecutionNode
     {
+        /// <summary>
+        /// Initializes a new instance for the specified root graph type.
+        /// </summary>
+        /// <param name="graphType"></param>
```
sungam3r, line 295:
```suggestion
```

src/GraphQL/Execution/IExecutionContext.cs
```diff
@@ -7,6 +7,9 @@
 
 namespace GraphQL.Execution
 {
+    /// <summary>
+    /// Provides information regarding the currently executing document.
```
sungam3r, line 11: executed?

Shane32: the context is (typically) only used while the document is being executed, not afterwards.

sungam3r: right

sungam3r (approved):

PR #2133, opened 2021-01-05
Change CoreToVanillaConverter members to be protected

Shane32:


sungam3r:
What for? We've already made them private.

PR #2134, opened 2021-01-05
Validation rework

sungam3r:
The main goal of this PR is reduce memory allocations during validation. The basic idea is that `ValidationContext` stops being captured in closures. Visitors generated by validation rules now store their state in `ValidationContext`.

Also fixes benchmarks.

Results before:

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.16299.726 (1709/FallCreatorsUpdate/Redstone3)
Intel Core i7 CPU 920 2.67GHz (Nehalem), 1 CPU, 8 logical and 4 physical cores
Frequency=2607421 Hz, Resolution=383.5207 ns, Timer=TSC
.NET Core SDK=5.0.101
  [Host]     : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT
  DefaultJob : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT


```
|        Method |        Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------- |------------:|----------:|----------:|-------:|------:|------:|----------:|
| Introspection | 5,080.81 μs | 58.344 μs | 54.575 μs |      - |     - |     - | 863.53 KB |
|          Hero |    70.00 μs |  0.136 μs |  0.121 μs | 5.3711 |     - |     - |  22.04 KB |

Results after:

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.16299.726 (1709/FallCreatorsUpdate/Redstone3)
Intel Core i7 CPU 920 2.67GHz (Nehalem), 1 CPU, 8 logical and 4 physical cores
Frequency=2607421 Hz, Resolution=383.5207 ns, Timer=TSC
.NET Core SDK=5.0.101
  [Host]     : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT
  DefaultJob : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT


```
|        Method |        Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------- |------------:|----------:|----------:|-------:|------:|------:|----------:|
[...]

PR #2136, opened 2021-01-05
Validation improvements

Shane32:
Makes many validation rules allocation-free instances

### DIFF
--- a/src/GraphQL/Validation/TypeInfo.cs
+++ b/src/GraphQL/Validation/TypeInfo.cs
@@ -1,5 +1,4 @@
 using System.Collections.Generic;
-using System.Linq;
 using GraphQL.Language.AST;
 using GraphQL.Types;
 
@@ -34,7 +33,14 @@ namespace GraphQL.Validation
         /// <returns></returns>
         public INode[] GetAncestors()
         {
-            return _ancestorStack.Skip(1).Reverse().ToArray();
+            var ancestorArray = _ancestorStack.ToArray();
+            var c = ancestorArray.Length;
+            var ret = new INode[c - 1];
+            for (int i = c-- - 1; i > 0; --i) {
+                ret[c - i] = ancestorArray[i];
+            }
+            return ret;
+            // aka: return _ancestorStack.Skip(1).Reverse().ToArray();
         }
 
         /// <summary>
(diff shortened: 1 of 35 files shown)


Shane32:
I finished my code.  Feel free to add the improvements you wanted to.

sungam3r:
Well I see that the idea of this PR the same as #2134. I see that in some places you don't use statics - where there is state. I went further and transferred the state to `ValidationContext`.

src/GraphQL/Validation/TypeInfo.cs
```diff
@@ -34,7 +33,14 @@ public TypeInfo(ISchema schema)
         /// <returns></returns>
         public INode[] GetAncestors()
         {
-            return _ancestorStack.Skip(1).Reverse().ToArray();
+            var ancestorArray = _ancestorStack.ToArray();
+            var c = ancestorArray.Length;
+            var ret = new INode[c - 1];
+            for (int i = c-- - 1; i > 0; --i) {
```
sungam3r, line 39: 😮

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -72,15 +72,12 @@ public List<VariableUsage> GetVariables(IHaveSelectionSet node)
...
-                    varRef => usages.Add(new VariableUsage(varRef, info.GetInputType()))
-                );
-            });
+            var listener = new MatchingNodeVisitor<VariableReference>(
+                (varRef, __) => usages.Add(new VariableUsage(varRef, info.GetInputType()))
+            );
 
             var visitor = new BasicVisitor(info, listener);
-            visitor.Visit(node);
+            visitor.Visit(node, null);
```
sungam3r, line 80: Why null?

Shane32:
Suggest closing in favor of #2139

sungam3r:
As you wish.

src/GraphQL/Validation/ValidationContext.cs
Shane32: The context was not used in the lambda (the code is up just a few lines), nor was it available in the method.

Shane32:
> Well I see that the idea of this PR the same as #2134. I see that in some places you don't use statics - where there is state. I went further and transferred the state to `ValidationContext`.

Yeah I think most of our code is exactly the same.

sungam3r:
Closing?

Shane32:
Closed in favor of #2139

PR #2137, opened 2021-01-05
Benchmark updates

Shane32:
Code by sungam3r in #2134


sungam3r:
It could be merged even into master.

PR #2143, opened 2021-01-06
IsValidLiteralValue small optimization

sungam3r:


src/GraphQL/GraphQLExtensions.cs
```diff
@@ -153,7 +153,9 @@ public static IGraphType BuildNamedType(this Type type, Func<Type, IGraphType> r
                        $"Expected non-null value, {nameof(resolve)} delegate return null for \"${type}\"");
         }
 
-        public static IEnumerable<string> IsValidLiteralValue(this IGraphType type, IValue valueAst, ISchema schema)
+        private static readonly string[] _foundNull = new[] { "Expected non-null value, found null" };
```
sungam3r, line 156: I do not think that the ability to change the value of an array element should be considered a threat.

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -237,7 +239,7 @@ public static IEnumerable<string> IsValidLiteralValue(this IGraphType type, IVal
                     errors.AddRange(result.Select(err => $"In field \"{field.Name}\": {err}"));
                 }
 
-                return errors;
+                return errors.ToArray();
```
sungam3r, line 242: In theory, this adds allocations, but in practice there should still be less of them due to the removal of `ToList` from other places 😕

Shane32 (approved):
Looks great!

sungam3r:
but formatting...

PR #2144, opened 2021-01-06
Remove Regex

Shane32:
I ran some performance tests. It only really helps people with transient/scoped schemas...but it does help.  What do you think?

```
Before: (transient schema)

|        Method |     Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|----------:|----------:|---------:|--------:|------:|----------:|
| Introspection | 5.842 ms | 0.0275 ms | 0.0257 ms | 484.3750 | 31.2500 |     - |    2.9 MB |
|          Hero | 4.563 ms | 0.0329 ms | 0.0292 ms | 367.1875 | 31.2500 |     - |    2.2 MB |

After: (transient schema)

|        Method |     Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|----------:|----------:|---------:|--------:|------:|----------:|
| Introspection | 5.751 ms | 0.0460 ms | 0.0408 ms | 476.5625 | 31.2500 |     - |   2.87 MB |
|          Hero | 4.409 ms | 0.0335 ms | 0.0297 ms | 359.3750 | 31.2500 |     - |   2.17 MB |
```


Shane32:
FYI, if you compare the ToConstantCase method against the regex equivalent,

```

|          Method |       Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |-----------:|----------:|----------:|-------:|------:|------:|----------:|
|           RegEx | 5,721.6 ns | 110.25 ns | 122.55 ns | 0.3662 |     - |     - |    1592 B |
| CustomFunction1 |   408.7 ns |   4.79 ns |   4.48 ns | 0.2027 |     - |     - |     848 B |
| CustomFunction2 |   267.9 ns |   1.79 ns |   1.59 ns | 0.2027 |     - |     - |     848 B |
```

`CustomFunction2` looks for A-Z and a-z, etc, whereas `CustomFunction1` checks `char.IsUpper` and `char.IsLower`

Shane32:
By the way, even though this implementation is 20x faster, it could be a lot faster yet, even without using System.Buffers, which could make it faster still.

sungam3r:
Let's merge #2146 first and retest.

src/GraphQL.Tests/StringExtensionTests.cs
```diff
@@ -0,0 +1,29 @@
...
+        [InlineData("a9", "A9")]
+        [InlineData("ABc", "A_BC")]
+        [InlineData("Abc", "ABC")]
+        [InlineData("ABC", "ABC")]
+        [InlineData("aBC", "A_BC")]
+        [InlineData("TestABCHelloTest", "TEST_ABC_HELLO_TEST")]
+        [InlineData("AbCDefGh", "AB_C_DEF_GH")]
+        public void ToConstantCase(string input, string expected)
+        {
+            Assert.Equal(expected, input.ToConstantCase());
```
sungam3r, line 26:
```suggestion
            input.ToConstantCase().ShouldBe(expected);
```

src/GraphQL.Tests/Utilities/NameValidatorTests.cs
```diff
@@ -23,14 +23,22 @@ public class NameValidatorTests
...
+        [InlineData("goodName")]
+        [InlineData("name3")]
+        [InlineData("Test")]
+        [InlineData("Test_Name")]
+        [InlineData("_test")]
+        public void ValidateName_whenNameIsCorrect_DoesntthrowsArgumentOutOfRange(string validName)
         {
-            var ex = Record.Exception(() => ValidateName("goodName"));
+            var ex = Record.Exception(() => ValidateName(validName));
             ex.ShouldBeNull();
```
sungam3r, line 42:
```suggestion
            ValidateName(validName);
```

sungam3r: Also I'm not abig fan of `using static` because it's not immediately clear where the method came from. There is little savings, but questions arise.

src/GraphQL/Execution/ErrorInfoProvider.cs
```diff
@@ -133,30 +132,13 @@ private static string NormalizeErrorCode(Type exceptionType)
                 code = code.Substring("GraphQL".Length);
             }
 
-            return GetAllCapsRepresentation(code);
-        }
-
-        private static string GetAllCapsRepresentation(string str)
-        {
-            return Regex
-                .Replace(NormalizeString(str), @"([A-Z])([A-Z][a-z])|([a-z0-9])([A-Z])", "$1$3_$2$4")
```
sungam3r, line 142: It's funny. I saw this place many times but did not think to change it to `ToConstantCase`.

src/GraphQL/StringExtensions.cs
```diff
@@ -50,5 +50,51 @@ public static string ToPascalCase(this string s)
...
+                    if (char.IsUpper(s[i + 1]))
+                    {
+                        s = s.Substring(0, ++i) + '_' + s.Substring(i);
+                    }
+                }
+            }
+
+            return s.ToUpperInvariant();
+        }
+
```
sungam3r, line 98:
```suggestion
```

sungam3r:
> What do you think?

I like the changes. What is your confidence that you can just delete all the calls to `TrimGraphQLTypes`? I wanted to do it myself some time ago in one of the PRs, when I was discussing the possibility of finding special characters in the name. Apparently they shouldn't be there, but I didn't dare to just delete the calls. 

For the rest of the edits, I can say thank you for removing this garbage.

src/GraphQL/Utilities/NameValidator.cs
```diff
@@ -52,12 +48,24 @@ public static void ValidateDefault(string name, string type)
         {
             ValidateNameNotNull(name, type);
 
-            if (name.Length > 1 && name.StartsWith(RESERVED_PREFIX, StringComparison.InvariantCulture))
+            if (name.Length > 1 && name[0] == '_' && name[1] == '_')
```
sungam3r, line 51: 👍

src/GraphQL/Utilities/NameValidator.cs
```diff
@@ -52,12 +48,24 @@ public static void ValidateDefault(string name, string type)
...
             {
                 throw new ArgumentOutOfRangeException(nameof(name),
                     $"A {type} name: {name} must not begin with \"__\", which is reserved by GraphQL introspection.");
```
sungam3r, line 54:
```suggestion
                    $"A {type} name: '{name}' must not begin with \"__\", which is reserved by GraphQL introspection.");
```

src/GraphQL/Utilities/NameValidator.cs
```diff
@@ -52,12 +48,24 @@ public static void ValidateDefault(string name, string type)
...
+            {
+                c = name[i];
+                if ((c < 'a' || c > 'z') && (c < 'A' || c > 'Z') && (c < '0' || c > '9') && c != '_')
+                    ThrowMatchError();
+            }
+
+            void ThrowMatchError()
             {
                 throw new ArgumentOutOfRangeException(nameof(name),
                     $"A {type} name must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but {name} does not.");
```
sungam3r, line 71:
```suggestion
                    $"A {type} name must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but '{name}' does not.");
```

src/GraphQL/Utilities/NameValidator.cs
```diff
@@ -8,9 +7,6 @@ namespace GraphQL.Utilities
     /// </summary>
     public static class NameValidator
     {
-        private const string RESERVED_PREFIX = "__";
-        private const string NAME_RX = @"^[_A-Za-z][_0-9A-Za-z]*$";
```
sungam3r, line 12: I can only guess why there was an underscore in the regular expression, if it is already checked separately at the beginning of the method.

sungam3r: Ah, `__` != `_`

src/GraphQL/Utilities/NameValidator.cs
```diff
@@ -52,12 +48,24 @@ public static void ValidateDefault(string name, string type)
...
-            if (name.Length > 1 && name.StartsWith(RESERVED_PREFIX, StringComparison.InvariantCulture))
+            if (name.Length > 1 && name[0] == '_' && name[1] == '_')
             {
                 throw new ArgumentOutOfRangeException(nameof(name),
                     $"A {type} name: {name} must not begin with \"__\", which is reserved by GraphQL introspection.");
             }
-            if (!Regex.IsMatch(name, NAME_RX))
+
+            var c = name[0];
+            if ((c < 'a' || c > 'z') && (c < 'A' || c > 'Z') && c != '_')
```
sungam3r, line 58: I didn't get it right away, but it looks good.

src/GraphQL/StringExtensions.cs
```diff
@@ -50,5 +50,51 @@ public static string ToPascalCase(this string s)
 
             return newFirstLetter + s.Substring(1);
         }
+
+        /// <summary>
+        /// Returns a constant case version of this string. For example, converts 'StringError' into 'STRING_ERROR'.
+        /// </summary>
+        public static string ToConstantCase(this string s)
```
sungam3r, line 57: I hope that the algorithm is correct, but it is complex and I would not like to support it in this form. Also, I don't like the calls to `Substring`.

Shane32:
> I like the changes. What is your confidence that you can just delete all the calls to `TrimGraphQLTypes`? I wanted to do it myself some time ago in one of the PRs, when I was discussing the possibility of finding special characters in the name. Apparently they shouldn't be there, but I didn't dare to just delete the calls.

I'm very confident.  However, just in case there is some issue we are unaware of, I think we should put this in 4.0.  I don't see any reason to work on 3.3 when there may be a question of how it functions, and especially since we are so close to releasing 4.0.  We want the 3.2 -> 3.3 transition to be 100% seamless.

If someone is going to go through the effort of migrating to 4.0, and they encounter a bug, so be it.  We can fix it and will just release a 4.0.1.  No big deal.  Besides, we can test 4.0 beta ourselves with our own large scale projects if we want extra confidence that everything is good.  But I'm pretty confident that the call can be removed entirely.

Besides all that, with a singleton schema, it doesn't matter how long it takes the the function to execute.  It's just for the few people that have scoped schemas.  So there's no reason to put these optimizations into 3.3

src/GraphQL/StringExtensions.cs
Shane32: I can eliminate the substring calls easily, but it will make it a little more complex.  If it's the naming you don't like, I can fix that as well.  If you just want to forget the whole thing, that's fine, although this code is 20 times faster than the regex code.  Not just 5% or 10%, but a full 20x improvement, and it does make a measurable difference for scoped schemas as can be seen from the performance results for the introspection and hero tests on scoped schemas.

src/GraphQL.Tests/Utilities/NameValidatorTests.cs
Shane32: I agree, and I also agree with this change.  I was just copying what was in there... I'll fix it.

src/GraphQL/StringExtensions.cs
Shane32: I can also write it using system.buffers if we are going to add that reference.... but it won't make the code any less complex.

Shane32: I can add a whole ton of comments so people know what the code actually does .....

sungam3r: I find using `System.Memory` necessary. Comments will be very helpful.

sungam3r: Or we leave this algorithm as it is (enriching with comments), given that its results are still cached.

src/GraphQL.Tests/StringExtensionTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        [InlineData("ABc", "A_BC")]
+        [InlineData("Abc", "ABC")]
+        [InlineData("ABC", "ABC")]
+        [InlineData("aBC", "A_BC")]
+        [InlineData("TestABCHelloTest", "TEST_ABC_HELLO_TEST")]
+        [InlineData("AbCDefGh", "AB_C_DEF_GH")]
+        public void ToConstantCase(string input, string expected)
+        {
+            input.ToConstantCase().ShouldBe(expected);
+        }
```
sungam3r, line 28:
```suggestion
        }
        
        [Theory]
        [InlineData(typeof(void), "Void")]
        [InlineData(typeof(int), "Int32")]
        [InlineData(typeof(Dictionary<string, bool>), "Dictionary<String,Boolean>")]
        [InlineData(typeof(List<Dictionary<string, HashSet<DateTime>>>), "List<Dictionary<String,HashSet<DateTime>>>")]
        [InlineData(typeof(Dictionary<,>), "Dictionary<TKey,TValue>")]       
        [InlineData(typeof(List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<int>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>), "List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<List<Int32>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>")]
        public void GetFriendlyName_Should_Return_Expected_Results(Type source, string expected)
        {
            source.GetFriendlyName().ShouldBe(expected);
        }
```

This is a test from another PR that I've experimented with `System.Memory`.

Shane32:
Looks like I need to update all the test results after you changed the test messages....

Shane32:
I'll postpone this after we merge changes for ToCamelCase.  Or, if you want to pick up this PR that's fine too.  I was just trying to get rid of regex, knowing how slow it is.

Shane32:
Waiting for #2146

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/pull/2150

Shane32:
@sungam3r If you want to finish this PR up, rewriting it with stackalloc or whatever, let me know. I've just been a bit busy.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -321,8 +320,9 @@ namespace GraphQL
     public static class StringExtensions
     {
         public static string ToCamelCase(this string s) { }
-        public static string ToFormat(this string format, params object[] args) { }
+        public static string ToConstantCase(this string s) { }
         public static string ToPascalCase(this string s) { }
+        public static string TrimGraphQLTypes(this string name) { }
```
Shane32, line 325: It seemed to me that this method (TrimGraphQLTypes) pertains to the library and could be useful, and as it was only a couple lines of code, I left it in the codebase.  However, it is not being used, so we can delete it.

Shane32:
> Or we leave this algorithm as it is (enriching with comments), given that its results are still cached.

Done.  Simplified the algorithm a little so it's easier to follow also.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: 50/50

src/GraphQL/StringExtensions.cs
```diff
@@ -64,5 +64,47 @@ private static string FastChangeFirstLetter(char newFirstLetter, string s)
...
+
+        /// <summary>
+        /// Returns a constant case version of this string. For example, converts 'StringError' into 'STRING_ERROR'.
+        /// </summary>
+        public static string ToConstantCase(this string value)
+        {
+            //aka: return Regex.Replace(s, @"([A-Z])([A-Z][a-z])|([a-z0-9])([A-Z])", "$1$3_$2$4").ToUpperInvariant();
+            int i;
+            int strLength = value.Length;
+            // iterate through each character in the string, stopping a character short of the end
```
sungam3r, line 76: _stopping a character short of the end_ ?

src/GraphQL/StringExtensions.cs
```diff
@@ -64,5 +64,47 @@ private static string FastChangeFirstLetter(char newFirstLetter, string s)
             s.AsSpan().Slice(1).CopyTo(buffer.Slice(1));
             return buffer.ToString();
         }
+
+        /// <summary>
+        /// Returns a constant case version of this string. For example, converts 'StringError' into 'STRING_ERROR'.
+        /// </summary>
+        public static string ToConstantCase(this string value)
```
sungam3r, line 71: The algorithm has become much clearer. As I see it, it is not difficult to change it for using spans. Will you do it? You just need to create a stackalloc/heap buffer and copy symbols to it one-by-one each step.

sungam3r: Of cource in this case there will be no need to concatenate `value` and use `.Substring`.

sungam3r: Buffer size can be roughly calculated in advance as `value.Length + value.Length/2`.

Shane32: I'm putting my effort into allowing input graph types to have resolvers.

Feel free to edit this branch as you see fit.

Shane32: > Buffer size can be roughly calculated in advance as `value.Length + value.Length/2`.

I was thinking that it would probably be very fast for the CPU to scan for matching patterns and then know exactly how large to make the buffer.  Performance testing and exact implementation would matter.  But with stackalloc it should not be necessary to know the exact length.

Shane32: Keep in mind that there's a 40% improvement to be made by switching IsUpper to `(char >= 'A' && char <= 'Z')` and similar, but then it does not match on foreign characters.  I opted to keep the function generic for all uses.

src/GraphQL/StringExtensions.cs
Shane32:
```
Offset: 0123456
String: TEST
```
Because the iterator needs to always check the current and following character, it scans from index 0 to index 2.

If the comment is unclear please update it as you see fit.

src/GraphQL/StringExtensions.cs
Shane32: I really just wanted to remove regex, as even this "slow" code represents a 20x improvement when using a string with 5 words.  (It is optimal if the string has one word, and should be nearly optimal if the string has two words, depending on how much .NET optimizes substring and concatenation.)

Shane32: It actually works better than the regex.  The regex fails one of the tests, and the regex does not match foreign upper/lowercase characters.

docs2/site/docs/migrations/migration4.md
```diff
@@ -24,6 +24,7 @@
 * `IHaveDefaultValue.Type` has been moved to `IProvideResolvedType.Type`
 * `Connection<TNode, TEdge>.TotalCount` has been changed from an `int` to an `int?`. This allows for returning `null` indicating that the total count is unknown.
 * By default fields returned by introspection query are no longer sorted by their names. `LegacyV3SchemaComparer` can be used to switch to the old behavior.
+* Various methods within `StringUtils` have been removed; please use extension methods within `StringExtensions` instead.
```
Shane32, line 27:
```suggestion
```

Shane32:
> I intentionally moved your added lines at the end of file from here because of example code below, which correspond to the note above.

:+1: Fixed

sungam3r (approved):

Issue #1127, opened 2019-05-03
Default Values for inputs?


scott-mac:
Many of my input types share the same fields (Created and LastModifed). So I created a base class for my input types to inherit from. I'm trying to set the default values on these fields as follows:

 ```
public class BaseInputType : InputObjectGraphType {
        public BaseInputType() {
            Field<DateTimeGraphType>("Created");
            Field<DateTimeGraphType>("LastModified");

            this.Fields.SingleOrDefault(f => f.Name == "Created").DefaultValue = DateTime.UtcNow;
            this.Fields.SingleOrDefault(f => f.Name == "LastModified").DefaultValue = DateTime.UtcNow;

    }
```

If the user doesn't pass in values for Created and LastModified then I'd like to have default values set but it doesn't seem to honor this. Is there something else I need to tweak? Thanks

sungam3r:
Please show how you read such input argument from query arguments in your resolver.

sungam3r:
@scott-mac, it seems to me that the fix relates specifically to the problem you are describing.

PR #2160, opened 2021-01-09
Don't override argument null literals with default values

### LINKED ISSUE #2159: Default values override supplying an argument literal of null
See rule for input coercion for scalars:

> For all types below, with the exception of Non‐Null, **if the explicit value null is provided, then the result of input coercion is null.**

See rule for input objects:

> If no value is provided for a defined input object field and that field definition provides a default value, the default value should be used. If no default value is provided and the input object field’s type is non‐null, an error should be thrown. Otherwise, if the field is not required, then no entry is added to the coerced unordered map.
>
> **If the value null was provided for an input object field, and the field’s type is not a non‐null type, an entry in the coerced unordered map is given the value null**. In other words, there is a semantic difference between the explicitly provided value null versus having not provided a value.
>
> If a literal value is provided for an input object field, an entry in the coerced unordered map is given the result of coercing that value according to the input coercion rules for the type of that field.
>
> If a variable is provided for an input object field, the runtime value of that variable must be used. If the runtime value is null and the field type is non‐null, a field error must be thrown. If no runtime value is provided, the variable definition’s default value should be used. If the variable definition does not provide a default value, the input object field definition’s default value should be used.

Also see this pseudo-code for Coercing Field Arguments: https://spec.graphql.org/June2018/#sec-Coercing-Field-Arguments

> ...
> Let hasValue be true if argumentValues provides a value for the name argumentName.
> ...
> **If hasValue is not true** and defaultValue exists (including null):
> - Add an entry to coercedValues named argumentName with the value defaultValue.

Currently what happens is that the null value is overridden by the field default value.  This is incorrect according to the spec.

@sungam3r I assume we should fix in 3.3?

Shane32:
Fixes #2159

Requires a signature change to CoerceValue and ValueFor.  They are added optional fields, so although not binary-compatible, it is "compile-compatible".

Changed one test that relied on incorrect behavior.  Added a number of tests that verify this fix.

Should note this behavior and signature change in the release notes.

### DIFF
--- a/docs2/site/docs/guides/known-issues.md
+++ b/docs2/site/docs/guides/known-issues.md
@@ -183,7 +183,7 @@ data from your database within the fetch delegate, use an inner join to retrieve
 ## Known Issues
 
 `IResolveFieldContext.HasArgument` will return `true` for all arguments where `GetArgument` does not return `null`.
-It cannot identify which arguments have been provided a null value compared to arguments which were not provided.
+It cannot identify which arguments have been provided a `null` value compared to arguments which were not provided.
 
 ## Common Errors
 
--- a/docs2/site/docs/guides/known-issues.md
+++ b/docs2/site/docs/guides/known-issues.md
@@ -180,6 +180,11 @@ This is done within your database queries; it is not a function of the dataloade
 `CollectionBatchDataLoader` as you would for a one-to-many relationship; then when you are loading
 data from your database within the fetch delegate, use an inner join to retrieve the proper data.
 
+## Known Issues
+
+`IResolveFieldContext.HasArgument` will return `true` for all arguments where `GetArgument` does not return `null`.
+It cannot identify which arguments have been provided a null value compared to arguments which were not provided.
+
 ## Common Errors
 
 ### Synchronous operations are disallowed.
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -93,8 +93,12 @@ namespace GraphQL.Tests.Bugs
         [Fact]
         public void Input_Enum_OverrideDefault() => AssertQuerySuccess("{ inputWithDefault(arg: SLEEPY) }", @"{ ""inputWithDefault"": ""Sleepy"" }");
 
+        // For non-nullable struct types, GetArgument coerces null to default(T), which is Grumpy
         [Fact]
         public void Input_Enum_NullDefault() => AssertQuerySuccess("{ inputWithDefault(arg: null) }", @"{ ""inputWithDefault"": ""Grumpy"" }");
+
+        [Fact]
+        public void Input_Enum_NullDefault_Nullable() => AssertQuerySuccess("{ inputWithDefaultNullable(arg: null) }", @"{ ""inputWithDefaultNullable"": null }");
 
         [Fact]
         public void Input_Enum_MissingRequired() => AssertQueryWithError(@"{ inputRequired }", null, "Argument \u0022arg\u0022 of type \u0022Bug1699Enum!\u0022 is required for field \u0022inputRequired\u0022 but not provided.", 1, 3, (object[])null, code: "PROVIDED_NON_NULL_ARGUMENTS", number: ProvidedNonNullArgumentsError.NUMBER);
@@ -163,6 +167,10 @@ namespace GraphQL.Tests.Bugs
                 "inputWithDefault",
                 arguments: new QueryArguments(new QueryArgument<EnumerationGraphType<Bug1699Enum>> { Name = "arg", DefaultValue = Bug1699Enum.Happy }),
                 resolve: ctx => ctx.GetArgument<Bug1699Enum>("arg").ToString());
+            Field<StringGraphType>(
+                "inputWithDefaultNullable",
+                arguments: new QueryArguments(new QueryArgument<EnumerationGraphType<Bug1699Enum>> { Name = "arg", DefaultValue = Bug1699Enum.Happy }),
+                resolve: ctx => ctx.GetArgument<Bug1699Enum?>("arg")?.ToString());
             Field<StringGraphType>(
                 "inputRequired",
                 arguments: new QueryArguments(new QueryArgument<NonNullGraphType<EnumerationGraphType<Bug1699Enum>>> { Name = "arg" }),
--- a/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
@@ -94,6 +94,9 @@ namespace GraphQL.Tests.Bugs
         public void Input_Enum_OverrideDefault() => AssertQuerySuccess("{ inputWithDefault(arg: SLEEPY) }", @"{ ""inputWithDefault"": ""Sleepy"" }");
 
         [Fact]
+        public void Input_Enum_NullDefault() => AssertQuerySuccess("{ inputWithDefault(arg: null) }", @"{ ""inputWithDefault"": ""Grumpy"" }");
+
+        [Fact]
         public void Input_Enum_MissingRequired() => AssertQueryWithError(@"{ inputRequired }", null, "Argument \u0022arg\u0022 of type \u0022Bug1699Enum!\u0022 is required for field \u0022inputRequired\u0022 but not provided.", 1, 3, (object[])null, code: "PROVIDED_NON_NULL_ARGUMENTS", number: ProvidedNonNullArgumentsError.NUMBER);
 
         [Fact]
--- a/src/GraphQL.Tests/Bugs/Bug2159.cs
+++ b/src/GraphQL.Tests/Bugs/Bug2159.cs
@@ -1,0 +1,118 @@
+using System;
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using GraphQL.Validation;
+using GraphQL.Validation.Errors;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/2159
+    public class Bug2159 : QueryTestBase<Bug2159Schema>
+    {
+        private void AssertQueryWithError(string query, string result, string message, int line, int column, string path, Exception exception = null, string code = null, string inputs = null, string number = null)
+            => AssertQueryWithError(query, result, message, line, column, new object[] { path }, exception, code, inputs, number);
+
+        private void AssertQueryWithError(string query, string result, string message, int line, int column, object[] path, Exception exception = null, string code = null, string inputs = null, string number = null)
+        {
+            ExecutionError error;
+            if (number != null)
+                error = new ValidationError(null, number, message);
+            else
+                error = exception == null ? new ExecutionError(message) : new ExecutionError(message, exception);
+            if (line != 0)
+                error.AddLocation(line, column);
+            error.Path = path;
+            if (code != null)
+                error.Code = code;
+            var expected = CreateQueryResult(result, new ExecutionErrors { error });
+            AssertQueryIgnoreErrors(query, expected, inputs?.ToInputs(), renderErrors: true, expectedErrorCount: 1);
+        }
+
+        [Fact]
+        public void Direct_Literal_Default() => AssertQuerySuccess("{ testValue }", @"{ ""testValue"": ""defaultValue"" }");
+
+        [Fact]
+        public void Direct_Literal_Specified() => AssertQuerySuccess("{ testValue(arg: \"hello\") }", @"{ ""testValue"": ""hello"" }");
+
+        [Fact]
+        public void Direct_Literal_Null() => AssertQuerySuccess("{ testValue(arg: null) }", @"{ ""testValue"": null }");
+
+        // If a variable is provided for an input object field, the runtime value of that variable must be used.
+        // If the runtime value is null and the field type is non‐null, a field error must be thrown. If no runtime
+        // value is provided, the variable definition’s default value should be used. If the variable definition
+        // does not provide a default value, the input object field definition’s default value should be used.
+        [Fact]
+        public void Direct_Variable_FieldDefault() => AssertQuerySuccess("query($input: String) { testValue(arg: $input) }", @"{ ""testValue"": ""defaultValue"" }");
+
+        [Fact]
+        public void Direct_Variable_VarDefault() => AssertQuerySuccess("query($input: String = \"varDefault\") { testValue(arg: $input) }", @"{ ""testValue"": ""varDefault"" }");
+
+        [Fact]
+        public void Direct_Variable_Specified() => AssertQuerySuccess("query($input: String = \"varDefault\") { testValue(arg: $input) }", @"{ ""testValue"": ""hello"" }", "{\"input\":\"hello\"}".ToInputs());
+
+        [Fact]
+        public void Direct_Variable_Null() => AssertQuerySuccess("query($input: String = \"varDefault\") { testValue(arg: $input) }", @"{ ""testValue"": null }", "{\"input\":null}".ToInputs());
+
+
+        [Fact]
+        public void Object_Literal_Default() => AssertQuerySuccess("{ testObject }", @"{ ""testObject"": ""defaultValue"" }");
+
+        [Fact]
+        public void Object_Literal_Specified() => AssertQuerySuccess("{ testObject(arg: {value:\"hello\"}) }", @"{ ""testObject"": ""hello"" }");
+
+        [Fact]
+        public void Object_Literal_Null() => AssertQuerySuccess("{ testObject(arg: {value:null}) }", @"{ ""testObject"": null }");
+
+        // If a variable is provided for an input object field, the runtime value of that variable must be used.
+        // If the runtime value is null and the field type is non‐null, a field error must be thrown. If no runtime
+        // value is provided, the variable definition’s default value should be used. If the variable definition
+        // does not provide a default value, the input object field definition’s default value should be used.
+        [Fact]
+        public void Object_Variable_Default() => AssertQuerySuccess("query($input: Bug2159Object) { testObject(arg: $input) }", @"{ ""testObject"": ""defaultFieldValue"" }", "{\"input\":{}}".ToInputs());
+
+        [Fact]
+        public void Object_Variable_Specified() => AssertQuerySuccess("query($input: Bug2159Object) { testObject(arg: $input) }", @"{ ""testObject"": ""hello"" }", "{\"input\":{\"value\":\"hello\"}}".ToInputs());
+
+        [Fact]
+        public void Object_Variable_Null() => AssertQuerySuccess("query($input: Bug2159Object) { testObject(arg: $input) }", @"{ ""testObject"": null }", "{\"input\":{\"value\":null}}".ToInputs());
+    }
+
+    public class Bug2159Schema : Schema
+    {
+        public Bug2159Schema()
+        {
+            Query = new Bug2159Query();
+        }
+    }
+
+    public class Bug2159Query : ObjectGraphType
+    {
+        public Bug2159Query()
+        {
+            Field<StringGraphType>(
+                "testValue",
+                resolve: ctx => ctx.GetArgument<string>("arg"),
+                arguments: new QueryArguments(
+                    new QueryArgument(typeof(StringGraphType)) { Name = "arg", DefaultValue = "defaultValue" }));
+            Field<StringGraphType>(
+                "testObject",
+                resolve: ctx => ctx.GetArgument<Bug2159Object>("arg")?.Value,
+                arguments: new QueryArguments(
+                    new QueryArgument(typeof(Bug2159ObjectGraphType)) { Name = "arg", DefaultValue = new Bug2159Object { Value = "defaultValue" } }));
+        }
+    }
+
+    public class Bug2159Object
+    {
+        public string Value { get; set; }
+    }
+
+    public class Bug2159ObjectGraphType : InputObjectGraphType<Bug2159Object>
+    {
+        public Bug2159ObjectGraphType()
+        {
+            Field(x => x.Value, true).DefaultValue("defaultFieldValue");
+        }
+    }
+}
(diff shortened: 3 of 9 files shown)


src/GraphQL.Tests/Bugs/Issue1127.cs
```diff
@@ -49,11 +49,11 @@ public Issue127Query()
                 resolve: ctx =>
                 {
                     ctx.Arguments["s1"].ShouldBe("def1");
-                    ctx.Arguments["s2"].ShouldBe("def2");
+                    ctx.Arguments["s2"].ShouldBeNull();
```
Shane32, line 52: Since the query specified `s2: null`, it was testing for incorrect behavior.  Also, issue #1127 was opened because `ctx.Arguments` contained entries without the default values when an argument was not specified.  This is still true, so this change isn't really breaking issue #1127.

src/GraphQL.Tests/Bugs/Issue1127.cs
```diff
@@ -49,11 +49,11 @@ public Issue127Query()
...
                     ctx.Arguments["s3"].ShouldBe("aaa");
 
                     ctx.Arguments["input1"].ShouldBe(1);
-                    ctx.Arguments["input2"].ShouldBe(2);
+                    ctx.Arguments["input2"].ShouldBeNull();
```
Shane32, line 56: Ditto here.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -125,9 +132,9 @@ public static object GetVariableValue(Document document, ISchema schema, Variabl
                 throw;
             }
 
-            if (input == null && variable.DefaultValue != null)
+            if (input == null)
             {
-                return variable.DefaultValue.Value;
+                return null;
             }
```
Shane32, line 138: This change represents a semantic change from 3.2 to 3.3.  Specifically, GetVariableValue now returns `null` for variables set to null rather than the variable's default value.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -97,9 +97,16 @@ public static Variables GetVariableValues(Document document, ISchema schema, Var
...
+                    if (inputs.TryGetValue(v.Name, out var variableValue))
+                    {
+                        variable.Value = GetVariableValue(document, schema, v, variableValue);
+                    }
+                    else
+                    {
+                        var value = GetVariableValue(document, schema, v, v.DefaultValue?.Value);
+                        if (value != null)
+                            variable.Value = value;
+                    }
```
Shane32, line 109: Checking for a default variable value has now been moved out of `GetVariableValue` into `GetVariableValues`.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -283,21 +285,29 @@ public static void AssertValidVariableValue(ISchema schema, IGraphType type, obj
         /// <summary>
         /// Coerces a variable value to a compatible .NET type for the variable's graph type.
         /// </summary>
-        public static object CoerceValue(ISchema schema, IGraphType type, IValue input, Variables variables = null)
+        public static object CoerceValue(ISchema schema, IGraphType type, IValue input, Variables variables = null, object fieldDefault = null)
```
Shane32, line 288: CoerceValue now accepts an additional argument specifying the default field value if none was specified.  Previously, it would return null if none was specified.  As such, the semantics of `CoerceValue` have not changed with this PR unless you supply new additional argument.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -269,12 +276,7 @@ public static void AssertValidVariableValue(ISchema schema, IGraphType type, obj
                 var value = astArguments?.ValueFor(arg.Name);
                 var type = arg.ResolvedType;
 
-                var coercedValue = CoerceValue(schema, type, value, variables) ?? arg.DefaultValue;
-
-                if (coercedValue != null)
-                {
-                    values[arg.Name] = coercedValue;
-                }
+                values[arg.Name] = CoerceValue(schema, type, value, variables, arg.DefaultValue);
```
Shane32, line 279: Previously, only arguments that were assigned a value were stored in the dictionary.  Now all arguments are stored in the dictionary after being assigned the correct default value.  This could represent a semantic difference, if someone is attempting to use `HasArgument` to determine if an argument was supplied or not.

However, not storing null values is semantically wrong as well, since supplying a value of `null` should result in `HasArgument` returning `true`.

And finally, not storing values that are not supplied doesn't work either, because it breaks the issue raised in #1127 where `context.Arguments` did not contain entries for arguments that should be set to their default values.

Ideally, I think that all fields should be stored in `context.Arguments`, but `HasArgument` should somehow be able to return `false` if the value was not supplied to the field (whether a default value was supplied or not) or, when it references a variable, not specified by either the inputs or a variable's default value.  This would be a bit more complicated to implement.

Shane32:
@sungam3r I've finished my self-review.  I'd be glad of your comments.  If you think we should move this to 4.0 and/or make additional changes, please let me know.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -63,7 +63,8 @@ public static bool HasArgument(this IResolveFieldContext context, string name)
         {
             var isIntrospection = context.ParentType == null ? context.FieldDefinition.IsIntrospectionField() : context.ParentType.IsIntrospectionType();
             var argumentName = isIntrospection ? name : (context.Schema?.NameConverter.NameForArgument(name, context.ParentType, context.FieldDefinition) ?? name);
-            return context.Arguments?.ContainsKey(argumentName) ?? false;
+            object value = null;
+            return ((context.Arguments?.TryGetValue(argumentName, out value) ?? false) && value != null);
```
Shane32, line 67: This patch should result in unchanged behavior for `HasArgument`

sungam3r:
I will review tonight. A bit busy with https://github.com/graphql-dotnet/parser/pull/101.

Shane32:
> I will review tonight. A bit busy with [graphql-dotnet/parser#101](https://github.com/graphql-dotnet/parser/pull/101).

Thanks.  I'm waiting on this to continue my work on allowing input resolvers, so please take a look when you can.

docs2/site/docs/guides/known-issues.md
```diff
@@ -180,6 +180,11 @@ This is done within your database queries; it is not a function of the dataloade
 `CollectionBatchDataLoader` as you would for a one-to-many relationship; then when you are loading
 data from your database within the fetch delegate, use an inner join to retrieve the proper data.
 
+## Known Issues
+
+`IResolveFieldContext.HasArgument` will return `true` for all arguments where `GetArgument` does not return `null`.
+It cannot identify which arguments have been provided a null value compared to arguments which were not provided.
```
sungam3r, line 186:
```suggestion
It cannot identify which arguments have been provided a `null` value compared to arguments which were not provided.
```

sungam3r: Is it? The parser has this information. If you mean `field(arg1: null, arg2: 5)` vs `field(arg2: 5)`

src/GraphQL/Language/AST/Variables.cs
```diff
@@ -27,12 +27,14 @@ public void Add(Variable variable)
         }
 
         /// <summary>
-        /// Returns the first variable with a matching name, or <see langword="null"/> if none are found.
+        /// Returns the first variable with a matching name, or <paramref name="defaultValue"/> if none are found.
         /// </summary>
-        public object ValueFor(string name)
+        public object ValueFor(string name, object defaultValue = null)
         {
             var variable = _variables?.FirstOrDefault(v => v.Name == name);
```
sungam3r, line 34: I'm not against LINQ, but I've already had a strong reaction to such comparisons during my experiments with performance. Are we going to get rid of such calls?

src/GraphQL/Language/AST/Variables.cs
```diff
@@ -27,12 +27,14 @@ public void Add(Variable variable)
...
-            return variable?.Value;
+            if (variable != null && variable.ValueSpecified)
+                return variable.Value;
+            return defaultValue;
```
sungam3r, line 37:
```suggestion
            return variable != null && variable.ValueSpecified
                ? variable.Value
                : defaultValue;
```

src/GraphQL/Language/AST/Variables.cs
Shane32: I would hope so. Now I’m just trying to fix a bug in 3.3.

docs2/site/docs/guides/known-issues.md
Shane32: It is now, anyway. I can check it if was this way in 3.2

sungam3r: In my opinion, this is just a flaw that can be corrected, and not some kind of fundamental design flaw.

sungam3r:
I will continue reviewing in the morning, now it's time to sleep. And tomorrow is a first working day after new year holidays. The vacation is over. Now there won't be so much (relatively) time for OS. We managed to do a lot these days, but we didn't manage to do even more things. No matter how long it takes, it's still not enough.

docs2/site/docs/guides/known-issues.md
Shane32: Agreed. That’s why it’s in “known issues” and not “faq”

sungam3r: I didn't understand what you want to say. Are you documenting this behavior or will it be fixed?

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -93,6 +93,9 @@ private void AssertQueryWithError(string query, string result, string message, i
         [Fact]
         public void Input_Enum_OverrideDefault() => AssertQuerySuccess("{ inputWithDefault(arg: SLEEPY) }", @"{ ""inputWithDefault"": ""Sleepy"" }");
 
+        [Fact]
+        public void Input_Enum_NullDefault() => AssertQuerySuccess("{ inputWithDefault(arg: null) }", @"{ ""inputWithDefault"": ""Grumpy"" }");
```
sungam3r, line 97: I think I understand the logic, but it seems to me contradictory (at least for value types now). We now distinguish between cases when the argument is specified in request as `null` explicitly and when it is not specified at all (so default value will be used either from variable or field definition). In case of nullable enumeration graphtype for input I expect that `context.GetArgument<MyEnum>("arg")` **should** throw if `null` was explicitly specified instead of return default value (0/Grumpy in this case). Note that if your argument is required (`NonNullGraphType<EnumerationGraphType<MyEnum>>` then you free to call `context.GetArgument<MyEnum>("arg")` and **always** expect some valid value. But if your argument defined as `EnumerationGraphType<MyEnum>` (optional) then it will be more correct to call `context.GetArgument<MyEnum?>("arg")`. In my opinion this is logical.

This also applies to all value types generic parameters passed to `GetArgument`.

Also please add test for `{ inputRequired(arg: null) }` and `{ inputRequiredWithDefault(arg: null) }`

sungam3r: Well... Now I see that `GetArgument` has additional ` TType defaultValue = default` parameter. Then that explains everything. Ok. 

So with provided `arg: null`
`context.GetArgument<MyEnum>("arg")` should yield `(MyEnum)0`
and
`context.GetArgument<MyEnum?>("arg")` should yield `null`

Do you agree? If so, let's add tests for nullable enums (and probably for other nullable value types - `int?`, `bool?`).

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: After looking at these methods, my intuition tells me that they should be rewritten in v4.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -283,21 +285,29 @@ public static void AssertValidVariableValue(ISchema schema, IGraphType type, obj
...
-        public static object CoerceValue(ISchema schema, IGraphType type, IValue input, Variables variables = null)
+        public static object CoerceValue(ISchema schema, IGraphType type, IValue input, Variables variables = null, object fieldDefault = null)
         {
             if (type is NonNullGraphType nonNull)
             {
-                return CoerceValue(schema, nonNull.ResolvedType, input, variables);
+                return CoerceValue(schema, nonNull.ResolvedType, input, variables, fieldDefault);
             }
 
-            if (input == null || input is NullValue)
```
sungam3r, line 293: Excellent! I remember looking at this place once and I didn't like it because it merges two different cases into one.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -331,7 +341,7 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
                     var objectField = objectValue.Field(field.Name);
                     if (objectField != null)
                     {
-                        obj[field.Name] = CoerceValue(schema, field.ResolvedType, objectField.Value, variables) ?? field.DefaultValue;
+                        obj[field.Name] = CoerceValue(schema, field.ResolvedType, objectField.Value, variables);
```
sungam3r, line 344:
```suggestion
                        obj[field.Name] = CoerceValue(schema, field.ResolvedType, objectField.Value, variables, field.DefaultValue);
```

?

If so, then we do not have the required test.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
Shane32: Exactly.  Here with the enum tests, I'll add one with nullable equals `null` and change the existing one to say `(MyEnum)0` so the logic is clearer.  I added a whole bunch of tests for nullability in another file.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: > However, not storing null values is semantically wrong as well, since supplying a value of `null` should result in `HasArgument` returning `true`.

Agree.

> Ideally, I think that all fields should be stored in context.Arguments, but HasArgument should somehow be able to return false if the value was not supplied to the field (whether a default value was supplied or not) or, when it references a variable, not specified by either the inputs or a variable's default value. This would be a bit more complicated to implement.

`Dictionary<string, object> GetArgumentValues` 
->
`Dictionary<string, ArgumentValue> GetArgumentValues`

```c#
public struct ArgumentValue
{
  object Value { get; }
  ArgumentSource Source { get; } (or ArgumentKind)
}

public enum ArgumentSource (or ArgumentKind)
{
  Argument, (or Literal, Explicit)
  FieldDefault,
  Variable
}
```

` public static bool HasArgument(this IResolveFieldContext context, string name)`
->
` public static bool HasArgument(this IResolveFieldContext context, string name, ArgumentSource source)`

\+
` public static bool HasArgument(this IResolveFieldContext context, string name)` **may** be just an overload for returning any argument kind.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Not in this case.  I looked at this a bunch of times also to be sure.  But if the field exists in the object array (`objectField != null`), then we want to capture it's value, even if it's null.  So the default value does not apply.

Shane32: A few lines below you can see if the field was not specified, and if a default value was specified, then add it to the dictionary (aka map).

Shane32: The behavior according to the spec is exceedingly specific here:

> If no value is provided for a defined input object field and that field definition provides a default value, the default value should be used. If no default value is provided and the input object field’s type is non‐null, an error should be thrown. Otherwise, if the field is not required, then no entry is added to the coerced unordered map.

So as you can see we are required to not add a null value to the dictionary if there is no dictionary and no default value, etc.

Shane32: But if it's null:

> If the value null was provided for an input object field, and the field’s type is not a non‐null type, an entry in the coerced unordered map is given the value null. In other words, there is a semantic difference between the explicitly provided value null versus having not provided a value.

sungam3r: > But if the field exists in the object array (objectField != null), then we want to capture it's value, even if it's null. So the default value does not apply.

Better to add in comments.

sungam3r: Also see this https://github.com/graphql/graphql-spec/pull/793

docs2/site/docs/guides/known-issues.md
```diff
@@ -180,6 +180,11 @@ This is done within your database queries; it is not a function of the dataloade
 `CollectionBatchDataLoader` as you would for a one-to-many relationship; then when you are loading
 data from your database within the fetch delegate, use an inner join to retrieve the proper data.
 
+## Known Issues
```
sungam3r, line 183: This file is already named as `known issues` 😄

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I think I didn't cover this rule, so I'll make a test and fix it:

> If no default value is provided and the input object field’s type is non‐null, an error should be thrown.

docs2/site/docs/guides/known-issues.md
```diff
@@ -180,6 +180,11 @@ This is done within your database queries; it is not a function of the dataloade
...
+
+`IResolveFieldContext.HasArgument` will return `true` for all arguments where `GetArgument` does not return `null`.
+It cannot identify which arguments have been provided a `null` value compared to arguments which were not provided.
```
sungam3r, line 186:
```suggestion
It cannot identify which arguments have been provided a `null` value compared to arguments which were not provided.
This issue should supposedly be resolved in version 4.
```

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: `inputRequired` field in test seems to fit to such a test

src/GraphQL.Tests/Bugs/Bug2159.cs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    public class Bug2159Query : ObjectGraphType
+    {
+        public Bug2159Query()
+        {
+            Field<StringGraphType>(
+                "testValue",
+                resolve: ctx => ctx.GetArgument<string>("arg"),
```
sungam3r, line 120: Could you please swap `arguments` and `resolve` parameters?  It's more common to read.

sungam3r:
I'll finish the review today. I have one file left to check. Besides, you will still have changes.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: It's covered by `AssertValidVariableValue`

Shane32: Comments added.

src/GraphQL.Tests/Bugs/Bug2159.cs
Shane32: Done

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
Shane32: Added a comment and a test

docs2/site/docs/guides/known-issues.md
Shane32: I'm documenting the behavior.  It's too much work for me to fix for 3.3.  I'll plan to fix that for 4.0.

src/GraphQL.Tests/Bugs/Bug1699InvalidEnum.cs
```diff
@@ -93,6 +93,13 @@ private void AssertQueryWithError(string query, string result, string message, i
         [Fact]
         public void Input_Enum_OverrideDefault() => AssertQuerySuccess("{ inputWithDefault(arg: SLEEPY) }", @"{ ""inputWithDefault"": ""Sleepy"" }");
 
+        // For non-nullable struct types, GetArgument coerces null to default(T), which is Grumpy
```
sungam3r, line 96: 👍

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r:
```c#
/// <summary>
/// Return the specified variable's value for the document from the attached <see cref="Inputs"/> object.
/// Since v3.3 returns null for variables set to null rather than the variable's default value.
/// </summary>

Let's document it as is.
```

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: OK

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Done

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: This looks great!  For 4.0 I assume?  I can incorporate these changes.

Shane32: I mean, we COULD put these changes behind internal members so HasArgument works properly for 3.3 ... but I was think I'd just implement it in 4.0

Shane32:
The tests seem to have a temporary failure.  But you can see that it had a successful test run immediately before the most recent change, which was only an xml comment difference.

Shane32:
@sungam3r Anything else on this?

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: Let's keep it for v4.

sungam3r:
Give me 1-2 hours to finish review. Everything looks good, just want to make sure.

src/GraphQL.Tests/Bugs/Bug2159.cs
```diff
@@ -0,0 +1,174 @@
...
+    public class Bug2159Query : ObjectGraphType
+    {
+        public Bug2159Query()
+        {
+            Field<StringGraphType>(
+                "testValue",
+                arguments: new QueryArguments(
+                    new QueryArgument(typeof(StringGraphType)) { Name = "arg", DefaultValue = "defaultValue" }),
+                resolve: ctx => ctx.GetArgument<string>("arg"));
+            Field<StringGraphType>(
```
sungam3r, line 129: It would be more interesting to return not a scalar, but an object.

sungam3r (approved):

Issue #2163, opened 2021-01-11
Fix HasArgument semantics


Shane32:
> However, not storing null values is semantically wrong as well, since supplying a value of `null` should result in `HasArgument` returning `true`.

Agree.

> Ideally, I think that all fields should be stored in context.Arguments, but HasArgument should somehow be able to return false if the value was not supplied to the field (whether a default value was supplied or not) or, when it references a variable, not specified by either the inputs or a variable's default value. This would be a bit more complicated to implement.

`Dictionary<string, object> GetArgumentValues` 
->
`Dictionary<string, ArgumentValue> GetArgumentValues`

```c#
public struct ArgumentValue
{
  object Value { get; }
  ArgumentSource Source { get; } (or ArgumentKind)
}

public enum ArgumentSource (or ArgumentKind)
{
  Argument, (or Literal, Explicit)
  FieldDefault,
  Variable
}
```

` public static bool HasArgument(this IResolveFieldContext context, string name)`
->
` public static bool HasArgument(this IResolveFieldContext context, string name, ArgumentSource source)`

\+
` public static bool HasArgument(this IResolveFieldContext context, string name)` **may** be just an overload for returning any argument kind.

_Originally posted by @sungam3r in https://github.com/graphql-dotnet/graphql-dotnet/pull/2160#discussion_r555076001_

sungam3r:
Done.

Issue #1176, opened 2019-06-10
InputObjectGraphType resolve function ignored


shoe-diamente:
## Description

By defining something like:

```c#
    public class CreateUserInputValue
    {
        public int? OtherName { get; set; }
    }

    public class CreateUserInput : InputObjectGraphType<CreateUserInputValue>
    {
        public CreateUserInput()
        {
            Field<IdGraphType>()
                .Name("id")
                .Resolve(context =>
                {
                    return context.Source.OtherName;
                });
        }
    }
```

it doesn't seem like `id` is correctly mapped to `OtherName`.

### Steps to reproduce

With the code above and by calling:

```graphql
mutation {
  test(input:{id:1244}) {
    id
  }
}
```

with:

```c#
    public class Mutation : ObjectGraphType
    {
        public Mutation()
        {
            Field<NonNullGraphType<User>>()
                .Name("test")
                .Argument<NonNullGraphType<CreateUserInput>>("input", "")
                .ResolveAsync(async context =>
                {
                    var input = context.GetArgument<CreateUserInputValue>("input");
                    // ... breakpoint
                    return null;
                });
        }
    }
```

the `OtherName` value is `null`.

### Expected result

I would expect `OtherName` to be `1244`.

### Actual result

`OtherName` is `null`.

### Environment

```xml
<PackageReference Include="GraphQL" Version="2.4.0" />
```

joemcbride:
This is currently not supported.  This is related to #850 #851 #864 #1045 (and some others I'm sure I'm missing).

gitmustashio:
Can you explain a bit what this feature is ? I can't figure out what he's trying to do from the code example.

shoe-diamente:
@gitmustashio I'm trying to map a GraphQL input field to a specific C# input property which is not necessarily the one with a similar name.

stephenlautier:
Was expecting the same, we always done resolvers, and luckily always worked by luck; Now just had a property with a different name and it wasn't working

flieks:
i tried this workaround for mutations
overriding the base class which contains an enum called DataType
![image](https://user-images.githubusercontent.com/5032389/65499920-b696cf00-debe-11e9-990e-0d3d6f9d20c1.png)

but then you would get an error
`Ambiguous match found.`
casting the input type
` var question = context.GetArgument<QuestionInput>("questionCreate");`

sungam3r:
[...]

PR #2165, opened 2021-01-12
Change IResolveFieldContext.Arguments to ArgumentValue, fix HasArguments, add ParseDictionary

Shane32:
* Adds `ArgumentValue` struct per #2163 and fixes `HasArgument` semantics
  * Changes `IResolveFieldContext.Arguments` dictionary to `Dictionary<string, ArgumentValue>`.  This represents a large number of changes in the tests, and could potentially be a large number of changes for users.  However, anyone using `GetArgument` and `HasArgument` should not be able to tell a difference.
* Redo variable coercion (#1176 part 1)
  * Old code would run AssertValidVariableValue and then coerce the variable value
  * New code does both in one pass
  * Old code would convert variable objects to AST and then execute `CoerceValue`, so `ParseLiteral` and `ParseValue` would often both run
  * New code does not call `CoerceValue` unless parsing the variable default value (a literal), so `ParseLiteral` is only called for literals and `ParseValue` is only called for variables.  Never is a single value converted multiple times.
  * Old code would store an `IDictionary<string, object>` in the `Variables` object, and is coerced to an object when `GetArgument` is called
  * New code will coerce the `IDictionary<string, object>` to the generic type specified within `InputObjectGraphType<T>`, and store the coerced value in the `Variables` object.  This means that when `GetArgument` is called, typically the object is already the correct type and it is returned directly.
  * For `InputObjectGraphType` objects (no `<T>`), the default implementation runs as previously, with a dictionary being stored in the `Variables` object.
  * `ParseDictionary` is a method provided on any `InputObjectGraphType` that can be overridden to provide customized deserialization behavior.  This point in the execution is after the child object fields have been deserialized via `ParseLiteral` or `ParseValue`, so at this point there is no difference between literals and variables.  So there is a single method to override, regardless if the dictionary was generated from a literal or a variable.
  * The new code is not performance optimized, but should perform better due to less iterations and conversions of the same data.  A benchmark with some variable inputs is necessary to confirm.
  * These changes allow, with additional changes, "resolver" functions on actual input fields.

Discuss:
* Namespace for `ArgumentValue`
* Expose a variation of `GetArgument` that returns `ArgumentValue`? (I don't think it's necessary.)
* Add any type of context argument to `ParseDictionary`? (I don't think so.)


src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -592,6 +592,18 @@ namespace GraphQL.DataLoader
 }
 namespace GraphQL.Execution
 {
+    public enum ArgumentSource
+    {
+        FieldDefault = 0,
+        Literal = 1,
+        Variable = 2,
```
sungam3r, line 599: \+ VariableDefault ?

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -592,6 +592,18 @@ namespace GraphQL.DataLoader
...
+    }
+    public struct ArgumentValue
```
sungam3r, line 601: readonly?

Shane32:
## Before

|          Method | Validation |      Mean |    Error |   StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------------- |----------------- |----------:|---------:|---------:|--------:|-------:|------:|----------:|
|         Literal |            False |  43.18 us | 0.381 us | 0.356 us |  5.7983 |      - |     - |  23.78 KB |
| DefaultVariable |            False |  79.56 us | 1.042 us | 0.975 us | 14.5264 |      - |     - |  59.67 KB |
|        Variable |            False |  73.04 us | 0.650 us | 0.576 us | 12.6953 | 0.2441 |     - |   52.2 KB |
|         Literal |             True | 210.04 us | 4.073 us | 4.001 us | 12.2070 |      - |     - |   50.4 KB |
| DefaultVariable |             True | 259.25 us | 2.947 us | 2.757 us | 21.9727 |      - |     - |  90.78 KB |
|        Variable |             True |  65.04 us | 0.751 us | 0.703 us | 12.9395 |      - |     - |  52.97 KB |

## After


|          Method | Validation |      Mean |    Error |   StdDev |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |----------------- |----------:|---------:|---------:|--------:|------:|------:|----------:|
|         Literal |            False |  45.73 us | 0.510 us | 0.452 us |  5.7983 |     - |     - |   23.8 KB |
| DefaultVariable |            False |  44.99 us | 0.500 us | 0.468 us |  5.9814 |     - |     - |   24.9 KB |
|        Variable |            False |  52.21 us | 0.442 us | 0.413 us |  6.1035 |     - |     - |  25.03 KB |
|         Literal |             True | 216.21 us | 1.573 us | 1.394 us | 12.2070 |     - |     - |  50.42 KB |
| DefaultVariable |             True | 228.43 us | 1.204 us | 1.067 us | 13.6719 |     - |     - |  56.01 KB |
|        Variable |             True |  76.88 us | 0.477 us | 0.422 us |  8.0566 |     - |     - |  33.04 KB |

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: Done

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: Done

src/GraphQL/Language/AST/Variable.cs
```diff
@@ -28,5 +28,10 @@ public object Value
         /// Indicates if the variable value has been set.
         /// </summary>
         public bool ValueSpecified { get; private set; }
+
+        /// <summary>
+        /// Indicates if the variable's value is the variable default.
+        /// </summary>
+        public bool IsDefault { get; set; }
```
Shane32, line 35: I could combine these two properties into an `ArgumentSource` property called `Source`.  But what if it has no source?  Add `None` enumeration member?

src/GraphQL/Language/AST/Variable.cs
```diff
@@ -28,5 +28,10 @@ public object Value
         /// Indicates if the variable value has been set.
         /// </summary>
         public bool ValueSpecified { get; private set; }
+
+        /// <summary>
+        /// Indicates if the variable's value is the variable default.
```
sungam3r, line 33: How does this property correlate with `ValueSpecified`? What combinations are valid and what do they mean?

Shane32: Presently, it’s “set” either explicitly or via the default specified in the query. So if IsDefault is false then ValueSpecified is false.

src/GraphQL/Execution/ArgumentValue.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// A variable referenced by the argument.
+        /// </summary>
+        Variable,
+
+        /// <summary>
+        /// A default value for a variable referenced by the argument.
+        /// </summary>
+        VariableDefault,
+    }
+
```
sungam3r, line 53:
```suggestion
```

src/GraphQL/Language/AST/Variable.cs
Shane32: IsDefault is if the variable wasn’t set in the inputs but has a configured default value in the query

Shane32: Although there is information contained within IsDefault, it should be completely invisible to the execution (as it was specified by the requester) which is why I had not previously included it.

src/GraphQL/Execution/ArgumentValue.cs
```diff
@@ -0,0 +1,54 @@
...
+        public ArgumentValue(object value, ArgumentSource source)
+        {
+            Value = value;
+            Source = source;
+        }
+
+        /// <summary>
+        /// Returns the value of the argument.
+        /// </summary>
+        public object Value { get; }
```
sungam3r, line 20: Do I understand correctly that this property cannot contain `null`?

src/GraphQL/Language/AST/Variable.cs
sungam3r: > IsDefault is if the variable wasn’t set in the inputs but has a configured default value in the query

is _true_ ? Also worth to add in comments.

src/GraphQL/Execution/ArgumentValue.cs
Shane32: No. If a null literal is set for a field argument then it will be null.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -63,8 +64,8 @@ public static bool HasArgument(this IResolveFieldContext context, string name)
         {
             var isIntrospection = context.ParentType == null ? context.FieldDefinition.IsIntrospectionField() : context.ParentType.IsIntrospectionType();
             var argumentName = isIntrospection ? name : (context.Schema?.NameConverter.NameForArgument(name, context.ParentType, context.FieldDefinition) ?? name);
-            object value = null;
-            return ((context.Arguments?.TryGetValue(argumentName, out value) ?? false) && value != null);
+            ArgumentValue value = default;
+            return (context.Arguments?.TryGetValue(argumentName, out value) ?? false) && value.Source != ArgumentSource.FieldDefault;
```
sungam3r, line 68: Why only `FieldDefault` and not `VariableDefault` ?

src/GraphQL/Execution/ArgumentValue.cs
Shane32: Also if there is no default for a nullable field and the value is not specified then it will be stored in the dictionary as a null value.

sungam3r: Ditto for variable `{"a": null}` ?

sungam3r: > Also if there is no default for a nullable field and the value is not specified then it will be stored in the dictionary as a null value.

What `Source` it have in this case?

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
Shane32: The value was specified by the request, in the request document.  It is not a default value, as seen by the graph/execution.  How the value was supplied should not matter (literal, variable within inputs, or literal default specified for a variable).  This is why I question the validity of including `VariableDefault` at all.

src/GraphQL/Execution/ArgumentValue.cs
Shane32: Yes.  Every argument defined on a field is listed in the `Arguments` dictionary, regardless of whether the value was specified, default, etc.

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
```diff
@@ -42,14 +42,15 @@ public void argument_converts_long_to_int()
         public void long_to_int_should_throw_for_out_of_range()
         {
             long val = 89429901947254093;
-            _context.Arguments["a"] = val;
+            _context.Arguments["a"] = new ArgumentValue(val, ArgumentSource.Literal);
```
sungam3r, line 45:
```c#
public readonly struct ArgumentValue
{
   public static implicit operator ArgumentValue(string value) => new ArgumentValue(value, ArgumentSource.Literal);
   public static implicit operator ArgumentValue(int value) => new ArgumentValue(value, ArgumentSource.Literal);
...
}
```

At least fewer changes will be required in tests. Although there is some convention in the fact that the value of `Source` is fixed as literal.

src/GraphQL/Language/AST/Variable.cs
Shane32:
```suggestion
        /// Indicates if the variable's value is the variable's configured default value.
```

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
sungam3r: And `_context.Arguments["a"] = val;` just works again.

sungam3r:
```c#
   var d = new Dictionary<string, ArgumentValue>
            {
                ["ddd"] = "adfafaf",
                ["adfaf"] = 5,
                ["afafaf"] = new object[] { true, 7, "f" }
            };
```

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
sungam3r: Understood.

>  This is why I question the validity of including VariableDefault at all.

I would leave this information. It may be useful to us later.

src/GraphQL/Language/AST/Variable.cs
sungam3r: OK

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
Shane32: I feel like this leaves open an easy possibility that someone can set an `ArgumentValue` as a literal without realizing it.  Of course it simplifies the tests, but it hurts the codebase.

src/GraphQL.Tests/Bugs/Bug256NullableEnumTests.cs
```diff
@@ -29,7 +30,7 @@ public void nullable_enum_returns_null()
         {
             var ctx = new ResolveFieldContext
             {
-                Arguments = new Dictionary<string, object> { { "value", null } }
+                Arguments = new Dictionary<string, ArgumentValue> { { "value", new ArgumentValue(null, ArgumentSource.Literal) } }
```
sungam3r, line 33:
```c#
public readonly struct ArgumentValue
{
    public static readonly ArgumentValue NullLiteral = new ArgumentValue(null, ArgumentSource.Literal);
    public static readonly ArgumentValue NullVariable = new ArgumentValue(null, ArgumentSource.Variable);
}
```

`Arguments = new Dictionary<string, ArgumentValue> { { "value", new ArgumentValue(null, ArgumentSource.Literal) } }`
->
`   Arguments = new Dictionary<string, ArgumentValue> { ["value"] = ArgumentValue.NullLiteral }`

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
sungam3r: > I feel like this leaves open an easy possibility that someone can set an ArgumentValue as a literal without realizing it.

This can be said for any implicit cast operator. Nevertheless, they are used quite widely. If there is no loss of accuracy, then why not? In this case, there is no loss of precision, we just chose the default behavior. Moreover, consumers are very unlikely to **create** this structs. They will **consume** them already populated by us.

src/GraphQL.Tests/Bugs/Bug256NullableEnumTests.cs
Shane32: Looks fine, but I'm not going to go change all the tests because the code looks prettier.  It took me an hour to fix all the tests the first time!  I'll add the variables into the struct if you like.

Shane32: I'll do a search-and-replace and catch whatever that turns up.

sungam3r: Well, if there is a way to make the code no worse than it was, then why not?. At least for simple int/string values.

src/GraphQL/Execution/ArgumentValue.cs
sungam3r: Then I only had a question with `Source` property on args with unspecified values.

src/GraphQL/ObjectExtensions.cs
```diff
@@ -306,5 +306,44 @@ public static bool IsDefinedEnumValue(Type type, object value)
...
             return false;
         }
+
+        /*
+        public static IDictionary<string, object> ToObjectDictionary(this IDictionary<string, ArgumentValue> source)
+        {
+            return new ObjectDictionary(source);
+        }
+
+        private class ObjectDictionary : IDictionary<string, object>
```
sungam3r, line 316: What is it?

src/GraphQL.Tests/Bugs/Bug256NullableEnumTests.cs
Shane32: I changed them to readonly properties so they do not cause a CPU cache miss.  It should be equally as fast otherwise.

src/GraphQL/ObjectExtensions.cs
Shane32: Backwards compatibility - if we want to provide a method to retrieve an `IDictionary<string, object>` that contains the argument values (like 3.x did) rather than `ArgumentValue` values.  If there is no reason to include this code, I'll remove it entirely.  What do you think?

src/GraphQL.Tests/Bugs/Bug256NullableEnumTests.cs
sungam3r: Of course. I wrote fields out of habit.

src/GraphQL/ObjectExtensions.cs
sungam3r: `IResolveFieldContext.Arguments` ?

sungam3r: `IResolveFieldContext.Arguments` type should be changed to `IDictionary<string, ArgumentValue>`

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
Shane32: You cannot write an implicit operator for `object`. There are an infinite number of types, so now we need to select which types we are writing implicit operators for.  This does not help the codebase and only helps tests.  Moreover, all the required code changes for the tests have already been completed.  If it had not been, that's one thing.  But they are done.  To be of any value, they would all need to change back.  And finally, it is better for tests to explicitly define the behavior they are testing.

If we want a helper to set argument values, let's write it within the test project.  We could write an extension method that converts `IDictionary<string, object>` to `Dictionary<string, ArgumentValue>`.  Then code might look like this:

```cs
   var d = new Dictionary<string, object>
            {
                ["ddd"] = "adfafaf",
                ["adfaf"] = 5,
                ["afafaf"] = new object[] { true, 7, "f" }
            }.ToArgumentValues();
```

Either way, if you want to change it, go ahead.  I don't see any reason to spend time on it.

src/GraphQL/ObjectExtensions.cs
Shane32: So then I'll delete this code.

src/GraphQL.Tests/Execution/ResolveFieldContextTests.cs
sungam3r: ok

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -1,10 +1,16 @@
...
 {
     /// <summary>
     /// Represents an input object graph type.
     /// </summary>
     public interface IInputObjectGraphType : IComplexGraphType
     {
+        /// <summary>
+        /// Converts a supplied dictionary of keys and values to an object.
+        /// </summary>
+        object ParseDictionary(IDictionary<string, object> value);
```
sungam3r, line 13: ConvertDictionary ?

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -592,6 +592,21 @@ namespace GraphQL.DataLoader
...
+        Variable = 2,
+        VariableDefault = 3,
+    }
+    public readonly struct ArgumentValue
+    {
+        public ArgumentValue(object value, GraphQL.Execution.ArgumentSource source) { }
+        public GraphQL.Execution.ArgumentSource Source { get; }
+        public object Value { get; }
+        public static GraphQL.Execution.ArgumentValue NullLiteral { get; }
+        public static GraphQL.Execution.ArgumentValue NullVariable { get; }
```
Shane32, line 608: Keep in mind that these properties are not used within GraphQL, and `default` can be used for any tests that need a `null` value and don't care about `Source`.  I probably didn't think of that when I changed the tests previously.  I'd remove these properties and change `ArgumentValue.NullLiteral` to `default`.  But I'm ok if you want to leave them.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -308,37 +363,33 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
...
+                        .ToList(), ArgumentSource.Literal);
                 }
                 else
                 {
-                    return new[] { CoerceValue(schema, listItemType, input, variables) };
+                    return new ArgumentValue(new[] { CoerceValue(schema, listItemType, input, variables).Value }, ArgumentSource.Literal);
                 }
             }
 
-            if (type is IObjectGraphType || type is IInputObjectGraphType)
```
sungam3r, line 321: What about `type is IObjectGraphType` condition?

src/GraphQL/Types/Composite/InputObjectGraphType.cs
Shane32: I chose `ParseDictionary` because it "matches" `ParseLiteral` and `ParseValue`.  It's only called while parsing arguments and variables.  Do you think `ConvertDictionary` is better?

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: `IObjectGraphType` is invalid for input types so it is not applicable.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: > and don't care about Source

This can be a source of errors. I want to leave them because such properties may be useful.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
sungam3r: OK, ParseDictionary

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: WOW 🤦 

For today I'm done, I have looked at almost all the files, tomorrow I will finish.

sungam3r:
изображение

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Thanks!! 👍

Shane32:
OHHH I misunderstood the error.  I can fix that

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
+                        // is a non-null graph type, the NonNullGraphType.ParseValue method will throw an error
+                        newDictionary[field.Name] = ParseValue(field.ResolvedType, childFieldVariableName, fieldValue);
+                    }
+                    else if (field.DefaultValue != null)
+                    {
+                        // RULE: If no value is provided for a defined input object field and that
+                        // field definition provides a default value, the default value should be used.
+                        newDictionary[field.Name] = field.DefaultValue;
+                    }
+                    else if (field.ResolvedType is NonNullGraphType nonNullGraphType)
```
Shane32, line 273:
```suggestion
                    else if (field.ResolvedType is NonNullGraphType)
```

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -15,6 +21,19 @@ public class InputObjectGraphType : InputObjectGraphType<object>
     /// <inheritdoc cref="IInputObjectGraphType"/>
     public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType>, IInputObjectGraphType
     {
+        /// <inheritdoc/>
+        public virtual object ParseDictionary(IDictionary<string, object> value)
```
sungam3r, line 25: Are you positioning that method as alternative of `Resolve` for inputs ?

sungam3r: If so I would like to mention this in comments to `IInputObjectGraphType.ParseDictionary`. And if I understand correctly  adding a special _parse_ (_resolve-like_) method for inputs serves as another argument in favor of separating inheritance hierarchies for inputs and outputs, where the `Resolver` property will remain only on outputs.

Shane32: > Are you positioning that method as an alternative of `Resolve` for inputs ?

For now, at least.  In a later revision maybe we can add/implement a `Resolve` method for each field.  This perhaps, would be called by the default implementation of `ParseDictionary` before calling `ToObject`.  Or perhaps more likely it would be executed by the `CoerceValue / GetVariable` prior to calling `ParseDictionary`.  I'm not sure if it's necessary since each field will be parsed already according to the scalar or object's `ParseLiteral` or `ParseValue` (or now, `ParseDictionary`) method.

> If so I would like to mention this in comments to IInputObjectGraphType.ParseDictionary.

Sounds good; I'll add additional comments, or feel free to make suggestions.

> And if I understand correctly adding a special parse (resolve-like) method for inputs serves as another argument in favor of separating inheritance hierarchies for inputs and outputs, where the Resolver property will remain only on outputs.

Completely agree.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -15,6 +21,19 @@ public class InputObjectGraphType : InputObjectGraphType<object>
     /// <inheritdoc cref="IInputObjectGraphType"/>
     public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType>, IInputObjectGraphType
     {
+        /// <inheritdoc/>
```
Shane32, line 24:
```suggestion
        /// <summary>
        /// Converts a supplied dictionary of keys and values to an object.
        /// The default implementation uses <see cref="ObjectExtensions.ToObject"/> to convert the field values
        /// of the input object into an object of type <typeparamref name="TSourceType"/>.
        /// Overriding this method allows for customizing the deserialization process of input objects,
        /// much like a field resolver for output objects.
        /// </summary>
```
Uhh...how about this.???

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -1,10 +1,16 @@
...
+
 namespace GraphQL.Types
 {
     /// <summary>
     /// Represents an input object graph type.
     /// </summary>
     public interface IInputObjectGraphType : IComplexGraphType
     {
+        /// <summary>
+        /// Converts a supplied dictionary of keys and values to an object.
```
Shane32, line 11:
```suggestion
        /// Converts a supplied dictionary of keys and values to an object.
        /// Overriding this method allows for customizing the deserialization process of input objects,
        /// much like a field resolver does for output objects.
```

sungam3r: edited

src/GraphQL/Types/Composite/InputObjectGraphType.cs
Shane32: 👍

src/GraphQL/Types/Composite/InputObjectGraphType.cs
Shane32: 👍

Shane32:
If you have no further comments, I'll merge this

sungam3r:
A bit busy with other work. I have to review the last file. It will take no more than 1 day.

sungam3r:
> Namespace for ArgumentValue

`GraphQL` or `GraphQL.Execution` ? I'm OK with `GraphQL.Execution`.

sungam3r:
> Add any type of context argument to ParseDictionary?

I didn’t see anything that could be used there.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
         {
             var variables = new Variables();
 
             if (variableDefinitions != null)
             {
-                foreach (var v in variableDefinitions)
+                foreach (var variableDef in variableDefinitions)
                 {
+                    // find the IGraphType instance for the variable type
+                    var graphType = variableDef.Type.GraphTypeFromType(schema);
```
sungam3r, line 86: Side note - #2102

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
+                        variable.IsDefault = true;
                     }
+                    else if (graphType is NonNullGraphType)
+                    {
+                        ThrowNullError(variable.Name);
+                    } 
+
+                    // if the variable was not specified and no default was specified, do not add the variable to the Variables object
 
+                    // add the variable to the list of parsed variables defined for the operation
```
sungam3r, line 122: I don't understand how this comment correlates to the line above.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
         /// Return the specified variable's value for the document from the attached <see cref="Inputs"/> object.
+        /// <br/><br/>
+        /// Validates and parses the supplied input object according to the variable's type, and converts the object
+        /// with <see cref="ScalarGraphType.ParseValue(object)"/> and
+        /// <see cref="IInputObjectGraphType.ParseDictionary(IDictionary{string, object})"/> as applicable.
+        /// <br/><br/>
         /// Since v3.3, returns null for variables set to null rather than the variable's default value.
         /// </summary>
-        public static object GetVariableValue(Document document, ISchema schema, VariableDefinition variable, object input)
+        public static object GetVariableValue(Document document, IGraphType graphType, VariableDefinition variable, object input)
```
sungam3r, line 143: Do we need this method in public API?

sungam3r:
> Old code would run AssertValidVariableValue and then coerce the variable value
New code does both in one pass

Good. This code was causing me questions.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
-            if (type is ListGraphType listType)
+            static object ParseValueList(ListGraphType listGraphType, string variableName, object value)
             {
-                var listItemType = listType.ResolvedType;
+                if (value == null)
+                    return null;
 
-                if (input is IEnumerable list && !(input is string))
+                // note: a list can have a single child element which will automatically be intrepreted as a list of 1 elements. (see below rule)
+                // so, to prevent a string as being intrepreted as a list of chars (which get converted to strings), we ignore considering a string as an IEnumerable
```
sungam3r, line 212:
```suggestion
                // so, to prevent a string as being interpreted as a list of chars (which get converted to strings), we ignore considering a string as an IEnumerable
```

Thanks for meaningful comments.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
-                    var index = -1;
-                    foreach (var item in list)
-                        AssertValidVariableValue(schema, listItemType, item, $"{variableName}[{++index}]", hasDefaultValue);
+                    // create a list containing the parsed elements in the input list
+                    var valueOutputs = new List<object>(values is ICollection collection ? collection.Count : 0);
+                    int index = 0;
+                    foreach (var val in values)
+                    {
+                        // parse/validate values as required by graph type
+                        valueOutputs.Add(ParseValue(listGraphType.ResolvedType, variableName + "[" + index++ + "]", val));
```
sungam3r, line 221: Wow. What will happen on arrays of hundreds of elements? It seems to me that `variableName` is used only in error messages so let's pass additional `int?` argument in `ParseValue` like we did it for `BuildExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, int? indexInParentNode = null)`

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
+                // so, to prevent a string as being intrepreted as a list of chars (which get converted to strings), we ignore considering a string as an IEnumerable
+                if (value is IEnumerable values && !(value is string))
                 {
-                    var index = -1;
-                    foreach (var item in list)
-                        AssertValidVariableValue(schema, listItemType, item, $"{variableName}[{++index}]", hasDefaultValue);
+                    // create a list containing the parsed elements in the input list
+                    var valueOutputs = new List<object>(values is ICollection collection ? collection.Count : 0);
+                    int index = 0;
+                    foreach (var val in values)
```
sungam3r, line 218: `values` can be checked for `IList` and iterated via indexer so we save a few allocations for `List<>+Iterator` etc.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
+                    return valueOutputs;
                 }
                 else
                 {
-                    AssertValidVariableValue(schema, listItemType, input, variableName, hasDefaultValue);
+                    // RULE: If the value passed as an input to a list type is not a list and not the null value,
+                    // then the result of input coercion is a list of size one, where the single item value is the
+                    // result of input coercion for the list’s item type on the provided value (note this may apply
+                    // recursively for nested lists).
+                    var result = ParseValue(listGraphType.ResolvedType, variableName, value);
```
sungam3r, line 231: Does this variable name need index 0?

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
 
-            if (type is IObjectGraphType || type is IInputObjectGraphType)
+            static object ParseValueObject(IInputObjectGraphType graphType, string variableName, object value)
             {
-                var complexType = (IComplexGraphType)type;
+                if (value == null)
+                    return null;
 
-                if (!(input is Dictionary<string, object> dict))
+                if (!(value is IDictionary<string, object> dic))
```
sungam3r, line 241: Why not dict ? 😄

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
+                }
 
-                if (type is IInputObjectGraphType)
+                var newDictionary = new Dictionary<string, object>(dic.Count);
+                foreach (var field in graphType.Fields)
                 {
-                    unknownFields = dict.Keys
-                        .Except(complexType.Fields.Select(f => f.Name))
-                        .ToList();
+                    var childFieldVariableName = variableName + "." + field.Name;
```
sungam3r, line 253: The same concern here about string concatenation.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
+                    {
+                        // RULE: If no default value is provided and the input object field’s type is non‐null,
+                        // an error should be thrown.
+                        ThrowNullError(childFieldVariableName);
+                    }
+
+                    // RULE: Otherwise, if the field is not required, then no
+                    // entry is added to the coerced unordered map.
+
+                    // so do not do this:    else { newDictionary[field.Name] = null; }
```
sungam3r, line 283: 👍

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -72,195 +72,248 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
...
                 }
 
-                if (unknownFields?.Count > 0)
+                // RULE: The value for an input object should be an input object literal
+                // or an unordered map supplied by a variable, otherwise a query error
+                // must be thrown. In either case, the input object literal or unordered
+                // map must not contain any entries with names not defined by a field
+                // of this input object type, ***otherwise an error must be thrown.***
+                var unknownFields = dic.Keys
```
sungam3r, line 291: link

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -274,32 +327,34 @@ public static void AssertValidVariableValue(ISchema schema, IGraphType type, obj
...
+        /// Coerces a literal value to a compatible .NET type for the variable's graph type.
+        /// Typically this is a value for a field argument or default value for a variable.
         /// </summary>
-        public static object CoerceValue(ISchema schema, IGraphType type, IValue input, Variables variables = null, object fieldDefault = null)
+        public static ArgumentValue CoerceValue(ISchema schema, IGraphType type, IValue input, Variables variables = null, object fieldDefault = null)
         {
             if (type is NonNullGraphType nonNull)
             {
-                // validation rules and/or AssertValidVariableValue have verified that this is not null
+                // validation rules have verified that this is not null
```
sungam3r, line 337: But they still can be disabled.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -274,32 +327,34 @@ public static void AssertValidVariableValue(ISchema schema, IGraphType type, obj
...
             if (input == null)
             {
-                return fieldDefault;
+                return new ArgumentValue(fieldDefault, ArgumentSource.FieldDefault);
             }
 
             if (input is NullValue)
             {
-                return null;
+                return new ArgumentValue(null, ArgumentSource.Literal);
```
sungam3r, line 348:
```suggestion
                return ArgumentValue.NullLiteral;
```

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -308,37 +363,33 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
 
                 if (input is ListValue list)
                 {
-                    return list.Values
-                        .Select(item => CoerceValue(schema, listItemType, item, variables))
-                        .ToList();
+                    return new ArgumentValue(list.Values
```
sungam3r, line 366: link2

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -308,37 +363,33 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
...
                 }
             }
 
-            if (type is IObjectGraphType || type is IInputObjectGraphType)
+            if (type is IInputObjectGraphType inputObjectGraphType)
             {
                 if (!(input is ObjectValue objectValue))
                 {
-                    return null;
+                    return new ArgumentValue(null, ArgumentSource.Literal);
```
sungam3r, line 380:
```suggestion
                    return ArgumentValue.NullLiteral;
```

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -308,37 +363,33 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
...
                 }
 
-                var complexType = (IComplexGraphType)type; // both IObjectGraphType and IInputObjectGraphType inherit from IComplexGraphType
                 var obj = new Dictionary<string, object>();
 
-                foreach (var field in complexType.Fields)
+                foreach (var field in inputObjectGraphType.Fields)
```
sungam3r, line 385: link3

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Good idea. I’ll put this in a subsequent PR of optimizations.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Good idea. I’ll put this in a subsequent PR of optimizations.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I would not. It is used for the error message and would be a more accurate error message without it.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Lol - I rewrote the code from scratch and it must have just turned out that way.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: This will be more tricky to implement without allocations. Either way it will go in a separate PR.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -349,27 +400,30 @@ public static object CoerceValue(ISchema schema, IGraphType type, IValue input,
...
             }
 
             if (type is ScalarGraphType scalarType)
             {
-                return scalarType.ParseLiteral(input) ?? throw new ArgumentException($"Unable to convert '{input}' to '{type.Name}'");
+                return new ArgumentValue(scalarType.ParseLiteral(input) ?? throw new ArgumentException($"Unable to convert '{input}' to '{type.Name}'"), ArgumentSource.Literal);
             }
 
-            return null;
+            throw new ArgumentOutOfRangeException(nameof(input), $"Unknown type of input object '{type.GetType()}'");
```
sungam3r, line 426: Сould the `type` be `null` here?

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Yeah I noticed how this worked. I figured that if you disabled the rule then presumably you didn’t want the check.

Shane32:
Please note that I’m specifically not making performance related changes in this PR. There’s enough changes here and having an outstanding PR this large prevents further work on the codebase. 

I’ll need to check the rest of your comments once I’m at a pc. I’m not sure I understand what you mean by “link” yet.

Shane32:
There’s some use of LINQ in there also that probably can be optimized away. I just used the old code for now so it is easily readable what was changed in this PR.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: OK

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: This consideration in the same form should be written in the comments so as not to mislead the reader.

sungam3r:
> I’m not sure I understand what you mean by “link” yet.

Just a http link to mention in another issue. See #2182.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32, line 122:
```suggestion
                    // if the variable was not specified and no default was specified, do not set variable.Value

                    // add the variable to the list of parsed variables defined for the operation
```

Shane32: The other comment has been fixed.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32:
```suggestion
                // validation rules have verified that this is not null; if the validation rule was not executed, it
                // is assumed that the caller does not wish this check to be executed
```

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Probably not.  Do you think it should be made private?

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I added a null check at the top of the method.  It certainly shouldn't be null.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Commit if this looks okay to you.

Shane32:
Todo in subsequent PRs:
- Remove schema from CoerceValue (completely unused)
- Eliminate string concatenation
- Make GetVariableValue private?

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: Yes. I tend to reduce the public API if we do not see an obvious need for it. This will make life easier for us.

src/GraphQL/Language/AST/Variable.cs
sungam3r: not worth it

sungam3r (approved):

PR #2166, opened 2021-01-12
Memory allocation optimizations

sungam3r:
1. Eliminate enumerators boxing (List<>, Dictionary<,>).
2. Avoid LINQ.
3. Avoid closures allocations.
4. Introduce array pools and eliminate `new`-ing.
5. **Avoid ResolveFieldContext allocations** - one of the main pain points
5. Moderate changes in public APIs.

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.16299.726 (1709/FallCreatorsUpdate/Redstone3)
Intel Core i7 CPU 920 2.67GHz (Nehalem), 1 CPU, 8 logical and 4 physical cores
Frequency=2607427 Hz, Resolution=383.5198 ns, Timer=TSC
.NET Core SDK=5.0.101
  [Host]     : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT
  DefaultJob : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT


```

`develop`:

|        Method | UseCaching |        Mean |    Error |   StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |----------- |------------:|---------:|---------:|---------:|--------:|------:|----------:|
| **Introspection** |      **False** | **2,631.82 μs** | **8.290 μs** | **7.349 μs** | **160.1563** | **46.8750** |     **-** | **713.46 KB** |
|          Hero |      False |    42.94 μs | 0.066 μs | 0.062 μs |   2.3193 |       - |     - |   9.59 KB |
| **Introspection** |       **True** | **2,073.45 μs** | **7.159 μs** | **6.346 μs** | **156.2500** | **50.7813** |     **-** |  **673.5 KB** |
|          Hero |       True |    10.48 μs | 0.010 μs | 0.010 μs |   1.1902 |       - |     - |   4.92 KB |

`develop+tune`:  83ce6a7

| **Introspection** |      **False** | **2,191.837 μs** | **2.6996 μs** | **2.3931 μs** | **125.0000** | **35.1563** |     **-** | **544.08 KB** |
|          Hero |      False |    40.949 μs | 0.0254 μs | 0.0238 μs |   2.1362 |       - |     - |   8.86 KB |
| **Introspection** |       **True** | **1,652.144 μs** | **2.3289 μs** | **2.1784 μs** | **117.1875** | **39.0625** |     **-** | **504.38 KB** |
|          Hero |       True |     9.290 μs | 0.0086 μs | 0.0076 μs |   1.0223 |       - |     - |   4.19 KB |

`develop+tune`: 057cd2b

| **Introspection** |      **False** | **2,213.526 μs** | **3.3816 μs** | **2.9977 μs** | **121.0938** | **19.5313** |     **-** | **533.57 KB** |
|          Hero |      False |    40.952 μs | 0.0310 μs | 0.0275 μs |   2.1362 |       - |     - |   8.86 KB |
| **Introspection** |       **True** | **1,715.587 μs** | **3.4338 μs** | **3.2120 μs** | **109.3750** | **37.1094** |     **-** | **493.86 KB** |
|          Hero |       True |     8.925 μs | 0.0198 μs | 0.0165 μs |   1.0223 |       - |     - |   4.19 KB |

`develop+tune`: 14b6092

| **Introspection** |      **False** | **2,102.830 μs** | **5.8698 μs** | **4.9016 μs** | **109.3750** | **27.3438** |     **-** | **498.22 KB** |
|          Hero |      False |    40.719 μs | 0.0286 μs | 0.0268 μs |   2.1362 |       - |     - |   8.81 KB |
| **Introspection** |       **True** | **1,519.596 μs** | **1.9548 μs** | **1.8285 μs** | **103.5156** | **37.1094** |     **-** | **458.52 KB** |
|          Hero |       True |     8.917 μs | 0.0200 μs | 0.0177 μs |   1.0071 |       - |     - |   4.14 KB |

14b6092 was my last commit in this PR. The PR turned out to be large and contains several areas at once in which improvements have occurred. I did not break it down into several smaller ones, because the changes are interconnected and this would only complicate the work. I have tested every significant commit to ensure consistent performance improvements. Above, I have published not all of my results, but only for the main stages. The difference compared to the beginning is immediately visible - both execution time and memory allocations.


sungam3r:
@Shane32 For now I demonstrated the work with pooled arrays only in `__Type.cs`.

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -217,6 +217,15 @@ public interface IParentExecutionNode
...
         IEnumerable<ExecutionNode> GetChildNodes();
+
+        /// <summary>
+        /// Applies the specified delegate to child execution nodes.
+        /// </summary>
+        /// <typeparam name="TState">Type of the provided state.</typeparam>
+        /// <param name="action">Delegate to execute on every child node of this node.</param>
+        /// <param name="state">An arbitrary state passed by the caller.</param>
+        /// <param name="reverse">Specifies the direct or reverse direction of child nodes traversal.</param>
+        void ApplyToChildren<TState>(Action<ExecutionNode, TState> action, TState state, bool reverse = false);
```
sungam3r, line 228: The same trick here as for nodes traversal at validation stage. `GetChildNodes()` are not used now.

src/GraphQL/Execution/ExecutionNode.cs
```diff
@@ -227,7 +236,7 @@ public class ObjectExecutionNode : ExecutionNode, IParentExecutionNode
         /// <summary>
         /// Returns a dictionary of child execution nodes, with keys set to the names of the child fields that the child nodes represent.
         /// </summary>
-        public IDictionary<string, ExecutionNode> SubFields { get; set; }
+        public Dictionary<string, ExecutionNode> SubFields { get; set; }
```
sungam3r, line 239: There was **A LOT** of allocations here due to repeatedly traversed dictionaries and `Dictionary<>+Iterator` boxing.

src/GraphQL/Introspection/__Type.cs
```diff
@@ -47,36 +48,73 @@ public __Type()
...
-
-                        fields ??= Enumerable.Empty<FieldType>();
-                        fields = await fields.WhereAsync(f => context.Schema.Filter.AllowField(type, f)).ConfigureAwait(false);
+                        var type = (IComplexGraphType)context.Source;
+                        var fields = context.GetPooledArray<FieldType>(type.Fields.Count);
+
+                        bool includeDeprecated = context.GetArgument<bool>("includeDeprecated");
+
+                        int index = 0;
+                        foreach (var field in type.Fields.List)
```
sungam3r, line 57: **A LOT** of LINQ stuff here before : chained enumerators, closures...

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
```diff
@@ -16,6 +16,15 @@ namespace GraphQL
...
     public interface IResolveFieldContext : IProvideUserContext
     {
+        /// <summary>
+        /// Gets an array of the specified minimum length from the pool.
+        /// This array does not need to be returned to the pool yourself.
+        /// </summary>
+        /// <typeparam name="TElement">Array element type.</typeparam>
+        /// <param name="minLength">The minimum length of the array.</param>
+        /// <returns>Array from pool.</returns>
+        TElement[] GetPooledArray<TElement>(int minLength);
```
sungam3r, line 26: I have not found a better place for this method.

src/GraphQL/TaskExtensions.cs
```diff
@@ -42,14 +42,14 @@ internal static object GetResult(this Task task)
             }
         }
 
-        internal static async Task<IEnumerable<T>> WhereAsync<T>(this IEnumerable<T> items, Func<T, Task<bool>> predicate)
+        internal static async Task<IEnumerable<T>> WhereAsync<T, TState>(this IEnumerable<T> items, Func<T, TState, Task<bool>> predicate, TState state)
```
sungam3r, line 45: I will gradually remove this method altogether, but for now I have added a parameter to it to transfer the state so that there are fewer closures.

src/GraphQL/Types/GraphTypesLookup.cs
```diff
@@ -506,46 +506,42 @@ private void ApplyTypeReference(IGraphType type)
 
             if (type is IObjectGraphType objectType)
             {
-                objectType.ResolvedInterfaces = objectType
```
sungam3r, line 509: Now there is no need to create a new list and replace the old one with it (with LINQ stuff).

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -1767,6 +1779,14 @@ namespace GraphQL.Types
         public GraphQL.Resolvers.IAsyncEventStreamResolver AsyncSubscriber { get; set; }
         public GraphQL.Resolvers.IEventStreamResolver Subscriber { get; set; }
     }
+    public class FieldFields : System.Collections.Generic.IEnumerable<GraphQL.Types.FieldType>, System.Collections.IEnumerable
```
Shane32, line 1782: This sounds funny.  How about `FieldList`?

src/GraphQL/Introspection/__Type.cs
Shane32: Changing the introspection queries' resolvers will skew performance metrics because now the introspection query is going to show a performance improvement for _introspection_ queries, regardless of whether an equally complex query's performance has improved or not.  An introspection query executed in a live environment is relatively rare.  So now any benchmarks have less meaning.  No conclusion, just FYI.

src/GraphQL/MemoryExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// Provides extension methods for working with arrays and pools.
+    /// </summary>
+    public static class MemoryExtensions
+    {
+        private static readonly ConcurrentDictionary<Type, Action<Array>> _delegates = new ConcurrentDictionary<Type, Action<Array>>();
+        private static readonly Func<Type, Action<Array>> _factory = CreateDelegate;
+
+        internal static void Return(this Array array) => _delegates.GetOrAdd(array.GetType(), _factory)(array);
```
Shane32, line 16:
```suggestion
        internal static void Return(this Array array) => _delegates.GetOrAdd(array.GetType(), CreateDelegate)(array);
```

Why create `_factory`? Maybe rename `CreateDelegate` to `CreateDelegateFactory` if you want.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
Shane32: I wish we could keep any memory pool usage `internal`.

Shane32: At minimum, I think we should document the lifetime explicitly.  e.g. "This array will be returned to the pool once the execution completes".  It is important that somebody does not try to use memory from this pool and store it in, for example, `Extensions`.

src/GraphQL/Types/Collections/FieldFields.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace GraphQL.Types
+{
+    /// <summary>
+    /// A class that represents a set of fields for <see cref="IComplexGraphType"/> i.e <see cref="ComplexGraphType{TSourceType}"/>.
+    /// </summary>
+    public class FieldFields : IEnumerable<FieldType>
```
Shane32, line 10:
```suggestion
    public class FieldList : IEnumerable<FieldType>
```
I think `FieldFields` is confusing.

Shane32:
I took a look though it.  Seems fine.  I really don't think we should expose the shared ArrayPool stuff outside the project.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: At first I named it just `Fields` but we already have this name `GraphQL.Language.AST.Fields`. If I call it `FieldList`, then it will be logical to give the `List` suffix to all such classes. I don't know.

src/GraphQL/Introspection/__Type.cs
sungam3r: I know. I wanted to add "real" tests, but I haven't had time yet. If you want, you can do it yourself, then I will re-experiment. In any case, sooner or later we will need them, we have a very small set and we do not cover the whole variety of types of requests with current benchmarks.

src/GraphQL/MemoryExtensions.cs
sungam3r: https://github.com/dotnet/roslyn/issues/39869

I also wrote this comment for BasicVisitor:
изображение

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
sungam3r: > I wish we could keep any memory pool usage internal.

This method is deliberately placed in the public API, because it makes sense for consumers to use it for theis resolvers, and not just inside GraphQL.NET internal bits. I added more comments.

src/GraphQL/Types/Collections/FieldFields.cs
sungam3r: Changed to `TypeFields`.

src/GraphQL/Introspection/__Type.cs
sungam3r: Introspection resolvers were a good example to test.

sungam3r:
Now there are only two methods for work with arrays: `IResolveFieldContext.GetPooledArray` + `MemoryExtensions.Constrained`.  They work even not with ROM/Span, but with "ordinary" arrays. If people want to use an ArrayPool then they will have to independently solve exactly the same problems of managing the lifetime of an array and limiting its size. They will not be able to simply return an array from the resolver, since the size of the allocated array is always larger than necessary. The added pair of methods does not greatly complicate the public API and at the same time provides ready-made tools for working with pools for simple scenarios when you need to return just an array of data.

src/GraphQL/ConstrainedArray.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public bool IsReadOnly => true;
+
+        public bool IsFixedSize => true;
+
+        public int IndexOf(T item) => throw new NotSupportedException();
+
+        public int IndexOf(object value) => throw new NotSupportedException();
+
+        public bool Contains(T item) => throw new NotSupportedException();
```
sungam3r, line 40: This and other methods that do not change the state of the array can be implemented. Until I just didn’t do it, because it’s not necessary.

Shane32:
> Now there are only two methods for work with arrays: `IResolveFieldContext.GetPooledArray` + `MemoryExtensions.Constrained`. They work even not with ROM/Span, but with "ordinary" arrays. If people want to use an ArrayPool then they will have to independently solve exactly the same problems of managing the lifetime of an array and limiting its size. They will not be able to simply return an array from the resolver, since the size of the allocated array is always larger than necessary. The added pair of methods does not greatly complicate the public API and at the same time provides ready-made tools for working with pools for simple scenarios when you need to return just an array of data.

OK

Shane32 (commented):
This PR seems fine to me.  I'll review it again after you've removed the WIP tag and it's ready to merge.

src/GraphQL/Introspection/__Type.cs
Shane32: Right.  It would just be nice to know what the impact is.  You can't tell now.

Shane32: Well, we can see some difference in the hero test. 👍

sungam3r:
I will still make changes to all types of introspection and remove some LINQ.

sungam3r:
In my opinion GH CI is experiencing some problems.

sungam3r:
@Shane32 Could you please run tests on your machine. After upgrading VS to 16.8.4 I can not run tests locally anymore. I think I make some infinite loop in one of the tests, that's why the tests take so long and don't complete.

sungam3r:
OK. I managed to reproduce infinite loop on other machine.

src/GraphQL.Tests/Types/GraphTypesLookupTests.cs
```diff
@@ -1,38 +0,0 @@
...
-    public class GraphTypesLookupTests
-    {
-        private class TestType : GraphType { }
-
-        private readonly ManualResetEvent _inIteration = new ManualResetEvent(false);
-        private readonly ManualResetEvent _lookupModified = new ManualResetEvent(false);
-        private readonly GraphTypesLookup _lookup = new GraphTypesLookup();
-
-        [Fact]
-        public void is_thread_safe()
```
sungam3r, line 17: This test doesn't make sense anymore because Lookup doesn't allow you to change its content after the schema is created.

Shane32:
> @Shane32 Could you please run tests on your machine. After upgrading VS to 16.8.4 I can not run tests locally anymore. I think I did some infinite loop in one of the tests, that's why the tests take so long and don't complete.

The .NET 4.8 tests have not been working for me recently.  I have to terminate the `testhost.net48.exe` process to get tests to run.  Note that this occurs before I start any tests.  Once I terminate that process, then I can run .Net Core 2.1/3.1 tests just fine.

sungam3r:
Side note - `SchemaTypes` (`GraphTypesLookup` earlier) will be changed more, but in a different PR. Now that this is internal, it opened the way for a simplification of the code.

Shane32:
See PR #2181 to fix tests (on my machine...)

Shane32:
@sungam3r I think you are making too many changes in a single PR. For instance, you recently added a commit that reworked how resolve field contexts work. It was so large and invasive that it should certainly be its own PR. And there’s no way to test/prove individual improvements with so many changes in a single PR.

Shane32:
Those changes should be in a separate PR

sungam3r:
> Those changes should be in a separate PR

@Shane32 I edited my initial comment with PR summary. I finished here.

> And there’s no way to test/prove individual improvements with so many changes in a single PR.

Let's consider the final results. They've gotten better for each case. Also, I actually tested every step.

Shane32:
Changes to ResolveFieldContext must go in another PR. I dislike the changes and want to explore other alternatives.

Shane32:
Moreover I had planned to try them but was waiting for you to finish this PR. Again, you must move the resolve field context changes to another PR

sungam3r:
> Changes to ResolveFieldContext

What exactly? One way or another there is a lot tied to it. I am ready to discuss the applied approaches and naming issues. To be honest, I didn't know how to better name the new interfaces.

Shane32:
If you create a branch, revert those changes, wrap up the original changes here, it should be easy to review and approve the changes and then you can easily merge your resolve field context changes branch to develop so that the changes are separated

sungam3r:
What changes are you talking about? What exactly to revert? There are many of them and they are connected.

Shane32:
It’s not about naming. I don’t like the optimized resolve methods and related that were added. Once you have separated that out I’ll be glad to do a proper review and compare to some other ideas.

Shane32:
The resolve field context changes did not seem to be related to any of the prior work in this PR

sungam3r:
Today I still need to have time to finish #2165, because I promised you. I would not want to rolling back my work piece by piece, especially since I don't have much experience with git. Let this PR be under review for a while, we have time.

sungam3r:
Perhaps later, after the review #2165, I will create a bit of small PRs, taking away changes from this big PR in small chunks that raise no questions. But that's later. I've been programming for about 7 hours straight and I'm tired.

sungam3r:
@Shane32 #2183 is ready for review (before changes to resolvers)

sungam3r:
Not actual anymore.

PR #2167, opened 2021-01-13
Add Variable benchmarks

Shane32:


src/GraphQL.Benchmarks/Benchmarks/VariableBenchmark.cs
```diff
@@ -0,0 +1,267 @@
...
+            services.AddSingleton<MyQueryGraphType>();
+            services.AddSingleton<MyInputObjectGraphType>();
+            services.AddSingleton<MySubInputObjectGraphType>();
+            services.AddSingleton<ISchema, MySchema>();
+
+            _provider = services.BuildServiceProvider();
+            _schema = _provider.GetRequiredService<ISchema>();
+            _schema.Initialize();
+            _executer = new DocumentExecuter();
+            _queryLiteral = @"
```
sungam3r, line 42: Let's keep all such requests in separate files like here:
изображение

I moved them there to make it easier to edit them separately from the code. In addition, then syntax highlighting becomes available. Also no need in escaping like `""`.

sungam3r: https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql

src/GraphQL.Benchmarks/Benchmarks/VariableBenchmark.cs
```diff
@@ -0,0 +1,267 @@
...
+            Variable();
+            EnableValidation = val;
+        }
+
+        public IEnumerable<bool> TrueFalse => new[] { true, false };
+
+        [ParamsSource(nameof(TrueFalse))]
+        public bool EnableValidation { get; set; }
+
+        [Benchmark]
```
sungam3r, line 150:
```suggestion
        [Benchmark(Baseline = true)]
```

sungam3r:
👍 👍 👍

src/GraphQL.Benchmarks/Benchmarks/VariableBenchmark.cs
Shane32: That's a great idea.  For now, I just consolidated the queries in the same fashion as is done in the rest of this project.

src/GraphQL.Benchmarks/Benchmarks/VariableBenchmark.cs
Shane32: Committed, then reverted.  Produces meaningless output.

sungam3r (approved):

PR #2173, opened 2021-01-14
Add document caching interface - IDocumentCache

Shane32:
| Introspection |      False | 1,137.269 us | 6.0073 us | 5.6192 us | 115.2344 | 39.0625 |     - | 713.35 KB |
|          Hero |      False |    17.480 us | 0.1275 us | 0.1193 us |   1.5564 |       - |     - |   9.59 KB |
| Introspection |       True |   878.678 us | 4.5961 us | 4.2992 us | 109.3750 | 37.1094 |     - | 673.33 KB |
|          Hero |       True |     4.098 us | 0.0273 us | 0.0256 us |   0.7858 |       - |     - |   4.85 KB |

@sungam3r  Thoughts? Document caching can of course be implemented as a wrapper around one or a pair of document executors. But this is cleaner. I'm thinking:
* If pre-compiling execution nodes (#2016) is going to help, which I think it will, it will help notably more if document caching is used. So if we want to implement caching, I'd do that first.
* Once the parser is updated, it will reduce the improvement with caching. So I'd like to see the results after that PR is merged.

For my own use, I know that all queries are going to hit the cache, since there is a fixed set of queries coming from the front end code.  However, others that are exposing a graphql endpoint to the public may not find a cache to be helpful, and would probably reduce performance.  So caching defaults to disabled.


src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -74,11 +89,23 @@ public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options
...
 
                 var document = options.Document;
+                bool saveInCache = false;
+                bool analyzeComplexity = true;
+                var validationRules = options.ValidationRules;
                 using (metrics.Subject("document", "Building document"))
                 {
+                    if (document == null && _documentCache != null && (document = _documentCache[options.Query]) != null)
+                    {
+                        // none of the default validation rules are dependent on the inputs, and the
```
sungam3r, line 99:
```suggestion
                        // none of the default validation rules yet are dependent on the inputs, and the
```

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -74,11 +89,23 @@ public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options
                 }
 
                 var document = options.Document;
+                bool saveInCache = false;
+                bool analyzeComplexity = true;
+                var validationRules = options.ValidationRules;
                 using (metrics.Subject("document", "Building document"))
```
sungam3r, line 95: The line `  using (metrics.Subject("document", "Analyzing complexity"))` now is under `if analyzeComplexity` condition. Should it be done for `using (metrics.Subject("document", "Building document"))` ?

Shane32: I was thinking about that also. Either way is fine to me. Since the old behavior when supplying options.Document still logged the metric, I left it that way. There certainly is a very little time required for the dictionary lookup so maybe it should stay this way. What do you suggest?

src/GraphQL/Caching/DocumentCache.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Collections.Concurrent;
+using System.Linq;
+using GraphQL.Language.AST;
+
+namespace GraphQL.Caching
+{
+    /// <summary>
+    /// The default implementation of a document cache, limited by a configured amount of memory.
+    /// </summary>
+    public class DocumentCache : IDocumentCache
```
sungam3r, line 11: I don't think it makes sense to invent your own implementations. It is enough to delegate all the work to `IMemoryCache`. It will be easier, less error prone and more customizable. I suggest to use 
```c#
public sealed class DefaultDocumentCache : IDocumentCache
{
  public Document this[string query] { get => null; } { set => { } }
}
```
and get rid of some null checks in `DocumentExecuter` and alse pass `new DefaultDocumentCache()` into `DocumentExecuter` ctor (or even static single instance since default impl is stateless), not null + change ` _documentCache = documentCache;` to ` _documentCache = documentCache ?? throw new ArgumentNullException(nameof(documentCache));;

sungam3r: Of course `MemoryDocumentCache` should be placed in new project since it depends on custom implementation from `Microsoft.Extensions.Caching.Memory`

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: OK. Let's keep it as is since dictionary lookup is in fact some arbitrary code and in case of distributed dictionary (Redis, etc.) it _may_ run indefinitely long.

src/GraphQL/Caching/DocumentCache.cs
Shane32: There’s no dependency on caching right now, is there? It seems a shame to add another dependency for such a simple implementation.

Another idea: We could move the caching implementation to a separate library and add a dependency there. 

I’ll make all the rest of your suggestions.

sungam3r: > Another idea: We could move the caching implementation to a separate library and add a dependency there.

That's what I wanted.

sungam3r: And _default_ implementation is just "No caching".

sungam3r:
> @sungam3r Thoughts?

I think this is the right direction. Caching is disabled by default and does not consume resources. It will be easy to enable + we will provide an implementation of in-memory caching in a new project. I suggest `GraphQL.MemoryCache`.

docs2/site/docs/migrations/migration4.md
```diff
@@ -4,6 +4,7 @@
 
 * Extension methods to configure authorization requirements for GraphQL elements: types, fields, schema.
 * New property `GraphQL.Introspection.ISchemaComparer ISchema.Comparer { get; set; }`
+* Document caching supported via `IDocumentCache` and a default implementation within `DocumentCache`.
```
sungam3r, line 7:
```suggestion
* Document caching supported via `IDocumentCache` and a default implementation within `DefaultDocumentCache`.
```

sungam3r (changes requested):
LGTM except `MemoryDocumentCache` naive implementation. I repeat that I see no point in its existence as it is now. I would not like to maintain such code. In addition, if I need to use the cache, then I will 100% use the implementation that works through `IMemoryCache`. It has many more options https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheentryoptions?view=dotnet-plat-ext-3.1, is time-tested and is better integrated into the NET Core.

src/GraphQL.MemoryCache/GraphQL.MemoryCache.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <Description>GraphQL memory cache implementation on top of Microsoft.Extensions.Caching.Memory</Description>
+    <PackageTags>GraphQL;json;api;subscriptions</PackageTags>
```
sungam3r, line 6:
```suggestion
    <PackageTags>GraphQL;memory</PackageTags>
```

src/GraphQL.MemoryCache/GraphQL.MemoryCache.csproj
```diff
@@ -0,0 +1,17 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\GraphQL\GraphQL.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="3.1.11" />
```
sungam3r, line 14:
```suggestion
    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="3.1.0" />
```

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,58 @@
...
+            set => _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value, new MemoryCacheEntryOptions { SlidingExpiration = _slidingExpiration, Size = query.Length });
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _memoryCache.Dispose();
+        }
+    }
+
```
sungam3r, line 57:
```suggestion
```

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private readonly MemoryCache _memoryCache;
+        private readonly TimeSpan _slidingExpiration;
+
+        /// <summary>
+        /// Initializes a new instance with the specified parameters.
+        /// </summary>
+        /// <param name="maxTotalQueryLength">The total length of all queries cached in this instance. Assume maximum memory used is about 10x this value. Will not cache queries larger than 1/3 of this value. During cache compression, reduces cache size by 1/3 of this value.</param>
+        /// <param name="slidingExpiration">The maximum lifetime of queries cached within this instance.</param>
+        public MemoryDocumentCache(int maxTotalQueryLength, TimeSpan slidingExpiration)
```
sungam3r, line 21:
```suggestion
        public MemoryDocumentCache(IMemoryCache cache, MemoryDocumentCacheOptions options)
```

`IMemoryCache` is external (not owned) dependency. In fact it can be even distributed cache.

So `Dispose` method is not needed.

src/GraphQL/Caching/DocumentCache.cs
Shane32: Done

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
Shane32: A distributed cache cannot contain managed instances of classes, so deserialization is necessary.  At that point we might as well deserialize the query and save the slow network calls!  It certainly would be faster.  Besides, a distributed cache would not implement `IMemoryCache`; it would implement `IDistributedCache`.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.distributed.idistributedcache?view=dotnet-plat-ext-5.0

sungam3r: I said wrong, forget about `IDistributedCache`. My idea is about taking `IMemoryCache` as dependency so allowing _DI-friendly_ style.

Shane32: The challenging thing about sharing a `IMemoryCache` instance, besides the fact that there can be key collisions, is that the size parameter of entries relates only to how the cache was initialized.  For instance, here the number of bytes allocated is about 10x larger than the value provided to `Size`, due to the amount of memory consumed by the AST copy of the query.  Other users of a cache may use a `Size` value as a simple object counter (all entries have a `Size` of 1).  And other users may have a `Size` value that relates closely to the amount of memory required.  So typically `IMemoryCache` instances are not shared unless the `Size` parameter is not used (which is not recommended due to easy abuse by DOS attacks).

Finally, dictionary lookups should be faster with a private memory cache versus a shared one.  I really don't see a reason to allow an external `IMemoryCache`.  I can add a protected constructor in case someone wants to customize behavior.

sungam3r: Well...OK, then let's do that:
```c#
 public class MemoryDocumentCacheOptions : MemoryCacheOptions, IOptions<MemoryDocumentCacheOptions>
        {
            public TimeSpan SlidingExpiration { get; set; }

            MemoryDocumentCacheOptions IOptions<MemoryDocumentCacheOptions>.Value => this;
        }

        public MemoryDocumentCache(MemoryDocumentCacheOptions options) { }

        public MemoryDocumentCache(IOptions<MemoryDocumentCacheOptions> options) { }
```

to be DI-friendly and resistant to adding new options.

Shane32: I'm in the middle of a sample; take a look first; hold on a sec, ok?

sungam3r: 👍

Shane32: Ok I just pushed some protected members.  Take a look first ok?

Shane32: > I said wrong, forget about `IDistributedCache`. My idea is about taking `IMemoryCache` as dependency so allowing _DI-friendly_ style.

So I think you're suggesting we move the required options to a new options class so it's DI friendly.  That makes sense.

sungam3r: Yes. I would like to keep `GetMemoryCacheEntryOptions` method but remove all protected ctors in favor of ` public MemoryDocumentCache(IOptions<MemoryDocumentCacheOptions> options) { }` since this one ctor covers all cases.

Shane32: Ok, options class done.  We can change the additional constructors to `private`, or leave them `public`.  There's also the `protected` members if someone wants to customize behavior further.

Shane32: > Yes. I would like to keep `GetMemoryCacheEntryOptions` method but remove all protected ctors in favor of ` public MemoryDocumentCache(IOptions<MemoryDocumentCacheOptions> options) { }` since this one ctor covers all cases.

I will remove the other `public` constructors then, leaving the `protected` constructors, as they are necessary to override behavior.

src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using Microsoft.Extensions.Options;
+
+namespace GraphQL.Caching
+{
+    /// <summary>
+    /// Provides configuration options for <see cref="MemoryDocumentCache"/>
```
sungam3r, line 7:
```suggestion
    /// Provides configuration options for <see cref="MemoryDocumentCache"/>.
```

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,88 @@
...
+                slidingExpiration)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance with the specified parameters.
+        /// </summary>
+        /// <param name="options">The memory cache configuration settings. Note that the <see cref="MemoryCacheOptions.SizeLimit"/> field is required.</param>
+        /// <param name="slidingExpiration">The maximum lifetime of queries cached within this instance.</param>
+        public MemoryDocumentCache(IOptions<MemoryCacheOptions> options, TimeSpan slidingExpiration)
```
sungam3r, line 44: ?

`MemoryDocumentCacheOptions` already has `SlidingExpiration` property.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace GraphQL.Caching
+{
+    /// <summary>
+    /// A basic implementation of a document cache, limited by a configured amount of memory.
+    /// </summary>
+    public class MemoryDocumentCache : IDocumentCache, IDisposable
+    {
+        private readonly IMemoryCache _memoryCache;
+        private readonly TimeSpan _slidingExpiration;
```
sungam3r, line 14:
```suggestion
        private readonly MemoryDocumentCacheOptions _options;
```

```c#
   public MemoryDocumentCache(IOptions<MemoryDocumentCacheOptions> options)
   { 
        _options = options.Value;
   }
```

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
Shane32: Done.  I could remove the `protected` members also, but they shouldn't hurt anything.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
Shane32: Updated

docs2/site/docs/migrations/migration4.md
```diff
@@ -4,6 +4,7 @@
 
 * Extension methods to configure authorization requirements for GraphQL elements: types, fields, schema.
 * New property `GraphQL.Introspection.ISchemaComparer ISchema.Comparer { get; set; }`
+* Document caching supported via `IDocumentCache` and a default implementation within `DefaultDocumentCache`.
```
Shane32, line 7:
```suggestion
* Document caching supported via `IDocumentCache` and a default implementation within `DefaultDocumentCache`.
  Within the `GraphQL.Caching` nuget package, a memory-backed implementation is available which is backed by `Microsoft.Extensions.Caching.Memory.IMemoryCache`.
```

src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Provides configuration options for <see cref="MemoryDocumentCache"/>.
+    /// </summary>
+    public class MemoryDocumentCacheOptions : IOptions<MemoryDocumentCacheOptions>
+    {
+        /// <summary>
+        /// The maximum total length of all queries cached. Assume maximum memory used is about 10x this value.
+        /// </summary>
+        public int MaxTotalQueryLength { get; set; }
```
sungam3r, line 14: It is always encouraged to give some meaningful defaults to all options. I suggest `100;` for `MaxTotalQueryLength` and `TimeSpan.FromMinutes(10)` for `SlidingExpiration`. The default configuration is currently broken.

sungam3r, line 14:
```suggestion
        public long? MaxTotalQueryLength { get; set; }
```

By the way, `MemoryCacheOptions.SizeLimit` is nullable long.

Shane32: I know

Shane32: Setting the value to null will crash the code when attempting to add an entry with a `Size` value.  Not really sure what value there is to allowing a cache over 20GB.

Shane32: Fine, but 100 ? An introspection query is 1.3kb.  The kitchensink query is 800 bytes.  I would have to say that it is not meaningful if set to 100.  Perhaps 100,000?

sungam3r: Then why to limit it to not nullable int ?
`MemoryCacheOptions.SlidingExpiration` is also nullable which means that all checks in the ctor are not needed.

sungam3r: What if someone wants to have a cache without size limit and no expiration?

Shane32: Making `SlidingExpiration` nullable makes sense.

Shane32: We can set the both to nullable, but the default options would then result in a cache without limit whose entries are never removed from the cache.

Shane32: I think we should force people to set a size limit.  Do you not agree?  They can always derive from the class for unique situations, utilizing the protected members.

sungam3r: > Setting the value to null will crash the code when attempting to add an entry with a Size value.

?

sungam3r: > I think we should force people to set a size limit. Do you not agree? They can always derive from the class for unique situations, utilizing the protected members.

I do not agree. We **should not** force to derive from our base class just in case of allowing some option. We **should** provide good defaults.

sungam3r: Any value != `null` suits me.

Shane32: I could be wrong...

src/GraphQL.Tests/Caching/MemoryCacheTests.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace GraphQL.Tests.Caching
+{
+    public class MemoryCacheTests
+    {
+        [Fact]
+        public void ValidateEntryIsCached()
+        {
+            var doc = new Document();
+            var query = "test";
+            var memoryCache = new MemoryDocumentCache(new MemoryDocumentCacheOptions);
```
sungam3r, line 15:
```suggestion
            var memoryCache = new MemoryDocumentCache(new MemoryDocumentCacheOptions());
```

src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Provides configuration options for <see cref="MemoryDocumentCache"/>.
+    /// </summary>
+    public class MemoryDocumentCacheOptions : IOptions<MemoryDocumentCacheOptions>
+    {
+        /// <summary>
+        /// The maximum total length of all queries cached. Assume maximum memory used is about 10x this value.
+        /// </summary>
+        public long MaxTotalQueryLength { get; set; } = 100000;
```
sungam3r, line 14:
```suggestion
        public long? MaxTotalQueryLength { get; set; } = 100000;
```

src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace GraphQL.Caching
+{
+    /// <summary>
+    /// Provides configuration options for <see cref="MemoryDocumentCache"/>.
+    /// </summary>
+    public class MemoryDocumentCacheOptions : IOptions<MemoryDocumentCacheOptions>
+    {
+        /// <summary>
+        /// The maximum total length of all queries cached. Assume maximum memory used is about 10x this value.
```
sungam3r, line 12:
```suggestion
        /// The maximum total length of all queries cached. Assume maximum memory used is about 10x this value. Defaults to 100000.
```

src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
Shane32: Ok

Shane32: > What if someone wants to have a cache without size limit and no expiration?

Presumably they will crash their computer.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// </summary>
+    public class MemoryDocumentCache : IDocumentCache, IDisposable
+    {
+        private readonly IMemoryCache _memoryCache;
+        private readonly MemoryDocumentCacheOptions _options;
+
+        /// <summary>
+        /// Initializes a new instance with the default options: 100,000 maximum total query size and no expiration time.
+        /// </summary>
+        public MemoryDocumentCache()
```
sungam3r, line 19: 👍 I was already ready to write this with another comment.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        /// <summary>
+        /// Initializes a new instance with the specified memory cache and sliding expiration time period.
+        /// Note that by overriding <see cref="GetMemoryCacheEntryOptions(string)"/>, the sliding expiration
+        /// time specified within <paramref name="options"/> can be ignored.
+        /// </summary>
+        protected MemoryDocumentCache(IMemoryCache memoryCache, IOptions<MemoryDocumentCacheOptions> options)
+        {
+            _memoryCache = memoryCache ?? throw new ArgumentNullException(nameof(memoryCache));
+            _options = options?.Value;
```
sungam3r, line 43:
```suggestion
            _options = options?.Value ?? throw new ArgumentNullException(nameof(options));;
```

Shane32: Options is not necessary if they are overriding `GetMemoryCacheEntryOptions`

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance with the specified options.
+        /// </summary>
+        /// <param name="options">A value containing the <see cref="MemoryDocumentCacheOptions"/> to use.</param>
+        public MemoryDocumentCache(IOptions<MemoryDocumentCacheOptions> options)
+            : this(
+                new MemoryCache(new MemoryCacheOptions { SizeLimit = options.Value.MaxTotalQueryLength }),
```
sungam3r, line 30:
```suggestion
                new MemoryCache(options),
```

Shane32: These are not the same class.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,71 @@
...
+            get => _memoryCache.TryGetValue<Document>(query, out var value) ? value : null;
+            set => _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value, GetMemoryCacheEntryOptions(query));
+        }
+
+        /// <summary>
+        /// Disposes of the underlying <see cref="IMemoryCache"/> instance.
+        /// </summary>
+        public virtual void Dispose()
+        {
+            _memoryCache.Dispose();
```
sungam3r, line 68:
```suggestion
            if (_owned)
                _memoryCache.Dispose();
```

```c#
private readonly bool _owned;
```
\+ set `_owned` to `true/false` in appropriate ctors.

src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
sungam3r: Well, so be it.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
sungam3r: Yeah, OK.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
Shane32: This is not typical for .NET.  For example, `StreamWriter`'s `Dispose` method disposes of it's underlying `Stream` regardless of whether it was constructed with a `Stream` or a `string`.  If the user does not want the underlying object disposed, they should not call `Dispose` on this object.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
sungam3r: But....

```c#
 _options = options?.Value ?? (GetType() == typeof(MemoryDocumentCache) ? null : throw new ArgumentNullException(nameof(options)));
```
😄

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
Shane32: (I don't disagree with your logic, but it doesn't match the .NET standard.)

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/pull/2173#discussion_r557690734
изображение

I didn't even notice 😄

Shane32: We can..... but members are not virtual so we cannot override `SizeLimit` with a meaningful xml comment.

Personally I don't like inheriting option classes.  It caused problems in the serialization libraries here for the same reason.

Do you want to add a `MemoryCacheOptions` property?

Shane32: I really don't care much but it's just a bad design in my opinion.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
sungam3r: >  If the user does not want the underlying object disposed, they should not call Dispose on this object.

😕 This question is outside the scope of his desires when it comes to work within the DI-container. Imagine that `DocumentCache` is configured like scoped/transient. Then after some request `DocumentCache.Dispose()` kills `IMemoryCache` dependency configured as singleton and oooooops.

Shane32: I understand the problems with that design in those circumstances (typically the scope would be the same, of course).  I am stating that it is not the .NET standard that programmers are accustomed to.  We have to deviate off from the standard if we implement the change.

sungam3r: Example with `StreamWriter` is bad example. The base rule is `DO NOT DISPOSE OBJECTS WHICH YOU DIDN'T CREATE` and it works in the vast majority of cases, especially in the realities of modern .NET Core apps, when 99% of apps use DI.

Shane32: Do you think this change is worth breaking the well-developed .NET standard for how Dispose works here?

Shane32: Are you saying that the example is bad or `StreamWriter` is bad?  I use `StreamWriter` all the time.  And I agree that it's a flawed design.

Shane32: If you say that we should break off from the standard, I'll make the change.  I don't have a problem changing it.

Shane32: Since the constructor in question is protected, it is very likely that it IS owned.  So I will add an argument to the constructor as well.

sungam3r: >  well-developed .NET standard for how Dispose works

Shane32: However `StreamWriter` was written. I do not know if it is documented.  It just "is".

sungam3r: > Are you saying that the example is bad or StreamWriter is bad? 

`StreamWriter` design _was_ bad. So one day MS team added one more overload to mitigate this https://stackoverflow.com/questions/2666888/is-there-any-way-to-close-a-streamwriter-without-closing-its-basestream

And since `StreamWriter` is bad then example based on it is bad too...

Shane32: Never said it wasn't.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
sungam3r: >  so we cannot override SizeLimit with a meaningful xml comment.

No problem.

>  It caused problems in the serialization libraries here for the same reason.

What problems? I suggested inheriting because only one new property actually needed - SlidingExpiration. This property already exists on the `CacheEntry` level but we "pick it up" to ctor options level so we can access it from `GetMemoryCacheEntryOptions`.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
sungam3r: Then I don't understand the point of giving a bad design counter-argument that itself uses bad design. Well, or I translated something wrong.

src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace GraphQL.Caching
+{
+    /// <summary>
+    /// Provides configuration options for <see cref="MemoryDocumentCache"/>.
+    /// </summary>
+    public class MemoryDocumentCacheOptions : MemoryCacheOptions, IOptions<MemoryDocumentCacheOptions>
+    {
+        public MemoryDocumentCacheOptions()
+        {
+            SizeLimit = 100000;
```
sungam3r, line 14: `SizeLimit` and `MaxTotalQueryLength` both means the same. When I showed my example at the very beginning of the review, I had it in mind even then, just then I didn't notice that inheritance was lost.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Initializes a new instance with the specified memory cache and sliding expiration time period.
+        /// Note that by overriding <see cref="GetMemoryCacheEntryOptions(string)"/>, the sliding expiration
+        /// time specified within <paramref name="options"/> can be ignored.
+        /// </summary>
+        /// <param name="memoryCache">The memory cache instance to use.</param>
+        /// <param name="disposeMemoryCache">Indicates if the memory cache is disposed when this instance is disposed.</param>
+        /// <param name="options">Provides option values for use by <see cref="GetMemoryCacheEntryOptions(string)"/>; optional.</param>
+        protected MemoryDocumentCache(IMemoryCache memoryCache, bool disposeMemoryCache, IOptions<MemoryDocumentCacheOptions> options)
```
sungam3r, line 45: Not DI-friendly but it's OK since we force user to write descendant with appropriate ctor.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
Shane32: > No problem.

?

sungam3r: I mean that the inability to write a comment should not be an obstacle to avoiding duplicate meaning of options. These are not new options, but an extension of existing ones. We could use `MemoryCacheOptions` directly if not `SlidingExpiration` setting.

src/GraphQL.Tests/Caching/MemoryCacheTests.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace GraphQL.Tests.Caching
+{
+    public class MemoryCacheTests
+    {
+        [Fact]
+        public void ValidateEntryIsCached()
+        {
+            var doc = new Document();
+            var query = "test";
+            var memoryCache = new MemoryDocumentCache(new MemoryDocumentCacheOptions());
```
Shane32, line 15:
```suggestion
            var memoryCache = new MemoryDocumentCache();
```

src/GraphQL.Benchmarks/Benchmarks/ExecutionBenchmark.cs
```diff
@@ -34,6 +39,7 @@ public void GlobalSetup()
             _schema = _provider.GetRequiredService<ISchema>();
             _schema.Initialize();
             _executer = new DocumentExecuter();
+            _cachedExecuter = new DocumentExecuter(new GraphQLDocumentBuilder(), new DocumentValidator(), new ComplexityAnalyzer(), new MemoryDocumentCache(new MemoryDocumentCacheOptions()));
```
Shane32, line 42:
```suggestion
            _cachedExecuter = new DocumentExecuter(new GraphQLDocumentBuilder(), new DocumentValidator(), new ComplexityAnalyzer(), new MemoryDocumentCache());
```

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            _memoryCache = memoryCache ?? throw new ArgumentNullException(nameof(memoryCache));
+            _options = options?.Value;
+            _memoryCacheIsOwned = disposeMemoryCache;
+        }
+
+        /// <summary>
+        /// Returns a <see cref="MemoryCacheEntryOptions"/> instance for the specified query.
+        /// Defaults to setting the <see cref="MemoryCacheEntryOptions.SlidingExpiration"/> value as specified
+        /// in the constructor, and the <see cref="MemoryCacheEntryOptions.Size"/> value to the length of the query.
```
sungam3r, line 57:
```suggestion
        /// in options, and the <see cref="MemoryCacheEntryOptions.Size"/> value to the length of the query.
```

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -0,0 +1,77 @@
...
+        public MemoryDocumentCache(IOptions<MemoryDocumentCacheOptions> options)
+            : this(
+                new MemoryCache(options),
+                true,
+                options)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance with the specified memory cache and sliding expiration time period.
```
sungam3r, line 40:
```suggestion
        /// Initializes a new instance with the specified memory cache and options.
```

sungam3r:
Just two little edits in comments and this PR looks perfect to me.

sungam3r:
Feel free to merge if suggestions are good. I'm done for today. The time budget is limited and today it was spent on this PR, I will look at #2165 tomorrow.

sungam3r (approved):

PR #2175, opened 2021-01-14
SchemaPrinter can print inputobject default values

sungam3r:
before
```graphql

type Query {
  str(argOne: SomeInput! = GraphQL.Tests.Utilities.SchemaPrinterTests+SomeInput): String!
}

input SomeInput {
  age: Int!
  name: String!
  isDeveloper: Boolean!
}
```

after
```graphql
type Query {
  str(argOne: SomeInput! = { age: 42, name: "Tom", isDeveloper: true }): String!
}

input SomeInput {
  age: Int!
  name: String!
  isDeveloper: Boolean!
}
```

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>3.3.0-preview</VersionPrefix>
+    <VersionPrefix>3.3.1-preview</VersionPrefix>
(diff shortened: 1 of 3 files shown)


Shane32 (approved):

sungam3r:
That bug affects my work.

Shane32:
Release 3.3.1 then

sungam3r:
Yes, I will do so a little later.

PR #2177, opened 2021-01-15
[WIP] 'Precompile' an execution tree to optimize performance on large requests

Shane32:
This contains the precompilation code from PR #2016 retargeted to the `develop` branch.  Thanks to @IdeaHunter for supplying this PR!

### Before

| Introspection |      False | 1,116.985 us | 4.4080 us | 3.9076 us | 115.2344 | 39.0625 |     - | 713.36 KB |
|          Hero |      False |    17.541 us | 0.2305 us | 0.2156 us |   1.5564 |       - |     - |   9.59 KB |
| Introspection |       True |   881.542 us | 3.6298 us | 3.3953 us | 109.3750 | 38.0859 |     - | 673.39 KB |
|          Hero |       True |     3.990 us | 0.0340 us | 0.0318 us |   0.8011 |       - |     - |   4.92 KB |

### After

|        Method | UseCaching |       Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |----------- |-----------:|----------:|----------:|---------:|--------:|------:|----------:|
| Introspection |      False | 956.568 us | 3.4724 us | 2.8996 us | 102.5391 | 35.1563 |     - | 632.03 KB |
|          Hero |      False |  19.071 us | 0.0882 us | 0.0825 us |   1.8311 |       - |     - |  11.24 KB |
| Introspection |       True | 722.251 us | 2.8166 us | 2.4968 us |  95.7031 | 30.2734 |     - | 592.07 KB |
|          Hero |       True |   4.979 us | 0.0330 us | 0.0257 us |   1.0681 |  0.0076 |     - |   6.57 KB |

As can be seen from the above statistics, speed and memory changes drastically for the better when processing lists.  However, extremely small queries, such as the hero query, demonstrate a performance reduction.

I believe much more optimization is possible with cached queries, as the "compiled" query can typically be cached as well.  However, compilation changes currently if there is a directive based on a variable or if a different operation is selected for the same document.  I'm sure these are challenges that can be overcome.

Currently this PR:
* Needs xml comments
* Fails some tests due to changes to federated results.  See https://github.com/graphql-dotnet/graphql-dotnet/pull/2016#discussion_r527193359

I suggest we review this PR before making changes to allow caching of the compiled representation.  However, here are some preliminary results if the compilation is cached with the document:

### With cached compilation

|        Method | UseCaching |       Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |----------- |-----------:|----------:|----------:|---------:|--------:|------:|----------:|
| Introspection |      False | 946.143 us | 2.8453 us | 2.6615 us | 102.5391 | 35.1563 |     - | 632.05 KB |
|          Hero |      False |  18.566 us | 0.0924 us | 0.0772 us |   1.8311 |       - |     - |  11.25 KB |
| Introspection |       True | 664.145 us | 3.4373 us | 3.2153 us |  86.9141 | 27.3438 |     - | 536.36 KB |
|          Hero |       True |   3.441 us | 0.0136 us | 0.0127 us |   0.6943 |  0.0038 |     - |   4.26 KB |

As can be seen, for cached queries, the 'hero' sample now also demonstrates a benefit, in addition to an even greater benefit for large cached queries.

Note that these changes don't actually change much code, but they are breaking changes and would need to be added during a major release, such as 4.0.


sungam3r:
I propose to postpone this. We should to deal with other (more understandable) PRs first.

Shane32:
> I propose to postpone this. We should to deal with other (more understandable) PRs first.

Fine with me.  As can be seen in the performance metrics, I do see a lot of potential here, especially when coupled with caching.

Shane32:
@IdeaHunter 

> What obstacles do you have?

To test precompiled queries, we can easily change a few lines of code to add a `CompiledNode` member to `Document` which doesn't compile if already set.  You can then see the performance metrics that I listed above. However, there are a few problems with this implementation:
1. Evaluation of directives changes compliation results.  Since directive arguments can be based off of variable values, it is not certain that the compilation result may be cached.  There are two potential fixes I foresee:
    * Compile the nodes regardless of directives, adding a `Func<Variables, bool>` delegate to the node which executes to determine if the node should be included.  For nodes with directives, the delegate can process the directive arguments and return a boolean.  For nodes without directives, this can be `() => true`.
    * Do not cache queries that have directives and where those directives are based on variables.  As long as any directive values are not based on variables, the compiled query can still be cached.
2. Queries are compiled only for the operation requested. There are a few potential solutions:
    * When caching is enabled, compile all of the operations for a query.
    * Add a ConcurrentDictionary to the cache entry corresponding to each of the named operations, caching the compiled query individually when the named operation is requested.
    * Separate cache entries based on both query string and operation name.
    * Do not cache entries containing more than one operation.

In addition to the above, we need confirmation that there are no problems with federation.

> And what can be done to put in 4.0?

@sungam3r is still hoping to release 4.0 by the end of the month.  In order to get it into 4.0, we would probably need all of the problems solved and implemented "yesterday" so we have time to review, add migration notes, etc.

Note that we have been making a lot of performance related improvements for 4.0. It's probably 30% faster and consumes 50% less memory already compared to 3.x.  I think that it would be nice to get this PR into 4.0 along with the rest of the performance upgrades, especially since otherwise it would have to wait for 5.0, but it would take a bit of effort immediately.  I just don't have any more time to spend on it right now.  If you write the code, I can review it.

Finally, it must be noted that at the end of the day, even if we figure out all the issues, this PR appears to provide performance benefits to (a) all cached queries (b) uncached queries with lists, but not uncached queries without lists.  Seems fine to me.

IdeaHunter:
@Shane32 

Thanks for big evaluation of the problem. 
The idea behind a compilation PR was to cache fields/types in context of single execution without any forward looking primary because of directives messing things around. 

And in general if im not mistaken we talking about 3 things:
1) Document caching - which have little to nothing to do with compilation because compilation as it is now works of the document+variables(as original code without one did)
2) Fields/Types caching (aka compilation) - which shouldnt do the difference we observing for simple queries because it does whatever code would be doing anyway even without this PR, but also allocating a little bit of memory to cache the info.
3) Caching of Fields/Types caching (aka compilation caching) - which is not sound quite right once you spell full name of it. IMHO it should rather be the caching that superseeds (1) and (2) together with some twist.

Also the objection to (3) is GC pressure which still quite the problem and adding more caching only adds more fuel to fire. 

So in the light of this discussion do you still think that caching of compilation is good idea as of this PR? Because if slowness on small queries caused by mostly GC it would be better solve the GC problem first and get overral performance improvement first, and then think how to cache it properly to give even more easier life to GC

Shane32:
This PR splits the execution phase into two: compiling and executing.  Common sense says that two passes are going to be slower than one pass, requiring more objects to be allocated, unless we re-use the compiled nodes (either via lists or document caching), or unless there is some other way to gain performance.  For example, if `ExecutionNode` could become a struct.

> Because if slowness on small queries caused by mostly GC it would be better solve the GC problem first and get overral performance improvement first

Yes, ideally we would be able to demonstrate performance gains whether or not caching is enabled, and then even greater performance with caching enabled.  I have not done enough experimentation and testing to know if that is possible.

For my own website that uses a graphql back-end, all queries are predictable and therefore cached, so this PR would demonstrate performance gains as soon as we can integrate it with caching, no matter if the query returned lists or not.

Other people that expose their graphql endpoint to the public may not be able to benefit from caching, and so would not see the same results.  They would see a performance increase or decrease with this PR depending on how many queries returned lists.

> So in the light of this discussion do you still think that caching of compilation is good idea as of this PR?

Yes, I think it is a necessary part of this PR.  Integrating this with caching should not be hard to implement, and no matter how much of a performance increase we can attain with compiled queries, it should demonstrate an even better increase in performance if the compiled queries can be cached.  (Even if we do something extremely basic, like only cache compiled queries that have a single operation and do not have directives.)  I just don't have the time to work on it just now.

sungam3r:
> Because if slowness on small queries caused by mostly GC it would be better solve the GC problem first and get overral performance improvement first, and then think how to cache it properly to give even more easier life to GC

The same thought is constantly running through my head.

sungam3r:
> In addition to the above, we need confirmation that there are no problems with federation.

I know nothing about federation 😞

sungam3r:
Do we need this?

Shane32:
> Do we need this?

I would like to run some more benchmarks after our latest updates.  Previously there was a definite performance benefit for large queries.  The hard part is fixing all the conflicts.....

sungam3r:
@Shane32 what conclusion will you have on this PR?

Shane32:
> @Shane32 what conclusion will you have on this PR?

I don't have time to work on it at the moment.  If the conflicts can be fixed, it's pretty easy to run performance tests and assess the benefit.

sungam3r:
I'm looking through old issues and checked this one. Perhaps the effect from these changes is no longer like a year ago. Personally, these changes seem to me as complicate.

Shane32:
As these changes affect execution (which has not changed much), rather than parsing/serialization (which has), I do not believe the effect will have changed substantially from my comments and performance comparisons above.  However, it would be complicated to implement, and I have no plans to attempt/complete an implementation at this time.

Shane32:
Closing PR as abandoned

PR #2181, opened 2021-01-15
Fix tests on .NET 4.8

Shane32:


src/GraphQL.Tests/StringExtensionTests.cs
```diff
@@ -33,7 +33,9 @@ public void ToConstantCase(string input, string expected)
         [InlineData(typeof(void), "Void")]
         [InlineData(typeof(int), "Int32")]
         [InlineData(typeof(Dictionary<string, bool>), "Dictionary<String,Boolean>")]
+#if !NET48 //hangs test process after successfully compiling, prior to actually running the test
```
sungam3r, line 36:
```suggestion
#if !NET48 //HACK: hangs test process after successfully compiling, prior to actually running the test
```

sungam3r:
WOW

sungam3r (approved):

Shane32:
Yeah I just kept deleting folders and files from GraphQL.Tests until it finally worked.  Traced it to that line of code.  No idea why.

PR #2183, opened 2021-01-16
Memory allocation optimizations [part 1]

sungam3r:
See #2166


src/GraphQL/TaskExtensions.cs
```diff
@@ -41,17 +38,5 @@ internal static object GetResult(this Task task)
                 }
             }
         }
-
-        internal static async Task<IEnumerable<T>> WhereAsync<T>(this IEnumerable<T> items, Func<T, Task<bool>> predicate)
```
sungam3r, line 45: Not used anymore.

src/GraphQL/Types/Collections/TypeFields.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+
+namespace GraphQL.Types
+{
+    /// <summary>
+    /// A class that represents a set of fields for <see cref="IComplexGraphType"/> i.e <see cref="ComplexGraphType{TSourceType}"/>.
+    /// </summary>
+    public class TypeFields : IEnumerable<FieldType>
+    {
+        internal List<FieldType> List { get; } = new List<FieldType>();
```
Shane32, line 12: May be able to remove this.  See below note.

Shane32 (changes requested):

src/GraphQL/ResolveFieldContext/ResolveFieldContext.cs
```diff
@@ -89,6 +94,8 @@ public class ResolveFieldContext : IResolveFieldContext
         /// </summary>
         public ResolveFieldContext(IResolveFieldContext context)
         {
+            _baseContext = context;
```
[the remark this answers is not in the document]
Shane32: See #2184 for a suggested set of changes.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
```diff
@@ -16,6 +16,17 @@ namespace GraphQL
     /// </summary>
     public interface IResolveFieldContext : IProvideUserContext
     {
+        /// <summary>
+        /// Gets an array of the specified minimum length from the pool. This array does not need
+        /// to be returned to the pool yourself because it will be returned to the pool once the
+        /// execution completes. It is important that you should not use this array anyway after
+        /// execution, otherwise its contents will most likely be overwritten at any point in time.
```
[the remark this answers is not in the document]
Shane32: See #2184 for some additional changes suggested here.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -119,14 +128,14 @@ private void CheckSealed()
         /// <param name="resolveType">A delegate which returns an instance of a graph type from its .NET type.</param>
         /// <param name="nameConverter">A name converter to use for the specified graph types.</param>
         /// <param name="seal">Prevents additional types from being added to the lookup table.</param>
-        public static GraphTypesLookup Create(
+        internal static SchemaTypes Create(
```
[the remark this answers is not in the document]
Shane32:
```suggestion
        public static SchemaTypes Create(
```

src/GraphQL/Types/Collections/TypeFields.cs
```diff
@@ -0,0 +1,37 @@
...
+            if (!List.Contains(field))
+                List.Add(field);
+        }
+
+        public bool Contains(FieldType field) => List.Contains(field ?? throw new ArgumentNullException(nameof(field)));
+
+        public bool Contains(IFieldType field) => List.Contains((FieldType)field ?? throw new ArgumentNullException(nameof(field)));
+
+        /// <inheritdoc cref="IEnumerable.GetEnumerator"/>
+        public IEnumerator<FieldType> GetEnumerator() => List.GetEnumerator();
```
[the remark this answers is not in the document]
Shane32: I took a look in more detail.  It seems that this idea might work well for some of the new classes, and not so much for others.  Your call if you want to implement it or not.

src/GraphQL/ResolveFieldContext/ResolveFieldContext.cs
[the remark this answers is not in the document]
sungam3r: Ok. I only doubt about naming - _collective_.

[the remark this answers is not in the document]
sungam3r: Done

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -1901,10 +1895,10 @@ namespace GraphQL.Types
     public interface ISchema : GraphQL.Types.IProvideMetadata
     {
         System.Collections.Generic.IEnumerable<System.Type> AdditionalTypes { get; }
-        System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphType> AllTypes { get; }
+        GraphQL.Types.SchemaTypes AllTypes { get; }
```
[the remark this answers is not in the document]
sungam3r: Done. Also see #2185

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -61,20 +63,15 @@ public class GraphTypesLookup
...
-        /// Initializes a new instance with the <see cref="CamelCaseNameConverter"/>.
+        /// Initializes a new instance with the specified <see cref="INameConverter"/>.
         /// </summary>
-        public GraphTypesLookup() : this(CamelCaseNameConverter.Instance) { }
-
-        /// <summary>
-        /// Initalizes a new instance with the specified <see cref="INameConverter"/>.
-        /// </summary>
-        public GraphTypesLookup(INameConverter nameConverter)
+        private SchemaTypes(INameConverter nameConverter)
```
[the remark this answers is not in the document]
sungam3r: >  So right now I'd say that we should expose only Create as public (no constructors)

That's exactly what I did after review. Intuition tells me not to expose "naked" `IEnumerable<>` properties. Separate classes like `SchemaTypes`, `PossibleTypes`, `ResolvedInterfaces` etc. will allow to group the corresponding APIs in them and also use some internal bits as you can see for `List` properties.

src/GraphQL/Types/Collections/SchemaTypes.cs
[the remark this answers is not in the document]
sungam3r: 🤦  Of course. Done.

src/GraphQL/Types/Collections/TypeFields.cs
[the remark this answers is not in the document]
sungam3r: > so we may (or may not) be able to remove the internal List<FieldType> List property

Why delete it? Moreover `List<FieldType>.Enumerator` exposes implementation details. All these classes are needed by us rather than by the library clients. It is unlikely that anyone uses these APIs often. I deliberately left the compatibility for `IEnumerable<>` with external code but all internal GraphQL.NET bits began to use internal properties (List).

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -60,18 +60,21 @@ public class ExecutionContext : IExecutionContext
...
+            return array;
         }
 
-        internal void ReturnArrays()
+        private readonly List<Array> _trackedArrays = new List<Array>();
+
+        /// <summary>
+        /// Releases any rented arrays back to the backing memory pool.
+        /// </summary>
+        public void Dispose()
```
Shane32, line 77: It is extremely important that ANY implementation of `IDisposable` be able to be called multiple times.  Doing so in this instance will return the same arrays to the pool again.  That is why it is very important to clear `_trackedArrays`.  It has nothing to do with clearing the variables or not.

If you'd rather implement a pattern where there's an `_disposed` property and `if (_disposed) return;` and etc, be my guest.  But you must anticipate that `Dispose` may be called multiple times.

src/GraphQL/Execution/IArrayPool.cs
```diff
@@ -0,0 +1,22 @@
+namespace GraphQL.Execution
+{
+    /// <summary>
+    /// Provides a resource pool of temporary arrays during query execution.
+    /// Can be used to return lists of data from field resolvers.
+    /// </summary>
+    /// <inheritdoc cref="System.Buffers.ArrayPool{T}"/>
```
Shane32, line 7:
```suggestion
```
I probably forgot to remove this line lol.

src/GraphQL/Execution/IArrayPool.cs
```diff
@@ -0,0 +1,22 @@
...
+    public interface IArrayPool
```
Shane32, line 8:
```suggestion
    public interface IExecutionArrayPool
```
I suggested `IExecutionArrayPool` because the naming helps reinforce to a user that it is for the lifetime of the execution, rather than a generic shared array pool.  It also matches the comments.  If you really don't like that, it's okay though.

Shane32 (changes requested):

src/GraphQL/Types/Collections/TypeFields.cs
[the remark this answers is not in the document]
Shane32: Well, we could write our own struct enumerator, but the one provided by `List` works just fine.  I don't really care, but I thought this would provide a cleaner codebase for us to work within, as we can use a standard `foreach` syntax without having to access `List`.  And less chance of error within future changes.  Moreover, anyone else who may ever enumerate through arguments or fields on a graph will also benefit. It's still fully `IEnumerator` compatible, and the members are public bits of .NET Standard 2.0.  I'm a bit surprised you don't like the idea, that's all.  It doesn't make much difference to me.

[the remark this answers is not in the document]
Shane32: > `List<FieldType>.Enumerator` exposes implementation details. 

Yeah, if we did implement it, it would be hard to change the internals to anything except a list.  And writing custom struct enumerators for each class isn't worth it.

src/GraphQL/Execution/ExecutionContext.cs
sungam3r: 🤦 Done. I decided to clear the context completely. This is a relatively large object with 17 properties. Let's apply pooling to it.

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -28,13 +28,13 @@ public class ExecutionContext : IExecutionContext
         public Operation Operation { get; set; }
 
         /// <inheritdoc/>
-        public Fragments Fragments { get; set; } = new Fragments();
+        public Fragments Fragments { get; set; }
```
sungam3r, line 31: `DocumentExecuter.BuildExecutionContext` sets `Fragments` to `document.Fragments` so this allocation was redundant.

Shane32: Can we eliminate the property’s memory use and make it just point to document.Fragments ?

src/GraphQL/Types/Collections/TypeFields.cs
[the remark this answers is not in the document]
sungam3r: Yes. It much easier to type `.List`.

src/GraphQL/Execution/IArrayPool.cs
sungam3r: 50/50 . It's already in `GraphQL.Execution` namespace so...  but on other hand I see this:
изображение

😕

src/GraphQL/Execution/ExecutionContext.cs
sungam3r: And what's the point in this property then? Let's remove it.

src/GraphQL/Execution/IArrayPool.cs
Shane32: If you’re 50/50, let’s change it to IExecutionArrayPool

sungam3r:
@Shane32 I see very interesting consequences after clearing execution context.
изображение

sungam3r:
Of course I can store `context.Document` in local variable and pass it further.

sungam3r:
And this method should be rewritten because we can not use `ExecutionContext` anymore after it has been disposed
изображение

sungam3r:
Someone can save a reference to this context. In principle, I see no obstacles to its destruction, because `ExecutionContext` lifetime is highly deterministic. For now, we just have to honor the "wrong" usage. Do you agree?

Shane32:
Not sure exactly what you're saying.  Of course someone can save a reference to a disposed object. -- ?  In any case, I'd personally work on these changes in another PR so we can get this one finished.  Simply add the one line of code to clear the array pool, rename it to IExecutionArrayPool and we can merge this.  I have no problem making `ExecutionContext` come from a pool but it is clear that it needs a little work to implement.  Not really sure it's worth the benefit, but that's okay if you want to spend the time on it.  Just saying that this PR is already big enough and it's holding back other work by not merging it.

sungam3r:
See `protected virtual async Task<IObservable<ExecutionResult>> ResolveEventStreamAsync(ExecutionContext context, ExecutionNode node)` - `Select/SelectMany/Catch` both work **after** `ResolveEventStreamAsync` returned.

sungam3r:
>  Of course someone can save a reference to a disposed object

Shane32:
> See `protected virtual async Task<IObservable<ExecutionResult>> ResolveEventStreamAsync(ExecutionContext context, ExecutionNode node)` - `Select/SelectMany/Catch` both work **after** `ResolveEventStreamAsync` returned.

Yeah that needs to be fixed obviously -- at least, if we do pooling. If the context is disposed (which means it may be returned to a pool), it shouldn't be used again. Clearing the context helps to identify that type of coding error.

Agree that the `With` helper probably needs to be rewritten.

Shane32 (approved):

sungam3r:
> Clearing the context helps to identify that type of coding error.

I thought so too.

sungam3r:
I have resolved conflicts.

sungam3r:
Please check again `ExecutionHelper.cs`

Shane32:
ok...one second

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -367,9 +367,14 @@ public static ArgumentValue CoerceValue(ISchema schema, IGraphType type, IValue
...
-                        .Select(item => CoerceValue(schema, listItemType, item, variables).Value)
-                        .ToList(), ArgumentSource.Literal);
+                    var count = list.ValuesList.Count;
+                    if (count == 0)
+                        return new ArgumentValue(Array.Empty<object>(), ArgumentSource.Literal);
+
+                    var values = new object[count];
+                    for (int i = 0; i < count; ++i)
+                        values[i] = CoerceValue(schema, listItemType, list.ValuesList[i], variables).Value;
+                    return new ArgumentValue(values, ArgumentSource.Literal);
```
Shane32, line 377: This now returns an array instead of a `List<object>`.  I think that's fine...

sungam3r: ok

Shane32: I had written it so that `ParseDictionary` only needs to process nested sets of `Dictionary<string, object>` or `List<object>`.

PR #2184, opened 2021-01-17
Suggested changes to array pools within Memory Optimizations part 1

Shane32:
These are my suggested changes to array pools.  If merged, it will merge into your branch, @sungam3r

### DIFF
--- a/src/GraphQL/Execution/ExecutionContext.cs
+++ b/src/GraphQL/Execution/ExecutionContext.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Threading;
 using GraphQL.Instrumentation;
@@ -10,7 +11,7 @@ namespace GraphQL.Execution
     /// <summary>
     /// Provides a mutable instance of <see cref="IExecutionContext"/>.
     /// </summary>
-    public class ExecutionContext : IExecutionContext
+    public class ExecutionContext : IExecutionContext, IExecutionArrayPool, IDisposable
     {
         /// <inheritdoc/>
         public Document Document { get; set; }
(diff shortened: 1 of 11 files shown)


src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -62,20 +63,27 @@ public class ExecutionContext : IExecutionContext
...
+        /// <summary>
+        /// Releases any rented arrays back to the backing memory pool.
+        /// </summary>
+        public void Dispose()
         {
             // lock is not required because at this time work with ExecutionContext has already been completed
             foreach (var array in _trackedArrays)
                 array.Return();
+
+            _trackedArrays.Clear();
```
sungam3r, line 86: At first I also wrote that, but therefore I thought that it does not make sense, because `ExecutionContext` contains references to a huge variety of other information, see properties. Either all or nothing needs to be cleared. I decided not to clear anything.

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -62,20 +63,27 @@ public class ExecutionContext : IExecutionContext
...
             lock (_trackedArrays)
                 _trackedArrays.Add(array);
+            return array;
         }
 
-        internal void ReturnArrays()
+        /// <summary>
+        /// Releases any rented arrays back to the backing memory pool.
+        /// </summary>
+        public void Dispose()
```
sungam3r, line 80: I deliberately didn't inherit `ExecutionContext` from `IDisposable`. Do you think it's worth doing this?

sungam3r: OK. I think indeed it is worth.

sungam3r:
Already done in tune-part1.

PR #2188, opened 2021-01-17
Add ReadonlyResolveFieldContext "pool"

Shane32:
Updated PR of #1584 

This re-uses `ReadonlyResolveFieldContext` for synchronous resolvers.

### `develop` branch (before)

| Introspection |      False | 949.071 us | 3.9364 us | 3.6821 us | 87.8906 | 23.4375 |     - | 544.37 KB |
|          Hero |      False |  17.149 us | 0.0353 us | 0.0330 us |  1.4343 |       - |     - |   8.84 KB |
| Introspection |       True | 707.605 us | 6.1738 us | 5.7750 us | 82.0313 | 27.3438 |     - | 504.67 KB |
|          Hero |       True |   3.509 us | 0.0137 us | 0.0129 us |  0.6790 |  0.0038 |     - |   4.16 KB |

### `pool2` branch (after)

| Introspection |      False | 938.379 us | 3.5908 us | 3.1832 us | 79.1016 | 24.4141 |     - | 487.07 KB |
|          Hero |      False |  16.611 us | 0.0504 us | 0.0472 us |  1.4038 |       - |     - |   8.75 KB |
| Introspection |       True | 709.847 us | 2.8191 us | 2.6370 us | 72.2656 | 22.4609 |     - | 447.38 KB |
|          Hero |       True |   3.480 us | 0.0160 us | 0.0150 us |  0.6638 |  0.0038 |     - |   4.08 KB |

### `tune` branch

| Introspection |      False | 882.251 us | 4.0359 us | 3.7752 us | 81.0547 | 24.4141 |     - | 498.52 KB |
|          Hero |      False |  16.831 us | 0.0744 us | 0.0696 us |  1.4343 |       - |     - |   8.79 KB |
| Introspection |       True | 647.674 us | 2.8377 us | 2.6544 us | 74.2188 | 23.4375 |     - | 458.83 KB |
|          Hero |       True |   3.522 us | 0.0229 us | 0.0215 us |  0.6714 |  0.0038 |     - |   4.12 KB |

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -309,6 +309,8 @@ namespace GraphQL
         public static TType GetArgument<TType>(this GraphQL.IResolveFieldContext context, string name, TType defaultValue = default) { }
         public static object GetExtension(this GraphQL.IResolveFieldContext context, string path) { }
         public static bool HasArgument(this GraphQL.IResolveFieldContext context, string name) { }
+        public static void Preserve(this GraphQL.IResolveFieldContext context) { }
+        public static void Preserve<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
         public static void SetExtension(this GraphQL.IResolveFieldContext context, string path, object value) { }
     }
     public class ResolveFieldContext<TSource> : GraphQL.ResolveFieldContext, GraphQL.Execution.IProvideUserContext, GraphQL.IResolveFieldContext, GraphQL.IResolveFieldContext<TSource>
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -309,8 +309,8 @@ namespace GraphQL
         public static TType GetArgument<TType>(this GraphQL.IResolveFieldContext context, string name, TType defaultValue = default) { }
         public static object GetExtension(this GraphQL.IResolveFieldContext context, string path) { }
         public static bool HasArgument(this GraphQL.IResolveFieldContext context, string name) { }
-        public static void Copy(this GraphQL.IResolveFieldContext context) { }
-        public static void Copy<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
+        public static GraphQL.IResolveFieldContext Copy(this GraphQL.IResolveFieldContext context) { }
+        public static GraphQL.IResolveFieldContext<TSource> Copy<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
         public static void SetExtension(this GraphQL.IResolveFieldContext context, string path, object value) { }
     }
     public class ResolveFieldContext<TSource> : GraphQL.ResolveFieldContext, GraphQL.Execution.IProvideUserContext, GraphQL.IResolveFieldContext, GraphQL.IResolveFieldContext<TSource>
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -305,12 +305,12 @@ namespace GraphQL
     {
         public static GraphQL.IResolveFieldContext<TSourceType> As<TSourceType>(this GraphQL.IResolveFieldContext context) { }
         public static GraphQL.Subscription.IResolveEventStreamContext<T> As<T>(this GraphQL.Subscription.IResolveEventStreamContext context) { }
+        public static GraphQL.IResolveFieldContext Copy(this GraphQL.IResolveFieldContext context) { }
+        public static GraphQL.IResolveFieldContext<TSource> Copy<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
         public static object GetArgument(this GraphQL.IResolveFieldContext context, System.Type argumentType, string name, object defaultValue = null) { }
         public static TType GetArgument<TType>(this GraphQL.IResolveFieldContext context, string name, TType defaultValue = default) { }
         public static object GetExtension(this GraphQL.IResolveFieldContext context, string path) { }
         public static bool HasArgument(this GraphQL.IResolveFieldContext context, string name) { }
-        public static GraphQL.IResolveFieldContext Copy(this GraphQL.IResolveFieldContext context) { }
-        public static GraphQL.IResolveFieldContext<TSource> Copy<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
         public static void SetExtension(this GraphQL.IResolveFieldContext context, string path, object value) { }
     }
     public class ResolveFieldContext<TSource> : GraphQL.ResolveFieldContext, GraphQL.Execution.IProvideUserContext, GraphQL.IResolveFieldContext, GraphQL.IResolveFieldContext<TSource>
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -309,8 +309,8 @@ namespace GraphQL
         public static TType GetArgument<TType>(this GraphQL.IResolveFieldContext context, string name, TType defaultValue = default) { }
         public static object GetExtension(this GraphQL.IResolveFieldContext context, string path) { }
         public static bool HasArgument(this GraphQL.IResolveFieldContext context, string name) { }
-        public static void Preserve(this GraphQL.IResolveFieldContext context) { }
-        public static void Preserve<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
+        public static void Copy(this GraphQL.IResolveFieldContext context) { }
+        public static void Copy<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
         public static void SetExtension(this GraphQL.IResolveFieldContext context, string path, object value) { }
     }
     public class ResolveFieldContext<TSource> : GraphQL.ResolveFieldContext, GraphQL.Execution.IProvideUserContext, GraphQL.IResolveFieldContext, GraphQL.IResolveFieldContext<TSource>
--- a/src/GraphQL/Execution/ExecutionContext.cs
+++ b/src/GraphQL/Execution/ExecutionContext.cs
@@ -90,7 +90,7 @@ namespace GraphQL.Execution
             // once a field resolver finishes executing. However, a ReadonlyResolveFieldContext instance is not re-used
             // when an exception within a field resolver is thrown, and the FAQ says that calls to UnhandledExceptionDelegate
             // will be provided with a context that is not re-used. If we clear or re-use execution contexts, we should
-            // at least provide UnhandledExecptionDelegate with a copy (e.g. create one with ReadonlyResolveFieldContext
+            // at least provide UnhandledExceptionDelegate with a copy (e.g. create one with ReadonlyResolveFieldContext
             // and then Copy it) so that it is unaffected by clearing the execution context. Also note that subscription
             // execution will be affected by clearing the execution context.
 
--- a/src/GraphQL/Execution/ExecutionContext.cs
+++ b/src/GraphQL/Execution/ExecutionContext.cs
@@ -110,5 +110,12 @@ namespace GraphQL.Execution
                 _trackedArrays.Clear();
             }
         }
+
+        /// <summary>
+        /// Provides a method for an execution strategy to reuse an instance of <see cref="ReadonlyResolveFieldContext"/>.
+        /// Although in theory this field should not be accessed by multiple threads at the same time,
+        /// access is restricted to <see cref="System.Threading.Interlocked.Exchange{T}(ref T, T)"/>.
+        /// </summary>
+        internal ReadonlyResolveFieldContext ReusableReadonlyResolveFieldContext = null;
     }
 }
--- a/src/GraphQL/Execution/ExecutionContext.cs
+++ b/src/GraphQL/Execution/ExecutionContext.cs
@@ -85,6 +85,15 @@ namespace GraphQL.Execution
         /// </summary>
         protected virtual void ClearContext()
         {
+            // clearing or re-using the context will break any instances of ReadonlyResolveFieldContext from being
+            // able to access many of their properties. This is not typically a problem since the context is re-used
+            // once a field resolver finishes executing. However, a ReadonlyResolveFieldContext instance is not re-used
+            // when an exception within a field resolver is thrown, and the FAQ says that calls to UnhandledExceptionDelegate
+            // will be provided with a context that is not re-used. If we clear or re-use execution contexts, we should
+            // at least provide UnhandledExecptionDelegate with a copy (e.g. create one with ReadonlyResolveFieldContext
+            // and then Copy it) so that it is unaffected by clearing the execution context. Also note that subscription
+            // execution will be affected by clearing the execution context.
+
             //TODO:
             //Document = null;
             //Schema = null;
@@ -103,6 +112,8 @@ namespace GraphQL.Execution
             //Extensions = null;
             //RequestServices = null;
 
+            // arrays rented after the execution context has been 'disposed' will still rent just fine, but will
+            // not be returned to the pool (since Dispose has already been run) and will be garbage collected.
             lock (_trackedArrays)
             {
                 foreach (var array in _trackedArrays)
--- a/src/GraphQL/Execution/ExecutionContext.cs
+++ b/src/GraphQL/Execution/ExecutionContext.cs
@@ -113,7 +113,7 @@ namespace GraphQL.Execution
 
         /// <summary>
         /// Allows for an execution strategy to reuse an instance of <see cref="ReadonlyResolveFieldContext"/>.
-        /// Although in theory this field should not be accessed by multiple threads at the same time,
+        /// This field may be accessed by multiple threads at the same time, so
         /// access is restricted to <see cref="System.Threading.Interlocked.Exchange{T}(ref T, T)"/>.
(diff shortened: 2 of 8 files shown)


Shane32:
@sungam3r I think you should take a look here. It's based on your own PR from awhile back, #1584.  It makes a significant impact on memory without making the API any more complex.  In fact, the only API change adds a `Preserve()` extension method to "preserve" a copy an `IResolveFieldContext` class.  It's not used in the codebase, but I certainly would use it in my own codebase and it seems to be a logical extension method to provide.

The code doesn't use an "object pool", since it's not necessary.  Most resolvers run synchronously and so the "pool" only needs to contain a single object.  A unique `IResolveFieldContext` instance will be provided for any code that actually runs asynchronously.  (And, in fact, if it's only a single field, such as a "hero"-type query, the instance will still be pooled.)

Unfortunately it can't pool `ResolveFieldContextAdapater`, so `FuncFieldResolver<TSource>` instances will still cause an allocation (but only one and not two as previously).  But `NameFieldResolver` and `ExpressionFieldResolver` should be allocation-free, which are the majority of the introspection fields, and could also be the majority of user-defined fields.  And anyone that wants to implement their own `IFieldResolver` implementation can do so and take full advantage of allocation-free resolving while still having access to the full `IResolveFieldContext` interface, and without having to reference execution classes/interfaces like `ExecutionContext` or `ExecutionNode`.

The key here, as compared to #2166, is that it has much less changes to the codebase and much less API changes.  On top of that, it requires even less memory.  Unfortunately it does not execute any faster than before, whereas #2166 runs about 7% faster.

I would hope that we could take the slim-API design of this PR and see if we can implement other optimizations from #2166 so that it also runs faster.

Another idea is perhaps to take the concept of `IResolveFieldContextProvider` and see if we can tailor it to facilitate object pooling, so `ResolveFieldContextAdapter` can also provide "pooled" instances.

A separate idea is to have `ExecutionNode` implement `IResolveFieldContext<object>`.

I just think we have better options available than adding `FieldOptimized` methods.  There's already too many `Field` overloads already, in my opinion.  So let's find an optimal solution without `FieldOptimized`, and then we can decide if it makes sense to add more methods/overloads.

sungam3r:
I drew attention to this PR. I understand that we have many optimization options. Everything you wrote about above was already spinning in my head. I'll see when I have time.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -184,5 +184,17 @@ public static void SetExtension(this IResolveFieldContext context, string path,
                 }
             }
         }
+
+        /// <summary>
+        /// Preserve a copy of the specified <see cref="IResolveFieldContext"/> instance so it can be
```
sungam3r, line 189:
```suggestion
        /// Make a copy of the specified <see cref="IResolveFieldContext"/> instance so it can be
```

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -184,5 +184,17 @@ public static void SetExtension(this IResolveFieldContext context, string path,
...
+        /// accessed at a later time.
+        /// </summary>
+        public static void Preserve(this IResolveFieldContext context) => new ResolveFieldContext(context);
+
+        /// <summary>
+        /// Preserve a copy of the specified <see cref="IResolveFieldContext"/> instance so it can be
```
sungam3r, line 195:
```suggestion
        /// Make a copy of the specified <see cref="IResolveFieldContext"/> instance so it can be
```

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -184,5 +184,17 @@ public static void SetExtension(this IResolveFieldContext context, string path,
...
+        /// accessed at a later time.
+        /// </summary>
+        public static void Preserve<TSource>(this IResolveFieldContext<TSource> context) => new ResolveFieldContext<TSource>(context);
```
sungam3r, line 198:
```suggestion
        public static void Copy<TSource>(this IResolveFieldContext<TSource> context) => new ResolveFieldContext<TSource>(context);
```

sungam3r:
The main danger of this change is that when updated, the clients will not even notice if their context is reused somewhere. They will get weird floating errors at runtime.

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -110,5 +110,12 @@ protected virtual void ClearContext()
                 _trackedArrays.Clear();
             }
         }
+
+        /// <summary>
+        /// Provides a method for an execution strategy to reuse an instance of <see cref="ReadonlyResolveFieldContext"/>.
+        /// Although in theory this field should not be accessed by multiple threads at the same time,
+        /// access is restricted to <see cref="System.Threading.Interlocked.Exchange{T}(ref T, T)"/>.
+        /// </summary>
+        internal ReadonlyResolveFieldContext ReusableReadonlyResolveFieldContext = null;
```
sungam3r, line 119:
```suggestion
        internal ReadonlyResolveFieldContext ReusableReadonlyResolveFieldContext;
```

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -110,5 +110,12 @@ protected virtual void ClearContext()
                 _trackedArrays.Clear();
             }
         }
+
+        /// <summary>
+        /// Provides a method for an execution strategy to reuse an instance of <see cref="ReadonlyResolveFieldContext"/>.
```
sungam3r, line 115:
```suggestion
        /// Allows for an execution strategy to reuse an instance of <see cref="ReadonlyResolveFieldContext"/>.
```

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -110,5 +110,12 @@ protected virtual void ClearContext()
...
+        /// Although in theory this field should not be accessed by multiple threads at the same time,
```
sungam3r, line 116: Why should not ? What about `ParallelExecutionStrategy` and async resolvers?

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -236,6 +237,8 @@ protected virtual async Task ExecuteNodeAsync(ExecutionContext context, Executio
                 if (!(result is IDataLoaderResult))
                 {
                     CompleteNode(context, node);
+                    // for non-dataloader nodes that completed without throwing an error, we can re-use the context
+                    System.Threading.Interlocked.Exchange(ref context.ReusableReadonlyResolveFieldContext, resolveContext);
```
sungam3r, line 241: or

```suggestion
                    System.Threading.Interlocked.CompareExchange(ref context.ReusableReadonlyResolveFieldContext, resolveContext, null);
```
?

Although I think that the current version is also suitable, because in any case, the second instance of the context has already been created. We'll either replace it with our own or not, but one of them will go to garbage collection.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
```diff
@@ -13,6 +13,9 @@ namespace GraphQL
 {
     /// <summary>
     /// Contains parameters pertaining to the currently executing <see cref="IFieldResolver"/>.
+    /// This object is only valid during the execution of the field; it is re-used once the field
```
sungam3r, line 16: A warning about this should be written in the largest font in the readme. This is a very dangerous change.

sungam3r:
Brilliant solution! 👏 Although dangerous. 😕  I'm sure we can still apply some of the #2166  optimizations.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
```diff
@@ -13,6 +13,9 @@ namespace GraphQL
...
+    /// has resolved. Use <see cref="ResolveFieldContextExtensions.Preserve(IResolveFieldContext)"/>
```
Shane32, line 17:
```suggestion
    /// has resolved. Use <see cref="ResolveFieldContextExtensions.Copy(IResolveFieldContext)"/>
```

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -309,6 +309,8 @@ namespace GraphQL
         public static TType GetArgument<TType>(this GraphQL.IResolveFieldContext context, string name, TType defaultValue = default) { }
         public static object GetExtension(this GraphQL.IResolveFieldContext context, string path) { }
         public static bool HasArgument(this GraphQL.IResolveFieldContext context, string name) { }
+        public static void Preserve(this GraphQL.IResolveFieldContext context) { }
+        public static void Preserve<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
```
Shane32, line 313:
```suggestion
        public static void Copy(this GraphQL.IResolveFieldContext context) { }
        public static void Copy<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
```

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
Shane32: I completely agree, although I do want to point out that this concept is not foreign to ASP.Net users.  HttpContext is also pooled.

src/GraphQL/Execution/ExecutionContext.cs
Shane32: Hmm......actually, you're right.  Multiple resolvers can run in parallel, and a resolver can run in parallel with ExecutionStrategy, and the code after the await within ExecuteNodeAsync could execute at the same time as other code within ExecutionStrategy.  So it actually is necessary.

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -110,5 +110,12 @@ protected virtual void ClearContext()
...
+        /// access is restricted to <see cref="System.Threading.Interlocked.Exchange{T}(ref T, T)"/>.
```
Shane32, line 117:
```suggestion
        /// This field may be accessed by multiple threads at the same time, so
        /// access is restricted to <see cref="System.Threading.Interlocked.Exchange{T}(ref T, T)"/>.
```

src/GraphQL/Execution/ExecutionContext.cs
Shane32: Comment updated.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
Shane32: I did prevent reuse of resolvers that throw exceptions and resolvers that return data loaders.  It is very easy to write data loader code that uses the context during data loader postponed execution (and it would be normal to do so).  And for exceptions, it seems like it would be possible for error handlers to store a copy of the context in logging code or whatever.  (I know my code does.)  So for an extra allocation in case of error, it is not reused.

sungam3r: This also needs to be written in the readme.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -184,5 +184,17 @@ public static void SetExtension(this IResolveFieldContext context, string path,
                 }
             }
         }
+
+        /// <summary>
+        /// Make a copy of the specified <see cref="IResolveFieldContext"/> instance so it can be
+        /// accessed at a later time.
+        /// </summary>
+        public static void Copy(this IResolveFieldContext context) => new ResolveFieldContext(context);
```
sungam3r, line 192: Wait a second... `void`?

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -110,5 +110,12 @@ protected virtual void ClearContext()
                 _trackedArrays.Clear();
             }
         }
+
+        /// <summary>
+        /// Allows for an execution strategy to reuse an instance of <see cref="ReadonlyResolveFieldContext"/>.
+        /// This field may be accessed by multiple threads at the same time, so
+        /// access is restricted to <see cref="System.Threading.Interlocked.Exchange{T}(ref T, T)"/>.
```
sungam3r, line 117:
```suggestion
        /// access is restricted to <see cref="System.Threading.Interlocked.CompareExchange{T}(ref T, T, T)"/>.
```

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
Shane32: LOL

Shane32:
```suggestion
        public static IResolveFieldContext Copy(this IResolveFieldContext context) => new ResolveFieldContext(context);
```

src/GraphQL/Execution/ExecutionContext.cs
Shane32: Well, it's using both...

Shane32: And we can't use CompareExchange in the other place.

Shane32:
```suggestion
        /// access is restricted to <see cref="System.Threading.Interlocked.Exchange{T}(ref T, T)"/>
        /// and <see cref="System.Threading.Interlocked.CompareExchange{T}(ref T, T, T)"/>.
```

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
Shane32: Yup

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -184,5 +184,17 @@ public static void SetExtension(this IResolveFieldContext context, string path,
...
+        /// Make a copy of the specified <see cref="IResolveFieldContext"/> instance so it can be
+        /// accessed at a later time.
+        /// </summary>
+        public static IResolveFieldContext Copy(this IResolveFieldContext context) => new ResolveFieldContext(context);
+
+        /// <summary>
+        /// Make a copy of the specified <see cref="IResolveFieldContext"/> instance so it can be
+        /// accessed at a later time.
+        /// </summary>
+        public static void Copy<TSource>(this IResolveFieldContext<TSource> context) => new ResolveFieldContext<TSource>(context);
```
Shane32, line 198:
```suggestion
        public static IResolveFieldContext<TSource> Copy<TSource>(this IResolveFieldContext<TSource> context) => new ResolveFieldContext<TSource>(context);
```

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -309,6 +309,8 @@ namespace GraphQL
         public static TType GetArgument<TType>(this GraphQL.IResolveFieldContext context, string name, TType defaultValue = default) { }
         public static object GetExtension(this GraphQL.IResolveFieldContext context, string path) { }
         public static bool HasArgument(this GraphQL.IResolveFieldContext context, string name) { }
+        public static void Copy(this GraphQL.IResolveFieldContext context) { }
+        public static void Copy<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
```
Shane32, line 313:
```suggestion
        public static GraphQL.IResolveFieldContext Copy(this GraphQL.IResolveFieldContext context) { }
        public static GraphQL.IResolveFieldContext<TSource> Copy<TSource>(this GraphQL.IResolveFieldContext<TSource> context) { }
```

src/GraphQL/Execution/ExecutionContext.cs
Shane32: I had used Exchange in both places to simply the notes more than anything else.  But if you want CompareExchange where it returns a copy to the "pool", then it is necessary to use both Exchange and CompareExchange.

Shane32:
Do you want me to add migration notes so we can merge this PR, and we can make more changes in subsequent PRs?  If we are going to add more changes within this PR, then I'll wait.

src/GraphQL/Execution/ExecutionContext.cs
sungam3r: I forgot about both methods. Do as you see fit.

sungam3r:
Yes. Add migration notes and merge.

sungam3r (approved):

Shane32:
@sungam3r Do you want to review the comments I added?

src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -85,6 +85,15 @@ public void Dispose()
         /// </summary>
         protected virtual void ClearContext()
         {
+            // clearing or re-using the context will break any instances of ReadonlyResolveFieldContext from being
+            // able to access many of their properties. This is not typically a problem since the context is re-used
+            // once a field resolver finishes executing. However, a ReadonlyResolveFieldContext instance is not re-used
+            // when an exception within a field resolver is thrown, and the FAQ says that calls to UnhandledExceptionDelegate
+            // will be provided with a context that is not re-used. If we clear or re-use execution contexts, we should
+            // at least provide UnhandledExecptionDelegate with a copy (e.g. create one with ReadonlyResolveFieldContext
```
sungam3r, line 93:
```suggestion
            // at least provide UnhandledExceptionDelegate with a copy (e.g. create one with ReadonlyResolveFieldContext
```

sungam3r:
Done

PR #2191, opened 2021-01-18
[WIP] Introduce ArgumentBuilder

sungam3r:
I think the concept is clear. I was always annoyed by the cumbersomeness of setting arguments.

### DIFF
--- a/src/GraphQL.Tests/Bugs/ArrayOfArrayTest.cs
+++ b/src/GraphQL.Tests/Bugs/ArrayOfArrayTest.cs
@@ -60,7 +60,7 @@ mutation {
         {
             Field<ArrayOfArrayType>(
                 "create",
-                arguments: new QueryArguments(new QueryArgument<ArrayOfArrayInput> { Name = "input" }),
+                arguments: Arg.Next("input").Type<ArrayOfArrayInput>(),
                 resolve: ctx =>
                 {
                     var arg = ctx.GetArgument<ArrayOfArrayModel>("input");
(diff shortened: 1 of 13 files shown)


src/GraphQL.Tests/Bugs/ArrayOfArrayTest.cs
```diff
@@ -60,7 +60,7 @@ public ArrayOfArrayMutation()
         {
             Field<ArrayOfArrayType>(
                 "create",
-                arguments: new QueryArguments(new QueryArgument<ArrayOfArrayInput> { Name = "input" }),
+                arguments: Arg.Next("input").Type<ArrayOfArrayInput>(),
```
sungam3r, line 63: I have randomly added some demos throughout the repository.

src/GraphQL/Types/DirectiveGraphType.cs
```diff
@@ -142,11 +142,7 @@ public IncludeDirective()
             })
         {
             Description = "Directs the executor to include this field or fragment only when the 'if' argument is true.";
-            Arguments = new QueryArguments(new QueryArgument<NonNullGraphType<BooleanGraphType>>
-            {
-                Name = "if",
-                Description = "Included when true."
-            });
+            Arguments = Arg.Next("if").Boolean().NonNull().Description("Included when true.");
```
sungam3r, line 145: This is much easier to read than before.

Shane32:
So far I’m not a fan. I’d have to think about it and look at the code some more.  Maybe if the argument builder was part of the field builder... Also the names you picked seems confusing. Arg? Next? Boolean? If we go with a builder for arguments I think we should consider some different naming options and then pick something.

sungam3r:
One of my goals was to keep writing short.

>  Maybe if the argument builder was part of the field builder...

Yes, why not.

> Also the names you picked seems confusing. Arg? Next?

It is because I need something to start from. If I make argument builder as part of the field builder then there is no need in `Arg`.

Naming is under development.

sungam3r:
Not actual anymore.

PR #2195, opened 2021-01-18
Remove TypeExtensions.As

Shane32:


sungam3r (approved):

sungam3r (changes requested):
@Shane32  see  `internal static class SchemaExtensions.As`. I did it on purpose to fix the LGTM warnings. I just forgot to write `this` so this method was not used 🤦

Shane32:
I can't find `SchemaExtensions.As`.

Faulty LGTM warnings is not an excuse for bad code.  Writing an alias for `as` that does the same thing as `as` is bad code.

Shane32:
You can probably put some type of comment at the top of the file to have lgtm ignore that specific warning, if you're concerned about it.

Shane32:
I see you approved the changes, so I'll merge it.  I'm not sure if you wanted other changes here, but if so, we can just make them in another PR.

sungam3r:
> I can't find SchemaExtensions.As.
https://github.com/graphql-dotnet/graphql-dotnet/blob/develop/src/GraphQL/Utilities/SchemaBuilder.cs#L538

PR #2197, opened 2021-01-19
Re-use ResolveFieldContextAdapter<T> instances

Shane32:
This PR is an implementation to reuse `ResolveFieldContextAdapter<T>` instances created by `FuncFieldResolver<TSourceType, TReturnType>` whenever possible, as already described by the FAQ for `ReadonlyResolveFieldContext` instances.

In order to optimize for execution speed, upon creating a `FuncFieldResolver<,>`, a number of checks are run to determine the most optimal execution path, and then the resolver is built and stored.

The introspection benchmark makes heavy use of `NameFieldResolver` and `ExpressionFieldResolver` which is unaffected by these changes.  Even so, with the few uses of `FuncFieldResolver`, the benchmarks demonstrate a 2% decrease in memory and 2% faster execution speed.  The hero sample does not use a typed `FuncFieldResolver` so there is no change there.

For implementations which rely heavily on `FuncFieldResolver<TSourceType, TReturnType>`, there is likely to a be a much greater amount of benefit.

BEFORE

AFTER

| Introspection |      False | 946.462 us | 4.7088 us | 4.4046 us | 78.1250 | 30.2734 |     - | 478.64 KB |
|          Hero |      False |  16.927 us | 0.1174 us | 0.1098 us |  1.4038 |       - |     - |   8.73 KB |
| Introspection |       True | 722.684 us | 4.4643 us | 3.9575 us | 71.2891 | 21.4844 |     - | 438.95 KB |
|          Hero |       True |   3.583 us | 0.0140 us | 0.0124 us |  0.6599 |  0.0038 |     - |   4.05 KB |


src/GraphQL/Resolvers/FuncFieldResolver.cs
```diff
@@ -30,16 +32,44 @@ public FuncFieldResolver(Func<IResolveFieldContext, TReturnType> resolver)
...
+                    if (adapter == null)
+                    {
+                        adapter = new ResolveFieldContextAdapter<TSourceType>(context);
+                    }
+                    else
+                    {
+                        adapter.Set(context);
+                    }
+                    var ret = resolver(adapter);
+                    if (!(ret is Task task && !task.IsCompleted) && !(ret is IDataLoaderResult))
```
Shane32, line 61:
```suggestion
                    // only re-use contexts that completed synchronously
                    if (!(ret is Task task && !task.IsCompleted) && !(ret is IDataLoaderResult))
```

src/GraphQL.Benchmarks/Properties/launchSettings.json
```diff
@@ -1,8 +1,7 @@
 {
   "profiles": {
     "Profiler": {
-      "commandName": "Project",
-      "commandLineArgs": "test"
+      "commandName": "Project"
```
sungam3r, line 4:
```suggestion
      "commandName": "Project",
      "commandLineArgs": "test"
```

The point is to run the profiler, not benchmarks. If you wish then add second profile like "Bench" without args.

sungam3r:
I don't think it's worth doing that. The gain is minimal compared to the complexity of the code. Basically, we should apply optimizations from #2166 , namely, consider the possibility of abandoning `ResolveFieldContext`/`ResolveFieldContextAdapter`.

Shane32:
It's only minimal because the introspection query rarely uses FuncFieldResolver.  I think in real-world scenarios the gain will be much greater.

Unfortunately it is even more complex than the current sample.

Shane32:
I have a very large graphql application running, and there is very little, if any, use of `ExpressionFieldResolver`, and no use of `NameFieldResolver`.  Applying internal optimizations to those specific classes would be of little benefit to me.  Most of my resolvers are custom implementations of `IFieldResolver` and/or `FuncFieldResolver<TSource>`.  I also have extensive code that takes advantage of the fact that `IResolveFieldContext` is an interface and can be extended.  Completely doing away with `IResolveFieldContext` isn't an option.  Providing optimizations to `IResolveFieldContext` and `IResolveFieldContext<TSource>` can definitely improve performance for my use.  Just because the introspection/hero tests don't use `FuncFieldResolver` much doesn't mean that this isn't a very valuable change for some people like me.

Shane32:
Simplification isn't possible since the typical field builders don't strongly type `TReturnType` 😢

sungam3r:
>  Just because the introspection/hero tests don't use FuncFieldResolver much doesn't mean that this isn't a very valuable change for some people like me.

Let's add another specific bench.

sungam3r:
>  Completely doing away with IResolveFieldContext isn't an option. 

Not completely but doing something that avoids this "intermediate" object.

sungam3r:
Describe your main idea. What is the code intended to do?

Shane32:
> Describe your main idea. What is the code intended to do?

This code is intended to reuse `IResolveFieldContext<TSourceType>` instances created by `FuncFieldResolver<TSourceType, TReturnType>` whenever possible, as already described by the FAQ for `IResolveFieldContext` instances.

Shane32:
@sungam3r I suggest you review this code.  I don't see any reason not to merge it.  It clearly provides a measurable impact for the better.  Many other performance changes we have made have provided smaller benefits, and at 100 lines of code, this is not so very complicated of a change.  There are no public API changes.

sungam3r:
I would like to get rid of `ResolveFieldContextAdapter` at all and simplify code.

> For implementations which rely heavily on FuncFieldResolver<TSourceType, TReturnType>, there is likely to a be a much greater amount of benefit.

Let's prove it.

Shane32:
> I would like to get rid of `ResolveFieldContextAdapter` at all and simplify code.

When that happens, we can delete this code. Not a reason not to merge it now.

sungam3r:
I'm reviewing right now.

sungam3r:
I would not to maintain code like this. It has become more difficult to perceive with all these `if/else` branches. The winnings are minimal. Most likely, `ResolveFieldContextAdapter` will be removed soon and there will be no point in saving on it. I have not yet found the time to apply optimizations from my other branch. So... if you want, merge it. I hope that I have time to change this code before the v4 release.

Shane32:
> So... if you want, merge it. I hope that I have time to change this code before the v4 release.

I added a lot of tests to prove functionality.  I'm ready to merge this now.

sungam3r:
OK. I don't know how long this code will stay. Maybe forever. It is difficult for me to understand it and it will be difficult to maintain it.

PR #2198, opened 2021-01-20
ROM from parser

sungam3r:
Draft of `string -> ROM` migration for https://github.com/graphql-dotnet/parser/pull/101

Unfortunately, the optimizations at the parser level are leveled by the fact that the API of GraphQL.NET is built on strings and it cannot be changed without degrading the design. The first attempt led me to a dead end and I rolled back my changes. Then I added additional properties to the parser project. See https://github.com/graphql-dotnet/parser/pull/101/commits/a2367b87e5648e81564a8a6a8a4eb525e0029d54 . After that, it became possible to dock the new parser to GraphQL.NET with minor code changes. For obvious reasons, materializing ROM into a string is not what I wanted at all. But now it is a necessary evil to move on. I haven't run benchmarks yet, but I have a well-founded suspicion that the performance will be worse.


src/GraphQL/GraphQL.csproj
```diff
@@ -1,15 +1,15 @@
...
+    <TargetFrameworks>netstandard2.1;netstandard2.0</TargetFrameworks>
     <Description>GraphQL for .NET</Description>
     <Authors>Joe McBride</Authors>
     <Copyright>Copyright 2015-2017 Joseph T. McBride et al. All rights reserved.</Copyright>
     <PackageTags>GraphQL;json;api</PackageTags>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="GraphQL-Parser" Version="6.0.0" />
+    <PackageReference Include="GraphQL-Parser" Version="7.0.0-preview-8" />
```
sungam3r, line 12: @Shane32 I tested this on my local machine. I made this package for myself for debugging. What to do to pick up the real package? I understand that now it does not exist at all.

Shane32: I would just push the PR to develop (in parser). Then we can use the package from here. If we need to revert or something in the parser library, so be it.

Shane32: It won’t be on nuget but we will have access to the version within GitHub Packages

Shane32:
So far, it's about on par compared to where it used to be:


BEFORE

AFTER


| Introspection |      False | 940.107 us | 2.4035 us | 2.2482 us | 79.1016 | 25.3906 |     - | 487.95 KB |
|          Hero |      False |  16.930 us | 0.0830 us | 0.0777 us |  1.4038 |       - |     - |   8.76 KB |
| Introspection |       True | 708.697 us | 3.1031 us | 2.7508 us | 72.2656 | 22.4609 |     - | 447.35 KB |
|          Hero |       True |   3.602 us | 0.0098 us | 0.0087 us |  0.6599 |  0.0038 |     - |   4.05 KB |


And all tests pass on my machine.

src/GraphQL/GraphQL.csproj
sungam3r: OK

sungam3r:
WOW! Not so bad. Chances are, after a closer look at the code, things will get better.

src/GraphQL/GraphQL.csproj
Shane32:
```suggestion
    <PackageReference Include="GraphQL-Parser" Version="7.0.0-preview-20" />
```
So now this is all that is necessary, and the CI builds can retrieve the package.  You may need to set up your machine to pull from that source however.

Shane32:
I think that if we eliminate `ROM.ValueString`, memory requirements will decrease, since GraphQL.NET always creates a new `NameNode` to store the string representation of the name anyway -- it never gets `ValueString` more than once.  I would suggest we eliminate `ValueString` entirely and cast `ROM` to `string` when needed.

If in the future we can eliminate the `Language` namespace entirely, then we may want `ValueString`.  But in that case it is extremely likely that we will need other changes to the parser library anyway, so we can add it back in at that time.

sungam3r:
> I think that if we eliminate ROM.ValueString, memory requirements will decrease

Yes, but not much.

> it never gets ValueString more than once.

For now. I understand it all. I am afraid that somebody will need to take the string from this and other nodes somewhere else. Anything in the validation rules or whatever else we pass the document from parser to. What then? Each time there will be a cast to a string. Allocation every time. Better to let this string be stored in the node itself and cached. Also, the memory gain is not that big compared to design problems when we force clients to either work with ROM directly or cast it to a string.

Shane32:
But right now the validation rules do not have access to the parser library...

src/GraphQL/GraphQL.csproj
sungam3r: Already done. Locally I use hand-made package because I don't want to deal with GitHub Packages auth.

sungam3r:
I think I get it. We convert the document received from the parser into our own document. If we continue to do this further, then the string properties can be removed.

sungam3r:
I see a lot of usages of `TextString`/`ValueString` in `SchemaBuilder`.

Shane32:
> I see a lot of usages of `TextString`/`ValueString` in `SchemaBuilder`.

I'm unfamiliar with the `SchemaBuilder` class and do not use it in my codebase.

sungam3r:
Let's assume that there are no problems with `CoreToVanillaConverter`. What about `SchemaBuilder` and `DirectiveVisitorSelector`?

Shane32:
> Let's assume that there are no problems with `CoreToVanillaConverter`. What about `SchemaBuilder` and `DirectiveVisitorSelector`?

I can't say.  I don't use directives either.

Shane32:
Isn't `ValueString` read-only?  If so, `SchemaBuilder` can't possibly write to it, so eliminating the property (or at least eliminating the caching) shouldn't pose an "issue".  And isn't `SchemaBuilder` only used to build a schema, not to execute a query?  So any negative performance of the `SchemaBuilder` class shouldn't be the main consideration here.

Shane32:
I think we should at least give it a shot and perform some benchmarks to see the impact.

sungam3r:
I am looking at this code now and think that I can do everything well.

sungam3r:
@Shane32 I think it this can be considered as working version.

Shane32:
BEFORE


| Introspection |      False | 970.689 us | 3.0713 us | 2.8729 us | 78.1250 | 23.4375 |     - | 487.05 KB |
|          Hero |      False |  17.070 us | 0.0456 us | 0.0380 us |  1.4038 |       - |     - |   8.73 KB |
| Introspection |       True | 718.527 us | 2.2545 us | 2.1089 us | 72.2656 | 22.4609 |     - | 447.35 KB |
|          Hero |       True |   3.625 us | 0.0093 us | 0.0087 us |  0.6599 |  0.0038 |     - |   4.05 KB |


AFTER

| Introspection |      False | 969.671 us | 6.5559 us | 5.8116 us | 78.1250 | 21.4844 |     - | 487.34 KB |
|          Hero |      False |  17.242 us | 0.0561 us | 0.0525 us |  1.4038 |       - |     - |   8.73 KB |
| Introspection |       True | 741.120 us | 2.1262 us | 1.8848 us | 72.2656 | 22.4609 |     - | 447.35 KB |
|          Hero |       True |   3.635 us | 0.0110 us | 0.0103 us |  0.6599 |  0.0038 |     - |   4.05 KB |

Shane32:
I don't understand why the cached queries are any slower.  All the changes made here should not affect cached queries, since the parser does not execute for cached queries.

src/GraphQL/Shim/Bool.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            if (IsFalseStringIgnoreCase(value))
+            {
+                result = false;
+                return true;
+            }
+
+            // Special case: Trim whitespace as well as null characters.
+            value = TrimWhiteSpaceAndNull(value);
```
Shane32, line 32: Parser library cannot provide whitespace or null characters within a boolean value.

```cs
            if (token.Value == "true" || token.Value == "false")
            {
                return ParseBooleanValue(token);
            }
```

The 'true' or 'false' string also must be lower case.

Shane32 (approved):

Shane32:
Keep in mind that before we release 4.0 we will need to publish the parser library and update the nuget reference here.

sungam3r:
> I don't understand why the cached queries are any slower.

This needs to be dealt with. We have some kind of flaw.

src/GraphQL/Shim/Bool.cs
sungam3r: Done. I was going to do it a little later.

Shane32:
Technically we could reduce the condition to `span.Length == 4`, actually.  It will always be "true" or "false", and we don't even need to check what characters are in the span.

sungam3r:
Then it can be removed from the parser completely.

sungam3r:
Ah, `GraphQLScalarValue` handles any scalar, not bool.

Shane32:
> > I don't understand why the cached queries are any slower.
> 
> This needs to be dealt with. We have some kind of flaw.

Right, but I just don't understand how it's possible.  You've only changed in this PR:
* GraphQLDocumentBuilder
* CoreToVanillaConverter
* FederatedSchemaPrinter
* DirectiveVisitorSelector
* SchemaBuilder

The first two should not be used for a cached query.  The last 3 should not be used at all for introspection queries, are they?

sungam3r:
I understand. This means that as a result of the investigation, we learn something interesting.

Shane32:
> I understand. This means that as a result of the investigation, we learn something interesting.

What did you learn?  I missed something.

sungam3r:
I mean this can happen when some part of the system works completely differently than we think, and we did not know it all the time. And now we have a chance to find out about it. Well, either we just made a mistake somewhere.

sungam3r:
@Shane32 Could you please rerun benchmarks one more time to compare current results with https://github.com/graphql-dotnet/graphql-dotnet/pull/2198#issuecomment-764767843 ?

Shane32:
| Introspection |      False | 962.413 us | 4.2813 us | 4.0047 us | 78.1250 | 25.3906 |     - |  486.5 KB |
|          Hero |      False |  17.290 us | 0.0560 us | 0.0496 us |  1.4038 |       - |     - |   8.69 KB |
| Introspection |       True | 728.312 us | 2.4642 us | 2.1844 us | 72.2656 | 22.4609 |     - | 447.35 KB |
|          Hero |       True |   3.776 us | 0.0073 us | 0.0064 us |  0.6599 |  0.0038 |     - |   4.05 KB |

Shane32:
Need to bump dependency and fix all the tests due to `ignoreComments` parameter replaced, and now missing extension method also.

sungam3r:
Maybe we will provide access to `IgnoreCommentsAndLocations` "mode"?

Shane32:
Fine with me but I recommend doing so in another PR as it may require a lot of changes to the error messages. Up to you.

sungam3r:
OK. Another PR.

Shane32:
Are you trying to wrap this up or is it on hold? Just wondering @sungam3r

sungam3r:
I'm done here. OK to merge?

Shane32:
> I'm done here. OK to merge?

Good to go! 👍

PR #2199, opened 2021-01-20
[WIP] ROM from parser - phase 2

Shane32:
Continuation of #2198 

Draft to change validation and execution code to use ROM.  Initial reports are not looking good.  I have not spent enough time to figure out why.  I'm guessing the memory usage is because every reference to a ROM now takes 16 bytes instead of 8 bytes for a string.  Interestingly, with caching (and hence no validation), it still uses more memory.  It's very possible (or even likely) that I missed something and everything is getting converted to strings anyway.

BEFORE

AFTER

|        Method | UseCaching |         Mean |     Error |    StdDev |   Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |----------- |-------------:|----------:|----------:|--------:|--------:|------:|----------:|
| Introspection |      False | 1,033.944 us | 4.2911 us | 4.0139 us | 82.0313 | 25.3906 |     - | 507.89 KB |
|          Hero |      False |    17.279 us | 0.0773 us | 0.0723 us |  1.4343 |       - |     - |   8.79 KB |
| Introspection |       True |   799.612 us | 4.8677 us | 4.5532 us | 76.1719 | 24.4141 |     - | 468.61 KB |
|          Hero |       True |     3.611 us | 0.0090 us | 0.0084 us |  0.6676 |  0.0038 |     - |    4.1 KB |

### DIFF
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using GraphQL.Language.AST;
 using GraphQL.Types;
+using GraphQLParser;
 
 namespace GraphQL.Execution
 {
@@ -92,7 +93,7 @@ namespace GraphQL.Execution
 
                     if (graphType == null)
                     {
-                        var error = new InvalidVariableError(variableDef.Name, $"Variable has unknown type '{variableDef.Type.Name()}'");
+                        var error = new InvalidVariableError((string)variableDef.Name, $"Variable has unknown type '{variableDef.Type.Name()}'");
                         error.AddLocation(variableDef, document);
                         throw error;
                     }
@@ -104,7 +105,7 @@ namespace GraphQL.Execution
                     };
 
                     // attempt to retrieve the variable value from the inputs
-                    if (inputs.TryGetValue(variableDef.Name, out var variableValue))
+                    if (inputs.TryGetValue((string)variableDef.Name, out var variableValue))
                     {
                         // parse the variable via ParseValue (for scalars) and ParseDictionary (for objects) as applicable
                         variable.Value = GetVariableValue(document, graphType, variableDef, variableValue);
@@ -119,7 +120,7 @@ namespace GraphQL.Execution
                     }
                     else if (graphType is NonNullGraphType)
                     {
-                        ThrowNullError(variable.Name);
+                        ThrowNullError((string)variable.Name);
                     } 
 
                     // if the variable was not specified and no default was specified, do not set variable.Value
@@ -142,9 +143,9 @@ namespace GraphQL.Execution
             {
                 Name = variableName;
                 Index = index;
-                ChildName = null;
+                ChildName = default;
             }
-            public VariableName(VariableName variableName, string childName)
+            public VariableName(VariableName variableName, ROM childName)
             {
                 if (variableName.ChildName == null)
                 {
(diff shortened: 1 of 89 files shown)


src/GraphQL/Resolvers/NameFieldResolver.cs
```diff
@@ -25,7 +25,7 @@ public class NameFieldResolver : IFieldResolver
         public static NameFieldResolver Instance { get; } = new NameFieldResolver();
 
         /// <inheritdoc/>
-        public object Resolve(IResolveFieldContext context) => Resolve(context?.Source, context?.FieldAst?.Name);
+        public object Resolve(IResolveFieldContext context) => Resolve(context.Source, context.FieldDefinition.Name);
```
Shane32, line 28: I think this change is sound, but it does break a bunch of tests where `FieldDefinition` isn't set.  I didn't spend time to figure it out for sure.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -158,12 +159,21 @@ public VariableName(VariableName variableName, string childName)
                 }
                 ChildName = childName;
             }
-            public string Name { get; set; }
+            public ROM Name { get; set; }
             public int? Index { get; set; }
-            public string ChildName { get; set; }
-            public override string ToString() => (!Index.HasValue ? Name : Name + "[" + Index.Value + "]") + (ChildName != null ? '.' + ChildName : null);
+            public ROM ChildName { get; set; }
+            public override string ToString() => (!Index.HasValue ? (string)Name : (string)Name + "[" + Index.Value + "]") + (ChildName.Length > 0 ? '.' + (string)ChildName : null);
```
sungam3r, line 165: @Shane32 Every time you cast ROM to a string, that string is allocated on the heap. That is why I added cached string properties to the AST classes from parser project. And that's why I made minimal edits in my first commit. Your changes are much more extensive, but the main thing is that they do not optimize, but, on the contrary, degrade performance. I myself first followed this path and received dozens of modified files, although at the very beginning I began to suspect something was wrong. I rolled back my changes and started over. I suggest to rollback the last commit. I will run bencmarks on my commit. Perhaps things are not so bad and can be left as it is. Otherwise, I will need to think about how to improve the situation.

sungam3r: `ROM.ToString()`:
изображение

sungam3r: So `ROM.ToString()) does not allocate **if and only if** it points to `string` and `_index = 0 and _length = string.Length`.
изображение

Shane32: Right, I get that.  But the code you pointed to only executes when an input variable's name is nested deeper than one level deep, or an exception occurs.  Neither of which happen during an introspection query.

Shane32:
> I suggest to rollback the last commit. I will run bencmarks on my commit.

??  I ran benchmarks on your commit and left it in a comment there.  The only commit here is on top of your changes from #2198 .

This PR is just for review and discussion, and to help you, if it does so.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: If you add a breakpoint in `ROM.ToString()` and run the `validate_core_schema` test, you will find that it only executes `ROM.ToString()` once, where the operation name ("IntrospectionQuery") is cast to a string so it can be stored in `Metrics`.

sungam3r: >  But the code you pointed to only executes when an input variable's name is nested deeper than one level deep, or an exception occurs.

I wrote in the first place I saw.

sungam3r:
I didn't notice and thought it was my branch.

sungam3r:
Closing?

PR #2209, opened 2021-01-22
Add lock for multithreaded initialization

### LINKED ISSUE #2208: Schema initialisation does not appear to be thread safe
## Description

FieldMiddleware may execute multiple times for the same request unless the schema is allowed to be initialised with a single request. From digging it appears that Schema initialisation is not thread safe.

### Steps to reproduce

I have a GraphQLController which is executing my queries and mutations by creating an instance of ExecutionOptions and calling ExecuteAsync on a new instance of DocumentExecuter.

I am adding middleware that simple logs each execution and adding it via ExecutionOptions.

If I target my API with a known number of concurrent calls, I log more executions of the middleware than I do if I target the API with a single call first followed by the same number of concurrent calls.

### Expected result

For each incoming request the middleware should be executed once for each request (assuming the request only contains a single operation) and once for each field.

I would therefore expect to see the same number of middleware executions for a given number of requests, regardless of whether those requests are serialised or concurrent.

### Workaround

I am able to workaround this by using the field middleware builder to apply the middleware in the constructor of my Schema followed by a call to Initialize.

### Environment

GraphQL 3.3.1
Native Windows 10 & Dockerised Alpine Linux

Shane32:
Fixes #2208 

Adds test which proves bug exists and fails prior to this fix.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -71,8 +71,14 @@ namespace GraphQL
                 {
                     using (metrics.Subject("schema", "Initializing schema"))
                     {
-                        options.FieldMiddleware.ApplyTo(options.Schema);
-                        options.Schema.Initialize();
+                        lock (options.Schema)
+                        {
+                            if (!options.Schema.Initialized)
+                            {
+                                options.FieldMiddleware.ApplyTo(options.Schema);
+                                options.Schema.Initialize();
+                            }
+                        }
                     }
                 }
 
--- a/src/GraphQL.Tests/Execution/MultithreadedTests.cs
+++ b/src/GraphQL.Tests/Execution/MultithreadedTests.cs
@@ -1,0 +1,63 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using GraphQL.Instrumentation;
+using GraphQL.Tests.Introspection;
+using GraphQL.Tests.StarWars;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Execution
+{
+    public class MultithreadedTests
+    {
+        [Fact]
+        public async Task test()
+        {
+            // create a middleware that increments a variable when every field is resolved
+            var middleware = new FieldMiddlewareBuilder();
+            int count = 0;
+            middleware.Use((schema, d) => context => {
+                Interlocked.Increment(ref count);
+                return d(context);
+            });
+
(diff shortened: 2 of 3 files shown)


src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -71,8 +71,14 @@ public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
                 {
                     using (metrics.Subject("schema", "Initializing schema"))
                     {
-                        options.FieldMiddleware.ApplyTo(options.Schema);
-                        options.Schema.Initialize();
+                        lock (options.Schema)
```
acraven, line 74: Have you considered using double-check locking here? I have no figures to hand but as I understand it, adding another `if (!options.Schema.Initialized)` outside of the lock would be more efficient. There is a small chance of race condition so the second check is performed inside the scope of the lock (as you current have it).

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -71,8 +71,14 @@ public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
                 {
                     using (metrics.Subject("schema", "Initializing schema"))
```
sungam3r, line 72: Move this using under lock too.

acraven (commented):
Speedy response - thanks very much.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: Just look a few lines above.

acraven: Yeah, we commented at the same time

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: That would be inappropriate.  If there is lock contention, it would be because the schema is being initialized.  Without the metric, the time spent waiting for the lock would not be timed or reported in the metrics.  And if there is no lock contention then it does not matter.

sungam3r: OK. One can look at it from this side too.

Shane32: True.  But it would seem that you'd want to know why your query froze for a period of time.

sungam3r (approved):

PR #2210, opened 2021-01-23
Move FieldMiddleware from ExecutionOptions to Schema

Shane32:
Similar to how `NameConverter` and `SchemaFilter` were moved from `ExecutionOptions` to `Schema`, the same is now true for `FieldMiddleware`, since that also defines the schema configuration, rather than being an execution option.

Also simplifies middleware `Use` methods and overall design.

### DIFF
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -99,7 +99,7 @@ Also, the middleware itself should be registered in DI:
 services.AddSingleton<InstrumentFieldsMiddleware>();
 ```
 
-Alternatively, you can use an enumerable in your constructor to add all DI-registered middleware:
+Alternatively, you can use an enumerable in your constructor to add all DI-registered middlewares:
 
 ```csharp
 public MySchema : Schema
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -99,6 +99,30 @@ Also, the middleware itself should be registered in DI:
 services.AddSingleton<InstrumentFieldsMiddleware>();
 ```
 
+Alternatively, you can use an enumerable in your constructor to add all DI-registered middleware:
+
+```csharp
+public MySchema : Schema
+{
+  public MySchema(
+    IServiceProvider services,
+    MyQuery query,
+    IEnumerable<IFieldMiddleware> middlewares)
+    : base(services)
+  {
+    Query = query;
+    foreach (var middleware in middlewares)
+      FieldMiddleware.Use(middleware);
+  }
+}
+
+// within Startup.cs
+services.AddSingleton<ISchema, MySchema>();
+services.AddSingleton<IFieldMiddleware, InstrumentFieldsMiddleware>();
+services.AddSingleton<IFieldMiddleware, MyMiddleware>();
+...
+```
+
 ## Known issues
 
 Perhaps the most important thing with Field Middlewares that you should be aware of is that the
--- a/docs2/site/docs/getting-started/field-middleware.md
+++ b/docs2/site/docs/getting-started/field-middleware.md
@@ -31,20 +31,18 @@ public class InstrumentFieldsMiddleware : IFieldMiddleware
 }
 ```
 
-Then register your Field Middleware in `ExecutionOptions`.
+Then register your Field Middleware on the schema.
 
 ```csharp
-await schema.ExecuteAsync(options =>
-{
-  options.Query = "...";
-  options.FieldMiddleware.Use<InstrumentFieldsMiddleware>();
-});
+var schema = new Schema();
+schema.Query = new MyQuery();
+schema.FieldMiddleware.Use(new InstrumentFieldsMiddleware());
 ```
 
 Or, you can register a middleware delegate directly:
 
 ```csharp
-_.FieldMiddleware.Use((schema, next) =>
+schema.FieldMiddleware.Use(next =>
 {
   return context =>
   {
@@ -75,11 +73,25 @@ public delegate Task<object> FieldMiddlewareDelegate(IResolveFieldContext contex
 
 First, you are advised to read the article about [Dependency Injection](Dependency-Injection).
 
-If you use `IFieldMiddlewareBuilder.Use` overloads which accept type parameter (that is,
-those that do not accept a `IFieldMiddleware` instance or a middleware delegate) then your
-middleware creation will be delegated to DI-container. Thus, you can pass any dependencies to
+Typically you will want to set the middleware within the schema constructor.
+
+```csharp
+public MySchema : Schema
+{
+  public MySchema(
+    IServiceProvider services,
+    MyQuery query,
+    InstrumentFieldsMiddleware middleware)
+    : base(services)
+  {
+    Query = query;
+    FieldMiddleware.Use(middleware);
+  }
+}
+```
+
+Then your middleware creation will be delegated to DI-container. Thus, you can pass any dependencies to
 the Field Middleware constructor, provided that you have registered them correctly in DI.
-Note that it is required to have a schema that inherits from `Schema` or implements `IServiceProvider`.
 
 Also, the middleware itself should be registered in DI:
 
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -15,6 +15,9 @@
   These properties can be set in the constructor of the `Schema` instance, or within your DI registration delegate, or at any time before
   any query is executed. Once a query has been executed, changes to these fields is not allowed, and adding middleware via the field middleware
   builder has no effect.
+* The signature of `IFieldMiddlewareBuilder.Use` has been changed to remove the schema from delegate. Since the schema is now known, there is no
+  need for it to be passed to the middleware builder.
+* The middleware `Use<T>` extension method has been removed. Please use the `Use` method with a middleware instance instead.
 * `GraphQL.Utilities.ServiceProviderExtensions` has been made internal. This affects usages of its extension method `GetRequiredService`. Instead, reference the `Microsoft.Extensions.DependencyInjection.Abstractions` NuGet package and use extension method from `Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions` class.
 * `GraphQL.Instrumentation.StatsReport` and its associated classes have been removed. Please copy the source code into your project if you require these classes.
 * When used, Apollo tracing will now convert the starting timestamp to UTC so that `StartTime` and `EndTime` are properly serialized as UTC values.
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -11,7 +11,10 @@
 
 ## Breaking Changes
 
-* `NameConverter` and `SchemaFilter` have been removed from `ExecutionOptions` and are now properties on the `Schema`.
+* `NameConverter`, `SchemaFilter` and `FieldMiddleware` have been removed from `ExecutionOptions` and are now properties on the `Schema`.
+  These properties can be set in the constructor of the `Schema` instance, or within your DI registration delegate, or at any time before
+  any query is executed. Once a query has been executed, changes to these fields is not allowed, and adding middleware via the field middleware
+  builder has no effect.
 * `GraphQL.Utilities.ServiceProviderExtensions` has been made internal. This affects usages of its extension method `GetRequiredService`. Instead, reference the `Microsoft.Extensions.DependencyInjection.Abstractions` NuGet package and use extension method from `Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions` class.
 * `GraphQL.Instrumentation.StatsReport` and its associated classes have been removed. Please copy the source code into your project if you require these classes.
 * When used, Apollo tracing will now convert the starting timestamp to UTC so that `StartTime` and `EndTime` are properly serialized as UTC values.
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -12,7 +12,7 @@
 ## Breaking Changes
 
 * `NameConverter`, `SchemaFilter` and `FieldMiddleware` have been removed from `ExecutionOptions` and are now properties on the `Schema`.
-  These properties can be set in the constructor of the `Schema` instance, or within your DI registration delegate, or at any time before
+  These properties can be set in the constructor of the `Schema` instance, or within your DI composition root, or at any time before
   any query is executed. Once a query has been executed, changes to these fields is not allowed, and adding middleware via the field middleware
   builder has no effect.
 * The signature of `IFieldMiddlewareBuilder.Use` has been changed to remove the schema from delegate. Since the schema is now known, there is no
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -913,7 +913,7 @@ namespace GraphQL.Instrumentation
     public class FieldMiddlewareBuilder : GraphQL.Instrumentation.IFieldMiddlewareBuilder
     {
         public FieldMiddlewareBuilder() { }
-        public void ApplyTo(GraphQL.Types.ISchema schema) { }
+        public System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> Build() { }
         public GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware) { }
     }
     public static class FieldMiddlewareBuilderExtensions
@@ -931,7 +931,7 @@ namespace GraphQL.Instrumentation
     }
     public interface IFieldMiddlewareBuilder
     {
-        void ApplyTo(GraphQL.Types.ISchema schema);
+        System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> Build();
         GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware);
     }
     public class InstrumentFieldsMiddleware : GraphQL.Instrumentation.IFieldMiddleware
@@ -2135,6 +2135,8 @@ namespace GraphQL.Types
     public class SchemaTypes : System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphType>, System.Collections.IEnumerable
     {
         public int Count { get; }
+        public void ApplyMiddleware(GraphQL.Instrumentation.IFieldMiddlewareBuilder fieldMiddlewareBuilder, GraphQL.Types.ISchema schema) { }
+        public void ApplyMiddleware(System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> transform, GraphQL.Types.ISchema schema) { }
         public System.Collections.Generic.IEnumerator<GraphQL.Types.IGraphType> GetEnumerator() { }
         public static GraphQL.Types.SchemaTypes Create(System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphType> types, System.Collections.Generic.IEnumerable<GraphQL.Types.DirectiveGraphType> directives, System.Func<System.Type, GraphQL.Types.IGraphType> resolveType, GraphQL.Conversion.INameConverter nameConverter) { }
     }
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -2093,7 +2093,7 @@ namespace GraphQL.Types
         public GraphQL.Introspection.ISchemaComparer Comparer { get; set; }
         public string Description { get; set; }
         public GraphQL.Types.SchemaDirectives Directives { get; }
-        public GraphQL.Instrumentation.IFieldMiddlewareBuilder FieldMiddleware { get; set; }
+        public GraphQL.Instrumentation.IFieldMiddlewareBuilder FieldMiddleware { get; }
         public GraphQL.Introspection.ISchemaFilter Filter { get; set; }
         public bool Initialized { get; }
         public GraphQL.Types.IObjectGraphType Mutation { get; set; }
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -913,16 +913,12 @@ namespace GraphQL.Instrumentation
     public class FieldMiddlewareBuilder : GraphQL.Instrumentation.IFieldMiddlewareBuilder
     {
         public FieldMiddlewareBuilder() { }
-        public System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> Build() { }
-        public GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware) { }
+        public System.Func<GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> Build() { }
+        public GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(System.Func<GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware) { }
     }
     public static class FieldMiddlewareBuilderExtensions
     {
         public static GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(this GraphQL.Instrumentation.IFieldMiddlewareBuilder builder, GraphQL.Instrumentation.IFieldMiddleware middleware) { }
-        public static GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(this GraphQL.Instrumentation.IFieldMiddlewareBuilder builder, System.Func<GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware) { }
-        public static GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(this GraphQL.Instrumentation.IFieldMiddlewareBuilder builder, System.Type middleware) { }
-        public static GraphQL.Instrumentation.IFieldMiddlewareBuilder Use<T>(this GraphQL.Instrumentation.IFieldMiddlewareBuilder builder)
-            where T : GraphQL.Instrumentation.IFieldMiddleware { }
     }
     public delegate System.Threading.Tasks.Task<object> FieldMiddlewareDelegate(GraphQL.IResolveFieldContext context);
     public interface IFieldMiddleware
@@ -931,8 +927,8 @@ namespace GraphQL.Instrumentation
     }
     public interface IFieldMiddlewareBuilder
     {
-        System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> Build();
-        GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware);
+        System.Func<GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> Build();
+        GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(System.Func<GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware);
     }
     public class InstrumentFieldsMiddleware : GraphQL.Instrumentation.IFieldMiddleware
     {
@@ -2135,8 +2131,8 @@ namespace GraphQL.Types
     public class SchemaTypes : System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphType>, System.Collections.IEnumerable
     {
         public int Count { get; }
-        public void ApplyMiddleware(GraphQL.Instrumentation.IFieldMiddlewareBuilder fieldMiddlewareBuilder, GraphQL.Types.ISchema schema) { }
-        public void ApplyMiddleware(System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> transform, GraphQL.Types.ISchema schema) { }
+        public void ApplyMiddleware(GraphQL.Instrumentation.IFieldMiddlewareBuilder fieldMiddlewareBuilder) { }
+        public void ApplyMiddleware(System.Func<GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> transform) { }
         public System.Collections.Generic.IEnumerator<GraphQL.Types.IGraphType> GetEnumerator() { }
         public static GraphQL.Types.SchemaTypes Create(System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphType> types, System.Collections.Generic.IEnumerable<GraphQL.Types.DirectiveGraphType> directives, System.Func<System.Type, GraphQL.Types.IGraphType> resolveType, GraphQL.Conversion.INameConverter nameConverter) { }
     }
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -88,7 +88,6 @@ namespace GraphQL
         public GraphQL.Validation.Complexity.ComplexityConfiguration ComplexityConfiguration { get; set; }
         public GraphQL.Language.AST.Document Document { get; set; }
         public bool EnableMetrics { get; set; }
-        public GraphQL.Instrumentation.IFieldMiddlewareBuilder FieldMiddleware { get; set; }
         public GraphQL.Inputs Inputs { get; set; }
         public System.Collections.Generic.List<GraphQL.Execution.IDocumentExecutionListener> Listeners { get; }
         public int? MaxParallelExecutionCount { get; set; }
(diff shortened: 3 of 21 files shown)


src/GraphQL/Types/Schema.cs
```diff
@@ -61,6 +62,18 @@ public static ISchema For(string typeDefinitions, Action<SchemaBuilder> configur
...
         public INameConverter NameConverter { get; set; } = CamelCaseNameConverter.Instance;
 
+        private IFieldMiddlewareBuilder _fieldMiddlewareBuilder = new FieldMiddlewareBuilder();
+        /// <inheritdoc/>
+        public IFieldMiddlewareBuilder FieldMiddleware
+        {
+            get => _fieldMiddlewareBuilder;
+            set
+            {
+                CheckInitialized();
```
sungam3r, line 72: Let's change `CheckInitialized`:
```c#
  private void CheckInitialized([System.Runtime.CompilerServices.CallerMemberName] string name = "")
        {
            if (Initialized)
                throw new InvalidOperationException($"Schema is already initialized and sealed for modifications. You should call '{name}' only when Schema.Initialized = false.");
        }
```

src/GraphQL/Types/Schema.cs
```diff
@@ -314,6 +327,8 @@ private IEnumerable<IGraphType> GetRootTypes()
 
         private SchemaTypes CreateSchemaTypes()
         {
+            FieldMiddleware.ApplyTo(this);
```
sungam3r, line 330: Let's move `ApplyTo` into `SchemaTypes` to break recursive call to `Lazy.Value`.

sungam3r: Also no need for `ApplyTo` to be `public`.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -914,7 +913,7 @@ namespace GraphQL.Instrumentation
     public class FieldMiddlewareBuilder : GraphQL.Instrumentation.IFieldMiddlewareBuilder
     {
         public FieldMiddlewareBuilder() { }
-        public void ApplyTo(GraphQL.Types.ISchema schema) { }
+        public System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> Build() { }
         public GraphQL.Instrumentation.IFieldMiddlewareBuilder Use(System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware) { }
```
Shane32, line 917: Now `Build()` returns a combination of all the delegates passed to `Use`, or `null` if no middleware was defined.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -2134,6 +2135,8 @@ namespace GraphQL.Types
     public class SchemaTypes : System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphType>, System.Collections.IEnumerable
     {
         public int Count { get; }
+        public void ApplyMiddleware(GraphQL.Instrumentation.IFieldMiddlewareBuilder fieldMiddlewareBuilder, GraphQL.Types.ISchema schema) { }
+        public void ApplyMiddleware(System.Func<GraphQL.Types.ISchema, GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> transform, GraphQL.Types.ISchema schema) { }
```
Shane32, line 2139: I really don't like the fact that the schema has to passed into `SchemaTypes` to configure the middleware, since logically `SchemaTypes` is a 'child' class of the schema.  It seems as though `ApplyMiddleware` would ideally be an extension method.  However, as an extension method of `ISchema`, it would need access to `AllTypes`, which would pose a "deadlock" during initialization as seen before, unless we create our own lock within `Schema` and remove `Lazy`.

In the end I left it within `SchemaTypes` anyway.

Now that I think about it, this still may not work, since the middleware delegate may attempt to access `AllTypes` while configuring the resolvers.  I will write a test for this and see if it fails.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: ok

src/GraphQL/Instrumentation/FieldMiddlewareBuilderExtensions.cs
```diff
@@ -10,7 +10,7 @@ namespace GraphQL.Instrumentation
     public static class FieldMiddlewareBuilderExtensions
     {
         /// <summary>
-        /// Adds middleware to the list of delegates that will be applied to the schema when invoking <see cref="IFieldMiddlewareBuilder.ApplyTo(ISchema)"/>.
+        /// Adds middleware to the list of delegates that will be applied to the schema when invoking <see cref="SchemaTypes.ApplyMiddleware(IFieldMiddlewareBuilder, ISchema)"/>.
```
sungam3r, line 13: >  will be applied to the **schema**

I would like to change the emphasis in all comments of this kind. Let's say delegates apply to **all field resolvers, not schema**. This is more accurate and understandable.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -183,6 +185,53 @@ private void CheckSealed()
...
+            }
+        }
+
+        /// <summary>
+        /// Applies the speicifed middleware transform delegate to the schema.
+        /// <br/><br/>
+        /// When applying to the schema, modifies the resolver of each field of each graph type adding required behavior.
+        /// Therefore, as a rule, this method should be called only once during schema initialization.
+        /// </summary>
+        public void ApplyMiddleware(Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate> transform, ISchema schema)
```
sungam3r, line 211: Thus, technically we still give the simple way to apply middlewares event after creating a schema. I think it's not bad.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: > middleware delegate may attempt to access AllTypes while configuring the resolvers

No as `delegate` not executed itself.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Yeah, although you can't remove middleware, and SchemaTypes is supposed to be "sealed", and the schema is now "initialized".  I would have to say it actually should throw an exception, but I didn't add code to do that.  I'll leave it as-is.

Shane32:
@sungam3r My last commit fixes the test demonstrated in the prior commit.  Can you review it?  I'll go back over the xml comments later.  Unless you have additional comments, I think all the code is done and I just need to review and update the xml comments.

src/GraphQL/Types/Schema.cs
```diff
@@ -89,7 +90,20 @@ public void Initialize()
...
+            lock (_allTypesInitializationLock)
+            {
+                if (Initialized)
+                    return;
+
+                CreateSchemaTypes();
+
+                Initialized = true;
+
+                FindType("____");
```
Shane32, line 105: Can we remove this line?  I think it was only there to trigger `Lazy`, and is not needed now.

Shane32: All tests pass without this line of code.  I'll remove it.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: > I really don't like the fact that the schema has to passed into `SchemaTypes` to configure the middleware, since logically `SchemaTypes` is a 'child' class of the schema. 

I understand your feelings :) In fact, there is nothing wrong with that, but we can improve the situation. `schema` parameter is only used for one(ok, two) overload - ` public static IFieldMiddlewareBuilder Use<T>(this IFieldMiddlewareBuilder builder) where T : IFieldMiddleware => Use(builder, typeof(T));`/`  public static IFieldMiddlewareBuilder Use(this IFieldMiddlewareBuilder builder, System.Type middleware)` . We _can_ get rid of it now.

sungam3r: This parameter is optional, the calling code can take it from the closure.

src/GraphQL.Tests/Types/SchemaTests.cs
```diff
@@ -160,6 +160,22 @@ private void DoesNotContainTypeNames(Schema schema, params string[] typeNames)
...
+            {
+                schema.AllTypes.Count.ShouldNotBe(0);
+                return async context =>
+                {
+                    var res = await next(context);
+                    return "One " + res.ToString();
+                };
+            });
+            schema.Initialize();
+        }
```
Shane32, line 178: If `schema` was only used within the actual field resolver, it could be pulled from the execution context.  So `schema` must only be necessary to build the middleware.

If I remember correctly, I think you added `schema` to the builder primarily so that it could be used to resolve a scoped middleware instance while building a scoped schema with the `Use<MyMiddleware>` syntax.  If that's the only reason for it, we could potentially replace `schema` with `serviceProvider`.  With `Middleware` being defined on the schema, maybe there's another way to handle this scenario.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r:
```c#
 options.FieldMiddleware
                        .Use<CountFieldMiddleware>()
                        .Use<InstrumentFieldsMiddleware>();
```
->
```c#
 options.FieldMiddleware
                        .Use(someserviceProvider.GetService<CountFieldMiddleware>())
                        .Use(new InstrumentFieldsMiddleware());
```

Shane32: I don't think we should eliminate the `Use<MyMiddleware>` syntax if we can help it.

Shane32: we could write an extension method on Schema, if we want....not ideal though.....

```cs
schema.UseMiddlware<CountFieldMiddleware>();
```

which is an extension method for:

```cs
schema.FieldMiddleware.Use(schema.ServiceProvider.GetRequiredService<CountFieldMiddleware>());
```

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: We _can_ write a feature to remove/add middlewares on-fly. `ApplyMiddleware` could return `IDisposable` and "track" applied resolvers. `Dispose` will eventually set stored values back. In this way, you can dynamically add and remove some functionality to all fields. Although the same can be achieved by adding the desired middleware once, and already inside it operate with some kind of on-off condition.

src/GraphQL/Types/Schema.cs
sungam3r: Yes, it was here only to trigger Lazy.

Shane32:
We may want additional changes to middleware, but perhaps it should be in another PR. As the goal of this PR was simply to move Middleware from ExecutionOptions to Schema, I suggest we wrap up this PR with the current design. I still need to review the xml comments a little of course.

Shane32:
I don’t mind additional changes to middleware, but this PR is getting complex enough!

src/GraphQL.Tests/Types/SchemaTests.cs
sungam3r: > If I remember correctly, I think you added schema to the builder primarily so that it could be used to resolve a scoped middleware instance while building a scoped schema with the Use<MyMiddleware> syntax.

Yes, exactly. But now I saw that the schema can be useful on its own when **constructiong** delegate. 

> If that's the only reason for it, we could potentially replace schema with serviceProvider.

50/50.  On the one hand, I see cases where the caller can get a reference to schema through a closure without any problems. On the other hand, there are probably cases when it is not so easy. Now the schema for building a middleware is, in fact, an auxiliary parameter that is not required in 100% of cases.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: 50/50 Let's leave it as it is. In some next PR we will think about it better.

src/GraphQL.Tests/Types/SchemaTests.cs
Shane32: I guess the middleware can be pulled from the schema constructor for DI injection. That makes sense to me.

sungam3r:
Agree. Then write comments and I'll look it all together one more time.

src/GraphQL.Tests/Types/SchemaTests.cs
sungam3r: Do you suggest to get rid of `FieldMiddlewareBuilder` ?

sungam3r: Or let's take `IFieldMiddlewareBuilder` as schema ctor.

sungam3r: As I understand it, now there is no special sense in its existence as a property of the schema after it is initialized. It is needed at creation - exactly as a parameter of the constructor. You don't even need to store it.

sungam3r: But not. It needs to be stored until the schema is initialized, but then it is no longer needed.

Shane32: What I meant is that it is now easy to define middleware like this:

```cs
public MySchema : Schema
{
    public MySchema(IServiceProvider services, MyQuery query, MyMutation mutation, MyMiddleware middleware) : base(services)
    {
        Query = query;
        Mutation = mutation;
        Middleware.Use(middleware);
    }
}
```

Shane32:
> Agree. Then write comments and I'll look it all together one more time.

Sorry, I didn't see this, so now the changes are in this PR.  Still have not got to updating the comments yet.  Also the documentation needs updating now.

Shane32:
We could add this extension method, but I did not do so:
```cs
        public static void UseMiddleware<T>(this Schema schema) where T : IFieldMiddleware
        {
            var middleware = ((IServiceProvider)schema).GetRequiredService<T>();
            schema.FieldMiddleware.Use(middleware);
        }
```

Shane32:
We can also add this extension method for backwards compatibility, but I did not do so:

```cs
        public static IFieldMiddlewareBuilder Use(this IFieldMiddlewareBuilder builder, ISchema schema, Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware)
            => builder.Use(next => middleware(schema, next));
```

Shane32:
I think this PR is ready for review @sungam3r

src/GraphQL.Tests/Types/SchemaTests.cs
sungam3r: Ok. Nice to document.

Shane32: I already did so. See updated middleware docs

sungam3r: OK

src/GraphQL/Types/ISchema.cs
```diff
@@ -34,6 +34,13 @@ public interface ISchema : IProvideMetadata
         /// </summary>
         INameConverter NameConverter { get; }
 
+        /// <summary>
+        /// Note that field middlewares apply only to an uninitialized schema. If the schema is initialized
```
sungam3r, line 38:
```suggestion
        /// Note that field middlewares from this property apply only to an uninitialized schema. If the schema is initialized
```

src/GraphQL/Types/ISchema.cs
```diff
@@ -34,6 +34,13 @@ public interface ISchema : IProvideMetadata
...
+        /// then adding additional middleware through the builder does nothing. The schema is initialized (if not yet)
+        /// at the beginning of the first call to <see cref="DocumentExecuter"/>.<see cref="DocumentExecuter.ExecuteAsync(ExecutionOptions)">ExecuteAsync</see>.
```
sungam3r, line 40:
```suggestion
        /// at the beginning of the first call to <see cref="DocumentExecuter"/>.<see cref="DocumentExecuter.ExecuteAsync(ExecutionOptions)">ExecuteAsync</see>.
        /// However, you can also apply middlewares at any time in runtime using SchemaTypes.ApplyMiddleware method.
```

Also let's make `ApplyMiddleware` method syncronized - lock or `[MethodImpl(MethodImplOptions.Synchronized)]`.

docs2/site/docs/getting-started/field-middleware.md
```diff
@@ -75,11 +73,25 @@ public delegate Task<object> FieldMiddlewareDelegate(IResolveFieldContext contex
...
+    MyQuery query,
+    InstrumentFieldsMiddleware middleware)
+    : base(services)
+  {
+    Query = query;
+    FieldMiddleware.Use(middleware);
+  }
+}
+```
+
```
sungam3r, line 92: Let's add one more example with `IEnumerable`:
```csharp
public MySchema : Schema
{
  public MySchema(
    IServiceProvider services,
    MyQuery query,
    IEnumerable<IFieldMiddleware> middlewares)
    : base(services)
  {
    Query = query;
    foreach (var middleware in middlewares)
        FieldMiddleware.Use(middleware);
  }
}
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -11,7 +11,13 @@
 
 ## Breaking Changes
 
-* `NameConverter` and `SchemaFilter` have been removed from `ExecutionOptions` and are now properties on the `Schema`.
+* `NameConverter`, `SchemaFilter` and `FieldMiddleware` have been removed from `ExecutionOptions` and are now properties on the `Schema`.
+  These properties can be set in the constructor of the `Schema` instance, or within your DI registration delegate, or at any time before
```
sungam3r, line 15: What is "DI registration delegate" ?

src/GraphQL/Instrumentation/IFieldMiddlewareBuilder.cs
```diff
@@ -9,21 +9,18 @@ namespace GraphQL.Instrumentation
...
-        IFieldMiddlewareBuilder Use(Func<ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware);
+        IFieldMiddlewareBuilder Use(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware);
 
         /// <summary>
-        /// Applies all delegates specified by the <see cref="Use(Func{ISchema, FieldMiddlewareDelegate, FieldMiddlewareDelegate})"/> method to the schema.
-        /// <br/><br/>
-        /// When applying to the schema, modifies the resolver of each field of each graph type adding required behavior.
-        /// Therefore, as a rule, this method should be called only once during schema initialization. See <see cref="DocumentExecuter.ExecuteAsync(ExecutionOptions)"/>.
+        /// Returns a transform for field resolvers, or <see langword="null"/> if no middleware is defined.
+        /// The transform is a cumulation of all middleware configured within this builder.
```
sungam3r, line 22:
```suggestion
        /// The transform is a cumulation of all middlewares configured within this builder.
```

src/GraphQL/Instrumentation/IFieldMiddlewareBuilder.cs
```diff
@@ -9,21 +9,18 @@ namespace GraphQL.Instrumentation
     public interface IFieldMiddlewareBuilder
     {
         /// <summary>
-        /// Adds the specified delegate to the list of delegates that will be applied to the schema when invoking <see cref="ApplyTo(ISchema)"/>.
+        /// Adds the specified delegate to the list of delegates that will be applied to all field resolvers when invoking <see cref="SchemaTypes.ApplyMiddleware(IFieldMiddlewareBuilder)"/>.
         /// <br/><br/>
         /// The delegate is used to unify the different ways of specifying middleware. See additional methods in <see cref="FieldMiddlewareBuilderExtensions"/>.
```
sungam3r, line 14: Only one method left :)

src/GraphQL.Harness/Startup.cs
```diff
@@ -49,7 +49,18 @@ public void ConfigureServices(IServiceCollection services)
...
+                var settings = services.GetRequiredService<IOptions<GraphQLSettings>>();
+                var schema = new StarWarsSchema(services);
+                if (settings.Value.EnableMetrics)
+                {
+                    schema.FieldMiddleware
+                        .Use(services.GetRequiredService<CountFieldMiddleware>())
+                        .Use(services.GetRequiredService<InstrumentFieldsMiddleware>());
+                }
+                return schema;
+            });
```
sungam3r, line 63: If the schema constructor accepts `IEnumerable<IFieldMiddleware>`, then this example can be simplified.

sungam3r:
```
 // add Field Middlewares
            services.AddSingleton<CountFieldMiddleware>();
            services.AddSingleton<InstrumentFieldsMiddleware>();
```
->
```
 // add Field Middlewares
            services.TryAddEnumerable(ServiceDescriptor.Singleton<IFieldMiddleware, CountFieldsMiddleware>());
            services.TryAddEnumerable(ServiceDescriptor.Singleton<IFieldMiddleware, InstrumentFieldsMiddleware>());
```

sungam3r: Well, you still need to pull `IOptions` from somewhere (unless you replace them with a primitive flag), but you can't do this in the `Configure` method itself, so I don't think thechanged example will be easier.

src/GraphQL.Tests/Execution/MultithreadedTests.cs
```diff
@@ -17,7 +17,7 @@ public async Task test()
             // create a middleware that increments a variable when every field is resolved
             var middleware = new FieldMiddlewareBuilder();
             int count = 0;
-            middleware.Use((schema, d) => context => {
+            middleware.Use((d) => context => {
```
sungam3r, line 20:
```suggestion
            middleware.Use(d => context => {
```

src/GraphQL.Tests/Types/SchemaTests.cs
```diff
@@ -160,6 +160,22 @@ private void DoesNotContainTypeNames(Schema schema, params string[] typeNames)
                 type.ShouldBe(null, $"Found {typeName} in type lookup.");
             });
         }
+
+        [Fact]
+        public void middleware_can_reference_SchemaTypes()
+        {
+            var schema = new Schema() { Query = new SomeQuery() };
```
sungam3r, line 167:
```suggestion
            var schema = new Schema { Query = new SomeQuery() };
```

src/GraphQL.Tests/Types/SchemaTests.cs
```diff
@@ -160,6 +160,22 @@ private void DoesNotContainTypeNames(Schema schema, params string[] typeNames)
...
+            schema.FieldMiddleware.Use((next) =>
```
sungam3r, line 168:
```suggestion
            schema.FieldMiddleware.Use(next =>
```

src/GraphQL/Types/Schema.cs
```diff
@@ -61,22 +63,45 @@ public static ISchema For(string typeDefinitions, Action<SchemaBuilder> configur
...
+            get => _fieldMiddlewareBuilder;
+            set
+            {
+                CheckInitialized();
+                _fieldMiddlewareBuilder = value ?? throw new ArgumentNullException();
+            }
+        }
+
+        /// <inheritdoc/>
+        public bool Initialized { get; private set; } = false;
```
sungam3r, line 79:
```suggestion
        public bool Initialized { get; private set; }
```

src/GraphQL/Types/Schema.cs
```diff
@@ -312,17 +345,21 @@ private IEnumerable<IGraphType> GetRootTypes()
...
 
-            return SchemaTypes.Create(
+            _allTypes = SchemaTypes.Create(
                 types,
                 Directives,
                 type => (IGraphType)_services.GetRequiredService(type),
                 NameConverter);
+
+            //at this point, Initialized will return false, and Initialize will still lock while waiting for initialization to complete
+            //however, AllTypes and similar properties will return a reference to SchemaTypes without waiting for a lock
```
sungam3r, line 361:
```suggestion
            // At this point, Initialized will return false, and Initialize will still lock while waiting for initialization to complete.
            // However, AllTypes and similar properties will return a reference to SchemaTypes without waiting for a lock.
```

src/GraphQL/Types/Schema.cs
```diff
@@ -251,7 +286,7 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                if (_allTypes != null)
+                if (_disposed)
```
sungam3r, line 289:
```suggestion
                if (!_disposed)
```
😮

src/GraphQL/Types/Schema.cs
```diff
@@ -122,19 +147,29 @@ public void Initialize()
         public SchemaDirectives Directives { get; }
 
         /// <inheritdoc/>
-        public SchemaTypes AllTypes => _allTypes?.Value;
+        public SchemaTypes AllTypes {
```
sungam3r, line 150:
```suggestion
        public SchemaTypes AllTypes
        {
```

src/GraphQL/Types/Schema.cs
```diff
@@ -122,19 +147,29 @@ public void Initialize()
...
+            get
+            {
+                if (_allTypes != null)
+                    return _allTypes;
+
+                Initialize();
+
+                return _allTypes;
```
sungam3r, line 158:
```suggestion
                if (_allTypes == null)
                    Initialize();

                return _allTypes;
```

src/GraphQL/Types/Schema.cs
```diff
@@ -61,22 +63,45 @@ public static ISchema For(string typeDefinitions, Action<SchemaBuilder> configur
         /// <inheritdoc/>
         public INameConverter NameConverter { get; set; } = CamelCaseNameConverter.Instance;
 
+        private IFieldMiddlewareBuilder _fieldMiddlewareBuilder = new FieldMiddlewareBuilder();
         /// <inheritdoc/>
-        public bool Initialized => _allTypes?.IsValueCreated == true;
+        public IFieldMiddlewareBuilder FieldMiddleware
+        {
+            get => _fieldMiddlewareBuilder;
+            set
```
sungam3r, line 71: Do we really need a setter?

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -183,6 +185,50 @@ private void CheckSealed()
             return lookup;
         }
 
+        /// <summary>
+        /// Applies all delegates specified by the middleware builder to the schema.
+        /// <br/><br/>
+        /// When applying to the schema, modifies the resolver of each field of each graph type adding required behavior.
+        /// Therefore, as a rule, this method should be called only once during schema initialization.
```
sungam3r, line 192:
```suggestion
        /// Therefore, as a rule, this method should be called only once - during schema initialization.
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -183,6 +185,50 @@ private void CheckSealed()
...
+
+            // allocation free optimization if no middlewares are defined
+            if (transform != null)
+            {
+                ApplyMiddleware(transform);
+            }
+        }
+
+        /// <summary>
+        /// Applies the speicifed middleware transform delegate to the schema.
```
sungam3r, line 206:
```suggestion
        /// Applies the specified middleware transform delegate to the schema.
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -183,6 +185,50 @@ private void CheckSealed()
...
+        /// <br/><br/>
+        /// When applying to the schema, modifies the resolver of each field of each graph type adding required behavior.
+        /// Therefore, as a rule, this method should be called only once during schema initialization.
```
sungam3r, line 209:
```suggestion
        /// Therefore, as a rule, this method should be called only once - during schema initialization.
```

src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
```diff
@@ -160,6 +193,16 @@ public Task<object> Resolve(IResolveFieldContext context, FieldMiddlewareDelegat
...
     internal static class TestExtensions
     {
-        public static FieldMiddlewareDelegate Build(this FieldMiddlewareBuilder builder) => builder.Build(null, null);
+        public static FieldMiddlewareDelegate BuildResolve(this FieldMiddlewareBuilder builder)
+        {
+            var transform = builder.Build();
+            return transform != null ? transform(null) : null;
+        }
+
+        public static FieldMiddlewareDelegate BuildResolve(this FieldMiddlewareBuilder builder, FieldMiddlewareDelegate start)
```
sungam3r, line 202: Is this method used?

src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
```diff
@@ -12,77 +9,36 @@ namespace GraphQL.Instrumentation
...
-                    _singleMiddleware,
-                    middleware
-                };
+                _middleware = middleware;
             }
             else
             {
-                _middlewares.Add(middleware);
+                var firstMiddleware = _middleware;
+                _middleware = (next) => firstMiddleware(middleware(next));
```
sungam3r, line 27:
```suggestion
                _middleware = next => firstMiddleware(middleware(next));
```

Thanks for rewriting this builder. For some reason, it did not occur to me to combine middleware immediately after adding them. So I had to iterate over them in reverse order.

src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
```diff
@@ -12,77 +9,36 @@ namespace GraphQL.Instrumentation
...
-            else if (_singleMiddleware != null)
-            {
-                middlewareDelegate = _singleMiddleware(schema, middlewareDelegate);
-            }
-
-            return middlewareDelegate;
-        }
-
-        private bool Empty => _singleMiddleware == null;
+        private readonly FieldMiddlewareDelegate _defaultDelegate = context => Task.FromResult(NameFieldResolver.Instance.Resolve(context));
```
sungam3r, line 33:
```suggestion
        private static readonly FieldMiddlewareDelegate _defaultDelegate = context => Task.FromResult(NameFieldResolver.Instance.Resolve(context));
```

src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
```diff
@@ -12,77 +9,36 @@ namespace GraphQL.Instrumentation
...
-                            var fieldMiddlewareDelegate = Build(context => inner.ResolveAsync(context), schema);
+            if (_middleware == null)
+                return null;
 
-                            field.Resolver = new FuncFieldResolver<object>(fieldMiddlewareDelegate.Invoke);
-                        }
-                    }
-                }
-            }
+            return (start) => _middleware(start ?? _defaultDelegate);
```
sungam3r, line 41:
```suggestion
            return start => _middleware(start ?? _defaultDelegate);
```

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: Removed. OK.

sungam3r:
Regarding https://github.com/graphql-dotnet/graphql-dotnet/pull/2210#issuecomment-766243357 : not worth it

sungam3r:
I'm done here.

docs2/site/docs/migrations/migration4.md
Shane32:
```cs
Services.AddSingleton<ISchema>(
    //this delegate -- I don't know what to call it
    serviceProvider => {
        var schema = new Schema();
        schema.Query = serviceProvider.GetRequiredService<MyQuery>();
        schema.FieldMiddleware.Use(serviceProvider.GetRequiredService<MyMiddleware>());
    }
);
```
I'm open to any suggestions you have as to the documentation here.

sungam3r: I used to call it _composition root_: https://blog.ploeh.dk/2011/07/28/CompositionRoot/

src/GraphQL/Types/Schema.cs
Shane32: That's why we do reviews!!! :+1:

src/GraphQL.Harness/Startup.cs
Shane32: I don't think we should have the `Schema` class's constructor accept `IEnumerable<IFieldMiddleware>`.  If we add that, what about `INameConverter`, `ISchemaFilter`, and `ISchemaComparer`?  Not to mention `Query`, `Mutation`, and `Subscription`.  Better just let the user inherit from `Schema` and add whatever they want to their constructor, as we can demonstrate in the documentation.

sungam3r: I meant to add this ctor to the example, not to the main project.

src/GraphQL/Types/ISchema.cs
Shane32: I don't see a reason for that, unless we lock executions from happening simultaneously via `ReaderWriterLockSlim`, which I wouldn't do.  If someone is adding middleware at runtime, it's their responsibility to be sure that there are no active executions or other threads also adding middleware.

sungam3r:
3.30 am here, I'm done for today.

Shane32:
> 3.30 am here, I'm done for today.

No problem.  Too bad it needs more work; I was hoping we could get this merged today.  Oh well.

src/GraphQL/Types/Schema.cs
Shane32: No; I'll remove it.

src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
Shane32: No; I'll delete it.

src/GraphQL.Harness/Startup.cs
Shane32: I see what you mean.  If we move it into the ctor then we need `GraphQLSettings` from the `Harness` project moved into the `StarWars` project and add a reference to the MS options package.  So I'm leaving it for now... but whatever you want is fine...

docs2/site/docs/migrations/migration4.md
Shane32: Uhh...ok so this???
```suggestion
  These properties can be set in the constructor of the `Schema` instance, or within your DI composition root, or at any time before
```

docs2/site/docs/getting-started/field-middleware.md
Shane32: Done; please review.

src/GraphQL/Types/ISchema.cs
sungam3r: Agree. Yes, I thought about it yesterday too. Middleware is installed at the same time as the executing request, so the result is still undefined. It was very late and in a lot of reviews I just forgot to write about it.

docs2/site/docs/migrations/migration4.md
sungam3r: Yes.

src/GraphQL.Harness/Startup.cs
sungam3r: Let's keep it as is. Perhaps later we publish more easier axample.

Shane32:
> internal for tests?

The setter on the one property on Schema? Yes there was one test somewhere that needed it. I didn’t work to find an alternative. Like I could have made it virtual and mocked it or something ....

docs2/site/docs/getting-started/field-middleware.md
```diff
@@ -75,18 +73,56 @@ public delegate Task<object> FieldMiddlewareDelegate(IResolveFieldContext contex
...
 the Field Middleware constructor, provided that you have registered them correctly in DI.
-Note that it is required to have a schema that inherits from `Schema` or implements `IServiceProvider`.
 
 Also, the middleware itself should be registered in DI:
 
 ```csharp
 services.AddSingleton<InstrumentFieldsMiddleware>();
 ```
 
+Alternatively, you can use an enumerable in your constructor to add all DI-registered middleware:
```
sungam3r, line 102:
```suggestion
Alternatively, you can use an enumerable in your constructor to add all DI-registered middlewares:
```

src/GraphQL.Tests/Types/SchemaTests.cs
```diff
@@ -160,6 +162,50 @@ private void DoesNotContainTypeNames(Schema schema, params string[] typeNames)
...
+                };
+            });
+            schema.Initialize();
+        }
+
+        [Fact]
+        public void disposed_schema_throws_errors()
+        {
+            var schema = new Schema();
+            schema.Dispose();
```
sungam3r, line 186:
```suggestion
            schema.Dispose();
            schema.Dispose();
```

Just to test idempotency.

sungam3r (approved):

PR #2215, opened 2021-01-24
Make ResolveFieldContext implement IResolveFieldContext<object>

Shane32:
This allows `As` to pass through the context instead of creating a new instance.  Probably not really used much since we typically use `ReadonlyResolveFieldContext` and stuff.


sungam3r:
It seems to be nothing wrong.

> This allows As to pass through the context instead of creating a new instance.

What do the benchmarks show?

Shane32:
There is no point to running benchmarks. It will not test this change.

sungam3r (approved):

PR #2221, opened 2021-01-26
Reduce memory usage by making shared instances static. 

SebastianStehle:
Useful when you have multiple graphql schemas (I have hundreds).

### DIFF
--- a/src/GraphQL.Tests/Execution/MultithreadedTests.cs
+++ b/src/GraphQL.Tests/Execution/MultithreadedTests.cs
@@ -1,0 +1,63 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using GraphQL.Instrumentation;
+using GraphQL.Tests.Introspection;
+using GraphQL.Tests.StarWars;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Execution
+{
+    public class MultithreadedTests
+    {
+        [Fact]
+        public async Task test()
+        {
+            // create a middleware that increments a variable when every field is resolved
+            var middleware = new FieldMiddlewareBuilder();
+            int count = 0;
+            middleware.Use((schema, d) => context => {
+                Interlocked.Increment(ref count);
+                return d(context);
+            });
+
+            // prep a test execution using that middleware
+            StarWarsTestBase starWarsTest = null;
+            Func<Task> testExecution = async () =>
+            {
+                var result = await starWarsTest.Executer.ExecuteAsync(new ExecutionOptions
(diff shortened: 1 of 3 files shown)


sungam3r:
These fields are not `static` intentionally. We intend to make it possible to replace built-in and "system" (provided by GraphQL.NET) scalars. Static fields will not allow this for individual schemas. @Shane32 ?

Shane32:
Correct.  Those instance are mutable, which causes a problem if there were two schemas configured on a single app.  What happens is that when the schema is initialized, the `ResolvedType` properties are set on the graph types, which will overwrite the values in the static instance.  Other issues are also possible, such as field middleware applying to a static instance repeatedly upon each initialization of a schema.

Shane32:
As it applies here, although the list is readonly, the instances contained therein must not be shared between schemas.  So the list cannot be static.

SebastianStehle:
Okay, this is interesting, you can close it now. What can I not share? I am creating graphql types on the fly for my headless CMS and I issues with memory usage and graphql is a big part of the allocations. Therefore I try to reduce as many instances as possible. 

I resolve all types manually, but now I am worried about the middleware thing you mentioned.

Shane32:
The scalar graph types are immutable classes, I believe.  They could be overridden, but the default implementation is immutable.  This presents an opportunity for using static instances, and in this case, perhaps `_builtInScalars` could also be static.  I'd have to check further.

sungam3r:
@SebastianStehle We are preparing v4, in which the memory consumption has decreased quite significantly. Can you provide some working sample code to demonstrate your memory consumption scenario?

Shane32:
> @SebastianStehle We are preparing v4, in which the memory consumption has decreased quite significantly. Can you provide some working sample code to demonstrate your memory consumption scenario?

@sungam3r FYI, we can now benchmark schema build time and memory consumption, which is what I believe @SebastianStehle is trying to help with.  Although many objects are unused after the schema build process (and garbage collection), this PR would reduce the static memory use even after that is complete.  It would be great if we could benchmark the amount of memory in use after garbage collection is complete, but I'm not sure how to do that with the benchmarks.  I could add a console app that uses `GC.Collect` and `GC.GetTotalMemory` if that is what you are looking for, but it may not provide reliable values.  What kind of sample did you have in mind?

SebastianStehle:
Btw: One of the main sources of memory usage (but not allocations of course) is the Metadata Provider. Even for small schemas it takes 1-2 MB for all the ConcurrentDictionary instances. I am cleaning all metadata now to get the memory usage down.

PR #2222, opened 2021-01-26
Lazy initialization of Metadata

SebastianStehle:
The metadata is responsible for a large amount of the memory usage.

image

I have made an attempt to set all unused metadata to null and use normal dictionaries and the result is like this:

image

I don't want to change the current behavior and semantics, therefore I just initialize the metadata when it is needed the first time. The setter is not used anyway, so I have removed it.

### DIFF
--- a/src/GraphQL/Utilities/MetadataProvider.cs
+++ b/src/GraphQL/Utilities/MetadataProvider.cs
@@ -11,24 +11,44 @@ namespace GraphQL.Utilities
     /// </summary>
     public class MetadataProvider : IProvideMetadata
     {
+        private readonly object _metadataLock = new object();
+        private IDictionary<string, object> _metadata;
+
         /// <inheritdoc />
-        public IDictionary<string, object> Metadata { get; set; } = new ConcurrentDictionary<string, object>();
+        public IDictionary<string, object> Metadata
+        {
+            get
+            {
+                if (_metadata == null)
+                {
+                    lock (_metadataLock)
+                    {
+                        if (_metadata == null)
+                        {
+                            _metadata = new ConcurrentDictionary<string, object>();
+                        }
+                    }
+                }
+
+                return _metadata;
+            }
+        }
 
         /// <inheritdoc />
         public TType GetMetadata<TType>(string key, TType defaultValue = default)
         {
-            var local = Metadata;
+            var local = _metadata;
             return local != null && local.TryGetValue(key, out object item) ? (TType)item : defaultValue;
         }
 
         /// <inheritdoc />
         public TType GetMetadata<TType>(string key, Func<TType> defaultValueFactory)
         {
-            var local = Metadata;
+            var local = _metadata;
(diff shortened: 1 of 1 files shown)


sungam3r:
Yeah. I already paid attention to `ConcurrentDictionary` from metadata today.

sungam3r:
Please retarget PR to `develop`.

sungam3r:
@Shane32 I just a few hours ago wanted to ask a question about whether we need `IProvideMetadata.Metadata` as `ConcurrentDictionary`? It is not "request related" storage. All other types and schema collections are not thread safe. We initialize schema once and then it just works. Of course, one **can** modify many parts of schema in runtime but then he has to provide synchronization.

sungam3r:
So I suggest making metadata as `Dictionary`, not `ConcurrentDictionary` + lock on `this` instead of new field.

SebastianStehle:
I followed this guideline: https://docs.microsoft.com/de-de/dotnet/csharp/language-reference/keywords/lock-statement#guidelines

I did not want to introduce a breaking change, therefore I kept the ConcurrentDictionary. I think only a few instances will actually have metadata, especially because the schemas are so deeply nested in many cases. So it is acceptable for these cases.

sungam3r:
You did introduce breaking changes. You remove piblic setter.

sungam3r:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/1451/commits/7af9c1d9643a68983da0354765ad861d1f4ab11b

sungam3r:
Locking on `this` in our case is not something dangerous.

Shane32:
@sungam3r Do we need to use `ConcurrentDictionary`?  Are we supporting writing to metadata during runtime?  It would seem that a standard `Dictionary` would be fine, assuming that it is only written to during schema initialization, as `Dictionary` fully supports multiple readers so long as there are no writers.  We could, by the way, have a static global property to enable `ConcurrentDictionary`, but use `Dictionary` by default.

Shane32:
I would agree that in this instance it would be better to lock on `this`.  We are trying to eliminate allocations and there shouldn't be user code locking on the graph that would interfere.

Shane32:
I think this PR is looking good!

> @Shane32 I just a few hours ago wanted to ask a question about whether we need `IProvideMetadata.Metadata` as `ConcurrentDictionary`? It is not "request related" storage. All other types and schema collections are not thread safe. We initialize schema once and then it just works. Of course, one **can** modify many parts of schema in runtime but then he has to provide synchronization.

🤦‍♂️ you said the same thing lol!  yes let's use `Dictionary` instead, or provide global option if we want to allow concurrent dictionaries

sungam3r:
Done in #1451

Shane32:
Thanks for the suggestion @SebastianStehle !  It's been merged as you initially suggested in #2223 with no locks and lazy-loading regular dictionaries.  Also by disabling the automatic scanning for xml comments in #2224 we have drastically reduced load time and memory requirements during the schema build process (thanks @sungam3r !).  Hopefully it will make a significant difference in your application.

PR #2225, opened 2021-01-27
Prevent API tests from running twice

Shane32:
Any particular reason the API tests run for both net5 and net3.1 ?


sungam3r:
Of course not, it should have been done for a long time.

sungam3r (approved):

PR #2229, opened 2021-01-27
Eliminate ExpressionFieldResolver

sungam3r:
This is done following an investigation for #1451.

### DIFF
--- a/src/GraphQL/Introspection/__Directive.cs
+++ b/src/GraphQL/Introspection/__Directive.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Linq;
 using GraphQL.Types;
 
@@ -18,13 +17,16 @@ namespace GraphQL.Introspection
             Description =
                 "A Directive provides a way to describe alternate runtime execution and " +
                 "type validation behavior in a GraphQL document." +
-                $"{Environment.NewLine}{Environment.NewLine}In some cases, you need to provide options to alter GraphQL's " +
+               @"
+
+" +
+                "In some cases, you need to provide options to alter GraphQL's " +
                 "execution behavior in ways field arguments will not suffice, such as " +
                 "conditionally including or skipping a field. Directives provide this by " +
                 "describing additional information to the executor.";
 
-            Field(f => f.Name);
-            Field(f => f.Description, nullable: true);
+            Field<NonNullGraphType<StringGraphType>>("name", resolve: context => context.Source.Name);
+            Field<StringGraphType>("description", resolve: context => context.Source.Description);
 
             Field<NonNullGraphType<ListGraphType<NonNullGraphType<__DirectiveLocation>>>>("locations");
             Field<NonNullGraphType<ListGraphType<NonNullGraphType<__InputValue>>>>("args",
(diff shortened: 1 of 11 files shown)


src/GraphQL/Introspection/__Directive.cs
```diff
@@ -18,13 +17,16 @@ public __Directive()
             Description =
                 "A Directive provides a way to describe alternate runtime execution and " +
                 "type validation behavior in a GraphQL document." +
-                $"{Environment.NewLine}{Environment.NewLine}In some cases, you need to provide options to alter GraphQL's " +
```
sungam3r, line 21: String allocation in each ctor call. Now it is constant.

src/GraphQL/Introspection/__Type.cs
```diff
@@ -19,14 +19,17 @@ public __Type()
             Description =
                 "The fundamental unit of any GraphQL Schema is the type. There are " +
                 "many kinds of types in GraphQL as represented by the `__TypeKind` enum." +
-                $"{Environment.NewLine}{Environment.NewLine}Depending on the kind of a type, certain fields describe " +
+               @"
```
sungam3r, line 22: String allocation in each ctor call. Now it is constant.

src/GraphQL/Resolvers/ExpressionFieldResolver.cs
```diff
@@ -5,6 +5,8 @@ namespace GraphQL.Resolvers
 {
     /// <summary>
     /// Returns a value from the field's graph type's source object, based on a predefined expression.
+    /// Note: this class uses dynamic compilation and therefore allocates a relatively large amount of
+    /// memory in managed heap, ~1KB. Do not use this class in cases with limited memory requirements.
```
sungam3r, line 9: I will also need to write similar comments to the methods that create this resolver.

Shane32 (approved):
As long as we are not actually removing `Field(x => x.Name)` syntax or `ExpressionFieldResolver`, I'm good with these changes.  We could investigate https://github.com/dadhi/FastExpressionCompiler but I probably would not at this time.

Shane32:
> @Shane32 I don't have time to make tests. I found another place that allocates huge amounts of memory - **AstPrinter**.

No problem.  AstPrinter isn't used during build is it?

Shane32:
|        Method | Stage |     Mean |     Error |    StdDev |   Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |------ |---------:|----------:|----------:|--------:|--------:|------:|----------:|
| Before        | Build | 1.390 ms | 0.0085 ms | 0.0080 ms | 27.3438 | 13.6719 |     - | 172.19 KB |
| After         | Build | 286.6 us |   1.29 us |   1.21 us | 15.1367 |  1.9531 |     - |  94.63 KB |

Shane32:
With these changes, there is much more use of `FuncFieldResolver<TSource, TReturn>` (since almost all the `ExpressionFieldResolver` uses are now `FuncFieldResolver<TSource, object>` uses), and so the improvements merged in #2197 will be much more valuable than previously.

sungam3r:
Yes. And regarding fastexpressioncompiler. One can use any resolver he likes and we provide a couple of "built-in" resolvers. Expression resolver is one of them. Let's keep fastexpressionresolver as external thing. If somebody wants to use it then OK, he just writes small adapter. I think you understand that sing compile or even fast compile in our simple cases like "x => x.prop" is ineffective.

sungam3r:
> AstPrinter isn't used during build is it?

It's used at the execution stage. I want to rewrite it.

Issue #2237, opened 2021-01-30
Possible redesign of the request execution stage for GraphQL.NET v5


sungam3r:
In version 4 (which we planned to release on February 1), we made many optimizations over version 3. Memory consumption has dropped quite dramatically, which is good news. It is enough to look at the PRs marked with `performance` label. However, all these changes are limited to the current design of the project. The last optimization for v4 release - #2235 - significantly reduced memory consumption, but was limited by optimization of a purely internal data storage structure. I will try to describe the problem as I see it.

It's about the graphql request execution stage. Parsing and validation are beyond the scope of this issue, although you can think about their impact too. So, the current design relies on two large buffers to process a request. The first buffer is a tree of `ExecutionNode` objects. It is formed within the execution strategy starting from the root - `RootExecutionNode`. The execution strategy calculates the required fields at each level, creates tree nodes, calculates them, storing the calculation result in the `ExecutionNode.Result` property, and proceeds to the next level.

```c#
 public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionContext context)
        {
            var rootType = ExecutionHelper.GetOperationRootType(context.Document, context.Schema, context.Operation);
            var rootNode = BuildExecutionRootNode(context, rootType);

            await ExecuteNodeTreeAsync(context, rootNode)                  <--- STAGE 1
                .ConfigureAwait(false);

            // After the entire node tree has been executed, get the values
            object data = rootNode.ToValue();                              <--- STAGE 2

            return new ExecutionResult
            {
                Data = data,
                Query = context.Document.OriginalQuery,
                Document = context.Document,
                Operation = context.Operation,
                Extensions = context.Extensions
            };
        }
```
 As a result, we have a fully built tree. OK. The first buffer is built. Then comes the stage of collecting calculated results starting from the root, when the strategy collects all the values ​​from all nodes into an `ExecutionResult` object. This object is the second buffer and main "product" produced by the library. Further, this result, one way or another, will be serialized by the appropriate serializer (we have two of them out of the box - NewtonsoftJson and SystemTextJson) and given to the client via some transport, which is usually a HTTP.

These buffers are the main contributors to memory consumption - 16 MB of 23 MB  /  199K of 365K instances. Here is example of running the `introspection` query from the `ExecutionBencmark` 100 times:
![изображение](https://user-images.githubusercontent.com/21261007/106361997-7278d180-6331-11eb-96cf-a2b501613a34.png)

[...]

PR #2239, opened 2021-01-31
Remove ExecutionNode.IsResultSet and memory for Source

Shane32:
Should provide 16 bytes of savings for every instance of ExecutionNode, including ValueExecutionNode instances.  Have not yet run benchmarks

### Concept

* `Source` is only set during subscription execution.  It is left `null` for any other execution.  To reduce memory, `_source` has been removed from `ExecutionNode` along with the setter, and for subscriptions, another class is used that overrides the source property.
* `IsResultSet` is unnecessary, since the execution strategy almost always is required to process nodes like this: 1. create node 2. execute node 3. execute data loader, if any.   The only exception is that root nodes are not executed, and children of array nodes are not executed.  A simple type check (`node is RootExecutionNode` or `node.Parent is ArrayExecutionNode`) can determine this and skip execution.  Step 3 (execute data loader, if any) does not use `IsResultSet` but simply performs a type check on the result (`result is IDataLoaderResult`).  So, `IsResultSet` is unnecessary.  In order to verify this concept, I added `if (!IsResultSet) throw new Exception();` code to all the locations where `IsResultSet` was accessed, ran all the tests, and none of the exceptions were triggered.
* These two properties consumes 9 bytes on a 64-bit machine, and probably saves 16 bytes due to alignment.
* As `ValueExecutionNode` is one of the most memory-consuming objects overall during execution (per #2237), this represents a significant savings during execution.

### DIFF
--- a/src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
+++ b/src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
@@ -105,8 +105,7 @@ namespace GraphQL.Execution
                 return subscription
                     .Select(value =>
                     {
-                        var executionNode = BuildSubscriptionExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value);
-                        return executionNode;
+                        return BuildSubscriptionExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value);
                     })
                     .SelectMany(async executionNode =>
                     {
--- a/src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
+++ b/src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
@@ -105,7 +105,7 @@ namespace GraphQL.Execution
                 return subscription
                     .Select(value =>
                     {
-                        var executionNode = BuildExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value);
+                        var executionNode = BuildSubscriptionExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value);
                         return executionNode;
                     })
                     .SelectMany(async executionNode =>
@@ -157,6 +157,24 @@ namespace GraphQL.Execution
                 context.Errors.Add(error);
                 return null;
             }
+        }
+
+        /// <summary>
+        /// Builds an execution node with the specified parameters.
+        /// </summary>
+        protected ExecutionNode BuildSubscriptionExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, int? indexInParentNode, object source)
+        {
+            if (graphType is NonNullGraphType nonNullFieldType)
+                graphType = nonNullFieldType.ResolvedType;
+
+            return graphType switch
+            {
+                ListGraphType _ => new SubscriptionArrayExecutionNode(parent, graphType, field, fieldDefinition, indexInParentNode, source),
+                IObjectGraphType _ => new SubscriptionObjectExecutionNode(parent, graphType, field, fieldDefinition, indexInParentNode, source),
+                IAbstractGraphType _ => new SubscriptionObjectExecutionNode(parent, graphType, field, fieldDefinition, indexInParentNode, source),
+                ScalarGraphType scalarGraphType => new SubscriptionValueExecutionNode(parent, scalarGraphType, field, fieldDefinition, indexInParentNode, source),
+                _ => throw new InvalidOperationException($"Unexpected type: {graphType}")
+            };
         }
 
         private ExecutionError GenerateError(
--- a/src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
+++ b/src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
@@ -108,8 +108,7 @@ namespace GraphQL.Execution
                 return subscription
                     .Select(value =>
                     {
-                        var executionNode = BuildExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode);
-                        executionNode.Source = value;
+                        var executionNode = BuildExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value);
                         return executionNode;
                     })
                     .SelectMany(async executionNode =>
--- a/src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
+++ b/src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
@@ -103,10 +103,7 @@ namespace GraphQL.Execution
                 }
 
                 return subscription
-                    .Select(value =>
-                    {
-                        return BuildSubscriptionExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value);
-                    })
+                    .Select(value => BuildSubscriptionExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value))
                     .SelectMany(async executionNode =>
                     {
                         if (context.Listeners != null)
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -211,6 +211,7 @@ namespace GraphQL.Execution
         {
             context.CancellationToken.ThrowIfCancellationRequested();
 
+            // these are the only conditions upon which a node has already been executed when this method is called
             if (node is RootExecutionNode || node.Parent is ArrayExecutionNode)
                 return;
 
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -207,6 +207,24 @@ namespace GraphQL.Execution
         }
 
         /// <summary>
+        /// Builds an execution node with the specified parameters.
+        /// </summary>
+        public static ExecutionNode BuildExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, int? indexInParentNode, object source)
+        {
+            if (graphType is NonNullGraphType nonNullFieldType)
+                graphType = nonNullFieldType.ResolvedType;
+
+            return graphType switch
+            {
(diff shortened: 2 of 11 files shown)


src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -262,8 +280,6 @@ protected virtual async Task ExecuteNodeAsync(ExecutionContext context, Executio
         /// </summary>
         protected virtual async Task CompleteDataLoaderNodeAsync(ExecutionContext context, ExecutionNode node)
         {
-            if (!node.IsResultSet)
-                throw new InvalidOperationException("This execution node has not yet been executed");
```
Shane32, line 266: It is up to the execution strategy to run the method at the correct time.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -214,7 +232,7 @@ protected virtual async Task ExecuteNodeAsync(ExecutionContext context, Executio
         {
             context.CancellationToken.ThrowIfCancellationRequested();
 
-            if (node.IsResultSet)
+            if (node is RootExecutionNode || node.Parent is ArrayExecutionNode)
```
Shane32, line 235: These are the only conditions upon which a node has already been executed when this method is called.  It is up to the execution strategy to not call this method twice for the same node.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -298,9 +314,6 @@ protected virtual async Task CompleteDataLoaderNodeAsync(ExecutionContext contex
         /// </summary>
         protected virtual void CompleteNode(ExecutionContext context, ExecutionNode node)
         {
-            if (!node.IsResultSet)
-                throw new InvalidOperationException("This execution node has not yet been executed");
-
```
Shane32, line 303: It is up to the execution strategy to run the method at the correct time.

src/GraphQL/Execution/Nodes/ExecutionNode.cs
```diff
@@ -62,36 +62,16 @@ internal IGraphType ResolvedType
...
-        private object _source;
         /// <summary>
         /// Returns the parent node's result. If set, the set value will override the parent node's result.
         /// </summary>
-        public object Source
-        {
-            get => _source ?? Parent?.Result;
-            set => _source = value;
-        }
+        public virtual object Source => Parent?.Result;
```
Shane32, line 74: Only subscription executions change the source value.  Note that field resolvers do not have access to the setter, since `IResolveFieldContext.Source` is a read-only property.

src/GraphQL/Execution/Nodes/ExecutionNode.cs
```diff
@@ -125,7 +105,7 @@ public IObjectGraphType GetParentType(ISchema schema)
             if (parentType is IObjectGraphType objectType)
                 return objectType;
 
-            if (parentType is IAbstractGraphType abstractType && Parent.IsResultSet)
+            if (parentType is IAbstractGraphType abstractType)
```
Shane32, line 108: I'm not sure exactly when this executes, but after running all the tests, the parent always had its result set.  I believe that is implicit, as the child nodes are not constructed until the parent node is executed (so that if the result is null, for example, no child node is constructed).

src/GraphQL/Execution/Nodes/ObjectExecutionNode.cs
```diff
@@ -33,7 +33,7 @@ public IObjectGraphType GetObjectGraphType(ISchema schema)
         {
             var objectGraphType = GraphType as IObjectGraphType;
 
-            if (GraphType is IAbstractGraphType abstractGraphType && IsResultSet)
+            if (GraphType is IAbstractGraphType abstractGraphType)
```
Shane32, line 36: I'm not exactly sure when this executes (unions or interfaces???), but I ran all the tests and the result was always set.

src/GraphQL/Execution/Nodes/SubscriptionArrayExecutionNode.cs
```diff
@@ -0,0 +1,21 @@
+using GraphQL.Language.AST;
+using GraphQL.Types;
+
+namespace GraphQL.Execution
+{
+    /// <inheritdoc/>
+    public class SubscriptionArrayExecutionNode : ArrayExecutionNode
```
Shane32, line 7: These classes were added to the main project because they would be necessary for any subscription implementation.

Shane32:
The other stages were unchanged.  Seems about 1% slower and 3-7% faster

### before

|        Method |           Stage |           Mean |         Error |   Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |---------------- |---------------:|--------------:|--------:|--------:|------:|----------:|
| Introspection |         Execute | 660,155.618 ns | 4,428.2779 ns | 50.7813 | 15.6250 |     - |  322306 B |
|          Hero |         Execute |   2,165.165 ns |     9.6959 ns |  0.3586 |       - |     - |    2264 B |
|     Variables |         Execute |   2,811.039 ns |    12.9771 ns |  0.3204 |       - |     - |    2024 B |
|       Literal |         Execute |  29,800.789 ns |   169.0502 ns |  2.0752 |       - |     - |   13127 B |

### after

|        Method |           Stage |           Mean |         Error |   Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------- |---------------- |---------------:|--------------:|--------:|--------:|------:|----------:|
| Introspection |         Execute | 669,300.467 ns | 3,322.1543 ns | 47.8516 | 14.6484 |     - |  300314 B |
|          Hero |         Execute |   2,196.080 ns |     5.2176 ns |  0.3471 |       - |     - |    2200 B |
|     Variables |         Execute |   2,737.731 ns |    10.0248 ns |  0.3166 |       - |     - |    1992 B |
|       Literal |         Execute |  29,981.630 ns |    74.8146 ns |  2.0752 |       - |     - |   13095 B |

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -203,6 +203,24 @@ protected static ExecutionNode BuildExecutionNode(ExecutionNode parent, IGraphTy
             };
         }
 
+        /// <summary>
+        /// Builds an execution node with the specified parameters.
+        /// </summary>
+        public static ExecutionNode BuildExecutionNode(ExecutionNode parent, IGraphType graphType, Field field, FieldType fieldDefinition, int? indexInParentNode, object source)
```
sungam3r, line 209: Let's rename it to `BuildSubscriptionExecutionNode`, make `protected` and move into `SubscriptionExecutionStrategy`.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Please add these comments above that check.

src/GraphQL/Execution/Nodes/ExecutionNode.cs
```diff
@@ -62,36 +62,16 @@ internal IGraphType ResolvedType
...
-            {
-                IsResultSet = true;
-                _result = value;
-            }
-        }
+        public object Result { get; set; }
 
-        private object _source;
         /// <summary>
         /// Returns the parent node's result. If set, the set value will override the parent node's result.
```
sungam3r, line 72: Old comment.

src/GraphQL/Execution/Nodes/SubscriptionArrayExecutionNode.cs
sungam3r: So, maybe move all them with the newly added `Build` method to `GraphQL.SystemReactive` project ?

sungam3r:
I'm finished review. Are you going to do benchmarks again after all the changes from updated `develop`?

Shane32:
> I'm finished review. Are you going to do benchmarks again after all the changes from updated `develop`?

I hadn't planned to.  These changes have nothing to do with the other changes.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: ok

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
```diff
@@ -105,8 +105,7 @@ protected virtual async Task<IObservable<ExecutionResult>> ResolveEventStreamAsy
                 return subscription
                     .Select(value =>
                     {
-                        var executionNode = BuildExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode);
-                        executionNode.Source = value;
+                        var executionNode = BuildSubscriptionExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value);
                         return executionNode;
```
sungam3r, line 109:
```suggestion
                        return BuildSubscriptionExecutionNode(node.Parent, node.GraphType, node.Field, node.FieldDefinition, node.IndexInParentNode, value);
```

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: ping

src/GraphQL/Execution/Nodes/SubscriptionArrayExecutionNode.cs
sungam3r: ping

Shane32: well we can but you these are not specific to systemreactive.  i thought you said we were only putting SystemReactive specific implementations in that project?

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32:
```suggestion
            // these are the only conditions upon which a node has already been executed when this method is called
            if (node is RootExecutionNode || node.Parent is ArrayExecutionNode)
```

src/GraphQL/Execution/Nodes/SubscriptionArrayExecutionNode.cs
sungam3r: You are right. I'm 50/50. On the one hand, yes. On the other hand, the new classes are only used by this implementation, the only one yet :). We can think of them as implementation details. It is possible that another subscription implementation will create completely different nodes or not create them at all. Who knows?

Shane32: It seemed obvious to me that they would be necessary for any implementation, considering the current design of the project.

Shane32: i don't really care.  i just want to finish these PRs up, finish the documentation up, and release 4.0

sungam3r (approved):

PR #2240, opened 2021-01-31
Additional memory reductions during serialization

Shane32:
This PR makes `ExecutionResult.Data` return `RootExecutionNode`.  To implement compatibility with any serializer, the object nodes implement `IReadOnlyDictionary<string, object>`, and the array node implements `IEnumerable<object>`, which works perfectly, but it creates many allocations and is quite a slow implementation.  Nevertheless, it is fully compatible with any serializer that supports reading `IReadOnlyDictionary<TKey, TValue>` interfaces, such as `System.Text.Json` or `Newtonsoft.Json`.  I also added special cases for execution nodes to the `System.Text.Json` serializer; the `Newtonsoft.Json` one is not done yet.  The `ObjectProperty` struct can also be removed with these changes.

All tests currently pass.

Todo:
* [x] Update Newtonsoft.Json serializer
* [x] Update xml comments and migration notes

### Results

As can be seen below, execution is faster but serialization is slower.  However, it's overall faster for the introspection query and a tad slower for hero.  Memory is reduced.

|Method|Stage|Mean [base]|Mean [current]|Mean [% of base]|Allocated [base]|Allocated [current]|Allocated [% of base]|
|---|---|--:|--:|--:|--:|--:|--:|
|Introspection|Build|281,260.094 ns|282,340.303 ns|100|96528 B|96528 B|100|
|Introspection|Parse|17,523.912 ns|17,548.621 ns|100|11920 B|11920 B|100|
|Introspection|Convert|9,253.560 ns|9,348.068 ns|<span style="color:red">101</span>|16552 B|16552 B|100|
|Introspection|Validate|173,310.135 ns|172,034.408 ns|<span style="color:green">99</span>|11104 B|11104 B|100|
|Introspection|DeserializeVars|1.501 ns|1.711 ns|<span style="color:red">113</span>|-|-|-|
|Introspection|ParseVariables|2.860 ns|2.619 ns|<span style="color:green">91</span>|-|-|-|
|Introspection|Execute|582,945.951 ns|556,940.586 ns|<span style="color:green">95</span>|207335 B|177027 B|<span style="color:green">85</span>|
|Introspection|Serialize|107,162.553 ns|111,683.320 ns|<span style="color:red">104</span>|27880 B|27880 B|100|
|||||||||
|Hero|Build|279,297.943 ns|278,755.163 ns|<span style="color:green">99</span>|96528 B|96531 B|100|
|Hero|Parse|801.924 ns|801.048 ns|<span style="color:green">99</span>|784 B|784 B|100|
|Hero|Convert|461.641 ns|459.236 ns|<span style="color:green">99</span>|1056 B|1056 B|100|
|Hero|Validate|11,006.777 ns|11,422.321 ns|<span style="color:red">103</span>|2464 B|2464 B|100|
|Hero|DeserializeVars|1.735 ns|1.734 ns|<span style="color:green">99</span>|-|-|-|
|Hero|ParseVariables|3.031 ns|2.843 ns|<span style="color:green">93</span>|-|-|-|
|Hero|Execute|1,716.775 ns|1,724.945 ns|100|1600 B|1504 B|<span style="color:green">94</span>|
|Hero|Serialize|513.271 ns|515.936 ns|100|504 B|504 B|100|
|||||||||
|Variables|Build|191,344.038 ns|190,870.461 ns|<span style="color:green">99</span>|65281 B|65272 B|<span style="color:green">99</span>|
|Variables|Parse|1,540.209 ns|1,534.266 ns|<span style="color:green">99</span>|1008 B|1008 B|100|
|Variables|Convert|515.959 ns|513.607 ns|<span style="color:green">99</span>|1208 B|1208 B|100|
|Variables|Validate|20,086.668 ns|19,651.892 ns|<span style="color:green">97</span>|7704 B|7704 B|100|
|Variables|DeserializeVars|9,477.746 ns|9,315.755 ns|<span style="color:green">98</span>|8288 B|8288 B|100|
|Variables|ParseVariables|30,644.252 ns|29,713.257 ns|<span style="color:green">96</span>|13013 B|13013 B|100|
|Variables|Execute|2,546.086 ns|2,491.273 ns|<span style="color:green">97</span>|1680 B|1640 B|<span style="color:green">97</span>|
|Variables|Serialize|369.021 ns|392.786 ns|<span style="color:red">106</span>|504 B|504 B|100|
|||||||||
|Literal|Build|188,807.993 ns|194,132.687 ns|<span style="color:red">102</span>|65272 B|65276 B|100|
|Literal|Parse|18,528.523 ns|18,408.929 ns|<span style="color:green">99</span>|9016 B|9016 B|100|
|Literal|Convert|7,465.206 ns|7,574.813 ns|<span style="color:red">101</span>|12768 B|12768 B|100|
|Literal|Validate|132,329.197 ns|131,758.906 ns|<span style="color:green">99</span>|27200 B|27202 B|100|
|Literal|DeserializeVars|1.737 ns|1.287 ns|<span style="color:green">74</span>|-|-|-|
|Literal|ParseVariables|3.037 ns|2.569 ns|<span style="color:green">84</span>|-|-|-|
|Literal|Execute|29,396.838 ns|29,422.399 ns|100|12782 B|12742 B|<span style="color:green">99</span>|
|Literal|Serialize|398.607 ns|379.813 ns|<span style="color:green">95</span>|504 B|504 B|100|


src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
```diff
@@ -56,6 +59,38 @@ public override object ToValue()
...
 
+        public override bool ClearErrorNodes()
+        {
+            if (Items == null)
+                return true;
+
+            if (Items.Count == 0)
+                return false;
+
+            var ok = false;
```
sungam3r, line 70: ?

Shane32: Short circuit so the type doesn’t have to be checked for every iteration

Shane32:
An additional speed improvement is possible: if there are no execution errors, it can be assumed that there were no null values provided for non-null graphs, so ClearErrorNodes can be skipped. I’m not positive but I think so. Quitting now as it’s 5am. Go ahead and work on the branch if you like @sungam3r

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
Shane32: Note that ClearErrorNodes still needs to be called for the descendants so can’t just return false.

Shane32:
Also the IReadOnlyDictionary implementation is purely optional, not used by the custom serializer.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
```diff
@@ -56,6 +59,38 @@ public override object ToValue()
...
+            for (int i = 0; i < Items.Count; ++i)
+            {
+                var item = Items[i];
+                bool valueIsNull = item.ClearErrorNodes();
+
+                if (valueIsNull && !ok)
+                {
+                    if (((ListGraphType)GraphType).ResolvedType is NonNullGraphType)
+                    {
+                        Items = null;
```
sungam3r, line 81: I suspect changing `Items` and `SubFields` might cause problems. I'm not sure.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
sungam3r: Please give a better name to this variable.

sungam3r:
I feel like we should postpone the v4 release for a week. We don't have documentation ready. We have this PR and I would like to finish it. In addition, when I started preparing an example of using applied directives, in the process of preparing it, I changed the design of the visitors and fixed several bugs. I'll do a PR today or tomorrow. All of these changes are backwards incompatible and should be included in version 4.

Shane32:
> I feel like we should postpone the v4 release for a week. 

sure

> We don't have documentation ready. We have this PR and I would like to finish it. 

I’ll go back to the documentation if you want to finish this up. 

> In addition, when I started preparing an example of using applied directives, in the process of preparing it, I changed the design of the visitors and fixed several bugs. I'll do a PR today or tomorrow. All of these changes are backwards incompatible and should be included in version 4.

Sounds good

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
Shane32: It might change the effect of accessing certain properties of a copied IResolveFieldContext after execution...

Shane32: Or might not since maybe sub fields is cloned also...

SebastianStehle:
Are you planning to provide a pre-release to nuget? I would be happy to try it out.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
sungam3r: I would think about how to achieve a one-time setting of this property. It may be that this is not yet possible.

sungam3r:
I think we will publish v4 release after not later than a week. I don't see much point in publishing a preview.

Shane32:
> I think we will publish v4 release after not later than a week. I don't see much point in publishing a preview.

I agree.

sungam3r:
Waiting for #2239

sungam3r:
изображение
?

src/GraphQL/Execution/Nodes/ExecutionNode.cs
```diff
@@ -95,6 +95,11 @@ protected ExecutionNode(ExecutionNode parent, IGraphType graphType, Field field,
         /// </summary>
         public abstract object ToValue();
 
+        /// <summary>
+        /// Returns true if the node should return null
```
sungam3r, line 99:
```suggestion
        /// Returns <see langword="true"/> if the node should return <see langword="null"/>.
```

sungam3r:
> Also the IReadOnlyDictionary implementation is purely optional, not used by the custom serializer.

So, maybe delete then?

Shane32:
> изображение
> ?

Lol I saw that too. I’m pretty tired and must have typed the wrong thing.

Shane32:
> > Also the IReadOnlyDictionary implementation is purely optional, not used by the custom serializer.
> 
> So, maybe delete then?

Sure but how should we type ExecutionResult.Data ? object ? Or should we create/use some interface rather than typing it as ExecutionNode.

sungam3r:
It _may_ stay as object as long as we provide serializers to convert that arbitrary object into json response. Also we _may_ provide some explicit method to convert that object into "well-known old format" when data is built from nested dictionaries if someone still needs to handle ExecutionResult as before.

Shane32:
|Method|Stage|Mean<br/>[base]|Mean<br/>[current]|Mean<br/>[% of base]|Allocated<br/>[base]|Allocated<br/>[current]|Allocated<br/>[% of base]|
|---|---|--:|--:|--:|--:|--:|--:|
|Introspection|Execute|598,626.8 ns|579,447.0 ns|<span style="color:green">96</span>|185327 B|155012 B|<span style="color:green">83</span>|
|Introspection|Serialize|113,659.7 ns|112,065.0 ns|<span style="color:green">98</span>|27880 B|27880 B|100|
|||||||||
|Hero|Execute|1,921.8 ns|1,896.1 ns|<span style="color:green">98</span>|1512 B|1416 B|<span style="color:green">93</span>|
|Hero|Serialize|538.2 ns|521.0 ns|<span style="color:green">96</span>|504 B|504 B|100|
|||||||||
|Variables|Execute|2,588.1 ns|2,665.7 ns|<span style="color:red">102</span>|1624 B|1584 B|<span style="color:green">97</span>|
|Variables|Serialize|380.9 ns|405.4 ns|<span style="color:red">106</span>|504 B|504 B|100|
|||||||||
|Literal|Execute|29,170.9 ns|30,480.8 ns|<span style="color:red">104</span>|12726 B|12686 B|<span style="color:green">99</span>|
|Literal|Serialize|386.2 ns|383.2 ns|<span style="color:green">99</span>|504 B|504 B|100|

src/GraphQL/Execution/Nodes/ExecutionNode.cs
```diff
@@ -95,6 +95,11 @@ protected ExecutionNode(ExecutionNode parent, IGraphType graphType, Field field,
         /// </summary>
         public abstract object ToValue();
 
+        /// <summary>
+        /// Returns <see langword="true"/> if the node should return <see langword="null"/>.
```
Shane32, line 99:
```suggestion
        /// Prepares this node and children nodes for serialization. Returns <see langword="true"/> if this node should return <see langword="null"/>.
```

sungam3r:
Codecov works!

sungam3r:
> and the array node implements IEnumerable<object>, which works perfectly, but it creates many allocations and is quite a slow implementation

Is it still so?

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
```diff
@@ -12,34 +12,95 @@ public class ExecutionResultJsonConverter : JsonConverter
...
+                    writer.WriteEndArray();
+                }
+            }
+            else if (node == null)
+            {
+                writer.WriteNull();
+            }
+            else
+            {
+                serializer.Serialize(writer, node.ToValue());
```
sungam3r, line 99: That is, in fact, this line will never be executed. In general, the whole benefit is to call `ToValue` as rarely as possible. Now it only happens for `ValueExecutionNode`. Right?

src/GraphQL.Tests/TestExtensions.cs
```diff
@@ -1,19 +1,66 @@
...
+
+            if (data is IReadOnlyDictionary<string, object> properties)
+            {
+                return properties;
+            }
+
+            throw new ArgumentException($"Unknown type {data.GetType()}. Parameter must be of type ObjectExecutionNode or IDictionary<string, object>.", nameof(data));
+        }
+
+        public static RootExecutionNode ToData(this IReadOnlyDictionary<string, object> dictionary)
```
sungam3r, line 27: `ToExecutionTree` ?

src/GraphQL.Tests/TestExtensions.cs
```diff
@@ -1,19 +1,66 @@
...
+        private static ExecutionNode CreateExecutionNode(string name, object value)
+        {
+            if (value is IEnumerable<KeyValuePair<string, object>> dict)
+            {
+                return new ObjectExecutionNode(null, null, new GraphQL.Language.AST.Field(new GraphQL.Language.AST.NameNode(name), default), null, default)
+                {
+                    SubFields = dict.Select(x => CreateExecutionNode(x.Key, x.Value)).ToArray(),
+                };
+            }
+            else if (value?.GetType() != typeof(string) && value is IEnumerable list)
```
sungam3r, line 45: What if value is `byte[]` ?

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
```diff
@@ -12,34 +12,95 @@ public class ExecutionResultJsonConverter : JsonConverter
...
+                {
+                    writer.WriteStartArray();
+                    foreach (var childNode in items)
+                    {
+                        WriteExecutionNode(writer, childNode, serializer);
+                    }
+                    writer.WriteEndArray();
+                }
+            }
+            else if (node == null)
```
sungam3r, line 93:
```suggestion
            else if (node == null || node is NullExecutionNode)
```

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
```diff
@@ -40,7 +40,65 @@ private static void WriteData(Utf8JsonWriter writer, ExecutionResult result, Jso
...
+                {
+                    writer.WriteStartArray();
+                    foreach (var childNode in items)
+                    {
+                        WriteExecutionNode(writer, childNode, options);
+                    }
+                    writer.WriteEndArray();
+                }
+            }
+            else if (node == null)
```
sungam3r, line 95:
```suggestion
            else if (node == null || node is NullExecutionNode)
```

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
Shane32: It should not, unless another execution strategy sets Data to something other than a RootExecutionNode. There are probably various tests that still use this code, but I didn’t check other than to change the tests I saw and ensure that both serializers were being exercised. 

Yes, that is correct that ToValue is only called for ValueExecutionNode instance (or a derived instance that does not derive from ObjectExecutionNode or ArrayExecutionNode).

src/GraphQL.Tests/TestExtensions.cs
Shane32: Sure

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
```diff
@@ -25,35 +25,54 @@ public ArrayExecutionNode(ExecutionNode parent, IGraphType graphType, Field fiel
...
 
-                if (value == null)
+                if (valueIsNull && !isNullableType)
                 {
-                    var listType = item.FieldDefinition.ResolvedType;
-                    if (listType is NonNullGraphType nonNull)
-                        listType = nonNull.ResolvedType;
-
-                    if (((ListGraphType)listType).ResolvedType is NonNullGraphType)
+                    if (((ListGraphType)GraphType).ResolvedType is NonNullGraphType)
```
sungam3r, line 63: Strange. It seems that evaluation of this expression can be taken out of the loop : 
```
    var isNullableType = ((ListGraphType)GraphType).ResolvedType is NonNullGraphType;
```

src/GraphQL.Tests/TestExtensions.cs
Shane32: Byte arrays are not native JSON types. So it should work correctly. 

I wrote the code to mimic the behavior in ObjectExtensions.GetPropertyValue and TypeExtensions.IsAnIEnumerable and AutoRegisteringGraphType.GetRealType, and so on. They all just check for a string and continue otherwise.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
sungam3r: I think this is a separate issue that requires attention. We need to stop execution as early as possible. Imagine `ArrayExecutionNode` of type `[Person!]` with 100000 child nodes. First element yields `null` in `SetArrayItemNodes`, so...

Shane32:
> > And the array node implements `IEnumerable<object>`, which works perfectly, but it creates many allocations and is quite a slow implementation.
> 
> Is it still so?

No, per your direction, I removed that capability.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
Shane32: Yes it can, but the loop still needs to executed. I benchmarked it a few different ways and found this to be the fastest. Of course it would really depend on what exact data was being returned.

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
```diff
@@ -128,12 +128,19 @@ await listener.AfterExecutionAsync(context)
                             }
                         }
 
+                        // Set the execution node's value to null if necessary
+                        executionNode.ClearErrorNodes();
```
sungam3r, line 132: If it returns `true` ? In `ExecutionStrategy` you use `  object data = rootNode.ClearErrorNodes() ? null : rootNode;`

Shane32: I thought long and hard about this.  Pretty sure it's correct.  Let's say there's a problem.  (Ignoring the fact that errors are not being returned, not sure why.)  What happens is that the executionNode will clear itself so ToValue will return null.  The Root node will then serialize this result:

```json
{
  "cart": null
}
```

rather than what it would normally do:

```json
{
  "cart": {
    "count": 3
  }
}
```

So basically the code is designed to assume that the subscription field is nullable.  Maybe we should check the spec on that.

Shane32: I can't find it in the spec (although it's probably there somewhere).  I've never used subscriptions so if you have any insight, let me know.  I'd feel pretty confident that the implementation is correct, and also, it matches what was there before.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
sungam3r: Also what about `protected virtual void ValidateNodeResult(ExecutionContext context, ExecutionNode node)` ? I have a suspicion that all checks in `ClearNodeErrors` will never work...

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
sungam3r: Strange. Why the fastest? Maybe only for arrays for size=0? :) But in general, I doubt that these checks are necessary. It seems to me that they are already executed earlier at the execution stage. Or at least we need to add them there.

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
sungam3r: Let's keep `//TODO: check` with a link to this comment https://github.com/graphql-dotnet/graphql-dotnet/pull/2240#discussion_r570631402 in the code. We'll check it someday.

Shane32: So I guess a subscription field can be defined as a non-null object type within GraphQL.Net.  I don't know if that meets spec or not.  Spec says that a subscription field must be an object type.  But let's say that was allowed.  Then the response would be:
```json
{
}
```
Does that make any sense?  How would the receiving end of the subscription know which subscription failed?  I don't know enough about subscriptions, but my gut feeling is that it doesn't make sense.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
Shane32: It is the fastest because often there are no situations where the value is null, and the arrays are often small.  So a couple quick null checks ends up being faster than a (relatively) costly type check -- followed by the same null check.  At least that's my theory.

I don't know how it can determine it earlier in the execution.  The child nodes have not been executed at that point in time.  Child nodes may have non-null "grandchildren" which return null, causing a cascade up to the "child".  That only happens after the nodes have been executed.  Sure if the children are value execution nodes then it could be checked at the time it is set, but it's got to be checked at some point in time.

The only thing that could happen is that when a value execution node is set, it could check all the parent nodes at the time to clear the children.  But this may prevent some child nodes from executing.  So you might need to add a boolean property.  It just seems like we're going in a loop here.  If you want to try it, there may be some way to get it to run a tad faster.

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
Shane32:
```suggestion
                        // Note: assumes that the subscription field is nullable, regardless of how it was defined
                        // See https://github.com/graphql-dotnet/graphql-dotnet/pull/2240#discussion_r570631402
                        //TODO: check if a non-null subscription field is allowed per the spec
                        //TODO: check if errors should be returned along with the data
                        executionNode.ClearErrorNodes();
```

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
Shane32: SetArrayItemNodes executes prior to the execution of child object nodes. Checking if the child is null is impossible in that case. 

ValidateNodeResult runs at its appropriate time. It is not supposed to throw errors for each parent node that is coerced to null because of a child execution error. 

The code executes the same as previously and the checks are essential. They were within ToValue, and I moved them out and into a method that you are required to run immediately prior to calling ToValue.

Shane32: (Serializing the node itself is essentially the same as calling ToValue)

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
sungam3r: > How would the receiving end of the subscription know which subscription failed?

Every subscription has a unique identifier and state bot server and client side.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
sungam3r: > But this may prevent some child nodes from executing.

Is it bad? I think this is a win.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
sungam3r: OK. I published another issue regarding this.

src/GraphQL/Execution/Nodes/ArrayExecutionNode.cs
Shane32: > > But this may prevent some child nodes from executing.
> 
> Is it bad? I think this is a win.

Your suggestion meets spec.

> If during `ExecuteSelectionSet()` a field with a non‐null fieldType throws a field error then that error must propagate to this entire selection set, either resolving to null if allowed or further propagated to a parent field.
>
> If this occurs, any sibling fields which have not yet executed or have not yet yielded a value may be cancelled to avoid unnecessary work.

sungam3r: OK. #2253

Shane32:
@sungam3r Are there any other changes required here?

sungam3r:
Give me a 1-2 hours, I have one thought left that I wanted to check

src/GraphQL/Execution/Nodes/ObjectExecutionNode.cs
```diff
@@ -42,28 +42,46 @@ public IObjectGraphType GetObjectGraphType(ISchema schema)
...
         /// within a <see cref="Dictionary{TKey, TValue}"/>.
+        /// <see cref="ClearErrorNodes"/> must be called prior to calling this method.
         /// </summary>
         public override object ToValue()
         {
             if (SubFields == null)
                 return null;
 
-            var fields = new ObjectProperty[SubFields.Length];
+            var fields = new Dictionary<string, object>(SubFields.Length);
```
sungam3r, line 52: Theoretically there could be an empty dictionary, SubFields.Length = 0, but it's OK.

src/GraphQL/Execution/Nodes/ObjectExecutionNode.cs
```diff
@@ -42,28 +42,46 @@ public IObjectGraphType GetObjectGraphType(ISchema schema)
...
                 var child = SubFields[i];
-                object value = child.ToValue();
+                fields.Add(child.Name, child.ToValue());
+            }
+
+            return fields;
+        }
+
+        /// <inheritdoc/>
+        public override bool ClearErrorNodes()
```
sungam3r, line 64: Let's change `ClearErrorNodes` to `PropagateNull`. This name is clearer.

sungam3r:
I was thinking about the relationship between `Result`, `Items` and `SubFields` properties. I was wondering if it is possible to leave the `Result` property only for `ValueExecutionNode`. It seems that the answer is no.

sungam3r:
@Shane32 By the way, what do you think about renaming `ValueExecutionNode` to `ScalarExecutionNode`?

sungam3r:
I'm done here.

Shane32:
> @Shane32 By the way, what do you think about renaming `ValueExecutionNode` to `ScalarExecutionNode`?

50/50, but let's get this merged first.

sungam3r (approved):
Really good job.

Shane32:
Thanks!

PR #2243, opened 2021-01-31
Fix name validator test

### LINKED ISSUE #2145: CI tests intermittently fail due to static variable usage
See https://github.com/graphql-dotnet/graphql-dotnet/issues/1964#issuecomment-754693085

sungam3r:
Fixes #2145


Shane32 (approved):
Ok, if it works.  By the way, can we use an enum for `type` instead of a string?

sungam3r:
It is not so important.

PR #2244, opened 2021-01-31
Only net5 for local tests

sungam3r:


Shane32 (commented):
Why would you do this?  I always want all the platforms to test locally.  I certainly would want to if we ever ran into an issue.  If you want to run a single platform, click on that platform and hit run.

I suggest you figure out how to configure your local computer to only run a certain platform, if that's how you want it to work for your PC.  But it should not be default for the repo.

Shane32:
By the way, we have had multiple issues that only applied to a single platform.  This would be very annoying to debug if we had to mess with the setting every time we wanted to run on .NET 3 / 5 / 2.2

sungam3r:
I am not comfortable with choosing some platforms every time. In addition, due to many platforms, test detection is slower, noticeably slower. I can make all platforms default and replace one setting locally.

Shane32:
> I am not comfortable with choosing some platforms every time. In addition, due to many platforms, test detection is slower, noticeably slower. I can make all platforms default and replace one setting locally.

Just configure a local github setting to ignore a config file if you want it different.  The projects in the repo should be configured for all the supported platforms.

sungam3r:
Done. What is "local github setting" ?

Shane32:
e.g. `Directory.Build.local.Props`.  Then you can set it however you want in there.

Shane32:
Edit `.gitignore` to ignore `Directory.Build.local.Props`.  Configure `Directory.Build.Props` to import `Directory.Build.local.Props`

sungam3r:
Did you mean  _git_ settings?

Shane32:
something like that.  then you can add the file to your local machine, and it will always be ignored no matter what branch you're in,etc.

Shane32:
> Did you mean _git_ settings?

yes

sungam3r:
Done

Shane32 (approved):

PR #2251, opened 2021-02-03
The ExecutionResult.Data property is serialized only if the request was executed.

sungam3r:
See https://spec.graphql.org/June2018/#sec-Data
> If an error was encountered before execution begins, the data entry should not be present in the result.
> If an error was encountered during the execution that prevented a valid response, the data entry in the response should be null.

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -73,6 +73,7 @@ namespace GraphQL
 
             ExecutionResult result = null;
             ExecutionContext context = null;
+            bool executionOccurred = false;
 
             try
             {
@@ -230,6 +231,8 @@ namespace GraphQL
                     };
                 }
 
+                executionOccurred = true;
+
                 using (metrics.Subject("execution", "Executing operation"))
                 {
                     if (context.Listeners != null)
@@ -266,10 +269,7 @@ namespace GraphQL
                         }
                 }
 
-                if (context.Errors.Count > 0)
-                {
-                    result.Errors = context.Errors;
-                }
+                result.AddErrors(context.Errors);
             }
             catch (OperationCanceledException) when (options.CancellationToken.IsCancellationRequested)
             {
@@ -277,13 +277,7 @@ namespace GraphQL
             }
             catch (ExecutionError ex)
             {
-                result = new ExecutionResult
-                {
-                    Errors = new ExecutionErrors
-                    {
-                        ex
-                    }
-                };
+                (result ??= new ExecutionResult()).AddError(ex);
             }
             catch (Exception ex)
             {
@@ -298,18 +292,14 @@ namespace GraphQL
                     ex = exceptionContext.Exception;
                 }
 
-                result = new ExecutionResult
-                {
-                    Errors = new ExecutionErrors
-                    {
-                        ex is ExecutionError executionError ? executionError : new UnhandledError(exceptionContext?.ErrorMessage ?? "Error executing document.", ex)
-                    }
-                };
+                (result ??= new ExecutionResult()).AddError(ex is ExecutionError executionError ? executionError : new UnhandledError(exceptionContext?.ErrorMessage ?? "Error executing document.", ex));
             }
             finally
             {
                 result ??= new ExecutionResult();
                 result.Perf = metrics.Finish();
+                if (executionOccurred)
+                    result.Executed = true;
                 context?.Dispose();
             }
 
--- a/src/GraphQL/Execution/ExecutionResult.cs
+++ b/src/GraphQL/Execution/ExecutionResult.cs
@@ -96,8 +96,10 @@ namespace GraphQL
         {
             if (errors?.Count > 0)
             {
-                foreach (var error in errors)
-                    AddError(error);
+                if (Errors == null)
+                    Errors = new ExecutionErrors(errors.Count);
+                foreach (var error in errors.List)
+                    Errors.Add(error);
             }
 
             return this;
--- a/src/GraphQL/Execution/ExecutionResult.cs
+++ b/src/GraphQL/Execution/ExecutionResult.cs
@@ -12,6 +12,13 @@ namespace GraphQL
     public class ExecutionResult
     {
         /// <summary>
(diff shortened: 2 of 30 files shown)


src/GraphQL/Execution/ExecutionResult.cs
```diff
@@ -11,6 +11,13 @@ namespace GraphQL
     /// </summary>
     public class ExecutionResult
     {
+        /// <summary>
+        /// Indicates if the operation included execution. If an error was encountered BEFORE execution begins,
+        /// the data entry SHOULD NOT be present in the result. If an error was encountered DURING the execution
+        /// that prevented a valid response, the data entry in the response SHOULD BE null.
+        /// </summary>
+        public bool Executed { get; set; }
```
sungam3r, line 19: @Shane32 If there is a better way to distinguish between the errors before and after execution stage, then suggest.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -266,24 +269,15 @@ await listener.AfterExecutionAsync(context)
                         }
                 }
 
-                if (context.Errors.Count > 0)
-                {
-                    result.Errors = context.Errors;
-                }
+                result.AddErrors(context.Errors);
```
Shane32, line 272: Why are we making these changes?  It just adds more allocations during execution.  Plus the code is harder to read.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -298,18 +292,14 @@ await listener.AfterExecutionAsync(context)
...
                 }
 
-                result = new ExecutionResult
-                {
-                    Errors = new ExecutionErrors
-                    {
-                        ex is ExecutionError executionError ? executionError : new UnhandledError(exceptionContext?.ErrorMessage ?? "Error executing document.", ex)
-                    }
-                };
+                (result ??= new ExecutionResult()).AddError(ex is ExecutionError executionError ? executionError : new UnhandledError(exceptionContext?.ErrorMessage ?? "Error executing document.", ex));
```
Shane32, line 295: And, this is now harder to read than it was before.

src/GraphQL/Execution/ExecutionResult.cs
Shane32: This seems fine to me.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: It's fine though.  Optimizing runtime when errors are present isn't a priority.

sungam3r: Why more ?

sungam3r: My thought was that errors from the context should complement the errors already recorded in the result, and not replace them. This is not true?

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: My thought was that errors from the context should complement the errors already recorded in the result, and not replace them. This is not true? What if `result` is not null?

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: What if `result.Errors` is not null/empty?

Shane32: With `AddErrors`, it has to iterate over the errors in `context.Errors` and add them to a new instance within `result`.  So it is more allocations.

Shane32: > My thought was that errors from the context should complement the errors already recorded in the result, and not replace them. This is not true?

Makes sense, although there is no existing possibility of this occurring, unless perhaps a completely custom executer was used.

src/GraphQL/Execution/ExecutionResult.cs
sungam3r: What about default value? Both `true`/`false` options have disadvantages.

Shane32 (approved):

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: Ok.  Currently that is only possible if a document listener throws an error after an execution.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -230,6 +231,8 @@ await listener.AfterValidationAsync(context, validationResult)
                     };
                 }
 
+                executionOccurred = true;
```
Shane32, line 234: The positioning here seems to make sense.  The document listeners could throw an exception, but they could be considered part of the execution stage....

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: > It's fine though. Optimizing runtime when errors are present isn't a priority.

I have not optimized.

> With AddErrors, it has to iterate over the errors in context.Errors and add them to a new instance within result. So it is more allocations.

OK. Possible one `List` instance.

> Makes sense, although there is no existing possibility of this occurring, unless perhaps a completely custom executer was used.

I just didn't like that I saw the potential for losing errors. I made sure that they are not lost in any situation, without understanding whether it is possible now or in the future. Is it good?

Shane32: 👍

src/GraphQL/Execution/ExecutionResult.cs
Shane32: `false`

Shane32:
Good to merge.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: Yes. Listeners "belong" to execution.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: Indeed.

PR #2252, opened 2021-02-04
Change NameValidator's type argument to an enum

Shane32:


src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -2507,11 +2507,19 @@ namespace GraphQL.Utilities
         public TType GetMetadata<TType>(string key, TType defaultValue = default) { }
         public bool HasMetadata(string key) { }
     }
+    public enum NameType
```
sungam3r, line 2510: Will there be confusion with `NamedType`?

Shane32: I didn't think of that.  If you have a suggestion, I'm all ears.  I couldn't think of a good name.

Shane32: If you would rather skip this PR, that's fine too, as it shouldn't really matter.  But in principle I think an enum is the right choice for the method.

src/GraphQL/Utilities/NameType.cs
```diff
@@ -0,0 +1,33 @@
+namespace GraphQL.Utilities
+{
+    /// <summary>
+    /// Indicates a field, type, argument, enum or directive.
+    /// </summary>
+    public enum NameType
```
sungam3r, line 6: NamedElement ? "Type" can be confusing.

src/GraphQL/Utilities/NameType.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// An argument
+        /// </summary>
+        Argument,
+
+        /// <summary>
+        /// An enum
+        /// </summary>
+        Enum,
```
sungam3r, line 26: EnumValue ?

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: 50/50

joemcbride: I like the use of the enum.  👍🏼

src/GraphQL/Utilities/NameType.cs
Shane32: I like that

src/GraphQL/Utilities/NameType.cs
Shane32: Ok

src/GraphQL/Utilities/NamedElement.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        Type,
+
+        /// <summary>
+        /// An argument
+        /// </summary>
+        Argument,
+
+        /// <summary>
+        /// An enum
```
sungam3r, line 24:
```suggestion
        /// An enum value
```

sungam3r (approved):

Issue #1927, opened 2020-10-20
 ListGraphType issue in 3.0.0.2026


OpenSpacesAndPlaces:
I'm not sure if my problem is related this pending fix or different.
https://github.com/graphql-dotnet/graphql-dotnet/issues/1831

I'm getting this when I try to add a ListGraphType on a Query Resolver:
```
Output type 'Query' can have fields only of output types: ScalarGraphType, ObjectGraphType, InterfaceGraphType, UnionGraphType or EnumerationGraphType. Field 'foo_list' has an input type.
Parameter name: fieldType 
```

This used to work OK for me in 2.4 (and prior).

OpenSpacesAndPlaces:
@sungam3r If you have a minute for this one - I'd appreciate it. Thanks much!

Shane32:
That's the designed behavior in 3.0.  Output graphs and input graphs are different types - the error message is fairly descriptive.  Is there some way I can help further?

Shane32:
The error message is saying that your output graph `Query` has a field `foo_list` that returns an input graph type.  So for example you can return `ListGraphType<StringGraphType>` for a list of strings, but for complex objects you'd have `ListGraphType<FooGraphType>`, and `FooGraphType` needs to inherit from `ObjectGraphType` (or one of the other ones listed there in the error message).  It can't inherit from `InputObjectGraphType`.

Shane32:
Complex argument objects can inherit from `InputObjectGraphType`, however.

Shane32:
I would feel pretty sure that your problem relates to the changes in PR #1436

OpenSpacesAndPlaces:
Maybe the usage has changed, but the reason I was doing this was because I was splitting out query resolver for requests that only returns objects {} and ones that only return lists [{}].

e.g. I'm using a custom ObjectGraphType and then wrapping in ListGraphType for the ListResolver.

OpenSpacesAndPlaces:
```
AddField(new FieldType
 {
             Name = friendlyTableName, 
	     ResolvedType = tableType, 
   .....
});

//Definition of List Query/Return Type
var listType = new ListGraphType(tableType);
AddField(new FieldType
{
       Name = $"{friendlyTableName}_list",
       ResolvedType = listType, 
....
});
```

Shane32:
Can you provide a sample of what `tableType` might be?  It would seem that `tableType` is an `InputObjectGraphType`, which isn't allowed in 3.0.

[...]

PR #2258, opened 2021-02-05
Patch for issue 1927 - adding a field where Type = ListGraphType

Shane32:
See issue #1927 

This does not fix the underlying issue, that `IsInputType` sometimes cannot accurately identify if the supplied type is an input or output type, but rather this code does not call `IsInputType` on `Type` if `ResolvedType` is set.

Note: both tests failed before and both pass now.  Also, the original code sample within #1927 was tested and now works.

Note: these are almost the only uses of `IsInputType(Type type)` and `IsOutputType(Type type)`.  The one other use does not suffer the same issue.

Also added check to `FieldType.Type` setter.

### DIFF
--- a/src/GraphQL/Types/Composite/ComplexGraphType.cs
+++ b/src/GraphQL/Types/Composite/ComplexGraphType.cs
@@ -97,13 +97,13 @@ namespace GraphQL.Types
             {
                 if (this is IInputObjectGraphType)
                 {
-                    if (fieldType.ResolvedType?.IsInputType() == false || fieldType.Type?.IsInputType() == false)
+                    if (fieldType.ResolvedType != null ? fieldType.ResolvedType.IsInputType() == false : fieldType.Type?.IsInputType() == false)
                         throw new ArgumentOutOfRangeException(nameof(fieldType),
                             $"Input type '{Name ?? GetType().GetFriendlyName()}' can have fields only of input types: ScalarGraphType, EnumerationGraphType or IInputObjectGraphType. Field '{fieldType.Name}' has an output type.");
                 }
                 else
                 {
-                    if (fieldType.ResolvedType?.IsOutputType() == false || fieldType.Type?.IsOutputType() == false)
+                    if (fieldType.ResolvedType != null ? fieldType.ResolvedType.IsOutputType() == false : fieldType.Type?.IsOutputType() == false)
                         throw new ArgumentOutOfRangeException(nameof(fieldType),
                             $"Output type '{Name ?? GetType().GetFriendlyName()}' can have fields only of output types: ScalarGraphType, ObjectGraphType, InterfaceGraphType, UnionGraphType or EnumerationGraphType. Field '{fieldType.Name}' has an input type.");
                 }
--- a/src/GraphQL/Types/Composite/NonNullGraphType.cs
+++ b/src/GraphQL/Types/Composite/NonNullGraphType.cs
@@ -6,7 +6,7 @@ namespace GraphQL.Types
     public class NonNullGraphType<T> : NonNullGraphType
         where T : GraphType
     {
-        /// <inheritdoc cref="NonNullGraphType.NonNullGraphType(Type)"/>
+        /// <inheritdoc cref="NonNullGraphType(Type)"/>
         public NonNullGraphType()
             : base(typeof(T))
         {
@@ -27,18 +27,18 @@ namespace GraphQL.Types
             if (type is NonNullGraphType)
             {
                 // http://spec.graphql.org/draft/#sec-Type-System.Non-Null.Type-Validation
-                throw new ArgumentException("Cannot nest NonNull inside NonNull.", nameof(type));
+                throw new ArgumentOutOfRangeException(nameof(type), "Cannot nest NonNull inside NonNull.");
             }
 
             ResolvedType = type;
         }
 
-        /// <inheritdoc cref="NonNullGraphType.NonNullGraphType(IGraphType)"/>
+        /// <inheritdoc cref="NonNullGraphType(IGraphType)"/>
         protected NonNullGraphType(Type type)
         {
-            if (type == typeof(NonNullGraphType))
+            if (typeof(NonNullGraphType).IsAssignableFrom(type))
             {
-                throw new ArgumentException("Cannot nest NonNull inside NonNull.", nameof(type));
+                throw new ArgumentOutOfRangeException(nameof(type), "Cannot nest NonNull inside NonNull.");
             }
 
             Type = type;
--- a/src/GraphQL/Types/Fields/FieldType.cs
+++ b/src/GraphQL/Types/Fields/FieldType.cs
@@ -68,12 +68,11 @@ namespace GraphQL.Types
             set
             {
                 if (value != null && !value.IsGraphType())
-                    ThrowInvalidType(value);
+                    throw new ArgumentOutOfRangeException("value", $"Type '{value}' is not a graph type");
                 _type = value;
             }
         }
 
-        private void ThrowInvalidType(Type type) => throw new ArgumentOutOfRangeException("value", $"Type '{type}' is not a graph type");
 
         /// <summary>
         /// Gets or sets the graph type of this field.
--- a/src/GraphQL/Types/Fields/FieldType.cs
+++ b/src/GraphQL/Types/Fields/FieldType.cs
@@ -58,10 +58,22 @@ namespace GraphQL.Types
             }
         }
 
+        private Type _type;
         /// <summary>
         /// Gets or sets the graph type of this field.
         /// </summary>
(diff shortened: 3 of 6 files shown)


src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -58,10 +58,22 @@ public object DefaultValue
...
+            get => _type;
+            set
+            {
+                if (value != null && !value.IsGraphType())
+                    throw Create(value);
+                _type = value;
+            }
+        }
+
+        private ArgumentOutOfRangeException Create(Type type) => new ArgumentOutOfRangeException("value", $"Type '{type}' is not a graph type");
```
Shane32, line 76: I assume this pattern (that I've seen elsewhere) is for performance??? Hope I did that right.

sungam3r: https://reubenbond.github.io/posts/dotnet-perf-tuning

Almost.

Shane32: I don't have time to read up on this today.  Suggest a change if you think it should change.

sungam3r: `throw` needs to be inside helper method so JIT can inline the rest.

Shane32: Makes sense.  Why is it this way elsewhere then?

Shane32, line 76:
```suggestion
                if (value != null && !value.IsGraphType())
                    ThrowInvalidType(value);
                _type = value;
            }
        }

        private void ThrowInvalidType(Type type) => throw new ArgumentOutOfRangeException("value", $"Type '{type}' is not a graph type");
```

Shane32: Is this wrong:
https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/QueryArgument.cs#L104..L127

Shane32: Looks like CheckType couldn't be inlined, so why it's in a private method I don't know.

sungam3r: This is a low-level optimization and only makes sense on a hot path. There is absolutely no need to apply this approach everywhere. I want to say that it is quite possible to do without it here. I used throw helpers in the parser because it showed the real effect there. Not very big.

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -58,10 +58,22 @@ public object DefaultValue
...
         /// Gets or sets the graph type of this field.
         /// </summary>
-        public Type Type { get; set; }
+        public Type Type
+        {
+            get => _type;
+            set
+            {
+                if (value != null && !value.IsGraphType())
+                    ThrowInvalidType(value);
```
sungam3r, line 71:
```suggestion
                    throw new ArgumentOutOfRangeException("value", $"Type '{value}' is not a graph type");
```

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -58,10 +58,22 @@ public object DefaultValue
...
+                _type = value;
+            }
+        }
+
+        private void ThrowInvalidType(Type type) => throw new ArgumentOutOfRangeException("value", $"Type '{type}' is not a graph type");
```
sungam3r, line 76:
```suggestion
```

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -97,13 +97,13 @@ public virtual FieldType AddField(FieldType fieldType)
             {
                 if (this is IInputObjectGraphType)
                 {
-                    if (fieldType.ResolvedType?.IsInputType() == false || fieldType.Type?.IsInputType() == false)
+                    if (fieldType.ResolvedType != null ? fieldType.ResolvedType.IsInputType() == false : fieldType.Type?.IsInputType() == false)
```
sungam3r, line 100: I don't undestand the difference.

Shane32: If `ResolvedType` contains an object, `Type` is ignored

sungam3r: If `fieldType.ResolvedType` is `null` then `fieldType.ResolvedType?.IsInputType()` is `null` too and `null` == `false` yields `false`.

sungam3r: The same below.

Shane32: That's not the part of the equation that has changed.  The point is to skip checking `Type` if `ResolvedType` contains an object, whether it is compatible or not, since `Type` is not used in that circumstance.

sungam3r: > If ResolvedType contains an object, Type is ignored

So it happened. This is bad? After changes I see the same behavior.

Shane32: Yes, that's what fixes the issue posted by OpenSpacesAndPlaces in #1927 .  He had set both `ResolvedType` and `Type`.

Shane32: See https://github.com/graphql-dotnet/graphql-dotnet/issues/1927#issuecomment-774198841

Shane32: I added tests to verify that code like his would work correctly now.  See the test I wrote, which would fail without this change.  Notice this line of code:

```cs
obj.AddField(new FieldType { Name = "list", ResolvedType = list, Type = list.GetType() });
```

And particularly `Type = list.GetType()`.  Seems to make perfect sense from a user's point of view.

Shane32: However, it should not matter, since the execution would ignore the `Type` property as long as `ResolvedType` was set.  So the fix here is a valid fix.

sungam3r: Understood, finally...

src/GraphQL/Types/Composite/NonNullGraphType.cs
```diff
@@ -27,18 +27,18 @@ public NonNullGraphType(IGraphType type)
             if (type is NonNullGraphType)
```
sungam3r, line 27: See `ResolvedType` setter. It's `public`. `Type` has `private` setter. Let's make `ResolvedType` setter as `internal`

sungam3r: Also I would like to move 
```
       protected NonNullGraphType(Type type)
        {
            if (type == typeof(NonNullGraphType))
            {
                throw new ArgumentException("Cannot nest NonNull inside NonNull.", nameof(type));
            }

            Type = type;
        }
```
into `NonNullGraphType<T>` and make `public virtual Type Type => null`. We can set `Type` only from `NonNullGraphType<T>`. Then in `NonNullGraphType<T>` :  `public override Type Type => _type`

sungam3r: After these changes the API will become less controversial.

Shane32: That has no bearing on this issue/PR.

Besides, doesn't `ResolvedType` need to be settable for any implementation of `SchemaTypes`?  I'm not sure what the point of forcing people to use `BuildNamedType` is.

sungam3r: Ok. I will do another PR.

sungam3r (approved):

src/GraphQL/Types/Composite/NonNullGraphType.cs
sungam3r: Why forcing?

sungam3r: They can use one or another constructor.

Shane32: It is impossible to write a custom `SchemaTypes` implementation if the `ResolvedType` setter is `internal`.  None of the other `ResolvedType` properties on classes are read-only for the same reason.  It should stay read/write at this time.  We should, however, design a fully read-only implementation of `ISchema` that is used during execution, so there are no inadvertent changes to the schema at runtime.

sungam3r: I'm talking about `ResolvedType` on `ListGraphType` and `NonNullGraphType` and not about `FieldType.ResolvedType`.

PR #2259, opened 2021-02-05
Better type checks for NonNull and List

sungam3r:


src/GraphQL/Types/Composite/ListGraphType.cs
```diff
@@ -27,23 +16,48 @@ public ListGraphType(IGraphType type)
...
+        /// Initializes a new instance for the specified inner graph type.
+        /// </summary>
+        public ListGraphType()
+            : base(null)
+        {
+            _type = typeof(T);
+        }
+
+        /// <inheritdoc/>
+        public override Type Type => _type;
```
Shane32, line 61: `Type` is only used when the schema is built, correct?  There should be no reason to cache the value.
```suggestion
        public override Type Type => typeof(T);
```

src/GraphQL/Types/Composite/NonNullGraphType.cs
```diff
@@ -24,37 +13,62 @@ public class NonNullGraphType : GraphType, IProvideResolvedType
...
+            if (typeof(NonNullGraphType).IsAssignableFrom(typeof(T)))
+            {
+                throw new ArgumentOutOfRangeException("type", "Cannot nest NonNull inside NonNull.");
+            }
+
+            _type = typeof(T);
+        }
+
+        /// <inheritdoc/>
+        public override Type Type => _type;
```
Shane32, line 72:
```suggestion
        public override Type Type => typeof(T);
```
See above.  Also need to remove `_type` if this is changed.

src/GraphQL/Types/Composite/ListGraphType.cs
sungam3r: 🤦

Shane32:
We should make it standard practice to write a test for added checks.  It would only take a few lines a code.  I would not check for the exact error message but simply that `ArgumentOutOfRangeException` occurred.

src/GraphQL/Types/Composite/NonNullGraphType.cs
sungam3r: Also 🤦

sungam3r:
OK. Red Codecov report motivates me to do it now.

Shane32 (approved):

Issue #2257, opened 2021-02-05
How to change output of decimal scalar (removing tailing zeros)


grean:
In my database, i have a field of type decimal(12,6)
and when i want to get it with graphql endpoint, have an output of : 
"field" : 2.000000 with all the zeros..
How can i remove all of those trailing zeros in order to reduce the HTTP payload ?

In my REST WepApi, to achieved this, i used to take this class instead

```
public class DecimalJsonConverter : JsonConverter
    {
        public override bool CanConvert(Type objectType)
        {
            return objectType == typeof(decimal);
        }

        public override object ReadJson(JsonReader reader, Type objectType, object existingValue,
            JsonSerializer serializer)
        {
            if (reader.TokenType == JsonToken.Null)
            {
                return null;
            }
            else
            {
                return Convert.ToDecimal(reader.Value, CultureInfo.InvariantCulture);
            }

            //|| reader.TokenType == JsonToken.Float
            //|| reader.TokenType == JsonToken.Integer)
            //{

            //}
            //return reader.Value;
        }

        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
        {
            writer.WriteRawValue(((decimal)value).ToString("G6", CultureInfo.InvariantCulture));
        }
    }
```

And i apply the convertion to the DTO´s field with an attribute like this : 
```
[JsonConverter(typeof(DecimalJsonConverter))]
 public decimal? Field { get; set; }
```
Thanks for your advices :)

sungam3r:
> and when i want to get it with graphql endpoint, have an output of :
"field" : 2.000000 with all the zeros..

It depends on the serializer you are using. We provide two. Please show how you configure your server (startup.cs / composition root). Both serializers have almost unlimited customization options. See constructors with serialization options [here](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.NewtonsoftJson/DocumentWriter.cs) and [there](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.SystemTextJson/DocumentWriter.cs).  

Also we have corresponding extensions [here](https://github.com/graphql-dotnet/server/blob/master/src/Transports.AspNetCore.NewtonsoftJson/GraphQLBuilderExtensions.cs) and [there](https://github.com/graphql-dotnet/server/blob/master/src/Transports.AspNetCore.SystemTextJson/GraphQLBuilderExtensions.cs) if you use server project.

[...]

PR #2270, opened 2021-02-09
Add schema types validation

sungam3r:


README.md
```diff
@@ -31,9 +31,8 @@ by [Marek Magdziak](https://github.com/mkmarek) and released with a MIT license.
 
 ## Documentation
 
-http://graphql-dotnet.github.io
-
-This site is in sync with `master` branch.
+1. http://graphql-dotnet.github.io - documentation site that is built from the [docs](/docs2/site/) folder in the `master` branch.
+2. https://graphql.org/learn - learn about GraphQL, how it works, and how to use it.
```
sungam3r, line 35: We get asked a lot of questions about the GraphQL basics.

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
```diff
@@ -48,41 +48,15 @@ public async Task validate_core_schema_pascal_case(IDocumentWriter documentWrite
 
             var json = await documentWriter.WriteToStringAsync(executionResult);
 
-            ShouldBe(json, "IntrospectionResult".ReadJsonResult());
-        }
-
-        [Theory]
-        [ClassData(typeof(DocumentWritersTestData))]
-        public async Task validate_core_schema_doesnt_use_nameconverter(IDocumentWriter documentWriter)
```
sungam3r, line 56: After adding schema validation and fixing tests, it is no longer possible for name converter not to be called. Query must be defined and must have at least one field.

src/GraphQL.Tests/Introspection/SchemaIntrospectionTests.cs
```diff
@@ -48,41 +48,15 @@ public async Task validate_core_schema_pascal_case(IDocumentWriter documentWrite
...
-            public string NameForField(string fieldName, IComplexGraphType parentGraphType) => throw new Exception();
+            ShouldBe(json, "IntrospectionResult".ReadJsonResult().Replace("\"test\"", "\"Test\""));
         }
 
         public class TestQuery : ObjectGraphType
         {
             public TestQuery()
             {
                 Name = "TestQuery";
+                Field<StringGraphType>("test");
```
sungam3r, line 59: Every Object type **must** have fields.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -412,6 +412,9 @@ private void HandleField(IComplexGraphType parentType, FieldType field, TypeColl
 
             if (field.ResolvedType == null)
             {
+                if (field.Type == null)
+                    throw new InvalidOperationException($"Both ResolvedType and Type properties on field '{parentType?.Name}.{field.Name}' are null.");
```
sungam3r, line 416: While fixing the tests, I discovered that NRE was happening a bit later in `AddTypeIfNotRegistered`

src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
```diff
@@ -0,0 +1,186 @@
+using System;
+using System.Linq;
+using GraphQL.Types;
+
+namespace GraphQL.Utilities
+{
+    internal sealed class SchemaValidationVisitor : BaseSchemaNodeVisitor
```
sungam3r, line 7: Many checks in this class are not achievable in practice (for example names starting with "__"), because we try to control the correctness of the schema definition as early as possible. Nevertheless, I'm sure it would be correct to have in this class the full set of checks required by the specification. Perhaps more will be added over time. Probably not if we go to have this class reflect standard spec requirements. Other checks can always be placed in other visitors.

README.md
```diff
@@ -31,9 +31,14 @@ by [Marek Magdziak](https://github.com/mkmarek) and released with a MIT license.
...
 ## Documentation
 
-http://graphql-dotnet.github.io
+1. http://graphql-dotnet.github.io - documentation site that is built from the [docs](/docs2/site/) folder in the `master` branch.
+2. https://graphql.org/learn - learn about GraphQL, how it works, and how to use it.
 
-This site is in sync with `master` branch.
+## Debugging
+
+All packages generated from this repository come with embedded pdb and support [Source Link](https://github.com/dotnet/sourcelink).
```
sungam3r, line 39: Ping @grean for https://github.com/graphql-dotnet/graphql-dotnet/issues/2257#issuecomment-776213185

src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
Shane32: Even if sealed, it should still be public so that a custom schema implementation can call it.

Shane32:
Are we going to move name validation in here?

src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
sungam3r: done

sungam3r:
I think not. At least not now. If some errors can be detected earlier, then that's good. We may have additional checks that are not in the specification. We can place them not only at the initialization stage of the schema.

Shane32 (approved):

Issue #1874, opened 2020-09-22
Byte Array sent through Variables(graphql .net client library) receive invalid_value


eyaldar:
## Description

Hi, 

I'm trying to create an InputObjectGraphType/ObjectGraphType with a property of byte array.
I'm using Json .NET for serialization.
I'm trying to receive a byte array property to my server query field resolver but it fails with INVALID_VALUE
I'd appreciate your help in understanding what am i doing wrong
Thanks a lot in advance!

### Steps to reproduce

I have this basic setup:
```
   public class BytesHolder
    {
        public byte[] Bytes { get; set; }
    }
```

```
   public class OutputBytesType : ObjectGraphType<BytesHolder>
    {
        public OutputBytesType()
        {
            Field(x => x.Bytes);
        }
    }
```

```
    public class InputBytesType : InputObjectGraphType<BytesHolder>
    {
        public InputBytesType()
        {
            Field(x => x.Bytes);
        }
    }
```

```
   public class DemoQuery : ObjectGraphType
    {
        public DemoQuery()
        {
            Field<OutputBytesType>(
                "bytes",
                arguments: new QueryArguments(new QueryArgument<InputBytesType> { Name = "bytesObject" }),
                resolve: context => 
                {
                    var bytesObject = context.GetArgument<BytesHolder>("bytesObject");
                    return bytesObject;
                }
            );
        }
    }
```

I'm sending this request through a console application client
```
            var bytesRequest = new GraphQLRequest
            {
                Query = @"
                query BytesRequest($bytesHolder: InputBytesType) {
                    bytes(bytesObject: $bytesHolder) {
                        bytes
                    }
                }",
                Variables = new
                {
                    bytesHolder = new BytesHolder
                    {
                        Bytes = new byte[] { 1, 2, 3, 4 }
                    }
                }
            };

            var graphQLResponse = await graphQLClient.SendQueryAsync<BytesHolder>(bytesRequest);
```

Upon receiving the request, the server fails to parse the request (it gets to this line:
```            
Register<string, byte>(value => byte.Parse(value, NumberFormatInfo.InvariantInfo));
```
instead of this line:
```
Register<string, byte[]>(value => Convert.FromBase64String(value)); // such a built-in conversion for string->byte[] seems useful
```

### Expected result

The server should succeed in parsing the byte array property.

### Actual result

The server fails with INVALID_VALUE due to miscalculation in the value converter

### Environment

[...]

PR #2273, opened 2021-02-10
Add test for #1874

sungam3r:


Shane32:
So it looks like bug #1874 was fixed with the updated variable parsing code?

src/GraphQL.Tests/Bugs/Bug1874.cs
```diff
@@ -0,0 +1,67 @@
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    public class Bug11874 : QueryTestBase<Bug11874Schema>
```
Shane32, line 7: Issue1874 ?  Or at least Bug1874 ?

src/GraphQL.Tests/Bugs/Bug1874.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    public class Bug11874Schema : Schema
+    {
+        public Bug11874Schema()
+        {
+            Query = new DemoQuery();
+        }
+    }
+
+    public class DemoQuery : ObjectGraphType
```
Shane32, line 31: Bug1874Query

src/GraphQL.Tests/Bugs/Bug1874.cs
```diff
@@ -0,0 +1,67 @@
...
+                resolve: context =>
+                {
+                    var bytesObject = context.GetArgument<BytesHolder>("bytesObject");
+                    return bytesObject;
+                }
+            );
+        }
+    }
+
+    public class BytesHolder
```
Shane32, line 47: Bug1874BytesHolder

src/GraphQL.Tests/Bugs/Bug1874.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        public byte[] Bytes { get; set; }
+    }
+
+    public class OutputBytesType : ObjectGraphType<BytesHolder>
```
Shane32, line 52: etc (since all these classes are in the same namespace as all the rest of the bugs)

src/GraphQL.Tests/Bugs/Bug1874.cs
Shane32: Or maybe these classes can be moved inside Bug1874Schema

src/GraphQL.Tests/Bugs/Bug1874.cs
sungam3r: oops, fixed

src/GraphQL.Tests/Bugs/Bug1874.cs
sungam3r: done

sungam3r:
> So it looks like bug #1874 was fixed with the updated variable parsing code?

Nope, just misuse.

Shane32 (approved):

Shane32:
> > So it looks like bug #1874 was fixed with the updated variable parsing code?
> 
> Nope, just misuse.

So it works on the master branch also?

sungam3r:
Well, yes... but after fixing `InputBytesType` type in query text to `Bug1874InputBytesType`

Shane32:
Knowing how badly the variable parsing was coded before, it's amazing any variables were able to be parsed.

PR #2274, opened 2021-02-10
Demo: Base64 conversion of byte arrays

Shane32:
@sungam3r we can merge this, or not.  Just wanted to check that this worked the way I thought it did.


sungam3r:
Why not?

sungam3r (approved):

Issue #1460, opened 2019-12-20
String with limited size


emulicus:
## Summary
Need to be able to limit size of the string field on arguments or mutation input graphs. Ideally, this would be something graphql client would be able to see in the schema and also show errors in graphiQL when they enter the value that's over the limit.
## Relevant information
 Also, since we base most of our input graph types on the entity framework entities ideally this length would come from that StringLength attribute on the EF entity property.

We're using graph types first approach. 
2.4 version.

joemcbride:
I'm not sure about showing info in the schema (maybe you can just add the requirement to the description for the field), but here is a validation framework that @SimonCropp put together.  It currently targets 2.4.

https://github.com/SimonCropp/GraphQL.Validation

sungam3r:
@emulicus This is a great task to solve with graphql directives.
```graphql
directive @length(max: Int!) on FIELD_DEFINITION

input Payload {
  name: String! @length(max: 50)
}
```

There are no problems in establishing our own directive in schema. The problem is to provide information about this directive to the client. Now the specification does not explicitly describe the possibility of returning directives through introspection, but this is a fairly popular request from the community. See  graphql/graphql-spec#300 

Also you may be interested in #1332 and #1451. We have been successfully using this fork in production for some time. In fact, we were interested in providing information about authentication and authorization directives. Maclean is the next directive that interests us.

emulicus:
Thank you, guys. I will check out your resources.

emulicus:
> I'm not sure about showing info in the schema (maybe you can just add the requirement to the description for the field), but here is a validation framework that @SimonCropp put together. It currently targets 2.4.
> 
> https://github.com/SimonCropp/GraphQL.Validation

This is a little rough for us, because most of our graph types are based on entity framework entities. Not sure if we could have a single validator that would check the length of the input and compare it to the StringLength attribute from the entity.

emulicus:
> @emulicus This is a great task to solve with graphql directives.
> 
> ```graphql
> directive @length(max: Int!) on FIELD_DEFINITION
> 
> input Payload {
>   name: String! @length(max: 50)
> }
> ```
> 
> There are no problems in establishing our own directive in schema. The problem is to provide information about this directive to the client. Now the specification does not explicitly describe the possibility of returning directives through introspection, but this is a fairly popular request from the community. See [graphql/graphql-spec#300](https://github.com/graphql/graphql-spec/issues/300)
> 
[...]

PR #2276, opened 2021-02-11
Length directive

### LINKED ISSUE #1002: How to set a custom directive on field?

sungam3r:
Fixes #1002 and #1460.

I added the first "non-standard" `length` directive and a validation rule for it.


sungam3r:
@Shane32 I suggest using this custom directive as an example in the documentation.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
```diff
@@ -0,0 +1,81 @@
...
+            if (node.Value is NullValue)
+            {
+                if (min != null)
+                    context.ReportError(new InputFieldsAndArgumentsOfCorrectLengthError(context, node, null, (int?)min, (int?)max));
+            }
+            else if (node.Value is StringValue strLiteral)
+            {
+                CheckStringLength(strLiteral.Value);
+            }
+            else if (node.Value is VariableReference vRef && context.Inputs != null)
```
sungam3r, line 57: @Shane32 This is the first rule that uses variables.

src/GraphQL/Validation/TypeInfo.cs
```diff
@@ -28,43 +28,52 @@ public TypeInfo(ISchema schema)
...
 
-            return e.Current; // throws if index is out of range
+                int i = index;
+                do
+                {
+                    _ = e.MoveNext();
+                }
+                while (i-- > 0);
+
+                return e.Current; // throws if index is out of range
```
Shane32, line 48: It should throw, but the behavior of `IEnumerable.Current` in that case is "undefined".  We should check the return value of `MoveNext`.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                var operation = string.IsNullOrWhiteSpace(context.OperationName)
+                    ? context.Document.Operations.FirstOrDefault()
+                    : context.Document.Operations.WithName(context.OperationName);
+
+                //TODO: change to get single variable
+                var values = ExecutionHelper.GetVariableValues(context.Document, context.Schema, operation?.Variables, context.Inputs);
```
Shane32, line 64: This looks like bad practice.  We should never call `GetVariableValues` within a validation rule.  It would have to parse all variables a second time.  Maybe multiple times depending on how many times this ran.  Besides, does this even work in practice?  Assuming there is a length directive on a string as part of an input object, doesn't this get skipped because there is no directive applied to the input argument?  At first glance it seems like it would only work if an argument type was a `StringGraphType`.

It would seem that we need to design a validation framework for variables, which for speed purposes should be processed at the same time that the variables are being parsed.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
```diff
@@ -0,0 +1,81 @@
...
+namespace GraphQL.Validation.Rules
+{
+    /// <summary>
+    /// Validation rule that checks minimum and maximum length of provided values for input fields and
+    /// arguments that marked with <see cref="LengthDirective"/> directive. Doesn't check default values.
+    ///
+    /// <br/><br/>
+    /// This is not a standard validation rule that is not in the specification.
+    /// </summary>
+    public class InputFieldsAndArgumentsOfCorrectLength : IValidationRule
```
Shane32, line 16: This is great demo code, and we should include it as a test to demonstrate a fully-functional directive, and it can serve to help us realize where the framework should change to suit these types of use cases .... but I don't think it should be included in the library.  Especially not with what appears to be some major design flaws as previously noted.

src/GraphQL/Validation/Errors.Custom/InputFieldsAndArgumentsOfCorrectLengthError.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using GraphQL.Language.AST;
+
+namespace GraphQL.Validation.Errors
+{
+    /// <inheritdoc cref="Rules.InputFieldsAndArgumentsOfCorrectLength"/>
+    [Serializable]
+    public class InputFieldsAndArgumentsOfCorrectLengthError : ValidationError
```
Shane32, line 8: Would not include in library.

src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
```diff
@@ -37,13 +37,19 @@ internal sealed class AppliedDirectivesValidationVisitor : ISchemaNodeVisitor
...
         {
             if (provider.HasAppliedDirectives())
             {
-                var applied = provider.GetAppliedDirectives().List;
+                var appliedDirectives = provider.GetAppliedDirectives().List;
 
-                foreach (var appliedDirective in applied)
+                foreach (var directive in schema.Directives.List)
+                {
+                    if (!directive.Repeatable && appliedDirectives.Count(applied => applied.Name == directive.Name) > 1)
```
Shane32, line 48: What's a repeatable directive?

src/GraphQL/Types/Directives/LengthDirective.cs
```diff
@@ -0,0 +1,50 @@
+using System;
+
+namespace GraphQL.Types
+{
+    /// <summary>
+    /// Used to specify the minimum and/or maximum length for an input field or argument.
+    /// </summary>
+    public class LengthDirective : DirectiveGraphType
```
Shane32, line 8: Once we figure out all the details of the implementation, such as not calling `GetVariableValues` more than once for a document, then we could consider including this directive.  I'm not sure if it should be included directly in the main namespace or even in the same nuget package.  Perhaps it should be in a namespace called `GraphQL.Extensions.Directives` or something.  The thing is that it currently cannot be used without updating the validation rule set to include the proper rule.  So unlike `UrlGraphType` and similar extensions, additional configuration is necessary for this directive to function.  Thoughts?

src/GraphQL.Tests/Validation/InputFieldsAndArgumentsOfCorrectLengthTests.cs
```diff
@@ -0,0 +1,287 @@
...
+
+        [Fact]
+        public void below_min_variable_input_field_length()
+        {
+            ShouldFailRule(_ =>
+            {
+                _.Query = @"query q($value: String)
+                {
+                  complicatedArgs {
+                    complexArgField(complexArg: { requiredField: true, stringField: $value })
```
Shane32, line 63: There should be a test for:

```gql
query q($complex: ...)
{
  complicatedArgs {
    complexArgField(complexArg: $complex)
  }
}
```

Where the stringField within the complex field is "aa".  So far I don't see how that will work with the current implementation.

Shane32 (changes requested):
It still seems like the sample directive needs some work before it's ready to include within the project.

Shane32:
> @Shane32 I suggest using this custom directive as an example in the documentation.

👍 

As noted above, I believe it still needs some work, but we need to release v4, and as an example, it's not critical that it's production-ready.

Btw, is this a sample of what the new 'applied directives' feature allows?  Or was this available previously?

Shane32:
> Fixes #1002 and #1460.
> 
> I added the first "non-standard" `length` directive and a validation rule for it.

I am glad you are working on this for the benefit of the project/community.  It will really help people understand how directives can be used. 🥇

src/GraphQL/Validation/TypeInfo.cs
sungam3r: fixed

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
sungam3r: >  It would have to parse all variables a second time. 

I know.

> Besides, does this even work in practice? Assuming there is a length directive on a string as part of an input object, doesn't this get skipped because there is no directive applied to the input argument? 

It works. See `HasAppliedDirectives` / `GetAppliedDirectives().Find` above.

> it's possible that we would need 3 separate validation passes

Something like this. Not a small job. Not my goal now.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
sungam3r: >  design flaws

It's not that bad. Existing code will not be harmed at all. No performance loss.

> I don't think it should be included in the library

Why so? Think of many scalar types that we have included in the library - `Uri` for example. I do not mind moving this directive into a separate project, but is it necessary now? There are other parts of the core library that could be moved to other projects to make a more modular solution. This is a lot of work and you can't do it quickly.

src/GraphQL/Validation/Errors.Custom/InputFieldsAndArgumentsOfCorrectLengthError.cs
sungam3r: Let's think about the name of a new project to put this into.

src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
sungam3r:
```graphql
type Query {
  secretInfo: String @allow("SuperUser") @allow("Tom")
  publicInfo: String @allow("Readers") @allow("Guest")
}
```

изображение

src/GraphQL/Types/Directives/LengthDirective.cs
sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/pull/2276#discussion_r574480038

> additional configuration is necessary for this directive to function

Not necessary, although most likely. The directive acts as metainformation. It can be used at execution stage, not just validation. Also this information is useful to the client himself. For example, in my case, I could completely do without the validation rule, because my clients are generated automatically by introspection. I get information about the `@length` directive and do checks in the clients themselves, that is, there is not even a network call and validation.

Also think about something like `@cached(expiration: Int)` directive. The server can declare through introspection directives that it does not make sense to query individual fields again (some global/static data). The server only marks the fields with a directive, but does not have any additional processing logic. The directive serves as a way to tell the client what to do and what not to do. This is food for thought for client.

src/GraphQL.Tests/Validation/InputFieldsAndArgumentsOfCorrectLengthTests.cs
sungam3r: You are right 👍 , fixed. See the new `IVariableVisitor` design.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
sungam3r: > design a validation framework for variables

See my first steps with `IVariableVisitor`.

sungam3r:
> As noted above, I believe it still needs some work, but we need to release v4, and as an example, it's not critical that it's production-ready.

If I had not started doing this, I would not have found all the errors that I found. This required a design change, backwards incompatible changes. So it needs to be finished. Directives should work in v4.

> Btw, is this a sample of what the new 'applied directives' feature allows? Or was this available previously?

Of course this is an example of 'applied directives' too. `@length` directive is introspectable and client will see it from introspection query now if being in experimental mode.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
Shane32: Now that we have a `IVariableVisitor` framework, it is at least possible to provide a fully working sample.  I would like to review further.

Shane32: Back to this particular line of code ... this would only run if there was a length directive applied to a string type that is a variable reference, correct?  So as long as we can make it only run `GetVariableValue` to retrieve only the variable desired, it should not be unreasonably slower even for the most complex of queries/mutations.  This should be implemented before this PR is merged.  (It is of course possible to manually pull the string value from the inputs, but there is no good reason to do so.)

Shane32: Once everything is done we can write some complex mutations and see the performance impact, as well as verifying that there is little/no performance impact for code without directives.

Shane32:
> If I had not started doing this, I would not have found all the errors that I found. This required a design change, backwards incompatible changes. So it needs to be finished. Directives should work in v4.

I agree.

src/GraphQL/Types/Directives/LengthDirective.cs
Shane32: We should at least make it clear within the xml comments that the directive does not perform validation without adding the proper validation rule to the rule set.  Note that this rule will be required to run on cached queries also, so `ExecutionOptions.CachedDocumentValidationRules` needs to be set as well (if using caching).

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        private static readonly Task<INodeVisitor> _nodeVisitor = new NodeVisitors(
+            new MatchingNodeVisitor<Document>((document, context) =>
+            {
+                var operation = string.IsNullOrWhiteSpace(context.OperationName)
+                   ? context.Document.Operations.FirstOrDefault()
+                   : context.Document.Operations.WithName(context.OperationName);
+
+                //TODO: change to get single variable / rework validation framework for variables
+                ExecutionHelper.GetVariableValues(context.Document, context.Schema, operation?.Variables, context.Inputs, new VariableVisitor(context));
```
Shane32, line 84: We need to allow validation rules so add variable visitors to a list within the validation context, which the document executer will pass to the variable parser.  There's probably a memory-optimized way to do that, but in any case, it should not run `GetVariableValues` here.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
sungam3r: > this would only run if there was a length directive applied to a string type that is a variable reference, correct?

Yes.

src/GraphQL/Types/Directives/LengthDirective.cs
sungam3r: Done.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
sungam3r: @Shane32 Done. Today I was able to achieve this. The changes have had strong consequences. I have changed some of the tests. Along the way, I discovered erroneous tests. It's too late now, I'll do a PR tomorrow. We will need to discuss it.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
sungam3r: > it's possible that we would need 3 separate validation passes

Done. As a result, I came to the conclusion that it would be logical to build variables at the validation stage. Before my changes, they were created when building ExecutionContext right afted validation stage.  A side effect is now more validation rule triggers due to missing variables or setting them incorrectly. I had to edit a lot of tests. Variable errors were previously thrown as exceptions after direct validation. I just changed `throw` to `context.ReportError` inside `GetVariableValues`. I believe that overall the code has become better and more logical, despite the fact that now, at the stage of validation, new returned errors seem to be similar to the existing ones. Look at the tests, it will be clear there.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
sungam3r: Design flaws were mitigated.

src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
sungam3r: I updated PR.

src/GraphQL.Tests/Bugs/Bug1831.cs
```diff
@@ -35,8 +35,14 @@ public void TestVariableObject_InvalidType(string param)
                 Code = "KNOWN_TYPE_NAMES"
             };
             error2.AddLocation(1, 13);
-            var expected = CreateQueryResult(null, new ExecutionErrors { error1, error2 }, executed: false);
-            AssertQueryIgnoreErrors("query($arg: abcdefg) { test1 (arg: $arg) }", expected, inputs: $"{{ \"arg\": {param} }}".ToInputs(), expectedErrorCount: 2, renderErrors: true);
+            var error3 = new ValidationError(null, "5.6.1",
+               "Variable \u0027$arg\u0027 is invalid. Variable has unknown type \u0027abcdefg\u0027")
```
sungam3r, line 39: @joemcbride @Shane32 Compare this message with one from `KnownTypeNamesError.UnknownTypeMessage`. They are similar in meaning, but still different. One speaks about an unknown type as such, and the second about an incorrectly set variable of an unknown type. The context is different. I think this is more an advantage than a disadvantage. You will find similar new validation messages in many more tests.

src/GraphQL.Tests/Validation/DefaultValuesOfCorrectTypeTests.cs
```diff
@@ -23,7 +23,8 @@ public void required_variables_without_default_values()
               query RequiredValues($a: Int!, $b: String!) {
                 dog { name }
               }
-            ");
+            ",
+            "{ \"a\": 1, \"b\": \"\" }");
```
sungam3r, line 27: I had to add variables for all successful tests to keep them that way, since getting the variables is now part of the validation. Previously, the validation completed successfully, and the error would have occurred immediately after that at the time of building `ExecutionContext`.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -71,298 +69,6 @@ public static FieldType GetFieldDefinition(ISchema schema, IObjectGraphType pare
             return parentType.GetField(field.Name);
         }
 
-        /// <summary>
-        /// Returns all of the variable values defined for the operation from the attached <see cref="Inputs"/> object.
-        /// </summary>
-        public static Variables GetVariableValues(Document document, ISchema schema, VariableDefinitions variableDefinitions, Inputs inputs)
```
sungam3r, line 77: This code was moved into `ValidationContext`.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -447,7 +153,7 @@ public static ArgumentValue CoerceValue(IGraphType type, IValue input, Variables
             {
                 if (!(input is ObjectValue objectValue))
                 {
-                    return ArgumentValue.NullLiteral;
```
sungam3r, line 450: @Shane32 Why so? I believe this was a bug. See `fail_on_incorrect_value` test. I renamed it from `does_not_use_incorrect_value`. Now such query fails.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -142,74 +141,31 @@ public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options
...
-                        options.ThrowOnUnhandledException,
-                        options.UnhandledExceptionDelegate,
-                        options.MaxParallelExecutionCount,
-                        options.RequestServices);
-                }
-                catch (InvalidVariableError)
+                context = new ExecutionContext
                 {
-                    // error parsing variables
-                    // attempt to run AfterValidationAsync with null for the 'ExecutionContext.Variables' property
```
sungam3r, line 165: I think we don't need this block of code anymore.

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -13,10 +13,11 @@ namespace GraphQL.Validation
     public interface IDocumentValidator
     {
         /// <inheritdoc cref="IDocumentValidator"/>
-        Task<IValidationResult> ValidateAsync(
+        Task<(IValidationResult validationResult, Variables variables)> ValidateAsync(
```
sungam3r, line 16: @Shane32 I still doubt this way of returning values. `out` does not work for async methods.

sungam3r: To be honest, I generally doubt that an async/Task is needed here. I would generally make the method synchronous.

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -79,22 +93,20 @@ public class DocumentValidator : IDocumentValidator
             }
             else if (!rules.Any())
             {
-                return SuccessfullyValidatedResult.Instance;
+                var ctx = Initialize();
+                var variables = ctx.GetVariableValues(schema, operation?.Variables, inputs);
```
sungam3r, line 97: No variable visitors here.

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -107,25 +119,36 @@ public class DocumentValidator : IDocumentValidator
...
+                        return x.ValidateAsync(context);
+                    }).Where(x => x != null);
                     visitors = (await Task.WhenAll(awaitedVisitors)).ToList();
                 }
 
                 visitors.Insert(0, context.TypeInfo);
 
                 new BasicVisitor(visitors).Visit(document, context);
 
+                var variables = context.GetVariableValues(schema, operation?.Variables, inputs ?? Inputs.Empty, variableVisitors == null ? null : new CompositeVariableVisitor(variableVisitors));
```
sungam3r, line 139: Previously, this happened when building `ExecutionContext`.

src/GraphQL/Validation/IValidationRule.cs
```diff
@@ -11,6 +11,11 @@ public interface IValidationRule
         /// Prepares and returns a node visitor to be used to validate a document (via a node walker) against this
         /// validation rule. Validation failures are added then by this visitor to a list stored within <see cref="ValidationContext.Errors"/>.
         /// </summary>
-        Task<INodeVisitor> ValidateAsync(ValidationContext context);
+        Task<INodeVisitor> ValidateAsync(ValidationContext context); //TODO: make sync?
+    }
+
+    public interface IVariableValidation
```
sungam3r, line 17: @Shane32 @joemcbride Any better name?

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -204,31 +206,279 @@ public List<FragmentDefinition> GetRecursivelyReferencedFragments(Operation oper
...
+                    else if (variableDef.DefaultValue != null)
+                    {
+                        // if the variable was not specified in the inputs, and a default literal value was specified, use the specified default variable value
+
+                        // parse the variable literal via ParseLiteral (for scalars) and ParseDictionary (for objects) as applicable
+                        try
+                        {
+                            variable.Value = ExecutionHelper.CoerceValue(graphType, variableDef.DefaultValue, variables, null).Value;
+                        }
+                        catch (Exception ex)
```
sungam3r, line 256: @Shane32 Note that I added `try/catch` here.

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -204,31 +206,279 @@ public List<FragmentDefinition> GetRecursivelyReferencedFragments(Operation oper
...
+                }
+            }
+
+            // return the list of parsed variables defined for the operation
+            return variables;
+        }
+
+        private object ReportNullError(VariableDefinition variable, VariableName variableName)
+        {
+            ReportError(new InvalidVariableError(this, variable, variableName, "Received a null input for a non-null variable."));
```
sungam3r, line 281: Before these changes `throw new InvalidVariableError` was here.

sungam3r:
I look forward to further comments and suggestions.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -2725,11 +2743,17 @@ namespace GraphQL.Validation
     {
         public static readonly System.Collections.Generic.IEnumerable<GraphQL.Validation.IValidationRule> CoreRules;
         public DocumentValidator() { }
-        public System.Threading.Tasks.Task<GraphQL.Validation.IValidationResult> ValidateAsync(string originalQuery, GraphQL.Types.ISchema schema, GraphQL.Language.AST.Document document, System.Collections.Generic.IEnumerable<GraphQL.Validation.IValidationRule> rules = null, System.Collections.Generic.IDictionary<string, object> userContext = null, GraphQL.Inputs inputs = null) { }
+        [return: System.Runtime.CompilerServices.TupleElementNames(new string[] {
+                "validationResult",
+                "variables"})]
+        public System.Threading.Tasks.Task<System.ValueTuple<GraphQL.Validation.IValidationResult, GraphQL.Language.AST.Variables>> ValidateAsync(string originalQuery, GraphQL.Types.ISchema schema, GraphQL.Language.AST.Document document, GraphQL.Language.AST.Operation operation, System.Collections.Generic.IEnumerable<GraphQL.Validation.IValidationRule> rules = null, System.Collections.Generic.IDictionary<string, object> userContext = null, GraphQL.Inputs inputs = null) { }
```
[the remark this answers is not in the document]
Shane32: I looked through the rest of the code.  It seems the operation is only used to retrieve the `Variables` from it.  I think we should change the signature of `ValidateAsync` to include the variables directly and not include the `Operation`.  It will be clearer that we are validating given a set of variables, and less likely for any future changes to the context to accidentally be based on the selected operation.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I agree.

src/GraphQL/Validation/DocumentValidator.cs
Shane32: Hmm.  I'd tend to agree, given the rest of the design of the validation code.  However, could this be used to asynchronously retrieve some information from a database prior to the validation?  Let's say someone wanted the validation code to be based on the permission set of the authenticated user (JWT/cookie/querystring/etc).  And assume that an async call is necessary for some reason to retrieve the permissions, whether it's from a redis cache or a database.  Would this be how it's done?  Is there another way to do it?  I think this is a single `Task` - a single allocation - per request, and as such, I'd probably leave it as-is.

Shane32 (commented):
I've been reviewing this over the last couple days, so hopefully all my comments make sense!

src/GraphQL/Validation/DocumentValidator.cs
Shane32: I think that I liked `GetVariableValues` back in `ExecutionHelper` better.  But it's okay either way.

Shane32: We should consider that with document caching, now a validation context is required even though no rules are executed (correct?).  👎   At least with context reuse this can be minimized.

src/GraphQL/Validation/DocumentValidator.cs
Shane32: I guess maybe a few allocations - one for each validation rule.

src/GraphQL.ApiTests/GraphQL.approved.txt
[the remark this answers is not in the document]
sungam3r: > Adding the operation to the validation context can break document caching

How?

> to include the variables directly and not include the Operation

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -204,31 +206,279 @@ public List<FragmentDefinition> GetRecursivelyReferencedFragments(Operation oper
...
+                {
+                    ret = scalarGraphType.ParseValue(value);
+                }
+                catch (Exception ex)
+                {
+                    throw new InvalidVariableError(this, variable, variableName, $"Unable to convert '{value}' to '{scalarGraphType.Name}'", ex);
+                }
+
+                if (ret == null)
+                    throw new InvalidVariableError(this, variable, variableName, $"Unable to convert '{value}' to '{scalarGraphType.Name}'");
```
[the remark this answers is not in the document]
sungam3r: You are right. Fixed.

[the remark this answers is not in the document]
sungam3r: > Also, what if it adds the error to the list and returns null, and then further assume this is an integer field of a object -- now ParseDictionary could throw an error because null is invalid for the field.

now this will not happen

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: I understand, I just grumbled a little about the asynchronous code, which is 99% synchronous.

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: > I think that I liked GetVariableValues back in ExecutionHelper better

I need access to the `ValidationContext` to be able to call `ReportError`. Code from `ExecutionHelper` just throws exceptions.

> now a validation context is required even though no rules are executed (correct?)

Yes. This was a necessary measure because it was necessary to have access to `ReportError`. In principle, it would be possible to leave the methods static, but then it would require working with a large number of variables in order to accumulate errors.

src/GraphQL.ApiTests/GraphQL.approved.txt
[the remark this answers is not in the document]
Shane32: Let's say `Operation` is available for validation rules to operate against.  Let's assume there is a simple custom validation rule: "if operation type is query then pass else fail".  When document caching is used, and if a document with a query operation (no variables) is validated, the document will be cached.  If the same document is executed again with a mutation operation (no variables), the document will be pulled from the cache and assumed to be valid.

Now I understand:
1. Currently the `Operation` is not cached in the validation context and hence not accessible by a validation rule (although it is available by a `IDocumentValidator` implementation).
2. The rule can be added to the `CachedDocumentValidationRules` to alleviate this issue.

However, the purpose of `Operation` being passed is simply to retrieve the variables, so I suggested that it would be better to do that prior to the `DocumentValidator` execution.

src/GraphQL/Validation/DocumentValidator.cs
Shane32: As indicated elsewhere, I do not think `GetVariableValue` should call `ReportError` at all.  Exceptions should be handled within `GetVariableValues` and `ReportError` handled from there, so any error will immediately invalidate the entire variable.

I think `GetVariableValue` can return to being a static method, not that it matters much.

As a side effect, this allows us the option of modifying/cloning `GetVariableValues` (which is not much code) so that it can be used without a `ValidationContext`.

Shane32:
I see now you have added boolean returns and out variables to all of the Parse methods.  There is no reason for this.  It worked perfectly with try..catch before with no change in functionality.  It should be reverted.  `GetVariableValues` should have a try..catch to catch `ValidationError`s and call `ReportError`.

src/GraphQL.ApiTests/GraphQL.approved.txt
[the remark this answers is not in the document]
sungam3r: >  it would be better to do that prior to the DocumentValidator execution.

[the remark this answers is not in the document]
sungam3r: There is a chicken and egg problem.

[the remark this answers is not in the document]
Shane32: I just meant to pass the variables directly in. operation.Variables instead of operation

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: > Exceptions should be handled within GetVariableValues

Handled _how_ ? `try/catch` with capturing throwed exception into `ReportError` call ? It doesn't look very good.

sungam3r:
This makes sense if the `GetVariableValue` method has other consumers. It is now `private`. Throwing an exception is an expensive operation. Too costly to just catch immediately this exception and save it to the collection. Such flow control scenarios should be avoided. If you can just save exception object there, then you should do it.

src/GraphQL/Validation/DocumentValidator.cs
Shane32: I think it’s easier to understand than having every method having Boolean return values and out parameters. It’s very clear what’s going on. An error occurs. It is caught and added to the validation error list — exactly the same way that would happen for a field resolver.

src/GraphQL.ApiTests/GraphQL.approved.txt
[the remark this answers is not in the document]
sungam3r: Oooooooops 👍

Shane32:
> This makes sense if the `GetVariableValue` method has other consumers. It is now `private`. Throwing an exception is an expensive operation. Too costly to just catch immediately this exception and save it to the collection. Such flow control scenarios should be avoided. If you can just save exception object there, then you should do it.

Not applicable. This is an error situation. Not hot path. Perfectly acceptable to have better code over faster code.  Everywhere else we have exceptions thrown for error conditions. This is no different. And if it allows us to have faster code for cached queries, then it’s faster also.

src/GraphQL.ApiTests/GraphQL.approved.txt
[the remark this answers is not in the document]
sungam3r: Done.

sungam3r:
>  Everywhere else we have exceptions thrown for error conditions. 

Validation stage differs from execution stage. The `ReportError` method was specially made for it. Validation aggregates all errors found. For the execution stage, it is quite normal to simply throw an error, thereby interrupting the control flow and returning it to its original point. By the way, it is quite easy to change the exception throwing code at execution stage to call `REsolveFieldContext.Errors.Add`. Technically, result will be the same, just without throwing an exception.

> And if it allows us to have faster code for cached queries, then it’s faster also.

?

Shane32:
Right but in this case we don’t want to aggregate the errors for a single variable. We want it to cancel out.

Shane32:
If it doesn’t have to create a validation context it can be faster. And the validation context is a very large object.

sungam3r:
Okay, I get the idea and I'll see if it works out better if I return throwing exceptions. 4 am now, I'm done.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -120,12 +117,14 @@ public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options
...
                 {
-                    validationResult = await _documentValidator.ValidateAsync(
+                    (validationResult, variables) = await _documentValidator.ValidateAsync(
                         options.Query,
                         options.Schema,
                         document,
+                        operation.Variables,
                         validationRules,
                         options.UserContext,
                         options.Inputs);
```
Shane32, line 130: What do you think about the order of the parameters?  Might it be better if Variables and Inputs were the last two?  It doesn't really matter of course, so whatever you think is fine.

sungam3r: изображение

Shane32: Yeah...  of course if inputs is null variables could be null. But sure, it’s fine as is.

sungam3r: I removed originalQuery.

sungam3r:
> If it doesn’t have to create a validation context it can be faster. And the validation context is a very large object.
> Right but in this case we don’t want to aggregate the errors for a single variable. We want it to cancel out.

src/GraphQL/Validation/Errors.Custom/InputFieldsAndArgumentsOfCorrectLengthError.cs
sungam3r: What will we decide?

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: OK. Done. 

> so that it can be used without a `ValidationContext`

Nope, eventually `IVariableVisitor` needs `ValidationContext`.

src/GraphQL/Validation/IValidationRule.cs
sungam3r: Pending question.

src/GraphQL/Validation/ValidationContext.cs
sungam3r: Now it is aligned with try/catch around `GetVariableValue`

sungam3r:
I'm done here for now. Waiting for further comments.

Shane32:
Sorry I couldn’t look at this PR today. Might take me another day or two before I can look at it.

sungam3r:
@Shane32 I think I've added most of the directives documentation I should have done. In the process of writing it, I fixed a few more bugs. PR is very large, alas.

docs2/site/docs/getting-started/directives.md
```diff
@@ -71,6 +71,203 @@ when the same directive should be used with different arguments at a single loca
...
+        RegisterVisitor(new UppercaseDirectiveVisitor());
+        // or
+        RegisterVisitor(typeof(UppercaseDirectiveVisitor));
+    }
+}
+```
+
+Note that a schema visitor, unlike a directive, can be registered not only as an instance but also as
+a type. In this case, when initializing the schema, schema visitor will be created according to how
+you configure the DI container. In other words, schema visitors support dependency injection.
```
Shane32, line 207: Add note about how schema visitors, when created via DI, have same lifetime as schema, and refer to dependency injection notes in middleware guide.

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -61,40 +61,61 @@ public class DocumentValidator : IDocumentValidator
...
+
             if (useOnlyStandardRules)
             {
                 rules = CoreRules;
             }
             else if (!rules.Any())
             {
-                return SuccessfullyValidatedResult.Instance;
+                context = Initialize();
+                try
```
Shane32, line 96: I think the `finally` block is the same.  Can we move the `if (!rules.Any())` inside the `try` block and then eliminate the duplicate try..catch block and inline `MakeResult` so it reads easier?

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -107,27 +128,43 @@ public class DocumentValidator : IDocumentValidator
...
 
                 visitors.Insert(0, context.TypeInfo);
 
                 new BasicVisitor(visitors).Visit(document, context);
 
-                success = !context.HasErrors;
-                return success
-                    ? SuccessfullyValidatedResult.Instance
-                    : new ValidationResult(context.Errors);
+                var variables = context.GetVariableValues(schema, variableDefinitions, inputs ?? Inputs.Empty, variableVisitors == null ? null : new CompositeVariableVisitor(variableVisitors));
```
Shane32, line 148: I know performance along this path isn't critical, but it would be nice if it skipped initialization of `CompositeVariableVisitor` when there was only a single `variableVisitor`

src/GraphQL/Validation/Errors.Custom/InputFieldsAndArgumentsOfCorrectLengthError.cs
Shane32: No opinion.  If we have worked out all the "bugs", then I'd probably leave it in the main project for now.  It may be useful, and at least it serves as an excellent sample.

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -34,11 +35,6 @@ internal void Reset()
             Inputs = null;
         }
 
-        /// <summary>
-        /// Returns the original GraphQL query string.
-        /// </summary>
-        public string OriginalQuery { get; set; }
```
Shane32, line 40: This could have been changed to a property, so as to not break existing applications, but I'm okay to remove it.
```cs
public string OriginalQuery => Document.OriginalQuery;
```

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -204,31 +200,277 @@ public List<FragmentDefinition> GetRecursivelyReferencedFragments(Operation oper
...
+                        catch (Exception ex)
+                        {
+                            ReportError(new InvalidVariableError(this, variableDef, variableDef.Name, "Error coercing default value.", ex));
+                            continue;
+                        }
+                        variable.IsDefault = true;
+                    }
+                    else if (graphType is NonNullGraphType)
+                    {
+                        ReportError(new InvalidVariableError(this, variableDef, variable.Name, "Received a null input for a non-null variable."));
```
Shane32, line 267: Is this error more specifically that no variable value was provided for a non-null variable?
```suggestion
                        ReportError(new InvalidVariableError(this, variableDef, variable.Name, "No value provided for a non-null variable."));
```

Shane32 (commented):
Looking good!  I've written down a few final comments.

Shane32:
@sungam3r Take a look at my comments and/or revise, then go ahead and merge if there's no other changes you think I should review. 👍

docs2/site/docs/getting-started/directives.md
sungam3r: Done. Also see #2299

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: Done

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: Done

src/GraphQL/Validation/Errors.Custom/InputFieldsAndArgumentsOfCorrectLengthError.cs
sungam3r: Then I will keep it here for now.

src/GraphQL/Validation/ValidationContext.cs
sungam3r: I remove such properties if I see they go against the single source of truth.

src/GraphQL/Validation/ValidationContext.cs
sungam3r: 👍 

The same for 
>  // RULE: If no default value is provided and the input object field’s type is non‐null,
                        // an error should be thrown.
                        throw new InvalidVariableError(this, variableDef, childFieldVariableName, "Received a null input for a non-null variable.");


?

sungam3r: Done for both.

sungam3r:
1. https://github.com/graphql-dotnet/graphql-dotnet/pull/2276#discussion_r577238887
2. https://github.com/graphql-dotnet/graphql-dotnet/pull/2276#discussion_r579315300
3. https://github.com/graphql-dotnet/graphql-dotnet/pull/2276#discussion_r579294295

src/GraphQL/Validation/ValidationContext.cs
Shane32: Yes, correct.

Shane32 (approved):

src/GraphQL/Validation/IValidationRule.cs
sungam3r: Changed to IVariableVisitorProvider.

Shane32 (approved):

sungam3r:
I will keep this PR open for another day, no more. Maybe someone that I mentioned in the related issues will have comments.

docs2/site/docs/getting-started/directives.md
```diff
@@ -16,3 +18,463 @@ query HeroQuery($id: ID, $withFriends: Boolean!) {
...
+                {
+                    Name = "id"
+                }
+                .ApplyDirective("length", "min", 2, "max", 5)
+            ));
+    }
+}
+```
+
+> Above was an example of how to create and how to apply the `@length` directive. Also, for the directive
```
mikelneo, line 151: Probably `>` not needed here.

sungam3r: done

lorenz068:
Hi @sungam3r
I'm interested by this feature. Do you know when it will be released?
Thank you

sungam3r:
At first we wanted to make a release on February 1, then postponed for a week, then a lot of work appeared, as a result, it was all stretched for a month. See https://github.com/graphql-dotnet/graphql-dotnet/issues?q=is%3Aopen+is%3Aissue+milestone%3A4.0
All issues within v4 milestone should be done. We are working on it every day. But it turns out that on some day we can detect an error in an unexpected place. And then we get more work. In any case, the countdown is already on the days.

PR #2283, opened 2021-02-14
GraphTypeTypeRegistry rework

### LINKED ISSUE #2189: GraphTypeTypeRegistry.Register design flaw
I ran into one error while working on #2102. One of the tests started to fail - auto_register_object_graph_type but only as part of the rest of the tests. A separate launch works fine. In the end, I found out the reason. It's funny, but it turned out that this is one of the tasks I had postponed, which was discovered long ago. See `auto_register_object_graph_type` and `auto_register_input_object_graph_type`. The first registers `Money` as `AutoRegisteringObjectGraphType`, the second does the same but as `AutoRegisteringInputObjectGraphType`. `GraphTypeTypeRegistry.Register` rewrites stored value. When tests are run in parallel, the state changes, which leads to an error. But the problem seems to me wider. Now the design of the library does not provide for the registration of the same type at the same time as input and output.

sungam3r:
Fixes #2189

This work started just as additional test for ##2275 to make sure GraphQL.NET functions properly. I've changed the error messages to make them more meaningful. And then, quite by accident, I came across an error that I have known for a long time - #2189.  I must admit that it took me a long time to make a series of corrections.


sungam3r:
Ping @Frost84

Shane32:
I would like to review this, but it looks complicated, so it will probably have to be this weekend.

src/GraphQL/GlobalSwitches.cs
```diff
@@ -54,5 +55,23 @@ public static class GlobalSwitches
         /// member will be unusable.
         /// </summary>
         public static Action<string, NamedElement> ValidationOnSchemaInitialize = NameValidator.ValidateDefault;
+
+        /// <summary>
+        /// This setting improves performance if your schema uses only scalar types or types marked with
```
Shane32, line 60: This only improves performance during schema build time, correct?  I think that is important to note here, as there is no practical reason to turn it off unless you have a scoped schema.  Can we update the note to indicate as such?

src/GraphQL/SchemaExtensions.cs
```diff
@@ -205,4 +259,12 @@ public enum ExperimentalIntrospectionFeaturesMode
...
         IntrospectionAndExecution
     }
+
+    [Flags]
+    public enum AutoRegisteringMode
+    {
+        Input = 1,
+        Output = 2,
+        Both = Input | Output
+    }
```
Shane32, line 269: Needs xml comments for the enum and each member.

src/GraphQL/SchemaExtensions.cs
```diff
@@ -205,4 +259,12 @@ public enum ExperimentalIntrospectionFeaturesMode
         /// </summary>
         IntrospectionAndExecution
     }
+
+    [Flags]
+    public enum AutoRegisteringMode
+    {
+        Input = 1,
+        Output = 2,
+        Both = Input | Output
```
Shane32, line 268:
```suggestion
        Both = Input | Output,
```

src/GraphQL/TypeExtensions.cs
```diff
@@ -290,4 +312,22 @@ public static object DefaultValue(this MemberInfo memberInfo)
                 : null;
         }
     }
+
+    /// <summary>
+    /// Mode to use when mapping CLR type to GraphType.
+    /// </summary>
+    public enum TypeMappingMode
+    {
+        UseBuiltInScalarMappings,
```
Shane32, line 321: Missing xml comment

src/GraphQL/TypeExtensions.cs
```diff
@@ -290,4 +312,22 @@ public static object DefaultValue(this MemberInfo memberInfo)
...
+
+        /// <summary>
+        /// Map type as input type.
+        /// </summary>
+        InputType,
+
+        /// <summary>
+        /// Map type as output type.
+        /// </summary>
+        OutputType
```
Shane32, line 331:
```suggestion
        OutputType,
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -35,18 +33,13 @@ public AutoRegisteringObjectGraphType(params Expression<Func<TSourceType, object
         /// <summary>
         /// Returns a list of properties that should have fields created for them.
         /// </summary>
-        protected virtual IEnumerable<PropertyInfo> GetRegisteredProperties()
-        {
-            return typeof(TSourceType)
-                .GetProperties(BindingFlags.Public | BindingFlags.Instance)
-                .Where(p => AutoRegisteringHelper.IsEnabledForRegister(p.PropertyType, true));
```
Shane32, line 42: I see the `Where` portion is gone from a number of these methods.  Was that functionality moved somewhere else?  I don't actually use this class so not real familiar with this.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -501,6 +535,53 @@ private void AddTypeIfNotRegistered(IGraphType type, TypeCollectionContext conte
             }
         }
 
+        private object RebuildType(Type type, bool input, List<(Type, Type)> typeMappings)
```
Shane32, line 538: This looks like an ugly hack -- returning an error as a string or a value as a `Type`.  How about returning an error as a string, or null if no error, with an out variable for the type.  Or return a bool with two out variables.  Or return a tuple with 2/3 variables.  But not an object that could be either a `string` or a `Type`.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -501,6 +535,53 @@ private void AddTypeIfNotRegistered(IGraphType type, TypeCollectionContext conte
...
+            {
+                return GetGraphType(type.GetGenericArguments()[0], input, typeMappings);
+            }
+            else
+            {
+                return type;
+            }
+        }
+
+        private object GetGraphType(Type clrType, bool input, List<(Type clr, Type graph)> typeMappings)
```
Shane32, line 563: ditto

src/GraphQL/Types/ISchema.cs
```diff
@@ -110,6 +110,19 @@ public interface ISchema : IProvideMetadata, IProvideDescription
...
+        /// <br/>
+        /// These mappings are used for type inference when constructing fields using expressions:
+        /// <br/>
+        /// <c>
+        /// Field(x => x.Filters);
+        /// </c>
+        /// </summary>
+        /// <param name="clrType">The CLR property type from which to infer the GraphType.</param>
+        /// <param name="graphType">Inferred GraphType.</param>
+        void RegisterTypeMapping(Type clrType, Type graphType);
```
Shane32, line 124: I see we have removed all the `GraphTypeTypeRegistry` methods and replaced it with a single method here on `ISchema`, but no way to enumerate those registrations.  I think we need `AllTypeRegistrations` or something also.  My code is extremely reliant on the `GraphTypeTypeRegistry`, and I know I will have some code that will need to do something like this:

```
if not already registered clr type x, then register clr type x with graph type y
```

Shane32 (changes requested):

Shane32:
In general I have no problem with the changes here.  I like that the registrations are per-schema rather than static (per-application).

docs2/site/docs/migrations/migration4.md
```diff
@@ -60,6 +60,53 @@ For untyped `InputObjectGraphType` classes, like shown above, the default behavi
 will be to return the dictionary. `GetArgument<T>` will still attempt to convert a dictionary to the
 requested type via `ObjectExtensions.ToObject` as it did before.
 
+### Ability to map CLR types to GraphTypes
```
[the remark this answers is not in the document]
sungam3r: Done. I tried to write the description as detailed as possible.

docs2/site/docs/migrations/migration4.md
```diff
@@ -339,3 +386,4 @@ lock (field)
 * `IDocumentValidator.ValidateAsync` does not take `originalQuery` parameter; use `Document.OriginalQuery` instead
 * `IDocumentValidator.ValidateAsync` now returns `(IValidationResult validationResult, Variables variables)` tuple instead of single `IValidationResult` before
 * `IResolveFieldContext.FieldName` and `IResolveFieldContext.ReturnType` properties have been removed, use `IResolveFieldContext.FieldAst.Name` and `IResolveFieldContext.FieldDefinition.ResolvedType` instead
+* `GraphQLMetadataAttribute.Type` property was renamed to `ResolverType`
```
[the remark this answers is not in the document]
sungam3r: Done.

src/GraphQL/GlobalSwitches.cs
sungam3r: Done

docs2/site/docs/migrations/migration4.md
```diff
@@ -278,6 +278,149 @@ lock (field)
...
+```
+
+In this case, a call to the registration method is not required, since the schema
+will use information from the provided attribute.
+
+> Keep in mind that you can register type mappings even for built-in/primitive types if you want to change their behavior:
+> <br/><br/>
+> schema.RegisterTypeMapping<int, MyIntGraphType>()
+> <br/>
+> schema.RegisterTypeMapping<string, MySpecialFormattedStringGraphType>()
```
Shane32, line 411: Needs to be properly formatted.

src/GraphQL/SchemaExtensions.cs
sungam3r: Done

src/GraphQL/TypeExtensions.cs
sungam3r: fixed

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: This code is now simply not needed. Previously, it could filter some properties when it thought they could not be matched. Even if this happens, it means that the user must add a type mapping. It is better to tell him about this through an exception than to quietly exclude a property so that he does not waste time figuring out why some property does not exist in autogenerated graph type.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: Believe me, I did it first. It turns out to be even worse.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: The situation is the same here, because the result is simply passed up to `RebuildType` and further.

src/GraphQL/Types/ISchema.cs
sungam3r: See `EnumerateTypeMappings`.

docs2/site/docs/migrations/migration4.md
sungam3r: how?

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Then we should add this to the migration notes, saying that previously, unmatched properties would be ignored, and now an exception will be thrown.

src/GraphQL/Types/ISchema.cs
```diff
@@ -110,6 +110,25 @@ public interface ISchema : IProvideMetadata, IProvideDescription
...
+
+        /// <summary>
+        /// Returns all registered by <see cref="RegisterTypeMapping"/> type mappings.
+        /// </summary>
+        /// <param name="includeBuiltInScalarMappings">Is it required to return built-in mappings for scalars.</param>
+        IEnumerable<(Type clrType, Type graphType)> EnumerateTypeMappings(bool includeBuiltInScalarMappings = false);
```
Shane32, line 130: Can we do better?  The other collections properties on the schema are: `Directives`, `AllTypes`, `AdditionalTypes`.  Why does this have to be a method, with an argument, prefixed with `Enumerate`?  How about simply:

```cs
IEnumerable<(Type clrType, Type graphType)> TypeMappings { get; }
```

It can return built in scalar mappings (unless they have been overridden).  Or we can choose to not include built-in mappings.  Or we can add another property:

```cs
IEnumerable<(Type clrType, Type graphType)> BuiltInTypeMappings { get; }
```

src/GraphQL/Types/Schema.cs
```diff
@@ -350,6 +374,7 @@ IEnumerable<ISchemaNodeVisitor> GetVisitors()
 
             _allTypes = SchemaTypes.Create(
                 GetTypes(),
+                 _clrToGraphTypeMappings,
```
Shane32, line 377:
```suggestion
                _clrToGraphTypeMappings,
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: done

src/GraphQL/Types/ISchema.cs
sungam3r: done

docs2/site/docs/migrations/migration4.md
Shane32:
```suggestion
Keep in mind that you can register type mappings even for built-in/primitive types if you want to change their behavior:

```csharp
schema.RegisterTypeMapping<int, MyIntGraphType>()
schema.RegisterTypeMapping<string, MySpecialFormattedStringGraphType>()
```

(needs closing triple back quotes also)

Or was it supposed to look like a quote?

docs2/site/docs/migrations/migration4.md
```diff
@@ -278,6 +278,165 @@ lock (field)
...
+the type mapping with `ISchema.RegisterTypeMapping` method. Alternatively, you can inherit from these classes and
+override the `GetRegisteredProperties` method. Consider the following example:
+
+```csharp
+public class MyAutoType : AutoRegisteringObjectGraphType<SomeClassWithManyProperties>
+{
+    protected override IEnumerable<PropertyInfo> GetRegisteredProperties() => typeof(SomeClassWithManyProperties)
+        .GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(p => Attribute.IsDefined(p, typeof(ForExportAttribute)));
+}
+```
```
Shane32, line 427: This explains current/new behavior very well, and how the `GetRegisteredProperties` method works, but doesn't explain why it's a breaking change, or how to fix old code so it will compile/run under v4.  It should say that previously it would skip unmatched properties, and then provide a sample of how to fix old code.  In this case, the sample should show how to pass the unmatchable property into the `excludedProperties` field of the constructor.  That, I would think, is the simplest fix for someone with properties that should not be matched.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Well...at least it's private.  Just feels wrong.

docs2/site/docs/migrations/migration4.md
sungam3r: done

docs2/site/docs/migrations/migration4.md
sungam3r: done

sungam3r: After the v4 release, we will need to copy all of this (with some edits) into the main documentation.

docs2/site/docs/migrations/migration4.md
```diff
@@ -414,10 +414,20 @@ schema.RegisterTypeMapping<string, MySpecialFormattedStringGraphType>();
...
-classes by default use all properties from the provided `TSourceType` to generate GraphType's fields. If no matching
-is found for some of the properties, then an exception will be thrown during schema initialization. You need to register
-the type mapping with `ISchema.RegisterTypeMapping` method. Alternatively, you can inherit from these classes and
-override the `GetRegisteredProperties` method. Consider the following example:
+classes by default use all properties from the provided `TSourceType` to generate GraphType's fields (previously they
+may skip unmatched properties). If no matching is found for some of the properties, then an exception will be thrown
+during schema initialization.
+
+You have three options to fix this.
+
```
Shane32, line 422: Four? Multiple?

Shane32:
Is that it? I’ll review again when I’m home.

docs2/site/docs/migrations/migration4.md
sungam3r: 🤦

sungam3r:
There are no comments left on which I have not made fixes.

src/GraphQL/Types/Schema.cs
```diff
@@ -249,6 +249,37 @@ public void RegisterTypes(params Type[] types)
...
+        /// <inheritdoc/>
+        public IEnumerable<(Type clrType, Type graphType)> TypeMappings
+        {
+            get
+            {
+                if (_clrToGraphTypeMappings != null)
+                {
+                    foreach (var (clrType, graphType) in _clrToGraphTypeMappings)
+                        yield return (clrType, graphType);
+                }
```
Shane32, line 269: Does this work?

```suggestion
                if (_clrToGraphTypeMappings != null)
                {
                    return _clrToGraphTypeMappings;
                }
                else
                {
                    return Enumerable.Empty<(Type, Type)>();
                }
```

src/GraphQL/Types/Schema.cs
```diff
@@ -249,6 +249,37 @@ public void RegisterTypes(params Type[] types)
...
+            }
+        }
+
+        /// <inheritdoc/>
+        public IEnumerable<(Type clrType, Type graphType)> BuiltInTypeMappings
+        {
+            get
+            {
+                foreach (var pair in SchemaTypes.BuiltInScalarMappings)
+                    yield return (pair.Key, pair.Value);
```
Shane32, line 279: Does this work?

```suggestion
                return SchemaTypes.BuiltInScalarMappings;
```

Shane32 (approved):
Two minor comments; merge as you see fit.

src/GraphQL/Types/Schema.cs
Shane32: Keep in mind that this change allows optimizations by LINQ methods for any LINQ queries executed against the returned IEnumerable.

sungam3r: OK

PR #2284, opened 2021-02-14
A little note about traces

### LINKED ISSUE #1839: Detailed performance data with Azure Application Insights
##  Summary
Azure Application Insights automatically shows the performance of REST API's per endpoint (ex. /users, /settings/general, etc...) but with graphQL we only get one endpoint (/graphql) which makes sense, but has anyone found an easy way to track performance per GraphQL query?

For now my plan is to manually make custom telemetry measurements for each query type that I have. Which will be quite long to make. Do you know of any existing library or do you have a more efficient way?

Thanks!
If I find a better way I'll add it here below.

sungam3r:
fixes #1839

### DIFF
--- a/docs2/site/docs/getting-started/metrics.md
+++ b/docs2/site/docs/getting-started/metrics.md
@@ -1,6 +1,10 @@
 # Metrics
 
-Metrics are captured during execution.  This can help you determine performance issues within a resolver or validation.  Field metrics are captured using Field Middleware and the results are returned as a `PerfRecord` array on the `ExecutionResult`.  You can then generate [Apollo Tracing](https://github.com/apollographql/apollo-tracing) data with the `EnrichWithApolloTracing()` extension method.
+Metrics are captured during execution. This can help you determine performance issues within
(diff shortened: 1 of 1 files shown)


docs2/site/docs/getting-started/metrics.md
```diff
@@ -1,6 +1,10 @@
 # Metrics
 
-Metrics are captured during execution.  This can help you determine performance issues within a resolver or validation.  Field metrics are captured using Field Middleware and the results are returned as a `PerfRecord` array on the `ExecutionResult`.  You can then generate [Apollo Tracing](https://github.com/apollographql/apollo-tracing) data with the `EnrichWithApolloTracing()` extension method.
+Metrics are captured during execution. This can help you determine performance issues within
```
Shane32, line 3: > Metrics are captured during execution.

Should this sentence be updated now that `EnableMetrics` defaults to `false`?

PR looks good otherwise.

sungam3r: done

Shane32 (approved):

PR #2293, opened 2021-02-16
Add ScalarGraphType.CanParseXXX methods

### LINKED ISSUE #2292: GraphQL.Execution.UnhandledError
I mistakenly defined a string field (name) as decimal in my input type

```csharp    
public partial class DepartmentInputType : InputObjectGraphType
    {
        public DepartmentInputType()
        {
            Name = "DepartmentInput";
            Field<GuidGraphType>("uuid");
            Field<StringGraphType>("code");
            Field<NonNullGraphType<DecimalGraphType>>("name");
            Field<StringGraphType>("description");
            Field<StringGraphType>("comments");
        }
    }
```
and got the following error when setting *name* with a string value

```json
{
  "errors": [
    {
      "message": "GraphQL.Execution.UnhandledError: Error executing document.\r\n ---> System.FormatException: Input string was not in a correct format.\r\n   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)\r\n   at System.Number.ParseDecimal(ReadOnlySpan`1 value, NumberStyles styles, NumberFormatInfo info)\r\n   at System.Decimal.Parse(String s, NumberStyles style, IFormatProvider provider)\r\n   at GraphQL.ValueConverter.<>c.<.cctor>b__1_11(String value) in /_/src/GraphQL/Conversion/ValueConverter.cs:line 39\r\n   at GraphQL.ValueConverter.<>c__DisplayClass7_0`2.<Register>b__0(Object v) in /_/src/GraphQL/Conversion/ValueConverter.cs:line 224\r\n   at GraphQL.ValueConverter.TryConvertTo(Object value, Type targetType, Object& result, Type sourceType) in /_/src/GraphQL/Conversion/ValueConverter.cs:line 184\r\n   at GraphQL.ValueConverter.ConvertTo(Object value, Type targetType) in /_/src/GraphQL/Conversion/ValueConverter.cs:line 158\r\n   at GraphQL.Types.DecimalGraphType.ParseValue(Object value) in /_/src/GraphQL/Types/Scalars/DecimalGraphType.cs:line 25\r\n   at GraphQL.Types.DecimalGraphType.ParseLiteral(IValue value) in /_/src/GraphQL/Types/Scalars/DecimalGraphType.cs:line 16\r\n   at GraphQL.GraphQLExtensions.IsValidLiteralValue(IGraphType type, IValue valueAst, ISchema schema) in /_/src/GraphQL/GraphQLExtensions.cs:line 283\r\n   at GraphQL.GraphQLExtensions.IsValidLiteralValue(IGraphType type, IValue valueAst, ISchema schema) in /_/src/GraphQL/GraphQLExtensions.cs:line 213\r\n   at GraphQL.GraphQLExtensions.IsValidLiteralValue(IGraphType type, IValue valueAst, ISchema schema)\r\n   at GraphQL.GraphQLExtensions.IsValidLiteralValue(IGraphType type, IValue valueAst, ISchema schema) in /_/src/GraphQL/GraphQLExtensions.cs:line 213\r\n   at GraphQL.Validation.Rules.ArgumentsOfCorrectType.<>c__DisplayClass1_0.<ValidateAsync>b__1(Argument argAst) in /_/src/GraphQL/Validation/Rules/ArgumentsOfCorrectType.cs:line 33\r\n   at GraphQL.Validation.EnterLeaveListener.GraphQL.Validation.INodeVisitor.Enter(INode node) in /_/src/GraphQL/Validation/EnterLeaveListener.cs:line 34\r\n   at GraphQL.Validation.BasicVisitor.Visit(INode node) in /_/src/GraphQL/Validation/BasicVisitor.cs:line 41\r\n   at GraphQL.Validation.BasicVisitor.Visit(INode node) in /_/src/GraphQL/Validation/BasicVisitor.cs:line 50\r\n   at GraphQL.Validation.BasicVisitor.Visit(INode node) in /_/src/GraphQL/Validation/BasicVisitor.cs:line 55\r\n   at GraphQL.Validation.BasicVisitor.Visit(INode node) in /_/src/GraphQL/Validation/BasicVisitor.cs:line 50\r\n   at GraphQL.Validation.BasicVisitor.Visit(INode node) in /_/src/GraphQL/Validation/BasicVisitor.cs:line 55\r\n   at GraphQL.Validation.BasicVisitor.Visit(INode node) in /_/src/GraphQL/Validation/BasicVisitor.cs:line 50\r\n   at GraphQL.Validation.MyDocumentValidator.ValidateAsync(String originalQuery, ISchema schema, Document document, IEnumerable`1 rules, IDictionary`2 userContext, Inputs inputs) in C:\\code\\dev\\live\\Meercat.RiskView\\Ehs.Risk.GraphQL\\MyDocumentValidator.cs:line 84\r\n   at Ehs.Risk.GraphQL.MyDocumentExecuter.ExecuteAsync(ExecutionOptions options) in C:\\code\\dev\\live\\Meercat.RiskView\\Ehs.Risk.GraphQL\\MyDocumentExecuter.cs:line 111\r\n   --- End of inner exception stack trace ---",
      "extensions": {
        "code": "FORMAT",
        "codes": [
          "FORMAT"
        ]
      }
    }
  ],
  "extensions": {
    "tracing": {
      "version": 1,
      "startTime": "2021-02-16T19:22:48.6791269Z",
      "endTime": "2021-02-16T19:22:50.5831269Z",
      "duration": 1903948599,
      "parsing": {
        "startOffset": 4200,
        "duration": 32599
      },
      "validation": {
        "startOffset": 38600,
        "duration": 1903875299
      },
      "execution": {
        "resolvers": []
      }
    }
  }
}
```

The error message doesn't mention which field is getting parsed.

After several minutes I managed to figure out which field wasn't correctly defined by debugging GraphQL.NET source code.

Wouldn't it be possible to add the field name to the error message?

Cheers,

Sil

sungam3r:
fixes #2293
Trivial fix, no new tests.

[UPDATE]
Introduces `ICanParseScalar` interface. Please, suggest a better name, I don't know.

Features:
1. Allows to eliminate generating exceptions from `ParseXXX` methods.
2. Allows to eliminate boxing from `Guid`, `decimal`, etc. to `object`. 

Also I deleted calls to `ValueConverter`. We have already talked about this, @Shane32 . I only made changes to two scalars to demonstrate the proposed approach.


src/GraphQL/Types/Scalars/ScalarGraphType.cs
```diff
@@ -50,4 +50,31 @@ public abstract class ScalarGraphType : GraphType
         /// <returns> Internal scalar representation. Returning null indicates a failed conversion. </returns>
         public abstract object ParseValue(object value);
     }
+
+    /// <summary>
+    /// This (optional) interface can be implemented by descendants of <see cref="ScalarGraphType"/> to provide
+    /// the ability to validate input values without directly getting those values, i.e. without boxing.
+    /// </summary>
+    public interface ICanParseScalar
```
sungam3r, line 58: 😕  Naming?

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -281,7 +281,23 @@ public static string[] IsValidLiteralValue(this IGraphType type, IValue valueAst
...
             var scalar = (ScalarGraphType)type;
 
-            var parseResult = scalar.ParseLiteral(valueAst);
+            object parseResult = null;
+            try
+            {
+                if (scalar is ICanParseScalar parseable)
+                {
+                    if (parseable.CanParseLiteral(valueAst)) // it still can throw
+                        return Array.Empty<string>();
```
sungam3r, line 290: Returns empty error array in a good case, otherwise `parseResult` just remains `null` without generating any exception as it was before.

Shane32:
I have not done a review yet of the code (time constraints), but I would like to.  Maybe this weekend.

Quick comment however --- is this necessary?  Variable parsing and validation basically happens concurrently, so if there is a problem reading a variable, it is already thrown at the proper point of execution.  The current validation rules already prevent an integer scalar being passed to a string scalar, correct?  So is this just for custom scalars?  How do validation rules handle custom scalars?  Then there's the parsing.  Parsing happens during `GetArgument`; however, the type should already validated by then.  So now we are just adding code so that the validation step can validate custom scalars before the execution step (and thereby know that `GetArgument` will succeed).

Just trying to think this through so I understand what's being fixed.

sungam3r:
It is not about custom scalars. Everything is simpler. Any exception from `ParseLiteral` should not blow up `IsValidLiteralValue`.

sungam3r:
> Just trying to think this through so I understand what's being fixed.

Fixed behavior where the real cause of the error is hidden.

Shane32:
I'll try to review this weekend.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -1985,6 +1989,11 @@ namespace GraphQL.Types
         GraphQL.Language.AST.IValue Convert(object value, GraphQL.Types.IGraphType type);
         bool Matches(object value, GraphQL.Types.IGraphType type);
     }
+    public interface ICanParseScalar
+    {
+        bool CanParseLiteral(GraphQL.Language.AST.IValue value);
+        bool CanParseValue(object value);
+    }
```
Shane32, line 1996: Why don't we just require these methods on `ScalarGraphType` ?  We can include a default implementation as follows:

```cs
public virtual CanParseLiteral(IValue value)
{
    try {
        return ParseLiteral(value) != null;
    } catch {
        return false;
    }
}
```

That's basically what happens now anyway within `IsValidLiteralValue`.

(And similar of course for `CanParseValue`).

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -281,7 +281,23 @@ public static string[] IsValidLiteralValue(this IGraphType type, IValue valueAst
...
+                {
+                    parseResult = scalar.ParseLiteral(valueAst);
+                }
+            }
+            catch (Exception)
+            {
+                // do nothing, just do not throw from IsValidLiteralValue
+            }
 
             if (parseResult == null)
```
Shane32, line 302: All this should simply be:

```suggestion
            if (!scalar.CanParseLiteral(valueAst))
```

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -1985,6 +1989,11 @@ namespace GraphQL.Types
         GraphQL.Language.AST.IValue Convert(object value, GraphQL.Types.IGraphType type);
         bool Matches(object value, GraphQL.Types.IGraphType type);
     }
+    public interface ICanParseScalar
+    {
+        bool CanParseLiteral(GraphQL.Language.AST.IValue value);
+        bool CanParseValue(object value);
```
Shane32, line 1995: I see the project is not using `CanParseValue` just now.  I think this is fine, and I do not think we should make any other changes to variable parsing to use it.

Shane32 (changes requested):

src/GraphQL/Types/Scalars/DecimalGraphType.cs
```diff
@@ -7,21 +9,66 @@ namespace GraphQL.Types
...
+        {
+            return value switch
+            {
+                decimal _ => true,
+                int _ => true,
+                long _ => true,
+                string s => decimal.TryParse(s, NumberStyles.Float, NumberFormatInfo.InvariantInfo, out _),
+                float f => Ret(checked((decimal)f)),
+                BigInteger b => Ret(checked((decimal)b)),
+                double d => Ret(checked((decimal)d)),
```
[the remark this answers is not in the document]
sungam3r: In principle, yes, but nevertheless we cannot guarantee this. At least for a custom scalar for sure. That's why I use `try/catch` in `IsValidLiteralValue`. After all, the reason for this bug is the fact that the scalar throws an exception. This is our scalar. What can we say about custom scalars, even if our code is error prone?

I **would** write a code here that does not throw exceptions, but I do not know how 😄 .

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: Obviously I didn't think about it. Done.

src/GraphQL/GraphQLExtensions.cs
sungam3r: Done

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: So? Let new method be or remove?

Shane32: I suggest we leave it there.  We may find a use for it later, and it matches `CanParseLiteral`.

src/GraphQL/Types/Scalars/DecimalGraphType.cs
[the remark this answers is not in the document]
Shane32: Just put it in a try catch block.

```cs
float f => CanConvertFloat(f),

private bool CanConvertFloat(float f)
{
    try {
        _ = checked((decimal)f);
        return true;
    } catch {
        return false;
    }
}
```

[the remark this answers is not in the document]
Shane32: > In principle, yes, but nevertheless we cannot guarantee this. At least for a custom scalar for sure. That's why I use `try/catch` in `IsValidLiteralValue`. After all, the reason for this bug is the fact that the scalar throws an exception. This is our scalar. What can we say about custom scalars, even if our code is error prone?

The default behavior of `CanConvertLiteral` will catch errors.  If someone overrides this with their own buggy implementation, then it is a server error, not an input error.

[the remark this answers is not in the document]
Shane32: Honestly we do not need to override `CanParseLiteral` for any of the classes.  But it will be a faster implementation if we do since there is one less boxing of integers parsed as decimals.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: Agree.

src/GraphQL/Types/Scalars/DecimalGraphType.cs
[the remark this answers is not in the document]
sungam3r: The proposal is understandable but will require more code jumping around `try/catch`. Also, even in the case of `their own buggy implementation`, I prefer to give a meaningful error. This is the essence of this PR. I mean that `try/catch` in `IsValidLiteralValue` should stay.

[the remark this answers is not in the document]
sungam3r: > But it will be a faster implementation if we do since there is one less boxing of integers parsed as decimals.

Of course. Think about larger structures - `Guid`, some custom scalars.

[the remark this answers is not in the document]
Shane32: I don't think `CanParseLiteral` should throw an error.  Nowhere in any .NET library would you find that a `TryParse` / `CanParse` implementation is allowed to throw an error.

Given that, if you want a duplicate try/catch in `IsValidLiteralValue`, I fail to see the reason why.  In case someone else writes a custom scalar with a buggy `CanParseLiteral` implementation?  I don't think it is necessary, but okay.  That doesn't mean we should write our other `CanParseLiteral` implemenations buggy knowning that in OUR consumer of the method it doesn't matter.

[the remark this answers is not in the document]
sungam3r: Then in `DecimalGraphType` I should write 2 or even 3 methods like this one:
```
private bool CanConvertFloat(float f)
{
    try {
        _ = checked((decimal)f);
        return true;
    } catch {
        return false;
    }
}
```

for `float`, `BigInteger`, `double` . Likewise in other built-in scalars.

> Given that, if you want a duplicate try/catch in IsValidLiteralValue, I fail to see the reason why. 

In order not to do the unnecessary work of writing `try/catch` blocks/methods as showed above in all scalars.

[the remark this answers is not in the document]
Shane32: Well here's another option:

```cs
public class ScalarGraphType
{
    public bool CanParseLiteral(IValue value)
    {
        try {
             return OnCanParseLiteral(value);
        } catch {
             return false;
        }
    }

    protected virtual OnCanParseLiteral(IValue value)
    {
        return ParseLiteral(value) != null;
    }
}
```

Now `CanParseLiteral` will never throw an exception and you don't have to do all that "unnecessary work".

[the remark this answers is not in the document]
sungam3r: done

Shane32:
Looks good. I’ll look through it one more time when I’m back at a pc.

Shane32:
Considering that a custom scalar should be common enough, should we write a dedicated paragraph for it in the migration guide also? It wouldn’t be a breaking change because of the default implementation. If we do write something, it wouldn’t need to say much.

Ok either way but we should update the main docs after v4 release where it talks about custom scalars.

sungam3r:
Done

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -279,16 +279,15 @@ public static string[] IsValidLiteralValue(this IGraphType type, IValue valueAst
...
             {
-                return new[] { $"Expected type \"{type.Name}\", found {AstPrinter.Print(valueAst)}." };
+                return scalar.CanParseLiteral(valueAst)
+                    ? Array.Empty<string>()
+                    : new[] { $"Expected type \"{type.Name}\", found {AstPrinter.Print(valueAst)}." };
             }
 
-            return Array.Empty<string>();
+            // the code shouldn't go here
+            return new[] { $"Value {AstPrinter.Print(valueAst)} of unknown type {type?.Name}." };
```
Shane32, line 290: Would this be more appropriate?

```suggestion
            throw new ArgumentOutOfRangeException(nameof(type), $"Type {type?.Name} is not a valid input graph type.");
```

Shane32 (approved):
One minor comment; merge as you see fit. 👍

PR #2294, opened 2021-02-17
Introduce ExecutionContext.ExecutionStrategy

### LINKED ISSUE #2179: Create custom directive being able to skip/include nodes other than existing "skip" and "include"
## Summary

Currently, to skip/include fields, the code will check directive name and boolean argument's value for "if". My idea is to make the code being able to do this with new custom directives. My idea is to add a boolean virtual property to Field class to check if we do the "ignoreNode", this boolean would take into account the "if" argument for existing include and skip.

### Basic example

- We have a type "Character" with 1 scalar fields of type string: "name"
- We have 2 instances of Character. For user rights reason for example, the first has a "readonly" name (we cannot modify it), the second has a "readwrite" name (we can modify it)
- characters is an array fields returning all character

query
{
  characters
  {
    readonlyName: name @isReadOnly
    readwriteName: name @isReadWrite
  }
}

Expected: 
data: {
  characters: [ readonlyName: "Goku", readwriteName: "Vegeta" ]
}

readwriteName for the first character and readonlyName for the second would be skipped

### Motivation

It would be helpful to be able to do this with customized conditions, I could show my concrete idea by coding on the existing project on my computer with a pull request for example.

sungam3r:
Fixes #2179


src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -51,6 +51,9 @@ public class ExecutionContext : IExecutionContext, IExecutionArrayPool, IDisposa
         /// <inheritdoc/>
         public Action<UnhandledExceptionContext> UnhandledExceptionDelegate { get; set; }
 
+        /// <inheritdoc/>
+        public Func<ExecutionContext, IHaveDirectives, bool> ShouldIncludeNode { get; set; } = (context, node) => ExecutionHelper.ShouldIncludeNode(context, node.Directives);
```
sungam3r, line 55: Ping @jeangsc . Will such property suit you?

jeangsc:
Yes, I haven't tested it yet but watching this code, this feature seems to
match my suggestion.
Thanks

Le mer. 17 févr. 2021 à 16:18, Ivan Maximov <notifications@github.com> a
écrit :

Shane32:
I have not looked at the source code and/or tests a whole lot, but my gut feeling is that this is a bad design.  If this is changing the execution, shouldn't it be a protected method?  Probably a protected method on `ExecutionStrategy`, not on the context. 
 The context can hold the pointer to the protected method if that's really necessary, but there should be a cleaner way to implement this.  Thoughts? @jeangsc @sungam3r 

Keep in mind that the point of directives are as follows (to quote the spec):

> Directives provide a way to describe alternate runtime execution and type validation behavior in a GraphQL document.

In .NET, typically altering runtime execution is done by inheriting from a class and overriding its behavior.  Let's say we wanted a directive to change the runtime behavior from parallel execution to serial execution.  This would be something that directives were explicitly designed for.  And to implement it, you'd override `DocumentExecuter.SelectExecutionStrategy` to inspect the document and return the proper class.

So I think that we should provide this capability by allowing users to override protected methods, rather than providing hooks inside data classes.  (Not to mention that performance is better this way.)

I've said before that we should eliminate these large static methods, and replace them with protected methods.  This is a prime example of why this is the case.  (I've wanted to do similar things in the past, but with static methods, it wasn't possible without copying large parts of the codebase.)

Now, we do have parts of the code, like the validation rules, which is configured through `ExecutionOptions`.  However, I might argue that most of those options also would be better designed as protected members of the `DocumentExecuter`, and not as "options" provided to the executer (since they are typically constants at runtime).  But I don't see that adding a delegate to the execution context is similar or proper.  Perhaps adding the delegate to `ExecutionOptions` (and thereby to the execution context), but I wouldn't do that either.

sungam3r:
I made this feature in the way that I considered the most understandable and non-destructive to the design.

> shouldn't it be a protected method

Protected method of _what_ ?

Shane32:
> Probably a protected method on E`xecutionStrategy`, not on the context.

sungam3r:
Then `Fields.CollectFrom` should somehow take `ExecutionStrategy` argument. It will look strange, but let it be ok for now.
Note that `Fields.CollectFrom` is called not only from `ExecutionStrategy` bu also from `ReadonlyResolveFieldContext`:
```c#
  private Fields GetSubFields()
        {
            return _executionNode.Field?.SelectionSet?.Selections?.Count > 0
                ? new Fields().CollectFrom(_executionContext, _executionNode.FieldDefinition.ResolvedType, _executionNode.Field.SelectionSet)
                : null;
        }
```

It is extremely problematic to get a reference to `ExecutionStrategy` here. How to be?

Shane32:
I don’t know. Probably the execution context needs to point to the execution strategy.

sungam3r:
This will strengthen the tight coupling of the classes. Not good. Now I see the current version as the simplest. One more options -  change the new property to a virtual method. But then we need to let the client code to inherit from `ExecutionContext`. Although this will not fit well with all other properties from it. After all, the mission of `ExecutionContext` is to transfer state. The mission of the `ExecutionStrategy` is to execute nodes using this state.

Shane32:
Then leave the property as-is, but default it to reference a new protected member on `ExecutionStrategy` called `ShouldIncludeNode`, which contains the default implementation.  The property on `ExecutionContext` can have an xml comment that says that is for internal use only.  This slightly "loosens" the coupling, while encouraging users to write proper OOP code.

sungam3r:
> but default it to reference a new protected member on ExecutionStrategy called ShouldIncludeNode

How can it be done from lambda delegate? I don't understand.

Shane32:
Hmm..maybe it can't be protected...

```cs
public class ExecutionContext
{
    public Func<ExecutionContext context, Directives directives, bool> ShouldIncludeNode { get; set; }
}

public class ExecutionStrategy
{
    public virtual bool ShouldIncludeNode(ExecutionContext context, Directives directives)
    {
        return ExecutionHelper.ShouldIncludeNode(context, directives);
    }
}

public class DocumentExecuter
{
                    ///line 245:
                    IExecutionStrategy executionStrategy = SelectExecutionStrategy(context);
                    ///add this line immediately after:
                    context.ShouldIncludeNode = (context, directives) => executionStrategy.ShouldIncludeNode(context, directives);
}
```

sungam3r:
It all looks very strange. A strategy declares a method that it does not use itself. The method only delegates control to another static method. There is a transfer of state from one place to another, moreover you want the client not to change the contents of the public property. For what is this all? The code is getting more and more confusing.

Shane32:
You're right.  Context should not hold this property at all.  It has nothing to do with the context; it is part of the execution.  I'm trying to suggest alternatives.  Most likely there's a whole lot of code here that needs to be reworked to be done properly.

sungam3r:
Let's change property to `virtual` method and allow `DocumentExecuter`'s descendants to create their own `ExecutionContext`.

sungam3r:
At least the ability to create execution context descendants will be useful. Then clients can use the normal properties in the descendants instead of `UserContext` dictionary.

sungam3r:
And yet I don't understand your objection to adding a new property. How does it differ from, for example, `ExecutionContext.MaxParallelExecutionCount`? After all, I can also say that `MaxParallelExecutionCount` should refer to strategy, not to context.

sungam3r:
Keeping these properties in context allows us to make other parts of the system as stateless. Otherwise, it would be required to create `ParallelExecutionStrategy` each request and transfer `MaxParallelExecutionCount` to it from options.

sungam3r:
Although I understand your point of view. Now only `GetSubFields` bothers me, otherwise everything is going well.

Shane32:
> And yet I don't understand your objection to adding a new property. How does it differ from, for example, `ExecutionContext.MaxParallelExecutionCount`? After all, I can also say that `MaxParallelExecutionCount` should refer to strategy, not to context.

This isn't an option interpreted by code.  It's code itself.  It's changing how the execution strategy executes.  The context should be a DATA model, not holding code executed by the execution strategy.

> Although I understand your point of view

👍

sungam3r:
As an idea - what if we make the `SubFields` as method. This property is rarely required. Never in our code :). And if someone needs it in the resolver, then there is no problem to save the result to a variable.

Shane32:
I’m fine with sub fields as method. Keep in mind that with recycling the resolve field context, extra fields are not so terribly bad.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -51,7 +50,7 @@ await ExecuteNodeTreeAsync(context, rootNode)
         /// <summary>
         /// Builds the root execution node.
         /// </summary>
-        protected static RootExecutionNode BuildExecutionRootNode(ExecutionContext context, IObjectGraphType rootType)
+        protected RootExecutionNode BuildExecutionRootNode(ExecutionContext context, IObjectGraphType rootType)
```
Shane32, line 53: Virtual?

src/GraphQL/Language/AST/Fields.cs
```diff
@@ -17,22 +24,31 @@ public class Fields : Dictionary<string, Field>
...
+        /// If set to <see langword="null"/>, <see cref="ExecutionHelper.ShouldIncludeNode"/> will be used to work
+        /// as required by the specification. Set this parameter if you understand exactly what you are doing,
+        /// because your actions may lead to the fact that the server's behavior ceases to comply with the
+        /// specification requirements.
+        /// </param>
+        public Fields CollectFrom(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet, IFieldCollectionStrategy fieldCollectionStrategy = null)
         {
             List<string> visitedFragmentNames = null;
-            CollectFields(context, specificType, selectionSet, ref visitedFragmentNames);
+            CollectFields(context, specificType, selectionSet, fieldCollectionStrategy ?? DefaultFieldCollectionStrategy.Instance, ref visitedFragmentNames);
```
Shane32, line 39: Why not store the field collection strategy instance in the context and/or use the execution strategy which is stored there? Sorry I’m not at a pc just now so maybe it’s obvious. I guess at first glance I wouldn’t add this argument.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -60,23 +59,25 @@ protected static RootExecutionNode BuildExecutionRootNode(ExecutionContext conte
...
-            SetSubFieldNodes(context, root, fields.CollectFrom(context, rootType, context.Operation.SelectionSet));
+            SetSubFieldNodes(context, root, fields.CollectFrom(context, rootType, context.Operation.SelectionSet, this));
 
             fields.Clear();
             System.Threading.Interlocked.CompareExchange(ref context.ReusableFields, fields, null);
 
             return root;
         }
 
+        bool IFieldCollectionStrategy.ShouldIncludeNode(ExecutionContext context, IHaveDirectives directives) => ExecutionHelper.ShouldIncludeNode(context, directives.Directives);
```
Shane32, line 70: I think IExecutionStrategy should inherit IFieldCollectionStrategy and this should simply be a protected virtual member of the class.

Shane32: And then add another IFieldCollectionStrategy.ShouldIncludeNode(c) => ShouldIncludeNode(c); so that it doesn’t need to be public

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
```diff
@@ -39,7 +39,7 @@ internal ReadonlyResolveFieldContext Reset(ExecutionNode node, ExecutionContext
         private Fields GetSubFields()
         {
             return _executionNode.Field?.SelectionSet?.Selections?.Count > 0
-                ? new Fields().CollectFrom(_executionContext, _executionNode.FieldDefinition.ResolvedType, _executionNode.Field.SelectionSet)
+                ? new Fields().CollectFrom(_executionContext, _executionNode.FieldDefinition.ResolvedType, _executionNode.Field.SelectionSet, _executionContext.ExecutionStrategy as IFieldCollectionStrategy)
```
Shane32, line 42: Then we can remove the cast

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Maybe. Note that other Build method is `static`. I still can't decide whether to make both static or both instance and (virtual).

src/GraphQL/Language/AST/Fields.cs
sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/pull/2294/files#diff-43fbf64bffe7f823cf1060bfa739a3ec85efdd804dd414e11fadef48cbb04c82

sungam3r: Ooops , understood.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: I think we should do private or protected virtual. If we do protected, then it breaks compatibility to make it virtual. In general I think many of these methods should be protected virtual instead of static, for the same reason. 

Also might consider moving the default ShouldIncludeNode implementation into ExecutionStrategy.ShouldIncludeNode

As much as I like interfaces, the execution strategy could be something where the interface really only helps for mocking and testing but in real life is not beneficial, assuming we use all virtual methods so it can be completely customized.

Shane32: (I am not suggesting we remove the IExecutionStrategy interface)

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
Shane32: And the argument altogether.

src/GraphQL/Language/AST/Fields.cs
sungam3r: I got lost in my own changes. Done!

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -217,24 +217,18 @@ public static bool ShouldIncludeNode(ExecutionContext context, Directives direct
...
                 if (directive != null)
                 {
-                    var values = GetArgumentValues(
-                        DirectiveGraphType.Skip.Arguments,
-                        directive.Arguments,
-                        context.Variables);
+                    var arg = DirectiveGraphType.Skip.Arguments.Find("if");
 
-                    if (values.TryGetValue("if", out ArgumentValue ifObj) && bool.TryParse(ifObj.Value?.ToString() ?? string.Empty, out bool ifVal) && ifVal)
+                    if ((bool)CoerceValue(arg.ResolvedType, directive.Arguments?.ValueFor(arg.Name), context.Variables, arg.DefaultValue).Value)
```
sungam3r, line 222: `Dictionary<,>` allocation optimization here and below.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: > Also might consider moving the default ShouldIncludeNode implementation into ExecutionStrategy.ShouldIncludeNode

Already done.

src/GraphQL/Language/AST/Fields.cs
```diff
@@ -24,31 +28,22 @@ private sealed class DefaultFieldCollectionStrategy : IFieldCollectionStrategy
...
-        /// as required by the specification. Set this parameter if you understand exactly what you are doing,
-        /// because your actions may lead to the fact that the server's behavior ceases to comply with the
-        /// specification requirements.
-        /// </param>
-        public Fields CollectFrom(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet, IFieldCollectionStrategy fieldCollectionStrategy = null)
+        public Fields CollectFrom(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet)
         {
             List<string> visitedFragmentNames = null;
-            CollectFields(context, specificType, selectionSet, fieldCollectionStrategy ?? DefaultFieldCollectionStrategy.Instance, ref visitedFragmentNames);
+            CollectFields(context, specificType, selectionSet, context.ExecutionStrategy ?? DefaultExecutionStrategy.Instance, ref visitedFragmentNames);
```
Shane32, line 34: Do we need to make context.ExecutionStrategy optional? I’d remove the “DefaultExecutionStrategy” altogether. If nothing else just make it pull the ParallelExecutionStrategy instance.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Done

src/GraphQL/Language/AST/Fields.cs
sungam3r: Done

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: I removed `IFieldCollectionStrategy `. It is not needed at all.

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
sungam3r: Done

sungam3r:
Anything else?

Shane32 (approved):

Shane32 (approved):

PR #2295, opened 2021-02-17
Introduce IResolveFieldContext.Parent

### LINKED ISSUE #899: Passing data down the tree
I'm gonna open a new issues specifically for that. #637

What about this scenario. Our API is scoped. We have an array of customers that we'd like to resolve. 
Underneath the customer node we have a fairly deep graphql tree. 
As far as i know the only way to pass down properties is to drill them down the tree.
```csharp
           FooBarModel  Customer(string customer).getFooBar();
                    ->  FooBarModelDeep FooBarModel(string customer).GetFooBarModelDeep();
```
That leads to unnecessary models and a higher complexity over all. 

```graphql
query Sample($f: [String]){
    foo {
        customers(f: $f) {
            fooBar {
               fooBarDeep {
                      name
                   }
            }
        }
    }
}
``` 

input ["FirstEntry", "SecondEntry"]  
result:
```
{
   foo: {
       customers: [
          {
               fooBar: {
                    fooBarDeep: {
                          name:"FirstEntry"
                    }
               }
          },
             {
               fooBar: {
                    fooBarDeep: {
                          name:"SecondEntry"
                    }
               }
          }

        ]
    }
}
```

Is there a way around it?

sungam3r:
fixes #899


sungam3r:
@Shane32 I am thinking about removing the `IResolveFieldContext` entirely and working with `ExecutionNode` and `ExecutionContext` directly or via readonly interfaces. This idea attracts me with its simplicity of implementation. Maybe it should be done in version 5. Now I do not want to make such a change, because it will require rewriting the documentation, there will be a delay again.

docs2/site/docs/migrations/migration4.md
```diff
@@ -322,3 +322,4 @@ lock (field)
 * `ExecutionNode.Source` is read-only; additional derived classes have been added for subscriptions
 * `NameValidator.ValidateName` and `NameValidator.ValidateNameOnSchemaInitialize` accept an enum instead of a string for their second argument
 * `ExecutionNode.PropagateNull` must be called before `ExecutionNode.ToValue`; see reference implementation
+* `IResolveFieldContext.FieldName` and `IResolveFieldContext.ReturnType` properties have been removed, use `IResolveFieldContext.FieldAst.Name` and `IResolveFieldContext.FieldDefinition.ResolvedType` instead
```
Shane32, line 325: This has nothing to do with this PR.  I do not think they should be removed, but if we want to discuss it, we should do it separately from this PR.

Further, there is no comment that `Parent` has been added in the 'new features' section.

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
```diff
@@ -49,20 +53,17 @@ private Fields GetSubFields()
...
         /// <inheritdoc/>
         public FieldType FieldDefinition => _executionNode.FieldDefinition;
 
         /// <inheritdoc/>
-        public IGraphType ReturnType => _executionNode.FieldDefinition.ResolvedType;
+        public IObjectGraphType ParentType => _executionNode.GetParentType(_executionContext.Schema);
 
         /// <inheritdoc/>
-        public IObjectGraphType ParentType => _executionNode.GetParentType(_executionContext.Schema);
+        public IResolveFieldContext Parent => _executionNode.Parent is RootExecutionNode || _executionNode.Parent == null ? null : _parent ??= new ReadonlyResolveFieldContext(_executionNode.Parent, _executionContext);
```
Shane32, line 66: Will this work for all situations, specifically array item nodes and children of array item nodes?

Shane32 (changes requested):

docs2/site/docs/migrations/migration4.md
sungam3r: Two removed properties are always calculated and do not carry their own semantic meaning. They should be removed.
I definitely added a note about `Parent` in some other PR.

sungam3r: Yes, in #2276 . I don't want to aggravate the situation with merge conflicts.

sungam3r: >  I do not think they should be removed

See here:
изображение

These properties should be removed to avoid the problem of invalid data. They are not just overkill, they are the cause of potential error. Now you can set values that contradict each other.

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
sungam3r: Right. Fixed.

docs2/site/docs/migrations/migration4.md
sungam3r: I added note about `Parent`.

Shane32: The properties are convenience properties on `IResolveFieldContext`, just like almost all of the properties are.  We both know that `ReadonlyResolveFieldContext` only holds a few members.  But many properties are provided (mapped) so the user can quickly access the fields they care about.  Further, `IResolveFieldContext` is one of the most used classes in the project.  We're not talking about some obscure part of the guts of the project.  We're talking about code that everyone deals with in every graph in every project.  If we want to eliminate these properties, let's mark them as obsolete for v4 and remove them in v5.  It does not benefit anything to remove them now.  It is not faster, it does not save memory, it does not help reorganize the API for the better, nothing.  I agree they are redundant, but they are not out of place.  Removing them only servers one purpose: adding to user frustration when they convert to v4.

Shane32: I have no problem if you want to change `ResolveFieldContext.FieldName` to be a mapping instead of a read/write property.  That class is almost exclusively used for tests.

sungam3r: > it does not help reorganize the API for the better

It does. We declared interface with two superfluous properties. This interface smells.

I don't see a problem in removing them, we made a lot of changes and we have a dood migration guide. Why drag into the new version something that is absolutely unnecessary now? Customers will still need to recompile their code.

Shane32: Ok

Shane32: Let's at least mark them obsolete in 3.x then, with a note saying the properties to use instead.

sungam3r: #2301

sungam3r:
Tested with Arrays and Arrays of Arrays.

Shane32:
I merged master -> develop -> parent-subtree

sungam3r:
Anything else here?

Shane32 (approved):

sungam3r:
After merge from `develop` turned out to be a rather compact PR.

PR #2301, opened 2021-02-20
Make two properties Obsolete

sungam3r:


Shane32 (approved):

Shane32:
Would suggest something like this where needed:
image

Shane32:
See #1827

PR #2304, opened 2021-02-22
Type names are easier to get

### LINKED ISSUE #2102: Rework TypeExtensions.GraphTypeFromType

sungam3r:
fixes #2102

### DIFF
--- a/src/GraphQL/Types/TypeExtensions.cs
+++ b/src/GraphQL/Types/TypeExtensions.cs
@@ -9,6 +9,21 @@ namespace GraphQL.Types
     public static class TypeExtensions
     {
         /// <summary>
+        /// Searches a schema for a graph type specified by an AST type after unwrapping any
+        /// <see cref="NonNullType"/> or <see cref="ListType"/> layers. If the type cannot be
+        /// found, returns <see langword="null"/>.
+        /// </summary>
+        /// <param name="type">The AST type to search for.</param>
+        /// <param name="schema">The schema to search within.</param>
+        public static IGraphType NamedGraphTypeFromType(this IType type, ISchema schema) => type switch
+        {
+            NonNullType nonnull => NamedGraphTypeFromType(nonnull.Type, schema),
+            ListType list => NamedGraphTypeFromType(list.Type, schema),
+            NamedType named => schema.AllTypes[named.Name],
+            _ => null
+        };
+
+        /// <summary>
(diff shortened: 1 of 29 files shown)


src/GraphQL/Types/TypeExtensions.cs
```diff
@@ -8,6 +8,21 @@ namespace GraphQL.Types
...
     public static class TypeExtensions
     {
+        /// <summary>
+        /// Searches a schema for a graph type specified by an AST type after unwrapping any
+        /// <see cref="NonNullType"/> or <see cref="ListType"/> layers. If the type cannot be
+        /// found, returns <see langword="null"/>.
+        /// </summary>
+        /// <param name="type">The AST type to search for.</param>
+        /// <param name="schema">The schema to search within.</param>
+        public static IGraphType NamedGraphTypeFromType(this IType type, ISchema schema) => type switch
```
sungam3r, line 18: Moved here from `VariablesAreInputTypes`.

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -191,16 +190,6 @@ public List<FragmentDefinition> GetRecursivelyReferencedFragments(Operation oper
...
-        /// <summary>
-        /// Returns a string representation of the specified node.
-        /// </summary>
-        public string Print(INode node) => AstPrinter.Print(node);
-
-        /// <summary>
-        /// Returns the name of the specified graph type.
-        /// </summary>
-        public string Print(IGraphType type) => SchemaPrinter.ResolveName(type);
-
```
Shane32, line 203: I was planning to see if we could decouple one or more of the AstPrinter / SchemaPrinter / SchemaBuilder libraries.  If that happens, these methods might need to move back here.  So I'd probably not remove them, but ok I guess....

Shane32 (approved):

src/GraphQL/Validation/ValidationContext.cs
sungam3r: These methods created unnecessary dependencies.

PR #2306, opened 2021-02-22
Add ISchema.ValueConverters

### LINKED ISSUE #885: Expose IAstFromValueConverters Registered with Schema
## Summary

It would be nice if all IAstFromValueConverters registered with a Schema were made available in a public, 
 read-only collection on Schema. These seem to be the only types the user can register that aren't made available this way.

### Basic example

```
public class Schema
{
    //All Existing functionality...

    public IEnumerable<IAstFromValueConverter> ValueConverters { get; }
}
```
### Motivation

I'm writing a utility to allow for the definition of 'partial schemas' that can be rolled up into a single composite. The structure of Schema makes this trivial for the most part, but the lack of access to Value Converters means I can't register these in my composite without resorting to reflection.

sungam3r:
fixes #885

### DIFF
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -172,6 +172,7 @@ This may be needed to get the parameters of parent nodes.
 
 ### Other Features
 
+* New property `ISchema.ValueConverters`
 * New method `IParentExecutionNode.ApplyToChildren`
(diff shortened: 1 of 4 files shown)


Shane32 (approved):
Should we add a bullet point in the migration document under new features?  Ok either way.

PR #2307, opened 2021-02-23
Cleanup ScalarGraphType

Shane32:
This cleans up the API and eliminates boxing for many types of scalar graph types when accessed by `CanParseLiteral`.

I did not add code for `CanParseValue` because in almost all instances, `ValueConverter` is called, and so it cannot be determined which types can be supported.  Besides, that code path is not used by GraphQL.Net.

A `try..catch` block was only needed by a couple scalars.

### DIFF
--- a/src/GraphQL/Types/Scalars/LongGraphType.cs
+++ b/src/GraphQL/Types/Scalars/LongGraphType.cs
@@ -18,5 +18,8 @@ namespace GraphQL.Types
 
         /// <inheritdoc/>
         public override object ParseValue(object value) => ValueConverter.ConvertTo(value, typeof(long));
+
+        /// <inheritdoc/>
+        public override bool CanParseLiteral(IValue value) => value is IntValue || value is LongValue;
     }
 }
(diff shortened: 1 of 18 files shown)


sungam3r:
> because in almost all instances, ValueConverter is called, and so it cannot be determined which types can be supported. 

I thought that one of our goals is to replace these calls. We already talked about this.

src/GraphQL/Types/Scalars/LongGraphType.cs
```diff
@@ -18,5 +18,8 @@ public class LongGraphType : ScalarGraphType
 
         /// <inheritdoc/>
         public override object ParseValue(object value) => ValueConverter.ConvertTo(value, typeof(long));
+
+        /// <inheritdoc/>
+        public override bool CanParseLiteral(IValue value) => value is IntValue || value is LongValue;
```
sungam3r, line 23: I would prefer to keep `switch value` style everywhere even in the simplest cases.

sungam3r: OK, nevermind.

sungam3r (approved):
We can delete `ValueConverter.ConvertTo` calls in another PR.

Shane32:
> > because in almost all instances, ValueConverter is called, and so it cannot be determined which types can be supported.
> 
> I thought that one of our goals is to replace these calls. We already talked about this.

I still agree.

Issue #2279, opened 2021-02-12
Generic Type Is Generated Only Once Per Schema - Resulting into having Incorrect Return Types for Fields


JakubSlonka:
## Description

While using generic types in your queries/mutations, they appear only once per schema. During schema generation, only last occurrence of the generic type is kept with its underlying type, other occurences are ignored (overwritten) resulting in incorrect schema types.

The issue can be reproduced when using any of the graphql UI.

Voyager
GraphqiQl
GraphQlPlayground
Altair

### Steps to reproduce

Create Graphql Generic type 
```
 public class GeneralRequestResponseType<T, TType> : ObjectGraphType<GeneralRequestResponse<T>>
    {
        public GeneralRequestResponseType()
        {
            Name = "GenericRequestResponse";
            Field(name: "Data", type: typeof(TType), resolve: context => context.Source.Data);
            Field(name: "Messages", type: typeof(ListGraphType<StringGraphType>), resolve: context => context.Source.Messages);
        }
    }
```

Create Multiple Object Graph Types from DTOs

```
public class SomeDtoType: ObjectGraphType<SomeDto>
    {
        public SomeDtoType()
        {
            Field(x => x.Id);
            Field(x => x.SomeProperty);
        }
}

public class SomeOtherDtoType: ObjectGraphType<SomeOtherDto>
    {
        public SomeOtherDtoType()
        {
            Field(x => x.Id);
            Field(x => x.SomeOtherProperty);
        }
}
```

Use Generic Type with Multiple Graph Types In Your Query/Mutation

```
FieldAsync<GeneralRequestResponseType<SomeDto, SomeDtoType>>(
                "someDto",
                resolve: async context => await resolve);

FieldAsync<GeneralRequestResponseType<SomeOtherDto, SomeOtherDtoType>>(
                "someOtherDto",
                resolve: async context => await resolve);
```

Run Project and Access UI endpoint
See Queries and its types

### Expected result

For each field using generic type there should be generic object graph type with correct underlying object graph type

### Actual result

All fields have the same generic object graph type ( type from last occurance during generation / not respecting actual type)

### Additional Info:
This behaviour breaks all of the UIs and makes them useless for projects using GenericTypes.

I have checked the schema in GrapQlHttpMiddleware and the schema is already generated with incorrect types.

I have tracked the issue and it indeed occurs during schema type generation.

When generating _lookUp table for types, there is Union method used, which filters out duplicities. 

[...]

PR #2309, opened 2021-02-23
Test for duplicate type name for generic graph type

Shane32:
This passes both against `master` and `develop`... ???

See #2279


sungam3r:
Then just merge it in develop.

sungam3r (approved):

Shane32:
GitHub is having issues running the build checks.  I don't think there's anything wrong with the code, but better to wait until GitHub Actions is working again and we have a successful build.

Shane32:
I don't think we should have newlines within error messages.  And any uses of the newline should be platform-specific!

PR #2312, opened 2021-02-23
ExecutionStrategy cleanup

Shane32:
I'd like to clean up the `ExecutionStrategy` design a bit, so everything execution-related is contained within `ExecutionStrategy`.  Please take a look.

In general, execution-related code was moved from `ExecutionHelper` and `Fields` into `ExecutionStrategy`, and methods were changed to be `protected virtual` rather than `public` or `static`.  A few methods' signatures changed a very little.  Also, `protected virtual DocumentExecuter.BuildExecutionContext` was added.

All in, this allows for a user to derive from `ExecutionOptions`, `DocumentExecuter`, `ExecutionContext` and `ExecutionStrategy`, with minimal changes to the codebase, and allows for changing nearly any part of the underlying execution process.  Most importantly, if a user finds a bug inside one of the methods, it is much easier to make a change than having most of the code statically linked together in an all-or-none approach.

Consider this example:

```cs
    class MyExecutionOptions : ExecutionOptions
    {
        public bool IgnoreDirectives { get; set; }
    }

    class MyExecutionContext : ExecutionContext
    {
        public bool IgnoreDirectives { get; set; }
    }

    class MyExecutionStrategy : ParallelExecutionStrategy
    {
        protected override bool ShouldIncludeNode(ExecutionContext context, IHaveDirectives node)
        {
            return ((MyExecutionContext)context).IgnoreDirectives || base.ShouldIncludeNode(context, node);
        }
    }

    class MyDocumentExecuter : DocumentExecuter
    {
        private IExecutionStrategy _executionStrategy;
        public MyDocumentExecuter()
        {
            _executionStrategy = new MyExecutionStrategy();
        }

        protected override ExecutionContext BuildExecutionContext(ExecutionOptions options, Document document, Operation operation, Variables variables, Metrics metrics)
        {
            var context = new MyExecutionContext
            {
                Document = document,
                Schema = options.Schema,
                RootValue = options.Root,
                UserContext = options.UserContext,

                Operation = operation,
                Variables = variables,
                Fragments = document.Fragments,
                Errors = new ExecutionErrors(),
                Extensions = new Dictionary<string, object>(),
                CancellationToken = options.CancellationToken,

                Metrics = metrics,
                Listeners = options.Listeners,
                ThrowOnUnhandledException = options.ThrowOnUnhandledException,
                UnhandledExceptionDelegate = options.UnhandledExceptionDelegate,
                MaxParallelExecutionCount = options.MaxParallelExecutionCount,
                RequestServices = options.RequestServices,

                IgnoreDirectives = ((MyExecutionOptions)options).IgnoreDirectives,
            };

            context.ExecutionStrategy = SelectExecutionStrategy(context);

            return context;
        }

        {
            return _executionStrategy;
        }
    }
```

I also cleaned up `ReadonlyResolveFieldContext` a little, providing a better `public` surface of `ExecutionStrategy`:

```cs
    /// <summary>
    /// Processes a parsed GraphQL request, resolving all the nodes and returning the result; exceptions
    /// are unhandled. Should not run any <see cref="IDocumentExecutionListener">IDocumentExecutionListener</see>s except
    /// for <see cref="IDocumentExecutionListener.BeforeExecutionStepAwaitedAsync(IExecutionContext)">BeforeExecutionStepAwaitedAsync</see>.
    /// </summary>
    public interface IExecutionStrategy
    {
        /// <summary>
        /// Executes a GraphQL request and returns the result
        /// </summary>
        /// <param name="context">The execution parameters</param>
        Task<ExecutionResult> ExecuteAsync(ExecutionContext context);

        /// <summary>
        /// Returns the children fields for a specified node.
        /// </summary>
        Fields GetSubFields(ExecutionContext executionContext, ExecutionNode executionNode);
    }
```

`GetSubFields` is still public, as it needs to be accessed by `IResolveFieldContext`, but at least now 100% of the related code is contained within the execution strategy, and can be modified at will there.

`IExecutionStrategy` holds no fields and could probably be removed.  But I left it alone, and would not remove it at this time.

I could not replace references to `ExecutionContext` with `IExecutionContext` because currently there are `internal` fields on `ExecutionContext`.  So for now I left `IExecutionContext` simply as a read-only view of `ExecutionContext`.

Another idea would be to remove `ExecutionContext` altogether; `BuildExecutionContext` becomes `BuildExecutionStrategy`.  All context-related fields are stored within the execution strategy itself.  All methods within the strategy now do not need to be passed a reference to `ExecutionContext` at all.  It would clean up the API very nicely I think, without any change to memory usage.  I did not do so within this PR; I can provide a separate PR for review.  We can still retain `IExecutionContext` for use by events and so on that require it.  Biggest negative is that it will probably complicate `SelectExecutionStrategy` a little, which is typically overridden by users.

If everything looks okay I will add migration notes.


src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -47,22 +47,49 @@ public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionContext context
...
                 Result = context.RootValue
             };
 
-            var fields = System.Threading.Interlocked.Exchange(ref context.ReusableFields, null) ?? new Fields();
-
-            SetSubFieldNodes(context, root, fields.CollectFrom(context, rootType, context.Operation.SelectionSet));
-
-            fields.Clear();
-            System.Threading.Interlocked.CompareExchange(ref context.ReusableFields, fields, null);
+            SetSubFieldNodes(context, root);
```
Shane32, line 92: By adding the `SelectionSet` to the `RootExecutionNode`, duplicate code can be removed.

src/GraphQL/Execution/UnhandledExceptionContext.cs
```diff
@@ -10,7 +10,7 @@ public class UnhandledExceptionContext
         /// <summary>
         /// Initializes a new instance with the specified properties.
         /// </summary>
-        public UnhandledExceptionContext(ExecutionContext context, IResolveFieldContext fieldContext, Exception originalException)
+        public UnhandledExceptionContext(IExecutionContext context, IResolveFieldContext fieldContext, Exception originalException)
```
Shane32, line 13: It seemed that in this situation, a read-only view of the execution context would be appropriate.

src/GraphQL/Language/AST/Fields.cs
```diff
@@ -10,85 +10,6 @@ namespace GraphQL.Language.AST
...
-        /// <see href="http://spec.graphql.org/June2018/#sec-Field-Collection"/> and <see href="http://spec.graphql.org/June2018/#CollectFields()"/>
-        /// </summary>
-        public Fields CollectFrom(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet)
-        {
-            List<string> visitedFragmentNames = null;
-            CollectFields(context, specificType.GetNamedType(), selectionSet, context.ExecutionStrategy ?? ParallelExecutionStrategy.Instance, ref visitedFragmentNames);
-            return this;
-        }
-
-        private void CollectFields(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet, IExecutionStrategy strategy, ref List<string> visitedFragmentNames) //TODO: can be completely eliminated? see Fields.Add
```
Shane32, line 28: Execution-related code has been removed from `Fields`.

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
```diff
@@ -40,13 +40,6 @@ internal ReadonlyResolveFieldContext Reset(ExecutionNode node, ExecutionContext
             return this;
         }
 
-        private Fields GetSubFields()
-        {
-            return _executionNode.Field?.SelectionSet?.Selections?.Count > 0
-                ? new Fields().CollectFrom(_executionContext, _executionNode.FieldDefinition.ResolvedType, _executionNode.Field.SelectionSet)
-                : null;
-        }
-
```
Shane32, line 49: Moved to `ExecutionStrategy`.

src/GraphQL/Execution/Nodes/ObjectExecutionNode.cs
```diff
@@ -86,6 +86,11 @@ public override bool PropagateNull()
 
         IEnumerable<ExecutionNode> IParentExecutionNode.GetChildNodes() => SubFields ?? Enumerable.Empty<ExecutionNode>();
 
+        /// <summary>
+        /// Returns the selected fields.
+        /// </summary>
+        public virtual SelectionSet SelectionSet => Field?.SelectionSet;
```
Shane32, line 92: Adding this member simplified `ExecutionStrategy.SetSubFieldNodes`.  No additional memory requirements here, but one more field in `RootExecutionNode`.

src/GraphQL/Execution/IExecutionStrategy.cs
```diff
@@ -17,8 +17,8 @@ public interface IExecutionStrategy
         Task<ExecutionResult> ExecuteAsync(ExecutionContext context);
 
         /// <summary>
-        /// This methods allows you to control the set of fields that the strategy will execute.
+        /// Returns the children fields for a specified node.
         /// </summary>
-        bool ShouldIncludeNode(ExecutionContext context, IHaveDirectives directives);
+        Fields GetSubFields(ExecutionContext executionContext, ExecutionNode executionNode);
```
Shane32, line 22: This seemed better to me. It simplifies `ReadonlyResolveFieldContext` and only exposes a method to retrieve the subfields, rather than a method with implementation details (requiring all of `CollectFrom` to be `static`, and hence unable to be changed my a derived execution strategy).

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -134,13 +177,131 @@ private void SetSubFieldNodes(ExecutionContext context, ObjectExecutionNode pare
...
+
+            fields.Clear();
+            System.Threading.Interlocked.CompareExchange(ref context.ReusableFields, fields, null);
+        }
+
+        /// <summary>
+        /// Returns a <see cref="FieldType"/> for the specified AST <see cref="Field"/> within a specified parent
+        /// output graph type within a given schema. For meta-fields, returns the proper meta-field field type.
+        /// </summary>
+        protected virtual FieldType GetFieldDefinition(ISchema schema, IObjectGraphType parentType, Field field)
```
Shane32, line 189: Moved from `ExecutionHelper`; no changes.  I think it belongs here, unless there is a reason this method would be needed outside of the execution strategy.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -134,13 +177,131 @@ private void SetSubFieldNodes(ExecutionContext context, ObjectExecutionNode pare
...
+            return parentType.GetField(field.Name);
+        }
+
+        /// <inheritdoc/>
+        public virtual Fields GetSubFields(ExecutionContext context, ExecutionNode node)
+        {
+            return node.Field?.SelectionSet?.Selections?.Count > 0
+                ? CollectFieldsFrom(context, node.FieldDefinition.ResolvedType, node.Field.SelectionSet)
+                : null;
+        }
```
Shane32, line 218: Moved from `ReadonlyResolveFieldContext`

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -134,13 +177,136 @@ private void SetSubFieldNodes(ExecutionContext context, ObjectExecutionNode pare
...
+        /// <param name="fields">A list to append the collected list of fields to; if null, a new list will be created.</param>
+        /// <returns>A list of collected fields</returns>
+        protected virtual Fields CollectFieldsFrom(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet, Fields fields = null)
+        {
+            fields ??= new Fields();
+            List<string> visitedFragmentNames = null;
+            CollectFields(context, specificType.GetNamedType(), selectionSet, fields, ref visitedFragmentNames);
+            return fields;
+
+            void CollectFields(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet, Fields fields, ref List<string> visitedFragmentNames) //TODO: can be completely eliminated? see Fields.Add
```
Shane32, line 240: Moved from `Fields` (previously in `ExecutionHelper`).  I think it belongs here, as execution-related code, rather than in `ExecutionHelper` as a static method, or in `Fields`, which should be a simple collection class.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -134,13 +177,136 @@ private void SetSubFieldNodes(ExecutionContext context, ObjectExecutionNode pare
...
+            }
+        }
+
+        /// <summary>
+        /// This method calculates the criterion for matching fragment definition (spread or inline) to a given graph type.
+        /// This criterion determines the need to fill the resulting selection set with fields from such a fragment.
+        /// <br/><br/>
+        /// <see href="http://spec.graphql.org/June2018/#DoesFragmentTypeApply()"/>
+        /// </summary>
+        protected virtual bool DoesFragmentConditionMatch(ExecutionContext context, string fragmentName, IGraphType type /* should be named type*/)
```
Shane32, line 280: Also moved from `Fields`.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -47,22 +47,49 @@ public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionContext context
         /// </summary>
         protected abstract Task ExecuteNodeTreeAsync(ExecutionContext context, ObjectExecutionNode rootNode);
 
+        /// <summary>
+        /// Returns the root graph type for the execution -- for a specified schema and operation type.
+        /// </summary>
+        protected virtual IObjectGraphType GetOperationRootType(Document document, ISchema schema, Operation operation)
```
Shane32, line 53: Moved from `ExecutionHelper`; no changes.  Could be changed to accept `ExecutionContext` rather than `Document`, `ISchema`, and `Operation` but I did not do so.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r, line 280: The behavior of this method is described in the specification and should not change.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Why virtual?

src/GraphQL/Execution/Nodes/ObjectExecutionNode.cs
sungam3r: OK

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Let's change to single ExecutionContext parameter.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: I didn't like that either.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Correct; it _should_ not change.  In fact the majority of this PR _should_ not change.  The point of this PR is to allow modifications to _any_ part of the execution strategy, regardless of whether it meets spec or not.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: The point of this PR is to allow modifications to any part of the execution strategy, regardless of the reason.

What if we needed to override the meta field to properly support the `@oneOf` directive (specifically, the new `oneField` field on `__Type`)?  That doesn't meet spec, but may be something someone wants to support.  With this change, you could change what graph type is returned without changing any other part of the code.

I'm not saying it is necessary in that case, but it is a good example of a situation where could be necessary to change this method.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -10,65 +10,6 @@ namespace GraphQL.Execution
     /// </summary>
     public static class ExecutionHelper
```
sungam3r, line 11: Only two methods left in `ExecutionHelper`. I wouldn't say they belong to "execution" stage 100%.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -263,6 +241,38 @@ await listener.AfterExecutionAsync(context)
...
+            var context = new ExecutionContext
+            {
+                Document = document,
+                Schema = options.Schema,
+                RootValue = options.Root,
+                UserContext = options.UserContext,
+
+                Operation = operation,
+                Variables = variables,
+                Fragments = document.Fragments,
```
sungam3r, line 258: Let's remove `Fragments` property from `ExecutionContext` and `IResolveFieldContext`.  For some reason I missed it last time.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -263,6 +241,38 @@ await listener.AfterExecutionAsync(context)
...
+                Errors = new ExecutionErrors(),
+                Extensions = new Dictionary<string, object>(),
```
sungam3r, line 260: Also I suggest to remove this initialization line. Anyone who needs extensions can now override `BuildExecutionContext` ans set it.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: I'll separately do a pair of PRs for this, after this is merged.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: I'll separately do a PR for this after this PR is merged.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -134,13 +177,136 @@ private void SetSubFieldNodes(ExecutionContext context, ObjectExecutionNode pare
...
+        /// fragments are executed at the same time.
+        /// <br/><br/>
+        /// <see href="http://spec.graphql.org/June2018/#sec-Field-Collection"/> and <see href="http://spec.graphql.org/June2018/#CollectFields()"/>
+        /// </summary>
+        /// <param name="context">The execution context.</param>
+        /// <param name="specificType">The graph type to compare the selection set against.</param>
+        /// <param name="selectionSet">The selection set from the document.</param>
+        /// <param name="fields">A list to append the collected list of fields to; if null, a new list will be created.</param>
+        /// <returns>A list of collected fields</returns>
+        protected virtual Fields CollectFieldsFrom(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet, Fields fields = null)
```
sungam3r, line 233:
```suggestion
        protected virtual Fields CollectFieldsFrom(ExecutionContext context, IGraphType specificType, SelectionSet selectionSet, Fields fields)
```

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -134,13 +177,136 @@ private void SetSubFieldNodes(ExecutionContext context, ObjectExecutionNode pare
...
+        /// Each entry in the grouped field set is a list of fields that share a response key (the alias if defined,
+        /// otherwise the field name). This ensures all fields with the same response key included via referenced
+        /// fragments are executed at the same time.
+        /// <br/><br/>
+        /// <see href="http://spec.graphql.org/June2018/#sec-Field-Collection"/> and <see href="http://spec.graphql.org/June2018/#CollectFields()"/>
+        /// </summary>
+        /// <param name="context">The execution context.</param>
+        /// <param name="specificType">The graph type to compare the selection set against.</param>
+        /// <param name="selectionSet">The selection set from the document.</param>
+        /// <param name="fields">A list to append the collected list of fields to; if null, a new list will be created.</param>
```
sungam3r, line 231:
```suggestion
        /// <param name="fields">A list to append the collected list of fields to; if <see langword="null"/>, a new list will be created.</param>
```
> A list to append

Not a list, dictionary.

Shane32:
```suggestion
        /// <param name="fields">A dictionary to append the collected list of fields to; if <see langword="null"/>, a new dictionary will be created.</param>
```

Shane32: Please confirm updated xml comment.

src/GraphQL/Language/AST/Fields.cs
```diff
@@ -10,85 +7,6 @@ namespace GraphQL.Language.AST
     /// </summary>
     public class Fields : Dictionary<string, Field>
```
sungam3r, line 8: Let's remove this class. I see only one usage of ` public void Add(Field field)`. Also anyone can call "native" `Add`/`[]` methods from dictionary ignoring "special" `Add(Field field)` method.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: Right.  `CoerceValue` is also used by the validation context.  `GetArgumentValues` is used mainly by `ReadonlyResolveFieldContext`, but also `SubscriptionExecutionStrategy`.

src/GraphQL/Language/AST/Fields.cs
Shane32: Sounds good.  I'll do it in a separate PR.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Maybe we shouldn't make absolutely everything `virtual`? Some small parts can be kept private, they are fixed. I agree that it's good to make `virtual` `CollectFields` for example.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: OK. https://github.com/graphql-dotnet/graphql-dotnet/issues/2327

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: OK

sungam3r:
I finished the review. I have a question whether two methods should really be `virtual`. Otherwise, seems good.

Shane32:
> If someone needs to change this method, then we can make it virtual later.

That should be true in this case, since the method is `protected`.  (We are not discussing `public` vs `public virtual`.)

> I finished the review. I have a question whether two methods should really be `virtual`. Otherwise, seems good.

I think this merits further discussion.  However, I can make appropriate changes so we can merge this PR and continue progressing our other PRs towards a v4 release.

Shane32:
(I still need write migration notes.)

sungam3r:
OK. Waiting for migration notes.

sungam3r (approved):

src/GraphQL/Language/AST/Fields.cs
Shane32: #2327

PR #2314, opened 2021-02-24
Ignore duplicates in SchemaTypes.SetGraphType

### LINKED ISSUE #1046: Code first registered interfaces cause errors if out of order.
## Description
When doing code first, the registration order of interfaces and graph types can cause run time errors when trying to serialize an object that implements an interface. The errors always seem to relate to receiving null for the graph parent type, instead of the concrete instance.

### Steps to reproduce

The error occurs when you use the code first approach and register the interface graph type before registering an implementer of the interface.

Here is one of the smallest examples I have where the error occurs:
```
using GraphQL;
using GraphQL.Types;

namespace GraphQLInterfaceErrorTest
{
  class Program
  {
    static void Main()
    {
      var schema = new Schema { Query = new QueryGraphType() };
      // Comment this line out, or reverse it and ImplementationGraphType and it works as expected.
      schema.RegisterType<InterfaceGraphType>();
      schema.RegisterType<ImplementationGraphType>();
      var response = schema.Execute(_ =>
      {
        _.Query = @"
query {
  inst {
    id
  }
}
";
      });
      // expect response does not have errors.
     }
  }

  public class QueryGraphType : ObjectGraphType
  {
    public QueryGraphType()
    {
      Field<InterfaceGraphType>("inst", resolve: c => new Implementation());
    }
  }

  public interface IInterface
  {
    string Id { get; }
  }
  public class InterfaceGraphType : InterfaceGraphType<IInterface>
  {
    public InterfaceGraphType()
    {
      Field(i => i.Id);
    }
  }

  public class Implementation : IInterface
  {
    public string Id => "Data!";
  }

  public class ImplementationGraphType : ObjectGraphType<Implementation>
  {
    public ImplementationGraphType()
    {
      Field(i => i.Id);
      Interface<InterfaceGraphType>();
    }
  }
}
```

### Expected result

```
{
  "data": {
    "inst": {
      "id": "Data!"
    }
  }
}
```

### Actual result

```
{
  "data": {
    "inst": null
  },
  "errors": [
    {
      "message": "Abstract type InterfaceGraphType must resolve to an Object type at runtime for field QueryGraphType.inst with value 'GraphQLInterfaceErrorTest.Implementation', received 'null'.",
      "locations": [
        {
          "line": 3,
          "column": 3
        }
      ],
      "path": [
        "inst"
      ]
    }
  ]
}
```

### Workaround

If you swap the order of these two lines, or remove the first one, no error occurs.
```
schema.RegisterType<InstanceGraphType>();
schema.RegisterType<ImplementationGraphType>();
```
Workarounds are easy, but the bug is not straightforward to figure out based on the error message provided.

### Environment

GraphQL 2.4.0

.NET Core 3.0
.NET Core 2.2
.NET Core 2.1
Happens on different systems and platforms including windows 10 and the asp.net core container, the error is reliably repeatable.

sungam3r:
Fixes #1046

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1156.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1156.cs
@@ -23,8 +23,8 @@ namespace GraphQL.Tests.Bugs
 }";
             var result = AssertQueryWithErrors(query, null, expectedErrorCount: 1, executed: false);
             result.Errors[0].Message.ShouldBe("Error executing document.");
-            result.Errors[0].InnerException.Message.ShouldBe(@"Unable to register GraphType 'GraphQL.Tests.Bugs.Type2' with the name 'MyType';
-the name 'MyType' is already registered to 'GraphQL.Tests.Bugs.Type1'.");
+            result.Errors[0].InnerException.Message.ShouldBe(@"Unable to register GraphType 'GraphQL.Tests.Bugs.Type2' with the name 'MyType'.
+The name 'MyType' is already registered to 'GraphQL.Tests.Bugs.Type1'. Check your schema configuration.");
         }
     }
 
--- a/src/GraphQL.Tests/Bugs/Bug2279.cs
+++ b/src/GraphQL.Tests/Bugs/Bug2279.cs
@@ -15,10 +15,8 @@ namespace GraphQL.Tests.Bugs
             var e = Should.Throw<InvalidOperationException>(() => s.Initialize());
             var t1 = typeof(Bug2279GraphType<int>).FullName;
             var t2 = typeof(Bug2279GraphType<string>).FullName;
-            //nl will be determined at compile time dictated by how git is configured to handle line endings, not what environment the code was ran in
-            var nl = @"
-";
-            e.Message.ShouldBe($"Unable to register GraphType '{t1}' with the name 'Bug2279GraphType_1';{nl}the name 'Bug2279GraphType_1' is already registered to '{t2}'.");
+            e.Message.ShouldBe(@$"Unable to register GraphType '{t1}' with the name 'Bug2279GraphType_1'.
+The name 'Bug2279GraphType_1' is already registered to '{t2}'. Check your schema configuration.");
         }
     }
 
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -697,18 +697,18 @@ Make sure that your ServiceProvider is configured correctly.");
 
             if (Dictionary.TryGetValue(typeName, out var existingGraphType))
             {
-                if (ReferenceEquals(existingGraphType, type))
+                if (ReferenceEquals(existingGraphType, type) || existingGraphType.GetType() == type.GetType())
                 {
-                    // nothing to do
-                }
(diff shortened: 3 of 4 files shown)


src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -697,18 +697,18 @@ private void SetGraphType(string typeName, IGraphType type)
...
             if (Dictionary.TryGetValue(typeName, out var existingGraphType))
             {
-                if (ReferenceEquals(existingGraphType, type))
+                if (ReferenceEquals(existingGraphType, type) || existingGraphType.GetType() == type.GetType())
                 {
-                    // nothing to do
-                }
-                else if (existingGraphType.GetType() == type.GetType())
-                {
-                    Dictionary[typeName] = type; // this case worked before overwriting the old value
```
sungam3r, line 706: I remember this place, I never liked it.

src/GraphQL.Tests/Bugs/Bug2279.cs
```diff
@@ -15,10 +15,8 @@ public void NoDuplicateTypeNames()
             var e = Should.Throw<InvalidOperationException>(() => s.Initialize());
             var t1 = typeof(Bug2279GraphType<int>).FullName;
             var t2 = typeof(Bug2279GraphType<string>).FullName;
-            //nl will be determined at compile time dictated by how git is configured to handle line endings, not what environment the code was ran in
-            var nl = @"
```
sungam3r, line 19: @Shane32 I didn't understand why it didn't work for you. As you can see, there was already another test with the same message. It worked before. I changed both tests and they both work after.

Shane32: I dunno. 🤷‍♂️   👍

Shane32: But you're still using `@` which of course works.  I tried using `Environment.NewLine` and had a problem.  I didn't know there was a `@$` syntax for literals. 👍

src/GraphQL.Tests/Bugs/Bug1156.cs
```diff
@@ -23,8 +23,8 @@ public void duplicated_type_names_should_throw_error()
 }";
             var result = AssertQueryWithErrors(query, null, expectedErrorCount: 1, executed: false);
             result.Errors[0].Message.ShouldBe("Error executing document.");
-            result.Errors[0].InnerException.Message.ShouldBe(@"Unable to register GraphType 'GraphQL.Tests.Bugs.Type2' with the name 'MyType';
-the name 'MyType' is already registered to 'GraphQL.Tests.Bugs.Type1'.");
+            result.Errors[0].InnerException.Message.ShouldBe(@"Unable to register GraphType 'GraphQL.Tests.Bugs.Type2' with the name 'MyType'.
+The name 'MyType' is already registered to 'GraphQL.Tests.Bugs.Type1'. Check your schema configuration.");
```
Shane32, line 27: This test also uses the `@` syntax, which matches the way it was thrown.

Shane32 (approved):

src/GraphQL.Tests/Bugs/Bug2279.cs
sungam3r: OK

PR #2316, opened 2021-02-24
Remove ValidationOnSchemaInitialize

### LINKED ISSUE #2204: ValidationOnSchemaInitialize is useless, and `Validation` comments is misleading
`NameValidator.ValidationOnSchemaInitialize` is useless and misleading because if someone tried to use a field/argument name that has invalid characters, the parser library would not read a query with those fields.

`NameValidator.Validation` has misleading comments, because it only allows bypassing validation when there is an `INameConverter` which fixes the names to conform to the GraphQL specification (so that queries with those names can be read by the parser library).

I suggest making the former `internal` for now, and updating the comments on the latter.

Shane32:
Fixes #2204 

Also renamed `GlobalSwitches.Validation` to `GlobalSwitches.NameValidation` to better indicate what this delegate is validating.

### DIFF
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -530,7 +530,7 @@ default continues to be `ResolverType.Resolver`.
 * `ObjectExecutionNode.SubFields` property type was changed from `Dictionary<string, ExecutionNode>` to `ExecutionNode[]`
 * `ExecutionNode.IsResultSet` has been removed
 * `ExecutionNode.Source` is read-only; additional derived classes have been added for subscriptions
-* `NameValidator.ValidateName` accept an enum instead of a string for their second argument
+* `NameValidator.ValidateName` accepts an enum instead of a string for its second argument
 * `NameValidator.ValidateNameOnSchemaInitialize` has been made internal and `ValidationOnSchemaInitialize` has been removed
 * `ExecutionNode.PropagateNull` must be called before `ExecutionNode.ToValue`; see reference implementation
 * `IDocumentValidator.ValidateAsync` does not take `originalQuery` parameter; use `Document.OriginalQuery` instead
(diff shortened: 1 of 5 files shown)


src/GraphQL/Utilities/NameValidator.cs
```diff
@@ -12,14 +12,14 @@ public static class NameValidator
...
-        public static void ValidateName(string name, NamedElement type) => GlobalSwitches.Validation(name, type);
+        public static void ValidateName(string name, NamedElement type) => GlobalSwitches.NameValidation(name, type);
 
         /// <summary>
         /// Validates a specified name during schema initialization.
         /// </summary>
         /// <param name="name">GraphQL name.</param>
         /// <param name="type">Type of element: field, type, argument, enum.</param>
-        public static void ValidateNameOnSchemaInitialize(string name, NamedElement type) => GlobalSwitches.ValidationOnSchemaInitialize(name, type);
+        public static void ValidateNameOnSchemaInitialize(string name, NamedElement type) => ValidateDefault(name, type);
```
sungam3r, line 22: Remove ? Only two usages from `private` method.

sungam3r (approved):
See my comment.

src/GraphQL/Utilities/NameValidator.cs
Shane32: I left it there because then it was an obvious what the intent of the code was versus calling `ValidateDefault` directly.  We could make it `internal` if you like.  I'm sure JIT would optimize out this call so there shouldn't be a performance difference.

sungam3r: OK with internal.

docs2/site/docs/migrations/migration4.md
```diff
@@ -532,7 +530,8 @@ default continues to be `ResolverType.Resolver`.
 * `ObjectExecutionNode.SubFields` property type was changed from `Dictionary<string, ExecutionNode>` to `ExecutionNode[]`
 * `ExecutionNode.IsResultSet` has been removed
 * `ExecutionNode.Source` is read-only; additional derived classes have been added for subscriptions
-* `NameValidator.ValidateName` and `NameValidator.ValidateNameOnSchemaInitialize` accept an enum instead of a string for their second argument
+* `NameValidator.ValidateName` accept an enum instead of a string for their second argument
```
Shane32, line 533:
```suggestion
* `NameValidator.ValidateName` accepts an enum instead of a string for its second argument
```

PR #2320, opened 2021-02-24
Remove ValueConverter from ParseValue implementations

### LINKED ISSUE #2149: Fix for variable coercion
See `Add test for variable coercion` (#1919)

sungam3r:
fixes #2149

Calling `ValueConverter.ConvertTo` from scalars is evil. This method should only be called from `GetArgument<T>` to convert values to user supplied types.


sungam3r:
We need to change all scalars in a similar way.

sungam3r:
Thanks to this fix, I changed several tests. Honestly, error messages now look more logical.

src/GraphQL.Tests/Bugs/Bug1773.cs
```diff
@@ -60,7 +60,7 @@ public void list_throws_for_invalid_type()
         {
             // TODO: does not yet fully meet spec (does not return members of lists that are able to be serialized, with nulls and individual errors for unserializable values)
             AssertQueryWithError("{testListInvalidType}", "{\"testListInvalidType\": null}", "Error trying to resolve field 'testListInvalidType'.", 1, 2, new[] { "testListInvalidType" },
-                new FormatException("Input string was not in a correct format."), localizedMessage: "Входная строка имела неверный формат.");
+                new InvalidOperationException("Unable to serialize 'test' to 'Int' for list index 0."));
```
sungam3r, line 63: Look at the tests above and below - all now return `InvalidOperationException`.

src/GraphQL.Tests/Bugs/ScalarsInputTest.cs
```diff
@@ -172,21 +180,40 @@ public void Should_Accept_Long_As_Long_In_Literal()
...
+        }
+
         [Fact]
         public void Should_Not_Accept_String_As_Long_In_Literal()
         {
             var query = @"mutation { long(number: ""100"") }";
             string expected = null;
-            AssertQueryWithErrors(query, expected, expectedErrorCount: 1, executed: false);
+            var result = AssertQueryWithErrors(query, expected, expectedErrorCount: 1, executed: false);
+            result.Errors[0].Message.ShouldBe("Argument \"number\" has invalid value \"100\".\nExpected type \"Long\", found \"100\".");
```
sungam3r, line 207: @Shane32 Also I would like to change all `"` to `'` from error messages in separate PR.

Shane32:
Looks good, but due to the amount of changes in #2321, I would prefer to merge #2321 first.  I'll be happy to fix the merge conflicts here afterwards, as well as make similar changes for the rest of the scalars.

sungam3r:
Is it really all in one PR? The review will not be quick.

Shane32:
@sungam3r I propose renaming this PR to be "Remove ValueConverter from scalar graph types"

Do you want to continue with it or should I take it over?  Or do you want me to create a new PR to that effect?

sungam3r:
Take it.

Shane32:
@sungam3r Before I continue, we need to decide what types will be allowed for numeric scalar types?  Similar to answering what will be allowed for a string scalar type.

For example, this code would be most similar to what is currently supported:
```cs
    public class IntGraphType : ScalarGraphType
    {        public override object ParseValue(object value) => value switch
        {
            int _ => value, // no boxing
            null _ => null,
            _ => Convert.ToInt32(value)
        };
    }
```

Shane32:
Alternatively, we can disallow any non-matching types:

```cs
    public class IntGraphType : ScalarGraphType
    {
        public override object ParseValue(object value) => value switch
        {
            int _ => value, // no boxing
            _ => null
        };
    }
```

Shane32:
We can disallow non-matching types except for certain cases, like specifically when you return an `int` for a `LongGraphType`.

Another option would be to explicitly support the full range of integer numeric types, but not use `Convert.ToInt32`.  Have to decide if we want to support floating point types, etc.

```cs
    public class IntGraphType : ScalarGraphType
    {
        public override object ParseValue(object value) => value switch
        {
            int _ => value, // no boxing
            long => checked((int)value),
            byte => (int)value,
            sbyte => (int)value,
            //etc
            _ => null,
        };
    }
```

Now, `ParseValue` probably will need to support the types that `System.Text.Json` deserializes to by default.  In practice, probably never any other type.  But we should check -- for longs, it might return a float.

So this question is really more for `Serialize`..... anyway, thoughts???

sungam3r:
I would prefer explicit cases with numeric types (maybe without floating ones for integers) without Convert calls.

Shane32:
> I would prefer explicit cases with numeric types (maybe without floating ones for integers) without Convert calls.

All numeric types?  So like you can pass a sbyte to a IntGraphType and it will convert it?

sungam3r:
Yes. I understand that this may look redundant. But in fact, this is how it worked for `ValueConverter` calls.

Shane32:
Propose to close in favor of #2341.  This branch was merged into that one.

sungam3r:
OK.

PR #2324, opened 2021-02-24
Change generic type constraint of NonNullGraphType

Shane32:
Of note, `ResolvedType` is of type `IGraphType`, and `ListGraphType<>`'s constraint is already `IGraphType`


sungam3r:
Ooops.

sungam3r (approved):

PR #2325, opened 2021-02-25
Fix for GraphQLExtensions.GetNamedType/IsInputType/IsOutputType

### LINKED ISSUE #1837: InputGraphType checks field types based on Resolved Type only.
## Description

When you add a field to an InputObjectGraphType it verifies if the type and field is also an input type. But it uses only the ResolvedType property for that. If you use the generic version of the types the ResolvedType is not resolved yet and therefore throws an exception.

### Steps to reproduce

```
public class InputDroidType : InputObjectGraphType
{
  public InputDroidType()
  {
    Field<NonNullGraphType<IdGraphType>>("id");
    Field<ListGraphType<StringGraphType>>("weapons");
  }
}
```

Version: 3.0.0

sungam3r:
fixes #1837

### DIFF
--- a/src/GraphQL/GraphQLExtensions.cs
+++ b/src/GraphQL/GraphQLExtensions.cs
@@ -36,8 +36,9 @@ namespace GraphQL
         /// </summary>
         public static bool IsLeafType(this IGraphType type)
         {
-            var namedType = type.GetNamedType();
-            return namedType is ScalarGraphType;
+            var (namedType, namedType2) = type.GetNamedTypes();
+            return namedType is ScalarGraphType ||
+                   typeof(ScalarGraphType).IsAssignableFrom(namedType2);
         }
 
         // https://graphql.github.io/graphql-spec/June2018/#sec-Input-and-Output-Types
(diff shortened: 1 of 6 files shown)


src/GraphQL/GraphQLExtensions.cs
```diff
@@ -107,33 +110,54 @@ public static bool IsOutputType(this Type type)
...
-            {
-                NonNullGraphType nonNull => GetNamedType(nonNull.ResolvedType),
-                ListGraphType list => GetNamedType(list.ResolvedType),
-                _ => type
-            };
+            if (type == null)
+                return null;
+
+            var (namedType, _) = type.GetNamedTypes();
+            return namedType ?? throw new NotSupportedException("Please set ResolvedType property before calling this method or call GetNamedType(this Type type) instead");
```
sungam3r, line 160: A key part of this PR. The `GetNamedType(this IGraphType type)` method was only intended for use by unwrapping other resolved types. The `GetNamedType(this Type type)` method has no problems in this regard. After throwing the exception, I saw that the calling methods needed to be changed. Typically, they didn't need the type itself. Only in one case did I have to actually create a type. See `SchemaTypes`. We had no errors all this time because we did not configure the graph types this way.

Shane32 (approved):

PR #2332, opened 2021-02-28
SchemaBuilder.AllowUnknownTypes/AllowUnknownFields 

### LINKED ISSUE #1155: Schema first should fail on schema buildup when types not match schema
## Summary

Schema first should fail on schema build up (application start) if schema does not match the given types (even if the types are not used).

### Basic example
```csharp
public class Query
{
    public Test Test()=> new Test();
}

public class Test
{
    public string Id => "foo";
}
               
{
    var schema = Schema.For(@"
        type Test{
            id: ID!
            name: String!
        }

        type Query {
            test: Test!
        }
        ", _ =>
    {
        _.Types.Include<Query>();
    });

    {
        _.Query = "{ test { id } }";
    });
}
```
In the given example I would expect  `Schema.For` or `schema.Execute` fail with an exception that tells my that there is a property `name` on type `Test` which has no resolver assigned.

### Motivation

While playing with the newer schema first stuff, I discovered that its pretty easy to get stuff wrong without knowing it.

My expectation is that it fails on start/schema build up or that there is a method like `schema.validate()` which tells me that I missed that my schema is wrong. 

The current approach means that I or my users have to do a query to know that the schema fails.

sungam3r:
fixes #1155 

I also fixed a few tests. We actually have a lot of tests that will fail if we set new properties as `false`. At first I wanted to change all of them, but then I changed my mind. It's not worth it.


src/GraphQL.Tests/Bugs/Issue1189.cs
```diff
@@ -30,6 +30,7 @@ public class Issue1189 : SchemaBuilderTestBase
         public void Issue1189_Should_Work(Type resolverType, string errorMessage, string code)
         {
             Builder.Types.Include<Issue1189_Query>();
+            Builder.Types.For("Character").Type = typeof(Issue1189_Character);
```
sungam3r, line 33: Ping @lanwin . This is a good example of the case you described.

lanwin: Yes it seems so. But my wish would be that I know that this is wrong before someone issues the first query against the api.

sungam3r: Of course there will be an error while building the schema.

sungam3r:
Ping @Shane32

Shane32 (approved):

PR #2341, opened 2021-03-01
Allow converting scalar values to/from null, don't call ValueConverter, fix list handling of errors

Shane32:
Fixes #2333 and replaces #2320

* [x] Update calls to ParseValue
* [x] Update calls to ParseLiteral
* [x] Update xml comments for all methods
* [x] No change to calls to CanParseLiteral / CanParseValue
* [x] Update calls to Serialize
* [x] Update calls to ToAST
* [x] Fix SchemaBuilder's handling of default values
* [x] Confirm new behavior: Scalar numeric types do not allow strings
* [x] Confirm new behavior: Scalar integer numeric types do not allow floating-point values
* [x] Confirm new behavior: Enumeration types do not allow strings
* [x] Confirm new behavior: String type does not allow non-string values
* [x] Confirm new behavior: Id type allows any numeric or string type of incoming value; serialize always returns string
* [x] Add more tests

Scalars:
* [x] All redone

~~Note: this PR does not include removing the calls to `ValueConverter`. See #2320~~

PR #2346, opened 2021-03-01
Fix IsValidDefault for single-item list + add tests

### LINKED ISSUE #2334: Fix schema-first coercion of default values
Wait. What about coersion from a value to single-item list ?
```graphql
type Query {
  field1(arg: [String] = "aaa")
  field2(arg: [String] = ["aaa"])
}
```

field1 and field2 are effectively the same as I think.

_Originally posted by @sungam3r in https://github.com/graphql-dotnet/graphql-dotnet/pull/2321#discussion_r584118420_

sungam3r:
fixes #2334
also fixes behavior for non-null graph types in `ToAST`

### DIFF
--- a/src/GraphQL.Tests/Extensions/GraphQLExtensionsTests.cs
+++ b/src/GraphQL.Tests/Extensions/GraphQLExtensionsTests.cs
@@ -4,8 +4,10 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using GraphQL.Dummy;
+using GraphQL.Language.AST;
 using GraphQL.StarWars.Types;
 using GraphQL.Types;
+using GraphQL.Utilities;
 using Shouldly;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


sungam3r:
@Shane32 I also added tests for `ToAST`. Look at my `TODO:` comments. What to do when calling `ToAST` with non-null graph type and `null` value? Exception?

Shane32 (approved):

src/GraphQL.Tests/Extensions/GraphQLExtensionsTests.cs
```diff
@@ -4,20 +4,173 @@
...
+            public IEnumerator<object[]> GetEnumerator()
+            {
+                yield return new object[] { null, true, new NullValue() };
+
+                yield return new object[] { new BooleanGraphType(), true, new BooleanValue(true) };
+                yield return new object[] { new BooleanGraphType(), false, new BooleanValue(false) };
+                yield return new object[] { new BooleanGraphType(), null, new NullValue() };
+
+                yield return new object[] { new NonNullGraphType(new BooleanGraphType()), false, new BooleanValue(false) };
+                //yield return new object[] { new NonNullGraphType(new BooleanGraphType()), null, false }; //TODO: exception ?
```
sungam3r, line 97: What about that case?

Shane32: It seems reasonable to throw an `InvalidOperationException` if you are trying to coerce a value that is invalid for the type.  In this case, trying to cast a null value to a non-null type, so-to-speak.  Not much different than trying to convert the string "abc" to a BooleanGraphType AST representation.

Shane32:
> @Shane32 I also added tests for `ToAST`. Look at my `TODO:` comments. What to do when calling `ToAST` with non-null graph type and `null` value? Exception? Something like that:
> 
> ```cs
>  public static IValue ToAST(this IGraphType type, object value)
>         {
>             if (type is NonNullGraphType nonnull)
>             {
>                 if (value == null)
>                     throw new InvalidArgumentException(nameof(value), "Unable to convert null value of the non-null type '{nonNull.ResolvedType}' to an AST representation.);
>                 return ToAST(nonnull.ResolvedType, value);
>             }
> ```
> 
> You can choose a better message.

Let's use `InvalidOperationException` to match the other exceptions in #2341 .  Or change them in both places.

sungam3r:
OK, InvalidOperation.

sungam3r:
@Shane32 I changed error messages a bit.

Shane32 (approved):

Issue #2253, opened 2021-02-04
Stop execution as early as possible


sungam3r:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/2240/files?file-filters%5B%5D=.cs&w=1#r570626595

Perhaps after that the `ClearErrorNodes` method will not be required.

PR #2349, opened 2021-03-02
return members of lists that are able to be serialized, with nulls and individual errors for unserializable values

sungam3r:
Also see #2253

### DIFF
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -361,7 +361,8 @@ namespace GraphQL.Execution
 
                             try
                             {
-                                serializedResult = valueNode.GraphType.Serialize(d);
+                                serializedResult = valueNode.GraphType.Serialize(d)
+                                    ?? throw new InvalidOperationException($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}' for list index {index}.");
                             }
                             catch (Exception ex)
                             {
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -357,8 +357,41 @@ namespace GraphQL.Execution
                         }
                         else if (node is ValueExecutionNode valueNode)
                         {
-                            node.Result = valueNode.GraphType.Serialize(d)
-                                ?? throw new InvalidOperationException($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}' for list index {index}.");
+                            object serializedResult = null;
+
+                            try
+                            {
+                                serializedResult = valueNode.GraphType.Serialize(d);
+                            }
+                            catch (Exception ex)
+                            {
+                                // Stop execution as early as possible, it doesn't make sense to set the rest of the array elements and then execute nodes
+                                if (listType.ResolvedType is NonNullGraphType)
+                                {
+                                    throw new InvalidOperationException($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}' for list index {index}.", ex);
+                                }
+                                else
+                                {
+                                    SetNodeError(context, node, new ExecutionError($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}'.", ex));
+                                    arrayItems.Add(node);
+                                    return;
+                                }
+                            }
+
+                            node.Result = serializedResult;
+
+                            if (serializedResult == null)
+                            {
+                                // Stop execution as early as possible, it doesn't make sense to set the rest of the array elements and then execute nodes
+                                if (listType.ResolvedType is NonNullGraphType)
+                                {
+                                    throw new InvalidOperationException($"Cannot return a null member within a non-null list for list index {index}.");
+                                }
+                                else
+                                {
+                                    SetNodeError(context, node, new ExecutionError("Cannot return a null member within a non-null list."));
+                                }
+                            }
                         }
                     }
 
@@ -366,6 +399,7 @@ namespace GraphQL.Execution
                 }
                 else
                 {
+                    // Stop execution as early as possible, it doesn't make sense to set the rest of the array elements and then execute nodes
                     if (listType.ResolvedType is NonNullGraphType)
                     {
                         throw new InvalidOperationException($"Cannot return a null member within a non-null list for list index {index}.");
@@ -487,6 +521,7 @@ namespace GraphQL.Execution
                     }
                     else if (node is ValueExecutionNode valueNode)
                     {
+                        //TODO: need test here as for https://github.com/graphql-dotnet/graphql-dotnet/issues/2348 , exception not always should prevents execution
                         node.Result = valueNode.GraphType.Serialize(node.Result)
                             ?? throw new InvalidOperationException($"Unable to serialize '{node.Result}' to '{valueNode.GraphType.Name}'.");
                     }
@@ -498,7 +533,7 @@ namespace GraphQL.Execution
             }
             catch (Exception ex)
             {
-                if (ProcessNodeUnhandledException(context, node, ex))
+                if (ProcessNodeUnhandledException(context, node, ex)) //TODO: change? No exception from user resolver here
                     throw;
             }
         }
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -364,7 +364,21 @@ namespace GraphQL.Execution
                                 serializedResult = valueNode.GraphType.Serialize(d)
                                     ?? throw new InvalidOperationException($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}' for list index {index}.");
                             }
-                            catch (Exception ex)
+            catch (OperationCanceledException) when (context.CancellationToken.IsCancellationRequested)
+            {
+                throw;
+            }
+            catch (Exception ex) when (listType.ResolvedType is NonNullGraphType)
+            {
+                // Stop execution as early as possible, it doesn't make sense to set the rest of the array elements and then execute nodes
+                throw;
+            }
+            catch (ExecutionError error)
+            {
+                SetNodeError(context, node, error);
+                arrayItems.Add(node);
+            }
+            catch (Exception ex)
                             {
                                 // Stop execution as early as possible, it doesn't make sense to set the rest of the array elements and then execute nodes
                                 if (listType.ResolvedType is NonNullGraphType)
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -340,6 +340,8 @@ namespace GraphQL.Execution
             // local function uses 'struct closure' without heap allocation
             void SetArrayItemNode(object d)
             {
+                context.CancellationToken.ThrowIfCancellationRequested();
+
                 if (d != null)
                 {
                     var node = BuildExecutionNode(parent, itemType, parent.Field, parent.FieldDefinition, index);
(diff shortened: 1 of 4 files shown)


src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -487,6 +521,7 @@ protected virtual void CompleteNode(ExecutionContext context, ExecutionNode node
                     }
                     else if (node is ValueExecutionNode valueNode)
                     {
+                        //TODO: need test here as for https://github.com/graphql-dotnet/graphql-dotnet/issues/2348 , exception not always should prevents execution
                         node.Result = valueNode.GraphType.Serialize(node.Result)
```
sungam3r, line 525: One more `Serialize` call with `?? throw`.

sungam3r:
@Shane32 I just noticed your `//TODO:` comments at the very end. I had already forgotten about them, but when I read it, I remembered what kind of bug it was.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +357,49 @@ void SetArrayItemNode(object d)
...
+                            {
+                                // Stop execution as early as possible, it doesn't make sense to set the rest of the array elements and then execute nodes
+                                if (listType.ResolvedType is NonNullGraphType)
+                                {
+                                    throw new InvalidOperationException($"Cannot return a null member within a non-null list for list index {index}.");
+                                }
+                                else
+                                {
+                                    SetNodeError(context, node, new ExecutionError("Cannot return a null member within a non-null list."));
+                                }
```
sungam3r, line 393: After #2341 this block of code may change a bit.

sungam3r:
This fix should be documented, among other changes about scalars and execution strategy. Although only the internal logic changes without the public API, the format of the responses changes for consumers (it has become correct).

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +357,49 @@ void SetArrayItemNode(object d)
...
                         {
-                            node.Result = valueNode.GraphType.Serialize(d)
-                                ?? throw new InvalidOperationException($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}' for list index {index}.");
+                            object serializedResult = null;
+
+                            try
+                            {
+                                serializedResult = valueNode.GraphType.Serialize(d);
+                            }
+                            catch (Exception ex)
```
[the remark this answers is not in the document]
Shane32: - If the http request has been canceled, it must throw.  Appears in `ExecuteAsync` and `CompleteDataLoaderNodeAsync` but also missing from `CompleteNode`.
- If list is a non-null type, it should `throw` now, with the exception being handled by the caller (typically `CompleteNode`)
- If the error is an `ExceptionError`, the error should not be wrapped

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +357,49 @@ void SetArrayItemNode(object d)
...
+                                if (listType.ResolvedType is NonNullGraphType)
+                                {
+                                    throw new InvalidOperationException($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}' for list index {index}.", ex);
+                                }
+                                else
+                                {
+                                    SetNodeError(context, node, new ExecutionError($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}'.", ex));
+                                    arrayItems.Add(node);
+                                    return;
+                                }
```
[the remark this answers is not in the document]
Shane32: Otherwise, the error should be wrapped by `ProcessNodeUnhandledException`, which depending on the execution options may instruct the exception to be thrown.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +357,49 @@ void SetArrayItemNode(object d)
...
+                                // Stop execution as early as possible, it doesn't make sense to set the rest of the array elements and then execute nodes
+                                if (listType.ResolvedType is NonNullGraphType)
+                                {
+                                    throw new InvalidOperationException($"Cannot return a null member within a non-null list for list index {index}.");
+                                }
+                                else
+                                {
+                                    SetNodeError(context, node, new ExecutionError("Cannot return a null member within a non-null list."));
+                                }
+                            }
```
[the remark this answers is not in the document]
Shane32: This code is unnecessary.  `?? throw new....` handles this scenario perfectly with a single line of code.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: > After #2341 this block of code may change a bit.

Yes it will chance significantly.  But this PR applies to the current state of the project.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: > One more `Serialize` call with `?? throw`.

Which is correct at this time.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -498,7 +533,7 @@ protected virtual void CompleteNode(ExecutionContext context, ExecutionNode node
             }
             catch (Exception ex)
             {
-                if (ProcessNodeUnhandledException(context, node, ex))
+                if (ProcessNodeUnhandledException(context, node, ex)) //TODO: change? No exception from user resolver here
```
Shane32, line 536: This is necessary.  Any time that a resolver returns a value that is invalid for the graph type, it causes an error to be thrown within Serialize.  This is an unhandled "server" exception and must be processed by the unhandled exception delegate, or depending on the configuration of the execution options, throw immediately back to the caller of ExecuteAsync with no wrapping.

Shane32:
```suggestion
                if (ProcessNodeUnhandledException(context, node, ex))
```

Shane32 (changes requested):
It is very important that error messages thrown from `Serialize` not be wrapped by anything other than `ProcessNodeUnhandledException`, and only if the exception is not an `ExecutionError`.  This maintains the ability of the unhandled exception delegate to manually wrap errors, perform logging, and so on.

Of course `Serialize` would never normally throw an `ExecutionError`, but if it does, that is because someone wrote a custom scalar to explicitly throw that type of error for certain conditions, and we should process it just as it would be handled anywhere else.

Shane32:
Please note that my code suggestions are cut-and-paste snippets (and hence are not tabbed properly).

src/GraphQL/Execution/ExecutionStrategy.cs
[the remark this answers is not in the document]
sungam3r: I understand. I guessed that this code can be shortened.

src/GraphQL/Execution/ExecutionStrategy.cs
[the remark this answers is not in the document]
sungam3r: > If the http request has been canceled, it must immediately throw

OK, but I added check in the beginngn of method. Anyway inside `Serialize` we have no access to cancellation token.

src/GraphQL/Execution/ExecutionStrategy.cs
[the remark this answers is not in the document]
Shane32: If you agree, let's merge the code suggestion, or wait on resolving the conversation until we agree on a resolution.

src/GraphQL/Execution/ExecutionStrategy.cs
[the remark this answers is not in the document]
Shane32: If you agree, let's merge the code suggestion, or wait on resolving the conversation until we agree on a resolution.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: It seems now I understood. I was misled for the comment on the ProcessNodeUnhandledException method. I changed it.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -516,7 +544,7 @@ protected virtual void SetNodeError(ExecutionContext context, ExecutionNode node
         }
 
         /// <summary>
-        /// Processes unhandled field resolver exceptions.
+        /// Processes unhandled field resolver exceptions as well as any other exceptions when serialized value is invalid for the graph type.
```
sungam3r, line 547: Right?

src/GraphQL/Execution/ExecutionStrategy.cs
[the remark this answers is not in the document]
sungam3r: I made changes.

sungam3r:
@Shane32 Please review again. I tried to simplify the branching of the conditions in `SetArrayItemNode` and minimize it. I am still not sure about using certain texts of messages and exceptions:
1. `ExecutionError` vs `InvalidOperation`
2. `Cannot return a null member within a non-null list.` vs `Unable to serialize 'value' to 'type'`
3. `for list index {index}.` or `(empty)`

Now I use `for list index {index}.` only for non-null cases ("fatal" cases) since otherwise index is already present in `Path`. We can always write `for list index {index}.` though.

sungam3r:
And maybe also add `ProcessNodeUnhandledException` somewhere?

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Looks good to me. 👍

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +359,40 @@ void SetArrayItemNode(object d)
...
+                            try
+                            {
+                                object result = valueNode.GraphType.Serialize(d);
+                                if (result == null)
+                                {
+                                    defaultException = true;
+                                    if (listType.ResolvedType is NonNullGraphType)
+                                        throw new InvalidOperationException($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}' for list index {index}.");
+                                    else
+                                        throw new ExecutionError($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}'.");
```
Shane32, line 372: I do not understand this.  Why the difference?  The problem is that `Serialize` was unable to serialize a value returned from the field resolver.  For example, "abc" was returned for an `IntGraphType`.  It does not matter if it is a non-null list type or not.  The problem lies in the resolver; it did not return data in the proper format.

`ExecutionError` indicates there is a client-side problem.  That is not the case as described above; the returned value from the field resolver cannot be coerced to the proper type.  Therefor, it is an `InvalidOperationException` either way.  And there is no reason to have different error messages.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +359,40 @@ void SetArrayItemNode(object d)
...
+                                }
+                                node.Result = result;
+                            }
+                            catch (Exception error) when (listType.ResolvedType is NonNullGraphType)
```
Shane32, line 376: This should first catch `ObjectDisposedExceptions` when cancellation is requested, and throw immediately.  `ThrowIfCancellationRequested` is not a suitable replacement.

For example, if you wrote graphql code to call a REST API, and your call to graphql was hanging, and you pressed "stop" in your browser or similar to stop the call, you would want the full stack trace returned to the caller.  It should not be capturing and rethrowing any other type of exception, or returning a failed execution document.  It should throw with the original stack trace so it can be seen where the execution died.

Shane32: Actually, it's impossible for Serialize to throw (due to the cancellation token) because it's not an async call.  So nevermind.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +359,40 @@ void SetArrayItemNode(object d)
...
+                            {
+                                // Stop execution as early as possible, it doesn't make sense to set the rest of the array elements and then execute nodes
+                                if (error is ExecutionError || defaultException)
+                                    throw;
+                                else
+                                    throw new InvalidOperationException($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}' for list index {index}.", error);
```
Shane32, line 382:
```suggestion
                                throw;
```
Exceptions are properly wrapped by the unhandled exception handler.  They should never be double-wrapped.  Perhaps it is too bad that they do not have the list index inside the message, but that's simply too bad.  Please remove.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +359,40 @@ void SetArrayItemNode(object d)
...
+                            }
+                            catch (Exception error) // listType.ResolvedType allows null
```
Shane32, line 384: This needs to catch ExecutionError and SetNodeError directly without wrapping, before resorting to this catch block

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -357,15 +359,40 @@ void SetArrayItemNode(object d)
...
+                            {
+                                SetNodeError(context, node, error is ExecutionError e ? e : new ExecutionError($"Unable to serialize '{d}' to '{valueNode.GraphType.Name}'.", error));
```
Shane32, line 386: Exceptions must be wrapped by the unhandled exception handler, not directly.

Shane32:
I am very disappointed that our tests are not covering these scenarios.  Perhaps I will write more tests.  It should be pretty well defined that:
- ExecutionError exceptions are not wrapped at all
- OperationCancelledException bubbles up to the caller
- Exceptions are not wrapped twice
- Exceptions thrown within Serialize are handled by the unhandled exception handler

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: If there is an issue with the error's path property not returning the proper index for errors within non-null lists, then perhaps we need to solve this another way.  But we cannot double-wrap exceptions.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: `ExecutionError` because in that case it will be used only for `SetNodeError`. `SetNodeError` takes `ExecutionError`, not `Exception`. OK?

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: > it's impossible for Serialize to throw 

I spoke about it.

sungam3r:
> ExecutionError exceptions are not wrapped at all

Add `if (exception is ExecutionError) throw ...` inside `  public ExecutionError(string message, Exception exception)`

It's 3 am here. I'm done. You can change this PR at your discretion.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Modify `Exception.Data` ?

sungam3r:
Closing?

Shane32:
I would say so.  PR #2341 by necessity will overwrite the work done here due to the new handling of scalar types.

Shane32:
I did review the code and add the new test into #2341 .

PR #2357, opened 2021-03-05
Scalar documentation

Shane32:


docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -48,16 +54,12 @@ mutation AddVector($vector3: Vector3!) {
 }
 ```
 
-Vectors should be received in a more structured format:
+Vectors should be returned in the same format:
 
 ```json
 {
     "data": {
-        "getVector": {
```
sungam3r, line 56: Strange. Is it a scalar? This is an object. Documentation was incorrect?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -126,49 +128,75 @@ public void ConfigureServices(IServiceCollection services)
...
 
 public override object ParseLiteral(IValue value)
 {
-    return value is StringValue stringValue
-        ? ParseValue(stringValue.Value)
-        : null;
+    if (value is NullValue)
+        return null;
+
+    if (value is StringValue stringValue) return ParseValue(stringValue.Value);
```
sungam3r, line 159:
```suggestion
    if (value is StringValue stringValue)
        return ParseValue(stringValue.Value);
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +211,216 @@ public class ExampleSchema : Schema
             throw new FormatException($"Failed to parse {nameof(Vector3)} from input '{vector3InputString}'. Input should be a string of three comma-separated floats in X Y Z order, ex. 1.0,2.0,3.0");
         }
     }
+
+    return ThrowValueConversionError(value);
 }
 ```
 
-6. Prepare to support conversion of `Vector3` to an AST node. This allows GraphQL.NET to treat values parsed from variables the same as arguments, which arrive for execution as AST nodes. 
+5. Implement `ScalarGraphType.Serialize` so `Vector3` instances can be sent to the client.
```
sungam3r, line 219:
```suggestion
6. Implement `ScalarGraphType.Serialize` so `Vector3` instances can be sent to the client.
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +211,216 @@ public class ExampleSchema : Schema
...
+Keep in mind that a null value also required to be handled.
```
sungam3r, line 220:
```suggestion
Keep in mind that a `null` value also required to be handled.
```

I suggest to highlight `null` everywhere in the text.

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +211,216 @@ public class ExampleSchema : Schema
...
     }
 
-    protected override bool Equals(ValueNode<Vector3> node)
-    {
-        return Value.Equals(node.Value);
+    return ThrowSerializationError(value);
+}
+```
+
+6. Override `ScalarGraphType.ToAST` if necessary.
```
sungam3r, line 243:
```suggestion
7. Override `ScalarGraphType.ToAST` if necessary.
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +211,216 @@ public class ExampleSchema : Schema
...
+    }
+}
+```
+
+And a sample of a response with a vector in a more structured format:
+
+```json
+{
+    "data": {
+        "getVector": {
```
sungam3r, line 291: I'm not sure that this is the right example (although he was here before). Scalar is to determine the leaf nodes of the request/response tree. Yes, custom scalar can have some kind of formatting, but they cannot add additional levels in the tree. These are no longer scalar, but objects.

src/GraphQL.Tests/Types/Vector3ScalarTests.cs
```diff
@@ -0,0 +1,190 @@
...
+        [Fact]
+        public void test_default()
+        {
+            AssertQuerySuccess(@"{ input }", @"{ ""input"": ""=7=8=9="" }");
+        }
+
+        [Fact]
+        public void test_output()
+        {
+            AssertQuerySuccess(@"{ output }", @"{ ""output"": { ""x"": 4, ""y"": 5, ""z"": 6 } }");
```
sungam3r, line 40: It works... thanks to the serializer. But I really doubt that the scalars were conceived to work like that.

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -8,9 +8,15 @@ You can extend your schema with your own custom scalars. Conceptually, a scalar
 
 - Literal Parsing: Transforms a scalar from its client-side representation as an argument to its server-side representation.
 
-Parsing for arguments and variables are handled separately because while arguments must always be expressed in GraphQL query syntax, variable format is transport-specific (usually JSON).
+Parsing for arguments and variables are handled separately because while arguments must always be expressed in GraphQL
+query syntax, variable format is transport-specific (usually JSON).
```
sungam3r, line 12:
```suggestion
query syntax, variable format is transport-specific (usually JSON). More information about these methods
you can find [here](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/Scalars/ScalarGraphType.cs).
```

docs2/site/docs/getting-started/custom-scalars.md
Shane32: No, the documentation was demonstrating how to have a scalar return a complex object.  I moved the demonstration further down the page, after a more basic demonstration.

docs2/site/docs/getting-started/custom-scalars.md
Shane32:
```suggestion
query syntax, variable format is transport-specific (usually JSON). You can find more information about
these methods [here](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/Scalars/ScalarGraphType.cs).
```

src/GraphQL/Types/Scalars/ScalarGraphType.cs
```diff
@@ -20,9 +20,10 @@ public abstract class ScalarGraphType : GraphType
         /// responsible for preparing the scalar for transport to the client. It is only responsible
         /// for generating an object which can eventually be serialized by some transport-focused API.
         /// <br/><br/>
-        /// This method must handle a value of <see langword="null"/>.
+        /// This method should handle a value of <see langword="null"/>, but may throw an exception
+        /// if <see langword="null"/> is an invalid internal scalar representation.
         /// </summary>
-        /// <param name="value">Resolved value (internal scalar representation). May be <see langword="null"/>.</param>
+        /// <param name="value">Resolved value (internal scalar representation). Might be <see langword="null"/> if returned from a field resolver.</param>
```
sungam3r, line 26: ?

docs2/site/docs/getting-started/custom-scalars.md
Shane32: The old 'custom scalars' page explicitly gave this sample, with a structured object being returned.

https://graphql-dotnet.github.io/docs/getting-started/custom-scalars

🤷‍♂️

I actually think it's a pretty neat sample, and demonstrates the power of custom scalars, including a need for `ToAST`.  But maybe we should add a note that the resulting schema is not spec-compliant.

Or perhaps we come up with another sample that can demonstrate a need for `ToAST`, and remove this one.

> These are no longer scalar, but object

True...  but whose fault is that?

Do we say that serialization and variables are outside the scope of graphql (as per spec), and therefor Vector3 is a scalar?  Along that logic, we look and see that ParseValue is explicitly looking for an object (dictionary) rather than a Vector3.  Why?  Because the json serialization engine cannot properly handle deserialization of Vector3.  Nor serialization for that matter (since it gets serialized as an object).  So we added a patch to ParseValue to patch a broken deserialization engine (e.g. `System.Text.Json`) that does not support deserializing Vector3 natively.

Or do we say that the graphql scalar should be written from the premise that a json serialization engine is used?  Then we can say that we should not be returning a Vector3 from Serialize because that type will be serialized as an object, which is "illegal" from graphql's POV.  (It's not a scalar anymore.)

We can even say that we "invented" a custom transport -- MyJSON -- where `{x:1,y:2,z:3}` is a Vector3 scalar.

I would generally write all of the code from the POV that a json serialization engine will be used.  Which is why I would remove native types (e.g. Guid or Uri) from being handled by ParseValue.  But in those cases at least it is a valid external representation of the value, and no harm will come from supporting it.

Anyway, back to the original question: I think we should provide a sample of a custom scalar with a reasonable premise and enough complexity so that the tooling can be demonstrated well.  It does not have to be a complex version of Vector3 of course.  But hopefully something that requires a `ToAST` implementation.

src/GraphQL/Types/Scalars/ScalarGraphType.cs
Shane32: I realized that it is the AST representation that is enforced by the graphql spec.  The spec does not control what internal values might be.  So when writing the sample above where null external values are internal (db) 0 values, I realized it was entirely valid for Serialize to throw an error upon `Serialize(null)`.

I also considered the reverse logic -- can a scalar refuse an AST with a value of NullValue?  Well maybe, actually, if that is never a valid value for the scalar.  But AST nodes must support null, as they are coerced to null when errors occur and so on.  And if they should not support null, it should be represented as a non-null graph type -- not a scalar that throws an exception when it's passed NullValue.  So I only changed Serialize.  Thoughts?

Shane32:
@sungam3r I believe you posted a code snippet for how to replace a built-in scalar type, but I can't remember where.  Can you add the necessary code to complete the samples in this documentation?  I feel that it is important to document this for v4, as we are instituting a lot of restrictions and people may want to be able to revert to the old behavior -- and this would allow them that capacity.  We should also add a test to ensure that our patch is valid.  And/or provide a more convenient way to replace a built-in type.

sungam3r:
I will continue review tomorrow, today I was engaged in other work.

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: > But maybe we should add a note that the resulting schema is not spec-compliant.

Probably. See http://spec.graphql.org/June2018/#sec-Scalars:
изображение


> True... but whose fault is that?

I would say ours, because we show people how to use scalar incorrectly.

> Or do we say that the graphql scalar should be written from the premise that a json serialization engine is used?

GraphQL response is **always** JSON.

> Which is why I would remove native types (e.g. Guid or Uri) from being handled by ParseValue. But in those cases at least it is a valid external representation of the value, and no harm will come from supporting it.

We should not delete them.

> But hopefully something that requires a ToAST implementation.

I already thought about JSON scalar. It will take some kind of library to implement it, so it cannot be placed in the main project. We can place that scalar in our projects for serialization. But for now just try to make an example.

sungam3r: https://github.com/graphql/graphql-spec/issues/826

src/GraphQL/Types/Scalars/ScalarGraphType.cs
sungam3r: I had more questions when I saw this change. `value` maybe `null`? Yes. This is a completely understandable description without any additional `if...`.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: > GraphQL response is always JSON.

And variables is not?

Shane32: Well, we can simply delete the second half of the sample.

Shane32: > GraphQL response is always JSON.

Not according to the specification.

http://spec.graphql.org/June2018/#sec-JSON-Serialization

> JSON is the most common serialization format for GraphQL. Though as mentioned above, GraphQL does not require a specific serialization format.

Shane32: The GraphQL spec defines the GraphQL query language, but does not make any requirements on the format of variables or result.  JSON is the most common format for variables and results to be formatted as.  I suggest limiting us to only JSON.  Then Guid and Uri and Enum are types that are not supported by JSON and hence have no need to be included within ParseValue.  And returning an object is invalid for a scalar to do.

But if we are designing the library where JSON is not a required input/output format -- for instance, perhaps the binary serializer within `System.Runtime.Serialization` was used -- then perhaps Guid and Uri and Enum are types that should be supported by ParseValue.  But also Vector3 is a completely valid scalar value, just as much as Guid or Uri.  Just like Uri, it can be serialized to a string, or can be split into parts (e.g. Scheme, Host, Port, Path, Query, etc).  There is no requirement that it MUST be serialized to a string.  It is simply a scalar that can be serialized by the binary transport in its native format.

So we can design this from a POV that only JSON transports are used, or not.

sungam3r: Good news in https://github.com/graphql/graphql-spec/issues/826 It seems that it is quite possible to serialize the scalar to objects:

> the value returned by custom scalar for the result coercion (in GraphQL.js this function is called serialize) can by anything that the Programming language/environment allows. For example it can be a Money object.

> GraphQL in general doesn't not limit at all this aspect: it is an implementation detail of the GraphQL implementation.

> To phrase it a bit different: The result of an GraphQL execution is tree, where Scalars or Enums are the leaves from a type perspective. But what the actual values of these types are at runtime can't be predicted or described by the spec.

I suggest you copy these phrases as is into documentation (with some changes). You can do this at the very beginning of a more complex second example to explain the motivation.

Please also add somewhere in text links to https://www.graphql-scalars.com/scalars-guide and https://www.graphql.de/blog/scalars-in-depth/ .

src/GraphQL.Tests/Types/Vector3ScalarTests.cs
```diff
@@ -0,0 +1,190 @@
...
+                if (value == null)
+                    return null;
+
+                if (value is Vector3 vector3)
+                    return vector3;
+
+                return ThrowSerializationError(value);
+            }
+
+            // this code is tested when the schema is built; tests fail without it
```
sungam3r, line 175:
```suggestion
```

?

src/GraphQL/Types/Scalars/ScalarGraphType.cs
sungam3r, line 26:
```suggestion
        /// <param name="value">Resolved value (internal scalar representation). May be <see langword="null"/>.</param>
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -62,6 +62,20 @@ For untyped `InputObjectGraphType` classes, like shown above, the default behavi
 will be to return the dictionary. `GetArgument<T>` will still attempt to convert a dictionary to the
 requested type via `ObjectExtensions.ToObject` as it did before.
 
+### Scalar null value handling
+
+Custom scalars can now handle serialization or deserialization of `null` values. This can be useful if
+you have a need to coerce certain internal values to null, such as serializing empty strings to null.
```
sungam3r, line 68:
```suggestion
you have a need to coerce certain internal values to `null`, such as serializing empty strings to `null`.
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -62,6 +62,20 @@ For untyped `InputObjectGraphType` classes, like shown above, the default behavi
...
+It can also be used to control deserialization of external null values, such as deserializing null to
```
sungam3r, line 69:
```suggestion
It can also be used to control deserialization of external `null` values, such as deserializing `null` to
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -183,6 +197,83 @@ This may be needed to get the parameters of parent nodes.
 
 ## Breaking Changes
 
+### Scalar Deserialization Type Enforcement
+
+Scalars do not coerce values if passed an incompatible type during deserialization from a variable. Previously, values would pass
+through the `ValueConverter` while being deserialized. Now the `ValueConverter` is ignored for deserialization of built-in fields.
```
sungam3r, line 203:
```suggestion
through the `ValueConverter` while being deserialized. Now the `ValueConverter` is ignored for deserialization of built-in scalars.
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -183,6 +197,83 @@ This may be needed to get the parameters of parent nodes.
...
+
+`IAstNodeConverter` has been completely removed along with the properties relating to it on the `Schema`. Code that had been configured
+for a custom scalar may need to be moved into the new `ToAST` virtual member of the custom scalar. `ValueNode` implementations are not
+recommended; the `ToAST` member should return one of the base value node types present in the library, such as `StringValue` or `IntValue`.
+If the `Serialize` method returns a basic .NET type (such as `int` or `string`), the default `ToAST` implementation should suffice.
+
+Code within `ValueConverter` registrations should be moved directly into the `ParseLiteral` and/or `ParseValue` methods. Having a
+`ValueConverter` registration should no longer be necessary for custom scalars.
+
+`ParseLiteral` and `ParseValue` must handle null values (`NullValue` for `ParseLiteral` and `null` for `ParseValue`). Typically
```
sungam3r, line 258:
```suggestion
`ParseLiteral` and `ParseValue` must handle `null` values (`NullValue` for `ParseLiteral` and `null` for `ParseValue`). Typically
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -183,6 +197,83 @@ This may be needed to get the parameters of parent nodes.
...
+`Serialize`'s default behavior still calls `ParseValue`. With the other changes, it should be verified if this is still valid
+for the custom scalar.
+
+If `ToAST` is overridden, it must process a value of `null` (typically by returning a new instance of `NullValue`) and throw an
+exception if there are serialization errors. The `ThrowASTConversionError` convenience method is provided for this purpose but
+any exception is valid to throw.
+
+You may wish to add implementations for the new `CanParseValue`, `CanParseLiteral` and `IsValidDefault` methods. This is not necessary
+as the default implementations will call `ParseValue`, `ParseLiteral` and `ToAST` respectively, returning `true` unless the method
+throws an exception. Adding a custom implementation of `CanParseLiteral` can improve performance if `ParseLiteral` causes boxing.
```
sungam3r, line 274:
```suggestion
throws an exception. Adding a custom implementation of `CanParseLiteral` can improve performance if `ParseLiteral` causes memory allocation / boxing.
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -183,6 +197,83 @@ This may be needed to get the parameters of parent nodes.
...
+`CanParseValue` is not used by the framework at this time, and `ToAST` is only called during schema initialization.
```
sungam3r, line 275:
```suggestion
`CanParseValue` is not used by the framework at this time, and `ToAST` is only called during schema initialization and schema printing.
```

sungam3r:
> I believe you posted a code snippet for how to replace a built-in scalar type, but I can't remember where. 

Not certainly in that way. Built-in scalars can not be replaced now. I showed an example when you can map `int` CLR type to  some user defined `MyScalar` when using expressions - _Ability to map CLR types to GraphTypes_ section in `migration4.md`. So it is actually a usage of the new type, not replacing existing scalar.

docs2/site/docs/migrations/migration4.md
```diff
@@ -183,6 +197,83 @@ This may be needed to get the parameters of parent nodes.
...
+
+- `StringGraphType` does not serialize integers to strings
+- Numeric scalars such as `IntGraphType` do not deserialize strings to their numeric type
+- Integer numeric scalars such as `IntGraphType` do not coerce floating-point values to their numeric type
+- `BooleanGraphType` does not deserialize/serialize strings or integers (e.g. "true", "false", 0, 1, etc)
+- `EnumGraphType` is stricter, requiring internal values for serialization, and external values for deserialization
+- `IdGraphType` (which allows any basic type) does not coerce variable values to trimmed strings during deserialization
+- `IdGraphType` does not trim serialized values (but does convert them to strings)
+
+For situations where it is necessary to revert scalars to previous behavior, you can override the built-in scalar by following the
```
sungam3r, line 217: Is it possible?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+```
+
+7. Override `ScalarGraphType.ToAST` if necessary.
+
+The infrastructure converts default field values to AST representations during initialization
+in order to verify that the default values are valid within an AST tree. The default implementation
+calls `Serialize` to convert the value to its client-side equivalent, then embeds it into an
+appropriate AST node based on the data type returned from `Serialize`. It is likely you will
+only need to override this method if you are creating a custom scalar that returns enumeration
+values, or if you are returning structured data.
```
sungam3r, line 252: > returning _structured_ data

Somewhere here may be note about the nature of scalars - scalars are not required to appear as primitive types i.e. GraphQL "supports" composite scalars on _serialization level_. See https://github.com/graphql-dotnet/graphql-dotnet/pull/2357#discussion_r589441410

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+
+In this example, you created a custom scalar. In summary:
+
+- Create a class for the server-side representation of the scalar
```
sungam3r, line 256: What class ? ScalarGraphType or data-class ?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+- Implement a `ScalarGraphType` which handles variable parsing, literal parsing, and serialization
+- Register the `ScalarGraphType` with the DI container
```
sungam3r, line 258:
```suggestion
- Register the `ScalarGraphType` within the DI container
```
?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+You can also choose to override `CanParseLiteral`, `CanParseValue` or `IsValidDefault` for
+enhanced performance. The default implementations call `ParseLiteral`, `ParseValue` and
+`ToAST` respectively, returning `false` if an exception is caught, or `true` otherwise.
+If you do choose to implement these methods, note that those methods must not throw an
+exception, and that they are not always called when executing a document.
+
+## Vector3 sample with combined string/structured parsing and serialization
+
+Now to extend our sample, let's assume that we want the Vector3 scalar to instead accept and
+return data in a more structured format, in addition to support the string format for variables.
```
sungam3r, line 269: > Somewhere here may be note about the nature of scalars

Or here.

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
 {
-    // new test
-    if (value is Vector3Value vector3Value)
-        return ParseValue(vector3Value.Value);
+    if (value == null)
+        return new NullValue();
+
+    if (value is Vector3 vector3)
+    {
+        return new StringValue($"{vector3.X},{vector3.Y},{vector3.Z}");
```
sungam3r, line 380: AST should be a string or a structure (object)?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+    }
+}
 ```
 
-This is necessary since the query executor converts all arguments and variables to `IValue` instances before coercing them to their server-side representation using `ParseLiteral`. `Vector3` instances parsed from variables will be converted to the more specific `Vector3Value` type.
+However, your code may use integer identifiers in the data access layer. So when you call `context.GetArgument<int>("id")`,
+GraphQL.NET calls the value converter to convert the string to an integer.
 
-7. Implement `ScalarGraphType.Serialize` so `Vector3` instances can be sent to the client.
+The value converter can be extended globally by calling the static method `Register` as follows:
```
sungam3r, line 409: This API is one to think about changing from static to instance method (v5).

docs2/site/docs/migrations/migration4.md
Shane32: #2388

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Yeah I didn't write this...

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+
+    ...
 }
 ```
 
-This implementation may surprise you. Why is `Serialize`, which is used for output, implemented identically to `ParseValue`, which is used for input? Why does `Serialize` return an object, rather than a string or byte array? It helps to understand a few internals of the library.
+## Null values
 
-- `Serialize` will be called during query execution, and should be passed an instance of `Vector3` from a field resolver.
+Custom scalars process and handle null values during serialization and deserialization. This allows
```
sungam3r, line 432: > This allows
for custom scalars that can assist when you have database values such as 0 that should represent null
when exposed outside the schema.

It is difficult for me to understand. There is no mistake here?

Shane32:
I'll review your comments and update the PR later, etc.  Can't tonight.

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+    public DbIdGraphType()
+    {
+        Name = "DbId";
+    }
 
-- `ValueConverter.ConvertTo` handles the case when `value` is an instance of the requested type by returning `value`. Therefore, `ValueConverter.ConvertTo` neatly handles both input and output representations of the scalar.
+    public override object ParseLiteral(IValue value) => value switch
+    {
+        StringValue s => int.TryParse(s.Value, out int i) && i > 0 ? i : throw new FormatException($"'{s.Value}' is not a valid identifier."),
+        null => 0,
```
sungam3r, line 448: `null` or `NullValue` or both ?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+
+```csharp
+public class MySchema : Schema
+{
+    ...
+
+    private readonly MyNodeWalker _myNodeWalker = new MyNodeWalker();
+    protected override void Validate()
+    {
+        _myNodeWalker.Run(this);
```
sungam3r, line 509: Before or after `base.Validate`? 🤔

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
-- Implement an `IAstFromValueConverter` for wrapping the server-side representation in its `ValueNode<T>` implementation
-- Register the `IAstFromValueConverter` with the schema using `Schema.RegisterValueConverter`
+## Replacing built-in scalar types
+
+In some cases you may want or need to replace the functionality of the built-in graph types. This can
+be accomplished by running a schema node visitor on the schema after it has been initialized. Keep in
+mind that the node visitor will walk through the introspection portion of the schema, so replacing
+a built-in type may affect the operation of introspection queries.
+
+Below is a sample of how to replace the built-in `IdGraphType` with the `DbIdGraphType` shown above.
```
sungam3r, line 474: The thought is correct, but I'm not sure yet it will work.

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+    }
+
+    ...
+}
+```
+
+2. Create a node walker
+
+```csharp
+// sungam3r to provide sample
```
sungam3r, line 493: https://github.com/graphql-dotnet/graphql-dotnet/issues/1952#issuecomment-786177240  Found!

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +213,303 @@ public class ExampleSchema : Schema
...
+// perhaps we should provide a node walker for this purpose, or provide a more convenient method to replace types
+// note: don't we need to fix up schematypes also so that if a type name is provided to a query variable argument, it gets a proper reference?
```
sungam3r, line 495: This is an interesting question. Perhaps not because _replacing_  involves replacement of behavior but not name. Name should stay the same.

Shane32: We are testing for this scenario with the solution in #2388 as it tests variables, literals, and serialization, for both schema-first and code-first.

sungam3r: So the names match. Right?

docs2/site/docs/getting-started/custom-scalars.md
Shane32:
```suggestion
- Create a data class for the server-side representation of the scalar
```

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Must be a string, as it needs to be able to be parsed by ParseLiteral.  From what I understand of the graphql spec, and based on the comments in the spec repo, scalar representations within graphql language cannot be structured.  Only the representations as variables or as serialized may be structured.

In my mind, `ParseLiteral` and `ToAST` are inverse functions of each other, and `ParseValue` and `Serialize` are inverse functions of each other.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Or delete it altogether...

docs2/site/docs/getting-started/custom-scalars.md
Shane32: I have a database where that is the case.  Another scenario I run into often is null vs empty strings.  Specifically, some databases hold null where they should represent empty strings, or some databases that hold empty strings where they should represent null.

If you have a suggestion of how I can explain this better, or a better sample to show altogether, I'm listening.

docs2/site/docs/getting-started/custom-scalars.md
Shane32:
```suggestion
        NullValue _ => 0,
```
You're right; `NullValue` is proper.  `value` cannot be `null` here.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Removed.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: I added a test to verify that it works.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Right.

Shane32: Sample replaced, so resolving this not-applicable comment.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Sorry, I added a test to show that the `DbIdGraphType` works with nullability coercion.

Shane32: This sample was replaced with a more common scenario -- replacing the boolean graph type (which is more strict in 4.0) to a version that allows coercion from any type of value (similar to how it was supported in 3.x).

docs2/site/docs/migrations/migration4.md
```diff
@@ -184,6 +198,83 @@ This may be needed to get the parameters of parent nodes.
...
+        NullValue _ => null,
+        _ => ThrowLiteralConversionException(value)
+    };
+
+    public override object ParseValue(object value) => value switch
+    {
+        null => null,
+        _ => ValueConverter.ConvertTo(value, typeof(bool)) ?? ThrowValueConversionException(value)
+    };
+}
```
Shane32, line 243:
```suggestion

    public virtual bool CanParseValue(object value)
    {
        try
        {
            _ = ParseValue(value);
            return true;
        }
        catch
        {
            return false;
        }
    }
}
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -184,6 +198,96 @@ This may be needed to get the parameters of parent nodes.
...
+- `IdGraphType` (which allows any basic type) does not coerce variable values to trimmed strings during deserialization
+- `IdGraphType` does not trim serialized values (but does convert them to strings)
+
+For situations where it is necessary to revert scalars to previous behavior, you can override the built-in scalar by following the
+instructions within the [Custom Scalars](https://graphql-dotnet.github.io/docs/getting-started/custom-scalars) documentation page.
+
+Below is a sample replacement for the `BooleanGraphType` which will restore the previous behavior.
+
+```csharp
+public class MyBooleanGraphType : ScalarGraphType
```
Shane32, line 224:
```suggestion
public class MyBooleanGraphType : BooleanGraphType
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -184,6 +198,96 @@ This may be needed to get the parameters of parent nodes.
...
+- Integer numeric scalars such as `IntGraphType` do not coerce floating-point values to their numeric type
+- `BooleanGraphType` does not deserialize/serialize strings or integers (e.g. "true", "false", 0, 1, etc)
+- `EnumGraphType` is stricter, requiring internal values for serialization, and external values for deserialization
+- `IdGraphType` (which allows any basic type) does not coerce variable values to trimmed strings during deserialization
+- `IdGraphType` does not trim serialized values (but does convert them to strings)
+
+For situations where it is necessary to revert scalars to previous behavior, you can override the built-in scalar by following the
+instructions within the [Custom Scalars](https://graphql-dotnet.github.io/docs/getting-started/custom-scalars) documentation page.
+
+Below is a sample replacement for the `BooleanGraphType` which will restore the previous behavior.
```
Shane32, line 221:
```suggestion
Below is a sample replacement for the `BooleanGraphType` which will restore the previous behavior exactly
as it was in version 3.x.
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -184,6 +198,97 @@ This may be needed to get the parameters of parent nodes.
...
+        {
+            _ = ParseValue(value);
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+}
```
Shane32, line 257:
```suggestion

    public virtual object Serialize(object value) => ParseValue(value);
}
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -184,6 +198,97 @@ This may be needed to get the parameters of parent nodes.
...
+        Name = "Boolean";
+    }
+
+    public override object ParseLiteral(IValue value) => value switch
+    {
+        BooleanValue b => b.Value,
+        NullValue _ => null,
+        _ => ThrowLiteralConversionException(value)
+    };
+
```
Shane32, line 238:
```suggestion
```

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: I think technically there is nothing preventing us to parse `ObjectValue` as literal. What is the difference at least for us parsing input object or scalar from this structured data? Both will work. Right?

sungam3r: I think technically there is nothing preventing us to parse `ObjectValue` as literal. What is the difference at least for us parsing input object or scalar from this structured data? Both will work. Right?

sungam3r: Try one more test supplying custom scalar as structured (object) literal.

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: Resolvefieldcontext.GetArgument still should be rather agile.

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: I'm talking about a long sentence to read and understand to me as not a native speaker. Maybe split to shorter ones?

sungam3r: Two that, two when.

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: Ok

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: Done?

docs2/site/docs/migrations/migration4.md
```diff
@@ -184,6 +198,92 @@ This may be needed to get the parameters of parent nodes.
...
+            _ = ParseValue(value);
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+
+    public virtual object Serialize(object value) => ParseValue(value);
```
sungam3r, line 251:
```suggestion
```

from base class

docs2/site/docs/migrations/migration4.md
```diff
@@ -184,6 +198,92 @@ This may be needed to get the parameters of parent nodes.
...
+        Name = "Boolean";
+    }
+
+    public override object ParseValue(object value) => value switch
+    {
+        null => null,
+        _ => ValueConverter.ConvertTo(value, typeof(bool)) ?? ThrowValueConversionException(value)
+    };
+
+    public virtual bool CanParseValue(object value)
```
sungam3r, line 238:
```suggestion
    public override bool CanParseValue(object value)
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -48,21 +60,17 @@ mutation AddVector($vector3: Vector3!) {
...
-            "Y":"43",
-            "Z":"66"
-        }
+        "getVector": "23, 43, 66"
     }
 }
 ```
 
-1. Create the class for the server-side representation.
+### 1. Create the class for the server-side representation.
```
sungam3r, line 73:
```suggestion
### 1. Create the class/struct for the server-side representation.
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +132,215 @@ public void ConfigureServices(IServiceCollection services)
...
+calls `Serialize` to convert the value to its client-side equivalent, then embeds it into an
+appropriate AST node based on the data type returned from `Serialize`. It is likely you will
+only need to override this method if you are creating a custom scalar that returns enumeration
+values, or if you are returning structured data.
 
-public class ExampleSchema : Schema
+### 8. Code-first: Register the .NET type mapping with the schema (optional).
+
+In order to allow simple field mappings like `Field(x => x.Vector)`, you need to register the
+custom scalar with the schema so that the proper graph type can be selected for the field.
```
sungam3r, line 262:
```suggestion
custom scalar within the schema so that the proper graph type can be selected for the field.
```
?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +132,215 @@ public void ConfigureServices(IServiceCollection services)
...
+Call `RegisterTypeMapping` with the CLR data model type and the graph type you want it mapped to.
```
sungam3r, line 263:
```suggestion
Call `RegisterTypeMapping` with the CLR data model type and the graph type you want it map to.
```

?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +132,215 @@ public void ConfigureServices(IServiceCollection services)
...
+    {
+        ...
+
+        this.RegisterTypeMapping<Vector3, Vector3Type>();
+    }
+}
+```
+
+This is not necessary if you use the alternate `Field<T>` syntax which specifies the graph type
+to be used for the field.
```
sungam3r, line 278:
```suggestion
to be used for the field or if your scalar data type is marked with `GraphQLMetadata` attribute
setting `InputType` and/or `OutputType` properies.
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +132,215 @@ public void ConfigureServices(IServiceCollection services)
...
+
+In this example, you created a custom scalar. In summary:
+
+- Create a data class for the server-side representation of the scalar
```
sungam3r, line 282:
```suggestion
- Create (if necessary) a data class/struct for the server-side representation of the scalar
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +132,215 @@ public void ConfigureServices(IServiceCollection services)
...
+`ToAST` respectively, returning `false` if an exception is caught, or `true` otherwise.
+If you do choose to implement these methods, note that those methods must not throw an
+exception, and that they are not always called when executing a document.
 
-        // Other schema assignments...
+## Vector3 sample with combined string/structured parsing and serialization
+
+Keep in mind that the serialized value returned by custom scalar can by anything that the
+environment allows. For example it can be a structured object, rather than a simple value.
+This is also true of variables, however literals can only be the represented by the native
```
sungam3r, line 296:
```suggestion
This is also true for variables, however literals can only be the represented by the native
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +132,215 @@ public void ConfigureServices(IServiceCollection services)
...
+scalar types supported by the GraphQL query language -- integers, floating-point values,
+booleans and null values. You cannot parse a structured literal object into a scalar value.
```
sungam3r, line 298: > You cannot parse a structured literal object into a scalar value.

That's a question. I think it's possible and in general does not violate logic.

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +132,215 @@ public void ConfigureServices(IServiceCollection services)
...
+        "getVector": {
+            "X":"23",
+            "Y":"43",
+            "Z":"66"
+        }
+    }
+}
+```
+
+### 1. Change `ParseValue` to accept strings or structured data:
```
sungam3r, line 334: The same for `ParseLiteral` 🤔

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +132,215 @@ public void ConfigureServices(IServiceCollection services)
...
+```
 
-    private Vector3 ParseVector3(string vector3InputString)
+And a sample of a response with a vector in a more structured format:
+
+```json
+{
+    "data": {
+        "getVector": {
+            "X":"23",
```
sungam3r, line 326: By the way... should we change "X" to "x" in all requests/responses?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+public override IValue ToAST(object value)
 {
-    public Vector3Value(Vector3 value)
+    if (value == null)
+        return new NullValue();
+
+    if (value is Vector3 vector3)
     {
-        Value = value;
+        return new StringValue($"{vector3.X},{vector3.Y},{vector3.Z}");
```
sungam3r, line 412:
```suggestion
        return new ObjectValue(
            new ObjectField("x", new FloatValue(vector3.X)),
            new ObjectField("y", new FloatValue(vector3.Y)),
            new ObjectField("z", new FloatValue(vector3.Z)),
        );
```

?

sungam3r: And if so please add new constuctor in `ObjectValue` with `params` keyword.

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
 
-    protected override bool Equals(ValueNode<Vector3> node)
-    {
-        return Value.Equals(node.Value);
+    return ThrowASTConversionError(value);
+}
+```
+
+With these changes, a variable can be parsed as a string or as a structured object, and is always returned
+as a structured object. Again, note that as an input literal, it must be parsed as a string.
```
sungam3r, line 420: > Again, note that as an input literal, it must be parsed as a string.

I doubt about that (_must_).

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+
+## ValueConverter
+
+When `GetArgument<T>` is called, the argument value is coerced to the requested type via the `ValueConverter`.
+No conversion takes place when the requested type matches the type of the object or scalar (the type returned from
+`ParseLiteral` or `ParseValue`). But you can also use the value converter to assist with input deserialization.
+
+For instance, you may be using `IdGraphType` within your schema as unique identifiers for your data objects. Pursuant
+to the GraphQL specification, these identifiers should be passed as strings such as in the below example:
```
sungam3r, line 429: Should they?
изображение

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+GraphQL.NET calls the value converter to convert the string to an integer.
+
+The value converter can be extended globally by calling the static method `Register` as follows:
 
 ```csharp
-public class Vector3AstValueConverter : IAstFromValueConverter
+ValueConverter.Register<Vector3, string>(v => $"{v.X},{v.Y},{v.Z}");
+```
+
+The above method registers a conversion from the `Vector3` format to a `string`. Since the registration is static,
```
sungam3r, line 448:
```suggestion
The above method registers a conversion from the `Vector3` struct to a `string`. Since the registration is static,
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+{
+    public DbIdGraphType()
     {
-        return value is Vector3;
+        Name = "DbId";
     }
+
+    public override object ParseLiteral(IValue value) => value switch
+    {
+        StringValue s => int.TryParse(s.Value, out int i) && i > 0 ? i : throw new FormatException($"'{s.Value}' is not a valid identifier."),
```
sungam3r, line 480: So passing 0 in inputs is forbidden?

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+1. inherit from the scalar graph type it is replacing; and
+2. have the `Name` property set to the name of the built-in graph type.
+
+You may then override any of the members to provide custom implementations. Note that most of the
+built-in scalars override `CanParseLiteral`, so it may be necessary to override that method if you
+override `ParseLiteral`. Check the source code for the built-in scalar type you are overriding for
+further reference.
+
+Below is a sample of how to replace the built-in `BooleanGraphType` so it will accept 0 and non-zero
+values to represent false and true.
```
sungam3r, line 517:
```suggestion
values to represent `false` and `true`.
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+        LongValue l => ParseValue(l.Value),
+        BigIntValue bi => ParseValue(bi.Value),
+        StringValue s => ParseValue(s.Value),
+        FloatValue f => ParseValue(f.Value),
+        DecimalValue d => ParseValue(d.Value),
+        NullValue _ => null,
+        _ => ThrowLiteralConversionError(value)
+    }
+
+    public virtual bool CanParseLiteral(IValue value)
```
sungam3r, line 555:
```suggestion
    public override bool CanParseLiteral(IValue value)
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+        BigInteger bi => bi != 0,
+        float f => f != 0,
+        double d => d != 0,
+        decimal d => d != 0,
+        string s => bool.Parse(s),
+        null => null,
+        _ => ThrowValueConversionError(value)
+    }
 
+    public virtual bool CanParseValue(object value)
```
sungam3r, line 588:
```suggestion
    public override bool CanParseValue(object value)
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+
+    public virtual object Serialize(object value) => ParseValue(value);
+
```
sungam3r, line 602:
```suggestion
```

from "base.base"

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+    public MyBooleanGraphType()
+    {
+        Name = "Boolean";
+    }
 }
-``` 
+```
 
-Update the implementation of `Vector3Type.ParseLiteral`:
+### 2. Override the methods as necessary; in this case we must override all of them except `IsValidDefault`.
```
sungam3r, line 532:
```suggestion
### 2. Override the methods as necessary; in this case we must override all of them except `Serialize` and `IsValidDefault`.
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+        _ => ThrowASTConversionError(value)
+    };
+}
 ```
 
-This is necessary since the query executor converts all arguments and variables to `IValue` instances before coercing them to their server-side representation using `ParseLiteral`. `Vector3` instances parsed from variables will be converted to the more specific `Vector3Value` type.
+### 3. Register the custom scalar within your schema.
 
-7. Implement `ScalarGraphType.Serialize` so `Vector3` instances can be sent to the client.
+The final step is to register an instance of the custom scalar to the schema. This can be
```
sungam3r, line 614:
```suggestion
The final step is to register an instance of the custom scalar within the schema. This can be
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
+    {
+        Query = ....;
+
+        RegisterType(new MyBooleanGraphType());
+    }
 }
 ```
 
-This implementation may surprise you. Why is `Serialize`, which is used for output, implemented identically to `ParseValue`, which is used for input? Why does `Serialize` return an object, rather than a string or byte array? It helps to understand a few internals of the library.
+For schema-first schemas, register it immediately after calling `Schema.For` to create the schema.
```
sungam3r, line 630:
```suggestion
For schema-first schemas, register it immediately after calling `Schema.For` to create the schema.
Immediately after calling `Schema.For` the schema is not yet initialized, therefore allows registration of types.
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
 
-- `Serialize` will be called during query execution, and should be passed an instance of `Vector3` from a field resolver.
+```csharp
+var schema = Schema.For(...);
+schema.RegisterType(new MyBooleanGraphType());
+```
 
-- `Serialize` is _also_ called when reading variables from the client so that variables can be converted to `IValue` instances. In the case of `Vector3Type`, `value` will be a string during this process.
+Now all `BooleanGraphType` references in your schema will utilize the new `MyBooleanGraphType`
+registered with the schema. This technique can be used to replace any of the built-in graph types.
```
sungam3r, line 638:
```suggestion
registered within the schema. This technique can be used to replace any of the built-in graph types.
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +355,310 @@ public class ExampleSchema : Schema
...
 
-- `ValueConverter.ConvertTo` handles the case when `value` is an instance of the requested type by returning `value`. Therefore, `ValueConverter.ConvertTo` neatly handles both input and output representations of the scalar.
+Note that if you set the `ResolvedType` property of a field or argument to an instance of a built-in
+type, or provide an instance of a built-in type to an applicable constructor, it will not be replaced
```
sungam3r, line 641: > it will **not be** replaced with your registered replacement built-in type
> This is **by design**.

😮 It was a surprise for me. Why is that?

sungam3r:
I made a full review again.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Have not tried it yet, and I think it would work, except pretty sure it fails validation (as it should). It would be interesting to try though.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: No. "..it mapped to" is correct.  "it to map to" would also be correct, but "it map to" is not.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Yes

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: Yes, interesting.

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: ok

docs2/site/docs/getting-started/custom-scalars.md
Shane32: It works!!!  Tests added.

sungam3r: 👍

docs2/site/docs/getting-started/custom-scalars.md
Shane32:
```suggestion
to the GraphQL specification, these identifiers may be passed as strings such as in the below example:
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -183,105 +385,321 @@ public class ExampleSchema : Schema
...
+    ...
+}
+```
+
+## Null values
+
+Custom scalars process and handle null values during serialization and deserialization. This allows
+for custom scalars that can assist when you have database values such as 0 that should represent null
+when exposed outside the schema. Below is an example of a scalar intended to represent a database
+autoincrementing numeric identifier internally, where null values are stored as 0.
```
Shane32, line 510:
```suggestion
Custom scalars process and handle null values during serialization and deserialization. For instance,
let's say you have a database where database ID values of 0 should be represented as graphql 'null'
values.  You could of course write code within the field resolver to translate the values, or you may
utilize a custom scalar to translate the values for you.  See below:
```

Shane32: Is this better?  I'm not sure.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: That's what is demonstrated, yes.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: Setting the `ResolvedType` is, by design, supposed to use the exact instance of the graph type that was passed to it.  This is necessary when writing code like this:

```csharp
var s = new Schema();
var query = new ObjectGraphType();
query.Name = "Query";
query.Field(...);
s.Query = query;
var mutation = new ObjectGraphType();
mutation.Name = "Mutation";
...
s.Mutation = mutation;
```

Now scalars typically don't have any properties on them, so they don't typically change as in the above example.

But the `RegisterType` is designed to, basically, register a type that can be used when setting all the `ResolvedType` instances throughout the schema.  If an object already has the `ResolvedType` property set, it should not be reset.

So that's why I say that it is by design.  Due to this design, we could change all the introspection types to be 'locked' onto the built-in scalar types by simply setting `ResolvedType` during their initialization, ensuring they always work the same regardless of the user's schema.  The same could be done by users within their own schema if there was a necessary place to do so.

If someone needs to change ALL scalars of a certain type after the schema is built, they can use `ReplaceScalar` afterwards as documented below.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: See suggestion here: https://github.com/graphql-dotnet/graphql-dotnet/pull/2357/files#r593837457

Modify or merge as desired.

docs2/site/docs/getting-started/custom-scalars.md
Shane32: I guess I don't know what I should do here.  We have documentation and an extensive sample regarding structured data a few paragraphs below.

docs2/site/docs/migrations/migration4.md
```diff
@@ -184,6 +198,90 @@ This may be needed to get the parameters of parent nodes.
...
+        {
+            _ = ParseValue(value);
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+}
```
Shane32, line 250:
```suggestion
}

// Code-first: add the following line of code to your schema's constructor
RegisterType(new MyBooleanGraphType());

// Schema-first: add the following line of code after your schema is built, before it is initialized
schema.RegisterType(new MyBooleanGraphType());
```

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +135,242 @@ public void ConfigureServices(IServiceCollection services)
...
+
+You can also choose to override `CanParseLiteral`, `CanParseValue` or `IsValidDefault` for
+enhanced performance. The default implementations call `ParseLiteral`, `ParseValue` and
+`ToAST` respectively, returning `false` if an exception is caught, or `true` otherwise.
+If you do choose to implement these methods, note that those methods must not throw an
+exception, and that they are not always called when executing a document.
+
+## Vector3 sample with combined string/structured parsing and serialization
+
+Keep in mind that the serialized value returned by custom scalar can by anything that the
```
sungam3r, line 298:
```suggestion
Keep in mind that the serialized value returned by custom scalar can be anything that the
```

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: better

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: I think I understand. So you are talking about `if ([field/arg].ResolvedType != null) { set_resolved_type }` blocks inside `SchemaTypes.cs`. Right?

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: done

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: ok

docs2/site/docs/getting-started/custom-scalars.md
```diff
@@ -124,51 +135,242 @@ public void ConfigureServices(IServiceCollection services)
...
+
+For any type that is not handled, or when the value cannot be parsed, you must throw an exception.
+`ThrowValueConversionError` is provided as a convenient method to facilitate throwing an exception
+when the type does not match.
+
+In the sample below, only `null` and `string` types need to be handled.
 
 ```csharp
 // In Vector3Type
 public override object ParseValue(object value)
```
sungam3r, line 204:
```suggestion
// In Vector3Type

public override object ParseValue(object value)
```

docs2/site/docs/getting-started/custom-scalars.md
sungam3r: Thus, the `ReplaceScalar` method takes its place because it is necessary :)

sungam3r (approved):
@Shane32 Some transient troubles with tests: https://github.com/graphql-dotnet/graphql-dotnet/runs/2105809570?check_suite_focus=true#step:10:487

I looked at the documentation again and I like it. Congratulations on the completion of great work.

PR #2360, opened 2021-03-06
Exclude projects from code coverage

sungam3r:


sungam3r:
Changes are simple, so I merge them.

PR #2361, opened 2021-03-06
SchemaPrinter prints default values of scalars + arguments uniqueness check in SchemaValidationVisitor

### LINKED ISSUE #2194: SchemaPrinter generates default values for mutations
Hello

We use SchemaPrinter in unit tests to validate schema integrity.

In queries and graphtypes we have a lot of mandatory scalar types which SchemaPrinter displays without default values like this:

> type Client {
>   Id: Long!
>   addresses: [Address]!
>   birthDay: Date!
> }

but for mutations it generates default values like this:

> parametersCorrection(applicationId: Guid! = 00000000-0000-0000-0000-000000000000, date: Date! = 01/01/0001 00:00:00, paymentDay: Int! = 0): Application!

We found this when unit test on local machine and build server generated different default date values due to different locales - _Date! = 01/01/0001_ and _Date! = 01.01.0001_
Unfortunately, we were not able to find any options to control this behavior and it looks like bug.

sungam3r:
fixes #2194 , also fixes `ComplicatedArgs.multipleOptAndReq` field in `ValidationSchema`.

@Shane32 Also see https://marketplace.visualstudio.com/items?itemName=RomanKvasov.GraphQL-Textmate&ssr=false#overview . It's rather usefull.
изображение

Before changes (my locale is `ru-RU`):
изображение


sungam3r:
What is interesting when writing the test, I copied the `data` argument, but the schema was still printed (initialized and validated). I realized that we do not have to test the uniqueness of the arguments. After that, I decided to see what was written in the specification, but I did not find such a requirement there. Perhaps I was poorly looking for.

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -313,12 +313,7 @@ public string FormatDefaultValue(object value, IGraphType graphType)
                 ListGraphType list => "[{0}]".ToFormat(string.Join(", ", ((IEnumerable<object>)value).Select(i => FormatDefaultValue(i, list.ResolvedType)))),
                 IInputObjectGraphType input => FormatInputObjectValue(value, input),
                 EnumerationGraphType enumeration => (enumeration.Serialize(value) ?? throw new ArgumentOutOfRangeException(nameof(value), $"Unknown value '{value}' for enumeration '{enumeration.Name}'")).ToString(),
-                ScalarGraphType _ => value switch
-                {
-                    string s => $"\"{s}\"",
-                    bool b => b ? "true" : "false",
-                    _ => value.ToString() // TODO: how to print custom scalars ("") ?
-                },
+                ScalarGraphType scalar => AstPrinter.Print(scalar.ToAST(value)),
```
Shane32, line 316: Isn't it nice to see how this old bug can be fixed with a single line of code, now with the new scalar design?

Shane32 (approved):

src/GraphQL/Utilities/SchemaPrinter.cs
sungam3r: 🥇

sungam3r: Did you try graphql textmate ?

PR #2362, opened 2021-03-06
Change double quotes to single quotes and remove new line in error messages

### LINKED ISSUE #2353: Change double quotes to single quotes in error messages
Example
изображение

### LINKED ISSUE #2356: Allow parsing enum variable by native .NET type
> There's just no reason for any external values to be parsed as internal values.

My point is that for variables we cannot say for sure what is `internal` and `external` values.

> There's no way for json -- the transport of choice for graphql -- to carry an .NET enumeration type to be parsed at all. 

Forget about JSON. This is just one of the sources. Think about the fact that the variables can simply be created programmatically. In the end, this is the essence of the runtime - the runtime/platform on which the application is running.

> So if you allowed ParseValue to contain the internal value 1, you should also allow the literal to parse the value 1. 

Absolutely no. Literal is a part of GraphQL language. Variables are transport/platform specific. Also keep in mind that there may be no any transport at all.

> I don't mind allowing native types within ParseValue for Guid/Uri/DateTime/DateTimeOffset because you can consider the values as external values -- **in another format**. Just like receiving a double for a decimal graph type -- no problem, coerce the type. But parsing an enum against an internal list of values is not the same thing.

Why that argument about _just another format_ cannot be applied to enums ?

I suggest to add a new issue pointing to that thread and go on with this PR.

_Originally posted by @sungam3r in https://github.com/graphql-dotnet/graphql-dotnet/pull/2341#discussion_r588407993_

sungam3r:
fixes #2353

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
@@ -36,7 +36,7 @@ namespace GraphQL.Tests.Bugs
             {
                 Errors = new ExecutionErrors
                 {
-                    new ValidationError(null, ArgumentsOfCorrectTypeError.NUMBER, "Argument \"in\" has invalid value 636474637870330463.\nExpected type \"Int\", found 636474637870330463.")
+                    new ValidationError(null, ArgumentsOfCorrectTypeError.NUMBER, "Argument 'in' has invalid value. Expected type 'Int', found 636474637870330463.")
                     {
                         Code = "ARGUMENTS_OF_CORRECT_TYPE"
                     }
@@ -92,7 +92,7 @@ namespace GraphQL.Tests.Bugs
             {
                 Errors = new ExecutionErrors
                 {
-                    new ValidationError(null, ArgumentsOfCorrectTypeError.NUMBER, "Argument \"in\" has invalid value 636474637870330463636474637870330463636474637870330463.\nExpected type \"Long\", found 636474637870330463636474637870330463636474637870330463.")
+                    new ValidationError(null, ArgumentsOfCorrectTypeError.NUMBER, "Argument 'in' has invalid value. Expected type 'Long', found 636474637870330463636474637870330463636474637870330463.")
                     {
                         Code = "ARGUMENTS_OF_CORRECT_TYPE"
                     }
--- a/src/GraphQL.Tests/Validation/ArgumentsOfCorrectTypeTests.cs
+++ b/src/GraphQL.Tests/Validation/ArgumentsOfCorrectTypeTests.cs
@@ -1,4 +1,3 @@
-using System.Collections.Generic;
 using GraphQL.Validation.Errors;
 using GraphQL.Validation.Rules;
 using Xunit;
@@ -543,11 +542,7 @@ namespace GraphQL.Tests.Validation
             ShouldFailRule(_ =>
             {
                 _.Query = query;
-                Rule.badValue(_, "stringListArg", "[String]", "[\"one\", 2]", 3, 36,
-                    new[]
-                    {
-                        "In element #1: Expected type \"String\", found 2."
-                    });
+                Rule.badValue(_, "stringListArg", "[String]", "[\"one\", 2]", 3, 36, "In element #2: [Expected type 'String', found 2.]");
             });
         }
 
@@ -741,7 +736,7 @@ namespace GraphQL.Tests.Validation
             ShouldFailRule(_ =>
             {
                 _.Query = query;
-                Rule.badValue(_, "req1", "Int", "null", 3, 30, new[] { "Expected \"Int!\", found null." });
+                Rule.badValue(_, "req1", "Int", "null", 3, 30, "Expected 'Int!', found null.");
             });
         }
 
@@ -758,7 +753,7 @@ namespace GraphQL.Tests.Validation
             ShouldFailRule(_ =>
             {
                 _.Query = query;
-                Rule.badValue(_, "req2", "Int", "null", 3, 30, new[] { "Expected \"Int!\", found null." });
+                Rule.badValue(_, "req2", "Int", "null", 3, 30, "Expected 'Int!', found null.");
             });
         }
 
@@ -775,8 +770,8 @@ namespace GraphQL.Tests.Validation
             ShouldFailRule(_ =>
             {
                 _.Query = query;
-                Rule.badValue(_, "req2", "Int", "null", 3, 30, new[] { "Expected \"Int!\", found null." });
-                Rule.badValue(_, "req1", "Int", "null", 3, 42, new[] { "Expected \"Int!\", found null." });
+                Rule.badValue(_, "req2", "Int", "null", 3, 30, "Expected 'Int!', found null.");
+                Rule.badValue(_, "req1", "Int", "null", 3, 42, "Expected 'Int!', found null.");
             });
         }
     }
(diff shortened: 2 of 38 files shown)


src/GraphQL.Tests/Bugs/Bug1205VeryLongInt.cs
```diff
@@ -36,7 +36,7 @@ public void Very_Long_Number_In_Input_Should_Return_Error_Without_Data_For_Int()
             {
                 Errors = new ExecutionErrors
                 {
-                    new ValidationError(null, ArgumentsOfCorrectTypeError.NUMBER, "Argument \"in\" has invalid value 636474637870330463.\nExpected type \"Int\", found 636474637870330463.")
+                    new ValidationError(null, ArgumentsOfCorrectTypeError.NUMBER, "Argument 'in' has invalid value. Expected type 'Int', found 636474637870330463.")
```
sungam3r, line 39: In new error messages value does not repeat twice.

src/GraphQL.Tests/Validation/ArgumentsOfCorrectTypeTests.cs
```diff
@@ -543,11 +542,7 @@ public void list_value_incorrect_item_type()
             ShouldFailRule(_ =>
             {
                 _.Query = query;
-                Rule.badValue(_, "stringListArg", "[String]", "[\"one\", 2]", 3, 36,
-                    new[]
-                    {
-                        "In element #1: Expected type \"String\", found 2."
-                    });
+                Rule.badValue(_, "stringListArg", "[String]", "[\"one\", 2]", 3, 36, "In element #2: [Expected type 'String', found 2.]");
```
sungam3r, line 545: Notice change in index. Was 1, became 2. This is due to bug in `IsValidLiteralValue` before:
```csharp
 return listValue.Values
                        .SelectMany(value =>
                            IsValidLiteralValue(ofType, value, schema)
                                .Select((err, index) => $"In element #{index + 1}: {err}")
                        )
                        .ToArray();
```

Index should be taken over "external" list, not over internal `string[]` array of errors.

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -283,12 +281,18 @@ public static string[] IsValidLiteralValue(this IGraphType type, IValue valueAst
 
                 if (valueAst is ListValue listValue)
                 {
-                    return listValue.Values
```
sungam3r, line 286: No more bug here. No more LINQ. No more closures and allocations.

src/GraphQL/Validation/Errors/ArgumentsOfCorrectTypeError.cs
```diff
@@ -14,16 +12,16 @@ public class ArgumentsOfCorrectTypeError : ValidationError
...
 
-        internal static string BadValueMessage(string argName, string value, IEnumerable<string> verboseErrors)
+        internal static string BadValueMessage(string argName, string verboseErrors)
         {
-            var message = verboseErrors != null ? $"\n{string.Join("\n", verboseErrors)}" : "";
-
-            return $"Argument \"{argName}\" has invalid value {value}.{message}";
+            return string.IsNullOrEmpty(verboseErrors)
+                ? $"Argument '{argName}' has invalid value."
+                : $"Argument '{argName}' has invalid value. {verboseErrors}";
```
Shane32, line 24: Simplify?  Either way is fine.
```suggestion
            return $"Argument '{argName}' has invalid value.{(string.IsNullOrEmpty(verboseErrors) ? null : " " + verboseErrors)}";
```

src/GraphQL/Validation/Errors/DefaultValuesOfCorrectTypeError.cs
```diff
@@ -15,15 +14,16 @@ public class DefaultValuesOfCorrectTypeError : ValidationError
...
         }
 
-        internal static string BadValueForDefaultArgMessage(string varName, string type, string value, IEnumerable<string> verboseErrors)
+        internal static string BadValueForDefaultArgMessage(string varName, string type, string value, string verboseErrors)
         {
-            var message = verboseErrors != null ? "\n" + string.Join("\n", verboseErrors) : "";
-            return $"Variable \"{varName}\" of type \"{type}\" has invalid default value {value}.{message}";
+            return string.IsNullOrEmpty(verboseErrors)
+                ? $"Variable '{varName}' of type '{type}' has invalid default value '{value}'."
+                : $"Variable '{varName}' of type '{type}' has invalid default value '{value}'. {verboseErrors}";
```
Shane32, line 26: Simplify?

src/GraphQL/Validation/Errors/NoUndefinedVariablesError.cs
```diff
@@ -19,7 +19,7 @@ public NoUndefinedVariablesError(ValidationContext context, Operation node, Vari
 
         internal static string UndefinedVarMessage(string varName, string opName)
             => !string.IsNullOrWhiteSpace(opName)
-                ? $"Variable \"${varName}\" is not defined by operation \"{opName}\"."
-                : $"Variable \"${varName}\" is not defined.";
+                ? $"Variable '${varName}' is not defined by operation '{opName}'."
+                : $"Variable '${varName}' is not defined.";
```
Shane32, line 23: Simplify?

Shane32: Maybe not here...

src/GraphQL/Validation/Errors/ArgumentsOfCorrectTypeError.cs
sungam3r: This option is more difficult to read and there are more allocations in it, although they are not on the hot path.

Shane32:
Looks great!  What about errors ending in periods?  I know we have some of those.

src/GraphQL/Validation/Errors/NoUndefinedVariablesError.cs
sungam3r: It is better to save readability, I inverted the condition.

Shane32:
Newlines still present at SchemaTypes.cs:487 and SchemaTypes.cs:712 - want to catch those also?

```cs
                throw new InvalidOperationException($@"A loop has been detected while registering schema types.
There was an attempt to re-register '{namedType.FullName}' with instance of '{resolvedType.GetType().FullName}'.
Make sure that your ServiceProvider is configured correctly.");
```
```cs
                    throw new InvalidOperationException($@"Unable to register GraphType '{type.GetType().FullName}' with the name '{typeName}'.
The name '{typeName}' is already registered to '{existingGraphType.GetType().FullName}'. Check your schema configuration.");
```

sungam3r:
Is it bad?

Shane32:
I just thought we were going to consolidate so all errors ended in a period?  Can be done later of course.

sungam3r:
> Newlines still present at SchemaTypes.cs:487 and SchemaTypes.cs:712 - want to catch those also?

No, those are not execution errors serialized into response.

Shane32:
Ok

sungam3r:
> I just thought we were going to consolidate so all errors ended in a period?

🤔 How do I find them all?

Shane32 (approved):

Shane32:
> > I just thought we were going to consolidate so all errors ended in a period?
> 
> 🤔 How do I find them all?

Yeah that's the hard part...

PR #2365, opened 2021-03-07
Add ISchema.AdditionalTypeInstances

Shane32:
Realized this property was missing when trying to enumerate the additionally registered graph type instances.

### DIFF
--- a/src/GraphQL/Types/ISchema.cs
+++ b/src/GraphQL/Types/ISchema.cs
@@ -77,12 +77,12 @@ namespace GraphQL.Types
         SchemaTypes AllTypes { get; }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/GraphQL/Types/ISchema.cs
```diff
@@ -81,6 +81,11 @@ public interface ISchema : IProvideMetadata, IProvideDescription
         /// </summary>
         IEnumerable<Type> AdditionalTypes { get; }
 
+        /// <summary>
+        /// A list of additional graph type instances manually added to the schema by RegisterType call.
```
sungam3r, line 85: Add a see tag on `RegisterType`.

sungam3r (approved):

PR #2366, opened 2021-03-07
Allow SchemaTypes to be overridable

Shane32:
Since `SchemaTypes` is not an interface (there is no `ISchemaTypes`), I would like to add the very minimum to the codebase to allow it to be replaced if desired.  I would like to experiment with doing so, but cannot as it is now because it is effectively a `sealed` class -- the constructor is private and even if it were not, the `Create` method is static.

In doing so, I took advantage of the fact that we are now passing `ISchema` to `Create`, and so there is no need to pass most of the rest of the arguments.  The code that executes is exactly the same; no changes were made.  It was simply reorganized a little.

The API changes to `SchemaTypes` are simple also:
- Add protected constructor that does nothing
- Make the `internal` members 'protected internal virtual' so they can be overridden:
  - Dictionary
  - SchemaMetaFieldType
  - TypeMetaFieldType
  - TypeNameMetaFieldType

The public members all access the dictionary, so I did not make them virtual at this time.  I would like to look at that, but it can be done after v4 has been released.  We also could/should make some members of `Schema` virtual.  However, it is an interface, so it can be replaced.

The only other change is that I did remove `CheckDisposed` because it was pointless.  There are no public members to allow adding types after initialization.  Retaining it would have required additional protected members.

This code requires PR #2365 in order to compile.

This PR is necessary by the fact that, unlike v3, `SchemaTypes` is referenced throughout the project.  There is no way with v4 to get around having to use `SchemaTypes`.  So this PR is designed to let `SchemaTypes` be replaced as needed.  Unfortunately since there are no protected methods yet, all of the code would need to be duplicated to the new class.  I hope to rectify that in a future non-breaking PR, but it will take time and I don't want to hold up v4.

@sungam3r If you feel that adding an interface is a better way to go, please let me know.

### DIFF
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -82,10 +82,31 @@ namespace GraphQL.Types
         private readonly TypeCollectionContext _context;
         private readonly INameConverter _nameConverter;
         private readonly object _lock = new object();
-        private bool _sealed;
 
-        private SchemaTypes(ISchema schema, List<(Type clrType, Type graphType)> typeMappings)
+        protected SchemaTypes()
         {
+        }
+
+        /// <summary>
+        /// Initializes a new instance for the specified schema, and with the specified type resolver.
+        /// </summary>
+        /// <param name="schema">A schema for which this instance is created.</param>
+        /// <param name="types">A list of graph type instances to register in the lookup table.</param>
+        /// <param name="graphTypeProvider">A service provider used to resolve graph types.</param>
+        public SchemaTypes(ISchema schema, IEnumerable<IGraphType> types, IServiceProvider graphTypeProvider)
+        {
+            if (schema == null)
+                throw new ArgumentNullException(nameof(schema));
+            if (types == null)
+                throw new ArgumentNullException(nameof(types));
+            if (graphTypeProvider == null)
+                throw new ArgumentNullException(nameof(graphTypeProvider));
+
+            var typeMappingsEnumerable = schema.TypeMappings ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.TypeMappings));
+            var typeMappings = typeMappingsEnumerable is List<(Type, Type)> typeMappingsList ? typeMappingsList : typeMappingsEnumerable.ToList();
+            var directives = schema.Directives ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.Directives));
+            Func<Type, IGraphType> resolveType = t => (IGraphType)graphTypeProvider.GetRequiredService(t);
+
             _introspectionTypes = CreateIntrospectionTypes(schema.Features.AppliedDirectives, schema.Features.RepeatableDirectives);
 
             _context = new TypeCollectionContext(
@@ -109,6 +130,51 @@ namespace GraphQL.Types
 
             // set the name converter properly
             _nameConverter = schema.NameConverter ?? CamelCaseNameConverter.Instance;
+
+            var ctx = new TypeCollectionContext(
+                t => _builtInScalars.TryGetValue(t, out var graphType) ? graphType : resolveType(t),
+                (name, graphType, context) =>
+                {
+                    if (this[name] == null)
+                    {
+                        AddType(graphType, context);
+                    }
+                },
+                typeMappings);
+
+            foreach (var type in types)
+            {
+                AddType(type, ctx);
+            }
+
+            // these fields must not have their field names translated by INameConverter; see HandleField
+            HandleField(null, SchemaMetaFieldType, ctx, false);
+            HandleField(null, TypeMetaFieldType, ctx, false);
+            HandleField(null, TypeNameMetaFieldType, ctx, false);
+
+            foreach (var directive in directives)
+            {
+                if (directive.Arguments?.Count > 0)
+                {
+                    foreach (var arg in directive.Arguments.List)
+                    {
+                        if (arg.ResolvedType != null)
+                        {
+                            AddTypeIfNotRegistered(arg.ResolvedType, ctx);
+                            arg.ResolvedType = ConvertTypeReference(directive, arg.ResolvedType);
+                        }
+                        else
+                        {
+                            AddTypeIfNotRegistered(arg.Type, ctx);
+                            arg.ResolvedType = BuildNamedType(arg.Type, ctx.ResolveType);
+                        }
+                    }
+                }
+            }
+
+            ApplyTypeReferences();
+
+            Debug.Assert(ctx.InFlightRegisteredTypes.Count == 0);
         }
 
         private static Dictionary<Type, IGraphType> CreateIntrospectionTypes(bool allowAppliedDirectives, bool allowRepeatable)
@@ -143,7 +209,7 @@ namespace GraphQL.Types
             .ToDictionary(t => t.GetType());
         }
 
-        internal Dictionary<string, IGraphType> Dictionary { get; } = new Dictionary<string, IGraphType>();
+        protected internal virtual Dictionary<string, IGraphType> Dictionary { get; } = new Dictionary<string, IGraphType>();
 
         /// <inheritdoc cref="IEnumerable.GetEnumerator"/>
         public IEnumerator<IGraphType> GetEnumerator() => Dictionary.Values.GetEnumerator();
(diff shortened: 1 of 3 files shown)


src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -109,8 +133,78 @@ private SchemaTypes(ISchema schema, List<(Type clrType, Type graphType)> typeMap
...
+                    }
+                }
+            }
+
+            ApplyTypeReferences();
+
+            Debug.Assert(ctx.InFlightRegisteredTypes.Count == 0);
         }
 
+        private static IEnumerable<IGraphType> GetSchemaTypes(ISchema schema, IServiceProvider graphTypeProvider)
```
Shane32, line 185: Moved from `Schema` with no practical changes.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -82,10 +82,34 @@ public class SchemaTypes : IEnumerable<IGraphType>
...
+            if (schema == null)
+                throw new ArgumentNullException(nameof(schema));
+            if (graphTypeProvider == null)
+                throw new ArgumentNullException(nameof(graphTypeProvider));
+
+            var types = GetSchemaTypes(schema, graphTypeProvider);
+            var typeMappingsEnumerable = schema.TypeMappings ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.TypeMappings));
+            var typeMappings = typeMappingsEnumerable is List<(Type, Type)> typeMappingsList ? typeMappingsList : typeMappingsEnumerable.ToList();
+            var directives = schema.Directives ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.Directives));
+            Func<Type, IGraphType> resolveType = t => (IGraphType)graphTypeProvider.GetRequiredService(t);
```
Shane32, line 109: Create old argument values from simplified constructor syntax, so rest of code does not need to change.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -155,80 +249,8 @@ private SchemaTypes(ISchema schema, List<(Type clrType, Type graphType)> typeMap
...
 
-        /// <summary>
-        /// Initializes a new instance for the specified graph types and directives, and with the specified type resolver and name converter.
-        /// </summary>
-        /// <param name="types">A list of graph type instances to register in the lookup table.</param>
-        /// <param name="typeMappings">CLR-GraphType type mappings.</param>
-        /// <param name="directives">A list of directives to register.</param>
-        /// <param name="resolveType">A delegate which returns an instance of a graph type from its .NET type.</param>
-        /// <param name="schema">A schema for which this instance is created.</param>
-        public static SchemaTypes Create(
```
Shane32, line 174: Moved all code within `Create` to the constructor, with no practical changes.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -292,10 +314,8 @@ public void ApplyMiddleware(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegat
                 }
                 return type;
             }
-            internal set
+            private set
```
Shane32, line 317: Was not used outside of this class

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -307,7 +327,7 @@ internal set
         /// Returns a graph type instance from the lookup table by its .NET type.
         /// </summary>
         /// <param name="type">The .NET type of the graph type.</param>
-        internal IGraphType this[Type type]
+        private IGraphType this[Type type]
```
Shane32, line 330: Not used outside this class

sungam3r:
> The only other change is that I did remove `CheckDisposed` because it was pointless.

CheckSealed ?

Shane32:
> > The only other change is that I did remove `CheckDisposed` because it was pointless.
> 
> CheckSealed ?

Right. CheckSealed.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -82,10 +82,34 @@ public class SchemaTypes : IEnumerable<IGraphType>
         private readonly TypeCollectionContext _context;
         private readonly INameConverter _nameConverter;
         private readonly object _lock = new object();
-        private bool _sealed;
 
-        private SchemaTypes(ISchema schema, List<(Type clrType, Type graphType)> typeMappings)
+        /// <summary>
+        /// Initializes a new instance with no types registered
```
sungam3r, line 87:
```suggestion
        /// Initializes a new instance with no types registered.
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -82,10 +82,34 @@ public class SchemaTypes : IEnumerable<IGraphType>
...
+                throw new ArgumentNullException(nameof(graphTypeProvider));
+
+            var types = GetSchemaTypes(schema, graphTypeProvider);
+            var typeMappingsEnumerable = schema.TypeMappings ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.TypeMappings));
+            var typeMappings = typeMappingsEnumerable is List<(Type, Type)> typeMappingsList ? typeMappingsList : typeMappingsEnumerable.ToList();
+            var directives = schema.Directives ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.Directives));
+            Func<Type, IGraphType> resolveType = t => (IGraphType)graphTypeProvider.GetRequiredService(t);
+
+            // old constructor -- creates introspection types
+
```
sungam3r, line 112:
```suggestion
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -109,8 +131,76 @@ private SchemaTypes(ISchema schema, List<(Type clrType, Type graphType)> typeMap
...
+            if (schema.Mutation != null)
+                yield return schema.Mutation;
+
+            if (schema.Subscription != null)
+                yield return schema.Subscription;
+
+            foreach (var type in schema.AdditionalTypes)
+                yield return (IGraphType)graphTypeProvider.GetRequiredService(type.GetNamedType());
+        }
+
```
Shane32, line 202:
```suggestion
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -82,10 +82,34 @@ public class SchemaTypes : IEnumerable<IGraphType>
...
+        protected SchemaTypes()
         {
+        }
+
+        /// <summary>
+        /// Initializes a new instance for the specified schema, and with the specified type resolver.
+        /// </summary>
+        /// <param name="schema">A schema for which this instance is created.</param>
+        /// <param name="graphTypeProvider">A service provider used to resolve graph types.</param>
+        public SchemaTypes(ISchema schema, IServiceProvider graphTypeProvider)
```
sungam3r, line 98:
```suggestion
        public SchemaTypes(ISchema schema, IServiceProvider serviceProvider)
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -82,10 +82,34 @@ public class SchemaTypes : IEnumerable<IGraphType>
...
+        /// </summary>
+        protected SchemaTypes()
         {
+        }
+
+        /// <summary>
+        /// Initializes a new instance for the specified schema, and with the specified type resolver.
+        /// </summary>
+        /// <param name="schema">A schema for which this instance is created.</param>
+        /// <param name="graphTypeProvider">A service provider used to resolve graph types.</param>
```
sungam3r, line 97:
```suggestion
        /// <param name="serviceProvider">A service provider used to resolve graph types.</param>
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -109,6 +131,73 @@ private SchemaTypes(ISchema schema, List<(Type clrType, Type graphType)> typeMap
...
+                    }
+                }
+            }
+
+            ApplyTypeReferences();
+
+            Debug.Assert(ctx.InFlightRegisteredTypes.Count == 0);
+        }
+
+        private static IEnumerable<IGraphType> GetSchemaTypes(ISchema schema, IServiceProvider graphTypeProvider)
```
sungam3r, line 181:
```suggestion
        private static IEnumerable<IGraphType> GetSchemaTypes(ISchema schema, IServiceProvider serviceProvider)
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -109,6 +131,73 @@ private SchemaTypes(ISchema schema, List<(Type clrType, Type graphType)> typeMap
...
+                yield return schema.Query;
+
+            if (schema.Mutation != null)
+                yield return schema.Mutation;
+
+            if (schema.Subscription != null)
+                yield return schema.Subscription;
+
+            foreach (var type in schema.AdditionalTypes)
+                yield return (IGraphType)graphTypeProvider.GetRequiredService(type.GetNamedType());
```
sungam3r, line 200:
```suggestion
                yield return (IGraphType)serviceProvider.GetRequiredService(type.GetNamedType());
```

sungam3r (approved):
```
   var types = GetSchemaTypes(schema, serviceProvider);
            var typeMappingsEnumerable = schema.TypeMappings ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.TypeMappings));
            var typeMappings = typeMappingsEnumerable is List<(Type, Type)> typeMappingsList ? typeMappingsList : typeMappingsEnumerable.ToList();
            var directives = schema.Directives ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.Directives));
            Func<Type, IGraphType> resolveType = t => (IGraphType)serviceProvider.GetRequiredService(t);
```

You can cut this block if it is made only to reduce the diff.

Shane32:
I'll review that block before merging.  Thanks.

PR #2367, opened 2021-03-07
Make type reference classes public

### LINKED ISSUE #2364: GraphQLTypeReference missing in 4.0
I've just been updating my consumer code to use GraphQL 4.0.0-preview-240 in development in order to further test it and anticipate any gotchas that might come up with migrating to a new major version. So far it all looks fine (though I haven't gotten it to compile yet), except one thing: `GraphQLTypeReference` has been made internal.

This bit of functionality was integral to my use case of GraphQL.Net, I found it very useful when building schemas with circular references (which is most schemas). I can elaborate more if desired. Could this internalisation please be reversed or equivalent replacement functionality released?

Shane32:
Fixes #2364 

These three classes seem necessary (or at the very least, useful) for people writing dynamic schema builder code.
- `GraphQLTypeReference`
- `GraphQLClrOutputTypeReference`
- `GraphQLClrInputTypeReference`

This PR makes them public.

### DIFF
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -502,7 +502,7 @@ default continues to be `ResolverType.Resolver`.
 * `CoreToVanillaConverter.Convert` method now requires only one `GraphQLDocument` argument.
 * `GraphTypesLookup` has been renamed to `SchemaTypes` with a significant decrease in public APIs 
 * `TypeCollectionContext` class is now internal, also all methods with this parameter in `GraphTypesLookup` (now `SchemaTypes`) are private.
-* `GraphQLTypeReference` class is now internal, also `GraphTypesLookup.ApplyTypeReferences` is now private.
+* `GraphTypesLookup.ApplyTypeReferences` is now private.
 * `IHaveDefaultValue.Type` has been moved to `IProvideResolvedType.Type`
 * `ErrorLocation` struct became `readonly`.
 * `DebugNodeVisitor` class has been removed.
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -436,6 +436,10 @@ schema.RegisterTypeMapping<int, MyIntGraphType>();
 schema.RegisterTypeMapping<string, MySpecialFormattedStringGraphType>();
 ```
 
+If you have dynamic code that relies on a call to `GraphTypeTypeRegistry.Get<T>` then you will need to instead utilize
+a graph type of `GraphQLClrOutputTypeReference<T>` or `GraphQLClrInputTypeReference<T>` where `T` is the CLR type.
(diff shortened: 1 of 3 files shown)


sungam3r:
I would not change access modifier to public.

Shane32:
Why not?  How can you write a dynamic schema builder without them?  Did you read #2364 ?  Seems pretty important.

sungam3r:
Because of it's library internals. First let's see what _dynamic schema builder_ are. It is possible that the task is solved in other ways. Ping @adamjones1

Shane32:
It is absolutely essential for any library that had utilized the `GraphTypeTypeRegistry` directly that `GraphQLClrOutputTypeReference` and `GraphQLClrInputTypeReference` be made public.

Here is a link to where it is used in @SimonCropp 's library:

https://github.com/SimonCropp/GraphQL.EntityFramework/blob/master/src/GraphQL.EntityFramework/GraphTypeFinder.cs

Here is a link to where it is used in my "DI" project, which I use in production applications:

https://github.com/Shane32/GraphQL.DI/blob/master/src/DIObjectGraphType.cs#L196..L204

I also use it in our company software that I cannot post here, but it is similar in nature to the two examples above.  Because the `GraphTypeTypeRegistry` is not `static` (and does not even exist anymore), the only way to create a graph type reference to a .NET type while the field is being constructed is to use these reference classes, letting them be resolved when the schema is built.  Basically, the need for these classes is identical to the need within the library!

Although I do not build schemas dynamically the way @adamjones1 described in #2364, the change is in the same nature and I feel that `GraphQLTypeReference` should be provided with the library.  Without it, I'm believe there's a workaround, but what I'm thinking of would be quite cumbersome.  I just do not see any reason to remove this useful class from v4.  Again, the need for these classes is the same as the need within the library -- to dynamically build a schema.  Just because someone does not choose to use the `SchemaBuilder` to build their schema does not mean that we should not provide the same toolset to be able to construct the schema with.

It is unfortunate that I did not think of this limitation when we implemented PR #2283.  But we must rectify it.  We will also undoubtedly have other users that need part of the library `public` which we made `internal` for v4.  We must be sure that the library is open and customizable enough that anyone can tailor it to their needs, reverting those changes or providing substitutions if necessary.  For example, the changes within `ExecutionStrategy` now allow complete customization through its `protected` members, which retaining only two `public` members.

sungam3r:
OK. Then let's add a new paragraph into `schema-types.md` like 
```
##References
```

and describe these 3 classes. These are special "service" types that are worth mentioning among all the others types.

Shane32:
Good idea; I can do some documentation for it.

Shane32:
If we can merge this PR, it will let @adamjones1 continue developing towards the v4 beta.  I should be able to get the documentation done within a couple days.

sungam3r (approved):

PR #2375, opened 2021-03-10
Change AstPrinter guts to internal

Shane32:
If someone needs access to the `AstPrinter` internals, we can make it public again.  Or refactor and then make it public again.  Seems that the only thing needed might be a hook in case someone adds another `ValueNode<T>` implementation, but with the new scalar types design, it should not be necessary.  All scalar nodes should be one of the predefined types.


Shane32:
Thanks I forgot.  Want to check it again?

sungam3r (approved):

PR #2379, opened 2021-03-10
Fix formatting for AstPrinter value nodes

Shane32:
Fixed bugs:
- Strings are not escaped
- Negative integers are not printed with the invariant culture, so the negative symbol can be incorrect
- Floating-point `double` values are not printed in the shortest form possible, either using an integer if possible or a "1.0E+44" syntax when necessary to represent very small or very large numbers.  Note that it is explicitly allowed that integer values can represent floating point input fields/arguments.
- Floating-point `double` values would not print very small values (e.g. less than 0.000000000001) at all.  Per the spec it should hold any value that can be represented by IEEE 754.
- Floating-point `decimal` values are not printed with the proper amount of precision, or in the shortest form possible
- Floating-point NaN/infinity values are printed instead of throwing an error
- AnyValue nodes are ignored, rather than throwing an exception
- EnumValue nodes with invalid characters in their names are printed instead of throwing an error

### DIFF
--- a/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
@@ -1,7 +1,9 @@
+using System;
 using System.Globalization;
 using GraphQL.Execution;
 using GraphQL.Language.AST;
 using GraphQL.Utilities;
+using GraphQL.Utilities.Federation;
 using Shouldly;
 using Xunit;
 
@@ -78,7 +80,7 @@ namespace GraphQL.Tests.Utilities
             int value = 3;
             var val = new IntValue(value);
             var result = _printer.Visit(val);
-            result.ShouldBe(value);
+            result.ShouldBe("3");
         }
 
         [Fact]
@@ -87,7 +89,7 @@ namespace GraphQL.Tests.Utilities
             long value = 3;
             var val = new LongValue(value);
             var result = _printer.Visit(val);
-            result.ShouldBe(value);
+            result.ShouldBe("3");
         }
 
         [Fact]
@@ -112,5 +114,69 @@ namespace GraphQL.Tests.Utilities
                 .Replace("\r\n", "\n")
                 .Replace("\r", "\n");
         }
+
+        [Fact]
+        public void anynode_throws()
+        {
+            Should.Throw<InvalidOperationException>(() => AstPrinter.Print(new AnyValue("")));
+        }
+
+        [Fact]
+        public void invalid_floatvalue_throws()
+        {
+            Should.Throw<InvalidOperationException>(() => AstPrinter.Print(new FloatValue(double.PositiveInfinity)));
+            Should.Throw<InvalidOperationException>(() => AstPrinter.Print(new FloatValue(double.NegativeInfinity)));
+            Should.Throw<InvalidOperationException>(() => AstPrinter.Print(new FloatValue(double.NaN)));
+        }
+
+        [Fact]
+        public void null_stringvalue_throws()
+        {
+            Should.Throw<InvalidOperationException>(() => AstPrinter.Print(new StringValue(null)));
+        }
+
+        [Theory]
+        [MemberData(nameof(NodeTests))]
+        public void prints_node(INode node, string expected)
+        {
+            AstPrinter.Print(node).ShouldBe(expected);
+        }
+
+        [Theory]
+        [MemberData(nameof(NodeTests))]
+        public void prints_node_cultures(INode node, string expected)
+        {
+            CultureTestHelper.UseCultures(() => prints_node(node, expected));
+        }
+
+        public static object[][] NodeTests = new object[][]
+        {
+            new object[] { new StringValue("test"), "\"test\"" },
+            new object[] { new StringValue("ab\"cd"), "\"ab\\\"cd\"" },
+            new object[] { new StringValue("ab\u0019cd"), "\"ab\\u0019cd\"" },
+            new object[] { new StringValue("\"abcd\""), "\"\\\"abcd\\\"\"" },
+            new object[] { new IntValue(int.MinValue), "-2147483648" },
+            new object[] { new IntValue(0), "0" },
+            new object[] { new IntValue(int.MaxValue), "2147483647" },
+            new object[] { new LongValue(long.MinValue), "-9223372036854775808" },
+            new object[] { new LongValue(0), "0" },
+            new object[] { new LongValue(long.MaxValue), "9223372036854775807" },
+            new object[] { new BigIntValue(new System.Numerics.BigInteger(decimal.MaxValue) * 1000 + 2), "79228162514264337593543950335002" },
+            new object[] { new FloatValue(double.MinValue), "-1.79769313486232E+308" },
+            new object[] { new FloatValue(double.Epsilon), "4.94065645841247E-324" },
+            new object[] { new FloatValue(double.MaxValue), "1.79769313486232E+308" },
+            new object[] { new FloatValue(0.00000001256), "1.256E-08" },
+            new object[] { new FloatValue(12.56), "12.56" },
+            new object[] { new FloatValue(3.33), "3.33" },
+            new object[] { new FloatValue(1.0), "1" },
+            new object[] { new FloatValue(34), "34" },
+            new object[] { new DecimalValue(1.0000m), "1"},
+            new object[] { new DecimalValue(decimal.MinValue), "-79228162514264337593543950335"},
+            new object[] { new DecimalValue(decimal.MaxValue), "79228162514264337593543950335"},
+            new object[] { new DecimalValue(0.00000000000000001m), "1E-17"},
+            new object[] { new BooleanValue(false), "false"},
+            new object[] { new BooleanValue(true), "true"},
+            new object[] { new EnumValue("TEST"), "TEST"},
+        };
     }
 }
--- a/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
@@ -127,6 +127,8 @@ namespace GraphQL.Tests.Utilities
         {
             var sample = new string(Enumerable.Range(0, 256).Select(x => (char)x).ToArray());
             var ret = AstPrinter.Print(new StringValue(sample));
+            foreach (var c in ret)
+                c.ShouldBeGreaterThanOrEqualTo(' ');
             var deserialized = System.Text.Json.JsonSerializer.Deserialize<string>(ret);
             deserialized.ShouldBe(sample);
         }
--- a/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
@@ -165,7 +165,7 @@ namespace GraphQL.Tests.Utilities
         public static object[][] NodeTests = new object[][]
         {
             new object[] { new StringValue("test"), @"""test""" },
-            new object[] { new StringValue("ab/cd"), @"""ab\/cd""" },
+            new object[] { new StringValue("ab/cd"), @"""ab/cd""" },
             new object[] { new StringValue("ab\bcd"), @"""ab\bcd""" },
             new object[] { new StringValue("ab\fcd"), @"""ab\fcd""" },
             new object[] { new StringValue("ab\rcd"), @"""ab\rcd""" },
--- a/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
@@ -127,17 +127,32 @@ namespace GraphQL.Tests.Utilities
         {
             var sample = new string(Enumerable.Range(0, 256).Select(x => (char)x).ToArray());
             var ret = AstPrinter.Print(new StringValue(sample));
-            foreach (var c in ret)
+
+            foreach (char c in ret)
                 c.ShouldBeGreaterThanOrEqualTo(' ');
+
             var deserialized = System.Text.Json.JsonSerializer.Deserialize<string>(ret);
             deserialized.ShouldBe(sample);
+
+            var token = GraphQLParser.Lexer.Lex(ret);
+            token.Kind.ShouldBe(GraphQLParser.TokenKind.STRING);
+            token.Value.ShouldBe(sample);
         }
 
         [Theory]
         [MemberData(nameof(NodeTests))]
         public void prints_node(INode node, string expected)
         {
-            AstPrinter.Print(node).ShouldBe(expected);
+            var printed = AstPrinter.Print(node);
+
+            printed.ShouldBe(expected);
+
+            if (node is StringValue str)
+            {
+                var token = GraphQLParser.Lexer.Lex(printed);
+                token.Kind.ShouldBe(GraphQLParser.TokenKind.STRING);
+                token.Value.ShouldBe(str.Value);
+            }
         }
 
         [Theory]
@@ -150,6 +165,9 @@ namespace GraphQL.Tests.Utilities
         public static object[][] NodeTests = new object[][]
         {
             new object[] { new StringValue("test"), @"""test""" },
+            new object[] { new StringValue("ab/cd"), @"""ab\/cd""" },
+            new object[] { new StringValue("ab\bcd"), @"""ab\bcd""" },
+            new object[] { new StringValue("ab\fcd"), @"""ab\fcd""" },
             new object[] { new StringValue("ab\rcd"), @"""ab\rcd""" },
             new object[] { new StringValue("ab\ncd"), @"""ab\ncd""" },
             new object[] { new StringValue("ab\tcd"), @"""ab\tcd""" },
--- a/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/AstPrinterTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Globalization;
+using System.Linq;
 using GraphQL.Execution;
 using GraphQL.Language.AST;
 using GraphQL.Utilities;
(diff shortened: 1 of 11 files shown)


Shane32:
Since `ArgumentsOfCorrectType` uses `IsValidLiteralValue` which uses `AstPrinter` to print the values of invalid nodes, and since the new default for printing floating-point values is to not print decimal places if they are not necessary, some tests have changed.

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -319,7 +325,7 @@ public AstPrintVisitor()
                 c.Print(f =>
                 {
                     var val = (decimal)f.Arg(x => x.Value);
-                    return val.ToString("0.0##############", CultureInfo.InvariantCulture);
+                    return val.ToString("G29", CultureInfo.InvariantCulture); //prints most compact form of value without losing any precision
```
Shane32, line 328: `decimal` values are designed to have no internal/unseen rounding errors.  So this prints all decimal values with full precision.  Truthfully, this is a bit pointless per spec, as a value with 29 decimal places cannot be stored within an IEEE 754 number.  However, it does not break the spec either.

sungam3r:
Do I understand correctly that all the `not` words in the initial comment should be replaced by `now`?

Shane32:
> Do I understand correctly that all the `not` words in the initial comment should be replaced by `now`?

It's a bug list, not describing new behavior.

But yeah...

src/GraphQL.Tests/Validation/ArgumentsOfCorrectTypeTests.cs
```diff
@@ -151,7 +151,7 @@ public void float_into_string()
             ShouldFailRule(_ =>
             {
                 _.Query = query;
-                Rule.badValue(_, "stringArg", "String", "1.0", 3, 32);
+                Rule.badValue(_, "stringArg", "String", "1", 3, 32);
```
sungam3r, line 154: Original query contains `1.0`. It will be incorrectly to show another value in the error message. The question is completely different. Why does `IsValidLiteralValue` call a `AstPrinter.Print` method? `valueAst` is `INode` and contains `SourceLocation` (start, end) so to get the required text we just need to call `originalquery.Subscting(valueAst.SourceLocation.Start, valueAst.SourceLocation.End - valueAst.SourceLocation.Start)`. Nothing to "print" here. But where to take `originalquery` 😄 ? Look at `ISchema schema`. It is not used 😮 ! Just change it to `ValidationContext`. There are only two usages from validation rules.

sungam3r:
`AstPrinter` is not the part of the code that I know well and understand. I will need more time to review.

src/GraphQL.Tests/Validation/ArgumentsOfCorrectTypeTests.cs
Shane32: Is the originalQuery always provided? Or is it optional if document is provided? I think this demonstrates a flaw in the design of the AST nodes - they should not be deserialized and in fact should not be different than the ones provided by parser. Separate discussion of course.

Shane32: Anyway it would seem you’ve found a separate bug, which could be reproduced without these changes. For example if the number was 2E+0 then it would have printed 2.0 before which would also be incorrect.

sungam3r: > Is the originalQuery always provided? Or is it optional if document is provided?

Yes - `GraphQLDocumentBuilder.Build`. There is no more separate `OriginalQuery` property since #2276.

Shane32:
??? I don’t understand the AstPrinter, and have no intention of doing so. The code seems very weird. But I simply made a very few changes that formats values appropriately based on the GraphQL spec.

src/GraphQL.Tests/Validation/ArgumentsOfCorrectTypeTests.cs
Shane32: The issue you noted with the error messages is fixed in #2397

sungam3r: ok

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
...
+                    return sb;
                 });
+
             });
 
             Config<BooleanValue>(c =>
             {
                 c.Field(x => x.Value);
-                c.Print(f => f.Arg(x => x.Value)?.ToString().ToLower(CultureInfo.InvariantCulture));
+                c.Print(f => (bool)f.Arg(x => x.Value) ? "true" : "false");
```
sungam3r, line 373: 👍

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -368,6 +405,8 @@ public AstPrintVisitor()
                 c.Print(p => $"{p.Arg(x => x.Name)}: {p.Arg(x => x.Value)}");
             });
 
+            Config<AnyValue>(c => c.Print(x => throw new InvalidOperationException("Cannot print AnyValue representations.")));
```
sungam3r, line 408: Somewhere there is such a requirement?

sungam3r, line 408:
```suggestion
            Config<AnyValue>(c => c.Print(x => throw new InvalidOperationException("Cannot print AnyValue representation.")));
```

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
...
             Config<EnumValue>(c =>
             {
                 c.Field(x => x.Name);
-                c.Print(p => p.Arg(x => x.Name));
+                c.Print(p =>
+                {
+                    var name = (string)p.Arg(x => x.Name);
+                    if (name == null)
+                        throw new InvalidOperationException("EnumValue name cannot be null");
+                    NameValidator.ValidateDefault(name, NamedElement.EnumValue);
```
sungam3r, line 384: I doubt that this check is required here. You need to print what is already there. It is possible that someone needs to print values like `KEBAB-ENUM`. Also I would move null check to `EnumValue` ctor from here.

sungam3r: And possible add null check into `NameNode` ctor.

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
                 c.Field(x => x.Value);
                 c.Print(f =>
                 {
-                    var val = f.Arg(x => x.Value);
-                    if (!string.IsNullOrWhiteSpace(val?.ToString()) && !val.ToString().StartsWith("\"", StringComparison.InvariantCulture))
+                    var val = (string)f.Arg(x => x.Value);
+                    if (val == null)
+                        throw new InvalidOperationException("StringValue cannot contain null");
```
sungam3r, line 341: Here is the same situation. This check is not the responsibility of `AstPrinter`. Let's move it into `StringValue` ctor.

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
...
+                    var sb = new System.Text.StringBuilder(val.Length + 2);
+                    sb.Append('\"');
```
sungam3r, line 343:
```suggestion
                    sb.Append('"');
```

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
...
+                    foreach (var ch in val)
                     {
-                        val = $"\"{val}\"";
+                        if (ch < ' ')
+                        {
+                            if (ch == '\n')
+                                sb.Append("\\\n");
```
sungam3r, line 349:
```suggestion
                                sb.Append("\\n");
```

```
StringBuilder b = new StringBuilder();
            b.Append("abc");
            b.Append("\\\n");
            b.Append("def");
            var s = b.ToString();
```
yields 
```
abc\
def
```

but we need `abc\ndef`

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
...
+                            else if (ch == '\r')
+                                sb.Append("\\\r");
```
sungam3r, line 351:
```suggestion
                                sb.Append("\\r");
```

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
...
+                            else if (ch == '\t')
+                                sb.Append("\\\t");
```
sungam3r, line 353:
```suggestion
                                sb.Append("\\t");
```

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
...
+                            else
+                                sb.Append("\\u" + ((int)ch).ToString("X4"));
+                        }
+                        else if (ch == '\\')
+                            sb.Append("\\\\");
+                        else if (ch == '\"')
```
sungam3r, line 359:
```suggestion
                        else if (ch == '"')
```

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -328,25 +336,54 @@ public AstPrintVisitor()
...
+                            sb.Append("\\\"");
+                        else
+                            sb.Append(ch);
                     }
-                    return val;
+                    sb.Append('\"');
```
sungam3r, line 364:
```suggestion
                    sb.Append('"');
```

src/GraphQL/Utilities/AstPrinter.cs
```diff
@@ -309,7 +311,13 @@ public AstPrintVisitor()
                 c.Print(f =>
                 {
                     var val = (double)f.Arg(x => x.Value);
-                    return val.ToString("0.0##############", CultureInfo.InvariantCulture);
+                    if (double.IsNaN(val) || double.IsInfinity(val))
```
sungam3r, line 314: I would move this check into `FloatValue` ctor.

src/GraphQL.Tests/Utilities/AstPrinterTests.cs
```diff
@@ -112,5 +114,69 @@ private static string MonetizeLineBreaks(string input)
...
         }
+
+        [Fact]
+        public void anynode_throws()
+        {
+            Should.Throw<InvalidOperationException>(() => AstPrinter.Print(new AnyValue("")));
+        }
+
+        [Fact]
+        public void invalid_floatvalue_throws()
```
sungam3r, line 125: I would move this into `FloatGraphTypeTests`

sungam3r: Or some other test for value nodes, not scalars.

sungam3r: Also for null_stringvalue_throws

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: Please add tests for \n \r \t.

sungam3r: Also what about other escaped symbols?
изображение

By the way I do not remember that there was `/` when I last read the specification.

src/GraphQL/Utilities/AstPrinter.cs
Shane32: I don't know how `AnyValue` is used for federation, but I know it's needed.  My best guess is that it's a placeholder for a part of the AST that's not part of the local instance, or something.  In any case, we certainly can't print it, without knowing what type it's supposed to be.

I figured throwing an error would be better than before, where it silently skipped printing the value (probably resulting in an invalid syntax).

src/GraphQL/Utilities/AstPrinter.cs
Shane32: Any enum value with invalid characters cannot be printed, as it does not meet the graphql language syntax.  What if the character had a curly brace in it?  Correct me if I'm wrong, but this ASTPrinter should print a string that conforms to the GraphQL syntax.  Allowing malformed documents to be created opens up security holes similar to the classic SQL injection attack.

src/GraphQL/Utilities/AstPrinter.cs
Shane32: Good catch.  I should have added tests :-(

Shane32: I would not encode the forward slash.  There is no reason to; it makes reading the resulting string harder.  As for the backspace character and form-feed characters, I have not seen those characters used since I wrote code for DOS!  Probably there's some use with telnet now, but I think we can live with those character's unicode encodings here!

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: Ok

Shane32:
Adding `IsInfinity` checks to the `FloatValue` constructor breaks some existing tests, where we check that parsing documents with values that overflow a double's maximum value is parsed as `double.Infinity`.  This is the default parsing algorithm for C# and we had elected to maintain that behavior.

Shane32:
I am partially reverting the change to the `FloatValue` constructor

Shane32:
See `GraphQL.Tests.Bugs.CoreToVanillaConverterTests.DocumentExecuter_really_big_double_valid`

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: Let's change this assertion `Any enum value with invalid characters cannot be printed` a bit -> `Any enum value with invalid characters cannot be created` . I suggest to move check inside `EnuValue` ctor.

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: > It is not required

See cut of the spec grammar above. It is required by the spec itself. Is not it so?

> we can live with those character's unicode encodings here

Why not do it as required in the specification?

sungam3r:
>  This is the current default parsing algorithm for .NET Core and we had elected to maintain that behavior.

I'm not sure that it was the right decision. Now I am not ready to say that we need to throw an exception in case of infinity values, but the probability of this is not zero.

src/GraphQL/Utilities/AstPrinter.cs
sungam3r, line 314:
```suggestion
                    if (double.IsInfinity(val))
```

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: I made changes myself not to delay the release.

src/GraphQL/Language/AST/VariableReference.cs
```diff
@@ -19,7 +18,7 @@ public VariableReference(NameNode name)
         /// <summary>
         /// Returns the name of the variable being referenced.
         /// </summary>
-        public string Name { get; }
+        public string Name => NameNode.Name;
```
sungam3r, line 21: This property implementation was aligned among all AST nodes.

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: From parser:
изображение
and
изображение

изображение

So forward slash **should** be escaped to work correctly. I'll add tests.

sungam3r: Well, I figured out that our parser works will with both options, both `"ab/cd"` and `"ab\/cd"` are parsed as `"ab/cd"`. But I think that we should print 100% compatible AST that can be parsed by any parser so I added forward slash to escaped symbols. Tests added.

sungam3r:
@Shane32 I'm done here. If everything is OK then merge and feel free to release v4. Finally! 👍

sungam3r (approved):

src/GraphQL/Utilities/AstPrinter.cs
Shane32: Reread the spec. String character is any source character the except three noted. That includes a forward slash.

Shane32: So printing a forward slash is 100% valid in any parser guaranteed.

Shane32: The parser MUST parse all characters because it may be present in a document. This is a serializer not a parser.

Shane32:
> @Shane32 I'm done here. If everything is OK then merge and feel free to release v4. Finally! 👍
> 
> Probably you wish to add some notes about this bugfix. In some sense it is a breaking (but the right) change.

Sounds good.

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: I don't understand. Where is it?

Shane32: Look up “Source Text” in the spec. SourceCharacter is
`/[\u0009\u000A\u000D\u0020-\uFFFF]/`

Which includes a forward slash. So string character also includes a forward slash because it is not a backslash or quote or line terminator

sungam3r: Aaa, I think I understood. SourceCharacter is `/[\u0009\u000A\u000D\u0020-\uFFFF]/`

Shane32: A forward slash MAY be escaped of course. 

And the backspace must be encoded as either `\b` or `\u0008` but not it is not required that it be one vs the other

Shane32: We can leave the changes you pushed by the way. I just personally wouldn’t encode the forward slash.

sungam3r: So only `/` MAY BE unescaped. Right? And all other `" \ b f n r t` SHOULD BE escaped.

sungam3r: Well, let's remove escaping for forward slash.

Shane32: > Well, let's remove escaping for forward slash.

Done

sungam3r (approved):

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: Make some notes if you wish and release. 👍

Shane32: I didn't add notes; it's just really a bugfix and listed along with the many other bugfixes for 4.0 in the search-prs-by-milestone list.  Too busy today to spend much time on it unfortunately.

Shane32:
Thanks for the help wrapping up this PR @sungam3r .  Really busy today.

src/GraphQL/Utilities/AstPrinter.cs
sungam3r: ok

PR #2380, opened 2021-03-10
Fix NameValidator

Shane32:
> All **types and directives** defined within a schema must not have a name which begins with "__" (two underscores), as this is used exclusively by GraphQL’s introspection system.

See: https://spec.graphql.org/June2018/#sec-Schema

> All fields defined within an Object type must not have a name which begins with "__" (two underscores), as this is used exclusively by GraphQL’s introspection system.

See: https://spec.graphql.org/June2018/#sec-Objects

> All arguments defined within a field must not have a name which begins with "__" (two underscores), as this is used exclusively by GraphQL’s introspection system.

See: https://spec.graphql.org/June2018/#sec-Field-Arguments

> Types and fields required by the GraphQL introspection system that are used in the same context as user‐defined types and fields are prefixed with "__" two underscores. This in order to avoid naming collisions with user‐defined GraphQL types. Conversely, GraphQL type system authors must not define any types, fields, arguments, or any other type system artifact with two leading underscores.

See: https://spec.graphql.org/June2018/#sec-Reserved-Names

Please review @sungam3r .  There does not seem to be any reason why an enum cannot be named "__TEST".  It is not a "type system artifact" -- it's a value.  Would you agree??  Whatever you think.

### DIFF
--- a/src/GraphQL.Tests/Utilities/NameValidatorTests.cs
+++ b/src/GraphQL.Tests/Utilities/NameValidatorTests.cs
@@ -31,8 +31,19 @@ namespace GraphQL.Tests.Utilities
             Should.Throw<ArgumentOutOfRangeException>(() => NameValidator.ValidateName(" ", NamedElement.Field));
 
         [Fact]
-        public void ValidateName_whenNameStartsWithReservedCharacters_throwsArgumentOutOfRange() =>
+        public void ValidateName_whenNameStartsWithReservedCharacters_throwsArgumentOutOfRange()
+        {
+            Should.Throw<ArgumentOutOfRangeException>(() => NameValidator.ValidateName("__dede", NamedElement.Type));
+            Should.Throw<ArgumentOutOfRangeException>(() => NameValidator.ValidateName("__dede", NamedElement.Directive));
             Should.Throw<ArgumentOutOfRangeException>(() => NameValidator.ValidateName("__dede", NamedElement.Field));
+            Should.Throw<ArgumentOutOfRangeException>(() => NameValidator.ValidateName("__dede", NamedElement.Argument));
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


sungam3r:
> There does not seem to be any reason why an enum cannot be named "__TEST".

enum _value_

sungam3r:
I think it's possible though it's better to ask from the spec authors - https://github.com/graphql/graphql-spec/issues/827

sungam3r:
https://github.com/graphql/graphql-spec/issues/827#issuecomment-824945827

😕

src/GraphQL.Tests/Utilities/NameValidatorTests.cs
```diff
@@ -31,8 +31,19 @@ public void ValidateName_whenNameIsEmpty_throwsArgumentOutOfRange()
             Should.Throw<ArgumentOutOfRangeException>(() => NameValidator.ValidateName(" ", NamedElement.Field));
 
         [Fact]
-        public void ValidateName_whenNameStartsWithReservedCharacters_throwsArgumentOutOfRange() =>
+        public void ValidateName_whenNameStartsWithReservedCharacters_throwsArgumentOutOfRange()
```
sungam3r, line 34: @Shane32 These tests could be merged.

Shane32: Yeah... I got too much else to do to mess with it, both at work, at home, and here lol.  Go ahead though if you have time.  I really want to finish up these serializer enhancements and write some documentation for it.

I really think the answer benjie gave should be wrong.  But it seems like the graphql spec progresses at a snail's pace.  This is also the second time when the answer given is partially driven by a "reference implementation" even when the implementation is demonstrably incorrect (from my point of view of course).

sungam3r: I know, I understand.

PR #2386, opened 2021-03-10
Add ArgumentConfig and FieldConfig.ArgumentFor

### LINKED ISSUE #1867: Field arguments not available on FieldConfig
I'm using the `Action<SchemaBuilder>` parameter of `Schema.For` to customise some GraphQL types in a schema. In particular I'd like to set the descriptions of some field arguments this way, but it seems `FieldConfig` (from `builder.Types.For("myType").FieldFor("myField", services)`) doesn't have an `Arguments` field exposing this metadata. Is there a particular reason for this? It seems like an omission to me.

sungam3r:
fixes #1867

### DIFF
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -171,6 +171,7 @@ This may be needed to get the parameters of parent nodes.
 
 ### Other Features
 
+* New method `FieldConfig.ArgumentFor`
 * New property `ISchema.ValueConverters`
 * New method `IParentExecutionNode.ApplyToChildren`
 * `ExecutionStrategy` exposes a number of `protected virtual` methods that can be used to alter the execution
@@ -540,6 +541,8 @@ default continues to be `ResolverType.Resolver`.
 * `IResolveFieldContext.Fragments` was removed; use `IResolveFieldContext.Document.Fragments` instead
 * `ExecutionContext.Fragments` was removed; use `ExecutionContext.Document.Fragments` instead
 * `AbstractGraphTypeExtensions.GetTypeOf` was removed; use `AbstractGraphTypeExtensions.GetObjectType` instead
+* `TypeConfig.FieldFor(string, IServiceProvider)` and `TypeConfig.SubscriptionFieldFor(string, IServiceProvider)` 
+  methods were merged into single `TypeConfig.FieldFor(string)` method and just return the required configuration without its initialization
 
 ### Other Breaking Changes (including but not limited to)
 
--- a/src/GraphQL/Utilities/SchemaBuilder.cs
+++ b/src/GraphQL/Utilities/SchemaBuilder.cs
@@ -3,7 +3,10 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+using System.Threading.Tasks;
 using GraphQL.Introspection;
+using GraphQL.Reflection;
+using GraphQL.Resolvers;
 using GraphQL.Types;
 using GraphQLParser;
 using GraphQLParser.AST;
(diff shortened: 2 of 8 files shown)


src/GraphQL/Utilities/TypeConfig.cs
```diff
@@ -44,53 +58,11 @@ public void IsTypeOf<T>()
...
-                    config.AsyncSubscriber = new AsyncEventStreamResolver(config.SubscriberAccessor, serviceProvider);
-                }
-                else
-                {
-                    config.Subscriber = new EventStreamResolver(config.SubscriberAccessor, serviceProvider);
-                }
-            }
-
-            return config;
-        }
```
sungam3r, line 93: Code was moved into `SchemaBuilder`. Note that `IServiceProvider` parameter was removed from public API here. It is absolutely not needed here.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -280,20 +283,62 @@ protected virtual IObjectGraphType ToObjectGraphType(GraphQLObjectTypeDefinition
             return type;
         }
 
+        private void InitializeField(FieldConfig config, Type parentType)
```
Shane32, line 286:
```suggestion
        protected virtual void InitializeField(FieldConfig config, Type parentType)
```

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -280,20 +283,62 @@ protected virtual IObjectGraphType ToObjectGraphType(GraphQLObjectTypeDefinition
...
+                var attrs = config.ResolverAccessor.GetAttributes<GraphQLAttribute>();
+                if (attrs != null)
+                {
+                    foreach (var a in attrs)
+                        a.Modify(config);
+                }
+            }
+        }
+
+        private void InitializeSubscriptionField(FieldConfig config, Type parentType)
```
Shane32, line 302:
```suggestion
        protected virtual void InitializeSubscriptionField(FieldConfig config, Type parentType)
```

Shane32 (changes requested):

src/GraphQL/Utilities/SchemaBuilder.cs
sungam3r: Both methods are called from other virtual methods. I do not consider it right to make them virtual. In fact, it is just a piece of private implementation. If the user wants to change the implementation, it will override the calling virtual method.

Shane32: Ok.

Shane32 (approved):

docs2/site/docs/migrations/migration4.md
```diff
@@ -540,6 +541,8 @@ default continues to be `ResolverType.Resolver`.
 * `IResolveFieldContext.Fragments` was removed; use `IResolveFieldContext.Document.Fragments` instead
 * `ExecutionContext.Fragments` was removed; use `ExecutionContext.Document.Fragments` instead
 * `AbstractGraphTypeExtensions.GetTypeOf` was removed; use `AbstractGraphTypeExtensions.GetObjectType` instead
+* `TypeConfig.FieldFor(string, IServiceProvider)` and `TypeConfig.SubscriptionFieldFor(string, IServiceProvider)` 
+  methods were merged into single `TypeConfig.FieldFor(string)` method and just return the required configuration without its initialization
```
Shane32, line 545: This would more appropriately say that the methods were removed, if they are not going to be protected.

Shane32 (changes requested):

docs2/site/docs/migrations/migration4.md
sungam3r: This method remained public. It is needed to set the configuration, not internal things.

Shane32 (approved):

docs2/site/docs/migrations/migration4.md
Shane32: Oh ok 👍

PR #2388, opened 2021-03-10
Allow overriding built-in scalar graph type

Shane32:
Includes tests for both code-first and schema-first.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Issue2387.cs
+++ b/src/GraphQL.Tests/Bugs/Issue2387.cs
@@ -64,7 +64,6 @@ namespace GraphQL.Tests.Bugs
             var json = await schema.ExecuteAsync(_ =>
             {
                 _.Query = "{ testInput(arg:123) }";
-                _.Root = new SchemaFirstRoot();
             });
             json.ShouldBeCrossPlatJson("{\"data\":{\"testInput\": \"122\"}}");
         }
--- a/src/GraphQL.Tests/Bugs/Issue2387.cs
+++ b/src/GraphQL.Tests/Bugs/Issue2387.cs
@@ -100,6 +100,7 @@ type Query {
             {
                 return 123;
             }
+            
             public string TestInput(IResolveFieldContext context)
             {
                 return context.GetArgument<int>("arg").ToString();
--- a/src/GraphQL.Tests/Bugs/Issue2387.cs
+++ b/src/GraphQL.Tests/Bugs/Issue2387.cs
@@ -52,7 +52,6 @@ namespace GraphQL.Tests.Bugs
             var json = await schema.ExecuteAsync(_ =>
             {
                 _.Query = "{ testOutput }";
-                _.Root = new SchemaFirstRoot();
             });
             json.ShouldBeCrossPlatJson("{\"data\":{\"testOutput\": 124}}");
         }
--- a/src/GraphQL.Tests/Bugs/Issue2387.cs
+++ b/src/GraphQL.Tests/Bugs/Issue2387.cs
@@ -1,15 +1,109 @@
+using System.Threading.Tasks;
+using GraphQL.Language.AST;
+using GraphQL.SystemTextJson;
 using GraphQL.Types;
 using Xunit;
 
 namespace GraphQL.Tests.Bugs
 {
     // https://github.com/graphql-dotnet/graphql-dotnet/issues/2387
-    public class Issue2387 : QueryTestBase<Issue2387.MySchema>
+    public class Issue2387_OverrideBuiltInScalars : QueryTestBase<Issue2387_OverrideBuiltInScalars.MySchema>
     {
         [Fact]
-        public void can_override_built_in_scalar_types()
+        public void codefirst_output()
         {
-            AssertQuerySuccess("{testInt}", "{\"testInt\": 124}");
+            AssertQuerySuccess("{ testOutput }", "{\"testOutput\": 124}");
+        }
+
+        [Fact]
+        public void codefirst_parseliteral()
+        {
+            AssertQuerySuccess("{ testInput(arg:123) }", "{\"testInput\": \"122\"}");
+        }
+
+        [Fact]
+        public void codefirst_parsevalue()
+        {
+            AssertQuerySuccess("query ($arg: Int) { testInput(arg:$arg) }", "{\"testInput\": \"122\"}", "{\"arg\":123}".ToInputs());
+        }
+
+        [Fact]
+        public void codefirst_nonnull_output()
+        {
+            AssertQuerySuccess("{ testNonNullOutput }", "{\"testNonNullOutput\": 124}");
+        }
+
+        [Fact]
+        public void codefirst_nonnull_parseliteral()
+        {
+            AssertQuerySuccess("{ testNonNullInput(arg:123) }", "{\"testNonNullInput\": \"122\"}");
+        }
+
+        [Fact]
+        public void codefirst_nonnull_parsevalue()
+        {
+            AssertQuerySuccess("query ($arg: Int!) { testNonNullInput(arg:$arg) }", "{\"testNonNullInput\": \"122\"}", "{\"arg\":123}".ToInputs());
+        }
+
+        [Fact]
+        public async Task schemafirst_output()
+        {
+            var schema = BuildSchemaFirst();
+            var json = await schema.ExecuteAsync(_ =>
+            {
+                _.Query = "{ testOutput }";
+                _.Root = new SchemaFirstRoot();
+            });
+            json.ShouldBeCrossPlatJson("{\"data\":{\"testOutput\": 124}}");
+        }
+
+        [Fact]
+        public async Task schemafirst_parseliteral()
+        {
+            var schema = BuildSchemaFirst();
+            var json = await schema.ExecuteAsync(_ =>
+            {
+                _.Query = "{ testInput(arg:123) }";
+                _.Root = new SchemaFirstRoot();
+            });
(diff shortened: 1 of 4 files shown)


src/GraphQL.Tests/Bugs/Issue2387.cs
```diff
@@ -0,0 +1,157 @@
...
+            }
+
+            public override object ParseValue(object value)
+                => value is int i ? i - 1 : value;
+
+            public override object Serialize(object value)
+                => value is int i ? (object)(i + 1) : null;
+        }
+    }
+
```
sungam3r, line 156:
```suggestion
```

src/GraphQL.Tests/Bugs/Issue2387.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        [Fact]
+        public async Task schemafirst_parseliteral()
+        {
+            var schema = BuildSchemaFirst();
+            var json = await schema.ExecuteAsync(_ =>
+            {
+                _.Query = "{ testInput(arg:123) }";
+                _.Root = new SchemaFirstRoot();
```
sungam3r, line 67:
```suggestion
```

Created by `DefaultServiceProvider` ?

src/GraphQL.Tests/Bugs/Issue2387.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        [Fact]
+        public async Task schemafirst_output()
+        {
+            var schema = BuildSchemaFirst();
+            var json = await schema.ExecuteAsync(_ =>
+            {
+                _.Query = "{ testOutput }";
+                _.Root = new SchemaFirstRoot();
```
sungam3r, line 55:
```suggestion
```

src/GraphQL.Tests/Bugs/Issue2387.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        [GraphQLMetadata("Query")]
+        public class SchemaFirstRoot
+        {
+            public int TestOutput()
+            {
+                return 123;
+            }
+            public string TestInput(IResolveFieldContext context)
```
sungam3r, line 103:
```suggestion
            }
            
            public string TestInput(IResolveFieldContext context)
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -232,6 +235,7 @@ private static IEnumerable<IGraphType> GetSchemaTypes(ISchema schema, IServicePr
         }
 
         protected internal virtual Dictionary<string, IGraphType> Dictionary { get; } = new Dictionary<string, IGraphType>();
+        private readonly Dictionary<Type, IGraphType> _typeDictionary = null;
```
sungam3r, line 238:
```suggestion
        private readonly Dictionary<Type, IGraphType> _typeDictionary;
```

src/GraphQL.Tests/Bugs/Issue2387.cs
```diff
@@ -0,0 +1,157 @@
...
+                return context.GetArgument<int>("arg").ToString();
+            }
+        }
+
+        public class MySchema : Schema
+        {
+            public MySchema()
+            {
+                Query = new MyQuery();
+                RegisterType(new MyIntGraphType());
```
sungam3r, line 115: Change this line to `RegisterType(typeof(MyIntGraphType));` and tests will fail. See `GetSchemaTypes` . To your idea to work, you need to register specially added types/instances in the first place. And it would be nice to add a comment on this so that no one then changed the order of returned types.

sungam3r: And it is also necessary to write in the documentation "in large letters" -  _to change the scalars, you need to register them first_.

Shane32: Right. We will need to be very explicit about how this is supported. Was going to update the scalar docs PR once we finish this PR.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -698,32 +696,44 @@ private IGraphType ConvertTypeReference(INamedType parentType, IGraphType type)
...
                 {
                     // Soft schema configuration error.
                     // Intentionally or inadvertently, a situation may arise when the same GraphType is registered more that one time.
                     // This may be due to the simultaneous registration of GraphType instances and the GraphType types. In this case
                     // the duplicate MUST be ignored, otherwise errors will occur.
                 }
+                else if (type.IsAssignableFrom(existingGraphType.GetType()) && typeof(ScalarGraphType).IsAssignableFrom(type))
+                {
+                    // This can occur when a built-in scalar graph type is overridden by preregistering a replacement graph type that
+                    // has the same name and inherits from it
```
sungam3r, line 719:
```suggestion
                    // has the same name and inherits from it.
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -330,13 +334,7 @@ private set
 
                 lock (_lock)
```
sungam3r, line 335: Why do we need these lock? I deleted it and all tests are green.

sungam3r: I suggest to remove it at all. Also remove private setter for ` public IGraphType this[string typeName]` and inline its call to `  SetGraphType(typeName, value);` in `ConvertTypeReference`.

sungam3r: And even remove
```
if (type == null)
                    throw new ArgumentOutOfRangeException(nameof(type), "A type is required to lookup.");
```
check since it's private method.

sungam3r: And maybe change `  private IGraphType this[Type type] ` indexer to method like `Find(Type type)` because it's always hard to search indexer usages`.

Shane32: The locks should not be needed. Types cannot be added after SchemaTypes is initialized and a dictionary can be read from concurrently as long as there are no writers

sungam3r: Seems to be a good consequence of refactoring.

Shane32: I would like to spend some extensive time cleaning up SchemaTypes. The suggested changes are all private and do not affect v4 release.

Shane32: I might rewrite the whole thing. There’s a lot of duplicate code in the constructor alone! But maybe it’s all fine and just needs a few tweaks

Shane32: If you have the code open already, go ahead and push some cleanup to it. I’ll be busy until this weekend.

sungam3r: Done

Shane32:
Changes look great @sungam3r  !

Shane32:
@sungam3r Since introspection types did not utilize `IntGraphType`, the tests were not completely valid, as attempting to override `StringGraphType` would fail.  Added tests and bugfix for overriding `StringGraphType`.

sungam3r:
Do you think that option with schema visitor to override scalars is simpler for us to maintain ? It does not require to jump over fences.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -107,20 +106,26 @@ public SchemaTypes(ISchema schema, IServiceProvider serviceProvider)
...
-                   {
-                       SetGraphType(name, type);
-                   }
+                   SetGraphType(name, type);
                    ctx.AddType(name, type, null);
                },
                typeMappings);
 
+            // Add manually-added scalar types. To allow overriding of built-in scalars, these must be added
+            // prior to adding introspection types.
```
sungam3r, line 122:
```suggestion
            // prior to adding any other types (including introspection types).
```

sungam3r (approved):

Shane32:
> Do you think that option with schema visitor to override scalars is simpler for us to maintain ? It does not require to jump over fences.

I'm concerned with the schema visitor option in that variable values are coerced based on the type name provided in the document - see `ValidationContext.cs:208` within `GetVariableValues()`:

```cs
                foreach (var variableDef in variableDefinitions.List)
                {
                    // find the IGraphType instance for the variable type
                    var graphType = variableDef.Type.GraphTypeFromType(schema);

                    ...
```

So unless `schema.AllTypes[...]` returns a reference to the replaced scalar type (see `TypeExtensions.cs:51` within `GraphTypeFromType()`), the built-in scalar type will be used.  So this PR becomes necessary for variables whether or not the schema visitor option is used.

All in all, I would say that we have a good solution here, and it should be very easy to document, as it pretty much just requires someone to inherit from the scalar type they want to replace (e.g. `class MyIntGraphType : IntGraphType`) and then register it with a single line of code within the schema constructor (`RegisterType(...)`) or after the schema is built (`schema.RegisterType(...)`)  but before it is initialized.  So it makes it easy to provide documentation for and easy to use.  Fortunately it only required a very few changes to `SchemaTypes` to implement also.

sungam3r:
> So unless schema.AllTypes[...] returns a reference to the replaced scalar type (see TypeExtensions.cs:51 within GraphTypeFromType()), the built-in scalar type will be used. 

You are right. I forgot about SchemaTypes. I' ll post PR to fix it.

PR #2392, opened 2021-03-12
Replace scalar after schema initialization

sungam3r:
Alternative to #2388. See https://github.com/graphql-dotnet/graphql-dotnet/pull/2388#issuecomment-797649109

### DIFF
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -487,6 +487,12 @@ This property was renamed. If you have explicitly set this property in an attrib
 directly anywhere, then just change its name. If you did not explicitly set this property, the
 default continues to be `ResolverType.Resolver`.
 
+### No more static predefined directives
+
+`DirectiveGraphType.Deprecated`, `DirectiveGraphType.Include` and `DirectiveGraphType.Skip` static properties were moved
+into corresponding virtual properties within `SchemaDirectives` class. This is done in order to be able to independently change
+the directives implementation without mutual influence on the schemas using them.
+
 ### API Cleanup
 
 * `GraphQL.Instrumentation.StatsReport` and its associated classes have been removed. Please copy the source code into
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -1711,9 +1711,6 @@ namespace GraphQL.Types
     }
     public class DirectiveGraphType : GraphQL.Utilities.MetadataProvider, GraphQL.Types.INamedType, GraphQL.Types.IProvideDescription
     {
-        public static readonly GraphQL.Types.DeprecatedDirective Deprecated;
-        public static readonly GraphQL.Types.IncludeDirective Include;
-        public static readonly GraphQL.Types.SkipDirective Skip;
         public DirectiveGraphType(string name, params GraphQL.Types.DirectiveLocation[] locations) { }
         public DirectiveGraphType(string name, System.Collections.Generic.IEnumerable<GraphQL.Types.DirectiveLocation> locations) { }
         public GraphQL.Types.QueryArguments Arguments { get; set; }
@@ -2173,6 +2170,9 @@ namespace GraphQL.Types
     {
         public SchemaDirectives() { }
         public int Count { get; }
+        public virtual GraphQL.Types.DeprecatedDirective Deprecated { get; }
+        public virtual GraphQL.Types.IncludeDirective Include { get; }
+        public virtual GraphQL.Types.SkipDirective Skip { get; }
         public GraphQL.Types.DirectiveGraphType Find(string name) { }
         public System.Collections.Generic.IEnumerator<GraphQL.Types.DirectiveGraphType> GetEnumerator() { }
         public void Register(GraphQL.Types.DirectiveGraphType directive) { }
@@ -2387,7 +2387,6 @@ namespace GraphQL.Utilities
     }
     public class DeprecatedDirectiveVisitor : GraphQL.Utilities.BaseSchemaNodeVisitor
     {
-        protected static readonly string DeprecatedDefaultValue;
         public DeprecatedDirectiveVisitor() { }
         public static GraphQL.Utilities.DeprecatedDirectiveVisitor Instance { get; }
         public override void VisitEnumValue(GraphQL.Types.EnumValueDefinition value, GraphQL.Types.EnumerationGraphType type, GraphQL.Types.ISchema schema) { }
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -377,6 +377,8 @@ namespace GraphQL
             where TSchema : GraphQL.Types.ISchema { }
         public static void RegisterVisitor<TVisitor>(this GraphQL.Types.ISchema schema)
             where TVisitor : GraphQL.Utilities.ISchemaNodeVisitor { }
+        public static TSchema ReplaceScalar<TSchema>(this TSchema schema, GraphQL.Types.ScalarGraphType scalar)
+            where TSchema : GraphQL.Types.ISchema { }
         public static void Run(this GraphQL.Utilities.ISchemaNodeVisitor visitor, GraphQL.Types.ISchema schema) { }
     }
     public static class StringExtensions
@@ -1918,8 +1920,7 @@ namespace GraphQL.Types
     }
     public interface IProvideResolvedType
     {
-        GraphQL.Types.IGraphType ResolvedType { get; }
-        System.Type Type { get; }
+        GraphQL.Types.IGraphType ResolvedType { get; set; }
     }
     public interface ISchema : GraphQL.Types.IProvideDescription, GraphQL.Types.IProvideMetadata
     {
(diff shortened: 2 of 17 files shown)


src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -1918,8 +1920,7 @@ namespace GraphQL.Types
     }
     public interface IProvideResolvedType
     {
-        GraphQL.Types.IGraphType ResolvedType { get; }
-        System.Type Type { get; }
```
sungam3r, line 1922: Was not used. In addition, the name of the interface hints that this property shoul not be here.

sungam3r: No breaking change aslo since there is no `IProvideResolvedType.Type` property in `master`.

src/GraphQL.Tests/Bugs/Issue2387_OverrideBuiltInScalars.Alt.cs
```diff
@@ -0,0 +1,259 @@
...
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/2387
+    public class Issue2387_OverrideBuiltInScalars_Alt : QueryTestBase<Issue2387_OverrideBuiltInScalars_Alt.MySchema>
+    {
+        public Issue2387_OverrideBuiltInScalars_Alt()
+        {
+            Services.Singleton<MySchema>();
```
sungam3r, line 15: I was worth me some time to understand why the tests failed, and the replacement of the scalars did not work. It turned out that by default schema works as a transient dependency and a new one is created every time accessing `QueryTestBase.Schema`.

src/GraphQL/SchemaExtensions.cs
```diff
@@ -245,6 +246,52 @@ public static void Run(this ISchemaNodeVisitor visitor, ISchema schema)
...
+        private sealed class ReplaceScalarVisitor : BaseSchemaNodeVisitor
+        {
+            private readonly ScalarGraphType _replacement;
+
+            public ReplaceScalarVisitor(ScalarGraphType replacement)
+            {
+                _replacement = replacement ?? throw new ArgumentNullException(nameof(replacement));
+            }
+
+            public override void VisitSchema(ISchema schema) => schema.AllTypes.Dictionary[_replacement.Name] = _replacement;
```
sungam3r, line 273: @Shane32 Fixed https://github.com/graphql-dotnet/graphql-dotnet/pull/2388#issuecomment-797574002

sungam3r:
😕 Somehow scalar replacement "leaks" into other tests:
изображение

I cannot find any static data into schema...

sungam3r:
Found! 
изображение

src/GraphQL/SchemaExtensions.cs
Shane32: Notice that it was not possible to write this line of code outside of the library because `Dictionary` is internal.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -1918,8 +1917,7 @@ namespace GraphQL.Types
...
+        GraphQL.Types.IGraphType ResolvedType { get; set; }
```
Shane32, line 1920: I don't like the fact that the field is writable.
```suggestion
        GraphQL.Types.IGraphType ResolvedType { get; internal set; }
```

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: I would prefer to leave it here so the type, when available, can be examined easily.  But it will not _always_ be present and is not required at runtime, so it makes sense to remove it if you really think we should.

docs2/site/docs/migrations/migration4.md
```diff
@@ -487,6 +487,12 @@ This property was renamed. If you have explicitly set this property in an attrib
 directly anywhere, then just change its name. If you did not explicitly set this property, the
 default continues to be `ResolverType.Resolver`.
 
+### No more static predefined directives
+
+`DirectiveGraphType.Deprecated`, `DirectiveGraphType.Include` and `DirectiveGraphType.Skip` static properties were moved
+into corresponding virtual properties within `SchemaDirectives` class. This is done in order to be able to independently change
+the directives implementation without mutual influence on the schemas using them.
```
Shane32, line 494: 👍

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: Our goal should be to not let runtime code have write-access to the schema.  I understand we are a long way from that now, but I do not think we need this writable at this time.  Even the visitor you added _could_ have been written without it.

Shane32 (commented):
I have listed a few comments, but to summarize, it is nice to have this available within the schema, but it feels just a little like a hack.  For example, schema configuration should be available within the schema constructor -- is that possible here?  We also are modifying the schema after it's been initialized and verified.  There's no protection against modifying an initialized schema.  And finally, it's using internal bits of the library to modify `SchemaTypes`.

So I'm 50-50 on including it.  Fine if you want to, probably nice to have for certain scenarios, but probably not necessary.

src/GraphQL/SchemaExtensions.cs
Shane32: This line of code needs to verify that the type being replaced was indeed a scalar type.  Otherwise it would be possible to overwrite an `InputObjectGraphType` reference with a `ScalarGraphType` reference.

src/GraphQL.Tests/Bugs/Issue2387_OverrideBuiltInScalars.Alt.cs
```diff
@@ -0,0 +1,259 @@
...
+using GraphQL.Language.AST;
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/2387
+    public class Issue2387_OverrideBuiltInScalars_Alt : QueryTestBase<Issue2387_OverrideBuiltInScalars_Alt.MySchema>
```
Shane32, line 11:
```suggestion
    // https://github.com/graphql-dotnet/graphql-dotnet/issues/2392
    public class Issue2392_OverrideBuiltInScalars_Alt : QueryTestBase<Issue2392_OverrideBuiltInScalars_Alt.MySchema>
```
???  either way is fine with me.

src/GraphQL/SchemaExtensions.cs
sungam3r: > Notice that it was not possible to write ...

I know so I wrote special method.

> This line of code needs to verify...

Right. Done.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: I would like to remove it from here. This property is associated with another problem. It does not necessarily correspond to the value in `ResolvedType` (sometimes absolutely does not match as if the scalars are replaced).

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: изображение

Breaking change for all implementations.

sungam3r:
> We also are modifying the schema after it's been initialized and verified. There's no protection against modifying an initialized schema.

It's by design. The schema allows modification at any time. Nothing prevents, for example, change the descriptions of any field on the go. It's just for the responsibility of the caller. Changing the schema after initialization is an advanced scenario and I believe that the caller must understand the consequences. It is worth noting this in the documentation.

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: Ok

src/GraphQL.Tests/Bugs/Issue2387_OverrideBuiltInScalars.Alt.cs
sungam3r: Done

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: Oh I forgot about that.  There's a feature request for that, since it's allowed by the .NET runtime, but not C# -- I can't find the link now.

Of course we could add explicit implementations for all classes that use `ResolvedType`, and then it would not be a breaking change -- either now or to make it settable in the future.

But here's the thing: do we really want to make this settable???

Shane32: > > We also are modifying the schema after it's been initialized and verified. There's no protection against modifying an initialized schema.
> 
> It's by design. The schema allows modification at any time. Nothing prevents, for example, change the descriptions of any field on the go. It's just for the responsibility of the caller. Changing the schema after initialization is an advanced scenario and I believe that the caller must understand the consequences. It is worth noting this in the documentation.

Ok

Shane32 (approved):
I'm fine with including this, although I would not suggest this as the preferred method of changing a scalar type, due to the fact that the call to `ReplaceScalar` does not reside in the schema constructor, and as mentioned before, the lack of validating the schema after the schema is configured.  I would say this alternate code can be used if necessary to replace scalars that were explicitly set via `ResolvedType`.

sungam3r:
OK. You can directly write in the documentation.

Shane32:
> OK. You can directly write in the documentation.

Done (review in #2357 )

PR #2397, opened 2021-03-14
Less AstPrinter.Print calls

Shane32:
Fixes issue noted in #2379


src/GraphQL/AstExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Provides extension methods for AST nodes.
+    /// </summary>
+    public static class AstExtensions
+    {
+        /// <summary>
+        /// Returns the original string that was parsed into the specified node.
```
sungam3r, line 12:
```suggestion
        /// Returns the original string that was parsed from provided document into the specified node.
```

src/GraphQL/AstExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// </summary>
+        public static string ToString(this INode node, Document document)
```
sungam3r, line 14:
```suggestion
        public static string TakeFrom(this INode node, Document document)
```

It seems to me that the special name here instead of `ToString()` will be more appropriate.

sungam3r: But I'm not sure.

sungam3r:
`using GraphQL.Utilities;` became unnecessary.

src/GraphQL/Validation/Rules/OverlappingFieldsCanBeMerged.cs
```diff
@@ -628,6 +628,7 @@ private static bool SameArguments(Arguments arguments1, Arguments arguments2)
 
         private static bool SameValue(Argument arg1, Argument arg2)
         {
+            // normalize values prior to comparison by using AstPrinter.Print rather than INode.ToString(document)
```
sungam3r, line 631:
```suggestion
            // normalize values prior to comparison by using AstPrinter.Print rather than INode.TakeFrom(document)
```
?

sungam3r:
And what about fix for `IsValidLiteralValue` ?

Shane32:
> And what about fix for `IsValidLiteralValue` ?

Somehow I missed that.  I moved the method into `ValidationContext`.

Shane32:
> ToString / TakeFrom

I don't mind considering another name, but `node.TakeFrom(document)` doesn't seem to signify what the method does.

Perhaps we might consider an method or extension method on `Document` rather than `INode`...

I don't really have any other ideas offhand.

sungam3r:
>  I moved the method into ValidationContext.

Yes, this is also a good option.

> Perhaps node.StringFrom(document)?

OK.

> Perhaps we might consider an method or extension method on Document rather than INode...

No special difference.

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -455,5 +455,123 @@ object ParseValueObject(IInputObjectGraphType graphType, VariableDefinition vari
...
         }
+
+        /// <summary>
+        /// Validates that the specified AST value is valid for the specified scalar or input graph type.
+        /// Graph types that are lists or non-null types are handled appropriately by this method.
+        /// Returns a string representing the errors encountered while validating the value.
+        /// </summary>
+        public string IsValidLiteralValue(IGraphType type, IValue valueAst)
+        {
+            // see also ExecutionHelper.AssertValidVariableValue
```
sungam3r, line 466: Does `AssertValidVariableValue` still exist?

sungam3r:
Breaking, so needs some notes.

Shane32:
> Breaking, so needs some notes.

I wrote some; they are gone! 🤷‍♂️

src/GraphQL/Validation/ValidationContext.cs
Shane32: Nope

sungam3r (approved):
👍

PR #2403, opened 2021-03-15
Directives cleanup

sungam3r:
Inspired by https://github.com/graphql-java/graphql-java/pull/2221#issuecomment-799582628 and further discussion.

### DIFF
--- a/src/GraphQL.Tests/Files/JSON/AppliedDirectivesResult.json
+++ b/src/GraphQL.Tests/Files/JSON/AppliedDirectivesResult.json
@@ -8,7 +8,7 @@
           "args": [
             {
               "name": "quality",
-              "value": "high"
+              "value": "\"high\""
             }
           ]
         }
@@ -329,7 +329,7 @@
                   "args": [
                     {
                       "name": "quality",
-                      "value": "low"
+                      "value": "\"low\""
                     }
                   ]
                 }
@@ -343,7 +343,7 @@
                       "args": [
                         {
                           "name": "quality",
-                          "value": "moderate"
+                          "value": "\"moderate\""
                         }
                       ]
                     }
--- a/src/GraphQL.Tests/Files/JSON/AppliedDirectivesResult.json
+++ b/src/GraphQL.Tests/Files/JSON/AppliedDirectivesResult.json
@@ -317,12 +317,7 @@
           "appliedDirectives": [
             {
               "name": "traits",
-              "args": [
-                {
-                  "name": "quality",
-                  "value": "unknown"
-                }
-              ]
+              "args": []
             }
           ],
           "fields": [
--- a/src/GraphQL/Introspection/__AppliedDirective.cs
+++ b/src/GraphQL/Introspection/__AppliedDirective.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Linq;
 using GraphQL.Types;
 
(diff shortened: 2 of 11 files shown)


src/GraphQL/Introspection/__DirectiveArgument.cs
```diff
@@ -22,25 +29,24 @@ public __DirectiveArgument()
...
-            Field<StringGraphType>(
+            Field<NonNullGraphType<StringGraphType>>(
                 "value",
                 "A GraphQL-formatted string representing the value for argument.",
                 resolve: context =>
                 {
                     var argument = context.Source;
                     if (argument.Value == null)
-                        return null;
+                        return "null";
```
sungam3r, line 39: GraphQL-formatted value for `null`.

src/GraphQL/Introspection/__AppliedDirective.cs
```diff
@@ -24,36 +23,8 @@ public __AppliedDirective()
...
-                            return new DirectiveArgument(arg.Name) //TODO: return QueryArgument instead of DirectiveArgument?
-                            {
-                                Value = arg.DefaultValue,
-                                ResolvedType = arg.ResolvedType
-                            };
-                        }
-                        else
-                        {
-                            return null;
-                        }
```
sungam3r, line 54: It was wrong 😞 . The things are much easier.

src/GraphQL/Types/Directives/DirectiveArgument.cs
```diff
@@ -35,7 +35,5 @@ public string Name
         /// Argument value.
         /// </summary>
         public object Value { get; set; }
-
-        internal IGraphType ResolvedType { get; set; } //TODO: possible remove after AstFromValue and AstPrinter rework
```
sungam3r, line 39: Finally! See `__DirectiveArgument.cs` and "climbing up" to the tree.

src/GraphQL.Tests/Files/JSON/AppliedDirectivesResult.json
```diff
@@ -8,7 +8,7 @@
           "args": [
             {
               "name": "quality",
-              "value": "high"
+              "value": "\"high\""
```
Shane32, line 11: Are we sure this is correct?  I assume this should match the introspection DefaultValue for input field types and similar.  Does it?  I see the spec says:

> defaultValue may return a String encoding (using the GraphQL language) of the default value used by this input value in the condition a value is not provided at runtime. If this input value has no default value, returns null.

Does a string need to be double encoded to meet this spec?

sungam3r: Yes. Code from `__InputValue`:
изображение
AstPrinter:
изображение

sungam3r: Consider default int value 42. In result introspection response it will be 
```json
"defaultValue": "42"
```

Now default value "abc":
```json
"defaultValue": "\"abc\""
```

It is GraphQL-formatted value, i.e. AST value. `"defaultValue": "abc"` will be wrong, it is some enum value. See AstPrinter:
изображение

sungam3r: In other words in JSON we always have outer quotes since value is always string. And in case of actual graphql string values we also have inner quotes.

sungam3r:
3 am here, I'm done for today.

Shane32 (approved):

PR #2407, opened 2021-03-17
Add ValidationError constructor

Shane32:


sungam3r:
@Shane32 I see conflict and malformed XML comments.

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -14,6 +14,12 @@ public class ValidationError : DocumentError
     {
         private readonly List<INode> _nodes = new List<INode>();
 
+        /// <inheritdoc cref="ValidationError(string, string, string, INode[])">
```
Shane32, line 17:
```suggestion
        /// <inheritdoc cref="ValidationError(string, string, string, INode[])"/>
```

src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>4.0.0-preview</VersionPrefix>
+    <VersionPrefix>4.1.0-preview</VersionPrefix>
```
Shane32, line 4:
```suggestion
    <VersionPrefix>4.0.2-preview</VersionPrefix>
```

src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>4.0.0-preview</VersionPrefix>
+    <VersionPrefix>4.0.2-preview</VersionPrefix>
```
sungam3r, line 4: 🤔 I see 4.0.2 in master already.

Shane32: Correct; I just set it to 4.0.2 to get rid of the merge conflict.  But wouldn't this go in 4.1.0 anyway?  Probably won't merge it quite yet in case there's any other changes we want in 4.0.2, such as #2416

Shane32: If this can go in 4.0.2 then it can be merged as-is.

sungam3r: My question is why I see this diff at all.

Shane32: Because it isn't merged with master.  GitHub is allowing it because the files are now identical, so it can be merged automatically without creating a conflict.

sungam3r: Sorry, but I don't understand. I see `4.0.2-preview` in master. Your PR targets master as well. So why I see this:
изображение

sungam3r (approved):

src/Directory.Build.props
Shane32: Because this branch was "branched" off of the master branch when it said 4.0.0-preview.  GitHub is showing changes that I've made to this branch as it compares to when it was "branched", not as compared to what it is now.  Does that make sense?

sungam3r: 🤦‍♂️ Yes, thanks!

PR #2408, opened 2021-03-17
Fix to serialize ExecutionResult from subscriptions

sungam3r:
Fixes `Subscribe_and_mutate_messages` test for https://github.com/graphql-dotnet/server/pull/518 . The error is that the data from the subscription comes with empty body. 
изображение

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


sungam3r:
@Shane32 I would like to merge this ASAP.

sungam3r:
I fixed all (except one) failed tests from https://github.com/graphql-dotnet/server/pull/518 . The last failed test showed that `Executed` property from subscription data stream was `false`.

PR #2412, opened 2021-03-18
Add IServiceProvider wrapper to auto-create instances

Shane32:
This feature allows the schema to create the instances of the graph types without them needing to be registered in the DI provider.  Since `SchemaTypes` only ever requests a given type be created once (and thereafter uses the cached instance), there's no real reason to have to register each type within your service provider.  They are all required to be registered, and you don't have a choice as to lifetime:

- With a singleton schema, you can register the types as any lifetime, but regardless of the lifetime, the types will only be requested once when the schema is initialized.

- With a scoped schema, you MUST register the types as scoped/transient or you will run into concurrency issues such as field middleware rewriting a field resolver multiple times.  Registering the types as scoped or transient will make no difference, as the types are only requested once per constructed schema.

Personally, I would recommend that this be the default behavior for `SchemaTypes` in v5 -- however it requires us to copy code from the MS DI Abstractions library, or import the package reference.  The code will work for any underlying DI provider as the underlying implementation is only dependent on `IServiceProvider`.  But whether we want to make that change or not for v5, I would like to add this feature to v4.1.


Shane32:
Note that the surface area of this PR is extremely small.  One new class with a single method, which contains two lines of executable code.  Mostly what we have here is documentation (and tests).

Shane32:
Also note that this similar to how ASP.NET Core creates MVC controllers; you don't need to register each of them individually.  So here within GraphQL.NET this PR allows it to work similarly -- the fact that the graph type was referenced within the schema is enough to dictate that the type should be instantiated.

sungam3r:
I will review this in the next days.

src/GraphQL.MicrosoftDI/AutoCreatingServiceProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Creates a new instance with the specified underlying service provider.
+        /// </summary>
+        public AutoCreatingServiceProvider(IServiceProvider serviceProvider)
+        {
+            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+        }
+
+        /// <inheritdoc cref="ActivatorUtilities.GetServiceOrCreateInstance(IServiceProvider, Type)"/>
+        public object GetService(Type serviceType) => ActivatorUtilities.GetServiceOrCreateInstance(_serviceProvider, serviceType);
```
sungam3r, line 36: From time to time I meet this method. I understand what it is doing. Its name speaks for itself :). But I was always interested in another question - what happens in the event of creating `IDisposable` instances. I got the impression that they were not disposed. Now I once again looked at the source code and my assumption seems true, although I could miss something 🤔 . Please write a test with this provider, which checks the fact of calling the `Dispose` method when destroying the container. If this method calls, it means that I carefully study the source code. Otherwise, it will be necessary to clearly note in the documentation that the types unregistered in the container should not implement the `IDisposable` interface (more precisely, they can do it, but the method will not be called).

sungam3r: Yep. I could not resist and did the test.
```csharp
    public class Repository : IDisposable
        {
            public Repository()

            {

            }
            public void Dispose()
            {
                Console.WriteLine("Called");
            }
        }

        public class My : IDisposable
        {
            public My(Repository r)
            {

            }

            public void Dispose()
            {
                Console.WriteLine("Not called");
            }
        }

 [Fact]
        public void aaa()
        {
            var services = new ServiceCollection();
            services.AddSingleton(typeof(Repository));
            var provider = services.BuildServiceProvider();
            var inst = ActivatorUtilities.GetServiceOrCreateInstance(provider, typeof(My));

            provider.Dispose();
```

> there's no real reason to have to register each type within your service provider. 

**IF** this type does not implement `IDisposable`. This is a weighty remark.

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
```diff
@@ -0,0 +1,38 @@
...
+using GraphQL.StarWars.Types;
+using Moq;
+using Xunit;
+
+namespace GraphQL.Tests.Types.Collections
+{
+    public class SchemaTypesTests
+    {
+        [Fact]
+        public void does_not_request_instance_more_than_once()
```
sungam3r, line 13: This test checks the current behavior and does not test a new feature at all. Right?

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
```diff
@@ -1,5 +1,10 @@
 namespace GraphQL.MicrosoftDI
 {
+    public class AutoCreatingServiceProvider : System.IServiceProvider
```
sungam3r, line 3: ActivatorUtilitiesServiceProvider 🤔  ?

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -65,6 +65,35 @@ public void ConfigureServices(IServiceCollection services)
 }
 ```
 
+To avoid having to register all of the individual graph types in your project, you can
```
sungam3r, line 68: Do you know about [AddGraphTypes](https://github.com/graphql-dotnet/server/blob/319764047fcc248d5e4055f43b06ecfb2d79e972/src/Core/GraphQLBuilderExtensions.cs#L80) ? One line of code and all types are registered. It will be logical to tell here about this method, if we stopped limited by the functionality of the main package.

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
Shane32: Correct. Obviously we would not want multiple instances for a single schema, but already SchemaTypes only requests a single instance per type.

docs2/site/docs/getting-started/dependency-injection.md
sungam3r: And... I would like to move `IGraphQLBuilder` into `GraphQL.MicrosoftDI` 😄 . It will be right. Previously there was no necessary project, where this interface would take its place.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
Shane32: AutoCreatingServiceProvider doesn’t sound that great but ActivatorUtilitiesServiceProvider doesn’t even say what it does. Maybe SelfActivatingServiceProvider.

docs2/site/docs/getting-started/dependency-injection.md
sungam3r:
```csharp
services.AddGraphTypes(); //  Add all types that implement IGraphType in the calling assembly
services.AddGraphTypes(typeof(MySchema)); // Add all types that implement IGraphType in the assembly which specified type belongs to
```

Shane32: Isn’t that in the server project? I don’t use the server project and usually have to rewrite similar code in all my projects. I also automatically add CLR type mappings for all automatically registered types. Does server do that? The issue is that some types must not be mapped so I have had to create a `DoNotRegister` attribute to prevent mapping it.

Shane32: We certainly can mention it in the docs though

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
sungam3r: SelfActivatingServiceProvider sounds a bit better than AutoCreatingServiceProvider.

docs2/site/docs/getting-started/dependency-injection.md
sungam3r: I do not want to create a situation when we write the same code in different places. I use the server project. I know that it has a lot of code to fix/refactor. Part of the code is generally worth transferring to another place. Now I see a good reason for this.

sungam3r: > I also automatically add CLR type mappings for all automatically registered types. Does server do that?

It scans assembly for types implementing `IGraphType`. It `can` do more. Let's add missing features. I suggest to move all these classes/methods not specific to ASP.NET Core into GraphQL.MicrosoftDI.

>  The issue is that some types must not be mapped so I have had to create a DoNotRegister attribute to prevent mapping it.

Agree. It's easy to add too.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
Shane32: Maybe `GetOrCreateServiceProvider`

src/GraphQL.MicrosoftDI/AutoCreatingServiceProvider.cs
Shane32: Yeah; all the dependencies will be properly disposed.

docs2/site/docs/getting-started/dependency-injection.md
Shane32: 👍  I can work on it.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
sungam3r: Any name will fit, let's go on.

docs2/site/docs/getting-started/dependency-injection.md
sungam3r: It will be a breaking change to the server project and a new feature for main project (4.1/4.2). I'm OK with fast major version changes in the server project.

Shane32:
More notes added @sungam3r .  Please review,

src/GraphQL.MicrosoftDI/AutoCreatingServiceProvider.cs
Shane32: I added documentation.  I don't understand what kind of test you would add.  One that demonstrates that desired behavior does NOT occur? There no failure condition.  Either it does not occur, which we know about, or it does which we would want. I would not add a `[Fact(Skip = true)]` test either, because this is expected behavior.

docs2/site/docs/getting-started/dependency-injection.md
Shane32: > 👍 I can work on it.

Well, I probably won't be.  However, I do agree that `IGraphQLBuilder` and its related methods should be moved to this repo, if you want to work on it.  If we do move it, the new methods should be reviewed and properly documented, and I just don't have the time for it right now.  Adding automatic CLR type mappings will actually be a separate feature, because it needs to operate on a schema instance rather than on a service collection.  And so it actually might not belong in the GraphQL.MicrosoftDI package at all.  Maybe I will do that part if I have time...

sungam3r: Ok

src/GraphQL.MicrosoftDI/AutoCreatingServiceProvider.cs
sungam3r: > One that demonstrates that desired behavior does NOT occur?

Yes. This must be at least in order to protect against unintentional changes caused by either by our code or by .NET runtime code.

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -65,6 +65,45 @@ public void ConfigureServices(IServiceCollection services)
...
+        Query = serviceProvider.GetRequiredService<StarWarsQuery>();
+        Mutation = serviceProvider.GetRequiredService<StarWarsMutation>();
+    }
+}
+```
+
+No other graph types will need to be registered. Graph types will only be instantiated once, during schema initialization
+as usual. Graph types can also pull in any services registered with dependency injection as usual.
+
+Note that if any of the graph types directly implement `IDisposable`, be sure to register those types with your dependency
```
sungam3r, line 97:
```suggestion
Note that if any of the graph types directly or indirectly implement `IDisposable`, be sure to register those types with your dependency
```

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -65,6 +65,45 @@ public void ConfigureServices(IServiceCollection services)
...
+injection provider, or their `Dispose` methods will not be called. Any dependencies of graph types that implement
+`IDisposable` will be disposed of properly, regardless of whether the graph type is registered within the service provider.
+
+You can also use the `services.AddGraphTypes()` extension method within the [server](https://github.com/graphql-dotnet/server)
```
sungam3r, line 101:
```suggestion
You can also use the `services.AddGraphTypes()` extension method from the [server](https://github.com/graphql-dotnet/server)
```

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -65,6 +65,45 @@ public void ConfigureServices(IServiceCollection services)
...
+project to scan the calling assembly for classes that implement `IGraphType` and register them all as singletons
+within the service provider. For additional options, see
```
sungam3r, line 103:
```suggestion
within the service provider. For additional options and overloads of this method, see
```

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -65,6 +65,45 @@ public void ConfigureServices(IServiceCollection services)
...
+[GraphQLBuilderCoreExtensions.cs](https://github.com/graphql-dotnet/server/blob/master/src/Core/Extensions/GraphQLBuilderCoreExtensions.cs)
+for additional overloads of this method.
```
sungam3r, line 105:
```suggestion
[GraphQLBuilderCoreExtensions.cs](https://github.com/graphql-dotnet/server/blob/master/src/Core/Extensions/GraphQLBuilderCoreExtensions.cs).
```

src/GraphQL.MicrosoftDI.Tests/AutoCreatingServiceProviderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.MicrosoftDI.Tests
+{
+    public class AutoCreatingServiceProviderTests
+    {
+        private readonly IServiceProvider _scopedServiceProvider1;
+        private readonly IServiceProvider _scopedServiceProvider2;
+        private readonly IServiceProvider _autoCreatingServiceProvider2;
```
sungam3r, line 12:
```suggestion
        private readonly IServiceProvider _selfActivatingServiceProvider2;
```

src/GraphQL.MicrosoftDI.Tests/AutoCreatingServiceProviderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public AutoCreatingServiceProviderTests()
+        {
+            var services = new ServiceCollection();
+            services.AddSingleton<TestSingleton>();
+            services.AddScoped<TestScoped>();
+            var rootServiceProvider = services.BuildServiceProvider();
+            _scopedServiceProvider1 = rootServiceProvider.CreateScope().ServiceProvider;
+            _scopedServiceProvider2 = rootServiceProvider.CreateScope().ServiceProvider;
+            _autoCreatingServiceProvider2 = new SelfActivatingServiceProvider(_scopedServiceProvider2);
```
sungam3r, line 22:
```suggestion
            _selfActivatingServiceProvider2 = new SelfActivatingServiceProvider(_scopedServiceProvider2);
```

src/GraphQL.MicrosoftDI.Tests/AutoCreatingServiceProviderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        [Fact]
+        public void prefers_pulling_from_service_provider()
+        {
+            var obj1 = _scopedServiceProvider1.GetRequiredService<TestSingleton>();
+            var obj2 = _autoCreatingServiceProvider2.GetRequiredService<TestSingleton>();
```
sungam3r, line 29:
```suggestion
            var obj2 = _selfActivatingServiceProvider2.GetRequiredService<TestSingleton>();
```

src/GraphQL.MicrosoftDI.Tests/AutoCreatingServiceProviderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            obj1.ShouldBe(obj2);
+        }
+
+        [Fact]
+        public void pulls_from_scoped_service_provider()
+        {
+            var obj1 = _scopedServiceProvider1.GetRequiredService<TestScoped>();
+            var obj2 = _scopedServiceProvider2.GetRequiredService<TestScoped>();
+            var obj3 = _autoCreatingServiceProvider2.GetRequiredService<TestScoped>();
```
sungam3r, line 38:
```suggestion
            var obj3 = _selfActivatingServiceProvider2.GetRequiredService<TestScoped>();
```

src/GraphQL.MicrosoftDI.Tests/AutoCreatingServiceProviderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            obj1.ShouldNotBe(obj2);
+            obj2.ShouldBe(obj3);
+        }
+
+        [Fact]
+        public void creates_when_not_registered()
+        {
+            _autoCreatingServiceProvider2.GetRequiredService<TestUnregistered>();
```
sungam3r, line 46:
```suggestion
            _selfActivatingServiceProvider2.GetRequiredService<TestUnregistered>();
```

src/GraphQL.MicrosoftDI.Tests/AutoCreatingServiceProviderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        [Fact]
+        public void creates_when_not_registered_with_dependencies()
+        {
+            _autoCreatingServiceProvider2.GetRequiredService<TestUnregisteredWithDependencies>();
```
sungam3r, line 52:
```suggestion
            _selfActivatingServiceProvider2.GetRequiredService<TestUnregisteredWithDependencies>();
```

src/GraphQL.MicrosoftDI.Tests/AutoCreatingServiceProviderTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        [Fact]
+        public void fails_with_missing_dependencies()
+        {
+            Should.Throw<InvalidOperationException>(() => _autoCreatingServiceProvider2.GetRequiredService<TestUnregisteredWithMissingDependencies>());
```
sungam3r, line 58:
```suggestion
            Should.Throw<InvalidOperationException>(() => _selfActivatingServiceProvider2.GetRequiredService<TestUnregisteredWithMissingDependencies>());
```

src/GraphQL.MicrosoftDI/SelfActivatingServiceProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// To use this, simply register the <see cref="ISchema"/> itself as follows:
+    /// <br/><br/>
+    /// <code>services.AddSingleton&lt;ISchema, MySchema&gt;(services => new MySchema(new SelfActivatingServiceProvider(services)));</code>
+    /// <br/><br/>
+    /// Within your <see cref="Schema"/> constructor, you may need to set your <see cref="Schema.Query"/>, <see cref="Schema.Mutation"/> and
+    /// <see cref="Schema.Subscription"/> fields to pull from the <see cref="SelfActivatingServiceProvider"/>:
+    /// <br/><br/>
+    /// <code>Query = services.GetRequiredService&lt;MyQuery&gt;();</code>
+    /// <br/><br/>
+    /// None of the graph types will need to be registered.
```
sungam3r, line 21:
```suggestion
    /// None of the graph types will need to be registered. Note that if any of the graph types implement
    /// IDisposable, be sure to register those types with your dependency injection provider, or their Dispose
    /// methods will not be called. Any dependencies of graph types that implement IDisposable will be
    /// disposed of properly, regardless of whether the graph type is registered within the service provider.
```

sungam3r (approved):

PR #2416, opened 2021-03-18
Fix to SchemaTypes.RebuildType

### LINKED ISSUE #2415: Can't resolve generic types that use GraphQL clr reference types
My schema builder constructed the following graph type:
- `ConnectionType<GraphQLClrOutputTypeReference<Widget>>`

(Note: `ConnectionType<>` is provided with GraphQL.)

However, during schema initialization, the "wrapped" reference type was not resolved, since it was not recognized as a GraphQL reference type.

This breaks my use case.  I will provide a fix within the library as a PR in which it will examining each generic graph type to determine if it needs to be reconstructed.  I will also (eventually) provide a PR to `SchemaTypes` that has enough protected members to allow overriding portions of it as necessary to fix unforeseen problems like this in the future.

Shane32:
Fixes #2415 

@sungam3r I would like to merge this asap.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>4.0.1-preview</VersionPrefix>
+    <VersionPrefix>4.0.2-preview</VersionPrefix>
     <LangVersion>8.0</LangVersion>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageIcon>logo.64x64.png</PackageIcon>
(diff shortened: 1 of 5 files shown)


sungam3r (approved):
Good fix. I did not think of this obvious thing when I wrote this method. One suggestion. At the beginning of `RebuildType` I would like to add additional check like this
```csharp
if (type.IsGenericTypeDefinition)
  throw new InvalidOperationException($"GraphType {type} should not be an open generic type definition");
```

To test this change `test` field to
```csharp
 var field = query.Field(typeof(ConnectionType<>), "test");
```

Or even change `FieldType.Type.set` to throw if `value` is generic type definition.

sungam3r:
And remember to bump VersionPrefix to 4.0.2

Shane32:
I won't release 4.0.2 just yet

sungam3r:
OK

Shane32:
@sungam3r I ran into an issue in my code where an enumeration type is not resolved by the type mapper.  I need a little more time to investigate.  It may be related to this bug here, .... still checking.  It's also likely enough that I need to fix something else in my own code.

PR #2419, opened 2021-03-19
Delete UseRuntimeTypeMappings property

Shane32:
I'm not sure how this got set to false.  This should be fixed **_immediately_**.

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -160,7 +160,6 @@ namespace GraphQL
         public static bool EnableReadDeprecationReasonFromAttributes { get; set; }
         public static bool EnableReadDescriptionFromAttributes { get; set; }
         public static bool EnableReadDescriptionFromXmlDocumentation { get; set; }
-        public static bool UseRuntimeTypeMappings { get; set; }
     }
     [System.AttributeUsage(System.AttributeTargets.Class | System.AttributeTargets.Method | System.AttributeTargets.Property | System.AttributeTargets.All, AllowMultiple=true)]
     public abstract class GraphQLAttribute : System.Attribute
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -251,16 +251,6 @@ namespace GraphQL.Tests.Types
         }
 
         [Fact]
-        public void throws_when_type_is_not_inferable()
-        {
-            var type = new ComplexType<TestObject>();
-
-            var exp = Should.Throw<ArgumentException>(() => type.Field(d => d.valuePair));
-
-            exp.Message.ShouldStartWith("The GraphQL type for field 'TestObject.valuePair' could not be derived implicitly from expression 'd => d.valuePair'.");
-        }
-
-        [Fact]
         public void throws_when_type_is_incompatible()
         {
             var type = new ComplexType<TestObject>();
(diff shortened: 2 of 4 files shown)


Shane32:
I must have had the impression that it would be set to true to "improve" performance.

Shane32:
Also, I just did a full benchmark.  Setting the property to false/true has absolutely no effect on performance.

Shane32:
The switch isn't mentioned anywhere in the migration documentation about `RegisterTypeMapping`.  It isn't mentioned in our documentation for custom scalars, which also mentions `RegisterTypeMapping`.  It isn't even listed along with the rest of the global switches in the migration document.

Shane32:
Pretty sure we should be setting this back to `true`.  A number of the field builders won't work at all with it set to false.  The only "benefit" this property has is to discourage the use of expressions in field builders, which only indirectly makes a practical difference  for schemas with a scoped lifetime, which is already strongly discouraged throughout the document.  It would be more appropriate to mention with the documentation about scoped schema lifetime that setting this to false can help performance.  But if the user was concerned about performance, they probably wouldn't have a scoped schema lifetime to begin with.

Shane32:
Note that the xml comment is wrong here as well.  It says "if you are using a scoped schema with many field expressions, then this setting will help speed up the initialization of your schema when set to false".  This is inaccurate.  It only helps speed up building the schema by the nature of preventing the use of field expressions.  It cannot speed up building the schema if you HAVE field expressions, as the slow part is compiling the expression, not converting a type reference.

Shane32:
Finally, having the `[GraphQLMetadataAttribute]` attribute marked on a type does not allow an field builder via expression syntax to run any faster than using the type mappings, since it still needs to compile the expression.

sungam3r:
When `UseRuntimeTypeMappings` is `true` then every _primitive_ type like `int`, `string`, `DateTime`, etc. is wrapped into `GraphQLClrOutputTypeReference<>`/`GraphQLClrInputTypeReference<>` and then rebuilded instead of directly returning graphtype from `SchemaTypes.BuiltInScalarMappings`.

Shane32:
It is fortunate that this is not a breaking change.

sungam3r:
> Also, I just did a full benchmark. Setting the property to false/true has absolutely no effect on performance.

It depends on your scenario.

Shane32:
> > Also, I just did a full benchmark. Setting the property to false/true has absolutely no effect on performance.
> 
> It depends on your scenario.

No, it doesn't.  Give me any scenario.  It cannot improve performance.  All it can do is cause exceptions when attempting to use the field builder with expression syntax.

Shane32:
> When `UseRuntimeTypeMappings` is `true` then every _primitive_ type like `int`, `string`, `DateTime`, etc. is wrapped into `GraphQLClrOutputTypeReference<>`/`GraphQLClrInputTypeReference<>` and then rebuilded instead of directly returning graphtype from `SchemaTypes.BuiltInScalarMappings`.

Ok I'll grant you that, but only if you're using an expression syntax.  And still it won't have any appreciable effect on build performance.  Compiling the expression syntax far outweighs any effect that above would have.

sungam3r:
> as the slow part is compiling the expression, not converting a type reference.

Then let's do it simple - remove this property at all.

sungam3r:
> Ok I'll grant you that, but only if you're using an expression syntax. And still it won't have any appreciable effect on build performance. Compiling the expression syntax far outweighs any effect that above would have.

OK, I'm OK to remove this property. No need in bump major version. Nobody uses this property yet.

sungam3r (approved):

PR #2421, opened 2021-03-19
Fix GetGraphTypeFromGraphType and enumerable mappings

Shane32:


src/GraphQL.Tests/Utilities/GetGraphTypeFromTypeTests.cs
```diff
@@ -75,5 +77,127 @@ public void supports_uint_type()
...
+        [InlineData(typeof(Uri), false, TypeMappingMode.OutputType, typeof(NonNullGraphType<GraphQLClrOutputTypeReference<Uri>>))]
+        [InlineData(typeof(MyClass), false, TypeMappingMode.OutputType, typeof(NonNullGraphType<GraphQLClrOutputTypeReference<MyClass>>))]
+        [InlineData(typeof(MyEnum), false, TypeMappingMode.OutputType, typeof(NonNullGraphType<GraphQLClrOutputTypeReference<MyEnum>>))]
+        //output mapping mode - nullable structs
+        [InlineData(typeof(int?), false, TypeMappingMode.OutputType, null)]
+        [InlineData(typeof(decimal?), false, TypeMappingMode.OutputType, null)]
+        [InlineData(typeof(Guid?), false, TypeMappingMode.OutputType, null)]
+        [InlineData(typeof(MyEnum?), false, TypeMappingMode.OutputType, null)]
+        //output mapping mode - various types of lists
+        [InlineData(typeof(IEnumerable), true, TypeMappingMode.OutputType, typeof(ListGraphType<NonNullGraphType<GraphQLClrOutputTypeReference<object>>>))]
```
Shane32, line 166: Does this even make sense?  See below.

src/GraphQL/TypeExtensions.cs
```diff
@@ -206,9 +211,7 @@ internal static string GetFriendlyName(this Type type)
         /// </summary>
         public static Type GetEnumerableElementType(this Type type)
```
Shane32, line 212: This is only ever used by `GetTypeFromType`

src/GraphQL/TypeExtensions.cs
```diff
@@ -218,11 +221,23 @@ public static Type GetEnumerableElementType(this Type type)
...
+            {
+                if (iface.IsGenericType && iface.GetGenericTypeDefinition() == typeof(IEnumerable<>))
+                {
+                    return iface.GenericTypeArguments[0];
+                }
+            }
+
+            // see if the type supports IEnumerable
+            if (typeof(IEnumerable).IsAssignableFrom(type))
+                return typeof(object);
```
Shane32, line 236: If it doesn't make sense to create a `GraphQlClrOutputTypeMapping` for an `object` type, then this code can be removed.  Except it's public, and it would make sense for it to return `object` here -- just not for its use.  So maybe the whole method should be private.  Maybe mark as obsolete now, and make private in v5.  But for now I left it as-is (except expanded the support for `IEnumerable<T>` interfaces).

src/Tests.props
```diff
@@ -2,7 +2,7 @@
   <Import Project="$(MSBuildThisFileDirectory)Tests.local.props" Condition="Exists('$(MSBuildThisFileDirectory)Tests.local.props')" />
 
   <PropertyGroup>
-    <NoWarn>$(NoWarn);1591;IDE1006</NoWarn>
+    <NoWarn>$(NoWarn);1591;IDE1006;CS0618</NoWarn>
```
Shane32, line 5: This ignores Obsolete warnings in tests.  Ok?

sungam3r: ok

src/GraphQL/TypeExtensions.cs
```diff
@@ -37,6 +37,7 @@ public static bool IsConcrete(this Type type)
         /// <returns>
         ///   <c>true</c> if the specified type is a subclass of Nullable&lt;T&gt;; otherwise, <c>false</c>.
         /// </returns>
+        [Obsolete("This member will be removed in GraphQL v5.")]
```
sungam3r, line 40:
```suggestion
        [Obsolete("This member will be removed in GraphQL.NET v5.")]
```

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -400,6 +400,7 @@ namespace GraphQL
         public static bool ImplementsGenericType(this System.Type type, System.Type genericType) { }
         public static bool IsConcrete(this System.Type type) { }
         public static bool IsGraphType(this System.Type type) { }
+        [System.Obsolete("This member will be removed in GraphQL v5.")]
```
sungam3r, line 403:
```suggestion
        [System.Obsolete("This member will be removed in GraphQL.NET v5.")]
```

src/GraphQL/TypeExtensions.cs
sungam3r: Obsolete?

src/GraphQL.Tests/Utilities/GetGraphTypeFromTypeTests.cs
```diff
@@ -75,5 +77,127 @@ public void supports_uint_type()
...
+        //built-in mapping mode - not nullable types
+        [InlineData(typeof(int), false, TypeMappingMode.UseBuiltInScalarMappings, typeof(NonNullGraphType<IntGraphType>))]
+        [InlineData(typeof(string), false, TypeMappingMode.UseBuiltInScalarMappings, typeof(NonNullGraphType<StringGraphType>))]
+        [InlineData(typeof(decimal), false, TypeMappingMode.UseBuiltInScalarMappings, typeof(NonNullGraphType<DecimalGraphType>))]
+        [InlineData(typeof(Guid), false, TypeMappingMode.UseBuiltInScalarMappings, typeof(NonNullGraphType<IdGraphType>))]
+        [InlineData(typeof(Uri), false, TypeMappingMode.UseBuiltInScalarMappings, typeof(NonNullGraphType<UriGraphType>))]
+        [InlineData(typeof(MyClass), false, TypeMappingMode.UseBuiltInScalarMappings, null)]
+        [InlineData(typeof(MyEnum), false, TypeMappingMode.UseBuiltInScalarMappings, typeof(NonNullGraphType<EnumerationGraphType<MyEnum>>))]
+        //built-in mapping mode - nullable structs
+        [InlineData(typeof(int?), false, TypeMappingMode.UseBuiltInScalarMappings, null)]
```
sungam3r, line 118: These combinations look strange, but even if we add NRT feature, it will not automatically mean that we can remove `nullable` parameter, because one can have cases when some `int` property **should** map to nullable IntGraphType.

src/GraphQL.Tests/Utilities/GetGraphTypeFromTypeTests.cs
```diff
@@ -75,5 +77,127 @@ public void supports_uint_type()
...
+        [InlineData(typeof(decimal?), false, TypeMappingMode.UseBuiltInScalarMappings, null)]
+        [InlineData(typeof(Guid?), false, TypeMappingMode.UseBuiltInScalarMappings, null)]
+        [InlineData(typeof(MyEnum?), false, TypeMappingMode.UseBuiltInScalarMappings, null)]
+        //built-in mapping mode - various types of lists
+        [InlineData(typeof(IEnumerable), true, TypeMappingMode.UseBuiltInScalarMappings, null)]
+        [InlineData(typeof(IEnumerable<int>), true, TypeMappingMode.UseBuiltInScalarMappings, typeof(ListGraphType<NonNullGraphType<IntGraphType>>))]
+        [InlineData(typeof(IEnumerable<string>), true, TypeMappingMode.UseBuiltInScalarMappings, typeof(ListGraphType<StringGraphType>))]
```
sungam3r, line 125: изображение

Indeed not good... The goal to fix in v5.

src/GraphQL.Tests/Utilities/GetGraphTypeFromTypeTests.cs
```diff
@@ -75,5 +77,127 @@ public void supports_uint_type()
...
+        [InlineData(typeof(Guid?), false, TypeMappingMode.OutputType, null)]
+        [InlineData(typeof(MyEnum?), false, TypeMappingMode.OutputType, null)]
+        //output mapping mode - various types of lists
+        [InlineData(typeof(IEnumerable), true, TypeMappingMode.OutputType, typeof(ListGraphType<NonNullGraphType<GraphQLClrOutputTypeReference<object>>>))]
+        [InlineData(typeof(IEnumerable<int>), true, TypeMappingMode.OutputType, typeof(ListGraphType<NonNullGraphType<GraphQLClrOutputTypeReference<int>>>))]
+        [InlineData(typeof(IEnumerable<string>), true, TypeMappingMode.OutputType, typeof(ListGraphType<GraphQLClrOutputTypeReference<string>>))]
+        [InlineData(typeof(IEnumerable<decimal>), true, TypeMappingMode.OutputType, typeof(ListGraphType<NonNullGraphType<GraphQLClrOutputTypeReference<decimal>>>))]
+        [InlineData(typeof(IEnumerable<Guid>), true, TypeMappingMode.OutputType, typeof(ListGraphType<NonNullGraphType<GraphQLClrOutputTypeReference<Guid>>>))]
+        [InlineData(typeof(IEnumerable<Uri>), true, TypeMappingMode.OutputType, typeof(ListGraphType<NonNullGraphType<GraphQLClrOutputTypeReference<Uri>>>))]
```
sungam3r, line 171: 🤔 `Uri` is class so like `string` it should be nullable.

sungam3r: And the case with `MyClass` too.

sungam3r: >             // note: when parsing lists of class types (including lists of lists), the inner type is treated as non-nullable, except for strings


Yes, I remember...

src/GraphQL.Tests/Utilities/GetGraphTypeFromTypeTests.cs
sungam3r: 50/50. Let's keep it.

src/GraphQL/TypeExtensions.cs
```diff
@@ -137,32 +140,34 @@ public static Type GetGraphTypeFromType(this Type type, bool isNullable = false,
                     if (mode == TypeMappingMode.UseBuiltInScalarMappings)
                     {
                         SchemaTypes.BuiltInScalarMappings.TryGetValue(type, out graphType);
+
+                        if (graphType == null)
```
sungam3r, line 144:
```suggestion
                        if (!SchemaTypes.BuiltInScalarMappings.TryGetValue(type, out graphType))
```

src/GraphQL/TypeExtensions.cs
```diff
@@ -137,32 +140,34 @@ public static Type GetGraphTypeFromType(this Type type, bool isNullable = false,
...
-            }
-
             if (!isNullable)
             {
                 graphType = typeof(NonNullGraphType<>).MakeGenericType(graphType);
             }
 
             return graphType;
+
+            static bool IsNullableType(Type type) => type == typeof(string) || type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>);
```
sungam3r, line 170:
```suggestion

            //TODO: rewrite nullability condition in v5
            static bool IsNullableType(Type type) => type == typeof(string) || type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>);
```

src/GraphQL/TypeExtensions.cs
sungam3r: Ok with Obsolete

src/GraphQL/TypeExtensions.cs
```diff
@@ -137,32 +140,34 @@ public static Type GetGraphTypeFromType(this Type type, bool isNullable = false,
                     if (mode == TypeMappingMode.UseBuiltInScalarMappings)
                     {
                         SchemaTypes.BuiltInScalarMappings.TryGetValue(type, out graphType);
+
+                        if (graphType == null)
+                        {
+                            if (type.IsEnum)
+                            {
+                                graphType = typeof(EnumerationGraphType<>).MakeGenericType(type);
```
sungam3r, line 148: So we consider enums as built-in types (scalars) as well. Worth to add a note about it somewhere in xml comments or docs.

src/GraphQL.Tests/Utilities/GetGraphTypeFromTypeTests.cs
Shane32: Ok so I'm deleting this note, and making string & MyClass so it will return a nullable list.  Correct?

sungam3r: Mmmmm... it would be a "semi-breaking" change to schema: `SomeType!` -> `SomeType` for both input and output cases.
Two options:
1. Start `develop` branch and put it there.
2. Consider this as non-breaking change and keep it here.

I tend to the second option.

sungam3r (approved):

PR #2424, opened 2021-03-19
Fix to constant case

Shane32:


Shane32:
@sungam3r Want to review it out now?

Shane32:
I can condense the `if` statements like this if you like it better:
```cs
public static string ToConstantCase(this string value)
{
    int i;
    int strLength = value.Length;
    // iterate through each character in the string, stopping a character short of the end
    for (i = 0; i < strLength - 1; ++i)
    {
        var curChar = value[i];
        var nextChar = value[i + 1];
                
        // look for the pattern [a-z][A-Z]
        if ((char.IsLower(curChar) && char.IsUpper(nextChar)) ||
            // or the pattern [0-9][A-Za-z]
            (char.IsDigit(curChar) && char.IsLetter(nextChar)) ||
            // or the pattern [A-Za-z][0-9]
            (char.IsLetter(curChar) && char.IsDigit(nextChar)) ||
            // or if there's enough characters left, look for the pattern [A-Z][A-Z][a-z]
            (i < strLength - 2 && char.IsUpper(curChar) && char.IsUpper(nextChar) && char.IsLower(value[i + 2])))
        {
            // add an underscore between the two characters, increment i to skip the underscore, and increase strLength because the string is longer now
            value = value.Substring(0, ++i) + '_' + value.Substring(i);
            ++strLength;
            // then skip the following match check since we already found a match here via the 'continue' statement
        }
    }
    // convert the resulting string to uppercase
    return value.ToUpperInvariant();
}
```

Shane32:
dotnet-format won't let me format it the other way.  Don't know why.

src/GraphQL/StringExtensions.cs
```diff
@@ -70,33 +70,47 @@ private static string FastChangeFirstLetter(char newFirstLetter, string s)
...
             }
             // convert the resulting string to uppercase
             return value.ToUpperInvariant();
+
+            void InsertUnderscore()
+            {
+                // add an underscore between the two characters, increment i to skip the underscore, and increase strLength because the string is longer now
+                value = value.Substring(0, ++i) + '_' + value.Substring(i);
+                ++strLength;
+                // then skip the following match check since we already found a match here via the 'continue' statement
```
sungam3r, line 112: This comment does not look here in its place.

src/GraphQL/StringExtensions.cs
```diff
@@ -70,33 +70,47 @@ private static string FastChangeFirstLetter(char newFirstLetter, string s)
...
+                    InsertUnderscore();
                     continue;
                 }
                 // if there's enough characters left, look for the pattern [A-Z][A-Z][a-z]
                 if (i < strLength - 2 && char.IsUpper(curChar) && char.IsUpper(nextChar) && char.IsLower(value[i + 2]))
                 {
-                    // add an underscore between the two characters, increment i to skip the underscore, and increase strLength because the string is longer now
-                    value = value.Substring(0, ++i) + '_' + value.Substring(i);
-                    ++strLength;
+                    InsertUnderscore();
```
sungam3r, line 101:
```suggestion
                    InsertUnderscore();
                    continue;
```

to unify all cases ?

src/GraphQL/StringExtensions.cs
```diff
@@ -70,33 +70,47 @@ private static string FastChangeFirstLetter(char newFirstLetter, string s)
         /// </summary>
         public static string ToConstantCase(this string value)
```
sungam3r, line 71:
```suggestion
        public static string ToConstantCase(this string value) //TODO: rewrite to stackalloc/ string.Create()
```

sungam3r (approved):

Issue #2422, opened 2021-03-19
NU1102 error on windows-latest


sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/pull/2421/checks?check_run_id=2150854275 

> D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.Dummy\GraphQL.Dummy.csproj : error NU1102: Unable to find package NETStandard.Library with version (>= 2.0.3) [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.Dummy\GraphQL.Dummy.csproj : error NU1102:   - Found 1 version(s) in Microsoft Visual Studio Offline Packages [ Nearest version: 1.6.0 ] [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]
D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.Dummy\GraphQL.Dummy.csproj : error NU1102:   - Found 0 version(s) in Source [D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln]

sungam3r:
`Re-run all jobs` button in GitHub CI UI solves this problem.

Shane32:
I was experiencing the same issue and fix.  I don't think there's a problem with our scripts.

sungam3r:
Windows:
![изображение](https://user-images.githubusercontent.com/21261007/111885606-d76bb200-89d9-11eb-9a4f-f8d37641705f.png)

Ubuntu:
![изображение](https://user-images.githubusercontent.com/21261007/111885620-ec484580-89d9-11eb-9931-0ac054942055.png)

I think we may try to disable `Microsoft Visual Studio Offline Packages` source.

sungam3r:
Waiting 1-2 months after #2426 to see if it happens again.

Shane32:
If anything, it’s a CI issue. Unlikely it’s anything with our CI scripts. Certainly not a bug.

sungam3r:
Also found here https://github.com/graphql-dotnet/parser/pull/126/checks?check_run_id=2282874117

sungam3r:
Closing.

PR #2426, opened 2021-03-20
Disable MSVS Nuget feed on Windows build

sungam3r:
Helps to investigate #2422

### DIFF
--- a/.github/workflows/build-artifacts-code.yml
+++ b/.github/workflows/build-artifacts-code.yml
@@ -53,7 +53,7 @@ jobs:
         run: dotnet build --no-restore -c Release -p:VersionSuffix=$GITHUB_RUN_NUMBER
       - name: Pack solution [Release]
         working-directory: src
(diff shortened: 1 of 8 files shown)


sungam3r:
@Shane32 I suggest to merge PR and see what happens. If after one-twoo months error will not happen again, then I can remove unnecessary steps and keep only one step that disables MSVS feed.

Shane32:
Have you tested a failure yet? If not, I suggest simulating an error by replacing “dotnet restore” in the first step with “exit 1”. Be sure the workflow runs properly, continuing to execute after the workflow step fails, retrying to restore the packages. I’m not sure, but you may need another setting in there to make it continue. 

If you’ve tested it, sure I’m fine with some extra steps to see if it helps.

sungam3r:
Actually I think that retry step is not necessary at all because of removal of the problem feed.

sungam3r:
I removed not necessary steps. Remember to ping me if one of your builds fails again with NU1102.

PR #2427, opened 2021-03-21
Add scoped connection builders

### LINKED ISSUE #2425: GraphQL.MicrosoftDI not support Connection
Thank you for great work 👍 . I only think will be good when GraphQL.MicrosoftDI will support ConnectionBuilder as well

Shane32:
Fixes #2425


Shane32:
I'm not sure that any extra documentation is necessary.  The design for connection builders is the same as field builders, which is documented at https://graphql-dotnet.github.io/docs/getting-started/dependency-injection#thread-safety-with-scoped-services 

I did add `.Returns<T>()` to the `Resolve()` connection builder design, since this is not supported by the connection builders.  There was two ways to implement this -- where `Returns<T>()` would require resolvers to return a `Connection<T>` object, or where it would require resolvers to return a `T` object.  Of the latter option, I can add a `where T : Connection<T>` constraint or not.  I opted for the least constraints, since it seemed to me likely enough that people would define their own connection object, and then I added a bunch of xml comments.

We could go with the most restrictive option (where `.Returns<T>()` requires returning a `Connection<T>` object), and then if people want to use their own type, they simply must not call `.Returns<T>()` so that `object` can be returned from the resolver.  Another option would be to add another method `.NodeType<T>()` which simply calls `.Resolve<Connection<T>>()`.  Any thoughts here?

Finally, the existing `Connection(...).Resolve(...)` returns `void`.  This is probably due to the fact that `.Unidirectional()` must appear before `.Resolve(...)` or else it will build the resolver incorrectly.  So this design is carried over to the new connection builders, where the final `.Resolve(...)` returns `void`.  If we change this in the main project, we should change it here at the same time.

Shane32:
Again, even though there are a few different ways to implement `.Returns<T>()`, it probably does not need much explanation in the main documentation.  If it is coded incorrectly, the resolver will be requesting the wrong type, and it will be rather obvious how to fix it.

sungam3r:
Need 1-2 days to review.

src/GraphQL.MicrosoftDI/ScopedResolveConnectionContextAdapter.cs
```diff
@@ -0,0 +1,78 @@
...
+using GraphQL.Language.AST;
+using GraphQL.Types;
+
+namespace GraphQL.MicrosoftDI
+{
+    internal sealed class ScopedResolveConnectionContextAdapter<TSource> : IResolveConnectionContext<TSource>
+    {
+        private readonly IResolveConnectionContext<TSource> _baseContext;
+
+        public ScopedResolveConnectionContextAdapter(IResolveConnectionContext<TSource> baseContext, IServiceProvider serviceProvider)
```
sungam3r, line 16: Why `IResolveConnectionContext<TSource>` ? I see non-generic `IResolveFieldContext` in the corresponding `ScopedResolveFieldContextAdapter<TSource>`.

src/GraphQL.MicrosoftDI/ScopedConnectionBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <summary>
+    /// Extension methods for creating connection resolver builders.
+    /// </summary>
+    public static class ScopedConnectionBuilderExtensions
+    {
+        /// <summary>
+        /// Sets the resolver for the connection field. A dependency injection scope is created for the duration of the resolver's execution
+        /// and the scoped service provider is passed within <see cref="IResolveFieldContext.RequestServices"/>.
+        /// </summary>
+        public static void ResolveScoped<TSourceType, TReturnType>(this ConnectionBuilder<TSourceType> builder, Func<IResolveConnectionContext<TSourceType>, TReturnType> resolver)
```
sungam3r, line 17: For corresponding methods for FieldBuilder we have `ScopedFieldResolver` and `ScopedAsyncFieldResolver`. Why there are no such?

src/GraphQL.MicrosoftDI.Tests/ScopedContextBase.cs
```diff
@@ -27,10 +29,11 @@ public ScopedContextBase()
             _requestServicesMock = new Mock<IServiceProvider>(MockBehavior.Strict);
             _requestServicesMock.Setup(x => x.GetService(It.Is<Type>(t => t == typeof(IServiceScopeFactory)))).Returns(serviceScopeFactory).Verifiable();
             var requestServices = _requestServicesMock.Object;
-            _scopedContext = new ResolveFieldContext
+            _scopedContext = new ResolveFieldContext<object>
```
sungam3r, line 32: Why did you add generic parameter here?

src/GraphQL.MicrosoftDI/ConnectionResolverBuilder.cs
```diff
@@ -0,0 +1,405 @@
...
+        /// <summary>
+        /// Specifies the delegate to execute when the field is being resolved.
+        /// </summary>
+        public void Resolve(Func<IResolveConnectionContext<TSourceType>, TReturnType> resolver)
+        {
+            if (_scoped)
+                _builder.ResolveScoped(resolver);
+            else
+                _builder.Resolve(context => resolver(context));
+        }
```
sungam3r, line 83:
```suggestion
        => _scoped ? _builder.ResolveScoped(resolver) : _builder.Resolve(resolver);
```

src/GraphQL.MicrosoftDI/ConnectionResolverBuilder.cs
```diff
@@ -0,0 +1,405 @@
...
+
+        /// <inheritdoc cref="Resolve(Func{IResolveConnectionContext{TSourceType}, TReturnType})"/>
+        public void ResolveAsync(Func<IResolveConnectionContext<TSourceType>, Task<TReturnType>> resolver)
+        {
+            if (_scoped)
+                _builder.ResolveScopedAsync(resolver);
+            else
+                _builder.ResolveAsync(async context => await resolver(context));
+        }
```
sungam3r, line 92:
```suggestion
        => _scoped ? _builder.ResolveScopedAsync(resolver) : _builder.ResolveAsync(resolver);
```

sungam3r: The same suggestions for other builders in this file.

src/GraphQL.MicrosoftDI.Tests/ScopedContextBase.cs
Shane32: Not sure; perhaps just to match `_scopedConnectionContext`.  It works without the change so I reverted it.

src/GraphQL.MicrosoftDI/ConnectionResolverBuilder.cs
Shane32: image

Shane32: Also cannot reduce `.ResolveAsync(async context => await resolver(context))` to `.ResolveAsync(resolver)` because the `Resolve` method expects a `Task<object>`, not a `Task<T>`.

src/GraphQL.MicrosoftDI/ScopedConnectionBuilderExtensions.cs
Shane32: There did not seem to be a reason to add them.  Since `IFieldResolver.Resolve` only accepts `IResolveFieldContext`, the connection context would need to be built within the `ScopedConnectionResolver`.  And I can't access the private `PageSize` or `Unidirectional` properties of the builder to be able to build a new connection context.  And people cannot use `.Resolve(IFieldResolver)` on a connection; they would need to access the underlying `FieldType`.

And if someone actually does want to use their own `IFieldResolver` for a connection context, they can write this:

```cs
var resolver = new ScopedFieldResolver<int>(context =>
{
    var connectionContext = new GraphQL.Builders.ResolveConnectionContext<object>(context, isUnidirectional: false, defaultPageSize: 100);
    //code here
});
```
just about as easily as this:
```cs
var resolver = new ScopedConnectionResolver<int>(connectionContext =>
{
    //code here
}, isUnidirectional: false, defaultPageSize: 100);
```

In short, I felt this code is more for the connection builders rather than fields in general, and since it is not needed by the builders, I did not feel the need to add them.  I can add them if you think I should -- but again, they cannot be used by the connection builders without additional changes to the codebase.

src/GraphQL.MicrosoftDI/ScopedResolveConnectionContextAdapter.cs
Shane32: This class is `internal`, and can only ever be used in a typed context, so I simplified the code.

`ScopedResolveFieldContextAdapter` (also `internal`) is used by `ScopedFieldResolver` & `ScopedAsyncFieldResolver` and requires an untyped `IResolveFieldContext` as its input.

src/GraphQL.MicrosoftDI/ConnectionResolverBuilder.cs
Shane32: See other notes on this.  Same issues are present, where the return value of `.Resolve()` is `object` rather than `T`, so a lambda is necessary.

Note that if `.Returns<TNewReturnType>()` was removed from the builder, it would always accept an `object` and then it could be passed directly through.  I favored a typed context.  We could move the `.Returns<TNewReturnType>()` method to the connection builder in GraphQL.NET (to match the field builder's `.Return<TNewReturnType>()` already present there), but this will change a bunch of APIs in the main repo.  Not sure if they would be breaking or not; I can look into this now if you want me to, or we can just mark this as to-do for v5.

Shane32:
@sungam3r Are you waiting for me on anything here?  I thought I answered your questions, and there isn't really any changes to be made.

sungam3r:
I wanted to look again, but I was busy. Now I can't, because I want to sleep, sorry. Could it wait a little?  I promise to finish the review this weekend.

Shane32:
Sure no problem

sungam3r:
> Finally, the existing Connection(...).Resolve(...) returns void. This is probably due to the fact that .Unidirectional() must appear before .Resolve(...) **or else it will build the resolver incorrectly**. So this design is carried over to the new connection builders, where the final .Resolve(...) returns void. If we change this in the main project, we should change it here at the same time.

What do you mean? I think that `Resolve` should return reference to `this`, not `void`.

src/GraphQL.MicrosoftDI/ScopedResolveConnectionContextAdapter.cs
sungam3r: I promised to review this PR again. To do this, I once again looked at the code of existing builders. I don't use connection builder at my work so I can misunder about the needs. I believe that all builders must have the most similar design, including for generic constraints (or not having them). I made some changes and show them in another PR. Most likely, some of them will need to roll back.

src/GraphQL.MicrosoftDI/ScopedConnectionBuilderExtensions.cs
sungam3r: IMO
1. `ConnectionBuilder` should expose 3 properties that now are private fields: _isUnidirectional, _isBidirectional, _pageSize
2. `ResolveScoped` and `ResolveScopedAsync` should return `ConnectionBuilder<>`, not void.

src/GraphQL.MicrosoftDI/ConnectionResolverBuilder.cs
sungam3r: > We could move the .Returns<TNewReturnType>() method to the connection builder in GraphQL.NET (to match the field builder's .Return<TNewReturnType>() already present there)

I would have done so.

sungam3r:
> We could go with the most restrictive option (where .Returns<T>() requires returning a Connection<T> object), and then if people want to use their own type, they simply must not call .Returns<T>() so that object can be returned from the resolver. 

Even more... Why `ConnnectionBuilder` has only one generic constraint for type from `IResolveFieldContext.Source` ? `FieldBuilder` has the second constraint `TReturnType` for resolver's return value. I think that the second type wss missed from the initial design.

See my other comment. `ConnectionBuilder` need to be rewritten a little more. Yes, a bunch of breaking changes but if we change this place, then let's change it so that the design of all builders was the same.

Also I don't like implementation of `Unidirectional` and `Bidirectional` methods:
```
builder.Unidirectional().Bidirectional().Unidirectional().Bidirectional()
```
yields 8 arguments 😞 on field.

src/GraphQL.MicrosoftDI/ScopedResolveConnectionContextAdapter.cs
sungam3r: >  and can only ever be used in a typed context, so I simplified the code.

OK. I agree.

>  and requires an untyped IResolveFieldContext as its input.

See #2427 . I would like to get such a solution when the design all builders was the same.

Shane32:
1. _`ConnectionBuilder` should expose 3 properties that now are private fields: `_isUnidirectional`, `_isBidirectional`, `_pageSize`_

Builders usually don't have `get` properties, but ok...

2. _`ResolveScoped` and `ResolveScopedAsync` should return `ConnectionBuilder<>`, not void._

Fine with me

3. _Move `Return<>` to the connection builder and add a new typed connection builder class_

I agree

4. _`Resolve` should return reference to this, not void._

That's fine, but I don't think we should be making breaking changes within v4.  It should be a separate PR targeting develop for v5.

5. _`ConnectionBuilder` need to be rewritten a little more. Yes, a bunch of breaking changes..._

I do not want to make any breaking changes within v4.  I do want to add scoped connection builders for v4.  We can consider breaking changes for v5 (even if they are breaking against code that was added within this PR).

6. _Implementation of `Unidirectional` and `Bidirectional` methods yields 8 arguments on field._

We can probably fix that without breaking changes.  It can't really do that now, can it?  Isn't it just going to throw an exception because of a duplicate argument name?  I'd have to check.

src/GraphQL.MicrosoftDI/ScopedConnectionBuilderExtensions.cs
Shane32: We have reviewed and addressed your suggested changes within #2458 

I have carefully revisited the idea of adding a `ScopedConnectionResolver`.  There still does not appear to be any reason to add this class.  While we can pull the page size from metadata and infer the `Bidirectional` property from `FieldType.Arguments`, there is no matching `ConnectionResolver` in the base project.  Also there is no `.Resolve(IFieldResolver resolver)` method exposed on the connection builder (although `FieldType` is exposed as a workaround).  And finally, it is very easy, as described above, to implement your own connection resolver using the `ScopedFieldResolver` method in conjunction with constructing a new `ResolveConnectionContext`.  So for those reasons I still am not adding a `ScopedConnectionResolver` class in this PR.

The only benefit would occur if we ALSO integrate `ScopedResolveConnectionContextAdapter` (an internal class) with `ResolveConnectionContext<T>`.  This would result in one less allocation per execution, at the cost of a more complex code design.  Implementing either one without the other does not remove the allocation.

Would you agree with this thought process?

src/GraphQL.MicrosoftDI/ConnectionResolverBuilder.cs
Shane32: With #2458 this is now able to be completed and has been done.

src/GraphQL.MicrosoftDI/ConnectionResolverBuilder.cs
Shane32: With #2458 this is now able to be completed and has been done.

Shane32:
Tests run fine locally.  Seems that the Windows tests are failing again.

src/GraphQL.MicrosoftDI/ScopedConnectionBuilderExtensions.cs
sungam3r: I don't know yet. I need to see. Recently, I have less time to work on the project, so the review is delayed.

Shane32: gentle reminder @sungam3r

sungam3r: I did not read  email for a few days, was busy. I will do the review tomorrow. As I remember, we have already closed most questions.

Shane32: Why did you release 4.3 without this? I thought you were going to review it today? It’s all done.

Shane32: Now we’d need to release 4.4 on the same day as 4.3

sungam3r: I forgot. I do the review right now.

.github/workflows/test-code.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Setup .NET Core 5.0 SDK
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: '5.0.x'
+          dotnet-version: '5.0.201'
```
sungam3r, line 51: Needs pull from master.

src/GraphQL.MicrosoftDI/ScopedConnectionBuilderExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+                    return await resolver(new ScopedResolveConnectionContextAdapter<TSourceType>(context, scope.ServiceProvider));
+                }
+            });
+            return builder;
+        }
+
+        /// <summary>
+        /// Creates a resolve builder for the connection field.
+        /// </summary>
+        public static ConnectionResolverBuilder<TSourceType, object> Resolve<TSourceType>(this ConnectionBuilder<TSourceType> builder)
```
sungam3r, line 49: Will this method then be deleted along with the `ConnectionBuilder<TSourceType>`?

sungam3r (approved):

src/GraphQL.MicrosoftDI/ScopedConnectionBuilderExtensions.cs
Shane32: I expect so.

.github/workflows/test-code.yml
sungam3r: @Shane32 You pushed `5.0.201` back into master.

Shane32: Whoops I thought it would be fixed with the merge

Shane32: Fixed

PR #2430, opened 2021-03-21
Apply NRT annotations to scalars

sungam3r:
I suggest gradually add `#nullable enable` by individual PRs. It will be easier. When we add nullable annotations throughout the code, then we can add `<Nullable>enable</Nullable>` to Directory.build.props and delete all `#nullable enable` from files.

### DIFF
--- a/src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
@@ -23,6 +23,8 @@ namespace GraphQL.Tests.Types
          *    BooleanGraphType
          *    FloatGraphType
          *    StringGraphType
+         *
+         *  Does not test:
          *    IdGraphType
          *    DecimalGraphType
          *    UriGraphType
--- a/src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
@@ -19,20 +19,18 @@ namespace GraphQL.Tests.Types
          *    UIntGraphType
          *    LongGraphType
          *    ULongGraphType
-         *    
+         *
          *    BooleanGraphType
          *    FloatGraphType
          *    StringGraphType
-         *    
-         *  Does not test:
          *    IdGraphType
          *    DecimalGraphType
          *    UriGraphType
          *    date/time graph types
          *    enumeration graph types
-         *    
+         *
          *  Does test ALL scalars' handling of null
-         *  
+         *
          */
 
         [Theory]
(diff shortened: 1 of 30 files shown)


src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -78,25 +80,25 @@ public void AddValue(EnumValueDefinition value)
...
             null => true,
             _ => false
         };
 
         /// <inheritdoc/>
-        public override object Serialize(object value)
+        public override object? Serialize(object? value)
         {
-            if (value == null)
+            if (value == null) // TODO: why? null as internal value may be mapped to some external enumeration name
```
sungam3r, line 101: @Shane32 Pay attention to this, please.

Shane32: Because in general, enumerations are value types and null would be a nullable enum. If null was able to be mapped, it would break the normal use case and throw an exception because null is unable to be matched. We should not support mapping null for enums as a built-in feature. People can override Serialize and the other methods if they want to serialize to/from null.

src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
```diff
@@ -19,20 +19,18 @@ public class AllScalarGraphTypeTests
          *    UIntGraphType
          *    LongGraphType
          *    ULongGraphType
-         *    
+         *
          *    BooleanGraphType
          *    FloatGraphType
          *    StringGraphType
-         *    
-         *  Does not test:
```
Shane32, line 27: Why remove this comment? Does it test all the semantics of the mentioned types? For example, that the proper exception is returned for deserializing an invalid date type? Or overflow conditions for a decimal? I don’t think so. Initially it didn’t test these hardly at all because these types cannot be inlined as an attribute.  (Not at a pc just now but going from memory. )

Shane32 (commented):
See comments. All other changes look good

src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
sungam3r: Was this comment for single test or a whole file?

Shane32: For the four types listed right under it.

Shane32: For the whole file

sungam3r: ok

PR #2442, opened 2021-03-24
update Newtonsoft to v13

SimonCropp:


Shane32:
Duplicate of #2440 . Should we be upgrading the newtonsoft dependency? Or might some users need to stay on the old one? It seems that GraphQL runs fine on either.

SimonCropp:
i would prefer to reduce friction for people who are actively keeping references up to date

Shane32:
> i would prefer to reduce friction for people who are actively keeping references up to date

I'm not sure how that answers the question.  To "reduce friction", we would stick with the older version (perhaps testing with the new version or even testing against both versions), because then people are not required to upgrade newtonsoft if they are using another library that relies on semantics in the old version.  But then you say "who are actively keeping references up to date", which means the opposite.

SimonCropp:
ok

SimonCropp:
newtonsoft follow "assembly version is major of the nuget version". so all of v13.x will have an assembly version of 13.0.0.0. so be prepared for the questions about binding redirects.

Shane32:
I’ve never been able to get binding redirects to work. Which is why I always target the oldest compatible version that’s compatible with the current version, so that binding redirects are not necessary - ??? Maybe I don’t understand what binding redirects are or how they are used.

sungam3r:
Binding redirects make sense only for .NET Framework, not .NET Core.

sungam3r:
I would to keep v12 for GraphQL.NET v4 and bump it for v13 for GraphQL.NET v5 in `develop` branch.

Shane32:
> I would to keep v12 for GraphQL.NET v4 and bump it for v13 for GraphQL.NET v5 in `develop` branch.

I agree

PR #2449, opened 2021-03-25
Rewrite auth docs

sungam3r:
The example is outdated and I deleted it. I considered that it makes no sense to try to give here a manually written example. We have two full-fledged authorization projects and it is much easier to refer to them. It is unlikely that someone will use their own authorization. An old example could confuse the reader because he could have the impression that this example should be copied and used as is.

### DIFF
--- a/docs2/site/docs/getting-started/authorization.md
+++ b/docs2/site/docs/getting-started/authorization.md
@@ -1,114 +1,41 @@
 # Authorization
 
-> See the [Authorization](https://github.com/graphql-dotnet/authorization) project for a
-> more in depth implementation of the following idea. Keep in mind that alongside this
-> project there is a similar [Authorization.AspNetCore](https://github.com/graphql-dotnet/server/tree/develop/src/Authorization.AspNetCore)
-> project specifically for ASP.NET Core apps.
+The GraphQL operation authorization problem is out of scope of GraphQL.NET project but we
+provide some [extension methods](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/AuthorizationExtensions.cs)
+to configure authorization requirements to fields, types and other shema elements.
 
-You can write validation rules that will run before a query is executed. You can use this
-pattern to check that the user is authenticated or has permissions for a specific field.
-This example uses the `Metadata` dictionary available on Fields to set permissions per field.
+You can write validation rules that will run before a GraphQL document is executed. You
+can use this pattern to check that the user is authenticated or has permissions for a
+specific field or type.
+
+See the [Authorization](https://github.com/graphql-dotnet/authorization) project for an
+implementation of authorization framework on top of GraphQL.NET. Keep in mind that alongside
+this project there is a similar [Authorization.AspNetCore](https://github.com/graphql-dotnet/server/tree/develop/src/Authorization.AspNetCore)
+project designed specifically for ASP.NET Core apps.
+
+This example shows how to mark type and field with the specified authorization policy.
 
 ```csharp
 public class MyGraphType : ObjectGraphType
 {
   public MyGraphType()
   {
-    this.RequirePermission("READ_ONLY");
-    Field(x => x.Secret).RequirePermission("Admin");
+    this.AuthorizeWith("ReadOnlyPolicy");
+    Field(x => x.Secret).AuthorizeWith("Admin");
   }
 }
 ```
+
+Authorization policy may consist from various authorization requirements and is configured
+outside your GraphType and schema. `AuthorizeWith` method just marks the type or field that
+this schema element should be checked against the specified policy before executing GraphQL
+document. Both frameworks mentioned above support the concept of policies and authorization
+requirements. See the examples in the corresponding repositories.
 
 ## Validation Rule
 
-```csharp
-public class RequiresAuthValidationRule : IValidationRule
-{
-  public Task<INodeVisitor> ValidateAsync(ValidationContext context)
-  {
-    var userContext = context.UserContext as GraphQLUserContext;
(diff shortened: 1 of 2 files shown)


docs2/site/docs/getting-started/authorization.md
```diff
@@ -1,114 +1,41 @@
...
-  public MyGraphType()
-  {
-    this.RequirePermission("READ_ONLY");
-    Field(x => x.Secret).RequirePermission("Admin");
-  }
-}
-```
+See the [Authorization](https://github.com/graphql-dotnet/authorization) project for an
+implementation of authorization framework on top of GraphQL.NET. Keep in mind that alongside
+this project there is a similar [Authorization.AspNetCore](https://github.com/graphql-dotnet/server/tree/develop/src/Authorization.AspNetCore)
```
Shane32, line 13: I think this link should point to the home of the server repo, which should contain documentation within the readme file, or to a dedicated documentation site.  Perhaps it could point to a documentation file within the repo.  But essentially, it should point to other documentation/samples so people can learn how to implement it, not to source code.

Shane32:
> It is unlikely that someone will use their own authorization

I almost always write my own authorization code.  I've tried ASP.Net Identity framework multiple times and rejected it almost every time.

While reading this PR, I was thinking that I might write some authorization code almost exactly like the sample, rather than using either of the two other available options.  The beauty of it is in its simplicity -- there's a single validation rule to be written and that executes, and there's only a single authorization-related method in the main project (`AuthorizeWith`).  I need not have to deal with claims or principals or anything else.  I would probably even copy `AuthorizeWith` to accept an enum rather than a string.  The validation rule can simply check the `UserContext` to determine the permission level available, and return a validation error if it doesn't match the required access level.

Now perhaps if I see a good demonstration using the authorization project that requires a very small amount of code (less than writing a simple validation rule), I would use it.  I'll review again after the new changes that are being made to the project.

Anyway, I just wanted to point out that some people prefer their own authentication code, before this sample is deleted.  For instance, this might make a great sample for a custom validation rule.  I would suggest we move this sample into the https://graphql-dotnet.github.io/docs/getting-started/query-validation page (after updating the sample), not because we want to suggest this as a good authentication framework, but rather as a good sample of a basic validation rule which is also potentially useful -- and that is the best type of sample we can provide.

Jonatthu:
@sungam3r  I am using my own authorization framework tho :(

sungam3r:
I'll be back to work on this later.

Jonatthu:
This is my current file

```
using GraphQL;
using GraphQL.Types;
using GraphQL.Language.AST;
using Generatthor.Designer.Core.Api.Models;
using Generatthor.Designer.Core.Api.Services;
using Generatthor.Designer.Core.Api.Extensions;
using GraphQL.Validation;
using System.Threading.Tasks;

namespace Generatthor.Designer.Core.Api.ValidationRules
{
	public class RequiresAuthValidationRule<TPermissionsEnum> : IValidationRule
		where TPermissionsEnum : struct
	{
		private readonly IUserContext<TPermissionsEnum> userContext;
		private readonly IAuthPermissionsService<TPermissionsEnum> authPermissionsService;

		public RequiresAuthValidationRule(
			IUserContext<TPermissionsEnum> userContext,
			IAuthPermissionsService<TPermissionsEnum> authPermissionsService
		)
		{
			this.userContext = userContext;
			this.authPermissionsService = authPermissionsService;
		}

		public async Task<INodeVisitor> ValidateAsync(ValidationContext context)
		{
			return await Task.FromResult(new EnterLeaveListener(_ =>
			{

				dynamic inputType = null;


				_.Match<Operation>(op =>
				{
					// TODO: Analizy if we still need to do checks on operational level.
				});


				_.Match<Argument>(f =>
				{
					inputType = context.TypeInfo.GetInputType().GetNamedType();
				});

				_.Match<ObjectField>(fieldAst =>
				{
					if (inputType.GetField(fieldAst.Name) is IProvideMetadata fieldDef)
					{
						TPermissionsEnum? permission = fieldDef.GetPermissionEnum<TPermissionsEnum>()?.Get;

						if (permission.HasValue)
						{
							if (authPermissionsService.TryGetPermissionValues(permission.Value, out var permissionValues))
							{
								if (userContext.DenyPermission(permission.Value, out DenyPermissionReasonEnum reason))
								{
									bool strictFlag = reason.HasFlag(DenyPermissionReasonEnum.Strict);
									bool notAuthenticated = reason == DenyPermissionReasonEnum.NotAuthenticated;

									if (strictFlag || notAuthenticated == false)
									{
										context.ReportError(
											new ValidationError(
												context.OriginalQuery,
												reason.ToString(),
												$"You are not authorized to run this query.",
												fieldAst
											)
										);
									}
								}
							}
						}

					}
				});

				_.Match<Field>(fieldAst =>
				{
					FieldType fieldDef = context.TypeInfo.GetFieldDef();

					if (fieldDef != null)
					{
						TPermissionsEnum? permission = fieldDef.GetPermissionEnum<TPermissionsEnum>()?.Get;

						if (permission.HasValue)
						{
							if (authPermissionsService.TryGetPermissionValues(permission.Value, out var permissionValues))
							{
								if (userContext.DenyPermission(permission.Value, out DenyPermissionReasonEnum reason))
								{
									bool strictFlag = reason.HasFlag(DenyPermissionReasonEnum.Strict);
									bool notAuthenticated = reason == DenyPermissionReasonEnum.NotAuthenticated;

									if (strictFlag || notAuthenticated == false)
									{
										context.ReportError(
											new ValidationError(
												context.OriginalQuery,
												reason.ToString(),
												$"You are not authorized to run this query.",
												fieldAst
											)
										);
									}
								}
							}
						}
					}
				});
			}));
		}

	}
}
```

PR #2451, opened 2021-03-26
Add GraphQLAuthorizeAttribute

sungam3r:
This attribute exists in both server and authorization projects. No new dependencies, attribute just class `AuthorizeWith` extension which is already here.


Shane32 (changes requested):
Why no constructor with `string policy` argument?

sungam3r:
I did not think, just copied the code. Fixed.

Shane32 (approved):

PR #2452, opened 2021-03-28
[WIP] ConnectionBuilder and  FieldBuilder rework [DRAFT]

sungam3r:
Some ideas related to #2427

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
@@ -51,11 +51,7 @@ namespace GraphQL.MicrosoftDI
         public GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ResolveAsync(System.Func<GraphQL.IResolveFieldContext<TSourceType>, T1, T2, T3, T4, T5, System.Threading.Tasks.Task<TReturnType>> resolver) { }
         public GraphQL.MicrosoftDI.ResolverBuilder<TSourceType, TReturnType, T1, T2, T3, T4, T5> WithScope() { }
     }
-    public class ScopedAsyncFieldResolver<TReturnType> : GraphQL.Resolvers.AsyncFieldResolver<TReturnType>
-    {
-        public ScopedAsyncFieldResolver(System.Func<GraphQL.IResolveFieldContext, System.Threading.Tasks.Task<TReturnType>> resolver) { }
-    }
-    public class ScopedAsyncFieldResolver<TSourceType, TReturnType> : GraphQL.Resolvers.AsyncFieldResolver<TReturnType>
+    public class ScopedAsyncFieldResolver<TSourceType, TReturnType> : GraphQL.Resolvers.AsyncFieldResolver<TSourceType, TReturnType>
     {
         public ScopedAsyncFieldResolver(System.Func<GraphQL.IResolveFieldContext<TSourceType>, System.Threading.Tasks.Task<TReturnType>> resolver) { }
     }
@@ -65,11 +61,7 @@ namespace GraphQL.MicrosoftDI
         public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ResolveScoped<TSourceType, TReturnType>(this GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> builder, System.Func<GraphQL.IResolveFieldContext<TSourceType>, TReturnType> resolver) { }
         public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ResolveScopedAsync<TSourceType, TReturnType>(this GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> builder, System.Func<GraphQL.IResolveFieldContext<TSourceType>, System.Threading.Tasks.Task<TReturnType>> resolver) { }
     }
-    public class ScopedFieldResolver<TReturnType> : GraphQL.Resolvers.FuncFieldResolver<TReturnType>
-    {
-        public ScopedFieldResolver(System.Func<GraphQL.IResolveFieldContext, TReturnType> resolver) { }
-    }
-    public class ScopedFieldResolver<TSourceType, TReturnType> : GraphQL.Resolvers.FuncFieldResolver<TReturnType>
+    public class ScopedFieldResolver<TSourceType, TReturnType> : GraphQL.Resolvers.FuncFieldResolver<TSourceType, TReturnType>
     {
         public ScopedFieldResolver(System.Func<GraphQL.IResolveFieldContext<TSourceType>, TReturnType> resolver) { }
     }
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -441,28 +441,22 @@ namespace GraphQL
 }
 namespace GraphQL.Builders
 {
-    public static class ConnectionBuilder
-    {
-        public static GraphQL.Builders.ConnectionBuilder<TSourceType> Create<TNodeType, TSourceType>()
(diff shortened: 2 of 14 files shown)


src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
```diff
@@ -51,11 +51,7 @@ namespace GraphQL.MicrosoftDI
         public GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ResolveAsync(System.Func<GraphQL.IResolveFieldContext<TSourceType>, T1, T2, T3, T4, T5, System.Threading.Tasks.Task<TReturnType>> resolver) { }
         public GraphQL.MicrosoftDI.ResolverBuilder<TSourceType, TReturnType, T1, T2, T3, T4, T5> WithScope() { }
     }
-    public class ScopedAsyncFieldResolver<TReturnType> : GraphQL.Resolvers.AsyncFieldResolver<TReturnType>
-    {
-        public ScopedAsyncFieldResolver(System.Func<GraphQL.IResolveFieldContext, System.Threading.Tasks.Task<TReturnType>> resolver) { }
-    }
-    public class ScopedAsyncFieldResolver<TSourceType, TReturnType> : GraphQL.Resolvers.AsyncFieldResolver<TReturnType>
+    public class ScopedAsyncFieldResolver<TSourceType, TReturnType> : GraphQL.Resolvers.AsyncFieldResolver<TSourceType, TReturnType>
```
Shane32, line 54: I designed this to be flexible and reduce allocations.  With this design there is always an extra allocation required.  I will need to review my thought process and use cases.  Same thought applies to `ScopedFieldResolver`.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -501,8 +490,11 @@ namespace GraphQL.Builders
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Directive(string name, System.Action<GraphQL.Types.AppliedDirective> configure) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Directive(string name, string argumentName, object argumentValue) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Name(string name) { }
-        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Resolve(GraphQL.Resolvers.IFieldResolver resolver) { }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Resolve(GraphQL.Resolvers.IFieldResolver<TReturnType> resolver) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Resolve(System.Func<GraphQL.IResolveFieldContext<TSourceType>, TReturnType> resolve) { }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ResolveAsync(GraphQL.Resolvers.IFieldResolver<GraphQL.DataLoader.IDataLoaderResult<TReturnType>> resolver) { }
```
Shane32, line 495: It is important not to have `Async` in the method name here.  An asynchronous method returning a `Task<IDataLoaderResult>` is distinctly different than a synchronous method returning a `IDataLoaderResult`.  This is one of the bugs that people had been complaining about that was fixed in the new data loader design.  See #945 .

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -441,28 +441,22 @@ namespace GraphQL
...
     public class ConnectionBuilder<TSourceType>
     {
         public GraphQL.Types.FieldType FieldType { get; set; }
-        public GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType>(string name, string description)
+        public GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType>(string name, System.Action<GraphQL.Types.QueryArgument> configure = null)
+            where TArgumentGraphType : GraphQL.Types.IGraphType { }
+        public GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType>(string name, string description, System.Action<GraphQL.Types.QueryArgument> configure = null)
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
-        public GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType, TArgumentType>(string name, string description, TArgumentType defaultValue = default)
+        public GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType, TArgumentType>(string name, string description, TArgumentType defaultValue = default, System.Action<GraphQL.Types.QueryArgument> configure = null)
```
Shane32, line 451: This is good so long as there is no issue with the compiler questioning which overload to use.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -479,11 +473,6 @@ namespace GraphQL.Builders
             where TEdgeType : GraphQL.Types.Relay.EdgeType<TNodeType>
             where TConnectionType : GraphQL.Types.Relay.ConnectionType<TNodeType, TEdgeType> { }
     }
-    public static class FieldBuilder
-    {
-        public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Create<TSourceType, TReturnType>(GraphQL.Types.IGraphType type) { }
-        public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Create<TSourceType, TReturnType>(System.Type type = null) { }
-    }
```
Shane32, line 486: I understand that this "duplicates existing code", but I like `FieldBuilder.Create<>` better than `FieldBuilder<>.Create`.  Not a big deal though.

Shane32 (commented):
I don't have time for a full review today, but I did review the api diff and made a few comments.

I don't think any breaking changes should be made for v4, but we can make any changes that are not breaking, and breaking changes can go in a PR for v5.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
sungam3r: Does the same apply to connection builder? I'm ok either way but both builders should act the same IMO - with or without TSourceType.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: One static method calls another one. It seems pointless to me.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
Shane32: Yes if it was public, although I had made it internal for reasons previously stated (and then eliminated redundant code). We can make it public if we expose bidirectional and similar as `get;` properties on the builder. This also eliminates an allocation.

sungam3r: By the way, as for builders I'm not so concerned about allocations.

sungam3r: And regarding these 3 properties in ConnectionBuilder. In fact, these are calculated properties - `_isUnidirectional` and `_isBidirectional`. The real object that is configured by builder is `FieldType`. `_pageSize` is not calculated but could be stored in `FieldType.Metadata` so `ConnectionBuilder` will store only reference to `FieldType` as `FieldBuilder` does. `IsUnidirectional` and `IsBidirectional` can be implemented in such a way that they set the arguments only if necessary.

Shane32: The allocations I was referring to occur during execution. I am not counting allocations during build time.

PR #2457, opened 2021-03-29
Nested input types with the same not supported

SimonCropp:
seems input types are keyed off the name, which is not unique. I assume the same would happen for input types in different namespaces

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug_nested_type_names_dont_conflict.cs
+++ b/src/GraphQL.Tests/Bugs/Bug_nested_type_names_dont_conflict.cs
@@ -36,10 +36,26 @@ mutation M($input_0: MyInput!) {
                     "run",
                     arguments: new QueryArguments(new QueryArgument<MyInputClass.MyInput> {Name = "input"}),
                     resolve: ctx => ctx.GetArgument<MyInputClass>("input").Id);
+                Field<StringGraphType>(
+                    "run2",
+                    arguments: new QueryArguments(new QueryArgument<MyInputClass2.MyInput> { Name = "input" }),
+                    resolve: ctx => ctx.GetArgument<MyInputClass2>("input").Id);
             }
         }
 
         public class MyInputClass
+        {
+            public string Id { get; set; }
+
+            public class MyInput : InputObjectGraphType
+            {
+                public MyInput()
+                {
+                    Field<NonNullGraphType<StringGraphType>>("id");
+                }
+            }
+        }
+        public class MyInputClass2
         {
             public string Id { get; set; }
 
--- a/src/GraphQL.Tests/Bugs/Bug_nested_type_names_dont_conflict.cs
+++ b/src/GraphQL.Tests/Bugs/Bug_nested_type_names_dont_conflict.cs
@@ -34,7 +34,7 @@ mutation M($input_0: MyInput!) {
             {
                 Field<StringGraphType>(
                     "run",
-                    arguments: new QueryArguments(new QueryArgument<MyInput> {Name = "input"}),
+                    arguments: new QueryArguments(new QueryArgument<MyInputClass.MyInput> {Name = "input"}),
                     resolve: ctx => ctx.GetArgument<MyInputClass>("input").Id);
             }
         }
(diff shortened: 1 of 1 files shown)


src/GraphQL.Tests/Bugs/Bug_nested_type_names_dont_conflict.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+        public class MyInputClass2
+        {
+            public string Id { get; set; }
+
+            public class MyInput : InputObjectGraphType
+            {
+                public MyInput()
+                {
+                    Field<NonNullGraphType<StringGraphType>>("id");
```
Shane32, line 66:
```suggestion
                    Name = "MyInput2";
                    Field<NonNullGraphType<StringGraphType>>("id");
```
Is this not satisfactory?

SimonCropp: shouldnt this be supported OOTB?

sungam3r: OK

sungam3r: In general, the problem is wider and inherent to all grapj types. By default, the type name is defined in `GraphType.cs` as:
```
  private string GetDefaultName()
        {
            var type = GetType();

            var attr = type.GetCustomAttribute<GraphQLMetadataAttribute>();

            if (!string.IsNullOrEmpty(attr?.Name))
            {
                return attr.Name;
            }

            var name = type.Name.Replace('`', '_');
            if (name.EndsWith(nameof(GraphType), StringComparison.InvariantCulture))
                name = name.Substring(0, name.Length - nameof(GraphType).Length);

            return name;
        }
```

That is, only the name type is used without namespace and without parent types. This is done intentionally for ease of names. We _can_ change the algorithm to take into account paren types and namespaces, but it will be a breaking change. I suggest to use `GlobalSwitches`.

sungam3r: I'll post a PR.

sungam3r:
@SimonCropp Thanks for the repro. Closing this as issue now is tracked in #2459 .

Shane32:
I'm not exactly sure how this is a bug.  The default name of a type (including input object graph types) is derived from the class name.  There are other ways to create an duplicate implicit class name.  When a duplicate occurs, you should manually set the name of one or the other of the classes to a specified name, either via code or an attribute.  Besides the sample shown in the PR (putting the classes in different namespaces or as subclasses of different classes), here are four methods to create a graph type with the same name:

```cs
public class Widget : ObjectGraphType { }  // Name defaults to "Widget"

public class WidgetGraphType : ObjectGraphType { } // Name defaults to "Widget"

{
    public MyGraphType() {
        Name = "Widget";
    }
}

[GraphQLMetadata("Widget")]
public class MyGraphType2 : ObjectGraphType { }
```

Further, even if we wanted to "fix" it, the name is set before the schema is constructed.  You'd have to generate a longer name (e.g. from `FullName`), which doesn't make sense to me, since the graph type's constructor has no way of knowing what other types might or might not be added to the schema.

@sungam3r 's solution in #2459 does not address namespace differences.  However, at least it ONLY affect nested types.  But let's consider the consequences: For the above sample, now the input graph type names are `MyInputClass_MyInput` and `MyInputClass2_MyInput`.  The type names are necessary when used as a variable, so you'd be writing:

```gql
mutation ($arg: MyInputClass_MyInput) {
   doSomeStuff(arg: $arg)
}
```

It seems to me if someone runs into a duplicate type name, they should define the name they want, not rely on a global switch to make longer names.

In short, the only time I see this being either desirable or useful would be for writing tests.  And I don't think we should be adding this option for tests.

sungam3r:
In general I agree. I thought it would not hurt to give a global option. In the end, the default behavior does not change. Perhaps someone will come in handy behavior.

PR #2458, opened 2021-03-29
Add typed connection builders

Shane32:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/2427#issuecomment-808909199

This PR addresses issue 3, 4 and 6:

> Move Return<> to the connection builder and add a new typed connection builder class

> Resolve should return reference to this, not void.

> Implementation of Unidirectional and Bidirectional methods yields 8 arguments on field.

We can add the `IsBidirectional` in another PR if desired.  I would not add `IsUnidirectional` and instead made `Unidirectional()` obsolete (and didn't add it to the new class).  We cannot add a property called `PageSize` because `PageSize()` already exists as a method.

We cannot change the existing methods to return the typed connection builders because that would be a breaking change.  But I did add a `Returns<T>()` method to the existing connection builder which will return a typed connection builder.  For v5, we can change the `ComplexFieldType<TSourceType>.Connection<TNodeType>()` method to return a typed connection builder (even if it defaults to type `object` as the return type).  This would "fix" the existing methods that still return `void` - since they would not exist anymore.

Once we decide on this PR and the other changes we have discussed, I can update the scoped extensions.


Shane32:
Added features from #2452 
- Added non-breaking versions of additional `Argument(...)` overloads to old class
- Added suggested revisions to `Argument(...)` overloads to new class
- Added `Configure()` method
- Added `Directive()` methods
- Added NRT annotations

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -1,8 +1,13 @@
 using System;
+#if NETSTANDARD2_1
+using System.Diagnostics.CodeAnalysis;
+#endif
```
Shane32, line 4: Thanks to our code formatting rules...  🙄

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -52,8 +57,6 @@ public static class ConnectionBuilder
     /// </summary>
     public class ConnectionBuilder<TSourceType>
     {
-        private bool _isUnidirectional;
-
         private bool _isBidirectional;
```
sungam3r, line 60: I suggest to make it as computed, not a field.

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -52,8 +57,6 @@ public static class ConnectionBuilder
...
 
         private int? _pageSize;
```
sungam3r, line 62: Can be moved into `FieldType.Metadata`

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: Computed from what?  Save/get from metadata?  Or are you thinking something else?

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32, line 62:
```suggestion
        internal const string _bidirectionalMetadataKey = "__ConnectionBuilder_Bidirectional";
        internal const string _pageSizeMetadataKey = "__ConnectionBuilder_PageSize";

        private bool _isBidirectional
        {
            get => FieldType.GetMetadata<bool>(_bidirectionalMetadataKey);
            set => FieldType.WithMetadata(_bidirectionalMetadataKey, value.Boxed());
        }

        private int? _pageSize
        {
            get => FieldType.GetMetadata<int?>(_pageSizeMetadataKey);
            set => FieldType.WithMetadata(_pageSizeMetadataKey, value);
        }
```

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,293 @@
...
+namespace GraphQL.Builders
+{
+    /// <summary>
+    /// Builds a connection field for graphs that have the specified source type.
+    /// </summary>
+    public class ConnectionBuilder<TSourceType, TReturnType>
+    {
+        private bool _isBidirectional;
+
+        private int? _pageSize;
```
Shane32, line 20:
```suggestion
        private bool _isBidirectional
        {
            get => FieldType.GetMetadata<bool>(ConnectionBuilder<TSourceType>._bidirectionalMetadataKey);
            set => FieldType.WithMetadata(ConnectionBuilder<TSourceType>._bidirectionalMetadataKey, value.Boxed());
        }

        private int? _pageSize
        {
            get => FieldType.GetMetadata<int?>(ConnectionBuilder<TSourceType>._pageSizeMetadataKey);
            set => FieldType.WithMetadata(ConnectionBuilder<TSourceType>._pageSizeMetadataKey, value);
        }
```

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: ?

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: From the current arguments. These properties are no more that reflection of some arguments set on the `FieldType`.

sungam3r: after+first / before+last . Just check for args with taht name and type.

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -222,27 +248,101 @@ public ConnectionBuilder<TSourceType> Argument<TArgumentGraphType>(string name,
         /// <param name="name">The name of the argument.</param>
         /// <param name="description">The description of the argument.</param>
         /// <param name="defaultValue">The default value of the argument.</param>
-        public ConnectionBuilder<TSourceType> Argument<TArgumentGraphType, TArgumentType>(string name, string description,
-            TArgumentType defaultValue = default)
+        public ConnectionBuilder<TSourceType> Argument<TArgumentGraphType, TArgumentType>(string name, string? description,
+#if NETSTANDARD2_1
+            [AllowNull]
```
sungam3r, line 253: 🤔 Try to remove this and just use `TArgumentType? defaultValue = default`. It should work for C#9.

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: Only for page size. Why internal ?

sungam3r: Bidirectional or Unidirectional can be calculated from the presence or absence of required arguments.

Shane32: `internal` so that it can be accessed from `ConnectionBuilder<TSourceType, TReturnType>`.  In v5 we will remove this class altogether of course, and then the const can be private.

`_bidirectional` can be inferred as you say, like this I assume:

```cs
private bool _isBidirectional => FieldType.Arguments.Any(x => x.Name == "before");
```

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: It compiles, but does not provide the same semantics as the prior code.  If `TArgumentType` is `int`, then it will become `int? defaultValue = null` rather than `int defaultValue = default`.  Perhaps this is more desirable, but in any case, it is different (and has a different method signature), and so it is a breaking change.

I would suggest changing both the field and connection builders for v5 to use `TArgumentType?` so that no default is assigned to the argument if the argument is not specified.  But for v4, I would just add the appropriate NRT annotations and keep the semantics uniform.

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: Do you think the const should be public?

Shane32: Please review updated code.

sungam3r: Internal is ok for now.

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: Ok

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+using System.Threading.Tasks;
+using GraphQL.Types;
+using GraphQL.Types.Relay;
+
+#nullable enable
+
+namespace GraphQL.Builders
+{
+    /// <summary>
+    /// Builds a connection field for graphs that have the specified source type.
```
sungam3r, line 15:
```suggestion
    /// Builds a connection field for graphs that have the specified source and return type.
```

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -112,27 +115,27 @@ public static ConnectionBuilder<TSourceType> Create<TNodeType>(string name = "de
...
-                .Unidirectional();
+            fieldType.Arguments.Add(new QueryArgument(typeof(StringGraphType))
+            {
+                Name = "after",
+                Description = "Only look at connected edges with cursors greater than the value of `after`.",
+            });
+            fieldType.Arguments.Add(new QueryArgument(typeof(IntGraphType))
+            {
+                Name = "first",
+                Description = "Specifies the number of edges to return starting from `after` or the first entry if `after` is not specified.",
```
sungam3r, line 126: Seems a bit weird. https://relay.dev/graphql/connections.htm#sec-Forward-pagination-arguments

```suggestion
                Description = "Specifies the number of edges to return (at most) starting from `after` or just the first number of edges if `after` is not specified.",
```

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -112,27 +115,27 @@ public static ConnectionBuilder<TSourceType> Create<TNodeType>(string name = "de
...
+            });
+            return new ConnectionBuilder<TSourceType>(fieldType);
         }
 
         /// <summary>
         /// Configure the connection to be forward-only.
         /// </summary>
+        [Obsolete("Calling Unidirectional is unnecessary and will be removed in future versions.")]
```
sungam3r, line 134: I'm not sure but OK for now.

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: This change will affect existing schemas.  Is this okay?  If not, we should only make this change for the added `ConnectionBuilder_Typed.cs`.  Note that we will be deleting the untyped versions in v5.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+            };
+            fieldType.Arguments.Add(new QueryArgument(typeof(StringGraphType))
+            {
+                Name = "after",
+                Description = "Only look at connected edges with cursors greater than the value of `after`.",
+            });
+            fieldType.Arguments.Add(new QueryArgument(typeof(IntGraphType))
+            {
+                Name = "first",
+                Description = "Specifies the number of edges to return starting from `after` or the first entry if `after` is not specified.",
```
Shane32, line 82:
```suggestion
                Description = "Specifies the maximum number of edges to return, starting after the cursor specified by `after`, or the first number of edges if `after` is not specified.",
```
I rewrote this a bit more as it was inaccurate (does not start at 'after').

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: Affect how? This is just  description.

Shane32: Unlike xml comments, the description is part of the generated schema. I understand there is no functionality difference.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: If we are okay making changes to the description field in between versions, we can make the same change to the untyped builder. I’d say it’s fine for a new minor version (4.x) but not a new revision version (4.2.x).

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32:
```suggestion
                Description = "Specifies the maximum number of edges to return, starting after the cursor specified by `after`, or the first number of edges if `after` is not specified.",
```

Shane32: I would suggest that we are okay making schema changes of this nature for new minor versions.  (e.g. 4.3.0, but not 4.2.2)  Since this will be a new minor version anyway (new API), I would say this is fine to merge.

sungam3r: ok

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
sungam3r: OK

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+#nullable enable
+
+namespace GraphQL.Builders
+{
+    /// <summary>
+    /// Builds a connection field for graphs that have the specified source and return type.
+    /// </summary>
+    public class ConnectionBuilder<TSourceType, TReturnType>
+    {
+        private bool _isBidirectional => FieldType.Arguments.Any(x => x.Name == "before");
```
sungam3r, line 19: What if some one add `before` field of other type ?

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+            fieldType.Arguments.Add(new QueryArgument(typeof(StringGraphType))
+            {
+                Name = "after",
+                Description = "Only look at connected edges with cursors greater than the value of `after`.",
+            });
+            fieldType.Arguments.Add(new QueryArgument(typeof(IntGraphType))
+            {
+                Name = "first",
+                Description = "Specifies the maximum number of edges to return, starting after the cursor specified by `after`, or the first number of edges if `after` is not specified.",
+            });
```
sungam3r, line 83:
```suggestion
                Arguments = new QueryArguments(
                    new QueryArgument<StringGraphType>
                    {
                        Name = "after",
                        Description = "Only look at connected edges with cursors greater than the value of `after`.",
                    },
                    new QueryArgument<IntGraphType>
                    {
                        Name = "first",
                        Description = "Specifies the maximum number of edges to return, starting after the cursor specified by `after`, or the first number of edges if `after` is not specified.",
                    }
                ),
            };
```

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+        /// </summary>
+        public ConnectionBuilder<TSourceType, TReturnType> Bidirectional()
+        {
+            if (_isBidirectional)
+                return this;
+
+            Argument<StringGraphType, string>("before",
+                "Only look at connected edges with cursors smaller than the value of `before`.");
+            Argument<IntGraphType, int?>("last",
+                "Specifies the number of edges to return counting reversely from `before`, or the last entry if `before` is not specified.");
```
sungam3r, line 99: This comment also looks weird.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
sungam3r, line 19:
```suggestion
        private bool IsBidirectional => FieldType.Arguments.Any(x => x.Name == "before");
```

It is a property, not field.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+namespace GraphQL.Builders
+{
+    /// <summary>
+    /// Builds a connection field for graphs that have the specified source and return type.
+    /// </summary>
+    public class ConnectionBuilder<TSourceType, TReturnType>
+    {
+        private bool _isBidirectional => FieldType.Arguments.Any(x => x.Name == "before");
+
+        private int? _pageSize
```
sungam3r, line 21:
```suggestion
        private int? PageSize
```

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        /// <summary>
+        /// Clears the default page size, so all records are returned by default.
+        /// </summary>
+        public ConnectionBuilder<TSourceType, TReturnType> ReturnAll()
+        {
+            _pageSize = null;
+            return this;
+        }
```
sungam3r, line 141:
```suggestion
        /// <summary>
        /// Sets the default page size or clears (if null) the default page size, so all records are returned by default.
        /// </summary>
        public ConnectionBuilder<TSourceType, TReturnType> PageSize(int? pageSize)
        {
            _pageSize = pageSize;
            return this;
        }
```

?

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+
+        private void CheckForErrors(IResolveConnectionContext<TSourceType> args)
+        {
+            if (args.First.HasValue && args.Last.HasValue)
+            {
+                throw new ArgumentException("Cannot specify both `first` and `last`.");
+            }
+            if (args.IsUnidirectional && args.Last.HasValue)
+            {
+                throw new ArgumentException("Cannot use `last` with unidirectional connections.");
```
sungam3r, line 287: Maybe change all ` to ' ?

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        /// <summary>
+        /// Sets the resolver method for the connection field.
+        /// </summary>
+        public ConnectionBuilder<TSourceType, TReturnType> ResolveAsync(Func<IResolveConnectionContext<TSourceType>, Task<TReturnType>> resolver)
+        {
+            FieldType.Resolver = new Resolvers.AsyncFieldResolver<TReturnType>(context =>
+            {
+                var args = new ResolveConnectionContext<TSourceType>(context, !_isBidirectional, _pageSize);
```
sungam3r, line 272:
```suggestion
                var context = new ResolveConnectionContext<TSourceType>(context, !_isBidirectional, _pageSize);
```
?

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        /// <summary>
+        /// Sets the resolver method for the connection field.
+        /// </summary>
+        public ConnectionBuilder<TSourceType, TReturnType> Resolve(Func<IResolveConnectionContext<TSourceType>, TReturnType> resolver)
+        {
+            FieldType.Resolver = new Resolvers.FuncFieldResolver<TReturnType>(context =>
+            {
+                var args = new ResolveConnectionContext<TSourceType>(context, !_isBidirectional, _pageSize);
```
sungam3r, line 258:
```suggestion
                var context = new ResolveConnectionContext<TSourceType>(context, !_isBidirectional, _pageSize);
```

?

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+            FieldType.ApplyDirective(name, argumentName, argumentValue);
+            return this;
+        }
+
+        /// <summary>
+        /// Apply directive to connection field specifying configuration delegate.
+        /// </summary>
+        /// <param name="name">Directive name.</param>
+        /// <param name="configure">Configuration delegate.</param>
+        public virtual ConnectionBuilder<TSourceType, TReturnType> Directive(string name, Action<AppliedDirective> configure)
```
sungam3r, line 236: Why `Configure` and three `Directive` methods are `virtual`?

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,291 @@
...
+            FieldType.Resolver = new Resolvers.AsyncFieldResolver<TReturnType>(context =>
+            {
+                var args = new ResolveConnectionContext<TSourceType>(context, !_isBidirectional, _pageSize);
+                CheckForErrors(args);
+                return resolver(args);
+            });
+            return this;
+        }
+
+        private void CheckForErrors(IResolveConnectionContext<TSourceType> args)
```
sungam3r, line 279:
```suggestion
        private void CheckForErrors(IResolveConnectionContext<TSourceType> context)
```

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -52,26 +58,23 @@ public static class ConnectionBuilder
     /// </summary>
     public class ConnectionBuilder<TSourceType>
```
sungam3r, line 59: Mark obsolete?

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: I copied and pasted your code suggestions from the other PR and you had written them as virtual. I thought I had removed virtual but apparently I didn’t. I can fix it.

sungam3r: Ooops. I remember! I did them `virtual` because all other methods in `FieldBuilder` are `virtual` too.

sungam3r: See #1351. So I would to do all methods in `ConnectionBuilder` as `virtual`.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: Impossible.  Naming conflict.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: It is private.  And renaming `_pageSize` to `PageSize` is not possible due to a naming conflict.

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: This class is necessary within v4 so it cannot be marked obsolete.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: It was a property before; you suggested I replace it with a calculated property, so I did so.  If you think it should be different, please suggest a change.  I don't really care and I don't think it matters.  It made more sense to me the other way.  I do like that PageSize is saved within metadata so that extension methods can create a resolver properly.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: I agree.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: Suggest something else, if you like, and I will implement it.  I don't care.

sungam3r: `PageSizeFromMetadata`

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
sungam3r: It's private but it is stiill propperty. Property should follow PascalCase naming.

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: OK, then `//TODO: mark obsolete in v5` or post a issue about it.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
sungam3r: I talked about `before` _argument_, not _field_, sorry. Now property checks only one argument by name, even not by type.

Shane32: True. Seems fine to me.

If you want to suggest different code, I’ll merge your suggestion. I’m not changing it otherwise.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: I agree but don’t have time to think about it. It’s not always easy to write text. I read it but nothing came to mind.

Maybe if I have time I’ll work on it. But I’ve not even been able to merge your prior suggestions in the last 4 days.

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
Shane32: Ok

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32: It will be removed in v5. I can add a comment to that effect

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -112,27 +115,27 @@ public static ConnectionBuilder<TSourceType> Create<TNodeType>(string name = "de
                 Type = typeof(TConnectionType),
                 Arguments = new QueryArguments(),
             };
-            return new ConnectionBuilder<TSourceType>(fieldType, false, false, null)
-                .Unidirectional();
+            fieldType.Arguments.Add(new QueryArgument(typeof(StringGraphType))
+            {
+                Name = "after",
+                Description = "Only look at connected edges with cursors greater than the value of 'after'.",
```
Shane32, line 121:
```suggestion
                Description = "Only return edges after the specified cursor.",
```

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -148,32 +151,29 @@ public ConnectionBuilder<TSourceType> Bidirectional()
             }
 
             Argument<StringGraphType, string>("before",
-                "Only look at connected edges with cursors smaller than the value of `before`.");
+                "Only look at connected edges with cursors smaller than the value of 'before'.");
```
Shane32, line 154:
```suggestion
                "Only return edges prior to the specified cursor.");
```

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -148,32 +151,29 @@ public ConnectionBuilder<TSourceType> Bidirectional()
...
             Argument<IntGraphType, int?>("last",
-                "Specifies the number of edges to return counting reversely from `before`, or the last entry if `before` is not specified.");
-
-            _isUnidirectional = false;
-            _isBidirectional = true;
+                "Specifies the number of edges to return counting reversely from 'before', or the last entry if 'before' is not specified.");
```
Shane32, line 156:
```suggestion
                "Specifies the maximum number of edges to return, starting prior to the cursor specified by 'before', or the last number of edges if 'before' is not specified.");
```

Shane32:
I made some more text suggestions, but they could still be better.

src/GraphQL/Builders/ConnectionBuilder.cs
Shane32:
```suggestion
    // TODO: Remove in v5
    public class ConnectionBuilder<TSourceType>
```

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
sungam3r:
```csharp
private bool IsBidirectional => FieldType.Arguments.Find("before")?.ResolvedType is StringGraphType && FieldType.Arguments.Find("last")?.ResolvedType is IntGraphType;
```

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
sungam3r: "Specifies the maximum number of edges to return counting reversely from the cursor specified by `before`, or the last number of edges if `before` is not specified.");

src/GraphQL/Builders/ConnectionBuilder.cs
sungam3r: I think you fixed an error in that comment :) https://relay.dev/graphql/connections.htm#sec-Edge-order , "counting reversely" was wrong.

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -50,28 +56,26 @@ public static class ConnectionBuilder
...
     public class ConnectionBuilder<TSourceType>
     {
-        private bool _isUnidirectional;
+        internal const string PAGE_SIZE_METADATA_KEY = "__ConnectionBuilder_PageSize";
 
-        private bool _isBidirectional;
+        private bool IsBidirectional => FieldType.Arguments.Any(x => x.Name == "before");
 
-        private int? _pageSize;
+        private int? PagrSizeFromMetadata
```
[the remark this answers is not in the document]
Shane32: Good catch.  Fixed

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,282 @@
...
+        {
+            var fieldType = new FieldType
+            {
+                Name = name,
+                Type = typeof(TConnectionType),
+                Arguments = new QueryArguments(
+                    new QueryArgument<StringGraphType>
+                    {
+                        Name = "after",
+                        Description = "Only look at connected edges with cursors greater than the value of 'after'.",
```
sungam3r, line 75:
```suggestion
                        Description = "Only return edges after the specified cursor.",
```

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,282 @@
...
+        /// <summary>
+        /// Configure the connection to be bi-directional.
+        /// </summary>
+        public virtual ConnectionBuilder<TSourceType, TReturnType> Bidirectional()
+        {
+            if (IsBidirectional)
+                return this;
+
+            Argument<StringGraphType, string>("before",
+                "Only look at connected edges with cursors smaller than the value of 'before'.");
```
sungam3r, line 97:
```suggestion
                "Only return edges prior to the specified cursor.");
```

src/GraphQL/Builders/ConnectionBuilder_Typed.cs
```diff
@@ -0,0 +1,282 @@
...
+            Argument<IntGraphType, int?>("last",
+                "Specifies the number of edges to return counting reversely from 'before', or the last entry if 'before' is not specified.");
```
sungam3r, line 99:
```suggestion
                "Specifies the maximum number of edges to return, starting prior to the cursor specified by 'before', or the last number of edges if 'before' is not specified.");
```

sungam3r (approved):
OK to merge after last comments about descriptions.

PR #2459, opened 2021-03-29
Introduce GlobalSwitches.UseDeclaringTypeNames

sungam3r:
Replaces #2457


Shane32:
I would not merge this option.  It seems only useful for tests.  But it is a good implementation if we do want to add it -- only addressing nested types (and not namespace differences).

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/pull/2457#issuecomment-809743387

Such an implementation just slightly expands behavior, without harm.

Shane32:
Ok

PR #2466, opened 2021-03-30
Fix when building a schema with many type references

sungam3r:
Before :
> System.ArgumentException: 'Type ObjectGraphType with name 'Commit' does not
 implement interface InterfaceGraphType with name 'Subscribable'.
 Field 'viewerSubscription' must be of type '__GraphQLTypeReference'
 or covariant from it, but in fact it is of type 'SubscriptionState'.'

I also added a check for '__GraphQLTypeReference' resolved type into `SchemaValidationVisitor`.


sungam3r:
I want to merge and publish 4.2.1 ASAP as this bug blocks some my code when migrating to v4.1

Shane32:
I'll review right away.

Shane32 (approved):
Will this be 4.1.1 or 4.2.0 ?   Either way is fine

Shane32:
It would be nice if we had a test that isolated the bug that caused the problem here.  Loading all of github's schema doesn't really help anyone understand what situation can cause the bug.

sungam3r:
> Will this be 4.1.1 or 4.2.0 ?

#2451 introduced new API, so 4.2.0

> Loading all of github's schema doesn't really help anyone understand what situation can cause the bug.

Yes but now I have no time to do so. This error occurred on my tests with this big schema.

Shane32:
If we decided to move forward with #2459 then let's merge that also before releasing 4.2.0.  Unless there's a reason not to.

PR #2473, opened 2021-04-02
FieldBuilder and ConnectionBuilder implement IProvideMetadata

sungam3r:
This allows to call any extension method defined on `IProvideMetadata`.

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -454,7 +454,7 @@ namespace GraphQL.Builders
             where TEdgeType : GraphQL.Types.Relay.EdgeType<TNodeType>
             where TConnectionType : GraphQL.Types.Relay.ConnectionType<TNodeType, TEdgeType> { }
     }
-    public class ConnectionBuilder<TSourceType>
+    public class ConnectionBuilder<TSourceType> : GraphQL.Types.IProvideMetadata
     {
         public GraphQL.Types.FieldType FieldType { get; set; }
         public GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType>(string name, string description)
@@ -485,7 +485,7 @@ namespace GraphQL.Builders
         public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Create<TSourceType, TReturnType>(GraphQL.Types.IGraphType type) { }
         public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Create<TSourceType, TReturnType>(System.Type type = null) { }
     }
-    public class FieldBuilder<TSourceType, TReturnType>
+    public class FieldBuilder<TSourceType, TReturnType> : GraphQL.Types.IProvideMetadata
     {
         public GraphQL.Types.EventStreamFieldType FieldType { get; }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Argument<TArgumentGraphType>(string name, System.Action<GraphQL.Types.QueryArgument> configure = null)
--- a/src/GraphQL/Builders/ConnectionBuilder.cs
+++ b/src/GraphQL/Builders/ConnectionBuilder.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using GraphQL.Types;
 using GraphQL.Types.Relay;
@@ -50,7 +51,7 @@ namespace GraphQL.Builders
     /// <summary>
     /// Builds a connection field for graphs that have the specified source type.
     /// </summary>
-    public class ConnectionBuilder<TSourceType>
+    public class ConnectionBuilder<TSourceType> : IProvideMetadata
     {
         private bool _isUnidirectional;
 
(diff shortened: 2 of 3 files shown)


sungam3r:
Relates to #2458

Shane32 (commented):
I don't like this idea.  Why should a field builder have `GetPolicies` on it, for example?  In fact, I don't see any extension methods that would be useful for a builder - ???

And besides, those extension methods are not designed for a field builder -- they don't return the builder, so they are not chainable.  We already have a `FieldType` property on the field builder so you can access the fields (and extension fields) available on the field.  This is perfectly adequate.

If we want to change the builders to use interfaces, we can do so, but I do not think it is necessary unless we want to chain different types of builders together.  For example, the LINQ `OrderBy` returns an `IOrderedQueryable` interface so that you can chain `ThenBy` (which only accepts `IOrderedQueryable`) or `Where` (which accepts any `IQueryable)`.  The `FluentMigrator` library uses this a lot.

sungam3r:
>Why should a field builder have GetPolicies on it

FieldBuilder wraps original field and provides some APIs on top of it constructiong that field. Nothing more. What if I want to call first some FieldBuilder's API, then some well-known existing API dofined as extension method on IProviderMetadata and then continue with FieldBuilder's API? Now I can not "return back" from FieldType from FieldBuilder.

> And besides, those extension methods are not designed for a field builder -- they don't return the builder, so they are not chainable.

They should not. Meaning to give the opportunity to use such signatures and allow chaining:

```csharp
 public static TMetadataProvider XXX<TMetadataProvider>(this TMetadataProvider provider, other parameters)
            where TMetadataProvider : IProvideMetadata
{
  // do something
  return provider;
}
```

The only thing needed - implement ` where TMetadataProvider : IProvideMetadata` constraint.

Shane32:
> FieldBuilder wraps original field and provides some APIs on top of it constructiong that field. Nothing more. What if I want to call first some FieldBuilder's API, then some well-known existing API dofined as extension method on IProviderMetadata and then continue with FieldBuilder's API? Now I can not "return back" from FieldType from FieldBuilder.

> They should not return builder.

If you do not return the builder, then you cannot "return back" in either case.

```cs
// basic sample
Field<IntGraphType, int>()
    .Name("hero")
    .Description("test");

// now, accessing metadata a few times
var builder = Field<IntGraphType, int>()
    .Name("hero")
    .Description("test");
builder.FieldType.MyAuthorizationExtension("test");
builder.FieldType.MyAuthorizationExtension("test2");

// now, alternate design
var field = Field<IntGraphType, int>()
    .Name("hero")
    .Description("test")
    .FieldType;
field.MyAuthorizationExtension("test");
field.MyAuthorizationExtension("test2");

// after, accessing metadata a few times
var builder = Field<IntGraphType, int>()
    .Name("hero")
    .Description("test");
builder.MyAuthorizationExtension("test");
builder.MyAuthorizationExtension("test2");
```

So if you don't design builder methods then this suggestion is pointless.  You still can't "return back".

And (in my opinion) non-builder methods should not be designed like builders; see below.

> extension method on IProviderMetadata

What methods? I still fail to see ANY extension methods defined within the library that this change would be useful for.  None!

> _code sample_

Why cannot users write an extension method like this:

```cs
public static FieldBuilder<TSourceType, TReturnType> XXX<TSourceType, TReturnType>(this FieldBuilder<TSourceType, TReturnType> provider, other parameters)
{
  // do something
  return provider;
}
```

This is already supported and requires no changes.  It is a builder method and is not mixed with the non-builder methods.

We could add interfaces, like I previously stated, if we want to eliminate having duplicate code between the field builder and connection builders.  It can be extremely simple:

```cs
public interface IFieldBuilder
{
    FieldType FieldType { get; }
}
```

Then you can write code like you showed before:

```cs
 public static TMetadataProvider XXX<TMetadataProvider>(this TMetadataProvider provider, other parameters)
            where TMetadataProvider : IFieldBuilder
{
  // do something
  return provider;
}
```

But by merging this PR, it would add a whole bunch of undesirable methods on the builder -- methods that do not return itself and cannot be chained:

- `List<string> GetPolicies()`
- `bool RequiresAuthorization()`
- `bool HasAppliedDirectives()`
- `AppliedDirectives GetAppliedDirectives()`
- `AppliedDirective FindAppliedDirective(name)`
- `Dictionary<string, object> Metadata`
- `TType GetMetadata<TType>(string key, TType defaultValue = default)`
- `TType GetMetadata<TType>(string key, Func<TType> defaultValueFactory)`
- `bool HasMetadata(string key)`

Those are each another reason not to merge this PR.  So I see 9 reasons not to merge this PR and 0 reasons to merge it.

Again, ONLY builder methods --- methods you'd want to use when defining a field --- should be shown by intellisense when you write `Field<IntGraphType, int>()` and press `.` to see the list of methods.  So this PR breaks the whole concept of a builder.  The exception is the `FieldType` property which allows access to the underlying `FieldType` instance.  The property essentially ends the builder, as it should.  It should not be designed to allow jumping back into the builder.

The root of the problem is the fact that we have signatures with and without field builders.  Half of the methods return a `FieldType` and the other half return a `FieldBuilder`.  It should never have been designed this way to begin with.  All `Field(...)` methods should return a field builder.  Only `AddField` should be exposed to add a field type to a graph type.  There should also be a way to infer the graph type from the return type for builders / resolvers.  It should not be required to use an expression in order to infer the return type.  So in my opinion, we should be able to write:

```cs
//field type: NonNullGraphType<ListGraphType<GraphQLClrOutputTypeReference<Hero>>>
Field("heros", resolve: context => dataSource.GetHeros(), nullable: false);

//field type: NonNullGraphType<GraphQlClrOutputTypeReference<int>>
Field()
    .Name("count")
    .Returns<int>()  //perhaps a nullable argument would be appropriate to add here
    .Resolve(context => dataSource.GetCount());
```

Neither of the above syntaxes are supported presently.

sungam3r:
ok

PR #2487, opened 2021-04-08
Allow SchemaBuilder to configured typed enums

### LINKED ISSUE #2485: Unable to serialize '<enum value>' to the scalar type '<enum type>'
## Description

With update to 4.x the enum conversion is broken. This is likely due to this PR

https://github.com/graphql-dotnet/graphql-dotnet/pull/1836

In file 
https://github.com/graphql-dotnet/graphql-dotnet/blob/2e49888da2393b2991bb52a5306b110dc948b224/src/GraphQL/Types/Scalars/EnumerationGraphType.cs#L230 

there is a check to match UnderlyingValue (which is of type int in C#, default). When assigning this value in https://github.com/graphql-dotnet/graphql-dotnet/blob/2e49888da2393b2991bb52a5306b110dc948b224/src/GraphQL/Types/Scalars/EnumerationGraphType.cs#L276
 the value will always be of string type. 

We are using schema first and define enum in schema like

```
enum SourceType {
    UnknownSource,
    Universal,
    Local,
    Standard
}
```

In C# the type defined is

```
public enum SourceType {
    UnknownSource = 0,
    Universal = 1,
    Local = 2,
    Standard = 3
}
```

### Steps to reproduce

Use schemafirst to create schema that uses above enumeration. Also create C# enum and then use schema to run a query.

### Expected result

enum value should be able to cast to enum type. This works as expected in 3.x

### Actual result

enum value is unable to cast to enum type. This is a break in 4.x

Shane32:
Fixes #2485 

Possible breaking change: If previously someone attempted to map a .NET enumeration type to an schema enumeration type via `TypeSettings.Include` or `TypeConfig.Type`, the value was completely ignored; basically it was an invalid configuration setting.  In that scenario, this PR would probably break their application.  I expect this to be rare.

Besides that possibility -- a user having set an invalid / ignored configuration setting -- there are no breaking changes.


docs2/site/docs/getting-started/schema-types.md
```diff
@@ -285,6 +285,25 @@ be implemented as:
...
 ```
 
+**Schema-First Enumeration Types**
+
+If you have defined your schema with the schema-first syntax, the backing value of each of the enumeration
+values will default to a string matching the name of the enumeration value. If you wish to use a C#
+enumeration type instead, configure the type as demonstrated in one of the following examples:
+
+```csharp
+var schema = Schema.For(definitions, c => {
```
sungam3r, line 295:
```suggestion
var schema = Schema.For(definitions, c =>
{
```

docs2/site/docs/migrations/migration4.md
```diff
@@ -215,6 +215,26 @@ Here are some of the situations that you may run into with version 4:
...
 
+If you have a schema-first schema, you may run into an issue with enumeration types, since the `SchemaBuilder` uses the name of each
+enumeration value as its value also. In other words, you must return a string corresponding to the enumeration value (e.g, `"Cat"` or
+`"Dog"`) rather than a matching C# enumeration value (e.g. `Animal.Cat` or `Animal.Dog`). You can configure the `SchemaBuilder` to
+match the defined enumeration values to a C# enumeration type in this manner demonstrated below. Then when used as an input type,
+the values will be parsed into the matching C# enumeration values, and when used as an output type, you must return the C# enumeration
+value (e.g. `Animal.Cat`) or its underlying value (typically an `int`). Below are a few examples of how this is configured:
+
+```csharp
+var schema = Schema.For(definitions, c => {
```
sungam3r, line 226:
```suggestion
var schema = Schema.For(definitions, c =>
{
```

sungam3r:
Please rebase to master, I fixed CI.

sungam3r (approved):

PR #2492, opened 2021-04-08
Bump GraphQL-Parser

sungam3r:


src/GraphQL.Tests/Files/SDL/GitHub.graphql
```diff
@@ -647,7 +647,7 @@ type BlameRange {
 }
 
 # Represents a Git blob.
-type Blob implements Node, GitObject {
+type Blob implements Node & GitObject {
```
sungam3r, line 650: There is a bug in my SDL printer where I took this example from.

PR #2498, opened 2021-04-11
ValidationContext.ParseValueScalar: Convert input scalar type into correct one.

JanezKr:
Migration from 3.x on 4.x have some issiue in my case.
Version 3 was not respected scalar input types.  But i think that is this good in most cases.

Especially javascript  (which is most common client  for graphql) in not most cases  treatment value '10' (string) and 10  (numeric) as same value.

### DIFF
--- a/src/GraphQL/Validation/ValidationContext.cs
+++ b/src/GraphQL/Validation/ValidationContext.cs
@@ -325,7 +325,14 @@ namespace GraphQL.Validation
                     //  example.: type is int. value is '10'. 
                     //     This value can be parsed from string into int and solution is very practical for wide usage.
                     // Maybe best practice should be add new global switch which enable/disable this option.
-                    value = Convert.ChangeType(value, scalarGraphType);
+                    if (SchemaTypes.BuiltInScalarMappings.Where(m => m.Value == scalarGraphType.GetType()).Any())
+                    {
+                        var type = SchemaTypes.BuiltInScalarMappings.Where(m => m.Value == scalarGraphType.GetType()).FirstOrDefault().Key;
+                        if (type != null)
+                        {
+                            value = Convert.ChangeType(value, type);
+                        }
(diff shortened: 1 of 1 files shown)


Shane32:
The migration guide covers this topic here:

https://graphql-dotnet.github.io/docs/migrations/migration4#scalar-deserialization-type-enforcement

I suggest you copy the `IntGraphType` source code, and make your own `MyIntGraphType` class with some additional code in the `ParseValue` method to convert strings to integers.  If you have the need, you can also override `CanParseLiteral` and `ParseLiteral`.

```cs
public class MyIntGraphType : IntGraphType
{
    public MyIntGraphType()
    {
        Name = "Int";
    }

    public override object? ParseValue(object? value)
        => value is string str ? int.Parse(str) : base.ParseValue(value);
}

// Code-first: add the following line of code to your schema's constructor
RegisterType(new MyIntGraphType());

// Schema-first: add the following line of code after your schema is built, before it is initialized
schema.RegisterType(new MyIntGraphType());
```

Shane32:
You should be aware that if you are intending to return/consume identifiers, `IdGraphType` is more appropriate, which will accept and coerce both strings and integers the same as it did in 3.x.

> Especially javascript (which is most common client for graphql) in not most cases treatment value '10' (string) and 10 (numeric) as same value.

Javascript is fully capable of holding numeric values in variables.  If a numeric value was stored as a string in one of my applications, I would consider it a bug.  I use Typescript so that this type of problem is near-impossible.  Typescript also prevents many many other types of bugs, mainly by identifying references that might be null and identifying spelling errors.  But for legacy applications you can still choose to override scalar graph types as described above to loosen the server-side type checks.

JanezKr:
Yes i solved it on client side and cast value in type which is expected by schema.  Thanks @Shane32  for answer.

PR #2507, opened 2021-04-16
DeprecationReason as computed property stored in Metadata (in applieddirectives), also changes " to ' in some error messages

sungam3r:
This PR addresses a problem with duplication of `@deprecated` directive set as `DeprecationReason` property and as applied directive in schema element metadata. Now it is stored in metadata. `DeprecationReason` became computed property. This also saves a bit of memory for each field, type and enum value if they don't have `DeprecationReason` which is expected in the most cases (wasn't a goal, but 😄).  No breaking changes.

### DIFF
--- a/src/GraphQL/Types/GraphType.cs
+++ b/src/GraphQL/Types/GraphType.cs
@@ -83,7 +83,11 @@ namespace GraphQL.Types
         public string Description { get; set; }
 
         /// <inheritdoc/>
-        public string DeprecationReason { get; set; }
+        public string? DeprecationReason
+        {
+            get => this.GetDeprecationReason();
+            set => this.SetDeprecationReason(value);
+        }
 
         /// <inheritdoc />
         public override string ToString() => Name;
--- a/src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
+++ b/src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
@@ -83,7 +83,12 @@ namespace GraphQL.Utilities
                         throw new InvalidOperationException($"Unknown directive '{appliedDirective.Name}'.");
 
                     if (location != null && !schemaDirective.Locations.Contains(location.Value))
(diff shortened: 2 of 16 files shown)


src/GraphQL/Types/GraphType.cs
```diff
@@ -83,7 +83,11 @@ public string Name
         public string Description { get; set; }
 
         /// <inheritdoc/>
-        public string DeprecationReason { get; set; }
+        public string? DeprecationReason
```
Shane32, line 86: NRT annotation in class without NRTs

src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
```diff
@@ -83,7 +83,12 @@ private void ValidateAppliedDirectives(IProvideMetadata provider, ISchema schema
                         throw new InvalidOperationException($"Unknown directive '{appliedDirective.Name}'.");
 
                     if (location != null && !schemaDirective.Locations.Contains(location.Value))
-                        throw new InvalidOperationException($"Directive '{schemaDirective.Name}' is applied in the wrong location '{location.Value}'. Allowed locations: {string.Join(", ", schemaDirective.Locations)}");
+                    {
+                        // TODO: think about strict check; needs to rewrite some tests (5)
```
Shane32, line 87: Add issue to track this for v5?

Shane32 (approved):
See comments also.

src/GraphQL/Types/GraphType.cs
sungam3r: It inherits `string? IProvideDeprecationReason.DeprecationReason { get; set; }`

src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
sungam3r: #2508

src/GraphQL/Types/GraphType.cs
Shane32: Whatever you think. It will just produce compiler warnings for now

sungam3r: If I continue to set NRT annotation, it will become avalanche-like process.

Issue #2518, opened 2021-04-21
PageSize not populating context.First


SimonCropp:
in version 4.2.0. when i set `ConnectionBuilder<TSource>.PageSize(10);`, then in the resolve `context.First` would return 10

in version 4.3.0. when i set `ConnectionBuilder<TSource>.PageSize(10);`, then in the resolve `context.First` now returns null

Any ideas what changed to cause this?

Shane32:
#2458 changed the connection builders' code a bit, but there should not have been any changes in functionality to old code.

Previously `.PageSize(10)` would set a private variable `_pagesize` which was captured by the `FuncFieldResolver` delegate within `.Resolve()` to create `ResolveConnectionContext`.

Now, `PageSize(10)` sets some field metadata, and within `.Resolve()` the delegate instead accesses a property that accesses field metadata to determine the page size.

This can cause two unintended side-effects:
1. If metadata is changed (e.g. cleared) after the resolver is built, it will change the resolver's behavior.
2. Another dictionary lookup within the field's metadata is required each time the resolver executes.

There are similar unintended side-effects of the `.Bidirectional()` method.  We should probably add a PR to create a local copy of these variables within `Resolve` to fix both of the above unintended changes.  However, it would not seem that this should break correct operation of the code, at least under normal circumstances.

Any chance that the field metadata is cleared after the connection resolver is built?

Can you try calling `field.GetMetadata<int?>("__ConnectionBuilder_PageSize")` and verify it returns 10?  I'd probably check this within your resolver to be sure it isn't changed after the fact somehow.

FYI, the reason the metadata was added is so that extension methods to the connection builders can read the page size and whether the connection is configured as bidirectional.

sungam3r:
Fixed in v4.4

Shane32:
As of now, we have not determined that there was a bug in 4.3, or that the changes in 4.4 would fix the issue.  @SimonCropp to confirm findings.

SimonCropp:
`var metadata = context.FieldDefinition.GetMetadata<int?>("__ConnectionBuilder_PageSize")!;`  returned null

here is my usage https://github.com/SimonCropp/GraphQL.EntityFramework/blob/master/src/GraphQL.EntityFramework/GraphApi/EfGraphQLService_QueryableConnection.cs#L92

Shane32:
Maybe SetField needs to move up a line - before the pagesize and bidirectional methods ?

Shane32:
[...]

PR #2519, opened 2021-04-21
Optimize connection builder

Shane32:
Captures IsBidirectional and PageSize properties when creating the connection resolver, preventing a metadata dictionary lookup and enumeration of field arguments each time that the connection is resolved.  This issue did not exist prior to 4.3 so this is a performance regression fix caused by #2458 .

Added test to verify that the PageSize value is utilized.  (There was no test for this previously.)

Also see #2518.

### DIFF
--- a/src/GraphQL.Tests/Builders/ConnectionBuilderTests.cs
+++ b/src/GraphQL.Tests/Builders/ConnectionBuilderTests.cs
@@ -401,12 +401,8 @@ namespace GraphQL.Tests.Builders
             graph.Connection<ChildType>()
                 .Name("connection")
                 .PageSize(10)
-                .Resolve(context =>
-                {
-                    context.First.ShouldBe(10);
-                    return null;
-                });
-            graph.Fields.Find("connection").Resolver.Resolve(new ResolveFieldContext());
+                .Resolve(context => context.First);
+            graph.Fields.Find("connection").Resolver.Resolve(new ResolveFieldContext()).ShouldBe(10);
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


Shane32:
Let’s merge this before 4.4

src/GraphQL.Tests/Builders/ConnectionBuilderTests.cs
```diff
@@ -394,6 +394,28 @@ public void bidirectional_called_twice_creates_proper_arguments()
...
         }
 
+        [Fact]
+        public void should_use_pagesize()
+        {
+            var graph = new ObjectGraphType();
+            graph.Connection<ChildType>()
+                .Name("connection")
+                .PageSize(10)
+                .Resolve(context => context.First);
```
sungam3r, line 404: What if you swap these lines?

sungam3r: Setting `PageSize` and other property after calling `Resolve` now does nothing.

Shane32: This is not possible in the connection builders included in 4.3 because Resolve returns void.  That's probably why they returned void.

The typed connection builders return the builder, but they have not been released yet.  I suggest altering them to return void.

I really don't like the fact that every time a connection is resolved it evaluates both these lines:

```cs
FieldType.GetMetadata<int?>(ConnectionBuilder<TSourceType>.PAGE_SIZE_METADATA_KEY);

FieldType.Arguments.Find("before")?.Type == typeof(StringGraphType) && FieldType.Arguments.Find("last")?.Type == typeof(IntGraphType);
```

which is a dictionary lookup and two list enumerations every time that the resolver executes.  This did not occur in 4.2.

sungam3r: OK, then make it `void` and put some notes into docs/xml.

sungam3r: Although this approach looks strange and will cause other questions.

Shane32: I was wrong; the typed connection builders are included in 4.3 and do return the connection builder.  It is the scoped connection builders that did not make it into 4.3.  The untyped connection builders from 4.2 (and also in 4.3) return void.

In this case I would make a breaking change and change the typed connection builders to return void.  4.3 is only 5 days old.  We just need to note the change in the release notes.

I have made the changes as described and are ready for your review.

sungam3r (approved):

Issue #2510, opened 2021-04-18
Azure functions v3, .Net Core 5 : Variable '$dish' is invalid. Unable to parse input as a 'DishInputType' type. Did you provide a List or Scalar value accidentally?


tsansome:
## Description
I'm trying to get my mutations working but always prompted with this error. Can't seem to find what's wrong.

> Variable '$dish' is invalid. Unable to parse input as a 'DishInputType' type. Did you provide a List or Scalar value accidentally?

Any suggestions? I've breakpointed and checked that it definitely parses to the ~~Dictionary<string, object> and object is a JObject.~~
```
System.Collections.Generic.Dictionary`2[System.String,System.Object]
```

### Steps to reproduce

Clear steps describing how to reproduce the issue.

Using Postman v7.36.5. Send the Http Post request with JSON as content type to trigger my mutation to create a new dish.

```
{
  "query": "mutation ($dish:DishInputType!){ createDish(dish: $dish) { name} }",
  "variables": {
    "dish": {
        "name": "Pad See Ew Chicken"
    }
  }
}
```

### Actual result
Doesn't enter into resolver but instead states error
"Variable '$dish' is invalid. Unable to parse input as a 'DishInputType' type. Did you provide a List or Scalar value accidentally?"

### Environment
Azure functions v3
.Net Core 5

other packages
```
<PackageReference Include="GraphQL" Version="4.2.0" />
    <PackageReference Include="GraphQL.SystemTextJson" Version="4.2.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.12" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.1" OutputItemType="Analyzer" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.0.0" />
    <PackageReference Include="SqlKata" Version="2.3.2" />
    <PackageReference Include="SqlKata.Execution" Version="2.3.2" />
```

### Graphql main
```
var logger = executionContext.GetLogger("graphql");
            logger.LogInformation("C# HTTP trigger function processed a request.");

            var schema = new Schema { 
                Query = new Query(), 
                Mutation= new Mutation() };

            var request = await JsonSerializer.DeserializeAsync<GraphQLRequest>(req.Body, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
        
            var executer = new DocumentExecuter();
[...]

PR #2521, opened 2021-04-21
Add additional deserialization extensions to GraphQL

Shane32:
See #2510 

- `FromJson` and `FromJsonAsync` are new methods for `GraphQL.SystemTextJson`.  They cannot be implemented for Newtonsoft due to its nature.  Or least, I'm not sure how at the moment.
- `ToInputs(this JsonElement element)` is a new method in `GraphQL.SystemTextJson`.  This mimics a similar method within `GraphQL.NewtonsoftJson` that already exists there (as `ToInputs(this JObject obj)`).  It uses a local `GetValue` function.  There exists a similar `GetValue` function in the `GraphQL.NewtonsoftJson` library for the same purpose which is public.  I do not see why this needs to be public, so it remains private in `GraphQL.SystemTextJson`.

### DIFF
--- a/src/GraphQL.Harness/GraphQLMiddleware.cs
+++ b/src/GraphQL.Harness/GraphQLMiddleware.cs
@@ -1,12 +1,12 @@
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using GraphQL;
 using GraphQL.Instrumentation;
-using GraphQL.Types;
 using GraphQL.SystemTextJson;
+using GraphQL.Types;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
-using System.Threading;
 
 namespace Example
 {
--- a/src/GraphQL.Harness/GraphQLMiddleware.cs
+++ b/src/GraphQL.Harness/GraphQLMiddleware.cs
@@ -1,11 +1,12 @@
 using System;
-using System.Text.Json;
 using System.Threading.Tasks;
 using GraphQL;
 using GraphQL.Instrumentation;
 using GraphQL.Types;
+using GraphQL.SystemTextJson;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
+using System.Threading;
 
 namespace Example
 {
@@ -50,11 +51,7 @@ namespace Example
         {
             var start = DateTime.UtcNow;
 
-            var request = await JsonSerializer.DeserializeAsync<GraphQLRequest>
-            (
-                context.Request.Body,
-                new JsonSerializerOptions { PropertyNameCaseInsensitive = true }
-            );
+            var request = await context.Request.Body.FromJsonAsync<GraphQLRequest>(context.RequestAborted);
 
             var result = await _executer.ExecuteAsync(options =>
             {
@@ -65,6 +62,7 @@ namespace Example
                 options.UserContext = _settings.BuildUserContext?.Invoke(context);
                 options.EnableMetrics = _settings.EnableMetrics;
                 options.RequestServices = context.RequestServices;
+                options.CancellationToken = context.RequestAborted;
             });
 
             if (_settings.EnableMetrics)
@@ -72,15 +70,15 @@ namespace Example
                 result.EnrichWithApolloTracing(start);
             }
 
-            await WriteResponseAsync(context, result);
+            await WriteResponseAsync(context, result, context.RequestAborted);
         }
 
-        private async Task WriteResponseAsync(HttpContext context, ExecutionResult result)
+        private async Task WriteResponseAsync(HttpContext context, ExecutionResult result, CancellationToken cancellationToken)
         {
             context.Response.ContentType = "application/json";
             context.Response.StatusCode = 200; // OK
 
-            await _writer.WriteAsync(context.Response.Body, result);
+            await _writer.WriteAsync(context.Response.Body, result, cancellationToken);
         }
     }
 }
--- a/src/GraphQL.SystemTextJson/StringExtensions.cs
+++ b/src/GraphQL.SystemTextJson/StringExtensions.cs
@@ -63,7 +63,7 @@ namespace GraphQL.SystemTextJson
         public static Inputs ToInputs(this JsonElement obj)
         {
             if (obj.ValueKind == JsonValueKind.Null)
-                return null;
+                return Inputs.Empty;
 
             if (obj.ValueKind != JsonValueKind.Object)
                 throw new InvalidOperationException("This element is not an object element");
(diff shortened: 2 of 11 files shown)


src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -88,7 +88,7 @@ public class DocumentValidator : IDocumentValidator
 
                 if (!rules.Any())
                 {
-                    variables = context.GetVariableValues(schema, variableDefinitions, inputs); // can report errors even without rules enabled
+                    variables = context.GetVariableValues(schema, variableDefinitions, inputs ?? Inputs.Empty); // can report errors even without rules enabled
```
Shane32, line 91: Bug fix.  Without the fix it can cause a NRE if variables are specified in the document but null was provided to inputs.  Already correct in the `else` condition of the above `if (!rules.Any())`.

https://github.com/graphql-dotnet/graphql-dotnet/blob/d9b96431d9962d6f0d1dd1070da5e4546522e065/src/GraphQL/Validation/DocumentValidator.cs#L128

src/GraphQL.Tests/Serialization/NewtonsoftJsonTests.cs
```diff
@@ -0,0 +1,101 @@
...
+using GraphQL.NewtonsoftJson;
+using Newtonsoft.Json.Linq;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests.Serialization
+{
+    public class NewtonsoftJsonTests
+    {
+        private readonly object _example = new
```
sungam3r, line 14: Why object?

sungam3r:
Mark `GetValue` from Newtonsoft as obsolete.

src/GraphQL.Tests/Serialization/NewtonsoftJsonTests.cs
sungam3r: And `dynamic` below on line 98.

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: Seems so.

src/GraphQL.Tests/Serialization/NewtonsoftJsonTests.cs
Shane32: var doesn't work here.  I can copy the type def from the other class...

Shane32: done

Shane32:
> Mark `GetValue` from Newtonsoft as obsolete.

done

src/GraphQL.Harness/GraphQLMiddleware.cs
```diff
@@ -50,11 +51,7 @@ private async Task ExecuteAsync(HttpContext context, ISchema schema)
         {
             var start = DateTime.UtcNow;
 
-            var request = await JsonSerializer.DeserializeAsync<GraphQLRequest>
-            (
-                context.Request.Body,
-                new JsonSerializerOptions { PropertyNameCaseInsensitive = true }
```
sungam3r, line 56: `PropertyNameCaseInsensitive = true` was lost

sungam3r: I'm OK with this change in general. It's just a sample program.

src/GraphQL.SystemTextJson/StringExtensions.cs
```diff
@@ -36,5 +40,81 @@ public static Inputs ToInputs(this string json)
...
+        /// Any <see cref="Dictionary{TKey, TValue}">Dictionary&lt;string, object&gt;</see> objects will be deserialized
+        /// into the proper format for passing to <see cref="InputsExtensions.ToInputs(Dictionary{string, object})"/>.
+        /// </summary>
+        public static ValueTask<T> FromJsonAsync<T>(this System.IO.Stream stream, CancellationToken cancellationToken = default)
+            => JsonSerializer.DeserializeAsync<T>(stream, _jsonOptions, cancellationToken);
+
+        /// <summary>
+        /// Converts a JSON element into an <see cref="Inputs"/> object.
+        /// </summary>
+        public static Inputs ToInputs(this JsonElement obj)
```
sungam3r, line 63: The corresponding method from Newtonsoft project returns `Inputs.Empty` if `obj` is `null`.

Shane32: Updated that and also added tests.

sungam3r: If `obj` is `null` then this method throws NRE.

src/GraphQL.SystemTextJson/StringExtensions.cs
```diff
@@ -36,5 +40,81 @@ public static Inputs ToInputs(this string json)
...
+        {
+            if (obj.ValueKind == JsonValueKind.Null)
```
sungam3r, line 65:
```suggestion
            if (obj == null)
```

sungam3r (approved):
Please make final edits, merge and publish v4.4, I'm done for today.

src/GraphQL.SystemTextJson/StringExtensions.cs
Shane32: JsonElement is a struct.  It can't be null.

Shane32: Right now if `default` was passed in to the method, it would execute this code:
```cs
            if (obj.ValueKind != JsonValueKind.Object)
                throw new InvalidOperationException("This element is not an object element");
```

Shane32: But it does check for a `null` element and the test show this works correctly.

src/GraphQL.SystemTextJson/StringExtensions.cs
Shane32: n/a - `JsonElement` is a struct here.

src/GraphQL.SystemTextJson/StringExtensions.cs
Shane32: Note that for Newtonsoft.Json, the object is a class and would actually be `null`.  This is properly tested for in the tests I added.

Shane32:
> Please make final edits, merge and publish v4.4, I'm done for today.

:+1:

I'd like you to review #2522 before publishing v4.4 if you don't mind.  It's all done.  No big deal if you want to publish v4.4 in the morning though and let it go to v4.5.  But it is also part of the serializers.

src/GraphQL.SystemTextJson/StringExtensions.cs
sungam3r: > JsonElement is a struct. It can't be null.

🤦‍♂️

src/GraphQL.SystemTextJson/StringExtensions.cs
sungam3r: 🤦‍♂️

PR #2522, opened 2021-04-21
Add classes to deserialize Inputs objects directly

Shane32:
- Enhances `FromJson` methods of `GraphQL.SystemTextJson` to parse `Inputs` objects
- Updates Harness sample to demonstrate the simpler design.
- Adds `FromJson` methods to `GraphQL.NewtonsoftJson` with similar signatures which will parse `Inputs` objects (but does not parse `Dictionary<string, object>` objects).
- Updates tests to verify all above functionality

Suggested additional change:
- Mark `GraphQL.SystemTextJson.ObjectDictionaryConverter` as obsolete, with `InputsConverter` as the suggested replacement. This will unify the supported features of both serialization engines.  I will submit a separate PR for us to consider this change.

Also:
- Need to add serialization documentation (separate PR).

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
@@ -32,6 +32,7 @@ namespace GraphQL.NewtonsoftJson
     }
     public static class StringExtensions
     {
+        [System.Obsolete("This method will be removed in a future version of GraphQL.NET.")]
         public static object GetValue(this object value) { }
         public static System.Collections.Generic.Dictionary<string, object> ToDictionary(this string json) { }
         public static GraphQL.Inputs ToInputs(this Newtonsoft.Json.Linq.JObject obj) { }
--- a/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
@@ -26,12 +26,23 @@ namespace GraphQL.NewtonsoftJson
         public override object ReadJson(Newtonsoft.Json.JsonReader reader, System.Type objectType, object existingValue, Newtonsoft.Json.JsonSerializer serializer) { }
         public override void WriteJson(Newtonsoft.Json.JsonWriter writer, object value, Newtonsoft.Json.JsonSerializer serializer) { }
     }
+    public class InputsConverter : Newtonsoft.Json.JsonConverter
+    {
+        public static readonly GraphQL.NewtonsoftJson.InputsConverter Instance;
+        public InputsConverter() { }
+        public override bool CanWrite { get; }
+        public override bool CanConvert(System.Type objectType) { }
+        public override object ReadJson(Newtonsoft.Json.JsonReader reader, System.Type objectType, object existingValue, Newtonsoft.Json.JsonSerializer serializer) { }
+        public override void WriteJson(Newtonsoft.Json.JsonWriter writer, object value, Newtonsoft.Json.JsonSerializer serializer) { }
+    }
     public static class SchemaExtensions
     {
         public static System.Threading.Tasks.Task<string> ExecuteAsync(this GraphQL.Types.ISchema schema, System.Action<GraphQL.ExecutionOptions> configure) { }
     }
     public static class StringExtensions
     {
+        public static T FromJson<T>(this string json) { }
+        public static System.Threading.Tasks.Task<T> FromJsonAsync<T>(this System.IO.Stream stream) { }
         [System.Obsolete("This method will be removed in a future version of GraphQL.NET.")]
         public static object GetValue(this object value) { }
         public static System.Collections.Generic.Dictionary<string, object> ToDictionary(this string json) { }
--- a/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
@@ -28,7 +28,6 @@ namespace GraphQL.NewtonsoftJson
     }
     public class InputsConverter : Newtonsoft.Json.JsonConverter
     {
-        public static readonly GraphQL.NewtonsoftJson.InputsConverter Instance;
         public InputsConverter() { }
         public override bool CanWrite { get; }
         public override bool CanConvert(System.Type objectType) { }
--- a/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
@@ -41,7 +41,7 @@ namespace GraphQL.NewtonsoftJson
     public static class StringExtensions
     {
         public static T FromJson<T>(this string json) { }
-        public static System.Threading.Tasks.Task<T> FromJsonAsync<T>(this System.IO.Stream stream) { }
+        public static T FromJson<T>(this System.IO.Stream stream) { }
         [System.Obsolete("This method will be removed in a future version of GraphQL.NET.")]
         public static object GetValue(this object value) { }
         public static System.Collections.Generic.Dictionary<string, object> ToDictionary(this string json) { }
--- a/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
@@ -40,8 +40,8 @@ namespace GraphQL.NewtonsoftJson
     }
     public static class StringExtensions
     {
-        public static T FromJson<T>(this string json) { }
         public static T FromJson<T>(this System.IO.Stream stream) { }
+        public static T FromJson<T>(this string json) { }
         [System.Obsolete("This method will be removed in a future version of GraphQL.NET.")]
         public static object GetValue(this object value) { }
         public static System.Collections.Generic.Dictionary<string, object> ToDictionary(this string json) { }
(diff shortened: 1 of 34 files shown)


src/GraphQL.NewtonsoftJson/InputsConverter.cs
```diff
@@ -0,0 +1,96 @@
...
+            => reader.TokenType switch
+            {
+                JsonToken.StartArray => ReadArray(reader),
+                JsonToken.StartObject => ReadDictionary(reader),
+                JsonToken.Integer => ReadNumber(reader),
+                JsonToken.Float => reader.Value,
+                JsonToken.Boolean => reader.Value,
+                JsonToken.String => reader.Value,
+                JsonToken.Null => null,
+                JsonToken.None => null,
```
sungam3r, line 54:
```suggestion
```

 //
        // Summary:
        //     This is returned by the Newtonsoft.Json.JsonReader if a read method has not been
        //     called.
        None = 0,

src/GraphQL.NewtonsoftJson/InputsConverter.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// This JSON converter does not support writing.
+        /// </summary>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+            => throw new NotImplementedException();
+
+        /// <inheritdoc/>
+        public override bool CanWrite => false;
+
+        /// <inheritdoc/>
+        public override bool CanConvert(Type objectType) => typeof(Inputs).IsAssignableFrom(objectType);
```
sungam3r, line 94: Why `IsAssignableFrom`? It makes sense for writing derived values but not for reading them. You always create `Inputs` instance above, not derived one.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -11,6 +12,8 @@ namespace GraphQL.NewtonsoftJson
     /// </summary>
     public static class StringExtensions
     {
+        private static readonly InputsConverter _inputsConverter = new();
```
sungam3r, line 15: I'm not a fan of this new C#9 feature.

src/GraphQL.NewtonsoftJson/InputsConverter.cs
Shane32: Right it should just be equals. I’m falling asleep now though.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
Shane32: Really? I think it’s pretty neat. Prevents the type from being specified twice. Sorta like var.

Shane32:
If that’s all your comments, update it as desired and merge it. I’m half asleep. Otherwise I’ll do it tomorrow

src/GraphQL.NewtonsoftJson/StringExtensions.cs
sungam3r: I know but it looks a bit weird.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -50,6 +53,38 @@ public static Inputs ToInputs(this JObject obj)
             return GetValueInternal(values) as Dictionary<string, object>;
         }
 
+        /// <summary>
+        /// Deserializes a JSON-formatted string of data into the specified type.
+        /// Any <see cref="Inputs"/> objects will be deserialized into the proper format.
+        /// </summary>
+        public static T FromJson<T>(this string json)
+        {
+            var settings = new JsonSerializerSettings
```
sungam3r, line 62: This can be moved into static JsonSerializer field:
```
var settings = new JsonSerializerSettings
            {
                DateParseHandling = DateParseHandling.None
            }
 settings.Converters.Add(_inputsConverter);
 JsonSerializer jsonSerializer = JsonSerializer.Create(settings); <-- jsonSerializer  may be static
```
`JsonConvert.DeserializeObject` is just a wrapper that allocates.

sungam3r: The same for `ToDictionary(this string json)`

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -50,6 +53,38 @@ public static Inputs ToInputs(this JObject obj)
...
+        }
+
+        /// <summary>
+        /// Deserializes a JSON-formatted stream of data into the specified type.
+        /// Any <see cref="Inputs"/> objects will be deserialized into the proper format.
+        /// </summary>
+        public static async Task<T> FromJsonAsync<T>(this System.IO.Stream stream)
+        {
+            using var streamReader = new System.IO.StreamReader(stream ?? throw new ArgumentNullException(nameof(stream)), System.Text.Encoding.UTF8, false, 1024, true);
+            var text = await streamReader.ReadToEndAsync();
```
sungam3r, line 78: It defeats the purpose of stream. See https://www.newtonsoft.com/json/help/html/DeserializeWithJsonSerializerFromFile.htm and use StreamReader instead with JsonSerializer.

src/GraphQL.Tests/Serialization/SystemTextJsonTests.cs
```diff
@@ -119,6 +143,12 @@ private class TestClass2
             public JsonElement variables { get; set; }
         }
 
+        private class TestClass3
+        {
+            public string query { get; set; }
```
sungam3r, line 148: Try to check that `Query` and `Variables` are deserialized too.

sungam3r:
I will publish v4.4 without that PR. OK for v4.5

src/GraphQL.NewtonsoftJson/InputsConverter.cs
Shane32: Fixed

src/GraphQL.NewtonsoftJson/StringExtensions.cs
Shane32: Well, it does read asynchronously, at the cost of buffering the whole input.  I made the suggested change and changed the method to a synchronous method here.

src/GraphQL.Tests/Serialization/SystemTextJsonTests.cs
Shane32: Done, and also added tests to verify case sensitivity behavior.  However, the default behavior differs between System.Text.Json and Newtonsoft.Json.  I opted to leave them as the default for the platform, as that would be expected behavior.  I did add additional xml comments.

Shane32:
Also fixed a couple issues dealing with null/missing `Inputs` fields and added tests for that as well.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
Shane32: Done

src/GraphQL.NewtonsoftJson/StringExtensions.cs
Shane32: This line does not exist anymore.

Shane32:
> I will publish v4.4 without that PR. OK for v4.5

Now I want to make a change to the FromJson definition for System.Text.Json but it is too late without it being a breaking change.

src/GraphQL.NewtonsoftJson/InputsConverter.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// This JSON converter does not support writing.
+        /// </summary>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+            => throw new NotImplementedException();
+
+        /// <inheritdoc/>
+        public override bool CanWrite => false;
+
+        /// <inheritdoc/>
+        public override bool CanConvert(Type objectType) => typeof(Inputs) == objectType;
```
sungam3r, line 96:
```suggestion
        public override bool CanConvert(Type objectType) => objectType == typeof(Inputs);
```
https://en.wikipedia.org/wiki/Yoda_conditions

Shane32: Seriously?

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -41,15 +51,39 @@ public static Inputs ToInputs(this JObject obj)
         /// <returns>Returns a <c>null</c> if the object cannot be converted into a dictionary.</returns>
```
sungam3r, line 51: Is it still true?

src/GraphQL.NewtonsoftJson/InputsConverter.cs
Shane32: If you want to reorder the expression feel free to do so. I’m not going to. It’s a complete waste of time.

src/GraphQL.ApiTests/GraphQL.NewtonsoftJson.approved.txt
```diff
@@ -26,12 +26,22 @@ namespace GraphQL.NewtonsoftJson
...
+        public override void WriteJson(Newtonsoft.Json.JsonWriter writer, object value, Newtonsoft.Json.JsonSerializer serializer) { }
+    }
     public static class SchemaExtensions
     {
         public static System.Threading.Tasks.Task<string> ExecuteAsync(this GraphQL.Types.ISchema schema, System.Action<GraphQL.ExecutionOptions> configure) { }
     }
     public static class StringExtensions
     {
+        public static T FromJson<T>(this string json) { }
+        public static System.Threading.Tasks.Task<T> FromJsonAsync<T>(this System.IO.Stream stream) { }
```
sungam3r, line 44:
```suggestion
        public static T FromJson<T>(this System.IO.Stream stream) { }
```

src/GraphQL.NewtonsoftJson/StringExtensions.cs
Shane32: It is as true or false as it was before (lol).

Shane32: It looked like invalid json would throw while a different type of json element would return null. I didn’t change that functionality.

Shane32:
Apiapprovals have been fixed (whoops!)

sungam3r (approved):

PR #2526, opened 2021-04-23
Mark ObjectDictionaryConverter as obsolete

Shane32:
- Mark `SystemTextJson.ObjectDictionaryConverter` as obsolete, as the `InputsConverter` has replaced it
- Mark `ToDictionary` as obsolete as `ToInputs` has replaced it
- Enable writing for `SystemTextJson.InputsConverter` (although it is not necessary to use it since `Inputs` implements `IReadOnlyDictionary<string, object>`), and enable tests
- Duplicate `SystemTextJson.ObjectDictionaryConverter` tests to `InputsConverter` tests for both `SystemTextJson` and `NewtonsoftJson` serialization engines.

This PR concludes my suggested changes to the serialization engines. (Also see #2522 and #2521). Updated documentation to follow in a separate PR.

PR #2530, opened 2021-04-28
Remove deprecated functionaliaty from the new SystemTextJson.FromJson methods

Shane32:
History:
- In PR #2521 we added `FromJson` extension methods to `SystemTextJson`, which deserializes `Dictionary<string, object>` .  They were not added to `NewtonsoftJson`.
- These methods were released with v4.4.0 on Apr 21.
- Then in #2522 we added `FromJson` extension methods to `NewtonsoftJson` along with `Inputs` parsing for both serialization engines.  The `Newtonsoft.FromJson` method does not deserialize `Dictionary<string, object>` types.
- Finally in #2526 we deprecated deserialization of `Dictionary<string, object>` types.

I suggest we remove the deprecated functionality from `SystemTextJson.FromJson`, unifying the method with the functionality with `NewtonsoftJson.FromJson`.  This functionality is a breaking change, but has only been available for the last 7 days within 4.4.0.  I think we should then release 4.5.0

Deserialization of `Dictionary<string, object>` types will still be available via `ToDictionary` extension methods and `ObjectDictionaryConverter` during the lifetime of v4.  This PR is only a breaking change from 4.4.0, which was released 7 days ago.

The new release (v4.5.0) will contain all of the changes to the serialization engines made over the last couple weeks:
- `ToDictionary` added for `SystemTextJson` (matching existing `ToDictionary` for `NewtonsoftJson`)
- `InputsConverter` added for both `SystemTextJson` and `NewtonsoftJson`
- `FromJson` added for both `SystemTextJson` and `NewtonsoftJson` which interprets `Inputs` values correctly
- `ObjectDictionaryConverter` deprecated for `SystemTextJson`
- `ToDictionary` deprecated for both `SystemTextJson` and `NewtonsoftJson`

Then I will work on adding documentation (which of course does not require a release).

### DIFF
--- a/src/GraphQL.SystemTextJson/StringExtensions.cs
+++ b/src/GraphQL.SystemTextJson/StringExtensions.cs
@@ -41,7 +41,7 @@ namespace GraphQL.SystemTextJson
         /// <param name="json">A JSON formatted string.</param>
         /// <returns>Inputs.</returns>
         public static Inputs ToInputs(this string json)
-            => json != null ? JsonSerializer.Deserialize<Inputs>(json, _jsonOptions) : Inputs.Empty;
+            => json != null ? JsonSerializer.Deserialize<Inputs>(json, _jsonOptions2) : Inputs.Empty;
 
         /// <summary>
         /// Converts a JSON-formatted string into a dictionary of objects of their actual type.
--- a/src/GraphQL.SystemTextJson/StringExtensions.cs
+++ b/src/GraphQL.SystemTextJson/StringExtensions.cs
@@ -25,6 +25,16 @@ namespace GraphQL.SystemTextJson
             }
         };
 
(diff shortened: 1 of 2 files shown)


src/GraphQL.SystemTextJson/StringExtensions.cs
```diff
@@ -25,6 +25,16 @@ public static class StringExtensions
             }
         };
 
+        private static readonly JsonSerializerOptions _jsonOptions2 = new JsonSerializerOptions
```
sungam3r, line 28: Why new field ?

Shane32: The other one is used by `ToDictionary`.

sungam3r (approved):

Shane32:
I’ll release 4.5.0 if you have no objection @sungam3r

sungam3r:
ok

Issue #1320, opened 2019-10-01
IAsyncEnumerable


MikeKenyon:
###  Summary

.NET Core 3.0/Standard 2.1 introduces the IAsyncEnumerable<T> type in order to support async pulls of data back from remote servers in efficient ways.  A typical pattern would feel to return an IAsyncEnumerable<T> back when performing an async lookup as this is doable for a single entity, for example.  However, the executor code [here](https://github.com/graphql-dotnet/graphql-dotnet/blob/952b4ef6950e51bbfe009354f0520180c343e8d4/src/GraphQL/Execution/ExecutionStrategy.cs#L97) is such that it checks for derivation from IEnumerable and not from IAsyncEnumerable and there is no common parentage between the two.

### Basic example

The following code does not properly execute when the returned type of `manager.FindAsync()` is an `IAsyncEnuerable`.
```
            Field<ListGraphType<Graph.MediaEntity>>("findMedia",
                arguments: new QueryArguments
                {
                    new QueryArgument<MediaType>{
                        Name = "type",
                        Description = "The type of media to search for."
                    },
                    new QueryArgument<NonNullGraphType<StringGraphType>>
                    {
                        Name = "query",
                        Description = "The text to search for."
                    }
                },
                resolve: c =>
                {
                    Models.MediaType? type = null;
                    if (c.HasArgument("type"))
                    {
                        type = c.GetArgument<Models.MediaType>("type");
                    }
                    var query = c.GetArgument<string>("query");
                    return manager.FindAsync(query, type);
                });

```
`
### Motivation

While this feature is new, it's a fairly impressive performance gain over prior execution patterns and likely to obtain significant attention.  This is particularly true among database calls and web service calls which are likely to be exposed in this fashion.  

Further, it seems to be working properly when the query in question has no arguments.   This discrepancy presents itself as a fragility in an otherwise stable system.

MikeKenyon:
Current workaround I've come up with this.  It's not at all pretty, but does work.   I get around it with an extension method ... 
```
                    return media.FindAsync(query, type).ToEnumerable().Result;

        public static async Task<IEnumerable<T>> 
            ToEnumerable<T>(this IAsyncEnumerable<T> listing)
        {
            var list = new List<T>();

            await foreach(var entry in listing)
            {
                list.Add(entry);
            }
            return list;
        }

```

Not at all performant, but it does get you past the problem.

sungam3r:
[...]

PR #2532, opened 2021-05-03
WIP: Add support for IAsyncEnumerable (#1320)

kuzn-ilya:


Shane32:
This is a pretty good implementation of basic support of an `IAsyncEnumerable`.  It validates nodes as they are being received and it does not require a `List<T>` to be created prior to the `List<ExecutionNode>` being created.  But that's about it.  Child nodes still cannot execute until the complete list has been received, so the performance increase is small at best.  Due to other issues (see below), I would guess that it runs slower for small result sets, and slightly better for very large result sets.  Certainly a minor difference either way.

Let's consider some other issues.  First, using reflection to examine supported interfaces is relatively slow, and as coded here, creates use of LINQ on the 'hot path' causing extra allocations and execution time.  `IAsyncEnumerable` support could be configured to be a fallback if `IEnumerable` is not matched, but this would not work for linq2db whose `IQueryable` objects support both `IAsyncEnumerable<T>` and `IEnumerable` interfaces.

Second, constructing a generic method is extremely slow -- a few orders of magnitude.  A concurrent dictionary should be used to cache these delegates.  Even after the method is cached, this does cause a dictionary lookup, which consumes a measurable amount of time, even if small.  This is just yet another reason why using `ToListAsync` before the call to GraphQL.NET will almost certainly perform better than this PR for small data sets.

Third, public API method signatures changed, so it could not be implemented until GraphQL.NET v5.

To solve some of these issues, I would investigate adding an interface and/or wrapper for `IAsyncEnumerable`s to make it more of an opt-in scenario.  A wrapper is a single allocation, but one is certainly less than the many allocations currently needed to check for `IAsyncEnumerable` support of an object.  So then the execution code could be something like `return is IAsyncResult resultAsync`.  Field builders can detect `IAsyncEnumerable` types (when strongly typed) and automatically use the correct field resolver and/or wrapper.  The wrapper can implement `IAsyncEnumerable<object>` so that there is no need to use reflection during execution.

Note that all of the above is true regardless of support of `IAsyncEnumerable` via `Microsoft.Bcl.AsyncInterfaces` for .NET Standard 2.0.  As of now, I still do not think it would be worth adding support to .NET Standard 2.0 projects at the cost of the extra dependency.  There is so little to be gained even if all the issues above were fixed and/or optimized.  Any new projects would likely be built on .NET Standard 2.1 anyway (which is .NET Core 3.0+).  However, as a counter-argument, it could also be said that old projects would likely continue using an old version of GraphQL.NET, and so it would not matter if `Microsoft.Bcl.AsyncInterfaces` was required for .NET Standard 2.0 within GraphQL.NET v5.

In conclusion, I would not recommend merging this PR (for v5) until benchmarks can confirm that there is no/negligible speed and memory difference for non-async queries.  I would also expect that we should investigate revising the execution strategy to start executing nodes as they are received from the source before making this change.

sungam3r:
Agree with @Shane32 . In general, the support of `IAsyncEnumerable` is not so easy. In addition, if we go in this direction, then sure without the use of reflection and LINQ. I close this PR because there was no feedback for the last comment for almost a year.

PR #2546, opened 2021-05-13
Correctly serialize Uris.

ddunkin:
`Uri.ToString()` may return a corrupt URI. See https://faithlife.codes/blog/2010/08/uritostring_must_die/.


Shane32 (approved):
@sungam3r Any comments with this change? It seems like we certainly should not be using `ToString()` for the `Uri` type.  I would suggest incrementing the minor version number as this could potentially break user code if it relied on unescaped urls being returned.  But to me this seems like the correct design.

sungam3r (approved):

sungam3r:
I'm OK either way.

Issue #2552, opened 2021-05-20
System.Text.Json ExecutionResultJsonConverter not serializing Extensions with provided JsonSerializerOptions


kyleadams-fm:
## Description

System.Text.Json ExecutionResultJsonConverter not serializing Extensions object based on the provided JsonSerializerOptions.

### Steps to reproduce

Using JsonSerializerOptions with a camelCase naming strategy 
```
var options = new JsonSerializerOptions
{
    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
    PropertyNameCaseInsensitive = true,
    DictionaryKeyPolicy = null,
    IgnoreNullValues = true,
    AllowTrailingCommas = true
};
```

Serialize an error object through WriteErrors with object in the extensions 
```
internal class ValidationViolation
{
    public string Message { get; }
    public string Property { get; }

    public ValidationViolation(string message, string property)
    {
        Message = message;
        Property = property;
    }
}
```

I believe this is due to the fact that options is not passed to the serializer at this point 
https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs#L274

### Expected result

With this I would expect to see this object serialized as 
``` 
{
     "message": "....",
     "property": "....."
}
```

### Actual result

Serialization is 
```
{
  "Message": "User with given Email already exists",
  "Property": "Email"
}
```

Shane32:
I agree. Do you want to make a PR with a fix?

PR #2553, opened 2021-05-21
Serialize Error Extensions in System.Text.Json with provided options

kyleadams-fm:
Extensions in errors should be serialized along with other values in the ExecutionResultJsonConverter

Addresses issue: https://github.com/graphql-dotnet/graphql-dotnet/issues/2552

### DIFF
--- a/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
+++ b/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
@@ -271,7 +271,7 @@ namespace GraphQL.SystemTextJson
                 if (info.Extensions?.Count > 0)
                 {
                     writer.WritePropertyName("extensions");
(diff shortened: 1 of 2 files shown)


sungam3r:
@kyleadams-fm Thanks for fix. To confirm the error correction, we usually add a test. As I understand, it will not take much time. After that, I will be glad to take your change.

kyleadams-fm:
Cheers @sungam3r, I've added a test for this case

Shane32 (approved):

kyleadams-fm:
@sungam3r can I get approval on the workflows above? Looks like LGTM fails on all PRs when trying to build?

Shane32:
> @sungam3r can I get approval on the workflows above? Looks like LGTM fails on all PRs when trying to build?

Yeah just ignore LGTM

Issue #2557, opened 2021-05-27
Why the `IsValueType` check?


kimipsen:
When using `context.GetArgument<TType>(name, defaultValue)` with non-value types (such as `List<string>`), the default value is ignored. Why is this check there? Is it needed for anything? 

https://github.com/graphql-dotnet/graphql-dotnet/blob/3c6749f5cc2dd87735ce27571f0765b8501b6d03/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs#L28

Shane32:
It looks wrong to me. It is there most likely because otherwise getargument could not distinguish between an unsupplied argument and a supplied value of null. It should be changed to use defaultValue when the value is not supplied by the caller or a default argument value, regardless of whether it is a value type or not.

Shane32:
But if that's the reason, then it begs the question: should a value of null passed to an argument of a value type be superseded by `defaultValue`?  Obviously yes (considering the existing design).  So ... then why not simply remove the IsValueType check?

kimipsen:
@Shane32 Removing the check would work for me just as fine as adding a check on `FieldDefault`.
Based on the Star Wars sample, the following code shows an issue we're currently looking at (v4.5.0):

```csharp
Connection<CharacterInterface>()
                .Name("characters")
                .Argument<ListGraphType<EpisodeEnum>>("excludedEpisodes", "Which episodes to exclude")
                .Resolve(context =>
                {
                    var excludedEpisodes = context.GetArgument("excludedEpisodes", new List<Episodes>());
                    return null; //irrelevant for sample
                });
```

`excludedeEpisodes` is null when I would expect an empty list.

Shane32:
Keep in mind that `context.GetArgument("excludedEpisodes") ?? new List<Episodes>()` will perform identically.

kimipsen:
True, but considering our codebase - that we finally got around to upgrade from 2.4 - is at ~1500 classes in the grapql layer, we're not very keen on going through each file to check for this potential issue :)

Shane32:
Of course.  Is this a regression from 2.4 ?

kimipsen:
Yes, we had this code working up until this upgrade.

Shane32:
Note: Version 3.x had the same `IsValueType` check.

Shane32:
@kimipsen I've added a PR - #2559 .  I'd like @sungam3r to review the PR before it gets merged.

PR #2558, opened 2021-05-27
Add check for `ArgumentValue` source

kimipsen:
#2557 
If we always have an  `ArgumentValue` we'll never hit the default value (for valuetypes).


Shane32:
That is by design. The defaultValue is used only when no field default value is specified. This is typically used when your internal logic requires a 0 instead of null or similar.

Variable default values are supplied by the caller. Overriding them would be incorrect in any case. 

If you have a need to examine the argument source, you can inspect the arguments dictionary on the context directly.

kimipsen:
You're absolutely right here. It won't work, as this way it won't be possible to specify default values on arguments. I'm closing this PR.

PR #2559, opened 2021-05-27
Allow GetArgument with default value to operate on reference types

### LINKED ISSUE #2557: Why the `IsValueType` check?

Shane32:
Fixes #2557 

The bug is a regression against v3.2.0 and prior, but exists in 3.3.0 and newer.  The bug was introduced in ~#1456~ #2160.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug2557.cs
+++ b/src/GraphQL.Tests/Bugs/Bug2557.cs
@@ -1,0 +1,81 @@
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using Xunit;
+
+namespace GraphQL.Tests.Bugs
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/2557
+    public class Bug2557 : QueryTestBase<Bug2557.MySchema>
+    {
+        [Fact]
+        public void NoDefault_NoArg() => AssertQuerySuccess("{ noDefault }", @"{ ""noDefault"": ""getArgumentDefault"" }");
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


sungam3r:
I think that this "bug" was not introduced in #1456. 

Look at `exists` flag.
изображение

Before changes in `ExecutionHelper.GetArgumentValues` and introduction of `ArgumentValue` this `exists` flag works well. Now it is always `true` (@Shane32 I think you can check it easily) so execution **always** flows to the first branch, not just returning `defaultValue`. I already do not remember exactly the progress of my thoughts, but this check was completely added to properly process defaults for ref types.

Shane32:
> I think that this "bug" was not introduced in #1456.

You're right; I did a whole bunch of before/after testing and traced it to #2160 .  After #2160 breaks many of the tests in this PR that passed before.  #2160 was included with version 3.3.0 so this bug has existed since then but not before.

Is there any reason not to merge this PR?  The new tests exercise as many combinations of default values as I can think of, and everything works properly now.

Shane32:
@sungam3r if you have no further comments, I’ll merge this.

sungam3r:
Just reviewed. Well, I don't remember all our refactorings around arguments and default values, I just hope that huge testing eventually allows to achive the right behavior.

Issue #2563, opened 2021-06-02
Federated Schema generates an empty query type in the SDL


myermian:
## Description

Creating a federated subgraph using graphql dotnet fails if you do not extend the Query type.

### Steps to reproduce

1. Download the [apollographql/federation-demo](https://github.com/apollographql/federation-demo) repository.
2. Run the `npm install` command to install the dependencies.
3. Run the `npm run start-services` command to run the services in terminal window A.
4. Run the `npm run start-gateway` command to run the gateway in terminal window B.
5. Download and run the [sample solution](https://github.com/graphql-dotnet/graphql-dotnet/files/6583970/Sample.zip) provided.
6. Modify the gateway.js file to include the new service endpoint. Save the file, the gateway will automatically try to update and fail.

### Expected result

The gateway should accept the new service endpoint and federate the schema properly.

### Actual result

The gateway fails with the error: `UnhandledPromiseRejectionWarning: Error: Couldn't load service definitions for "user-service" at http://localhost:5000/graphql: Syntax Error: Expected Name, found "}".` (full error below)

> (node:7676) UnhandledPromiseRejectionWarning: Error: Couldn't load service definitions for "user-service" at http://localhost:5000/graphql: Syntax Error: Expected Name, found "}".
>     at X:\GitHub\apollographql\federation-demo\node_modules\@apollo\gateway\dist\loadServicesFromRemoteEndpoint.js:45:19
>     at processTicksAndRejections (internal/process/task_queues.js:97:5)
>     at async Promise.all (index 4)
>     at async Object.getServiceDefinitionsFromRemoteEndpoint (X:\GitHub\apollographql\federation-demo\node_modules\@apollo\gateway\dist\loadServicesFromRemoteEndpoint.js:48:32)
>     at async ApolloGateway.updateSchema (X:\GitHub\apollographql\federation-demo\node_modules\@apollo\gateway\dist\index.js:266:24)
>     at async ApolloGateway.loadDynamic (X:\GitHub\apollographql\federation-demo\node_modules\@apollo\gateway\dist\index.js:250:13)
>     at async ApolloGateway.load (X:\GitHub\apollographql\federation-demo\node_modules\@apollo\gateway\dist\index.js:221:15)

<hr>

If you navigate to the new service (http://localhost:5000/ui/playground) and run the following query, you will see that the SDL is returning an empty type query.

```gql
query {
  _service {
    sdl
  }
}
```

```json
{
  "data": {
[...]

PR #2576, opened 2021-06-16
Updated the FederatedSchemaPrinter file to prevent adding an empty Query { } as part of the SDL.

myermian:
Updated the FederatedSchemaPrinter file to prevent adding an empty Query { } as part of the SDL. This occurs if the query type only contains federated fields.

This does not prevent the Query type from being part of the SDL if the type includes non-federated fields.

Added an xUnit test case for this, and as there was no associated test class for FederatedSchemaPrinter a new test class was created.

See issue #2563.

### DIFF
--- a/src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
+++ b/src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
@@ -52,6 +52,10 @@ namespace GraphQL.Utilities.Federation
 
         public override string PrintObject(IObjectGraphType type)
         {
+            // Do not return an empty query type: "Query { }" as it is not valid as part of the sdl.
+            if (type != null && String.Equals(type.Name, "Query", StringComparison.Ordinal) && !type.Fields.Any(x => !IsFederatedType(x.ResolvedType.GetNamedType().Name)))
+               return String.Empty;
+
             var isExtension = type.IsExtensionType();
 
             var interfaces = type.ResolvedInterfaces.List.Select(x => x.Name).ToList();
--- a/src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
+++ b/src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
@@ -53,8 +53,8 @@ namespace GraphQL.Utilities.Federation
         public override string PrintObject(IObjectGraphType type)
         {
             // Do not return an empty query type: "Query { }" as it is not valid as part of the sdl.
-            if (type != null && String.Equals(type.Name, "Query", StringComparison.Ordinal) && !type.Fields.Any(x => !IsFederatedType(x.ResolvedType.GetNamedType().Name)))
-                return String.Empty;
+            if (type != null && string.Equals(type.Name, "Query", StringComparison.Ordinal) && !type.Fields.Any(x => !IsFederatedType(x.ResolvedType.GetNamedType().Name)))
+                return string.Empty;
 
             var isExtension = type.IsExtensionType();
 
--- a/src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
+++ b/src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
@@ -54,7 +54,7 @@ namespace GraphQL.Utilities.Federation
         {
             // Do not return an empty query type: "Query { }" as it is not valid as part of the sdl.
             if (type != null && String.Equals(type.Name, "Query", StringComparison.Ordinal) && !type.Fields.Any(x => !IsFederatedType(x.ResolvedType.GetNamedType().Name)))
-               return String.Empty;
+                return String.Empty;
 
             var isExtension = type.IsExtensionType();
 
--- a/src/GraphQL.Tests/Utilities/Federation/FederatedSchemaPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/Federation/FederatedSchemaPrinterTests.cs
@@ -1,4 +1,3 @@
-using System;
 using GraphQL.Types;
 using GraphQL.Utilities;
 using GraphQL.Utilities.Federation;
@@ -24,7 +23,7 @@ namespace GraphQL.Tests.Utilities.Federation
             string result = federatedSchemaPrinter.PrintObject(query);
 
             // Assert
-            Assert.Equal(String.Empty, result);
+            Assert.Equal(string.Empty, result);
         }
     }
 }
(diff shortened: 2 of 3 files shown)


myermian:
Some help here? Not sure how I could possibly have failed the analysis with just these lines of code I added? Again, apologies but I rarely check in code against GitHub.

Shane32:
LGTM analysis isn’t working so don’t worry about that. I approved the workflow run so it should run the remainder of the checks now.

Shane32:
It failed a formatting check just now. If you look at the log it will say which line of code isn’t formatted correctly.

src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
```diff
@@ -52,6 +52,10 @@ public string PrintFederatedDirectivesFromAst(IProvideMetadata type)
 
         public override string PrintObject(IObjectGraphType type)
         {
+            // Do not return an empty query type: "Query { }" as it is not valid as part of the sdl.
+            if (type != null && String.Equals(type.Name, "Query", StringComparison.Ordinal) && !type.Fields.Any(x => !IsFederatedType(x.ResolvedType.GetNamedType().Name)))
+               return String.Empty;
```
Shane32, line 57: Needs to be indented 4 spaces (not 3 spaces. )

Shane32:
```suggestion
                return String.Empty;
```

Shane32: You can commit the suggestion right here by the way.

myermian: I downloaded the dotnet tool used, ran the tool, it modified the file, I checked in the change and pushed it to my forked branch.

myermian:
There still appears to be formatting errors, apparently it does not like 'String' and prefers 'string'. That's fine, I can change it to use the alias version of the String class.

I don't know why it mentions the `ResolveFieldContextExtensions.cs` file, I never touched that, but if I need to, I can fix that as well.

Shane32:
Sometimes the formatter seems a bit overzealous I know. I’m not sure why it would give errors with the extensions file. Maybe the rules were updated since the last PR? I wouldn’t think you’d need to change it but I’m not sure.

myermian:
Either way, I think I resolved that code without losing the logic. The suggestion that VS was giving does not seem like the logic would be the same. This seems like the logic will be the same without a complaint from an IDE rule.

Shane32:
Now it says the test failed....

```
[xUnit.net 00:00:02.59]     GraphQL.Tests.Utilities.Federation.FederatedSchemaPrinterTests.PrintObject_ReturnsEmptyString_GivenQueryTypeHasOnlyFederatedFields [FAIL]
  Failed GraphQL.Tests.Utilities.Federation.FederatedSchemaPrinterTests.PrintObject_ReturnsEmptyString_GivenQueryTypeHasOnlyFederatedFields [7 ms]
  Error Message:
   Assert.Equal() Failure
          ↓ (pos 0)
Expected: 
Actual:   type Query {\n  _entities: [__GraphQLTypeR···
          ↑ (pos 0)
  Stack Trace:
     at GraphQL.Tests.Utilities.Federation.FederatedSchemaPrinterTests.PrintObject_ReturnsEmptyString_GivenQueryTypeHasOnlyFederatedFields() in /home/runner/work/graphql-dotnet/graphql-dotnet/src/GraphQL.Tests/Utilities/Federation/FederatedSchemaPrinterTests.cs:line 26
```

Shane32:
You should be able to run tests locally via the Test Explorer - just hit the 'run all' button.  Do they work locally for you?

image

myermian:
> Now it says the test failed....
> 
> ```
> [xUnit.net 00:00:02.59]     GraphQL.Tests.Utilities.Federation.FederatedSchemaPrinterTests.PrintObject_ReturnsEmptyString_GivenQueryTypeHasOnlyFederatedFields [FAIL]
>   Failed GraphQL.Tests.Utilities.Federation.FederatedSchemaPrinterTests.PrintObject_ReturnsEmptyString_GivenQueryTypeHasOnlyFederatedFields [7 ms]
>   Error Message:
>    Assert.Equal() Failure
>           ↓ (pos 0)
> Expected: 
> Actual:   type Query {\n  _entities: [__GraphQLTypeR···
>           ↑ (pos 0)
>   Stack Trace:
>      at GraphQL.Tests.Utilities.Federation.FederatedSchemaPrinterTests.PrintObject_ReturnsEmptyString_GivenQueryTypeHasOnlyFederatedFields() in /home/runner/work/graphql-dotnet/graphql-dotnet/src/GraphQL.Tests/Utilities/Federation/FederatedSchemaPrinterTests.cs:line 26
> ```

Yea, I am running it locally. I guess I didn't understand how the code is built as well as I did. Manually testing it, it worked, but there is something interesting, and I can't hunt down where it happens.

I'm utterly lost in how the schema resolves all the named types. Mine are resolving to '__GraphQLTypeReference' instead of '_Service' for example. When I run it on my web application, everything works fine, but I can't configure the test properly.

Shane32:
Perhaps I can take a look at it.  I don't use federation, but I have a decent understanding of how the schema is built.

myermian:
I identified the problem after hours of debugging. I utilized the `FederatedSchema` class to generate the schema, but more importantly I had to call `Initialize()` on the schema for it to get all the properly resolved names.

Shane32:
> I identified the problem after hours of debugging. I utilized the `FederatedSchema` class to generate the schema, but more importantly I had to call `Initialize()` on the schema for it to get all the properly resolved names.

Ahh; makes sense.  The document executer always calls `Initialize();` on the schema; probably the schema printer classes should also.  Added issue #2577

Shane32:
@joemcbride Do you have time for a quick review here?  This looks fine to me, but I don't use federation.

joemcbride (approved):

myermian:
> > I identified the problem after hours of debugging. I utilized the `FederatedSchema` class to generate the schema, but more importantly I had to call `Initialize()` on the schema for it to get all the properly resolved names.
> 
> Ahh; makes sense. The document executer always calls `Initialize();` on the schema; probably the schema printer classes should also. Added issue #2577

I'm not sure if this is an issue outside of testing. Under normal usage, you will host an ASP.NET Core site and use the GraphQL middleware. Somewhere in that chain, the `Initialize()` method is called (perhaps from the `AllTypes` property?) which has everything resolved in terms of naming.

This was an issue in testing because I was trying to minimize the test code; I couldn't set up an entire site for a unit test.

Shane32:
Thanks @myermian !!!

Issue #2588, opened 2021-07-06
ExecutionStrategy tracker


mbenedykconfigit:
## Summary

It would be nice to have documentation for ExecutionStrategies, how they work and how nodes are processed in differend strategies like `SerialExecutionStrategy`, 'ParallelExecutionStrategy'.

Or having execution tracker could be a nice idea, it would be helpful with more advanced investigations and reproductions

### Basic example

Genera idea for tracker:
```csharp
public class TrackableSerialExecutionStrategy : SerialExecutionStrategy
    {
        protected override Task ExecuteNodeAsync(ExecutionContext context, ExecutionNode node)
        {
            var result =  base.ExecuteNodeAsync(context, node);
            Console.WriteLine(GetNodeInfo(node));
            return result;
        }

        private string GetNodeInfo(ExecutionNode node)
        {
            if (node is RootExecutionNode)
            {
                return "root";
            }

            if (node.Parent is ArrayExecutionNode arrayExecutionNode)
            {
                var nodeIndex = arrayExecutionNode.Items.IndexOf(node);
                return $"{GetNodeInfo(node.Parent)}[{nodeIndex}]";
            }

            var accumulator = new StringBuilder();
            if (node.Parent != null)
            {
                accumulator.Append($"{GetNodeInfo(node.Parent)}");
            }

            if (node.Name != node.FieldDefinition.Name)
            {
                accumulator.Append($".{node.Name}.{node.FieldDefinition?.Name}");
                return accumulator.ToString();
            }

            accumulator.Append($".{node.FieldDefinition?.Name}");

            if (node.Result is IDataLoaderResult)
            {
                accumulator.Append(" (DataLoader)");
            }

            return accumulator.ToString();
        }
    }
```

Example query:
```graphql
query MyQuery {
  families {
    leader_dataLoader {
      lastName
      name
    }
    leader {
      lastName
      name
    }
  }
}
```

Example output:
```
root
root.families
root.families[0]
root.families[0].leader_dataLoader (DataLoader)
root.families[0].leader
root.families[0].leader.lastName
root.families[0].leader.name
root.families[1]
root.families[1].leader_dataLoader (DataLoader)
root.families[1].leader
root.families[1].leader.lastName
root.families[1].leader.name
root.families[2]
root.families[2].leader_dataLoader (DataLoader)
root.families[2].leader
root.families[2].leader.lastName
root.families[2].leader.name
root.families[3]
[...]

PR #2589, opened 2021-07-06
Fix execution order of data loader nodes with serial execution strategy

Shane32:
See #2588

### DIFF
--- a/src/GraphQL/Execution/SerialExecutionStrategy.cs
+++ b/src/GraphQL/Execution/SerialExecutionStrategy.cs
@@ -10,6 +10,7 @@ namespace GraphQL.Execution
         // frequently reused objects
         private Stack<ExecutionNode> _reusableNodes;
         private Queue<ExecutionNode> _reusableDataLoaderNodes;
+        private Stack<ExecutionNode> _reusableAddlNodes;
 
         /// <summary>
         /// Gets a static instance of <see cref="SerialExecutionStrategy"/> strategy.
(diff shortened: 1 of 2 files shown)


src/GraphQL/Execution/SerialExecutionStrategy.cs
```diff
@@ -10,6 +10,7 @@ public class SerialExecutionStrategy : ExecutionStrategy
         // frequently reused objects
         private Stack<ExecutionNode> _reusableNodes;
         private Queue<ExecutionNode> _reusableDataLoaderNodes;
+        private Stack<ExecutionNode> _reusableAddlNodes;
```
sungam3r, line 13: addl ?

sungam3r: Worth to add comments: _reusableNodes are for ....  _reusableDataLoaderNodes are for ...  _reusableAddlNodes are for...

src/GraphQL/Execution/SerialExecutionStrategy.cs
```diff
@@ -75,19 +77,27 @@ await OnBeforeExecutionStepAwaitedAsync(context)
                         }
                         else if (node is IParentExecutionNode parentNode)
                         {
-                            // Add in reverse order so fields are executed in the correct order
-                            parentNode.ApplyToChildren((node, state) => state.Push(node), nodes, reverse: true);
+                            // Do not reverse the order of the nodes here
+                            parentNode.ApplyToChildren((node, state) => state.Push(node), addlNodes, reverse: false);
```
sungam3r, line 81: Why `reverse` is `false` here?

Shane32: It's hard to explain.  It's like, if nodes A B are pushed in reverse order, but then under a separate node, nodes C D are pushed in reverse order, then you have B A D C instead of D C B A.  Something like that.  So the order of the data loader nodes had to be reversed immediately prior to execution of those nodes, rather than while they were being queued.

The test does test for the situation and the fix does resolve it, and it's pretty evident when looking at test output that the execution was incorrect before.

If you're still curious, I'd suggest reversing the change and running the test again to see the output.  I can't really remember all the details of this PR.

src/GraphQL/Execution/SerialExecutionStrategy.cs
Shane32: > addl ?

Add'l is short for 'additional'

PR #2592, opened 2021-07-08
Allow schema printer to print descriptions according to 2018 GraphQL spec

### LINKED ISSUE #2574: Is there a way to add description (not the comment) to field or type in schema?
## Summary
According to official graphql specification, field description is surrounded by double-quotes, type description is surrounded by triple double-quotes. 
In turn, comments start from the # character.

Value set via `GraphQL.Builders.Description`  method results in comment (not a description) in the schema.
```
Field<SomeGraphType>()
    .Name("fieldName")
    .Description("field description")
```

Value set via `Description` property of `GraphQL.Types.GraphType` also results in comment (not a description) in the schema.
```
public class MyGraphType : ObjectGraphType<MyModel>
{
    public MyGraphType()
    {
        Name = "TypeName";
        Description = "type description";
    }
}
```

Is there a way to specify description (not the comment) for the type (`GraphQL.Types.ObjectGraphType`) or for its fields in the "GraphType First Approach"?

## Relevant information
Assembly GraphQL, Version=4.1.0.0
GraphQL.Server.All, Version=5.0.2

https://spec.graphql.org/June2018/#sec-Descriptions
https://spec.graphql.org/June2018/#sec-Comments

Shane32:
Fixes #2574

### DIFF
--- a/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
@@ -39,7 +39,7 @@ namespace GraphQL.Tests.Utilities
 
         private string print(ISchema schema)
         {
-            return print(schema, new SchemaPrinterOptions { IncludeDescriptions = true, IncludeDeprecationReasons = true });
+            return print(schema, new SchemaPrinterOptions { IncludeDescriptions = true, IncludeDeprecationReasons = true, PrintDescriptionsAsComments = true });
         }
 
         private string print(ISchema schema, SchemaPrinterOptions options)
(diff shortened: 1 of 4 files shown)


src/GraphQL/Utilities/SchemaPrinterOptions.cs
```diff
@@ -19,5 +19,10 @@ public class SchemaPrinterOptions
         /// Indicates whether to use ',' instead of '&amp;' when inheriting a type from multiple interfaces.
         /// </summary>
         public bool OldImplementsSyntax { get; set; }
+
+        /// <summary>
+        /// Indicates whether to print descriptions as comments for compatibility with the 2016 GraphQL specification.
+        /// </summary>
+        public bool PrintDescriptionsAsComments { get; set; } = true;
```
sungam3r, line 26: @Shane32 Hi! I continue to view summer emails. Is this a candidate for removing in v5?

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -386,6 +400,48 @@ public string PrintDescription(string description, string indentation = "", bool
             return desc;
         }
 
+        public string PrintDescription(string description, string indentation = "", bool firstInBlock = true)
```
sungam3r, line 403: Reading this, I do not cease to remember the parser code, where it saved every allocation.

src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
```diff
@@ -1010,6 +1220,259 @@ enum __TypeKind {
...
+                Query = new ObjectGraphType
+                {
+                    Name = "Root"
+                }
+            };
+            schema.Query.Fields.Add(new FieldType { Name = "unused", ResolvedType = new StringGraphType() });
+            var printer = new SchemaPrinter(schema, new SchemaPrinterOptions { IncludeDescriptions = true, PrintDescriptionsAsComments = false });
+            var result = Environment.NewLine + printer.PrintIntrospectionSchema();
+
+            const string expected = @"
```
sungam3r, line 1241: I would place such text into file.

src/GraphQL/Utilities/SchemaPrinterOptions.cs
Shane32: Fine with me, but we should mark as obsolete in v4.

PR #2595, opened 2021-07-08
Highlight the importance of using Singleton Schema

EmmanuelPonnudurai:
This is my attempt to make it absolutely clear on the importance of using Singleton Schema and the kind of performance benefits it has. We have seen first-hand the enormous improvement in performance once we moved to singleton schema and used service locator pattern. 

The idea with the doc update is to make it clear to the reader that we know it's an anti-pattern but in this specific situation, it's a necessity considering the huge performance degradation, if we don't go this route. 

May be the next reader will go away with less ambiguity because it did happen to me when I first read it, I didn't pay attention to it and paid the price dearly.

### DIFF
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -159,11 +159,10 @@ lifetimes are as follows:
 * **Scoped** services are created per scope. In a web application, every web request creates a new unique service scope. That means scoped services are generally created per web request.
 * **Singleton** services are created per DI container. That generally means that they are created only one time per application and then used for whole the application life time.
 
-> It is _highly_ recommended that the schema is registered as a singleton. This provides the best performance as
-the schema does not need to be built for every request. This also means that your server is doing way less work and 
-hence reduced CPU utilization. As all graph types are constructed at the
+> It is _highly_ recommended that the schema is registered as a singleton. As all graph types are constructed at the
 same time as the schema, all graph types will effectively have a singleton lifetime, regardless
-of how it is registered with the DI framework.
+of how it is registered with the DI framework. This is most performant appraoch. Having a scoped schema can degrade performance
+by a huge margin. For instance, even a small schema execution can slow down by 100x, and much more with a large schema.
 
 Scoped lifetime can be used to allow the schema and all its graph types access to the current DI scope.
 This is not recommended; please see [Scoped Services](#scoped-services-with-a-singleton-schema-lifetime)
@@ -257,15 +256,13 @@ public class MyGraphType : ObjectGraphType<Category>
 }
 ```
 
-> Be aware that using the service locator in this fashion described in this section could be considered an
+Be aware that using the service locator in this fashion described in this section could be considered an
 Anti-Pattern. See [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/).
-However, the performance benefits far outweigh the anti-pattern aspect so prefer going ahead with this approach to avoid heavy performance costs.
+However, the performance benefits far outweigh the anti-pattern idealogy. 
 
-> Another approach to resolve scoped services is to use the SteroidsDI project, as described below. Whlie
-StreoidsDI approach also uses service locator pattern, it does provide a clear dependency visibility via constructor injection, 
-which can be mocked if you ever want to write unit tests on type generation and set up process.
-
-Within the GraphQL.MicrosoftDI package, there is also a builder approach to adding scoped dependencies:
+Within the `GraphQL.MicrosoftDI` package, there is also a builder approach to adding scoped dependencies.
+This makes for a concise and declarative approach. Each field clearly states the service it needs
+and thereby, the anti-pattern argument does not apply anymore.
 
 ```csharp
 public class MyGraphType : ObjectGraphType<Category>
@@ -281,6 +278,8 @@ public class MyGraphType : ObjectGraphType<Category>
     }
 }
 ```
+
+Another approach to resolve scoped services is to use the SteroidsDI project, as described below.
 
 ## Using SteroidsDI
 
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -159,8 +159,9 @@ lifetimes are as follows:
 * **Scoped** services are created per scope. In a web application, every web request creates a new unique service scope. That means scoped services are generally created per web request.
 * **Singleton** services are created per DI container. That generally means that they are created only one time per application and then used for whole the application life time.
 
-It is highly recommended that the schema is registered as a singleton. This provides the best performance as
-the schema does not need to be built for every request. As all graph types are constructed at the
+> It is _highly_ recommended that the schema is registered as a singleton. This provides the best performance as
+the schema does not need to be built for every request. This also means that your server is doing way less work and 
+hence reduced CPU utilization. As all graph types are constructed at the
 same time as the schema, all graph types will effectively have a singleton lifetime, regardless
 of how it is registered with the DI framework.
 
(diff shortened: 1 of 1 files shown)


docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -256,9 +257,14 @@ public class MyGraphType : ObjectGraphType<Category>
...
-Be aware that using the service locator in this fashion described in this section could be considered an
+> Be aware that using the service locator in this fashion described in this section could be considered an
 Anti-Pattern. See [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/).
-Another approach to resolve scoped services is to use the SteroidsDI project, as described below.
+However, the performance benefits far outweigh the anti-pattern aspect so prefer going ahead with this approach to avoid heavy performance costs.
+
+> Another approach to resolve scoped services is to use the SteroidsDI project, as described below. Whlie
+StreoidsDI approach also uses service locator pattern, it does provide a clear dependency visibility via constructor injection, 
+which can be mocked if you ever want to write unit tests on type generation and set up process.
+
```
Shane32, line 267: I like the changes here; however I would de-emphasize these paragraphs.  (Eliminate the `>`).  Using the SteroidsDI project is not recommended, and we do not want to raise undue concern over this so-called "anti-pattern" described here.  It is simply noted as an option for people that are already concerned about it.

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -159,8 +159,9 @@ lifetimes are as follows:
 * **Scoped** services are created per scope. In a web application, every web request creates a new unique service scope. That means scoped services are generally created per web request.
 * **Singleton** services are created per DI container. That generally means that they are created only one time per application and then used for whole the application life time.
 
-It is highly recommended that the schema is registered as a singleton. This provides the best performance as
-the schema does not need to be built for every request. As all graph types are constructed at the
+> It is _highly_ recommended that the schema is registered as a singleton. This provides the best performance as
+the schema does not need to be built for every request. This also means that your server is doing way less work and 
+hence reduced CPU utilization. As all graph types are constructed at the
```
Shane32, line 164: I think the added sentence is a bit redundant, but fine.  I do like the emphasis placed here.  It would be better if we could give a performance comparison, such as "100 times slower", but this is entirely dependent on the complexity of the schema, and whether they are using the `EnableReadDescriptionFromXmlDocumentation` option and/or expression builders to create their schema.

:+1:

Shane32 (changes requested):

docs2/site/docs/getting-started/dependency-injection.md
Shane32: Note that GraphQL.NET does include the GraphQL.MicrosoftDI project, which although does not expose the dependent services in the constructor, allows the `WithService<>` resolver pattern so the schema designer can consider the schema itself as defining the services.  Basically, the schema should never contain any logic (aka resolver code) but basically only exists to define a schema definition and wire fields to the a method of a service.  In that sense, it is perfectly logical that the dependent services are noted for each individual field, rather than on the schema/graph as a whole, as each field may require different services.  So in my mind the anti-pattern logic does not apply anymore.

This is how I write most of my graphs; so basically the resolver code only ever contains a few lines of code, namely to retrieve the arguments (if any) and execute the service method.

If you agree and see a better way to clarify this, that might help others as well.  It's been a while since I last worked on those docs.

docs2/site/docs/getting-started/dependency-injection.md
Shane32: Perhaps something like "Having a scoped schema can easily slow down execution by 100x with even a small schema, and much more with a large schema."  Just for reference, benchmarks indicate that running the "hero" query on a very simple schema only takes 1% of the build time, but running an introspection request on the same simple schema takes twice as long to execute as it does to build.  So while build time can be slow, it's certainly possible that execution time can be just as long for a complex query (meaning that even then it will run twice as fast if it didn't need to build).  But that's assuming a small schema with no expression builders.

docs2/site/docs/getting-started/dependency-injection.md
EmmanuelPonnudurai: The main reason for emphasis is to make this stand out. Often, once someone reads about this service locator and anti-pattern thing, they usually avoid it. Then they usually chose to do constructor injection and turn schema into scoped and realized later that this was a mistake. So to clip that at the bud, I felt that the emphasis and additional information is needed. 

I didn't know that `SteroidsDI` is not recommended. Let me remove that thing then.

I will reword this along with the information of `WithService` via `GraphQL.MicrosoftDI`. But then, this still is an encapsulation over the service locator. Because I think this also uses `RequestServices` which then uses `IServiceProvider` to get the service of required Type. As opposed to injecting the specified service in the constructor. 

I do agree that this is better because its declarative in nature but is eventually the same thing. No?

Agreed that the Graph building code should be just a slim layer, however the query would need to get the required business logic service, which is usually the dependency that gets injected into the Query class' constructor.  In my example for instance. I have a whole bunch of separate classes for each query. They are not within one big class, in which case it would be disastrous cos I would end up injecting 100's of dependencies. So we have split the code in such a way that we have one class per query. and the graph types( input and output types) are their own classes. This kind of structure is necessary when we have a lot of queries and lots of types. 

For example, 

```cs
public class GetPersonQuery : IQuery
{
    private readonly IPersonService _personService;

    public GetPersonQuery(IPersonService personService)
    {
        _personService = personService;
    }

    public void PopulateFields()
    {
        Field<PersonType>("person", 
            arguments: new List<QueryArgument>{ new QueryArgument<NonNullGraphType<PersonRequestType>> {  Name = "personRequest" } }
            resolve: (context) => _personService.GetPerson(context.source.id))
    }
}
```

The `IQuery` interface has the `PopulateFields` method. All these `IQuery` type of classes are invoked from another place which helps pick up all these queries as part of the schema building process that we have. This ensures we have an isolated flow and easy for developers to add their own queries and follow along in this pattern.

And this is where the issue starts. Conventional wisdom is to use as above. So, you can clearly see that the constructor is asking for `IPersonService` so it depends on it. But this leads to worst performance because `IPersonService` is a scoped dependency and so I change my type and the schema to be scoped. 

After analyzing, we found this to be the culprit and now we removed this and went with the pattern of `RequestServices`. While this does mean that the constructor is not asking for any dependency and is resolving `IPersonService` internally, which is the tenet of the service locator anti pattern. But is not of significance considering the performance issues.

Now, the same thing, I want to make clear in the docs so that the next person, should know this well ahead and not make this mistake and start crying foul that performance is bad. Basically. 

Apologies for the long write up.

docs2/site/docs/getting-started/dependency-injection.md
EmmanuelPonnudurai: Got it. Yes, I think I ended up being redundant here. I will reword this and push the changes. Thanks for noticing it.

EmmanuelPonnudurai:
@Shane32 I have made some changes. Can you take another look when you get a chance. Thank you.

Shane32 (approved):

Shane32:
Looks good; thanks!

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -258,8 +258,11 @@ public class MyGraphType : ObjectGraphType<Category>
 
 Be aware that using the service locator in this fashion described in this section could be considered an
 Anti-Pattern. See [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/).
-Another approach to resolve scoped services is to use the SteroidsDI project, as described below.
-Within the GraphQL.MicrosoftDI package, there is also a builder approach to adding scoped dependencies:
+However, the performance benefits far outweigh the anti-pattern idealogy.
```
sungam3r, line 261: Not agree. One completely does not follow from the other.

sungam3r: #2739

PR #2596, opened 2021-07-08
NRT changes part 1

Shane32:


Shane32:
@sungam3r @joemcbride I will be posting and merging these NRT-related PRs quickly, as subsequent sets of changes depend on the ones within the previous PR.  Please let me know if you have any comments.

src/GraphQL/Utilities/MetadataProvider.cs
```diff
@@ -10,20 +12,20 @@ namespace GraphQL.Utilities
     /// </summary>
     public class MetadataProvider : IProvideMetadata
     {
-        private Dictionary<string, object> _metadata;
+        private Dictionary<string, object>? _metadata;
```
sungam3r, line 15: Maybe ` private Dictionary<string, object?>? _metadata;`, not sure.

src/GraphQL/Language/AST/ValueNodes/ValueNode.cs
```diff
@@ -1,14 +1,35 @@
...
-        public T Value { get; protected set; }
+        public T Value
+        {
+            get;
+            [Obsolete]
+            protected set;
+        }
 
-        object IValue.Value => Value;
+        object IValue.Value => Value!;
```
sungam3r, line 32: `NullValue` returns `null`.

sungam3r: In master I see ` object? Value { get; }` for `IValue` so need change here.

src/GraphQL/Language/AST/Variable.cs
```diff
@@ -1,20 +1,41 @@
...
     /// Represents a variable name and value tuple that has been gathered from the document and attached <see cref="Inputs"/>.
     /// </summary>
     public class Variable
     {
+        [Obsolete]
+        public Variable() : this(null!)
+        {
+        }
+
+        public Variable(string name)
```
sungam3r, line 17: It is possible that the best solution would be to use init-only properties.

src/GraphQL/Language/AST/VariableDefinition.cs
```diff
@@ -45,17 +64,17 @@ public override IEnumerable<INode> Children
...
             }
         }
 
         /// <inheritdoc/>
         public override void Visit<TState>(Action<INode, TState> action, TState state)
         {
             action(Type, state);
-            action(DefaultValue, state);
+
+            if (DefaultValue != null)
```
sungam3r, line 76: 👍

src/GraphQL/Language/AST/VariableDefinition.cs
```diff
@@ -11,15 +13,27 @@ public class VariableDefinition : AbstractNode, IHaveName
...
         /// </summary>
-        public VariableDefinition(NameNode node)
+        [Obsolete]
+        public VariableDefinition(NameNode node) : this(node, null!)
         {
             NameNode = node;
         }
 
+        /// <summary>
+        /// Initializes a new variable definition node with the specified <see cref="NameNode"/> containing the name of the variable.
```
sungam3r, line 23: .... and variable type.

src/GraphQL/Language/AST/VariableReference.cs
```diff
@@ -18,7 +20,7 @@ public VariableReference(NameNode name)
         /// <summary>
         /// Returns the name of the variable being referenced.
         /// </summary>
-        public string Name => NameNode.Name;
+        public string Name => NameNode.Name!;
```
sungam3r, line 23: Why does `NameNode.Name` nullable?

src/GraphQL/Language/AST/ValueNodes/ObjectValue.cs
```diff
@@ -54,9 +56,9 @@ public object Value
         {
             get
             {
-                var obj = new Dictionary<string, object>(ObjectFieldsList.Count);
+                var obj = new Dictionary<string, object?>(ObjectFieldsList.Count);
                 foreach (var item in ObjectFieldsList)
-                    obj.Add(item.Name, Field(item.Name).Value.Value);
```
sungam3r, line 59: Do you have thoughts, why was it so written here?

src/GraphQL/Language/AST/ValueNodes/FloatValue.cs
```diff
@@ -10,14 +12,18 @@ public class FloatValue : ValueNode<double>
...
+        public FloatValue(double value) : base(ValidateValue(value))
+        {
+        }
+
+        private static double ValidateValue(double value)
         {
             // TODO: see https://github.com/graphql-dotnet/graphql-dotnet/pull/2379#issuecomment-800828568 and https://github.com/graphql-dotnet/graphql-dotnet/pull/2379#issuecomment-800906086
             // if (double.IsNaN(value) || double.IsInfinity(value))
             if (double.IsNaN(value))
                 throw new ArgumentOutOfRangeException(nameof(value), "Value cannot be NaN."); // Value cannot be NaN or Infinity.
```
sungam3r, line 24: `or Infinity` from comment should be removed

src/GraphQL/Language/AST/VariableReference.cs
Shane32: I don't know.  It isn't now, so maybe I already updated it in a subsequent PR.

src/GraphQL/Language/AST/Variable.cs
Shane32: I think I had come to the conclusion that it would be a breaking change.  But for v5, sure.

src/GraphQL/Language/AST/ValueNodes/ValueNode.cs
Shane32: That is because `NullNode` returns `null`.  `ValueNode<T>.Value` cannot return `null`.  So this code should be correct.

src/GraphQL/Utilities/MetadataProvider.cs
Shane32: Already changed in a subsequent PR.

src/GraphQL/Language/AST/ValueNodes/ObjectValue.cs
Shane32: Nope.  It made no sense to me, so I fixed it.

src/GraphQL/Language/AST/ValueNodes/ValueNode.cs
sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Language/AST/IValue.cs#L11

I see explicit `IValue` (not `IValue<T>`) interface implementation that returns `object` instead of `object?`. So was it planned?

src/GraphQL/GraphQLExtensions.cs
```diff
@@ -151,7 +153,7 @@ internal static (IGraphType resolvedType, Type type) GetNamedTypes(this IGraphTy
         public static IGraphType GetNamedType(this IGraphType type)
         {
             if (type == null)
-                return null;
+                return null!;
```
sungam3r, line 156: :)

src/GraphQL/Language/AST/Directive.cs
```diff
@@ -29,16 +31,24 @@ public Directive(NameNode node)
...
-        public Arguments Arguments { get; set; }
+        public Arguments? Arguments { get; set; }
 
         /// <inheritdoc/>
         public override IEnumerable<INode> Children
         {
-            get { yield return Arguments; }
+            get
+            {
+                if (Arguments != null)
```
sungam3r, line 41: 👍

src/GraphQL/Language/AST/Directive.cs
```diff
@@ -29,16 +31,24 @@ public Directive(NameNode node)
...
+                    yield return Arguments;
+            }
         }
 
         /// <inheritdoc/>
-        public override void Visit<TState>(Action<INode, TState> action, TState state) => action(Arguments, state);
+        public override void Visit<TState>(Action<INode, TState> action, TState state)
+        {
+            if (Arguments != null)
```
sungam3r, line 49: 👍

src/GraphQL/Language/AST/FragmentDefinition.cs
```diff
@@ -11,15 +13,27 @@ public class FragmentDefinition : AbstractNode, IDefinition, IHaveSelectionSet
...
         /// Initializes a new fragment definition node with the specified <see cref="NameNode"/> containing the name of this fragment definition.
         /// </summary>
-        public FragmentDefinition(NameNode node)
+        [Obsolete]
+        public FragmentDefinition(NameNode node) : this(node, null!, null!)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new fragment definition node with the specified <see cref="NameNode"/> containing the name of this fragment definition and its selection set.
```
sungam3r, line 22: and type

src/GraphQL/Language/AST/ValueNodes/ValueNode.cs
Shane32: I think I see what you're saying.  Since it's an explicit `IValue` implementation, not exposed as a public property on the class, it might as well be `object? IValue.Value => Value;`.

That's true, it can be.  However, you could also say that this is more accurate source-code, depicting that a non-null value will always be returned, even if `IValue.Value` is defined as `object?`.  If you want to change it, that's fine; I don't care either way.

src/GraphQL/GraphQLExtensions.cs
Shane32: For backwards compatibility I left it like this.  For v5 we could make this method throw an exception if `type` is null.  Or we could add nullability attributes that specify that when `type` is null, the return value will be also be null.

src/GraphQL/Language/AST/ValueNodes/ValueNode.cs
Shane32: Perhaps there should be a non-null type restriction added to the generic `T` type for v5.

sungam3r: >  If you want to change it, that's fine; I don't care either way.

50/50, nevermind

PR #2601, opened 2021-07-11
Nullable reference type annotations part 4

Shane32:


src/GraphQL/Validation/Rules/5.8 - Variables/2. VariablesAreInputTypes.cs
```diff
@@ -26,7 +28,7 @@ public class VariablesAreInputTypes : IValidationRule
         {
             var type = varDef.Type.NamedGraphTypeFromType(context.Schema);
 
-            if (!type.IsInputType())
+            if (type == null || !type.IsInputType())
```
sungam3r, line 31: `IsInputType` extension method actually can handle `null` from `this` but I'm fine with this change.

src/GraphQL/Validation/Rules/OverlappingFieldsCanBeMerged.cs
```diff
@@ -741,24 +743,24 @@ private static bool SameValue(Argument arg1, Argument arg2)
 
         private sealed class FieldDefPair
         {
-            public IGraphType ParentType { get; set; }
-            public ISelection Field { get; set; }
-            public FieldType FieldDef { get; set; }
+            public IGraphType? ParentType { get; set; } = null!;
```
sungam3r, line 746: ?

sungam3r:
Reviewed, only one "issue" found.

PR #2603, opened 2021-07-12
Update Sample to use a scoped dependency for data, to reflect real world usage

EmmanuelPonnudurai:
I have updated the `StarWarsData` sample to use a more real world sample of how things are usually set up in applications. This can serve as a sample reference implementation.

I noticed that we were using static `StarWarsData` class as a provider of data. 

I've converted the same into an `IStarWarsDataRespository` which mimics a layer responsible for actual business layer for apps.

### DIFF
--- a/src/GraphQL.StarWars/DataRepository/IStarWarsDataRespository.cs
+++ b/src/GraphQL.StarWars/DataRepository/IStarWarsDataRespository.cs
@@ -1,0 +1,19 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using GraphQL.StarWars.Types;
+
+namespace GraphQL.StarWars.DataRepository
+{
+    public interface IStarWarsDataRespository
+    {
+        IEnumerable<StarWarsCharacter> GetFriends(StarWarsCharacter character);
+
+        StarWarsCharacter AddCharacter(StarWarsCharacter character);
+
+        Task<Human> GetHumanByIdAsync(string id);
+
+        Task<Droid> GetDroidByIdAsync(string id);
+
+        Task<List<StarWarsCharacter>> GetCharactersAsync(List<string> guids);
+    }
+}
--- a/src/GraphQL.StarWars/DataRepository/IStarWarsDataRespository.cs
+++ b/src/GraphQL.StarWars/DataRepository/IStarWarsDataRespository.cs
@@ -15,5 +15,7 @@ namespace GraphQL.StarWars.DataRepository
         Task<Droid> GetDroidByIdAsync(string id);
 
         Task<List<StarWarsCharacter>> GetCharactersAsync(List<string> guids);
+
+        Task<List<StarWarsCharacter>> GetAllCharactersAsync();
     }
 }
--- a/src/GraphQL.StarWars/DataRepository/StarWarsDataRespository.cs
+++ b/src/GraphQL.StarWars/DataRepository/StarWarsDataRespository.cs
@@ -92,5 +92,12 @@ namespace GraphQL.StarWars.DataRepository
 
             return Task.FromResult(results);
         }
+
+        public async Task<List<StarWarsCharacter>> GetAllCharactersAsync()
+        {
+            var results = await Task.FromResult(_characters.ToList());
+
+            return results;
+        }
     }
 }
--- a/src/GraphQL.StarWars/DataRepository/StarWarsDataRespository.cs
+++ b/src/GraphQL.StarWars/DataRepository/StarWarsDataRespository.cs
@@ -3,15 +3,13 @@ using System.Linq;
 using System.Threading.Tasks;
 using GraphQL.StarWars.Types;
 
-namespace GraphQL.StarWars
+namespace GraphQL.StarWars.DataRepository
 {
-    public class StarWarsData
+    public class StarWarsDataRespository : IStarWarsDataRespository
     {
-        private readonly List<StarWarsCharacter> _characters = new List<StarWarsCharacter>();
-
-        public StarWarsData()
+        private static readonly List<StarWarsCharacter> _characters = new()
         {
-            _characters.Add(new Human
+            new Human
             {
                 Id = "1",
                 Name = "Luke",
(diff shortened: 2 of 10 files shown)


Shane32:
@EmmanuelPonnudurai I think this is a great idea in concept.  A few thoughts:

1. Should `IStarWarsDataRespository` be registered as scoped?  If this is a sample with scoped services (I see you use the resolver builders with CreateScope) then the repository should probably be registered as scoped.

2. Should we extend this to actually using Entity Framework as the database provider?  There's a way to easily initialize a temporary database, which persists as long as the application is running, using the LocalDb nuget package.

See the sample project within https://github.com/Shane32/GraphQL.DI - relevant code below:

```cs
//startup file
services.AddSingleton(_ => new SqlInstance<TodoDbContext>(builder => new TodoDbContext(builder.Options)));
services.AddSingleton(serviceProvider => serviceProvider.GetRequiredService<SqlInstance<TodoDbContext>>().Build().GetAwaiter().GetResult());
services.AddScoped(serviceProvider => serviceProvider.GetRequiredService<SqlDatabase<TodoDbContext>>().NewDbContext());
```

Then add data initialization code like shown here:
https://github.com/Shane32/GraphQL.DI/blob/master/Sample/TodoDbContext.cs

3. Should we retain the existing sample, and have this be a second duplicate Harness sample?  It seems that a lot of people use singletons exclusively, and the existing sample would be a better (and different) sample.

EmmanuelPonnudurai:
@Shane32 Thanks for checking. 

> 1. Should `IStarWarsDataRespository` be registered as scoped?  If this is a sample with scoped services (I see you use the resolver builders with CreateScope) then the repository should probably be registered as scoped.

This is a total miss on my end. Its odd that I created this and left the registration as Singleton as opposed to Scoped. I shall fix this.

> 2. Should we extend this to actually using Entity Framework as the database provider?  There's a way to easily initialize a temporary database, which persists as long as the application is running, using the LocalDb nuget package.
> 
> See the sample project within https://github.com/Shane32/GraphQL.DI - relevant code below:
> 
> ```cs
> //startup file
> services.AddSingleton(_ => new SqlInstance<TodoDbContext>(builder => new TodoDbContext(builder.Options)));
> services.AddSingleton(serviceProvider => serviceProvider.GetRequiredService<SqlInstance<TodoDbContext>>().Build().GetAwaiter().GetResult());
> services.AddScoped(serviceProvider => serviceProvider.GetRequiredService<SqlDatabase<TodoDbContext>>().NewDbContext());
> ```
> 
> Then add data initialization code like shown here:
> https://github.com/Shane32/GraphQL.DI/blob/master/Sample/TodoDbContext.cs

Ok that does make sense. Might as well spin up an entity framework DB sample implementation to make it even easier to follow. The reason I didn't do this in first cut is that there could be several implementations of ORM's within different apps so didn't want to pick one. But might be useful cos EF Core seems to be at the top (not backed by any statistics I'm just assuming this)

> 1. Should we retain the existing sample, and have this be a second duplicate Harness sample?  It seems that a lot of people use singletons exclusively, and the existing sample would be a better (and different) sample.

I thought the exact opposite. But I'm thinking, how can a real world app have the data provider as a singleton. Seems strange if that is the case.

Shane32:
> But I'm thinking, how can a real world app have the data provider as a singleton.

I believe a common scenario is for people to be writing a GraphQL API layer that simply calls other services via http or similar.  Like a wrapper for a legacy REST API.  As such, they have no need for scoped services.

Another common scenario, I believe, is to use an underlying ORM that does not require a scoped database provider, like linq2db, or if they create and dispose of the database connection within the database layer code.  Even EF can do this, to a certain degree, if the dbcontext is registered as a transient.

My code doesn't work this way, but I think there are a lot of users that do fall into one of those categories, where they do not need scoped services.  There have certainly been multiple of both the above users that have posted questions here.

EmmanuelPonnudurai:
> I believe a common scenario is for people to be writing a GraphQL API layer that simply calls other services via http or similar. Like a wrapper for a legacy REST API. As such, they have no need for scoped services.
> 
> Another common scenario, I believe, is to use an underlying ORM that does not require a scoped database provider, like linq2db, or if they create and dispose of the database connection within the database layer code. Even EF can do this, to a certain degree, if the dbcontext is registered as a transient.
> 
> My code doesn't work this way, but I think there are a lot of users that do fall into one of those categories, where they do not need scoped services. There have certainly been multiple of both the above users that have posted questions here.

Got it. Thanks.

src/GraphQL.StarWars/DataRepository/IStarWarsDataRespository.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using GraphQL.StarWars.Types;
+
+namespace GraphQL.StarWars.DataRepository
+{
+    public interface IStarWarsDataRespository
```
sungam3r, line 7:
```suggestion
    public interface IStarWarsDataRepository
```

src/GraphQL.StarWars/DataRepository/StarWarsDataRespository.cs
```diff
@@ -0,0 +1,103 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using GraphQL.StarWars.Types;
+
+namespace GraphQL.StarWars.DataRepository
+{
+    public class StarWarsDataRespository : IStarWarsDataRespository
```
sungam3r, line 8:
```suggestion
    public class StarWarsDataRepository : IStarWarsDataRepository
```

sungam3r:
> Should we extend this to actually using Entity Framework as the database provider? There's a way to easily initialize a temporary database, which persists as long as the application is running, using the LocalDb nuget package.

EF may be simply targeted against memory storage aw well - https://docs.microsoft.com/en-us/ef/core/providers/in-memory/?tabs=dotnet-core-cli

Shane32:
> EF may be simply targeted against memory storage

True, but queries don't execute the same.  Probably not an issue here, if we want to use EF simply as a sample.  And LocalDb only works on Windows systems.

sungam3r:
@EmmanuelPonnudurai Will you still work on it?

sungam3r:
Closed due to inactivity.

EmmanuelPonnudurai:
I have added a more succinct sample here - https://github.com/graphql-dotnet/examples/pull/121 

I also realized that I totally missed this part. So which ever one you want me to move forward with, im good. I can do both as well. Pls let me know @sungam3r @Shane32

PR #2606, opened 2021-07-13
Add relay and generic graph types to default registrations

Shane32:


src/GraphQL/DI/GraphQLBuilderBase.cs
```diff
@@ -38,6 +41,18 @@ protected virtual void Initialize()
             this.TryRegister<IDocumentCache>(DefaultDocumentCache.Instance);
             this.TryRegister<IErrorInfoProvider, ErrorInfoProvider>(ServiceLifetime.Singleton);
 
+            // configure relay graph types
+            TryRegister(typeof(EdgeType<>), typeof(EdgeType<>), ServiceLifetime.Transient);
+            TryRegister(typeof(ConnectionType<>), typeof(ConnectionType<>), ServiceLifetime.Transient);
+            TryRegister(typeof(ConnectionType<,>), typeof(ConnectionType<,>), ServiceLifetime.Transient);
+            this.TryRegister<PageInfoType>(ServiceLifetime.Transient);
```
sungam3r, line 48: Why do you use extension method here but not above ? I would like uniformity.

Shane32: The compiler will not accept an open generic as a type argument. So the above cannot change.  Since the generic method is much cleaner, I use it whenever possible.

sungam3r: ok

PR #2607, opened 2021-07-13
Expose Services within GraphQLBuilder

Shane32:
Allows extension methods to cast `IGraphQLBuilder` to `GraphQLBuilder` and retrieve the `IServiceCollection` via `Services`.  Works for only Microsoft-based DI of course, but this is needed for authorization extension methods that integrate with ASP.Net within the server project.


sungam3r:
`GraphQLBuilder` may be reverted to `internal` again after #2771

PR #2611, opened 2021-07-15
Nullable reference type annotations part 6

Shane32:
This concludes the first pass of NRT annotations for GraphQL.NET.  Many changes should be made, but would be breaking.  Most specifically, things like ResolvedType and Source which are typically non-null, yet can be at other times.  (e.g. ResolvedType before schema is initialized, and Source when it equals the root object).  There's also lots of data objects which start off null even though null is invalid, such as the name of a type.  Most of those I set as non-null.  There is very little code that I changed, rather I just added `!` where necessary to skip NRT checks.  So almost every reference to ResolvedType has now changed to `ResolvedType!`.  Another property of that nature is `Parent` -- almost always non-null, but not always.  So that's nullable.

I'm sure there will be some incorrect NRT annotations, but this is a start.  Again, the code should perform the same since it was not changed (except in a few rare circumstances).  So we can just pull requests to fix any incorrect NRT annotations.  Maybe for the next version we can use `init` properties so these NRT situations are less likely to occur.

### DIFF
--- a/src/GraphQL/GraphQL.csproj
+++ b/src/GraphQL/GraphQL.csproj
@@ -6,6 +6,7 @@
     <Authors>Joe McBride</Authors>
     <Copyright>Copyright (c) 2015-2021 Joseph T. McBride, Ivan Maximov, Shane Krueger, et al. All rights reserved.</Copyright>
     <PackageTags>GraphQL;json;api</PackageTags>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/GraphQL/Introspection/__Schema.cs
+++ b/src/GraphQL/Introspection/__Schema.cs
@@ -57,7 +57,7 @@ namespace GraphQL.Introspection
                 "If this server supports mutation, the type that mutation operations will be rooted at.",
                 resolve: async context =>
                 {
-                    return await context.Schema.Filter.AllowType(context.Schema.Mutation).ConfigureAwait(false)
+                    return context.Schema.Mutation != null && await context.Schema.Filter.AllowType(context.Schema.Mutation).ConfigureAwait(false)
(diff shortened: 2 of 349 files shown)


sungam3r:
изображение
😮

src/GraphQL/GraphQL.csproj
```diff
@@ -6,6 +6,7 @@
     <Authors>Joe McBride</Authors>
     <Copyright>Copyright (c) 2015-2021 Joseph T. McBride, Ivan Maximov, Shane Krueger, et al. All rights reserved.</Copyright>
     <PackageTags>GraphQL;json;api</PackageTags>
+    <Nullable>enable</Nullable>
```
sungam3r, line 9: Worth to move into Directory.Build.props

src/GraphQL/Introspection/__Schema.cs
```diff
@@ -67,7 +67,7 @@ public __Schema(bool allowAppliedDirectives = false)
                 "If this server supports subscription, the type that subscription operations will be rooted at.",
                 resolve: async context =>
                 {
-                    return await context.Schema.Filter.AllowType(context.Schema.Subscription).ConfigureAwait(false)
+                    return context.Schema.Subscription != null && await context.Schema.Filter.AllowType(context.Schema.Subscription).ConfigureAwait(false)
```
sungam3r, line 70: 👍

src/GraphQL/GraphQL.csproj
Shane32: Still not applied to every project 100%. It should be. Probably not much work left

sungam3r: Can be disabled for tests.

sungam3r:
reviewed

PR #2612, opened 2021-07-19
IGraphQLBuilder improvements

Shane32:
- Allow casting IGraphQLBuilder to IServiceCollection so that a reference to GraphQL.MicrosoftDI is not needed if it is necessary to access the underlying IServiceCollection
- Allow asynchronous execution configurations
- Allow service registrations that replace existing registrations


sungam3r:
reviewed

PR #2619, opened 2021-08-06
Allow overriding SchemaTypes.GetGraphType

Shane32:
This allows overriding `SchemaTypes.GetGraphType` for a custom implementation.  For example, one may want to automatically map unmapped CLR types to `AutoRegisteringObjectGraphType<>`.  With this change, it is possible with just a few lines of code.

### DIFF
--- a/src/GraphQL/Types/Schema.cs
+++ b/src/GraphQL/Types/Schema.cs
@@ -145,7 +145,7 @@ namespace GraphQL.Types
                 if (Initialized)
                     return;
 
-                CreateSchemaTypes();
+                CreateAndInitializeSchemaTypes();
 
                 Initialized = true;
             }
@@ -354,7 +354,7 @@ namespace GraphQL.Types
                 throw new ObjectDisposedException(nameof(Schema));
         }
 
-        private void CreateSchemaTypes()
+        private void CreateAndInitializeSchemaTypes()
         {
             IEnumerable<ISchemaNodeVisitor> GetVisitors()
             {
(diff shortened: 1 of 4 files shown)


src/GraphQL/Types/Schema.cs
```diff
@@ -371,16 +371,36 @@ IEnumerable<ISchemaNodeVisitor> GetVisitors()
...
+        /// <summary>
+        /// Creates and returns a new instance of <see cref="SchemaTypes"/> for this schema.
+        /// Does not apply middleware, apply schema visitors, or validate the schema.
+        /// </summary>
+        /// <remarks>
+        /// This executes within a lock in <see cref="Initialize"/>.
+        /// </remarks>
+        protected virtual SchemaTypes CreateSchemaTypes()
+        {
+            return new SchemaTypes(this, _services);
```
sungam3r, line 403: Overridden method still can not get _services.

sungam3r:
Did you document this feature?

src/GraphQL/Types/Schema.cs
Shane32: True but if they inherit from Schema they can capture the service provider in their own constructor

Shane32:
Probably not, other than in xml comments. Not really sure where such documentation might belong. Since this really all has to do with custom implementations of SchemaTypes, and since we have no documentation on creating new implementations of GraphQL.NET inner workings … although it of course would be nice if this was all documented with examples and stuff.

PR #2634, opened 2021-08-23
fix 'DateTimeOffset's Offset loss

paul804:
`DateTimeOffsetGraphType.ParseDate` drops the offset and there is no way to get the original client's offset. I've removed flag ` DateTimeStyles.AdjustToUniversal` and changed test: later it checked using `DateTimeOffset.Equals` that compares UTC time. Now it checks properties separately to avoid data loss.

### DIFF
--- a/src/GraphQL.Tests/Types/DateTimeOffsetGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/DateTimeOffsetGraphTypeTests.cs
@@ -51,8 +51,10 @@ namespace GraphQL.Tests.Types
         {
             CultureTestHelper.UseCultures(() =>
             {
-                _type.ParseValue("2015-11-21T19:59:32.987+0200").ShouldBe(
-                    new DateTimeOffset(2015, 11, 21, 19, 59, 32, 987, TimeSpan.FromHours(2)));
+                var dateTimeOffset = (DateTimeOffset)_type.ParseValue("2015-11-21T19:59:32.987+0200");
+                dateTimeOffset.Date.ShouldBe(new DateTime(2015, 11, 21));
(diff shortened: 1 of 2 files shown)


Shane32:
Thanks for the feedback @ificator .  It's important to consider both viewpoints, especially when making a change that could go unnoticed and cause data errors with existing implementations.

Just for reference, we do have a documented method of replacing built-in scalars with custom scalars.  For example, the 4.x boolean graph type only allows native `true` and `false` values, while the 3.x version allowed interpreting strings such as "True" and "False" as boolean values.  So even if this change was made without an easy way to opt-in/opt-out, one can still revert to the prior behavior by implementing a custom scalar (probably by copying the source code of the prior version), which would require no changes to any existing graphs or logic.

https://graphql-dotnet.github.io/docs/getting-started/custom-scalars

This type of change, if implemented, would certainly be documented in the migration notes, which are always very important to be carefully read by anyone migrating to a new major version.

Personally, if your application must store `DateTimeOffset` values based on UTC time, I would suggest using the `DateTimeOffset.UtcDateTime` property to explicitly retrieve the UTC date/time value to be stored from the input argument.  I understand of course that there may be reasons that you do not wish to make these changes.

> Provide a means to opt-in to the existing behavior

This would make sense if this change was implemented within v4, but if it is changed in v5, then I would think that it should be opt-out rather than opt-in, and noted as a breaking change.  As described above, the 'opt-out' behavior might be replacing the built-in scalar with a custom scalar, but if we receive enough feedback, including a static property to opt-out is another option.

Shane32:
@sungam3r Please review this to merge into develop for v5

sungam3r (approved):

PR #2660, opened 2021-09-21
Update InputFieldsAndArgumentsOfCorrectLength.cs

bamancio-futura:
Check against interface instead of concrete implementation type.


Shane32 (approved):

sungam3r:
@bamancio-futura Thanks! I have long noticed this mistake, but I lacked time to make this one-symbol change. 👍

PR #2679, opened 2021-11-01
SchemaValidationVisitor now checks that ResolvedType was specified for Lists and NonNulls

### LINKED ISSUE #2675: Dynamically Generating Complex Input Arguments not Generating a Valid Schema.
I am dynamically generating my GraphQL schema from the equivalent of a database. It is important that I can dynamically generate so that it is easy to design the look and feel of our real GraphQL API by changing the database structure, putting some sample data in and then letting my content owners play with the output. I know, API first would be better, but, alas, this is data first design for good or bad.

I am finding different behaviour when trying to dynamically set complex input types. I think that I might have figured out how I can work around this issue, but there still seems to be a bug.

I am using GraphQL 4.6.1

If I do something like this:

```cs
new QueryArgument<ListGraphType<NonNullGraphType<StringGraphType>>> { Name = "objectIds", Description = "Unique id of the record" })
```

Which is not completely dynamic, there are certain queries that always have this 'objectIds' argument. The definition of this argument is static, it is dynamically added to my queries. This list of not null strings looks fine in my generated schema (using the pretty printer):

```gql
type Root {
  TestOT(objectIds: [String!], limit: Int, filter: InputString): [TestOT]
}
```

If I try and do something more dynamic that looks like:

```cs
InputObjectGraphType stringFilterInputType = new InputObjectGraphType { Name = "InputString" };

stringFilterInputType.AddField(new FieldType
{
    Name = "EQ",
    ResolvedType = new StringGraphType()
});
stringFilterInputType.AddField(new FieldType
{
   Name = "IN",
   ResolvedType = new ListGraphType<StringGraphType>()
});
stringFilterInputType.AddField(new FieldType
{
   Name = "NOT",
   ResolvedType = new NonNullGraphType<StringGraphType>()
});
```

This generates a broken schema that also breaks GraphiQL 'docs' when it does introspection. The generated type for this InputObjectType looks like:

```gql
input InputString {
  EQ: String
  IN: []
  NOT: !
}
```

What happened to my ListGraphType<StringGraphType> or my NonNullGraphType<StringGraphType>?

If I don't use the generic constructor for ListGraphType or NonNullGraphType, then I can get this to work. So, if I change my code to:

```cs
var stringType = new StringGraphType();
InputObjectGraphType stringFilterInputType = new InputObjectGraphType { Name = "InputString" };
stringFilterInputType.AddField(new FieldType
{
    Name = "EQ",
    ResolvedType = new StringGraphType()
});
stringFilterInputType.AddField(new FieldType
{
     Name = "IN",
     ResolvedType = new ListGraphType(stringType)
});
stringFilterInputType.AddField(new FieldType
{
     Name = "NOT",
     ResolvedType = new NonNullGraphType(stringType)
});
```

When I run the server now and look at the pretty printed SDL, I get something much better:

```gql
input InputString {
  EQ: String
  IN: [String]
  NOT: String!
}
```

Looking at the code I thought that:

```cs
var stringType = new StringGraphType();
var argument = new ListGraphType(stringType)
```

should be equivalent to:

```cs
var argument = new ListGraphType<StringGraphType>()
```

And, I also think that it should be impossible for the server to start if the schema is invalid, so, no matter how I managed to get a schema that looked like:

```gql
input InputString {
  EQ: String
  IN: []
  NOT: !
}
```

Then there should have been an exception thrown somewhere upon startup.

sungam3r:
fixes #2675

### DIFF
--- a/src/GraphQL.Tests/Bugs/Issue1837.cs
+++ b/src/GraphQL.Tests/Bugs/Issue1837.cs
@@ -70,7 +70,7 @@ namespace GraphQL.Tests.Bugs
 
             AddField(new FieldType
             {
-                ResolvedType = new ListGraphType<IntGraphType>(),
+                ResolvedType = new ListGraphType<IntGraphType> { ResolvedType = new IntGraphType() },
                 Name = "abc"
             });
         }
--- a/src/GraphQL/Validation/ValidationContext.cs
+++ b/src/GraphQL/Validation/ValidationContext.cs
@@ -570,7 +570,7 @@ namespace GraphQL.Validation
                     : $"Expected type '{type.Name}', found {valueAst.StringFrom(Document)}.";
             }
 
-            throw new ArgumentOutOfRangeException(nameof(type), $"Type {type?.Name} is not a valid input graph type.");
+            throw new ArgumentOutOfRangeException(nameof(type), $"Type {type?.Name ?? "<NULL>"} is not a valid input graph type.");
         }
     }
 }
--- a/src/GraphQL.Tests/Initialization/SchemaInitializationTests.cs
+++ b/src/GraphQL.Tests/Initialization/SchemaInitializationTests.cs
@@ -60,6 +60,12 @@ namespace GraphQL.Tests.Initialization
         {
             ShouldThrow<SchemaWithArgumentsOnInputField, InvalidOperationException>("The field 'id' of an Input Object type 'MyInput' must not have any arguments specified.");
         }
+
+        [Fact]
+        public void SchemaWithNotFullSpecifiedResolvedType_Should_Throw()
+        {
+            ShouldThrow<SchemaWithNotFullSpecifiedResolvedType, InvalidOperationException>("The field 'in' of an Input Object type 'InputString' must have non-null 'ResolvedType' property for all types in the chain.");
+        }
     }
 
     public class EmptyQuerySchema : Schema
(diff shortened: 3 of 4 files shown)


src/GraphQL.Tests/Bugs/Issue1837.cs
```diff
@@ -70,7 +70,7 @@ public Issue1837ArrayInputType()
 
             AddField(new FieldType
             {
-                ResolvedType = new ListGraphType<IntGraphType>(),
+                ResolvedType = new ListGraphType<IntGraphType> { ResolvedType = new IntGraphType() },
```
sungam3r, line 73: Caught!

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -570,7 +570,7 @@ public string IsValidLiteralValue(IGraphType type, IValue valueAst)
                     : $"Expected type '{type.Name}', found {valueAst.StringFrom(Document)}.";
             }
 
-            throw new ArgumentOutOfRangeException(nameof(type), $"Type {type?.Name} is not a valid input graph type.");
+            throw new ArgumentOutOfRangeException(nameof(type), $"Type {type?.Name ?? "<NULL>"} is not a valid input graph type.");
```
sungam3r, line 573: @Shane32  Previously, there was an strange message with empty type, because `type` was `null`. I added `<NULL>` literal to explicitly state this but now with new checks inside schema validation visitor it should not happen. Of cource one may "turn off" our validation so better safe than sorry.

Shane32 (approved):

PR #2693, opened 2021-11-17
Add SchemaTypes.Initialize

Shane32:
The current design of `SchemaTypes` is such that no properties can be set on derived classes before initialization occurs.  For example, this code doesn't work:

```cs
private class MySchemaTypes : SchemaTypes
{
    private readonly Dictionary<Type, Type> _mapping;

    public MySchemaTypes(ISchema schema, IServiceProvider serviceProvider, Dictionary<Type, Type> customMapping)
        : base(schema, serviceProvider)
    {
        _mapping = customMapping;
    }

    protected override Type? GetGraphTypeFromClrType(Type clrType, bool isInputType, List<(Type ClrType, Type GraphType)> typeMappings)
    {
        var ret = base.GetGraphTypeFromClrType(clrType, isInputType, typeMappings);
        if (ret != null) return ret;

        //custom logic here, which requires values passed into the constructor
        if (!isInputType)
            return _mapping[clrType]; //throws null reference exception!!!

        return null;
    }
}
```

This PR fixes that by adding a protected `Initialize(schema, serviceProvider)` method, which can be called by the default constructor like so:

```cs
private class MySchemaTypes : SchemaTypes
{
    private readonly Dictionary<Type, Type> _mapping;

    public MySchemaTypes(ISchema schema, IServiceProvider serviceProvider, Dictionary<Type, Type> customMapping)
    {
        _mapping = customMapping;
        Initialize(schema, serviceProvider);
    }

    protected override Type? GetGraphTypeFromClrType(Type clrType, bool isInputType, List<(Type ClrType, Type GraphType)> typeMappings)
    {
        var ret = base.GetGraphTypeFromClrType(clrType, isInputType, typeMappings);
        if (ret != null) return ret;

        //custom logic here, which requires values passed into the constructor
        if (!isInputType)
            return _mapping[clrType]; //works now

        return null;
    }
}
```

### DIFF
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -104,18 +104,22 @@ namespace GraphQL.Types
             Initialize(schema, serviceProvider);
         }
 
+        private bool _initialized = false;
         /// <summary>
         /// Initializes the instance for the specified schema, and with the specified type resolver.
         /// </summary>
         /// <param name="schema">A schema for which this instance is created.</param>
         /// <param name="serviceProvider">A service provider used to resolve graph types.</param>
         /// </summary>
-        protected virtual void Initialize(ISchema schema, IServiceProvider serviceProvider)
+        protected void Initialize(ISchema schema, IServiceProvider serviceProvider)
         {
             if (schema == null)
                 throw new ArgumentNullException(nameof(schema));
             if (serviceProvider == null)
                 throw new ArgumentNullException(nameof(serviceProvider));
+            if (_initialized)
+                throw new InvalidOperationException("SchemaTypes has already been initialized.");
+            _initialized = true;
 
             var types = GetSchemaTypes(schema, serviceProvider);
             var typeMappingsEnumerable = schema.TypeMappings ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.TypeMappings));
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -45,7 +45,7 @@ namespace GraphQL.Types
         };
 
         // Introspection types http://spec.graphql.org/June2018/#sec-Schema-Introspection
-        private readonly Dictionary<Type, IGraphType> _introspectionTypes;
+        private Dictionary<Type, IGraphType> _introspectionTypes;
 
         // Standard scalars https://graphql.github.io/graphql-spec/June2018/#sec-Scalars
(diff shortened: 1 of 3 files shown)


src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -234,7 +247,7 @@ private static IEnumerable<IGraphType> GetSchemaTypes(ISchema schema, IServicePr
         /// Returns a dictionary that relates type names to graph types.
         /// </summary>
         protected internal virtual Dictionary<string, IGraphType> Dictionary { get; } = new Dictionary<string, IGraphType>();
-        private readonly Dictionary<Type, IGraphType> _typeDictionary;
```
sungam3r, line 237: I don't like that fields lost `readonly` modifier.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -97,7 +97,20 @@ protected SchemaTypes()
...
+            Initialize(schema, serviceProvider);
+        }
+
+        /// <summary>
+        /// Initializes the instance for the specified schema, and with the specified type resolver.
+        /// </summary>
+        /// <param name="schema">A schema for which this instance is created.</param>
+        /// <param name="serviceProvider">A service provider used to resolve graph types.</param>
+        /// </summary>
+        protected virtual void Initialize(ISchema schema, IServiceProvider serviceProvider)
```
sungam3r, line 113: 1. virtual ?
2. Add some sort of check if `Initialize` has already been called.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Agreed, but there's nothing I can do about it.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: While I could remove `virtual`, as a matter of principle, all `protected` members should be `virtual`.  It doesn't really matter here, as you can run code in the constructor before or after `Initialize()`, or replace `Initialize()`.  Do you want me to remove `virtual` here?  (I plan to rewrite this entire class at some point anyway, hopefully before v5, but not sure if I have time.)

Shane32: I will add the check for Initialize.

sungam3r: > Do you want me to remove virtual here?

Yes.

Shane32: Both changes done

sungam3r (approved):

PR #2694, opened 2021-11-17
Enable SchemaPrinter to sort output

### LINKED ISSUE #2685: SchemaPrinter does not sort field names
`SchemaPrinter` sorts directives and types by name, but does not sort field names, argument names, enumeration members, and so on.  It does not respect the `ISchemaComparer` set on the schema, and there is no setting to change this.

I suggest that a new property on `SchemaPrinterOptions` be added called `Comparer` which accepts a `ISchemaComparer` and uses it to print the schema.  This can be added to v4 without a breaking change.  Probably 4.7 since this changes the exposed API.

Perhaps in v5, the default behavior should be to respect the schema's comparer property, but allow for overriding the comparer if necessary.  I expect that in most of my applications, I will want the introspection verification test to use a sorted schema, but an introspection query would remain unsorted.

Shane32:
Fixes #2685 

Note: due to the limitations of `ISchemaComparer`, there is no way to sort directive locations or directive arguments.  Changing an interface signature is a breaking change and will need to be done in v5; an issue will be created to address this.  I think this is still worthwhile to include the other changes in v4.

Another temporary solution would be to evaluate `Options.Comparer is AlphabeticalSchemaComparer`, and if so, sort the members alphabetically.  I did not do so.

### DIFF
--- a/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
@@ -1750,7 +1750,7 @@ enum Beta {
 }
 ");
             var printer = new SchemaPrinter(schema, new SchemaPrinterOptions { Comparer = new GraphQL.Introspection.AlphabeticalSchemaComparer() });
-            var ret = printer.Print();
+            var actual = printer.Print();
             var expected = @"directive @test1(
   arg2: Boolean!
   arg1: Boolean!
@@ -1787,9 +1787,8 @@ type Zebra {
   field3: String
 }
 ";
-            ret.ShouldBe(expected, StringCompareShould.IgnoreLineEndings);
+            actual.ShouldBe(expected, StringCompareShould.IgnoreLineEndings);
         }
-
 
         public class FooType : ObjectGraphType
         {
--- a/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
@@ -1700,6 +1700,97 @@ enum __TypeKind {
             printer.PrintDescription("Th\\is \"is\" a \"\"\"test\n\tline2\n line3\u0003").ShouldBeCrossPlat("\"\"\"\nTh\\is \"is\" a \\\"\"\"test\n\tline2\n line3\n\"\"\"\n");
         }
 
+        [Fact]
+        public void sorts_schema_correctly()
+        {
+            var schema = Schema.For(@"
+# test sorting type names
+type Zebra {
+  # test sorting field names on object types
+  field3: String
+  field2: Int
+}
+
+type Query {
+  field1(arg1: Rutabaga, arg2: Beta): Zebra
+  # test sorting arguments
+  field2(arg2: Rutabaga, arg1: Beta): Tango
+  # test sorting fields of default values
+  field3(arg3: Rutabaga = { field3: ""hello"", field2: 2 }): String
+}
+
+type Tango {
+  field1: Int
+  field2: Int
+}
+
+input Rutabaga {
+  # test sorting field names on input types
+  field3: String
+  field2: Int
+}
+
+# test sorting directives
+directive @test2(
+  arg1: Boolean!
+  arg2: Boolean!
+  # test sorting directive locations -- does not work yet
+) on INLINE_FRAGMENT | FIELD | FRAGMENT_SPREAD
+
+directive @test1(
+  # test sorting fields within directives -- does not work yet
+  arg2: Boolean!
+  arg1: Boolean!
+) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+enum Beta {
+  # test sorting of enum value names
+  VALUE_3
+  VALUE_2
+}
+");
+            var printer = new SchemaPrinter(schema, new SchemaPrinterOptions { Comparer = new GraphQL.Introspection.AlphabeticalSchemaComparer() });
+            var ret = printer.Print();
+            var expected = @"directive @test1(
+  arg2: Boolean!
+  arg1: Boolean!
+) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+directive @test2(
+  arg1: Boolean!
+  arg2: Boolean!
+) on INLINE_FRAGMENT | FIELD | FRAGMENT_SPREAD
+
+enum Beta {
+  VALUE_2
+  VALUE_3
+}
+
+type Query {
+  field1(arg1: Rutabaga, arg2: Beta): Zebra
+  field2(arg1: Beta, arg2: Rutabaga): Tango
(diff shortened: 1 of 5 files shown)


src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -338,21 +347,25 @@ private string FormatInputObjectValue(object value, IInputObjectGraphType input)
...
             {
                 string propertyName = field.GetMetadata<string>(ComplexGraphType<object>.ORIGINAL_EXPRESSION_PROPERTY_NAME) ?? field.Name;
-                PropertyInfo propertyInfo;
 
-                try
-                {
-                    propertyInfo = value.GetType().GetProperty(propertyName, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance);
-                }
-                catch (AmbiguousMatchException)
+                if (!(value is IDictionary<string, object?> dic && dic.TryGetValue(propertyName, out var propertyValue)))
```
Shane32, line 354: This is a bug fix: SchemaPrinter would not print a schema with a default input object value that was parsed by Schema.For.

src/GraphQL/Utilities/SchemaPrinterOptions.cs
```diff
@@ -24,5 +26,11 @@ public class SchemaPrinterOptions
         /// Indicates whether to print descriptions as comments for compatibility with the 2016 GraphQL specification.
         /// </summary>
         public bool PrintDescriptionsAsComments { get; set; } = true;
+
+        /// <summary>
+        /// Provides the ability to order the schema elements upon printing.
```
sungam3r, line 31: Lets change comment for `ISchemaComparer`:
>  Provides the ability to order the schema elements upon introspection.
->
>  Provides the ability to order the schema elements upon introspection or printing.

src/GraphQL/Utilities/SchemaPrinter.cs
sungam3r: Could you explain in more details?

Shane32:
> @Shane32 How do you think to get rid of `SchemaPrinter` "internal printing" at all? I mean to switch to implementation from parser. Here we duplicate essentially the same code. Note that there will have to solve the same problems with descriptions, block strings, comments, string escaping and so on. And now printing is far from optimal here - too many concatenations.
> 
> What I suggest is to switch to building AST (using appropriate order of sibling nodes) and then write this AST via `SDLWriter`. `SchemaPrinter` as a class may stay.

In concept, I'm fine with it, but we may run into problems trying to implement the FederatedSchemaPrinter.  Bottom line is that I think we should retain the SchemaPrinter class and existing feature set, even if behind-the-scenes it builds an AST and uses the parser to print the AST.

However, let's not hold up this PR/feature because we want to rewrite the SchemaPrinter.

src/GraphQL/Utilities/SchemaPrinterOptions.cs
Shane32: Done.

src/GraphQL/Utilities/SchemaPrinter.cs
Shane32: Given this schema:

```gql
input SomeInput1 {
  field1: SomeInput2 = { age: 22 }
}
input SomeInput2 {
  age: Int
}
```

When creating the schema in code-first, the default value would likely be `new SomeInput2 { age = 22 }`.  However, when parsing the above text via `Schema.For`, the default value is `new Dictionary<string, object?> { { "age", 22 } }`.  The default value is valid and `GetArgument<SomeInput2>` will coerce it to the proper type.  However, the schema printer currently expects a code-first object and does not expect a dictionary.  This code change fixes that problem so it can be printed.

sungam3r:
> let's not hold up this PR/feature

Of course.

src/GraphQL/Utilities/SchemaPrinter.cs
sungam3r: So the if branch is for code-first "object" value to take some property from it. And then we have either that value or just "plain" value from dictionary. Right? Worth to add comments.

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -338,28 +347,37 @@ private string FormatInputObjectValue(object value, IInputObjectGraphType input)
...
+                    catch (AmbiguousMatchException)
+                    {
+                        propertyInfo = value.GetType().GetProperty(propertyName, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
+                    }
+
+                    propertyValue = propertyInfo.GetValue(value);
                 }
 
-                object propertyValue = propertyInfo.GetValue(value);
-                if (propertyValue != null)
```
Shane32, line 356: Skipping null members can produce invalid schema, as seen by the added test.

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -338,28 +347,37 @@ private string FormatInputObjectValue(object value, IInputObjectGraphType input)
...
                 string propertyName = field.GetMetadata<string>(ComplexGraphType<object>.ORIGINAL_EXPRESSION_PROPERTY_NAME) ?? field.Name;
-                PropertyInfo propertyInfo;
 
-                try
+                // if 'value' is stored as a dictionary of key/value pairs, pull the field value from the dictionary by the property name
+                object? propertyValue;
+                if (value is IDictionary<string, object?> dic)
                 {
-                    propertyInfo = value.GetType().GetProperty(propertyName, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance);
+                    // note: per spec, unspecified properties may not exist within the dictionary
```
Shane32, line 358: Had to change the logic due to this condition.  Updated test.

src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
```diff
@@ -854,6 +854,75 @@ public void prints_input_type_with_default()
...
+        public void prints_input_type_with_default_as_dictionary_null_values()
+        {
+            var schema = Schema.For(@"
+input SomeInput {
+  age: Int = 2
+}
+
+type Query {
+  str(arg: SomeInput = { age: null }): String!
+}
```
Shane32, line 905: As seen here, `age` is necessary to be printed in the schema.

src/GraphQL/Utilities/SchemaPrinter.cs
Shane32: Comments added; please review.  Had to update logic because found two more bugs.

src/GraphQL/Utilities/SchemaPrinter.cs
sungam3r: 👍

sungam3r (approved):

PR #2696, opened 2021-11-17
Fix for 2695 - Schema.For does not always decode default values correctly

### LINKED ISSUE #2695: Schema.For does not always decode default values correctly
If a default value is dependent on another default value, Schema.For may decode (convert from AST to value representation) default values in the wrong order.

For example:

```gql
type Query {
    post(arg: SomeInputType = { name: ""John"" }): String
}

input SomeInputType {
    id: ID! = 1
    name: String!
}
```

In the above sample, the default value of `post` is coerced from an AST to a value prior to the default value of `id`.  So the default value for input argument `arg` is a `Dictionary<string, object>` with `name: "John"` but `id` equals an AST node instead of `1`.

At present, this is due to the validation walker coercing the default values in the order that it walks the node.

To fix this, the schema builder should do the following prior to the schema validation:

1. Scan all input types for fields with default values.
2. Sort the list as necessary based on dependencies.
3. Coerce default values in proper order, or throw an exception if there is a circular dependency based on default values.

So long as all the input types have been coerced, the remaining default values (for directives, arguments, variables, etc) will parse correctly.

Shane32:
Fixes #2695

### DIFF
--- a/src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
@@ -548,6 +548,21 @@ namespace GraphQL.Tests.Utilities
             Should.Throw<InvalidOperationException>(() => schema.Initialize()).Message.ShouldBe("Default values in input types cannot contain a circular dependency loop. Please resolve dependency loop between the following types: 'SomeInputType3', 'SomeInputType2', 'SomeInputType1'.");
         }
 
+        [Fact(Skip = "Not yet supported")]
+        public void input_types_default_value_loops_pass()
+        {
+            // see: https://github.com/graphql-dotnet/graphql-dotnet/pull/2696#discussion_r764975585
+
+            var definitions = @"
+                input SomeInputType1 {
+                  test: SomeInputType1 = { test: null }
+                }
+            ";
+
+            var schema = Schema.For(definitions);
+            schema.Initialize();
+        }
+
         [Fact]
         public void builds_directives()
         {
--- a/src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using GraphQL.Language.AST;
--- a/src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
 using GraphQL.Language.AST;
 using GraphQL.StarWars.Types;
@@ -227,6 +230,39 @@ namespace GraphQL.Tests.Utilities
         }
 
         [Fact]
+        public void builds_type_with_arguments_with_dependent_default_values()
+        {
+            var definitions = @"
+                type Query {
+                  post(arg: [SomeInputType] = [{ name: ""John"" }]): String
+                }
+                input SomeInputType {
+                  id: ID! = 1
+                  name: String!
+                }
+            ";
+
+            var schema = Schema.For(definitions);
+            schema.Initialize();
+
+            var query = schema.Query;
+            query.Fields.Count.ShouldBe(1);
+
+            var field = query.Fields.Single();
+            field.Name.ShouldBe("post");
+            field.Arguments.Count.ShouldBe(1);
+
+            var arg = field.Arguments.First();
+            arg.Name.ShouldBe("arg");
+            var list = arg.DefaultValue.ShouldBeAssignableTo<IEnumerable<object>>();
+            list.Count().ShouldBe(1);
+            var item = list.First().ShouldBeAssignableTo<IDictionary<string, object>>();
+            item.Count.ShouldBe(2);
+            item["name"].ShouldBeOfType<string>().ShouldBe("John");
+            item["id"].ShouldBeOfType<int>().ShouldBe(1);
+        }
+
+        [Fact]
         public void builds_interface()
         {
             var definitions = @"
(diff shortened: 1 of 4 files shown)


src/GraphQL/Types/Schema.cs
```diff
@@ -411,8 +412,41 @@ protected virtual void Validate()
...
+        {
+            var completed = new HashSet<IInputObjectGraphType>();
+            var inProcess = new Stack<IInputObjectGraphType>();
+            foreach (var type in AllTypes.Dictionary.Values)
+            {
+                if (type is IInputObjectGraphType inputType)
+                    ExamineType(inputType, completed, inProcess);
+            }
+
+            static void ExamineType(IInputObjectGraphType inputType, HashSet<IInputObjectGraphType> completed, Stack<IInputObjectGraphType> inProcess)
```
Shane32, line 430: Is there a speed difference between this and:

```suggestion
            void ExamineType(IInputObjectGraphType inputType)
```

?  If the compiler produces the same output either way, then the suggestion would be cleaner.

Shane32:
@sungam3r Do you have any comments on this PR before I merge it?  It seems like a pretty simple fix, although there are probably a few different ways it could be implemented.

sungam3r:
Did you see https://github.com/graphql/graphql-spec/pull/793 ?

src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
```diff
@@ -432,6 +467,87 @@ Stars description
...
+
+            var inputType2 = schema.AllTypes["SomeInputType2"] as InputObjectGraphType;
+            inputType2.ShouldNotBeNull();
+            inputType2.Fields.Count.ShouldBe(2);
+            inputType2.Fields.Find("arg1").ShouldNotBeNull().DefaultValue.ShouldBeNull();
+            inputType2.Fields.Find("arg2").ShouldNotBeNull().DefaultValue.ShouldBe(30);
+        }
+
+        [Fact]
+        public void input_types_default_value_loops_throw1()
```
sungam3r, line 520: Did you see https://github.com/graphql/graphql-spec/pull/793#issuecomment-726811961 ?

Worth to asd test like 
```graphql
input O {
  o: O = { o: null }
  a: Int
}
```

src/GraphQL/Types/Schema.cs
sungam3r: If compiler in both ways generate static method with struct arg then I would prefer this suggestion. I think that explicit `static` keyword for local functions helps to see unintentional arguments closure.

src/GraphQL/Types/Schema.cs
```diff
@@ -411,8 +412,41 @@ protected virtual void Validate()
...
+                    throw new InvalidOperationException($"Default values in input types cannot contain a circular dependency loop. Please resolve dependency loop between the following types: {string.Join(", ", inProcess.Select(x => $"'{x.Name}'"))}.");
+                inProcess.Push(inputType);
+                foreach (var field in inputType.Fields)
+                {
+                    if (field.DefaultValue is GraphQLValue value)
+                    {
+                        var baseType = field.ResolvedType!.GetNamedType();
+                        if (baseType is IInputObjectGraphType inputFieldType)
+                            ExamineType(inputFieldType, completed, inProcess);
+                        field.DefaultValue = Execution.ExecutionHelper.CoerceValue(field.ResolvedType!, Language.CoreToVanillaConverter.Value(value)).Value;
```
sungam3r, line 444: What about `ExecutionHelper.CoerceValue` calls from `SchemaValidationVisitor` ? Keep as is?

src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
sungam3r: Also I see in their discussion those defaults : `= { }` . What does that mean ?

Shane32: The example you stated is not currently supported.  At present, it will throw a friendly error message stating that circular dependencies are not supported.

Supporting circular dependencies (when possible) is outside of my goal for this PR.  Rather, the goal is to prevent out-of-order dependencies from crashing the GraphQL.NET infrastructure.  This PR solves that.  You can add an issue or PR to support circular dependencies.

Shane32: > Also I see in their discussion those defaults : `= { }` . What does that mean ?

It means an object of the specified type with default values.  This is supported by GraphQL.NET.

src/GraphQL/Types/Schema.cs
Shane32: I don't know if it does or not, so I'll leave the code as-is.  Feel free to merge the suggestion if you think it would be better the other way.

src/GraphQL/Types/Schema.cs
Shane32: Yes, I would.

src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
Shane32: > Did you see [graphql/graphql-spec#793 (comment)](https://github.com/graphql/graphql-spec/pull/793#issuecomment-726811961) ?

I started to thumb through it.  It seems they are dealing with the spec, trying to decide if the spec should explicitly describe coercion rules.  But coercion must take place within GraphQL.NET whether it's in the spec or not, or GraphQL.NET will not run correctly.  So I don't think it really applies to us, unless they decide to change a coercion rule (for example, if they decide that circular coercion should not be allowed).

Shane32: If and when they merge a change to the spec, I'll review at that time.  Regardless, this PR is necessary for GraphQL.NET to operate correctly.

sungam3r: What about adding test for 
```graphql
input O {
  o: O = { o: null }
  a: Int
}
```
even if it fails for now. Mark as ignored with comment pointing this thread.

src/GraphQL/Types/Schema.cs
sungam3r: ok

src/GraphQL/Types/Schema.cs
```diff
@@ -411,8 +412,41 @@ protected virtual void Validate()
             //TODO: add different validations, also see SchemaBuilder.Validate
             //TODO: checks for parsed SDL may be expanded in the future, see https://github.com/graphql/graphql-spec/issues/653
             // Do not change the order of these validations.
+            CoerceInputTypeDefaultValues();
             SchemaValidationVisitor.Instance.Run(this);
             AppliedDirectivesValidationVisitor.Instance.Run(this);
         }
+
+        protected virtual void CoerceInputTypeDefaultValues()
```
sungam3r, line 420: It's `protected` so should have xml comments. Right?

sungam3r (approved):
Only xml comments left.

src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
Shane32: sure

src/GraphQL/Types/Schema.cs
Shane32: yes; done

src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
Shane32: done

src/GraphQL/Types/Schema.cs
```diff
@@ -411,8 +412,45 @@ protected virtual void Validate()
...
+            CoerceInputTypeDefaultValues();
             SchemaValidationVisitor.Instance.Run(this);
             AppliedDirectivesValidationVisitor.Instance.Run(this);
         }
+
+        /// <summary>
+        /// Coerces input types' default values when those values are GraphQLValue nodes.
+        /// This is applicable when the SchemaBuilder is used to build the schema.
+        /// </summary>
+        protected virtual void CoerceInputTypeDefaultValues()
```
Shane32, line 424: @sungam3r Review xml comment?

sungam3r: I would wrap GraphQLValue and SchemaBuilder  into `<see>`

sungam3r:
By the way, does approval test works well for you now?

Shane32:
> By the way, does approval test works well for you now?

so far

PR #2702, opened 2021-11-23
Nulls passed to DataLoader LoadAsync pass through as default results

YarekTyshchenko:
I have a usecase where chained dataloaders are used to check for a valid permission. A dataloader of type `IDataLoader<T, T?>` (input and output of same type), if permission is missing, the value is null.

I've tried several different ways to approach this and settled on returning null in the results, however, DataLoader's `LoadAsync` doesn't accept nulls.

### DIFF
--- a/src/GraphQL.DataLoader/DataLoaderExtensions.cs
+++ b/src/GraphQL.DataLoader/DataLoaderExtensions.cs
@@ -28,7 +28,9 @@ namespace GraphQL.DataLoader
 
             foreach (var key in keys)
             {
-                results.Add(dataLoader.LoadAsync(key));
+                results.Add(key != null
+                    ? dataLoader.LoadAsync(key)
+                    : new DataLoaderResult<T>(Task.FromResult<T>(default)));
             }
 
             return new DataLoaderResultWhenAll<T>(results);


Shane32:
The underlying cause is because `Dictionary<TKey, TValue>` cannot hold null entries as keys.  Dictionaries are used within `DataLoaderBase<TKey, T>` to prevent fetching the same key value more than once.  `SimpleDataLoader` is not affected because it does not derive from `DataLoaderBase`.  However, `BatchDataLoader` and `CollectionBatchDataLoader` both suffer from the same issue.

We should consider adding `where TKey : notnull` to the `DataLoaderBase<TKey, T>` class definition.

I wouldn't change the `DataLoaderExtensions.LoadAsync` method, used for fetching multiple keys at once, as this does not fix the underlying problem, and the same problem would exist on `DataLoaderBase.LoadAsync` for a single null key.  We could change `DataLoaderBase.LoadAsync` to return a null value for null keys (and then the change to the extension method would not be necessary).  Or we could modify `DataLoaderBase` be compatible with null keys.  I do not see any other changes that would be necessary.

Finally, we could leave the project as-is with no changes, or perhaps add another extension method that wraps a data loader with a null check.

Shane32:
Fixed in #2708

YarekTyshchenko:
Fantastic, thats much better than what I was doing. This makes dataloaders much more composable. Thanks

PR #2706, opened 2021-11-24
Update nullability attribute of typed Source fields

### LINKED ISSUE #2705: IResolveFieldContext.Source nullable?
## Summary

I recently updated the package from 4.5.0 to 4.6.1. Found lots of build errors due to the nullable change [#2599](https://github.com/graphql-dotnet/graphql-dotnet/pull/2599). Especially on this `IResolveFieldContext.Source` [code](https://github.com/graphql-dotnet/graphql-dotnet/blame/master/src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs#L104), why is this nullable? I can also see some usages in the same commit having `context.Source!`, which assuming it's not null. Is there any scenario/reason that the source in the resolve context would be null?

Shane32:
Fixes #2705

With this PR, `IResolveFieldContext<T>.Source` is not marked as nullable (unless `T` is a nullable type).  This is useful because in almost every use case, `Source` is not nullable -- only field resolvers within the root node (Query, Mutation, or Subscription) can have a null `Source` property.

`IResolveFieldContext.Source` is still typed as `object?`, and `ObjectGraphType` derives from `ObjectGraphType<object?>` so the `Source` property is still marked as nullable in these cases (since `ObjectGraphType` is most likely to be used for a root node).

Please note that this is a NON-breaking change, as nullability reference annotations do not change method signatures.  ~At least for reference types -- could this be a breaking change for generic types where the generic type is a value type?  I did not treat these changes as breaking changes previously.~

### DIFF
--- a/src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
+++ b/src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
@@ -61,7 +61,7 @@ namespace GraphQL.Utilities
                 throw new InvalidOperationException($"The field '{field.Name}' of an Object type '{type.Name}' has '{nameof(GraphQLTypeReference)}' type. This type must be replaced with a reference to the actual GraphQL type before using the reference.");
 
             // 2.3
-            if (!field.ResolvedType.IsOutputType())
+            if (!field.ResolvedType!.IsOutputType())
                 throw new InvalidOperationException($"The field '{field.Name}' of an Object type '{type.Name}' must be an output type.");
 
             ValidateFieldArgumentsUniqueness(field, type);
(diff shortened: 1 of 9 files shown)


src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
```diff
@@ -61,7 +61,7 @@ public override void VisitObjectFieldDefinition(FieldType field, IObjectGraphTyp
                 throw new InvalidOperationException($"The field '{field.Name}' of an Object type '{type.Name}' has '{nameof(GraphQLTypeReference)}' type. This type must be replaced with a reference to the actual GraphQL type before using the reference.");
 
             // 2.3
-            if (!field.ResolvedType.IsOutputType())
+            if (!field.ResolvedType!.IsOutputType())
```
sungam3r, line 64: Why changed this file? I have not seen `FieldType.ResolvedType` to change.

sungam3r:
In principle, I agree with the amendments, although I still believe that in some cases there may be problems.

src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
Shane32: Just cleaning up some build warnings probably

sungam3r: ok

sungam3r (approved):

PR #2707, opened 2021-11-24
Fix SchemaPrinter bug printing null lists

Shane32:
NRE before.

### DIFF
--- a/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
@@ -855,6 +855,36 @@ union SingleUnion = Foo
         }
 
         [Fact]
+        public void prints_input_type_with_default_null_value()
+        {
+            var root = new ObjectGraphType { Name = "Query" };
+            root.Field<NonNullGraphType<StringGraphType>>(
+                "str",
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<SomeInputType2>> { Name = "argOne", DefaultValue = new SomeInput2 { Names = null } })
+                );
+
+            var schema = new Schema { Query = root };
+
+            var expected = new Dictionary<string, string>
+            {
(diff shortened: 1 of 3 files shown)


src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
```diff
@@ -854,6 +854,36 @@ public void prints_input_type_with_default()
...
         }
 
+        [Fact]
+        public void prints_input_type_with_default_null_value()
+        {
+            var root = new ObjectGraphType { Name = "Query" };
+            root.Field<NonNullGraphType<StringGraphType>>(
+                "str",
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<SomeInputType2>> { Name = "argOne", DefaultValue = new SomeInput2 { Names = null } })
```
sungam3r, line 864: Probably, yes, but `names` is `[String]!` and I suspect that the problem needs to be determined earlier - at the stage of the schema initialization. Surely in the specification there is a necessary requirement. We need to add one more check (for default values) in the `SchemaValidationVisitor`. See `InputObjectGraphType.IsValidDefault` too.

sungam3r: I mean that I'm not against adding
```c#
            if (value == null)
                return "null";
```

but this tests shows IMO another problem.

Shane32: Fixed test schema so that `names` is `[String]`.  NRE still exists without the fix in this PR.  This PR solves the NRE.

sungam3r (approved):

src/GraphQL.Tests/Utilities/SchemaPrinterTests.cs
sungam3r: ok

PR #2708, opened 2021-11-24
Allow nulls in data loaders to pass through

Shane32:
Replaces:
- #2702

Not a breaking change because before an ArgumentNullException would have been thrown by the underlying dictionary.  So null values (reference or value types) were not supported previously.

### DIFF
--- a/src/GraphQL.DataLoader/DataLoaderBase.cs
+++ b/src/GraphQL.DataLoader/DataLoaderBase.cs
@@ -79,6 +79,8 @@ namespace GraphQL.DataLoader
 
         /// <summary>
         /// Asynchronously load data for the provided given key.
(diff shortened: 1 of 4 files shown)


src/GraphQL.DataLoader/DataLoaderBase.cs
```diff
@@ -79,13 +79,20 @@ public DataLoaderBase(bool caching, IEqualityComparer<TKey> equalityComparer, in
 
         /// <summary>
         /// Asynchronously load data for the provided given key.
+        /// If the key is <see langword="null"/> then a <see cref="DataLoaderResult{T}"/> containing
+        /// <see langword="null"/> will be immediately be returned.
```
sungam3r, line 83:
```suggestion
        /// <see langword="null"/> will be immediately returned.
```

sungam3r (approved):

PR #2717, opened 2021-12-04
CodeQL support

sungam3r:
This is a success! 😄 
изображение

After 75badfc:
изображение


Shane32:
Can it be updated so that it does not analyze the tests?

sungam3r:
I think yes, but why not to analyze tests? Let's analyze all C# code.

sungam3r:
I added config file to try include/exclude paths.

sungam3r:
From docs:
> For compiled languages, if you want to limit code scanning to specific directories in your project, you must specify appropriate build steps in the workflow. The commands you need to use to exclude a directory from the build will depend on your build system. 

I read the documentation and so I understand that in our case we do not even need to build test projects.

sungam3r:
@Shane32 Are you sure that the tests should be excluded?

sungam3r:
изображение

Shane32:
I find it quite annoying to have to write tests that conform to all the coding standards as the main project. There’s no reason for it. The test is there simply to ensure the main project works properly. Each test is, generally, independent, and new tests can be written without any regard to existing tests. 

I’m not sure what codeql does but that’s my general thoughts. So if we want codeql analysis as warnings in the tests that’s fine but I’m not sure we should have it required to pass build checks. Perhaps if I knew how codeql was useful then maybe I would have another opinion. 

I also do not think we should merge this PR until it passes all build checks, showing a green checkmark by the build. Perhaps one option is to configure all the codeql tests as warnings only, until the errors can be fixed.

sungam3r:
> I’m not sure we should have it required to pass build checks.

CodeQL has never been mandatory to merge PR.

Shane32:
> > I’m not sure we should have it required to pass build checks.
> 
> CodeQL has never been mandatory to merge PR.

Regardless, it shows a red x by the build now.

Shane32:
When we added xml comments, we selectively enabled the xml warnings once we had the xml comments in place. We didn’t configure the build to fail when an xml comment was missing until all xml comments were in place.

sungam3r:
What do you think about single error left ? https://github.com/graphql-dotnet/graphql-dotnet/security/code-scanning/645?query=ref%3Arefs%2Fpull%2F2717%2Fmerge+ref%3Arefs%2Fpull%2F2717%2Fhead+ref%3Arefs%2Fheads%2Fcodeql

sungam3r:
Base `GraphType` overrides `Equals(object)` so I think to mark error as false positive.

Shane32:
Right, or we can probably change the code easily to prevent the false positive.

Shane32:
> Base `GraphType` overrides `Equals(object)` so I think to mark error as false positive.

Makes sense

sungam3r:
изображение

🎉

sungam3r:
The last question - add or not into branch protection rules ?

Shane32:
Well I wouldn’t just now probably.

Shane32 (approved):

Issue #2549, opened 2021-05-19
Extensions Property is missing in ExecutionOptions


Popeye4242:
## Description

Extensions Property is missing in ExectionOptions.

Background: When working with WebSockets you cannot work with Headers as they are set initially when the WebSocket connection is made only. Therefore it is required that I can use Extensions for sending extra variables that don't belong in the query variables (for example OpenTracing trace id). 

### Steps to reproduce

1. Implement a Custom `IDocumentExecutor` and inject it into `IServiceCollection`
2. Set `context.extensions` variable in Apollo client
3. `ExecutionOptions` are missing in the `ExecuteAsync` Method of the `IDocumentExecutor`. 
![image](https://user-images.githubusercontent.com/35311793/118770537-bf9d9680-b881-11eb-80ad-79a1cecbeffe.png)

### Expected result

3. `Extensions` property should be available in `ExecutionOptions`. 

### Actual result

They are not. They are being completely removed and inaccessible. Even tho they definitely make it to the server in the payload. 
### Environment

```		
<PackageReference Include="GraphQL.Conventions" Version="4.1.0" />
<PackageReference Include="GraphQL.Server.All" Version="5.0.2" />
```

Shane32:
@sungam3r The feature request has merit; see `includeExtensions` property here: https://www.apollographql.com/docs/react/networking/advanced-http-networking/

The reasoning is sound also; when using websockets, extensions may be necessary to provide any other parameters to the server.

However, @Popeye4242, as it is implemented in this pull request, I do not see any way that user code can utilize the extensions property.  The extensions property would need to be added to `ExecutionContext`, and probably added to `IResolveFieldContext` as well.  We would need to be sure it is not confused with the output extensions property.  Documentation may need to be added as well.  Note that a change to `IResolveFieldContext` would be a breaking change and would need to be implemented in v5.  Breaking changes need notes added to the v5 migration document (which does not yet exist).

Now if the extensions are only used as 'headers', it does not belong within `ExecutionOptions` or `ExecutionContext` as `IDocumentExecutor` simply executes a given query.  But in accordance with the GraphQL spec, we support writing extensions within field resolvers (and middleware/etc), and I see no reason why a field resolver could not access input "extensions" as well.  The GraphQL spec does not provide guidance upon the format of the GraphQL request beyond the document itself, so there is no conflict.

[...]

PR #2719, opened 2021-12-05
Rough draft for InputExtensions property

Shane32:
See:
- #2549 

Needs:
- XML comments
- Tests updated

### DIFF
--- a/src/GraphQL/Execution/ExecutionContext.cs
+++ b/src/GraphQL/Execution/ExecutionContext.cs
@@ -59,6 +59,8 @@ namespace GraphQL.Execution
         public Dictionary<string, object?> Extensions { get; set; }
 #pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
 
+        public Inputs? InputExtensions { get; set; }
+
         /// <inheritdoc/>
         public IServiceProvider? RequestServices { get; set; }
 
--- a/src/GraphQL/Execution/ExecutionOptions.cs
+++ b/src/GraphQL/Execution/ExecutionOptions.cs
@@ -73,5 +73,7 @@ namespace GraphQL
         /// from your dependency injection framework.
         /// </summary>
         public IServiceProvider? RequestServices { get; set; }
+
+        public Inputs? InputExtensions { get; set; }
     }
 }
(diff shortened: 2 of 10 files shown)


src/GraphQL/Execution/ExecutionContext.cs
```diff
@@ -59,6 +59,8 @@ public class ExecutionContext : IExecutionContext, IExecutionArrayPool, IDisposa
         public Dictionary<string, object?> Extensions { get; set; }
 #pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
 
+        public Inputs? InputExtensions { get; set; }
```
sungam3r, line 62: OK. I suggest to rename `Extensions` to `OutputExtensions` as well.

src/GraphQL/Execution/ExecutionOptions.cs
```diff
@@ -73,5 +73,7 @@ public class ExecutionOptions : IProvideUserContext
         /// from your dependency injection framework.
         /// </summary>
         public IServiceProvider? RequestServices { get; set; }
+
+        public Inputs? InputExtensions { get; set; }
```
sungam3r, line 77: See  https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#request-parameters:

query -> Query [OK]
operationName -> OperationName [OK]
variables -> Inputs [rename to Variables]
extensions -> InputExtensions [rename to Extensions]

?

sungam3r: `Inputs` class should not be renamed, only properties

Shane32: i agree with those suggestions

sungam3r:
Server project is already ready for this PR:
https://github.com/graphql-dotnet/server/blob/master/src/Core/GraphQLRequest.cs#L51

src/GraphQL/Execution/ExecutionOptions.cs
Shane32: just for clairifcation, `ExecutionOptions` will have just `Extensions`, but `IResolveFieldContext` will have `InputExtensions` and `OutputExtensions`?  Seems maybe we should keep `InputExtensions` named the same throughout.

sungam3r: Yes, `IResolveFieldContext` will have `InputExtensions` and `OutputExtensions`. `ExecutionOptions` essentially is an "input configuration", only input.  I would keep naming of properties closed to spec - `Extensions` for `ExecutionOptions`. Context already implies "input" interpretation.

Shane32:
Replaced by #2766

PR #2724, opened 2021-12-06
add dateonly and timeonly support

SimonCropp:


src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            NullValue _ => null,
+            StringValue stringValue => ParseDate(stringValue.Value),
+            _ => ThrowLiteralConversionError(value)
+        };
+
+        /// <inheritdoc/>
+        public override object? ParseValue(object? value) => value switch
+        {
+            DateOnly d => d, // no boxing
```
Shane32, line 35:
```suggestion
            DateOnly _ => value, // no boxing
```
(to prevent boxing)

src/GraphQL/Types/Scalars/TimeOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            NullValue _ => null,
+            StringValue stringValue => ParseTime(stringValue.Value),
+            _ => ThrowLiteralConversionError(value)
+        };
+
+        /// <inheritdoc/>
+        public override object? ParseValue(object? value) => value switch
+        {
+            TimeOnly d => d, // no boxing
```
Shane32, line 35:
```suggestion
            TimeOnly _ => value, // no boxing
```
(to prevent boxing)

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+            string stringValue => ParseDate(stringValue),
+            null => null,
+            _ => throw new FormatException($"Could not parse date. Expected either a string or a DateOnly. Value: {value}")
```
Shane32, line 38:
```suggestion
            _ => ThrowValueConversionError(value)
```

src/GraphQL/Types/Scalars/TimeOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+            string stringValue => ParseTime(stringValue),
+            null => null,
+            _ => throw new FormatException($"Could not parse time. Expected either a string or a TimeOnly. Value: {value}")
```
Shane32, line 38:
```suggestion
            _ => ThrowValueConversionError(value)
```

src/GraphQL/Types/Scalars/TimeOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+        };
+
+        /// <inheritdoc/>
+        public override object? Serialize(object? value) => value switch
+        {
+            TimeOnly d => d.ToString("HH:mm:ss.fffffff", DateTimeFormatInfo.InvariantInfo),
```
Shane32, line 44:
```suggestion
            TimeOnly d => d.ToString("HH:mm:ss.FFFFFFF", DateTimeFormatInfo.InvariantInfo),
```
So that fractional parts of a second are not serialized if they are zero.

src/GraphQL/Types/Scalars/TimeOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+            null => null,
+            _ => ThrowSerializationError(value)
+        };
+
+        private static TimeOnly ParseTime(string stringValue)
+        {
+            if (TimeOnly.TryParseExact(stringValue, "HH:mm:ss.fffffff", DateTimeFormatInfo.InvariantInfo, DateTimeStyles.None, out var date))
```
Shane32, line 51:
```suggestion
            if (TimeOnly.TryParseExact(stringValue, "HH:mm:ss.FFFFFFF", DateTimeFormatInfo.InvariantInfo, DateTimeStyles.None, out var date))
```

src/GraphQL/Types/Scalars/TimeOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                return date;
+            }
+
+            throw new FormatException($"Could not parse time. Expected HH:mm:ss.fffffff. Value: {stringValue}");
```
Shane32, line 56:
```suggestion
            throw new FormatException($"Could not parse time. Expected HH:mm:ss.FFFFFFF. Value: {stringValue}");
```

Shane32 (changes requested):

src/GraphQL/Types/Scalars/TimeOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    public class TimeOnlyGraphType : ScalarGraphType
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TimeOnlyGraphType"/> class.
+        /// </summary>
+        public TimeOnlyGraphType()
+        {
+            Description = "The `Time` scalar type represents a time in accordance with the " +
+                "[ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) standard. Format is `HH:mm:ss.fffffff`.";
```
Shane32, line 21:
```suggestion
                "[ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) standard. Format is `HH:mm:ss.FFFFFFF`.";
```
I missed this one. For consistency let’s change it too. Otherwise looks good.

Shane32 (changes requested):

SimonCropp:
seems the LGTM step doesnt support net6?

Shane32:
> seems the LGTM step doesnt support net6?

No

Shane32 (approved):

Shane32:
Note that we can’t issue a new release until @joemcbride updates the nuget tokens

src/GraphQL/GraphQL.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;net6.0</TargetFrameworks>
```
sungam3r, line 4:
```suggestion
    <TargetFrameworks>netstandard2.1;netstandard2.0;net6</TargetFrameworks>
```

to align with other usages

src/GraphQL/Conversion/ValueConverter.cs
```diff
@@ -37,6 +37,10 @@ static ValueConverter()
             Register<string, float>(value => float.Parse(value, NumberStyles.Float, NumberFormatInfo.InvariantInfo));
             Register<string, double>(value => double.Parse(value, NumberStyles.Float, NumberFormatInfo.InvariantInfo));
             Register<string, decimal>(value => decimal.Parse(value, NumberStyles.Float, NumberFormatInfo.InvariantInfo));
+#if NET6_0_OR_GREATER
+            Register<string, DateOnly>(value => DateOnly.Parse(value, DateTimeFormatInfo.InvariantInfo));
```
sungam3r, line 41: No DateTimeStyles ?

SimonCropp: dateonly has no timezone, hence no universal time, so `DateTimeStyles.AdjustToUniversal | DateTimeStyles.AssumeUniversal` doesnt make sense. should i have included other DateTimeStyles?

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            NullValue _ => null,
+            StringValue stringValue => ParseDate(stringValue.Value),
+            _ => ThrowLiteralConversionError(value)
+        };
+
+        /// <inheritdoc/>
+        public override object? ParseValue(object? value) => value switch
+        {
+            DateOnly _ => value, // no boxing
```
sungam3r, line 35: What about `DateTime` support here? Also what about supporting `DateOnly` for `DateGraphType`? I think that `DateGraphType` and `DateOnlyGraphType` now are some sort of twins.

src/GraphQL/Types/Scalars/TimeOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+
+using System;
+using System.Globalization;
+using GraphQL.Language.AST;
+
+namespace GraphQL.Types
+{
+    /// <summary>
+    /// The Time scalar graph type represents a time in accordance with the ISO-8601 standard.
+    /// Format is `HH:mm:ss.fffffff`.
```
sungam3r, line 11:
```suggestion
    /// Format is `HH:mm:ss.FFFFFFF`.
```

src/GraphQL/Types/Scalars/TimeOnlyGraphType.cs
```diff
@@ -0,0 +1,61 @@
...
+            TimeOnly d => d.ToString("HH:mm:ss.FFFFFFF", DateTimeFormatInfo.InvariantInfo),
+            null => null,
+            _ => ThrowSerializationError(value)
+        };
+
+        private static TimeOnly ParseTime(string stringValue)
+        {
+            if (TimeOnly.TryParseExact(stringValue, "HH:mm:ss.FFFFFFF", DateTimeFormatInfo.InvariantInfo, DateTimeStyles.None, out var date))
+            {
+                return date;
```
sungam3r, line 53:
```suggestion
            if (TimeOnly.TryParseExact(stringValue, "HH:mm:ss.FFFFFFF", DateTimeFormatInfo.InvariantInfo, DateTimeStyles.None, out var time))
            {
                return time;
```

src/GraphQL.Tests/Types/TimeOnlyGraphTypeTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            CultureTestHelper.UseCultures(() => Should.Throw<FormatException>(() => _type.ParseValue("11 AM")));
+        }
+
+        [Fact]
+        public void coerces_iso8601_formatted_string_to_time()
+        {
+            CultureTestHelper.UseCultures(() =>
+            {
+                _type.ParseValue("01:02:03.0040000").ShouldBe(new TimeOnly(1, 2, 3, 4));
```
sungam3r, line 46: Please add more tests with "01:02:03.00400009" and "01:02:03.0040009" values.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -573,9 +573,9 @@ namespace GraphQL.Builders
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
         public virtual GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType>(string name, string? description, System.Action<GraphQL.Types.QueryArgument>? configure)
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
-        public virtual GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType, TArgumentType>(string name, string? description, [System.Diagnostics.CodeAnalysis.AllowNull] TArgumentType defaultValue = default)
+        public virtual GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType, TArgumentType>(string name, string? description, TArgumentType defaultValue = default)
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
-        public virtual GraphQL.Builders.ConnectionBuilder<TSourceType> Argument<TArgumentGraphType, TArgumentType>(string name, string? description, [System.Diagnostics.CodeAnalysis.AllowNull] TArgumentType defaultValue, System.Action<GraphQL.Types.QueryArgument>? configure)
```
sungam3r, line 578: .NET 6 TFM side effect ? I think that in case of ifdefs there should be multiple approval files - for each TFM.

src/GraphQL/Conversion/ValueConverter.cs
sungam3r: I don't know. I just see that `Parse` method accepts `DateTimeStyles` argument.

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
Shane32: I would not parse DateTime here. If someone receives DateTime that would be a coding error. We don’t want to allow people to write errors in new code.

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: I wouldn’t. It will just make it more painful to update the approval files. And then whenever there is an api change, there will be three approval files to review. The approvals are supposed to help us, not annoy us.

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
sungam3r: Why coding error?

sungam3r: This is only a matter of ease of use. It seems to me quite natural that this scalar type will support several .NET types. Semantics is not broken.

joemcbride:
> Note that we can’t issue a new release until @joemcbride updates the nuget tokens

I thought I had done this ... will double check.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: First of all, it should reflect reality. Now it shows public API for NET6 only. As soon as we have different APIs for different TFM, we actually **have to** have different approval files. 

> The approvals are supposed to help us, not annoy us.

Agree. That is why we need to make different approval files in order not to get confused in subsequent independent changes for netstandard/NET TFMs.

joemcbride:
> I thought I had done this ... will double check.

Yes I updated the token in May.  I went ahead and regenerated it again and updated the Org secret.  `NUGET_AUTH_TOKEN`

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
Shane32: No, the scalar type should match the variable type.  There is absolutely no reason for the DateOnly scalar type to support any other .NET type as a variable.  If someone can think of a legitimate reason, we can change it upon request.

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: No, we don't.  The existing approval files are sufficient.  There is no confusion.  Nobody looks at the approval files to know what APIs are supported.  It's just a way for us to review changes to the public API base to be sure we don't issue a breaking change.  It's not a published document that shows that for such-and-such a TFM there are specific public members.

Shane32:
> > I thought I had done this ... will double check.
> 
> Yes I updated the token in May. I went ahead and regenerated it again and updated the Org secret. `NUGET_AUTH_TOKEN`

Thank you!!

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: I did not say that this is a public document.

>  Nobody looks at the approval files to know what APIs are supported for a specific TFM.

**We** look to know about breaking changes. It's for us of course. Imagine one adds `TimeOnlyGraphType` in netstandard2.0 TFM in v4.8. Then for some reason this type will be removed in v4.9. It's a breaking change obviously but we will not even notice this change, it will be hidden from our eyes because now we see only public API changes for NET6.

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
sungam3r: > No, the scalar type should match the variable type. 

What means _match_ ?
изображение

sungam3r: `IntGraphType` accepts `ushort` variable thought the is "dedicated" `UShortGraphType`. The same is true for a plethora of other types.

Shane32: Right, because it is not known what System.Text.Json will deserialize a number into.  In this case the value stored as json is a string.  It is not going to be DateTime, DateOnly, or any other .NET type.

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: I do not think this "issue" is worth the considerable trouble to implement.

sungam3r: I will try.

sungam3r: #2730

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
Shane32: Regardless, does not mean that DateTime should be coerced to DateOnly.

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: Perhaps you misunderstood me.  I was not talking about the effort to implement multiple approval files.  I was talking about the effort keep them up to date and review them on an ongoing basis in the future.

src/GraphQL/Conversion/ValueConverter.cs
SimonCropp: agreed. also i cant find any docs about how DateTimeStyles works in the context of DateOnly

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
SimonCropp: i would prefer that DateGraphType only supported dateonly and DateOnlyGraphType would not need to exist. but i could not work out how to do that gracefully. if it was my project i would drop support for older fwks and only support net6

src/GraphQL.ApiTests/GraphQL.approved.txt
SimonCropp: if we were using Verify i would use a  `UniqueFor`  to produce one approved file per 
 target fwk https://github.com/VerifyTests/Verify/blob/main/docs/naming.md#uniquefor 

> Perhaps you misunderstood me. I was not talking about the effort to implement multiple approval files. I was talking about the effort keep them up to date and review them on an ongoing basis in the future.

verify supports bulk approval (through a system wide kb shortcut or a UI)

SimonCropp: but IMO this topic should not block this PR. i can be handled in a diff issue/pr

src/GraphQL/Types/Scalars/DateOnlyGraphType.cs
sungam3r: Well, it will be completely radically 😄 . So... I give up, let it remain as it is.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: I fixed approval files bloat and `AllowNull` attributes side effect as well. See #2730 again, please.

sungam3r: I suggest merge this PR after #2730

src/GraphQL.Tests/Types/TimeOnlyGraphTypeTests.cs
SimonCropp: good catch. only 3 points of milliseconds are supported. i will fix all the FFFFFs https://docs.microsoft.com/en-us/dotnet/api/system.timeonly.-ctor?view=net-6.0#System_TimeOnly__ctor_System_Int32_System_Int32_System_Int32_System_Int32_

src/GraphQL/Conversion/ValueConverter.cs
sungam3r: So maybe use `DateTimeStyles.None` explicitly to highlight this decision and go on?

src/GraphQL.Tests/Types/TimeOnlyGraphTypeTests.cs
Shane32: What do you mean?  This prints 00:00:00.0000001

```cs
Console.WriteLine(new TimeOnly(1).ToString("HH:mm:ss.FFFFFFF"));
```

TimeOnly stores times in ticks: "A time of day expressed in the number of 100-nanosecond units since 00:00:00.0000000."

sungam3r: WOW!

SimonCropp: sorry. i crossed the streams. ignore my previous comment

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: OK, this goes first :)

sungam3r:
@SimonCropp Waiting for FFFFFFF tests. Otherwize looks good.

src/GraphQL.Tests/Types/TimeOnlyGraphTypeTests.cs
SimonCropp: so the difficulty is the readable timeonly constructors do not support anything smaller than milliseconds. u need to use tick overload, resulting in 

```
_type.ParseValue("01:02:03.0040009").ShouldBe(new TimeOnly(37230040009));
```

ugly but it works

sungam3r (approved):

SimonCropp:
@sungam3r 

> Waiting for FFFFFFF tests. Otherwize looks good.

pushed. let me know if it looks good

src/GraphQL.Tests/Types/TimeOnlyGraphTypeTests.cs
sungam3r: ok

SimonCropp:
awesome. any chance of getting a beta nuget pushed so i can smoke test?

sungam3r:
We use GitHub package registry for preview packages - https://github.com/orgs/graphql-dotnet/packages?repo_name=graphql-dotnet

SimonCropp:
@sungam3r so there should be a 4.7 beta? 
image

sungam3r:
These versions are from `Directory.Build.props`. We had no `4.7.-preview` version here (but we SHOULD, we just forgot to bump version). Only `4.6.0-preview` for `master` and `5.0.0-preview` for `develop` now. Just pull the latest `4.6.0-preview` with your changes.

PR #2730, opened 2021-12-06
Add multi-TFM support for API approval tests

sungam3r:
If some lib **really** has different APIs for different tfms then it should be tracked by different approval files.


Shane32:
I do not think this PR should be merged.

sungam3r:
I understand your reasoning well. Nevertheless, this is the only way to guarantee correct public API changes tracking. Actually I can merge txt files from different tfms if they match.

sungam3r:
@Shane32 OK now ? :)

By the way, GraphQL.NET files for 20 and 21 differ because of `AllowNull` attribute.

Shane32:
Ok so explain what this PR does now?  There's no additional approval files, so what's it doing?

Shane32:
Let's merge #2724 first and then see how this PR works when there is a change in the api for a specific TFM.

sungam3r:
You'll see additional files only when there are actual changed between TFMs.

sungam3r:
I will review #2724 again.

sungam3r:
So now there are 3 files for GraphQL.NET : net6 file has new time/date types, 20 and 21 not. Also 20 and 21 are identical because of `AllowNull` suppression. @Shane32 do you want to merge 20 and 21 ?

Shane32:
Well, my opinion hasn't changed.  I don't think we should have 3 API approval files just because DateOnlyGraphType and TimeOnlyGraphType are only available in .NET 6.

sungam3r:
Now only 2 files :)

sungam3r:
Do you understand that in the case of different TFMs and ifdefs the public API may differ fundamentally? It is very easy not to notice, forget and face the bad consequences later.

Shane32:
> Do you understand that in the case of different TFMs and ifdefs the public API may differ fundamentally? It is very easy not to notice, forget and face the bad consequences later.

Sure, but GraphQL.NET is not that complex.  There is very little 'bad consequences' that could really occur.  We have two people, you and I, both looking over every PR before they get merged.  This entire approval file is intended to help us, not create a bunch of extra work for us.  We don't need multiple approval files.  It's already a pain.  We already are stuck with LGTM that doesn't work, isn't useful, and we can't get rid of it.  As it is now, every time I run tests, VS opens an additional copy of VS with the api file comparisons visible.  (No, not just the file, but an entire copy of VS.)  Not only that, but it does so for every mismatched file.  Adding this in is just an example of how this unnecessary change is going to be even more annoying than it already is.  For now I'll have to wait for my pc to open **2** additional copies of visual studio.  There's just no good reason for this PR at this time.  Maybe if we had complex differences between the public api, that would be different.  But we don't.  It's a very simple and very obvious difference.

I'm not going to change my opinion on this topic.

sungam3r:
I'm not feeling that I perform extra work with approvals. It's not a pain for me at all.

> As it is now, every time I run tests, VS opens an additional copy of VS with the api file comparisons visible. (No, not just the file, but an entire copy of VS.) Not only that, but it does so for every mismatched file. Adding this in is just an example of how this unnecessary change is going to be even more annoying than it already is. For now I'll have to wait for my pc to open 2 additional copies of visual studio.

OK. I also do not like it. Then let's just fix it. See 8f0cb49. I hope that there will be less pain now.

sungam3r:
Just to clarify my reasoning. Yes, we are just people, not machines. I don't want to bother with manual public API change tracking. It seems to me that now the proposed solution in general is better than before. Let me know whether it becomes easier after using the `NoDiff` setting.

sungam3r:
@Shane32 I'm going to merge this. If after all changes (merge TFMs, suppress VS diff) it becomes more difficult for you to work with approvals, then I agree to revert my changes.

Shane32:
I'll try it

Shane32:
@sungam3r This isn't working.  First, it failed the GraphQL.SystemTextJson api approval, when nothing was changed.  Second, it's in the wrong folder.  Third, if I compare the text contents of that api approval, it's drastically different, with a bunch of stuff missing.

image

The only thing it should have done is updated the approval file for net6 -- and it's not checking the net6 tfm at all!

Shane32:
(Trying to update api approvals for #2720 )

Shane32:
The biggest problem is that it's using 'GetDirectories' to determine what files to compare.  VS doesn't clean up old folders upon build, so it matches old dlls indiscriminately.  Even "Build > Clean Solution" doesn't clean up all old bin folders.

Shane32:
I have no idea why it wouldn't compare the net6 approval file.  That must be another bug.  Maybe it doesn't check all the files it finds, and instead only checks one file at a time until it hits a failure.

Shane32:
You really should make GraphQL.ApiTests run with different TFMs, so that VS will automatically copy the proper dependency file in with the api test and run it without having to do directory searching.

sungam3r:
Hmm.. I am sure that some of the problems are due to old bin/debug files. Just delete old folders and repeat the experiment.

Shane32:
I did

Shane32:
This is what I mean -- all I wanted to do was update the api approvals and it's taken me 20 minutes.  This is totally not worth it.

sungam3r:
But on my machine I do not get any unnecessary files. How can I reproduce?

sungam3r:
What about opening VS instance? No more ?

Shane32:
> What about opening VS instance? No more ?

no

sungam3r:
OK

sungam3r:
изображение
I'm 100% sure that these tfms are just old artefacts from older builds. After manual folder deletion you should not see them anymore.

Shane32:
yes, that is correct

sungam3r:
> The biggest problem is that it's using 'GetDirectories' to determine what files to compare. VS doesn't clean up old folders upon build, so it matches old dlls indiscriminately. 

Obviously. But as you said earlier - you and I **know** how to work with approvals and what it can do for us. So I think that **one-time** deletion of old directories is acceptable solution.

sungam3r:
If you have any other problems **after** deletion then I would like to solve them.

Shane32:
and every time i need to check something on version 2.4 i'll need to go manually delete the folders again.

Shane32:
i don't know about you, but i always have master3 as one of the branches on my pc

Shane32:
(havn't used 2.4 in a while though)

sungam3r:
Can you just use the second repo clone to look at v2.4 / v3 ? I did so for me to be able to keep state where I had a break from project support. Now I make all review, "fresh" PRs from _current_ repo and use _old_ repo when checking changes from April-September emails.

sungam3r:
Then you will not have build artifacts.

sungam3r:
On the other hand, I can to modify tfms lookup to use only needed tfms.

Shane32:
That solution doesn't work for me.  I typically am writing a test, and then switching between versions of graphql to run the test on multiple versions, so i can see if the problem was solved in a certain version of graphql.

sungam3r:
WOW!

sungam3r:
OK. I will change tfms lookup. It's not hard.

Shane32:
sometimes between minor versions, to identify where a problem was fixed or broke.  i do this all the time among different repos.

Shane32:
i'm willing to try this again.  i still don't think this pr was needed.  but i will try it some more and see if i run into more issues.

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/pull/2738

sungam3r:
> sometimes between minor versions, to identify where a problem was fixed or broke. i do this all the time among different repos.

OK. I do such things very-very rarely, I think.

PR #2737, opened 2021-12-07
Add GraphQLRelayName extension

kmcgill88:
We ran into an issue of conflicting type name when Including a class named `OrderOccasion` and an enum named `OrderOccasionType`. The `Type` extension `.GraphQLName()` was removing the trailing text `Type` causing the server to fail with `System.InvalidOperationException : Unable to serialize 'DELIVERY' to the scalar type 'OrderOccasionType'.` and registering `OrderOccasion` twice.

```
config.Types.Include<OrderOccasion>();
config.Types.Include<OrderOccasionType>();
```

The work around is to explicitly name the type with bypasses `Type.GraphQLName()` extension altogether.
```
config.Types.Include<OrderOccasionType>("OrderOccasionType");
```

We added an explicit `.GraphQLRelayName()` extension for specific Relay use case and updated code and tests appropriately.

Cheers! 🍻

### DIFF
--- a/src/GraphQL.Tests/TypeExtensionsTests.cs
+++ b/src/GraphQL.Tests/TypeExtensionsTests.cs
@@ -32,6 +32,19 @@ namespace GraphQL.Tests
             TypeExtensions.IsNullable(type).ShouldBeTrue();
         }
 
+        [Theory]
+        [InlineData(typeof(TestEnum))]
+        [InlineData(typeof(FooType))]
+        [InlineData(typeof(FooTypes))]
+        [InlineData(typeof(Foo))]
+        public void allow_non_relay_types_to_have_type_in_name(Type type)
+        {
+            Assert.Equal(type.Name, TypeExtensions.GraphQLName(type));
+        }
+
         private enum TestEnum { }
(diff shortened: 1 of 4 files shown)


sungam3r:
You modified `GraphQLName` extension and I'm 100% sure that it will break some schemas.

kmcgill88:
> You modified `GraphQLName` extension and I'm 100% sure that it will break some schemas.

Ok. What do you suggest?

sungam3r:
It seems to me that it is worth keeping things as it is. ping @Shane32

Shane32:
Agree.  In this case it is a GraphQL.NET convention.  If there was a compelling reason to do so, we could add a 'global switch' to modify behavior, perhaps as a delegate much like the name validation delegate.  But as it is now, all types are named based on this convention, so I would not change the default behavior.

sungam3r:
@kmcgill88 Options are
1. Rename C# class 
2. Set `Name` manually or "semi-manually" (not for each class but in some base class).

kmcgill88:
I don't have time to continue working on this. Thanks anyway.

PR #2739, opened 2021-12-08
Update dependency-injection.md

sungam3r:
Review of #2595

### DIFF
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -321,7 +321,7 @@ public class MyGraphType : ObjectGraphType<Category>
 
 Be aware that using the service locator in this fashion described in this section could be considered an
 Anti-Pattern. See [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/).
-However, the ease of use benefits for not very complex project often far outweigh the anti-pattern idealogy. 
+However, the ease of use benefits for not very complex projects often far outweigh the anti-pattern idealogy. 
 
 Within the `GraphQL.MicrosoftDI` package, there is also a builder approach to adding scoped dependencies.
 This makes for a concise and declarative approach. Each field clearly states the services it needs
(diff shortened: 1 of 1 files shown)


docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -321,10 +321,10 @@ public class MyGraphType : ObjectGraphType<Category>
 
 Be aware that using the service locator in this fashion described in this section could be considered an
 Anti-Pattern. See [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/).
-However, the performance benefits far outweigh the anti-pattern idealogy. 
+However, the ease of use benefits for not very complex project often far outweigh the anti-pattern idealogy.
```
sungam3r, line 324:
```suggestion
However, the ease of use benefits for not very complex projects often far outweigh the anti-pattern idealogy. 
```

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -321,10 +321,10 @@ public class MyGraphType : ObjectGraphType<Category>
 
 Be aware that using the service locator in this fashion described in this section could be considered an
 Anti-Pattern. See [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/).
-However, the performance benefits far outweigh the anti-pattern idealogy. 
+However, the ease of use benefits for not very complex projects often far outweigh the anti-pattern idealogy.
```
Shane32, line 324: This text is comparing a scoped schema to a singleton schema.  It is not easier to use `context.RequestServices`; rather, it is easier to use DI in its usual fashion, in the constructor of the class.  Performance, however, is key here.  With a scoped schema, there is considerable overhead to create the schema for each request, not to mention that any service used by any part of the entire schema needs to be instianted regardless of whether it is necessary to fulfill the request.  Please revert this comment.  Perhaps additional clarification could be added to clarify the context of this sentence.

```suggestion
However, the performance benefits far outweigh the anti-pattern idealogy, when compared to creating a scoped schema.
```

There is additional paragraphs below that describe other alternatives to this so-called "anti-pattern".

Shane32 (changes requested):

Shane32 (approved):

PR #2751, opened 2021-12-13
Rework of FederatedSchemaPrinter fix

sungam3r:
See #2576. I think that it should be done in different way. First - type for queries may have any name, not only "Query".  Second - empty type definitions are allowed by the spec. See https://github.com/graphql-dotnet/parser/pull/117 fix. I changed both printers and test.

### DIFF
--- a/src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
+++ b/src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
@@ -52,10 +52,6 @@ namespace GraphQL.Utilities.Federation
 
         public override string PrintObject(IObjectGraphType type)
         {
-            // Do not return an empty query type: "Query { }" as it is not valid as part of the sdl.
-            if (type != null && string.Equals(type.Name, "Query", StringComparison.Ordinal) && !type.Fields.Any(x => !IsFederatedType(x.ResolvedType!.GetNamedType().Name)))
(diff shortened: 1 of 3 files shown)


Shane32 (approved):
I don't use federation.  If this seems to be better suited to the spec, it's fine with me.

src/GraphQL/Utilities/Federation/FederatedSchemaPrinter.cs
```diff
@@ -68,7 +64,10 @@ public override string PrintObject(IObjectGraphType type)
 
             var extended = isExtension ? "extend " : "";
 
-            return FormatDescription(type.Description) + "{1}type {2}{3}{4} {{{0}{5}{0}}}".ToFormat(Environment.NewLine, extended, type.Name, implementedInterfaces, federatedDirectives, PrintFields(type));
+            if (type.Fields.Any(x => !IsFederatedType(x.ResolvedType!.GetNamedType().Name)))
```
sungam3r, line 67: `FederatedSchemaPrinter` overrides `PrintFields` so this check correlates to `PrintFields` implementation.

sungam3r:
@myermian Could you please check this fix? If OK I would like to merge PR as "real" SDL fix.

sungam3r:
So... I'll merge it. I hope that fix is correct.

PR #2755, opened 2021-12-14
async UnhandledExceptionDelegate

### LINKED ISSUE #2616: Allow async UnhandledExceptionDelegate method
See #2615 

Suggest to change:

https://github.com/graphql-dotnet/graphql-dotnet/blob/419a558cf76adb5636fb1cc6bc1aae99b31773f1/src/GraphQL/Execution/ExecutionOptions.cs#L66

to:

```cs
public Func<UnhandledExceptionContext, Task> UnhandledExceptionDelegate { get; set; } = context => Task.CompletedTask;
```

sungam3r:
Fixes #2616. "Async infection" included.

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -957,7 +957,7 @@ namespace GraphQL.Execution
         object? RootValue { get; }
         GraphQL.Types.ISchema Schema { get; }
         bool ThrowOnUnhandledException { get; }
-        System.Func<GraphQL.Execution.UnhandledExceptionContext, System.Threading.Tasks.Task> UnhandledExceptionDelegate { get; set; }
+        System.Func<GraphQL.Execution.UnhandledExceptionContext, System.Threading.Tasks.Task> UnhandledExceptionDelegate { get; }
         GraphQL.Language.AST.Variables Variables { get; }
     }
     public interface IExecutionStrategy
--- a/src/GraphQL.ApiTests/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.approved.txt
@@ -126,7 +126,7 @@ namespace GraphQL
         public object? Root { get; set; }
         public GraphQL.Types.ISchema? Schema { get; set; }
         public bool ThrowOnUnhandledException { get; set; }
-        public System.Action<GraphQL.Execution.UnhandledExceptionContext> UnhandledExceptionDelegate { get; set; }
+        public System.Func<GraphQL.Execution.UnhandledExceptionContext, System.Threading.Tasks.Task> UnhandledExceptionDelegate { get; set; }
         public System.Collections.Generic.IDictionary<string, object?> UserContext { get; set; }
         public System.Collections.Generic.IEnumerable<GraphQL.Validation.IValidationRule>? ValidationRules { get; set; }
     }
@@ -861,7 +861,7 @@ namespace GraphQL.Execution
         public object? RootValue { get; set; }
         public GraphQL.Types.ISchema Schema { get; set; }
         public bool ThrowOnUnhandledException { get; set; }
-        public System.Action<GraphQL.Execution.UnhandledExceptionContext> UnhandledExceptionDelegate { get; set; }
+        public System.Func<GraphQL.Execution.UnhandledExceptionContext, System.Threading.Tasks.Task> UnhandledExceptionDelegate { get; set; }
         public System.Collections.Generic.IDictionary<string, object?> UserContext { get; set; }
         public GraphQL.Language.AST.Variables Variables { get; set; }
         protected virtual void ClearContext() { }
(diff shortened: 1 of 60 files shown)


src/GraphQL/Execution/IExecutionContext.cs
```diff
@@ -72,7 +73,7 @@ public interface IExecutionContext : IProvideUserContext
         /// A delegate that can override, hide, modify, or log unhandled exceptions before they are stored
         /// within <see cref="Errors"/> as an <see cref="ExecutionError"/>.
         /// </summary>
-        Action<UnhandledExceptionContext> UnhandledExceptionDelegate { get; }
+        Func<UnhandledExceptionContext, Task> UnhandledExceptionDelegate { get; set; }
```
Shane32, line 76:
```suggestion
        Func<UnhandledExceptionContext, Task> UnhandledExceptionDelegate { get; }
```

Shane32 (changes requested):

src/GraphQL/Execution/IExecutionContext.cs
Shane32: Everything else in `IExecutionContext` is readonly.

Shane32:
:+1:

src/GraphQL/Execution/IExecutionContext.cs
sungam3r: Of course, just copy-paste.

sungam3r:
Are you not confused by "async infection" in loops and other places? I'm a bit.

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -907,8 +907,8 @@ namespace GraphQL.Execution
         public virtual System.Collections.Generic.Dictionary<string, GraphQL.Language.AST.Field>? GetSubFields(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
         [System.Obsolete]
         protected virtual System.Threading.Tasks.Task OnBeforeExecutionStepAwaitedAsync(GraphQL.Execution.ExecutionContext context) { }
-        protected virtual bool ProcessNodeUnhandledException(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node, System.Exception ex) { }
-        protected virtual void SetArrayItemNodes(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ArrayExecutionNode parent) { }
+        protected virtual System.Threading.Tasks.Task<bool> ProcessNodeUnhandledException(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node, System.Exception ex) { }
+        protected virtual System.Threading.Tasks.Task SetArrayItemNodes(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ArrayExecutionNode parent) { }
```
Shane32, line 911: Rename to SetArrayItemNodesAsync

src/GraphQL.ApiTests/GraphQL.approved.txt
```diff
@@ -897,7 +897,7 @@ namespace GraphQL.Execution
         protected virtual GraphQL.Execution.RootExecutionNode BuildExecutionRootNode(GraphQL.Execution.ExecutionContext context, GraphQL.Types.IObjectGraphType rootType) { }
         protected virtual System.Collections.Generic.Dictionary<string, GraphQL.Language.AST.Field> CollectFieldsFrom(GraphQL.Execution.ExecutionContext context, GraphQL.Types.IGraphType specificType, GraphQL.Language.AST.SelectionSet selectionSet, System.Collections.Generic.Dictionary<string, GraphQL.Language.AST.Field>? fields) { }
         protected virtual System.Threading.Tasks.Task CompleteDataLoaderNodeAsync(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
-        protected virtual void CompleteNode(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
+        protected virtual System.Threading.Tasks.Task CompleteNode(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
```
Shane32, line 900: Rename to CompleteNodeAsync

Shane32 (changes requested):
We should probably rename the async methods

Shane32:
I'm fine leaving the name of `UnhandledExceptionDelegate` as-is, but if you want to rename it to something that indicates it's async, that would probably be better.

src/GraphQL.ApiTests/GraphQL.approved.txt
Shane32: And rename to ProcessNodeUnhandledExceptionAsync

sungam3r:
> that indicates it's async

It has no "async kind". It' just a delegate that returns `Task`. It may be async but may be not. I'll keep existing name for delegate.

Shane32:
To clarify, I was not thinking `UnhandledExceptionDelegateAsync`, but rather `UnhandledExceptionAsyncDelegate`.  The delegate isn't async itself, but it's a delegate of an async method.

sungam3r:
Or sync :) maybe change return type to `ValueTask` ?

Shane32:
> Or sync :)

Well, I would disagree, and say that if a method (or method delegate) returns `Task` or `Task<T>`, it is considered an asynchronous method (or asynchronous method delegate), regardless of what code comprises the method, be it synchronous or not.  That is why extension methods that pass through tasks from other methods still are named with `...Async` even though they contain neither `async` nor `await` keywords.  It has to do with the point of view; methods are named based on how they are used, regardless of how they are implemented.  And so a delegate for a method returning `Task` is therefore called an asynchronous method delegate.  Whether a user choses to implement the asynchronous method synchronously or not is up to them.

> maybe change return type to `ValueTask` ?

The GraphQL execution pipeline with parallel execution strategy uses `WhenAll`, necessitating a conversion of any `ValueTask`s to `Task`s.  Also, there is almost no benefit of a `ValueTask` over a `Task` when it does not return a value.  (Faulted `ValueTask`s wrap a `Task` containing an `Exception` anyway.)  It's only a very rare and specific set of circumstances.  Finally, as this is an exception handler, optimization to this degree is unnecessary.

We could change `ProcessNodeUnhandledExceptionAsync` to return a `ValueTask<bool>`.  This would have no detrimental effect, I believe.  Again, it would only be used when an exception occurred; it's not on the hot path.  So long as the delegate executes synchronously, it is possible that an allocation could be avoided.

The only other thing we could do is consider using `ValueTask<object>`s throughout the field resolver pipeline, rather than `object` and comparing `object` to `Task`.  However, in order to not break the semantics for `ValueTask` (can only await once, etc), it would be similar code, checking if the `ValueTask` completed synchronously and awaiting it immediately, or if not, converting it to a `Task`.  With the extra conversions necessary, it would probably increase memory usage (although perhaps not the number of allocations), and the overall benefit to the project would be questionable.  So as much as I would like to do this, I am guessing that benchmarks would show that it has a negative effect on performance for synchronous field resolvers.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: Done.

src/GraphQL.ApiTests/GraphQL.approved.txt
sungam3r: done

Shane32 (approved):

PR #2757, opened 2021-12-18
[WIP] Eliminate System.Reactive dependency

Shane32:
Notes:
- Appears that with about 230 lines of code, the System.Reactive dependency can be entirely eliminated
- The subscription support code could then be moved into the main project
- If we move forward, I'd probably make the change for v5 and not v4 (although it certainly would work with v4)
- No effect on other projects (e.g. GraphQL.Server), since only .NET Standard interfaces are used
- I should write a bunch of additional tests to prove that subscription events are raised in the correct order and so on


sungam3r:
Is it copied from System.Reactive.Core/Linq?

Shane32:
No I wrote it from scratch. It’s pretty simple code really. Only took me a couple hours. I have looked at copying the System.Reactive code but it was way too complex for what is needed here. 

FYI, one of my goals is to solve the issue with disconnected websockets connections from leaving partially open tcp connections. Writing this code from scratch allows me to understand exactly when each event can or cannot fire.

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
```diff
@@ -103,9 +103,10 @@ public override async Task<ExecutionResult> ExecuteAsync(ExecutionContext contex
                 }
 
                 return subscription
-                    .Select(value => BuildSubscriptionExecutionNode(node.Parent!, node.GraphType!, node.Field, node.FieldDefinition, node.IndexInParentNode, value!))
-                    .SelectMany(async executionNode =>
+                    .SelectAsync(async (value, token) =>
                     {
+                        //note: token is unused here
```
Shane32, line 108: Need to think about this: is it correct for the execution to use the ExecutionContext's cancellation token?  When would that trigger?  Seems like in this case it should cancel when the observable sequence is disconnected, which would be `token` here in my code.  Might need to wrap/duplicate the execution context.  Need to know more about where the execution context's cancellation token came from in order to answer this question.

Shane32:
Of course the tests fail in GitHub Actions since they are time-dependent.  I will have to rewrite the tests somehow.

Shane32:
> Of course the tests fail in GitHub Actions since they are time-dependent. I will have to rewrite the tests somehow.

Done.  The tests might look like a nightmare, but they do test the aspects that I designed it to.

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
sungam3r: `GraphQLHttpMiddleware.GetCancellationToken`

sungam3r:
I would prefer to do other things and postpone this PR.

Shane32:
I’m currently working on rewriting the entire websocket handling chain. Trying to solve the memory leak issue. So this is still WIP until I solve the entire issue.

Shane32:
@sungam3r2022 I'm going to rebase this PR onto `develop`, write xml comments, write tests, write documentation, and finish it up.  Ok?  The code seems quite straightforward such that we can include subscription support in the main project without the SystemReactive dependency.  I really want to spend time improving subscription support here in the next month before the v5 release - if I have the time of course - as I would like to start using subscriptions for server-to-server notifications in my projects.

If you have any concerns as to this code, please let me know.

sungam3r2022:
OK, rebase and I'' review.

Shane32:
Replaced by #3004

PR #2758, opened 2021-12-19
Makes IDocumentCache asynchronous.

gdlcf88:
Hi, I redesigned the `IDocumentCache` with async getter and setter methods since I want to create a document cache implementation myself with `IDistributedCache` which has async APIs.

### DIFF
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -70,8 +70,8 @@ namespace GraphQL.Caching
         }
 
         /// <inheritdoc/>
-        public virtual Task<Document?> GetAsync(string query) =>
-            Task.FromResult(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
+        public virtual ValueTask<Document?> GetAsync(string query) =>
+            new(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
 
         /// <inheritdoc/>
         public virtual Task SetAsync(string query, Document? value) =>
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -74,12 +74,14 @@ namespace GraphQL.Caching
             new(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
 
         /// <inheritdoc/>
-        public virtual ValueTask SetAsync(string query, Document? value)
+        public virtual ValueTask SetAsync(string query, Document value)
         {
-            if (value == null)
-                _memoryCache.Remove(query);
-            else
-                _memoryCache.Set(query, value, GetMemoryCacheEntryOptions(query));
+            if (value is null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            _memoryCache.Set(query, value, GetMemoryCacheEntryOptions(query));
 
             return default;
         }
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using GraphQL.Language.AST;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Options;
@@ -62,17 +63,19 @@ namespace GraphQL.Caching
         }
 
         /// <inheritdoc/>
-        public virtual Document? this[string query]
-        {
-            get => _memoryCache.TryGetValue<Document>(query, out var value) ? value : null;
-            set => _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value, GetMemoryCacheEntryOptions(query));
-        }
-
-        /// <inheritdoc/>
         public virtual void Dispose()
         {
             if (_memoryCacheIsOwned)
                 _memoryCache.Dispose();
         }
+
+        /// <inheritdoc/>
+        public virtual Task<Document?> GetAsync(string query) =>
+            Task.FromResult(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
+
+        /// <inheritdoc/>
+        public virtual Task SetAsync(string query, Document? value) =>
+            Task.FromResult(_memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value,
+                GetMemoryCacheEntryOptions(query)));
     }
 }
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -74,12 +74,12 @@ namespace GraphQL.Caching
             new(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
 
         /// <inheritdoc/>
-        public virtual Task SetAsync(string query, Document? value)
+        public virtual ValueTask SetAsync(string query, Document? value)
         {
             _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value,
                 GetMemoryCacheEntryOptions(query));
 
-            return Task.CompletedTask;
+            return new ValueTask(Task.CompletedTask);
         }
     }
 }
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -74,17 +74,12 @@ namespace GraphQL.Caching
             new(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
 
         /// <inheritdoc/>
-        public virtual ValueTask SetAsync(string query, Document value)
+        public virtual ValueTask SetAsync(string query, Document? value)
         {
-            _memoryCache.Set(query, value, GetMemoryCacheEntryOptions(query));
-
-            return default;
-        }
-
-        /// <inheritdoc/>
-        public virtual ValueTask RemoveAsync(string query)
-        {
-            _memoryCache.Remove(query);
+            if (value == null)
+                _memoryCache.Remove(query);
+            else
+                _memoryCache.Set(query, value, GetMemoryCacheEntryOptions(query));
 
             return default;
         }
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -76,10 +76,9 @@ namespace GraphQL.Caching
         /// <inheritdoc/>
         public virtual ValueTask SetAsync(string query, Document value)
         {
-            _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value,
-                GetMemoryCacheEntryOptions(query));
+            _memoryCache.Set(query, value, GetMemoryCacheEntryOptions(query));
 
-            return new ValueTask(Task.CompletedTask);
+            return default;
         }
 
         /// <inheritdoc/>
@@ -87,7 +86,7 @@ namespace GraphQL.Caching
         {
             _memoryCache.Remove(query);
 
-            return new ValueTask(Task.CompletedTask);
+            return default;
         }
     }
 }
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -74,10 +74,18 @@ namespace GraphQL.Caching
             new(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
 
         /// <inheritdoc/>
-        public virtual ValueTask SetAsync(string query, Document? value)
+        public virtual ValueTask SetAsync(string query, Document value)
         {
             _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value,
                 GetMemoryCacheEntryOptions(query));
+
+            return new ValueTask(Task.CompletedTask);
+        }
+
+        /// <inheritdoc/>
+        public virtual ValueTask RemoveAsync(string query)
+        {
+            _memoryCache.Remove(query);
 
             return new ValueTask(Task.CompletedTask);
         }
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -74,8 +74,12 @@ namespace GraphQL.Caching
             new(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
 
         /// <inheritdoc/>
-        public virtual Task SetAsync(string query, Document? value) =>
-            Task.FromResult(_memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value,
-                GetMemoryCacheEntryOptions(query)));
+        public virtual Task SetAsync(string query, Document? value)
+        {
+            _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value,
+                GetMemoryCacheEntryOptions(query));
+
+            return Task.CompletedTask;
+        }
     }
 }
--- a/src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
+++ b/src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
@@ -61,6 +61,8 @@ namespace GraphQL.Tests.Utilities
                 stack = stack.Remove(0, 5);
             if (stack.StartsWith("   at "))
                 stack = stack.Remove(0, 6);
+            if (stack.StartsWith("   在 "))
+                stack = stack.Remove(0, 5);
 
             stack.ShouldStartWith("GraphQL.Tests.Utilities.SchemaBuilderExecutionTests.Query.Method()");
         }
@@ -91,6 +93,8 @@ namespace GraphQL.Tests.Utilities
                 stack = stack.Remove(0, 5);
             if (stack.StartsWith("   at "))
                 stack = stack.Remove(0, 6);
+            if (stack.StartsWith("   在 "))
+                stack = stack.Remove(0, 5);
 
             stack.ShouldStartWith("GraphQL.Tests.Utilities.SchemaBuilderExecutionTests.Query.get_Property()");
         }
--- a/src/GraphQL/Caching/DefaultDocumentCache.cs
+++ b/src/GraphQL/Caching/DefaultDocumentCache.cs
@@ -18,6 +18,6 @@ namespace GraphQL.Caching
 
         public ValueTask<Document?> GetAsync(string query) => default;
 
-        public ValueTask SetAsync(string query, Document? value) => default;
+        public ValueTask SetAsync(string query, Document value) => default;
     }
 }
--- a/src/GraphQL/Caching/DefaultDocumentCache.cs
+++ b/src/GraphQL/Caching/DefaultDocumentCache.cs
@@ -16,10 +16,10 @@ namespace GraphQL.Caching
         /// </summary>
         public static readonly DefaultDocumentCache Instance = new DefaultDocumentCache();
 
-        public ValueTask<Document?> GetAsync(string query) => new((Document?)null);
+        public ValueTask<Document?> GetAsync(string query) => default;
 
-        public ValueTask SetAsync(string query, Document value) => new(Task.CompletedTask);
+        public ValueTask SetAsync(string query, Document value) => default;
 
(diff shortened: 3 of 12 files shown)


Shane32:
Wouldn’t data in a distributed cache need to be serialized and deserialized upon storage / retrieval?

Shane32:
It seems to me that a Document is simply a deserialized query. So a document cache only has a performance benefit if it is in memory where it skips the deserialization process. 

It does also allow for skipping some validation rules, which can be beneficial. But the overhead of the network request for the distributed cache would probably still outweigh the time it would take to validate the document. 

Have you done any performance comparisons?

We can still consider implementing the document cache interface as asynchronous methods for v5 but it would probably be a highly unique scenario to realize a performance benefit.

gdlcf88:
Hi @Shane32, I think you are right, parsing a query input may be cheaper than getting a distributed cache item and deserializing it (even in a cluster).

> We can still consider implementing the document cache interface as asynchronous methods for v5

Thanks, async methods are worth to be adopting. 😃

Shane32:
Can you update the pr to target the develop branch?

src/GraphQL/Caching/IDocumentCache.cs
```diff
@@ -8,8 +9,13 @@ namespace GraphQL.Caching
     public interface IDocumentCache
     {
         /// <summary>
-        /// Gets or sets a document in the cache. Must be thread-safe. Returns <see langword="null"/> if no entry is found.
+        /// Gets a document in the cache. Must be thread-safe. Returns <see langword="null"/> if no entry is found.
         /// </summary>
-        Document? this[string query] { get; set; }
+        Task<Document?> GetAsync(string query);
```
Shane32, line 14: Maybe ValueTask here???

gdlcf88:
It seems difficult to change the target branch. Should I recreate the PR?

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -61,18 +62,20 @@ protected virtual MemoryCacheEntryOptions GetMemoryCacheEntryOptions(string quer
...
                 _memoryCache.Dispose();
         }
+
+        /// <inheritdoc/>
+        public virtual ValueTask<Document?> GetAsync(string query) =>
+            new(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
+
+        /// <inheritdoc/>
+        public virtual Task SetAsync(string query, Document? value) =>
+            Task.FromResult(_memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value,
```
Shane32, line 78: Should return Task.CompletedTask so there is no heap allocation

Shane32:
> It seems difficult to change the target branch. Should I recreate the PR?

No but I probably need to merge master into develop for the changes to show up correctly.

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
```diff
@@ -61,6 +61,8 @@ public async Task schema_first_generate_exception_with_normal_stack_trace_for_me
                 stack = stack.Remove(0, 5);
             if (stack.StartsWith("   at "))
                 stack = stack.Remove(0, 6);
+            if (stack.StartsWith("   在 "))
+                stack = stack.Remove(0, 5);
```
Shane32, line 65: Did these come from a merge conflict?

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
```diff
@@ -91,6 +93,8 @@ public async Task schema_first_generate_exception_with_normal_stack_trace_for_pr
...
```
Shane32, line 97: And here

Shane32:
I have it targeting develop properly now.

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
gdlcf88: No, I add this code to ensure the test's success in the zh-hans environment.

Shane32: Oh ok

src/GraphQL/Caching/IDocumentCache.cs
```diff
@@ -8,8 +9,13 @@ namespace GraphQL.Caching
...
-        /// Gets or sets a document in the cache. Must be thread-safe. Returns <see langword="null"/> if no entry is found.
+        /// Gets a document in the cache. Must be thread-safe. Returns <see langword="null"/> if no entry is found.
         /// </summary>
-        Document? this[string query] { get; set; }
+        ValueTask<Document?> GetAsync(string query);
+
+        /// <summary>
+        /// Sets a document in the cache. Must be thread-safe.
+        /// </summary>
+        Task SetAsync(string query, Document? value);
```
sungam3r, line 19: `Document?` or `Document` ?

sungam3r: Also change `Task` to `ValueTask` to align with one from `GetAsync`. No performance reasons here, just unifying.

src/GraphQL/Caching/DefaultDocumentCache.cs
```diff
@@ -15,6 +16,8 @@ public sealed class DefaultDocumentCache : IDocumentCache
         /// </summary>
         public static readonly DefaultDocumentCache Instance = new DefaultDocumentCache();
 
-        Document? IDocumentCache.this[string query] { get => null; set { } }
+        public ValueTask<Document?> GetAsync(string query) => new((Document?)null);
```
sungam3r, line 19: @Shane32 Why we used explicit interface implementation here?

src/GraphQL/Caching/IDocumentCache.cs
sungam3r: Also this PR should change documentation/migration guide for v5 accordingly.

sungam3r: Can be a separate PR but it's better to handle documentation changes here as well.

gdlcf88: > `Document?` or `Document` ?

https://github.com/graphql-dotnet/graphql-dotnet/pull/2758/commits/a3301366655be6be93c1b14c5aef25e37ba2b289

gdlcf88: > Can be a separate PR but it's better to handle documentation changes here as well.

Need help on docs changes because I'm not familiar with graphql-dotnet. 🍰

src/GraphQL/Caching/DefaultDocumentCache.cs
Shane32: Probably a mistake

Shane32:
If we are using untyped ValueTask rather than Task, then the proper return value is `default` or `new ValueTask()`. 

See ms documentation:

> An instance created with the parameterless constructor or by the `default(ValueTask)` syntax (a zero-initialized structure) represents a synchronously, successfully completed operation.

src/GraphQL/Caching/IDocumentCache.cs
Shane32: If you can just add a note under the breaking changes section of this page:
- https://github.com/graphql-dotnet/graphql-dotnet/blob/develop/docs2/site/docs/migrations/migration5.md

I’d suggest saying that the indexer property of IDocumentCache has been replaced with GetAsync, SetAsync and RemoveAsync methods or similar.

Shane32: You can make the change within the same PR. It should be pretty easy to do.

Shane32: (Usually we will make another pass at the documentation before release so it doesn’t have to be perfect, but it helps remind us what changes were made and what needs to be documented. )

src/GraphQL/Caching/IDocumentCache.cs
```diff
@@ -8,8 +9,18 @@ namespace GraphQL.Caching
...
+
+        /// <summary>
+        /// Sets a document in the cache. Must be thread-safe.
+        /// </summary>
+        ValueTask SetAsync(string query, Document value);
+
+        /// <summary>
+        /// Remove a document cache. Must be thread-safe.
+        /// </summary>
+        ValueTask RemoveAsync(string query);
```
sungam3r, line 24: Why do we need this method here? This is an interface (contract) between GraphQL.NET and some external code. GraphQL.NET needs only to get document from cache or set document into cache. That's all.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -62,17 +63,31 @@ protected virtual MemoryCacheEntryOptions GetMemoryCacheEntryOptions(string quer
...
+
+        /// <inheritdoc/>
+        public virtual ValueTask SetAsync(string query, Document value)
         {
-            if (_memoryCacheIsOwned)
-                _memoryCache.Dispose();
+            _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value,
+                GetMemoryCacheEntryOptions(query));
+
+            return new ValueTask(Task.CompletedTask);
```
sungam3r, line 82:
```suggestion
            _memoryCache.Set(query, value, GetMemoryCacheEntryOptions(query));
            return default;
```

src/GraphQL/Caching/IDocumentCache.cs
gdlcf88: The value arg of the SetAsync method is non-nullable now. There is no longer a valid way to remove a cache item (or set its value to null, at least). I'm not sure if the RemoveAsync is needed, but I reserved it.

I can revert it if you're sure.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
gdlcf88: https://github.com/gdlcf88/graphql-dotnet/commit/0e8a12c3f5dd9158806069c89389205561aa5eba

src/GraphQL/Caching/IDocumentCache.cs
gdlcf88: > You can make the change within the same PR. It should be pretty easy to do.

https://github.com/gdlcf88/graphql-dotnet/commit/32e2d31a7c2c905d92fff88ef3734011ef6999b8

src/GraphQL/Caching/IDocumentCache.cs
Shane32: I think what @sungam3r is saying is that the interface is for GraphQL.NET to use, and GraphQL.NET does not currently need to clear data from the cache.  However, it would make sense for implementations, including MemoryCache, to have a Remove / RemoveAsync method.

@sungam3r Are you sure we should not include a method in the interface for clearing entries?  What about for a scenario where someone wanted to write custom middleware that removes cache entries when, for example, an authentication token has expired?

I can't think of a specific reason myself, but I might include the method anyway, perhaps with a note that its implementation is not required for GraphQL.NET.

sungam3r: If one needs `Remove` then he inherits from `MemoryDocumentCache` or introduce another interface containing `Remove`.  It makes no sense to add this method in `IDocumentCache` interface.

sungam3r: Even `SetAsync` **may** be implemented in a way that it actually removes supplied key from cache if `value` is `null`.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -62,17 +63,30 @@ protected virtual MemoryCacheEntryOptions GetMemoryCacheEntryOptions(string quer
...
+                _memoryCache.Dispose();
         }
 
         /// <inheritdoc/>
-        public virtual void Dispose()
+        public virtual ValueTask<Document?> GetAsync(string query) =>
+            new(_memoryCache.TryGetValue<Document>(query, out var value) ? value : null);
+
+        /// <inheritdoc/>
+        public virtual ValueTask SetAsync(string query, Document value)
```
sungam3r, line 77:
```csharp
        public virtual ValueTask SetAsync(string query, Document value)
        {
            if (value == null)
                _memoryCache.Remove(query);
            else
                _memoryCache.Set(query, value, GetMemoryCacheEntryOptions(query));

            return default;
        }
```

and remove `public virtual ValueTask RemoveAsync(string query)`

src/GraphQL/Caching/IDocumentCache.cs
gdlcf88: Should I roll it back from
```csharp
Task SetAsync(string query, Document value);
```
to
```csharp
Task SetAsync(string query, Document? value);
```
?

sungam3r: I think not:
```
ValueTask SetAsync(string query, Document value);
```

gdlcf88: > Even `SetAsync` **may** be implemented in a way that it actually removes supplied key from cache if `value` is `null`.

But the value is considered never to be null since the project is `<Nullable>enable</Nullable>`, at least the IDEs do.

sungam3r: Yes. I said that `SetAsync` **may** be implemented such way but not really **should**. GraphQL.NET needs `SetAsync` to set not null document. Nullable annotations help to understand that. However, it is technically nothing preventing pass `null!`.

Shane32: If the argument is passed as non-nullable, it should be expected that it may throw an ArgumentNullException if it is null. We should not disguise the method intent behind incorrect NRT annotations.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
Shane32: Then we should revert back to `Document?`

Shane32: Either the argument is nullable or it isn’t. Having it defined as nullable does not hurt any other code.

gdlcf88: > Either the argument is nullable or it isn’t. Having it defined as nullable does not hurt any other code.

Actually, the setter with a null value is expected to set a real cache item as `{key} => null` (not to remove an existing item). If the RemoveAsync method is considered no-need, we just do not provide any cache removing way and keep the setter as a single responsibility method.

gdlcf88:
```csharp
        public virtual Document? this[string query]
        {
            get => _memoryCache.TryGetValue<Document>(query, out var value) ? value : null;
            set => _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value, GetMemoryCacheEntryOptions(query));
        }
```

> If the argument is passed as non-nullable, it should be expected that it may throw an ArgumentNullException if it is null. We should not disguise the method intent behind incorrect NRT annotations.

@Shane32 By the way, the previous design allows setting a null cache value.

If we do this change, projects that need the caches removing feature may be hurt.

src/GraphQL/Caching/IDocumentCache.cs
sungam3r: I'm fine either way.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
sungam3r: > the setter with a null value is expected to set a real cache item as {key} => null (not to remove an existing item).

Why "expected". It's just an implementation detail. The client of this method gets identical behavior regardless of whether the cache keeps `null` values or not.

sungam3r: Moreover, the storage of null values implies keeping **keys**, which affects the overall memory consumption.

Shane32: > the previous design allows setting a null cache value.

Actually it doesn't.  See the null-coalescing operator where it throws an exception if a null value is passed in.

```cs
_memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query))
```

> Why "expected"? It's just an implementation detail. The client of this method gets identical behavior regardless of whether the cache keeps null values or not.

I agree.

Shane32: Perhaps for the simplest implementation, let's not support removing key values at all.  If a user requests a change, we can perform the following as a non-breaking change:

1. Change the NRT annotation on `IDocumentCache.SetAsync`'s `document` argument to indicate it is nullable.
2. Rewrite the `MemoryDocumentCache.SetAsync` code to clear an entry if `null` is passed in.

(Note: adding another member, such as `RemoveAsync`, to an interface is a breaking change.)

Shane32: This suggestion will also match up with the existing implementation, which it throws an `ArgumentNullException` if trying to set a value to the cache which is `null`.

Shane32: Right, I see.

Shane32: > Perhaps for the simplest implementation, let's not support removing key values at all. If a user requests a change, we can perform the following as a non-breaking change:
> 
> 1. Change the NRT annotation on `IDocumentCache.SetAsync`'s `document` argument to indicate it is nullable.
> 2. Rewrite the `MemoryDocumentCache.SetAsync` code to clear an entry if `null` is passed in.
> 
> (Note: adding another member, such as `RemoveAsync`, to an interface is a breaking change.)

Ok then let's go with this @gdlcf88 if we are all agreed.

gdlcf88: @Shane32 @sungam3r

Done.

https://github.com/graphql-dotnet/graphql-dotnet/pull/2758/commits/384e3c0d2ad67a70e8fb08395e422f6490bde966

Shane32: Well, what we wrote is to ** not ** support removing keys at all -- and change it later if someone requests it.  (lol)

gdlcf88: https://github.com/graphql-dotnet/graphql-dotnet/pull/2758/commits/ce8b43aa1f5da5a03ed53726a965891e33178ada

src/GraphQL/Caching/DefaultDocumentCache.cs
```diff
@@ -15,6 +16,8 @@ public sealed class DefaultDocumentCache : IDocumentCache
         /// </summary>
         public static readonly DefaultDocumentCache Instance = new DefaultDocumentCache();
 
-        Document? IDocumentCache.this[string query] { get => null; set { } }
+        public ValueTask<Document?> GetAsync(string query) => default;
+
+        public ValueTask SetAsync(string query, Document value) => default;
```
sungam3r, line 21: Please add inheritdoc xml comments for these new methods.

src/GraphQL/Caching/IDocumentCache.cs
```diff
@@ -8,8 +9,17 @@ namespace GraphQL.Caching
     public interface IDocumentCache
     {
         /// <summary>
-        /// Gets or sets a document in the cache. Must be thread-safe. Returns <see langword="null"/> if no entry is found.
+        /// Gets a document in the cache. Must be thread-safe.
         /// </summary>
-        Document? this[string query] { get; set; }
+        /// <param name="query">As the cache key</param>
```
sungam3r, line 14:
```suggestion
        /// <param name="query">As the cache key.</param>
```

src/GraphQL/Caching/IDocumentCache.cs
```diff
@@ -8,8 +9,17 @@ namespace GraphQL.Caching
...
+        /// <returns>The cached document object. Returns <see langword="null"/> if no entry is found.</returns>
+        ValueTask<Document?> GetAsync(string query);
+
+        /// <summary>
+        /// Sets a document in the cache. Must be thread-safe.
+        /// </summary>
+        /// <param name="query">As the cache key</param>
```
sungam3r, line 21:
```suggestion
        /// <param name="query">As the cache key.</param>
```

sungam3r (approved):

sungam3r:
@gdlcf88 Approval tests failed. Just remove ~.approved file and rename ~.received into ~.approved.

gdlcf88:
> @gdlcf88 Approval tests failed. Just remove ~.approved file and rename ~.received into ~.approved.

I believe that it will pass the tests now. My first time using PublicApiGenerator, sorry for the mistake. 🤣

Shane32 (approved):

Shane32:
Thanks for the effort @gdlcf88 !

PR #2760, opened 2021-12-20
[WIP] Integrate deserialization into base GraphQL serialization libraries

Shane32:
Goal: Implement deserialization functionality into existing GraphQL.SystemTextJson and GraphQL.NewtonsoftJson serialization libraries so that the server project's ASP.Net implementation code is not dependent on the serialization library in use.

- Adds generic `GraphQLRequest` type to GraphQL library (copied from GraphQL.Server.Core)
- Adds `IGraphQLRequestReader` interface to GraphQL library, with one `ReadAsync` method
- Adds implementation within GraphQL.SystemTextJson
- Todo: add GraphQL.NewtonsoftJson implementation (note: tests will fail until newtonsoft implementation is added)


Shane32:
@sungam3r Thoughts?

Another idea: combine `IDocumentWriter` and `IGraphQLRequestReader` into a single interface such as `ISerializer`.  I thought it might be better to have two separate interfaces though.

sungam3r:
What is the goal?

sungam3r:
I am extremely skeptical about such changes. The transport agnostic functionality is mixed with transport related.

Shane32:
We have a separate nuget package for serialization already. It is completely independent of GraphQL operation. (GraphQL.SystemTextJson). There is no reason why it should be only a serialization package and not a deserialization package as well. 

The goal is to combine the serialization and deserialization libraries that already exist into one package. For instance combine GraphQL.SystemTextJson (a serialization library) with GraphQL.Server.Transport.SystemTextJson (a deserialization library). 

It would make it a lot easier for those of us that are using GraphQL without the server project

The only code added to the main project is (1) a deserialization interface containing one method, similar to our existing serialization interface with one method, and (2) the GraphQLRequest class definition. Nothing else.

Shane32:
With these changes, GraphQL.SystemTextJson is still independent of the transport channel. There is no ASP.Net specific code or anything. Yes it is specific to JSON, as it was before. Yes, it has a special handler for the typical JSON request format (which is not defined by the GraphQL spec). But it can deserialize any type just like DocumentWriter can, and can be customized also. It’s basically the same as DocumentWriter, except a reader.

sungam3r:
> There is no reason why it should be only a serialization package and not a deserialization package as well.

The reason is the zones of responsibility.

> For instance combine GraphQL.SystemTextJson (a serialization library) with GraphQL.Server.Transport.SystemTextJson (a deserialization library).

I feared it... Always was against such merging. We once had already affected this topic with @joemcbride . `GraphQLRequest` is a transport class from [HTTP spec](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#request). GraphQL.NET is transport agnostic and **should** stay so. For example, my company actually has private GraphQL transport implementation working with Kafka. We use our own data format in that transport. This format is no better and no worse than one from HTTP spec. So why throw transport details of a separate implementation in the general-purpose code?

> The only code added to the main project is (1) a deserialization interface containing one method, similar to our existing serialization interface with one method, and (2) the GraphQLRequest class definition. Nothing else.

I understand and I see what you do. You mix the concepts in favor of "simplicity". This is not that "simplicity", which I would like to see. This is the penetration of a specific HTTP implementation in the general-purpose code. That is why I consider individual repositories as good. They support the boundaries of the responsibility zones.

> GraphQL.SystemTextJson is still independent of the transport channel. There is no ASP.Net specific code or anything.

I can not agree. Dependence is and it is brought in an implicit form. Not good.

Shane32:
> GraphQL.NET is transport agnostic and should stay so. 

Please explain how GraphQL.SystemTextJson is transport agnostic where GraphQL.Server.Transport.SystemTextJson is not.  They both serialize/deserialize to JSON.  The official GraphQL spec does not have any specifications requiring or suggesting the use of JSON.

I grant you the GraphQLRequest class is a transport class definition.  (It doesn't even contain any code!)  Are you against this entire PR just because of a single class definition?

I don't see any good reason to have the serialization and deserialization libraries separate.  They are both "transport" libraries. 
 They both are serializing or deserializing to a UTF-8 encoded JSON stream.  Serialization so happens to not have a wrapper class (unless used with websockets, where it does); deserialization does.  But they are both transport libraries.  Please explain.

> The reason is the zones of responsibility.

GraphQL.SystemTextJson should have the responsibility of serialization AND deserialization, eliminating separate SystemTextJson and NewtonsoftJson libraries within the server project.  That's what I'm trying to accomplish.  Let's discuss how.

sungam3r:
> I grant you the GraphQLRequest class is a transport class definition. (It doesn't even contain any code!)

This class actually doesn't contain any code too:
```c#
public class TransportAgnosticClass // really?
{
   public string[] HttpHeaders { get; set; } 
}
```
As you understand, it is not only in the code, but in **concepts**.

> Are you against this entire PR just because of a single class definition?

If I do not delve into details, then you can say so. Well, I can return to the review later after finishing with other PRs from email. I admit that there may be some sense in all this, but now it is not clear for me. I see the leaky abstractions now.

Shane32:
We could always create a nuget package called GraphQL.Transports which contains:
- `IDocumentWriter`
- `GraphQLRequest`
- `IGraphQLRequestReader`
- Accompanying `IGraphQLBuilder` extension methods

Seems a bit pitiful to create an entire nuget package for a few lines of code.  I wouldn't do it, but it would be better than having two different SystemTextJson nuget packages - one for serialization and one for deserialization.

src/GraphQL.Harness/GraphQLMiddleware.cs
```diff
@@ -51,7 +53,7 @@ private async Task ExecuteAsync(HttpContext context, ISchema schema)
         {
             var start = DateTime.UtcNow;
 
-            var request = await context.Request.Body.FromJsonAsync<GraphQLRequest>(context.RequestAborted);
+            var request = await _reader.ReadAsync<GraphQLRequest>(context.Request.Body, context.RequestAborted);
```
Shane32, line 56: Here you can see in the basic sample provided where the middleware code was tied directly to the `GraphQL.SystemTextJson` extension method using an unconfigurable JSON serializer.

Now the code uses the registered reader, which can be configured elsewhere in the system.

(Note: the old code works also.)

src/GraphQL.Harness/GraphQLMiddleware.cs
```diff
@@ -3,7 +3,6 @@
 using System.Threading.Tasks;
 using GraphQL;
 using GraphQL.Instrumentation;
-using GraphQL.SystemTextJson;
```
Shane32, line 6: And the dependency within the middleware upon GraphQL.SystemTextJson has been eliminated.

src/GraphQL.Harness/GraphQLMiddleware.cs
sungam3r: > where the middleware code was tied directly to the GraphQL.SystemTextJson extension method using an unconfigurable JSON deserializer.

It's just a sample project.

src/GraphQL.NewtonsoftJson/ExecutionResultContractResolver.cs
```diff
@@ -21,6 +21,9 @@ protected override JsonConverter ResolveContractConverter(Type objectType)
             if (typeof(ExecutionResult).IsAssignableFrom(objectType))
                 return new ExecutionResultJsonConverter(_errorInfoProvider);
 
+            if (typeof(GraphQLRequest) == objectType)
```
sungam3r, line 24: Let's say that `ExecutionResult` is _GraphQL core spec related_ thing . Then `GraphQLRequest` is not _GraphQL core spec related_ thing. It's a _HTTP transport related_ thing.

sungam3r:
I understand what you want to do. The goal is worthy. But the goal should not be achieved at the cost of mixing abstraction levels. It should be another way.

Shane32:
> I understand what you want to do. The goal is worthy. But the goal should not be achieved at the cost of mixing abstraction levels. It should be another way.

I already suggested adding another nuget package - GraphQL.Transports - to contain the few class and interface definitions that deal with transports.  If you think this is necessary, let's discuss it.

Do you have any other suggestions?

I think it is ABSOLUETLY ESSENTIAL that in v5 the JSON serialization libraries (already specific to a JSON transport encoding) handle both deserialization and serialization of the typical JSON-based transports in use via an abstraction layer.  This includes the typical HTTP request format for single queries, the typical HTTP request format for multiple queries, the typical HTTP response format, and the graphql-ws transport definitions currently in use.  The serialization packages are JSON-transport specific already; they will continue to be transport specific, but support more transports than simply the HTTP response transport.  These should specifically be designed to be abstracted away from any particular HTTP server implementation, so there are no dependencies on ASP.NET Core if you are using it as a client library, for example.  It should be able to be extended as desired and handle generic types.  Already the typical HTTP response format is covered, which is a JSON encoding of the map defined by the spec.  Already the packages can deserialize generic formats, but not in a configurable and an abstract way.  They cannot handle the HTTP request formats in an abstract manner.

If you have no suggestions, let's proceed with this PR.  Or even if you do, let's merge this PR and then discuss additional improvements, such as we have been doing with `IGraphQLBuilder`.

sungam3r:
If all code from here is covered by #2831 then I suggest to close it in favor of #2831.

Shane32:
Closed in favor of #2831

PR #2768, opened 2021-12-21
Add forgotten ConfigureAwait(false)

sungam3r:


Shane32:
That’s all great but we really should apply all these changes to master.

sungam3r:
I intentionally made changes to `develop`. The main development goes there, and not on the `master`. I do not want to perform double work.

Shane32:
I intend to support v4 until v5 is released, and perhaps after for critical bugs.  This is a v4 bug, and can be a critical bug.  It should be in v4.  There would have been no "double work" if this was merged to master.  As it is, now there is, since I have to recreate this PR for v4.

sungam3r:
2 changed files after #2770

Shane32 (approved):

PR #2773, opened 2021-12-22
EnumCaseAttribute

sungam3r:


Shane32:
I like the PR in general, but shouldn't we handle this by expanding the `INameConverter` in v5?  Then it would be easy to apply to the entire schema rather than having to mark each enum.

sungam3r:
It does not interfere. You can specify case for each enum type before this change. I made the same but without having to write additional types.

docs2/site/docs/getting-started/schema-types.md
```diff
@@ -120,14 +120,44 @@ The name of each member _is_ the value.
...
 
-You can use `EnumerationGraphType<TEnum>` to automatically generate values by providing a .NET
-`enum` for `TEnum`. The `Name` will default to the .NET Type name, which you can override in
-the constructor. The `Description` will default to any `System.ComponentModel.DescriptionAttribute`
-applied to the enum type. The `DeprecationReason` will default to any `System.ObsoleteAttribute`
-applied to the enum type. By default, the name of each enum member will be converted to CONSTANT_CASE.
-Override `ChangeEnumCase` to change this behavior. Apply a `DescriptionAttribute` to an enum member
-to set the GraphQL `Description`. Apply an `ObsoleteAttribute` to an enum member to set the GraphQL
-`DeprecationReason`.
+I. You can use `EnumerationGraphType<TEnum>` to automatically generate values by providing a .NET
```
Shane32: I would not use "I" and "II", especially as the primary/outer numbering.  I would use 1/2/3 first and a/b/c second.  Note that when I and II and etc are used, it is always done lowercase.  "i." and "ii." and "iii.".

Shane32: Default Microsoft Word numbering pattern:
image

sungam3r:
One may not necessarily want to change the case to all enums in schema. For example, due to backward compatibility. The part will be in one case, part in the other.

Shane32:
> One may not necessarily want to change the case to all enums in schema. For example, due to backward compatibility. The part will be in one case, part in the other.

Perhaps.

docs2/site/docs/getting-started/schema-types.md
sungam3r: reformatted

Shane32:
I think the question is: do you think we should expand `INameConverter` for v5?  If so, I think we should consider these changes with those now.  For example, if we want a 'raw' name converter (which does not convert names at all), how would that work if the EnumerationGraphType converts them to upper case by default?  Or perhaps it does not convert by default, and lets the INameConverter convert it to uppercase.  Then how would the name converter know to let the attribute override the default conversion?

> It does not interfere. 

It may not, if we are careful about implementation now.  It will interfere if we are not.

I think we should expand the INameConverter for v5, and we should consider those changes with these.  At least anticipated changes, even if we do not move forward.

Shane32:
FYI, everything in the PR looks good.  I am just concerned with how it will work with an INameConverter in v5 assuming we add functionality in the name converter to sanitize enum value names (which I think we should do).  I don't want to make it difficult for ourselves, being that we are at the end of the v4 lifespan with v5 "around the corner".  We should have a plan.

sungam3r:
I looked at `INameConverter` and its usage. It's like a global tool to adjust all names for schema parts when building it. I see no reason to not have both converter (as is now or expanded for v5) and this PR. Well, graph type (any type, enum in our case) **may** exist without schema. Schema just adds (or not in case of 'raw' name converter) some additional requirements on naming. These requirements **may** override those in the type (enum). And may not.

sungam3r:
Note that techically we _may_ add similar `ChangeTypeName` (or whatever) feature even for `GraphType`. Now it's implicit inside `GraphType.GetDefaultName`. So the behavior will be aligned among all graph types.

sungam3r:
Of course, I do not call for that way. I just show analogies.

Shane32:
> I looked at `INameConverter` and its usage. It's like a global tool to adjust all names for schema parts when building it. I see no reason to not have both converter (as is now or expanded for v5) and this PR. Well, graph type (any type, enum in our case) **may** exist without schema. Schema just adds (or not in case of 'raw' name converter) some additional requirements on naming. These requirements **may** override those in the type (enum). And may not.

Good.  Then let's figure out how they will work together before merging this.  I have already explained issues that I foresee.

Shane32:
#2774

sungam3r:
>  Then let's figure out how they will work together

It seemed to me that I already answered. First enum/graphtype set their `Name` property. Then `INameConverter` *may* rewrite it. It is already happening:
изображение
изображение

In this regard `EnumValueDefinition.Name` has the same logic - it's just a get/set property that **may** be rewrited.

Shane32:
I need a little more time to review this PR.

Shane32 (approved):

PR #2779, opened 2021-12-23
Prevent context re-use for enumerables

### LINKED ISSUE #2777: Source is null when resolve is used with an IEnumerable
I am currently trying to migrate our codebase to 4.6.1 and felt about the following problem.

`Field<ListGraphType<MyObjectType>>("test", resolve: x => List.Where(h => h.Id == x.Source.Id));`

This worked for the old versions but now it throws a NullReferenceException cause Source is null. This is cause `Where` returns a `IEnumerable` and so the `Func` in `Where` is executed when the `GetEnumerator` is called and not when the resolve is done.

My guess is that the IResolveFieldContext instance is reused and cleared right after the resolve is called but before the IEnumerable is done.

A simple fix would be to add an ToArray after the Where. But that is error prone and its also inefficient cause its wasted allocated memory.

It would be very helpful and less error prone if graphql-dotnet could help here and in example could keep the field context until the IEnumerable is enumerated. Or simply do a ToArray of the returned object if resolve is IEnumerable and the GraphType is a list (while that would increase memory usage).

I would guess that for a lot of users of graphql-dotnet its not obvious why source is null in that case.

Shane32:
Fixes #2777

### DIFF
--- a/src/GraphQL.Tests/Execution/FuncFieldResolverTests.cs
+++ b/src/GraphQL.Tests/Execution/FuncFieldResolverTests.cs
@@ -498,5 +498,103 @@ namespace GraphQL.Tests.Execution
             rfc2.ShouldNotBeSameAs(_context);
             rfc1.ShouldNotBe(rfc2);
         }
+        //====
+
+        [Fact]
+        public void Does_Not_Share_Enumerable_Typed()
+        {
+            IResolveFieldContext<int?> rfc1 = null;
+            IResolveFieldContext<int?> rfc2 = null;
+            var ffr1 = new FuncFieldResolver<int?, IEnumerable<int>>(context =>
+            {
+                rfc1 = context;
+                return new[] { 1, 2 };
+            });
+            var ffr2 = new FuncFieldResolver<int?, IEnumerable<int>>(context =>
+            {
+                rfc2 = context;
+                return new[] { 1, 2 };
+            });
+            ffr1.Resolve(_context);
+            ffr2.Resolve(_context);
+            rfc1.ShouldNotBeNull();
+            rfc1.ShouldNotBeSameAs(_context);
+            rfc2.ShouldNotBeNull();
+            rfc2.ShouldNotBeSameAs(_context);
+            rfc1.ShouldNotBe(rfc2);
+        }
+
+        [Fact]
+        public void Does_Not_Share_Enumerable_Untyped()
+        {
+            IResolveFieldContext<int?> rfc1 = null;
+            IResolveFieldContext<int?> rfc2 = null;
+            var ffr1 = new FuncFieldResolver<int?, object>(context =>
+            {
+                rfc1 = context;
+                return new[] { 1, 2 };
+            });
+            var ffr2 = new FuncFieldResolver<int?, object>(context =>
+            {
+                rfc2 = context;
+                return new[] { 1, 2 };
+            });
+            ffr1.Resolve(_context);
+            ffr2.Resolve(_context);
+            rfc1.ShouldNotBeNull();
+            rfc1.ShouldNotBeSameAs(_context);
+            rfc2.ShouldNotBeNull();
+            rfc2.ShouldNotBeSameAs(_context);
+            rfc1.ShouldNotBe(rfc2);
+        }
+
+        [Fact]
+        public void Does_Not_Share_Enumerable_Tasks_Typed()
+        {
+            IResolveFieldContext<int?> rfc1 = null;
+            IResolveFieldContext<int?> rfc2 = null;
+            var ffr1 = new FuncFieldResolver<int?, Task<IEnumerable<int>>>(context =>
+            {
+                rfc1 = context;
+                return Task.FromResult<IEnumerable<int>>(new[] { 1, 2 });
+            });
+            var ffr2 = new FuncFieldResolver<int?, Task<IEnumerable<int>>>(context =>
+            {
+                rfc2 = context;
+                return Task.FromResult<IEnumerable<int>>(new[] { 1, 2 });
+            });
+            ffr1.Resolve(_context);
+            ffr2.Resolve(_context);
+            rfc1.ShouldNotBeNull();
+            rfc1.ShouldNotBeSameAs(_context);
+            rfc2.ShouldNotBeNull();
+            rfc2.ShouldNotBeSameAs(_context);
+            rfc1.ShouldNotBe(rfc2);
+        }
+
+        [Fact]
+        public void Does_Not_Share_Enumerable_Tasks_Untyped()
+        {
+            IResolveFieldContext<int?> rfc1 = null;
+            IResolveFieldContext<int?> rfc2 = null;
+            var ffr1 = new FuncFieldResolver<int?, object>(context =>
+            {
+                rfc1 = context;
+                return Task.FromResult<IEnumerable<int>>(new[] { 1, 2 });
+            });
+            var ffr2 = new FuncFieldResolver<int?, object>(context =>
+            {
+                rfc2 = context;
+                return Task.FromResult<IEnumerable<int>>(new[] { 1, 2 });
+            });
+            ffr1.Resolve(_context);
+            ffr2.Resolve(_context);
+            rfc1.ShouldNotBeNull();
+            rfc1.ShouldNotBeSameAs(_context);
+            rfc2.ShouldNotBeNull();
+            rfc2.ShouldNotBeSameAs(_context);
+            rfc1.ShouldNotBe(rfc2);
+        }
+
     }
 }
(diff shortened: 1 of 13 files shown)


src/GraphQL/Resolvers/FuncFieldResolver.cs
```diff
@@ -104,12 +105,6 @@ public FuncFieldResolver(Func<IResolveFieldContext<TSourceType>, TReturnType?> r
                     var adapter = System.Threading.Interlocked.Exchange(ref _sharedAdapter, null);
                     adapter = adapter == null ? new ResolveFieldContextAdapter<TSourceType>(context) : adapter.Set(context);
                     var ret = resolver(adapter);
```
Shane32, line 107: This code really should never execute.  For instance, if the resolver returned `Task` (a void result), which makes no sense.  It is possible that someone derived from `Task<TResult>` and returned that here.  So for this code flow, since we can't anticipate the use case, re-using the adapter isn't a good idea.  We just don't know what value was really returned.  I really doubt this code will ever execute, however.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -408,9 +408,13 @@ protected virtual async Task ExecuteNodeAsync(ExecutionContext context, Executio
                 if (result is not IDataLoaderResult)
                 {
                     CompleteNode(context, node);
-                    // for non-dataloader nodes that completed without throwing an error, we can re-use the context
-                    resolveContext.Reset(null, null);
-                    System.Threading.Interlocked.CompareExchange(ref context.ReusableReadonlyResolveFieldContext, resolveContext, null);
+                    // for non-dataloader nodes that completed without throwing an error, we can re-use the context,
+                    // except for enumerable lists - in case the user returned a value with LINQ that is based on the context source
```
sungam3r, line 412: > that is based on the context source

maybe or maybe not

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -408,9 +408,13 @@ protected virtual async Task ExecuteNodeAsync(ExecutionContext context, Executio
...
+                    if (result is not IEnumerable)
```
sungam3r, line 413: One may have such resolver:
```
context => new MyObject { Values = context.Source.Where(x => x.Age > 42); }
```

sungam3r: The question of the context usage is much wider than just checking the return type.

sungam3r: Or `context => new MyObject(context.Source.Where(x => x.Age > 42));`

sungam3r: I am not sure that it makes sense to consider a separate case and adjust the code under it, especially since not 100% of cases of change will be useful. I can return `IEnumerable` without the use of context:
`context => Enumerable.Range(5);`

sungam3r: We have `Copy` method with clear semantic. It will work for any situation that I demonstated above.

sungam3r: I doubt it makes sense to take these changes. They are not full. The problem is not solved, questions remain.

Shane32: The only way to 'fully' fix context re-use is to not re-use the context.

This change covers an EXTREMELY COMMON use case: returning a LINQ IEnumerable from a field resolver for a ListGraphType.  It is also an explicit change, specifically designed to support list graph types.  And, it does fix ALL list graph types (since they are ALL required to return an IEnumerable).  It is not intended to fix issues with object graph types, nor does it do so.

We should not have code that breaks for normal use cases, even if such behavior is documented.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: This is merely a comment explaining the purpose of why this code was added.  Of course they may not have returned a LINQ enumerable at all.

src/GraphQL/Resolvers/FuncFieldResolver.cs
```diff
@@ -78,16 +79,16 @@ public FuncFieldResolver(Func<IResolveFieldContext<TSourceType>, TReturnType?> r
                         adapter.Set(context);
                     }
                     var ret = resolver(adapter);
-                    // only re-use contexts that completed synchronously and do not return an IDataLoaderResult
+                    // only re-use contexts that completed synchronously and do not return an IDataLoaderResult or an IEnumerable
```
sungam3r, line 82:
```suggestion
                    // only re-use contexts that completed synchronously and do not return an IDataLoaderResult or an IEnumerable (that may be based on the context source)
```

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: I understand. Well, OK, let this common case be considered explicitly.

src/GraphQL/Resolvers/FuncFieldResolver.cs
sungam3r: You know better.

sungam3r (approved):

src/GraphQL/Resolvers/FuncFieldResolver.cs
Shane32: I totally misread and botched this code.  Good thing we have tests for such sheer stupidity.

Shane32:
I also need to add tests

sungam3r:
Ready for review?

Shane32:
Yes, ready for review.

src/GraphQL.Tests/Execution/FuncFieldResolverTests.cs
```diff
@@ -498,5 +498,103 @@ public void Does_Not_Share_Dataloader_Tasks_Untyped_Derived()
             rfc2.ShouldNotBeSameAs(_context);
             rfc1.ShouldNotBe(rfc2);
         }
+        //====
```
Shane32, line 501:
```suggestion
```

src/GraphQL.Tests/Execution/FuncFieldResolverTests.cs
```diff
@@ -498,5 +498,102 @@ public void Does_Not_Share_Dataloader_Tasks_Untyped_Derived()
...
+            });
+            ffr1.Resolve(_context);
+            ffr2.Resolve(_context);
+            rfc1.ShouldNotBeNull();
+            rfc1.ShouldNotBeSameAs(_context);
+            rfc2.ShouldNotBeNull();
+            rfc2.ShouldNotBeSameAs(_context);
+            rfc1.ShouldNotBe(rfc2);
+        }
+
```
Shane32, line 597:
```suggestion
```

src/GraphQL/Resolvers/FuncFieldResolver.cs
```diff
@@ -45,24 +46,32 @@ public FuncFieldResolver(Func<IResolveFieldContext<TSourceType>, TReturnType?> r
...
+            // when source type is object, just pass the context through
             if (typeof(TSourceType) == typeof(object))
             {
                 // ReadonlyResolveFieldContext implements IResolveFieldContext<object>
                 return (context) => resolver(context.As<TSourceType>());
             }
 
-            if (typeof(IDataLoaderResult).IsAssignableFrom(typeof(TReturnType)))
+            // for return types of IDataLoaderResult or IEnumerable
+            if (typeof(IDataLoaderResult).IsAssignableFrom(typeof(TReturnType)) || (typeof(IEnumerable).IsAssignableFrom(typeof(TReturnType)) && typeof(TReturnType) != typeof(string)))
```
sungam3r, line 57: Worth to move into local function and reuse.

src/GraphQL/Resolvers/FuncFieldResolver.cs
```diff
@@ -78,16 +87,20 @@ public FuncFieldResolver(Func<IResolveFieldContext<TSourceType>, TReturnType?> r
...
+                    // only re-use contexts that completed synchronously and do not return an IDataLoaderResult or an IEnumerable (that may be based on the context source)
                     if (ret is Task task)
                     {
-                        if (task.IsCompleted && task.Status == TaskStatus.RanToCompletion && !(task.GetResult() is IDataLoaderResult))
+                        if (task.IsCompleted && task.Status == TaskStatus.RanToCompletion)
                         {
-                            adapter.Reset();
-                            System.Threading.Interlocked.CompareExchange(ref _sharedAdapter, adapter, null);
+                            var ret2 = task.GetResult();
+                            if (ret2 is not IDataLoaderResult && (ret2 is not IEnumerable || ret2 is string))
```
sungam3r, line 96: Worth to move into local function and reuse.

sungam3r (approved):

lanwin:
Thank you! The new version seems to work.

A comprehension question. Shouldn't it be simpler to reuse context only for the "next" query. So within a query, or as long as the queries runs a context is never reused?

sungam3r:
Query has many fields. Context was created for each field to run its resolver (before reuse).

PR #2780, opened 2021-12-23
Introduce simplified null check [debatable]

sungam3r:
If OK then I'll copy-paste this approach in more places.

In the meantime waiting for https://github.com/dotnet/csharplang/issues/2145
-
UPDATE
Available solution for NET6 is [`ArgumentNullException.ThrowIfNull`](https://endjin.com/blog/2021/11/csharp-10-net-6-argument-exceptions).


Shane32:
I already described why I cannot endorse these changes.

sungam3r:
Release of `!!` feature was postponed. Not in C# 11.

sungam3r:
Example of implementation in other repo https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/1511.

src/GraphQL.MemoryCache/GraphQL.MemoryCache.csproj
```diff
@@ -1,13 +1,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net5</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
sungam3r, line 4:
```suggestion
    <TargetFrameworks>netstandard2.0;net5</TargetFrameworks>
```

src/GraphQL.MemoryCache/GraphQL.MemoryCache.csproj
```diff
@@ -1,13 +1,14 @@
...
     <Description>GraphQL memory cache implementation on top of Microsoft.Extensions.Caching.Memory</Description>
     <PackageTags>GraphQL;memory</PackageTags>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\GraphQL\GraphQL.csproj" PackageVersion="[$(Version),$(NextVersion))" />
+    <ProjectReference Include="..\GraphQL\GraphQL.csproj" />
```
sungam3r, line 11:
```suggestion
    <ProjectReference Include="..\GraphQL\GraphQL.csproj" PackageVersion="[$(Version),$(NextVersion))" />
```

src/GraphQL.MemoryCache/GraphQL.MemoryCache.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFrameworks>netstandard2.0;net5</TargetFrameworks>
     <Description>GraphQL memory cache implementation on top of Microsoft.Extensions.Caching.Memory</Description>
     <PackageTags>GraphQL;memory</PackageTags>
+    <Nullable>enable</Nullable>
```
sungam3r, line 7:
```suggestion
```

Shane32:
I suggest closing this PR.

gao-artur:
@Shane32, what are your objections? I really like the concise form of the `ArgumentNullException.ThrowIfNull`. Unfortunately, it's not available in older TFM's, but `Guard.ThrowIfNull` can be a good alternative.

Shane32:
Well, I guess I'm just not a fan of it.  `?? throw new ArgumentNullException(nameof(arg))` seems really simple.  Breakpoints are going to occur within the ThrowIfNull method, messing up the call stack... although if we do it right then it should get inlined in the release build.  Another benefit is that if we move the `throw` to a separate method then the null check should get inlined.  Another benefit could be that small methods might get inlined where as before they might not have been.  So there are some benefits.

gao-artur:
And I'm happy the `!!` was not added. C# starts looking like Predator's language.

PR #2785, opened 2021-12-23
Misc GraphQLBuilder changes

Shane32:

PR #2787, opened 2021-12-24
Fix check annotations

sungam3r:
See check annotations here https://github.com/graphql-dotnet/graphql-dotnet/pull/2785/files

### DIFF
--- a/src/GraphQL/Introspection/ISchemaComparer.cs
+++ b/src/GraphQL/Introspection/ISchemaComparer.cs
@@ -91,12 +91,12 @@ namespace GraphQL.Introspection
 
         private sealed class EnumValueByNameComparer : IComparer<EnumValueDefinition>
         {
-            public int Compare(EnumValueDefinition x, EnumValueDefinition y) => x.Name.CompareTo(y.Name);
+            public int Compare(EnumValueDefinition? x, EnumValueDefinition? y) => (x?.Name ?? "").CompareTo(y?.Name ?? "");
         }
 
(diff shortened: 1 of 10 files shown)


src/GraphQL/Introspection/ISchemaComparer.cs
```diff
@@ -76,17 +76,17 @@ public class AlphabeticalSchemaComparer : ISchemaComparer
...
         private sealed class DirectiveByNameComparer : IComparer<DirectiveGraphType>
         {
-            public int Compare(DirectiveGraphType x, DirectiveGraphType y) => x.Name.CompareTo(y.Name);
+            public int Compare(DirectiveGraphType? x, DirectiveGraphType? y) => (x?.Name ?? "").CompareTo(y?.Name ?? "");
         }
 
         private sealed class ArgumentByNameComparer : IComparer<QueryArgument>
         {
-            public int Compare(QueryArgument x, QueryArgument y) => x.Name.CompareTo(y.Name);
+            public int Compare(QueryArgument? x, QueryArgument? y) => (x?.Name ?? "").CompareTo(y?.Name ?? "");
```
sungam3r, line 89: For some reason, there are annotations for some converters and there is no for others. Perhaps this is done specifically in GitHub UI in order not to throw huge lists of the annotations.

sungam3r: Seems so.

Shane32 (approved):

Issue #2797, opened 2021-12-24
Change 'double registration' so one registration pulls from the other [Discuss]


Shane32:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/2571#discussion_r774757822

### Sample existing code

```cs
builder.Register<TDocumentListener>(serviceLifetime);
builder.Register<IDocumentExecutionListener, TDocumentListener>(serviceLifetime);
```

### Sample alternate code

```cs
builder.Register<TDocumentListener>(serviceLifetime);
builder.Register<IDocumentExecutionListener>(services => services.GetRequiredService<TDocumentListener>(), serviceLifetime);
```

### Considerations

- In the first example, it is possible to create/pull two different instances; in the second it is not
- In the second example, requesting `IDocumentExecutionListener` will result in the DI framework disposing of the instance twice.
- In the second example, if the service lifetimes were mismatched, a disposed instance could be returned when requesting `IDocumentExecutionListener`.

### Comments by @Shane32

I thought about this and decided against it, since it will be registering a service for disposal that it actually doesn't own, if you understand what I mean. Perhaps since the lifetime is the same here between ISchema and TSchema then it would not matter. I was not sure so I figured it would be safer to simply duplicate the couple lines of code over again.

I think perhaps there was other areas where I did use the pattern you suggested - I can't remember now. I know I thought a while about it. But here I felt it was better not to do so.

This is an area where Autofac can do better than MS DI -- with Autofac you can register a single factory/instance as multiple service types. (Not that it helps us here, just FYI.)

Note: Part of the decision was also based on the fact that the only time TSchema or ISchema is typically pulled from DI is during execution. It's not like we have other classes that pull ISchema and might be pulling a different instance than the execution which might pull TSchema. At least I couldn't think of any scenario.

### Comments by @sungam3r

I understand. This is not the first place where I see this. I came across such doubled registration in other projects. As a rule, this happened in the result of searching for bugs. We need to choose the smallest evil. I am sure that there will be more problems from double registration and double creating an object (singleton!).

sungam3r:
> In the second example, if the service lifetimes were mismatched

Both calls to `Register` use the same `serviceLifetime` parameter.

sungam3r:
What is the consensus? Will we change this or not?

sungam3r:
Reviewing #3520 I noticed that `AddSelfActivatingSchema` double registers schema.

sungam3r:
[...]

PR #2799, opened 2021-12-24
Refactor prep for issue 2797

Shane32:
No change to existing functionality.  Prep to change `RegisterAsBoth` and `TryRegisterAsBoth` to map one type to the other.

Ref:
- #2797

### DIFF
--- a/src/GraphQL/GraphQLBuilderExtensions.cs
+++ b/src/GraphQL/GraphQLBuilderExtensions.cs
@@ -89,35 +89,35 @@ namespace GraphQL
         #endregion
 
         #region - RegisterAsBoth and TryRegisterAsBoth -
-        internal static IGraphQLBuilder RegisterAsBoth<TService, TImplementation>(this IGraphQLBuilder builder, ServiceLifetime serviceLifetime)
+        private static IGraphQLBuilder RegisterAsBoth<TService, TImplementation>(this IGraphQLBuilder builder, ServiceLifetime serviceLifetime)
             where TService : class
             where TImplementation : class, TService
             => builder.Register<TImplementation>(serviceLifetime).Register<TService, TImplementation>(serviceLifetime);
 
-        internal static IGraphQLBuilder RegisterAsBoth<TService, TImplementation>(this IGraphQLBuilder builder, Func<IServiceProvider, TImplementation> implementationFactory, ServiceLifetime serviceLifetime)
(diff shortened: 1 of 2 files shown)


Shane32:
Due to the complicated nature of the tests, I would like to merge this before making the other change.

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -88,6 +88,56 @@ public static IGraphQLBuilder Configure<TOptions>(this IGraphQLBuilder builder,
...
+            where TImplementation : class, TService
+            => builder.Register(implementationInstance).Register<TService>(implementationInstance);
+
+        /// <summary>
+        /// Calls Register for the implementation and TryRegister for the service
+        /// </summary>
+        private static IGraphQLBuilder TryRegisterAsBoth<TService, TImplementation>(this IGraphQLBuilder builder, ServiceLifetime serviceLifetime)
+            where TService : class
+            where TImplementation : class, TService
+            => builder.Register<TImplementation>(serviceLifetime).TryRegister<TService, TImplementation>(serviceLifetime);
```
sungam3r, line 122:
```suggestion
            => builder.TryRegister<TImplementation>(serviceLifetime).TryRegister<TService, TImplementation>(serviceLifetime);
```

?

Shane32: No; please read xml comments and changed code.

Shane32: Note that these are private methods, so users cannot be confused as to why `TryRegisterAsBoth` does not actually perform `TryRegister` on both.

Shane32: We could name the methods `RegisterAndTryRegister` !!

sungam3r (approved):

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: ok

Shane32: ok ... rename to `RegisterAndTryRegister` ?  Or ok to leave as-is ?

Shane32:
I guess I'll just leave it and merge it then.  No problem if we want to change the name of those TryRegisterAsBoth methods.

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: whatever

PR #2801, opened 2021-12-24
Fix bug: schema instance was registered incorrectly by AddSchema

Shane32:
It would be disposed when it should not have been because the wrong `Register` method was called


Shane32:
Looks like I need to check into why the test is failing.

Shane32:
I guess it just needed a merge

sungam3r (approved):

PR #2637, opened 2021-08-25
Ensure that all tasks are awaited before execution strategy ends

Shane32:
Fixes #2635

PR #2723, opened 2021-12-05
Remove obsolete members from v5

Shane32:

PR #2802, opened 2021-12-25
Eliminate empty catch

sungam3r:
Review for #2637


Shane32 (approved):
You realize that the whole AwaitSafe method will be removed in v5? See PR #2723

sungam3r:
I guess

PR #2803, opened 2021-12-25
Use System.Runtime.CompilerServices.Unsafe only for netstandard2.1

sungam3r:
Review for #2677

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug_nested_type_names_dont_conflict.cs
+++ b/src/GraphQL.Tests/Bugs/Bug_nested_type_names_dont_conflict.cs
@@ -4,7 +4,6 @@ using Xunit;
 
 namespace GraphQL.Tests.Bugs
 {
-    [Collection("Sequential_UseDeclaringTypeNames")]
     public class Bug_nested_type_names_dont_conflict :
         QueryTestBase<Bug_nested_type_names_dont_conflict.MutationSchema>
     {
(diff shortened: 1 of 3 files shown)


Shane32 (approved):

src/GraphQL.Tests/Bugs/Bug_nested_type_names_dont_conflict.cs
```diff
@@ -4,7 +4,6 @@
 
 namespace GraphQL.Tests.Bugs
 {
-    [Collection("Sequential_UseDeclaringTypeNames")]
```
sungam3r, line 7: It does not work. From time to time, some tests fail due to `GlobalSwitches.UseDeclaringTypeNames` setting during parallel execution. This can be avoided only performing most tests in serial i.e. `Collection` attribute should be added to practically each schema test. So it's easier to restart tests.

src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
```diff
@@ -281,7 +281,7 @@ public void parseValue_from_system_text_json_ok(Type graphType, object value)
             var g = Create(graphType);
             var valueString = ((IFormattable)value).ToString(null, CultureInfo.InvariantCulture);
             object converted = SystemTextJson.StringExtensions.ToDictionary($"{{ \"arg\": {valueString} }}")["arg"];
-            g.CanParseValue(converted).ShouldBeTrue();
+            g.CanParseValue(converted).ShouldBeTrue($"Converted value: {converted}, Type: {converted.GetType()}");
```
sungam3r, line 284: This test failed once 🤔

Shane32:
I have the most problems locally with the tests that alter the current culture settings.  I'm sure it's the same thing - parallel tests altering some static property.  We have had problems with other tests also in a similar fashion in the past.

I'm okay skipping some tests that would otherwise cause repeated false failures.  Or if it "usually" works, re-running the tests in case of failure, or whatever you suggest.

sungam3r:
It fail rather rarely. Maybe 1/20.

Shane32:
This PR broke the develop branch.  It can't find `ValueTask`.  I can't tell why it's different than the master branch.

Shane32:
Well, I wrote #2808 to fix the develop branch, whether or not it's due to this PR.

sungam3r:
It was due to my other merge-pr.

PR #2804, opened 2021-12-25
Boost IsValidValue/ToAST

sungam3r:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/2707/files#r757949833

\+ some NRT annotations

ref #2707


src/GraphQL/Types/IProvideMetadata.cs
```diff
@@ -22,7 +22,7 @@ public interface IProvideMetadata
         /// <param name="key"> String key. </param>
         /// <param name="defaultValue"> It is used if there is no value for the given key. </param>
         /// <returns> Value of the specified type. </returns>
-        TType? GetMetadata<TType>(string key, TType? defaultValue = default);
+        TType GetMetadata<TType>(string key, TType defaultValue = default!);
```
sungam3r, line 25: The same logic here as for `GetArgument`.

Shane32:
I like everything in this PR ... except boosting `IsValidValue`.  It seems like a huge waste of CPU power.  If a user manually specifies an instance of their object that does not conform to their own definition, that's their problem.  Perhaps they did so on purpose, so that their code can detect a default value different from a deserialized value.  I really don't think we should add that code.

Shane32:
I agree with all the NRT changes.

Shane32:
I'm going to take another look at how it is used throughout the project.

Shane32:
It seems to be only used by the schema validator, so CPU load shouldn't be a problem.  I'll look it over some more and think about it.

sungam3r:
>  If a user manually specifies an instance of their object that does not conform to their own definition, that's their problem

😮 

I would like my mistakes to be discovered as early as possible. It is better to get an error when initializing the schema than at the execution stage some day. Especially in the production. I have seen such errors. Developers simply forgot about default values they set.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -60,7 +61,41 @@ public virtual object ParseDictionary(IDictionary<string, object?> value)
         }
 
         /// <inheritdoc/>
-        public virtual bool IsValidDefault(object value) => value is TSourceType;
+        public virtual bool IsValidDefault(object value)
+        {
+            if (value is not TSourceType)
+                return false;
+
+            foreach (var field in Fields)
```
Shane32, line 69:
```suggestion
            if (GetType() != typeof(InputObjectGraphType<TSourceType>)
                return true;

            foreach (var field in Fields)
```

How about this to account for if people derive from this class in order to change `ParseDictionary`? Otherwise every derived instance probably won't work correctly.

Shane32:
Also if people are relying on the `ValueConverter` to register conversion registrations that do not work in the default manner, it may also break this code, correct?

We should add notes to the 'breaking changes' section to this effect.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
sungam3r: I would not.

ParseDictionary:

> Overriding this method allows for customizing the deserialization process of input objects, much like a field resolver does for output objects.

`IsValidDefault` works for some already constructed runtime value. I want to say that these two methods are not connected with each other and are not required to correlate. Such a default implementation is quite simple and makes sense. In the end, both of them are virtual, so that if there are problems, they can always be overridden.

sungam3r: > Otherwise every derived instance probably won't work correctly.

?

sungam3r:
> Also if people are relying on the ValueConverter to register conversion registrations that do not work in the default manner, it may also break this code, correct?

I looked at the use of the converter, but did not understand whether it was worth to use in `IsValidDefault`.

sungam3r:
> We should add notes to the 'breaking changes' section to this effect.

Section of what? Target is `master`.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
Shane32: Any time that ParseDictionary is used, it is for the purpose of mutating the inputs from the AST or variables into the target object.  IsValidDefault checks the value against the AST.  It would be required to "unmutate" the values in order to check it against the AST.  So pretty much by definition, IsDefaultValue will be incorrect if ParseDictionary is used.  Further, IsDefaultValue is not generally necessary, and it should not be necessary to implement a complicated IsDefaultValue method just to use ParseDictionary.

If anything, we should be implementing a ToDictionary method, which IsDefaultValue can use to check the dictionary against the AST.  This would be much clearer of its intent and how to properly implement it.  But it would only ever be used for validating a default value, and seems rather pointless.

Shane32: I'm not sure if you've used ParseDictionary much, or read up on how other people are using it.  But I believe it's often when the shape of the input object does not match the shape of TObjectType.  Of course it could be used just for validating field values or similar.

Shane32:
> > We should add notes to the 'breaking changes' section to this effect.
> 
> Section of what? Target is `master`.

We certainly cannot commit this to master without the above change, restricting the scope only to its default usage.  It would break most use cases when ParseDictionary is used.  And even then, this change should be listed in the release notes as a potentially breaking change.

As I said before, people may have default values that do not conform to the AST so that it can be checked to see if it's the default or not - or for any other reason.  Or perhaps they made a mistake in their code.  This change can break their code, as it will not accept an "invalid" default value.  This PR may prevent broken introspection requests, I know, but that does not mean that it's not a breaking change.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
sungam3r: >  we should be implementing a ToDictionary method, which IsDefaultValue can use to check the dictionary against the AST. 

How `IsDefaultValue` can use it if it's just a dictionary?

sungam3r:
So... since I don't understand what exactly I can do here I suggest:
1. Rebase to `develop`
2. Make note about that _possible_ breaking change.
3. Release v5 and wait for feedback if any. Then tune this method.

Shane32:
Ok…

src/GraphQL/GraphQL.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageReference Include="System.Memory" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```
sungam3r, line 17: Fix after rebase.

sungam3r: @Shane32 You already fixed that.

sungam3r:
Conflicts resolved.

sungam3r:
@Shane32 ?

Shane32:
I'll try to review again this weekend.

sungam3r:
@Shane32 I pushed code that was stashed ~ one month ago. Then I wrote it with other purposes, and now it is required to rewrite `SchemaPrinter`.

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -1799,7 +1799,7 @@ namespace GraphQL.Types
     {
         bool IsValidDefault(object value);
         object ParseDictionary(System.Collections.Generic.IDictionary<string, object?> value);
-        GraphQLParser.AST.GraphQLObjectValue? ToAST(object value);
+        GraphQLParser.AST.GraphQLValue ToAST(object value);
```
sungam3r, line 1802: GraphQLObjectValue -> GraphQLValue to be able to return `GraphQLNullValue`

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -2035,11 +2035,15 @@ namespace GraphQL.Types
...
+        [System.Diagnostics.CodeAnalysis.DoesNotReturn]
         protected GraphQLParser.AST.GraphQLValue ThrowASTConversionError(object? value) { }
+        [System.Diagnostics.CodeAnalysis.DoesNotReturn]
         protected object ThrowLiteralConversionError(GraphQLParser.AST.GraphQLValue input, string? description = null) { }
+        [System.Diagnostics.CodeAnalysis.DoesNotReturn]
         protected object ThrowSerializationError(object? value) { }
+        [System.Diagnostics.CodeAnalysis.DoesNotReturn]
         protected object ThrowValueConversionError(object? value) { }
-        public virtual GraphQLParser.AST.GraphQLValue? ToAST(object? value) { }
+        public virtual GraphQLParser.AST.GraphQLValue ToAST(object? value) { }
```
sungam3r, line 2046: Either returns object or throws + each usage looks like `blabla.ToAST() ?? throw new ...`

sungam3r: I mean that each method actually never returns `null` (though it _can_).

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -60,20 +61,79 @@ public virtual object ParseDictionary(IDictionary<string, object?> value)
...
+        /// is <see langword="null"/> and <see cref="GraphQLObjectValue"/> filled with the values
+        /// for all input fields except ones returning <see cref="GraphQLNullValue"/>.
         /// <br/><br/>
-        /// The default implementation always throws an exception. It is recommended that this method be
-        /// overridden to support introspection of fields of this type that have default values. This method
-        /// is not otherwise needed to be implemented.
+        /// Note that you may need to override this method if you have already overrided <see cref="ParseDictionary"/>.
         /// </summary>
-        public virtual GraphQLObjectValue? ToAST(object value)
+        public virtual GraphQLValue ToAST(object? value)
```
sungam3r, line 91: Such a default implementation is similar to the one that is now resides in `SchemaPrinter` where string value is concatenated manually - https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/SchemaPrinter.cs#L382

docs2/site/docs/migrations/migration5.md
```diff
@@ -399,3 +399,10 @@ string s = writer.ToString();
 `SDLPrinter` is a highly optimized visitor for asynchronous non-blocking SDL output
 into provided `TextWriter`. In the majority of cases it does not allocate memory in
 the managed heap at all.
+
+### 21. More strict `InputObjectGraphType<TSourceType>.IsValidDefault`
+
+This method began to perform more checks of the specified value. Current implementation
+can cause problems in some situations. Changes are made for earlier error detection - at
+the schema initialization stage. If you encountered that your code stopped working,
+then feel free to post an issue.
```
Shane32, line 408: I reviewed everything again and the code is okay.  But the migration documentation needs to be rewritten.  This seems rather vague and does not indicate the most important point of the new code - ToAST is supported and enables printing a code-first schema that uses `InputObjectGraphType`.  We need not instruct the user to post an issue if they have a problem; let's just tell them what's changed and what they need to change in their code.

I would suggest that in the 'new features' section, you discuss support of `ToAST` and `IsValidDefault` and the fact that schemas with default input objects can be printed, while in the breaking changes you say that if you override `ParseDictionary`, they may now be required to also override `ToAST`.

Shane32 (changes requested):

docs2/site/docs/migrations/migration5.md
Shane32: Ideally I would suggest a small snippet of the prior one-line `IsValidDefault` method to revert behavior if necessary, so they can cut and paste it if there's an issue in their situation.

sungam3r: OK. I'll add notes tomorrow, 3 AM here.

sungam3r: Done. Not sure about wordinds.

Shane32: Good enough.  Probably we will review entire migration document before releasing anyway. :+1:

Shane32 (approved):

PR #2809, opened 2021-12-26
Rename BuildNamedType into BuildGraphQLType

### LINKED ISSUE #2805: Rename BuildNamedType to BuildGraphQLType
See http://spec.graphql.org/October2021/#NamedType

From method comments:
> List and non-null graph types are instantiated

So this is not a _named type_.

sungam3r:
fixes #2805

### DIFF
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -270,7 +270,7 @@ namespace GraphQL.Types
         /// </summary>
         public int Count => Dictionary.Count;
 
-        private IGraphType BuildNamedType(Type type, Func<Type, IGraphType> resolver) => type.BuildNamedType(t => FindGraphType(t) ?? resolver(t));
+        private IGraphType BuildNamedType(Type type, Func<Type, IGraphType> resolver) => type.BuildGraphQLType(t => FindGraphType(t) ?? resolver(t));
 
         /// <summary>
         /// Applies all delegates specified by the middleware builder to the schema.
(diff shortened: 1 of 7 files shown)


src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -270,7 +270,7 @@ private static IEnumerable<IGraphType> GetSchemaTypes(ISchema schema, IServicePr
         /// </summary>
         public int Count => Dictionary.Count;
 
-        private IGraphType BuildNamedType(Type type, Func<Type, IGraphType> resolver) => type.BuildNamedType(t => FindGraphType(t) ?? resolver(t));
+        private IGraphType BuildNamedType(Type type, Func<Type, IGraphType> resolver) => type.BuildGraphQLType(t => FindGraphType(t) ?? resolver(t));
```
Shane32, line 273: This is the only place the extension method is used.  How about making it a `protected virtual` method here and removing the extension method?

Shane32 (changes requested):

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: done

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -270,7 +270,41 @@ private static IEnumerable<IGraphType> GetSchemaTypes(ISchema schema, IServicePr
...
-        private IGraphType BuildNamedType(Type type, Func<Type, IGraphType> resolver) => type.BuildNamedType(t => FindGraphType(t) ?? resolver(t));
+        /// <summary>
+        /// Returns a new instance of the specified graph type, using the specified resolver to
+        /// instantiate a new instance if the required type cannot be found from the lookup table.
+        /// Defaults to <see cref="Activator.CreateInstance(Type)"/> if no <paramref name="resolve"/>
+        /// parameter is specified. List and non-null graph types are instantiated and their
+        /// <see cref="IProvideResolvedType.ResolvedType"/> property is set to a new instance of
+        /// the base (wrapped) type.
+        /// </summary>
+        protected internal virtual IGraphType BuildGraphQLType(Type type, Func<Type, IGraphType> resolve)
```
sungam3r, line 281: `internal` for tests

Shane32 (approved):

PR #2811, opened 2021-12-26
Split IGraphQLBuilder to IServiceRegister

### LINKED ISSUE #2771: Split IGraphQLBuilder to IServiceRegister
See https://github.com/graphql-dotnet/graphql-dotnet/pull/2571/files/59c1188c0bc8868784afdc29a2e7cc3104778b48#r773423125

sungam3r:
fixes #2771

### DIFF
--- a/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
+++ b/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
@@ -16,12 +16,15 @@ namespace GraphQL.MicrosoftDI
     /// An implementation of <see cref="IGraphQLBuilder"/> which uses the Microsoft dependency injection framework
     /// to register services and configure options.
     /// </summary>
-    public class GraphQLBuilder : GraphQLBuilderBase, IServiceCollection
+    public class GraphQLBuilder : GraphQLBuilderBase, IServiceCollection, IServiceRegister
     {
+        /// <inheritdoc />
+        public override IServiceRegister Services => this;
+
         /// <summary>
         /// Returns the underlying <see cref="IServiceCollection"/> of this builder.
         /// </summary>
-        public IServiceCollection Services { get; }
+        public IServiceCollection ServiceCollection { get; }
 
         /// <summary>
         /// Initializes a new instance for the specified service collection.
@@ -32,7 +35,7 @@ namespace GraphQL.MicrosoftDI
         /// </remarks>
         public GraphQLBuilder(IServiceCollection services, Action<IGraphQLBuilder>? configure)
         {
-            Services = services ?? throw new ArgumentNullException(nameof(services));
+            ServiceCollection = services ?? throw new ArgumentNullException(nameof(services));
             configure?.Invoke(this);
             RegisterDefaultServices();
         }
@@ -40,7 +43,7 @@ namespace GraphQL.MicrosoftDI
         /// <inheritdoc/>
         protected override void RegisterDefaultServices()
         {
-            Services.AddOptions();
+            ServiceCollection.AddOptions();
             base.RegisterDefaultServices();
         }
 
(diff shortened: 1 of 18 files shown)


src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
```diff
@@ -16,12 +16,15 @@ namespace GraphQL.MicrosoftDI
...
+    public class GraphQLBuilder : GraphQLBuilderBase, IServiceCollection, IServiceRegister
     {
+        /// <inheritdoc />
+        public override IServiceRegister Services => this;
+
         /// <summary>
         /// Returns the underlying <see cref="IServiceCollection"/> of this builder.
         /// </summary>
-        public IServiceCollection Services { get; }
+        public IServiceCollection ServiceCollection { get; }
```
sungam3r, line 27: `GraphQLBuilder` implements `IServiceCollection` and at the same time has ` public IServiceCollection ServiceCollection { get; }` . Are they both required?

Shane32: It implements `IServiceCollection` so that the server project can cast it to `IServiceCollection` without needing a reference to `GraphQL.MicrosoftDI`.  (Since to use the ASP.Net authorize extension, it needs access to the original service collection.)

The other design pattern would be for the server project to cast to `GraphQL.MicrosoftDI.GraphQLBuilder` and then access the `Services` property.  However, this means that it would not support any other Microsoft DI implementation unless it derives from `GraphQL.MicrosoftDI.GraphQLBuilder`.

I opted for the former choice.  The `Services` property exists as a convenience for anyone using the class directly.

Shane32: The third choice would be to add an interface within the `GraphQL.MicrosoftDI` package that encapsulates the `Services` property, for example `IMicrosoftServices` with one property `ServiceCollection` which returns the `IServiceCollection`.  But this seems a bit pointless.  Might as well just cast to `IServiceCollection`.

src/GraphQL.Tests/DI/GraphQLBuilderTests.cs
```diff
@@ -30,12 +34,14 @@ private void Verify()
 
         private void MockSetupRegister<TService, TImplementation>(ServiceLifetime serviceLifetime = ServiceLifetime.Singleton, bool replace = false)
         {
+            _builderMock.Setup(b => b.Services).Returns(_builder).Verifiable();
```
Shane32, line 37: Let's move this into the constructor, since it's necessary by nearly every test.  If necessary, we can remove `.Verifiable()` since as long as the mocked methods on `IServiceRegister` execute, we are fine.

Shane32: (If you disagree, disregard this comment; it's fine as-is also.)

Shane32 (approved):
Looks great!

Shane32:
Coverage is up 👍 

PR scope is well defined 👍 

Maintains access to `IServiceCollection` via `ServiceCollection` property 👍 

Maintains access to `IServiceCollection` via cast 👍 

Splitting into multiple files todo in separate PR 👍

src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
sungam3r: Are you OK with name `ServiceCollection`? You wanted `Services` as the name for `IServiceRegister` property so I had to choose another name for this property.

src/GraphQL.Tests/DI/GraphQLBuilderTests.cs
sungam3r: moved

Shane32 (approved):
Note the windows build failed - might want to rerun the tests to be sure it was a fluke.

src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
Shane32: Yes, perfect.

PR #2812, opened 2021-12-26
Introduce file scoped namespaces [DEMONSTRATION]

sungam3r:
See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces

For the demonstration I chose two projects in which conflicts should not arise with other PRs.

1 indentation saved! 😮 

Looks unusually IMO.

### DIFF
--- a/src/GraphQL.ApiTests/ApiApprovalTests.cs
+++ b/src/GraphQL.ApiTests/ApiApprovalTests.cs
@@ -9,57 +9,57 @@ using PublicApiGenerator;
 using Shouldly;
 using Xunit;
 
-namespace GraphQL.ApiTests
+namespace GraphQL.ApiTests;
+
+/// <see href="https://github.com/JakeGinnivan/ApiApprover"/>
+public class ApiApprovalTests
 {
-    /// <see href="https://github.com/JakeGinnivan/ApiApprover"/>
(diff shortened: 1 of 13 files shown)


Shane32:
Excellent!  As you create PRs, please add a comment if you change any code other than as described above, as it can be difficult to read the diff.  (The "hide whitespace" option works pretty well, though.)

However..... if you change this, whether it is in develop or master, we cannot merge from master to develop without EVERY changed file showing as a conflict.  (I think).  We may want to delay this change until after v5 is released and we only need to worry about the master branch.  Otherwise, I would suggest we merge this to develop, to ease merge conflicts.

Thoughts?

sungam3r:
I thought about conflicts. In fact, I do not have a lot of desire to merge this PR. Perceive it is more like a demonstration and "feature reminder". It is better to finish the main work with feature-PRs first, and then engage in "code decorating" (although main work apparently never end 🤔).

Shane32:
@sungam3r Please note that I will be on the road for the next few days, so I may be a little slow to respond, but I'll try to review your PRs promptly as I'm able.

sungam3r:
There were conflicts, let's not do so more untill merging develop into master.

PR #2815, opened 2021-12-27
Bump test packages

sungam3r:
replaces #2440 

Also bumps Alba to 6.0 and drops netcoreapp3.1 TFM in `GraphQL.Harness.Tests`


sungam3r:
@Shane32  I think it's OK to drop 3.1 support here. Or even all GraphQL.Harness.Tests project. These tests are more for server project.

Shane32:
I think we should have a reason for bumping Newtonsoft.Json.  The code works just as well with v13 as v12 so why bump it?  Let the user decide.  Perhaps v13 is a breaking change in their project for something unrelated to GraphQL.NET.  I'd rather see us add a test matrix to the GitHub action scripts to cover both v13 and v12, such as we have done for GraphQL.NET in the authentication project.  Then we can be confident that it will work for both v13 and v12 without forcing the user to use v13.

sungam3r:
https://github.com/JamesNK/Newtonsoft.Json/releases

Bugfixes, security vulnerabilities.

Shane32:
Even if that were a valid reason, there are no security fixes at all, and there are no bugfixes that could apply in any way to this project.

> New feature - Add JsonSelectSettings with configuration for a regex timeout
> Change - Remove portable assemblies from NuGet package
> Change - JsonReader and JsonSerializer MaxDepth defaults to 64

Not bugfixes or a security issues

> Fix - Fixed throwing missing member error on ignored fields

Does not apply

> Fix - Fixed various nullable annotations

Does not apply

> Fix - Fixed annotations not being copied when tokens are cloned

Does not apply

> Fix - Fixed naming strategy not being used when deserializing dictionary enum keys

Does not apply

> Fix - Fixed serializing nullable struct dictionaries

Does not apply

> Fix - Fixed JsonWriter.WriteToken to allow null with string token

Does not apply

> Fix - Fixed missing error when deserializing JToken with a contract type mismatch

Does not apply

> Fix - Fixed JTokenWriter when writing comment to an object

Does not apply

Shane32:
If we redo serialization libraries in #2760 then we should discuss if it's a good idea to use the latest dependencies...

sungam3r:
For the next major version I do not see anything bad.

> Not bugfixes or security issues

Vulnerability report - https://alephsecurity.com/vulns/aleph-2018004

Shane32:
> Vulnerability report - https://alephsecurity.com/vulns/aleph-2018004

Which was fixed in v12 with the MaxDepth property.

If anything, we should address the same flaw in the parser, as it exists there also.

```cs
        // force a stack overflow exception
        [Fact]
        public void WontStackOverflow()
        {
            var count = 24000;
            var sb = new System.Text.StringBuilder(count * 3);
            for (int i = 0; i < count; i++)
                sb.Append("{a");
            sb.Append(new string('}', count));
            var query = sb.ToString();
            _ = query.Parse();
        }
```

Shane32:
> Which was fixed in v12 with the MaxDepth property.

Correction: probably only if set.  Whereas in v13 it defaults to 64 instead of unlimited.

Shane32:
> > Which was fixed in v12 with the MaxDepth property.
> 
> Correction: probably only if set. Whereas in v13 it defaults to 64 instead of unlimited.

InputsConverter ignores MaxDepth anyway, so whether v12 or v13 is in use is moot; the problem still exists in GraphQL.NET.

Shane32:
> > > Which was fixed in v12 with the MaxDepth property.
> > 
> > 
> > Correction: probably only if set. Whereas in v13 it defaults to 64 instead of unlimited.
> 
> InputsConverter ignores MaxDepth anyway, so whether v12 or v13 is in use is moot; the problem still exists in GraphQL.NET.

Nevermind.  It actually does.

Shane32:
Let's merge this if/when #2760 is merged.  In the current version there is no way to configure deserialization options, so we might as well just set it to a maxdepth of 64.

Shane32:
Can we close this?  Newtonsoft.Json has already been updated, I believe.

Shane32 (approved):

PR #2816, opened 2021-12-27
Prepare v4.8

sungam3r:


src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>4.6.0-preview</VersionPrefix>
+    <VersionPrefix>4.8.0-preview</VersionPrefix>
```
sungam3r, line 4: We forgot to bump to 4.7.0 some time ago.

Shane32: For my own projects I just leave the version as 1.0.0 or similar, and all the nuget package published to GitHub will be like 1.0.0-172

Then when I publish it shows up on nuget with the correct version due to the tag - but I never need to update the props file or decide what version to set it to

Shane32: More likely if we release another 4.x version it will be 4.7.1. But we will see.

Shane32: So the reason it still says 4.6.0 is because you have not been with the project since 4.5.x was current, and I saw no reason to update directory props after releasing 4.6.0 or 4.6.1

sungam3r: See from here https://github.com/graphql-dotnet/graphql-dotnet/pull/2724#issuecomment-987328150

PR #2818, opened 2021-12-27
Generate received files for all TFMs in one test run.

sungam3r:
Tested locally.


sungam3r:
If OK I'll port it into server project.

Shane32 (approved):

PR #2819, opened 2021-12-27
Add style checks

sungam3r:


sungam3r:
~60-70 warnings fixed

src/GraphQL.DataLoader/DataLoaderDocumentListener.cs
```diff
@@ -33,7 +33,7 @@ public Task BeforeExecutionAsync(IExecutionContext context)
         }
 
         /// <inheritdoc/>
-        public Task BeforeExecutionAwaitedAsync(IExecutionContext context)
+        public Task BeforeExecutionAwaitedAsync(IExecutionContext _)
```
Shane32, line 36: Public api should not have argument names called `_`

src/GraphQL.ApiTests/GraphQL.DataLoader.approved.txt
```diff
@@ -58,8 +58,8 @@ namespace GraphQL.DataLoader
         public System.Threading.Tasks.Task AfterExecutionAsync(GraphQL.Execution.IExecutionContext context) { }
         public System.Threading.Tasks.Task AfterValidationAsync(GraphQL.Execution.IExecutionContext context, GraphQL.Validation.IValidationResult validationResult) { }
         public System.Threading.Tasks.Task BeforeExecutionAsync(GraphQL.Execution.IExecutionContext context) { }
-        public System.Threading.Tasks.Task BeforeExecutionAwaitedAsync(GraphQL.Execution.IExecutionContext context) { }
-        public System.Threading.Tasks.Task BeforeExecutionStepAwaitedAsync(GraphQL.Execution.IExecutionContext context) { }
+        public System.Threading.Tasks.Task BeforeExecutionAwaitedAsync(GraphQL.Execution.IExecutionContext _) { }
+        public System.Threading.Tasks.Task BeforeExecutionStepAwaitedAsync(GraphQL.Execution.IExecutionContext _) { }
```
Shane32, line 62: Public api should not have arguments named `_`

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -199,8 +199,8 @@ protected virtual void PreConfigure(Schema schema)
 
         protected virtual IGraphType GetType(string name)
         {
-            _types.TryGetValue(name, out IGraphType type);
-            return type;
+            _types.TryGetValue(name, out var type);
+            return type!;
```
Shane32, line 203: Seems like either it should be nullable return value or else use the Item property so an exception is thrown if the key doesn’t exist. Probably do the former and then change to the latter in a separate PR

Shane32 (changes requested):

src/GraphQL.ApiTests/GraphQL.DataLoader.approved.txt
Shane32: We should check why this method BeforeExecutionStepAwaitedAsync exists in the develop branch - I thought it was removed when the obsolete members was removed. Maybe I missed it.

sungam3r:
Done

Shane32 (approved):

PR #2820, opened 2021-12-28
Auto approve API

sungam3r:
@Shane32 I think you like it.

The idea is pretty simple. We have long we use this test not as a test, but as a tool for controlling changes in API. We manually rename _approval_ files. Now this happens automatically, but only locally to accidentally not miss the changes.

### DIFF
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -68,11 +68,6 @@ namespace GraphQL.Caching
             set => _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value, GetMemoryCacheEntryOptions(query));
         }
 
-        /// <summary>
-        /// Do somme.
-        /// </summary>
-        public static void DoSomething() { }
-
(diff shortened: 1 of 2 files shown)


src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -68,6 +68,11 @@ protected virtual MemoryCacheEntryOptions GetMemoryCacheEntryOptions(string quer
             set => _memoryCache.Set(query ?? throw new ArgumentNullException(nameof(query)), value, GetMemoryCacheEntryOptions(query));
         }
 
+        /// <summary>
+        /// Do somme.
+        /// </summary>
+        public static void DoSomething() { }
```
sungam3r, line 74: Demonstration. Test should fail.

sungam3r:
It should help beginners who are not familiar with this mechanism. When they start tests, they will see the modified files (not failed test) and most likely will commit changed _approved_ files among other changes.

sungam3r:
If OK I'll port Pr to other repos.

Shane32 (approved):

PR #2828, opened 2022-01-02
Set Newtonsoft MaxDepth to 64 and add tests

Shane32:
Note;
- With Newtonsoft.Json v12, maximum depth defaults to unlimited.  The ToInputs method cannot be configured in this version (see PR #2760 for v5 fix).  However, the user can write their own extension method using custom deserialization options, which will honor the maximum depth value set in the options.
- With Newtonsoft.Json v13, the maximum depth defaults to 64.
- With System.Text.Json, the maximum depth defaults to 64.

As a breaking change (perhaps in v4.8) we can add code to set the maximum depth, when not set, to 64.

Either way, without an updated parser, GraphQL.NET is vulnerable to this same security issue within the query, regardless of whether or not the variables deserializer is fixed.


src/GraphQL.Tests/Serialization/NewtonsoftJson/InputsConverterTests.cs
```diff
@@ -36,6 +37,13 @@ private string Serialize<T>(T data)
             return stringWriter.ToString();
         }
 
+        [Fact]
+        public void Throws_For_Deep_Objects()
+        {
+            var value = "{\"a\":" + new string('[', 100) + new string(']', 100) + "}";
```
sungam3r, line 43:
```suggestion
            var value = "{\"a\":" + new string('[', 65) + new string(']', 65) + "}";
```
?

src/GraphQL.Tests/Serialization/SystemTextJson/InputsConverterTests.cs
```diff
@@ -30,6 +30,13 @@ public class InputsConverterTests
             }
         };
 
+        [Fact]
+        public void Throws_For_Deep_Objects()
+        {
+            var value = "{\"a\":" + new string('[', 100) + new string(']', 100) + "}";
```
sungam3r, line 36:
```suggestion
            var value = "{\"a\":" + new string('[', 65) + new string(']', 65) + "}";
```

sungam3r:
Maybe do all-all changes on `develop` at once?

> Either way, without an updated parser, GraphQL.NET is vulnerable to this same security issue within the query, regardless of whether or not the variables deserializer is fixed.

Parser has only `master` branch and `master` tracks all changes now - backward compatible or not. So why to bother with `master` here? I would use `develop` and not split changes into "phases".

Shane32:
I had planned to release a compatible version of parser as it would fix v4 - but it seems that we have already made breaking changes and may not be worth the trouble.

Sure, I’ll move these changes to develop.

sungam3r:
Too much changes from other commits in master.

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -15,6 +15,7 @@ public static class StringExtensions
         {
             DateFormatHandling = DateFormatHandling.IsoDateFormat,
             DateParseHandling = DateParseHandling.None,
+            //TODO: MaxDepth = 64,
```
Shane32, line 18:
```suggestion
            MaxDepth = 64, // already default for Newtonsoft.Json v13 -- but for v12, unlimited was default
```

src/GraphQL.Tests/Serialization/NewtonsoftJson/InputsConverterTests.cs
```diff
@@ -13,6 +13,7 @@ public class InputsConverterTests
             DateFormatHandling = DateFormatHandling.IsoDateFormat,
             DateParseHandling = DateParseHandling.None,
             Formatting = Formatting.Indented,
+            MaxDepth = 64, //default for Newtonsoft.Json v13, but for v12, unlimited is default
```
Shane32, line 16:
```suggestion
```

src/GraphQL/Utilities/Federation/FederatedSchemaBuilder.cs
```diff
@@ -157,7 +157,7 @@ private UnionGraphType BuildEntityGraphType()
 
             union.ResolveType = x =>
             {
-                if (x is Dictionary<string, object> dict && dict.TryGetValue("__typename", out object typeName))
+                if (x is Dictionary<string, object> dict && dict.TryGetValue("__typename", out object? typeName))
```
Shane32, line 160: The formatter wasn't happy without this change

Shane32:
@sungam3r Ready for review

src/GraphQL.NewtonsoftJson/StringExtensions.cs
```diff
@@ -15,6 +15,7 @@ public static class StringExtensions
         {
             DateFormatHandling = DateFormatHandling.IsoDateFormat,
             DateParseHandling = DateParseHandling.None,
+            MaxDepth = 64, // already default for Newtonsoft.Json v13 -- but for v12, unlimited was default
```
sungam3r, line 18:
```suggestion
            MaxDepth = 64, // already default since Newtonsoft.Json v13 -- but unlimited was default before
```

sungam3r (approved):
What about bumping Newtonsoft.Json?

Shane32:
> What about bumping Newtonsoft.Json?

It is unnecessary for the purposes of this PR.  The only way to deserialize data with this library is to use `ToInputs` or `FromJson`, which will now use a maximum depth of 64 even on the older version.

However, it should be bumped within #2760 or #2831 which supports deserialization because the `IGraphQLBuilder` extension method has a configuration delegate and the configuration options can't be set prior to the delegates being called.

So I bumped Newtonsoft.Json in PR #2831

Shane32:
Pretty much all these changes are obsolete with #2831

sungam3r (approved):

Issue #2837, opened 2022-01-04
[Proposal] Extend AutoRegisteringInputObjectGraphType<T>


Shane32:
### Proposed design and goals:

- Auto register public writable properties of generic type `T`

- Inherits `InputObjectGraphType<T>`

- Class features
  - Supports `[Name(...)]` on the class to set the graph type name
    - Defaults to the name of `T` after removing "Model" from the end of `T` if it exists.
  - Supports `[Description(...)]` and `[Obsolete(...)]` to set the description and deprecation reason
  - Can inherit from this class and add code to the derived class's constructor to add additional fields or modify existing ones.

- Field features
  - Skips properties marked with `[Ignore]`
  - Supports `[Name(...)]` on the property to set the field name
  - Supports `[Description(...)]` and `[Obsolete(...)]` to set the description and deprecation reason
  - Supports `[GraphType(...)]` to manually select an exact graph type (ignoring all type inference)
  - Supports `[DefaultValue(...)]` to set a default value for intrinsic values, and for complex types, `[DefaultValueAst(...)]` to set a complex default value from a string that is parsed as an AST and converted to the graph type's type just like the schema builder would do.
  - Without manually selecting a graph type, will auto generate a graph type using `GetGraphTypeFromType` method
    - Supports nullability attributes to determine nullability of type (and if applicable its list) -- for example:
      - `List<string?>` is `NonNullGraphType<ListGraphType<StringGraphType>>`
      - `List<string>?` is `ListGraphType<NonNullGraphType<StringGraphType>>`
    - For libraries without NRT annotations, or to override type nullability inference, supports `[Required]`, `[Optional]`, `[RequiredList]` and `[OptionalList]` attributes (applied to the property for ease of use)
    - Supports `[Id]` to indicate that a member should be an `IdGraphType`
      - Works with NRT annotations and lists, etc

- Class is completely extendable with overridable members used to build the graph.  Typically this would be meant for derived generic classes, as there should be enough extensibility with attributes to not require code to modify behavior.
  - `GetDefaultName()` to set the default name of the graph.  Derived generic classes may override this member to customize default naming.
  - `GetPropertiesToProcess()` returns the set of properties of `T` that should be added to the graph.  Derived classes may use this method to customize how property scanning is done.
  - `ProcessProperty(PropertyInfo property)` returns a `FieldType` for the property
[...]

PR #2840, opened 2022-01-04
Add nullability inference by examining NRT annotations

Shane32:
Part 1 of #2837 -- add the ability to interpret nullability reference type annotations from user code.


sungam3r:
Isn't there any ready-made package that provides nullable information?

Shane32:
I don't know.  Even if there were, this code would eliminate a dependency.  Coverage should be at 100%, and I've covered every scenario I can think of.  As we can see from the api tests, even third party code is not always dependable.

Shane32:
The actual executable code here is only 250 lines of code, and that includes about 80 lines of comments.

I've been using this code for quite a while in `Shane32.GraphQL.DI`, by the way.  It works very well from my experience.

Shane32:
Of course I could package the code into a separate nuget package...

Shane32:
Well coverage is about 100% except for the exceptions thrown if the NRT annotations are unable to be parsed... which makes it only 87.5%.  I forgot that I couldn't simulate those errors.

sungam3r:
> Of course I could package the code into a separate nuget package...

@Shane32 I really advise you to pack this code into **source only** nuget package. I use them from time to time when I want to reuse some technical small code and do not want to have an external dependency in the created package.

src/GraphQL/Reflection/Nullability.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public enum Nullability
+    {
+        /// <summary>
+        /// The member is a reference type not marked with nullable reference annotations.
+        /// </summary>
+        Unknown = 0,
+
+        /// <summary>
+        /// The member is a non-nullable value type or is a nullable refernce type marked as non-nullable.
```
sungam3r, line 14:
```suggestion
        /// The member is a non-nullable value type or is a nullable reference type marked as non-nullable.
```

src/GraphQL/Reflection/NullabilityInterpretationException.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Reflection;
+
+namespace GraphQL.Reflection
+{
+    /// <summary>
+    /// Exception when nullable reference annotations cannot be read from a parameter.
```
sungam3r, line 7:
```suggestion
    /// Exception when nullable reference annotations cannot be read from a member.
```

Shane32:
> source only nuget package

Makes sense but I have not heard of that for C#.  Do you know of a repo with CI set up and so on that I can look at, and a repo that uses it?

src/GraphQL/Reflection/NullabilityInterpretationException.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class NullabilityInterpretationException : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance for the specified <paramref name="parameter"/>.
+        /// </summary>
+        public NullabilityInterpretationException(ParameterInfo parameter) : base($"Unable to interpret nullability attributes for argument '{parameter.Name}' on method '{parameter.Member.DeclaringType.Name}.{parameter.Member.Name}'.")
```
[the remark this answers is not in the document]
sungam3r: Why only one special ctor for parameter?

sungam3r:
It will be easier for me to copy one of my projects and show.

src/GraphQL/Reflection/ReflectionExtensions.cs
```diff
@@ -0,0 +1,199 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace GraphQL.Reflection
+{
+    /// <summary>
+    /// Provides extension methods to <see cref="ParameterInfo"/> and <see cref="MemberInfo"/> instances to read nullable type reference annotations.
```
sungam3r, line 9: PropertyInfo too. What about `MethodInfo`?

Shane32:
```suggestion
    /// Provides extension methods to <see cref="ParameterInfo"/> and <see cref="PropertyInfo"/> instances to read nullable type reference annotations.
```
All other methods are private

Shane32: Both arguments and the return value are parameters of a member.  So the ParameterInfo extension method covers those cases while the PropertyInfo extension method covers properties.

sungam3r:
Sample csproj for building source-only package. Try it. For CI you may configure some props as we did it in Auth project to test different package versions.

```xml
<!--https://medium.com/@attilah/source-code-only-nuget-packages-8f34a8fb4738-->
<Project>
  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />

  <PropertyGroup>
    <NoBuild>true</NoBuild> <!-- Set false and comment Compile+CopyFilesToOutputDirectory targets to check build works -->
    <IsPackable>true</IsPackable>
    <TargetFramework>netstandard2.1</TargetFramework>
    <GeneratePackageOnBuild>true</GeneratePackageOnBuild> <!-- NETSDK1085 The 'NoBuild' property was set to true but the 'Build' target was invoked. -->
    <Description>YOUR TEXT</Description>
    <NoWarn>$(NoWarn);NU5128</NoWarn>
  </PropertyGroup>

  <PropertyGroup Condition="'$(NoBuild)' == 'True'">
    <IncludeBuildOutput>false</IncludeBuildOutput>
    <ContentTargetFolders>contentFiles</ContentTargetFolders>
    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
    <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
    <SonarQubeExclude>true</SonarQubeExclude>
  </PropertyGroup>

  <ItemGroup Condition="'$(NoBuild)' == 'True'">
    <Compile Update="@(Compile)">
      <Pack>true</Pack>
      <PackagePath>$(ContentTargetFolders)\cs\$(TargetFramework)\$(PackageId)\%(RecursiveDir)</PackagePath>
    </Compile>
  </ItemGroup>

  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />

  <!-- Comment 2 targets to check build works. Alas, Condition attribute does not work on Target. -->
  <Target Name="Compile" />
  <Target Name="CopyFilesToOutputDirectory" />

</Project>
```

Shane32:
Thanks!

sungam3r:
Usage:
```
<PackageReference Include="PackageName" Version="1.0.0" PrivateAssets="all" />
```

PrivateAssets should be _all_ to not have any dependencies in consuming package. So source only package allows you to distribute your source code without dependencies.

sungam3r:
>  <SonarQubeExclude>true</SonarQubeExclude>

You don't need this.

sungam3r:
Should I continue review here or wait to review another repo?

Shane32:
here

src/GraphQL/Reflection/ReflectionExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+        private static (Nullability, IList<CustomAttributeTypedArgument>?) GetMethodParameterNullability(ParameterInfo parameter)
+        {
+            var nullableContext = GetMethodDefaultNullability(parameter.Member);
+
+            return ExamineMemberAttributes(nullableContext, parameter.CustomAttributes)
+                ?? throw new NullabilityInterpretationException(parameter);
+        }
+
+        /// <summary>
+        /// Returns the default NRT annotation for the member, and a list of the NRT annotations for the type.
```
sungam3r, line 71: > for the type

What type? Type of member? Declaring type?

src/GraphQL/Reflection/ReflectionExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+            var list = new List<(Type, Nullability)>();
+            var index = 0;
+            Consider(property.PropertyType, false, list, nullabilityBytes, nullableContext, ref index);
+            if (nullabilityBytes != null && nullabilityBytes.Count != index)
+                throw new NullabilityInterpretationException(property);
+            return list;
+        }
+
+        /// <summary>
+        /// Returns the default NRT annotation for the parameter, and a list of the NRT annotations for the type.
```
sungam3r, line 60: What type? Type of member? Declaring type?

src/GraphQL/Reflection/ReflectionExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+{
+    /// <summary>
+    /// Provides extension methods to <see cref="ParameterInfo"/> and <see cref="PropertyInfo"/> instances to read nullable type reference annotations.
+    /// </summary>
+    public static class ReflectionExtensions
+    {
+        /// <summary>
+        /// Examines the parameter type and walks the type definitions, for each returning the nullability information for the type.
+        /// <para>
+        /// For instance, for Test&lt;string, int&gt;, first returns nullability for Test&lt;string, int&gt;, then string, then int.
```
sungam3r, line 16: It is not clear from this example, in what order you need to traverse the tree of types. You provides a link to nullable-metadata docs so whatever. One can open and read.

sungam3r (commented):
My thoughts are:
1. I understand what this code does.
2. I did not read it thoroughly.
3. A long time ago (in 2019 maybe) I added NRT inference support in [PublicApiApprover](https://github.com/PublicApiGenerator/PublicApiGenerator) and **it was a nightmare**. I read all that documentation and debugged the types traversal.
4. I definitely do not want to merge (and then maintain) such code in the main GraphQL.NET repository or in any other org repo. Similarly with tests. This is not a GraphQL.NET task.
5.  I'm fine with some external (preferably source-only) package (your or somebody else) to consume from GraphQL.NET.
6. Did you see https://github.com/SimonCropp/NullabilityInfo ?

So my conslusion is that you should to transfer all this code to another repo or use some other ready-made package. Search suggestions:
- https://www.nuget.org/packages?q=nullable
- https://www.nuget.org/packages?q=nullability

I can see a few source-only packages in that list.

Shane32:
I know my own code; I know how much time I spent on this and how many situations I carefully covered. Personally, I don’t want to debug someone else’s code in this case.

I can easily make this into a separate nuget package. I’d make it a binary package, but we could have it also packaged as source only. If we want to go this route we could work on it together. 

I can review SimonCropp’s code … but this code is already done and verified for exactly our needs, including extensive testing.

Shane32:
I looked around and in .NET 6 there is a built in class to read nullability information. (I wrote this code last year before .NET 6.) We should probably use the built in classes for .NET 6 and a source-compatible library for .NET Standard 2.0+ such as SimonCropp’s library. Or adjust mine to use the standardized interface. It seems that SimonCropp’s library is already in a source-only nuget package.

I suggest we keep the classes and tests here for two reasons: double check the third party library, and because the other tests I have written all use these same classes. We may eliminate the VerifyTestClasses tests if you like (since these tests verify the build process in order that nullability tests are testing what it was designed to. But I’d leave them in.

sungam3r:
> I suggest we keep the classes and tests here for two reasons: 

I would not. Why not just add nessesary tests in other repo?

> I looked around and in .NET 6 there is a built in class to read nullability information.

🤔 Perhaps we can copy-paste built-in bits or maybe someone already did netstandard compatible package. I saw source only packages for nullability attributes in aforementioned nuget links.

sungam3r:
By the way, what is the name of built-in nullability class?

Shane32:
https://docs.microsoft.com/en-us/dotnet/api/system.reflection.nullabilityinfocontext?view=net-6.0

Shane32:
> I would not. Why not just add nessesary tests in other repo?

Be fine if it was ours. We don’t necessarily have control of others’ repos. This is delicate code. I want to be sure it will work properly. Besides, I still need the test classes and most of the test setup for the rest of the tests. 

> 🤔 Perhaps we can copy-paste built-in bits or maybe someone already did netstandard compatible package. I saw source only packages for nullability attributes in aforementioned nuget links.

Like I said, SimonCropp’s package is a source only nuget package and is source compatible with the .NET 6 library. I’m not sure if he copied the source or not.

I just am quite busy to start investigating this. And then adapting all my code to use it.

Shane32:
I wrote code to utilize the .NET 6 nullability inference code.  I can't get it to pass my tests.

PR #2841, opened 2022-01-05
Support naming policy in serializers

### LINKED ISSUE #2839: Setting DocumentWriter options does not make any difference in serialization
## Description

We are migrating from GraphQL 3.2 to 4.7 and I noticed setting the options of the DocumentWriter does not have any impact on the response serialization. 

### Steps to reproduce

Consider the following code:
```csharp
void Main()
{
	var schema = new Schema { Query = new TestQuery() };

	var exec = new DocumentExecuter();

	var result = exec.ExecuteAsync(options =>
			{
				options.Schema = schema;
				options.Query = "{ test { thisIsAString, thisIsADateTime } }";
			}
	).Result;

	var writer = new DocumentWriter(options =>
	{
		options.Converters.Add(new IsoDateTimeConverter()
		{
		    DateTimeFormat = "yyyy-MMM-dd"
		});

		options.ContractResolver = new ExecutionResultContractResolver(new ErrorInfoProvider())
		{
			NamingStrategy = new KebabCaseNamingStrategy(true, false, false)
		};
	});
	
	writer.WriteToStringAsync(result).Dump();

}

public class TestQuery : ObjectGraphType
{
	public TestQuery()
	{
		Name = "testQuery";
		Description = "Test description";

		Field<TestResponseType>(
			name: "test",
			resolve: context =>
			{
				return new TestResponse();
			}
		);
	}
}

public class TestResponseType : ObjectGraphType<TestResponse>
{
	public TestResponseType()
	{
		Name = "TestReponse";
		Field<StringGraphType>("ThisIsAString");
		Field<DateTimeGraphType>("ThisIsADateTime");
	}
}

public class TestResponse
{
	public string ThisIsAString { get; set; }
	public DateTime ThisIsADateTime { get; set; }
	public TestResponse()
	{
		ThisIsAString = "String Value";
		ThisIsADateTime = new DateTime(2022, 1, 4, 10, 20, 30);
	}
}

```

### Expected result

```javascript
{"data":{"test":{"this-is-a-string":"String Value","this-is-a-date-time":"2022-Jan-04"}}}
```

### Actual result
```javascript
{"data":{"test":{"thisIsAString":"String Value","thisIsADateTime":"2022-01-04T10:20:30"}}}
```

It works as expected on version 3.2.0

Shane32:
Fixes both the Newtonsoft.Json and System.Text.Json serializers to honor the naming policy set in the serializer's options.

Since the _effective_ property naming policy for both System.Text.Json and Newtonsoft.Json was `null` (unchanged) for `ExecutionResult` nodes, I changed the default naming policy to `null` for System.Text.Json to prevent this bugfix from being a breaking bug fix.

However, with the System.Text.Json serializer only, child nodes of `extensions` previously used camelCase property naming policy whereas now it will use 'none' property naming policy by default.  (Apollo tracing nodes will continue to use camelCase property name encoding.)  This was the intended behavior and matches that of the Newtonsoft.Json converter.

### DIFF
--- a/docs2/site/docs/migrations/migration4.md
+++ b/docs2/site/docs/migrations/migration4.md
@@ -214,6 +214,7 @@ Here are some of the situations that you may run into with version 4:
 - `EnumGraphType` is stricter, requiring internal values for serialization, and external values for deserialization
 - `IdGraphType` (which allows any basic type) does not coerce variable values to trimmed strings during deserialization
 - `IdGraphType` does not trim serialized values (but does convert them to strings)
+- `DateTimeGraphType` serializes values to strings instead of letting the JSON serializer do so
 
 If you have a schema-first schema, you may run into an issue with enumeration types, since the `SchemaBuilder` uses the name of each
 enumeration value as its value also. In other words, you must return a string corresponding to the enumeration value (e.g, `"Cat"` or
--- a/src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
+++ b/src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
@@ -18,14 +18,14 @@ namespace GraphQL.NewtonsoftJson
         /// Initializes a new instance with the specified <see cref="IErrorInfoProvider"/>.
         /// </summary>
         public ExecutionResultJsonConverter(IErrorInfoProvider errorInfoProvider)
-            : this(errorInfoProvider, new CamelCaseNamingStrategy())
+            : this(errorInfoProvider, null)
         {
         }
 
         /// <summary>
-        /// Initializes a new instance with the specified <see cref="IErrorInfoProvider"/>.
+        /// Initializes a new instance with the specified <see cref="IErrorInfoProvider"/> and <see cref="NamingStrategy"/>.
         /// </summary>
-        public ExecutionResultJsonConverter(IErrorInfoProvider errorInfoProvider, NamingStrategy? namingStrategy)
+        public ExecutionResultJsonConverter(IErrorInfoProvider errorInfoProvider, NamingStrategy namingStrategy)
         {
             _errorInfoProvider = errorInfoProvider ?? throw new ArgumentNullException(nameof(errorInfoProvider));
             _namingStrategy = namingStrategy;
--- a/src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
+++ b/src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
@@ -2,6 +2,7 @@ using System;
 using System.Linq;
 using GraphQL.Execution;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
 
 namespace GraphQL.NewtonsoftJson
 {
@@ -11,13 +12,23 @@ namespace GraphQL.NewtonsoftJson
     public class ExecutionResultJsonConverter : JsonConverter
     {
         private readonly IErrorInfoProvider _errorInfoProvider;
+        private readonly NamingStrategy _namingStrategy;
 
         /// <summary>
         /// Initializes a new instance with the specified <see cref="IErrorInfoProvider"/>.
         /// </summary>
         public ExecutionResultJsonConverter(IErrorInfoProvider errorInfoProvider)
+            : this(errorInfoProvider, new CamelCaseNamingStrategy())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance with the specified <see cref="IErrorInfoProvider"/>.
+        /// </summary>
+        public ExecutionResultJsonConverter(IErrorInfoProvider errorInfoProvider, NamingStrategy? namingStrategy)
         {
             _errorInfoProvider = errorInfoProvider ?? throw new ArgumentNullException(nameof(errorInfoProvider));
+            _namingStrategy = namingStrategy;
         }
 
         /// <inheritdoc/>
@@ -67,7 +78,10 @@ namespace GraphQL.NewtonsoftJson
                     writer.WriteStartObject();
                     foreach (var childNode in objectExecutionNode.SubFields)
                     {
-                        writer.WritePropertyName(childNode.Name);
+                        var propName = childNode.Name;
+                        if (_namingStrategy != null)
+                            propName = _namingStrategy.GetPropertyName(propName, false);
+                        writer.WritePropertyName(propName);
                         WriteExecutionNode(writer, childNode, serializer);
                     }
                     writer.WriteEndObject();
--- a/src/GraphQL.SystemTextJson/ApolloTraceJsonConverter.cs
+++ b/src/GraphQL.SystemTextJson/ApolloTraceJsonConverter.cs
@@ -1,0 +1,33 @@
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using GraphQL.Instrumentation;
+
+namespace GraphQL.SystemTextJson
+{
+    /// <summary>
+    /// Converts an instance of <see cref="ApolloTrace"/> to/from JSON.
+    /// </summary>
+    public class ApolloTraceJsonConverter : JsonConverter<ApolloTrace>
+    {
+        private readonly JsonSerializerOptions _optionsNoIndent = new()
+        {
+            WriteIndented = false,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        };
+
+        private readonly JsonSerializerOptions _optionsIndent = new()
+        {
+            WriteIndented = true,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        };
+
+        /// <inheritdoc/>
+        public override void Write(Utf8JsonWriter writer, ApolloTrace value, JsonSerializerOptions options)
+            => JsonSerializer.Serialize(writer, value, options.WriteIndented ? _optionsIndent : _optionsNoIndent);
+
+        /// <inheritdoc/>
+        public override ApolloTrace Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+            => JsonSerializer.Deserialize<ApolloTrace>(ref reader);
+    }
+}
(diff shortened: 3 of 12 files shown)


src/GraphQL.SystemTextJson/DocumentWriter.cs
```diff
@@ -131,7 +131,7 @@ private void ConfigureOptions(IErrorInfoProvider errorInfoProvider)
         }
 
         private static JsonSerializerOptions GetDefaultSerializerOptions(bool indent)
-            => new JsonSerializerOptions { WriteIndented = indent, PropertyNamingPolicy = JsonNamingPolicy.CamelCase };
+            => new JsonSerializerOptions { WriteIndented = indent };
```
Shane32, line 134: Many tests failed without this change (since the naming policy was being ignored previously).

Shane32:
Code is failing now because no explicit support for ApolloTrace was added to System.Text.Json ... working on it

src/GraphQL.Tests/DocumentWritersTestData.cs
```diff
@@ -10,7 +10,6 @@ public class DocumentWritersTestData : IEnumerable<object[]>
             new SystemTextJson.DocumentWriter(new System.Text.Json.JsonSerializerOptions
             {
                 WriteIndented = true,
-                PropertyNamingPolicy = System.Text.Json.JsonNamingPolicy.CamelCase,
```
Shane32, line 13: CamelCase was and is not the default.  Tests need to examine the default situation.

Shane32:
@sungam3r Ready for review here.

src/GraphQL.Tests/DocumentWritersTestData.cs
rmeshksar: What will be the default naming policy after this? I thought camelCase was the default because when I define the field like this:
```
Field<DateTimeGraphType>("ThisIsADateTime");
```

The response is:
```
{"data":{"test":{"thisIsAString":"String Value"}}}
```

Shane32: The schema has a default naming policy of camelCase. This occurs within the schema before it hits the serializer

docs2/site/docs/migrations/migration4.md
```diff
@@ -214,6 +214,7 @@ Here are some of the situations that you may run into with version 4:
 - `EnumGraphType` is stricter, requiring internal values for serialization, and external values for deserialization
 - `IdGraphType` (which allows any basic type) does not coerce variable values to trimmed strings during deserialization
 - `IdGraphType` does not trim serialized values (but does convert them to strings)
+- `DateTimeGraphType` serializes values to strings instead of letting the JSON serializer do so
```
sungam3r, line 217: And it looks correct. Serializer settings should not necessarily affect (although they **can**) all returned results. Some scalars have _fixed_ format by their nature. Perhaps it is worth adding a note about it to the documentation.

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
```diff
@@ -11,13 +12,23 @@ namespace GraphQL.NewtonsoftJson
...
         /// </summary>
         public ExecutionResultJsonConverter(IErrorInfoProvider errorInfoProvider)
+            : this(errorInfoProvider, null)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance with the specified <see cref="IErrorInfoProvider"/> and <see cref="NamingStrategy"/>.
+        /// </summary>
+        public ExecutionResultJsonConverter(IErrorInfoProvider errorInfoProvider, NamingStrategy namingStrategy)
```
sungam3r, line 28:
```suggestion
        public ExecutionResultJsonConverter(IErrorInfoProvider errorInfoProvider, NamingStrategy? namingStrategy)
```
`DefaultContractResolver.NamingStrategy` is nullable.

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
```diff
@@ -11,13 +12,23 @@ namespace GraphQL.NewtonsoftJson
     public class ExecutionResultJsonConverter : JsonConverter
     {
         private readonly IErrorInfoProvider _errorInfoProvider;
+        private readonly NamingStrategy _namingStrategy;
```
sungam3r, line 15:
```suggestion
        private readonly NamingStrategy? _namingStrategy;
```

src/GraphQL.SystemTextJson/DocumentWriter.cs
sungam3r: New tests?

src/GraphQL.SystemTextJson/ApolloTraceJsonConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using GraphQL.Instrumentation;
+
+namespace GraphQL.SystemTextJson
+{
+    /// <summary>
+    /// Converts an instance of <see cref="ApolloTrace"/> to/from JSON.
+    /// </summary>
+    public class ApolloTraceJsonConverter : JsonConverter<ApolloTrace>
```
sungam3r, line 11: Why do this class need? I thought that ApolloTrace serialization just works.

sungam3r: This happened because of `NamingPolicy` removal?

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
Shane32: Yes, but this project does not have NRT annotations enabled.  We need to finish that up for all the other projects in the solution.  Only the main project has NRT enabled.

src/GraphQL.SystemTextJson/ApolloTraceJsonConverter.cs
Shane32: In a manner of speaking, the fact this did not exist was a bug.  If you look in the contract resolver for Newtonsoft.Json there exists code to detect properties of ApolloTrace and force a camelCase naming policy.  That code has never existed in this project for System.Text.Json.  So when using any other naming policy besides camelCase, serialization would be incorrect.  And the System.Text.Json serializer has always had a default of 'none' for the naming policy.  (At least since v4 for sure.)

Now the weird thing is that we had a test to ensure that ApolloTrace is serialized correctly.  But we had overridden the naming policy in the configuration of the test writer, so it was not testing a default scenario.  Moreover, it did not test what would happen when a custom naming policy was applied.  So it really wasn't testing what it was designed to test.  (Most likely the test was written for Newtonsoft.Json before we changed the tests to prefer System.Text.Json in 3.x)

Now the tests check both GraphQL.SystemTextJson and GraphQL.NewtonsoftJson.

> This happened because of `NamingPolicy` removal?

Yes -- in the tests project.  Not because the naming policy changed in the GraphQL.NewtonsoftJson project.

src/GraphQL.SystemTextJson/DocumentWriter.cs
Shane32: Well, there's plenty of tests that already fail if this isn't 'none'.  If you want I can write an explicit test that verifies that the default naming policy is 'none'.

src/GraphQL.SystemTextJson/ApolloTraceJsonConverter.cs
Shane32: > This happened because of `NamingPolicy` removal?

But also 'yes', since the default naming policy was set to camelcase in the converter, and even though it wasn't applying the camelCase conversion to `ExecutionResult` nodes, it was applying it to `extension` children.

So that's a breaking change if you consider that now extension dictionary entries will not have camelcase applied to them.  But I feel that to leave the default as camelCase is worse for two reasons:
1. It would not be unified behavior with the Newtonsoft converter
2. `ExecutionResult` nodes will now be converted to camelCase, even if another `NameConverter` was applied to the schema.

docs2/site/docs/migrations/migration4.md
Shane32: :+1: if you want to suggest a change to the docs, that'd be great.  I think my brain is fried right now, working on so much stuff at once!

We can probably just leave it though.  We are nearing the end of the 4.x lifecycle.

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
sungam3r: Please add //TODO comment here

src/GraphQL.SystemTextJson/DocumentWriter.cs
sungam3r: Already fail? Do you mean existing tests **after** your changes?

Shane32: I mean: once the naming policy is honored by the serializer (the original bug we were trying to fix) then there are a lot of tests that fail because they rely on assuming that there is no property name conversion.

is that clear?

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
Shane32: Todo what?!? This NRT annotation is obvious and won’t compile if incorrect once annotations are enabled. Further, many changes will need to be made.

src/GraphQL.SystemTextJson/ApolloTraceJsonConverter.cs
Shane32: Also, “none” naming strategy will need to be default for 5.x out of necessity because we cannot control the default serializer settings prior to the configuration delegates executing. Further, we cannot change after because there would be no way to distinguish a users desire for null versus the setting being unchanged from null.

sungam3r: OK

docs2/site/docs/migrations/migration4.md
sungam3r: #2842

src/GraphQL.NewtonsoftJson/ExecutionResultJsonConverter.cs
sungam3r: ok

src/GraphQL.SystemTextJson/DocumentWriter.cs
sungam3r: ok

sungam3r (approved):

PR #2843, opened 2022-01-06
Fix number precision for Newtonsoft.Json

sungam3r:
First commit demonstrates issue with Newtonsoft.

Inspired by changes in tests from #2841.

### DIFF
--- a/src/GraphQL.NewtonsoftJson/FixPrecisionConverter.cs
+++ b/src/GraphQL.NewtonsoftJson/FixPrecisionConverter.cs
@@ -8,6 +8,17 @@ namespace GraphQL.NewtonsoftJson
     // https://stackoverflow.com/questions/21153381/json-net-serializing-float-double-with-minimal-decimal-places-i-e-no-redundant
     public class FixPrecisionConverter : JsonConverter
     {
+        private readonly bool _decimal;
+        private readonly bool _double;
+        private readonly bool _float;
+
+        public FixPrecisionConverter(bool forDecimal, bool forDouble, bool forFloat)
+        {
+            _decimal = forDecimal;
+            _double = forDouble;
+            _float = forFloat;
+        }
+
         public override bool CanRead => false;
 
         public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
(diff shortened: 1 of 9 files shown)


Shane32:
I do not understand.  Can you explain what this PR is fixing?

sungam3r:
Newtonsoft.Json always adds `.0` when serializes float, double or decimal values. First commit shows it. I changed tests to always execute both serializers. Before changes tests used only STJ serializer.

Shane32:
> Newtonsoft.Json always adds `.0` when serializes float, double or decimal values. First commit shows it. I changed tests to always execute both serializers. Before changes tests used only STJ serializer.

And why is that an issue? It’s not wrong. There looks to be a lot of code that now has to execute for every single floating point number serialized. Seems like a complete waste of CPU power. There’s no reason why the serializers cannot perform differently so long as they meet spec.

We could add the converter to execute only during tests if we need to unify the testing. But it would be a shame not to test the default environment. 

I just don’t see that anyone would request this change. Except us, and only because of our tests that cover two serialization engines. (User code uses one serializer or the other.) So, it does not seem an appropriate change to make. 

Now personally this doesn’t affect me because I don’t use Newtonsoft.Json. But if I did I wouldn’t want this extra code added for no reason (as seen from the users perspective).

This code hurt anything except performance. So, we could add this code. But why injure performance for the sake of a few tests?

sungam3r:
> And why is that an issue?
> I just don’t see that anyone would request this change

https://github.com/JamesNK/Newtonsoft.Json/issues/1726 reasoning and reactions here

Shane32:
I read through that. It seems people have issues with large integer `decimal` values. Let’s at least restrict this converter to the `decimal` data type and not float or double. Since the default deserialization mode for numbers with a period is double, none of these issues would affect them.

src/GraphQL.NewtonsoftJson/FixPrecisionConverter.cs
```diff
@@ -0,0 +1,79 @@
...
+    public class FixPrecisionConverter : JsonConverter
+    {
+        public override bool CanRead => false;
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            throw new NotSupportedException();
+        }
+
+        public override bool CanConvert(Type objType) => objType == typeof(decimal) || objType == typeof(float) || objType == typeof(double);
```
Shane32, line 18:
```suggestion
        public override bool CanConvert(Type objType) => objType == typeof(decimal);
```

sungam3r:
Ready for review.

Shane32 (approved):
Nice

PR #2845, opened 2022-01-07
[Discussion] Demonstrate failure of .NET 6 nrt code

Shane32:
Demonstrates that the `NullabilityInfoContext` class of .NET 6 does not correctly report NRT annotations.  (My code in PR #2840 does.)

Note: this is not a compiler failure.  It is a failure of the interpretation code.  And it does not even deal with many of the complexities that can occur with NRT annotations -- it's simply a couple nested generic types.

### DIFF
--- a/src/GraphQL.Tests/NRTTests.cs
+++ b/src/GraphQL.Tests/NRTTests.cs
@@ -1,0 +1,46 @@
+#nullable enable
+#if NET6_0_OR_GREATER
+using System;
+using System.Reflection;
+using Shouldly;
+using Xunit;
+
+namespace GraphQL.Tests
+{
+    public class NRTTests
+    {
+        [Fact]
+        public void TestNRTField2()
+        {
+            var type = typeof(NullableClass18);
+            var field = type.GetMethod("Field2").ShouldNotBeNull();
+            var returnParameter = field.ReturnParameter;
+            var context = new NullabilityInfoContext();
+            var info = context.Create(returnParameter);
+
+            info.Type.ShouldBe(typeof(Tuple<Tuple<string, string>, string>));
+            info.ReadState.ShouldBe(NullabilityState.NotNull);
+            info.GenericTypeArguments.Length.ShouldBe(2);
+
+            info.GenericTypeArguments[0].Type.ShouldBe(typeof(Tuple<string, string>));
+            info.GenericTypeArguments[0].ReadState.ShouldBe(NullabilityState.NotNull);
+            info.GenericTypeArguments[0].GenericTypeArguments.Length.ShouldBe(2);
+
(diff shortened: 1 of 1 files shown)


Shane32:
This is why I don't want to rely on 3rd party code here.  Even MS can't get it right!!!!

Shane32:
The faulty code is here:

https://github.com/dotnet/runtime/blob/b1edcf125122488e1be32ffce54d909e99825d2d/src/libraries/System.Private.CoreLib/src/System/Reflection/NullabilityInfoContext.cs#L395

`GetNullabilityInfo` is a recursive call but `index` is not `ref`.  So the code only increments `offset` by 1 instead of by the proper quantity of NRT annotations within the nested type.

sungam3r:
>  Even MS can't get it right!!!!

Shame, what to say yet. So IMO the plan is:
1. Post a bug report to dotnet/runtime.
2. Create source only nullability package with all your code + tests.
3. Consume this package from GraphQL.NET + maybe from Auth/Examples/etc.

Shane32:
Ok

Shane32:
See:
- https://github.com/dotnet/runtime/pull/63556

Shane32:
@sungam3r As a secondary problem, there seems to be quite a few problems with `NullabilityInfoContext` when analyzing types with generics, such as in the following example:

```cs
class Test<T> //T may be nullable or non-nullable, so should be considered unknown
{
    T? GetStuff() { } //T is certainly nullable here - either a nullable reference type or a 
    T GetStuff2() { } //T may be nullable or non-nullable depending on the generic type, so it's really unknown
}

class Test2 : Test<string> { }
//Now Test2.GetStuff2() is non-nullable because T is string, which is non-nullable

class Test3<T> //T is non-nullable
    where T : notnull
{
    T? GetStuff() { } //T is nullable
    T GetStuff2() { } //T is not nullable
}
```

MS had written code to analyze those types of situations, but it seems the code doesn't work at all like it's supposed to.  Nearly all my tests that have to do with generics like that fail.  And since MS added the code to begin with, it won't even return the nullability flag on the generic type.  The return value is pretty useless in those situations.

I'd like to see the tests for the code they wrote.  It's a wonder they pass...

sungam3r:
👍

Shane32:
@sungam3r MS merged my pull request.  I issued a pull request to SimonCropp's source-only nuget repo for the same fix (which will match the code currently in MS's repo).  A number of other issues have been raised in MS's repo for problems with NullabilityInfoContext, and I also found an issue with the compiler not adding the proper attributes.  But this fix should take care of what we need.

I just realized that even if SimonCropp updates his nuget package, we still can't use SimonCropp's nuget package if it is only injected on pre-.NET 6 platforms, as there will be bugs on .NET 6.  Ugh.

Shane32:
Ref:
- https://github.com/SimonCropp/NullabilityInfo/pull/7

PR #2846, opened 2022-01-07
User context docs update

decembrya:


docs2/site/docs/getting-started/user-context.md
```diff
@@ -28,3 +28,20 @@ public class Query : ObjectGraphType
   }
 }
 ```
+
+If you need to access the User from the http request and populate this globally, you will need to add a User property to your custom `UserContext` class and modify your Startup configuration as follows:
```
sungam3r, line 32: > populate this globally

What does _globally_ mean?

sungam3r, line 32: > you will need to add a User property

or just
```
var context = new MyGraphQLUserContext(); // or even Dictionary<string, object?>();
context["User"] = user;
```

sungam3r: So I would change wording from _will need to_ to _may_.

docs2/site/docs/getting-started/user-context.md
```diff
@@ -28,3 +28,20 @@ public class Query : ObjectGraphType
...
+      public ClaimsPrincipal User { get; set; }
+
+      public MyGraphQLUserContext(ClaimsPrincipal user)
+      {
+          User = user;
+      }
+  }
+
+  services.AddGraphQL()
+          .AddUserContextBuilder(httpContext => new MyGraphQLUserContext(httpContext.User));
```
sungam3r, line 46: There is no such method in this repo. It's method from server project. Worth to note explicitly.

docs2/site/docs/getting-started/user-context.md
decembrya: Thank you for catching this, adjusted the wording and added the other changes suggested.

sungam3r (approved):

sungam3r:
@Shane32 Merge if OK.

PR #2847, opened 2022-01-07
Don't expose the disposed ServiceProvider instance

ganhammar:
The `IServiceProvider` instance in the `ResolveEventStreamContext` has been disposed at the point where it gets exposed to the subscription resolver and any data loaded fields included in the GraphQL request. With this PR it is set to `null` instead of the disposed instance. See discussion [here](https://github.com/graphql-dotnet/server/issues/706).

I don't know if you would like to see tests for this change?


ganhammar:
@Shane32 I think this is the change you suggested?

I don't know if the issue should be moved from the server project and be linked with this PR?

src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
```diff
@@ -101,6 +101,9 @@ protected virtual async Task<IObservable<ExecutionResult>> ResolveEventStreamAsy
                     throw new InvalidOperationException($"Subscriber not set for field '{node.Field.Name}'.");
                 }
 
+                // The IServiceProvider instance will be disposed at this point, set it to null insted of exposing the disposed object
```
sungam3r, line 104:
```suggestion
                // The IServiceProvider instance will be disposed at this point, set it to null instead of exposing the disposed object
```

sungam3r:
Rel: https://github.com/graphql-dotnet/server/issues/706

Shane32 (approved):

Shane32:
@ganhammar yes thanks.  It looks like the branch needs to be merged with the current 'master' branch so that we can let it run the tests and merge it.

ganhammar:
> @ganhammar yes thanks. It looks like the branch needs to be merged with the current 'master' branch so that we can let it run the tests and merge it.

Yeah, sorry, I missed that I already had forked the repo a while back, so didn't pull latest changes. Fixed now!

Shane32:
@sungam3r look good?

sungam3r (approved):

PR #2859, opened 2022-01-13
Configure serializers so they can be easily extended if need be

Shane32:
For Newtonsoft.Json, the method that builds the serializer is static and can't be made virtual.  So I exposed the constructor with the serializer as protected, so derived classes can build their own serializer.

For System.Text.Json, derived classes can place additional configuration within the derived constructor of the class.

### DIFF
--- a/src/GraphQL.NewtonsoftJson/GraphQLSerializer.cs
+++ b/src/GraphQL.NewtonsoftJson/GraphQLSerializer.cs
@@ -5,6 +5,8 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using GraphQL.Execution;
+using GraphQL.Instrumentation;
+using GraphQL.Transport;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
@@ -17,7 +19,7 @@ namespace GraphQL.NewtonsoftJson
     public class GraphQLSerializer : IGraphQLTextSerializer
     {
         private readonly JsonArrayPool _jsonArrayPool = new JsonArrayPool(ArrayPool<char>.Shared);
-        private readonly JsonSerializer _serializer;
+        protected JsonSerializer Serializer { get; }
         private static readonly Encoding _utf8Encoding = new UTF8Encoding(false);
 
         /// <summary>
@@ -104,9 +106,14 @@ namespace GraphQL.NewtonsoftJson
         {
         }
 
-        private GraphQLSerializer(JsonSerializer jsonSerializer)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="GraphQLSerializer"/> class with the specified <see cref="JsonSerializer"/>.
+        /// The specified <see cref="JsonSerializer"/> should support serializing and/or deserializing <see cref="ExecutionResult"/>,
+        /// <see cref="GraphQLRequest"/>, <see cref="Inputs"/>, <see cref="OperationMessage"/> and <see cref="ApolloTrace"/>.
+        /// </summary>
+        protected GraphQLSerializer(JsonSerializer jsonSerializer)
         {
-            _serializer = jsonSerializer;
+            Serializer = jsonSerializer;
         }
 
         private static JsonSerializerSettings GetDefaultSerializerSettings(bool indent, IErrorInfoProvider errorInfoProvider)
@@ -150,7 +157,7 @@ namespace GraphQL.NewtonsoftJson
                 AutoCompleteOnClose = false
             };
 
-            _serializer.Serialize(jsonWriter, value);
+            Serializer.Serialize(jsonWriter, value);
             await jsonWriter.FlushAsync(cancellationToken).ConfigureAwait(false);
         }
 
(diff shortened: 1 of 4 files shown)


src/GraphQL.NewtonsoftJson/GraphQLSerializer.cs
```diff
@@ -17,7 +19,7 @@ namespace GraphQL.NewtonsoftJson
     public class GraphQLSerializer : IGraphQLTextSerializer
     {
         private readonly JsonArrayPool _jsonArrayPool = new JsonArrayPool(ArrayPool<char>.Shared);
-        private readonly JsonSerializer _serializer;
+        protected JsonSerializer Serializer { get; }
```
sungam3r, line 22: Requires xml comment?

src/GraphQL.SystemTextJson/GraphQLSerializer.cs
```diff
@@ -17,7 +17,7 @@ namespace GraphQL.SystemTextJson
     /// </summary>
     public class GraphQLSerializer : IGraphQLTextSerializer
     {
-        private readonly JsonSerializerOptions _options;
+        protected JsonSerializerOptions SerializerOptions { get; }
```
sungam3r, line 20: xml comment

src/GraphQL.SystemTextJson/GraphQLSerializer.cs
```diff
@@ -81,7 +81,7 @@ public GraphQLSerializer(Action<JsonSerializerOptions> configureSerializerOption
         /// <param name="serializerOptions">Specifies the JSON serializer settings</param>
         public GraphQLSerializer(JsonSerializerOptions serializerOptions)
         {
-            _options = serializerOptions ?? throw new ArgumentNullException(nameof(serializerOptions));
+            SerializerOptions = serializerOptions ?? throw new ArgumentNullException(nameof(serializerOptions));
 
             // TODO: fix this: it modifies serializerOptions
```
sungam3r, line 86: @Shane32 Maybe it's time to fix TODO ?

For NSJ you wrote
изображение

So maybe to write the same here and remove `ConfigureOptions(null);` ? + Add a few words in migration notes.

sungam3r: \+ remove ` public GraphQLSerializer(JsonSerializerOptions serializerOptions, IErrorInfoProvider errorInfoProvider)`

Shane32: Well, we can't do that.  It is required by DI.  Besides, it has no bearing on this PR.

sungam3r: **If** we don't want to modify passed serializer (or anything else passed from outside - DI) then we remove `ConfigureOptions` call and then `errorInfoProvider` argument become unused.

Shane32: I have no intention of changing the behavior in that respect.  I will be happy to remove the "todo" comments if you like.

src/GraphQL.NewtonsoftJson/GraphQLSerializer.cs
Shane32: done

src/GraphQL.SystemTextJson/GraphQLSerializer.cs
Shane32: done

src/GraphQL.SystemTextJson/GraphQLSerializer.cs
sungam3r: Serializer for NSJ is actualy the same API part as JsonSerializerOptions for STJ. So why we don't allow to modify first and allow to modify second?

sungam3r: I would not modify the passed object (serializer or settings) for both cases.

sungam3r: >  I will be happy to remove the "todo" comments if you like.

These TODO are necessary. I remember that the modification of these settings caused the error. Someone posted an issue.

sungam3r (approved):

src/GraphQL.SystemTextJson/GraphQLSerializer.cs
Shane32: If you can explain how this will work with the `IOptions` DI pattern, I will consider your suggestions, but only in another PR.  I do not see any way this will work with the existing `IOptions` DI pattern.  If you want to remove support of the `IOptions` pattern, you can suggest it in another PR.

I agree that the constructor _should_ not alter incoming options classes.  I do not see a easy way to fix the problem.

Shane32: Keep in mind this "issue" applies equally to STJ and NJ.

sungam3r: > If you want to remove support of the IOptions pattern

No. Why?

sungam3r: #2865

PR #2872, opened 2022-01-22
Refactor existing classes to separate files

Shane32:
No code changes.


sungam3r (approved):

Shane32:
Apparently the VS refactor wizard doesn't know to use the same text encoding!

sungam3r:
😕

PR #2873, opened 2022-01-22
Only register writable properties for auto input types

Shane32:

PR #2874, opened 2022-01-22
Only register readable properties for auto output graph types

Shane32:
This PR builds on #2873


src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            public string Name { get; set; }
+        }
+
+        private class TestClass
+        {
+            public int Field1 { get; set; }
+            public int Field2 { get; }
+            public int Field3 { set { } }
+            public int Field5() => 0;
```
sungam3r, line 47: again 4/5

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -10,15 +10,17 @@ namespace GraphQL.Types
 {
     internal static class AutoRegisteringHelper
     {
-        internal static void SetFields<TSourceType>(ComplexGraphType<TSourceType> type, IEnumerable<PropertyInfo> properties, params Expression<Func<TSourceType, object?>>[]? excludedProperties)
+        internal static IEnumerable<PropertyInfo> ExcludeProperties<TSourceType>(IEnumerable<PropertyInfo> properties, params Expression<Func<TSourceType, object?>>[]? excludedProperties)
```
sungam3r, line 13: Why do you need to split code into two methods?

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
```diff
@@ -17,6 +17,8 @@ namespace GraphQL.Types
     /// <typeparam name="TSourceType"></typeparam>
     public class AutoRegisteringInputObjectGraphType<TSourceType> : InputObjectGraphType<TSourceType>
     {
+        private readonly Expression<Func<TSourceType, object?>>[]? _excludedProperties;
```
sungam3r, line 20: Unused after `GetRegisteredProperties` call.

sungam3r: That is after ctor called.

Shane32: Correct.  No way around this without a poorer OOP design.

sungam3r: Probably because of this I made single method.

sungam3r:
```csharp
public AutoRegisteringInputObjectGraphType(params Expression<Func<TSourceType, object?>>[]? excludedProperties)
        {
            _excludedProperties = excludedProperties;
            AutoRegisteringHelper.SetFields(this, GetRegisteredProperties());
           _excludedProperties  = null;
        }
```

Shane32: make sense.

Shane32: well, then constructors of derived classes would not return the same data for `GetRegisteredProperties`.  I would leave it as is.

Shane32: Holding a few initialization properties in memory should not matter.

sungam3r: Why should they be kept if you can pass them into the method? If someone needs to keep them, then it will save them in the ctor of derived class, but in general `GetRegisteredProperties` method is called once.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: Read the next PR that contains migration doc changes.

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
Shane32: This special constructor as a whole is a patch. It should not change a good design to be specific for this use case.

Shane32: And, the design needs to be considered one more time.  A derived class's constructor cannot be passed arguments that change the behavior of methods in the class.  Probably nothing we can do about it, since there is no `Initialize` call to an `IGraphType` before it is used.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
sungam3r: OK

sungam3r (approved):

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
Shane32: I thought about this some more and will push another change.

Shane32:
I moved the call to `ExcludeProperties` into the constructor.  As such, there is no need to store the enumeration.  Does that look good?

This behavior is actually more similar to v4 behavior anyway.  Not that either design would matter for derived classes.

Shane32:
All it did is break the next PR. :-(

Shane32:
Merging as approved.

sungam3r:
ok

Issue #2876, opened 2022-01-23
[Discussion] Review design of GraphQLMetadataAttribute


Shane32:
Currently the attribute is used in several different ways, listed below.  It would seem this is a poor design, because it is not obvious which properties are available for the specific use case.  I think this should be cleaned up for v5, possibly with backwards compatibility via an obsolete attribute.

First, let's review the current situation:

## Code first supported scenarios

1. Applied to a graph type class

```cs
public class Person
{
}

[GraphQLMetadata("Person")]
public class PersonType : ObjectGraphType<Person>
{
   //...
}
```

Notes;
- Does not support any property besides `Name`
- Name can be set in the constructor just as easily
- Only one test that uses this attribute in this manner, and only to test an invalid name
- If supported, it should support `Description` and `DeprecationReason` also.  As of now, these properties are pulled from the CLR type's `Description` and `Obsolete` attributes (not attributes on the graph type itself).

2. Applied to a CLR type

```cs
internal class TestType : ObjectGraphType<TestTypeData>
{
    public TestType()
    {
        Field(x => x.Example, true);
    }
}

internal class TestTypeData
{
    public Test3 Example => null!;
}

[GraphQLMetadata(OutputType = typeof(Test3Type))]
internal class Test3
{
    public string Test { get; set; }
}

internal class Test3Type : ObjectGraphType<Test3>
{
    public Test3Type()
    {
        Field(x => x.Test);
    }
}
```

Notes:
- Only supports `InputType` and `OutputType`
- Somewhat unnecessary with `schema.RegisterTypeMappings()`
- May be present in existing code bases
- Does not support any other `[GraphQLMetadata]` properties such as `Description` and `DeprecationReason`
- The `[Description]` and `[Obsolete]` attributes are recognized from the CLR type, however.

## Schema first supported scenarios

1. Applied to a CLR type

```cs
var schema = Schema.For(definitions, b => b.Types.Include<TestType>());

[GraphQLMetadata("MyTestType", Description = "Test1", DeprecationReason = "Test2")]
internal class TestType
{
    public string Example() => "value";
}
```

Notes:
- Definitions must have the type defined already.  The `Name` property does not set the name but rather matches it up to the name in the definitions
- `Description` and `DeprecationReason` supported; although these could probably be supported within the definitions also
- `IsTypeOf` used for interfaces support
- `ResolverType` used to set if it is a regular or subscription field, or something like that
[...]

PR #2877, opened 2022-01-23
Remove support for GraphQLMetadata attribute to set graph type name

Shane32:
See #2876 

Suggest to remove support for this attribute on graph type classes in code-first scenarios since the attribute was designed for schema-first and is not fully supported for code-first in any case.  There is no good reason for this attribute, because if a user has derived from `ObjectGraphType` in order to add an attribute, then they certainly can add code to the constructor.

Only possibly way I could see someone wishing to use the attribute today would be as follows:

```cs
[GraphQLMetadata("Person")]
public class PersonType : AutoRegisteringObjectGraphType<Human> { }
```

Even if such a scenario was desired, it should have supported all relevant properties.  In any case, in a subsequent PR, I will be adding support for the attribute to be applied to the CLR type as in the following example:

```cs
//note that defining the class is now optional because the name can be applied to the CLR type
public class PersonType : AutoRegisteringObjectGraphType<Human> { }

[GraphQLMetadata("Person")]
public class Human { ... }
```

This will align the use case of the attribute with that of schema-first, which is that it is designed to be set on a CLR type.

In the cases where a user cannot set attributes on their data classes, they can still resort to setting the GraphQL type name within the constructor:

```cs
public class PersonType : AutoRegisteringObjectGraphType<Human>
{
    public PersonType()
    {
        Name = "Person";
    }
}
```

### DIFF
--- a/src/GraphQL.Tests/Types/ObjectGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ObjectGraphTypeTests.cs
@@ -9,8 +9,16 @@ namespace GraphQL.Tests.Types
     {
         private class TestInterface : InterfaceGraphType { }
 
-        [GraphQLMetadata(Name = ":::")]
-        private class TypeWithInvalidName : ObjectGraphType { }
+        private class TypeWithInvalidName : ObjectGraphType
+        {
+            public TypeWithInvalidName()
+            {
+                Name = ":::";
+            }
+        }
+
+        [GraphQLMetadata(Name = "testing")]
+        private class TypeWithAttribute : ObjectGraphType { }
 
         [Fact]
         public void can_implement_interfaces()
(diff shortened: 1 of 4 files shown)


src/GraphQL.Tests/Types/ObjectGraphTypeTests.cs
```diff
@@ -9,8 +9,16 @@ public class ObjectGraphTypeTests
     {
         private class TestInterface : InterfaceGraphType { }
 
-        [GraphQLMetadata(Name = ":::")]
-        private class TypeWithInvalidName : ObjectGraphType { }
+        private class TypeWithInvalidName : ObjectGraphType
```
Shane32, line 12: The only test in the entire project that utilized `[GraphQLMetadata]` on a graphtype class, which was only testing that invalid names cannot be set on a graph type.

src/GraphQL/Types/GraphType.cs
```diff
@@ -33,13 +33,6 @@ private string GetDefaultName()
         {
             var type = GetType();
 
-            var attr = type.GetCustomAttribute<GraphQLMetadataAttribute>();
-
-            if (!string.IsNullOrEmpty(attr?.Name))
-            {
-                return attr!.Name!;
-            }
```
Shane32, line 41: No other properties - such as `Description` or `DeprecationReason` - were ever supported.

src/GraphQL.Tests/Types/ObjectGraphTypeTests.cs
sungam3r: 👍

src/GraphQL/Types/GraphType.cs
sungam3r: 👍

sungam3r (approved):

PR #2882, opened 2022-01-26
Add Ignore attribute

Shane32:
Adds `[Ignore]` attribute so properties on data models can be easily skipped when auto-registering graph types:

```cs
public class ShoppingCartItem
{
    [OutputType(typeof(NonNullGraphType<IdGraphType>))]
    public int Id { get; set; }

    public int Quantity { get; set; }

    public decimal Price { get; set; }

    [Ignore]
    public decimal Total => Quantity * Price;
}
```

This new attribute does not inherit from `GraphQLAttribute`.  We could add another protected method to the `GraphQLAttribute` class -- something like `bool Ignore(PropertyInfo property, bool isInputType)` -- and then rewrite `IgnoreAttribute` to inherit `GraphQLAttribute`, but so far I don't see a good reason to do so.

### DIFF
--- a/src/GraphQL/IgnoreAttribute.cs
+++ b/src/GraphQL/IgnoreAttribute.cs
@@ -1,0 +1,12 @@
+using System;
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Does not add the marked property to the auto-registered GraphQL type as a field.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    public class IgnoreAttribute : Attribute
+    {
+    }
+}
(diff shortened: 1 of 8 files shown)


src/GraphQL/IgnoreAttribute.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Does not add the marked property to the auto-registered GraphQL type as a field.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
```
sungam3r, line 8: What about inherited? True by default?

sungam3r (approved):

src/GraphQL/IgnoreAttribute.cs
Shane32: I don't know.

PR #2863, opened 2022-01-16
More parser integration 

sungam3r:
I accidentally included several extra commits, but they contains only documentation changes.

~50 files from Language namespace were deleted. They are useless now.

Many changed like:
1) `string` -> `ROM`
2) `BlaBlaType` -> `GraphQLBlaBlaType`
3) `INode`/`AbstractNode` -> `ASTNode`
4) `IValue` -> `GraphQLValue`

Refactored: `ASTPrinter`, `CoreToVanillaConverter`, `ComplexityAnalyzer`

TODO:
1) remove/rewrite `SchemaPrinter` [in separate PR]

PR #2883, opened 2022-01-26
Tune validation rules

sungam3r:
Further optimizations after parser v8:
изображение
Ref from #2863:
изображение

### DIFF
--- a/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
@@ -2650,11 +2650,12 @@ namespace GraphQL.Validation
         public System.Collections.Generic.IDictionary<string, object?> UserContext { get; set; }
         public GraphQL.Inputs Variables { get; set; }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentSpread> GetFragmentSpreads(GraphQLParser.AST.GraphQLSelectionSet node) { }
-        public System.Collections.Generic.List<GraphQL.Validation.VariableUsage> GetRecursiveVariables(GraphQLParser.AST.GraphQLOperationDefinition operation) { }
+        public System.Collections.Generic.List<GraphQL.Validation.VariableUsage>? GetRecursiveVariables(GraphQLParser.AST.GraphQLOperationDefinition operation) { }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentDefinition>? GetRecursivelyReferencedFragments(GraphQLParser.AST.GraphQLOperationDefinition operation) { }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentDefinition>? GetRecursivelyReferencedFragments(System.Collections.Generic.List<GraphQLParser.AST.GraphQLOperationDefinition> operations) { }
         public GraphQL.Validation.Variables GetVariableValues(GraphQL.Validation.IVariableVisitor? visitor = null) { }
-        public System.Collections.Generic.List<GraphQL.Validation.VariableUsage> GetVariables(GraphQLParser.AST.IHasSelectionSetNode node) { }
+        public System.Collections.Generic.List<GraphQL.Validation.VariableUsage>? GetVariables<TNode>(TNode node)
+            where TNode : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasSelectionSetNode { }
         public string? IsValidLiteralValue(GraphQL.Types.IGraphType type, GraphQLParser.AST.GraphQLValue valueAst) { }
         public void ReportError(GraphQL.Validation.ValidationError error) { }
     }
@@ -2668,7 +2669,7 @@ namespace GraphQL.Validation
         public System.Collections.Generic.IEnumerable<GraphQLParser.AST.ASTNode> Nodes { get; }
         public string Number { get; set; }
     }
-    public class ValidationOptions
+    public readonly struct ValidationOptions
     {
         public ValidationOptions() { }
         public System.Threading.CancellationToken CancellationToken { get; set; }
--- a/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
@@ -2649,10 +2649,10 @@ namespace GraphQL.Validation
         public GraphQL.Validation.TypeInfo TypeInfo { get; set; }
         public System.Collections.Generic.IDictionary<string, object?> UserContext { get; set; }
         public GraphQL.Inputs Variables { get; set; }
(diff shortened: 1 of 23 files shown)


sungam3r:
I think that after work on memory optimizations it will be possible to deal with execution time.

Shane32 (approved):

sungam3r:
For info:
изображение

sungam3r:
изображение

sungam3r:
@Shane32 The idea of this PR is to move to the new `ASTVisitor` API from parser that allow to use static stateless visitor as well as passing context as struct into.

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -80,19 +80,36 @@ public async Task<(IValidationResult validationResult, Variables variables)> Val
                         context.TypeInfo
                     };
 
-                    foreach (var rule in rules)
+                    if (rules is List<IValidationRule> list) //TODO:allocation - optimization for List+Enumerator<IvalidationRule>
```
sungam3r, line 83: No code changes here - only duplicated branch for List iterator, in future PRs may be removed since design of `BasicVisitor` may be changed.

src/GraphQL/Validation/Rules/NoUnusedFragments.cs
```diff
@@ -11,18 +10,18 @@ namespace GraphQL.Validation.Rules
     /// No unused fragments:
     ///
     /// A GraphQL document is only valid if all fragment definitions are spread
-    /// within operations, or spread within other fragments spread within operations.
+    /// within operations, or spread within other fragment spreads within operations.
```
sungam3r, line 13: @Shane32 Did I write it correctly?

sungam3r:
I'm done here. I realized that all validation rules can be rewritten so as to use the benefits of a new parser design. Someday we can do it but not now. Possible changes:
- remove `BasicVisitor`
- remove `MatchingNodeVisitor`
- remove `NodeVisitors`
- write custom private `ASTVisitor` for each validation rule
- rewrite `TypeInfo` in some way to save memory and iterations over AST

src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
```diff
@@ -2668,7 +2669,7 @@ namespace GraphQL.Validation
         public System.Collections.Generic.IEnumerable<GraphQLParser.AST.ASTNode> Nodes { get; }
         public string Number { get; set; }
     }
-    public class ValidationOptions
+    public readonly struct ValidationOptions
```
Shane32, line 2672: Why changing to struct?  Why readonly struct?  Why setters if readonly struct? MS guidelines state to avoid a struct unless it contains < 16 bytes, and this struct is 144 bytes.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct

src/GraphQL/Validation/DocumentValidator.cs
Shane32: I think this should be eliminated.  There is no need to eliminate one allocation per GraphQL request.  It adds additional complexity with no benefit.

src/GraphQL/Validation/ValidationContext.GetVariables.cs
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+
+            _variables[operation] = usages;
+
+            return usages;
+        }
+
+        // struct intentionally - works only on stack
+        private readonly struct GetVariablesVisitorContext : IASTVisitorContext
```
Shane32, line 61: Good use of struct here

src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
Shane32: In the code I see you're using init-only properties.  Are you sure this works properly with older compilers that do not recognize the init-only attributes?  Since the struct is readonly and has no constructors, it seems like it could be a major problem.

sungam3r: > Why changing to struct?

To save allocation for such transitive object as options. 

> Why readonly struct? 

To be able to use `in` modifier to pass struct by reference without guard copy. Fixed in latest commit. I just forgot to use `in`.

> Why setters if readonly struct?

Because `init` != `set`.

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: I think about it in further PRs - either remove branch or completely rewrite visitors here. Now it helped me to reduce the number of objects in the memory consumption report.

Shane32: Add an issue then.  We should not be adding unnecessarily complex code for the sake of the "memory consumption report"

sungam3r: OK. Every allocation (specifically well-know one as for List iterator that can be fixed easily) on green path matters for me. #2886

src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
sungam3r: > Are you sure this works properly with older compilers that do not recognize the init-only attributes?

I'm 100% sure that this feature does not work for older compilers. Of course.

>  it seems like it could be a major problem.

Whose problem?

Shane32: We target .NET Standard 2.0 so people can use whatever type of environment they wish. Mono or VS, webassembly, .NET Framework, .NET Core, C#, F# whatever. We shouldn’t use features that require the newest software if we are targeting old frameworks.

I actually think I remember that init-only is fine with older compilers — but perhaps not when the struct is marked as read only. There is a document on it from MS but I’m not sure where it is at the moment.

Shane32: I have multiple active projects today that run on old or outdated frameworks. One project I was working on yesterday was a website in Classic ASP. We don’t support that, but we should make a conscience decision if we are breaking support for older frameworks or compilers, and then list it in the migration notes. 

I think it would be a shame to drop support for older tool sets or frameworks for the sake of one allocation per request. Especially when doing something as simple as adding a constructor with all the parameters can satisfy the needs of such a toolset.

sungam3r: Old runtime != old compiler. `init` property is a compiler feature. Old runtime **can** use classes compiled with `init` properties, it just a `set` property. Regarding `readonly struct` - i'm not 100% sure what happens. Well, `readonly` is a compiler feature as well so I think old compilers can just set any value of "readonly struct". Try and check if you work with outdated frameworks.

PR #2884, opened 2022-01-27
Auto registering graph types respect NRT annotations

Shane32:
Also see:
- https://github.com/SimonCropp/NullabilityInfo/issues/8
- https://github.com/dotnet/runtime/pull/64143


Shane32:
@sungam3r how to fix tests?

Shane32:
No, the source-only repo contains exact duplicates of the .NET source, so it is easy to update when the .NET source does. Besides, others’ formatting requirements are not the same as this repo. My projects are formatted much differently. There’s no good reason why the other repo should change to match this formatting style.

Shane32:
> How did you compile this code?

I just added the nuget package and compiled and ran tests. Everything worked perfectly. Tests pass on my machine for all frameworks.

Shane32:
> Hmm.. It isn't compiling now. Maybe I changed something....

I updated the nuget package to 1.0.5, and the new code in the source-only package isn't working.

Shane32:
See:
- https://github.com/SimonCropp/NullabilityInfo/issues/8
- https://github.com/SimonCropp/NullabilityInfo/issues/9

Shane32:
All tests pass now.  We can bump to the next version of NullabilityInfo when it's ready.  There's one notable flaw that was fixed, but mostly it was edge cases that we won't run into.

Shane32:
The formatter is disabled until/unless you can determine how to exclude source-only packages from formatting inspection @sungam3r 

We can probably replace most of its functionality with a proper editorconfig anyway.  That's what I do in my repos.  It's also nice because it tells you in the editor when there's a problem, rather than forcing you to push the code and wait for CI to run.

sungam3r:
dotnet-format respects .editorconfig, doesn't it?

Shane32:
> dotnet-format respects .editorconfig, doesn't it?

I don't know.  I don't use `dotnet-format`.  If `dotnet-format` just uses the editorconfig file, then we should just change the error levels from warning to error and delete the formatter.

sungam3r:
> how to fix tests?

Something like
`dotnet-format --no-restore --verify-no-changes --severity warn whitespace --exclude GraphQL\obj\**\*.*`

sungam3r:
изображение

Shane32:
Exactly.  So what's the point?  Just set the level from warning to error within editorconfig and call it a day.

Shane32:
If you have no comments here @sungam3r, I'll merge this and create an issue to review the editorconfig/formatter

sungam3r:
>  If dotnet-format just uses the editorconfig file, then we should just change the error levels from warning to error and delete the formatter.

.editorconfig can not fail CI, it works only for IDE and dotnet-format forces CI to fail applying editorconfig rules (and so own logic).

sungam3r:
I need to review. I did not look at the code yet, just understood format/build errors.

sungam3r:
Also I'm interested in that
> dotnet-format --no-restore --verify-no-changes --severity warn whitespace --exclude GraphQL\obj\**\*.*

to make formatter work for files from the new package.

I'll rewiew today or tomorrow.

Shane32:
I’ll be on the road for the next 2-3 days so go ahead and try pushing changes to the GitHub Action workflows as desired, or else we can fix the formatted in another PR.

Shane32:
I still don’t understand why the editorconfig has spacing and stuff set to warning if it won’t pass build checks. If they are warnings they should be warnings, and build checks should pass. If they are errors then they should be errors.

sungam3r:
Also this https://github.com/dotnet/roslyn/issues/33558 (https://github.com/dotnet/roslyn/issues/33558#issuecomment-690347050) and #2778

sungam3r:
> I still don’t understand why the editorconfig has spacing and stuff set to warning if it won’t pass build checks. If they are warnings they should be warnings, and build checks should pass. If they are errors then they should be errors.

You are right. The reason is the imperfect tools design. I watched when MS add support to what you are talking about. I tried in #2778, but it did not work out. I don't know why. Maybe I'll try one time later.

Shane32:
NullabilityInfo version 1.0.6 also is broken due to it relying on global using statements.
- https://github.com/SimonCropp/NullabilityInfo/pull/10

sungam3r:
I see you actively change the code. Now the formatter and the rest of the code work as it should and PR is ready for review?

Shane32:
Yes I am done with this PR.

sungam3r:
I'll review in 1-2 hours if I have time, otherwise tomorrow.

.editorconfig
```diff
@@ -255,4 +255,7 @@ dotnet_naming_style.end_in_async_style.required_suffix = Async
 dotnet_naming_rule.async_methods_end_in_async.severity = warning
 
 # Remove unnecessary import https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0005
-dotnet_diagnostic.IDE0005.severity = warning
+dotnet_diagnostic.IDE0005.severity = error
+
+# Enforce formatting
```
sungam3r, line 260:
```suggestion
# Enforce formatting https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#rule-id-ide0055-fix-formatting
```

.editorconfig
```diff
@@ -255,4 +255,7 @@ dotnet_naming_style.end_in_async_style.required_suffix = Async
 dotnet_naming_rule.async_methods_end_in_async.severity = warning
 
 # Remove unnecessary import https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0005
-dotnet_diagnostic.IDE0005.severity = warning
+dotnet_diagnostic.IDE0005.severity = error
```
sungam3r, line 258: 👍

.github/workflows/test-code.yml
```diff
@@ -52,9 +52,9 @@ jobs:
         working-directory: src
         run: |
           dotnet tool install -g dotnet-format --version 6.0.243104 --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
-          dotnet format --no-restore --verify-no-changes --severity warn whitespace || (echo "Run 'dotnet format' to fix whitespace issues" && exit 1)
-          dotnet format --no-restore --verify-no-changes --severity warn analyzers || (echo "Run 'dotnet format' to fix analyzers issues" && exit 1)
-          dotnet format --no-restore --verify-no-changes --severity warn style || (echo "Run 'dotnet format' to fix style issues" && exit 1)
+          dotnet format --no-restore --verify-no-changes --severity warn whitespace --exclude ./GraphQL/obj/**/*.cs || (echo "Run 'dotnet format' to fix whitespace issues" && exit 1)
```
sungam3r, line 55: I'm glad that we found the way to get it work.

src/GraphQL/TypeExtensions.cs
```diff
@@ -41,6 +41,30 @@ public static bool IsConcrete(this Type type)
         public static bool IsGraphType(this Type type)
             => typeof(IGraphType).IsAssignableFrom(type);
 
+        /// <summary>
+        /// Determines if the specified type represents a named graph type (not a wrapper type such as <see cref="ListGraphType"/>).
+        /// </summary>
+        internal static bool IsNamedGraphType(this Type type)
```
sungam3r, line 47: See `GraphQLExtensions.GetNamedType(this Type type)` and whole `GraphQLExtensions` class. I propose to merge two classes into one. I see no reason to keep 2 classes for methods that both work with graph types (`GraphQLType` from parser, `IGraphType` or just `Type`).

sungam3r: Also I propose to rename `IsNamedGraphType` to `IsNamedType` or rename `GetNamedType` to `GetNamedGraphType`.

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,258 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace GraphQL.Types
+{
+    /// <summary>
+    /// Contains type and nullability information for a method return type or argument type.
```
sungam3r, line 9: > for a method return type or argument type

I see only usages with `PropertyInfo`.

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,258 @@
...
+                }
+                else
+                {
+                    list.Add((info.Type, isInputProperty ? info.ReadState : info.WriteState));
+                }
+            }
+        }
+
+        /// <summary>
+        /// Applies <see cref="GraphQLAttribute"/> attributes for the specified member to this instance.
```
sungam3r, line 188: Not only `GraphQLAttribute`, `RequiredAttribute` too. I am not quite understand the responsibility of this class. Is it designed to handle **only** nullability information? If so why `IsList` here. If no then maybe to absorb other attributes into to handle all of them in one place?
изображение

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -75,5 +79,20 @@ protected virtual IEnumerable<PropertyInfo> GetRegisteredProperties()
...
+        /// containing information necessary to select a graph type. Any <see cref="GraphQLAttribute"/>
+        /// attributes marked on the property are applied.
+        /// <br/><br/>
+        /// Override this method to enforce specific graph types for specific CLR types, or to implement custom
+        /// attributes to change graph type selection behavior.
+        /// </summary>
+        protected virtual TypeInformation GetTypeInformation(PropertyInfo propertyInfo)
+        {
+            var typeInformation = new TypeInformation(propertyInfo, false);
+            typeInformation.ApplyAttributes();
```
sungam3r, line 94: Why do you need a separate (and public) method? Maybe call it from `TypeInformation` ctor?

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        /// <summary>
+        /// Returns a graph type constructed based on the properties set within this instance.
+        /// If <see cref="GraphType"/> is <see langword="null"/>, the graph type is generated via
+        /// <see cref="GraphQL.TypeExtensions.GetGraphTypeFromType(Type, bool, TypeMappingMode)"/>.
+        /// The graph type is then wrapped with <see cref="NonNullGraphType{T}"/> and/or
+        /// <see cref="ListGraphType{T}"/> as appropriate.
+        /// </summary>
+        public virtual Type ConstructGraphType()
```
sungam3r, line 218: `virtual` here and for `ApplyAttributes`?

Shane32: Yes. I did this to simplify the number of virtual members in AutoRegisteringObjectGraphType. I was going to make this class sealed but after further review it seemed most logical to make them virtual.

Note than in my current code there are a very great number of virtual members. They are all useful, even within my own projects, for different purposes. Now some of the burden is shifted here so if users want to fully customize behavior they might need to derive from this class. 

At the same time that the api is simpler to understand, it is still reasonably easy to replace one piece of functionality without replacing large portions of the codebase.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Designed so code can be called from AutoRegisteringObjectGraphType between the constructor and applying the attributes. This will be obviously necessary with my next pull request (unless I split it to the one after next).

src/GraphQL/Types/TypeInformation.cs
Shane32: Prep for the next pull request, which is mostly done now. I tried to eliminate references in here but must have missed this one.

src/GraphQL/TypeExtensions.cs
Shane32: I’m okay shifting things around a bit. I do like having them separate in the codebase so if we do merge them then I would keep them as separate files defined as partial classes.

Shane32: IsNamedGraphType: just let me know what you prefer. I had IsNamedType but as an extension on Type I figured it would be better if it was named something more specific. And it would better match IsGraphType. 

Unless you have a comment, if GetNamedType isn’t public, I’ll rename it, otherwise I’ll rename this method as you suggested.

sungam3r: OK for IsNamedType

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
```diff
@@ -160,10 +161,33 @@ public void Field_IgnoresOutputNameAttribute()
...
+        [InlineData(nameof(FieldTests.NotNullArrayNullableTupleField), typeof(NonNullGraphType<ListGraphType<GraphQLClrInputTypeReference<Tuple<int, string>>>>))]
+        [InlineData(nameof(FieldTests.NotNullArrayNotNullTupleField), typeof(NonNullGraphType<ListGraphType<NonNullGraphType<GraphQLClrInputTypeReference<Tuple<int, string>>>>>))]
+        [InlineData(nameof(FieldTests.NullableArrayNullableTupleField), typeof(ListGraphType<GraphQLClrInputTypeReference<Tuple<int, string>>>))]
+        [InlineData(nameof(FieldTests.NullableArrayNotNullTupleField), typeof(ListGraphType<NonNullGraphType<GraphQLClrInputTypeReference<Tuple<int, string>>>>))]
+        [InlineData(nameof(FieldTests.IdField), typeof(NonNullGraphType<IdGraphType>))]
+        [InlineData(nameof(FieldTests.EnumerableField), typeof(NonNullGraphType<ListGraphType<GraphQLClrInputTypeReference<object>>>))]
+        [InlineData(nameof(FieldTests.CollectionField), typeof(NonNullGraphType<ListGraphType<GraphQLClrInputTypeReference<object>>>))]
+        public void Field_DectectsProperType(string fieldName, Type expectedGraphType)
```
sungam3r, line 184: Forgot to add tests?
изображение

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
```diff
@@ -232,6 +256,26 @@ private class FieldTests
...
+            public List<string>? NullableListNotNullStringField { get; set; }
+            public IEnumerable<int?> NotNullEnumerableNullableIntField { get; set; } = null!;
+            public IEnumerable<int> NotNullEnumerableNotNullIntField { get; set; } = null!;
+            public IEnumerable<int?>? NullableEnumerableNullableIntField { get; set; }
+            public IEnumerable<int>? NullableEnumerableNotNullIntField { get; set; }
+            public Tuple<int, string>?[] NotNullArrayNullableTupleField { get; set; } = null!;
+            public Tuple<int, string>[] NotNullArrayNotNullTupleField { get; set; } = null!;
+            public Tuple<int, string>?[]? NullableArrayNullableTupleField { get; set; }
+            public Tuple<int, string>[]? NullableArrayNotNullTupleField { get; set; }
+            [Id]
```
sungam3r, line 275: [Id]
            public int IdField { get; set; }
            public IEnumerable EnumerableField { get; set; } = null!;
            public ICollection CollectionField { get; set; } = null!;

Would like to see tests for
            [Id]
            public int? IdField { get; set; }
            public IEnumerable? EnumerableField { get; set; } = null!;
            public ICollection? CollectionField { get; set; } = null!;

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
sungam3r: the same for `AutoRegisteringObjectGraphTypeTests`

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
sungam3r: the same for `AutoRegisteringObjectGraphTypeTests`

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -75,5 +79,20 @@ protected virtual IEnumerable<PropertyInfo> GetRegisteredProperties()
             => AutoRegisteringHelper.ExcludeProperties(
                 typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(x => x.CanRead),
                 _excludedProperties);
+
+        /// <summary>
+        /// Analyzes a property and returns an instance of <see cref="TypeInformation"/>
+        /// containing information necessary to select a graph type. Any <see cref="GraphQLAttribute"/>
```
sungam3r, line 85: > Any \<see cref="GraphQLAttribute"/>

Not only those ,
https://github.com/graphql-dotnet/graphql-dotnet/pull/2884/files#r795050710

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
```diff
@@ -76,5 +80,20 @@ protected virtual IEnumerable<PropertyInfo> GetRegisteredProperties()
             => AutoRegisteringHelper.ExcludeProperties(
                 typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(x => x.CanWrite),
                 _excludedProperties);
+
+        /// <summary>
+        /// Analyzes a property and returns an instance of <see cref="TypeInformation"/>
+        /// containing information necessary to select a graph type. Any <see cref="GraphQLAttribute"/>
```
sungam3r, line 86: >  Any \<see cref="GraphQLAttribute"/>

Not only those ,
https://github.com/graphql-dotnet/graphql-dotnet/pull/2884/files#r795050710

src/GraphQL.Tests/Types/TypeInformationTests.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        [Fact]
+        public void GraphType_Input_VerifyErrors()
+        {
+            var typeInfo = new TypeInformation(_member, true);
+            typeInfo.GraphType.ShouldBeNull();
+            typeInfo.GraphType = typeof(InputObjectGraphType);
+            typeInfo.GraphType = typeof(IdGraphType);
+            Should.Throw<ArgumentException>(() => typeInfo.GraphType = typeof(ObjectGraphType));
```
sungam3r, line 32: It's worth to verify exception messages here too. Also good to understand why setter throws.

src/GraphQL/Types/TypeInformation.cs
Shane32: I'm not sure exactly what class you are referring to, since you posted the question in the `TypeInformation` class, referenced the `GraphQLAttribute` class, and posted a screenshot of the `AutoRegisteringHelper` class.  But let me try to clarify:

## TypeInformation

For `TypeInformation`, the purpose is to read a .NET `Type` and deconstruct that into bits that can be used to construct a graph type.  Graph types separate list and nullability flags from the `Type` and the NRT annotations.  There is also distinct behavior depending on if it's used as an input or as an output. So as "inputs" to this class, we need at a minimum the `PropertyInfo` to be parsed and to know if it's being used as a `get;` property or as a `set;` property.  We see that in the constructor:

```cs
public TypeInformation(PropertyInfo propertyInfo, bool isInputProperty)
```

Then the output is only the graph type.  We see that here:

```cs
public virtual Type ConstructGraphType()
```

We also allow `GraphQLAttribute`s marked on the property to modify this type information.  That is applied within this method:

```cs
public virtual void ApplyAttributes()
```

That covers everything in this entire class (besides the transient properties).  Keep in mind that when the attributes are applied, it only executes one specific overload of `GraphQLAttribute` -- the one that modifies a `TypeInformation` class, not the one that modifies a `FieldType` class:

```cs
public virtual void Modify(TypeInformation typeInformation)
```

So as an example, the user may have written this:

```cs
[Name("Id")]
[Id]
public int PersonId { get; set; }
```

We understand this to be a field named "Id" which is an `NonNullGraphType<IdGraphType>`.  But this happens with two independent parts of code.  One attribute overrides  `Modify(TypeInformation)`:

```cs
public class IdAttribute : GraphQLAttribute
{
    public override void Modify(TypeInformation typeInformation)
        => typeInformation.GraphType = typeof(IdGraphType);
}
```

While the other overrides `Modify(FieldType, bool)`:

```cs
public class NameAttribute : GraphQLAttribute
{
    // ...
    public override void Modify(FieldType fieldType, bool isInputType)
        => fieldType.Name = Name;
}
```

It can allow for an attribute to emit complex behavior, such as a a `MyCustomIdAttribute` containing both behaviors, and they both would get applied (in different areas) to the same property.

Note that while it is possible for the selected graph type to be changed by the `Modify(FieldType, bool)` method (and in fact this is how the `OutputType` attribute works), this would be rather cumbersome to retain the nullability and/or list wrappers, and it would not have access to the original `PropertyInfo` instance or original `Type` of the property.

## GraphQLAttribute

The purpose of this class is to apply changes to the schema as it is being built -- either by `AutoRegisteringObjectGraphType`, `AutoRegisteringInputObjectGraphType`, or `SchemaBuilder`.

Depending on where the derived attribute is placed, it may override one or another, or multiple, of its overridable methods.  Each method affects a different part of the schema.

| Method | Description |
|-|-|
| `Modify(TypeConfig)` | Provides configuration to the `SchemaBuilder` for types |
| `Modify(FieldConfig)` | Provides configuration to the `SchemaBuilder` for fields |
| `Modify(IGraphType)` | Provides configuration to the auto-registering classes for types |
| `Modify(FieldType, bool)` | Provides configuration to the auto-registering classes for fields |
| `Modify(TypeInformation)` | Changes graphtype-inference behavior for auto-registering fields |

There is only one attribute that does not fit within the above -- the `[Ignore]` attribute.  It seemed to me that the attribute was designed to control how the auto-registering classes scanned for fields to be added, rather than modifying a method or field as it was being built.  If you think we should change it, I can add another overridable method to `GraphQLAttribute` such as `bool ShouldInclude(MemberInfo)` or some such thing.

## GraphQL attributes vs component model attributes

The design here is that types and fields are initialized with "defaults" consisting of the following:
- Name derived from class or property name
- Description derived from `[Description]` attribute or xml comments, if those global options are enabled
- Deprecation reason derived from `[Obsolete]` attribute, if the global option is enabled
- Default value derived from `[DefaultValue]` attribute, if the global option is enabled

Then `GraphQLAttribute` attributes are applied afterwards.  Of course we could add our own `DescriptionAttribute` which would derive from `GraphQLAttribute` and set the description on a type or field.  But I saw no need for that.

Likewise the `TypeInformation` constructor performs type parsing by default, and then the attributes are applied.

## Custom attribute examples

- `[Nullable]` - force a field to be nullable, overriding NRT annotations
- `[ListNullable]` - force the list wrapper to be nullable, overriding NRT annotations
- `[MemberDefault]` - for complex default values, set a default value based on the value of a static property
- `[Auto]` - force a graph type to be `AutoRegisteringObjectGraphType<T>` or similar
- `[Model]` - append the word "Model" (or "Input") to the end of the name as a convention

I have several additional uses for attributes that I will be presenting in additional PRs that will be really cool.

src/GraphQL.Tests/Types/TypeInformationTests.cs
Shane32: Done

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Well, this method doesn't have anything to do with the description attribute, but it does read NRT attributes and the Required attribute.  I updated the comment.

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
Shane32: I updated the comment to match.

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
Shane32: Done

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
Shane32: Tests added.

src/GraphQL/Types/TypeInformation.cs
Shane32: Rewrote this comment to read

> Parses a `System.Type` along with nullability information into its constituent parts
> in preparation for creating a graph type that represents such a type.

src/GraphQL/Types/TypeInformation.cs
sungam3r: >  If you think we should change it, I can add another overridable method to `GraphQLAttribute` such as bool ShouldInclude(MemberInfo) or some such thing.

70/30. I think that it's better to do so now then some time after since it allows to
1) dynamically ignore fields, for example, to inherit from base attribute and use `[EnvIgnore(Environment: "PROD")] 
```
public class EnvIgnoreAttribute: IgnoreAttribute
{
    public string Environment { get; set; }
    public override bool ShouldInclude(....) => Environment.Current != Environment;
}
```
2) have the same design for all attributes

> Of course we could add our own DescriptionAttribute which would derive from GraphQLAttribute and set the description on a type or field. But I saw no need for that.

50/50. I would like to have single understandable design for all "graphql-related" dattributes. The disadvantage of the use of system.componentmodel attributes  is the possible conflict of interest. I came across it several times. Not big deal but may be a problem. Also inheritance from the base `GraphQLAttribute` will simplify the discoverability of these attributes and their use as I think but I'm not sure 100%. Of course shifting to such new attributes for descriptions/deprecations/default values is a breaking change, therefore I am not sure whether it is worth it.

src/GraphQL/Types/TypeInformation.cs
sungam3r: OK

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,259 @@
...
+    /// </summary>
+    public class TypeInformation
+    {
+        /// <summary>
+        /// The member being inspected.
+        /// </summary>
+        public MemberInfo MemberInfo { get; }
+
+        /// <summary>
+        /// Indicates that this is an input type (an argument); false for output types.
```
sungam3r, line 20:
```suggestion
        /// Indicates that this is an input type (an argument or input field); false for output types.
```

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,259 @@
...
+                }
+                _graphType = value;
+            }
+        }
+
+        /// <summary>
+        /// Initializes an instance with the specified properties.
+        /// </summary>
+        /// <param name="memberInfo">The member being inspected.</param>
+        /// <param name="isInputType">Indicates that this is an input type (an argument); false for output types.</param>
```
sungam3r, line 82:
```suggestion
        /// <param name="isInputType">Indicates that this is an input type (an argument or input field); false for output types.</param>
```

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,259 @@
...
+        /// <param name="type">The underlying type.</param>
+        /// <param name="isNullable">Indicates that the underlying type is nullable.</param>
+        /// <param name="isList">Indicates that this represents a list of elements.</param>
```
sungam3r, line 85:
```suggestion
        /// <param name="isList">Indicates that this member represents a list of elements.</param>
```

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,259 @@
...
+        /// <param name="listIsNullable">Indicates that the list is nullable.</param>
+        /// <param name="graphType">The graph type of the underlying CLR type; null if not specified.</param>
```
sungam3r, line 87: > null if not specified

?

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,259 @@
...
+        /// The instance is populated based on inspecting the type and NRT annotations on the specified property.
+        /// </summary>
+        public TypeInformation(PropertyInfo propertyInfo, bool isInputProperty)
+            : this(propertyInfo, isInputProperty, propertyInfo.PropertyType, false, false, false, null)
+        {
+            var typeTree = Interpret(new NullabilityInfoContext().Create(propertyInfo), isInputProperty);
+
+            foreach (var type in typeTree)
+            {
+                //detect list types, but not lists of lists
```
sungam3r, line 110: How to deal with lists of lists?

Shane32: I think `int[]?[]?` should end up as `ListGraphType<GraphQLClrOutputTypeReference<int[]>>` - but probably doesn't now and I should fix that and add a test.

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,259 @@
...
+        /// <summary>
+        /// Indicates that this is an input type (an argument); false for output types.
+        /// </summary>
+        public bool IsInputType { get; }
+
+        /// <summary>
+        /// The underlying CLR type represented. This might be the underlying type of a <see cref="Nullable{T}"/>
+        /// or the underlying type of a <see cref="IEnumerable{T}"/>.
+        /// </summary>
+        public Type Type { get; set; }
```
sungam3r, line 28: It is assumed that you can install an arbitrary type here, even nulable and then everything will be calculated?

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        /// <summary>
+        /// Indicates if the underlying type is nullable.
+        /// </summary>
+        public bool IsNullable { get; set; }
+
+        /// <summary>
+        /// Indicates that this represents a list of elements.
+        /// </summary>
+        public bool IsList { get; set; }
```
sungam3r, line 38: IsList / ListIsNullable looks some strange to me but I nearly understand what happens here. I do not want to delay this PR more time. I will see later how this code works under the debugger with concrete values.

src/GraphQL/Types/TypeInformation.cs
sungam3r: GraphQL 100% supports lists of lists, I explicitly asked about it in the spec repo. OK to merge as is for now.

Shane32: Actually GraphQL allows for lists of lists.... so maybe I should leave it alone for now...

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -0,0 +1,254 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace GraphQL.Types
+{
+    /// <summary>
+    /// Contains type and nullability information for a method return type or argument type.
+    /// </summary>
+    public class TypeInformation
```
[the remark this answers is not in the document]
sungam3r: `TypeInformation` or `TypeNullabilityInformation`

src/GraphQL/Types/TypeInformation.cs
Shane32: Ok yeah, I forgot.  Ok I won't touch it, but I will add a test to show existing behavior, which does support lists of lists, but not lists of lists of lists, and the second level of lists won't have the right NRT attribute..

Shane32: Keep in mind that we can use `[InputType]` or `[OutputType]` to force a specific exact graph type, so all is not lost...

src/GraphQL/Types/TypeInformation.cs
Shane32: Updated comment

src/GraphQL/Types/TypeInformation.cs
Shane32: Sure let's put ShouldInclude in GraphQlAttribute.  Can't do it just now, but a little later today.

src/GraphQL/Types/TypeInformation.cs
Shane32: No this property should only be set to the underlying type.

Generally, the constructor disects the original type into parts:
- Type
- IsList
- IsNullable
- ListIsNullable

This makes it easy to perform the following by attributes:
- Make the type nullable or not
- Make the list type nullable or not
- Change the underlying data type without changing the nullability or list flags

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
```diff
@@ -288,7 +288,7 @@ private class FieldTests
             public ICollection CollectionField { get; set; } = null!;
             public IEnumerable? NullableEnumerableField { get; set; }
             public ICollection? NullableCollectionField { get; set; }
-            public int[]?[]? ListOfListOfIntsField { get; set; }
+            public int?[]?[]? ListOfListOfIntsField { get; set; }
```
Shane32, line 291: This is an example of the buggy NRT code by MS.  `int[]?[]?` does not parse correctly on .NET 6, but with the patched code present in the source-only repo, it works correctly on .NET 5 and will also on .NET 7.

However, `int?[]?[]?` parses the same on both .NET 5 and .NET 6, so I just changed the test to that.

src/GraphQL/Types/TypeInformation.cs
sungam3r: So maybe add some checks in `set` ? What if I set `typeof(List<int?>)` ?

Shane32:
@sungam3r Any more comments on this PR? Let me know if I missed anything.

src/GraphQL/Types/TypeInformation.cs
Shane32: Well as of now, it is useful in that it is able to produce lists like `int[][]`, because the value of this type is passed to `GetGraphTypeFromType`.  I think we should consider this in another PR.

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/pull/2884/files#r795091517
ShouldInclude

SimonCropp:
FYI https://github.com/SimonCropp/NullabilityInfo/issues/11

Shane32:
> https://github.com/graphql-dotnet/graphql-dotnet/pull/2884/files#r795091517 ShouldInclude

Done

Shane32:
> FYI [SimonCropp/NullabilityInfo#11](https://github.com/SimonCropp/NullabilityInfo/issues/11)

Thanks!

@sungam3r Do you want to stick with the source-only repo or bring in a nuget package which include fixes for .NET 6?  (Such as parsing `int[]?[]?` correctly.)

For single-depth arrays, I think the source-only repo will suffice.  Arrays of arrays are not very well supported with the current code design anyway, so maybe we revisit this decision if/when we change that.

sungam3r:
I would prefer source package.

Shane32:
@sungam3r ok to merge this now?

src/GraphQL/GraphQLAttribute.cs
```diff
@@ -39,5 +40,18 @@ public virtual void Modify(IGraphType graphType)
...
+        /// <summary>
+        /// Updates the properties of the specified <see cref="TypeInformation"/> as necessary.
+        /// </summary>
+        public virtual void Modify(TypeInformation typeInformation)
+        {
+        }
+
+        /// <summary>
+        /// Determines if a specified member should be included during automatic generation
+        /// of a graph type from a CLR class.
```
sungam3r, line 53:
```suggestion
        /// of a graph type from a CLR type.
```

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
```diff
@@ -54,7 +54,12 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
             foreach (var propertyInfo in GetRegisteredProperties())
             {
-                if (propertyInfo.IsDefined(typeof(IgnoreAttribute)))
+                bool include = true;
+                foreach (var attr in propertyInfo.GetCustomAttributes<GraphQLAttribute>())
+                {
+                    include &= attr.ShouldInclude(propertyInfo, true);
+                }
+                if (!include)
                     continue;
```
sungam3r, line 63:
```suggestion
                bool include = true;
                foreach (var attr in propertyInfo.GetCustomAttributes<GraphQLAttribute>())
                {
                    include = attr.ShouldInclude(propertyInfo, true);
                    if (!include)
                        break;
                }
                if (!include)
                    continue;
```

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
```diff
@@ -76,5 +85,22 @@ protected virtual IEnumerable<PropertyInfo> GetRegisteredProperties()
             => AutoRegisteringHelper.ExcludeProperties(
                 typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(x => x.CanWrite),
                 _excludedProperties);
+
```
sungam3r, line 88:
```suggestion
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -53,7 +53,12 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
             foreach (var propertyInfo in GetRegisteredProperties())
             {
-                if (propertyInfo.IsDefined(typeof(IgnoreAttribute)))
+                bool include = true;
+                foreach (var attr in propertyInfo.GetCustomAttributes<GraphQLAttribute>())
+                {
+                    include &= attr.ShouldInclude(propertyInfo, true);
+                }
+                if (!include)
                     continue;
```
sungam3r, line 62:
```suggestion
                bool include = true;
                foreach (var attr in propertyInfo.GetCustomAttributes<GraphQLAttribute>())
                {
                    include = attr.ShouldInclude(propertyInfo, false);
                    if (!include)
                        break;
                }
                if (!include)
                    continue;
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -53,7 +53,12 @@ protected virtual IEnumerable<FieldType> ProvideFields()
         {
             foreach (var propertyInfo in GetRegisteredProperties())
             {
-                if (propertyInfo.IsDefined(typeof(IgnoreAttribute)))
+                bool include = true;
+                foreach (var attr in propertyInfo.GetCustomAttributes<GraphQLAttribute>())
+                {
+                    include &= attr.ShouldInclude(propertyInfo, true);
```
sungam3r, line 59: copy-pasted `true`

sungam3r (approved):
Merge after final fixes.

PR #2888, opened 2022-01-28
Enforce formatting in Visual Studio rather than by dotnet-format

Shane32:
This should work much better.  No need to push bad code just to have CI tests flunk it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -256,4 +256,6 @@ dotnet_naming_rule.async_methods_end_in_async.severity = warning
 
 # Remove unnecessary import https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0005
 dotnet_diagnostic.IDE0005.severity = error
+
+# Enforce formatting
 dotnet_diagnostic.IDE0055.severity = error
--- a/.editorconfig
+++ b/.editorconfig
@@ -255,4 +255,5 @@ dotnet_naming_style.end_in_async_style.required_suffix = Async
 dotnet_naming_rule.async_methods_end_in_async.severity = warning
 
 # Remove unnecessary import https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0005
-dotnet_diagnostic.IDE0005.severity = warning
+dotnet_diagnostic.IDE0005.severity = error
+dotnet_diagnostic.IDE0055.severity = error
--- a/.github/workflows/test-code.yml
+++ b/.github/workflows/test-code.yml
@@ -47,14 +47,6 @@ jobs:
       - name: Install dependencies
         working-directory: src
         run: dotnet restore
-      - name: Check formatting
-        if: ${{ startsWith(matrix.os, 'ubuntu') }}
-        working-directory: src
-        run: |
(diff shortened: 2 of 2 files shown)


sungam3r:
Please try push some file with unnessessary usings and check if CI fails this commit.

Shane32:
It doesn’t, and nor is it designed to.   Nor would I have it do so.

Shane32:
I specifically tested it to be sure it wouldn’t.

sungam3r:
I specifically added dotnet-format to CI so that no one could merge the code that does not correspond to the defined rules. If you remove this check, then there is no protection from the fact that anyone can merge arbitrary code violating rules from any IDE (or even without any IDE). And then you or I have to correct this code yourself. Why? I don't want to bother with all that stuff.

sungam3r:
> No need to push bad code just to have CI tests flunk it.

You solve this issue in the wrong way. The right decision will be to add rules in editorconfig to make issues (whitespaces or something else) explicitly visible in IDE.

sungam3r:
> dotnet_diagnostic.IDE0005.severity = error

👍 Good, but it is not necessary to turn off dotnet-format.

Shane32:
> The right decision will be to add rules in editorconfig to make issues (whitespaces or something else) explicitly visible in IDE.

Huh? That is what this does.

> > dotnet_diagnostic.IDE0005.severity = error
>
> 👍 Good, but it is not necessary to turn off dotnet-format.

Well, it is if we want to use a source-only repo.

Shane32:
> I specifically added dotnet-format to CI so that no one could merge the code that does not correspond to the defined rules. If you remove this check, then there is no protection from the fact that anyone can merge arbitrary code violating rules from any IDE (or even without any IDE).

Yes, I know.  I know this is personal opinion, but I do not like the formatter.  I find it extremely annoying that it constantly rejects code changes because of a BOM or extra using or spacing that it isn't happy with.  So I often have to spend an extra 5-10 minutes waiting for code to be pushed tested and rejected, and I don't have that kind of time to waste.  I would much rather 'encourage' proper formatting than 'enforce' proper formatting.  I would much rather that the editor suggest changes, rather than having no idea that the code will be rejected until CI tests execute.  But I don't mind the formatter so terribly much if it's important to you.

Now in this case, you suggested that we use a source-only repo for nullability detection.  I am trying to suggest a way in which we can do so.  You suggested the exclude parameter; that didn't work.  This was another suggestion.

sungam3r:
You can setup dotnet-format to run locally if editorconfig does not work for your cases. Either as additional build step (conditional) or external command invoked by hotkey.

>  I would much rather 'encourage' proper formatting than 'enforce' proper formatting.

I would also do that, if in practice it was. Experience shows that it is not.

sungam3r:
> You suggested the exclude parameter; that didn't work.

I'll try myself.

sungam3r:
See #2890

Red: https://github.com/graphql-dotnet/graphql-dotnet/runs/4985533002?check_suite_focus=true
Green: https://github.com/graphql-dotnet/graphql-dotnet/runs/4985633359?check_suite_focus=true

Fixed in https://github.com/graphql-dotnet/graphql-dotnet/pull/2890/commits/548237e8fbb4c462fed94f4ed07fd4278e415113

PR #2891, opened 2022-01-28
AutoRegisteringObjectGraphType supports methods

Shane32:
After this, the last couple PRs will focus on parameter support for methods (auto generated field arguments).


docs2/site/docs/migrations/migration5.md
```diff
@@ -430,3 +449,33 @@ string s = writer.ToString();
...
 into provided `TextWriter`. In the majority of cases it does not allocate memory in
 the managed heap at all.
+
+### 21. `AutoRegisteringObjectGraphType` changes
+
+The protected method `GetRegisteredProperties` has been renamed to `GetRegisteredMembers`
+and now supports properties, methods and fields, although fields are not included
+with the default implementation. Override the method in a derived class to include fields.
+
+New for v5, methods are included by default.  To revert to v4 behavior, which does not
```
sungam3r, line 459:
```suggestion
New for v5, methods are included by default. To revert to v4 behavior, which does not
```

sungam3r, line 459: > New for v5, methods are included by default.

Are you sure?

docs2/site/docs/migrations/migration5.md
```diff
@@ -430,3 +449,33 @@ string s = writer.ToString();
...
+Register this class within your DI engine like this:
+
+```cs
+services.AddTransient(typeof(AutoRegisteringObjectGraphType<>), typeof(AutoRegisteringObjectGraphTypeWithoutMethods<>));
+```
+
+### 22. `AutoRegisteringInputObjectGraphType` changes
+
+The protected method `GetRegisteredProperties` has been renamed to `GetRegisteredMembers`
+and now supports returning both properties and fields, although fields are not included
```
sungam3r, line 480: Why without methods?

Shane32: The next PR enables support for methods
- #2891 

Of course not for input objects -- only output objects.

Shane32: This is an input object.  How could it support methods?

docs2/site/docs/migrations/migration5.md
Shane32: Yes

Shane32: Methods were supported by the schema builder, but not by `AutoRegisteringObjectGraphType`.

src/GraphQL/GraphQLMetadataAttribute.cs
```diff
@@ -107,6 +107,9 @@ public override void Modify(IGraphType graphType)
 
             if (DeprecationReason != null)
                 graphType.DeprecationReason = DeprecationReason == "" ? null : DeprecationReason;
+
+            if (graphType is IObjectGraphType objectGraphType && IsTypeOf != null)
```
sungam3r, line 111: Was lost in prev PRs?

Shane32: In a manner of speaking.  This entire method is new as of the last PR.  I missed adding this in, yes.  `IsTypeOf` has always continued to work for the schema builder where it was designed for.

src/GraphQL/IdAttribute.cs
```diff
@@ -6,7 +6,7 @@ namespace GraphQL
     /// <summary>
     /// Specifies that a property will be mapped to <see cref="IdGraphType"/>.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Property)]
+    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Method | AttributeTargets.Field, Inherited = false)]
```
sungam3r, line 9: Why do you use `Inherited=false` here and in other places?

```csharp
public class Person
{
  [Id]
  public virtual string Name {get;set; } // <-- IdGraphType
}

public class Manager : Person
{
  public override Name {get;set} // <-- StringGraphType
}

File: src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs [hunk 5]
```diff
@@ -52,7 +52,7 @@ public AutoRegisteringInputObjectGraphType(params Expression<Func<TSourceType, o
         /// </summary>
         protected virtual IEnumerable<FieldType> ProvideFields()
         {
-            foreach (var propertyInfo in GetRegisteredProperties())
+            foreach (var propertyInfo in GetRegisteredMembers())
```
sungam3r, line 55:
```suggestion
            foreach (var memberInfo in GetRegisteredMembers())
```

sungam3r:
I will review carefully tomorrow.

docs2/site/docs/migrations/migration5.md
sungam3r: How does the method differ from the property in this context? The property is the same method, even two.

Shane32: Yes, but how would it work?

```cs
//for output types
class Test
{
    public int Value1 => 3;
    public int Value2() => 3;
}
//usage within field resolver
var result1 = test.Value1;
var result2 = test.Value2();
//both read the value -- so far so good

//for input types
class Test
{
    public int Value { get; set; }
    //what is the signature?
    public void Value2(int newValue); //like this?? why?? very uncommon for data input
}
//usage internally
test.Value1 = 3;
test.Value2(3); //like this?  seems quite abnormal for such a use case
```

It just doesn't seem to make any sense to support methods for input objects.

src/GraphQL/IdAttribute.cs
Shane32: I misunderstood what the `Inherited` property was for.  Certainly we would want it set to `true`.  I'll fix it, and add tests.

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
Shane32: Done

src/GraphQL/IdAttribute.cs
Shane32: Done

Shane32:
This PR is waiting on #2884 before CI checks will execute.

docs2/site/docs/migrations/migration5.md
sungam3r: I imagined it differently. Perhaps I completely misunderstood the goal.

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -105,6 +107,39 @@ public TypeInformation(PropertyInfo propertyInfo, bool isInputProperty)
         {
             var typeTree = Interpret(new NullabilityInfoContext().Create(propertyInfo), isInputProperty);
 
+            ProcessTypeTree(typeTree, isInputProperty);
```
sungam3r, line 110:
```suggestion
            ProcessTypeTree(typeTree, isInput);
```

\+ Interpret(NullabilityInfo info, bool isInputProperty) -> Interpret(NullabilityInfo info, bool isInput)

sungam3r:
I look throught half of files, will continue tomorrow. I need to understand the idea of supporting methods.

Shane32:
> I look throught half of files, will continue tomorrow. I need to understand the idea of supporting methods.

Look at the description of #2892 which contains a sample of the end goal

src/GraphQL/Types/TypeInformation.cs
Shane32: done (both)

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
```diff
@@ -238,6 +247,31 @@ public void CanOverrideFieldList()
...
+            {
+                { "Field1", 11 },
+                { "Field3", 13 },
+                { "Field5", 15 },
+                { "Field6AltName", 16 },
+            };
+            var graph = new TestFieldSupport<TestClass>();
+            var actual = graph.ParseDictionary(dic).ShouldBeOfType<TestClass>();
+            actual.Field1.ShouldBe(11);
+            actual.Field3Value.ShouldBe(13);
```
sungam3r, line 270: 13 ? How does Field3 translates into Field3Value?

docs2/site/docs/migrations/migration5.md
sungam3r: I want to understand. Context of using output types - resolvers. You give the opportunity to work with CLR classes and GraphQL.NET maps them into graphtypes. So if method or property returns `int value` then this `value` translated to field with type `Int`. Right? What is the context of input types translation? Arguments?

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -238,6 +253,37 @@ public void CanOverrideFieldList()
...
+            var graph = new TestFieldSupport<TestClass>();
+            var field = graph.Fields.Find(fieldName).ShouldNotBeNull();
+            var resolver = field.Resolver.ShouldNotBeNull();
+            var obj = new TestClass();
+            var actual = resolver.Resolve(new ResolveFieldContext
+            {
+                Source = obj,
+                FieldDefinition = new FieldType { Name = fieldName },
+            });
+            if (actual is Task task)
```
sungam3r, line 280: Never hit?

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
Shane32: Field3 is a write-only property which contains code to populate Field3Value for purposes of testing only.

src/GraphQL/ObjectExtensions.cs
```diff
@@ -169,6 +169,25 @@ string GetPropertyName(string key, out FieldType? field)
                     object? value = GetPropertyValue(item.Value, propertyInfo.PropertyType, field?.ResolvedType);
                     propertyInfo.SetValue(obj, value, null); //issue: this works even if propertyInfo is ValueType and value is null
                 }
+                else
+                {
+                    FieldInfo? fieldInfo;
+
+                    try
+                    {
+                        fieldInfo = type.GetField(propertyName, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance);
```
sungam3r, line 178: Only public fields? May be surprise for someone who overrided `GetRegisteredMembers`. But the same true for properties so maybe just add a note into docs.

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
Shane32: Missed Field7 inlinedata; fixed.

src/GraphQL/ObjectExtensions.cs
Shane32: I plan to override ParseDictionary so this code isn't hit at all when using AutoRegisteringInputObjectGraphType.  I have not thought about exactly how this will be done yet.  However, this code will still be used for InputObjectGraphType.  As you can see, I added code for fields to match that of properties, which only supported public members.  Yes a comment is deserving.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -32,19 +33,27 @@ internal static IEnumerable<PropertyInfo> ExcludeProperties<TSourceType>(IEnumer
...
-                DeprecationReason = propertyInfo.ObsoleteMessage(),
+                Name = memberInfo.Name,
+                Description = memberInfo.Description(),
+                DeprecationReason = memberInfo.ObsoleteMessage(),
                 Type = graphType,
-                DefaultValue = isInputType ? propertyInfo.DefaultValue() : null,
+                DefaultValue = isInputType ? memberInfo.DefaultValue() : null,
+                Resolver = isInputType ? null : MemberResolver.Create(memberInfo),
             };
+            if (isInputType)
```
sungam3r, line 50: Why only fro inputs?

docs2/site/docs/migrations/migration5.md
```diff
@@ -448,3 +467,33 @@ the managed heap at all.
...
+with the default implementation. Override the method in a derived class to include fields.
+
+New for v5, methods are included by default. To revert to v4 behavior, which does not
+include methods, create a derived class as follows:
+
+```csharp
+public class AutoRegisteringObjectGraphTypeWithoutMethods<T> : AutoRegisteringObjectGraphType<T>
+{
+    public AutoRegisteringObjectGraphTypeWithoutMethods() : base() { }
+    public AutoRegisteringObjectGraphTypeWithoutMethods(params Expression<Func<T, object?>>[]? excludedProperties) : base(excludedProperties) { }
```
sungam3r, line 484: A little inaccuracy here. `AutoRegisteringInputObjectGraphType` may work with properties and fields, `AutoRegisteringObjectGraphType` may work with properties, fields and method, but both contain parameter to exclude  only properties. Is it right?

docs2/site/docs/migrations/migration5.md
sungam3r: OK. The discard my suggestion abot methods.

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
sungam3r: 🤦‍♂️ I see

src/GraphQL/Resolvers/MemberResolver.cs
```diff
@@ -0,0 +1,51 @@
...
+            var convertToObjectExpression = Expression.Convert(expression, typeof(object));
+            var lambda = Expression.Lambda<Func<object?, object?>>(convertToObjectExpression, parameter);
+            _resolver = lambda.Compile();
+        }
+
+        private static readonly ConcurrentDictionary<MemberInfo, MemberResolver> _resolverDictionary = new();
+        private static readonly Func<MemberInfo, MemberResolver> _createFunc
+            = (MemberInfo memberInfo) => new MemberResolver(memberInfo);
+        public static MemberResolver Create(MemberInfo memberInfo)
+            => _resolverDictionary.GetOrAdd(memberInfo, _createFunc);
```
sungam3r, line 47:
```suggestion
        public static MemberResolver Create(MemberInfo memberInfo)
            => _resolverDictionary.GetOrAdd(memberInfo, info => new MemberResolver(info));
```

Shoulf work the same.

src/GraphQL/Resolvers/MemberResolver.cs
```diff
@@ -0,0 +1,51 @@
...
+            var parameter = Expression.Parameter(typeof(object), "x");
+            Expression expression = memberInfo switch
+            {
+                MethodInfo methodInfo => Expression.Call(
+                    methodInfo.IsStatic
+                        ? null
+                        : Expression.Convert(parameter, methodInfo.DeclaringType!),
+                    methodInfo),
+                PropertyInfo propertyInfo => Expression.MakeMemberAccess(
+                    (propertyInfo.GetMethod ?? throw new ArgumentOutOfRangeException(nameof(memberInfo), "Property does not have get method.")).IsStatic
```
sungam3r, line 27:
```suggestion
                    (propertyInfo.GetMethod ?? throw new ArgumentOutOfRangeException(nameof(memberInfo), $"Property {propertyInfo.PropertyName} does not have get method.")).IsStatic
```

src/GraphQL/Resolvers/MemberResolver.cs
```diff
@@ -0,0 +1,51 @@
...
+                        ? null
+                        : Expression.Convert(parameter, propertyInfo.DeclaringType!),
+                    propertyInfo),
+                FieldInfo propertyInfo => Expression.MakeMemberAccess(
+                    propertyInfo.IsStatic
+                        ? null
+                        : Expression.Convert(parameter, propertyInfo.DeclaringType!),
+                    propertyInfo),
```
sungam3r, line 35:
```suggestion
                FieldInfo fieldInfo => Expression.MakeMemberAccess(
                    fieldInfo.IsStatic
                        ? null
                        : Expression.Convert(parameter, fieldInfo.DeclaringType!),
                    fieldInfo),
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -51,54 +52,74 @@ public AutoRegisteringObjectGraphType(params Expression<Func<TSourceType, object
...
+            var props = AutoRegisteringHelper.ExcludeProperties(
                 typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(x => x.CanRead),
                 _excludedProperties);
+            var methods = typeof(TSourceType).GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly)
+                .Where(x =>
+                    !x.ContainsGenericParameters && //exclude methods with open generics
+                    !x.IsSpecialName &&             //exclude methods generated for properties
+                    x.ReturnType != typeof(void) && //exclude methods which do not return a value
+                    x.ReturnType != typeof(Task) && //exclude methods which do not return a value
+                    x.GetParameters().Length == 0); //exclude methods which contain arguments
```
sungam3r, line 101: I guess it will be improved in the next PR?

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -51,54 +52,74 @@ public AutoRegisteringObjectGraphType(params Expression<Func<TSourceType, object
...
-        protected virtual IEnumerable<PropertyInfo> GetRegisteredProperties()
-            => AutoRegisteringHelper.ExcludeProperties(
+        protected virtual IEnumerable<MemberInfo> GetRegisteredMembers()
+        {
+            var props = AutoRegisteringHelper.ExcludeProperties(
                 typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(x => x.CanRead),
                 _excludedProperties);
+            var methods = typeof(TSourceType).GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly)
+                .Where(x =>
+                    !x.ContainsGenericParameters && //exclude methods with open generics
```
sungam3r, line 97:
```suggestion
                    !x.ContainsGenericParameters && // exclude methods with open generics
```

Let's use one whitespace indent for all comments.

sungam3r:
I looked through all files.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: It is only used for ToObject within ParseDictionary. I can apply it to output graphs also but it is not necessary; the field resolver doesn’t need this metadata set to work properly. 

Once I add an overload for ParseDictionary then it won’t be needed at all. 

Should I just set it for both inputs and outputs anyway?

docs2/site/docs/migrations/migration5.md
Shane32: You are correct; I suggest we mark that constructor obsolete. It’s not like it can be used by DI (very easily). Or we can just leave it.

Note that it is impractical to use it to exclude methods, since the expression would need a full method call, but it would make sense for fields - although fields require a derived class to be useful. So I didn’t bother to change the functionality here. 

What is inaccurate? I’ll have to reread the documentation when I’m at a pc again.

src/GraphQL/Resolvers/MemberResolver.cs
Shane32: Correct I’ll change it.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Yes

src/GraphQL/Resolvers/MemberResolver.cs
Shane32:
```suggestion
                    (propertyInfo.GetMethod ?? throw new ArgumentOutOfRangeException(nameof(memberInfo), $"Property '{propertyInfo.Name}' does not have get method.")).IsStatic
```

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: I was also thinking it would be more valuable to store the MemberInfo in the metadata rather than just the field name. That way we don’t have to search the class for a matching property name so parsing variables should be faster.

sungam3r: OK. Do as you see fit.

docs2/site/docs/migrations/migration5.md
sungam3r: > Or we can just leave it.

OK.

> What is inaccurate?

Expose option to ignore only properties, but OK.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -51,54 +52,74 @@ public AutoRegisteringObjectGraphType(params Expression<Func<TSourceType, object
...
+            var props = AutoRegisteringHelper.ExcludeProperties(
                 typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(x => x.CanRead),
                 _excludedProperties);
+            var methods = typeof(TSourceType).GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly)
+                .Where(x =>
+                    !x.ContainsGenericParameters && // exclude methods with open generics
+                    !x.IsSpecialName &&             //exclude methods generated for properties
+                    x.ReturnType != typeof(void) && //exclude methods which do not return a value
+                    x.ReturnType != typeof(Task) && //exclude methods which do not return a value
+                    x.GetParameters().Length == 0); //exclude methods which contain arguments
```
sungam3r, line 101:
```suggestion
                    !x.IsSpecialName &&             // exclude methods generated for properties
                    x.ReturnType != typeof(void) && // exclude methods which do not return a value
                    x.ReturnType != typeof(Task) && // exclude methods which do not return a value
                    x.GetParameters().Length == 0); // exclude methods which contain arguments
```

sungam3r (approved):
OK to merge after final fixes if any.

docs2/site/docs/migrations/migration5.md
```diff
@@ -168,9 +187,9 @@ automatically-generated graph types:
...
 | ConfigureGraph   | Configures default graph properties prior to applying attributes     | Applying a different default naming convention, such as appending "Input" or "Model" |
-| GetRegisteredProperties | Returns the set of properties to be automatically configured  | Filtering internal properties; sorting the property list |
+| GetRegisteredMembers | Returns the set of properties, methods and fields to be automatically configured | Filtering internal properties; sorting the property list; including fields; excluding methods |
 | ProvideFields    | Returns a set of generated fields                                    | Adding additional fields to the generated set |
-| CreateField      | Creates a `FieldType` from a `PropertyInfo`                          | Applying custom behavior to field generation |
+| CreateField      | Creates a `FieldType` from a `MemberInfo`                            | Applying custom behavior to field generation |
 
 If you utilize dependency injection within your schema, you can register your custom graph
```
Shane32, line 194:
```suggestion
Note that if you override `GetRegisteredMembers` to include private methods or fields, you
may also need to override `ParseDictionary` as well.

If you utilize dependency injection within your schema, you can register your custom graph
```

Shane32: I'll add this in now but probably it will change later once I redo ParseDictionary.

PR #2892, opened 2022-01-30
AutoRegisteringObjectGraphType supports field arguments

Shane32:
Method parameters are added as query arguments to the field.  However, no query argument is added in these cases:

- If the parameter is a `CancellationToken`, it is passed the cancellation token from `IResolveFieldContext`.
- If the parameter is a `IResolveFieldContext`, it is passed the `IResolveFieldContext`.

The field resolver is precompiled and should run as fast as any delegate passed to a `FuncFieldResolver`.  To populate the method parameters, it calls `GetArgument<T>` for each query argument necessary, and then calls the method.

GraphQL attributes can supply replacement behavior for an argument, such as the included `[FromServices]` which like its counterpart in ASP.Net can pull an argument from the `IServiceProvider` rather than a query argument.

Sample:

```cs
[Name("Person")]
public class Human
{
    [Id]
    public int Id { get; set; }

    [Name("Orders")]
    public async Task<IEnumerable<Order>> GetOrdersAsync(
        [FromServices] DbContext db,
        CancellationToken token,
        [Name("MaxResults")] int maxReturned = 100)
    {
        return await db.Orders
            .Where(x => x.HumanId == Id)
            .OrderByDesc(x => x.OrderDate)
            .Take(maxReturned)
            .ToListAsync(token);
    }
}
```

This would produce a GraphQL schema like this:

```gql
type Person {
  Id: ID!
  Name: String!
  Orders(MaxResults: Int! = 100): [Order!]!
}
```

The field resolver for the `Orders` field would execute a precompiled lambda that executes:

```cs
Func<IResolveFieldContext, object> resolverFunc = context => context.Source.GetOrdersAsync(
    context.RequestServices.GetRequiredService<DbContext>(),
    context.CancellationToken,
    context.GetArgument<int>("MaxResults", 100));
```

Todo:
- [x] Tests
- [x] XML comments
- [x] Documentation
- [x] Self-review


src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,115 @@
...
+                Expression<Func<IResolveFieldContext, CancellationToken>> expr = x => x.CancellationToken;
+                Expression = expr;
+            }
+        }
+
+        public ParameterInfo ParameterInfo { get; }
+
+        public Type SourceType { get; }
+
+        public FieldType FieldType { get; }
```
Shane32, line 38: `SourceType` is useful if the attribute logic needs to interact with `context.Source`.

`FieldType` may not really be needed.  It provides attributes with information about the field to which the argument is being added.  I have to say I can't think of a scenario where it would be useful, so we could remove it.

src/GraphQL/MissingRequestServicesException.cs
```diff
@@ -1,9 +1,13 @@
 using System;
 
-namespace GraphQL.MicrosoftDI
+namespace GraphQL
 {
-    internal class MissingRequestServicesException : InvalidOperationException
+    /// <summary>
+    /// Indicates that <see cref="IResolveFieldContext.RequestServices"/> was required but not set.
+    /// </summary>
+    public class MissingRequestServicesException : InvalidOperationException
```
Shane32, line 8: Needed by both MicrosoftDI and main library, so I moved it from MicrosoftDI and made it public.

src/GraphQL/TypeExtensions.cs
```diff
@@ -329,6 +329,32 @@ public static bool ImplementsGenericType(this Type type, Type genericType)
...
+            if (GlobalSwitches.EnableReadDescriptionFromAttributes)
+            {
+                description = (parameterInfo.GetCustomAttributes(typeof(DescriptionAttribute), false).FirstOrDefault() as DescriptionAttribute)?.Description;
+                if (description != null)
+                    return description;
+            }
+
+            if (GlobalSwitches.EnableReadDescriptionFromXmlDocumentation)
+            {
+                description = parameterInfo.GetXmlDocumentation();
```
Shane32, line 352: This overload of `GetXmlDocumentation` already existed although was unused.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -77,7 +78,89 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
+            // apply field attributes after resolver has been set
+            AutoRegisteringHelper.ApplyFieldAttributes(memberInfo, fieldType, false);
+            return fieldType;
+        }
+
+        /// <summary>
+        /// Retuns a list of query arguments within a <see cref="QueryArguments"/> instance, and a
+        /// <see cref="IFieldResolver"/> instance, for the specified field.
+        /// </summary>
+        protected (QueryArguments? QueryArguments, IFieldResolver Resolver) BuildMethodResolver(MethodInfo methodInfo, FieldType fieldType)
```
Shane32, line 105: This is not `virtual`.  I cannot think of any reason it should be virtual, and it is quite complex code that isn't easily replicated.  Any behavior that anyone would want to change has been moved to the `GetArgumentInformation` and `ApplyArgumentAttributes` virtual methods.  By overriding `CreateField`, which contains the only call to `BuildMethodResolver`, it is possible to "override" this method also.

sungam3r:
Conflicts after merging implicit usings. Will review tomorrow.

Shane32:
bump @sungam3r

sungam3r:
I was going to review it today, but I had to urgently fix broken CI at work. Will review tomorrow.

docs2/site/docs/migrations/migration5.md
```diff
@@ -150,6 +150,63 @@ and data loader methods are supported such as shown in the below example:
 The above example assumes that the `Person` CLR type was mapped to `PersonGraphType`
 in the schema CLR type mappings.
 
+Arguments are also added as query arguments to the field, and recognize default values
```
sungam3r, line 153: I still do not like the term `query arguments`. I hope we will eventually change it.

docs2/site/docs/migrations/migration5.md
```diff
@@ -150,6 +150,63 @@ and data loader methods are supported such as shown in the below example:
...
+
+public enum SortOrder
+{
+    Date
+}
+```
+
+The above code would generate a GraphQL schema like this:
+
+```gql
```
sungam3r, line 202:
```suggestion
```graphql
```

docs2/site/docs/migrations/migration5.md
```diff
@@ -180,6 +236,9 @@ private class CustomAutoObjectType<T> : AutoRegisteringObjectGraphType<T>
 Similarly, by overriding `CreateField` you can change the default name, description,
 graph type, or other information applied to each generated field.
 
+Most of these changes can be performed by attributes, but by creating a derived class you can
+change default behavior without needing to add attributes to all of your data models.
```
sungam3r, line 240:
```suggestion
Most of these changes can be performed declaratively by attributes, but by creating a derived class you can
change default behavior imperatively without needing to add attributes to all of your data models.
```

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
```diff
@@ -160,3 +160,12 @@ namespace GraphQL.MicrosoftDI
         public object GetService(System.Type serviceType) { }
     }
 }
+namespace GraphQL
+{
+    [System.AttributeUsage(System.AttributeTargets.Method | System.AttributeTargets.All)]
```
sungam3r, line 165: Bump PublicApiGenerator to 10.3.0 to get rid of `| System.AttributeTargets.All`.

src/GraphQL.MicrosoftDI/DynamicScopedFieldResolver.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace GraphQL.MicrosoftDI
+{
+    /// <summary>
+    /// When resolving a field, this implementation calls
+    /// <see cref="IResolveFieldContext.RequestServices"/>.<see cref="ServiceProviderServiceExtensions.CreateScope(IServiceProvider)">CreateScope</see>
+    /// to create a dependency injection scope. Then it calls the specified <see cref="IFieldResolver"/>, passing the scoped service provider
+    /// within <see cref="IResolveFieldContext.RequestServices"/>, and returns the result.
+    /// </summary>
+    internal class DynamicScopedFieldResolver : IFieldResolver<Task<object>>
```
sungam3r, line 12: We already have some similar resolvers. None of them fit?

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// Creates a dedicated service scope during the field resolver's execution.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public class ScopedAttribute : GraphQLAttribute
+    {
+        /// <inheritdoc/>
+        public override void Modify(FieldType fieldType, bool isInputType)
+        {
```
sungam3r, line 14:
```suggestion
        {
            Debug.Assert(isInputType == false);
```
?

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+            if (fieldType.Resolver == null)
```
sungam3r, line 15: Is this condition ever triggered?

src/GraphQL.MicrosoftDI/DynamicScopedFieldResolver.cs
Shane32: No

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
Shane32: Sure, if you like but it would only verify that the tests were written correctly. It won’t prevent users tagging input fields. But since an input field does not have a field resolver, this attribute will do nothing. (See if condition on next line of code.)

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
Shane32: Won’t we do that in a separate PR for the whole solution?

docs2/site/docs/migrations/migration5.md
Shane32: Fine but for now it is consistent, and helps to distinguishing from method arguments.

src/GraphQL.MicrosoftDI/DynamicScopedFieldResolver.cs
Shane32: I can explain, but the bottom line is simply “no”. Especially this one. The most similar code is within the middleware builder.

Shane32: Let me know if you think this should be public.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
sungam3r: ok

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
sungam3r: I would make condition expliciit: `if (isInputType) return;`

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: ok

src/GraphQL/MissingRequestServicesException.cs
sungam3r: ok

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
Shane32:
```suggestion
            if (isInputType || fieldType.Resolver == null)
```

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -42,21 +41,34 @@ internal static FieldType CreateField(MemberInfo memberInfo, Type graphType, boo
...
             };
             if (isInputType)
             {
                 fieldType.WithMetadata(ComplexGraphType<object>.ORIGINAL_EXPRESSION_PROPERTY_NAME, memberInfo.Name);
             }
+            if (!isInputType &&
+                memberInfo is MethodInfo methodInfo &&
+                fieldType.Name.EndsWith("Async") &&
+                methodInfo.ReturnType.IsGenericType &&
+                methodInfo.ReturnType.GetGenericTypeDefinition() == typeof(Task<>))
```
sungam3r, line 53: \+ ValueTask<> support?

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -69,5 +81,40 @@ private static string GetPropertyName<TSourceType>(Expression<Func<TSourceType,
...
+                resolveFieldContextParameter);
+        }
+
+        private static readonly MethodInfo _getArgumentMethod = typeof(AutoRegisteringHelper).GetMethod(nameof(GetArgumentInternal), BindingFlags.NonPublic | BindingFlags.Static)!;
+        /// <summary>
+        /// Returns the value for the specified query argument, or the default value of the query argument
+        /// if a value was not specified in the request.
+        /// <br/><br/>
+        /// Unlike <see cref="ResolveFieldContextExtensions.GetArgument{TType}(IResolveFieldContext, string, TType)"/>,
+        /// the default value is not returned if <see langword="null"/> was supplied.
```
sungam3r, line 109: > if \<see langword="null"/> was supplied

supplied where?

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -69,5 +81,40 @@ private static string GetPropertyName<TSourceType>(Expression<Func<TSourceType,
...
             throw new NotSupportedException($"Unsupported type of expression: {expression.GetType().Name}");
         }
+
+        /// <summary>
+        /// Constructs a lambda expression for a field resolver to return the specified query argument
+        /// from the resolve context.
+        /// </summary>
+        internal static LambdaExpression GetParameterExpression(Type parameterType, QueryArgument queryArgument)
+        {
+            //construct a call to ResolveFieldContextExtensions.GetArgument, passing in the appropriate default value
```
sungam3r, line 91: GetArgument or TryGetArgument?

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -69,5 +81,40 @@ private static string GetPropertyName<TSourceType>(Expression<Func<TSourceType,
...
+            var getArgumentMethodTyped = _getArgumentMethod.MakeGenericMethod(parameterType);
+            var resolveFieldContextParameter = Expression.Parameter(typeof(IResolveFieldContext), "context");
+            var queryArgumentExpression = Expression.Constant(queryArgument, typeof(QueryArgument));
+            //e.g. Func<IResolveFieldContext, int> = (context) => ResolveFieldContextExtensions.GetArgument<int>(context, queryArgument.Name, queryArgument.DefaultValue);
```
sungam3r, line 95: I see no default value here.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: As of now there is no ValueTask support for the entire GraphQL.NET execution pipeline. It is something I’d like to explore. It would apply to the entire solution at once, not just this method, I expect.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: As a literal or variable. 

I feel that the getargument behavior is wrong. Something we should consider changing. But certainly the argument support here doesn’t work properly with the normal GetArgument behavior.

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
```diff
@@ -0,0 +1,21 @@
+using GraphQL.MicrosoftDI;
+using GraphQL.Types;
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Creates a dedicated service scope during the field resolver's execution.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public class ScopedAttribute : GraphQLAttribute
```
sungam3r, line 10: Can I use [Scoped] with [FromServices]. If yes does the order matter?

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: Needs to be updated to AutoRegisteringHelper.GetArgumentInternal and etc.

sungam3r:
@Shane32 I looked through a half of files, a bit hard for me to understand sometimes. I will continue tomorrow.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: Yeah the xml comments need to be updated. 

This code was tricky to get working properly because attributes can change the query argument properties after the expression is constructed. So now it references the query argument directly.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
sungam3r: ok

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: So even if the name or default value changes via attributes, the behavior is correct.

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
Shane32: Scoped is applied to the method. FromServices is applied to the method argument. 

Certainly; that is the point. I think it is shown correctly in the docs? I can’t remember.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: XML comments have been updated.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: XML comments have been updated

src/GraphQL.MicrosoftDI/DynamicScopedFieldResolver.cs
Shane32: I looked at this again.  Technically any field resolver could use `FuncFieldResolver`.  But then the functionality encapsulated within the field resolver moves to a method somewhere and limits re-use.  In this case I think the better design is to create a dedicated field resolver rather than putting the code directly in the attribute or in a static method.

Shane32:
bump @sungam3r

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
sungam3r: > As a literal or variable.

Ah. I look for a null value in some method argument. Worth to add _was supplied in request as a literal or variable_.

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
sungam3r: OK.

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
```diff
@@ -0,0 +1,21 @@
+using GraphQL.MicrosoftDI;
+using GraphQL.Types;
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Creates a dedicated service scope during the field resolver's execution.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
```
sungam3r, line 9: AllowMultiple = false

src/GraphQL/FromServicesAttribute.cs
```diff
@@ -0,0 +1,20 @@
+using GraphQL.Types;
+using GraphQL.Utilities;
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Specifies that the method argument should be pulled from <see cref="IResolveFieldContext.RequestServices"/>.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
```
sungam3r, line 9: AllowMultiple = false

src/GraphQL.MicrosoftDI/DynamicScopedFieldResolver.cs
sungam3r: ok

src/GraphQL/FromServicesAttribute.cs
Shane32: https://docs.microsoft.com/en-us/dotnet/api/system.attributeusageattribute.allowmultiple?view=net-6.0

> true if more than one instance is allowed to be specified; otherwise, false. The default is false.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: It's already been updated.  Suggest a change if you like.

src/GraphQL.MicrosoftDI/ScopedAttribute.cs
Shane32: https://docs.microsoft.com/en-us/dotnet/api/system.attributeusageattribute.allowmultiple?view=net-6.0

> true if more than one instance is allowed to be specified; otherwise, false. The default is false.

sungam3r: ok

src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
```diff
@@ -0,0 +1,96 @@
+using GraphQL.Types;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GraphQL.MicrosoftDI.Tests
+{
+    public class ScopedAttributeTests
+    {
+        [Fact]
+        public async void ScopedMethodWorks()
```
sungam3r, line 9: Good test. I understand what is going on better.

src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            (await (Task<object>)scopedAsyncFieldResolver.Resolve(context)).ShouldBe("0 1");
+            Class1.DisposedCount.ShouldBe(4);
+            unscopedFieldResolver.Resolve(context).ShouldBe("3 4");
+            rootServiceProvider.Dispose();
+            Class1.DisposedCount.ShouldBe(5);
+        }
+
+        private class TestClass
+        {
+            public string UnscopedField([FromServices] Class1 arg1, [FromServices] Class2 arg2)
```
sungam3r, line 44: By the way, how to deal with a such (awkward) cases:
```
  [Scoped]
  public string BlaBla([FromServices] Class1 arg1, [FromServices] Class1 arg2,  [FromServices] Class1 arg3,  [FromServices] Class1 arg4)
```

Would be arg1..4 different references or the same?

sungam3r: I think the same.

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -203,33 +206,113 @@ public void Field_DectectsProperType(string fieldName, Type expectedGraphType)
...
+            }
+            if (arg2Value.HasValue)
+            {
+                context.Arguments.Add("arg2", new ArgumentValue(arg2Value, ArgumentSource.Variable));
+            }
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddSingleton<string>("testService");
+            context.RequestServices = serviceCollection.BuildServiceProvider();
+            fieldType.Resolver.ShouldNotBeNull();
+            fieldType.Resolver!.Resolve(context).ShouldBe(expected);
```
sungam3r, line 293:
```suggestion
            using var provider = serviceCollection.BuildServiceProvider();
            context.RequestServices = provider;
            fieldType.Resolver.ShouldNotBeNull();
            fieldType.Resolver!.Resolve(context).ShouldBe(expected);
```

src/GraphQL/ExpressionExtensions.cs
```diff
@@ -0,0 +1,44 @@
+using System.Linq.Expressions;
+
+namespace GraphQL
+{
+    internal static class ExpressionExtensions
+    {
+        /// <summary>
+        /// Replaces all occurrances of <paramref name="oldParameter"/> with <paramref name="newBody"/> within <paramref name="expression"/>.
```
sungam3r, line 8:
```suggestion
        /// Replaces all occurrences of <paramref name="oldParameter"/> with <paramref name="newBody"/> within <paramref name="expression"/>.
```

sungam3r:
Uh, 5 files left.

src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
Shane32: It would depend on if they were registered as transients (different) or scoped/singletons (same), exactly like if you did the same in a constructor.

sungam3r: ok

src/GraphQL/Resolvers/MethodResolver.cs
```diff
@@ -0,0 +1,67 @@
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace GraphQL.Resolvers
+{
+    /// <summary>
+    /// A field resolver for a specific <see cref="MemberInfo"/>.
+    /// Returns the value of the field or property, or for methods, calls the method and returns the value of the method.
```
sungam3r, line 8: Looks like comments are outdated.

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        private LambdaExpression? _expression;
+        /// <summary>
+        /// Gets or sets a delegate in the form of a <see cref="LambdaExpression"/> to be used to populate
+        /// this method argument while building the field resolver.
+        /// <br/><br/>
+        /// If not set, a query argument will be added to the field and the argument's value will be used
+        /// to populate the method argument while building the field resolver.
+        /// <br/><br/>
+        /// The delegate must of the type
```
sungam3r, line 80:
```suggestion
        /// The delegate must be of the type
```

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,179 @@
...
+        public LambdaExpression? Expression
+        {
+            get => _expression;
+            set
+            {
+                if (value != null && (value.ReturnType != ParameterInfo.ParameterType || value.Parameters.Count != 1 || value.Parameters[0].Type != typeof(IResolveFieldContext)))
+                {
+                    throw new ArgumentException($"Value must be a lambda expression delegate of type Func<IResolveFieldContext, {ParameterInfo.ParameterType.Name}>.");
+                }
+                _expression = value;
```
sungam3r, line 93: So we can reset value?

src/GraphQL/Resolvers/MethodResolver.cs
Shane32:
```suggestion
    /// A field resolver for a specific <see cref="MethodInfo"/>.
    /// Calls the specified method (with the specified arguments) and returns the value of the method.
```

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Builds a query argument or expression from this instance.
+        /// <br/><br/>
+        /// If a query argument is returned, it will be added to field type.
```
sungam3r, line 100:
```suggestion
        /// If a query argument is returned, it will be added to arguments of field type.
```

Also it worth to note _where_ it will be added. I understand what you wrote but class/method are public so comment seems a bit strange.

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// <br/><br/>
+        /// If an expression is returned, it will be used to populate the method argument within the field resolver;
+        /// if not, the query argument's value will be used to populate the method argument within the field resolver.
+        /// <br/><br/>
+        /// The default implemenatation will return either a <see cref="QueryArgument"/> or <see cref="LambdaExpression"/>
```
sungam3r, line 105:
```suggestion
        /// The default implementation will return either a <see cref="QueryArgument"/> or <see cref="LambdaExpression"/>
```

src/GraphQL/Types/ArgumentInformation.cs
Shane32: Yes (or clear it).  It's read/write, so long as it's valid.  A value of "null" is described in detail in the xml comments.

src/GraphQL/Types/ArgumentInformation.cs
Shane32:
```suggestion
        /// If a query argument is returned, it will be added to the arguments list of the field type.
```

src/GraphQL/FromServicesAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class FromServicesAttribute : GraphQLAttribute
+    {
+        /// <inheritdoc/>
+        public override void Modify<TReturnType>(ArgumentInformation<TReturnType> argumentInformation)
```
sungam3r, line 13: It seems that method is not needed to be generic.

src/GraphQL/Types/ArgumentInformation.cs
Shane32: I know it's a little odd, but it was written from the perspective of someone attempting to override the method, to help them understand what will happen when it is overridden.  I tried to be as clear as possible.

sungam3r: I'm fine to keep as is or you may even extend your comment. As you see fit.

src/GraphQL/FromServicesAttribute.cs
Shane32: No it is not "needed"; I didn't have it generic initially.  But it made the code MUCH more complex and hard to use.  You have to build the expression from scratch and return it.  Not pretty for anyone trying to write their own attribute.

With it as a generic method, writing a custom attribute is trivial, like this one we might even want to include:

```cs
[AttributeUsage(AttributeTargets.Parameter)]
public class FromUserContextAttribute : GraphQLAttribute
{
    public override void Modify<TReturnType>(ArgumentInformation<TReturnType> argumentInformation)
        => argumentInformation.Expression = context => (TReturnType)context.UserContext;
}
```

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            if (parameterInfo.ParameterType != typeof(TReturnType))
+            {
+                throw new ArgumentOutOfRangeException(nameof(parameterInfo), $"Parameter must have a return type of {typeof(TReturnType).Name}.");
+            }
+            return parameterInfo;
+        }
+
+        /// <summary>
+        /// Gets or sets a delegate in the form of a <see cref="LambdaExpression"/> to be used to populate
```
sungam3r, line 155: > in the form of a \<see cref="LambdaExpression"/>

copy-pasted

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// this method argument while building the field resolver.
+        /// <br/><br/>
+        /// If not set, a query argument will be added to the field and the argument's value will be used
```
sungam3r, line 158:
```suggestion
        /// If not set, a query argument will be added to the field arguments list and the argument's value will be used
```

src/GraphQL/Types/ArgumentInformation.cs
Shane32: Yeah, still, it's accurate.  I can add an additional note.

Shane32: I'm going to rename `TReturnType` to `TArgumentType` also to better describe it.

Shane32: or maybe `TParameterType`

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -73,7 +74,85 @@ protected virtual IEnumerable<FieldType> ProvideFields()
         {
             var typeInformation = GetTypeInformation(memberInfo);
             var graphType = typeInformation.ConstructGraphType();
-            return AutoRegisteringHelper.CreateField(memberInfo, graphType, false);
+            var fieldType = AutoRegisteringHelper.CreateField(memberInfo, graphType, false);
+            if (memberInfo is MethodInfo methodInfo)
```
sungam3r, line 78:
```suggestion
            if (memberInfo is MethodInfo methodInfo && methodInfo.GetParameters().Length == 0)
```
\+ delete
```
 if (methodInfo.GetParameters().Length == 0)
            {
                return (null, MemberResolver.Create(methodInfo));
            }
```
from  `BuildMethodResolver`. Or keep as is since ` BuildMethodResolver` is protected and may be called by somebody else.

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,179 @@
...
+        public new Expression<Func<IResolveFieldContext, TReturnType>>? Expression
+        {
+            get => (Expression<Func<IResolveFieldContext, TReturnType>>?)base.Expression;
+            set => base.Expression = value;
+        }
+
+        /// <summary>
+        /// Applies <see cref="GraphQLAttribute"/> attributes pulled from the <see cref="ArgumentInformation.ParameterInfo">ParameterInfo</see> onto this instance.
+        /// </summary>
+        public virtual void ApplyAttributes()
```
sungam3r, line 170: May be moved into base type. Actually I think entire `ArgumentInformation<TReturnType>` can be eliminated. I'll add more ideas later.

Shane32: We can experiment with that in another PR.  I already had it that way, and changed it because it is so much more useful this way.  I'd like to merge it as-is, and if you want to try some changes, that's fine.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: > Or keep as is since BuildMethodResolver is protected and may be called by somebody else.

I would like to leave as-is.

src/GraphQL/Types/ArgumentInformation.cs
Shane32: Changed.

src/GraphQL/FromServicesAttribute.cs
Shane32: Basically, it hides the complexity for ANY custom attribute that desires to return an expression delegate.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -119,5 +197,21 @@ protected virtual TypeInformation GetTypeInformation(MemberInfo memberInfo)
...
+        /// <summary>
+        /// Analyzes a method argument and returns an instance of <see cref="TypeInformation"/>
+        /// containing information necessary to select a graph type. Nullable reference annotations
+        /// are read, if they exist, as well as the <see cref="RequiredAttribute"/> attribute.
+        /// Then any <see cref="GraphQLAttribute"/> attributes marked on the property are applied.
+        /// <br/><br/>
+        /// Override this method to enforce specific graph types for specific CLR types, or to implement custom
+        /// attributes to change graph type selection behavior.
+        /// </summary>
+        protected virtual TypeInformation GetTypeInformation(ParameterInfo parameterInfo)
```
sungam3r, line 210: Alas `ParameterInfo` and `MemberInfo` do not have common base class, but they inherit from `ICustomAttributeProvider` so `GetTypeInformation` may be  merged into `GetTypeInformation` above into new type case inside switch.

\+ protected virtual TypeInformation GetTypeInformation(MemberInfo memberInfo) -> protected virtual TypeInformation GetTypeInformation(ICustomAttributeProvider provider)

As you see fit.

src/GraphQL/FromServicesAttribute.cs
sungam3r: > But it made the code MUCH more complex and hard to use. 

Are you 100% sure? Your code or client code? When I looked through your code I had an impression that all that jumps around `TReturnType` are not nessessary. I will not say I'm sure 100%, but intuition tells me that.

src/GraphQL/Types/ArgumentInformation.cs
sungam3r: I'm not sure that I dare to change this code myself. Maybe. Also I can just forget about it :)

src/GraphQL/FromServicesAttribute.cs
sungam3r: Also could I use block expression here, i.e. 
```
context =>
{
  Console.WriteLine("afagagsg");
  return (TReturnType)context.UserContext;
}
```
?

docs2/site/docs/migrations/migration5.md
```diff
@@ -150,6 +150,63 @@ and data loader methods are supported such as shown in the below example:
...
 in the schema CLR type mappings.
 
+Arguments are also added as query arguments to the field, and recognize default values
+and other attributes set on the parameter. Certain argument types are recognized and
+treated special; special fields do not add a query argument to the field:
+
+| Argument type                     | Value passed to the method   |
+|-----------------------------------|------------------------------|
+| `IResolveFieldContext`            | The field resolver's context |
+| `CancellationToken`               | The cancellation token from the resolve context |
```
sungam3r, line 160: Maybe add a note that "injecting" `CancellationToken` may be a common case for methods that don't have dependencies from GraphQL.NET and "injecting" `IResolveFieldContext` brings such dependency and should be treated as "advanced" usage?

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Well, I could.  I would think that `MemberInfo` and `ParameterInfo` better indicate what the methods would support.  I think I'd prefer to leave it, but if you really think it would be better, consolidating the methods would be fine.

docs2/site/docs/migrations/migration5.md
```diff
@@ -137,7 +137,7 @@ Custom attributes can also be added to perform the following behavior changes:
 - Override detected nullability or list nullability state
 - Override chosen underlying graph type
 
-#### Method support for `AutoRegisteringObjectGraphType<T>` instances
+#### Method and argument support for `AutoRegisteringObjectGraphType<T>` instances
```
sungam3r, line 140: All this paragraph looks like a separate section of the main documentation.

docs2/site/docs/migrations/migration5.md
Shane32: Like: `The field resolvers's context; useful in advanced scenarios` ?

docs2/site/docs/migrations/migration5.md
Shane32: Yes, it is on my todo list to update the main documentation - see #2893 

We can delete it out of here if desired at that time.

Shane32: (I want to finish building some other relevant functionality first.)

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: I understand, so 50/50 😕 .

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
sungam3r: fine

docs2/site/docs/migrations/migration5.md
sungam3r: Something like that, yes. Techically we can inject any property from `IResolveFieldContext` but you added only `CancellationToken` . Why? Because its simplicity and common usage.

docs2/site/docs/migrations/migration5.md
sungam3r: 👍

src/GraphQL/FromServicesAttribute.cs
Shane32: Client code for custom attributes.

This is the equivalent to the above sample if we do not use generics:

```cs
[AttributeUsage(AttributeTargets.Parameter)]
public class FromUserContextAttribute : GraphQLAttribute
{
    public void Modify(ArgumentInformation argumentInformation)
    {
        var parameter = Expression.Parameter(typeof(IResolveFieldContext));
        var userContextExpression = Expression.MakeMemberAccess(parameter, typeof(IResolveFieldContext).GetProperty(nameof(IResolveFieldContext.UserContext)));
        var castExpression = Expression.Convert(userContextExpression, argumentInformation.ParameterInfo.ParameterType);
        var lambda = Expression.Lambda(castExpression, parameter);
        argumentInformation.Expression = lambda;
    }
}
```

And anything more than a simple property access gets even more complex.  But with generics, you can just write a lambda with C# code and it will compile it into an expression tree for you.

Shane32: Block expressions are not supported by the C# compiler, however, they are supported if you construct them manually.  I have done so before.

image

Shane32: Here is the method for building a block of code:

https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression.block?view=net-6.0#system-linq-expressions-expression-block(system-linq-expressions-expression())

You can also write loops and other stuff, if you have the patience for it.

Shane32: Typically it is easier to have the expression call a compiled method and then you write the block of code in there.

image

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: > 50/50

lol!  Let's just leave it then.

src/GraphQL/Types/ArgumentInformation.cs
Shane32: This is sorta the point, is that it's too complex to write expression trees manually.  If you look at my "before/after" code above, you will see how much simpler it is with generics.  Unfortunately the simplification comes at a cost.  But nobody wants to write expression trees.  None of my programmers want to, anyway.  And I wouldn't want to write a sample in our docs saying, "oh yeah you can write your own attribute.... but only if you know how to create expression trees".  It's just not for the average joe.

src/GraphQL/FromServicesAttribute.cs
sungam3r: Fast coded example:
```
 public interface IResolveFieldContext
    {

    }

    public class ArgumentInformation
    {
        public Expression<Func<IResolveFieldContext, object>> Expression1 { get; set; }

        public LambdaExpression Expression2 { get; set; }
    }




  var i = new ArgumentInformation
        {
            Expression1 = c => "aaa",
            Expression2 = (Expression<Func<IResolveFieldContext, object>>)(c => "aaa"),
        };
```

Just single explicit cast and you get the full power of C# compiler. 
изображение

So my idea is that there is no _actual_ need to manipulate `TReturnType`.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: ok

src/GraphQL/Types/ArgumentInformation.cs
sungam3r: I added an example of what I try to tell in other thread.

Shane32: So with the design I have, all the details of working with reflection and expressions are carefully tucked away where nobody should need to deal with them.  The methods and classes that deal with that are almost all `internal`; it "just happens".  And still you can easily write a custom attribute that contains code that will get embedded directly into a precompiled field resolver for that specific field.

If you look at each of my attributes that I've written, you will see that the code is extremely simple.  Anyone can copy an attribute and change it for their needs without knowing almost anything about reflection or expressions.

sungam3r: Let's continue in another conversation.

src/GraphQL/FromServicesAttribute.cs
Shane32: It can't return a `Func<IResolveFieldContext, object>`.  It must return a `Func<IResolveFieldContext, TParameterType>`.  Without `Modify` as a generic method, it is not possible to write the code you described.

docs2/site/docs/migrations/migration5.md
```diff
@@ -150,6 +150,63 @@ and data loader methods are supported such as shown in the below example:
 The above example assumes that the `Person` CLR type was mapped to `PersonGraphType`
 in the schema CLR type mappings.
 
+Arguments are also added as query arguments to the field, and recognize default values
+and other attributes set on the parameter. Certain argument types are recognized and
+treated special; special fields do not add a query argument to the field:
+
+| Argument type                     | Value passed to the method   |
+|-----------------------------------|------------------------------|
+| `IResolveFieldContext`            | The field resolver's context |
```
Shane32, line 159:
```suggestion
| `IResolveFieldContext`            | The field resolvers's context; useful in advanced scenarios |
```

docs2/site/docs/migrations/migration5.md
Shane32: Added as I mentioned above.

src/GraphQL/FromServicesAttribute.cs
Shane32: And I do not want to cast a returned `object` back to the proper type.  I want it to be typed correctly to begin with.

Shane32: At least with all this discussion I have fixed an incorrect NRT annotation :+1:

sungam3r: See #2914

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        /// <summary>
+        /// Sets a delegate to be used to populate this method argument while building the field resolver.
+        /// <br/><br/>
+        /// The delegate must be of the type
+        /// <see cref="Func{T, TResult}">Func</see>&lt;<see cref="IResolveFieldContext"/>, <typeparamref name="TParameterType"/>&gt;
+        /// where <typeparamref name="TParameterType"/> matches <see cref="ParameterInfo">ParameterInfo</see>.<see cref="ParameterInfo.ParameterType">ParameterType</see>.
+        /// </summary>
+        public void SetExpression<TParameterType>(Func<IResolveFieldContext, TParameterType?> argumentDelegate)
```
sungam3r, line 126: SetExpression -> SetDelegate ?

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -73,7 +74,85 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
+            }
+        }
+
+        /// <summary>
+        /// Analyzes a method parameter and returns an instance of <see cref="ArgumentInformation"/>
+        /// containing information necessary to build a <see cref="QueryArgument"/> and <see cref="IFieldResolver"/>.
+        /// Also applies any <see cref="GraphQLAttribute"/> attributes defined on the <see cref="ParameterInfo"/>
+        /// to the returned <see cref="ArgumentInformation"/> instance.
+        /// </summary>
+        protected virtual ArgumentInformation GetArgumentInformation<TReturnType>(FieldType fieldType, ParameterInfo parameterInfo)
```
sungam3r, line 150: So `TReturnType` always equals to `ParameterInfo.ParameterType` and we provide both.

src/GraphQL.Tests/Types/ArgumentInformationTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            info.SetExpression(context => 23);
+            info.Expression!.Compile().DynamicInvoke(new object?[] { null }).ShouldBeOfType<int>().ShouldBe(23);
+        }
+
+        [Fact]
+        public void SetExpression_Throws_For_Invalid_Type()
+        {
+            var info = new ArgumentInformation(_testParameterInfo, typeof(object), new FieldType(), new TypeInformation(_testParameterInfo));
+            var error = Should.Throw<ArgumentException>(() => info.SetExpression(context => "hello"));
+            error.Message.ShouldBe("Delegate must be of type Func<IResolveFieldContext, Int32>. (Parameter 'argumentDelegate')");
```
sungam3r, line 70:
```suggestion
            error.Message.ShouldStartWith("Delegate must be of type Func<IResolveFieldContext, Int32>");
```

Behaviour differs among TFMs.

src/GraphQL.Tests/Types/ArgumentInformationTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        [Fact]
+        public void SetExpression_Throws_For_Null()
+        {
+            var info = new ArgumentInformation(_testParameterInfo, typeof(object), new FieldType(), new TypeInformation(_testParameterInfo));
+            var error = Should.Throw<ArgumentNullException>(() => info.SetExpression<int>(null!));
+        }
+
+        private readonly MethodInfo _testMethodInfo = typeof(ArgumentInformationTests).GetMethod(nameof(TestMethod), BindingFlags.NonPublic | BindingFlags.Instance)!;
```
sungam3r, line 80: unused

src/GraphQL.Tests/Types/ArgumentInformationTests.cs
Shane32:
```suggestion
            error.Message.ShouldStartWith("Delegate must be of type Func<IResolveFieldContext, Int32>.");
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: I had renamed all the `TReturnType` to `TParameterType` but must have missed this one.  I'll rename this one also for clarity.

src/GraphQL/Types/ArgumentInformation.cs
Shane32: Sure

Shane32: Done

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: > So `TReturnType` always equals to `ParameterInfo.ParameterType` and we provide both.

Correct

src/GraphQL.Tests/Types/ArgumentInformationTests.cs
Shane32: Removed

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        /// <summary>
+        /// Sets a delegate to be used to populate this method argument while building the field resolver.
+        /// An expression is generated that calls <paramref name="argumentDelegate"/> and stored within <see cref="Expression"/>.
+        /// <br/><br/>
+        /// The delegate must be of the type
+        /// <see cref="Func{T, TResult}">Func</see>&lt;<see cref="IResolveFieldContext"/>, <typeparamref name="TParameterType"/>&gt;
+        /// where <typeparamref name="TParameterType"/> matches <see cref="ParameterInfo">ParameterInfo</see>.<see cref="ParameterInfo.ParameterType">ParameterType</see>.
+        /// </summary>
```
Shane32, line 126: Updated this comment also

Shane32: this part:
```cs
/// An expression is generated that calls <paramref name="argumentDelegate"/> and stored within <see cref="Expression"/>.
```

Shane32:
Always the stupid formatter
image

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -73,7 +74,85 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
+                    parameterInfo.ParameterType,
+                    queryArgument ?? throw new InvalidOperationException("Invalid response from ConstructQueryArgument: queryArgument and expression cannot both be null"));
+                expressions.Add(expression);
+            }
+            var resolver = new MethodResolver(methodInfo, expressions);
+            return (queryArguments, resolver);
+        }
+
+        private static readonly MethodInfo _getArgumentInformationInternalMethodInfo = typeof(AutoRegisteringObjectGraphType<TSourceType>).GetMethod(nameof(GetArgumentInformationInternal), BindingFlags.NonPublic | BindingFlags.Instance);
+        private ArgumentInformation GetArgumentInformationInternal<TReturnType>(FieldType fieldType, ParameterInfo parameterInfo)
```
sungam3r, line 127: One more TReturnType

sungam3r (approved):

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Changed

PR #2899, opened 2022-02-01
EnumerationGraphType rework + optimizations + embedded readme

sungam3r:
изображение

### DIFF
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -1685,30 +1685,50 @@ namespace GraphQL.Types
     }
     public class EnumValueDefinition : GraphQL.Utilities.MetadataProvider, GraphQL.Types.IProvideDeprecationReason, GraphQL.Types.IProvideDescription
     {
-        public EnumValueDefinition() { }
+        public EnumValueDefinition(string name, object? value) { }
         public string? DeprecationReason { get; set; }
         public string? Description { get; set; }
-        public string Name { get; set; }
-        public object? Value { get; set; }
+        public string Name { get; }
+        public object? Value { get; }
     }
-    public class EnumValues : System.Collections.Generic.IEnumerable<GraphQL.Types.EnumValueDefinition>, System.Collections.IEnumerable
+    public class EnumValues : GraphQL.Types.EnumValuesBase
     {
         public EnumValues() { }
-        public int Count { get; }
+        public override int Count { get; }
+        public override void Add(GraphQL.Types.EnumValueDefinition value) { }
+        public override GraphQL.Types.EnumValueDefinition? FindByName(GraphQLParser.ROM name) { }
+        public override GraphQL.Types.EnumValueDefinition? FindByValue(object? value) { }
+        public override System.Collections.Generic.IEnumerator<GraphQL.Types.EnumValueDefinition> GetEnumerator() { }
+    }
+    public abstract class EnumValuesBase : System.Collections.Generic.IEnumerable<GraphQL.Types.EnumValueDefinition>, System.Collections.IEnumerable
+    {
+        protected EnumValuesBase() { }
+        public abstract int Count { get; }
         public GraphQL.Types.EnumValueDefinition? this[string name] { get; }
-        public void Add(GraphQL.Types.EnumValueDefinition value) { }
-        public GraphQL.Types.EnumValueDefinition? FindByName(string name, System.StringComparison comparison = 5) { }
-        public GraphQL.Types.EnumValueDefinition? FindByValue(object? value) { }
-        public System.Collections.Generic.IEnumerator<GraphQL.Types.EnumValueDefinition> GetEnumerator() { }
+        public abstract void Add(GraphQL.Types.EnumValueDefinition value);
+        public abstract GraphQL.Types.EnumValueDefinition? FindByName(GraphQLParser.ROM name);
+        public abstract GraphQL.Types.EnumValueDefinition? FindByValue(object? value);
+        public abstract System.Collections.Generic.IEnumerator<GraphQL.Types.EnumValueDefinition> GetEnumerator();
+    }
+    public class EnumValues<TEnum> : GraphQL.Types.EnumValuesBase
+        where TEnum : System.Enum
+    {
+        public EnumValues() { }
+        public override int Count { get; }
+        public override void Add(GraphQL.Types.EnumValueDefinition value) { }
+        public override GraphQL.Types.EnumValueDefinition? FindByName(GraphQLParser.ROM name) { }
+        public override GraphQL.Types.EnumValueDefinition? FindByValue(object? value) { }
+        public override System.Collections.Generic.IEnumerator<GraphQL.Types.EnumValueDefinition> GetEnumerator() { }
     }
     public class EnumerationGraphType : GraphQL.Types.ScalarGraphType
     {
         public EnumerationGraphType() { }
-        public GraphQL.Types.EnumValues Values { get; }
+        public GraphQL.Types.EnumValuesBase Values { get; }
         public void AddValue(GraphQL.Types.EnumValueDefinition value) { }
         public void AddValue(string name, string? description, object? value, string? deprecationReason = null) { }
         public override bool CanParseLiteral(GraphQLParser.AST.GraphQLValue value) { }
         public override bool CanParseValue(object? value) { }
+        protected virtual GraphQL.Types.EnumValuesBase CreateValues() { }
         public override object? ParseLiteral(GraphQLParser.AST.GraphQLValue value) { }
         public override object? ParseValue(object? value) { }
         public override object? Serialize(object? value) { }
@@ -1719,6 +1739,7 @@ namespace GraphQL.Types
     {
         public EnumerationGraphType() { }
         protected virtual string ChangeEnumCase(string val) { }
+        protected override GraphQL.Types.EnumValuesBase CreateValues() { }
     }
     public class EventStreamFieldType : GraphQL.Types.FieldType
     {
--- a/src/GraphQL.Tests/Types/EnumGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/EnumGraphTypeTests.cs
@@ -72,9 +72,9 @@ namespace GraphQL.Tests.Types
         }
 
         [Fact]
-        public void adds_values_from_enum_custom_casing()
+        public void adds_values_from_enum_custom_casing_should_throw()
         {
-            type.ParseValue("rED").ShouldBe(Colors.Red);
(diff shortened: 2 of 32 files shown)


src/GraphQL.Tests/Types/EnumGraphTypeTests.cs
```diff
@@ -72,9 +72,9 @@ public void description_and_obsolete_from_enum()
         }
 
         [Fact]
-        public void adds_values_from_enum_custom_casing()
+        public void adds_values_from_enum_custom_casing_should_throw()
         {
-            type.ParseValue("rED").ShouldBe(Colors.Red);
+            Should.Throw<InvalidOperationException>(() => type.ParseValue("rED")).Message.ShouldBe("Unable to convert 'rED' to the scalar type 'Colors'");
```
sungam3r, line 77: @Shane32 I believe that enums should parse exact values since GraphQL is case-sensitive.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -36,11 +36,9 @@ public EnumerationGraphType()
         /// <param name="deprecationReason">The reason this enumeration member has been deprecated; <see langword="null"/> if this member has not been deprecated.</param>
         public void AddValue(string name, string? description, object? value, string? deprecationReason = null)
```
sungam3r, line 37: @Shane32 I find this method a bit problematic - I would move description as optional argument before deprecationReason.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -97,10 +97,14 @@ public void AddValue(EnumValueDefinition value)
         /// <inheritdoc/>
         public override object? Serialize(object? value)
         {
-            if (value == null) // TODO: why? null as internal value may be mapped to some external enumeration name
```
sungam3r, line 100: TODO fixed here but see comments below. I'm not sure. 
`ScalarGraphType.Serialize` says that
изображение

sungam3r: I think that actually enumeration graph type should always throw if it has no `null` internal value.

sungam3r: All other scalars return `null` 😕

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -217,14 +229,12 @@ public class EnumValues : IEnumerable<EnumValueDefinition>
...
-        /// Returns an enumeration definition for the specified value.
-        /// </summary>
-        public EnumValueDefinition? FindByValue(object? value)
+        /// <inheritdoc/>
+        public override EnumValueDefinition? FindByValue(object? value)
         {
             if (value is Enum)
             {
-                value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType())); //TODO: allocation, move work with enums into new generic class
+                value = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType())); //ISSUE:allocation
```
sungam3r, line 237: No more allocation for `EnumerationGraphType<TEnum>`

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -249,10 +302,19 @@ public class EnumValues : IEnumerable<EnumValueDefinition>
     [DebuggerDisplay("{Name}: {Value}")]
     public class EnumValueDefinition : MetadataProvider, IProvideDescription, IProvideDeprecationReason
     {
+        public EnumValueDefinition(string name, object? value)
+        {
+            Name = name;
```
sungam3r, line 307: All properties are get-only now.

src/GraphQL.Tests/Types/EnumGraphTypeTests.cs
Shane32: I agree

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
Shane32: I agree; just be sure we have applicable migration documentation.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
Shane32: None of the other built-in scalars have definable internal values.  Only custom scalars might map a null to something else.  An enum in theory could map null to MyEnum.None - but I don't mind if this situation isn't supported.

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -1685,30 +1685,50 @@ namespace GraphQL.Types
...
     public class EnumValueDefinition : GraphQL.Utilities.MetadataProvider, GraphQL.Types.IProvideDeprecationReason, GraphQL.Types.IProvideDescription
     {
-        public EnumValueDefinition() { }
+        public EnumValueDefinition(string name, object? value) { }
         public string? DeprecationReason { get; set; }
         public string? Description { get; set; }
-        public string Name { get; set; }
-        public object? Value { get; set; }
+        public string Name { get; }
+        public object? Value { get; }
```
Shane32, line 1692: I would like to have `EnumerationGraphType<T>` scan for `GraphQLAttribute` instances and apply them to `EnumValueDefinition`.  The attribues should be able to set the name and/or value, description, deprecation reason, and metadata on the definition.  As such, the `Name` property should not be read-only, since the `EnumValueDefinition` will be passed to the attribute class to be modified.  The `value` could be read-only, I guess.

We should be able to do this:

```cs
[Name("SortOrderEnum")]
public enum SortOrders
{
    [Description("Last edit date")]
    [Name("LAST_EDIT_DATE")]
    LastModifiedDate = 1,
}
```

src/GraphQL.Tests/Types/EnumGraphTypeTests.cs
```diff
@@ -134,6 +134,17 @@ public void serialize_by_name_throws()
             Should.Throw<InvalidOperationException>(() => type.Serialize("RED"));
         }
 
+        [Fact]
+        public void serialize_should_work_with_null_values()
```
Shane32, line 138: Looks good to me.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -97,10 +97,14 @@ public void AddValue(EnumValueDefinition value)
...
         public override object? Serialize(object? value)
         {
-            if (value == null) // TODO: why? null as internal value may be mapped to some external enumeration name
+            var foundByValue = Values.FindByValue(value);
+
+            // NOTE: for backward compatibility - keep or remove?
+            // See public void allow_null(Type graphType)
+            // g.Serialize(null).ShouldBeNull();
+            if (foundByValue == null && value == null)
                 return null;
```
Shane32, line 106: Pretty sure this is required for any field resolver that returns a nullable enum value.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
sungam3r: So what to do with `null` in case of enumerations?

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
sungam3r: OK

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
sungam3r: So I return `null` if no match found.

sungam3r: Added. Also I fixed TODO for `ToAST`.

Shane32:
Null support code looks good :+1:

src/GraphQL.Tests/Types/EnumGraphTypeTests.cs
Shane32: This change should be noted in the migration document.  It may be important.

Shane32: I'm wondering if we shouldn't have a documented workaround here.  If someone had a very large GraphQL client written that does not consistently use the correct case, it could be absolutely necessary to have a workaround.

sungam3r: I'll add migration notes but I would not invent workaround for such case. If someone had a very large GraphQL client he can post an issue.

Shane32: Ok. I would be prepared for such an issue.

sungam3r: done

src/GraphQL/Types/Scalars/Enumeration/EnumValues.cs
```diff
@@ -0,0 +1,131 @@
+using System.Collections;
+using GraphQLParser;
+
+namespace GraphQL.Types
+{
+    public abstract class EnumValuesBase : IEnumerable<EnumValueDefinition>
```
Shane32, line 6: There's three classes in this file.  Let's move each to its own file, now that we have a whole folder for enumeration stuff.

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
sungam3r: done

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
sungam3r: OK, `Name` is read/write now, see comment:
>  //TODO: important to set this property only BEFORE EnumValuesBase.AddValue called

src/GraphQL/Types/Scalars/Enumeration/EnumValues.cs
sungam3r: done

Shane32:
I'm doing a review now and just thinking that changing the order of `AddValue` might require a whole lot of rework for someone.  Worse yet, it could compile without changes and unknowingly affect someone who didn't read the whole migration document (or just missed a change).

Is it reasonable to to add a new member (e.g. `Add`) with the new argument order and remove `AddValue`?  We could even retain the old method with the old signature and mark it as obsolete.  We would of course leave it listed in the migration document.

Shane32:
Apart from that comment, I'm good with all the changes here.

sungam3r:
Yes, I know. Already changed method name.

Shane32 (approved):
So no obsolete AddValue method with old signature?

sungam3r:
Yes.

PR #2902, opened 2022-02-02
EnumerationGraphType supports GraphQLAttributes

Shane32:
@sungam3r You may wish to review or incorporate this during your enum rework

### DIFF
--- a/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
+++ b/src/GraphQL/Types/Scalars/EnumerationGraphType.cs
@@ -140,16 +140,35 @@ namespace GraphQL.Types
                 name: ChangeEnumCase(e.name),
                 value: Enum.Parse(type, e.name),
                 description: e.member.Description(),
-                deprecation: e.member.ObsoleteMessage()
+                deprecation: e.member.ObsoleteMessage(),
+                member: e.member
             ));
 
             Name = type.Name.ToPascalCase();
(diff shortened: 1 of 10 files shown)


src/GraphQL/Types/Scalars/EnumerationGraphType.cs
```diff
@@ -140,16 +140,41 @@ public EnumerationGraphType()
...
+                {
+                    Name = name,
+                    Value = value,
+                    Description = description,
+                    DeprecationReason = deprecation,
+                };
+                bool ignore = false;
+                foreach (var attr in member.GetCustomAttributes<GraphQLAttribute>())
+                {
+                    if (!attr.ShouldInclude(member, true))
```
sungam3r, line 163: true/false? null? for second argument

sungam3r (approved):

src/GraphQL/Types/Scalars/EnumerationGraphType.cs
Shane32: As a scalar is both input and output equally. Do you think I should make the property a nullable bool or just write in the xml comment that it’s true for enums?

Shane32: Changing to null.

sungam3r: ok

PR #2903, opened 2022-02-03
Update middleware sample

Shane32:
- Make the middleware a singleton
- Pulls `ISchema` via DI
- Middleware implements ASP.Net `IMiddleware` interface

### DIFF
--- a/src/GraphQL.Harness/Startup.cs
+++ b/src/GraphQL.Harness/Startup.cs
@@ -56,6 +56,7 @@ namespace GraphQL.Harness
             // add infrastructure stuff
             services.AddHttpContextAccessor();
             services.AddLogging(builder => builder.AddConsole());
+            services.AddSingleton<GraphQLMiddleware>();
 
             // add options configuration
             services.Configure<GraphQLSettings>(Configuration);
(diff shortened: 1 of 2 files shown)


src/GraphQL.Harness/Startup.cs
```diff
@@ -56,6 +56,7 @@ public void ConfigureServices(IServiceCollection services)
             // add infrastructure stuff
             services.AddHttpContextAccessor();
             services.AddLogging(builder => builder.AddConsole());
+            services.AddSingleton<GraphQLMiddleware>();
```
sungam3r, line 59: How did it work before? I mean ` app.UseMiddleware<GraphQLMiddleware>();` without `services.Add`?

sungam3r: All services from ctor and Invaoke are discovered and pulled by ASP.NET subsystem, right?

sungam3r (approved):

src/GraphQL.Harness/Startup.cs
Shane32: Previously, yes that was correct, as it was "ASP.Net Core convention-based middleware".  It was essentially a scoped/transient service, and DI services could be added to either the constructor or Invoke method.

With this change, the middleware is pulled from DI and it is called "ASP.Net Core factory-based middleware".

I think it's just a clearer sample.

sungam3r: What about middleware from server repo?

Shane32: I'll review when I look at the serialization

PR #2907, opened 2022-02-06
Change field resolvers and middleware to use ValueTask (part 1)

Shane32:
Waiting on:
- #2978

Notes:
- This will have substantial benefits for anyone using field middleware
- Should have neglible effect on performance otherwise (tbd)
- Async fields must use `FieldAsync` or async field builder methods
- This PR does NOT focus on adding `ValueTask` support to `FieldAsync` or similar methods, but rather the minimal set of changes required to provide a `ValueTask`-based execution pipeline with as little changes to user code as possible.  Changes and/or additions to field builder methods will be done in a subsequent PR.
- Similarly, this PR does NOT remove or change the usages of `AsyncFieldResolver` and `FuncFieldResolver`.  As such, it is not yet possible to use either class to provide a `ValueTask`-based delegate, even though `IFieldResolver` is `ValueTask`-based.  Such changes will be discussed and done in a separate PR.
- While this PR does update `IEventStreamResolver` to be `ValueTask`-based (and removes `IAsyncEventStreamResolver`), again it does NOT change or remove either `EventStreamResolver` or `AsyncEventStreamResolver` or change any field builder methods.  Such changes will be discussed and done in a separate PR.


Shane32:
```
before

|                  Method |   Stage |       Mean |     Error |    StdDev |   Gen 0 |  Gen 1 | Allocated |
|------------------------ |-------- |-----------:|----------:|----------:|--------:|-------:|----------:|
|           Introspection | Execute | 546.413 us | 3.3760 us | 3.1580 us | 24.4141 | 5.8594 |    151 KB |
|                         |         |            |           |           |         |        |           |
|                    Hero | Execute |   1.566 us | 0.0046 us | 0.0043 us |  0.2460 |      - |      2 KB |
|                         |         |            |           |           |         |        |           |
| IntrospectionMiddleware | Execute | 596.889 us | 1.6224 us | 1.3547 us | 34.1797 | 7.8125 |    210 KB |
|                         |         |            |           |           |         |        |           |
|          HeroMiddleware | Execute |   1.691 us | 0.0072 us | 0.0068 us |  0.2804 |      - |      2 KB |

after

|                  Method |   Stage |       Mean |     Error |    StdDev |   Gen 0 |  Gen 1 | Allocated |
|------------------------ |-------- |-----------:|----------:|----------:|--------:|-------:|----------:|
|           Introspection | Execute | 559.101 us | 2.0880 us | 1.9531 us | 24.4141 | 5.8594 |    151 KB |
|                         |         |            |           |           |         |        |           |
|                    Hero | Execute |   1.622 us | 0.0072 us | 0.0068 us |  0.2575 |      - |      2 KB |
|                         |         |            |           |           |         |        |           |
| IntrospectionMiddleware | Execute | 628.019 us | 3.8196 us | 3.5729 us | 24.4141 | 5.8594 |    151 KB |
|                         |         |            |           |           |         |        |           |
|          HeroMiddleware | Execute |   1.879 us | 0.0056 us | 0.0049 us |  0.2575 |      - |      2 KB |
```

Shane32:
@sungam3r what do you think:
- 3% slower
- no more allocations or memory use
- significant memory and allocation savings when using middleware
- cleaner code design
- supports valuetask throughout pipeline

Shane32:
Note: middleware benchmarks are with CountFieldMiddleware

Shane32:
Note: benchmarks are for execution stage only, not parsing, validation or serialization stages

Shane32:
@sungam3r Any idea from the logs why the build checks would hang?  Re-ran them and again they hung.

sungam3r:
Please use merge tool. It will show colored diff easy to read.

sungam3r:
> Any idea from the logs why the build checks would hang? Re-ran them and again they hung.

Obviously because of 'WIP' .

Shane32:
> Please use merge tool. It will show colored diff easy to read.

I don't use it enough to know how to use it the few times a year it may be beneficial.  I'm certainly not going to here!  You may copy and paste the report into the merge tool if you like.

Shane32:
> > Any idea from the logs why the build checks would hang? Re-ran them and again they hung.
> 
> Obviously because of 'WIP' .

After 4 hours and 41 minutes?  Could be but I don't think so.  I don't recall ever having this issue before.  Perhaps though.

sungam3r:
изображение

> I don't use it enough to know how to use it

From csproj:
изображение

sungam3r:
You can find it relly useful. Just compile and install one time and then type `merge` in terminal.

sungam3r:
> but I think you get the idea of the changes necessary here.

I did not look into yet. I reviewed other PRs. Is this one ready?

Shane32:
It is not ready to be merged, if that’s what you mean. But I am not going to do any more work on this PR unless you think it is worth pursuing. So I am awaiting your thoughts. 

I would endorse this goal of using ValueTask for field resolvers.

sungam3r:
OK, I'll reread and look through today or tomorrow.

src/GraphQL/Resolvers/IFieldResolver.cs
```diff
@@ -27,15 +27,15 @@ public interface IFieldResolver
         /// <summary>
         /// Returns an object, <see langword="null"/>, or a <see cref="Task{TResult}"/> for the specified field. If a task is returned, then this task will be awaited to obtain the actual object.
         /// </summary>
-        object? Resolve(IResolveFieldContext context);
+        ValueTask<object?> ResolveAsync(IResolveFieldContext context);
```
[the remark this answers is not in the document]
sungam3r: And (more important) all resolvers are async now so we can remove ~ 1/2 of all resolver classes. Right?

sungam3r:
What about `private Func<IResolveFieldContext, TReturnType?> GetResolverFor(Func<IResolveFieldContext<TSourceType>, TReturnType?> resolver)` ? And a more general question - what about returning "non-result types" i.e "async-container types" from resolvers? Example: `ValueTask<Task<int>>`. If I understand correctly those container types become forbidden by design so a lot of code like one from `GetResolverFor` should be rewritten/deleted. In other words all resolvers **always** return `ValueTask` as universal container. That container may provide completed or incompleted value. In first case we just take this value. In second case we may convert `ValueTask`-container into `Task`-container. But we never work with `Task` as value inside `ValueTask` container. In the proposed design it becomes unnecessary to do such things.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -342,7 +342,7 @@ public void ApplyMiddleware(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegat
 
                         var fieldMiddlewareDelegate = transform(context => inner.ResolveAsync(context));
 
-                        field.Resolver = new FuncFieldResolver<object>(fieldMiddlewareDelegate.Invoke);
+                        field.Resolver = new AsyncFieldResolver<object>(fieldMiddlewareDelegate.Invoke);
```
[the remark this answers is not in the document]
sungam3r: I mean that for convinience we of course should continue to provide `AddField` and etc. overloads with non-async delegates but under the hood all delegates will be "wrapped" into `ValueTask` container.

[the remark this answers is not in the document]
Shane32: Makes sense to me.

src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
```diff
@@ -28,7 +28,7 @@ public IFieldMiddlewareBuilder Use(Func<FieldMiddlewareDelegate, FieldMiddleware
             return this;
         }
 
-        private static readonly FieldMiddlewareDelegate _defaultDelegate = context => Task.FromResult(NameFieldResolver.Instance.Resolve(context));
```
sungam3r, line 31: Now I see that `object`->`Task<object>` "boxing" is eliminated saving memory here for middlewares. We always use `ValueTask` container across the codebase and so in most cases work with stack memory. In this regard behaviour of resolvers is the same as middlewares - both work with `ValueTask` so we can just passthrough results.

src/GraphQL.Tests/StarWars/StarWarsSubFieldsTest.cs
```diff
@@ -108,7 +108,7 @@ public void subfields_is_not_null_for_single_InterfaceGraphType()
                ctx.SubFields.ShouldNotBeNull();
                ctx.SubFields.Keys.ShouldContain("id");
                ctx.SubFields.Keys.ShouldContain("friends");
-               return null;
+               return Task.FromResult<object>(null);
```
sungam3r, line 111: I think that this PR may allow to cut some APIs from `Field/FieldAsync` family. I'm not sure.

Shane32:
> What about `private Func<IResolveFieldContext, TReturnType?> GetResolverFor(Func<IResolveFieldContext<TSourceType>, TReturnType?> resolver)` ?

I expect it would be considerably cleaned up and be changed to `private Func<IResolveFieldContext, ValueTask<TReturnType?>> GetResolverFor(Func<IResolveFieldContext<TSourceType>, ValueTask<TReturnType?>> resolver)`.  I would have to review it a bit more.

> And a more general question - what about returning "non-result types" i.e "async-container types" from resolvers? Example: `ValueTask<Task<int>>`.

That would not be allowed.  We might create three constructors for `FuncFieldResolver`, like this:
- `Func<IResolveFieldContext, TReturnType> delegate`
- `Func<IResolveFieldContext, Task<TReturnType>> delegate`
- `Func<IResolveFieldContext, ValueTask<TReturnType>> delegate`

Any delegate that returns a Task would be directly wrapped into a valuetask via the `ValueTask<TResult>(Task<TResult>)` constructor.  That is why it will be necessary for people to use the correct method to create the field -- they must use `FieldAsync` for async resolvers and `Field` for non-async resolvers.  You will see a few lines of the sample project changed because the sample was using `Field` for an async call instead of `FieldAsync`.

There will be some additional code necessary for the schema builder and auto-registering graph type classes where it will need to build the proper type of field resolver based on whether the method returns a Task or ValueTask or neither.  But this would only occur once while the schema is being built.

> If I understand correctly those container types become forbidden by design so a lot of code like one from `GetResolverFor` should be rewritten/deleted. In other words all resolvers **always** return `ValueTask` as universal container. That container may provide completed or incompleted value. In first case we just take this value. In second case we may convert `ValueTask`-container into `Task`-container. But we never work with `Task` as value inside `ValueTask` container. In the proposed design it becomes unnecessary to do such things.

Exactly! 🎉🎉🎉

src/GraphQL/Instrumentation/FieldMiddlewareBuilder.cs
Shane32: Exactly 🎉🎉🎉

src/GraphQL/Resolvers/IFieldResolver.cs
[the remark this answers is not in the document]
Shane32: > The main change that saves memory

Well, the main change that saves memory is due to the fact that middleware does not wrap synchronous resolvers with a Task.  Because we use ValueTask through the entire pipeline, there is no wrapping, as you already noticed.

src/GraphQL.Tests/StarWars/StarWarsSubFieldsTest.cs
sungam3r: Anyway I would like to switch all `Field/FieldAsync` overloads across code base to prefer `ValueTask` over `Task`. One could just use a simple conversion like `return new ValueTask<int>(my_Task_Of_Int_From_Some_External_Source)` or just `return await my_Task_Of_Int_From_Some_External_Source` to make transition from `Task` to `ValueTask`.

Shane32:
Let me know if your conclusion is that I should continue with this PR.

sungam3r:
Yes, go on. I propose to drop support for `Func<IResolveFieldContext, Task<TReturnType>> delegate`. Any who returns `Task<X>` may switch to ValueTask:
```
private static Task<int> ReturnResult() => ....


.Field("f", resolver: context => ReturnResult());  // Task overload which I propose to remove
->
.Field("f", resolver: async context => await ReturnResult()); // new ValueTask API, compiler "unwraps" value from `Task` and then "wraps" it again into `ValueTask`
```

I remember this comment https://github.com/graphql-dotnet/graphql-dotnet/issues/369#issuecomment-314830399 about 6 overloads for `Field`.

src/GraphQL/Resolvers/IFieldResolver.cs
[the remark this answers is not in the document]
sungam3r: We will not have any problems due to `ValueTask` design limitations (await once, etc)?

Shane32:
Ok. We can decide about the amount of support for Task to retain as this PR progresses. It may be that we cannot include support for Task even if we wanted to or else the compiler wouldn’t be able to determine which overload to use.

src/GraphQL/Resolvers/IFieldResolver.cs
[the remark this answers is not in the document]
Shane32: I though there was but later I found that ExecuteNodeAsync returns Task (not `Task<T>`) and so basically if it runs asynchronously then it gets wrapped at that point. Which is necessary for ParallelExecutionStrategy to WaitAll. 

I will review further of course. It may be that performance is better with another approach. My gut feeling is that if there’s ever a true async operation, any allocations necessary are moot, and probably necessary with ValueTask also; it is really the 99% of sync operations we are optimizing for.

Shane32:
For testing purposes, attempting only GraphQL.Tests project with .NET 6 TFM and detailed verbosity...

Shane32:
```
[xUnit.net 00:00:36.71] GraphQL.Tests: [Long Running Test] 'GraphQL.Tests.Types.AutoRegisteringObjectGraphTypeTests.ThrowsWhenSourceNull_Struct', Elapsed: 00:00:32
[Long Running Test] 'GraphQL.Tests.Execution.ResolveFieldContextTests.ExecutionError_Should_Be_Thread_Safe', Elapsed: 00:00:34
[Long Running Test] 'GraphQL.Tests.Subscription.SubscriptionWithReflectionTests.SubscribeWithArgumentAsync', Elapsed: 00:00:34
[Long Running Test] 'GraphQL.Tests.Subscription.SubscriptionTests.SubscribeWithArgumentAsync', Elapsed: 00:00:34
[Long Running Test] 'GraphQL.Tests.Bugs.Bug2635.test_parallel', Elapsed: 00:00:34
[Long Running Test] 'GraphQL.Tests.Execution.MultithreadedTests.test', Elapsed: 00:00:33
[Long Running Test] 'GraphQL.Tests.Execution.MutationTests.evaluates_mutations_correctly_in_the_presence_of_a_failed_mutation', Elapsed: 00:00:33
[Long Running Test] 'GraphQL.Tests.Subscription.SubscriptionWithAutoGraphTypeTests.SubscribeAsync', Elapsed: 00:00:30
[Long Running Test] 'GraphQL.Tests.Execution.Cancellation.CancellationTests.cancellation_is_propagated_async', Elapsed: 00:00:30
```

I'll have to look at this stuff tomorrow.

Shane32:
Seems like it's hanging while running those tests

Shane32:
Of course it could be just one of the tests that is crashing the rest or something; not sure yet.

Shane32:
Skipping these two tests prevents the test runner from hanging:
- `MutationTests.evaluates_mutations_correctly_in_the_presence_of_a_failed_mutation`
- `MutationTests.evaluates_datetime_mutations_correctly_in_the_presence_of_a_failed_mutation`

Shane32:
Executing them asynchronously also seems to solve the problem (instead of via GetAwaiter.GetResult)

src/GraphQL.Tests/Execution/MutationTests.cs
```diff
@@ -362,7 +362,7 @@ public void evaluates_mutations_correctly_in_the_presence_of_a_failed_mutation()
                   ""sixth"": null
                 }";
 
-            var result = AssertQueryWithErrors(query, expected, root: new Root(6, DateTime.Now), expectedErrorCount: 2);
+            var result = await AssertQueryWithErrorsAsync(query, expected, root: new Root(6, DateTime.Now), expectedErrorCount: 2);
```
Shane32, line 365: This test hangs/deadlocks within GitHub Actions if not run asynchronously.

src/GraphQL.Tests/Execution/MutationTests.cs
```diff
@@ -457,7 +457,7 @@ public void evaluates_datetime_mutations_correctly_in_the_presence_of_a_failed_m
...
```
Shane32, line 460: This test hangs/deadlocks within GitHub Actions if not run asynchronously.

src/GraphQL.Tests/xunit.runner.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
+  "diagnosticMessages": true,
+  "longRunningTestSeconds": 10
+}
```
Shane32, line 5: In the event of a hung test, this allows determination of which test(s) caused the hang.

src/GraphQL/Resolvers/IFieldResolver.cs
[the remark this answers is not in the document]
sungam3r2022: Bad news - I can not resolve opened conversations. 🤦‍♂️

src/GraphQL/Types/Collections/SchemaTypes.cs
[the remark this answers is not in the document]
sungam3r2022: This PR on another?

src/GraphQL.Tests/Execution/MutationTests.cs
sungam3r2022: I saw such an issue in graphql-client repo. XUnit has special constrained (6 workers by defaukt) syncronization context.

sungam3r2022:
Working with it.

src/GraphQL/Resolvers/IFieldResolver.cs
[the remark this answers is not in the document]
Shane32: Perhaps you simply cannot here because "sungam3r" opened this conversation?  In any case, just put a 👍 on a conversation and I'll resolve it.  I'm sure we can work through this and hopefully it's only temporary.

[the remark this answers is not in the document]
Shane32: Probably can't resolve any of my comments though either :-(

.github/workflows/test-code.yml
```diff
@@ -51,7 +51,6 @@ jobs:
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         working-directory: src
         run: |
-          dotnet tool install -g dotnet-format --version 6.0.243104 --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
```
sungam3r2022, line 54: You may also revert 6.0.102 to 6.0.x in all other workflows.

docs2/site/docs/migrations/migration5.md
```diff
@@ -802,3 +816,80 @@ You may use the following classes and methods as replacements:
...
+- `FieldConfig.AsyncSubscriber`
+
+Any direct implementation of these interfaces or classes derived from the above list will need to be modified to fit the new design.
+
+In addition, it is required that any asynchronous fields must use an appropriate asynchronous field builder method or
+asynchronous field resolver, and inferred methods (built by the schema builder, `FieldDelegate`, or `AutoRegisteringObjectGraphType`)
+must be strongly typed.
+
+```csharp
+// works in v4, not in v5
```
sungam3r2022, line 854:
```suggestion
// works in v4, not in v5 (throws in runtime)
```

?

docs2/site/docs/migrations/migration5.md
```diff
@@ -802,3 +816,80 @@ You may use the following classes and methods as replacements:
...
+{
+    Name = "hero",
+    Resolver = new FuncFieldResolver<Task<Droid>>(context => data.GetDroidByIdAsync("3")),
+});
+
+// works in v4 or v5
+AddField(new FieldType
+{
+    Name = "hero",
+    Resolver = new AsyncFieldResolver<Droid>(context => data.GetDroidByIdAsync("3")),
```
sungam3r2022, line 872: I would remove all `Async` prefixed from resolvers since the fact is that all resolvers are async by design. In further PR.

.github/workflows/test-code.yml
Shane32: So 6.0.x works properly as long as we don't try to install the dotnet-format tool (because it is already included) ?

src/GraphQL.MicrosoftDI.Tests/ConnectionBuilderExtensionTests.cs
```diff
@@ -39,7 +39,7 @@ public void WithScope0()
                 .Resolve()
                 .WithScope()
                 .Resolve(context => "hello");
-            field.Resolver.Resolve(_scopedContext).ShouldBe("hello");
+            field.Resolver.ResolveAsync(_scopedContext).Result.ShouldBe("hello");
```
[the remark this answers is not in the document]
sungam3r2022: If resolver body is async of course.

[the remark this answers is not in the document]
Shane32: I was thinking of changing all those to run asynchronously.  Some of the unchanged code also is running synchronously. 
 But the bigger problem is that in GraphQL.Tests, every time we use QueryTestBase, it uses `GetAwaiter().GetResult()` for every validation call.  So while here I only need to change two files and it will all be async, in the main project we would have a lot of changes to make.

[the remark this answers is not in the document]
Shane32: At the moment, I would like to merge this code as-is, without further complicating this PR.  I'm trying to keep this PR to a minimum of changes -- just get ValueTask in place first before continuing work.

.github/workflows/test-code.yml
sungam3r2022: Yes. Also see this https://github.com/dotnet/format/issues/1540

src/GraphQL.MicrosoftDI.Tests/ConnectionBuilderExtensionTests.cs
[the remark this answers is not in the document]
sungam3r2022: OK.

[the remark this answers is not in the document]
sungam3r2022: I even can't resolve conversations that I started.

.github/workflows/test-code.yml
sungam3r2022: > So 6.0.x works properly as long as we don't try to install the dotnet-format tool (because it is already included) ?

Sorry. I misunderstood your question.

sungam3r2022: It does ot matter we use `dotnet-fromat` or `dotnet format`. It always fails on 6.0.200. Either version works on 6.0.102.

Shane32: So I see...

src/GraphQL.MicrosoftDI.Tests/TestExtensions.cs
```diff
@@ -2,9 +2,9 @@ namespace GraphQL.MicrosoftDI.Tests
 {
     internal static class TestExtensions
     {
-        public static void ShouldBeTask(this object value, object expected)
+        public static void ShouldBeTask(this ValueTask<object> value, object expected)
         {
-            value.ShouldBeAssignableTo<Task<object>>().Result.ShouldBe(expected);
+            value.AsTask().Result.ShouldBe(expected);
```
sungam3r2022, line 7:
```suggestion
            value.Result.ShouldBe(expected);
```

Why not?

src/GraphQL.StarWars/StarWarsQuery.cs
```diff
@@ -9,16 +9,16 @@ public StarWarsQuery(StarWarsData data)
     {
         Name = "Query";
 
-        Field<CharacterInterface>("hero", resolve: context => data.GetDroidByIdAsync("3"));
-        Field<HumanType>(
+        FieldAsync<CharacterInterface>("hero", resolve: async context => await data.GetDroidByIdAsync("3"));
```
sungam3r2022, line 12: I think we should consider unify all `Field` overloads before v5. I would prefer use `Field` instead of `FieldAsync` but you chose `ResolveAsync` instead of `Resolve` for resolvers. In that way `FieldAsync` seems more logical though.

src/GraphQL.MicrosoftDI.Tests/TestExtensions.cs
Shane32: > The following operations should never be performed on a `ValueTask<TResult>` instance:

> 
> Awaiting the instance multiple times.
> Calling AsTask multiple times.
> Using .Result or .GetAwaiter().GetResult() when the operation hasn't yet completed, or using them multiple times.

See item 3 -- if it hasn't completed, we cannot call Result

Shane32: Per: https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=net-6.0

src/GraphQL.Tests/Execution/Cancellation/CancellationTests.cs
```diff
@@ -16,9 +16,9 @@ public CancellationTestType()
         {
             Name = "CancellationTestType";
 
-            Field<StringGraphType>("one", resolve: GetOneAsync);
-            Field<StringGraphType>("two", resolve: GetTwoAsync);
-            Field<StringGraphType>("three", resolve: GetThreeAsync);
+            FieldAsync<StringGraphType>("one", resolve: async context => await GetOneAsync(context));
```
sungam3r2022, line 19: Good examples of `Task<> -> ValueTask<>` conversions.

src/GraphQL.Tests/GraphQL.Tests.csproj
```diff
@@ -19,6 +19,16 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
```
sungam3r2022, line 22: revert changes?

src/GraphQL.StarWars/StarWarsQuery.cs
Shane32: I'm open to suggestions.  Here's my thoughts so far and we can discuss:
1. All methods that return `Task` or `ValueTask` should end in `Async`.  `ResolveAsync` returns a `ValueTask<object>`, so it must be `ResolveAsync` (or `SubscribeAsync`).
2. `Field` isn't an asynchronous method, so it would be appropriate to name them all `Field`.  But we need to consider (a) backwards compatibility, (b) we may need a way to differentiate the method overload for synchronous and asynchronous methods.
3. Are we supporting both `Task` asynchronous methods and `ValueTask` asynchronous methods?  If only one, which one?

I think we need to create a '[Discussion]' issue and discuss this further before a change is made.

Shane32: Also we need to remember that the generic type for `ValueTask` and `Task` are not covariant.  (not `out T`)  This can be a pain when assigning methods to field resolvers.

src/GraphQL.Tests/GraphQL.Tests.csproj
Shane32: I'd leave it - ok?  It's useful and only affects the test project.  Ideally this would be added into the Test.props file so if a test hangs in any test project, we can identify it without 3 hours of work (like I did this morning).

src/GraphQL.Tests/Utilities/Visitors/UppercaseDirectiveVisitor.cs
```diff
@@ -51,13 +46,11 @@ public override void VisitObjectFieldDefinition(FieldType field, IObjectGraphTyp
             if (applied != null)
             {
                 var inner = field.Resolver ?? NameFieldResolver.Instance;
-                field.Resolver = new AsyncFieldResolver<object>(async context =>
+                field.Resolver = new FuncFieldResolver<object>(async context =>
```
sungam3r2022, line 49: Compare UppercaseDirectiveVisitor with AsyncUppercaseDirectiveVisitor. It's time to remove the latter :)

src/GraphQL.Tests/xunit.runner.json
sungam3r2022: Still needed?

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -436,13 +436,7 @@ protected virtual async Task ExecuteNodeAsync(ExecutionContext context, Executio
                 resolveContext = resolveContext != null ? resolveContext.Reset(node, context) : new ReadonlyResolveFieldContext(node, context);
 
                 var resolver = node.FieldDefinition!.Resolver ?? NameFieldResolver.Instance;
-                var result = resolver.Resolve(resolveContext);
-
-                if (result is Task task)
```
sungam3r2022, line 441: 👍

sungam3r2022: Do we need `TaskExtensions` still?

Shane32: No. I will remove it.

src/GraphQL.Tests/Utilities/Visitors/UppercaseDirectiveVisitor.cs
Shane32: Done

src/GraphQL.Tests/xunit.runner.json
Shane32: I recommend leaving this to be able to diagnose future hung tests.

src/GraphQL.MicrosoftDI.Tests/TestExtensions.cs
sungam3r2022: Yeah, still need to read all that stuff.

src/GraphQL.StarWars/StarWarsQuery.cs
sungam3r2022: OK

src/GraphQL.Tests/GraphQL.Tests.csproj
sungam3r2022: OK, move into tests.pros; ` <None Remove="xunit.runner.json" />` may be removed.

src/GraphQL.Tests/xunit.runner.json
sungam3r2022: ok

sungam3r2022: Do I need to look into the dough logs, if something hangs?

src/GraphQL.Tests/GraphQL.Tests.csproj
Shane32: Done

src/GraphQL/Resolvers/NameFieldResolver.cs
```diff
@@ -92,5 +93,7 @@ Expression CreateAccessorExpression()
                 }
             }
         }
+
+        private static readonly PropertyInfo _sourcePropertyInfo = typeof(IResolveFieldContext).GetProperty(nameof(IResolveFieldContext.Source));
```
sungam3r2022, line 97: Repeated across codebase many times.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -101,7 +101,7 @@ protected void BuildFieldType(EventStreamFieldType fieldType, MemberInfo memberI
                 var resolver = new MemberResolver(propertyInfo, BuildMemberInstanceExpression(memberInfo));
                 fieldType.Arguments = null;
                 fieldType.Resolver = resolver;
-                fieldType.AsyncSubscriber = null;
+                fieldType.Subscriber = null;
```
sungam3r2022, line 104: I see xml comments mentioning `AsyncSubscriber`.

src/GraphQL/Resolvers/NameFieldResolver.cs
Shane32: Yeah, I was thinking that also.  If you have a suggestion of where to store it as an `internal` property or field, I can consolidate.  `IResolveFieldContext` is an interface, so I can't add it there.  Really shouldn't be more than a half dozen instances however, so I don't think it's a real big deal.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: I found & fixed one but couldn't find any others.  Lmk if you find any.

src/GraphQL.Tests/xunit.runner.json
Shane32: The GitHub Action log will start repeating stuff like this over and over if it hangs.  It will be obvious because it will show the test running for 6 hours (max time) before it terminates the process.

```
[xUnit.net 00:00:36.71] GraphQL.Tests: [Long Running Test] 'GraphQL.Tests.Types.AutoRegisteringObjectGraphTypeTests.ThrowsWhenSourceNull_Struct', Elapsed: 00:00:32
[Long Running Test] 'GraphQL.Tests.Execution.ResolveFieldContextTests.ExecutionError_Should_Be_Thread_Safe', Elapsed: 00:00:34
[Long Running Test] 'GraphQL.Tests.Subscription.SubscriptionWithReflectionTests.SubscribeWithArgumentAsync', Elapsed: 00:00:34
[Long Running Test] 'GraphQL.Tests.Subscription.SubscriptionTests.SubscribeWithArgumentAsync', Elapsed: 00:00:34
[Long Running Test] 'GraphQL.Tests.Bugs.Bug2635.test_parallel', Elapsed: 00:00:34
[Long Running Test] 'GraphQL.Tests.Execution.MultithreadedTests.test', Elapsed: 00:00:33
[Long Running Test] 'GraphQL.Tests.Execution.MutationTests.evaluates_mutations_correctly_in_the_presence_of_a_failed_mutation', Elapsed: 00:00:33
[Long Running Test] 'GraphQL.Tests.Subscription.SubscriptionWithAutoGraphTypeTests.SubscribeAsync', Elapsed: 00:00:30
[Long Running Test] 'GraphQL.Tests.Execution.Cancellation.CancellationTests.cancellation_is_propagated_async', Elapsed: 00:00:30
```

src/GraphQL/Resolvers/NameFieldResolver.cs
sungam3r2022: Not a real big deal. May be internal property in `AutoregisteringHelper`.

sungam3r2022 (approved):

PR #2906, opened 2022-02-04
Fix labeler

sungam3r:
fixes #2439

PR #2908, opened 2022-02-06
test

sungam3r:
Rel: #2906


sungam3r:
@Shane32 At least labeler works as before for PR from non-fork. Need to check PR from fork / other account.

PR #2910, opened 2022-02-06
Fix reader.Skip and add test

Shane32:


Shane32:
Didn't realize that the one serializer already pointed to the next token, so advancing it again was improper.  In either case, however, Skip automatically advances it if it's pointing to the property name, so do need for Read.

Shane32:
Thought I had a test for this, but it didn't catch it properly.

sungam3r:
I was also interested in calling `Read`, when I looked through code. I thought that since the tests passed, everything was fine and did not ask.

Shane32:
> I was also interested in calling `Read`, when I looked through code. I thought that since the tests passed, everything was fine and did not ask.

Correct for one but not the other.... but not needed for either.

PR #2914, opened 2022-02-07
Eliminate TParameterType

sungam3r:
@Shane32 I simplified (as I think, I may be wrong) code a bit and removed generic version of  `ArgumentInformation`.

### DIFF
--- a/src/GraphQL/FromServicesAttribute.cs
+++ b/src/GraphQL/FromServicesAttribute.cs
@@ -10,11 +10,11 @@ namespace GraphQL
     public class FromServicesAttribute : GraphQLAttribute
     {
         /// <inheritdoc/>
-        public override void Modify<TParameterType>(ArgumentInformation<TParameterType> argumentInformation)
-            => argumentInformation.Expression = context => GetRequiredService<TParameterType>(context);
+        public override void Modify(ArgumentInformation argumentInformation)
+            => argumentInformation.Expression = context => GetRequiredService(context, argumentInformation.ParameterInfo.ParameterType);
 
-        private static TServiceType GetRequiredService<TServiceType>(IResolveFieldContext context)
+        private static object GetRequiredService(IResolveFieldContext context, Type serviceType)
             => (context.RequestServices ?? throw new MissingRequestServicesException())
-                .GetRequiredService<TServiceType>();
+                .GetRequiredService(serviceType);
     }
 }
--- a/src/GraphQL/Types/ArgumentInformation.cs
+++ b/src/GraphQL/Types/ArgumentInformation.cs
@@ -7,18 +7,18 @@ namespace GraphQL.Types
     /// Contains information pertaining to a method parameter in preparation for buliding an
     /// expression or query argument for it.
     /// <br/><br/>
-    /// If <see cref="ArgumentInformation.Expression"/> is set, a query argument will not be added
+    /// If <see cref="Expression"/> is set, a query argument will not be added
     /// and the expression will be used to build the method resolver.
     /// <br/><br/>
     /// If not, a query argument will be generated and added to the field; the field resolver will
     /// use the argument's value to populate the method parameter.
     /// </summary>
-    public abstract class ArgumentInformation
+    public class ArgumentInformation
     {
         /// <summary>
         /// Initializes a new instance with the specified parameters.
         /// </summary>
-        protected ArgumentInformation(ParameterInfo parameterInfo, Type sourceType, FieldType fieldType, TypeInformation typeInformation, LambdaExpression? expression)
+        public ArgumentInformation(ParameterInfo parameterInfo, Type sourceType, FieldType fieldType, TypeInformation typeInformation, Expression<Func<IResolveFieldContext, object>>? expression)
         {
             ParameterInfo = parameterInfo ?? throw new ArgumentNullException(nameof(parameterInfo));
             FieldType = fieldType ?? throw new ArgumentNullException(nameof(fieldType));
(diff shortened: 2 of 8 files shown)


src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -69,30 +67,16 @@ protected ArgumentInformation(ParameterInfo parameterInfo, Type sourceType, Fiel
...
-            set
-            {
-                if (value != null && (value.ReturnType != ParameterInfo.ParameterType || value.Parameters.Count != 1 || value.Parameters[0].Type != typeof(IResolveFieldContext)))
-                {
-                    throw new ArgumentException($"Value must be a lambda expression delegate of type Func<IResolveFieldContext, {ParameterInfo.ParameterType.Name}>.");
-                }
-                _expression = value;
-            }
-        }
+        public Expression<Func<IResolveFieldContext, object>>? Expression { get; set; }
```
sungam3r, line 79: It may be changed back to `LambdaExpression` but it has no much sense. Note that there are no additional boxing introduced, only unboxing/convert from object to method parameter type - see MethodResolver.cs:38 
>  expressionBodies[i] = replaced.Type == methodParameters[i].ParameterType ? replaced : Expression.Convert(replaced, methodParameters[i].ParameterType);

It now happens when using `FromServicesAttribute` that sets `Expression` property. But since it returns some service from DI I think this service is ref type in 99,99% of cases.

Shane32: Only because FromServices happens to be a reference type.

Shane32:
I do not understand why you dislike TParameterType. I want to keep the expression strongly typed. I already explained this. Do not think I have not already explored this code design. I do not want to merge this code. 

It is possible to eliminate the generic ArgumentInfo, but the generic version of Modify must remain. There’s just no benefit to simplifying the design when it works so well.

Shane32:
You want to eliminate every allocation in the parser. Fine. I want to eliminate every possible allocation here. I want this code to run equally as fast as if you wrote a method by hand. No boxing for value types. No casting to object and back again. You of all people should not be complaining.

src/GraphQL/Types/ArgumentInformation.cs
sungam3r: Yes.

sungam3r:
> I do not understand why you dislike TParameterType

I dislike it because I see that it gives no benefit. I deleted this part of code and the remaining code has not worse characteristics. If I'm wrong, then what advantages are lost?

> I want this code to run equally as fast as if you wrote a method by hand. 

So am I.

>  No boxing for value types. No casting to object and back again.

Where? Be more concrete, please.

Shane32:
> I dislike it because I see that it gives no benefit. 

I've been very explicit about where the benefit is.  I will try to explain again below.

> I'm wrong, then what advantages are lost?

First, I would like the expression to be type-safe.  This is not the case with your code.  You can set the expression to return a string for an integer parameter and not only will the it happily accept that expression, at runtime there will be exception in a delegate compiled at runtime.

Second, I would like to eliminate casting to object and back again.  If the parameter type was a value type, this would cause boxing.  If not, it's simply slower.

> Where?

In custom attributes that work on arguments that have value types.

Shane32:
See:
- #2915

sungam3r:
> First, I would like the expression to be type-safe.

This is clear.

> You can set the expression to return a string for an integer parameter and not only will the it happily accept that expression; at runtime there will be exception in a delegate compiled at runtime.

Only in custom attributes. In addition, similar errors can be achieved with a generic version too.

> Second, I would like to eliminate casting to object and back again. If the parameter type was a value type, this would cause boxing.

Again only in custom attributes that return value type from `ArgumentInformation.Expression`. And since this is _expression_ data structure, top-level expression node should be `Convert` so code could check that fact, unwrap body from `Convert` and check underlying type. If it equals to parameter type then just use it, so no perfomance penalty. In this regard, work with expressions seems comfortable - we can cut "sharp corners".

I mean that we have 2 entry points to build expression. First - `AutoRegisteringHelper.GetParameterExpression`. Let me call it _standard_. Second - `ArgumentInformation.Expression`'s setter. Let me call it _custom_. First works good always building expression that has the right return type that will be used further to call desired method. And second may vary because caller has a bit more freedom. One can actually return any expression (you are right, type safety lost). Honestly, I do not see a big problem here. So in second case there is single additional step to wrap/unwrap the right expression. Yes, I missed double conversion. Fixed in the last commit.

Shane32:
I do not wish to pursue this discussion further.  I have made my reasons known.  I do not wish to utilize this code.

Shane32:
There is nothing more to be said.  Of course I already considered extracting the underlying function out, assuming that the outermost layer was a `Expression.Convert`.  Your code only works because `GetRequiredService` is also available in an overload which accepts a `Type` and returns an `object`.  It would not work if it only worked as a generic type.  I don't have the time to discuss this.  I do not wish to discuss it further. I have work to get done.

src/GraphQL/FromServicesAttribute.cs
```diff
@@ -10,11 +10,11 @@ namespace GraphQL
...
-        public override void Modify<TParameterType>(ArgumentInformation<TParameterType> argumentInformation)
-            => argumentInformation.Expression = context => GetRequiredService<TParameterType>(context);
+        public override void Modify(ArgumentInformation argumentInformation)
+            => argumentInformation.Expression = context => GetRequiredService(context, argumentInformation.ParameterInfo.ParameterType);
 
-        private static TServiceType GetRequiredService<TServiceType>(IResolveFieldContext context)
+        private static object GetRequiredService(IResolveFieldContext context, Type serviceType)
             => (context.RequestServices ?? throw new MissingRequestServicesException())
-                .GetRequiredService<TServiceType>();
+                .GetRequiredService(serviceType);
```
Shane32, line 18: Even your new code here is not type-safe.  It simply relies on the fact that `GetRequiredService` will return the proper type.

Shane32:
Suggest to close, replaced by #2916

PR #2915, opened 2022-02-07
Alternative approach to typed ArgumentInformation

Shane32:
This is the only alternative I would consider.  It reduces the amount of code by a mere 30 lines.  I don't see how this is a benefit to the project.

### DIFF
--- a/src/GraphQL/FromServicesAttribute.cs
+++ b/src/GraphQL/FromServicesAttribute.cs
@@ -10,11 +10,8 @@ namespace GraphQL
     public class FromServicesAttribute : GraphQLAttribute
     {
         /// <inheritdoc/>
-        public override void Modify<TParameterType>(ArgumentInformation<TParameterType> argumentInformation)
-            => argumentInformation.Expression = context => GetRequiredService<TParameterType>(context);
-
-        private static TServiceType GetRequiredService<TServiceType>(IResolveFieldContext context)
-            => (context.RequestServices ?? throw new MissingRequestServicesException())
-                .GetRequiredService<TServiceType>();
+        public override void Modify<TParameterType>(ArgumentInformation argumentInformation)
+            => argumentInformation.SetExpression(context => (context.RequestServices ?? throw new MissingRequestServicesException())
+                .GetRequiredService<TParameterType>());
     }
 }
--- a/src/GraphQL/GraphQLAttribute.cs
+++ b/src/GraphQL/GraphQLAttribute.cs
@@ -57,7 +57,7 @@ namespace GraphQL
         /// <summary>
         /// Updates the properties of the specified <see cref="ArgumentInformation{TParameterType}"/> as necessary.
         /// </summary>
-        public virtual void Modify<TParameterType>(ArgumentInformation<TParameterType> argumentInformation)
+        public virtual void Modify<TParameterType>(ArgumentInformation argumentInformation)
         {
         }
 
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -147,7 +147,7 @@ namespace GraphQL
     public class FromServicesAttribute : GraphQL.GraphQLAttribute
     {
         public FromServicesAttribute() { }
-        public override void Modify<TParameterType>(GraphQL.Types.ArgumentInformation<TParameterType> argumentInformation) { }
+        public override void Modify<TParameterType>(GraphQL.Types.ArgumentInformation argumentInformation) { }
     }
     public sealed class FuncServiceProvider : System.IServiceProvider
     {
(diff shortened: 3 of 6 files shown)


src/GraphQL/FromServicesAttribute.cs
```diff
@@ -10,11 +10,8 @@ namespace GraphQL
...
         /// <inheritdoc/>
-        public override void Modify<TParameterType>(ArgumentInformation<TParameterType> argumentInformation)
-            => argumentInformation.Expression = context => GetRequiredService<TParameterType>(context);
-
-        private static TServiceType GetRequiredService<TServiceType>(IResolveFieldContext context)
-            => (context.RequestServices ?? throw new MissingRequestServicesException())
-                .GetRequiredService<TServiceType>();
+        public override void Modify<TParameterType>(ArgumentInformation argumentInformation)
+            => argumentInformation.SetExpression(context => (context.RequestServices ?? throw new MissingRequestServicesException())
+                .GetRequiredService<TParameterType>());
```
Shane32, line 15: In order to be able to EASILY set a type-safe expression, without having to write `(Expression<Func<IResolveFieldContext, TParameterType>>)`, I added a generic method called `SetExpression`.  I also made this method accept a delegate, which is honestly more flexible in its use, as shown here where we can use null coalescing operators, code blocks, and so on.

src/GraphQL/GraphQLAttribute.cs
```diff
@@ -57,7 +57,7 @@ public virtual void Modify(TypeInformation typeInformation)
         /// <summary>
         /// Updates the properties of the specified <see cref="ArgumentInformation{TParameterType}"/> as necessary.
         /// </summary>
-        public virtual void Modify<TParameterType>(ArgumentInformation<TParameterType> argumentInformation)
+        public virtual void Modify<TParameterType>(ArgumentInformation argumentInformation)
```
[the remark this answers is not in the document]
sungam3r: I understand your intentions. Even with generic argument one can write invalid expression using invalid cast or something else.

src/GraphQL/FromServicesAttribute.cs
sungam3r: > to be able to EASILY set a type-safe expression

OK. I'm also want to end with the most easy solution. Consider `HardcodedValueAttribute` from my latest commit. One wants to return hardcoded value 85. So he should to cast 85 to `TParameterType` , right? Something like `context => (TParameterType)(object)85`. Do you mean that this syntax is simple? Type-safety is ephemeral here. In my variant the code looks like `context => 85`.

Shane32: Exactly -- you are hiding the inherit loss of type safety.  It should be explicit.

Shane32:
There is a third approach: we could move the call to the typed `Modify` inside the attribute.  Perhaps with a derived attribute class.

Shane32:
> There is a third approach: we could move the call to the typed `Modify` inside the attribute. Perhaps with a derived attribute class.

If you want I can add another PR with this approach demonstrated.

sungam3r:
If I understand correctly, then all 3 of the proposed options work. The difference is only as we understand _type safety_. More precisely, even as we **evaluate** _type safety_. I understand that you want to provide easy means to set expression in _type safe manner_. OK, good goal. I'm trying to convey the idea that this simplicity is actually ephemeral and in practice does not carry a tangible benefit, only complicating our own code jumping over generics. Also in some cases that _type safety_ complicates user code as in `HardcodedValueAttribute` example.

sungam3r:
> we could move the call to the typed Modify inside the attribute. 

Hmmm, `Modify` is already a method inseide attrubute, I don't get it.

Shane32:
If you say so, then that's what I want, yes.

> Also in some cases that type safety complicates user code as in HardcodedValueAttribute example.

Then we can look for another approach to that example.  Or let it be.

sungam3r:
By the way, is it already so critical that my variant of code is not so _type safe_? After all error will occur at schema initialization stage, i.e. on app startup. It is practically the same as when compiling, in a certain sense. I think, you understood what I mean. We have a lot of such checks.

src/GraphQL/FromServicesAttribute.cs
sungam3r: 😕  Then there should be unwrapping code for `Convert` expression in your variant for such cases if you really want to eliminate boxing and achive the same performance as with direct method call.

sungam3r:
> If you want I can add another PR with this approach demonstrated.

😄 I want to achive consensus.

Shane32:
- #2916

Shane32:
> 😄 I want to achive consensus.

I'm trying the best I can.  I do not wish to lose generics support or type safety because of "simplification".  I'm trying to offer alternatives.

Shane32:
Replaced by #2916

PR #2916, opened 2022-02-07
Another TParameterType variation

Shane32:
Allows without generics:

```cs
    internal class HardCodedExampleAttribute : GraphQLAttribute
    {
        public override void Modify(ArgumentInformation argumentInformation)
        {
            // 'int' is inferred, and if it is the improper type for the argument,
            // an exception is thrown immediately within SetExpression
            argumentInformation.SetExpression(x => 33);
        }
    }
```

or with generics:

```cs
    public class FromServicesAttribute : GraphQLAttribute
    {
        /// <inheritdoc/>
            => argumentInformation.SetExpression(context => (context.RequestServices ?? throw new MissingRequestServicesException())
                .GetRequiredService<TParameterType>());
    }
```

- Full type safety, at least to the extent we have control of


Shane32:
I like this option.

sungam3r:
Yes, looks like the best of "two worlds".

Let's add one test from https://github.com/graphql-dotnet/graphql-dotnet/pull/2914/files#diff-6b5d3a4ed2ed5ee010d067e6f8bf7c6490499b666276306827cbc78493e86315 and additional check for unwrapping `Convert` expression to eliminate boxing. Do you mind?

OK to merge into support_argument and make a final review.

Shane32:
Added test and added ability to unwrap convert expression.  Added tests to confirm that exception is thrown if it is not the correct type and so on.

PR #2920, opened 2022-02-08
Introduce IResolveFieldContext.Directives

### LINKED ISSUE #1981: Complex directive arguments
## Summary

I was struggling to create a directive with complex arguments, so I made some workaround extension methods that make it possible to extract complex directive arguments the same way you would extract arguments for fields using the IResolveFieldContext interface. 
The downside is that you have to provide the schema to these methods and they have poor performance because they try to find the type information about the directive arguments straight from the schema. Also, they haven't been tested extensively, but I hope that they help someone struggling with the same issues.

## Relevant information

Here is the code:

    namespace GraphQL
    {
        public static class DirectiveExtensions
        {
            public static bool HasDirective(this IResolveFieldContext context, string name) =>
                context.GetDirective(name) != null;
                
            public static bool TryGetDirective(
                this IResolveFieldContext context,
                string                    name,
                out Directive?            directive)
            {
                switch (context.GetDirective(name))
                {
                    case null:
                        directive = null;
                        return false;
        
                    case { } result:
                        directive = result;
                        return true;
                }
            }
        
            public static Directive? GetDirective(this IResolveFieldContext context, string name) =>
                context.FieldAst
                       .Directives
                      ?.Find(name);
        
        
            public static bool HasArgument(
                this Directive directive,
                string         name) =>
                directive.GetArgument(name) != null;
        
            public static TArgument? GetRefArgument<TArgument>(
                this Directive directive,
                ISchema        schema,
                string         name,
                TArgument?     defaultValue = null)
                where TArgument : class =>
                directive.TryGetArgument(schema, typeof(TArgument), name, out var result) ?
                    result == null ?
                        defaultValue :
                        result as TArgument :
                    defaultValue;
        
            public static TArgument GetValueArgument<TArgument>(
                this Directive directive,
                ISchema        schema,
                string         name,
                TArgument      defaultValue = default)
                where TArgument : struct =>
                directive.TryGetArgument(schema, typeof(TArgument), name, out var result) ?
                    (TArgument?) result ?? defaultValue :
                    defaultValue;
        
            public static object? GetArgument(
                this Directive directive,
                ISchema        schema,
                Type           argumentType,
                string         name,
                object?        defaultValue = null) =>
                directive.TryGetArgument(schema, argumentType, name, out var result) ?
                    result == null && argumentType.IsValueType ?
                        defaultValue :
                        result :
                    defaultValue;
        
            public static bool TryGetArgument(
                this Directive directive,
                ISchema        schema,
                Type           argumentType,
                string         name,
                out object?    result)
            {
                switch (directive.GetArgument(name))
                {
                    case null:
                        result = null;
                        return false;
        
        
                    case IDictionary<string, object> inputObject when argumentType == typeof(object):
                        result = inputObject;
                        return true;
        
                    case IDictionary<string, object> _ when argumentType.IsPrimitive():
                        throw new InvalidOperationException(
                            $"Could not read primitive type '{argumentType.FullName}' " +
                            $"from complex argument '{name}'");
        
                    case IDictionary<string, object> inputObject:
                        result = inputObject.ToObject(
                            argumentType,
                            schema.FindArgumentType(directive.Name, name));
        
                        return true;
        
        
                    case { } arg:
                        result = arg.GetPropertyValue(
                            argumentType,
                            schema.FindArgumentType(directive.Name, name));
        
                        return true;
                }
            }
        
        
            private static object? GetArgument(this Directive directive, string name) =>
                directive
                   .Arguments
                   .ValueFor(name)
                  ?.Value;
        
        
            private static IGraphType FindArgumentType(this ISchema schema, string directiveName, string argumentName) =>
                schema
                   .Directives
                   .FirstOrDefault(
                        directive =>
                            directive.Name == directiveName)
                  ?.Arguments
                   .FirstOrDefault(
                        argument =>
                            argument.Name == argumentName)
                  ?.ResolvedType ??
                throw new InvalidOperationException(
                    $"{directiveName} argument {argumentName} not found on {schema}.");
        }
    }

sungam3r:
fixes #1981 

Suddenly, it was quite simple to add support for directives.

### DIFF
--- a/src/GraphQL/Execution/DirectiveInfo.cs
+++ b/src/GraphQL/Execution/DirectiveInfo.cs
@@ -44,7 +44,7 @@ namespace GraphQL.Execution
         {
             bool exists = TryGetArgument(argumentType, name, out object? result);
             return exists
-                ? result ?? defaultValue
+                ? result
                 : defaultValue;
         }
 
--- a/src/GraphQL/Execution/DirectiveInfo.cs
+++ b/src/GraphQL/Execution/DirectiveInfo.cs
@@ -29,8 +29,7 @@ namespace GraphQL.Execution
 
         /// <summary>
         /// Returns the value of the specified directive argument, or <paramref name="defaultValue"/> when
-        /// unspecified or when specified as <see langword="null"/>. Field and variable default values take
-        /// precedence over the <paramref name="defaultValue"/> parameter.
+        /// unspecified. Variable default values take precedence over the <paramref name="defaultValue"/> parameter.
         /// </summary>
         public TType GetArgument<TType>(string name, TType defaultValue = default!)
         {
--- a/src/GraphQL/Execution/DirectiveInfo.cs
+++ b/src/GraphQL/Execution/DirectiveInfo.cs
@@ -36,7 +36,7 @@ namespace GraphQL.Execution
         {
             bool exists = TryGetArgument(typeof(TType), name, out object? result);
             return exists
-                ? result == null ? defaultValue : (TType)result
+                ? (TType)result
                 : defaultValue;
         }
 
--- a/src/GraphQL/Execution/DirectiveInfo.cs
+++ b/src/GraphQL/Execution/DirectiveInfo.cs
@@ -35,7 +35,7 @@ namespace GraphQL.Execution
         {
             bool exists = TryGetArgument(typeof(TType), name, out object? result);
             return exists
-                ? (TType)result
+                ? (TType)result!
                 : defaultValue;
         }
 
--- a/src/GraphQL/Execution/DirectiveInfo.cs
+++ b/src/GraphQL/Execution/DirectiveInfo.cs
@@ -1,0 +1,77 @@
+using GraphQL.Types;
+
+namespace GraphQL.Execution
+{
+    /// <summary>
+    /// Represents information about directive that has been provided in the GraphQL query request.
+    /// </summary>
+    public class DirectiveInfo
+    {
+        /// <summary>
+        /// Creates an instance of <see cref="DirectiveInfo"/> with the specified
+        /// directive definition and directive arguments.
+        /// </summary>
+        public DirectiveInfo(DirectiveGraphType directive, IDictionary<string, ArgumentValue> arguments)
+        {
+            Directive = directive;
+            Arguments = arguments;
+        }
+
+        /// <summary>
+        /// Directive definition.
+        /// </summary>
+        public DirectiveGraphType Directive { get; }
+
+        /// <summary>
+        /// Dictionary of directive arguments.
+        /// </summary>
+        public IDictionary<string, ArgumentValue> Arguments { get; }
+
+        /// <summary>
+        /// Returns the value of the specified directive argument, or <paramref name="defaultValue"/> when
+        /// unspecified or when specified as <see langword="null"/>. Field and variable default values take
+        /// precedence over the <paramref name="defaultValue"/> parameter.
+        /// </summary>
+        public TType GetArgument<TType>(string name, TType defaultValue = default!)
+        {
+            bool exists = TryGetArgument(typeof(TType), name, out object? result);
+            return exists
+                ? result == null ? defaultValue : (TType)result
+                : defaultValue;
+        }
+
+        /// <inheritdoc cref="GetArgument{TType}(string, TType)"/>
+        public object? GetArgument(Type argumentType, string name, object? defaultValue = null)
+        {
+            bool exists = TryGetArgument(argumentType, name, out object? result);
+            return exists
+                ? result ?? defaultValue
+                : defaultValue;
+        }
+
+        // initially copy-pasted from ResolveFieldContextExtensions.TryGetArgument and then modified
+        private bool TryGetArgument(Type argumentType, string name, out object? result)
+        {
+            if (Arguments == null || !Arguments.TryGetValue(name, out var arg))
+            {
+                result = null;
+                return false;
+            }
+
+            if (arg.Value is IDictionary<string, object?> inputObject)
+            {
+                if (argumentType == typeof(object))
+                {
+                    result = arg.Value;
+                    return true;
+                }
+
+                result = inputObject.ToObject(argumentType, Directive.Arguments?.Find(name)?.ResolvedType);
+                return true;
+            }
+
+            result = arg.Value.GetPropertyValue(argumentType, Directive.Arguments?.Find(name)?.ResolvedType);
+            return true;
+        }
+    }
+}
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -46,9 +46,7 @@ namespace GraphQL.Execution
         public static Dictionary<string, ArgumentValue>? GetArgumentValues(QueryArguments? definitionArguments, GraphQLArguments? astArguments, Variables? variables)
         {
             if (definitionArguments == null || definitionArguments.Count == 0)
-            {
                 return null;
-            }
 
             var values = new Dictionary<string, ArgumentValue>(definitionArguments.Count);
 
--- a/src/GraphQL/Execution/ExecutionHelper.cs
+++ b/src/GraphQL/Execution/ExecutionHelper.cs
@@ -1,3 +1,4 @@
+using System.Collections.ObjectModel;
 using GraphQL.Types;
 using GraphQL.Validation;
 using GraphQLParser.AST;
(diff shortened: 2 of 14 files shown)


src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -9,9 +10,34 @@ namespace GraphQL.Execution
     /// </summary>
     public static class ExecutionHelper
     {
+        private static readonly IDictionary<string, ArgumentValue> _emptyDirectiveArguments = new ReadOnlyDictionary<string, ArgumentValue>(new Dictionary<string, ArgumentValue>());
```
sungam3r, line 13: Question 1: use empty inner dictionary or null ?

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -9,9 +10,34 @@ namespace GraphQL.Execution
...
+            if (field.Directives == null || field.Directives.Count == 0)
+                return null;
+
+            var directives = new Dictionary<string, IDictionary<string, ArgumentValue>>(field.Directives.Count);
+
+            foreach (var dir in field.Directives.Items)
+            {
+                var dirDefinition = schema.Directives.Find(dir.Name);
+                if (dirDefinition == null)
+                    throw new ArgumentOutOfRangeException(nameof(field), $"Unknown directive '{dir.Name.StringValue}' for field '{field.Name.StringValue}'.");
```
sungam3r, line 30: Question 2: wording of error message.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
```diff
@@ -39,6 +39,16 @@ public interface IResolveFieldContext : IProvideUserContext
...
 
+        //TODO: change comments or/and add new methods
+
+        /// <summary>
+        /// A dictionary of directives with their arguments passed to the field, or <see langword="null"/> if no directives were defined for the field.
+        /// It is recommended to use the <see cref="ResolveFieldContextExtensions.GetArgument{TType}(IResolveFieldContext, string, TType)">GetArgument</see>
+        /// and <see cref="ResolveFieldContextExtensions.HasArgument(IResolveFieldContext, string)">HasArgument</see> extension
+        /// methods rather than this dictionary, so the names can be converted by the selected <see cref="INameConverter"/>.
+        /// </summary>
+        IDictionary<string, IDictionary<string, ArgumentValue>>? Directives { get; }
```
sungam3r, line 50: Question 3: method signature 
`IDictionary<string, IDictionary<string, ArgumentValue>>?` or `IDictionary<string, IDictionary<string, ArgumentValue>?>?` or maybe something else?

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
```diff
@@ -39,6 +39,16 @@ public interface IResolveFieldContext : IProvideUserContext
         /// </summary>
         IDictionary<string, ArgumentValue>? Arguments { get; }
 
+        //TODO: change comments or/and add new methods
+
+        /// <summary>
+        /// A dictionary of directives with their arguments passed to the field, or <see langword="null"/> if no directives were defined for the field.
+        /// It is recommended to use the <see cref="ResolveFieldContextExtensions.GetArgument{TType}(IResolveFieldContext, string, TType)">GetArgument</see>
```
sungam3r, line 46: Question 4: Add methods to get directive arguments as well? I guess yes. So what name to choose?

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: I assume that validation rules should validate that directives are valid before executing the request?  If not, we should probably add that code.  Assuming that it does (or if we add a validation rule), maybe this would be better here:

```cs
if (dirDefinition == null)
    continue;
```

Then if someone purposely removed the validation rule, it would ignore unknown directives while executing the request.  I think that would be better.

Also, if by default the added validation rule is not part of the default set of validation rules, again I would think it would be better to skip unknown directives rather than throw an exception here.  For example:

```gql
{
  fieldWith2Literal @undefinedDirective @include(if: true)
  fieldWithoutDirectives @undefinedDirective
}
```

In the example above, executing `fieldWithoutDirectives` would not throw an exception assuming the field resolver does not attempt to read the directives at all, and `fieldWith2Literal` would always throw an exception, assuming the code was attempting to inspect the `include` directive.

So I think this code should ignore unknown directives and leave it to the validation rules to throw errors for unknown directives, if so desired.

src/GraphQL/Execution/ExecutionHelper.cs
Shane32: 50/50.  I'd say empty.  If someone wants to read arguments, it will save a null check.  If they don't attempt to read arguments, having the reference won't negatively affect anything.

src/GraphQL/Execution/ExecutionHelper.cs
```diff
@@ -9,9 +10,34 @@ namespace GraphQL.Execution
...
+        private static readonly IDictionary<string, ArgumentValue> _emptyDirectiveArguments = new ReadOnlyDictionary<string, ArgumentValue>(new Dictionary<string, ArgumentValue>());
+
+        /// <summary>
+        /// Returns a dictionary of directives with their arguments values for a field.
+        /// Values will be retrieved from literals or variables as specified by the document.
+        /// </summary>
+        public static IDictionary<string, IDictionary<string, ArgumentValue>>? GetDirectives(GraphQLField field, Variables? variables, ISchema schema)
+        {
+            if (field.Directives == null || field.Directives.Count == 0)
+                return null;
```
Shane32, line 22: We could return an empty list here also.  It just saves a null check later at the cost of looking for a directive inside an empty list.  50-50 on this as well.  What you have is fine, as it matches `IResolveFieldContext.Arguments`, I believe.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
Shane32: Have you thought of returning a `DirectiveInfo` or similar class like this?

```cs
class DirectiveInfo : Dictionary<string, ArgumentValue>
{
    public DirectiveDefinition DirectiveDefinition { get; }
    public object GetArgument(...);
}
```

Then on `IResolveFieldContext` we have:

```cs
IDictionary<string, DirectiveInfo>? Directives { get; }
DirectiveInfo? GetDirective(string directiveName);
```

Now you can write code like this:

```cs
if (context.GetDirective("include")?.GetArgument<bool>("if") ?? false)
{
    // code
}
```

And the names are all run through the name converter.

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: Ok

src/GraphQL/Execution/ExecutionHelper.cs
sungam3r: Agree, done.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
sungam3r: > And the names are all run through the name converter.

I think no. NameCoverter works for fields and fields arguments. Directive has no fields and directive is not a graph type.

sungam3r: Thanks, done.

src/GraphQL/ResolveFieldContext/IResolveFieldContext.cs
```diff
@@ -39,6 +39,14 @@ public interface IResolveFieldContext : IProvideUserContext
         /// </summary>
         IDictionary<string, ArgumentValue>? Arguments { get; }
 
+        /// <summary>
+        /// A dictionary of directives with their arguments passed to the field, or <see langword="null"/> if no directives were defined for the field.
+        /// It is recommended to use the <see cref="ResolveFieldContextExtensions.GetDirective(IResolveFieldContext, string)">GetDirective</see>
+        /// and <see cref="ResolveFieldContextExtensions.HasDirective(IResolveFieldContext, string)">HasDirective</see> extension
+        /// methods rather than this dictionary directly.
```
sungam3r, line 46: " so the names can be converted by the selected <see cref="INameConverter"/" was removed

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
```diff
@@ -42,6 +44,9 @@ internal ReadonlyResolveFieldContext Reset(ExecutionNode? node, ExecutionContext
         private IDictionary<string, ArgumentValue>? GetArguments()
             => ExecutionHelper.GetArgumentValues(_executionNode.FieldDefinition!.Arguments, _executionNode.Field!.Arguments, _executionContext.Variables);
```
sungam3r, line 45: Maybe rename to `GetArguments`?

src/GraphQL/Execution/DirectiveInfo.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <summary>
+        /// Returns the value of the specified directive argument, or <paramref name="defaultValue"/> when
+        /// unspecified or when specified as <see langword="null"/>. Field and variable default values take
+        /// precedence over the <paramref name="defaultValue"/> parameter.
+        /// </summary>
+        public TType GetArgument<TType>(string name, TType defaultValue = default!)
+        {
+            bool exists = TryGetArgument(typeof(TType), name, out object? result);
+            return exists
+                ? result == null ? defaultValue : (TType)result
```
Shane32, line 39:
```suggestion
                ? (TType)result
```

sungam3r: ok

src/GraphQL/Execution/DirectiveInfo.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        /// <summary>
+        /// Dictionary of directive arguments.
+        /// </summary>
+        public IDictionary<string, ArgumentValue> Arguments { get; }
+
+        /// <summary>
+        /// Returns the value of the specified directive argument, or <paramref name="defaultValue"/> when
+        /// unspecified or when specified as <see langword="null"/>. Field and variable default values take
+        /// precedence over the <paramref name="defaultValue"/> parameter.
```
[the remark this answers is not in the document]
sungam3r: > Do field default values take precedence?

Why not? The same code from `GetArgument`. Directive arguments can have default values for both simple args and complex args.

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
sungam3r: done

Shane32 (commented):
Code looks good.  Let's document the new feature in the 'new features' section of the migration guide.

src/GraphQL/Execution/DirectiveInfo.cs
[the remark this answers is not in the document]
sungam3r: Code from `ToObject`/`GetPropertyValue`?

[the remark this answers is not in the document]
sungam3r: Maybe. The task for another PR.

sungam3r:
Agree.

PR #2922, opened 2022-02-08
Cut new expression

sungam3r:


sungam3r:
Only `new BlaBla()` -> `new()` changes here.

PR #2924, opened 2022-02-08
NRT annotations for GraphQL.MicrosoftDI + cleanup

Shane32:


src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -130,7 +130,7 @@ private FieldBuilder(EventStreamFieldType fieldType)
         }
 
         /// <inheritdoc cref="Resolve(IFieldResolver)"/>
-        public virtual FieldBuilder<TSourceType, TReturnType> Resolve(Func<IResolveFieldContext<TSourceType>, TReturnType> resolve)
+        public virtual FieldBuilder<TSourceType, TReturnType> Resolve(Func<IResolveFieldContext<TSourceType>, TReturnType?> resolve)
```
sungam3r, line 133: If generic is marked as `TReturnType` here an I use `string` then it will be `string?`. But what if I use `string?`. Would it be `string??` ?

sungam3r (approved):

src/GraphQL/Builders/FieldBuilder.cs
Shane32: Well, it's still a `string?`.  Note that if `TReturnType` is `int`, then it's not `int?`.  The `?` here only applies to reference types (because there is no `where TReturnType : struct` clause).

PR #2927, opened 2022-02-08
Add ExecutionStrategySelector

Shane32:
Todo:
- [x] Add framework
- [x] Add migration notes
- [x] Add docs
- [x] Add tests
- [x] Consider removing SubscriptionDocumentExecuter, or marking as obsolete - marked obsolete
- [x] Consider removing SelectExecutionStrategy, or marking as obsolete - left as-is

### DIFF
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -105,7 +105,7 @@ with the underlying dependency injection framework. The `Services` property retu
 
 | Method         | Description |
 |----------------|-------------|
-| `Register`     | Registers a service within the DI framework |
+| `Register`     | Registers a service within the DI framework replacing existing registration if needed |
 | `TryRegister`  | Registers a service within the DI framework if it has not already been registered |
 
 To use the `AddGraphQL` method, you will need to install the proper nuget package for your DI provider. See list below:
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -45,14 +45,14 @@ How you integrate this into your system will depend on the dependency injection 
 ## Dependency Injection Registration Helpers
 
 GraphQL.NET provides an `IGraphQLBuilder` interface which encapsulates the configuration methods of a dependency injection framework, to provide an
-abstract method of configuring a dependency injection framework to work with GraphQL.NET. This interface is returned from a DI-provider-specific
-setup method (typically called `AddGraphQL()`), at which point you can call extension methods on the interface to configure this library. A simple
-example is below:
+abstract method of configuring a dependency injection framework to work with GraphQL.NET. This interface is provided through a configuration delegate
+from a DI-provider-specific setup method (typically called `AddGraphQL()`), at which point you can call extension methods on the interface to
+configure this library. A simple example is below:
 
 ```csharp
-services.AddGraphQL()
+services.AddGraphQL(builder => builder
     .AddSystemTextJson()
-    .AddSchema<MySchema>();
+    .AddSchema<MySchema>());
 ```
 
 The interface also allows configuration of the schema during initialization, and configuration of the execution at runtime. In this manner, adding
@@ -61,12 +61,13 @@ configuration.
 
 The `AddGraphQL()` method will register default implementations of the following services within the dependency injection framework:
 
-* `IDocumentExecuter` - which does not support subscriptions
+* `IDocumentExecuter`
 * `IDocumentBuilder`
 * `IDocumentValidator`
 * `IComplexityAnalyzer` - which is not used unless configured within `ExecutionOptions`
 * `IErrorInfoProvider`
 * `IDocumentCache` - an implemenation which does not cache documents
+* `IExecutionStrategySelector` - which does not support subscriptions by default
 
 A list of the available extension methods is below:
 
@@ -79,6 +80,8 @@ A list of the available extension methods is below:
 | `AddDocumentExecuter<>` | Registers the specified document executer; useful when needed to change the execution strategy utilized | |
 | `AddDocumentListener<>` | Registers the specified document listener and configures execution to use it | |
 | `AddErrorInfoProvider`  | Registers a custom error info provider or configures the default error info provider | |
+| `AddExecutionStrategy`  | Registers an `ExecutionStrategyRegistration` for the selected execution strategy and operation type | |
+| `AddExecutionStrategySelector` | Registers the specified execution strategy selector | |
 | `AddGraphTypes`         | Scans the specified assembly for graph types and registers them within the DI framework | |
 | `AddMemoryCache`        | Registers the memory document cache and configures its options | GraphQL.MemoryCache |
 | `AddMetrics`            | Registers and enables metrics depending on the supplied arguments | |
@@ -87,17 +90,23 @@ A list of the available extension methods is below:
 | `AddSchema<>`           | Registers the specified schema | |
 | `AddSelfActivatingSchema<>` | Registers the specified schema which will create instances of unregistered graph types during initialization | |
 | `AddSerializer<>`       | Registers the specified serializer | |
-| `AddSubscriptionDocumentExecuter` | Registers the document executer that has subscription support | GraphQL.SystemReactive | |
+| `AddSubscriptionExecutionStrategy` | Registers an `ExecutionStrategyRegistration` for subscription support | GraphQL.SystemReactive | |
 | `AddSystemTextJson`     | Registers the document writer that uses System.Text.Json as its underlying JSON serialization engine | GraphQL.SystemTextJson |
 | `AddValidationRule<>`   | Registers the specified validation rule and configures it to be used at runtime | |
 | `ConfigureExecutionOptions` | Configures execution options at runtime | |
 | `ConfigureSchema`       | Configures schema options when the schema is initialized | |
 | `Configure<TOptions>`   | Used by extension methods to configures an options class within the DI framework | |
-| `Register`              | Used by extension methods to register services within the DI framework | |
-| `TryRegister`           | Used by extension methods to register services within the DI framework when they have not already been registered | |
 
 The above methods will register the specified services typically as singletons unless otherwise specified. Graph types and middleware are registered
 as transients so that they will match the schema lifetime. So with a singleton schema, all services are effectively singletons.
+
+Custom `IGraphQLBuilder` extension methods typically rely on the `Services` property of the builder in order to register services
+with the underlying dependency injection framework. The `Services` property returns a `IServiceRegister` interface which has these methods:
+
+| Method         | Description |
+|----------------|-------------|
+| `Register`     | Registers a service within the DI framework |
+| `TryRegister`  | Registers a service within the DI framework if it has not already been registered |
 
 To use the `AddGraphQL` method, you will need to install the proper nuget package for your DI provider. See list below:
 
@@ -161,10 +170,9 @@ Note that if any of the graph types directly or indirectly implement `IDisposabl
 injection provider, or their `Dispose` methods will not be called. Any dependencies of graph types that implement
 `IDisposable` will be disposed of properly, regardless of whether the graph type is registered within the service provider.
 
-You can also use the `services.AddGraphTypes()` extension method from the [server](https://github.com/graphql-dotnet/server)
-project to scan the calling assembly for classes that implement `IGraphType` and register them all as singletons
-within the service provider. Mark your class with `DoNotRegisterAttribute` if you want to skip registration.
-For additional options and overloads of this method, see [GraphQLBuilderCoreExtensions.cs](https://github.com/graphql-dotnet/server/blob/master/src/Core/Extensions/GraphQLBuilderCoreExtensions.cs).
+You can also use the `.AddGraphTypes()` builder method to scan the calling or specified assembly for classes that implement
+`IGraphType` and register them all as transients within the service provider. Mark your class with `DoNotRegisterAttribute` if you
+want to skip registration.
 
 ## Nancy TinyIoCContainer
 
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -86,12 +86,12 @@ A list of the available extension methods is below:
 | `AddMemoryCache`        | Registers the memory document cache and configures its options | GraphQL.MemoryCache |
 | `AddMetrics`            | Registers and enables metrics depending on the supplied arguments | |
 | `AddMiddleware<>`       | Registers the specified middleware and configures it to be installed during schema initialization | |
-| `AddNewtonsoftJson`     | Registers the document writer that uses Newtonsoft.Json as its underlying JSON serialization engine | GraphQL.NewtonsoftJson |
+| `AddNewtonsoftJson`     | Registers the serializer that uses Newtonsoft.Json as its underlying JSON serialization engine | GraphQL.NewtonsoftJson |
 | `AddSchema<>`           | Registers the specified schema | |
 | `AddSelfActivatingSchema<>` | Registers the specified schema which will create instances of unregistered graph types during initialization | |
 | `AddSerializer<>`       | Registers the specified serializer | |
 | `AddSubscriptionExecutionStrategy` | Registers an `ExecutionStrategyRegistration` for subscription support | GraphQL.SystemReactive | |
-| `AddSystemTextJson`     | Registers the document writer that uses System.Text.Json as its underlying JSON serialization engine | GraphQL.SystemTextJson |
+| `AddSystemTextJson`     | Registers the serializer that uses System.Text.Json as its underlying JSON serialization engine | GraphQL.SystemTextJson |
 | `AddValidationRule<>`   | Registers the specified validation rule and configures it to be used at runtime | |
 | `ConfigureExecutionOptions` | Configures execution options at runtime | |
 | `ConfigureSchema`       | Configures schema options when the schema is initialized | |
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -344,6 +344,77 @@ Field<StringGraphType>("myField", resolve: context =>
 });
 ```
 
+### 12. The `ExecutionStrategy` selected for an operation can be configured through `IGraphQLBuilder`
+
+Previously, in order to change the execution strategy for a specific operation -- for instance,
+using a serial execution strategy for 'query' operation types -- required creating a custom
+document executer and overriding the `SelectExecutionStrategy` method.
+
+Now, for DI configurations, you can call the `.AddExecutionStrategy<T>(OperationType)` method to
+provide this configuration without overriding the method. See below for an example.
+
+```csharp
+// === GraphQL.NET v4 ===
+public class SerialDocumentExecuter : DocumentExecuter
+{
+    public SerialDocumentExecuter(IDocumentBuilder documentBuilder, IDocumentValidator documentValidator, IComplexityAnalyzer complexityAnalyzer, IDocumentCache documentCache, IEnumerable<IConfigureExecutionOptions> configurations)
+        : base(documentBuilder, documentValidator, complexityAnalyzer, documentCache, configurations)
+    {
+    }
+
+    protected override IExecutionStrategy SelectExecutionStrategy(ExecutionContext context)
+    {
+        return context.Operation.Operation switch
+        {
+            OperationType.Query => SerialExecutionStrategy.Instance,
+            _ => base.SelectExecutionStrategy(context)
+        };
+    }
+}
+
+// within Startup.cs
+services.AddGraphQL()
+    .AddSystemTextJson()
+    .AddSchema<StarWarsSchema>()
+    .AddDocumentExecuter<SerialDocumentExecuter>();
+
+
+// === GraphQL.NET v5 ===
+
+// within Startup.cs
+services.AddGraphQL(builder => builder
+    .AddSystemTextJson()
+    .AddSchema<StarWarsSchema>()
+    .AddExecutionStrategy<SerialExecutionStrategy>(OperationType.Query));
+```
+
+You can also register your own implementation of `IExecutionStrategySelector` which can inspect the
+`ExecutionContext` to make additional decisions before selecting an execution strategy.
+
+```csharp
+public class MyExecutionStrategySelector : IExecutionStrategySelector
+{
+    public virtual IExecutionStrategy Select(ExecutionContext context)
+    {
+        return context.Operation.Operation switch
+        {
+            OperationType.Query => ParallelExecutionStrategy.Instance,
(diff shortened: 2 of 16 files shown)


src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Initializes an instance that only returns the default registrations;
+        /// <see cref="ParallelExecutionStrategy"/> for <see cref="OperationType.Query"/> and
+        /// <see cref="SerialExecutionStrategy"/> for <see cref="OperationType.Mutation"/>.
+        /// </summary>
+        public DefaultExecutionStrategySelector()
+            : this(Array.Empty<ExecutionStrategyRegistration>())
+        {
+        }
```
Shane32, line 18: Could make this `internal` as it is only used when `DocumentExecuter` is not provided an instance of `IExecutionStrategySelector`.  Could also make this a static instance, but it shouldn't be used with DI at all, so I figured this was better.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Initializes an instance that returns the specified registrations.
+        /// </summary>
+        public DefaultExecutionStrategySelector(IExecutionStrategy queryExecutionStrategy, IExecutionStrategy mutationExecutionStrategy, IExecutionStrategy? subscriptionExecutionStrategy = null)
+            : this(GetRegistrations(queryExecutionStrategy, mutationExecutionStrategy, subscriptionExecutionStrategy))
+        {
+        }
```
Shane32, line 26: Could eliminate this.  Seems handy to manually select the execution strategy.  However, if an execution strategy is registered within the DI engine as an `IExecutionStrategy` (our code would not do this), the DI engine would select this constructor because it fulfills the most number of constructor arguments.

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -953,5 +954,101 @@ public static IGraphQLBuilder AddMetrics(this IGraphQLBuilder builder, Func<Exec
             return builder;
         }
         #endregion
+
+        #region - AddExecutionStrategySelector -
```
Shane32, line 958: I added both `AddExecutionStrategySelector` and `AddExecutionStrategy` builder methods.

src/GraphQL/Execution/IExecutionStrategySelector.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace GraphQL.Execution
+{
+    /// <summary>
+    /// Returns an instance of an <see cref="IExecutionStrategy"/> for a specified operation type.
+    /// </summary>
+    public interface IExecutionStrategySelector
+    {
+        /// <inheritdoc cref="IExecutionStrategySelector"/>
+        IExecutionStrategy Select(OperationType operationType);
```
sungam3r, line 11: I would prefer `ExecutionContext` over `OperationType`. Consider such code:
```
 IExecutionStrategy Select(ExecutionContext context)
{
    if (context.UserContext["forceParallel"] == true)
       return (parallel);
    else
       return (default);
}
```

`ExecutionContext` provides more _context_ for making decision.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
```diff
@@ -0,0 +1,80 @@
+using GraphQLParser.AST;
+
+namespace GraphQL.Execution
+{
+    /// <inheritdoc cref="IExecutionStrategySelector"/>
+    public class DefaultExecutionStrategySelector : IExecutionStrategySelector
```
sungam3r, line 6: Well, I thought that you won't add all this stuff to fill _default_ registrations. I think that `DefaultExecutionStrategySelector` should stay as simple as possible - just taking a set of registrations and then foreaching over them. Registrations for different operations may be added by extension methods and base set are added in `GraphQLBuilderBase`. Also `_registrationsarrray` actually **may** contain more registrations then 3 (for query, mutation and subscription). It's normal - just peek the first/last - what have the higher priority.

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: 👍

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: Well, the one line of code within the default constructor is required SOMEWHERE.  I can move it into the `DocumentExecuter` so it does `new DefaultExecutionStrategySelector(Array.Empty<ExecutionStrategyRegistration>)` when not using DI, but I thought a constructor here would be more organized, even if the constructor is `internal`.

src/GraphQL/Execution/IExecutionStrategySelector.cs
Shane32: Fine with me.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: Per your comments, I will remove this constructor.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -307,15 +319,6 @@ protected virtual ExecutionContext BuildExecutionContext(ExecutionOptions option
         /// for example it can be SubscriptionExecutionStrategy from GraphQL.SystemReactive.
         /// </summary>
         protected virtual IExecutionStrategy SelectExecutionStrategy(ExecutionContext context)
```
sungam3r, line 321: I think we don't need this method anymore. `ExecutionStrategySelector` is a right place to reconfigure decision.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: Per your comments, I will make this internal.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: It is, but that doesn't mean this method isn't proper.  `DocumentExecuter` should have virtual methods as necessary to split up functionality contained within and to allow overriding functionality as desired.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
sungam3r: I'm talking about that default selector should not bother with any registration at all. Registractions are external stuff. Think about _default_ as not "default concrete selection of known strategies" but about "default selection implementation" i.e simple foreaching.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: One could change:
1. Selector component
2. Registration order, add registrations , etc.

So I'm sure this method excessive. Extension points are already enough.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: I don't understand.

Are you saying that it should not provide default instances for Query and Mutation operation types?  That won't work.  DI does not guarantee the order of registrations in the enumerable and there would be no way to know what was a default registration from what was a custom registration.  Or else the DI configuration would be stupid -- like having to provide a factory that examines the enumerables and then adds on the missing operation types or something.  I don't get it.

Shane32: I really don't know what you're suggesting.  I've pretty much copied your suggested code, and only made changes that were necessary to get it to work at all, and you're talking about the selector 'bothering with registrations'.  I suggested a few constructors and then said ok I'd remove them.  So now I don't know what you mean.  You wrote the code, not me.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: Removed.  Lmk if you want it restored or resolve if not.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: Fine; it's excessive.

Shane32: It's also backwards compatible with any existing code.  It's also only two lines of code.  It's also good separation of logic.  It also doesn't hurt anything.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: Made internal.  Lmk if you want it public or resolve if not.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
sungam3r: > Are you saying that it should not provide default instances for Query and Mutation operation types? 

Yes.

>  DI does not guarantee the order of registrations in the enumerable and there would be no way to know what was a default registration from what was a custom registration.

Are you sure?

> Or else the DI configuration would be stupid  -- like having to provide a factory that examines the enumerables and then adds on the missing operation types or something. 

Now I do not understand.

I suggest to add default registrations into `GraphQLBuilderBase`.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: If the task can be solved using one extension point instead of two, then I will prefer the first. We can make another 33 virtual methods, spreading them evenly throughout the code. Theoretically, they will not harm, but in practice there will be only unnecessary burden.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: > Are you sure?

Look for yourself.  And when you're done, look for AutoFac.  And so on for every other DI provider.  I'm certainly not going to.  Designing a code pattern that relies on the order of a list of registered implementations by the DI provider would be a very poor coding practice, certainly subject to failure in the failure even if it worked today.

I'm not writing code that is subject to such subjective behavior.

sungam3r: I'll check and think about.

Shane32: In MS DI the order is guaranteed.  I don't know about the others.  Still I'm not changing it.  What if it was registered before GraphQLBuilderBase executed?  Code won't work.  Still would be a bad design.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -59,12 +60,23 @@ public DocumentExecuter(IDocumentBuilder documentBuilder, IDocumentValidator doc
         /// <see cref="IDocumentCache"/> and a set of <see cref="IConfigureExecutionOptions"/> instances.
         /// </summary>
         public DocumentExecuter(IDocumentBuilder documentBuilder, IDocumentValidator documentValidator, IComplexityAnalyzer complexityAnalyzer, IDocumentCache documentCache, IEnumerable<IConfigureExecutionOptions> configurations)
+            : this(documentBuilder, documentValidator, complexityAnalyzer, documentCache, configurations, new DefaultExecutionStrategySelector())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance with specified <see cref="IDocumentBuilder"/>,
```
sungam3r, line 68:
```suggestion
        /// Initializes a new instance with the specified <see cref="IDocumentBuilder"/>,
```

? I'm not a native speaker though.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: Nothing to think about.  Relying on the order of the list is a bad idea.

src/GraphQL/Execution/IExecutionStrategySelector.cs
```diff
@@ -0,0 +1,11 @@
+namespace GraphQL.Execution
+{
+    /// <summary>
+    /// Returns an instance of an <see cref="IExecutionStrategy"/> for a specified operation type.
```
sungam3r, line 4:
```suggestion
    /// Returns an instance of an <see cref="IExecutionStrategy"/> for a specified <see cref="ExecutionContext"/>.
```

sungam3r: The same question here - "a" or "the" before "specified"?

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
sungam3r: > Nothing to think about. 

😮 There is always something to think about. Even if my proposal is bad (most likely).

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        private readonly ExecutionStrategyRegistration[] _registrations;
+
+        /// <summary>
+        /// Initializes an instance that only returns the default registrations;
+        /// <see cref="ParallelExecutionStrategy"/> for <see cref="OperationType.Query"/> and
+        /// <see cref="SerialExecutionStrategy"/> for <see cref="OperationType.Mutation"/>.
+        /// </summary>
+        internal DefaultExecutionStrategySelector()
```
sungam3r, line 15: 60/40 to make it public.

src/GraphQL/Execution/IExecutionStrategySelector.cs
Shane32: I don't know for sure.  Either way I think would be proper English.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
sungam3r: I'm fine with currrent implementation here after removing other ctor and some code.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: Whatever you think is fine with me.  It is not really needed by anything; anyone can always call the other constructor with an empty list.  It's here because it's better documented here than if I was to embed the code directly into `DocumentExecuter`.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <summary>
+        /// Initializes a new instance with the specified registrations.
+        /// If no registration is specified for <see cref="OperationType.Query"/>, returns <see cref="ParallelExecutionStrategy"/>.
+        /// If no registration is specified for <see cref="OperationType.Mutation"/>, returns <see cref="SerialExecutionStrategy"/>.
+        /// </summary>
+        public DefaultExecutionStrategySelector(IEnumerable<ExecutionStrategyRegistration> registrations)
+        {
+            _registrations = registrations is ExecutionStrategyRegistration[] array
+                ? array
+                : (registrations ?? throw new ArgumentNullException(nameof(registrations))).ToArray();
```
sungam3r, line 29:
```suggestion
            _registrations = (registrations ?? throw new ArgumentNullException(nameof(registrations))).ToArray();
```

I don't think that optimization matters here. This will only cause extra questions. In other places (enumerable in ctor) we do not have such behavior.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: Well, it is pretty easy to replace if necessary.  A derived class can implement `IExecutionStrategySelector` and use `this` when calling the base's constructor.  Mainly I wouldn't just because of all the reasons I outlined before, mostly because it makes existing code compatible without changes, and there is no reason to remove it.  I know you don't, but I think a very large percentage of users use the serial execution strategy for queries, due to the use of scoped database engines like EF.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
sungam3r: Let there be public...

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: So for 2 lines of code, that don't interfere with anything, are not outdated, work well with the new enhancements, and all that, I wouldn't change it.  Prevents a breaking change.  I know it's just one less breaking change, but it's still one less breaking change, and one that I think will affect a large number of people.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: ok

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: Ok.   Just FYI, MS DI uses arrays under the hood.  But sure, it's only one allocation during startup.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: I give up :)

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
Shane32: When you're happy with the API diff and stuff, I'll add the necessary tests, and you can review again before merging.

src/GraphQL/Execution/DefaultExecutionStrategySelector.cs
sungam3r: Yes, I know.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: If you insist, I will remove it.  I really wouldn't.  Removing a single method, containing one line of code, just isn't worth it here for a breaking change.

If we want to, in a separate PR, clean up most all the constructors in `DocumentExecuter` and along with it, this method, then I think that makes sense.

sungam3r:
Reviewed. Good. I'm done for today.

Shane32:
@sungam3r Let's review and discuss again what should be removed/obsolete/remain.  Or if possible, let's do that in a follow-up PR.

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -79,26 +80,34 @@ A list of the available extension methods is below:
...
 
 The above methods will register the specified services typically as singletons unless otherwise specified. Graph types and middleware are registered
 as transients so that they will match the schema lifetime. So with a singleton schema, all services are effectively singletons.
 
+Custom `IGraphQLBuilder` extension methods typically rely on the `Services` property of the builder in order to register services
+with the underlying dependency injection framework. The `Services` property returns a `IServiceRegister` interface which has these methods:
+
+| Method         | Description |
+|----------------|-------------|
+| `Register`     | Registers a service within the DI framework |
```
sungam3r, line 108:
```suggestion
| `Register`     | Registers a service within the DI framework replacing existing registration if needed |
```

docs2/site/docs/migrations/migration5.md
```diff
@@ -344,6 +344,77 @@ Field<StringGraphType>("myField", resolve: context =>
...
+
+You can also register your own implementation of `IExecutionStrategySelector` which can inspect the
+`ExecutionContext` to make additional decisions before selecting an execution strategy.
+
+```csharp
+public class MyExecutionStrategySelector : IExecutionStrategySelector
+{
+    public virtual IExecutionStrategy Select(ExecutionContext context)
+    {
+        return context.Operation.Operation switch
```
sungam3r, line 399: Maybe show example with "force-serial" flag in `UserContext` here to reveal the meaning of _additional decisions_ ?

docs2/site/docs/migrations/migration5.md
```diff
@@ -622,3 +693,25 @@ invisible to the caller and lead to hardly detected bugs, the method name has be
...
+
+// v5
+services.AddGraphQL(builder => builder
+   .AddSchema<StarWarsSchema>()
+   .AddSubscriptionExecutionStrategy());
+```
+
+For more details on the new approach to execution strategy selection, please review the new feature above titled:
+
+> The `ExecutionStrategy` selected for an operation can be configured through `IGraphQLBuilder`
```
sungam3r, line 716: Markdown supports local links. We can use it for such things + also to provide some sort of table of contents.

sungam3r: migration2 : 76 lines
migration3: 479 lines
migration4: 704 lines
migration5: 716 lines and continues to grow 😮

sungam3r (approved):

sungam3r:
> Let's review and discuss again what should be removed/obsolete/remain.

I'm fine with current state. I can say that I like it more to delete the old code than to mark it as obsolete.

docs2/site/docs/migrations/migration5.md
Shane32: I know but since the migration file isn’t finished the link would likely break. This is the kind of thing I would do during cleanup immediately before release. 

The documentation for the new AutoRegistering graph types will likely all be moved into the main docs with a link pointing to the new functionality. This will probably save a lot of lines of text. Again, we should do documentation cleanup once the functionality has settled. 

I do prefer to keep documenting as the changes are being made. For the longest time the only thing holding back a 3.x release was the fact that no docs, of any sort, were added while the changes were being made. For instance I know little about directives. So if I were to issue a release without you having made docs for the changes made to directives, that would be a problem. But if we all make notes during the changes it is easy to reorder or reformat the migration docs at the end.

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -1024,6 +1043,23 @@ namespace GraphQL.Execution
         protected virtual bool ShouldIncludeNode(GraphQL.Execution.ExecutionContext context, GraphQLParser.AST.IHasDirectivesNode node) { }
         protected virtual void ValidateNodeResult(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
     }
+    public class ExecutionStrategyRegistration : System.IEquatable<GraphQL.Execution.ExecutionStrategyRegistration>
```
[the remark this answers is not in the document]
Shane32: > > Let's review and discuss again what should be removed/obsolete/remain.
> 
> I'm fine with current state. I can say that I like it more to delete the old code than to mark it as obsolete.

Just so you understand my thought process, using SubscriptionDocumentExecuter (a class we provide) can break AddExecutionStrategy (a method we provide). So that’s why I marked it as obsolete. Whereas the SelectExecutionStrategy does not interfere unless overridden, in which case the user presumably wanted to override the behavior.

PR #2932, opened 2022-02-10
Allow AutoRegisteringObjectGraphType to be used for Query or Mutation root

Shane32:
Basically, if `IResolveFieldContext.Source` is `null`, which is only possible if the graph is used as the root query/mutation graph, it will create a new instance of `TSourceType` and then call the applicable method within the field resolver (assuming that a default constructor exists for the type).

For example:

```cs
var schema = new Schema();
schema.Query = new AutoRegisteringObjectGraphType<Query>();
//etc

public class Query
{
    public string Example => "It works!";
    public IEnumerable<Person> People([FromServices] IDataContext db) => db.GetPeople();
}
```

Query:

```gql
{
  example
  people {
    name
  }
}
```

Response:

```json
{
  "example": "It works!",
  "people": [
    {
      "name": "John Doe"
    }
  ]
}
```

Of course the execution could set the root value, but then if you do the same for both query and mutation, you'd have to write code to figure out whether the query or the mutation was executed to know how to properly set the root value.  This eliminates that need.

There is also some cleanup and refactoring done.  In addition, a sample is added of how a derived class could use the new protected virtual method (along with others previously added) to add dependency injection support to graphs.


src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -363,6 +363,39 @@ public void FieldResolversWork(string fieldName, object expected)
...
+            graphType.Fields.Find("Example2")!.Resolver!.Resolve(context).ShouldBe("test");
+            graphType.Fields.Find("Example3")!.Resolver!.Resolve(context).ShouldBe(1);
+        }
+
+        [Fact]
+        public void ErrorWithNoDefaultConstructorWhenSourceNull()
+        {
+            var graphType = new TestFieldSupport<NoDefaultConstructorTest>();
+            var context = new ResolveFieldContext();
+            Should.Throw<NullReferenceException>(() => graphType.Fields.Find("Example1")!.Resolver!.Resolve(context).ShouldBe(true));
```
sungam3r, line 384: NRE?

sungam3r: Could this sort of type without default ctor be instantiated via DI ?

Shane32: > DI

As it is with this PR, you can override the BuildSourceExpression with `context => ActivatorUtilities.GetServiceOrCreateInstance<TSourceType>(context.RequestServices)` to always pull from DI and similar. Although that specific method is only with the MS DI package.

Shane32: I have added a sample of how a derived class may use the `AutoRegisteringObjectGraphType` for dependency injection.  I have code set up similar to this in some of my projects, and I would like the test to remain here, not so much as a sample, but so as to verify that the desired functionality can be achieved in a derived class.  The derived class makes use of almost everything that has been added to `AutoRegisteringObjectGraphType` so far.  We should probably make some of the `AutoRegisteringHelper` methods public to facilitate use such as is demonstrated.  I'll probably write a PR for that after other modifications are complete.

Shane32: Just FYI, I think that the functionality within `AutoRegisteringObjectGraphType`, once we add this PR, should be sufficient for most use cases.  While it is possible to go the next step with DI injection as demonstrated in the sample included within this PR, I don't think that will really be necessary to include as a built-in feature.  It's certainly worth discussing of course, but now that we have `[FromServices]` available, it should be really easy to inject services where needed, and has the benefit of only using the required service for the field that requires it -- of course that's at the cost of not having the services listed in the constructor.

Shane32:
Fixed exception bubbling in two locations and added test.

Shane32:
This PR should be ready for review.  It has some more bug fixes, tests, and a sample use case for a class derived from `AutoRegisteringObjectGraphType` that allows for DI injection of services to a graph.

Shane32:
@sungam3r Have you had a chance to look at this?

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
sungam3r: > Making this type of object the root of a graph would result in NRE.

I don't understand.

src/GraphQL.Tests/Bugs/Issue2932_DemoDIGraphType.cs
```diff
@@ -0,0 +1,209 @@
...
+        public class QueryGraph : ObjectGraphType
+        {
+            public QueryGraph()
+            {
+                Field<NonNullGraphType<DIGraphType<SampleGraph, SampleSource>>>(
+                    "example",
+                    arguments: new QueryArguments(
+                        new QueryArgument<NonNullGraphType<IntGraphType>>() { Name = "id" },
+                        new QueryArgument<NonNullGraphType<StringGraphType>>() { Name = "name" }),
+                    resolve: context => new SampleSource() { Id = context.GetArgument<int>("id"), Name = context.GetArgument<string>("name") });
```
sungam3r, line 89:
```suggestion
                    resolve: context => new SampleSource { Id = context.GetArgument<int>("id"), Name = context.GetArgument<string>("name") });
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -11,7 +11,6 @@ namespace GraphQL.Types
     /// Supports <see cref="DescriptionAttribute"/>, <see cref="ObsoleteAttribute"/>, <see cref="DefaultValueAttribute"/> and <see cref="RequiredAttribute"/>.
     /// Also it can get descriptions for fields from the XML comments.
     /// </summary>
-    /// <typeparam name="TSourceType"></typeparam>
```
sungam3r, line 14: Then remove it for `AutoRegisteringInputObjectGraphType` too.

src/GraphQL/GraphQLAttribute.cs
```diff
@@ -69,7 +69,8 @@ public virtual void Modify(ArgumentInformation argumentInformation)
             var typedMethod = _typedModifyMethods.GetOrAdd(
                 argumentInformation.ParameterInfo.ParameterType,
                 type => _modifyMethod.MakeGenericMethod(type));
-            typedMethod.Invoke(this, new object[] { argumentInformation });
```
sungam3r, line 72: Why did you change implementation? Faster? Safer?

src/GraphQL.Tests/GraphQL.Tests.csproj
```diff
@@ -21,6 +21,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\GraphQL.Dummy\GraphQL.Dummy.csproj" />
+    <ProjectReference Include="..\GraphQL.MicrosoftDI\GraphQL.MicrosoftDI.csproj" />
```
sungam3r, line 24: What do you need from this package? ScopedAttribute?

src/GraphQL.Tests/QueryTestBase.cs
```diff
@@ -30,7 +30,7 @@ public QueryTestBase()
             Executer = new DocumentExecuter(new TDocumentBuilder(), new DocumentValidator(), new ComplexityAnalyzer());
         }
 
-        public ISimpleContainer Services { get; set; }
+        public TIocContainer Services { get; set; }
```
sungam3r, line 33: So it allows you to use container-specific APIs like `Scoped` method for `Issue2932_DemoDIGraphType`. On the one hand it is good/correct. On the other hand, should we not write tests working for any DI container? Just as we did with serializers. I understand that it will require refactoring of QueryTestBase.

src/GraphQL.Tests/MsDiContainer.cs
```diff
@@ -64,6 +64,11 @@ public T Get<T>()
             _services.AddSingleton(provider => instanceCreator());
         }
 
+        public void Scoped<TService>() where TService : class
```
sungam3r, line 67: Is it worth adding to other containers?

sungam3r: Ooops, there are no other containers actually, only `SimpleContainer`. And I see that `MsDiContainer` is used only in single test - `BugWithCustomScalarsInDirective` 🤔. I woud like to remove synthetic `I/SimpleContainer` and switch to `IServiceRegister` and its implementations (that we should write, of course). A bunch of work indeed.

src/GraphQL.Tests/QueryTestBase.cs
sungam3r: #2937

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -94,33 +88,100 @@ protected virtual IEnumerable<FieldType> ProvideFields()
         /// Retuns a list of query arguments within a <see cref="QueryArguments"/> instance, and a
         /// <see cref="IFieldResolver"/> instance, for the specified field.
         /// </summary>
-        protected (QueryArguments? QueryArguments, IFieldResolver Resolver) BuildMethodResolver(MethodInfo methodInfo, FieldType fieldType)
+        protected (QueryArguments? QueryArguments, IFieldResolver Resolver) BuildFieldResolver(MemberInfo memberInfo, FieldType fieldType)
```
sungam3r, line 91: Do you mean _Field_ here as a _GraphQL field_ and not a _CLR field_ ?

src/GraphQL/Resolvers/MemberResolver.cs
```diff
@@ -1,48 +0,0 @@
...
-using System.Linq.Expressions;
-using System.Reflection;
-
-namespace GraphQL.Resolvers
-{
-    /// <summary>
-    /// A field resolver for a specific <see cref="MemberInfo"/>.
-    /// Returns the value of the field or property, or for methods, calls the method and returns the value of the method.
-    /// </summary>
-    internal class MemberResolver : IFieldResolver
```
sungam3r, line 11: Merged into `BuildFieldResolver`?

src/GraphQL/GraphQLAttribute.cs
Shane32: Exceptions were wrapped with TargetInvokeException previously

src/GraphQL.Tests/GraphQL.Tests.csproj
Shane32: Yes

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Yes. Now this method is used for any member. I thought of BuildMemberResolver or BuildResolver also. What do you think?

src/GraphQL/Resolvers/MemberResolver.cs
Shane32: Basically yes.

src/GraphQL/GraphQLAttribute.cs
Shane32: I also added a test which exercises both bits of changed code and verifies that exceptions bubble without being wrapped

src/GraphQL.Tests/Bugs/Issue2932_DemoDIGraphType.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        public class DIGraph<TSourceType>
+        {
+            public IResolveFieldContext Context { get; internal set; } = null!;
+            public TSourceType Source => (TSourceType)Context.Source!;
+            public CancellationToken RequestAborted => Context.CancellationToken;
+        }
+
+        [AttributeUsage(AttributeTargets.Parameter)]
+        public class FromSourceAttribute : GraphQLAttribute
```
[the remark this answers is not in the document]
Shane32: I can, but it doesn’t really serve a purpose under normal use, because the method is written on the instance of the source. In another way of speaking “this” always points to the source so the attribute is pointless. It’s only needed here because of the way AutoRegisteringObjectGraphType is adapted to fit another use case. 

Lmk if you still think I should add it.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Notice this also returns query arguments and a resolver, which is immediately set on the `FieldType`.  We could simply make this a void method that sets the arguments and resolver directly on the field.  Since both properties are read/write, a caller still has the opportunity to change the arguments or resolver on the field after calling this method.  In this case the method might be called something different.  Perhaps the order of the method arguments would be swapped.  For example:

```cs
public void BuildFieldType(FieldType fieldType, MemberInfo memberInfo)
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Done

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
Shane32: > I don't understand.

So historically, the point of AutoRegisteringObjectGraphType is this:

```gql
{
  people    # Query graph returns IEnumerable<Person>
  {         # this graph type is AutoRegisteringObjectGraphType<Person>
    id      # returns ((Person)IResolveFieldContext.Source).Id
    name    # returns ((Person)IResolveFieldContext.Source).Name
  }
}
```

Now we have added the capability to write methods, and even those with arguments, and change names via attributes and so on:

```gql
{
  people                  # Query graph returns IEnumerable<Person>
  {                       # this graph type is AutoRegisteringObjectGraphType<Person>
    id                    # returns ((Person)IResolveFieldContext.Source).Id
    name                  # returns ((Person)IResolveFieldContext.Source).Name
    children(minAge: 5)   # returns ((Person)IResolveFieldContext.Source).GetChildren(5)
    {
      id
      name
    }
  }
}
```

Now normally if this was used as the query root it would fail unless `ExecutionOptions.Root` was set

```gql
{                # schema.Query = typeof(AutoRegisteringObjectGraphType<Person>);
  id             # NRE trying to access IResolveFieldContext.Source
}
```

However, this is a useful scenario if you just want to easily write some methods on a class, for example let's go back to the original query:

```gql
{
  people
  {
    ...
  }
}
```

If you want to write this with C# code like this:

```cs
public class Query
{
    public IEnumerable<Person> People([FromServices] db) => db.GetPeople();
}
```

This is going to be a problem when the query executes:

```gql
{
  people      # returns ((Query)IResolveFieldContext.Source).People()    -- which is a NRE because Source is null
  {
    ...
  }
}
```

Obviously we could set the root to `new Query()`.  But then if we wrote a similar bit of code for Mutation, we would end up with `InvalidCastException`s because root is `new Query()` instead of `new Mutation()`.

Now if we think about it, when is `IResolveFieldContext.Source` null?  It can ONLY ever be in the above situation.  If null were returned for any other point in the graph, the execution strategy would not execute child nodes.  So `Source` always contains a value, except when used as the `Query`, `Mutation` or `Subscription` types, and only when `Root` is null.

This PR changes the resolver like this (going to the first example):

```gql
{
  people    # Query graph returns IEnumerable<Person>
  {         # this graph type is AutoRegisteringObjectGraphType<Person>
    id      # returns (((Person)IResolveFieldContext.Source) ?? new Person()).Id
    name    # returns (((Person)IResolveFieldContext.Source) ?? new Person()).Name
  }
}
```

Normally nothing is different.  The null check should not add any significant time to execution and no allocations are added.  However now we can use `AutoRegisteringObjectGraphType` as in the example above:

```gql
{
  people      # returns (((Query)IResolveFieldContext.Source) ?? new Query()).People()
  {
    ...
  }
}
```

Because now Query is instantiated.  It works with `ScopedAttribute` and stuff too if needed.

However, a NRE is still possible if `Query` does not have a default constructor, because then it will build the field resolver without the null-coalescing operator (just as it always did before, even before ANY of my changes).  I wrote a test for this condition.  I certainly could add code so that something like this runs, if you think I should:

```gql
{
  people      # returns (((Query)IResolveFieldContext.Source) ?? throw new InvalidOperationException($"Cannot use type 'Query' as a root graph type because no default constructor was found.")).People()
  {
    ...
  }
}
```

Shane32: Now another solution to the problem as a whole is to support static methods.  So then you would write this:

```cs
schema.Query = typeof(AutoRegisteringGraphType<Query>);

public class Query
{
    public static IEnumerable<Person> People([FromServices] db) => db.GetPeople();
}
```

Then it is not necessary to use null-coalescing operators within the field resolvers.  It would be worthwhile at this point to add `[FromSource]` into the codebase also.

Shane32: See:
- #2938

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: See:
- #2939

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -169,10 +230,12 @@ protected virtual IEnumerable<MemberInfo> GetRegisteredMembers()
...
-                    !x.ContainsGenericParameters && // exclude methods with open generics
-                    !x.IsSpecialName &&             // exclude methods generated for properties
-                    x.ReturnType != typeof(void) && // exclude methods which do not return a value
-                    x.ReturnType != typeof(Task));  // exclude methods which do not return a value
+                    !x.ContainsGenericParameters &&               // exclude methods with open generics
+                    !x.IsSpecialName &&                           // exclude methods generated for properties
+                    x.ReturnType != typeof(void) &&               // exclude methods which do not return a value
+                    x.ReturnType != typeof(Task) &&               // exclude methods which do not return a value
+                    x.GetBaseDefinition() == x &&                 // exclude methods which override an inherited class' method (e.g. GetHashCode)
+                    x.Name != "Equals" && x.Name != "<Clone>$");  // exclude methods generated for record types
```
Shane32, line 238: Bug fix: added two lines here to support record types properly.  I could not find a better way to identify the auto-generated methods. Tests added for both input and output graph types.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: OK for `protected void BuildFieldType(FieldType fieldType, MemberInfo memberInfo)`, maybe even `ConfigureFieldType`

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
Shane32: Merged #2938 - since you approved it, I assume you prefer that option.

Shane32: So now it does not implicitly create objects, but it does support static methods, and that is how you would create a graph from a clr type for a root object.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: `var isRecord = x.GetMethods().Any(m => m.Name == "<Clone>$");`

`
x.GetBaseDefinition() == x &&                 // exclude methods which override an inherited class' method (e.g. GetHashCode)
x.Name != "Equals" && x.Name != "<Clone>$");
`
->
`
x.GetMethods().Any(m => m.Name == "<Clone>$") &&
x.Name != "Equals" && x.Name != "<Clone>$");
`

Actually `Equals` may be a valid method to map into field so it should be excluded only for record types.

Shane32: ok

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
sungam3r: OK

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: how about excluding methods with this signature:

```cs
public bool Equals(TSourceType other)
```

where TSourceType is the type of the clr type being examined?

Shane32: I implemented as I suggested, excluding by convention/signature rather than more specifically by whether it's detected as a record type.  please review and lmk if this is okay.

Shane32: Ok; updated. Please review again.

sungam3r: I mean that you should not take the responsibility for all "standard" methods. Why only `Equals`? If you try to eliminate `Equals` then you should to deal with `GetHashCode`, `ToString`.

sungam3r: Record is a special case - we 100% know that records have special methods that are compiler-generated. So we 100% sure that such methods should not be considered for mapping.

Shane32: > Record is a special case - we 100% know that records have special methods that are compiler-generated. So we 100% sure that such methods should not be considered for mapping.

Are you requesting further changes?

Shane32: If anyone wants a field called GetHashCode or ToString, they can probably use `new`, and even if not certainly they can create a method with any other name and simply use the `[Name]` attribute to set the graphql name to "GetHashCode" or "ToString".  I would personally take this approach with Equals where it matches the well-known signature.  But okay if you would rather include by default.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -75,54 +76,90 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
         /// <summary>
-        /// Retuns a list of query arguments within a <see cref="QueryArguments"/> instance, and a
-        /// <see cref="IFieldResolver"/> instance, for the specified field.
+        /// Configures query arguments and a field resolver for the specified <see cref="FieldType"/>, overwriting
+        /// any existing configuration within <see cref="FieldType.Arguments"/> and <see cref="FieldType.Resolver"/>.
+        /// <br/><br/>
+        /// For fields and properties, no query arguments are added and the field resolver simply pulls the appropriate
+        /// member from <see cref="IResolveFieldContext.Source"/>.
+        /// <br/><br/>
+        /// For methods, method arguments are iterated and processed by (...crefs to other methods here...), building
```
sungam3r, line 92: > (...crefs to other methods here...)

?

Shane32: Fixed

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: > the virtual that overridden methods are always skipped

where?

Shane32: `x.GetBaseDefinition() == x` is false if it's a overridden virtual method.

Shane32: Note that methods from inherited classes are skipped because DeclaredOnly enum is used with GetMethods.  But if it's overridden then it shows up anyway in GetMethods, and that's where `x.BaseDefinition() == x` comes into play.

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -363,6 +363,62 @@ public void FieldResolversWork(string fieldName, object expected)
...
+            Should.Throw<NullReferenceException>(() => graphType.Fields.Find("Example2")!.Resolver!.Resolve(context).ShouldBe("test"));
+            Should.Throw<NullReferenceException>(() => graphType.Fields.Find("Example3")!.Resolver!.Resolve(context).ShouldBe(1));
```
sungam3r, line 386:
```suggestion
            Should.Throw<NullReferenceException>(() => graphType.Fields.Find("Example1")!.Resolver!.Resolve(context));
            Should.Throw<NullReferenceException>(() => graphType.Fields.Find("Example2")!.Resolver!.Resolve(context));
            Should.Throw<NullReferenceException>(() => graphType.Fields.Find("Example3")!.Resolver!.Resolve(context));
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: ok

src/GraphQL/Resolvers/MethodResolver.cs
```diff
@@ -42,11 +47,9 @@ public MethodResolver(MethodInfo methodInfo, List<LambdaExpression> methodArgume
...
                         ? null
-                        : Expression.Convert(
-                            Expression.MakeMemberAccess(
-                                resolveFieldContextParameter,
-                                _resolveFieldContextSourceParameter),
-                            methodInfo.DeclaringType!),
+                        : sourceExpression.Body.Replace(
+                            sourceExpression.Parameters[0],
+                            resolveFieldContextParameter),
                     methodInfo,
```
sungam3r, line 53: So NRE in case of null `Source` happens here, right? On `methodInfo` access. If so maybe change 
` private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression = context => (TSourceType)context.Source!;`
to
` private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression = context => (TSourceType)context.Source ?? throw new InvalidOperationException("some informative message");
`
?

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -386,6 +442,24 @@ public override void Modify(ArgumentInformation argumentInformation)
...
+
+            public bool Example1 { get; set; }
+            public string Example2() => "test";
+            public int Example3 = 1;
+        }
+
+        private class NullSourceTest
+        {
+            public static bool Example1 { get; set; } = true;
+            public static string Example2() => "test";
```
sungam3r, line 460: Would like to add  `public static int Example3 = 1;` field and test for it

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -75,54 +76,91 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
-                    parameterInfo.ParameterType,
-                    queryArgument ?? throw new InvalidOperationException("Invalid response from ConstructQueryArgument: queryArgument and expression cannot both be null"));
-                expressions.Add(expression);
+                var resolver = new MethodResolver(methodInfo, BuildMemberInstanceExpression(memberInfo), expressions);
+                fieldType.Arguments = queryArguments;
+                fieldType.Resolver = resolver;
+            }
+            else if (memberInfo is FieldInfo fieldInfo)
+            {
+                var sourceExpression = BuildMemberInstanceExpression(memberInfo);
```
sungam3r, line 131: Property/Method branches handle static methods (for property it is static get method). What about fields?

src/GraphQL.Tests/Bugs/Issue2932_DemoDIGraphType.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        public class QueryGraph : ObjectGraphType
+        {
+            public QueryGraph()
+            {
+                Field<NonNullGraphType<DIGraphType<SampleGraph, SampleSource>>>(
+                    "example",
+                    arguments: new QueryArguments(
+                        new QueryArgument<NonNullGraphType<IntGraphType>>() { Name = "id" },
+                        new QueryArgument<NonNullGraphType<StringGraphType>>() { Name = "name" }),
```
sungam3r, line 88:
```suggestion
                        new QueryArgument<NonNullGraphType<IntGraphType>> { Name = "id" },
                        new QueryArgument<NonNullGraphType<StringGraphType>> { Name = "name" }),
```

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -75,54 +76,91 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
+        /// <br/><br/>
+        /// Typically this is a lambda expression of type <see cref="Func{T, TResult}">Func</see>&lt;<see cref="IResolveFieldContext"/>, <typeparamref name="TSourceType"/>&gt;.
+        /// <br/><br/>
+        /// By default this returns the <see cref="IResolveFieldContext.Source"/> property.
+        /// </summary>
+        /// <param name="memberInfo">The member being called or accessed.</param>
+        protected virtual LambdaExpression BuildMemberInstanceExpression(MemberInfo memberInfo)
+            => _sourceExpression;
+
+        private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression = context => (TSourceType)context.Source!;
```
Shane32, line 162: I believe NullReferenceException is the proper exception type in this case.

```suggestion
        private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression
            = context => (TSourceType)context.Source! ?? ThrowSourceNullException();

        private static TSourceType ThrowSourceNullException()
        {
            throw new NullReferenceException($"IResolveFieldContext.Source is null; please use static methods when using an AutoRegisteringGraphType as a root graph type.");
        }
```

Note: throw is not permitted inside an expression

src/GraphQL.Tests/Bugs/Issue2932_DemoDIGraphType.cs
```diff
@@ -0,0 +1,198 @@
...
+            public SampleGraph(Service1 service1)
+            {
+                _service1 = service1;
+            }
+
+            // methods work with attributes and can use source property
+            [Id]
+            public int Id() => Source.Id;
+
+            // can write as static methods so that an instance of SampleGraph is not constructed at runtime
```
sungam3r, line 137:
```suggestion
            // can write a static method so that an instance of SampleGraph is not constructed at runtime
```
?

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
Shane32: Done

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Good catch; fixed.  Test added as you noted above.

src/GraphQL.Tests/Bugs/Issue2932_DemoDIGraphType.cs
```diff
@@ -0,0 +1,198 @@
...
+                        new QueryArgument<NonNullGraphType<StringGraphType>>() { Name = "name" }),
+                    resolve: context => new SampleSource { Id = context.GetArgument<int>("id"), Name = context.GetArgument<string>("name") });
+            }
+        }
+
+        public class Service1
+        {
+            private int _counter;
+
+            public Task<IEnumerable<string>> GetChildrenAsync(int sourceId, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
Shane32: This is an example service pulled from DI.  It's not mapped to graphql at all.

src/GraphQL.Tests/Bugs/Issue2932_DemoDIGraphType.cs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            Services.Singleton<TestSchema>();
+            Services.Singleton<DIGraphType<SampleGraph, SampleSource>>();
+            Services.Scoped<Service1>();
+            Services.Scoped<Service2>();
+            // note: in this example, SampleGraph is not registered, but is created for every field resolver (except static methods) -- see DIGraphType.MemberInstanceFunc
+        }
+
+        [Fact]
+        public void TestDIAutoRegisteringGraphType()
```
sungam3r, line 22: Pretty complex example.

src/GraphQL.Tests/Bugs/Issue2932_DemoDIGraphType.cs
Shane32: No; that would change the meaning of the sentence.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: I updated comment. Commit if OK.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
sungam3r: 👍

src/GraphQL/Resolvers/MethodResolver.cs
Shane32: Changed as discussed elsewhere.

src/GraphQL.Tests/QueryTestBase.cs
Shane32: I assume we will track this in #2937.  resolving comment here.

src/GraphQL.Tests/MsDiContainer.cs
Shane32: I assume we will track this in #2937.  resolving comment here.

src/GraphQL.Tests/Bugs/Issue2932_DemoDIGraphType.cs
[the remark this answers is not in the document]
sungam3r: 🤦‍♂️

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -75,52 +76,97 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
+        /// Typically this is a lambda expression of type <see cref="Func{T, TResult}">Func</see>&lt;<see cref="IResolveFieldContext"/>, <typeparamref name="TSourceType"/>&gt;.
+        /// <br/><br/>
+        /// By default this returns the <see cref="IResolveFieldContext.Source"/> property.
+        /// </summary>
+        /// <param name="memberInfo">The member being called or accessed.</param>
+        protected virtual LambdaExpression BuildMemberInstanceExpression(MemberInfo memberInfo)
+            => _sourceExpression;
+
+        private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression
+            = context => (TSourceType)context.Source! ?? ThrowSourceNullException();
```
sungam3r, line 165: `context.Source! -> context.Source`
?

Shane32:
```suggestion
            = context => (TSourceType?)context.Source ?? ThrowSourceNullException();
```

Note: `?` is required here

sungam3r:
I looked through all the code and have the last concern. I doubt about value-types as source and case when casting `null` to `TSourceType` so I suggest to add tests with methods returning some complex struct.

Shane32:
> I looked through all the code and have the last concern. I doubt about value-types as source and case when casting `null` to `TSourceType` so I suggest to add tests with methods returning some complex struct.

How about if I make the null-coalescing operator occur before the cast?

`= context => (TSourceType)(context.Source ?? ThrowSourceNullException());`

Shane32:
Please review last commit.  It should take care of your concern.

Structs are not properly supported because the AttributeUsage does not allow NameAttribute to be placed on a struct.  I'll fix that in a separate PR and will add tests for struct support.

sungam3r (approved):

Shane32:
I added the test you requested as well, so it will test for a correct null reference exception in the case that TSourceType is a struct type and when used as a root object.

PR #2938, opened 2022-02-12
Alt for 2932 (supporting root types): support static methods

Shane32:
This proposal might make more logical sense than #2932


src/GraphQL/FromSource.cs
```diff
@@ -0,0 +1,16 @@
+using GraphQL.Types;
+
+namespace GraphQL
+{
+
```
sungam3r, line 5:
```suggestion
```

sungam3r (approved):

PR #2943, opened 2022-02-13
Switch tests to work with real DI containers

### LINKED ISSUE #2937: Rewrite tests to use real DI containers
Ooops, there are no other containers actually, only `SimpleContainer`. And I see that `MsDiContainer` is used only in single test - `BugWithCustomScalarsInDirective` 🤔. I woud like to remove synthetic `I/SimpleContainer` and switch to `IServiceRegister` and its implementations (that we should write, of course). A bunch of work indeed.

_Originally posted by @sungam3r in https://github.com/graphql-dotnet/graphql-dotnet/pull/2932#r805143348_

sungam3r:
fixes #2937 

I tried to make it so to demand the minimum number of changes in the tests themselves. xUnit as it seemed to me, gives not much tools to parametrize classes (not methods).


src/GraphQL.Tests/Bugs/BugWithCustomScalarsInDirective.cs
```diff
@@ -1,11 +1,12 @@
...
 using GraphQL.Types;
 using GraphQLParser.AST;
 
 namespace GraphQL.Tests.Bugs
 {
-    public class BugWithCustomScalarsInDirective : QueryTestBase<BugWithCustomScalarsInDirectiveSchema, MsDiContainer>
+    public class BugWithCustomScalarsInDirective : QueryTestBase<BugWithCustomScalarsInDirectiveSchema>
     {
-        public BugWithCustomScalarsInDirective()
+        public override void RegisterServices(IServiceRegister Services)
```
sungam3r, line 9: I use `Services` instead of `services`/ `register` here to minimize diff. @Shane32 so I would like to rename it after your review.

src/GraphQL.Tests/Bugs/Issue2392_OverrideBuiltInScalars_Alt.cs
```diff
@@ -6,11 +6,6 @@ namespace GraphQL.Tests.Bugs
     // https://github.com/graphql-dotnet/graphql-dotnet/issues/2392
     public class Issue2392_OverrideBuiltInScalars_Alt : QueryTestBase<Issue2392_OverrideBuiltInScalars_Alt.MySchema>
     {
-        public Issue2392_OverrideBuiltInScalars_Alt()
-        {
-            Services.Singleton<MySchema>();
```
sungam3r, line 11: Schema registered in `QueryTestBase.RegisterServices` via `TryRegister`.

src/GraphQL.Tests/DI.xUnit/DependencyInjectionDataAttribute.cs
```diff
@@ -0,0 +1,8 @@
+namespace GraphQL.Tests.DI;
+
+internal sealed class DependencyInjectionDataAttribute : ClassDataAttribute
```
sungam3r, line 3: There are no usages of this attribute in tests now by I tested it. It works. This attribute should be used when someone adds one more DI provider.

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
```diff
@@ -13,12 +13,20 @@ public class SchemaTypesTests
...
+            services.AddSingleton<EpisodeEnum>();
+            using var provider = services.BuildServiceProvider();
 
             // mock it so we can verify behavior
             var mock = new Mock<IServiceProvider>(MockBehavior.Loose);
-            mock.Setup(x => x.GetService(It.IsAny<Type>())).Returns<Type>(type => services.Get(type));
+            mock.Setup(x => x.GetService(It.IsAny<Type>())).Returns<Type>(type => provider.GetService(type));
 
             // run test
             var schema = new StarWarsSchema(mock.Object);
```
Shane32, line 32: I would guess that we can change the code to just return `provider` now where it was using `mock.Object`.

src/GraphQL.Tests/DI.xUnit/ServiceRegisterExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using GraphQL.DI;
+
+namespace GraphQL.Tests
+{
+    internal static class ServiceRegisterExtensions
+    {
+        //TODO: rename to transient
+        public static void Register<TService>(this IServiceRegister register)
```
sungam3r, line 8: Again - I kept all names as before changes to minimize diff - the goal of whole this class. @Shane32 I would like to rename `Register` to `Transient` after your review. Another option - remove extension methods at all and use "plain" `Register` methods but I would not.

src/GraphQL.Tests/Execution/Performance/ThreadPerformanceTests.cs
```diff
@@ -1,16 +1,17 @@
...
 {
     public class ThreadPerformanceTests : QueryTestBase<ThreadPerformanceTests.ThreadPerformanceSchema>
     {
-        public ThreadPerformanceTests()
+        public override void RegisterServices(IServiceRegister Services)
         {
             Services.Register<PerfQuery>();
-
-            Services.Singleton(() => new ThreadPerformanceSchema(new SimpleContainerAdapter(Services)));
+            Services.Register<PerfMutation>();
```
sungam3r, line 13: In some tests (3 or 4), some registrations were lost. It worked because `SimpleContainer` created an unregistered services under the hood.

src/GraphQL.Tests/Execution/UnionInterfaceTests.cs
```diff
@@ -52,6 +46,16 @@ public UnionInterfaceTests()
             };
         }
 
+        public override void RegisterServices(IServiceRegister Services)
+        {
+            Services.Register<DogType>();
+            Services.Register<CatType>();
+            Services.Register<PetType>();
+            Services.Register<PersonType>();
+            Services.Register<NamedType>();
```
sungam3r, line 55: was lost

src/GraphQL.Tests/StarWars/StarWarsSubFieldsTest.cs
```diff
@@ -6,12 +6,7 @@ namespace GraphQL.Tests.StarWars
 {
     public class StarWarsSubFieldsTests : StarWarsTestBase
     {
-        public StarWarsSubFieldsTests() : base()
-        {
-            RootQuery = (StarWarsQuery)Schema.Query;
```
sungam3r, line 11: Such tests now cannot access objects from DI from ctor, i.e. before registration happens.

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
sungam3r: `
  Moq.MockException : 
    Expected invocation on the mock once, but was 0 times: x => x.GetService(GraphQL.StarWars.StarWarsQuery)
    
    Performed invocations:
    
       Mock<IServiceProvider:1> (x):
       No invocations performed.
    
`

Shane32: without looking at the code i'm not sure how it's used, so i assume you've got it figured out.

src/GraphQL.Tests/DI.xUnit/DependencyInjectionDataAttribute.cs
Shane32: i can add another di provider

src/GraphQL.Tests/Bugs/BugWithCustomScalarsInDirective.cs
Shane32: Understood

Shane32:
Everything looks pretty straightforward.  I don't care too much how tests are written so long as they test the functionality they were written for.  Of course improving the test base is always a good idea.

Is there anything in particular I should review here?

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
sungam3r: Mock verifies calls made through it so mock should be used, not real provider.

src/GraphQL.Tests/DI.xUnit/DependencyInjectionDataAttribute.cs
sungam3r: I'm experimenting with StructureMap` now. Some variants:
1. Autofac
2. LightInject
3. Ninject
4. SimpleInjector
5. StructureMap
6. CastleWindsor

Rel: https://github.com/EasyNetQ/EasyNetQ/pull/1108

sungam3r:
Naming maybe.

src/GraphQL.Tests/DI.xUnit/ServiceRegisterExtensions.cs
Shane32: Changing Register to Transient makes sense here.

src/GraphQL.Tests/Complexity/ComplexityTestBase.cs
```diff
@@ -18,9 +20,16 @@ public class ComplexityTestBase
...
                 options.Query = query;
                 options.ComplexityConfiguration = complexityConfig;
             });
+
+        //ISSUE: manually created test instance with ServiceProvider
+        private static StarWarsSchema CreateSchema()
+        {
+            var builder = new MicrosoftDI.GraphQLBuilder(new ServiceCollection(), b => new StarWarsTestBase().RegisterServices(b.Services));
+            return builder.ServiceCollection.BuildServiceProvider().GetRequiredService<StarWarsSchema>();
+        }
```
Shane32, line 33: I see you marked this and some other places with //ISSUE

Do you have a solution or for now is this just how it will be?

src/GraphQL.Tests/Execution/Performance/ThreadPerformanceTests.cs
Shane32: I'm glad that the registrations are now explicit.

src/GraphQL.Tests/DI.xUnit/DependencyInjectionAdapterData.cs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+internal sealed class MicrosoftAdapter : IDependencyInjectionAdapter
+{
+    private readonly MicrosoftDI.GraphQLBuilder _builder;
+    private IServiceProvider _provider;
+
+    public MicrosoftAdapter(Action<IServiceRegister> configure)
+    {
+        _builder = new MicrosoftDI.GraphQLBuilder(new ServiceCollection(), b => configure(b.Services));
```
Shane32, line 38: This is going to register a whole bunch of predefined default services.  Is that intended?

src/GraphQL.Tests/DI.xUnit/DependencyInjectionAdapterData.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        yield return new object[] { "MSDI" };
+        //yield return new object[] { "StructureMap" };
+        //yield return new object[] { "CastleWindsor" };
+    }
+
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+}
+
+internal interface IDependencyInjectionAdapter
```
Shane32, line 26: Let's move to a separate file.  We already have a folder just for this stuff..

src/GraphQL.Tests/DI.xUnit/DependencyInjectionAdapterData.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    public IServiceProvider ServiceProvider { get; }
+}
+
+internal sealed class MicrosoftAdapter : IDependencyInjectionAdapter
```
Shane32, line 31: Let's move to a separate file.

src/GraphQL.Tests/DI.xUnit/PrepareDependencyInjectionAttribute.cs
```diff
@@ -0,0 +1,71 @@
...
+
+/// <summary>
+/// Allows test methods from <see cref="QueryTestBase{TSchema, TDocumentBuilder}"/> and descendants
+/// work with different DI providers. Mark your test method with [Theory] and [DependencyInjectionData]
+/// attributes instead of [Fact] attribute. Also add 'string name' parameter (unused, the name of DI provider)
+/// into your test method. By default all methods from <see cref="QueryTestBase{TSchema, TDocumentBuilder}"/>
+/// and descendants (not marked with [DependencyInjectionData]) work with the first DI provider from
+/// <see cref="DependencyInjectionAdapterData"/>.
+/// </summary>
+internal sealed class PrepareDependencyInjectionAttribute : BeforeAfterTestAttribute
```
Shane32, line 19: This class looks really weird. How does it line up the names to the instances? Also I see a static concurrent dictionary holding references to every DI service collection for every test and every DI provider.  Isn’t there another way to make this work? Can we just use async local to temporarily hold the service provider for the specific test in question and eliminate the dictionary and only instantiate the specific instance needed? And what if we want to test multiple serializers with multiple DI providers - how is that done?

Shane32: Maybe we should just use something a bit more conventional and simply set a property on the query test base like “DIProviderName = diName;” so it’s more flexible when writing “theory” tests.

src/GraphQL.Tests/Complexity/ComplexityTestBase.cs
sungam3r: Just a note about "not so good" implementation in my opinion. I don't want to rewrite those tests completely, not a goal for this PR. These tests create schema object that should be properly configured with registered services.

src/GraphQL.Tests/DI.xUnit/PrepareDependencyInjectionAttribute.cs
sungam3r: > This class looks really weird.

Agree.

> How does it line up the names to the instances?

Nohow. We just should use the correct order of adapters and their names in `DependencyInjectionAdapterData`.

>  Isn’t there another way to make this work? 

Maybe, but I don't know.

> Can we just use async local to temporarily hold the service provider for the specific test in question and eliminate the dictionary and only instantiate the specific instance needed?

It was my first attempt. Alas xUnit doesn't use the same async-local context.

> And what if we want to test multiple serializers with multiple DI providers - how is that done?

I thought about that. I don't know yet. Maybe some further tricks.

sungam3r: > Maybe we should just use something a bit more conventional and simply set a property on the query test base like “DIProviderName = diName;” so it’s more flexible when writing “theory” tests.

Perhaps but I don't know. Again my goal was to integrate real DI containers with as little test changes as possible. We may continue further when writing other DI adapters implementations.

src/GraphQL.Tests/DI.xUnit/ServiceRegisterExtensions.cs
sungam3r: done

src/GraphQL.Tests/DI.xUnit/DependencyInjectionAdapterData.cs
sungam3r: done

src/GraphQL.Tests/DI.xUnit/ServiceRegisterExtensions.cs
sungam3r: done

src/GraphQL.Tests/DI.xUnit/PrepareDependencyInjectionAttribute.cs
Shane32: I'd like to suggest that we have a goal of removing this class or rewriting it to work better.  But we can merge in the changes now if you like.

```suggestion
// TODO: remove or rewrite this class; should support theories in conjuction with multiple DI providers
internal sealed class PrepareDependencyInjectionAttribute : BeforeAfterTestAttribute
```

sungam3r: > If there is a test that is a theory, and does not use DependencyInjectionAdapterData (the theory is for other data), is there any issue here? Will it still use just the first DI provider even if multiple are configured?

It uses the first provider, real provider, MSDI.

Shane32 (approved):

PR #2946, opened 2022-02-14
Introduce RegistrationCompareMode

sungam3r:
Rel: https://github.com/graphql-dotnet/server/issues/705

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
@@ -62,9 +62,9 @@ namespace GraphQL.MicrosoftDI
         public GraphQL.DI.IServiceRegister Register(System.Type serviceType, System.Func<System.IServiceProvider, object> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime, bool replace = false) { }
         public GraphQL.DI.IServiceRegister Register(System.Type serviceType, System.Type implementationType, GraphQL.DI.ServiceLifetime serviceLifetime, bool replace = false) { }
         protected override void RegisterDefaultServices() { }
-        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, object implementationInstance) { }
-        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Func<System.IServiceProvider, object> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime) { }
-        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Type implementationType, GraphQL.DI.ServiceLifetime serviceLifetime) { }
+        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, object implementationInstance, GraphQL.DI.RegistrationCompareMode mode = 0) { }
+        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Func<System.IServiceProvider, object> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime, GraphQL.DI.RegistrationCompareMode mode = 0) { }
+        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Type implementationType, GraphQL.DI.ServiceLifetime serviceLifetime, GraphQL.DI.RegistrationCompareMode mode = 0) { }
     }
     public static class GraphQLBuilderExtensions
     {
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -873,9 +873,14 @@ namespace GraphQL.DI
         GraphQL.DI.IServiceRegister Register(System.Type serviceType, object implementationInstance, bool replace = false);
         GraphQL.DI.IServiceRegister Register(System.Type serviceType, System.Func<System.IServiceProvider, object> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime, bool replace = false);
         GraphQL.DI.IServiceRegister Register(System.Type serviceType, System.Type implementationType, GraphQL.DI.ServiceLifetime serviceLifetime, bool replace = false);
-        GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, object implementationInstance);
-        GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Func<System.IServiceProvider, object> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime);
-        GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Type implementationType, GraphQL.DI.ServiceLifetime serviceLifetime);
+        GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, object implementationInstance, GraphQL.DI.RegistrationCompareMode mode = 0);
+        GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Func<System.IServiceProvider, object> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime, GraphQL.DI.RegistrationCompareMode mode = 0);
+        GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Type implementationType, GraphQL.DI.ServiceLifetime serviceLifetime, GraphQL.DI.RegistrationCompareMode mode = 0);
+    }
+    public enum RegistrationCompareMode
+    {
+        ServiceType = 0,
+        ServiceTypeAndImplementationType = 1,
     }
     public enum ServiceLifetime
     {
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -298,11 +298,14 @@ namespace GraphQL
             where TImplementation :  class, TService { }
         public static GraphQL.DI.IServiceRegister TryRegister<TService>(this GraphQL.DI.IServiceRegister services, GraphQL.DI.ServiceLifetime serviceLifetime)
             where TService :  class { }
-        public static GraphQL.DI.IServiceRegister TryRegister<TService>(this GraphQL.DI.IServiceRegister services, TService implementationInstance)
-            where TService :  class { }
         public static GraphQL.DI.IServiceRegister TryRegister<TService>(this GraphQL.DI.IServiceRegister services, System.Func<System.IServiceProvider, TService> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime)
             where TService :  class { }
-        public static GraphQL.DI.IServiceRegister TryRegister<TService, TImplementation>(this GraphQL.DI.IServiceRegister services, GraphQL.DI.ServiceLifetime serviceLifetime)
+        public static GraphQL.DI.IServiceRegister TryRegister<TService>(this GraphQL.DI.IServiceRegister services, TService implementationInstance, GraphQL.DI.RegistrationCompareMode mode = 0)
+            where TService :  class { }
+        public static GraphQL.DI.IServiceRegister TryRegister<TService, TImplementation>(this GraphQL.DI.IServiceRegister services, GraphQL.DI.ServiceLifetime serviceLifetime, GraphQL.DI.RegistrationCompareMode mode = 0)
+            where TService :  class
+            where TImplementation :  class, TService { }
+        public static GraphQL.DI.IServiceRegister TryRegister<TService, TImplementation>(this GraphQL.DI.IServiceRegister services, System.Func<System.IServiceProvider, TImplementation> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime, GraphQL.DI.RegistrationCompareMode mode = 0)
             where TService :  class
             where TImplementation :  class, TService { }
     }
--- a/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
+++ b/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
@@ -175,7 +175,7 @@ namespace GraphQL.MicrosoftDI
             else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
                 ServiceCollection.TryAddEnumerable(descriptor);
             else
-                throw new NotSupportedException(mode.ToString());
+                throw new ArgumentOutOfRangeException(nameof(mode));
 
             return this;
         }
--- a/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
+++ b/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
@@ -156,7 +156,7 @@ namespace GraphQL.MicrosoftDI
             else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
                 ServiceCollection.TryAddEnumerable(descriptor);
             else
-                throw new NotSupportedException(mode.ToString());
+                throw new ArgumentOutOfRangeException(nameof(mode));
 
             return this;
         }
--- a/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
+++ b/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
@@ -124,38 +124,59 @@ namespace GraphQL.MicrosoftDI
         }
 
         /// <inheritdoc/>
-        public IServiceRegister TryRegister(Type serviceType, Func<IServiceProvider, object> implementationFactory, ServiceLifetime serviceLifetime)
+        public IServiceRegister TryRegister(Type serviceType, Func<IServiceProvider, object> implementationFactory, ServiceLifetime serviceLifetime, RegistrationCompareMode mode = RegistrationCompareMode.ServiceType)
         {
             if (serviceType == null)
                 throw new ArgumentNullException(nameof(serviceType));
             if (implementationFactory == null)
                 throw new ArgumentNullException(nameof(implementationFactory));
 
-            ServiceCollection.TryAdd(new ServiceDescriptor(serviceType, implementationFactory, TranslateLifetime(serviceLifetime)));
+            var descriptor = new ServiceDescriptor(serviceType, implementationFactory, TranslateLifetime(serviceLifetime));
+            if (mode == RegistrationCompareMode.ServiceType)
+                ServiceCollection.TryAdd(descriptor);
+            else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
+                ServiceCollection.TryAddEnumerable(descriptor);
+            else
+                throw new NotSupportedException(mode.ToString());
+
             return this;
         }
 
         /// <inheritdoc/>
-        public IServiceRegister TryRegister(Type serviceType, Type implementationType, DI.ServiceLifetime serviceLifetime)
+        public IServiceRegister TryRegister(Type serviceType, Type implementationType, DI.ServiceLifetime serviceLifetime, RegistrationCompareMode mode = RegistrationCompareMode.ServiceType)
         {
             if (serviceType == null)
                 throw new ArgumentNullException(nameof(serviceType));
             if (implementationType == null)
                 throw new ArgumentNullException(nameof(implementationType));
 
-            ServiceCollection.TryAdd(new ServiceDescriptor(serviceType, implementationType, TranslateLifetime(serviceLifetime)));
+            var descriptor = new ServiceDescriptor(serviceType, implementationType, TranslateLifetime(serviceLifetime));
+            if (mode == RegistrationCompareMode.ServiceType)
+                ServiceCollection.TryAdd(descriptor);
+            else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
+                ServiceCollection.TryAddEnumerable(descriptor);
+            else
+                throw new NotSupportedException(mode.ToString());
+
             return this;
         }
 
         /// <inheritdoc/>
-        public IServiceRegister TryRegister(Type serviceType, object implementationInstance)
+        public IServiceRegister TryRegister(Type serviceType, object implementationInstance, RegistrationCompareMode mode = RegistrationCompareMode.ServiceType)
         {
             if (serviceType == null)
                 throw new ArgumentNullException(nameof(serviceType));
             if (implementationInstance == null)
                 throw new ArgumentNullException(nameof(implementationInstance));
 
-            ServiceCollection.TryAdd(new ServiceDescriptor(serviceType, implementationInstance));
+            var descriptor = new ServiceDescriptor(serviceType, implementationInstance);
+            if (mode == RegistrationCompareMode.ServiceType)
+                ServiceCollection.TryAdd(descriptor);
+            else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
+                ServiceCollection.TryAddEnumerable(descriptor);
+            else
+                throw new NotSupportedException(mode.ToString());
+
             return this;
         }
 
--- a/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
+++ b/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
@@ -137,7 +137,7 @@ namespace GraphQL.MicrosoftDI
             else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
                 ServiceCollection.TryAddEnumerable(descriptor);
             else
-                throw new NotSupportedException(mode.ToString());
+                throw new ArgumentOutOfRangeException(nameof(mode));
 
             return this;
         }
--- a/src/GraphQL/DI/IServiceRegister.cs
+++ b/src/GraphQL/DI/IServiceRegister.cs
@@ -34,7 +34,7 @@ namespace GraphQL.DI
         /// has not already been registered.
         /// <br/><br/>
         /// With <see cref="RegistrationCompareMode.ServiceTypeAndImplementationType"/>, it is required
-        /// that <paramref name="implementationFactory"/ is a strongly typed delegate with a return type
+        /// that <paramref name="implementationFactory"/> is a strongly typed delegate with a return type
         /// of a specific implementation type.
         /// </summary>
         IServiceRegister TryRegister(Type serviceType, Func<IServiceProvider, object> implementationFactory, ServiceLifetime serviceLifetime, RegistrationCompareMode mode = RegistrationCompareMode.ServiceType);
--- a/src/GraphQL/DI/IServiceRegister.cs
+++ b/src/GraphQL/DI/IServiceRegister.cs
@@ -27,7 +27,16 @@ namespace GraphQL.DI
         /// </summary>
         IServiceRegister TryRegister(Type serviceType, Type implementationType, ServiceLifetime serviceLifetime, RegistrationCompareMode mode = RegistrationCompareMode.ServiceType);
 
-        /// <inheritdoc cref="TryRegister(Type, Type, ServiceLifetime, RegistrationCompareMode)"/>
+        /// <summary>
+        /// Registers the service of type <paramref name="serviceType"/> with the dependency
+        /// injection provider if a service of the same type (and of the same implementation type
+        /// in case of <see cref="RegistrationCompareMode.ServiceTypeAndImplementationType"/>)
+        /// has not already been registered.
+        /// <br/><br/>
+        /// With <see cref="RegistrationCompareMode.ServiceTypeAndImplementationType"/>, it is required
+        /// that <paramref name="implementationFactory"/ is a strongly typed delegate with a return type
+        /// of a specific implementation type.
+        /// </summary>
         IServiceRegister TryRegister(Type serviceType, Func<IServiceProvider, object> implementationFactory, ServiceLifetime serviceLifetime, RegistrationCompareMode mode = RegistrationCompareMode.ServiceType);
 
         /// <inheritdoc cref="TryRegister(Type, Type, ServiceLifetime, RegistrationCompareMode)"/>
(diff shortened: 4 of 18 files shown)


Shane32:
We do not need this code.  There is ** no ** reason to add it to the library.

Shane32:
The more we make changes like this, the more we limit the ability to use any DI provider.  I do not think we should merge this code.

sungam3r:
😮 Now I migrate my projects to v5-dev versions and they are completely broken because `TryAddEnumerable` doesn't work.

Shane32:
> 😮 Now I migrate my projects to v5-dev versions and they are completely broken because `TryAddEnumerable` doesn't work.

The problem is simply that I wrote the code incorrectly.  It should have been Register instead of TryRegister.  There is ** no need ** for TryAddEnumerable.

sungam3r:
This is a strong assertion and even offensive.

Shane32:
Fix the bug -- removal of three characters "Try" -- and it will work just fine.

Shane32:
> This is a strong assertion and even offensive.

Then I apologize.  I would appreciate some basis for saying that Register is not entirely suitable for the method in question.

Shane32:
Pleas understand, my goal is to make `IServiceRegister` compatible with as many DI providers as possible.  Adding this code does the opposite.  We already have a problem with `register` with Autofac, the second-largest DI provider that exists, and I have not even looked at any other DI provider.  Why make it worse still?  Especially when -- from my point of view -- there is no need.

sungam3r:
No. You miss the idempotency of behavior out of sight. If I add a certain kind of currency multiplier in the DI multiplier, then I don't want it to be added twice and changed the currency twice. I need to repeat again - I write libraries, not end-user apps. App consists of many independent layers/components. **Each** component may add some sort of configuration in DI. If two or more components add the same registration then I wont it be duplicated if I designed it to be injected as `IEnumerable<>`.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
```diff
@@ -62,9 +62,9 @@ namespace GraphQL.MicrosoftDI
...
         protected override void RegisterDefaultServices() { }
-        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, object implementationInstance) { }
-        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Func<System.IServiceProvider, object> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime) { }
-        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Type implementationType, GraphQL.DI.ServiceLifetime serviceLifetime) { }
+        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, object implementationInstance, GraphQL.DI.RegistrationCompareMode mode = 0) { }
+        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Func<System.IServiceProvider, object> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime, GraphQL.DI.RegistrationCompareMode mode = 0) { }
+        public GraphQL.DI.IServiceRegister TryRegister(System.Type serviceType, System.Type implementationType, GraphQL.DI.ServiceLifetime serviceLifetime, GraphQL.DI.RegistrationCompareMode mode = 0) { }
```
Shane32, line 67: Isn't TryAddEnumerable only compatible with `Type serviceType, Type implementationType` registrations?  I have not tried to be sure, but I don't understand how it could be otherwise.

sungam3r:
> Pleas understand, my goal is to make IServiceRegister compatible with as many DI providers as possible. Adding this code does the opposite. We already have a problem with register with Autofac, the second-largest DI provider that exists, and I have not even looked at any other DI provider. Why make it worse still? Especially when -- from my point of view -- there is no need.

I know your intention. I myself tried to do so. I asked the same questions. Being ready for this task it is impossible to solve due to the **fundamental** differences between DI providers. If we design the smallest common denominator for ALL providers, then the solution will be uncomfortable for everyone.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
sungam3r: https://github.com/graphql-dotnet/server/issues/705#issuecomment-1005629696

Shane32:
> No. You miss the idempotency of behavior out of sight. If I add a certain kind of currency multiplier in the DI multiplier, then I don't want it to be added twice and changed the currency twice. I need to repeat again - I write libraries, not end-user apps. App consists of many independent layers/components. **Each** component may add some sort of configuration in DI. If two or more components add the same registration then I wont it be duplicated if I designed it to be injected as `IEnumerable<>`.

When we have a concrete need for "idempotency of behavior", then let's consider the PR.  I don't believe now is that time.  A one-line bug fix does not justify this massive of change, especially when many DI providers do not support this behavior.

The only thing this does is to make it so that you can call AddDefaultEndpointSelectorPolicy multiple times and it will not reissue a registration.  I find it very hard to believe that users are calling that method multiple times in their `Startup.cs` file.  It's not referenced anywhere else within the server project.  It's not like our default `AddGraphQL` calls it.  This example simply isn't one that has enough basis for making these changes.

sungam3r:
> When we have a concrete need for "idempotency of behavior", then let's consider the PR. 

I have the concrete need now.

Shane32:
> If we design the smallest common denominator for ALL providers, then the solution will be uncomfortable for everyone.

How so?  It seems to be working great.  I find no issue with anything we have provided.  Nothing other than a one-line bug in my code.

Shane32:
> I have the concrete need now.

You call `AddDefaultEndpointSelectorPolicy` multiple times in your code?  On purpose?

sungam3r:
The problem is not with `AddDefaultEndpointSelectorPolicy` at all. The problem is API that I use from other libraries - `IServiceRegister`.

Shane32:
> The problem is not with `AddDefaultEndpointSelectorPolicy` at all. The problem is API that I use from other libraries - `IServiceRegister`.

Then please explain.  If you'd like this PR merged, you should be prepared to explain your need and use case.

sungam3r:
> The only thing this does is to make it so that you can call `AddDefaultEndpointSelectorPolicy` multiple times and it will not reissue a registration. 

Yes + change `AddDefaultEndpointSelectorPolicy` word from the sentence above to any other and I think you may undestand my problem.

Shane32:
> > The only thing this does is to make it so that you can call `AddDefaultEndpointSelectorPolicy` multiple times and it will not reissue a registration.
> 
> Yes + change `AddDefaultEndpointSelectorPolicy` word from the sentence above to any other and I think you may undestand my problem.

Not yet I do not.  I am awaiting some type of example of a need here.

sungam3r:
I have a plethora of methods that SHOULD be idempotent. If they don't then BOOM! The same algorithm will be applied multiple times making weird things across codebase. Think of this as about applying field middleware multiple times like appending "10$" in resolver, so if you apply it 3 time from different components then you get something like 9910$10$10$.

Shane32:
> I have a plethora of methods that SHOULD be idempotent. If they don't then BOOM! The same algorithm will be applied multiple times making weird things across codebase. Think of this as about applying field middleware multiple times like appending "10$" in resolver, so if you apply it 3 time from different components then you get something like 9910$10$10$.

How does this apply to the IGraphQLBuilder, which is designed for GraphQL stuff?  You are free to design your plethora of methods to use any service provider you wish.  `IGraphQLBuilder` is designed for one purpose - to set up the services required for GraphQL.  It serves that purpose, as far as I can see.  It is not designed to service the needs of a general DI provider.

sungam3r:
One more example - I want to add some sort of logging interceptor in my DI. Component A does it. Component A can be wrapped by components B, C, D. So all of them now register A in DI. So should I have logs to be written 3 times?

Shane32:
> One more example - I want to add some sort of logging interceptor in my DI. Component A does it. Component A can be wrapped by components B, C, D. So all of them now register A in DI. So should I have logs to be written 3 times?

At least now we have an example.  And why does this code use `IServiceRegister` at all?

sungam3r:
> is designed for one purpose - to set up the services required for GraphQL. It serves that purpose, as far as I can see. It is not designed to service the needs of a general DI provider.

I think you are contradictory yourself. By your logic, no one will be able to facilitate `TryAddEnumerable`, having only `IGraphQLBuilder` on the hands. `AddDefaultEndpointSelectorPolicy` is just a one example of such situation (in out code).

Answer the question, please - how can I properly write an extension method for `IGraphQLBuilder`, which should add some sort of loger to some of our component that takes `IEnumerable<>` in ctor?

sungam3r:
> And why does this code use IServiceRegister at all?

😕 And what do you think should be used instead?

Shane32:
IServiceCollection ?

Shane32:
I would really like to understand what you're trying to do here.

Shane32:
I have a large solution with a bunch of services.  Each project has a Startup.cs which is called from the main project.  They don't use `IServiceRegister` at all.  Only within the startup.cs do I use it to set up the GraphQL services.  I don't use it to set up database services, printing services, or anything else.

sungam3r:
I doesn't have `IServiceCollection`. Well, OK. I'll try one more time - look at `AddDocumentListener` from the main repo.

Shane32:
Are you designing GraphQL-specific services that are interdependent, that are only for GraphQL.NET, such as a document listeners, middleware, type registrations and similar?  That should be the only reason to write anything that needs IGraphQLBuilder / IServiceRegister.

> Look at AddDocumentListener from the main repo.

Ok...I've got it open...

sungam3r:
Imagine that I want to add some listener that writes logs. It is some sort of "system component" like `services.AddOptions()`. I think `AddOptions` is a very good example here. So... you know that a lot of MS methods call `AddOptions`. So do I with my approaches - I have many wrapping methods across many libraries that call some "base" methods like this one - `AddOptions` or `AddSomeStandardDocumentListener` or whatever. They should be idempotent. What if `AddOptions` is not? Think about that.

sungam3r:
> Are you designing GraphQL-specific services that are interdependent, that are only for GraphQL.NET, such as a document listeners, middleware, type registrations and similar? 

Yes, including but not limited to.

sungam3r:
> Ok...I've got it open...

How to ensure that some sort of document listener was added exactly once? I mean that this method should be fixed as well. In fact I would prefer that ALL methods that inject services for `IEnumerable<>` be written with `TryRegister(...mode:ServiceAndImpl)` instead of `Register`.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
Shane32: I looked at the code.  It appears we can say with certainty that it will not work with factory methods without further changes to the API, as all factories will have an apparent implementation type of `object`.

sungam3r:
Please understand that now I have no any acceptable opportunity to write code correctly. Well, excluding dirty techniques with the casting of the `IGraphQLBuilder/IServiceRegister` object to the known type and then walking through its internals. By the way, it will require an additional dependency on the implementing package 🤦‍♂️ .

Shane32:
> > Ok...I've got it open...
> 
> How to ensure that some sort of document listener was added exactly once? I mean that this method should be fixed as well. In fact I would prefer that ALL methods that inject services for `IEnumerable<>` be written with `TryRegister(...mode:ServiceAndImpl)` instead of `Register`.

I don't know what you'd change.  The code is designed to 'build' your own execution engine.  `AddMiddleware`, for example, injects code to configure the schema during execution, and pulls the registered service directly.  It does not pull `IEnumerable<IFieldMiddleware>` at all, so whether or not double-registration was prevented would have no effect on the operation.

I think you need to realize that `IGraphQLBuilder` is more like building an execution pipeline than it is a DI provider.  Calls to all of the extension methods build the execution engine.  You would need many more changes in order to be able to, for example, call `AddDataLoader` twice without double-registering the `DataLoaderDocumentListener` within the schema.

Let's see:
- `AddSchema` - does not use `IEnumerable`
- `AddDocumentExecuter` does not use `IEnumerable`
- `AddComplexityAnalyzer` does not use `IEnumerable`
- `AddErrorInfoProvider` does not use `IEnumerable`
- `AddGraphTypes` does not use `IEnumerable`
- `AddClrTypeMappings` does not use `IEnumerable`
- `AddDocumentListener` does not use `IEnumerable`
- `AddMiddleware` does not use `IEnumerable`
- `AddDocumentCache` does not use `IEnumerable`
- `AddSerializer` does not use `IEnumerable`
- `ConfigureSchema` and `ConfigureExecutionOptions` - all are delegates of the same type so impossible to know which ones to deduplicate
- `AddValidationRule` does not use `IEnumerable`
- `AddMetrics` does not use `IEnumerable`
- `AddExecutionStrategySelector` does not use `IEnumerable`
- `AddExecutionStrategy` - all implementations are of the same type

And all the other extension methods for serialization, subscriptions, caching, and data loaders reference those base methods.

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
sungam3r: object? Let's consider ` public static IGraphQLBuilder AddErrorInfoProvider<TProvider>(this IGraphQLBuilder builder, Func<IServiceProvider, TProvider> errorInfoProviderFactory)
            where TProvider : class, IErrorInfoProvider`

`ImplementationFactory.GetType().GenericArguments[1]` is some `TProvider`

Shane32:
> an additional dependency on the implementing package

It can't be any more than what you had "last week" - the old `IGraphQLBuilder` just returned `IServiceCollection`.  Casting from `IServiceRegister` to `IServiceCollection` may not be preferred but it certainly isn't creating additional dependencies from the last version.

Shane32:
If this is really important to your projects, however, we can merge it as a courtesy to you -- but I would like tests to see that MS actually supports these methods. I'm pretty sure it will not work with the factory method, unless we rewrite the TryRegister factory method to use generics.  And I do not want to change the entire design of the builder methods and system so the calls can be "idempotent".  Is that okay?

src/GraphQL.ApiTests/GraphQL.MicrosoftDI.approved.txt
Shane32: Oh because the last parameter is `out`.  Ok so that will work.

sungam3r:
>  It can't be any more than what you had "last week" - the old IGraphQLBuilder just returned IServiceCollection. Casting from IServiceRegister to IServiceCollection may not be preferred but it certainly isn't creating additional dependencies from the last version.

No... We speak completely different languages. At all. I repeat - I **have no access** to the stuff like `IServiceCollection`. I build a library layer **on top of** GraphQL.NET package provided API - `IGraphQLBuilder/IServiceRegister` (not on top of `GraphQL.MicrosoftDI`). And so will also do any other library developer. I'm not talking about the app developer.

sungam3r:
Just look at `GraphQL.GraphQLBuilderExtensions` one more time - a lot of APIs on top of `IServiceRegister/IGraphQLBuilder`. **And so do I in my parts of my libraries stack**.

Shane32:
If you had a library written that was compatible with GraphQL Server v5, then `IGraphQLBuilder.Services` returns `IServiceCollection` and any library or nuget package would contain an indirect dependency on `Microsoft.Extensions.DependencyInjection.Abstractions`.

sungam3r:
)))) https://github.com/graphql-dotnet/graphql-dotnet/pull/2946#issuecomment-1039699108   Why server? Or even why **only** server?

sungam3r:
I wrote some methods on top of server project ONLY because there was no such interface in the main repo. That is all. Now it exists and interface from server was removed so it's **obviosly** that  I just switched to the "new" interface from main repo.

sungam3r:
If it will be clearer to explain - I **never** needed a dependency from the server project to do things that I did. I always needed some sort of fluent API to build my APIs on top of.

Shane32:
Ok, so by merging this you will help to eliminate a dependency, perhaps even two or more.  But you certainly had a dependency on MS-DI previously if you were using the IGraphQLBuilder there.

And I'm okay with a PR to help your libraries.  If we are all on the same page here.  I don't want you to expect that we are going to rewrite all the builder methods now because of this added functionality.

sungam3r:
> But you certainly had a dependency on MS-DI previously if you were using the IGraphQLBuilder there.

Of course! https://github.com/graphql-dotnet/graphql-dotnet/pull/2946#issuecomment-1039703858

sungam3r:
>  I don't want you to expect that we are going to rewrite all the builder methods

Rewrite in what way?

sungam3r:
Keep in mind that our own extension methods built on top of `IGraphQLBuilder/IServiceRegister` are subject to the same rules of simple logic as any other external methods written by me or somebodyy else. No one wants to write/use code that works like `AddOptions()` but adds a new registration each time being called. These are not my personal desires at all.

Shane32:
> > I don't want you to expect that we are going to rewrite all the builder methods
> 
> Rewrite in what way?

To be “idempotent”

sungam3r:
> I have a large solution with a bunch of services. Each project has a Startup.cs which is called from the main project. They don't use IServiceRegister at all. 

This is my world. I have a large set of independent or interdependent library repos and even more set of repos with microservices (400, maybe more, I don't count). Each service is built on top of some stack of libraries (20-30). There are 3 main stacks and GraphQL is one of them. So I work in libraries world, I have no any common sln and Startup.cs that I can just call as is from any project in that sln. I even don't own all these services but I provide API stacks for them. These services indeed don't use `IServiceRegister` at all in 99,9% of cases. But I use it (and a bunch of other "low-level" APIs) to provide top-level features to my consumers. Consumers facilitate a lot of features from theirs Startups. Of course some features have some common parts. Think about these parts as `AddOptions`. Some features are built on top of MSDI specific APIs, so I actually have access to `TryAdd` and other stuff. Some features - don't, for example, `IGraphQLBuilder` that became DI-agnostic.

src/GraphQL/DI/IServiceRegister.cs
```diff
@@ -20,16 +20,18 @@ public interface IServiceRegister
...
+        /// injection provider if a service of the same type (and of the same implementation type
+        /// in case of <see cref="RegistrationCompareMode.ServiceTypeAndImplementationType"/>)
+        /// has not already been registered.
         /// </summary>
-        IServiceRegister TryRegister(Type serviceType, Type implementationType, ServiceLifetime serviceLifetime);
+        IServiceRegister TryRegister(Type serviceType, Type implementationType, ServiceLifetime serviceLifetime, RegistrationCompareMode mode = RegistrationCompareMode.ServiceType);
 
-        /// <inheritdoc cref="TryRegister(Type, Type, ServiceLifetime)"/>
-        IServiceRegister TryRegister(Type serviceType, Func<IServiceProvider, object> implementationFactory, ServiceLifetime serviceLifetime);
+        /// <inheritdoc cref="TryRegister(Type, Type, ServiceLifetime, RegistrationCompareMode)"/>
```
Shane32, line 30:
```suggestion
        /// <summary>
        /// Registers the service of type <paramref name="serviceType"/> with the dependency
        /// injection provider if a service of the same type (and of the same implementation type
        /// in case of <see cref="RegistrationCompareMode.ServiceTypeAndImplementationType"/>)
        /// has not already been registered.
        /// <br/><br/>
        /// With <see cref="RegistrationCompareMode.ServiceTypeAndImplementationType"/>, it is required
        /// that <paramref name="implementationFactory"/ is a strongly typed delegate with a return type
        /// of a specific implementation type.
        /// </summary>
```
I think it is worthwhile to note that `Func<IServiceProvider, object>` is inadequate for `implementationFactory` when `mode = RegistrationCompareMode.ServiceTypeAndImplementationType`.  Does this sound good?  Edit as desired.

src/GraphQL/DI/RegistrationCompareMode.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Mode used for <see cref="IServiceRegister.TryRegister(Type, Type, ServiceLifetime, RegistrationCompareMode)">IServiceRegister.TryRegister</see>
+    /// methods family.
+    /// </summary>
+    public enum RegistrationCompareMode
+    {
+        /// <summary>
+        /// Registers the service with the dependency injection provider
+        /// if a service of the same type has not already been registered.
```
Shane32, line 11:
```suggestion
        /// if a service of the same service type has not already been registered.
```

src/GraphQL/DI/RegistrationCompareMode.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// </summary>
+        ServiceType,
+
+        /// <summary>
+        /// Registers the service with the dependency injection provider
+        /// if a service of the same type and of the same implementation type
```
Shane32, line 17:
```suggestion
        /// if a service of the same service type and same implementation type
```

src/GraphQL/DI/RegistrationCompareMode.cs
```diff
@@ -0,0 +1,22 @@
+namespace GraphQL.DI
+{
+    /// <summary>
+    /// Mode used for <see cref="IServiceRegister.TryRegister(Type, Type, ServiceLifetime, RegistrationCompareMode)">IServiceRegister.TryRegister</see>
+    /// methods family.
```
Shane32, line 5:
```suggestion
    /// Comparison mode used for <see cref="IServiceRegister.TryRegister(Type, Type, ServiceLifetime, RegistrationCompareMode)">IServiceRegister.TryRegister</see>
    /// methods.
```

src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
```diff
@@ -124,38 +124,59 @@ public IServiceRegister Register(Type serviceType, object implementationInstance
...
                 throw new ArgumentNullException(nameof(implementationInstance));
 
-            ServiceCollection.TryAdd(new ServiceDescriptor(serviceType, implementationInstance));
+            var descriptor = new ServiceDescriptor(serviceType, implementationInstance);
+            if (mode == RegistrationCompareMode.ServiceType)
+                ServiceCollection.TryAdd(descriptor);
+            else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
+                ServiceCollection.TryAddEnumerable(descriptor);
+            else
+                throw new NotSupportedException(mode.ToString());
```
Shane32, line 178:
```suggestion
                throw new ArgumentOutOfRangeException(nameof(mode));
```

src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
```diff
@@ -124,38 +124,59 @@ public IServiceRegister Register(Type serviceType, object implementationInstance
...
                 throw new ArgumentNullException(nameof(implementationType));
 
-            ServiceCollection.TryAdd(new ServiceDescriptor(serviceType, implementationType, TranslateLifetime(serviceLifetime)));
+            var descriptor = new ServiceDescriptor(serviceType, implementationType, TranslateLifetime(serviceLifetime));
+            if (mode == RegistrationCompareMode.ServiceType)
+                ServiceCollection.TryAdd(descriptor);
+            else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
+                ServiceCollection.TryAddEnumerable(descriptor);
+            else
+                throw new NotSupportedException(mode.ToString());
```
Shane32, line 159:
```suggestion
                throw new ArgumentOutOfRangeException(nameof(mode));
```

src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
```diff
@@ -124,38 +124,59 @@ public IServiceRegister Register(Type serviceType, object implementationInstance
...
                 throw new ArgumentNullException(nameof(implementationFactory));
 
-            ServiceCollection.TryAdd(new ServiceDescriptor(serviceType, implementationFactory, TranslateLifetime(serviceLifetime)));
+            var descriptor = new ServiceDescriptor(serviceType, implementationFactory, TranslateLifetime(serviceLifetime));
+            if (mode == RegistrationCompareMode.ServiceType)
+                ServiceCollection.TryAdd(descriptor);
+            else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
+                ServiceCollection.TryAddEnumerable(descriptor);
+            else
+                throw new NotSupportedException(mode.ToString());
```
Shane32, line 140:
```suggestion
                throw new ArgumentOutOfRangeException(nameof(mode));
```

src/GraphQL.MicrosoftDI/GraphQLBuilder.cs
```diff
@@ -124,38 +124,59 @@ public IServiceRegister Register(Type serviceType, object implementationInstance
...
                 throw new ArgumentNullException(nameof(serviceType));
             if (implementationFactory == null)
                 throw new ArgumentNullException(nameof(implementationFactory));
 
-            ServiceCollection.TryAdd(new ServiceDescriptor(serviceType, implementationFactory, TranslateLifetime(serviceLifetime)));
+            var descriptor = new ServiceDescriptor(serviceType, implementationFactory, TranslateLifetime(serviceLifetime));
+            if (mode == RegistrationCompareMode.ServiceType)
+                ServiceCollection.TryAdd(descriptor);
+            else if (mode == RegistrationCompareMode.ServiceTypeAndImplementationType)
+                ServiceCollection.TryAddEnumerable(descriptor);
```
Shane32, line 138: We should add tests to hit all new lines of code.  I've got the rest of the Register/TryRegister methods at 99% coverage and it should not be hard to duplicate some of the tests to check TryAddEnumerable and the ArgumentOutOfRange exceptions.

Shane32: https://app.codecov.io/gh/graphql-dotnet/graphql-dotnet/blob/develop/src/GraphQL.MicrosoftDI/GraphQLBuilder.cs

sungam3r: done

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -298,11 +298,14 @@ namespace GraphQL
             where TImplementation :  class, TService { }
         public static GraphQL.DI.IServiceRegister TryRegister<TService>(this GraphQL.DI.IServiceRegister services, GraphQL.DI.ServiceLifetime serviceLifetime)
             where TService :  class { }
-        public static GraphQL.DI.IServiceRegister TryRegister<TService>(this GraphQL.DI.IServiceRegister services, TService implementationInstance)
-            where TService :  class { }
         public static GraphQL.DI.IServiceRegister TryRegister<TService>(this GraphQL.DI.IServiceRegister services, System.Func<System.IServiceProvider, TService> implementationFactory, GraphQL.DI.ServiceLifetime serviceLifetime)
```
Shane32, line 301: Remove this overload in favor of the new one that has a default parameter?

sungam3r: I see only one overload with implementationInstance -         `public static GraphQL.DI.IServiceRegister TryRegister<TService>(this GraphQL.DI.IServiceRegister services, TService implementationInstance, GraphQL.DI.RegistrationCompareMode mode = 0)`.

Shane32: Nevermind I figured it out.

Shane32:
I'm looking for tests to be added to `GraphQL.MicrosoftDI.Tests.GraphQLBuilderTests`.

The method `TryRegister_InvalidParameters` should check for invalid arguments.

We have some `TryRegister` methods for instance, factory and so on.

Then we have `TryRegister_Succeed` and `TryRegister_Fail` that tests that attempting a registration when it already exists works properly.  We should probably do something similar for the service and implementation variant.

I can write the code if you want, but please take a look at it.  It is not easy code, but it fully tests the MicrosoftDI provider in almost every different combination.

sungam3r:
@Shane32 so? It's a blocker for me and my team now.

Shane32:
> > > I don't want you to expect that we are going to rewrite all the builder methods
> > 
> > 
> > Rewrite in what way?
> 
> To be “idempotent”

Ok?

sungam3r:
I can answer like this - if logic needs to make methods working with `IEnumerable<>` idempotent, then I would do them idempotent regardless of which part of the code they are located. https://github.com/graphql-dotnet/graphql-dotnet/pull/2946#issuecomment-1039707376 

I have already explained my arguments. It is better to have one working DI provider than 5/6 providers working equally badly because they can handle only thin subset of all features. And by the way, when I talk about feature sets I absolutely do not think that this PR adds some extraordinary feature that goes beyond common use-cases.

Shane32:
I really don't know what you're saying.  I do not want the builder methods rewritten so `AddMiddleware<CountFieldsMiddleware>()` and `AddDataLoader()` and so on is idempotent.  I've already listed each builder method in the main project and noted how they would not work (as currently written) to be idempotent.  If you want to change this design, then I do not want to merge this PR.  If you want this PR purely for your own needs, that's fine.

Either you agree or diagree; pick one.

If there's underlying logic like AddOptions that should logically be (and can easily be) idempotent, such as `AddDefaultEndpointSelectorPolicy`, we can discuss at that time, but I do not want it assumed we are going to rewrite the entire builder class to be idempotent.  The class is not designed to be used in this manner, and my opinion is that there is no need for it from GraphQL.NET's perspective.  (I believe GraphQL.NET would benefit more from having a simpler `IServiceRegister` design.)  If you have a project with "400 microservices", surely you can write a wrapper for your own needs.  It certainly is an advanced use case, not a typical one.

sungam3r:
> If there's underlying logic like `AddOptions` that should logically be (and can easily be) idempotent, such as AddDefaultEndpointSelectorPolicy, we can use the new RegistrationCompareMode

_that should logically be (and can easily be)_ - I'm talking about such cases.

> we are going to rewrite the entire builder class to be idempotent

Not my goal. I'm fine to reapply middlewares more and more times because middleware intentionally wraps resolver.

> Interesting that Autofac, the second-most-common DI provider

I have nothing to say about Autofac. I never used it and I know nobody who used it. I worked with Ninject some time ago.

Shane32 (approved):
I will likely add more tests later to fit with the rest of my tests.

PR #2949, opened 2022-02-15
Lock SDK version on 6.0.102

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/runs/5199930740?check_suite_focus=true works
https://github.com/graphql-dotnet/graphql-dotnet/runs/5207538082?check_suite_focus=true fails

### DIFF
--- a/.github/workflows/build-artifacts-code.yml
+++ b/.github/workflows/build-artifacts-code.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Setup .NET Core 6.0 SDK
         uses: actions/setup-dotnet@v1
         with:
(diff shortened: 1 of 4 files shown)


sungam3r:
Because something is broken in 6.0.200 SDK . See link from the first comment:
`
Unhandled exception: System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.
Could not load file or assembly 'Microsoft.CodeAnalysis.CSharp, Version=4.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
`

The same in server project and so on.

PR #2956, opened 2022-02-20
Updates.

sioanis:
To Do: 

- [ ] Change the base branch from `master` to `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,13 +364,13 @@ If you have a PR with some breaking changes, then please target it to the `devel
 
 ## Backers
 
-Thank you to all our backers! 🙏 [Become a backer](https://opencollective.com/graphql-net#backer).
+Thank you to all our backers! 🙏 [Back this project](https://opencollective.com/graphql-net#backer).
 
 <a href="https://opencollective.com/graphql-net#backers" target="_blank"><img src="https://opencollective.com/graphql-net/backers.svg?width=890"></a>
 
 ## Sponsors
 
-Support this project by becoming a sponsor. Your logo will show up here with a link to your website. [Become a sponsor](https://opencollective.com/graphql-net#sponsor).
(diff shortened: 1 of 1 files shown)


sungam3r:
What is the goal?

sioanis:
The goal is to complete the to-do before merging.

sioanis:
I created a fork to try this out.

sungam3r:
I mean what is the goal to rename master into main?

sioanis:
Yes.

sungam3r:
Sorry. Perhaps I was not clear, I am not a native speaker. My question is **why** to rename master in main at all?

sioanis:
Old pepper doesn't like this. We should keep the branch.

Screen Shot 2022-02-20 at 3 12 40 PM

sungam3r:
😕 I do not know who it is. Probably his opinion is important for you. Well, OK.

sioanis:
This is the Github reasoning for renaming. (https://github.com/github/renaming)

sioanis:
Whoever has the name old pepper probably has some good ideas.

sungam3r:
Thanks. I briefly read this documentation, but nevertheless I did not understand what is the reason. There is described **what** is changing and **where**, but it does not say **why** this is done. The only proposal from that document that should answer this question (as I think) sounds so
> We like it because it's short, it keeps your muscle memory intact, and it translates well across most languages.

I can say the same thing about `master`. Taking into account the amount of work to be done in addition to renaming (change  all CI workflows, change settings of git clients) and the dubiousness of the above arguments, I would not rename branch. In addition, following this renaming it needs to be performed in all the repositories of the graphql-dotnet org to have the same structure in all repos.

sioanis:
Sounds good.

PR #2959, opened 2022-02-21
Add SchemaBuilder.RunConfigurations

sungam3r:


sungam3r:
@Shane32 I got deadlock in one of my tests where I configured DI as
 `services.AddSchema<Schema>(p => p.GetRequiredService<SchemaFetcher>().GetSchemaAsync().GetAwaiter().GetResult())`.

 Another thread was waiting on
`if (services.GetService(typeof(IEnumerable<IConfigureSchema>)) is IEnumerable<IConfigureSchema> configurations)` from `Schema`'s ctor. 😕 

`SchemaFetcher` uses `SchemaBuilder.Build` and passes `ServiceProvider`.

sungam3r:
So I want to be able to set `RunConfigurations = false` to my `SchemaBuilder` to eliminate `GetService<>` call from ctor.

sungam3r:
I merge this one and go on.

sungam3r:
From ` Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver`:
```csharp
 private object VisitCache(ServiceCallSite callSite, RuntimeResolverContext context, ServiceProviderEngineScope serviceProviderEngine, RuntimeResolverLock lockType)
        {
            bool lockTaken = false;
            System.Collections.Generic.Dictionary<ServiceCacheKey, object> resolvedServices = serviceProviderEngine.ResolvedServices;

            // Taking locks only once allows us to fork resolution process
            // on another thread without causing the deadlock because we
            // always know that we are going to wait the other thread to finish before
            // releasing the lock
            if ((context.AcquiredLocks & lockType) == 0)
            {
                Monitor.Enter(resolvedServices, ref lockTaken);   // <---- second thread waits here waiting for  services.GetService(typeof(IEnumerable<IConfigureSchema>)) in Schema ctor
            }

            try
            {
                if (!resolvedServices.TryGetValue(callSite.Cache.Key, out object resolved))
                {
                    resolved = VisitCallSiteMain(callSite, new RuntimeResolverContext  // <---- first thread inside here waiting for .GetSchemaAsync().GetAwaiter().GetResult()  in my startup (test)
                    {
                        Scope = serviceProviderEngine,
                        AcquiredLocks = context.AcquiredLocks | lockType
                    });
```

😕

sungam3r:
`SchemaFetcher` - my class from my codebase. It's just an example. I still did not comprehend the essence of the problem.

sungam3r:
`SyncronizationContext.Current` is null. I checked.

Shane32:
It doesn’t have anything to do with synchronization contexts - they don’t exist in ASP.Net Core, which is why ConfigureAwait isn’t necessary in ASP.Net Core apps. 

You’re running an asynchronous method from within the DI factory method. The async method will be executed via the task scheduler on a separate thread while GetAwaiter.GetResult blocks the current thread. However the lock inside MSDI (Monitor.Enter aka lock) isn’t allowing another thread to initialize other objects. 

Now if the async method didn’t call GetRequiredService, you’d be fine, and MSDI would just block until initialization was complete. But since there are nested services to resolve, it won’t work within an asynchronous factory.

Shane32:
That’s probably why setting RunConfigurations to false appears to fix the problem - because the Schema doesn’t call GetService and trigger the deadlock. But there are likely other similar ways to trigger the deadlock, like if the SchemaFactory were to attempt to initialize the schema

Shane32:
Any call to GetService within the asynchronous method would likely trigger the deadlock

sungam3r:
> It doesn’t have anything to do with synchronization contexts - they don’t exist in ASP.Net Core, which is why ConfigureAwait isn’t necessary in ASP.Net Core apps.

I know, just a note.

> That’s probably why setting RunConfigurations to false appears to fix the problem - because the Schema doesn’t call GetService and trigger the deadlock.

Yeah. I got the same deadlock but a bit later 😖 

> However the lock inside MSDI (Monitor.Enter aka lock) isn’t allowing **another** thread to initialize other objects.
> Any call to GetService within the asynchronous method would likely trigger the deadlock

I agree. I also come to this conclusion. It turns out that we cannot use asynchronous method of obtaining a schema. But this is not our fault? After all, a similar error would be obtained in the case of any other service, not just a schema, right? So it's by design in MSDI world.

PR #2960, opened 2022-02-21
Add SchemaBuilder.CreateSchema

sungam3r:
Rel: #2959 . I continue to fight deadlocks in my tests. Now I want to be able to override `Schema.Validate` using schema built from `SchemaBuilder` so I need new protected method to return my subsclass with overridden method.


sungam3r:
@Shane32 Changes are backward compatible so I merge ASAP and go on.

sungam3r:
Well. I still get deadlock when I try to resolve any service from `ServiceProvider`.

PR #2962, opened 2022-02-22
Fix rounding of decimal numbers for STJ

### LINKED ISSUE #2958: Incorrect rounding of decimal numbers
## Description

The issue is reproduced for decimal numbers with max available precision, scale (16, 2)
For mutation model with decimal type field as a result of parsing the value is rounding incorrectly 
 e. g. 12345678901234.56 -> 12345678901234.6

 
### Steps to reproduce

- Request:

- Mutation Model:

```
public class MutationTestModel
    {
        public decimal Discount { get; set; }
    }
```

- Mutation Type:
```
 public sealed class MutationTestType : InputObjectGraphType<MutationTestModel>
    {
        public MutationTestType()
        {
            Field(x => x.Discount);
        }
    }
```

 For getting requests variables GraphQL.SystemTextJson.StringExtensions is used
 ```
     var request = await stream.FromJsonAsync<T>();
 ```
  
 The issue is related to *GraphQL.SystemTextJson.InputsConverter: private static object ReadNumber(ref Utf8JsonReader reader)*.
 **Any decimal number from the request is parsed as Double and *reader.TryGetDecimal(out decimal dm)* method is never used**.
 
 ``` 
        private static object ReadNumber(ref Utf8JsonReader reader)
        {
            if (reader.TryGetInt32(out int i))
                return i;
            else if (reader.TryGetInt64(out long l))
                return l;
            else if (JsonConverterBigInteger.TryGetBigInteger(ref reader, out var bi))
                return bi;
            else if (reader.TryGetDouble(out double d))
                return d;
            else if (reader.TryGetDecimal(out decimal dm))
                return dm;
            ...
 ```
 

Therefore, when the input is mapped to the query model, double is parsed to decimal, and for numbers with the maximum available precision, the scale (16, 2) is rounded incorrectly.
 

### Expected result

The decimal field value should be the same as in the request
12345678901234.56 -> 12345678901234.56

### Actual result

The decimal field value is incorrect
12345678901234.56 -> 12345678901234.6

### Environment

GraphQL Version: 4.7.0
FrameWork: .Net 6.0

sungam3r:
Fixes #2958 

@[dulcisclade](https://github.com/dulcisclade) note that for NSJ you may set `settings.FloatParseHandling = Newtonsoft.Json.FloatParseHandling.Decimal`

### DIFF
--- a/src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
+++ b/src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
@@ -8,6 +8,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="..\GraphQL\Shim\DecimalData.cs" Link="DecimalData.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference Include="..\GraphQL\GraphQL.csproj" />
     <PackageReference Include="System.Text.Json" Version="4.7.2" />
   </ItemGroup>
--- a/src/GraphQL.SystemTextJson/InputsJsonConverter.cs
+++ b/src/GraphQL.SystemTextJson/InputsJsonConverter.cs
@@ -82,15 +82,39 @@ namespace GraphQL.SystemTextJson
         private static object ReadNumber(ref Utf8JsonReader reader)
         {
             if (reader.TryGetInt32(out int i))
+            {
                 return i;
+            }
             else if (reader.TryGetInt64(out long l))
+            {
                 return l;
+            }
             else if (JsonConverterBigInteger.TryGetBigInteger(ref reader, out var bi))
+            {
                 return bi;
-            else if (reader.TryGetDouble(out double d))
-                return d;
-            else if (reader.TryGetDecimal(out decimal dm))
-                return dm;
+            }
+            else
+            {
+                bool isDouble = reader.TryGetDouble(out double dbl);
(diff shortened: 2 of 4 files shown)


sungam3r:
@Shane32 I would not backport this fix into v4.x

src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
```diff
@@ -7,6 +7,10 @@
     <PackageTags>GraphQL;json</PackageTags>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Include="..\GraphQL\Shim\DecimalData.cs" Link="DecimalData.cs" />
```
sungam3r, line 11: Now both serializers profit from `DecimalData` trick.

Shane32:
So we finally got rid of this horrid code from literal parsing and have to add it right back for variable parsing, eh?

Shane32 (approved):

Shane32:
> @Shane32 I would not backport this fix into v4.x

Any particular reason?

sungam3r:
> So we finally got rid of this horrid code from literal parsing and have to add it right back for variable parsing, eh?

No, this horrid code works well for both literal and variable parsing 😄 and for both serializers:
изображение

> Any particular reason?

I'm lazy.

Shane32:
Whether it works well or not, we are parsing the floating point variable three times plus a comparison every time a floating point number is encountered in the JSON.  It bugs me.

Issue #2865, opened 2022-01-18
Stop modify serializer options


sungam3r:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/2859/files#r786709872

sungam3r:
I ran into this porting my code on v5.

Shane32:
@sungam3r The above PRs were merged -- is this complete?

sungam3r:
Nope. Search for `// TODO: fix this: it modifies serializerOptions`.

PR #2965, opened 2022-02-24
Fix STJ bug

sungam3r:
Rel: #2865 #2797 

Before:
> System.InvalidOperationException: 'Serializer options cannot be changed once serialization or deserialization has occurred.'

### DIFF
--- a/src/GraphQL.SystemTextJson/GraphQLSerializer.cs
+++ b/src/GraphQL.SystemTextJson/GraphQLSerializer.cs
@@ -129,7 +129,7 @@ namespace GraphQL.SystemTextJson
                 SerializerOptions.Converters.Add(new ApolloTraceJsonConverter());
             }
 
-            if (!SerializerOptions.Converters.Any(c => c.CanConvert(typeof(JsonConverterBigInteger))))
+            if (!SerializerOptions.Converters.Any(c => c.CanConvert(typeof(System.Numerics.BigInteger))))
             {
                 SerializerOptions.Converters.Add(new JsonConverterBigInteger());
             }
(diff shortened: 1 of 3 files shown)


sungam3r:
I want to note that this error manifested itself thanks to double registration of `IGraphQLSerializer/IGraphQLTextSerializer`:

```csharp
    public static IGraphQLBuilder AddSerializer<TSerializer>(this IGraphQLBuilder builder)
            where TSerializer : class, IGraphQLSerializer
        {
            builder.Services.Register<IGraphQLSerializer, TSerializer>(ServiceLifetime.Singleton, true);
            if (typeof(IGraphQLTextSerializer).IsAssignableFrom(typeof(TSerializer)))
                builder.Services.Register(typeof(IGraphQLTextSerializer), typeof(TSerializer), ServiceLifetime.Singleton, true);
            // builder.Services.Register(services => (IGraphQLTextSerializer)services.GetRequiredService<IGraphQLSerializer>(), ServiceLifetime.Singleton);
            return builder;
        }
```

Initialization of json options occurred 2 times and it is still an issue. I believe we should
1) Stop modifying json options.
2) Remove double registration.

Even after fixing the current bug, one can create a situation where the same json options object will be changed after serialization yielding `InvalidOperationException`.

Shane32 (approved):

Shane32:
I’m available to review PRs that would fix the other issues you mentioned. I would fix them separately not together. I am not ready to forfeit the design of the options pattern if it doesn’t allow for a way to fix the first issue. However we discussed using inheritance to set defaults. I’m not sure if that works or not for IErrorInfoProvider.

PR #2966, opened 2022-02-24
Have schema-first utilize method builder from AutoRegisteringObjectGraphType

Shane32:
- Add `[FromUserContext]` attribute
- Change `MethodResolver` to `MemberResolver` and consolidate functionality therein
- Prep `MemberResolver.BuildFieldResolver` method for use with ValueTask (coming soon in PR #2907 )
- Precompiles field resolvers
- Requires `[FromSource]` and `[FromUserContext]` added to user's code, otherwise fully backwards-compatible (I believe)
- Supports `[FromServices]` and other custom/new attributes
- Supports a renamed method argument (e.g. method argument `id` pulls from query argument `nameId` by adding `[Name("nameId")]` attribute to method)
- Supports subscription methods (not added to `AutoRegisteringObjectGraphType` quite yet; future PR)
- Schema-first objects attempt to pull from `RequestServices`, if available, rather than always using the root service provider; this allows for scoped objects and does not change singleton/transient objects

Todo:
- [x] Add migration notes


src/GraphQL/Reflection/AccessorFieldResolver.cs
```diff
@@ -1,33 +0,0 @@
...
-                    ? context.Source
-                    : _serviceProvider.GetService(_accessor.DeclaringType);
-
-            if (target == null)
-            {
-                var parentType = context.ParentType != null ? $"{context.ParentType.Name}." : null;
-                throw new InvalidOperationException($"Could not resolve an instance of {_accessor.DeclaringType.Name} to execute {parentType}{context.FieldAst.Name}");
-            }
-
-            return _accessor.GetValue(target, arguments);
```
Shane32, line 30: This deleted code is now contained within `AutoRegisteringHelper.BuildInstanceExpressionForSchemaBuilder` and executes (precompiled) in the same fashion, with the exception that if `context.RequestServices` is not null, it is used rather than `_serviceProvider`.

src/GraphQL/Reflection/AccessorFieldResolver.cs
```diff
@@ -1,33 +0,0 @@
...
-
-        public AccessorFieldResolver(IAccessor accessor, IServiceProvider serviceProvider)
-        {
-            _accessor = accessor;
-            _serviceProvider = serviceProvider;
-        }
-
-        public object? Resolve(IResolveFieldContext context)
-        {
-            var arguments = ReflectionHelper.BuildArguments(_accessor.Parameters, context);
```
Shane32, line 18: My goal is to delete the BuildArguments method.  So far it is still used elsewhere, for instance by the `FieldDelegate` method.

src/GraphQL/Resolvers/EventStreamMethodResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Reflection;
+using GraphQL.Subscription;
+
+namespace GraphQL.Resolvers
+{
+    /// <summary>
+    /// A precompiled event stream resolver for a specific <see cref="MethodInfo"/>.
+    /// Calls the specified method (with the specified arguments) and returns the value of the method.
+    /// </summary>
+    internal class EventStreamMethodResolver : MemberResolver, IAsyncEventStreamResolver
```
Shane32, line 11: This resolver implements both `IAsyncEventStreamResolver` and `IFieldResolver` and inherits much of its functionality from `MemberResolver`.  I did not attempt to implement `IEventStreamResolver`, for two reasons (1) a single `Task` allocation when setting up a new event stream is not so bad, (2) once we switch to `ValueTask`, there will only be one interface anyway.

Shane32: The reason this resolver implements `IFieldResolver` is because a typical event stream resolver returns an `IObservable<T>` and then when there is a data push from the event stream, GraphQL executes the `IFieldResolver` for the same field, with `T` as the root, which in turn is immediately returned as the source for the child graph.  For `AutoRegisteringObjectGraphType` this will be simplified so that any method that returns `IObservable<T>` will automatically set up a subscription field and handle this legwork.  If a user really wants the child graph to use a child property of the observable, they can call Select on the observable before returning it.  All this allows for proper type inference.

Long story short: it's prepwork for a subsequent PR and not needed quite yet.

src/GraphQL/Resolvers/EventStreamMethodResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+                }
+            }
+
+            throw new InvalidOperationException("Method must return a IObservable<T> or Task<IObservable<T>> where T is a reference type.");
+
+            Func<IResolveFieldContext, object?> Complete(Expression bodyExpression)
+            {
+                var lambda = Expression.Lambda<Func<IResolveFieldContext, Task<IObservable<object?>>>>(bodyExpression, resolveFieldContextParameter);
+                _eventStreamResolver = lambda.Compile();
+                return context => context.Source;
```
Shane32, line 71: Here we can see the `IFieldResolver` implementation is always `context => context.Source` for a subscription field.  This is correct.

src/GraphQL/Resolvers/MemberResolver.cs
```diff
@@ -0,0 +1,168 @@
...
+                    methodInfo,
+                    expressionBodies);
+
+            _resolver = BuildFieldResolver(resolveFieldContextParameter, methodCallExpr);
+        }
+
+        /// <summary>
+        /// Creates an appropriate resolver function based on the return type of the expression body.
+        /// </summary>
+        protected virtual Func<IResolveFieldContext, object?> BuildFieldResolver(ParameterExpression resolveFieldContextParameter, Expression bodyExpression)
```
Shane32, line 138: The contents of this method will expand drastically once we implement ValueTask.  You can see a prototype of the code in the draft PR #2907 now.  Regardless, this protected method is overridden now by the EventStreamMethodResolver.

src/GraphQL/Resolvers/MemberResolver.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Reflection;
+
+namespace GraphQL.Resolvers
+{
+    /// <summary>
+    /// A precompiled field resolver for a specific <see cref="MethodInfo"/>, <see cref="PropertyInfo"/> or <see cref="FieldInfo"/>.
+    /// Returns the specified field or property, or for methods, calls the specified method (with the specified arguments)
+    /// and returns the value of the method.
+    /// </summary>
+    internal class MemberResolver : IFieldResolver
```
Shane32, line 11: Mostly this class is a renamed `MethodResolver` with added constructors for `FieldInfo` and `PropertyInfo`.  There will be significant code deduplication with this design once ValueTask is implemented in PR #2907 .

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -57,12 +63,12 @@ public ArgumentInformation(ParameterInfo parameterInfo, Type sourceType, FieldTy
...
         /// Should equal <c>TSourceType</c> within <see cref="AutoRegisteringObjectGraphType{TSourceType}"/>.
         /// </summary>
-        public Type SourceType { get; }
+        public Type? SourceType { get; }
 
         /// <summary>
         /// The <see cref="Types.FieldType"/> that the query argument will be added to.
         /// </summary>
-        public FieldType FieldType { get; }
+        public FieldType? FieldType { get; }
```
Shane32, line 71: The schema builder does not know what type the source is, as far as I can tell.  It also has not built the field at the time that it attempts to build the field resolver.  The latter we could likely change.  For now, to make as little changes in the schema builder as possible, these properties (which were really just informative and not absolutely necessary before) are now nullable.

src/GraphQL/Types/ArgumentInformation.cs
```diff
@@ -18,29 +19,34 @@ public class ArgumentInformation
...
             if (parameterInfo.ParameterType == typeof(IResolveFieldContext))
             {
                 Expression<Func<IResolveFieldContext, IResolveFieldContext>> expr = x => x;
                 Expression = expr;
             }
+            else if (parameterInfo.ParameterType == typeof(IResolveEventStreamContext))
+            {
+                Expression<Func<IResolveFieldContext, IResolveEventStreamContext>> expr = x => (IResolveEventStreamContext)x;
+                Expression = expr;
+            }
```
Shane32, line 49: Not ideal, but to resolve an `IResolveEventStreamContext` as a method argument the context gets implicitly cast to `IResolveFieldContext` and then explicitly cast back to `IResolveEventStreamContext`.

src/GraphQL/Reflection/AccessorFieldResolver.cs
Shane32: See `AutoRegisteringHelper.BuildSourceExpressionForSchemaBuilderInternal_GetSource<T>`

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -1,10 +1,116 @@
...
+
+            var target = typeof(T).IsInstanceOfType(source)
+                ? (T)source!
+                : (T?)(context.RequestServices ?? serviceProvider).GetService(typeof(T));
+
+            if (target == null)
+            {
+                var parentType = context.ParentType != null ? $"{context.ParentType.Name}." : null;
+                throw new InvalidOperationException($"Could not resolve an instance of {typeof(T).Name} to execute {parentType}{context.FieldAst.Name}");
+            }
```
Shane32, line 109: This code moved from `AccessorFieldResolver`; see notes there.

Shane32:
@sungam3r This PR is ready for review, besides my having to write migration documentation.  Please review it if you have time.  Once I write the migration documentation then I'll remove the [WIP] tag and request or re-request a review.

Shane32:
> @sungam3r This PR is ready for review, besides my having to write migration documentation. Please review it if you have time. Once I write the migration documentation then I'll remove the [WIP] tag and request or re-request a review.

Finished some documentation.

sungam3r:
I'll start tomorrow.

sungam3r:
Sorry for delay. I was busy and I will 100% start reviewing tomorrow.

src/GraphQL.Tests/Utilities/SchemaBuilderNestedTypesTests.cs
```diff
@@ -82,19 +82,15 @@ public class Character
...
         // [GraphQLMetadata("Droid", IsTypeOf = typeof(Droid))]
         public class DroidType
         {
-            public string Id(Droid droid) => droid.Id;
+            public string Id([FromSource] Droid droid) => droid.Id;
 
-            public string Name(Droid droid) => droid.Name;
+            public string Name([FromSource] Droid droid) => droid.Name;
 
             [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "for tests")]
```
sungam3r, line 92:
```suggestion
```

?

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -310,22 +314,24 @@ private void InitializeSubscriptionField(FieldConfig config, Type? parentType)
...
 
-                if (config.SubscriberAccessor.MethodInfo.ReturnType.GetGenericTypeDefinition() == typeof(Task<>))
-                {
-                    config.AsyncSubscriber = new AsyncEventStreamResolver(config.SubscriberAccessor, ServiceProvider);
-                }
-                else
-                {
-                    config.Subscriber = new EventStreamResolver(config.SubscriberAccessor, ServiceProvider);
-                }
+                config.AsyncSubscriber = AutoRegisteringHelper.BuildEventStreamResolver(
```
sungam3r, line 330: Where is Subscriber  (non async)?

docs2/site/docs/getting-started/introduction.md
```diff
@@ -217,12 +217,12 @@ public class Query
 [GraphQLMetadata("Droid", IsTypeOf=typeof(Droid))]
 public class DroidType
 {
-  public string Id(Droid droid) => droid.Id;
-  public string Name(Droid droid) => droid.Name;
+  public string Id([FromSource] Droid droid) => droid.Id;
```
sungam3r, line 220: What will happen if you do not specify this attribute?

src/GraphQL/FromSource.cs
```diff
@@ -11,7 +11,7 @@ public class FromSourceAttribute : GraphQLAttribute
         /// <inheritdoc/>
         public override void Modify<TParameterType>(ArgumentInformation argumentInformation)
         {
-            if (!argumentInformation.ParameterInfo.ParameterType.IsAssignableFrom(argumentInformation.SourceType))
+            if (argumentInformation.SourceType != null && !argumentInformation.ParameterInfo.ParameterType.IsAssignableFrom(argumentInformation.SourceType))
```
sungam3r, line 14: FromSource.cs -> FromSourceAttribute.cs

docs2/site/docs/getting-started/introduction.md
Shane32: It would be treated as an input argument, e.g. `GetArgument<Droid>(“droid”)`

src/GraphQL/Utilities/SchemaBuilder.cs
Shane32: Will not exist once conversion to ValueTask is completed so I did not bother to write code for it. Besides, as it exists in this PR, synchronous subscribers will get wrapped in Task FromResult. Which is not the worst thing for a single allocation when setting up a subscription.

src/GraphQL.Tests/Utilities/SchemaBuilderNestedTypesTests.cs
Shane32: 🤷🏻‍♂️

src/GraphQL/FromSource.cs
Shane32: Done

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
```diff
@@ -859,24 +859,24 @@ public Pet Pet(PetKind type)
...
         public string ResolveWithParam(IResolveFieldContext context, string id) => $"Resolved {id}";
 
-        public string UserContext(MyUserContext context) => context.Name;
+        public string UserContext([FromUserContext] MyUserContext context) => context.Name;
 
-        public string UserContextWithParam(MyUserContext context, string id) => $"{context.Name} {id}";
+        public string UserContextWithParam([FromUserContext] MyUserContext context, string id) => $"{context.Name} {id}";
 
-        public bool Three(IResolveFieldContext resolveContext, object source, MyUserContext context)
+        public bool Three(IResolveFieldContext resolveContext, [FromSource] object source, [FromUserContext] MyUserContext context)
```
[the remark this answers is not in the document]
sungam3r: ok

docs2/site/docs/migrations/migration5.md
```diff
@@ -415,6 +417,64 @@ servcies.AddGraphQL(builder => builder
...
+    public IEnumerable<Droid> Friends([FromSource] Droid source) => _repo.FriendsOf(source.Id);
+}
+
+// v5 alternate
+[GraphQLMetadata("Droid")]
+class DroidType
+{
+    public int Id([FromSource] Droid source) => source.Id;
+
+    // only inject Repostiory where needed
```
sungam3r, line 473:
```suggestion
    // only inject Repository where needed
```

docs2/site/docs/migrations/migration5.md
```diff
@@ -415,6 +417,64 @@ servcies.AddGraphQL(builder => builder
...
+    public int Id([FromSource] Droid source) => source.Id;
+
+    public IEnumerable<Droid> Friends([FromSource] Droid source) => _repo.FriendsOf(source.Id);
+}
+
+// v5 alternate
+[GraphQLMetadata("Droid")]
+class DroidType
+{
+    public int Id([FromSource] Droid source) => source.Id;
```
sungam3r, line 471: Of course, with the attribute, it is clear how to get value for method.

src/GraphQL/FromUserContextAttribute.cs
```diff
@@ -0,0 +1,18 @@
+using GraphQL.Execution;
+using GraphQL.Types;
+
+namespace GraphQL
```
sungam3r, line 4: Does your IDE suggest you to use file scoped namespaces?

Shane32: No, at least, not as a warning.  But I can use VS to easily refactor as a file-scoped namespace.  (As it is now this file matches the rest of the codebase.)

docs2/site/docs/migrations/migration5.md
sungam3r: `IsInstanceOfType` from existing code looks not very good. As well as whole `ReflectionHelper.BuildArguments`.

src/GraphQL/FromUserContextAttribute.cs
sungam3r: I configured this rule  as suggestion so VS should respect it when creating new files.

docs2/site/docs/migrations/migration5.md
Shane32: `ReflectionHelper.BuildArguments` will be removed once I finish PRs to alter/remove the methods that depend on it.

src/GraphQL/FromUserContextAttribute.cs
Shane32: Ok but let's keep the codebase uniform.  In this case, the file was copied from `FromSourceAttribute` and slightly modified.

Shane32:
> I will finish review tomorrow. Recently, it has become harder to check the code, the fatigue affects.

Ok.  Note that it will become increasingly difficult to work on the project.  Already there is a PR waiting on this one, and I will be pushing another PR soon that will also depend on prior PRs.

src/GraphQL/FromUserContextAttribute.cs
sungam3r: ok

src/GraphQL/Resolvers/MemberResolver.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Initializes an instance for the specified property, using the specified instance expression to access the instance of the property.
+        /// If <paramref name="instanceExpression"/> is <see langword="null"/> then a default instance expression is used as follows:
+        /// <code>context =&gt; (TSourceType)context.Source</code>
+        /// </summary>
+        public MemberResolver(PropertyInfo propertyInfo, LambdaExpression? instanceExpression)
+            : this((propertyInfo ?? throw new ArgumentNullException(nameof(propertyInfo))).GetMethod ?? throw new ArgumentException("No 'get' method for the supplied property.", nameof(propertyInfo)), Array.Empty<LambdaExpression>(), instanceExpression)
```
sungam3r, line 64:
```suggestion
            : this((propertyInfo ?? throw new ArgumentNullException(nameof(propertyInfo))).GetMethod ?? throw new ArgumentException($"No 'get' method for the supplied {propertyInfo.Name} property.", nameof(propertyInfo)), Array.Empty<LambdaExpression>(), instanceExpression)
```

src/GraphQL/Resolvers/MemberResolver.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <code>context =&gt; (TSourceType)context.Source</code>
+        /// </summary>
+        public MemberResolver(FieldInfo fieldInfo)
+            : this(fieldInfo, null)
+        {
+        }
+
+        /// <summary>
+        /// Initializes an instance for the specified field, using the specified instance expression to access the instance of the field.
+        /// If <paramref name="instanceExpression"/> is <see langword="null"/> then a default instance expression is used as follows:
```
sungam3r, line 26: > then a default instance expression is used as follows: <code>context =&gt; (TSourceType)context.Source</code>

Is such an expression used by default makes sense? Does it make any sense to use any default value? Is it worth demanding it always explicitly?

Shane32: Yes, I expect those constructors can be removed.  Although `ExpressionFieldResolver` and `NameFieldResolver` need to build a 'default instance expression of `context => context.Source`' in PR #2907, it does not do so with those constructors.  So I can remove them.

sungam3r: I would not say at all that ExpressionFieldResolver and NameFieldResolver build some "default" expressions. They build concrete expressions like `context => context.Source.MyProperty` and user can not configure those expressions. No defaults.

sungam3r:
I finished review. Hard to track/verify all changes but let's go on. At least I understand a whole goal.

src/GraphQL/Resolvers/MemberResolver.cs
Shane32: Done
- Excess constructors removed in `MemberResolver` and `EventStreamMethodResolver`
- Instance expression is not nullable for other constructors
- Constructor parameters reordered more appropriately considering that instance expression is not optional
- Instance expression is not nullable in `AutoRegisteringHelper.BuildFieldResolver`
- Xml comments updated.

sungam3r (approved):

PR #2974, opened 2022-02-28
IObservable support for AutoRegisteringObjectGraphType

Shane32:
Depends on
- #2966 

The tests are a duplicate of `SubscriptionTests.cs` and `SubscriptionWithReflectionTests.cs` with extremely minor changes to (a) use the new schema, and (b) register `Chat` as a singleton within a service provider which is passed within `RequestServices`.

With this PR and the added test schema you can see how incredibly simple the auto-registering schema is.  Just compare the 42 lines of code in that file to the 176 lines of code in `SubscriptionSchema.cs` that does the same thing.


Shane32:
@sungam3r This is the next PR waiting to be reviewed.

src/GraphQL.Tests/Subscription/SubscriptionSchemaWithAutoGraphType.cs
```diff
@@ -0,0 +1,42 @@
...
+
+public class SubscriptionSchemaWithAutoGraphType : Schema
+{
+    public SubscriptionSchemaWithAutoGraphType()
+    {
+        Query = new AutoRegisteringObjectGraphType<QueryType>();
+        Mutation = new AutoRegisteringObjectGraphType<MutationType>();
+        Subscription = new AutoRegisteringObjectGraphType<SubscriptionType>();
+
+        RegisterTypeMapping(typeof(Message), typeof(AutoRegisteringObjectGraphType<Message>));
```
sungam3r, line 16: Alternative syntax:

`this.AutoRegister<Message>()`

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -121,15 +123,27 @@ protected void BuildFieldType(FieldType fieldType, MemberInfo memberInfo)
                         queryArgument ?? throw new InvalidOperationException("Invalid response from ConstructQueryArgument: queryArgument and expression cannot both be null"));
                     expressions.Add(expression);
                 }
-                var resolver = new MemberResolver(methodInfo, BuildMemberInstanceExpression(memberInfo), expressions);
+                var memberInstanceExpression = BuildMemberInstanceExpression(methodInfo);
+                if (IsObservable(methodInfo.ReturnType))
+                {
+                    var resolver = new EventStreamMethodResolver(methodInfo, memberInstanceExpression, expressions);
+                    fieldType.Resolver = resolver;
```
sungam3r, line 130: Both branches contain this line.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -93,13 +94,14 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
         /// </summary>
-        protected void BuildFieldType(FieldType fieldType, MemberInfo memberInfo)
+        protected void BuildFieldType(EventStreamFieldType fieldType, MemberInfo memberInfo)
         {
             if (memberInfo is PropertyInfo propertyInfo)
             {
                 var resolver = new MemberResolver(propertyInfo, BuildMemberInstanceExpression(memberInfo));
                 fieldType.Arguments = null;
                 fieldType.Resolver = resolver;
+                fieldType.AsyncSubscriber = null;
```
sungam3r, line 104: Is it necessary?

src/GraphQL/Types/TypeInformation.cs
```diff
@@ -206,7 +206,7 @@ private void ProcessTypeTree(List<(Type Type, NullabilityState Nullable)> typeTr
                     if (type.Type.IsGenericType)
                     {
                         var genericType = type.Type.GetGenericTypeDefinition();
-                        if (genericType == typeof(Task<>) || genericType == typeof(IDataLoaderResult<>))
+                        if (genericType == typeof(Task<>) || genericType == typeof(IDataLoaderResult<>) || genericType == typeof(IObservable<>))
```
sungam3r, line 209: Plus ValueTask?

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Yeah.... I'd have to write `IFieldResolver resolver` to move it outside the `if`.  Or I can consolidate the line with the line above it in each branch.  Seemed best to leave it this way.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
Shane32: Not necessarily, but the xml comment says specifically that it overwrites the values in `Arguments`, `Resolver` and `AsyncSubscriber`.  So that's what I coded it to do.

src/GraphQL/Types/TypeInformation.cs
Shane32: That change exists in PR #2907

src/GraphQL.Tests/Subscription/SubscriptionSchemaWithAutoGraphType.cs
Shane32: done

src/GraphQL/Types/TypeInformation.cs
Shane32: (Returning a `ValueTask` is not supported by field resolvers at this point, so adding a condition for it here would be pointless.)

sungam3r (approved):
OK, let's go on.

PR #2975, opened 2022-02-28
Allow ScopedAttribute to work with subscription fields

Shane32:
Depends on
- #2974 

Note: although the field resolver is 'scoped', it is scoped only for the duration that it takes the resolver to return the `IObservable<T>`.  It does not maintain the scope until the client disconnects from the subscription.  Rather, the observable is expected to maintain its own lifetime and watch for a call to `IDisposable.Dispose` to detect a disconnection.  As such, this addition may have limited usefulness (e.g. database calls prior to returning the observable), but I think this should be the proper/expected behavior if someone were to tag a subscription method with the `[Scoped]` attribute.

I'm open to suggestions here, of course, and another option would be to create a separate attribute for a more lengthy service scope for subscription fields.

### DIFF
--- a/src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
+++ b/src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
@@ -1,3 +1,4 @@
+using GraphQL.Subscription;
 using GraphQL.Types;
 using Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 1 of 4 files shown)


Shane32:
@sungam3r This can be reviewed at your convenience.  It will not hold up any other PRs, so just review when you have time.  Thanks!

sungam3r (approved):

PR #2988, opened 2022-03-02
Allow method arguments with NameFieldResolver

Shane32:
Waiting on
- #2907

Uses same code as schema builder, `FieldDelegate`, and `AutoRegisteringObjectGraphType`, supporting `IResolveFieldContext` method parameters, method parameters with attributes like `[FromService]`, or method parameters from field arguments

Potential breaking change: an ambiguous match on the method name could occur where it did not before, if an overload with arguments existed alongside a method with no arguments.  Look for method with no parameters if this occurs for backwards compatibility?  Or document this scenario in the migration document?


Shane32:
@sungam3r2022 Ready for review

sungam3r2022:
Interesting behavior for such "primitive" resolver at least as it was designed before. Now it is not so primitive as before and I think of a better class name.

src/GraphQL/Resolvers/NameFieldResolver.cs
```diff
@@ -58,40 +59,31 @@ public class NameFieldResolver : IFieldResolver
...
             var param = Expression.Parameter(typeof(IResolveFieldContext), "context");
             var source = Expression.MakeMemberAccess(param, _sourcePropertyInfo);
             var cast = Expression.Convert(source, target);
-            var body = CreateAccessorExpression(cast, name, target);
-            // supports return types of T, Task<T> and ValueTask<T>
-            return MemberResolver.BuildFieldResolverInternal(param, body);
+            var instanceLambda = Expression.Lambda(cast, param);
 
-            static Expression CreateAccessorExpression(Expression member, string name, Type target)
+            MemberInfo? member = target.GetProperty(name, BindingFlags.IgnoreCase | BindingFlags.Instance | BindingFlags.Public | BindingFlags.FlattenHierarchy);
```
sungam3r2022, line 70: I think it may throw too in case of hiding property (with same name).

src/GraphQL.Tests/NameFieldResolverTests.cs
```diff
@@ -14,27 +16,38 @@ public class NameFieldResolverTests
...
+                    Arguments = new Dictionary<string, ArgumentValue>()
+                    {
+                        { "prefix", new ArgumentValue("test ", ArgumentSource.Literal) }
+                    },
+                    RequestServices = services.BuildServiceProvider(),
+                });
 
             if (throws)
-                Should.Throw<InvalidOperationException>(() => result());
+                await Should.ThrowAsync<InvalidOperationException>(async () => await result());
```
sungam3r2022, line 48: Test message text too.

sungam3r2022:
> Potential breaking change: an ambiguous match on the method name could occur where it did not before, if an overload with arguments existed alongside a method with no arguments. Look for method with no parameters if this occurs for backwards compatibility? Or document this scenario in the migration document?

NameFieldResolver initial desing was rather primitive - return property value by its name. NameFieldResolver works like a default resolver when one was not specified. Then we added support for methods. Now for custom attributes. What to do with custom parameters and ambiguous matches? I think nothing. We have already done too much for a resolver, which works by default. I propose to throw on ambiguous match and not complicate code further.

src/GraphQL.Tests/NameFieldResolverTests.cs
Shane32: This is a `[Theory]` and the messages vary depending on the failure.  In general, I really have no inclination to verify exception messages, and I do not believe that verifying the message adds value to the testing.  Changing an exception message is not generally considered a breaking change.  Besides, some exception messages change with the framework in use or language.  You may rework this method and add validation of exception messages if you choose to do so.

src/GraphQL/Resolvers/NameFieldResolver.cs
Shane32: Added test; no errors for this situation.

Shane32:
I like the fact that with this PR we have now unified the support of methods between the `AutoRegisteringObjectGraphType` class, the schema builder, the `FieldDelegate` method, and now the `NameFieldResolver`.  All are precompiled (to the extent possible); all support attributes; all support query arguments as method parameters; and any changes in the future to the logic will apply uniformly across those classes/methods.

Previously, the `NameFieldResolver` supported methods but no arguments, the `AutoRegisteringObjectGraphType` class did not support methods at all, and the `FieldDelegate` and schema builder used reflection at run-time to execute.

src/GraphQL.Tests/NameFieldResolverTests.cs
sungam3r2022: Exception messages are useful to see what situation happened (for test reader). If messages vary I use `Contains`.

sungam3r2022 (approved):

PR #2993, opened 2022-03-04
Remove IResolveEventStreamContext, EventStreamFieldType, typed IEventStreamResolver

### LINKED ISSUE #2983: [Discussion] Remove typed IFieldResolver and IEventStreamResolver ?
Should we remove `IFieldResolver<TReturnType>` ?  It really serves no purpose and just complicates the design of the field resolvers, especially with `ValueTask<T>` not being `ValueTask<out T>`.

### LINKED ISSUE #2984: [Discussion] Remove IResolveEventStreamContext ?
`IResolveEventStreamContext` is identical to `IResolveFieldContext`; do we really need a separate interface? Eliminating the interface would also allow us to remove `ResolveEventStreamContext`

### LINKED ISSUE #2985: [Discussion] Remove EventStreamFieldType ?
Why not just add `Subscriber` to `FieldType` and remove the `EventStreamFieldType` class for simplification?  (Note, after #2907 the `AsyncSubscriber` property will already have been removed.)

Shane32:
- Remove `IResolveEventStreamContext`
- Remove `EventStreamFieldType`
- Remove `IEventStreamResolver<T>`
- Add `Func<....., ValueTask<T>>` constructors to `FuncFieldResolver` and `EventStreamResolver`

This PR does not add overloaded methods for `Field` with resolver delegates that return a `ValueTask`, or similar changes.  That would be a separate PR.

Closes:
- #2983 
- #2984 
- #2985


Shane32:
@sungam3r2022 Ready for review.

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
```diff
@@ -50,45 +47,19 @@ public override async Task<ExecutionResult> ExecuteAsync(ExecutionContext contex
...
-                    CancellationToken = context.CancellationToken,
-                    Metrics = context.Metrics,
-                    Errors = context.Errors,
-                    Path = node.Path,
-                    RequestServices = context.RequestServices,
-                };
-
-                var eventStreamField = node.FieldDefinition as EventStreamFieldType;
-
+                var resolveContext = new ResolveFieldContext(new ReadonlyResolveFieldContext(node, context));
```
sungam3r2022, line 56: Wrapped because of ` resolveContext.RequestServices = null;` ?

Shane32: yes.

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
```diff
@@ -1,44 +1,54 @@
...
 
-        public AsyncEventStreamResolver(
-            Func<IResolveEventStreamContext, Task<IObservable<T?>>> subscriber)
+        public AsyncEventStreamResolver(Func<IResolveFieldContext, Task<IObservable<T?>>> subscriber)
         {
-            _subscriber = subscriber ?? throw new ArgumentNullException(nameof(subscriber));
+            if (subscriber == null)
+                throw new ArgumentNullException(nameof(subscriber));
+
+            if (typeof(T).IsValueType)
```
sungam3r2022, line 12: Add `where T : class` and remove runtime checks ?

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -75,5 +75,10 @@ internal void SetName(string name, bool validate)
         /// Gets or sets a field resolver for the field. Only applicable to fields of output graph types.
         /// </summary>
         public IFieldResolver? Resolver { get; set; }
+
+        /// <summary>
+        /// Gets or sets a subscription resolver for the field. Only applicable to subscription fields.
```
sungam3r2022, line 80: It's worth to add additional check into schema validation.

src/GraphQL/Resolvers/EventStreamResolver.cs
```diff
@@ -1,38 +1,40 @@
...
+        public EventStreamResolver(Func<IResolveFieldContext, IObservable<T?>> subscriber)
         {
-            _subscriber = subscriber ?? throw new ArgumentNullException(nameof(subscriber));
-        }
+            if (subscriber == null)
+                throw new ArgumentNullException(nameof(subscriber));
 
-        public ValueTask<IObservable<T?>> SubscribeAsync(IResolveEventStreamContext context)
-            => new ValueTask<IObservable<T?>>(_subscriber(context));
+            if (typeof(T).IsValueType)
```
sungam3r2022, line 12: where T : class

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
```diff
@@ -1,44 +1,54 @@
-using GraphQL.Subscription;
-
 namespace GraphQL.Resolvers
 {
-    public class AsyncEventStreamResolver<T> : IEventStreamResolver<T>
+    public class AsyncEventStreamResolver<T> : IEventStreamResolver
```
sungam3r2022, line 3: I would remove `AsyncEventStreamResolver` classes at all and add ctors in `EventStreamResolver` classes. Those are effectively the same concept - resolver returning IObservable.

src/GraphQL/Resolvers/EventStreamResolver.cs
Shane32: I had put those in, which works great here, but it breaks compatibility with the field builders since there is no `where T : class` on the field builders.  So I did this instead.

Shane32: Perhaps you'd like to download the branch and see if you have any other ideas ?

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
sungam3r2022: Or even just remove them because `Task<T>`  can be converter by `ValueTask<T>` by async/await trick.

src/GraphQL/Resolvers/EventStreamResolver.cs
Shane32: This is probably obvious, but it is necessary because `IObservable<int>` cannot be cast to `IObservable<object>`

src/GraphQL/Types/Fields/FieldType.cs
Shane32: I agree

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
Shane32: I'd like to work on that and `AsyncFieldResolver` in a separate PR.  There are some complications which we will need to consider.

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
Shane32: See conversation within `EventStreamResolver` below.

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
Shane32: > Or even just remove them because `Task<T>` can be converter by `ValueTask<T>` by async/await trick.

Yes that is one option.

src/GraphQL/Types/Fields/FieldType.cs
Shane32: - #2994

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
sungam3r2022: I would remove indeed. Compiler is smart enough to make things under the hood that we should not write manually complicated the code.

src/GraphQL/Resolvers/EventStreamResolver.cs
```diff
@@ -1,38 +1,40 @@
-using GraphQL.Subscription;
-
 namespace GraphQL.Resolvers
 {
-    public class EventStreamResolver<T> : IEventStreamResolver<T>
+    public class EventStreamResolver<T> : IEventStreamResolver
```
sungam3r2022, line 3: T -> TReturnType to simplify compare with `EventStreamResolver<TSourceType, TReturnType>`

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
Shane32: I'll do it in a subsequent PR.  I would like to make changes in smaller PRs when possible, and I do not believe anything in this PR requires this change.

src/GraphQL/Resolvers/EventStreamResolver.cs
Shane32: Of course I could use reflection to call the constructor and move the check into the field builder.

sungam3r2022: Let's keep as is for now.

sungam3r2022 (approved):
T -> TReturnType

src/GraphQL/Resolvers/EventStreamResolver.cs
Shane32: - #2995

src/GraphQL/Resolvers/AsyncEventStreamResolver.cs
Shane32: - #2996

PR #2999, opened 2022-03-05
Add IGraphQLBuilder method(s) to build an auto-registering schema

Shane32:
This PR is designed to allow code like this, eliminating all code-first set-up:

```cs
services.AddGraphQL(b => b
    .AddAutoSchema<Query>(s => s.WithMutation<Mutation>())
    .AddSystemTextJson());

public class Query
{
    public Droid Hero() => new Droid(1, "John Doe");
}

public record Droid(int Id, string Name);

public class Mutation
{
    public Droid Add(string name, [FromServices] MyRepo repo) => repo.AddDroid(name);
}
```

Todo:
- [x] Add xml comments
- [x] Add documentation & migration documentation
- [x] Reorganize code properly

@sungam3r2022 Can you do a preliminary review here of the API changes / design?


Shane32:
@sungam3r2022 If you have no comments as to the API changes here then I will finish up this PR.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -103,11 +105,22 @@ protected SchemaTypes()
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance for the specified schema, and with the specified type resolver.
+        /// </summary>
+        /// <param name="schema">A schema for which this instance is created.</param>
+        /// <param name="serviceProvider">A service provider used to resolve graph types.</param>
+        /// <param name="graphTypeMappings">A service used to map CLR types to graph types.</param>
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+        public SchemaTypes(ISchema schema, IServiceProvider serviceProvider, IEnumerable<IGraphTypeMapping>? graphTypeMappings)
```
sungam3r2022, line 120: I don't see usages of this ctor. `Schema` creates `SchemaTypes` explicitly.

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -219,6 +219,42 @@ public static IGraphQLBuilder AddSchema<TSchema>(this IGraphQLBuilder builder, F
...
+        {
+            builder.AddSchema<Schema>(ServiceLifetime.Singleton);
+            builder.Services.TryRegister<IGraphTypeMapping, AutoRegisteringGraphTypeMapping>(ServiceLifetime.Singleton);
+            builder.ConfigureSchema((schema, provider) => schema.Query = provider.GetRequiredService<AutoRegisteringObjectGraphType<TQueryClrType>>());
+            return new GraphQLBuilderOrConfigureAutoSchema(builder);
+        }
+
+        public static IGraphQLBuilderOrConfigureAutoSchema WithMutation<TMutationClrType>(this IGraphQLBuilderOrConfigureAutoSchema builder)
+        {
+            builder.ConfigureSchema((schema, provider) => schema.Query = provider.GetRequiredService<AutoRegisteringObjectGraphType<TMutationClrType>>());
```
sungam3r2022, line 233:
```suggestion
            builder.ConfigureSchema((schema, provider) => schema.Mutation = provider.GetRequiredService<AutoRegisteringObjectGraphType<TMutationClrType>>());
```
oops, for subscription as well

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -681,6 +696,17 @@ private object GetGraphType(Type clrType, bool input, List<(Type clr, Type graph
                 }
             }
 
+            // then an available IGraphTypeMapping instance
+            if (_graphTypeMappings != null)
```
sungam3r2022, line 700: I have a feeling that we may merge new API with `TypeMappings`. I mean delete `TypeMappings` and stop passing it down the `SchemaTypes` here to **unify** API. So `Schema.RegisterTypeMapping` registers another `IGraphTypeMapping` and then code works only with `IGraphTypeMapping` instances here. Drawback - someone may want to know what types are registered via `public Schema.TypeMappings` property. On the other hand, how will he find out that some type is also registered now, but "implicitly" through `AutoRegisteringGraphTypeMapping`?

I don't like that adding this feature can cause additional questions, although it works itself. Need to document such behavior - that `Schema.TypeMappings` does not reflect autoregistration added by new extension methods.

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -523,6 +559,14 @@ public static IGraphQLBuilder AddClrTypeMappings(this IGraphQLBuilder builder, A
         }
         #endregion
 
+        #region - AddAutoClrMappings -
+        public static IGraphQLBuilder AddAutoClrMappings(this IGraphQLBuilder builder, bool mapInputTypes = true, bool mapOutputTypes = true)
+        {
+            builder.Services.Register<IGraphTypeMapping>(_ => new AutoRegisteringGraphTypeMapping(mapInputTypes, mapOutputTypes), ServiceLifetime.Singleton);
```
sungam3r2022, line 565: `AddAutoSchema` uses `TryRegister` and adds `AutoRegisteringGraphTypeMapping` with `true, true` params. Depending on the registration order of `AddAutoClrMappings/ AddAutoSchema` code will work in different ways.

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -219,6 +219,42 @@ public static IGraphQLBuilder AddSchema<TSchema>(this IGraphQLBuilder builder, F
...
+        #region - AddAutoSchema / WithMutation / WithSubscription -
+        public static IGraphQLBuilderOrConfigureAutoSchema AddAutoSchema<TQueryClrType>(this IGraphQLBuilder builder)
+        {
+            builder.AddSchema<Schema>(ServiceLifetime.Singleton);
+            builder.Services.TryRegister<IGraphTypeMapping, AutoRegisteringGraphTypeMapping>(ServiceLifetime.Singleton);
+            builder.ConfigureSchema((schema, provider) => schema.Query = provider.GetRequiredService<AutoRegisteringObjectGraphType<TQueryClrType>>());
+            return new GraphQLBuilderOrConfigureAutoSchema(builder);
+        }
+
+        public static IGraphQLBuilderOrConfigureAutoSchema WithMutation<TMutationClrType>(this IGraphQLBuilderOrConfigureAutoSchema builder)
```
sungam3r2022, line 231: > services.AddGraphQL()
    .AddAutoSchema<Query>()
    .WithMutation<Mutation>()
    .AddSystemTextJson();

Typo?

```
services.AddGraphQL(b =>
   b.AddAutoSchema<Query>()
    .WithMutation<Mutation>()
    .AddSystemTextJson())
```

Also consider to change API as well as we did for `AddGraphQL`:

```
services.AddGraphQL(b =>
   b.AddAutoSchema<Query>(b2 => b2.WithMutation<Mutation>())
    .AddSystemTextJson())
```

The same reasons here as before - more focused API.

sungam3r2022:
Reviewed.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: This is why I did not remove the `TypeMappings` as you suggested above -- because then this would be an issue.  `AddAutoSchema` needs a way to register a catch-all, but cannot override any manually registered mappings, either by `AddAutoClrMappings` or elsewhere.

How about this - we design the `IGraphTypeMapping` delegate to be somewhat like middleware, where all of the delegates can decide whether or not to change the mapping?  So `AddAutoClrMappings` will only change the mapping if it's currently `null`, whereas mappings added elsewhere would always set the mapping.  Then it should generally not matter which order the mappings were registered in.

```cs
/// <summary>
/// Provides a mapping from CLR types to graph types.
/// </summary>
public interface IGraphTypeMapping
{
    /// <summary>
    /// Returns a graph type for a given CLR type, or <see langword="null"/> if no mapping is available.
    /// Should return <paramref name="preferredGraphType"/> if this instance does not wish to change the mapping.
    /// </summary>
    /// <param name="clrType">The CLR type to be mapped.</param>
    /// <param name="isInputType">Indicates whether the type is an input type.</param>
    /// <param name="preferredGraphType">The graph type that is suggested for this CLR type.</param>
    Type? GetGraphTypeFromClrType(Type clrType, bool isInputType, Type? preferredGraphType);
}
```

Then code would be something like this:

```cs
Type? mappingType = null;
foreach (var mapping in mappings)
{
    mappingType = mapping.GetGraphTypeFromClrType(clrType, isInputType, mappingType);
}

if (mappingType == null && clrType.IsEnum)
{
    mappingType = typeof(EnumerableGraphType<>).MakeGenericType(clrType);
}

//ditto for scalar types

return mappingType;
``

The other thing to consider is that I believe we should retain the ability to add mappings from within the `Schema` as is possible now.  I would suggest this be done by creating a list of `IGraphTypeMapping` implementations within `Schema` and then appending them to those retrieved from DI when passed to `SchemaTypes`.  Is this what you were thinking also?

File: src/GraphQL/GraphQLBuilderExtensions.cs (hunk 5 above)
Comment (Shane32) [reply src/GraphQL/GraphQLBuilderExtensions.cs:231]: Yes type-o.

We can, but I was thinking this was simpler and felt more like just registering additional types.  We could also make `b2` be the `Schema`, but I wanted this implementation to be extremely simple, so IntelliSense could easily help people register the mutation and/or subscription CLR types, as that is all that is likely needed for this use case.  Users can always call `ConfigureSchema` for additional configuration on the `Schema` instance.

Finally, we could have `AddAutoSchema` include the mutation or subscription type, but which one?  How to tell which one was intended?  Seemed better to only include the type that is required (TQuery) and use additional methods to configure the mutation and subscription types.

Considering the above, do you still think I should change it like this?

```cs
services.AddGraphQL(b => b
    .AddAutoSchema<Query>(b2 => b2.WithMutation<Mutation>())
    .AddSystemTextJson())
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -19,7 +20,7 @@ namespace GraphQL.Types
     /// </summary>
     public class SchemaTypes : IEnumerable<IGraphType>
     {
-        internal static readonly Dictionary<Type, Type> BuiltInScalarMappings = new Dictionary<Type, Type>
+        public static ReadOnlyDictionary<Type, Type> BuiltInScalarMappings { get; } = new(new Dictionary<Type, Type>
```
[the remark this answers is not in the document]
Shane32: Yes, I'm pretty sure I will need a custom implementation for some of my code.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: The standard constructor calls this one and populates the new argument from DI.

Shane32:
@sungam3r2022 I made some changes to the PR to reflect the discussion above.  Tests do not pass because the builder tests for AddClrTypeMappings need to be updated; but other than that it passes build checks.  (Of course I have not finished the PR, made tests for any new functionality, etc -- I still want to figure out the API with you first.)

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: See updated code to reflect this idea.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: All of your points are valid points.  Personally, I would conclude that we should leave `Schema.TypeMappings` and related `RegisterTypeMapping` methods alone, or it will be difficult to write code that builds a schema from code like this:

```cs
var queryType = new ObjectGraphType;
//configure query type

var schema = new Schema();
schema.Query = queryType;
schema.RegisterTypeMapping<MyObject, MyObjectGraphType>();
```

Of course the method could add the mapping to a list of `IGraphTypeMapping` implementations, but like you said, then you can enumerate the added mappings.  Just seems like a small change to leave it as-is, and better for the project overall.

Shane32: On the other hand there is no way to manually add a `IGraphTypeMapping` implementation to a schema without some sort of change here.  So maybe `TypeMappings` returns `IEnumerable<IGraphTypeMapping>` now?

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: done

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r2022: > Then it should generally not matter which order the mappings were registered in.

It will matter if more then one `IGraphTypeMapping` implementations want to change mappings.

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r2022: If you expect that `WithMutation/WithSubscription` methods should always be used with `AddAutoSchema` and "in the context" of `AddAutoSchema` then yes, I would switch to inner configuration delegate. I prefer to group extension methods into internal lambdas, and not to have a flat list. When evolving the API, it helps.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r2022: 🤦‍♂️

sungam3r2022:
OK. I'll finish review tomorrow, was a busy today.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r2022: The "problem" is a different nature of mappings. `IGraphTypeMapping` is a _factory_ that builds something. You can not say what before you call it. On the other hand `TypeMappings` now is a list of explicitly defined types.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: Ok.

Shane32: I'll wait until you finish your preliminary review before I make further changes.

sungam3r: OK. Today I will not be able to think thoughtfully. I will answer all incoming emails (~10 left) and go to sleep.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -127,8 +142,12 @@ protected void Initialize(ISchema schema, IServiceProvider serviceProvider)
             _initialized = true;
 
             var types = GetSchemaTypes(schema, serviceProvider);
-            var typeMappingsEnumerable = schema.TypeMappings ?? throw new ArgumentNullException(nameof(schema) + "." + nameof(ISchema.TypeMappings));
-            var typeMappings = typeMappingsEnumerable is List<(Type, Type)> typeMappingsList ? typeMappingsList : typeMappingsEnumerable.ToList();
+            if (schema.TypeMappings != null)
+            {
+                // this code could be moved into Schema
+                var additionalMappings = schema.TypeMappings.Select(x => new ManualGraphTypeMapping(x.clrType, x.graphType));
+                graphTypeMappings = graphTypeMappings != null ? graphTypeMappings.Concat(additionalMappings).ToList() : additionalMappings.ToList();
```
sungam3r, line 149:
```suggestion
                var manualMappings = schema.TypeMappings.Select(x => new ManualGraphTypeMapping(x.clrType, x.graphType));
                graphTypeMappings = graphTypeMappings != null ? graphTypeMappings.Concat(manualMappings).ToList() : additionalMappings.ToList();
```

sungam3r: Also I think manualMappings should go first for `Concat`.

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: I looked through code one more time. OK, let's try to go this way. Now, when calculating the graph type from clr type  you have to iterate the entire collection of mappings. Well... OK, it is one time operation when initializing a schema.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: I reversed it so the manual mappings would always take precedence over DI-configured mappings.  (Last-wins with this new design versus first-wins before.)

src/GraphQL/Types/TypeCollectionContext.cs
```diff
@@ -30,7 +30,7 @@ internal TypeCollectionContext(Func<Type, IGraphType> resolver, Action<string, I
         /// </summary>
         internal Action<string, IGraphType, TypeCollectionContext> AddType { get; private set; }
 
-        internal List<(Type, Type)> TypeMappings { get; private set; }
+        internal IEnumerable<IGraphTypeMapping>? TypeMappings { get; private set; }
```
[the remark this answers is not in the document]
Shane32: Sure

src/GraphQL/Types/Collections/ManualGraphTypeMappingResolver.cs
```diff
@@ -0,0 +1,20 @@
+namespace GraphQL.Types.Collections;
+
+internal class ManualGraphTypeMappingResolver : IGraphTypeMappingResolver
```
Shane32, line 3: internal or public?

Shane32: Schema still has methods to add CLR to graph type mappings without making this public

Shane32:
@sungam3r Ready for review here.  I can up the code coverage on GraphQLBuilderExtensions if you like; I'm out of time today.  Besides, if you request any changes, I'd rather add those tests afterwards.

src/GraphQL/Types/Collections/ManualGraphTypeMappingResolver.cs
sungam3r: internal

src/GraphQL/Types/Collections/IGraphTypeMappingResolver.cs
```diff
@@ -0,0 +1,16 @@
+namespace GraphQL.Types;
+
+/// <summary>
+/// Provides a mapping from CLR types to graph types.
+/// </summary>
+public interface IGraphTypeMappingResolver
```
sungam3r, line 6: "Resolver" in name may sound a bit confusing.

Shane32: I'm open to suggestions.  I added Resolver because it doesn't represent a mapping (with a concrete set of types) so much as a delegate / factory / resolver.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -19,7 +21,10 @@ namespace GraphQL.Types
     /// </summary>
     public class SchemaTypes : IEnumerable<IGraphType>
     {
-        internal static readonly Dictionary<Type, Type> BuiltInScalarMappings = new Dictionary<Type, Type>
+        /// <summary>
+        /// Returns a dictionary of default CLR type to graph type mappings.
```
sungam3r, line 25:
```suggestion
        /// Returns a dictionary of default CLR type to graph type mappings for a set of built-in (primitive) types.
```

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -103,11 +108,22 @@ protected SchemaTypes()
...
+            : this(schema, serviceProvider, (IEnumerable<IGraphTypeMappingResolver>?)serviceProvider.GetService(typeof(IEnumerable<IGraphTypeMappingResolver>)))
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance for the specified schema, and with the specified type resolver.
+        /// </summary>
+        /// <param name="schema">A schema for which this instance is created.</param>
+        /// <param name="serviceProvider">A service provider used to resolve graph types.</param>
+        /// <param name="graphTypeMappings">A service used to map CLR types to graph types.</param>
```
sungam3r, line 121: A service ?

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -103,11 +108,22 @@ protected SchemaTypes()
...
         /// <param name="schema">A schema for which this instance is created.</param>
         /// <param name="serviceProvider">A service provider used to resolve graph types.</param>
-#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         public SchemaTypes(ISchema schema, IServiceProvider serviceProvider)
+            : this(schema, serviceProvider, (IEnumerable<IGraphTypeMappingResolver>?)serviceProvider.GetService(typeof(IEnumerable<IGraphTypeMappingResolver>)))
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance for the specified schema, and with the specified type resolver.
```
sungam3r, line 117: Copy-pasted from prev ctor.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: Where?

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Oops; before I updated it to use an enumerable.

Shane32:
```suggestion
        /// <param name="graphTypeMappings">A list of <see cref="IGraphTypeMappingResolver"/> instances used to map CLR types to graph types.</param>
```

src/GraphQL/Types/Collections/IGraphTypeMappingResolver.cs
sungam3r: Resolver -> Provider

The only reason from me - do not mix this API with field resolvers.

docs2/site/docs/migrations/migration5.md
```diff
@@ -500,6 +500,50 @@ When adding a field by name only, such as `Field<StringGraphType>("Name");`, and
...
+    o.RequestServices = provider;
+    o.Schema = provider.GetRequiredService<ISchema>();
+    o.Query = "{hero}";
+});
+var resultString = provider.GetRequiredService<IGraphQLTextSerializer>().Serialize(result);
+// resultString returns the following JSON: {"data":{"hero":"Luke Skywalker"}}
+
+
+// sample schema
+private class Query
```
sungam3r, line 529: I suggest change private to public for this example.

src/GraphQL/Types/Collections/IGraphTypeMappingResolver.cs
Shane32: I agree.

docs2/site/docs/migrations/migration5.md
Shane32: Good idea.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Merged suggestion above.  Resolve if ok.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Merged suggestion above.  Resolve if okay.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: > I reversed it

Where is the priority of manual mappings over other ones?

Shane32: See `GetGraphTypeFromClrType`:

```cs
            Type? mappedType = null;

            // check custom mappings first
            if (typeMappings != null)
            {
                foreach (var mapping in typeMappings)
                {
                    mappedType = mapping.GetGraphTypeFromClrType(clrType, isInputType, mappedType);
                }
            }

            if (mappedType != null)
                return mappedType;
```

In this code, for any `ManualGraphTypeMappingProvider`, the LAST registration in `typeMappings` above is going to take effect.  So `SchemaTypes.Initialize` puts manual registrations in the schema (`schema.TypeMappings`) LAST in the enumerable by using concat:

```cs
            if (schema.TypeMappings != null)
            {
                // this code could be moved into Schema
                var additionalMappings = schema.TypeMappings.Select(x => new ManualGraphTypeMappingProvider(x.clrType, x.graphType));
                graphTypeMappings = graphTypeMappings != null ? graphTypeMappings.Concat(additionalMappings).ToList() : additionalMappings.ToList();
            }
```

Does this not make sense?

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -219,6 +220,100 @@ public static IGraphQLBuilder AddSchema<TSchema>(this IGraphQLBuilder builder, F
...
+        #endregion
+
+        #region - AddAutoSchema / WithMutation / WithSubscription -
+        /// <summary>
+        /// Registers an instance of the <see cref="Schema"/> class within the dependency injection framework as a singleton.
+        /// <see cref="ISchema"/> is also registered if it is not already registered within the dependency injection framework.
+        /// <see cref="Schema.Query"/> is set to an instance of <see cref="AutoRegisteringObjectGraphType{TSourceType}"/> with
+        /// <typeparamref name="TQueryClrType"/> as TSourceType.
+        /// <br/><br/>
+        /// Additionally, this method calls <see cref="AddAutoClrMappings(IGraphQLBuilder, bool, bool)">AddAutoClrMappings</see>
```
sungam3r, line 277: Technically it does not call AddAutoClrMappings.

docs2/site/docs/migrations/migration5.md
Shane32: done

src/GraphQL/Types/Collections/IGraphTypeMappingResolver.cs
Shane32: done.  I hope I didn't miss anything.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: True, but for the purposes of the documentation, it does.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: OK. I see, it does. I'm fine with excessive call to `MakeGenericType`.

docs2/site/docs/migrations/migration5.md
```diff
@@ -961,3 +1005,11 @@ with the specified name, and if none is found, looks for a method with the match
 now supports methods with arguments as well as methods without arguments, an `AmbigiousMatchException`
 can occur if the name refers to a public method with multiple overloads. Either specify a field
 resolver explicitly, or reduce the number of public methods with the same name to one.
+
+### 38. `SchemaTypes` updated to support DI-injected mapping resolvers
```
sungam3r, line 1009:
```suggestion
### 38. `SchemaTypes` updated to support DI-injected mapping providers
```

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: Keep as is ?

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: Keep as is?

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Just FYI, for any ManualGraphTypeMappingProvider registrations, there are no calls to MakeGenericType at all.  And for `AutoRegisteringGraphTypeMappingProvider`, if a prior registration has set the type, it also will not call MakeGenericType.  And if `AutoRegisteringGraphTypeMappingProvider` does return a type, it will be returned by `GetGraphTypeFromClrType` prior to any other code.  So really I cannot see how (with the included implementations) there will be multiple calls to `MakeGenericType`.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: I would keep as it is now, similar to the above sample, not as is was when the first comments here were written.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: I would keep as it is now, with the changes we added pursuant to this and the below conversation, not as is was when the first comments here were written.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: not multiple but excessive one

```
 var additionalMappings = schema.TypeMappings.Select(x => new ManualGraphTypeMappingProvider(x.clrType, x.graphType));
graphTypeMappings = graphTypeMappings != null ? graphTypeMappings.Concat(additionalMappings).ToList() : additionalMappings.ToList();
```

`graphTypeMappings` here may contain `AutoRegisteringGraphTypeMappingProvider` so it goes before manual mapping. First `AutoRegisteringGraphTypeMappingProvider` always creates some type via `MakeGenericType`, then this type flows into `ManualGraphTypeMappingProvider` that _may_ discard it, not always of course. I'm not against such behaviour. OK to keep as is.

Shane32: You’re right.

sungam3r (approved):

PR #3000, opened 2022-03-07
Reduce build warnings

Shane32:
- Added XML comments
- Added nullability information
- Specify build warnings-as-errors if various nullability checks or XML comments are missing
- Fixed directory.build.props and directory.build.targets and csproj files properly so that build conditions work as intended
  - Directory.Build.props executes prior to csproj, and should not have any conditions, and should act as 'defaults' only
  - Imported files like Tests.props executes within csproj at point of import
  - Directory.Build.targets executes after csproj, and that is where conditional properties belong

Note:
- Did not add XML comments for GraphQL federation stuff, obsolete classes, or the schema printer which is expected to change
- Some nullability attributes were changed, either because nullability was previously not enabled for the project (e.g. data loaders), or because it seemed that a different nullability attribute was more accurate
- It seems that the VS IDE doesn't always interpret nullability the same as the compiler, so in a couple instances I found that I was unable to satisfy both the IDE and the compiler.  Mostly I just decided to just skip that particular compiler warning.
- Probably could use more work to figure out the rest of the warnings; there probably is a solution.
- Hopefully these changes will reduce the amount of times the formatter fails a build check because of a warning

I may do another pass later and try to take care of more of the build warnings.  I think there are 45 or so left now.  It would also be nice if the test projects also had NRT enabled and were subject to the same rules.


src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -12,7 +12,9 @@ namespace GraphQL.Utilities
     /// </summary>
     public class SchemaBuilder
     {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
         protected readonly Dictionary<string, IGraphType> _types = new Dictionary<string, IGraphType>();
```
Shane32, line 16: This should be fixed!!!  Was it intended to be `internal protected` ?

src/GraphQL/Utilities/SchemaPrinter.cs
```diff
@@ -1,3 +1,5 @@
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
Shane32, line 1: Did not spend any time on the SchemaPrinter, as it is marked with a todo "this should be completely rewritten", and I know we discussed having it use the AST printer from the parser project

Shane32:
@sungam3r2022 This is ready for review.

src/GraphQL/Utilities/SchemaPrinter.cs
sungam3r2022: Yes.

sungam3r2022:
> Directory.Build.props executes prior to csproj, and **should not have any conditions**, and should act as 'defaults' only

Why so? Any documentation about that? I see many other projects use conditions in Directory.Build.props.

src/Directory.Build.targets
```diff
@@ -6,4 +6,15 @@
     <Exclude>[GraphQL.StarWars]*,[GraphQL.Harness]*,[GraphQL.Dummy]*</Exclude>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(IsPackable)' == 'true'">
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+    <WarningsAsErrors>$(WarningsAsErrors);CS1591;CS1574;CS8765;CS8601;CS8620;CS8600;CS8625;CS8767;CS8603;CS8618</WarningsAsErrors>
```
sungam3r2022, line 11: I suggest to switch to `<TreatWarningsAsErrors>true</TreatWarningsAsErrors> + <WarningsNotAsErrors>X;Y;Z</WarningsNotAsErrors>` instead.

src/Directory.Build.targets
```diff
@@ -6,4 +6,15 @@
...
+    <Nullable>enable</Nullable>
```
sungam3r2022, line 12: Do you want to exclude tests/examples for now?

Shane32:
> > Directory.Build.props executes prior to csproj, and **should not have any conditions**, and should act as 'defaults' only
> 
> Why so? Any documentation about that? I see many other projects use conditions in Directory.Build.props.

Yes. I looked it up on MS website and it’s quite clear. I’ll post a link. But regardless, it only seemed to work this way. Maybe there was something I missed; not sure.

src/Directory.Build.targets
Shane32: Yes, I only have so much time, and I ran out of time to do more. Ideally I would like it to apply to tests also.

Shane32:
> > > Directory.Build.props executes prior to csproj, and **should not have any conditions**, and should act as 'defaults' only
> > 
> > 
> > Why so? Any documentation about that? I see many other projects use conditions in Directory.Build.props.
> 
> Yes. I looked it up on MS website and it’s quite clear. I’ll post a link. But regardless, it only seemed to work this way. Maybe there was something I missed; not sure.

https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022

See:

> - Directory.Build.props is imported very early in Microsoft.Common.props, and properties defined later are unavailable to it. So, avoid referring to properties that are not yet defined (and will evaluate to empty).
> - Properties that are set in Directory.Build.props can be overridden elsewhere in the project file or in imported files, so you should think of the settings in Directory.Build.props as specifying the defaults for your projects.
> - Directory.Build.targets is imported from Microsoft.Common.targets after importing .targets files from NuGet packages. So, it can override properties and targets defined in most of the build logic, or set properties for all your projects regardless of what the individual projects set.

src/Directory.Build.targets
Shane32: Done

Shane32:
Error CS8604 usually relates to `ToString()` being thought of as possibly nullable.  We can change references to `ToString()!` and then remove the CS8604 exemption if you like.

sungam3r2022:
What problem did you solve moving properties around? And what about `ContinuousIntegrationBuild` with condition?

src/GraphQL.Harness/GraphQL.Harness.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>net6;net5;netcoreapp3.1</TargetFrameworks>
     <NoWarn>$(NoWarn);1591</NoWarn>
+    <IsPackable>false</IsPackable>
```
sungam3r2022, line 6: IsPackable is false by default for all Web projects.

src/GraphQL.NewtonsoftJson/FixPrecisionConverter.cs
```diff
@@ -42,7 +52,7 @@ public override void WriteJson(JsonWriter jWriter, object value, JsonSerializer
             }
         }
 
-        private static bool IsWholeValue(object value, out string result)
+        private static bool IsWholeValue(object value, out string? result)
```
sungam3r2022, line 55:
```suggestion
        private static bool IsWholeValue(object? value, out string? result)
```

src/GraphQL.NewtonsoftJson/FixPrecisionConverter.cs
```diff
@@ -5,34 +5,44 @@ namespace GraphQL.NewtonsoftJson
...
             objType == typeof(decimal) && _decimal ||
             objType == typeof(float) && _float ||
             objType == typeof(double) && _double;
 
-        public override void WriteJson(JsonWriter jWriter, object value, JsonSerializer jSerializer)
+        /// <inheritdoc/>
+        public override void WriteJson(JsonWriter jWriter, object? value, JsonSerializer jSerializer)
         {
-            if (IsWholeValue(value, out string result))
+            if (IsWholeValue(value!, out string? result))
```
sungam3r2022, line 45:
```suggestion
            if (IsWholeValue(value, out string? result))
```

src/GraphQL.NewtonsoftJson/OperationMessageJsonConverter.cs
```diff
@@ -23,9 +23,9 @@ public class OperationMessageJsonConverter : JsonConverter
         public override bool CanWrite => true;
 
         /// <inheritdoc/>
-        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        public override void WriteJson(JsonWriter writer, object? value, JsonSerializer serializer)
```
sungam3r2022, line 26: If they marked `value` as nullable then they suppose `WriteJson` _may_ accept `null` so we have NRE on line 30.

src/GraphQL.SystemTextJson/InputsJsonConverter.cs
```diff
@@ -14,18 +14,18 @@ public class InputsJsonConverter : JsonConverter<Inputs>
...
         public override void Write(Utf8JsonWriter writer, Inputs value, JsonSerializerOptions options)
-            => JsonSerializer.Serialize<IReadOnlyDictionary<string, object>>(writer, value, options);
+            => JsonSerializer.Serialize<IReadOnlyDictionary<string, object?>>(writer, value, options);
 
         /// <inheritdoc/>
         public override Inputs Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
             => ReadDictionary(ref reader).ToInputs();
 
-        private static Dictionary<string, object> ReadDictionary(ref Utf8JsonReader reader)
+        private static Dictionary<string, object?> ReadDictionary(ref Utf8JsonReader reader)
```
sungam3r2022, line 23: NSJ returns null:
```
  internal static Dictionary<string, object?>? ReadDictionary(JsonReader reader) { ...
            if (reader.TokenType == JsonToken.Null)
                return null;
```

src/Directory.Build.props
```diff
@@ -16,14 +16,7 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
-    <PackageReadmeFile Condition="'$(IsPackable)' == 'true'">README.md</PackageReadmeFile>
-    <NoWarn>$(NoWarn);1591</NoWarn> <!--TODO: temporary solution-->
+    <IsPackable>true</IsPackable>
```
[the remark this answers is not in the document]
Shane32: > What problem did you solve moving properties around?

As one example, since IsPackable is true for all library projects by default, there was no way to set IsPackable to false for GraphQL.Dummy prior to the conditions within Directory.Build.props that checked IsPackable.

Once the conditions were moved to Directory.Build.targets, then those conditions were skipped for GraphQL.Dummy

There was probably similar issues with other projects.  After reading the documentation, I configured the settings in the correct places, and it works properly now.

src/GraphQL.Harness/GraphQL.Harness.csproj
Shane32: I can remove it here, and I can probably remove IsPackable = true in the Directory.Build.props also

Shane32: Well, it didn't work.  If I remove IsPackable = true in Directory.Build.props, then Newtonsoft doesn't think IsPackable is true.  If I remove IsPackable = false from GraphQL.Harness then the setting within Directory.Build.props of IsPackable = true takes effect.  So I am leaving it as-is here.

src/GraphQL.NewtonsoftJson/FixPrecisionConverter.cs
Shane32: This is incorrect, but okay.

src/GraphQL.NewtonsoftJson/OperationMessageJsonConverter.cs
Shane32: Could be... I can add a check in any case.

src/GraphQL.SystemTextJson/InputsJsonConverter.cs
Shane32: As far as I know, STJ only executes a JsonConverter when there is an actual object to be decoded, and never in the case of nulls.  Perhaps Newtonsoft is different.  If you suggest a change anyway, lmk.

src/GraphQL.NewtonsoftJson/FixPrecisionConverter.cs
Shane32: done

src/GraphQL.NewtonsoftJson/FixPrecisionConverter.cs
Shane32: done

src/GraphQL/Resolvers/IEventStreamResolver.cs
```diff
@@ -1,7 +1,14 @@
 namespace GraphQL.Resolvers
 {
+    /// <summary>
+    /// Returns an <see cref="IObservable{T}"/> for a field. The <see cref="IObservable{T}"/> provides
+    /// a sequence of event notifications (aka 'an event stream') to the subscription execution engine,
+    /// which will resolve child fields using the event data as the source. Then the resolved graph is
+    /// returned to the client, and the process repeats for additional event notifications.
```
sungam3r2022, line 7:
```suggestion
    /// returned to the client, and the process repeats for further event notifications. 
```

src/GraphQL.Harness/GraphQL.Harness.csproj
sungam3r2022: Something wrong here. I'll contibue review and then pull this branch to check.

src/GraphQL.SystemTextJson/InputsJsonConverter.cs
sungam3r2022: If so I'm fine to keep as is.

src/GraphQL.Harness/GraphQL.Harness.csproj
Shane32: It's weird.  When I comment out IsPackable from Directory.Build.props, it acts as though the ItemGroup DOES realize that IsPackable is true, so the readme file is included for example, but the PropertyGroup does NOT realize it, so Nullable does not get set to enabled.
image

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -327,6 +327,17 @@ public virtual FieldType AddField(FieldType fieldType)
             });
         }
 
+        /// <summary>
+        /// Adds a subscription field to this graph type.
```
sungam3r2022, line 331:
```suggestion
        /// Adds a subscription field with the specified properties to this graph type.
```

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -327,6 +327,17 @@ public virtual FieldType AddField(FieldType fieldType)
...
+        /// </summary>
+        /// <typeparam name="TGraphType">The .NET type of the graph type of this field</typeparam>
```
sungam3r2022, line 333:
```suggestion
        /// <typeparam name="TGraphType">The .NET type of the graph type of this field.</typeparam>
```

src/GraphQL.Harness/GraphQL.Harness.csproj
Shane32: Please feel free to try to diagnose this.  But if I set IsPackable to true in Directory.Build.props, and IsPackable to false elsewhere as needed, everything works fine.  (With the conditions within Directory.Build.targets)

Shane32: It's also possible the VS IDE is applying IsPackable to the library projects and the compiler is not.  In this case, setting the property manually forces both to operate the same.

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -352,6 +363,17 @@ public virtual FieldType AddField(FieldType fieldType)
             });
         }
 
+        /// <summary>
+        /// Adds a subscription field to this graph type.
```
sungam3r2022, line 367:
```suggestion
        /// Adds a subscription field with the specified properties to this graph type.
```

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -352,6 +363,17 @@ public virtual FieldType AddField(FieldType fieldType)
...
+        /// </summary>
+        /// <typeparam name="TGraphType">The .NET type of the graph type of this field</typeparam>
```
sungam3r2022, line 369:
```suggestion
        /// <typeparam name="TGraphType">The .NET type of the graph type of this field.</typeparam>
```

src/GraphQL/Utilities/SchemaBuilder.cs
sungam3r2022: I don't know.

Shane32:
> And what about `ContinuousIntegrationBuild` with condition?

The property is set on the command line (or via environment variable) before Directory.Build.props is evaluated.  So it is fine there.

src/GraphQL/VariableName.cs
```diff
@@ -1,14 +1,26 @@
 namespace GraphQL
 {
+    /// <summary>
+    /// Represents the full path to a property of a variable, typically for
```
sungam3r2022, line 4: >to a _property_ of a variable

Sounds strange to me.

Shane32: Yeah, anything you suggest is fine with me.  That was the best I could come up with at the time - lol

src/GraphQL.Harness/GraphQL.Harness.csproj
sungam3r2022: I pulled branch, removed IsPackable=true,  added <GeneratePackageOnBuild>true</GeneratePackageOnBuild> in NSJ and package was built. Then I added IsPackable=false before GeneratePackageOnBuild and rebuilt project. No nupkg. So all works fine as I can see.

sungam3r2022: I would not change the structure of configuration files moving parts around. We have it the same among few repos. I like that they are the same, because then it is easy to change them at the same time. If the problem is only in one conditional property `nullable` (which you eventually want to apply to all projects) then I suggest move `nullable` under another condition into Directory.Build.props:
```
<PropertyGroup Condition="$(MSBuildProjectName.EndsWith('.Tests'))">
    <Nullable>disable</Nullable>
</PropertyGroup>
```

sungam3r:
@Shane32 I came back!

src/GraphQL/VariableName.cs
sungam3r: > Represents the full path to a variable

?

Shane32: The class might represent "variableA.products[4].categoryName".  In this case the name is now "variableA.products" and the index is 4 and the ChildName is "categoryName".  Somehow the XML comment needs to explain this concept to the user, that the class as a whole represents, as I called it, "the full path to a property of a variable".  I'm open to suggestions.  But it certainly isn't just a variable, and isn't just a variable's property.  It's a whole path....

src/GraphQL.SystemReactive/SubscriptionExecutionStrategy.cs
```diff
@@ -5,13 +5,22 @@
...
         /// Gets a static instance of <see cref="SubscriptionExecutionStrategy"/> strategy.
         /// </summary>
         public static new SubscriptionExecutionStrategy Instance { get; } = new SubscriptionExecutionStrategy();
 
+        /// <summary>
+        /// Executes a GraphQL subscription request and returns the result. The result consists
+        /// of one or more streams of GraphQL responses, returned within <see cref="SubscriptionExecutionResult.Streams"/>.
+        /// No serializable <see cref="ExecutionResult"/> is directly returned.
+        /// </summary>
+        /// <inheritdoc/>
```
Shane32, line 23:
```suggestion
```

src/GraphQL.Harness/GraphQL.Harness.csproj
sungam3r: > "add true for NSJ"

Sorry, I forgot to escape xml tags; fixed comment.

src/Directory.Build.targets
```diff
@@ -6,4 +6,16 @@
...
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(IsPackable)' == 'true'">
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningsNotAsErrors>CS8604</WarningsNotAsErrors>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(IsPackable)' == 'true'">
```
sungam3r, line 16: From https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022 : 
> Include items in .props files (conditioned on a property). All properties are considered before any item, so user-project property customizations get picked up, and this gives the user's project the opportunity to Remove or Update any item brought in by the import.

sungam3r: What do you think? Move back into .props? 

I have read documentation you provided and I'm fine to accept these changes.

src/GraphQL.Harness/GraphQL.Harness.csproj
sungam3r: > t's weird. When I comment out IsPackable from Directory.Build.props, it acts as though the ItemGroup DOES realize that IsPackable is true, so the readme file is included for example, but the PropertyGroup does NOT realize it, so Nullable does not get set to enabled.

Because of that ? https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022:
> Include items in .props files (conditioned on a property). All properties are considered before any item, so user-project property customizations get picked up, and this gives the user's project the opportunity to Remove or Update any item brought in by the import.

sungam3r: Nevermind, I'm fine to merge as is and go on.

src/GraphQL/VariableName.cs
sungam3r: 😕 OK

sungam3r (approved):

src/Directory.Build.targets
Shane32: So that's why I was experiencing the weirdness -- it processed all the property groups first then the item groups.  I missed that when I was reading the doc.

Sure, I'll move the item groups back to `Directory.Build.props`, leaving the property conditions in `Directory.Build.targets`, and then merge if it all builds okay.

Shane32: Actually, I'll just do a separate PR so we can both review.

sungam3r: I find all these MSBuild rules rather non-obvious. I have a nuget metapackage with props/targets that is used in all of our projects at my work. I spent a bunch of time and strength to place all the properties and items I need so that they work correctly. And I'm still not sure if I did it well.

PR #3007, opened 2022-03-11
Add IsAbstract check and Introduce EnableAbstractGenericParametersForInputObjectGraphType switch

sungam3r:
I faced a problem with abstract TSourceType in one of our services.

### DIFF
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -170,6 +170,7 @@ namespace GraphQL
     public static class GlobalSwitches
     {
         public static System.Action<string, GraphQL.Utilities.NamedElement> NameValidation;
+        public static bool EnableAbstractGenericParametersForInputObjectGraphType { get; set; }
         public static bool EnableReadDefaultValueFromAttributes { get; set; }
         public static bool EnableReadDeprecationReasonFromAttributes { get; set; }
         public static bool EnableReadDescriptionFromAttributes { get; set; }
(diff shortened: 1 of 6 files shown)


Shane32:
Just so I understand, how it could be useful to support abstract types for InputObjectGraphType?

Shane32:
I would suggest we do not add the global switch nor the check in the constructor, but do add the check in ToObject. Offhand it seems so unlikely that someone try to create an instance of InputObjectGraphType with an abstract type as TSourceType that they probably know what they are doing, and if not, it will be obvious the first time they attempt to parse a value if they have not overridden ParseDictionary or registered within the ValueConverter.

I might add even more text to the exception message to make it more clear where the likely spot to fix the problem would be. For instance “please register a conversion within the ValueConverter or for input graph types override ParseDictionary” or similar. 

I really don’t care for global switches if we can at all avoid them and here it seems unnecessary. If you feel strongly then we can merge as is.

sungam3r:
The problem was not immediately.

> it seems so unlikely that someone try to create an instance of InputObjectGraphType with an abstract type as TSourceType

But someone tried :)

> it will be obvious the first time they attempt to parse a value if they have not overridden ParseDictionary or registered within the ValueConverter.

For my colleagues it was not obvious. I had to deal with.

New check in `ToObject` is my main goal here. New switch is just a continuation of the main idea. It's backward compatible. I changed error text.

Shane32:
> For my colleagues it was not obvious. I had to deal with.

Because we did not throw a proper exception with a message stating how to solve it, which you added in ToObject.

Shane32:
If `EnableAbstractGenericParametersForInputObjectGraphType` defaults to `true`, then I would not add it.  Who would specifically set it to false?  They will run into the problem long before they poke around in the global settings and decide that it should be false for their needs.  Either default it to `false` or remove it.

sungam3r:
True. I wanted to shift error to app start rather then request execution but with proper message new switch is useless. Removed.

Shane32 (approved):

PR #3016, opened 2022-03-17
Add typed IDocumentExecuter

Shane32:
I would like to add this to better support multiple schemas.  It adds a single interface to the public API.  An example of its use is provided in the tests - see `StringExecuter` sample class.

### DIFF
--- a/src/GraphQL.Tests/Execution/DocumentExecuterTests.cs
+++ b/src/GraphQL.Tests/Execution/DocumentExecuterTests.cs
@@ -61,15 +61,15 @@ namespace GraphQL.Tests.Execution
                 .AddSchema<Schema1>()
                 .AddSchema<Schema2>()
                 .AddSystemTextJson());
-            services.AddSingleton(typeof(MyExecuter<>));
+            services.AddSingleton(typeof(StringExecuter<>));
             var provider = services.BuildServiceProvider();
 
             // verify executing with Schema1 works with custom class
-            var executer1 = provider.GetRequiredService<MyExecuter<Schema1>>();
+            var executer1 = provider.GetRequiredService<StringExecuter<Schema1>>();
             var result1 = await executer1.ExecuteAsync("{hero}");
             result1.ShouldBe("{\"data\":{\"hero\":\"hello\"}}");
 
-            // verify executing with Schema2 works with IDocumentExecuter<>
+            // verify executing with Schema2 works with IDocumentExecuter<> directly
             var executer2 = provider.GetRequiredService<IDocumentExecuter<Schema2>>();
             var result2 = await executer2.ExecuteAsync(new ExecutionOptions { Query = "{hero}", RequestServices = provider });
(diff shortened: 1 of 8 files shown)


src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -326,4 +328,25 @@ protected virtual ExecutionContext BuildExecutionContext(ExecutionOptions option
...
+        private readonly IDocumentExecuter _documentExecuter;
+        public DocumentExecuter(IDocumentExecuter documentExecuter)
+        {
+            _documentExecuter = documentExecuter ?? throw new ArgumentNullException(nameof(documentExecuter));
+        }
+
+        public Task<ExecutionResult> ExecuteAsync(ExecutionOptions options)
+        {
+            if (options.Schema != null)
+                throw new InvalidOperationException("Schema must be null when calling this typed IDocumentExecuter<> implementation; it will be pulled from the dependency injection provider.");
```
sungam3r, line 344:
```suggestion
                throw new InvalidOperationException("ExecutionOptions.Schema must be null when calling this typed IDocumentExecuter<> implementation; it will be pulled from the dependency injection provider.");
```

src/GraphQL.Tests/Execution/DocumentExecuterTests.cs
```diff
@@ -50,6 +53,77 @@ public async Task Uses_ExecutionStrategySelector()
...
+            result1.ShouldBe("{\"data\":{\"hero\":\"hello\"}}");
+
+            // verify executing with Schema2 works with IDocumentExecuter<> directly
+            var executer2 = provider.GetRequiredService<IDocumentExecuter<Schema2>>();
+            var result2 = await executer2.ExecuteAsync(new ExecutionOptions { Query = "{hero}", RequestServices = provider });
+            var serializer = provider.GetRequiredService<IGraphQLTextSerializer>();
+            serializer.Serialize(result2).ShouldBe("{\"data\":{\"hero\":\"hello2\"}}");
+
+            // verify that you cannot specify Schema with this implementation
+            await Should.ThrowAsync<InvalidOperationException>(async () => await executer2.ExecuteAsync(new ExecutionOptions { Schema = new Schema1(provider), Query = "{hero}", RequestServices = provider }));
```
sungam3r, line 79: I would test message text here. As I said earlier it helps to understand the reason of exception.

sungam3r (approved):

src/GraphQL.Tests/Execution/DocumentExecuterTests.cs
Shane32:
```suggestion
            var err = await Should.ThrowAsync<InvalidOperationException>(async () => await executer2.ExecuteAsync(new ExecutionOptions { Schema = new Schema1(provider), Query = "{hero}", RequestServices = provider }));
            err.Message.ShouldBe("ExecutionOptions.Schema must be null when calling this typed IDocumentExecuter<> implementation; it will be pulled from the dependency injection provider.");
```

PR #3018, opened 2022-03-21
Prevent mutating JsonSerializerOptions for .NET 5+

Shane32:


sungam3r:
Tests failed

src/GraphQL.Tests/Serialization/SystemTextJson/GraphQLSerializerTests.cs
```diff
@@ -1,10 +1,34 @@
 using System.Text.Json;
+using GraphQL.Execution;
```
Shane32, line 2:
```suggestion
```

Shane32:
Coverlet keeps failing.

Shane32:
Running attempt 4 now...

Shane32:
@sungam3r Finally passed.

src/GraphQL.SystemTextJson/GraphQLSerializer.cs
```diff
@@ -78,9 +78,14 @@ public GraphQLSerializer(Action<JsonSerializerOptions> configureSerializerOption
         /// <param name="serializerOptions">Specifies the JSON serializer settings</param>
         public GraphQLSerializer(JsonSerializerOptions serializerOptions)
         {
+#if NET5_0_OR_GREATER
+            // clone serializerOptions
+            SerializerOptions = new(serializerOptions ?? throw new ArgumentNullException(nameof(serializerOptions)));
```
sungam3r, line 83: Hmm, new API since NET5?

sungam3r: It worth to add some notes like this in summary or for `serializeOptions` argument:
> Note that `JsonSerializerOptions` is designed in such a way to reuse created objects. This leads to a massive speedup in subsequent calls to the serializer. The downside is you can’t change properties on the options object after you’ve passed it in a Serialize()/Deserialize() call. You’ll get the exception: System.InvalidOperationException: Serializer options cannot be changed once serialization or deserialization has occurred. To get around this problem we create a copy of options on NET5 platform and above. Passed options object remains unchanged and all possible changes you made are unobserved.

sungam3r: What about TFMs < 5 ?

Shane32: Well, I'm just fixing what I can easily.  Fixing the bug for older version is up for grabs.

Shane32: Sure we can add the comment in there somewhere.  Not really sure people care about 'what makes it tick', so long as it does "tick", but sure.

Shane32: I suggest a slight change: "and all possible changes you made" to "and any changes you make"

Shane32: Updated -- added to constructors that accept `JsonSerializerOptions` (not `Action<JsonSerializerOptions>`)

sungam3r: > Not really sure people care about 'what makes it tick', so long as it does "tick", but sure.

This comment is needed for those cases if someone thought to change the values of options later. This is a warning. Not everyone knows how STJ works undercover.

sungam3r (approved):

PR #3025, opened 2022-03-25
Add ScopedSubscriptionExecutionStrategy

Shane32:
I will add tests sometime this weekend, but if you want to review the API and code @sungam3r , please do.


src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
```diff
@@ -35,15 +35,15 @@ public SubscriptionExecutionStrategy(IExecutionStrategy baseExecutionStrategy)
...
     /// <br/><br/>
     /// Keep in mind that if a scoped context is passed into <see cref="ExecutionContext.RequestServices"/>,
     /// and if it is disposed after the initial execution, node executions of subsequent data events will contain
     /// the disposed <see cref="ExecutionContext.RequestServices"/> instance and hence be unusable.
     /// <br/><br/>
-    /// If scoped services are needed, it is recommended to override
-    /// <see cref="ProcessDataAsync(ExecutionContext, ExecutionNode, object?)"/>
-    /// with an implementation that creates a service scope during execution of the data events.
+    /// If scoped services are needed, it is recommended to utilize the ScopedSubscriptionExecutionStrategy
+    /// class from the GraphQL.MicrosoftDI package, which will create a service scope during execution of data events.
```
sungam3r, line 46:
```suggestion
    /// class from the GraphQL.MicrosoftDI package, which will create a service scope during processing of data events.
```

I think that GraphQL requests are _executed_ but data events are _processed_.

src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
```diff
@@ -122,33 +122,42 @@ public override async Task<ExecutionResult> ExecuteAsync(ExecutionContext contex
...
-            Errors = new ExecutionErrors(),
-            OutputExtensions = new Dictionary<string, object?>(),
-            Metrics = Instrumentation.Metrics.None,
-            CancellationToken = token,
-        };
     }
 
+    /// <summary>
+    /// Clones an execution context without stateful information -- errors, metrics, and output extensions.
+    /// Sets the cancellation token on the clonsed context to the specified value.
```
sungam3r, line 149:
```suggestion
    /// Sets the cancellation token on the cloned context to the specified value.
```

src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
```diff
@@ -122,33 +122,42 @@ public override async Task<ExecutionResult> ExecuteAsync(ExecutionContext contex
...
+                    using var childContext = CloneExecutionContext(preservedContext, token);
 
                     return await ProcessDataAsync(childContext, node, value).ConfigureAwait(false);
                 },
                 async (exception, token) =>
                 {
-                    using var childContext = CloneExecutionContext(context, token);
+                    using var childContext = CloneExecutionContext(preservedContext, token);
 
-                    return await ProcessErrorAsync(context, node, exception).ConfigureAwait(false);
```
sungam3r, line 140: bugfix? :)

src/GraphQL.MicrosoftDI/ScopedSubscriptionExecutionStrategy.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        var newContext = base.CloneExecutionContext(context, token);
+        // prevent the initial service provider, likely a scoped service provider
+        // which will be disposed after the initial execution completes,
+        // from having a reference held to it from the preserved execution context
+        // necessary for subsequent data event executions
+        newContext.RequestServices = null;
+        return newContext;
+    }
+    /// <summary>
```
sungam3r, line 52:
```suggestion
    }

    /// <summary>
```

src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
```diff
@@ -122,33 +122,42 @@ public override async Task<ExecutionResult> ExecuteAsync(ExecutionContext contex
...
+
         // cannot throw an exception here
         return source
             .SelectCatchAsync(
                 async (value, token) =>
                 {
                     // duplicate context to prevent multiple event streams from sharing the same context,
                     // and clear errors/metrics/extensions. Free array pool leased arrays after execution.
-                    using var childContext = CloneExecutionContext(context, token);
+                    using var childContext = CloneExecutionContext(preservedContext, token);
```
sungam3r, line 135: 🤔 Double clone here and for error. Why did you add `preservedContext` ?

src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
Shane32: yup

src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
Shane32: mostly so that the disposed service provider during the initial execution can be cleared, so no references are being held to it unnecessarily.  besides that, references to the other unused fields (output extensions, metrics, etc) are also released to the GC.

It is still necessary to clone for each processed data event so that it does not 'corrupt' original clone.

sungam3r: I understand general intention about clearing service provider and other stuff. I don't understand about **double** cloning.

Shane32: Because the original `context` is held indefinitely until the subscription ends.  It is not desirable to hold a reference to the service provider (which may be scoped) or other transient information.  Nor can we clear it at this point because the DocumentExecuter has not finished.  So we clone it prior to SelectCatchAsync so that only the references we need to hold are held.

With the changes you can hold your mouse over the delegate and see that it says "captured references" are "this" and "preservedContext" now, and not "context".

So far so good.  Then we move on to execution of processed data events.  Events are processed asynchronously and, in the case that processing does not occur synchronously (at some point it actually awaits something), it can occur simultaneously.  So it is necessary for each processing event to own its own context, and being able to create errors, output extensions and/or a service scope for each execution.  So it is cloned prior to the execution.

Note that it is not cloned twice per data event.  It is cloned once initially, and then cloned from the clone for each subsequent execution.

If we do not add the "additional" clone, then we run into the issue you pinned previously, where the SelectCatchAsync is going to hold a reference to a disposed service provider including all of the disposed objects that were created during initial execution.

sungam3r: >Note that it is not cloned twice per data event. It is cloned once initially, and then cloned from the clone for each subsequent execution.

🤦‍♂️  Understood.

sungam3r (approved):

PR #3031, opened 2022-03-28
Fix deserialization bug for batches containing null

Shane32:
Fixes a bug where if a batch request as follows was processed by the server:

```json
[null]
```

Then the server would only receive an empty array and respond like this:

```json
[]
```

Rather than the correct response which should be something of this nature:

```json
[
  {
    "errors": [{"message":"GraphQL request cannot be null."}]
  }
]
```

### DIFF
--- a/src/GraphQL.NewtonsoftJson/GraphQLRequestListJsonConverter.cs
+++ b/src/GraphQL.NewtonsoftJson/GraphQLRequestListJsonConverter.cs
@@ -9,7 +9,7 @@ namespace GraphQL.NewtonsoftJson
     /// <br/><br/>
     /// To determine if a single request is a batch request or not, deserialize to the type
     /// <see cref="IList{T}">IList</see>&lt;<see cref="GraphQLRequest"/>&gt; and examine the type
-    /// of the returned object to see if it <see cref="GraphQLRequest"/>[].
+    /// of the returned object to see if it is <see cref="GraphQLRequest"/>[].
     /// If the returned object is an array, then it is not a batch request.
     /// </summary>
     public class GraphQLRequestListJsonConverter : JsonConverter
--- a/src/GraphQL.NewtonsoftJson/GraphQLRequestListJsonConverter.cs
+++ b/src/GraphQL.NewtonsoftJson/GraphQLRequestListJsonConverter.cs
@@ -6,6 +6,11 @@ namespace GraphQL.NewtonsoftJson
     /// <summary>
     /// A custom JsonConverter for reading a list of <see cref="GraphQLRequest"/> objects.
     /// Will deserialize a single request into a list containing one request. Doesn't support writing.
+    /// <br/><br/>
+    /// To determine if a single request is a batch request or not, deserialize to the type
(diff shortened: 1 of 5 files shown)


Shane32:
Apparently there are situations where NSJ executes a JsonConverter in the case of nulls.

Shane32:
@sungam3r  Please let me know if you have any comments.  I'm merging this so I can continue.

sungam3r:
In practice, have you encountered this?

src/GraphQL.NewtonsoftJson/GraphQLRequestListJsonConverter.cs
```diff
@@ -60,9 +66,8 @@ public override object ReadJson(JsonReader reader, Type objectType, object? exis
                         : list;
                 }
 
-                var request = serializer.Deserialize<GraphQLRequest>(reader);
-                if (request != null)
-                    list.Add(request);
+                var request = serializer.Deserialize<GraphQLRequest>(reader)!;
```
sungam3r, line 69: Does `request` always `null`?

Shane32:
> In practice, have you encountered this?

Yes I ran into this bug while writing tests for my updated server project.

src/GraphQL.NewtonsoftJson/GraphQLRequestListJsonConverter.cs
Shane32: No -- it will be null in the example shown at the top of this PR.

Shane32:
Most likely this was a bug I introduced while writing the updated serialization engine for v5, not realizing the importance here to be able to provide a proper result.

src/GraphQL.NewtonsoftJson/GraphQLRequestListJsonConverter.cs
sungam3r: So list of non-nullable requests will contain `null` here but later this `null` will be properly handled, right? I doubt since I see these lines in server:
```
await RequestExecutingAsync(gqlRequestInBatch, i);
var result = await ExecuteRequestAsync(gqlRequestInBatch, userContext, executer, context.RequestServices, cancellationToken);
```

`gqlRequestInBatch` should be `null` if I understand correctly. So eventually code will throw NRE here:
```
 protected virtual ExecutionOptions GetOptions(GraphQLRequest request, IDictionary<string, object> context, IServiceProvider requestServices, CancellationToken cancellationToken)
        {
            {
                Schema = Schema,
                OperationName = request.OperationName, <--------------------- BOOM
                Query = request.Query,
                Variables = request.Variables,
                Extensions = request.Extensions,
                UserContext = context,
                CancellationToken = cancellationToken,
                ComplexityConfiguration = _options.ComplexityConfiguration,
                EnableMetrics = _options.EnableMetrics,
                UnhandledExceptionDelegate = _options.UnhandledExceptionDelegate,
                MaxParallelExecutionCount = _options.MaxParallelExecutionCount,
                RequestServices = requestServices,
            };

            return opts;
        }
```

Shane32: I agree.  I've already fixed this in my copy of the server project and added a test for it.

Issue #3030, opened 2022-03-28
Nested Fragments


ansulgoenka1:
## Summary

When I am trying to pass a query that contains nested fragments, the DocumentExecutor throws this Error: "Error executing the document. The given key was not present in the dictionary."

## Relevant information

Query: 

                fragment nameFragment on Dog {
                    name
                    ... nestedNameFragment
                }

                fragment nestedNameFragment on Dog {
                    name
                }

                {
                    dog {
                        name
                        ... nameFragment
                    }
                }

Is this query correct? If not how do we use nested fragments here or is this a bug? I could not find any documentation on this

Any help would be greatly appreciated

sungam3r:
I'll try to reproduce it.

ansulgoenka1:
I am using the 5.00 (preview) version.

I was debugging the code and found that document.Definition (after using DocumentBuilder) contains errors for this query. When we use DocumentExecutor and set Complexity Configuration, ComplexityAnalyzer Line 60 throws this error when it iterates over the document.Definition (probably because it cannot find error in that switch condition)

ansulgoenka1:
Your test suite passes this test as it uses DocumentValidator

sungam3r:
>  document.Definition (after using DocumentBuilder) **contains errors** for this query

I don't understand. How can GraphQL AST contain _errors_? Could you please post PR with test repro?

ansulgoenka1:
It can also be reproduced by debugging all_fragment_names_are_used in NoUsedFragmentTests

![image](https://user-images.githubusercontent.com/102316754/160473916-d58d1dae-47fb-4f16-ae65-c8b43bb96fe3.png)

sungam3r:
See linked PR, it's just a displayed text for debugger.

sungam3r:
@ansulgoenka1 

I'm interested in that:
> DocumentExecutor throws this Error: "Error executing the document. The given key was not present in the dictionary."

Could you write a repro after #3032 merged?

ansulgoenka1:
If you replace DocumentValidator with DocumentExecutor in ValidationTestBase line 89:

private ExecutionResult Validate(string query, ISchema schema, IEnumerable<IValidationRule> rules, Inputs variables)
        {
            var validator = new DocumentExecuter();
            return validator.ExecuteAsync(new ExecutionOptions
            {
                Schema = schema,
                ValidationRules = rules,
[...]

PR #3040, opened 2022-03-30
For 4.8, restrict parser to pre-8 versions

Shane32:
Parser 8.0.0 is not compatible with GraphQL.NET 4.8


sungam3r:
@Shane32 I fixed #3030 . I'll make PR soon.

sungam3r (approved):

PR #3051, opened 2022-03-31
Add RequestServices to ValidationContext

Shane32:
I tried to write a validation rule today and was quite surprised to find that I could not access the scoped service provider provided to the request.

Also, to add an error to the validation context, you have to add an instance of `ValidationError`, which currently requires that `Number` be provided.  This is merely a NRT error, as `Number` is not actually required.  Note that `Number` is defined as the following:

> Gets or sets the rule number of this validation error corresponding to the paragraph number from the official specification.

So obviously there is no "paragraph number from the official specification" for custom validation errors.  Hence `Number` should not be required for custom validation errors anyway.

### DIFF
--- a/src/GraphQL/Validation/ValidationContext.cs
+++ b/src/GraphQL/Validation/ValidationContext.cs
@@ -25,6 +25,7 @@ namespace GraphQL.Validation
             NonUserContext = null;
             Variables = null!;
             Extensions = null!;
+            RequestServices = null!;
         }
 
         /// <summary>
--- a/src/GraphQL/Validation/ValidationContext.cs
+++ b/src/GraphQL/Validation/ValidationContext.cs
@@ -66,6 +66,9 @@ namespace GraphQL.Validation
         /// <inheritdoc cref="ExecutionOptions.Extensions"/>
         public Inputs Extensions { get; set; } = null!;
 
+        /// <inheritdoc cref="ExecutionOptions.RequestServices"/>
+        public IServiceProvider? RequestServices { get; set; }
+
         /// <summary>
         /// <see cref="System.Threading.CancellationToken">CancellationToken</see> to cancel validation of request;
         /// defaults to <see cref="CancellationToken.None"/>
--- a/src/GraphQL/Validation/ValidationError.cs
+++ b/src/GraphQL/Validation/ValidationError.cs
@@ -21,9 +21,8 @@ namespace GraphQL.Validation
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="ValidationError"/> class with a specified error message. Sets the
-        /// <see cref="ExecutionError.Code">Code</see> property based on the inner exception.
-        /// Loads any exception data from the inner exception into this instance.
+        /// Initializes a new instance of the <see cref="ValidationError"/> class with a specified error message and code.
+        /// Sets additional codes based on the inner exception(s). Loads any exception data from the inner exception into this instance.
         /// </summary>
         public ValidationError(string message, Exception? innerException) : base(message, innerException)
         {
--- a/src/GraphQL/Validation/ValidationError.cs
+++ b/src/GraphQL/Validation/ValidationError.cs
@@ -12,8 +12,26 @@ namespace GraphQL.Validation
     {
         private readonly List<ASTNode> _nodes = new();
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ValidationError"/> class with a specified error message.
+        /// </summary>
+        public ValidationError(string message) : base(message)
+        {
+            Code = GetValidationErrorCode(GetType());
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ValidationError"/> class with a specified error message. Sets the
+        /// <see cref="ExecutionError.Code">Code</see> property based on the inner exception.
+        /// Loads any exception data from the inner exception into this instance.
+        /// </summary>
+        public ValidationError(string message, Exception? innerException) : base(message, innerException)
+        {
+            Code = GetValidationErrorCode(GetType());
+        }
+
         /// <inheritdoc cref="ValidationError(ROM, string, string, ASTNode[])"/>
-        public ValidationError(ROM originalQuery, string? number, string message, ASTNode node)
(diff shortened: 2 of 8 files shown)


sungam3r:
>  Hence Number should not be required for custom validation errors anyway.

I would keep it as non-null. Writing non-standard rules is not a frequent work. Standard rules should not suffer because of this forcing callers to check `Number` for `null`. For non-standard rules number _may_ be used as well. Code from auth project:
```csharp
 public AuthorizationError(ASTNode? node, ValidationContext context, string message, AuthorizationResult result, OperationType? operationType = null)
            : base(context.Document.Source, "6.1.1", message, node == null ? Array.Empty<ASTNode>() : new ASTNode[] { node })
```

6.1.1 points to the paragraph with generic explanation of validation process.

Shane32:
Then we should have an alternate way of adding a validation rule without a number.  There is no practical reason why a number should be required.  It is not part of the spec, and we already have codes, also not part of the spec.  Why should a custom validation rule require a number?  Clearly we have set up the system for custom validation rules, and clearly those will not typically have a number from the spec assigned to it.

I can add constructors that mirrors those of `DocumentError`.  That is a better plan anyway.

> Standard rules should not suffer because of this forcing callers to check Number for null. 

I seriously don't know what you're talking about.  The serializers already check `Number` for `null` as they always have.  No user code needs to do so.  Most likely this change is just the result of incorrect NRT annotations during the early part of 4.x.  If we want a specific class for validation errors that relate to the spec, then we should add a class like `SpecValidationError`.  Not require `Number` for EVERY validation error.

Shane32:
I'm not sure what other people do, but this is the second validation rule I have written in the last couple weeks.  Neither had to do with anything to do with the spec.  Could I have punched a string in the field?  Sure.  But that is not the intent of the field.

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -66,6 +66,9 @@ internal void Reset()
         /// <inheritdoc cref="ExecutionOptions.Extensions"/>
         public Inputs Extensions { get; set; } = null!;
 
+        /// <inheritdoc cref="ExecutionOptions.RequestServices"/>
+        public IServiceProvider? RequestServices { get; set; }
```
sungam3r, line 70: Forgot to reset it in `Reset` method.

src/GraphQL/Validation/ValidationOptions.cs
```diff
@@ -52,6 +52,9 @@ public ValidationOptions()
         /// </summary>
         public GraphQLOperationDefinition Operation { get; init; } = null!;
 
+        /// <inheritdoc cref="ExecutionOptions.RequestServices"/>
+        public IServiceProvider? RequestServices { get; init; } = null;
```
sungam3r, line 56: I was a bit confused about changing struct size but it seems as a non-breaking change
https://stackoverflow.com/questions/37865610/is-changing-the-size-of-a-struct-a-breaking-change-in-c

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -88,6 +106,6 @@ internal static string GetValidationErrorCode(Type type)
         /// <summary>
         /// Gets or sets the rule number of this validation error corresponding to the paragraph number from the official specification.
```
sungam3r, line 107:
```suggestion
        /// Gets or sets the rule number of this validation error corresponding to the paragraph number from the official specification if any.
```

src/GraphQL/Validation/ValidationError.cs
```diff
@@ -12,6 +12,24 @@ public class ValidationError : DocumentError
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ValidationError"/> class with a specified error message. Sets the
+        /// <see cref="ExecutionError.Code">Code</see> property based on the inner exception.
+        /// Loads any exception data from the inner exception into this instance.
+        /// </summary>
+        public ValidationError(string message, Exception? innerException) : base(message, innerException)
+        {
+            Code = GetValidationErrorCode(GetType());
```
sungam3r, line 30: > Code property based on the inner exception

You override code with one from current exception.

src/GraphQL/Validation/ValidationContext.cs
Shane32: Good catch.  Fixed.

src/GraphQL/Validation/ValidationError.cs
Shane32: This is copied from the existing constructors here and I believe it is valid.  The idea is to return the Code that points to the current derived of the validation error class -- for example an InvalidVariableError always returns INVALID_VALUE regardless of what inner exception caused the value conversion problem.  However Codes can still be examined to see the the inner exception code.  (Compared to UnhandledError which would always give the code for the underlying exception.)

sungam3r: > Sets additional codes based on the inner exception(s).

This remark no longer belongs to this class, it is better to remove it. No more `Codes` property here. `ErrorInfoProvider` as a default implementation of `IErrorInfoProvider` reads codes in `GetCodesForError` and then stores this info into `extensions`.

Shane32: While that is true, it helps the user to understand what the effective result of passing an exception to the constructor is.  That is why the comment exists that way in all of these classes.  Any changes to the comments should retain this concept, even if it is written as a remark (for example, "Inner exceptions are typically used to write the 'codes' property to the error's extension property.").

While I personally think it's fine, feel free to issue a separate PR to change the language of the xml comments throughout the source.  As of now this comment will retain the same language as the rest of the execution error class constructors.

Shane32: I will clarify this a bit.

Shane32: Well, I don't know what to do.  It could use clarification.  I'm leaving it.  Feel free to make a PR.

Shane32: The xml comment in DocumentError needs to be fixed too.

sungam3r: ok

sungam3r (approved):

Issue #2982, opened 2022-03-02
Build checks intermittently fail


Shane32:
Continuing build check failures:

- `GraphQL.Tests.Introspection.AppliedDirectivesTests.introspection_should_return_applied_directives`
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5384459140?check_suite_focus=true
  - Sometimes it returns `"name": "RootMutation"` (correct) and sometimes `"name": "AppliedDirectivesTests_RootMutation"` (incorrect)
- `GraphQL.Tests.Initialization.SchemaInitializationTests.SchemaWithArgumentsOnInputField_Should_Throw`
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5737387227?check_suite_focus=true
  - Also seems to have to do with a static option incorrectly set

Fixed:
- `GraphQL.Tests.Types.AllScalarGraphTypeTests.parseValue_from_system_text_json_ok`
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5378902286?check_suite_focus=true
  - Culture problem with negative value
- Fails when all tests pass
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5383976682?check_suite_focus=true
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5378827075?check_suite_focus=true
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5603665805?check_suite_focus=true
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5651194580?check_suite_focus=true
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5721750186?check_suite_focus=true
  - https://github.com/graphql-dotnet/graphql-dotnet/runs/5801756371?check_suite_focus=true
  - Seems to be issues with coverlet

Shane32:
I will continue to update this PR as more checks fail.

sungam3r2022:
> "RootMutation" (correct) and sometimes "name": "AppliedDirectivesTests_RootMutation"

Known issue with static setting.

sungam3r2022:
Note:

>Additional Info:
    Converted value: −1E+25, Type: **System.Double**

I added this trace some time ago to deal with failed results.

sungam3r2022:
Something wrong with `JsonConverterBigInteger.TryGetBigInteger` called from `InputsJsonConverter.ReadNumber`

sungam3r2022:
[...]

PR #3058, opened 2022-04-02
Small code style and xmldoc fixes

### LINKED ISSUE #3057: Rewrite xml comments for error codes
While that is true, it helps the user to understand what the effective result of passing an exception to the constructor is.  That is why the comment exists that way in all of these classes.  Any changes to the comments should retain this concept, even if it is written as a remark (for example, "Inner exceptions are typically used to write the 'codes' property to the error's extension property.").

While I personally think it's fine, feel free to issue a separate PR to change the language of the xml comments throughout the source.  As of now this comment will retain the same language as the rest of the execution error class constructors.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/pull/3051#discussion_r840978726_

sungam3r:
fixes #3057

### DIFF
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -76,7 +76,7 @@ namespace GraphQL
     {
         public ExecutionError(string message) { }
         public ExecutionError(string message, System.Collections.IDictionary data) { }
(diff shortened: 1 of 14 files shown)


Shane32 (approved):

Shane32:
Added failed test run to #2982

PR #3059, opened 2022-04-02
Switch to coverlet.collector

sungam3r:
fixes (probably) #2982

### DIFF
--- a/.github/workflows/test-code.yml
+++ b/.github/workflows/test-code.yml
@@ -66,7 +66,7 @@ jobs:
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         uses: codecov/codecov-action@v2.1.0
         with:
-          files: .coverage/GraphQL.DataLoader.Tests/coverage.net6.opencover.xml,.coverage/GraphQL.Harness.Tests/coverage.net6.opencover.xml,.coverage/GraphQL.Tests/coverage.net6.opencover.xml,.coverage/GraphQL.MicrosoftDI.Tests/coverage.net6.opencover.xml
+          files: "**/coverage.opencover.xml"
 
   buildcheck:
     needs:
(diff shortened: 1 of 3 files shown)


Shane32 (approved):

Shane32:
> fixes (probably) #2982

Fixes the coverlet problems anyway. I think I have a solution for some of the issues that have to do with static variables but I have not worked on it yet.

sungam3r:
@Shane32 I think that the reason for increasing the coverage (above) is that files for all TFMs have become transmitted to codecov (only net6 before).
изображение

sungam3r:
Or it is just a difference of coverage calculation between coverlet.msbuild and coverlet.collector.

.github/workflows/test-code.yml
```diff
@@ -61,12 +61,19 @@ jobs:
...
+          --results-directory .coverage
+          --
+          DataCollectionRunSettings.DataCollectors.DataCollector.Configuration.Format=opencover
+          DataCollectionRunSettings.DataCollectors.DataCollector.Configuration.Exclude=[GraphQL.StarWars]*,[GraphQL.Harness]*,[GraphQL.Dummy]*
       - name: Upload coverage to codecov
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         uses: codecov/codecov-action@v2.1.0
         with:
-          files: .coverage/GraphQL.DataLoader.Tests/coverage.net6.opencover.xml,.coverage/GraphQL.Harness.Tests/coverage.net6.opencover.xml,.coverage/GraphQL.Tests/coverage.net6.opencover.xml,.coverage/GraphQL.MicrosoftDI.Tests/coverage.net6.opencover.xml
+          files: 'src/.coverage/**/coverage.opencover.xml'
```
sungam3r, line 76: No need anymore to remember adding new projects here.

sungam3r: I think codecov had no globbing support before.

src/Directory.Build.targets
```diff
@@ -1,11 +1,5 @@
 <Project>
 
-  <PropertyGroup Condition="'$(IsTestProject)' == 'true'">
```
sungam3r, line 3: Moved into `dotnet test` arguments.

PR #3060, opened 2022-04-03
Make tests that modify static variables run without parallelism

Shane32:
Fixes remainder of #2982 issues (hopefully).

See:
- https://xunit.net/docs/shared-context
- https://xunit.net/docs/running-tests-in-parallel

> By default, each test class is a unique test collection. Tests within the same test class will not run in parallel against each other.

> If we need to indicate that multiple test classes should not be run in parallel against one another, then we place them into the same test collection. This is simply a matter of decorating each test class with an attribute that places them into the same uniquely named test collection:

> Parallel-capable test collections will be run first (in parallel), followed by parallel-disabled test collections (run sequentially).

So with these changes,
  1. All tests that modify static variables are now placed in the same "test collection" called "StaticTests", so that all those tests "will not run in parallel against each other.
  2. The new test collection (called "StaticTests") has parallelism disabled, which means it will run sequentially after all other tests complete

So basically all tests within a class marked with `[Collection("StaticTests")]` are guaranteed to not run in parallel with any other tests.

All tests that use the `CultureTestHelper` or modify `GlobalSwitches` have had their test classes marked with `[Collection("StaticTests")]`

### DIFF
--- a/src/GraphQL.Tests/CultureTestHelper.cs
+++ b/src/GraphQL.Tests/CultureTestHelper.cs
@@ -17,6 +17,7 @@ namespace GraphQL.Tests
         /// <summary>
         /// Executes the specified delegate with a variety of cultures.
         /// Be sure to mark the test class with <c>[Collection("StaticTests")]</c>
+        /// to avoid the interference of static variables.
         /// </summary>
         public static void UseCultures(Action scope)
         {
--- a/src/GraphQL.Tests/CultureTestHelper.cs
+++ b/src/GraphQL.Tests/CultureTestHelper.cs
@@ -18,7 +18,6 @@ namespace GraphQL.Tests
         /// Executes the specified delegate with a variety of cultures.
         /// Be sure to mark the test class with <c>[Collection("StaticTests")]</c>
         /// </summary>
-        /// <param name="scope"></param>
         public static void UseCultures(Action scope)
         {
             foreach (var culture in Cultures)
--- a/src/GraphQL.Tests/CultureTestHelper.cs
+++ b/src/GraphQL.Tests/CultureTestHelper.cs
@@ -4,7 +4,7 @@ namespace GraphQL.Tests
 {
     public static class CultureTestHelper
     {
-        public static IEnumerable<CultureInfo> Cultures => new[]
+        private static IEnumerable<CultureInfo> Cultures => new[]
         {
             new CultureInfo("ru-RU"),
             new CultureInfo("fi-FI"),
@@ -14,6 +14,11 @@ namespace GraphQL.Tests
             new CultureInfo(CultureInfo.CurrentCulture.Name)
         };
 
+        /// <summary>
+        /// Executes the specified delegate with a variety of cultures.
+        /// Be sure to mark the test class with <c>[Collection("StaticTests")]</c>
+        /// </summary>
+        /// <param name="scope"></param>
         public static void UseCultures(Action scope)
         {
             foreach (var culture in Cultures)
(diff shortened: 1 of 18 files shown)


src/GraphQL.Tests/CultureTestHelper.cs
```diff
@@ -14,6 +14,11 @@ public static class CultureTestHelper
             new CultureInfo(CultureInfo.CurrentCulture.Name)
         };
 
+        /// <summary>
+        /// Executes the specified delegate with a variety of cultures.
+        /// Be sure to mark the test class with <c>[Collection("StaticTests")]</c>
+        /// </summary>
+        /// <param name="scope"></param>
```
Shane32, line 21:
```suggestion
```

sungam3r:
🤔 I think some time long ago I have already tried to do the same and I failed apparently because I missed point 2 about disable parallelization for dedicated collection so moving it after others.

src/GraphQL.Tests/CultureTestHelper.cs
```diff
@@ -14,6 +14,10 @@ public static class CultureTestHelper
             new CultureInfo(CultureInfo.CurrentCulture.Name)
         };
 
+        /// <summary>
+        /// Executes the specified delegate with a variety of cultures.
+        /// Be sure to mark the test class with <c>[Collection("StaticTests")]</c>
+        /// </summary>
```
sungam3r, line 20:
```suggestion
        /// Be sure to mark the test class with <c>[Collection("StaticTests")]</c>
        /// to avoid the interference of static variables.
        /// </summary>
```

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -8,6 +8,7 @@
 
 namespace GraphQL.Tests.Types
 {
+    [Collection("StaticTests")]
```
sungam3r, line 11: Try to uncomment all lines after
> // disabled to make tests stable without touch GlobalSwitches

src/GraphQL.Tests/ObjectExtensionsTest.cs
```diff
@@ -1,5 +1,6 @@
 namespace GraphQL.Tests
 {
+    [Collection("StaticTests")]
```
sungam3r, line 3: As I understand, there is no need to capture all the tests from test classes to make them sequential - only ones that uses `UseCultures` but I'm OK to keep as is since performance drop is almost invisible.

src/GraphQL.Tests/StaticTestsCollection.cs
```diff
@@ -0,0 +1,11 @@
+namespace GraphQL.Tests;
+
+// decorate any test classes containing tests that modify global variables with
+//   [Collection("StaticTests")]
+// these tests will run sequentially after all other tests are complete
+// be sure to restore the global variables to their initial state with a finally block
+
+[CollectionDefinition("StaticTests", DisableParallelization = true)]
+public class StaticTestsCollection
```
sungam3r, line 9: Class name can be arbitrary?

sungam3r (approved):

src/GraphQL.Tests/ObjectExtensionsTest.cs
Shane32: That’s what I was thinking as well

src/GraphQL.Tests/StaticTestsCollection.cs
Shane32: Yes

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32: Done

PR #3067, opened 2022-04-08
Add role support to authorization extensions

Shane32:
This allows the role name (or multiple allowed role names) to be directly specified, bringing the `GraphQLAuthorizeAttribute` more in line with the ASP.Net `AuthorizeAttribute`.

It also will be simpler to use for those (like me) that often just wish to specify one or more roles rather than having to create policies that map to the configured roles.

Finally, as it is used in the GraphQL Server project, checking for roles is a much faster operation (probably zero-allocation), as the claim principal's `IsInRole` method can be easily called rather than having to call the asynchronous policy checker method and then using `GetAwaiter().GetResult()` to validate the policy.

### DIFF
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -608,6 +608,19 @@ This allows for specifying roles rather than just policies that can be used to v
 As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), it requires support by a third-party
 library to perform the validation.
 
+Similar to the ASP.Net Core `AuthorizeAttribute`, the new `AuthorizeWithRoles` method accepts
+a comma-separated list of role names that would allow access to the graph or field.
+
+```csharp
+graph.AuthorizeWithRoles("Administrators,Managers");
+```
+
+You may also supply a list of strings as in the following example:
+
+```csharp
+graph.AuthorizeWithRoles("Administrators", "Managers");
+```
+
 ## Breaking Changes
 
 ### 1. UnhandledExceptionDelegate
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -1118,4 +1118,4 @@ This was done for the purposes of the overall strategy for reducing memory consu
 
 ### 47. `AuthorizeWith` renamed to `AuthorizeWithPolicy`
 
-This change was made to clarify and differentiate between `AuthorizeWithRoles`
+This change was made to clarify and differentiate between `AuthorizeWithRoles`.
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -608,7 +608,7 @@ This allows for specifying roles rather than just policies that can be used to v
 As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), it requires support by a third-party
 library to perform the validation.
 
-Similar to the ASP.Net Core `AuthorizeAttribute`, the new `AuthorizeWithRoles` method accepts
+Similar to the ASP.NET Core `AuthorizeAttribute`, the new `AuthorizeWithRoles` method accepts
 a comma-separated list of role names that would allow access to the graph or field.
 
 ```csharp
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -602,6 +602,12 @@ There are a number of other minor issues fixed; see these links for more details
 - https://github.com/graphql-dotnet/graphql-dotnet/issues/3002
 - https://github.com/graphql-dotnet/graphql-dotnet/pull/3004
 
+### 20. `AuthorizeWithRoles` added in GraphQL 5.1.0
+
+This allows for specifying roles rather than just policies that can be used to validate a request.
+As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), it requires support by a third-party
+library to perform the validation.
+
 ## Breaking Changes
 
 ### 1. UnhandledExceptionDelegate
@@ -1109,3 +1115,7 @@ GraphQL.NET v5 it is enabled by default as part of the `DocumentValidator.CoreRu
 ### 46. `ValidationContext.GetRecursiveVariables` returns null instead of an empty list
 
 This was done for the purposes of the overall strategy for reducing memory consumption.
+
+### 47. `AuthorizeWith` renamed to `AuthorizeWithPolicy`
+
+This change was made to clarify and differentiate between `AuthorizeWithRoles`
--- a/src/GraphQL.Tests/AuthorizationTests.cs
+++ b/src/GraphQL.Tests/AuthorizationTests.cs
@@ -1,0 +1,116 @@
+using GraphQL.Types;
+
+namespace GraphQL.Tests;
+
+public class AuthorizationTests
+{
+    [Fact]
+    public void Field()
+    {
+        var field = new FieldType();
+        field.RequiresAuthorization().ShouldBeFalse();
+        field.AuthorizeWith("Policy1");
+        field.RequiresAuthorization().ShouldBeTrue();
+        field.AuthorizeWith("Policy2");
+        field.AuthorizeWith("Policy2");
+        field.AuthorizeWithRoles("Role1,Role2");
+        field.AuthorizeWithRoles("Role3, Role2");
+
+        field.RequiresAuthorization().ShouldBeTrue();
+        field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2" });
+        field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+    }
+
+    [Fact]
+    public void FieldBuilder()
+    {
+        var graph = new ObjectGraphType();
+        graph.Field<StringGraphType>("Field")
+            .AuthorizeWith("Policy1")
+            .AuthorizeWith("Policy2")
+            .AuthorizeWith("Policy2")
+            .AuthorizeWithRoles("Role1,Role2")
+            .AuthorizeWithRoles("Role3, Role2");
+
+        var field = graph.Fields.Find("Field");
+        field.RequiresAuthorization().ShouldBeTrue();
+        field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2" });
+        field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+    }
+
+    [Fact]
+    public void ConnectionBuilder()
+    {
+        var graph = new ObjectGraphType();
+        graph.Connection<StringGraphType>()
+            .Name("Field")
+            .AuthorizeWith("Policy1")
+            .AuthorizeWith("Policy2")
+            .AuthorizeWith("Policy2")
+            .AuthorizeWithRoles("Role1,Role2")
+            .AuthorizeWithRoles("Role3, Role2");
+
+        var field = graph.Fields.Find("Field");
+        field.RequiresAuthorization().ShouldBeTrue();
+        field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2" });
+        field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+    }
+
+    [Fact]
+    public void AutoOutputGraphType()
+    {
+        var graph = new AutoRegisteringObjectGraphType<Class1>();
+        graph.RequiresAuthorization().ShouldBeTrue();
+        graph.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
+        graph.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+
+        graph.Fields.Find("Id").RequiresAuthorization().ShouldBeFalse();
+
+        var field = graph.Fields.Find("Name");
+        field.RequiresAuthorization().ShouldBeTrue();
+        field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
+        field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+    }
+
+    [Fact]
+    public void SchemaBuilder()
+    {
+        var schema = Schema.For(
+            @"
+type Class1 {
+  id: String!
+  name: String!
+}",
+            configure => configure.Types.Include<Class1>());
+
+        var graph = (ObjectGraphType)schema.AllTypes["Class1"];
+        graph.RequiresAuthorization().ShouldBeTrue();
+        graph.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
+        graph.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+
+        graph.Fields.Find("id").RequiresAuthorization().ShouldBeFalse();
+
+        var field = graph.Fields.Find("name");
+        field.RequiresAuthorization().ShouldBeTrue();
+        field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
+        field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+    }
+
+    [GraphQLAuthorize("Policy1")]
+    [GraphQLAuthorize("Policy2")]
+    [GraphQLAuthorize("Policy2")]
+    [GraphQLAuthorize(Policy = "Policy3")]
+    [GraphQLAuthorize(Roles = "Role1,Role2")]
+    [GraphQLAuthorize(Roles = "Role3, Role2")]
+    private class Class1
+    {
+        public string Id { get; set; }
+        [GraphQLAuthorize("Policy1")]
+        [GraphQLAuthorize("Policy2")]
+        [GraphQLAuthorize("Policy2")]
+        [GraphQLAuthorize(Policy = "Policy3")]
+        [GraphQLAuthorize(Roles = "Role1,Role2")]
+        [GraphQLAuthorize(Roles = "Role3, Role2")]
+        public string Name { get; set; }
+    }
+}
--- a/src/GraphQL.Tests/AuthorizationTests.cs
+++ b/src/GraphQL.Tests/AuthorizationTests.cs
@@ -13,12 +13,15 @@ public class AuthorizationTests
         field.RequiresAuthorization().ShouldBeTrue();
         field.AuthorizeWith("Policy2");
         field.AuthorizeWith("Policy2");
+        field.AuthorizeWithPolicy("Policy3");
+        field.AuthorizeWithPolicy("Policy3");
         field.AuthorizeWithRoles("Role1,Role2");
         field.AuthorizeWithRoles("Role3, Role2");
+        field.AuthorizeWithRoles("Role1", "Role4");
 
         field.RequiresAuthorization().ShouldBeTrue();
-        field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2" });
-        field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+        field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
+        field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3", "Role4" });
     }
 
     [Fact]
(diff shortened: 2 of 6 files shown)


Shane32:
@sungam3r I'd appreciate it if you could review this today.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
```diff
@@ -4,10 +4,17 @@
 namespace GraphQL
 {
     /// <summary>
-    /// Attribute to apply authorization policy when using schema first approach.
```
sungam3r, line 7: > when using schema first approach.

No only schema first now?

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
```diff
@@ -19,18 +26,52 @@ public GraphQLAuthorizeAttribute(string policy)
...
+        /// Role names will be trimmed before adding.
+        /// </summary>
+        public string? Roles { get; set; }
 
         /// <inheritdoc />
-        public override void Modify(TypeConfig type) => type.AuthorizeWith(Policy);
+        public override void Modify(TypeConfig type)
+        {
+            if (Policy != null)
+                type.AuthorizeWith(Policy);
```
sungam3r, line 41: I would rename `AuthorizeWith` to `AuthorizeWithPolicy` to align with `AuthorizeWithRoles` name.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
```diff
@@ -4,10 +4,17 @@
 namespace GraphQL
 {
     /// <summary>
-    /// Attribute to apply authorization policy when using schema first approach.
+    /// Attribute to apply an authorization policy and/or roles to a graph or field.
     /// </summary>
     public class GraphQLAuthorizeAttribute : GraphQLAttribute
     {
+        /// <summary>
+        /// Creates an instance with the specified policy and/or roles.
```
sungam3r, line 12:
```suggestion
        /// Creates an empty instance of <see cref="GraphQLAuthorizeAttribute"/> with no policy/roles specified.
```

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
Shane32: No it is not.  It is used for "type-first" (i.e. the auto-registering object graph type) also.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
```diff
@@ -19,18 +26,52 @@ public GraphQLAuthorizeAttribute(string policy)
         /// <summary>
         /// The name of policy to apply.
         /// </summary>
-        public string Policy { get; }
+        public string? Policy { get; set; }
+
+        /// <summary>
+        /// A comma-separated list of the roles to apply.
```
sungam3r, line 32: Are you sure that it is better than `string[]` ? Do you want to align this API completely to https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authorization.authorizeattribute.roles?view=aspnetcore-6.0#microsoft-aspnetcore-authorization-authorizeattribute-roles ?

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
Shane32: That would be nice but it would be a breaking change.  I can add AuthorizeWithPolicy and optionally mark AuthorizeWith as obsolete, and probably update the v5 migration guide also since this change is so new.  Lmk.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
Shane32: I want to align this API with the attribute that most ASP.NET Core developers are used to, where the name "Roles" (not "Role") and they use a comma-separated list to indicate multiple roles, any of which are acceptable to execute.

So, `[GraphQLAuthorize(Roles = "Administrator, Power User")]` should function just like `[Authorize(Roles = "Administrator, Power User")]` does (in concept).

Note that as it relates to the policy, the API is already compatible -- `[GraphQLAuthorize("MyPolicy")]` and `[GraphQLAuthorize(Policy = "MyPolicy")]` aligns with ASP.Net Core `[Authorize("MyPolicy")]` and `[Authorize(Policy = "MyPolicy")]`

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
sungam3r: I know that it's breaking. Let's obsolete `AuthorizeWith`.

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            .AuthorizeWithRoles("Role3, Role2");
+
+        var field = graph.Fields.Find("Field");
+        field.RequiresAuthorization().ShouldBeTrue();
+        field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2" });
+        field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
+    }
+
+    [Fact]
+    public void AutoOutputGraphType()
```
Shane32, line 60: This test demonstrates and tests the manner in which the `GraphQLAuthorizeAttribute` can be used without schema-first.

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -63,6 +80,39 @@ public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataP
...
+        /// <typeparam name="TMetadataProvider"> The type of metadata provider. Generics are used here to
+        /// let compiler infer the returning type to allow methods chaining.
+        /// </typeparam>
+        /// <param name="provider">
+        /// Metadata provider. This can be an instance of <see cref="GraphType"/>,
+        /// <see cref="FieldType"/>, <see cref="Schema"/> or others.
+        /// </param>
+        /// <param name="roles"> Comma-separated list of authorization role name(s). </param>
+        /// <returns> The reference to the specified <paramref name="provider"/>. </returns>
+        public static TMetadataProvider AuthorizeWithRoles<TMetadataProvider>(this TMetadataProvider provider, string roles)
```
sungam3r, line 97: Add overload with `params string[] roles`? I'm a bit confused here about difference between _policy_ and _role_ and about chaining design:
```csharp
field.AuthorizeWith("mypolicy1").AuthorizeWith("mypolicy2");
```
but
```csharp
field.AuthorizeWithRoles("myrole1,myrole2").AuthorizeWithRoles("myrole3;myrole4;myrole5");
```

I do not like that in one case, a single number is used, and in another multiple. Techically we can implement both methods to accept multiple policies/roles. I would like to see clean API design here.

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -79,6 +129,23 @@ public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataP
...
         }
 
+        /// <summary>
+        /// Adds authorization role(s) to the specified field builder. Roles should
+        /// be comma-separated and role names will be trimmed. If the underlying field already
+        /// contains a role with the same name, then it will not be added twice.
+        /// </summary>
+        /// <typeparam name="TSourceType"></typeparam>
+        /// <typeparam name="TReturnType"></typeparam>
+        /// <param name="builder"></param>
```
sungam3r, line 139:
```suggestion
```

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -93,5 +160,21 @@ public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataP
...
             return builder;
         }
+
+        /// <summary>
+        /// Adds authorization role(s) to the specified connection builder. Roles should
+        /// be comma-separated and role names will be trimmed. If the underlying field already
+        /// contains a role with the same name, then it will not be added twice.
+        /// </summary>
+        /// <typeparam name="TSourceType"></typeparam>
+        /// <param name="builder"></param>
```
sungam3r, line 170:
```suggestion
```

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
Shane32: Just FYI, the authorize attribute has worked this way not only in ASP.Net Core since 1.0 but also in ASP.NET MVC 5.2 and likely prior versions as well.  So it is something people are likely used to, if they do any web programming in C#.

sungam3r: C# allows to set string arrays when initializing attributes:
```csharp
 [GraphQLAuthorize("mypolicy", Roles = new[] { "myrole1", "myrole2" })]
 public class GraphQLAuthorizeAttribute : GraphQLAttribute
```

src/GraphQL/Authorization/AuthorizationExtensions.cs
Shane32: Well typically you either authorize with a single policy, or a number of allowable roles.

I have no problem changing `AuthorizeWithRoles()` to use either params, or an overload for both params and comma-separated.  But I feel that GraphQLAuthorizeAttribute should match the ASP.Net Core attribute.  And since it does, I would leave the comma-separated overload.

So I'll quickly add overloads for params to `AuthorizeWithRoles()`.  There should be no need to add `AuthorizeWithPolicies(params string[])` -- this would be an edge case and the user can simply call `AuthorizeWithPolicy` twice if necessary.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
Shane32: Ok

sungam3r:
> Finally, as it is used in the GraphQL Server project, checking for roles is a much faster operation (probably zero-allocation), as the claim principal's IsInRole method can be easily called rather than having to call the asynchronous policy checker method and then using GetAwaiter().GetResult() to validate the policy.

Is it possible to say that roles authorization is a simplified case of policy-based authorization? As I understand policy-based authorization may check anything it wants and a scope of role-based authorization is much focused i.e. checking the presence of required roles (strings) in principal object.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
sungam3r: Well, all extension methods need to be obsoleted as well, I think.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
sungam3r: OK. So someone has already thought about such a signature and we will use it.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
```diff
@@ -19,18 +26,52 @@ public GraphQLAuthorizeAttribute(string policy)
         /// <summary>
         /// The name of policy to apply.
         /// </summary>
-        public string Policy { get; }
+        public string? Policy { get; set; }
```
sungam3r, line 29:
```suggestion
        public string? Policy { get; init; }
```

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
```diff
@@ -19,18 +26,52 @@ public GraphQLAuthorizeAttribute(string policy)
...
+
+        /// <summary>
+        /// A comma-separated list of the roles to apply.
+        /// Role names will be trimmed before adding.
+        /// </summary>
+        public string? Roles { get; set; }
```
sungam3r, line 35:
```suggestion
        public string? Roles { get; init; }
```

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
sungam3r: > I want to align this API with the attribute that most ASP.NET Core developers are used to

It's worth to add something like that into xml comments.

docs2/site/docs/migrations/migration5.md
```diff
@@ -1109,3 +1115,7 @@ GraphQL.NET v5 it is enabled by default as part of the `DocumentValidator.CoreRu
 ### 46. `ValidationContext.GetRecursiveVariables` returns null instead of an empty list
 
 This was done for the purposes of the overall strategy for reducing memory consumption.
+
+### 47. `AuthorizeWith` renamed to `AuthorizeWithPolicy`
+
+This change was made to clarify and differentiate between `AuthorizeWithRoles`
```
Shane32, line 1121:
```suggestion
This change was made to clarify and differentiate between `AuthorizeWithRoles`.
```

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -0,0 +1,125 @@
+using GraphQL.Types;
+
+namespace GraphQL.Tests;
```
sungam3r, line 3: Nice tests. I wonder how you have enough patience to write all that.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
sungam3r: >  So it is something people are likely used to, if they do any web programming in C#

This is a good argument.

Shane32: Added more notes and examples in the migration doc.

src/GraphQL/Authorization/AuthorizationExtensions.cs
sungam3r: > typically you either authorize with a single policy, or a number of allowable roles.

I think so too. Worth to mention that in xml comments.

> I feel that GraphQLAuthorizeAttribute should match the ASP.Net Core attribute

OK, I'm talking here more about extension methods for graphtype first approach.

>  There should be no need to add AuthorizeWithPolicies(params string[])

OK.

Shane32:
> > Finally, as it is used in the GraphQL Server project, checking for roles is a much faster operation (probably zero-allocation), as the claim principal's IsInRole method can be easily called rather than having to call the asynchronous policy checker method and then using GetAwaiter().GetResult() to validate the policy.
> 
> Is it possible to say that roles authorization is a simplified case of policy-based authorization? As I understand policy-based authorization may check anything it wants and a scope of role-based authorization is much focused i.e. checking the presence of required roles (strings) in principal object.

Basically, yes.  Typically a policy would just require certain roles, but can contain more complex behavior.  But when you're coding small projects, you usually just want to write something like `[Authorize(Roles = "Administrators")]` and not have to think about setting up policies at all.  Of course for large-scale deployments, policies allow remapping to new/other roles without reviewing all the code previously set up.

src/GraphQL.Tests/AuthorizationTests.cs
Shane32: Copy and paste!!

docs2/site/docs/migrations/migration5.md
```diff
@@ -602,6 +602,25 @@ There are a number of other minor issues fixed; see these links for more details
 - https://github.com/graphql-dotnet/graphql-dotnet/issues/3002
 - https://github.com/graphql-dotnet/graphql-dotnet/pull/3004
 
+### 20. `AuthorizeWithRoles` added in GraphQL 5.1.0
+
+This allows for specifying roles rather than just policies that can be used to validate a request.
+As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), it requires support by a third-party
+library to perform the validation.
+
+Similar to the ASP.Net Core `AuthorizeAttribute`, the new `AuthorizeWithRoles` method accepts
```
sungam3r, line 611:
```suggestion
Similar to the ASP.NET Core `AuthorizeAttribute`, the new `AuthorizeWithRoles` method accepts
```

Shane32:
I hope this looks good now.  I would like to release 5.1.0 that has this and the update for the ValidationError constructor.

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
Shane32: > Well, all extension methods need to be obsoleted as well, I think.

I'm not sure I follow.  I obsoleted all the AuthorizeWith methods.... didn't I?

sungam3r:
ValidationError constructor ?

Shane32:
> ValidationError constructor ?

- https://github.com/graphql-dotnet/graphql-dotnet/pull/3051

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
```diff
@@ -4,10 +4,17 @@
 namespace GraphQL
 {
     /// <summary>
-    /// Attribute to apply authorization policy when using schema first approach.
+    /// Attribute to apply an authorization policy and/or roles to a graph or field.
```
sungam3r, line 7:
```suggestion
    /// Attribute to apply an authorization policy and/or roles to a graph or field.
    /// This attribute mimics AuthorizeAttribute from ASP.NET Core so it is something
    /// people are likely used to, if they do any web programming in C#.
```

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -63,6 +80,76 @@ public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataP
             return provider;
         }
 
+        /// <inheritdoc cref="AuthorizeWithPolicy{TMetadataProvider}(TMetadataProvider, string)"/>
+        [Obsolete("Please use AuthorizeWithPolicy instead.")]
```
sungam3r, line 84:
```suggestion
        [Obsolete("Please use AuthorizeWithPolicy instead. Will be removed in v6.")]
```

I suggest to always write when obsoleted member will be removed.

Shane32: Agree

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -63,6 +80,76 @@ public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataP
...
+            where TMetadataProvider : IProvideMetadata
+        {
+            if (roles == null)
+                throw new ArgumentNullException(nameof(roles));
+
+            var list = GetRoles(provider) ?? new List<string>();
+
+            foreach (var role in roles.Split(','))
+            {
+                var role2 = role.Trim();
```
sungam3r, line 113: check for empty string?

src/GraphQL/Authorization/GraphQLAuthorizeAttribute.cs
sungam3r: Yes, you did that.

src/GraphQL/Authorization/AuthorizationExtensions.cs
Shane32: Done

src/GraphQL/Authorization/AuthorizationExtensions.cs
Shane32: Done

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -63,6 +80,76 @@ public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataP
...
+        /// <typeparam name="TMetadataProvider"> The type of metadata provider. Generics are used here to
+        /// let compiler infer the returning type to allow methods chaining.
+        /// </typeparam>
+        /// <param name="provider">
+        /// Metadata provider. This can be an instance of <see cref="GraphType"/>,
+        /// <see cref="FieldType"/>, <see cref="Schema"/> or others.
+        /// </param>
+        /// <param name="roles"> List of authorization role name(s). </param>
+        /// <returns> The reference to the specified <paramref name="provider"/>. </returns>
+        public static TMetadataProvider AuthorizeWithRoles<TMetadataProvider>(this TMetadataProvider provider, params string[] roles)
```
sungam3r, line 135: I think it is worth to add calls without `params` arg in tests to verify no-op for all such methods like `field.AuthorizeWithRoles()`

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -63,6 +80,76 @@ public static TMetadataProvider AuthorizeWith<TMetadataProvider>(this TMetadataP
...
+            where TMetadataProvider : IProvideMetadata
+        {
+            if (roles == null)
+                throw new ArgumentNullException(nameof(roles));
+
+            var list = GetRoles(provider) ?? new List<string>();
+
+            foreach (var role in roles)
+            {
+                if (!list.Contains(role))
```
sungam3r, line 145:
```suggestion
                if (role != "" && !list.Contains(role))
```

Shane32: whoops. thanks

src/GraphQL/Authorization/AuthorizationExtensions.cs
Shane32: I'm not sure I follow.  The tests run both overloads of AuthorizeWithRoles -- all 6 new methods.  They also run the old policy methods and the obsolete method.  What did I miss?

Shane32: Oh I think I know what you mean.

Shane32: Done

sungam3r (approved):

Shane32:
@sungam3r I'd like to release GraphQL v5.1.0 now

PR #3070, opened 2022-04-08
Add ConfigureAwait(false) to tests

Shane32:
Twice today different tests have failed for no apparent reason.  One seems to be a deadlock, presumably due to the synchronization context required for xUnit.  Adding `.ConfigureAwait(false)` should solve the problem.


Shane32:
https://github.com/graphql-dotnet/graphql-dotnet/runs/5889809239?check_suite_focus=true

https://github.com/graphql-dotnet/graphql-dotnet/runs/5889400224?check_suite_focus=true

PR #3073, opened 2022-04-08
More analyzers

sungam3r:
Continuation of #3070


sungam3r:
Not only tests fixed, some places without `ConfigureAwait` in main code too.

Shane32:
Is this just configure await false? Lmk if there’s more to look at. Otherwise merge as you please.

sungam3r:
`ConfigureAwait` is the main goal here. Other (not many) fixes: make class static, write summary for xml comment, avoid empty catch clause.

Shane32 (approved):

PR #3077, opened 2022-04-09
Authorization update

### LINKED ISSUE #3076: There is no way to tell that authorization is required without a role or policy specified
@sungam3r I messed up a bit in the last PR.  The ASP.NET Core `[Authorize]` attribute when specified without a role or policy indicates that the user should be authenticated (not a guest).  We should have matching functionality in GraphQL.

This means that the `RequiresAuthorization` extension method should return `true` when a field is marked with `[GraphQLAuthorize]`, which it does not now.

I suggest:
- Add a new extension method to indicate that the user must be authenticated.  For example, called `RequireAuthentication` or `WithAuthorization()`.
- Modify the attribute to call the above method when no policies or roles are specified
- Modify the `RequiresAuthorization` extension method to return `true` when the above method is called.  (breaking change from yesterday; not breaking from 5.0.0)

I would like to do this promptly, while GraphQL 5 is still new.  Hopefully release 5.1.1 (or 5.2.0) tomorrow with this change.

Shane32:
Closes #3076

This also fixes the attribute so that it applies the metadata to query arguments or enum definitions also, if it is marked on those type of objects.

### DIFF
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -602,9 +602,10 @@ There are a number of other minor issues fixed; see these links for more details
 - https://github.com/graphql-dotnet/graphql-dotnet/issues/3002
 - https://github.com/graphql-dotnet/graphql-dotnet/pull/3004
 
-### 20. `AuthorizeWithRoles` added in GraphQL 5.1.0
+### 20. `Authorize` and `AuthorizeWithRoles` added in GraphQL 5.1.0
 
 This allows for specifying roles rather than just policies that can be used to validate a request.
+`Authorize` can be used to specify that authentication is required, without specifying any specific roles or policies.
 As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), it requires support by a third-party
 library to perform the validation.
 
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -605,7 +605,7 @@ There are a number of other minor issues fixed; see these links for more details
 ### 20. `Authorize` and `AuthorizeWithRoles` added in GraphQL 5.1.0
 
 This allows for specifying roles rather than just policies that can be used to validate a request.
-`Authorize` can be used to specify that authentication is required, without specifying any specific roles or policies.
+`Authorize` can be used to specify that only authentication is required, without specifying any specific roles or policies.
 As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), it requires support by a third-party
 library to perform the validation.
 
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -602,7 +602,7 @@ There are a number of other minor issues fixed; see these links for more details
 - https://github.com/graphql-dotnet/graphql-dotnet/issues/3002
 - https://github.com/graphql-dotnet/graphql-dotnet/pull/3004
 
-### 20. `Authorize` and `AuthorizeWithRoles` added in GraphQL 5.1.0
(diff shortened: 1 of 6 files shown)


docs2/site/docs/migrations/migration5.md
```diff
@@ -602,9 +602,10 @@ There are a number of other minor issues fixed; see these links for more details
 - https://github.com/graphql-dotnet/graphql-dotnet/issues/3002
 - https://github.com/graphql-dotnet/graphql-dotnet/pull/3004
 
-### 20. `AuthorizeWithRoles` added in GraphQL 5.1.0
+### 20. `Authorize` and `AuthorizeWithRoles` added in GraphQL 5.1.0
```
sungam3r, line 605:
```suggestion
### 20. `Authorize` and `AuthorizeWithRoles` extension methods added in GraphQL 5.1.0
```

docs2/site/docs/migrations/migration5.md
```diff
@@ -602,9 +602,10 @@ There are a number of other minor issues fixed; see these links for more details
...
 
 This allows for specifying roles rather than just policies that can be used to validate a request.
+`Authorize` can be used to specify that authentication is required, without specifying any specific roles or policies.
```
sungam3r, line 608:
```suggestion
`Authorize` can be used to specify that only authentication is required, without specifying any specific roles or policies.
```

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -20,6 +20,12 @@ public static class AuthorizationExtensions
         /// </summary>
         public const string ROLE_KEY = "Authorization__Roles";
 
+        /// <summary>
+        /// Metadata key name for indicating that the user must be authorized to access the resource.
```
sungam3r, line 24:
```suggestion
        /// Metadata key name for indicating that the user must be authorized (strictly speaking, authenticated) to access the resource.
```

sungam3r (approved):
Only cosmetic changes suggestions.

Shane32:
5.1.1 or 5.2.0 ?

docs2/site/docs/migrations/migration5.md
```diff
@@ -602,9 +602,10 @@ There are a number of other minor issues fixed; see these links for more details
 - https://github.com/graphql-dotnet/graphql-dotnet/issues/3002
 - https://github.com/graphql-dotnet/graphql-dotnet/pull/3004
 
-### 20. `AuthorizeWithRoles` added in GraphQL 5.1.0
+### 20. `Authorize` and `AuthorizeWithRoles` extension methods added in GraphQL 5.1.0
```
Shane32, line 605:
```suggestion
### 20. `Authorize` and `AuthorizeWithRoles` extension methods added in GraphQL 5.1.1
```
?

sungam3r: Write both.

sungam3r:
5.1.1 since new APIs may be considered as a way to fix a bug

docs2/site/docs/migrations/migration5.md
Shane32:
```suggestion
### 20. `Authorize` and `AuthorizeWithRoles` extension methods added in GraphQL 5.1.0 and 5.1.1
```

Shane32: Like this?

Shane32:
> 5.1.1 since new APIs may be considered as a way to fix a bug

I agree

docs2/site/docs/migrations/migration5.md
sungam3r: yes

PR #3079, opened 2022-04-10
Add AllowAnonymous extension method and attribute

### LINKED ISSUE #3078: [Feature] Add AllowAnonymous metadata
To finish adding authorization metadata to align with ASP.NET Core, I suggest adding the equivalent of the `[AllowAnonymous]` attribute to allow this behavior:

- When a field is marked to allow anonymous access, and the field is the only one in the selection set of a parent protected graph type, then access to the graph will be allowed.

Example schema:

```gql
type Query {
  hero: Person!
}

type Person {     # [Authorize("MyPolicy")]
  id: ID!         # [AllowAnonymous]
  name: String!
}
```

Now requests like this fail validation for an anonymous use:

```gql
{
  hero {
    id
    name
  }
}
```

But this passes:

```gql
{
  hero {
    id
  }
}
```

I suggest making this functionality only apply "one level deep".

For example, for this schema:

```gql
type Query {
  hero: Person!   # [Authorize]
}

type Person {
  id: ID!         # [AllowAnonymous]
  name: String!
}
```

Then this request would still fail, because AllowAnonymous only relaxes authentication definitions on Person, not on Query.

```gql
{
  hero {
    id
  }
}
```

This feature will allow these scenarios to work:

1. Mark the query type as requiring authentication, but allow anonymous access to `__type` and `__schema`.
2. Mark the query type as requiring authentication, but allow anonymous access to a 'login' graph for authentication services.

This should help to solve:
- https://github.com/graphql-dotnet/server/issues/549

My specific suggestions are:
- Add `.AllowAnonymous()` extension method which sets a `Authorization__AllowAnonymous` metadata key to `true`
- Add `.AllowingAnonymous()` extension method which returns a boolean from the key
- Add `AllowAnonymousAttribute` for fields only which calls `AllowAnonymous` on the field

Shane32:
Closes #3078 

For 5.1.1

### DIFF
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -606,7 +606,7 @@ There are a number of other minor issues fixed; see these links for more details
 
 `AuthorizeWithRoles` allows for specifying roles rather than just policies that can be used to validate a request.
 `Authorize` can be used to specify that only authentication is required, without specifying any specific roles or policies.
-`AllowAnonymous` indicates that anonymous access should be allowed to a field of a graph type requiring authorization,
+`AllowAnonymous` typically indicates that anonymous access should be allowed to a field of a graph type requiring authorization,
 providing that no other fields were selected. As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), these
 new methods require support by a third-party library to perform the validation.
 
--- a/docs2/site/docs/migrations/migration5.md
+++ b/docs2/site/docs/migrations/migration5.md
@@ -602,12 +602,13 @@ There are a number of other minor issues fixed; see these links for more details
 - https://github.com/graphql-dotnet/graphql-dotnet/issues/3002
 - https://github.com/graphql-dotnet/graphql-dotnet/pull/3004
 
-### 20. `Authorize` and `AuthorizeWithRoles` extension methods added in GraphQL 5.1.0 and 5.1.1
+### 20. `Authorize`, `AuthorizeWithRoles` and `AllowAnonymous` extension methods added in GraphQL 5.1.0 and 5.1.1
 
-This allows for specifying roles rather than just policies that can be used to validate a request.
+`AuthorizeWithRoles` allows for specifying roles rather than just policies that can be used to validate a request.
 `Authorize` can be used to specify that only authentication is required, without specifying any specific roles or policies.
-As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), it requires support by a third-party
-library to perform the validation.
+`AllowAnonymous` indicates that anonymous access should be allowed to a field of a graph type requiring authorization,
+providing that no other fields were selected. As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), these
+new methods require support by a third-party library to perform the validation.
 
 Similar to the ASP.NET Core `AuthorizeAttribute`, the new `AuthorizeWithRoles` method accepts
 a comma-separated list of role names that would allow access to the graph or field.
@@ -621,6 +622,9 @@ You may also supply a list of strings as in the following example:
 ```csharp
 graph.AuthorizeWithRoles("Administrators", "Managers");
 ```
+
+For schema-first and "type-first" graphs, the `[GraphQLAuthorize]` has been updated to support roles and can now
+be used without any policy or role names, and an `[AllowAnonymous]` attribute has been added.
 
 ### 21. `RequestServices` added to `ValidationContext` in GraphQL 5.1.0
 
--- a/src/GraphQL.Tests/AuthorizationTests.cs
+++ b/src/GraphQL.Tests/AuthorizationTests.cs
@@ -8,9 +8,9 @@ public class AuthorizationTests
     public void Field()
     {
         var field = new FieldType();
-        field.RequiresAuthorization().ShouldBeFalse();
+        field.IsAuthorizationRequired().ShouldBeFalse();
         field.AuthorizeWith("Policy1");
-        field.RequiresAuthorization().ShouldBeTrue();
+        field.IsAuthorizationRequired().ShouldBeTrue();
         field.AuthorizeWith("Policy2");
         field.AuthorizeWith("Policy2");
         field.AuthorizeWithPolicy("Policy3");
@@ -20,7 +20,7 @@ public class AuthorizationTests
         field.AuthorizeWithRoles("Role1", "Role4");
         field.AuthorizeWithRoles("");
 
-        field.RequiresAuthorization().ShouldBeTrue();
+        field.IsAuthorizationRequired().ShouldBeTrue();
         field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
         field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3", "Role4" });
     }
@@ -33,25 +33,25 @@ public class AuthorizationTests
         field.AuthorizeWithRoles("");
         field.AuthorizeWithRoles(" ");
         field.AuthorizeWithRoles(",");
-        field.RequiresAuthorization().ShouldBeTrue();
+        field.IsAuthorizationRequired().ShouldBeTrue();
     }
 
     [Fact]
     public void AllowAnonymous()
     {
         var field = new FieldType();
-        field.AllowsAnonymous().ShouldBeFalse();
+        field.IsAnonymousAllowed().ShouldBeFalse();
         field.AllowAnonymous();
-        field.AllowsAnonymous().ShouldBeTrue();
+        field.IsAnonymousAllowed().ShouldBeTrue();
     }
 
     [Fact]
     public void Authorize()
     {
         var field = new FieldType();
-        field.RequiresAuthorization().ShouldBeFalse();
+        field.IsAuthorizationRequired().ShouldBeFalse();
         field.Authorize();
-        field.RequiresAuthorization().ShouldBeTrue();
+        field.IsAuthorizationRequired().ShouldBeTrue();
     }
 
     [Fact]
@@ -69,7 +69,7 @@ public class AuthorizationTests
             .AuthorizeWithRoles("Role1", "Role4");
 
         var field = graph.Fields.Find("Field");
-        field.RequiresAuthorization().ShouldBeTrue();
+        field.IsAuthorizationRequired().ShouldBeTrue();
         field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
         field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3", "Role4" });
     }
@@ -90,7 +90,7 @@ public class AuthorizationTests
             .AuthorizeWithRoles("Role1", "Role4");
 
         var field = graph.Fields.Find("Field");
-        field.RequiresAuthorization().ShouldBeTrue();
+        field.IsAuthorizationRequired().ShouldBeTrue();
         field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
         field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3", "Role4" });
     }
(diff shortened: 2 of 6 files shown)


docs2/site/docs/migrations/migration5.md
```diff
@@ -602,12 +602,13 @@ There are a number of other minor issues fixed; see these links for more details
...
 
-### 20. `Authorize` and `AuthorizeWithRoles` extension methods added in GraphQL 5.1.0 and 5.1.1
+### 20. `Authorize`, `AuthorizeWithRoles` and `AllowAnonymous` extension methods added in GraphQL 5.1.0 and 5.1.1
 
-This allows for specifying roles rather than just policies that can be used to validate a request.
+`AuthorizeWithRoles` allows for specifying roles rather than just policies that can be used to validate a request.
 `Authorize` can be used to specify that only authentication is required, without specifying any specific roles or policies.
-As with `AuthorizeWithPolicy` (renamed from `AuthorizeWith`), it requires support by a third-party
-library to perform the validation.
+`AllowAnonymous` indicates that anonymous access should be allowed to a field of a graph type requiring authorization,
```
sungam3r, line 609: I suggest not to narrow meaning of allowanonymous only to fields like for actions in ASP.NET controllers. It's GraphQL world.

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -151,5 +175,7 @@ private class Class1
         public string Name { get; set; }
         [GraphQLAuthorize]
         public string Value { get; set; }
+        [AllowAnonymous]
+        public string Test { get; set; }
```
sungam3r, line 179:
```suggestion
        public string Public { get; set; }
```

src/GraphQL/Authorization/AllowAnonymousAttribute.cs
```diff
@@ -0,0 +1,25 @@
+using GraphQL.Types;
+using GraphQL.Utilities;
+
+namespace GraphQL
```
sungam3r, line 4: Let's use file-scoped syntax for all new code.

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -48,6 +54,23 @@ public static class AuthorizationExtensions
         /// <returns> List of authorization role names applied to this metadata provider. </returns>
         public static List<string>? GetRoles(this IProvideMetadata provider) => provider.GetMetadata<List<string>>(ROLE_KEY);
 
+        /// <summary>
+        /// Returns a boolean indicating if anonymous access should be allowed to a field of a graph type
+        /// requiring authorization, providing that no other fields were selected.
+        /// </summary>
+        public static bool AllowingAnonymous(this IProvideMetadata provider) => provider.GetMetadata(ANONYMOUS_KEY, false);
```
sungam3r, line 61:
```suggestion
        public static bool AllowsAnonymous(this IProvideMetadata provider) => provider.GetMetadata(ANONYMOUS_KEY, false);
```

docs2/site/docs/migrations/migration5.md
Shane32: In this case, I'm not really thinking about how it relates to ASP.NET (apart from the naming and general concept), but rather the practical usage of this attribute in the GraphQL world.  I'm just not sure how AllowAnonymous would be useful to apply to a graph, since graphs are generally abstract from other graphs.  For instance, if Product is marked as AllowAnonymous, and if MasterProductsTable is protected, querying for all products within MasterProductsTable should still be refused.  The same thought process applies to input object graphs and scalars, leaving only fields at this time.

Within ASP.NET Core, only one action can ever execute, so AllowAnonymous simply overrides all Authorize attributes that are parent to it in the chain.  In this respect, my suggestion is unlike ASP.NET Core.  It's simply what works best for GraphQL.

I can remove the attribute usage restriction, but even if we did so, I figured that the xml comments should explain anticipated usage of the attributes, even knowing that implementation is not part of this library.  If nothing else, use the word "typically" and retain the wording/example.

Do you suggest I remove the attribute usage restrictions?  Do you have a suggestion of rewording here?

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -114,23 +130,31 @@ public void SchemaBuilder()
...
         field.RequiresAuthorization().ShouldBeTrue();
+        field.AllowingAnonymous().ShouldBeFalse();
         field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
         field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
 
         field = graph.Fields.Find("value");
         field.RequiresAuthorization().ShouldBeTrue();
+        field.AllowingAnonymous().ShouldBeFalse();
+
+        field = graph.Fields.Find("test");
```
Shane32, line 155:
```suggestion
        field = graph.Fields.Find("public");
```

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -114,23 +130,31 @@ public void SchemaBuilder()
   id: String!
   name: String!
   value: String!
+  test: String!
```
Shane32, line 133:
```suggestion
  public: String!
```

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -91,18 +100,25 @@ public void AutoOutputGraphType()
...
         field.RequiresAuthorization().ShouldBeTrue();
+        field.AllowingAnonymous().ShouldBeFalse();
         field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
         field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
 
         field = graph.Fields.Find("Value");
         field.RequiresAuthorization().ShouldBeTrue();
+        field.AllowingAnonymous().ShouldBeFalse();
+
+        field = graph.Fields.Find("Test");
```
Shane32, line 119:
```suggestion
        field = graph.Fields.Find("Public");
```

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -114,23 +130,31 @@ public void SchemaBuilder()
...
 
         var field = graph.Fields.Find("name");
         field.RequiresAuthorization().ShouldBeTrue();
+        field.AllowingAnonymous().ShouldBeFalse();
         field.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
         field.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
 
         field = graph.Fields.Find("value");
         field.RequiresAuthorization().ShouldBeTrue();
+        field.AllowingAnonymous().ShouldBeFalse();
```
Shane32, line 153:
```suggestion
        field.AllowsAnonymous().ShouldBeFalse();
```

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -114,23 +130,31 @@ public void SchemaBuilder()
...
         graph.RequiresAuthorization().ShouldBeTrue();
+        graph.AllowingAnonymous().ShouldBeFalse();
         graph.GetPolicies().ShouldBe(new string[] { "Policy1", "Policy2", "Policy3" });
         graph.GetRoles().ShouldBe(new string[] { "Role1", "Role2", "Role3" });
 
         graph.Fields.Find("id").RequiresAuthorization().ShouldBeFalse();
 
         var field = graph.Fields.Find("name");
         field.RequiresAuthorization().ShouldBeTrue();
+        field.AllowingAnonymous().ShouldBeFalse();
```
Shane32, line 147:
```suggestion
        field.AllowsAnonymous().ShouldBeFalse();
```

src/GraphQL.Tests/AuthorizationTests.cs
```diff
@@ -114,23 +130,31 @@ public void SchemaBuilder()
   id: String!
   name: String!
   value: String!
+  public: String!
 }",
             configure => configure.Types.Include<Class1>());
 
         var graph = (ObjectGraphType)schema.AllTypes["Class1"];
         graph.RequiresAuthorization().ShouldBeTrue();
+        graph.AllowingAnonymous().ShouldBeFalse();
```
Shane32, line 139:
```suggestion
        graph.AllowsAnonymous().ShouldBeFalse();
```

Shane32:
@sungam3r I'm working on an authorization implementation that works with all the PRs over the last couple days.  One question that is interesting to consider is: _Should the introspection field types be marked with `.AllowAnonymous()`?_  I think the answer is 'no', but rather the validation rule requires some special logic, and here's why:

Consider these queries:

```gql
# this query requires authentication because ProtectedType requires authentication
{
  protectedType {   # the type, not the field here, is marked as needing authentication
    unmarkedField
    anonymousField
  }
}

# this query does not require authentication becauase although ProtectedType requires
# authentication, all of the selected fields are marked with AllowAnonymous
{
  protectedType {
    anonymousField
  }
}

# this query also does not require authentication
{
  protectedType {
    __typeName
    anonymousField
  }
}

# this query SHOULD require authentication, because it could unintentionally reveal information to the caller,
# if for example the schema designer did not mark any fields with AllowAnonymous
{
  protectedType {
    __typeName
  }
}
```

So as you can see, in my implementation `__typeName` will only be treated as `AllowAnonymous` when all other selected fields are also marked with AllowAnonymous.  But if `__typeName` is the only selected field, then the authentication requirements of the type will take effect.

Now for `__type` and `__schema`, they can safely be marked with `.AllowAnonymous()` (or treated as such), because they only exist on the root Query type; if no other fields are requested, no secure information is returned.  I suggest we do so, and then users can override such behavior in their schema constructor by modifying the metadata.  But even that should not be required; users can just mark the schema itself with `.Authorize()` and then any unauthenticated request will fail, regardless of whether or not only introspection fields were requested.  On the other hand, since we are not marking `__typeName`, we can skip marking `__type` and `__schema`, leaving it up to the validation rule to treat them as `.AllowAnonymous()` (or as desired by the authentication implementation).

Shane32:
p.s. Dealing with fragments and the `AllowAnonymous()` concept is a real pain.  Everything else is pretty straightforward to implement.

docs2/site/docs/migrations/migration5.md
sungam3r: > implementation is not part of this library

This is true and we must leave some space for maneuver.

> word "typically" and retain the wording/example

Yes, let's insert _typically_ somewhere. The general meaning will be the same, but the phrase will sound more accurate.

Shane32:
```suggestion
`AllowAnonymous` typically indicates that anonymous access should be allowed to a field of a graph type requiring authorization,
```

Shane32: See above (merged).

src/GraphQL/Authorization/AllowAnonymousAttribute.cs
```diff
@@ -0,0 +1,24 @@
+using GraphQL.Types;
+using GraphQL.Utilities;
+
+namespace GraphQL;
+
+/// <summary>
+/// Attribute to indicate that anonymous access should be allowed to a field of a graph type
+/// requiring authorization, providing that no other fields were selected.
```
Shane32, line 8:
```suggestion
/// Attribute to typically indicate that anonymous access should be allowed to a field of a graph type
/// requiring authorization, providing that no other fields were selected.
```

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -26,6 +26,12 @@ public static class AuthorizationExtensions
         /// </summary>
         public const string AUTHORIZE_KEY = "Authorization__Required";
 
+        /// <summary>
+        /// Metadata key name for indicating if anonymous access should be allowed to a field of a graph type
+        /// requiring authorization, providing that no other fields were selected.
```
Shane32, line 31:
```suggestion
        /// Metadata key name for typically indicating if anonymous access should be allowed to a field of a graph type
        /// requiring authorization, providing that no other fields were selected.
```

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -48,6 +54,23 @@ public static class AuthorizationExtensions
         /// <returns> List of authorization role names applied to this metadata provider. </returns>
         public static List<string>? GetRoles(this IProvideMetadata provider) => provider.GetMetadata<List<string>>(ROLE_KEY);
 
+        /// <summary>
+        /// Returns a boolean indicating if anonymous access should be allowed to a field of a graph type
+        /// requiring authorization, providing that no other fields were selected.
```
Shane32, line 59:
```suggestion
        /// Returns a boolean typically indicating if anonymous access should be allowed to a field of a graph type
        /// requiring authorization, providing that no other fields were selected.
```

src/GraphQL/Authorization/AuthorizationExtensions.cs
```diff
@@ -48,6 +54,23 @@ public static class AuthorizationExtensions
...
+        /// </summary>
+        public static bool IsAnonymousAllowed(this IProvideMetadata provider) => provider.GetMetadata(ANONYMOUS_KEY, false);
+
+        /// <summary>
+        /// Adds metadata to indicate that anonymous access should be allowed to a field of a graph type
+        /// requiring authorization, providing that no other fields were selected.
```
Shane32, line 65:
```suggestion
        /// Adds metadata to typically indicate that anonymous access should be allowed to a field of a graph type
        /// requiring authorization, providing that no other fields were selected.
```

sungam3r:
>  It would help a TON if the document validator sorted the document so dependent fragments were processed prior to the fragments or query that needed those fragments.

:) I was biten by that when doing complexity analizer.

Regarding all other notes. Yes, auth validation is a pain. There are so much questions to think about. I have already thought about them, because in my work we use an extensive system for checking auth directives. I came to the conclusion that there is no ideal solution. Implementation can pursue different purposes.

Also remember of https://github.com/graphql-dotnet/authorization/pull/186 (abandoned).

src/GraphQL/Authorization/AllowAnonymousAttribute.cs
```diff
@@ -0,0 +1,24 @@
+using GraphQL.Types;
+using GraphQL.Utilities;
+
+namespace GraphQL;
+
+/// <summary>
+/// Attribute to typically indicate that anonymous access should be allowed to a field of a graph type
+/// requiring authorization, providing that no other fields were selected.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Method | AttributeTargets.Field)]
```
sungam3r, line 10: Someday in the future this list will expand. I came across some unusual cases when the non-standard logic of processing input parameters was required.

sungam3r (approved):

src/GraphQL/Authorization/AllowAnonymousAttribute.cs
Shane32: No problem; it will be a non-breaking change at that time.

Shane32:
@sungam3r I am going to finish my auth code tonight, add a million tests and all that stuff, and make sure it is operating as I am anticipating that it should.  I know you're wrapping up for the night, but by your morning, you'll be able to see my progress in https://github.com/Shane32/GraphQL.AspNetCore3/pull/6 .  (Don't look now, it's not nearly finished.)  I will plan to release 5.1.1 in the morning; just reply if you have any comments or objections between now and then.

Shane32:
If you would like to help review my implementation, particularly the public API, let me know.  I know you are busy and also support a lot of open-source projects, so if you don't have time, that's fine too.

Shane32:
@sungam3r I'm going to release 5.1.1 now.

sungam3r:
OK. I opened your PR. Now the main thing is not to lose it among the other 150 open tabs.

Issue #3093, opened 2022-04-21
IndexOutOfRangeException with more than 16 named fragment spread


cotillion:
## Description
If a query is executed with more than 16 named fragments and atleast 1 is not spread at the top level System.IndexOutOfRangeException is thrown.

GetFragmentSpreads() does not take into account the possibility of recursion in CollectFields

### Steps to reproduce

```

query {
  hero {
    ...first
    ...second
    ...third
    ...fourth
    ...fifth
    ...sixth
    ...seventh
    ...eight
    ...ninth
    ...tenth
    ...eleventh
    ...twelfth
    ...thirteenth
    ...fourteenth
    ...fifteenth
    ...sixteenth           
  }
}


fragment first on Character {
  id
}

fragment second on Character {
  id
}

fragment third on Character {
  id
}

fragment fourth on Character {
  id
}

fragment fifth on Character {
  id
}

fragment sixth on Character {
  id
}

fragment seventh on Character {
  id
}

fragment eight on Character {
  id
}

fragment ninth on Character {
  id
}

fragment tenth on Character {
  id    
}

fragment eleventh on Character {
  id  
}
fragment twelfth on Character {
  id  
}
fragment thirteenth on Character {
  id  
}

fragment fourteenth on Character {
  id  
}

fragment fifteenth on Character {
  id  
}

fragment sixteenth on Character {
  id  
  ...boom
}

fragment boom on Character {
  id
}

```

### Expected result

Result

### Actual result

Boom.

### Environment

v5

Shane32:
Note: the reason it is 16 is because even though the number is miscounted for any nested fragment, the array used is rented.  The MS array pool probably always allocates arrays with at least 16 elements minimum in it.

sungam3r:
https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/shared/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs#L31

PR #3094, opened 2022-04-21
Fix counting of nested named fragments

cotillion:
Fixes graphql-dotnet/graphql-dotnet#3093

It's possible there is a better way to do this but it seems to work. There might be some performance impact..

### DIFF
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -222,7 +222,7 @@ namespace GraphQL.Execution
             fields ??= new();
 
             // optimization for majority of cases: 0 or 1 fragment spread in selection set, so nothing to track
-            int countOfSpreads = GetFragmentSpreads(selectionSet);
+            int countOfSpreads = GetFragmentSpreads(context, selectionSet);
             ROM[]? visitedFragmentNames = null;
             if (countOfSpreads > 1)
                 visitedFragmentNames = ArrayPool<ROM>.Shared.Rent(countOfSpreads);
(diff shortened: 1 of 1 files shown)


sungam3r:
> with the revised code, a recursive query will cause a stack overflow exception.

`NoFragmentCycles` validation rule handles this case. Of course, one can turn it off 😄 .

> As for performance, so long as there are no fragment references, there is no performance impact. And if there is, it is important that the fragments be counted properly.

I agree. One more thing I want to say - it is strange that such an obvious error remained unnoticed. In other places recursion is used when dealing with fragments, and here we just forgot.

> if (count > 500) throw new InvalidOperationException("Too many nested fragments");

I would not. As I said, in other places recursion is also used. Validation rules are intended to catch broken requests. Adding such checks here and there in code base will only complicate and pollute the code.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -316,14 +316,27 @@ static void Add(Dictionary<string, (GraphQLField Field, FieldType FieldType)> fi
...
+                    if (selection is GraphQLFragmentSpread spread)
+                    {
                         ++count;
+
+                        var fragment = context.Document.FindFragmentDefinition(spread.FragmentName.Name);
+                        if (fragment != null)
+                        {
+                            count += GetFragmentSpreads(context, fragment.SelectionSet);
+                        }
+
```
sungam3r, line 334:
```suggestion
```

sungam3r (approved):

sungam3r:
@Shane32 merge if OK.

Shane32 (approved):

Shane32:
Thanks for the contribution @cotillion !!

PR #3098, opened 2022-04-24
Update ToInputs sample in migration document

Shane32:


docs2/site/docs/migrations/migration5.md
```diff
@@ -778,16 +778,16 @@ public static class StringExtensions
 {
     private static readonly GraphQLSerializer _serializer = new();
 
-    public static Inputs ToInputs(string json)
+    public static Inputs ToInputs(this string json)
```
sungam3r, line 781: Regarding line 773 - I would change cs to csharp. I know that some markdown plugins work wrong with cs. Also we use cshar across code base, I have found only 3 usages of cs.

sungam3r (approved):

docs2/site/docs/migrations/migration5.md
Shane32: You're correct; I believe our own doc pages do not work with cs; only csharp.  I always use cs while on github so I keep forgetting, especially since github works correctly for cs.

PR #3099, opened 2022-04-24
Restrict list types in GetGraphTypeFromType

Shane32:
I keep running into this problem in my code:  `GetGraphTypeFromType` takes classes that inherit or implement a list or dictionary type and assign it a graph type mapping of the `KeyValuePair` rather than the object as a whole.

For example:
```cs
private MyInfoClass : Dictionary<string, object>
{
    public string Name => this["Name"];
}

var type = GetGraphTypeFromType(typeof(MyInfoClass), true, TypeMappingMode.OutputType);
//expected:
type == typeof(GraphQLClrOutputGraphType<MyInfoClass>)
//actual:
type == typeof(ListGraphType<GraphQLClrOutputGraphType<KeyValuePair<string, object>>>);
```

My workaround is currently as follows:

1. Apply the patch in this PR to a custom version of the `GetGraphTypeFromType` method
2. Create a class derived from `TypeInformation` and override `ConstructGraphType` to use the new method
3. Create a class dervied from `AutoRegisteringObjectGraphType` and override `GetTypeInformation` to use the new class
4. Register the new class within DI to be used when `AutoRegisteringObjectGraphType` is requested.

My suggested fix is to only detect explicit enumerable types that we have already predefined in the `TypeInformation` class:

- `IEnumerable`
- `IEnumerable<>`
- `IList<>`
- `List<>`
- `ICollection<>`
- `IReadOnlyCollection<>`
- `IReadOnlyList<>`
- `HashSet<>`
- `ISet<>`

Any derived types from those types will not be detected as a list.  This is a breaking change, and if someone is relying on the prior behavior, they will need to cast their custom return type to a detected list type.  For example:

```cs
public class People : List<Person>
{
}

var type = GetGraphTypeFromType(typeof(People), true, TypeMappingMode.OutputType);
//previously:
type == typeof(ListGraphType<GraphQLClrOutputGraphType<Person>>)
//now:
type == typeof(GraphQLClrOutputGraphType<People>)

//breaking change affects uses like this:
Field(x => x.People); //where People is of type People not of type List<Person>
```

My workaround (described above) contains almost 200 lines of code to fix this problem (mostly copied from the main library).  I suggest we fix it in the main library.  However, it is a breaking change and there may be existing code that relies on the current behavior.  As much as I would like to see it in 5.1 for my own needs, I suggest this be a change for 6.0.  I can live with my existing workaround.

An alternative option would be to make this a global option.  Even though I generally discourage global options, this really seems like a pain point (for me) that we can get rid of rather easily.

### DIFF
--- a/src/GraphQL/TypeExtensions.cs
+++ b/src/GraphQL/TypeExtensions.cs
@@ -148,7 +148,7 @@ namespace GraphQL
             }
             else
             {
-#pragma warning disable CS0618 // Type or member is obsolete -- remove this block for v6
+#pragma warning disable CS0618 // Type or member is obsolete
                 var attr = type.GetCustomAttribute<GraphQLMetadataAttribute>();
                 if (attr != null)
                 {
--- a/src/GraphQL/TypeExtensions.cs
+++ b/src/GraphQL/TypeExtensions.cs
@@ -133,14 +133,21 @@ namespace GraphQL
                 var elementType = GetGraphTypeFromType(clrElementType, IsNullableType(clrElementType), mode); // isNullable from elementType, not from parent array
                 graphType = typeof(ListGraphType<>).MakeGenericType(elementType);
             }
-            else if (TryGetEnumerableElementType(type, out var clrElementType))
+#pragma warning disable CS0618 // Type or member is obsolete -- remove this block for v6
+            else if (GlobalSwitches.MapAllEnumerableTypes && IsAnIEnumerable(type))
+            {
+                var clrElementType = GetEnumerableElementType(type);
+                var elementType = GetGraphTypeFromType(clrElementType, IsNullableType(clrElementType), mode); // isNullable from elementType, not from parent container
+                graphType = typeof(ListGraphType<>).MakeGenericType(elementType);
+            }
+#pragma warning disable CS0618 // Type or member is obsolete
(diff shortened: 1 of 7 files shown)


Shane32:
Unfortunately my patch is required in multiple private libraries - ugh!

src/GraphQL/TypeExtensions.cs
```diff
@@ -132,15 +133,21 @@ public static Type GetGraphTypeFromType(this Type type, bool isNullable = false,
...
             }
-            else if (IsAnIEnumerable(type))
+#pragma warning disable CS0618 // Type or member is obsolete -- remove this block for v6
+            else if (GlobalSwitches.MapAllEnumerableTypes && IsAnIEnumerable(type))
             {
                 var clrElementType = GetEnumerableElementType(type);
                 var elementType = GetGraphTypeFromType(clrElementType, IsNullableType(clrElementType), mode); // isNullable from elementType, not from parent container
                 graphType = typeof(ListGraphType<>).MakeGenericType(elementType);
             }
+#pragma warning disable CS0618 // Type or member is obsolete
```
sungam3r, line 143: Doubled pragma warning disable / pragma warning restore

sungam3r (approved):

Shane32:
@sungam3r I added more details to the xml comment for the global switch if you want to review it.

Issue #3100, opened 2022-04-25
Can't override graph types within DI


Shane32:
When using dependency injection to replace a requested graph type with a derived type, `SchemaTypes` does not add the proper type to its internal dictionary and constructs the type multiple times (which should never happen), with the result that the `ResolvedType` property of a parent graph is a different instance than what is stored in `schema.AllTypes`.  The duplicate type is not initialized, causing various problems -- in my case, field names of the uninitialized instance were not converted to camel case, causing a query validation failure.

A sample of the code that would cause a problem is shown here:

```cs
// override AutoRegisteringObjectGraphType with MyAutoGraphType
services.AddTransient(typeof(AutoRegisteringObjectGraphType<>), typeof(MyAutoGraphType<>));

private class MyAutoGraphType<T> : AutoRegisteringObjectGraphType<T>
{
    // this class contains no code and should perform identical to the type it derives from
}
```

@sungam3r This is a critical bug for my code.  I'm not exactly sure how it worked in v4, or what I may have changed in my project during the transition to v5, but it doesn't work now.  I started modifying `SchemaTypes` to store the requested type (rather than the resolved type's `GetType()` value), but it seems to be a rather invasive change.  Do you have any ideas before I dive into this?  Again, this is critical for me, and if you have any ideas, please let me know.  Thanks!!

See #3101 for a demo of the bug

Note that I can implement workarounds in other classes to get around this failure, but it would probably take as much work as fixing the problem within GraphQL.NET.

Shane32:
> Note that I can implement workarounds in other classes to get around this failure, but it would probably take as much work as fixing the problem within GraphQL.NET.

I was able to work around the bug partially by using the new `IGraphTypeMappingProvider` to remap auto-generated graph types on the fly, but this doesn't cover explicitly specified graph types.  Fortunately there were few of these in my code base.  So I have a working solution for the time being.

sungam3r:
> Do you have any ideas

Shane32:
I would like to fix this bug, one way or another, and then release 5.1.  If you have time to look at it briefly tonight, that would be appreciated.

I'll likely try to work on a patch tonight (which is after your 'tonight').  It will be in the form of a patch to `SchemaTypes`, or possible a rework of the whole class.  (I've long intended to rewrite `SchemaTypes`, as it seems overly complex, but I have not had the motivation, which this bug amply supplies.)

sungam3r:
[...]

PR #3114, opened 2022-05-04
Fix SchemaTypes so it does not create instances twice

Shane32:
Fixes:
- #3100


Shane32:
@sungam3r `SchemaTypes` might still be a mess, but at least this bug should be fixed for now.  I also intend to release 5.2 once this PR is merged, unless you have an objection.

Shane32:
@sungam3r Any comments here?

Shane32:
@sungam3r I will be merging this PR tonight and releasing 5.2.  Please let me know if you have any comments.

sungam3r:
OK. I looked through the code and I understand the described problem slightly.

sungam3r:
Why did not you change line 429 as well ?
изображение

PR #3118, opened 2022-05-05
 Automatic Persisted Queries

SlavaUtesinov:
This PR is a preparation for implementation of [Automatic Persisted Queries](https://www.apollographql.com/docs/react/api/link/persisted-queries) (APQ) in a server project [PR](https://github.com/graphql-dotnet/server/pull/782) - graphql-dotnet/server#132. Briefly, in case of APQ, a query field can be omitted, execution of a request will be implemented by means of additional hash field (sha256Hash) which corresponds to the query which has been saved on server side in advance, The request will look this way:
```
{
  operationName: 'MyQuery',
  variables: null,
  extensions: {
    persistedQuery: {
      version: 1,
      sha256Hash: hashOfQuery
    }
  }
}
```


sungam3r:
Protocol here - https://www.apollographql.com/docs/react/api/link/persisted-queries/#protocol (Apollo JS stuff is an implementation).

src/GraphQL/Transport/GraphQLRequest.cs
```diff
@@ -12,9 +12,11 @@ public class GraphQLRequest
         public string? OperationName { get; set; }
 
         /// <summary>
-        /// A Document containing GraphQL Operations and Fragments to execute (required).
+        /// A Document containing GraphQL Operations and Fragments to execute.
+        /// It can be null in case of automatic persisted queries (https://www.apollographql.com/docs/apollo-server/performance/apq/)
+        /// when a client sends only SHA-256 hash of the query given that corresponding key-value pair has been persisted on a server beforehand.
```
sungam3r, line 17:
```suggestion
        /// when a client sends only SHA-256 hash of the query in <see cref="Extensions"/> given that corresponding key-value pair has been persisted on a server beforehand.
```

sungam3r:
@Shane32 Note that graphql-over-http spec implicitly says that query field is required
изображение
but I think that it is not a big deal to make it optional. We are going to use APQ in production pretty soon. It's a rather simple concept, requires ~20 lines of code (in server project) and it saves a lot of network/memory/CPU reources (being used with document caching). Also note that APQ != document caching implemented in GraphQL.MemoryCache. These are diffferent features that **may** be used together.

sungam3r (approved):

src/GraphQL/Transport/GraphQLRequest.cs
SlavaUtesinov: done

Shane32 (approved):

Shane32:
FYI, the persisted query implementation can easily be done with a custom `IDocumentExecuter` which essentially wraps another `IDocumentExecuter`.  Since the implementation returns a proper GraphQL response in the case that a match is not found, I'm pretty sure it does not _need_ to interact with the server project at all.  However, the server project needs to be modified to allow requests without a query specified.

Shane32:
@sungam3r I use something similar to APQ already - but only for logging purposes.  Each query is hashed and logged, allowing me to compute statistics for any query over a large period of time with very little database / storage space required.  Adding APQ would be rather trivial to implement, and would enhance performance.

In v6 of GraphQL we should consider allowing `IConfigureExecution` to 'wrap' a document executer execution, thereby being able to implement APQ, or the MemoryCache, or appending Apollo Trace statistics without having any special code within with document executer itself.  Some DI engines already allow such configuration built-in, called decorators, but MSDI does not have decorator support.

PR #3119, opened 2022-05-05
Allow execution configurations to set Schema and Query

Shane32:
Very simple change here, which allows `IConfigureExecution` instances to configure `Schema`, `Query` and `Document` properties of `ExecutionOptions`.


Shane32:
@sungam3r I will be merging this PR tonight and releasing 5.2.  Please let me know if you have any comments.

sungam3r:
Hah, simple change indeed.

PR #3123, opened 2022-05-09
connection should be non nullable

SimonCropp:
re submit of #2581


Shane32:
Whoops - this should have been merged for v5.

JTeeuwissen:
Hi again @SimonCropp, in light of #2581 I'm wondering the same here. I don't see changes to make the connection itself non nullable, as mentioned before. Or am I mistaken?

sungam3r:
@JTeeuwissen Please read discussion from #2581 around nullability of connection and other things. PR title is a bit confusing nnow.

JTeeuwissen:
> @JTeeuwissen Please read discussion from #2581 around nullability of connection and other things. PR title is a bit confusing now.

Hi, I already had. I added a comment there with the same question but that merge was closed. But I am still not sure why the fieldType of the connection is still `typeof(TConnectionType)` instead of `typeof(NonNullGraphType<TConnectionType>)`. I get the discussed nullable cursors. But I don't see a mention as to why this is not included.

PR #3126, opened 2022-05-09
Make FieldBuilder constructor public

Shane32:
I need this for some of my code.  Even though all the methods are `virtual`, since the constructor is `private`, I cannot inherit the class.

@sungam3r I will merge tonight if you do not get a chance to review before then; if you have comments we can always add another PR to address them.

Thanks!

### DIFF
--- a/src/GraphQL/Builders/ConnectionBuilder.cs
+++ b/src/GraphQL/Builders/ConnectionBuilder.cs
@@ -70,7 +70,10 @@ namespace GraphQL.Builders
         /// </summary>
         public FieldType FieldType { get; protected set; }
 
-        private ConnectionBuilder(FieldType fieldType)
+        /// <summary>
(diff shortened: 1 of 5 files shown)


src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -70,7 +70,10 @@ public class ConnectionBuilder<TSourceType>
         /// </summary>
         public FieldType FieldType { get; protected set; }
 
-        private ConnectionBuilder(FieldType fieldType)
+        /// <summary>
+        /// Initializes a new instance for the specified <see cref="Types.FieldType"/>.
```
sungam3r, line 74: Interesting comment on line 55.

sungam3r (approved):

PR #3132, opened 2022-05-11
[WIP] Update Apollo Tracing extension methods

Shane32:


src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -1029,13 +1029,52 @@ public static IGraphQLBuilder AddValidationRule<TValidationRule>(this IGraphQLBu
...
-        #region - AddMetrics -
+        #region - AddApolloTracing / AddMetrics -
+        /// <summary>
+        /// Registers <see cref="InstrumentFieldsMiddleware"/> within the dependency injection framework and
+        /// configures it to be installed within the schema, and configures responses to include Apollo
+        /// Tracing data when enabled via <see cref="ExecutionOptions.EnableMetrics"/>.
+        /// When <paramref name="enable"/> is <see langword="true"/>, configures execution to set
+        /// <see cref="ExecutionOptions.EnableMetrics"/> to <see langword="true"/>; otherwise leaves it unchanged.
+        /// </summary>
+        public static IGraphQLBuilder AddApolloTracing(this IGraphQLBuilder builder, bool enable = true)
```
sungam3r, line 1040: I'm a bit lean towards `collectMetrics` vs `enable` but I'm OK to leave `enable`.

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -1029,13 +1029,52 @@ public static IGraphQLBuilder AddValidationRule<TValidationRule>(this IGraphQLBu
...
+            => AddApolloTracing(builder, _ => enable);
+
+        /// <summary>
+        /// Registers <see cref="InstrumentFieldsMiddleware"/> within the dependency injection framework and
+        /// configures it to be installed within the schema, and configures responses to include Apollo
+        /// Tracing data when enabled via <see cref="ExecutionOptions.EnableMetrics"/>.
+        /// Configures execution to run <paramref name="enablePredicate"/> and when <see langword="true"/>, sets
+        /// <see cref="ExecutionOptions.EnableMetrics"/> to <see langword="true"/>; otherwise leaves it unchanged.
```
sungam3r, line 1048: > otherwise leaves it unchanged

🤔 Are you sure? Why not to set it to `false` if one wants to disable metrics to some subset of requests with default metrics enabled?

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -1029,13 +1029,52 @@ public static IGraphQLBuilder AddValidationRule<TValidationRule>(this IGraphQLBu
...
+        public static IGraphQLBuilder AddApolloTracing(this IGraphQLBuilder builder, Func<ExecutionOptions, bool> enablePredicate)
+        {
+            if (enablePredicate == null)
+                throw new ArgumentNullException(nameof(enablePredicate));
+
+            builder.AddMiddleware<InstrumentFieldsMiddleware>();
+            builder.ConfigureExecution(async (options, next) =>
+            {
+                if (enablePredicate(options))
+                    options.EnableMetrics = true;
```
sungam3r, line 1059:
```suggestion
                options.EnableMetrics = enablePredicate(options);
```

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -1029,13 +1029,52 @@ public static IGraphQLBuilder AddValidationRule<TValidationRule>(this IGraphQLBu
...
+                DateTime start = DateTime.UtcNow;
+                var ret = await next(options).ConfigureAwait(false);
+                if (options.EnableMetrics)
+                {
+                    ret.EnrichWithApolloTracing(start);
```
sungam3r, line 1064: I'm 70/30 to always add tracing key even if Metrics.Enabled is false and ExecutionResult.Perf is null if someone call AddApolloTracing.

sungam3r: Maybe I'm wrong and if the `tracing` key exists then it should have some content.

src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -1029,13 +1029,52 @@ public static IGraphQLBuilder AddValidationRule<TValidationRule>(this IGraphQLBu
...
+                throw new ArgumentNullException(nameof(enablePredicate));
+
+            builder.AddMiddleware<InstrumentFieldsMiddleware>();
+            builder.ConfigureExecution(async (options, next) =>
+            {
+                if (enablePredicate(options))
+                    options.EnableMetrics = true;
+                DateTime start = DateTime.UtcNow;
+                var ret = await next(options).ConfigureAwait(false);
+                if (options.EnableMetrics)
```
sungam3r, line 1062:
```suggestion
                if (enabled)
```
What should be the behavior in case when some middleware set `EnableMetrics` to other value?

sungam3r: I suggest to use the same calculated value inside this method to be consistent.

Shane32: No, it’s designed to allow for the user to set the EnableMetrics value anywhere else in the code, which may often be done when setting options from configuration settings before calling the document executer. In this manner it is consistent, and it is fully documented.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: Typically a user wants to add tracing results only based on a configuration setting. I surely wouldn’t want my production systems adding trace results, unless I (for example) set a HTTP header for debugging. Pretty sure we do not want any added results if EnableMetrics is false.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: Definitely not as described below and as written in the xml comments.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: Yes I am sure.

Shane32:
I would imagine that typically in production scenarios the only time metrics are enabled is when debugging information is requested.  The tracing data is a LOT of data, compared to without, and I don't imagine that anyone would want that in production unless specifically requested.

A simple query like this:

```gql
{
  hero {
    __typename
    id
    name
    appearsIn
    friends {
      __typename
      id
      name
      appearsIn
    }
  }
}
```

Is 36 formatted lines of JSON, but with tracing, it is 188 lines -- representing a 5-fold increase in data transmission -- and this is a relatively small query.  Not to mention the notable amount of allocations required to collect the data.

I seriously can't imagine anyone that wants this data in production -- certainly not to be returned to the user anyway.  Perhaps some kind of logging scenario.

So I imagine that one of the main goals of the `AddApolloTracing` method is to allow users to configure at runtime whether they want the added data.  Sure we have the default `AddApolloTracing()` to just enable all the time; that's probably a good default behavior to get a user start quickly, but not likely to be used in production.

Rather, you'd want a way to configure it.  Some users will just want this configured during development, and some would want it configured during runtime.  Some perhaps both.  So as it is now we support a variety of different methods to accomplish this:

### Configure during development

1. Pull from `IConfiguration` and set when calling the method.

```cs
var opts = new GraphQLOptions();
configuration.GetSection("GraphQLOptions").Bind(opts);
services.AddGraphQL(b => b
    .AddApolloTracing(opts.EnableMetrics));
```

2. Pull options from DI

```cs
services.AddGraphQL(b => b
    .AddApolloTracing(opts => opts.RequestServices.GetRequiredService<IOptions<GraphQLOptions>>().Value.EnableMetrics));
```

### Configure during runtime

1. Use within predicate

```cs
services.AddGraphQL(b => b
    .AddApolloTracing(opts => {
        return (opts.InputExtensions?.TryGetValue("enableMetrics", out var value) ?? false)
            && value is bool boolValue ? boolValue : false;
    }));
```

2. Configure within `ConfigureExecutionOptions`

```cs
services.AddGraphQL(b => b
    .ConfigureExecutionOptions(opts =>
    {
        opts.EnableMetrics = (opts.InputExtensions?.TryGetValue("enableMetrics", out var value) ?? false)
            && value is bool boolValue ? boolValue : false;
    })
    .AddApolloTracing(false));
```

3. Configure when calling document executer

```cs
services.AddGraphQL(b => b
    .AddApolloTracing(false));

// execution code
{
    Query = "{ hero }",
    EnableMetrics = true,  //or any other condition
};
var ret = await executer.ExecuteAsync(opts);
```

The current PR supports all of these scenarios.  Overwriting `EnableMetrics` when calling `AddApolloTracing(false)` would break two of the above scenarios.  And I cannot imagine that users would want tracing data added if no metrics were collected.

For any other scenario, I would let the user add the middleware via `.AddMiddleware<>` and call `ConfigureExecution` to append the results if desired.  If they want to add the key to the extensions property even when metrics data has not been collected, so be it.  But that would be an "advanced scenario", not default behavior.

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: What is the answer for second question?

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: OK. So tracing data **may** be calculated, but some middleware **after** actual document executing may set `EnableMetrics` to `false`, so there will be no tracing data in result. Right?

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: Right, OK.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: Please read my thought process below.

As for your second question, it does not make sense.  The default value is already false.  If it was set to true manually by user code it should not be changed by the code here; they can modify their other code so that it is set to true conditionally.  If you wish to post a code snippet of a realistic scenario you anticipate happening, we can review.  I have demonstrated a number of scenarios below and I believe the behavior in this PR is correct.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: That is a possible scenario, which makes sense.

```cs
services.AddGraphQL(b => b
    .AddApolloTracing() //note: order matters here
    .ConfigureExecution(async (opts, next) =>
    {
        var ret = await next(opts);
        // log captured metrics
        opts.EnableMetrics = false;
    }));
```

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: I see. Just reading emails in order.

src/GraphQL/GraphQLBuilderExtensions.cs
Shane32: I'm rather indifferent.  `enable` was used with `AddMetrics` to match up to `EnableMetrics`.  Perhaps this should be `enableMetrics` to more clearly reflect that it is setting the `EnableMetrics` option.

Shane32: Consider that, and then let me know your preference and I can change it as you feel would be best.

sungam3r:
Rel: #3135

sungam3r (approved):
Let's merge.

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: 👍

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: `enebleMetrics` then

PR #3134, opened 2022-05-12
Automatic Persisted Queries

SlavaUtesinov:
Implementation of [Automatic Persisted Queries](https://www.apollographql.com/docs/react/api/link/persisted-queries) (APQ) based on `IConfigureExecution` middleware and in-memory cache. It is a possible solution of the issue graphql-dotnet/server#132. The feature has been implemented in accordance with APQ protocol, [Apollo logic](https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-core/src/requestPipeline.ts#L144,L208)  and [general architecture](https://www.apollographql.com/blog/apollo-client/persisted-graphql-queries/):
image
Typical incoming request looks the following way:
```
{
  extensions: {
    persistedQuery: {
      version: '1',
      sha256Hash: 'd7b0dfafc61a1f0618f4f346911d5aa87bef97b134f2943383223bdac4410134'
    }
  }
}
```


Shane32:
Is this targeting v5.3.0 or v6?  Assuming we are targeting v5.3.0 then we cannot make breaking changes to existing classes and interfaces.

SlavaUtesinov:
I was trying to make breaking changes as little as possible. It seems like only `GetMemoryCacheEntryOptions` method's signature was changed. I think we can deal with it.

Shane32:
I think the biggest issue is that `IDocumentCache` has changed.  Interfaces can't be changed because adding, changing or removing methods in an interface are all breaking changes.  The inheritance tree of an interface might be able to be changed.... but certainly not if it implements any new methods.  So pretty much interfaces are locked once created.

Shane32:
I do see that `IDocumentCache` will have the same method signatures before as after.  I would need to check to see if it's a breaking change or not.  I think it's still a breaking change.  @sungam3r do you know?

For example -- would this compile anymore??

```cs
class test : IDocumentCache
{
    System.Threading.Tasks.ValueTask<GraphQLParser.AST.GraphQLDocument?> IDocumentCache.GetAsync(string key)
    {
        //do some stuff
    }
}
```

I can't remember -- wouldn't it have to change to this?

```cs
class test : IDocumentCache
{
    System.Threading.Tasks.ValueTask<GraphQLParser.AST.GraphQLDocument?> ICache<GraphQLParser.AST.GraphQLDocument>.GetAsync(string key)
    {
        //do some stuff
    }
}
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
```diff
@@ -0,0 +1,120 @@
...
+    /// Create <see cref="ExecutionResult"/> with provided error.
+    /// </summary>
+    protected virtual ExecutionResult CreateExecutionResult(ExecutionError error) => new ExecutionResult { Errors = new ExecutionErrors { error } };
+
+    /// <inheritdoc/>
+    public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
+    {
+        var apq = GetAPQProperties(options.Extensions);
+
+        if (apq.Enabled && apq.Version != SUPPORTED_VERSION)
```
Shane32, line 81:
```suggestion
        if (!apq.Enabled)
            return await next(options).ConfigureAwait(false);

        if (apq.Version != SUPPORTED_VERSION)
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            return CreateExecutionResult(new PersistedQueryUnsupportedVersionError(apq.Version));
+        }
+
+        if (string.IsNullOrWhiteSpace(options.Query))
+        {
+            if (string.IsNullOrWhiteSpace(apq.Hash))
+            {
+                return CreateExecutionResult(new ExecutionError("GraphQL query is missing."));
```
Shane32, line 90: Let's create a class for this situation and move the check into the `DocumentExecuter.CoreExecuteAsync` method.  This is what I wrote for the server side which probably should be added to the main project.  The other errors you've added probably can inherit from `RequestError` also.  @sungam3r should also review.

```cs
/// <summary>
/// Represents an error indicating that no GraphQL query was provided to the request.
/// </summary>
public class QueryMissingError : RequestError
{
    /// <inheritdoc cref="QueryMissingError"/>
    public QueryMissingError() : base("GraphQL query is missing.") { }
}

/// <summary>
/// Represents an error that occurred prior to the execution of the request.
/// </summary>
public class RequestError : ExecutionError
{
    /// <summary>
    /// Initializes an instance with the specified message.
    /// </summary>
    public RequestError(string message) : base(message)
    {
        Code = GetErrorCode();
    }

    /// <summary>
    /// Initializes an instance with the specified message and inner exception.
    /// </summary>
    public RequestError(string message, Exception? innerException) : base(message, innerException)
    {
        Code = GetErrorCode();
    }

    internal string GetErrorCode()
    {
        var code = ErrorInfoProvider.GetErrorCode(GetType());
        if (code != "REQUEST_ERROR" && code.EndsWith("_ERROR", StringComparison.Ordinal))
            code = code.Substring(0, code.Length - 6);
        return code;
    }
}
```

sungam3r: Do we need to do something else in the server after this PR?

Shane32: Yes; we need to allow requests with a blank Query.  Already done in https://github.com/Shane32/GraphQL.AspNetCore3/pull/28 .

Shane32: If we code DocumentExecuter to return a response with an ExecutionError when query is null, then we can eliminate the query-null checking from the server entirely.  But as of now it throws an `InvalidOperationException`.

sungam3r: Now I see I think. Do you want to check for `RequestError` type in server and return `HttpStatusCode.BadRequest` in that case?

sungam3r: > Represents an error that occurred prior to the execution of the request.

Validation also occures prior to the execution but this error is not about GraphQL validation, it is about client supplied argument on the transport/infrastructure layer.

Shane32: > Now I see I think. Do you want to check for `RequestError` type in server and return `HttpStatusCode.BadRequest` in that case?

That already occurs, generally speaking.  Meaning that any validation error (where `result.Executed == false`) will return a 400 bad request error.  In GraphQL.AspNetCore3 it is configurable, but the default behavior would be to return a 400 error.

> Validation also occures prior to the execution but this error is not about GraphQL validation, it is about client supplied arguments on the transport/infrastructure layer.

Right, so that's why I made a new class `RequestError` - more for 'good housekeeping' rather than any real implementation difference.

It certainly is possible that the server could return 400 for a `RequestError` exception -- even if it returns 200 for validation errors -- if we use this type of inheritance.

Shane32: Personally I think of a request without a query as a validation/syntax error.  Meaning, what's the difference between these?

```json
{
  "query": "",
  "variables": { "stuff": 123 }
}
```

and

```json
{
  "query": null,
  "variables": { "stuff": 123 }
}
```

and

```json
{
  "query": "{",
  "variables": { "stuff": 123 }
}
```

All are essentially the same issue -- the query is invalid.  It's not a problem deserializing the GraphQLRequest json.  It's a problem with the query.  I don't have an issue differentiating those errors, just saying, all are somewhat similar errors.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
```diff
@@ -0,0 +1,120 @@
...
+        string? versionResult = null;
+        bool enabled = false;
+
+        if (extensions?.TryGetValue("persistedQuery", out var persistedQueryObject) ?? false)
+        {
+            enabled = true;
+
+            if (persistedQueryObject is Dictionary<string, object> persistedQuery)
+            {
+                if (persistedQuery.TryGetValue("sha256Hash", out var sha256HashObject) && sha256HashObject is string sha256Hash && !string.IsNullOrWhiteSpace(sha256Hash))
```
sungam3r, line 40:
```suggestion
                if (persistedQuery.TryGetValue("sha256Hash", out var sha256HashObject) && sha256HashObject is string sha256Hash && !string.IsNullOrEmpty(sha256Hash))
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    hashResult = sha256Hash;
+                }
+
+                if (persistedQuery.TryGetValue("version", out var versionObject) && versionObject is string version && !string.IsNullOrWhiteSpace(version))
```
sungam3r, line 45:
```suggestion
                if (persistedQuery.TryGetValue("version", out var versionObject) && versionObject is string version && !string.IsNullOrEmpty(version))
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        return (hashResult, versionResult, enabled);
+    }
+
+    /// <summary>
+    /// Compute SHA256 hash.
+    /// </summary>
+    protected virtual string ComputeSHA256(string input)
+    {
+        var bytes = SHA256.Create().ComputeHash(Encoding.UTF8.GetBytes(input));
```
sungam3r, line 60: `SHA256.Create()` may be moved into static.

sungam3r: And a general note regarding this method - it's greedy for memory.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
```diff
@@ -0,0 +1,120 @@
...
+    public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
+    {
+        var apq = GetAPQProperties(options.Extensions);
+
+        if (apq.Enabled && apq.Version != SUPPORTED_VERSION)
+        {
+            return CreateExecutionResult(new PersistedQueryUnsupportedVersionError(apq.Version));
+        }
+
+        if (string.IsNullOrWhiteSpace(options.Query))
```
sungam3r, line 86: Why `OrWhiteSpece` everywhere?

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
Shane32: Are you saying that you can use a static instance of SHA256?  I do not see in the documentation that instance members can be called from multiple threads in parallel.  Do you believe that is the case?  If so, why?

I would expect that we can write some optimized code that reduces memory usage here.  I would probably write such an implementation in a separate PR if we choose to do so.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
```diff
@@ -0,0 +1,120 @@
...
+                }
+                else
+                {
+                    options.Query = queryFromCache;
+                }
+            }
+        }
+        else if (!string.IsNullOrWhiteSpace(apq.Hash))
+        {
+            if (apq.Hash!.Equals(ComputeSHA256(options.Query!), StringComparison.InvariantCultureIgnoreCase))
```
sungam3r, line 108: I suggest to switch from comparing characters to comparing bytes.

Shane32: `StringComparison.Ordinal` ?

src/GraphQL.MemoryCache/AutomaticPersistedQueries/PersistedQueryUnsupportedVersionError.cs
```diff
@@ -0,0 +1,18 @@
...
+/// An error in case of unsupported version.
+/// </summary>
+[Serializable]
+public class PersistedQueryUnsupportedVersionError : ValidationError
+{
+    /// <summary>
+    /// Initializes a new instance.
+    /// </summary>
+    public PersistedQueryUnsupportedVersionError(string? version)
+        : base($"Persisted queries with '{version}' version are not supported.")
```
sungam3r, line 15:
```suggestion
        : base($"Automatic persisted queries protocol of version '{version}' is not supported.")
```

sungam3r:
Regarding biggest issue. For me this implementation is overcomplicated and contains a lot of things that have no value. I'm completely against `ICache<TValue>`/`IQueryCache` "abstractions". I suggest to implement APQ with **embedded** memory cache. It's simple, it's small, it has no breaking changes, it does not touch existing interfaces. Moreover - almost any implementation different from memory cache will suffer from the same problem - big network data interchange pulling query text from some outer service/storage. So for me it makes no sense to provide `ICache<TValue>`/`IQueryCache` and try to build something around them. **If** someone **really** wants to use some other storage then he may copy-paste required code.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
sungam3r: The difference is whether or not we are inside GraphQL engine or outside.

> That already occurs, generally speaking. Meaning that any validation error (where result.Executed == false) will return a 400 bad request error. 

Where is that code?

Shane32: Here is the pointer to the code within my PR https://github.com/graphql-dotnet/server/pull/774 :

https://github.com/graphql-dotnet/server/blob/f2a7011c74ac5317f22fe4149c969d4ee3f241f7/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs#L402

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
sungam3r: > Are you saying that you can use a static instance of SHA256?

I believe yes but in general this method should be called rather rarely and bytes <-> string conversion takes much more memory.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
sungam3r: No, comparing raw bytes i.e remove byte[] -> string conversion from `ComputeSHA256` but convert apq.Hash to `byte[]`. It can be done with ArrayPool.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
sungam3r: Ah, new code. I no longer have the strength to review anything today, only tomorrow.

Shane32:
@sungam3r Don't you think that a REDIS implementation could be significant to users?  You do not think we should at least support some type of extension supporting such a scenario?

I suggest that even if we eliminate the interfaces you mentioned (okay with me), implement the APQ functionality as an abstract class, with the memory implementation as a derived class on top of it.  In that manner, anyone can implement their own caching implementation and there are less abstractions that need to be maintained without completely blocking derived implementations.

Something like this: (quick draft)

```cs
public abstract class APQBase : IConfigureExecution
{
    public virtual Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
    {
        //implementation here
    }

    // add some methods to generate errors, which can be overridden if logging is desired
    public virtual ValueTask<ExecutionResult> HandleNotInCacheErrorAsync(ExecutionOptions options)
        => new(new ExecutionResult { ExecutionErrors = new() { new NotInCacheError() } }); //or whatever
    // similar for invalid version and similar errors

    public virtual string HashQuery(string query) { } // default SHA hash code here, which can be overridden

    public abstract ValueTask<string?> GetQueryAsync(string hash)
    {
    }

    public abstract Task SetQueryAsync(string hash, string query)
    {
    }
}
```

Without a REDIS cache, APQ usefulness is somewhat limited to a single-server design, as a request through a load balancer to one server may have a cache miss, and the subsequent request containing the full query may hit a different server, which could already have a populated cache entry.  Of course we can optimize for an in-memory cache, but it seems that we should also support a custom cache (or `IDistributedCache` implementation) in some manner.  Some users may even want to stack a memory cache with a distributed cache.

sungam3r:
> You do not think we should at least support some type of _extension supporting_ such a scenario?

No if this causes a significant complication of the main scenario - keeping hash-query dictionary into memory.

> I suggest that even if we eliminate the interfaces you mentioned (okay with me), implement the APQ functionality as an abstract class, with the memory implementation as a derived class on top of it.

Of course, I argee with provided example. APQ may and should be abstracted itself as a whole concept. It seems that abstracting on top of `ICache<TValue>/IQueryCache` interfaces is a wrong place like abstracting for a small part inside a fixed Apollo implementation . 

> Without a REDIS cache, APQ usefulness is somewhat limited to a single-server design, as a request through a load balancer to one server may have a cache miss, and the subsequent request containing the full query may hit a different server, which could already have a populated cache entry

I understand but eventually (sooner or later) every app node's hash-query cache will be filled with all needed data.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
SlavaUtesinov: to increase rigor

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
sungam3r: Looks like error classes from #774 are needed for both PRs.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
sungam3r: I do not think that we need to interpret whitespaces in the query text in a special way. I did not see any place in codebase where we use this approach. @Shane32  ?

src/GraphQL/Execution/QueryMissingError.cs
```diff
@@ -0,0 +1,39 @@
...
+/// Represents an error indicating that no GraphQL query was provided to the request.
+/// </summary>
+public class QueryMissingError : RequestError
+{
+    /// <inheritdoc cref="QueryMissingError"/>
+    public QueryMissingError() : base("GraphQL query is missing.") { }
+}
+
+/// <summary>
+/// Represents an error that occurred prior to the execution of the request.
```
sungam3r, line 12:
```suggestion
/// Represents an error that occurred prior to the execution of the GraphQL request.
/// This refers to any errors that arise before passing the request inside the GraphQL engine, that is, even before its validation.
```

sungam3r: And please move into separate file in `Error` folder.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -134,7 +134,7 @@ private async Task<ExecutionResult> CoreExecuteAsync(ExecutionOptions options)
             if (options.Schema == null)
                 throw new InvalidOperationException("Cannot execute request if no schema is specified");
             if (options.Query == null && options.Document == null)
-                throw new InvalidOperationException("Cannot execute request if no query is specified");
+                return new ExecutionResult { Errors = new ExecutionErrors { new QueryMissingError() } };
```
sungam3r, line 137:
```suggestion
                throw new QueryMissingError();
```
I think that initial idea from @Shane32 was just like that - throw typed error to be able to catch it above and return 400 http code. Moreover, switching from throw to return is a breaking change in behavior.

src/GraphQL.Tests/Bugs/Bug1769.cs
```diff
@@ -31,14 +31,12 @@ public async Task DocumentExecuter_has_valid_options()
             Schema = Schema,
         }).ConfigureAwait(false);
         valid.Data.ShouldNotBeNull();
-        await Assert.ThrowsAsync<InvalidOperationException>(async () =>
+        var result = await de.ExecuteAsync(new ExecutionOptions()
```
sungam3r, line 34: https://github.com/graphql-dotnet/graphql-dotnet/pull/3134/files#r872080742

SlavaUtesinov:
In retrospect, `AutomaticPersistedQueriesExecutionBase` is the same thing as `IQueryCache` - just factory method vs strategy pattern, for me the last is more appropriate and comfortable because there is a strict separation of APQ logic and cache without mixing also test experience should be taken into account. Also I don't see any breaking changes - none of public API has been changed and for me it is better to reuse existing codebase i.e. `MemoryDocumentCache` than copy-paste it's parts, if someone wants another cache they could easily accomplish it on their own via custom implementation of `IQueryCache`. 
Nevertheless, I have made almost all you wanted to release this feature.

src/GraphQL/Execution/QueryMissingError.cs
Shane32: 👍

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: @sungam3r Uhhh....???  I would change this behavior here to return an ExecutionResult as @SlavaUtesinov already posted.  While your idea is somewhat logical, it seems like a make-work project.  Now the server project (and any other middleware usage) has to catch errors thrown from DocumentExecuter and check if it's an ExecutionError (or RequestError) and encapsulate it in an ExecutionResult.  Better just to write the code here in the DocumentExecuter.

We have never considered breaking changes to error conditions as preventing a minor version bump, and this change should have no negative effect on any existing code in the server project or provided samples.  For instance, GraphQL v6 (and GraphQL.AspNetCore3) will still return a 400 error and not even call the DocumentExecuter to begin with.

Rather, we are now "supporting a feature" we did not before -- namely, providing a response when there is no query provided.  Again, while this is breaking in the strictest sense, it should be allowed with a minor version bump.

Finally, as stated previously, situations where `result.Executed == false` should already return 400 according to spec, and does so and will continue to do so.  There is no reason why a missing (`null`) query should be a 400 and an empty (`""`) query should return 200.  If a user were returning 200 for empty (`""`) queries, then it should be acceptable to return a 200 for a null query as well.

Shane32: Note: actually the server v6 as it exists now does not return 400 in the case of validation errors.  That situation is fixed within my PR.  But it does check for null and return 400 regardless of the change to this code.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
Shane32: Well ... if I was coding it, I might use `== null` because the query specifically shouldn't exist at all in the request -- correct?  I would consider `IsNullOrEmpty` in the case that for some reason the transport always deserializes an empty string for the query, even if none is provided.  (Although I don't think that actually occurs.)  I don't see a reason to use `IsNullOrWhiteSpace`.

However, I feel this point is a bit moot no matter the choice, so any of the three options is fine with me.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// <summary>
+    /// Compute SHA256 hash.
+    /// </summary>
+    public virtual string ComputeSHA256(string input)
+    {
+        byte[] bytes;
+        lock (_lockObject)
+        {
+            bytes = _sha256.ComputeHash(Encoding.UTF8.GetBytes(input));
+        }
```
Shane32, line 61: By the way, I confirmed that the SHA256 instance can not be called concurrently from multiple threads.  The resulting hash is corrupt when two threads concurrently create a hash.

I suggest a small performance change here along with the removal of the lock object:

```suggestion
        var inputBytes = Encoding.UTF8.GetBytes(input);
        var shaShared = Interlocked.Exchange(ref _sha256, null) ?? SHA256.Create();
        var bytes = shaShared.ComputeHash(inputBytes);
        Interlocked.CompareExchange(ref _sha256, shaShared, null);
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        var builder = new StringBuilder();
+        foreach (var item in bytes)
+        {
+            builder.Append(item.ToString("x2"));
+        }
+
+        return builder.ToString();
```
Shane32, line 69: I would check and see if this works properly and include it if so.

```suggestion
#if NET5
        return Convert.ToHexString(bytes);
#else
        var builder = new StringBuilder();
        foreach (var item in bytes)
        {
            builder.Append(item.ToString("x2"));
        }

        return builder.ToString();
#endif
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// <summary>
+    /// Get query by hash. It is likely to be implemented via cache mechanism.
+    /// </summary>
+    public abstract ValueTask<string?> GetQueryAsync(string hash);
+
+    /// <summary>
+    /// Set query by hash. It is likely to be implemented via cache mechanism.
+    /// </summary>
+    public abstract ValueTask SetQueryAsync(string hash, string query);
```
Shane32, line 85:
```suggestion
    public abstract Task SetQueryAsync(string hash, string query);
```

`Task` should be used when no result is returned.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/PersistedQueryBadHashError.cs
```diff
@@ -0,0 +1,18 @@
+using GraphQL.Execution;
+
+namespace GraphQL.Caching;
+
+/// <summary>
+/// An error in case provided hash doesn't equal to calculated hash.
+/// </summary>
+[Serializable]
```
Shane32, line 8:
```suggestion
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/PersistedQueryNotFoundError.cs
```diff
@@ -0,0 +1,18 @@
+using GraphQL.Execution;
+
+namespace GraphQL.Caching;
+
+/// <summary>
+/// An error in case a query hasn't been found by hash.
+/// </summary>
+[Serializable]
```
Shane32, line 8:
```suggestion
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/PersistedQueryUnsupportedVersionError.cs
```diff
@@ -0,0 +1,18 @@
+using GraphQL.Execution;
+
+namespace GraphQL.Caching;
+
+/// <summary>
+/// An error in case of unsupported version.
+/// </summary>
+[Serializable]
```
Shane32, line 8:
```suggestion
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
Shane32: Note that ToHexString encodes as uppercase...

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: I see the suggestion was merged.  I think this is incorrect and should be reverted.

Shane32:
I completed my review.  I have a few minor performance comments that do not really matter and could be left as-is.

I do think that `DocumentExecuter` should return an `ExecutionResult` when `Query` and `Document` are both null.  It should not throw directly.  I understand this is a change in behavior, which I would consider a new feature for v5.3.0.  @sungam3r and I can review in a separate PR if we want to merge this as it is now, however.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: #3138

Shane32 (approved):

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
Shane32: With the current inheritance pattern, which we already discussed, I would leave as-is.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
Shane32: Confirmed it cannot be static.


```cs
    [Fact]
    public async Task sha256test()
    {
        var sha256shared = SHA256.Create();
        var tasks = new List<Task>();

        // spin up 8 concurrent threads
        for (int i = 0; i < 8; i++)
        {
            tasks.Add(Task.Factory.StartNew(() =>
            {
                var rng = RandomNumberGenerator.Create();
                var b = new byte[4];
                rng.GetBytes(b);
                var r = new Random(BitConverter.ToInt32(b, 0));
                for (int j = 0; j < 100; j++)
                {
                    b = new byte[r.Next(8192, 10240)];
                    rng.GetBytes(b);
                    var hash1 = SHA256.Create().ComputeHash(b);

                    // passes
                    var hash2 = sha.ComputeHash(b);
                    hash1.ShouldBe(hash2, customMessage: "new instance different than local instance");

                    // fails
                    var hash3 = sha256shared.ComputeHash(b);
                    hash1.ShouldBe(hash3, customMessage: "new instance different than shared instance");
                }
            }));
        }
        foreach (var t in tasks)
        {
            await t.ConfigureAwait(false);
        }
    }
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
SlavaUtesinov: Only `<TargetFramework>netstandard2.0</TargetFramework>` [available](https://github.com/graphql-dotnet/graphql-dotnet/blob/5.2.0/src/GraphQL.MemoryCache/GraphQL.MemoryCache.csproj#L4)

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: OK

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
sungam3r: > By the way, I confirmed that the SHA256 instance can not be called concurrently from multiple threads. 

Well, so be it, I was wrong.

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
sungam3r: Just add net5 tfm. For example, main package declares supported tfms like that
`  <TargetFrameworks>netstandard2.1;netstandard2.0;net6</TargetFrameworks>`

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecution.cs
sungam3r: >  var hash2 = sha.ComputeHash(b);

Where is `sha` variable?

Shane32: Supposed to be right after “rng= “

Shane32:
> > which I would consider a new feature for v5.3.0
> 
> I'm fine either way. In case v5.3.0 it should at least be mentioned in release notes.

I agree

Shane32 (approved):
Looks great!

src/GraphQL/Execution/Errors/RequestError.cs
```diff
@@ -0,0 +1,31 @@
+namespace GraphQL.Execution;
+
+/// <summary>
+/// Represents an error that occurred prior to the execution of the request.
```
sungam3r, line 4:
```suggestion
/// Represents an error that occurred prior to the execution of the GraphQL request.
/// This refers to any errors that arise before passing the request inside the GraphQL engine, that is, even before its validation.
```

src/GraphQL/Execution/Errors/QueryMissingError.cs
```diff
@@ -0,0 +1,11 @@
+namespace GraphQL.Execution;
+
+/// <summary>
+/// Represents an error that occurred prior to the execution of the GraphQL request.
+/// This refers to any errors that arise before passing the request inside the GraphQL engine, that is, even before its validation.
```
sungam3r, line 5:
```suggestion
/// Represents an error indicating that no GraphQL query was provided to the request.
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
sungam3r: resolved

src/GraphQL.Tests/Bugs/Bug1769.cs
```diff
@@ -31,14 +31,12 @@ public async Task DocumentExecuter_has_valid_options()
...
         {
-            await de.ExecuteAsync(new ExecutionOptions()
-            {
-                Query = null,
-                Schema = Schema,
-            }).ConfigureAwait(false);
+            Query = null,
+            Schema = Schema,
         }).ConfigureAwait(false);
+        result.Errors.Single().ShouldBeOfType(typeof(QueryMissingError));
```
sungam3r, line 39:
```suggestion
        result.Errors.Single().ShouldBeOfType(typeof(QueryMissingError));
        result.Errors.Single().Locations.ShouldBeNull();
        result.Errors.Single().Path.ShouldBeNull();
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// </summary>
+    public const string SUPPORTED_VERSION = "1";
+
+    private SHA256? _sha256;
+
+    /// <summary>
+    /// Searching APQ fields in <see cref="ExecutionOptions.Extensions"/> based on a protocol:
+    /// https://www.apollographql.com/docs/react/api/link/persisted-queries/#protocol
+    /// </summary>
+    public virtual (string? Hash, string? Version, bool Enabled) GetAPQProperties(Inputs? extensions)
```
sungam3r, line 24: Change to `protected`?

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,135 @@
...
+        var queryHash = builder.ToString();
+#endif
+
+        return hash.Equals(queryHash, StringComparison.InvariantCultureIgnoreCase);
+    }
+
+    /// <summary>
+    /// Create <see cref="ExecutionResult"/> with provided error.
+    /// </summary>
+    public virtual ExecutionResult CreateExecutionResult(ExecutionError error) => new ExecutionResult { Errors = new ExecutionErrors { error } };
```
sungam3r, line 78: Change to `protected`?

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+
+        return (hashResult, versionResult, enabledResult);
+    }
+
+    /// <summary>
+    /// Check equality of the provided hash and a hash computed from the query.
+    /// </summary>
+    public virtual bool CheckHash(string hash, string query)
```
sungam3r, line 54: Change to `protected`?

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        var inputBytes = Encoding.UTF8.GetBytes(query);
+        var shaShared = Interlocked.Exchange(ref _sha256, null) ?? SHA256.Create();
+        var bytes = shaShared.ComputeHash(inputBytes);
+        Interlocked.CompareExchange(ref _sha256, shaShared, null);
+
+#if NET5_0_OR_GREATER
+        var queryHash = Convert.ToHexString(bytes);
+#else
+        var builder = new StringBuilder();
```
sungam3r, line 64:
```suggestion
        var builder = new StringBuilder(bytes.Length * 2);
```

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// <summary>
+    /// Create <see cref="ExecutionResult"/> with provided error.
+    /// </summary>
+    public virtual ExecutionResult CreateExecutionResult(ExecutionError error) => new ExecutionResult { Errors = new ExecutionErrors { error } };
+
+    /// <summary>
+    /// Get query by hash. It is likely to be implemented via cache mechanism.
+    /// </summary>
+    public abstract ValueTask<string?> GetQueryAsync(string hash);
```
sungam3r, line 83: Change to `protected`?

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// <summary>
+    /// Set query by hash. It is likely to be implemented via cache mechanism.
+    /// </summary>
+    public abstract Task SetQueryAsync(string hash, string query);
```
sungam3r, line 88: Change to `protected`?

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesCacheOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace GraphQL.Caching;
+
+/// <summary>
+/// Provides configuration options for <see cref="AutomaticPersistedQueriesExecution"/>.
+/// </summary>
+public class AutomaticPersistedQueriesCacheOptions : MemoryCacheOptions, IOptions<AutomaticPersistedQueriesCacheOptions>
+{
+    /// <summary>
+    /// Initializes a default instance with the size limit set to 100,000.
```
sungam3r, line 12:
```suggestion
    /// Initializes a default instance with the size limit (sum of all requests lengths in symbols) set to 100,000.
```

sungam3r:
I'm fine with this PR. Only 1 thing - make methods from  AutomaticPersistedQueriesExecutionBase  protected

src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs
sungam3r: @Shane32 What do you think? I would change all methods to protected, merge and continue with PR from server project.

sungam3r: fix: merge and publish v5.3.0

Shane32: @sungam3r I agree the methods should be protected

sungam3r: @SlavaUtesinov Let's do this final change and I'm glad to merge this awesome feature.

SlavaUtesinov: done

sungam3r (approved):

PR #3143, opened 2022-05-18
Update api approvals for .NET SDK 6.0.300

Shane32:
See:
- https://github.com/PublicApiGenerator/PublicApiGenerator/issues/207

Without this PR, api approvals will generate differently locally than within GitHub Actions (assuming that the current version of Visual Studio is installed locally).


sungam3r:
Good)

PR #3147, opened 2022-05-20
Throw proper exception when RequestServices is null

Shane32:
Fixes issue reported at:
- https://github.com/graphql-dotnet/authorization/issues/235

### DIFF
--- a/src/GraphQL/GraphQLBuilderExtensions.cs
+++ b/src/GraphQL/GraphQLBuilderExtensions.cs
@@ -649,7 +649,11 @@ namespace GraphQL
             where TDocumentListener : class, IDocumentExecutionListener
         {
             builder.Services.RegisterAsBoth<IDocumentExecutionListener, TDocumentListener>(serviceLifetime);
-            builder.ConfigureExecutionOptions(options => options.Listeners.Add(options.RequestServices!.GetRequiredService<TDocumentListener>()));
+            builder.ConfigureExecutionOptions(options =>
+            {
+                var requestServices = options.RequestServices ?? throw new MissingRequestServicesException();
+                options.Listeners.Add(requestServices.GetRequiredService<TDocumentListener>());
+            });
             return builder;
         }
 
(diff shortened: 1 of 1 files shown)


src/GraphQL/GraphQLBuilderExtensions.cs
```diff
@@ -649,7 +649,11 @@ public static IGraphQLBuilder AddDocumentListener<TDocumentListener>(this IGraph
             where TDocumentListener : class, IDocumentExecutionListener
         {
             builder.Services.RegisterAsBoth<IDocumentExecutionListener, TDocumentListener>(serviceLifetime);
-            builder.ConfigureExecutionOptions(options => options.Listeners.Add(options.RequestServices!.GetRequiredService<TDocumentListener>()));
+            builder.ConfigureExecutionOptions(options =>
+            {
+                var requestServices = options.RequestServices ?? throw new MissingRequestServicesException();
```
sungam3r, line 654: @Shane32 I see a lot of places with such pattern (here and in server project). Let's add simple one line extension method:
```cs
public static IServiceProvider RequestServicesOrThrow(this IProvideRequestServices provider) => provider.RequestServices ?? throw new MissingRequestServicesException();

public interface IProvideRequestServices
{
  IServiceProvider? RequestServices { get; set; }
}
```

Then add `IProvideRequestServices` to `ExecutionOptions` and `IResolveFieldContext`.

Shane32: That is a breaking change.  There are a number of non-breaking changes we could do, but adding `IProvideRequestServices` to `IResolveFieldContext` is not one of them.

Non-breaking changes:
1. Add `RequestServicesOrThrow` extension method for `ExecutionOptions`, `ExecutionContext`, `IExecutionContext`, `ValidationContext`, and `IResolveFieldContext`
2. Add `GetRequiredService` extension method for the above classes, which will throw `RequestServicesMissingException` if `RequestServices` is null
3. Have the first statement within `DocumentExecuter` replace a null `RequestServices` property `NullServiceProvider` (which throws `RequestServicesMissingException` for any call to `GetService)

As a breaking change, `IProvideRequestServices` should be added to the aforementioned classes and interfaces for the next major version.

Then for the next major version, we can simplify the added extension methods, if any.

Other possibilities with the next major version:
- Make `ExecutionOptions.RequestServices` required, with an immediate check within `DocumentExecuter`

sungam3r: Option 1 (with notes/todo/issue) about new interface in future.

sungam3r: OK

Shane32:
@sungam3r Can we merge this until you have time to implement the extension methods you wanted?

src/GraphQL/GraphQLBuilderExtensions.cs
sungam3r: #3167

PR #3148, opened 2022-05-20
Add dependency-review-action

sungam3r:
I saw this workflow in one of the watched projects and decided to try. At least it is very fast.


sungam3r:
ping @Shane32

sungam3r:
https://github.com/marketplace/actions/dependency-review

sungam3r:
So ? New check is small and fast. I see no reason to reject this.

PR #3154, opened 2022-05-31
Release DataLoaderList for cached DataLoaderPair entries

Shane32:
If a DataLoaderBase is used as a singleton, so that entries are cached project-wide, then the individual DataLoaderPair entries retain unnecessary references to DataLoaderList classes.

This PR releases the reference to DataLoaderList after the result has been set.

Also added some memory barriers to be sure the result is read after the check to see if the result has been set.  This may be necessary now that the 'await' is not always called, which would have an implicit memory barrier.

The breaking change is that the `Loader` property has been removed from `DataLoaderPair`.  I can't really imagine a way that the property would be needed.

### DIFF
--- a/src/GraphQL.DataLoader/DataLoaderPair.cs
+++ b/src/GraphQL.DataLoader/DataLoaderPair.cs
@@ -85,6 +85,7 @@ namespace GraphQL.DataLoader
 
         async Task<object?> IDataLoaderResult.GetResultAsync(CancellationToken cancellationToken)
         {
+            // same code as above; prevents an additional allocation
             var loader = _loader;
             if (loader != null)
                 await loader.DispatchAsync(cancellationToken).ConfigureAwait(false);
--- a/src/GraphQL.DataLoader/DataLoaderPair.cs
+++ b/src/GraphQL.DataLoader/DataLoaderPair.cs
@@ -87,11 +87,7 @@ namespace GraphQL.DataLoader
         {
             var loader = _loader;
             if (loader != null)
-            {
-                // it does not matter if there are simultaneous calls to DispatchAsync as DataLoaderList
-                // protects against double calls to DispatchAsync
                 await loader.DispatchAsync(cancellationToken).ConfigureAwait(false);
-            }
             return Result;
         }
     }
--- a/src/GraphQL.DataLoader/DataLoaderPair.cs
+++ b/src/GraphQL.DataLoader/DataLoaderPair.cs
@@ -16,7 +16,7 @@ namespace GraphQL.DataLoader
         /// </summary>
         public DataLoaderPair(IDataLoader loader, TKey key)
         {
-            Loader = loader ?? throw new ArgumentNullException(nameof(loader));
+            _loader = loader ?? throw new ArgumentNullException(nameof(loader));
             Key = key;
         }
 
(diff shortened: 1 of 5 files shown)


src/GraphQL.DataLoader/DataLoaderPair.cs
```diff
@@ -30,17 +30,29 @@ public DataLoaderPair(IDataLoader loader, TKey key)
...
         /// Returns the result if it has been set, or throws an exception if not
         /// </summary>
-        public T Result => IsResultSet ? _result : throw new InvalidOperationException("Result has not been set");
+        public T Result
+        {
+            get
+            {
+                if (IsResultSet)
+                {
+                    Interlocked.MemoryBarrier(); // ensure that _loader is read before _result is read
```
sungam3r, line 44: 🤔 A rather rare API to see, I don’t even have experience working with it. What is its help here?

src/GraphQL.DataLoader/DataLoaderPair.cs
```diff
@@ -60,15 +73,22 @@ public void SetResult(T value)
         /// <param name="cancellationToken">Optional <seealso cref="CancellationToken"/> to pass to fetch delegate</param>
         public async Task<T> GetResultAsync(CancellationToken cancellationToken = default)
         {
-            if (!IsResultSet)
-                await Loader.DispatchAsync(cancellationToken).ConfigureAwait(false);
+            var loader = _loader;
+            if (loader != null)
```
sungam3r, line 77: Why not `IsResultSet` here and below?

src/GraphQL.DataLoader/DataLoaderPair.cs
```diff
@@ -60,15 +73,22 @@ public void SetResult(T value)
...
+                await loader.DispatchAsync(cancellationToken).ConfigureAwait(false);
+            }
             return Result;
         }
 
         async Task<object?> IDataLoaderResult.GetResultAsync(CancellationToken cancellationToken)
         {
-            if (!IsResultSet)
-                await Loader.DispatchAsync(cancellationToken).ConfigureAwait(false);
+            // same code as above; prevents an additional allocation
```
sungam3r, line 88: > prevents an additional allocation

where?

sungam3r (approved):
I don't use dataloaders. I trust your changes.

src/GraphQL.DataLoader/DataLoaderPair.cs
Shane32: Need to eliminate a race condition where another thread may clear _loader.  IsResultSet becomes redundant.

src/GraphQL.DataLoader/DataLoaderPair.cs
Shane32: If GetResultAsync calls the other overload, I believe it will cause an allocation while re-wrapping the value in a new Task.

src/GraphQL.DataLoader/DataLoaderPair.cs
Shane32: It’s use case is also very rare. Basically only is useful in situations like these where a lock is not necessary so long as the Intel CPU always reads from and writes to memory in the order stated. (Typically the compiler and even the CPU itself can reorder memory reads during its optimization)

Shane32: The code comment explains precisely why the line of code was added; hopefully it is clear.

Issue #1254, opened 2019-08-19
Mark objects not to cause impact during complexity analysis


SDLionas:
Is there a way to mark custom types and/or fields to be ignored in complexity analysis?

We have a complex structure like this:
```
price {
  value
  currency {
      identifier
  }
  units {
      identifier
      category {
          identifier
      }
  }
}
```
The problem we are facing is the fact that currency and units (plus their category) are cached and we just pick up and ID from the DB alongside value and then just get those from the cache.
However, all this nesting is calculated in both depth and complexity of the query. Is there a way to avoid this without implementing our analyzer?

sungam3r:
Now default complexity analyzer is very limited in its capabilities.

> Is there a way to avoid this without implementing our analyzer?

Not yet.

PR #3159, opened 2022-06-06
Field's impact for ComplexityAnalyzer

SlavaUtesinov:
An ability to specify a particular field's complexity impact by means of `Field<IntGraphType>("id").WithComplexityImpact(123)`, which will be taken into account by `ComplexityAnalyzer`. It is a possible solution of the issues #1254 and #298.

### DIFF
--- a/src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
+++ b/src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
@@ -93,7 +93,7 @@ public class ComplexityMetaDataFixture : IDisposable
         public ComplexityQuery()
         {
             Field<Hero>("hero", resolve: _ => 0)
-                .WithImpact(2);
+                .WithComplexityImpact(2);
         }
     }
 
@@ -102,12 +102,12 @@ public class ComplexityMetaDataFixture : IDisposable
         public Hero()
         {
             Field<IntGraphType>("id", resolve: context => context.Source)
-                .WithImpact(0);
+                .WithComplexityImpact(0);
 
             Field<StringGraphType>("name", resolve: context => $"Tom_{context.Source}");
 
             Field<ListGraphType<Hero>>("friends", resolve: context => new List<int> { 0, 1, 2 }.Where(x => x != context.Source).ToList())
-                .WithImpact(3);
+                .WithComplexityImpact(3);
         }
     }
 
--- a/src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
+++ b/src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
@@ -1,8 +1,8 @@
-using Microsoft.Extensions.DependencyInjection;
-using GraphQL.MicrosoftDI;
-using GraphQL.Validation.Complexity;
 using GraphQL.Execution;
+using GraphQL.MicrosoftDI;
 using GraphQL.Types;
+using GraphQL.Validation.Complexity;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace GraphQL.Tests.Complexity;
 
--- a/src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
+++ b/src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
@@ -1,6 +1,5 @@
 using Microsoft.Extensions.DependencyInjection;
 using GraphQL.MicrosoftDI;
-using GraphQL.StarWars;
 using GraphQL.Validation.Complexity;
 using GraphQL.Execution;
 using GraphQL.Types;
@@ -17,9 +16,9 @@ public class ComplexityMetaDataTests : IClassFixture<ComplexityMetaDataFixture>
     }
 
     [Fact]
-    public void MetaData_ShouldBe_Used()
+    public async Task MetaData_ShouldBe_Used()
     {
-        var result = _fixture.Analyze(@"
+        var result = await _fixture.AnalyzeAsync(@"
 query {
     hero { #2
         id #0
@@ -36,15 +35,16 @@ query {
         }
     }
 }
-");
+").ConfigureAwait(false);
+
         result.Complexity.ShouldBe(60);
         result.TotalQueryDepth.ShouldBe(4);
     }
 
     [Fact]
-    public void MetaData_With_Fragments_ShouldBe_Used()
+    public async Task MetaData_With_Fragments_ShouldBe_Used()
     {
-        var result = _fixture.Analyze(@"
+        var result = await _fixture.AnalyzeAsync(@"
 query {
     hero { #2
         id #0
@@ -71,7 +71,8 @@ fragment friendsIdName on Hero {
         name #3
     }
 }
-");
+").ConfigureAwait(false);
+
         result.Complexity.ShouldBe(136);
         result.TotalQueryDepth.ShouldBe(5);
     }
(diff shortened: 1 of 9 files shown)


SebastianStehle:
I would introduce a ComplexityAnalayzerMetadataExtensions  class and provide two methods to set and retrieve the metadata, so that you do not have to know the magic string.

Shane32:
How about a `float` rather than a `decimal`?

src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
```diff
@@ -0,0 +1,149 @@
...
+    public async Task<ComplexityResult> AnalyzeAsync(string query)
+    {
+        var result = await _executer.ExecuteAsync(o =>
+        {
+            o.Query = query;
+            o.RequestServices = _provider;
+        }).ConfigureAwait(false);
+        result.Errors.ShouldBeNull();
+
+        return _analyzer.Analyze(_documentBuilder.Build(query), _config.FieldImpact ?? 2f, _config.MaxRecursionCount, _schema);
```
sungam3r, line 145:
```suggestion
        return _analyzer.Validate(_documentBuilder.Build(query), _config, _schema);
```

sungam3r (changes requested):

Shane32:
> How about a `float` rather than a `decimal`?

Oh if it's `double` that's okay.  I mean `float` is probably sufficient, but I'm okay with `double`.  I must have misread it earlier.

src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
SlavaUtesinov: `Analyze` and `Validate` methods don't have the same return signature, the last returns `void`.

src/GraphQL/Validation/Complexity/ComplexityVisitor.cs
```diff
@@ -29,42 +53,46 @@
             var prevCurrentSubSelectionImpact = context.CurrentSubSelectionImpact;
             var prevCurrentEndNodeImpact = context.CurrentEndNodeImpact;
 
+            var fieldImpact = _visitor?.GetFieldDef()?.GetComplexityImpact() ?? context.AvgImpact;
+            var zeroImpact = fieldImpact == 0;
```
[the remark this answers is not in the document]
SlavaUtesinov: https://stackoverflow.com/questions/6598179/the-right-way-to-compare-a-system-double-to-0-a-number-int#comment7787303_6598240

src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
sungam3r: ok

sungam3r (approved):

sungam3r:
@Shane32 Merge if OK. In the future, I want to make a few more changes related to the calculation of complexity values.

src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
```diff
@@ -0,0 +1,149 @@
...
+    hero { #2
+        id #0
+        name #2
+        ...friendsRoot #132(4/2*66)
+    }
+}
+
+fragment friendsRoot on Hero {
+    friends { #3
+        ...friendsIdName #27(9/2*6)
```
Shane32, line 58: I don't understand what's going on here, but it doesn't seem correct.  It should just be `3*6` - where 3 is the value of the parent and 6 is the total of the fragment.  Can we fix this while we are making changes here?

src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    [Fact]
+    public async Task MetaData_With_Fragments_ShouldBe_Used()
+    {
+        var result = await _fixture.AnalyzeAsync(@"
+query {
+    hero { #2
+        id #0
+        name #2
+        ...friendsRoot #132(4/2*66)
```
Shane32, line 52: Similarly, this should be `2*66`, not `4/2*66`.

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
```diff
@@ -9,11 +10,11 @@ namespace GraphQL.Validation.Complexity
     public class ComplexityAnalyzer : IComplexityAnalyzer
     {
         /// <inheritdoc/>
-        public void Validate(GraphQLDocument document, ComplexityConfiguration complexityParameters)
+        public void Validate(GraphQLDocument document, ComplexityConfiguration complexityParameters, ISchema? schema = null)
         {
             if (complexityParameters == null)
                 return;
-            var complexityResult = Analyze(document, complexityParameters.FieldImpact ?? 2.0f, complexityParameters.MaxRecursionCount);
+            var complexityResult = Analyze(document, complexityParameters.FieldImpact ?? 2.0f, complexityParameters.MaxRecursionCount, schema);
```
Shane32, line 17: Why do we have a constant of 2.0f here?  Should it be 2.0f?  It seems like the default impact should be 1.0f.  Either way, perhaps we should convert this to a constant within this class.  For example:

```cs
const float DEFAULT_FIELD_IMPACT = 2.0f
```

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
```diff
@@ -9,11 +10,11 @@ namespace GraphQL.Validation.Complexity
     public class ComplexityAnalyzer : IComplexityAnalyzer
     {
         /// <inheritdoc/>
-        public void Validate(GraphQLDocument document, ComplexityConfiguration complexityParameters)
+        public void Validate(GraphQLDocument document, ComplexityConfiguration complexityParameters, ISchema? schema = null)
```
Shane32, line 13: If this is a breaking change anyway, perhaps we should make `schema` required.

Shane32 (commented):
The new code looks good, and I like the added tests.  I have a few questions before we merge, but they could be addressed in future PRs also.

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
SlavaUtesinov: https://github.com/graphql-dotnet/graphql-dotnet/blob/develop/src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs#L16

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
SlavaUtesinov: It still can work without scheme as before

src/GraphQL.Tests/Complexity/ComplexityMetaDataTests.cs
Shane32: Ok.  We can address it in a separate issue.  Thanks!

Shane32: #3164

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
sungam3r: @Shane32 Another observation from me - look at `GetImpactFromArgs`. If we have `id` argument then method returns 1 so `CurrentEndNodeImpact` will be less 2 times. Why so? I don't know.

sungam3r:
> My intention was to implement an ability to specify a specific field's complexity not to rewrite core calculation logic of complexity.

Agree. Complexity is not an easy area to design.

PR #3166, opened 2022-06-07
Convert IVariableVisitor to async

Shane32:
Sister to #3163


sungam3r (approved):
Looks like "async infection" 🤔

Shane32:
I’ll update the migration note before merging

PR #3170, opened 2022-06-09
Fix error message when no query is defined

Shane32:
Currently a `ARGUMENT_NULL` error is returned to the client when a query request arrives for a schema with no root query type.  This PR aligns the error message with the similar case for mutation and subscription types, until such time as we decide to disallow schemas without a root query type (in accordance with the spec).

Note that there is an upcoming RFC to the spec for a validation rule to address this condition:
https://github.com/graphql/graphql-spec/pull/955

### DIFF
--- a/src/GraphQL.Tests/Execution/ExecutionStrategyTests.cs
+++ b/src/GraphQL.Tests/Execution/ExecutionStrategyTests.cs
@@ -1,0 +1,27 @@
+using GraphQL.Types;
+
+namespace GraphQL.Tests.Execution;
+
+public class ExecutionStrategyTests
+{
+    [Theory]
+    [InlineData("query { test }", "Schema is not configured for queries")]
+    [InlineData("mutation { test }", "Schema is not configured for mutations")]
+    [InlineData("subscription { test }", "Schema is not configured for subscriptions")]
+    public async Task Refuses_Null_Root_Type(string query, string expectedErrorMessage)
+    {
+        var schema = new Schema();
+        schema.Initialize();
+        var executer = new DocumentExecuter();
+        var result = await executer.ExecuteAsync(new ExecutionOptions
+        {
+            Query = query,
+            Schema = schema,
+        }).ConfigureAwait(false);
+        result.Data.ShouldBeNull();
+        //todo: the document should fail validation and Executed should return false
+        //see: https://github.com/graphql/graphql-spec/pull/955
+        result.Executed.ShouldBeTrue();
+        result.Errors.ShouldHaveSingleItem().Message.ShouldBe(expectedErrorMessage);
+    }
+}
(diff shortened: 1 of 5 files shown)


Shane32:
Problem: this means that without a query type, introspection queries cannot be performed anymore with this PR.

Shane32:
We should consider revising our schema introspection to require a query root type, as the spec requires -- and changing all of the tests as required.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -50,6 +50,11 @@ protected virtual IObjectGraphType GetOperationRootType(ExecutionContext context
             {
                 case OperationType.Query:
                     type = context.Schema.Query;
+                    // note that per the GraphQL specification, a query root type is required; but currently
+                    // the schema validation check is disabled to allow for test schemas
```
sungam3r, line 54: Where is this check? I don't remember but I did something with it.

> to allow for test schemas 

I think not only for this, don't remember 😞

sungam3r:
> We should consider revising our schema introspection to require a query root type, as the spec requires

I remember how I read discussions regarding it ). Hot topic.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: https://github.com/graphql-dotnet/graphql-dotnet/blob/b6ef5f2d94dd1f22ac6cd3cad5f3bb98ee16b4e0/src/GraphQL/Types/Collections/SchemaTypes.cs#L231-L233

Shane32:
@sungam3r I fixed the tests.  I see you already approved this PR, but I'll wait to merge it until tomorrow, so you have a chance to see my additional changes if desired.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: Ah, yes. Exactly this place. Thanks.

src/GraphQL.Tests/Execution/ExecutionStrategyTests.cs
```diff
@@ -0,0 +1,27 @@
...
+        var schema = new Schema();
+        schema.Initialize();
+        var executer = new DocumentExecuter();
+        var result = await executer.ExecuteAsync(new ExecutionOptions
+        {
+            Query = query,
+            Schema = schema,
+        }).ConfigureAwait(false);
+        result.Data.ShouldBeNull();
+        //todo: the document should fail validation and Executed should return false
```
sungam3r, line 22:
```suggestion
        //TODO: the document should fail validation and Executed should return false
```

sungam3r (approved):
I reviewed again. 👍 

This dummy field for query looks pointless though. I'm still hope that spec wg will find better solution.

PR #3173, opened 2022-06-10
Replace DirectiveGraphType with Directive

anvu03:
Replace DirectiveGraphType with Directive in Directive.md since DirectiveGraphType has been renamed to Directive.


sungam3r (approved):

sungam3r:
Thanks.

micheltol:
@sungam3r The live-docs do not reflect this change yet. Am I missing something or does something else still need to be done? Can I do something to speed up the process?

PR #3184, opened 2022-06-17
Unify namespaces for all extension methods for IGraphQLBuilder

sungam3r:
I suppose that DI should be discoverable as much as possible so usual I place extension methods in such namespaces that caller already typed so IntelliSense shows all stuff when one types '.'. For this case it is either `GraphQL.DI` where `IGraphQLBuilder` resides, or `GraphQL`. Core package already has many extension methods for `IGraphQLBuilder` in `GraphQL` namespace so I chose it. I'm not sure about these changes still.


Shane32:
I like the concept, and I like that they are all within `GraphQL`, not `Microsoft.Extensions.DependencyInjection`.  However, I think each `GraphQLBuilderExtensions` class should be named different so they can be individually accessed if needed.  For instance `MicrosoftDIExtensions` and `SystemTextJsonExtensions` or something like that.  Similar to how `AddHttpContextAccessor` is located within `Microsoft.Extensions.DependencyInjection.HttpServiceCollectionExtensions` and `AddLogging` is within `Microsoft.Extensions.DependencyInjection.LoggingServiceCollectionExtensions`.

Alternatively, they could all be under `GraphQL.DI`, but if we are trying to simplify things, let's go with having everything under `GraphQL`.  It makes more sense from a user's perspective.

Shane32:
Originally I put all the extension methods within the namespace of the individual nuget package (which is how it is now), but now after a year it seems evident to me that it was not the best choice.  There is little benefit and places undue burden on users.

sungam3r:
> each GraphQLBuilderExtensions class should be named different so they can be individually accessed if needed.

I wrote TODO about that somewhere in codebase.

sungam3r:
So I will rename those classes.

Shane32 (approved):

PR #3186, opened 2022-06-17
Optimize hashing for APQ

sungam3r:
Saves 2 `byte[]` allocations:
1. `Encoding.UTF8.GetBytes`: string -> byte[]
2. return value of `ComputeHash` (len = 32)


sungam3r:
ping @SlavaUtesinov

Shane32 (approved):

PR #3189, opened 2022-06-20
Fix SystemTextJson serializer to allow custom json converters for any data type

### LINKED ISSUE #2257: How to change output of decimal scalar (removing tailing zeros)

Shane32:
Fixes #2257 and allows any registered JsonConverter to work properly.

It comes at a cost of an additional 50% of serialization time.  For the "middle" query below, this represents a half of a millionth of a second of processing time.  Even for the introspection query, it's only 50 millionths of a second.

"Alt" is the new code

|            Method |          Code |         Mean |     Error |    StdDev |  Gen 0 | Allocated |
|------------------ |-------------- |-------------:|----------:|----------:|-------:|----------:|
|    SystemTextJson | Introspection |  79,349.3 ns | 159.36 ns | 141.27 ns |      - |     161 B |
| SystemTextJsonAlt | Introspection | 124,444.6 ns | 467.13 ns | 436.95 ns |      - |     168 B |
|    SystemTextJson |        Middle |   1,442.3 ns |   2.38 ns |   1.98 ns | 0.0229 |     152 B |
| SystemTextJsonAlt |        Middle |   2,075.8 ns |   3.94 ns |   3.68 ns | 0.0229 |     152 B |
|    SystemTextJson |         Small |     397.0 ns |   1.11 ns |   0.98 ns | 0.0238 |     152 B |
| SystemTextJsonAlt |         Small |     535.5 ns |   2.46 ns |   2.18 ns | 0.0238 |     152 B |

### DIFF
--- a/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
+++ b/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
@@ -34,7 +34,7 @@ namespace GraphQL.SystemTextJson
                 }
                 else
                 {
-                    WriteValue(writer, result.Data, options);
+                    JsonSerializer.Serialize(writer, result.Data, options);
                 }
             }
         }
@@ -43,7 +43,7 @@ namespace GraphQL.SystemTextJson
         {
             if (node is ValueExecutionNode valueExecutionNode)
             {
-                WriteValue(writer, valueExecutionNode.ToValue(), options);
+                JsonSerializer.Serialize(writer, valueExecutionNode.ToValue(), options);
             }
             else if (node is ObjectExecutionNode objectExecutionNode)
             {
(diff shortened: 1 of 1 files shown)


Shane32:
> Interesting change 😄 . Any test to verify decimal bug fixed?

I'll do that.

PR #3190, opened 2022-06-20
Handle both string and integer versions

sungam3r:
https://www.apollographql.com/docs/apollo-server/performance/apq/:

изображение


sungam3r:
ping @SlavaUtesinov

Shane32 (approved):

PR #3192, opened 2022-06-21
Fix NestedFragmentsComparer

### LINKED ISSUE #3191: Deeply Nested Fragments Fail Based Upon Order of Fragments In Query
## Description

Using GraphQL DotNet 5.0+ (tried with 5.1 and 5.3), a query like the below fails (simplified example from a more complex project):

```query fragmentTesting
{
  car(id: 1)
  {
    ...lastUpdated
    ...optionsOnCar
  }
}

fragment optionsOnCar on Car
{
  options
  {
    edges
    {
      node
      {
        ...optionDetail
      }
    }
  }
}

fragment optionPrice on Option
{
  price
}

fragment lastUpdated on Car
{
  updatedAt
}

fragment optionDetail on Option
{
  name
  ...optionPrice
  optionContents(first: 9999999)
  {
    edges
    {
      node
      {
        optionContent
        {
          name
        }
      }
    }
  }
}
```

With an error that begins with, 

> "System.Collections.Generic.KeyNotFoundException: The given key 'optionDetail' was not present in the dictionary.\r\n   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)\r\n   at GraphQL.Validation.Complexity.ComplexityVisitor.VisitFragmentSpreadAsync(GraphQLFragmentSpread fragmentSpread, AnalysisContext context) in /_/src/GraphQL/Validation/Complexity/ComplexityVisitor.cs:line 79"

The full error goes on a while, but it appears related to  #3030 Nested Fragments.

However, when we re-arrange the fragments to move the lastUpdated fragment to the beginning or the end of the set of fragments, the query succeeds as expected. If we try a query that looks like this:

```query fragmentTesting
{
  car(id: 1)
  {
    ...lastUpdated
    ...optionsOnCar
  }
}

fragment optionsOnCar on Car
{
  options
  {
    edges
    {
      node
      {
        ...optionDetail
      }
    }
  }
}

fragment optionPrice on Option
{
  price
}

fragment optionDetail on Option
{
  name
  ...optionPrice
  optionContents(first: 9999999)
  {
    edges
    {
      node
      {
        optionContent
        {
          name
        }
      }
    }
  }
}

fragment lastUpdated on Car
{
  updatedAt
}
```

Where the nested fragments are in an unordered, but unbroken group, the query succeeds every time.

### Steps to reproduce
Unfortunately, I do not yet have reliable steps to reproduce from a simple example yet.

### Expected result

I would expect fragment order in the query to be irrelevant. A query would always parse and execute -- throwing errors only from the library consumer's code -- or always return an error from the GraphQL Dotnet subsystems because the query was not well-formed.

### Actual result

Fragment order seems to matter, which is unexpected. It is also a departure from previous GraphQL Dotnet versions from 2 to 4, as this error only appeared in queries as we upgrade from 4 to 5.

### Environment

OS: Windows 10
GraphQL DotNet Version: 5.1 and 5.3
Dotnet Core version: 6.0.3

sungam3r:
fixes #3191 

Comparison sequence before fix:

optionPrice optionsOnCar -1
lastUpdated optionsOnCar 0
optionDetail lastUpdated 0
optionPrice optionsOnCar -1
lastUpdated optionsOnCar 0
optionDetail lastUpdated 0

Result:

optionPrice
optionsOnCar
lastUpdated
optionDetail

Comparison sequence after fix:

optionPrice optionsOnCar -1
lastUpdated optionsOnCar -1
lastUpdated optionPrice 0
optionDetail optionsOnCar -1
optionDetail lastUpdated 1

Result:

optionPrice
lastUpdated
optionDetail
optionsOnCar

### DIFF
--- a/src/GraphQL.Tests/Complexity/ComplexityTests.cs
+++ b/src/GraphQL.Tests/Complexity/ComplexityTests.cs
@@ -143,6 +143,65 @@ fragment comparisonFields on Character {
         withFrag.TotalQueryDepth.ShouldBe(woFrag.TotalQueryDepth);
     }
 
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/3191
+    [Fact]
+    public void nested_fragments_2()
+    {
+        var result = AnalyzeComplexity(@"
+{
+  car(id: 1)
+  {
+    ...lastUpdated
+    ...optionsOnCar
+  }
+}
+
+fragment optionsOnCar on Car
+{
+  options
+  {
+    edges
+    {
+      node
+      {
+        ...optionDetail
+      }
+    }
+  }
+}
+
+fragment optionPrice on Option
+{
+  price
+}
+
+fragment lastUpdated on Car
+{
+  updatedAt
+}
+
+fragment optionDetail on Option
+{
+  name
+  ...optionPrice
+  optionContents(first: 9999999)
+  {
+    edges
+    {
+      node
+      {
+        optionContent
+        {
+          name
+        }
+      }
+    }
+  }
+}");
+
+        result.Complexity.ShouldBe(1839999841); // WOW! :)
+    }
+
     [Fact]
     public void absurdly_huge_query()
     {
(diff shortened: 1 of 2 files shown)


sungam3r:
The fix was ready almost immediately after I reproduced bug, but I needed to move away on household chores.

Shane32 (approved):

PR #3197, opened 2022-06-24
Add codeowners

sungam3r:


sungam3r:
Also I found this https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-pull-request-reviews-before-merging particularly
> Optionally, you can choose to require reviews from code owners. If you do, any pull request that affects code with a code owner must be approved by that code owner before the pull request can be merged into the protected branch.

but can not find a way to turn it on. I doubt only repo admin can do such changes:
> Repository administrators can require ...

ping @Shane32 @joemcbride

Shane32:
I don't understand the purpose here.  There's only two of us active @sungam3r , and both of us look at almost all PRs.  We have not had any issues between us.  I almost always wait for you to review before merging, and you likewise do the same.  Also we each sometimes push an obvious PR when there is a strong compelling reason to do so.

So I certainly wouldn't add restrictions to the repo at this time.  We could add the file, but why ???

When we have multiple team members that take care of different parts of the repo, then this will be a useful feature.

sungam3r:
I was curious how it would work. I came across this file in another repo, read the documentation and decided to make PR. Small automation, which in fact does not change anything serious.

Shane32:
> I was curious how it would work.

I wouldn't use GraphQL.NET as a place to run experiments.

> Small automation

To do what?  I'm still at a loss for how this can be a benefit.

sungam3r:
> To do what?

изображение

PR #3202, opened 2022-06-27
Fix complexity analyzer test

Shane32:
After merging `master` into `develop`, this test broke.  @sungam3r Please verify this change is correct.

The test was added into `master` recently in
- #3192 

It is probably conflicting with this code added to `develop`:
- #3159


sungam3r (approved):
This little difference doe not matter.

PR #3203, opened 2022-06-27
Add global GraphQL Attributes

### LINKED ISSUE #3198: Feature request: global GraphQL attributes
I propose adding within global options a list of GraphQL attributes which are applied to all types anywhere they could otherwise be applied.  This typically applies to schema-first and type-first, of course.

For an example use case, users could add a global attribute which converts query arguments of type DbContext to pull from services, like this:

```cs
public class DbContextFromServicesAttribute : GraphQLAttribute
{
    {
        if (typeof(TParameterType) == typeof(DbContext))
            argumentInformation.SetDelegate(context => (context.RequestServices ?? throw new MissingRequestServicesException())
                .GetRequiredService<TParameterType>());
    }
}

// in Startup.cs or Program.cs
static void Startup()
{
    GlobalSwitches.GlobalAttributes.Add(new DbContextFromServicesAttribute());
}
```

Although the above attribute is applied to other types as well, it would not affect them since it would be executing other overloads of `Modify()` which execute no code.

Ideally this would be configured per schema, but that is not possible with the current GraphQL.NET design, where graph types do not have access to schema properties in their constructors.

Shane32:
Closes #3198


docs2/site/docs/getting-started/global-switches.md
```diff
@@ -18,3 +18,11 @@ The current options list is presented below:
...
 
 For a detailed description of each option, see [GlobalSwitches](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/GlobalSwitches.cs).
+
+# Global GraphQL Attributes
+
+You may apply `GraphQLAttribute`s globally by applying them to the module or assembly.
+Code that utilizes `GraphQLAttribute`s, such as auto-registering graph types, will scan
+the CLR type's owning module and assembly and apply any globally-defined attributes found.
+Globally-defined attributes may be configured to execute before or after individually-specified
+attributes by changing the `GraphQLAttribute.Priority` value.
```
sungam3r, line 28:
```suggestion
attributes by changing the `GraphQLAttribute.Priority` value. When writing your globally-defined
attribute do not forget to mark it with `AttributeUsage` attribute and include `AttributeTargets.Assembly`
or `AttributeTargets.Module`.
```

docs2/site/docs/migrations/migration6.md
```diff
@@ -54,6 +54,37 @@ Field<IntGraphType>("id").WithComplexityImpact(123);
...
+will apply to all applicable CLR types within the module or assembly.
+
+This allow global changes to how the schema builder or auto-registering graph type builds graph types,
+field types or field arguments.
+
+For an example use case, users could add a global attribute which converts query arguments of type
+`DbContext` to pull from services, like this:
+
+```csharp
+public class DbContextFromServicesAttribute : GraphQLAttribute
```
sungam3r, line 69: I think modified `AttributeUsage` is required here.

sungam3r (approved):

PR #3205, opened 2022-06-28
AutoRegisteringObjectGraphType: Add recognition for inherited methods

psmolinsky:


Shane32:
@psmolinsky Looks good !  One last change: can you add a section under "new features" of the latest migration notes that describes the changes?  You'll need to merge `develop` into your branch first, and then the file will be here:

https://github.com/psmolinsky/graphql-dotnet/blob/AutoRegisteringObjectGraphType-inherited-methods/docs2/site/docs/migrations/migration6.md

It can be just a simple bullet point with a short note describing the added functionality.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -225,26 +225,27 @@ protected virtual ArgumentInformation GetArgumentInformation<TParameterType>(Fie
         /// <summary>
         /// Returns a list of properties, methods or fields that should have fields created for them.
         /// <br/><br/>
-        /// Unless overridden, returns a list of public instance readable properties (including properties declared on
-        /// inherited classes) and public instance methods (excluding methods declared on inherited classes) that
+        /// Unless overridden, returns a list of public instance readable properties and public instance methods
+        /// (including properties and methods declared on inherited classes) that
         /// do not return <see langword="void"/> or <see cref="Task"/>.
```
sungam3r, line 230: It's would be better to clarify tat part about void and Task corresponds only for methods.

sungam3r (approved):

sungam3r:
@Shane32 milestone 7 I think

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
psmolinsky: So should I change it? It would require a new PR, right?

```
/// Unless overridden, returns a list of public instance readable properties and public instance methods
/// that do not return <see langword="void"/> or <see cref="Task"/>
/// including properties and methods declared on inherited classes.
```

sungam3r: I would change to stste this explicitly. Yes, small PR that I'll merge ASAP.

PR #3213, opened 2022-07-05
Use SourceFieldResolver for top-level nodes of subscriptions

sungam3r:
Allows to keep `Resolver` property unspecified (`null`) for top-level nodes of subscriptions since resolvers for such nodes always return `source` object. The goal - simplify field configuration on the caller side.

### DIFF
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -1141,6 +1141,7 @@ namespace GraphQL.Execution
                 "fieldType"})]
         public virtual System.Collections.Generic.Dictionary<string, System.ValueTuple<GraphQLParser.AST.GraphQLField, GraphQL.Types.FieldType>>? GetSubFields(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
         protected virtual System.Threading.Tasks.Task<bool> ProcessNodeUnhandledExceptionAsync(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node, System.Exception ex) { }
+        protected virtual GraphQL.Resolvers.IFieldResolver SelectResolver(GraphQL.Execution.ExecutionNode node, GraphQL.Execution.ExecutionContext context) { }
         protected virtual System.Threading.Tasks.Task SetArrayItemNodesAsync(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ArrayExecutionNode parent) { }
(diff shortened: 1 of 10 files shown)


src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -1141,6 +1141,7 @@ namespace GraphQL.Execution
                 "fieldType"})]
         public virtual System.Collections.Generic.Dictionary<string, System.ValueTuple<GraphQLParser.AST.GraphQLField, GraphQL.Types.FieldType>>? GetSubFields(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node) { }
         protected virtual System.Threading.Tasks.Task<bool> ProcessNodeUnhandledExceptionAsync(GraphQL.Execution.ExecutionContext context, GraphQL.Execution.ExecutionNode node, System.Exception ex) { }
+        protected virtual GraphQL.Resolvers.IFieldResolver SelectResolver(GraphQL.Execution.ExecutionNode node, GraphQL.Execution.ExecutionContext context) { }
```
sungam3r, line 1144: I can make SelectResolver private and SourceFieldResolver internal.

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -342,7 +342,7 @@ public virtual FieldType AddField(FieldType fieldType)
             string name,
             string? description = null,
             QueryArguments? arguments = null,
-            Func<IResolveFieldContext<TSourceType>, object?>? resolve = null,
+            Func<IResolveFieldContext<TSourceType>, object?>? resolve = null, // TODO: remove?
```
sungam3r, line 345: I doubt this parameter makes sense. I see no usages of these two methods and for top-level node of subscription it should always return source object (I think). I propose to remove `resolve` parameter or even remove both methods since we have `ResolveStream/ResolveStreamAsync` extensions and these two methods IMO only clutter public API.

sungam3r: Remove in develop of course after merging this one into master.

Shane32 (approved):
Looks good as proposed

sungam3r:
What about FieldSubscribe methods?

Shane32:
I’d remove the argument as you suggested with a todo note. I really have no idea if people use the methods or not; I’d leave them probably as I don’t see it hurts anything.

Shane32:
(And it matches Field and FieldAsync, right?)

sungam3r:
Fluent extensions allow to chain needed configurations without combinatorial overloads.

Shane32:
Let's review all the API `Field...` methods as a whole, then, and thoughtfully consider the anticipated impact any changes would have on users.

PR #3217, opened 2022-07-10
Update publish-documentation.yml

Shane32:


.github/workflows/publish-documentation.yml
```diff
@@ -18,29 +18,40 @@ jobs:
     steps:
       - name: Checkout source
         uses: actions/checkout@v3
+        with:
+          path: main
```
sungam3r, line 22: main?

Shane32: arbitrary folder name because we need two different git repos downloaded.  i called one 'main' and one 'published'...

sungam3r: Ah, dest directory, OK.

.github/workflows/publish-documentation.yml
```diff
@@ -18,29 +18,40 @@ jobs:
...
-        with:
-          repository: graphql-dotnet/graphql-dotnet.github.io
-          branch: master
-          github_token: '${{ secrets.GITHUB_TOKEN  }}'
-          github_pat: '${{ secrets.ORG_PAT_TOKEN }}'
-          source_folder: docs2/public
+      - name: Commit docs
+        shell: bash
+        run: |
+          rsync -a --quiet --delete --exclude ".git" "main/docs2/public/" "published/" || exit 1
```
Shane32, line 51: I don't use rsync, but I copied this from `seanmiddleditch/gha-publish-to-git`, the previously used action here.

sungam3r:
So your changes are: 
1. Checkout source repo (our GraphQL.NET repo).
2. Checkout dest (repo for docs).
3. Run special command to build docs (yarn/gatsby as before) from source.
4. Bulk load (with clean/ovewrite) all generated files from source into dest.
5. Commit all changes in dest repo and push.

Right?

I see that seanmiddleditch/gha-publish-to-git@v2 does +- the same but is a subject of more security restrictions.

.github/workflows/publish-documentation.yml
sungam3r: I see.

Shane32:
Yeah that's about it.

sungam3r:
I suggest to add this yaml file into `paths` property.

PR #3222, opened 2022-07-11
Fix complexity analyzer

### LINKED ISSUE #3221: Fragment Cycle Detected Unexpectedly
Hello again! I'm still having a bit of an issue with fragments, but this one is new.

## Description

A query set up like the below returns an error of "Fragment Cycle Detected." Simplified example from a more complex query:

```
query carFragmentTest
{
  car(id: 1)
  {
    name
    ... carInfo
  }
}

fragment carInfo on Car
{
    ...pricing
    ... detail
}

fragment pricing on Car
{
  msrp
}

fragment detail on Car
{
  ... furtherDetail
}

fragment furtherDetail on Car
{
  ... pricing
}
```

Produces an error that looks like this:

> "System.InvalidOperationException: Fragments dependency cycle detected!

Note the pricing fragment is used twice, one at the first fragment level and another at the third fragment level. The error is unexpected, because pricing terminates with the field msrp, but that could also be my misconception of what a fragment cycle means and the difficulties of resolving them under the hood. The error persists if the pricing fragment is moved from the carInfo fragment to the detail fragment; location does not appear to matter so long as it is valid and not on the top level type.

If we move the Pricing fragment just one level up to the main query  like this:

```
query carFragmentTest
{
  car(rowId: 1)
  {
    name
    ... pricing
    ... carInfo
  }
}

fragment carInfo on Car
{
    ... detail
}

fragment pricing on Car
{
  msrp
}

fragment detail on Car
{
  ... furtherDetail
}

fragment furtherDetail on Car
{
  ... pricing
}
```

It works correctly, every time. I am unsure if this is a bug or my misunderstanding of the limitations of fragments in the GraphQL spec. If this cannot be resolved, would it be possible for the error to name the offending fragment to assist in diagnosing the issue?

### Steps to reproduce

Create a fragment chain on the same type, where one fragment within the chain is spread across two different levels of a fragment chain.

### Expected result

Since the pricing fragment terminates, I would expect it to resolve successfully to the msrp field, similarly to how it works when the pricing fragment is moved to the main type.

### Actual result

> System.InvalidOperationException: Fragments dependency cycle detected!

### Environment

Windows 10
Dotnet Core 6.0.2
GraphQL Dotnet 5.3.2

Shane32:
Fixes #3221 

Two tests added (both failed previously, both pass now)

### DIFF
--- a/src/GraphQL.Tests/Complexity/ComplexityTests.cs
+++ b/src/GraphQL.Tests/Complexity/ComplexityTests.cs
@@ -249,6 +249,75 @@ query fragmentTest
         //result.Complexity.ShouldBe(12345);
     }
 
+
+    [Fact]
+    public void duplicate_fragment_ok()
+    {
+        var query = @"
+query carFragmentTest
+{
+  car(id: 1)
+  {
+    name
+    ... carInfo
+    ... carInfo
+  }
+}
+
+fragment carInfo on Car
+{
+    ...pricing
+    ...pricing
+}
+
+fragment pricing on Car
+{
+  msrp
+}
+";
+        var result = AnalyzeComplexity(query);
+
+        //result.Complexity.ShouldBe(12345);
+    }
+
+    [Fact]
+    public void no_fragment_cycle()
+    {
+        var query = @"
+query carFragmentTest
+{
+  car(id: 1)
+  {
+    name
+    ... carInfo
+  }
+}
+
+fragment carInfo on Car
+{
+    ...pricing
+    ... detail
+}
+
+fragment pricing on Car
+{
+  msrp
+}
+
+fragment detail on Car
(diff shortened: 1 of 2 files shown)


Shane32:
@sungam3r I read your code again and noticed that the same fragment could be added multiple times to the `List` within `BuildDependencies` on line 135 but not removed on line 77.  This fixes that issue and provides a test.  After doing so, issue #3221 seemed to be solved, but it is not immediately apparent why to me.  Perhaps you would like to investigate issue #3221 further and see if my patch does indeed solve #3221 or if there is a further fix necessary.

sungam3r (approved):
Yep, silly mistake. Using `HashSet` is fine here.

PR #3226, opened 2022-07-12
Add IResolveFieldContext.User

Shane32:
See https://github.com/graphql-dotnet/server/issues/314

Matches `HttpContext.User` except that it is nullable.

It might be better to default `User` to a static implementation of an anonymous user, sorta like ASP.NET Core does, so it does not need to be nullable.

This feature allows separation of the execution which defines the user from the validation rule which validates the request against the user, rather than requiring AsyncLocal to provide this mechanism.

### DIFF
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -250,3 +250,7 @@ were moved into `GraphQL` namespace. Also class names were changed:
 - `GraphQL.SystemTextJson.GraphQLBuilderExtensions` -> `GraphQL.SystemTextJsonGraphQLBuilderExtensions`
 
 This change was done for better discoverability and usability of extension methods when configuring DI.
+
+### 8. `IResolveFieldContext.User` property added
+
+Custom implementations of `IResolveFieldContext` must implement the new `User` property.
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -96,7 +96,7 @@ Note that global attributes may also be added to the `GlobalSwitches.GlobalAttri
 
 ### 8. `ExecutionOptions.User` property added and available to validation rules and field resolvers
 
-You may pass an `IPrincipal` instance into `ExecutionOptions` and it will be fed through to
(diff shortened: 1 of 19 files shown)


Shane32:
I think this PR will be a very valuable addition to GraphQL.NET.  I'm not sure if `User` should be `IPrincipal` or `ClaimsPrincipal`.  You can't do much with an `IPrincipal`, but then again, it could be implemented by any type of authentication token, even if it did not have 'claims'.  `ClaimsPrincipal` was added in .NET Framework 4.5 and currently seems to be the primary mechanism for passing around identity tokens in .NET.  It's used exclusively by ASP.NET Core, for example, and never `IPrincipal`.  As such, I'm changing the definition of `User` to `ClaimsPrincipal`, as this seems to be more useful with today's software.

Shane32:
This PR will be useful for the `GraphQL.Authorization` project as well, which also uses `ClaimsPrincipal` for identities.

docs2/site/docs/migrations/migration7.md
```diff
@@ -94,6 +94,16 @@ adjust the return value of the `Priority` property of the attribute.
 
 Note that global attributes may also be added to the `GlobalSwitches.GlobalAttributes` collection.
 
+### 8. `ExecutionOptions.User` property added and available to validation rules and field resolvers
+
+You may pass an `IPrincipal` instance into `ExecutionOptions` and it will be fed through to
```
sungam3r, line 99:
```suggestion
You may pass a `ClaimsPrincipal` instance into `ExecutionOptions` and it will be fed through to
```

docs2/site/docs/migrations/migration7.md
```diff
@@ -94,6 +94,16 @@ adjust the return value of the `Priority` property of the attribute.
...
+`ValidationContext.User`, `IExecutionContext.User` and `IResolveFieldContext.User` so the value
+is accessible by validation rules, document listeners, field middleware and field resolvers.
+
+This property is similar in nature to the ASP.NET Core `HttpContext.User` propery, not being
```
sungam3r, line 103:
```suggestion
This property is similar in nature to the ASP.NET Core `HttpContext.User` property, not being
```

sungam3r (approved):

Issue #3229, opened 2022-07-15
Could not load type 'GraphQL.Server.Transports.AspNetCore.IGraphQLRequestDeserializer


eddyjl077:
System.TypeLoadException: 'Could not load type '### **GraphQL.Server.Transports.AspNetCore.IGraphQLRequestDeserializer**' from assembly 'GraphQL.Server.Transports.AspNetCore, Version=**6.1.0.0**, Culture=neutral, PublicKeyToken=null'.'

![image](https://user-images.githubusercontent.com/20169438/179227254-e94a9a73-dd2c-4ea8-8a3f-2d7843080baa.png)

Shane32:
Please provide a list of nuget packages you have referenced.

Shane32:
Please remove these references:
```
    <PackageReference Include="GraphQL.Server.Core" Version="5.2.1" />
    <PackageReference Include="GraphQL.Server.Transports.AspNetCore.SystemTextJson" Version="5.2.1" />
    <PackageReference Include="GraphQL.Server.Transports.AspNetCore.NewtonsoftJson" Version="5.2.1" />
```

PR #3231, opened 2022-07-15
Fix package versioning

Shane32:
This, for example, restricts GraphQL.DataLoader to use only GraphQL 5.x versions.  Specifically, the matching version that was published, and any version up to but excluding the next major version number.  This will prevent GraphQL.DataLoader 5.3.3 from being compatible with GraphQL 7.0.0, and indirectly prevents #3229 from happening in the future.  A similar change should be made in the server repo.

See:
- https://github.com/NuGet/Home/issues/5556
- https://github.com/NuGet/Home/issues/5525

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -15,6 +15,7 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
   </ItemGroup>
 
+  <!--Workaround for https://github.com/NuGet/Home/issues/5556 and https://github.com/NuGet/Home/issues/5525-->
   <Target Name="UseExplicitPackageVersions" BeforeTargets="GenerateNuspec">
     <ItemGroup>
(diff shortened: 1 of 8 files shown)


src/Directory.Build.targets
```diff
@@ -15,4 +15,19 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
   </ItemGroup>
 
+  <Target Name="UseExplicitPackageVersions" BeforeTargets="GenerateNuspec">
```
sungam3r, line 18:
```suggestion
  <!--Workaround for https://github.com/NuGet/Home/issues/5556 and https://github.com/NuGet/Home/issues/5525-->
  <Target Name="UseExplicitPackageVersions" BeforeTargets="GenerateNuspec">
```

sungam3r: Without an explanatory comment, an ordinary developer (i.e not MSBuild guru) is completely impossible to understand what is happening here.

sungam3r (approved):

PR #3242, opened 2022-07-21
Use fluent Field APIs instead of FieldXXX overloads

sungam3r:
First part of changes from #3237

### DIFF
--- a/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
@@ -1768,7 +1768,7 @@ namespace GraphQL.Types
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field(string name, GraphQL.Types.IGraphType type) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field(string name, System.Type type) { }
         public GraphQL.Types.FieldType Field(System.Type type, string name, string? description = null, GraphQL.Types.QueryArguments? arguments = null, System.Func<GraphQL.IResolveFieldContext<TSourceType>, object?>? resolve = null, string? deprecationReason = null) { }
-        public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field<TGraphType>(string name = "default")
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field<TGraphType>(string name)
             where TGraphType : GraphQL.Types.IGraphType { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TProperty> Field<TProperty>(System.Linq.Expressions.Expression<System.Func<TSourceType, TProperty>> expression, bool nullable = false, System.Type? type = null) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TProperty> Field<TProperty>(string name, System.Linq.Expressions.Expression<System.Func<TSourceType, TProperty>> expression, bool nullable = false, System.Type? type = null) { }
--- a/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
@@ -1766,7 +1766,7 @@ namespace GraphQL.Types
         protected virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> CreateBuilder<TReturnType>(GraphQL.Types.IGraphType type) { }
         protected virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> CreateBuilder<TReturnType>(System.Type type) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field(string name, GraphQL.Types.IGraphType type) { }
-        public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field(System.Type type, string name = "default") { }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field(string name, System.Type type) { }
         public GraphQL.Types.FieldType Field(System.Type type, string name, string? description = null, GraphQL.Types.QueryArguments? arguments = null, System.Func<GraphQL.IResolveFieldContext<TSourceType>, object?>? resolve = null, string? deprecationReason = null) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field<TGraphType>(string name = "default")
             where TGraphType : GraphQL.Types.IGraphType { }
--- a/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
@@ -834,12 +834,16 @@ namespace GraphQL.Builders
     {
         protected FieldBuilder(GraphQL.Types.FieldType fieldType) { }
         public GraphQL.Types.FieldType FieldType { get; }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Argument(System.Type type, string name, System.Action<GraphQL.Types.QueryArgument>? configure = null) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Argument<TArgumentGraphType>(string name, System.Action<GraphQL.Types.QueryArgument>? configure = null)
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Argument<TArgumentGraphType>(string name, string? description, System.Action<GraphQL.Types.QueryArgument>? configure = null)
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Argument<TArgumentGraphType, TArgumentType>(string name, string? description, TArgumentType? defaultValue = default, System.Action<GraphQL.Types.QueryArgument>? configure = null)
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Arguments(params GraphQL.Types.QueryArgument[] arguments) { }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Arguments(System.Collections.Generic.IEnumerable<GraphQL.Types.QueryArgument> arguments) { }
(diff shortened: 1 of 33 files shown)


src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
```diff
@@ -1757,8 +1763,12 @@ namespace GraphQL.Types
             where TNodeType : GraphQL.Types.IGraphType
             where TEdgeType : GraphQL.Types.Relay.EdgeType<TNodeType>
             where TConnectionType : GraphQL.Types.Relay.ConnectionType<TNodeType, TEdgeType> { }
+        protected virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> CreateBuilder<TReturnType>(GraphQL.Types.IGraphType type) { }
+        protected virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> CreateBuilder<TReturnType>(System.Type type) { }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field(string name, GraphQL.Types.IGraphType type) { }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field(System.Type type, string name = "default") { }
```
Shane32, line 1769: What's the thought process behind having differently ordered arguments here?

src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
```diff
@@ -1757,8 +1763,12 @@ namespace GraphQL.Types
...
         public GraphQL.Types.FieldType Field(System.Type type, string name, string? description = null, GraphQL.Types.QueryArguments? arguments = null, System.Func<GraphQL.IResolveFieldContext<TSourceType>, object?>? resolve = null, string? deprecationReason = null) { }
-        public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field<TGraphType>()
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, object> Field<TGraphType>(string name = "default")
```
Shane32, line 1771: Again, just a quick note that this causes a breaking change, as `Field<IntGraphType>("test")` previously returned FieldType and now will return a field builder.  Probably rather rare that this change would affect anyone.

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -139,6 +141,25 @@ protected FieldBuilder(FieldType fieldType)
         public virtual FieldBuilder<TSourceType, TReturnType> ResolveAsync(Func<IResolveFieldContext<TSourceType>, Task<TReturnType?>> resolve)
             => Resolve(new FuncFieldResolver<TSourceType, TReturnType>(context => new ValueTask<TReturnType?>(resolve(context))));
 
+        /// <inheritdoc cref="Resolve(IFieldResolver)"/>
+        public virtual FieldBuilder<TSourceType, TReturnType> Resolve(Delegate? resolve)
```
Shane32, line 145: I'm concerned about this.  This means that any delegate might unintentionally compile:

```cs
    public DroidType(StarWarsData data)
    {
        Field<StringGraphType>("help")
            .Returns<bool>()
            .Resolve(Test);
    }

    protected virtual string Test(IResolveFieldContext<Droid> context)
    {
        return "help"; // expecting bool
    }
```

Without the `Delegate` overload, this would have been a compilation error requiring a `bool` return type.

Also note that my concern does not affect inline functions; apparently they cannot resolve to `Delegate`.

Thoughts:
- Is this a "feature" that you can just pass any method to `Resolve`?
- Should we verify the return type and/or arguments against those defined in the builder?
- Should we automatically add arguments that were not defined with `Argument()`?
- Can we infer the graph type of the field?  (the underlying code already has this ability)

For example:

```cs
    public DroidType(StarWarsData data)
    {
        Field("help").Resolve(Help);
    }

    private Character Help(Assistance assistanceNeeded)
    {
        if (assistanceNeeded == Assistance.Fighting)
            return data.Find("Chewie");
        if (assistanceNeeded == Assistance.Force)
            return data.Find("Luke");
    }
```

Taking this thought process a step further, I've always been dissatisfied that the field builders ask for the graph type of the field.  Why?  The expression syntax does not need it (`Field(x => x.Name)`).  If we are specifying the return type of the field, we should not need to specify the graph type of the field.

```cs
    public DroidType(StarWarsData data)
    {
        // we should be able to do this
        Field("help")
            .Returns<bool>()
            .ResolveAsync(async context => true);

        // or better yet, this
        Field<bool>("help")
            .ResolveAsync(async context => true);

        // but this is annoying and stupid (personal opinion)
        Field<BooleanGraphType>("help")
            .Returns<bool>()
            .ResolveAsync(async context => true);
    }
```

I think if we are refactoring/enhancing the field builders we should figure this out.

Shane32: To carry on my suggestion, I'd suggest:

```cs
// already have this
public virtual FieldBuilder<TSourceType, object> Field<TGraphType>(string name = "default")
    where TGraphType : IGraphType { }

// suggesting to add this
/// <summary>
/// Adds a new field to the complex graph type and returns a builder for this newly added field.
/// </summary>
public virtual FieldBuilder<TSourceType, TReturnType> Field<TReturnType>(string name, bool nullable = false)
{
    var type = typeof(TReturnType);

    try
    {
        if (type == null)
            type = typeof(TReturnType).GetGraphTypeFromType(nullable, this is IInputObjectGraphType ? TypeMappingMode.InputType : TypeMappingMode.OutputType);
    }
    catch (ArgumentOutOfRangeException exp)
    {
        throw new ArgumentException($"The GraphQL type for field '{Name ?? GetType().Name}.{name}' could not be derived implicitly from type '{typeof(TReturnType).Name}'.", exp);
    }

    var builder = CreateBuilder<TReturnType>(type)
        .Name(name);

    AddField(builder.FieldType);
    return builder;
}

// both are valid:
Field<BooleanGraphType>("hello");
Field<bool>("hello");
```

Shane32: Re-thinking the above sample with a delegate, and assuming the graph type, attributes and query arguments are defined from the delegate, then I'm not sure the field builder is appropriate anyway:

```cs
    public DroidType(StarWarsData data)
    {
        // maybe this doesn't make sense, since no graph type can be inferred until Resolve is called:
        Field("help").Resolve(Help);

        // perhaps this is better
        Field("help", GetHelp);  // if we assume that functions typically have internal names, or if it makes sense to have the field names all listed in the constructor, this might be good

        // and/or this, where the name is inferred
        Field(Help);  // still could use `[Name]` attribute to define the name, even with no name defined here
    }

    private Character Help(Assistance assistanceNeeded) { }
```

Note that with a delegate, NRT attributes can be read from the delegate, whereas with inline functions they cannot.

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -418,9 +406,34 @@ public virtual FieldType AddField(FieldType fieldType)
...
+            AddField(builder.FieldType);
+            return builder;
+        }
+
+        /// <summary>
+        /// Adds a new field to the complex graph type and returns a builder for this newly added field.
+        /// </summary>
+        /// <param name="type">The graph type of this field.</param>
+        /// <param name="name">The name of the field.</param>
+        public virtual FieldBuilder<TSourceType, object> Field(string name, IGraphType type)
```
Shane32, line 431:
```suggestion
        public virtual FieldBuilder<TSourceType, object> Field(IGraphType type, string name)
```

I suggest we always have the type first -- so we have
- `Field<BooleanGraphType>("hello")`
- `Field(typeof(BooleanGraphType), "hello")`
- `Field(new BooleanGraphType(), "hello")`

Shane32 (commented):
If we push suggestions and comments about the delegate overload to another issue/PR, then I have only a single suggestion here -- reversing `Field(string, IGraphType)` to `Field(IGraphType, string)`

src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
sungam3r: I wanted to minimize required changes in tests when I fixed compiler errors. Then I forgot about this method. Fixed.

src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
sungam3r: `develop` is full of breaking changes. In that PR I tried to mini,ize them but not to zero.

Shane32: No problem, just a note for future reference.

sungam3r: By the way, I see that I can easily remove `= default`.

src/GraphQL/Builders/FieldBuilder.cs
sungam3r: Well, the things are simpler. At first attemt I named this method `ResolveDelegate`. And at some point of rewriting I decided to rename it into `Resolve` to align with other methods. I think the simple solution here - rename it to `ResolveDelegate` back to get rid of all potential problems and to not compilate the code. OK?

sungam3r: I'm sure very few people use this method.

src/GraphQL/Types/Composite/ComplexGraphType.cs
sungam3r: )))) I chose the best option for a long time. Now I would like to prefer name argument going first. It looks more naturally fo me to see field name first and then its type:

```graphql
type User {
  name: String
}
```

sungam3r:
So far I have answered all the comments.

src/GraphQL/Builders/FieldBuilder.cs
Shane32: Seems fine for this PR. Maybe consider further.

Shane32 (approved):

src/GraphQL/Builders/FieldBuilder.cs
sungam3r: Changed

src/GraphQL/Types/Composite/ComplexGraphType.cs
sungam3r: Of course in case of generic argument type goes first because of such C# syntax.

Shane32: Didn't we just change this so the type is always first?

Shane32: I think the type should always be first for consistency.

Shane32: I guess the other logic makes sense also.....

PR #3244, opened 2022-07-21
Use fluent Field APIs in tests + documentation fixes

sungam3r:
Part 2 of #3237

PR #3245, opened 2022-07-21
Obsolete old Field APIs + add docs about recommended ones

sungam3r:
Part 3 of #3237


Shane32:
I don't think we should deprecate the old `Field` methods without advance warning, and I don't think we should remove the methods until at least two versions later; perhaps 3 versions later.

Up until yesterday the `Field` methods were the suggested and documented method of adding fields to a graph type.  And the changes to the field builders within #3242 and #3244 were not even present.

Making this change isn't like most of the other changes we have made over time.  Typically a new version requires changing a little in the Startup.cs file, and perhaps a few modifications throughout the codebase.  Whereas this change would, for many users, require rewriting code for every graph type, easily hundreds or possibly thousands of lines of code.  We have had complaints before about the large number of changes required for every version bump.

I think we should see additional input on this before making this change.  @joemcbride do you have comments here?

Note that personally these changes would not bother me terribly much.  It would require quite a few changes in my code, but I can handle it.  Since many of my projects use extension methods, dynamic code, and derivitaves of `AutoRegisteringObjectGraphType`, the changes I would need to make are not nearly as bad as they would otherwise be.

I might suggest:

1. Adding a migration note with ⚠️ icons alerting that the non-field-builder methods will be deprecated in a _future_ version.  Perhaps keep all the migration text you already wrote in a separate document covering migrating `Field` methods, and link to it from the aforementioned ⚠️ note.  I'd also put this note near the top of the readme file.

2. State that they will be deprecated in 2023, and removed in 2025 -- not in a specific version.  It could be deprecated in a later 7.x version, not necessarily 8.0.

3. Future migration documents for v8/v9/etc can link to the same document that covers migrating `Field` methods.

4. Remove the `Field` methods in the first major version released in 2025.

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -825,9 +825,14 @@ namespace GraphQL.Builders
...
+    [System.Obsolete("This class will be removed in v8.")]
     public static class FieldBuilder
     {
+        [System.Obsolete("Please use FieldBuilder<TSourceType, TReturnType>.Create() method. This method wi" +
+            "ll be removed in v8.")]
         public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Create<TSourceType, TReturnType>(GraphQL.Types.IGraphType type) { }
+        [System.Obsolete("Please use FieldBuilder<TSourceType, TReturnType>.Create() method. This method wi" +
+            "ll be removed in v8.")]
         public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Create<TSourceType, TReturnType>(System.Type? type = null) { }
     }
```
Shane32, line 837: I'm good with these three changes in v7.

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -839,6 +844,9 @@ namespace GraphQL.Builders
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Argument<TArgumentGraphType>(string name, string? description, System.Action<GraphQL.Types.QueryArgument>? configure = null)
             where TArgumentGraphType : GraphQL.Types.IGraphType { }
+        [System.Obsolete("Please use Action<QueryArgument> parameter from other Argument() method overloads" +
+            " to set default value for parameter or use Arguments() method. This method will " +
+            "be removed in v8.")]
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Argument<TArgumentGraphType, TArgumentType>(string name, string? description, TArgumentType? defaultValue = default, System.Action<GraphQL.Types.QueryArgument>? configure = null)
```
Shane32, line 850: I'm good with this change in v7

sungam3r:
> and removed in 2025

😮

sungam3r:
I personaly do not see the problem at all in marking methods as obsolete.

> require rewriting code for every graph type, easily hundreds or possibly thousands of lines of code

No. This is an exaggeration. Obsolete does not assume that you need to rush to rewrite the code. A warning can be ignored or suppressed. These are sufficiently soft conditions for migration. No need to do them super-soft, it only complicates the support and evolution of the project.

Shane32:
> This is an exaggeration.

It certainly is not.  Just one of my projects has 500 `Field` calls - I checked.  And my GraphQL interface is only for end-users; the admin interface is a MVC interface.  I'm certain other users could have many more calls than me.

sungam3r:
Again I mean `obsolete` does not assume that you need to rush to rewrite the code. One more note - for those who **really** have 100500 `Fields` calls that **should be** changed they can just write 10-lines method copying it from here.

sungam3r:
We have already had such experience, for example, with methods for serialization in v5. I do not see the difference why it is impossible to do the same.

Shane32:
> they can just write 10-lines method copying it from here.

No it is not that easy.  Copying as extension methods would require adding `this.` in front of every call to `Field(...)`.  Or else they would need to derive from `ObjectGraphType` and add the methods, and then change every class to use the newly derived type.

Shane32:
Normally I have no problem immediately marking old methods as obsolete.  People can immediately use the code and as time presents itself, switch to the new code.

I just think that in this case, being that it will more often than not require touching every single graph of a user's schema, we should provide notification and opportunity for users to transition to the new code before they update and have hundreds of warnings in their build report -- and most people don't like warnings in their build report.

Think of it like this is a LTS feature -- we should support it longer than usual.  And as we have not had any notice or forewarning of this change, I don't think it is necessarily appropriate to make these old methods obsolete immediately.

The bigger issue to me is the plan to remove the methods in v8.  We still have people opening issues that are transitioning from 2.4.x (not often, but it happens).  They obviously have not upgraded in a few years.  If we remove the old methods in v8, it is very likely that many people on the current v5, or even last year's v4, will not have had any interim period to update their code.  And if they are coming from v3, there might be very little code that compiles at all -- seeing that scalars have changed, graph types would need to be rewritten, and the startup code would be completely different.  Not to mention that we rewrote the server project from scratch.  We'd have some pretty unhappy people, I can assure you.  I certainly would not remove the old methods anytime soon, even if we mark them obsolete now.

If some other large users endorse making the old methods obsolete immediately that's fine.  But without any additional input, I can't endorse making the changes this quickly.  I'd at least aim for marking the methods obsolete at year-end -- probably version 7.2 or something.

Shane32:
> We have already had such experience, for example, with methods for serialization in v5. I do not see the difference why it is impossible to do the same.

Many of those bigger changes have been constrained to overall setup.  Once the Startup.cs file is fixed, and perhaps a few graph types or a few calls are modified, all is well.  We are talking about every single graph type, and not only that, but every field definition on every graph type.  I think it is different.

sungam3r:
> No it is not that easy.

True.

> we should provide notification and opportunity for users to transition to the new code before they update and have hundreds of warnings in their build report -- **and most people don't like warnings in their build report**.

Most people don't like to read notifications either. I agree to change obsolete message to something like 
> Please use XXX method. This method **may** be removed in v8 or some time later. For now you can continue to use this API but we do not encourage this.


I believe that we **should** provide a clear signal via compiler warnings.

Shane32:
> Please use XXX method. This method may be removed in v8 or some time later. For now you can continue to use this API but we do not encourage this.

I would not say v8 specifically, and nor would I remove those methods if v8 is released soon.

I suggest:

> Please use XXX method. This method may be removed in a future release. For now you can continue to use this API but we do not encourage this.

sungam3r:
Done

Shane32:
I suggest we expand the examples a little more to show replacements for FieldDelegate, FieldSubscribe, FieldSubscribeAsync and the FieldAsync overload which includes the return CLR type.

sungam3r:
Done

docs2/site/docs/migrations/migration7.md
```diff
@@ -286,3 +286,110 @@ This change was done for better discoverability and usability of extension metho
...
 Custom implementations of `IResolveFieldContext` must implement the new `User` property.
+
+### 9. A bunch of FieldXXX APIs were deprecated
+
+After upgrade to v7 you will notice with ~100% probability compiler warnings with message
+> Please use one of the Field() methods returning FieldBuilder and then methods defined on it or just
+> use AddField() method directly. This method will be removed in v8.
+
+The goal of this [change](https://github.com/graphql-dotnet/graphql-dotnet/pull/3237) was to simplify
+APIs and guide developer with well-discovered APIs.
```
Shane32, line 297:
```suggestion
The goal of this [change](https://github.com/graphql-dotnet/graphql-dotnet/pull/3237) was to simplify
APIs and guide developers with well-discovered APIs.
```

docs2/site/docs/migrations/migration7.md
```diff
@@ -286,3 +286,110 @@ This change was done for better discoverability and usability of extension metho
...
+
+You will need to change a way of setting fields on your Graph Types. Instead of many `FieldXXX`
```
Shane32, line 299:
```suggestion
You will need to change a way of setting fields on your graph types. Instead of many `FieldXXX`
```

docs2/site/docs/migrations/migration7.md
```diff
@@ -286,3 +286,110 @@ This change was done for better discoverability and usability of extension metho
...
+overloads start configuring your field with one of the `Field` methods defined on `ComplexGraphType`.
```
Shane32, line 300:
```suggestion
overloads, start configuring your field with one of the `Field` methods defined on `ComplexGraphType`.
```

docs2/site/docs/migrations/migration7.md
```diff
@@ -286,3 +286,110 @@ This change was done for better discoverability and usability of extension metho
...
+All such methods define a new field and return an instance of `FieldBuilder<T,U>`. Then continue to
+configure defined field with rich APIs provided by the returned builder.
```
Shane32, line 302:
```suggestion
configure the field with rich APIs provided by the returned builder. 
```

Shane32 (changes requested):

docs2/site/docs/migrations/migration7.md
```diff
@@ -286,3 +286,110 @@ This change was done for better discoverability and usability of extension metho
 ### 8. `IResolveFieldContext.User` property added
 
 Custom implementations of `IResolveFieldContext` must implement the new `User` property.
+
+### 9. A bunch of FieldXXX APIs were deprecated
```
Shane32, line 290:
```suggestion
### 10. A bunch of FieldXXX APIs were deprecated
```
Probably requires manual merge now...

docs2/site/docs/migrations/migration7.md
```diff
@@ -286,3 +286,110 @@ This change was done for better discoverability and usability of extension metho
...
+
+After upgrade to v7 you will notice with ~100% probability compiler warnings with message
+> Please use one of the Field() methods returning FieldBuilder and then methods defined on it or just
+> use AddField() method directly. This method will be removed in v8.
```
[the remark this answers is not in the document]
Shane32: btw, "then" is incorrect ("and then methods") and should be "the".  Better check the `[Obsolete]` tags as I think I copied this from one of them.  It's 2am now so goodnight @sungam3r  !

docs2/site/docs/migrations/migration7.md
```diff
@@ -286,3 +286,110 @@ This change was done for better discoverability and usability of extension metho
...
+Field<MessageGraphType>("messages").ResolveStream(context => observable);
+
+
+
+// GraphQL 5.x
+Task<IObservable<object>> observable = null!;
+FieldSubscribeAsync<MessageGraphType>("messages", subscribeAsync: context => observable);
+
+
+
```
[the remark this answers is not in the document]
sungam3r: I did it intentionally to visually separate different examples.

docs2/site/docs/migrations/migration7.md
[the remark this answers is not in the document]
sungam3r: done

docs2/site/docs/migrations/migration7.md
sungam3r: done

Shane32:
@joemcbride would you mind reviewing this PR?

I’m ready to approve this, if @joemcbride agrees also.

Shane32:
I’m still concerned about deprecating the most commonly used methods in the library.

sungam3r:
> concerned about deprecating the most commonly used methods in the library.

This is not the first time I tried to do it. I started and retreated but sooner or later it had to be done. With fluent APIs on `FieldBuilder` configuring types is simpler then before.

STRATZ-Ken:
Overall, I agree with everything talked about here.  I would be one of those users with over 10,000 lines of code to update.  We have a VERY large project.

> Whereas this change would, for many users, require rewriting code for every graph type, easily hundreds or possibly thousands of lines of code. We have had complaints before about the large number of changes required for every version bump.

However, I am fine with obsolete as well. My job has me moving around to so many different project, I rarely have time to keep up with change logs.  Those little warnings tell me to come here and read the change log to see how its going.  

I also appreciate the expansion of the example sections.

Shane32 (approved):
I'll approve based on @STRATZ-Ken 's comments.

If @joemcbride has any comments, I'd like to hear them.  If not, let's merge.

sungam3r:
Waiting 1-2 days and then merge/release. This is the last issue from v7 milestone.

sungam3r:
No comments. I merge this one.

PR #3246, opened 2022-07-21
[WIP] Introduce JsonGraphType

### LINKED ISSUE #474: Support for JSON scalar type
Hi,
Is JSON scalar type should be added soon to the library?
If not, what is the best way to add it (inherit from ScalarGraphType)?

Thank you

sungam3r:
Closes #474. Needs counterpart in NSJ project.

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.SystemTextJson.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.SystemTextJson.approved.txt
@@ -68,6 +68,14 @@ namespace GraphQL.SystemTextJson
         public override void Write(System.Text.Json.Utf8JsonWriter writer, System.Numerics.BigInteger value, System.Text.Json.JsonSerializerOptions options) { }
         public static bool TryGetBigInteger(ref System.Text.Json.Utf8JsonReader reader, out System.Numerics.BigInteger bi) { }
     }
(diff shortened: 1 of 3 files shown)


Shane32:
I really don't think this is at all what was being requested.  This is deserializing a JSON string value, rather than a map.  Let me demonstrate:

```gql
# this PR
{
  example (json: "{\"test\":\"value\"}")
}

# what's been requested
{
  example (json: { test: "value" })
}
```

Or for variables:

```json
// this PR
{
  "json": "{\"test\":\"value\"}"
}

// what's been requested
{
  "json": {
    "test": "value"
  }
}
```

In addition:

- The scalar as shown in this PR can be implemented by pulling `IGraphQLTextSerializer` into the scalar via DI and then deserializing to type `object`.  If it's a map or array, users can then use `IGraphQLTextSerializer.ReadNode` to deserialize it into a specific type.  I might consider deserializing to the `Inputs` type within the scalar for objects, however.

- What I believe has been requested wouldn't contain any JSON parsing at all.  It would construct a dictionary from literals, and for variables, it would simply pass through the dictionary.

Shane32:
See also #669. Both that and #474 have sample code indicating that users are not attempting to deserialize a string value, but rather read a map as a scalar, either within a literal or within the JSON variables map.

PR #3247, opened 2022-07-22
Add Field overload with clr type

Shane32:
Allows for:

```cs
Field<int>("test")
    .Resolve(_ => 123);
```

### DIFF
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -104,6 +104,38 @@ This property is similar in nature to the ASP.NET Core `HttpContext.User` proper
 used by the GraphQL.NET engine internally but merely being a convenience property similar to
 `RequestServices` and `UserContext` for use by separate authentication packages.
 
+### 9. Add `Field<TReturnType>` overload to create field builder with an inferred graph type
+
+To define a field with a field builder, previously the graph type was always required, like this:
+
+```csharp
+Field<IntGraphType>("test")
+    .Resolve(_ => 123);
+
+// or
+
+Field<IntGraphType, int>("test")
+    .Resolve(_ => 123);
+```
+
+Now you can simply specify the return type, and the graph type will be inferred:
+
+```csharp
+Field<int>("test")        // assumes not-null
+    .Resolve(_ => 123);
+
+// or
+
+Field<int>("test", true)  // specify true or false to indicate nullability
+    .Resolve(_ => 123);
+```
+
+This is similar to the expression syntax (`Field(x => x.Name)`) which does not require
+the graph type to be specified in order to define a field.
+
+As with the expression syntax or the `AutoRegisteringObjectGraphType`,
+CLR type mappings can be tailored via the `schema.RegisterTypeMapping()` methods.
+
 ## Breaking Changes
 
 ### 1. `DataLoaderPair<TKey, T>.Loader` property removed
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -121,7 +121,7 @@ Field<IntGraphType, int>("test")
 Now you can simply specify the return type, and the graph type will be inferred:
 
 ```csharp
-Field<int>("test")        // assumes not-null
+Field<int>("test")        // by defaut assumes not-null
     .Resolve(_ => 123);
 
 // or
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System.Collections;
 using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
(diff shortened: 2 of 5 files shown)


docs2/site/docs/migrations/migration7.md
```diff
@@ -104,6 +104,38 @@ This property is similar in nature to the ASP.NET Core `HttpContext.User` proper
...
+// or
+
+Field<IntGraphType, int>("test")
+    .Resolve(_ => 123);
+```
+
+Now you can simply specify the return type, and the graph type will be inferred:
+
+```csharp
+Field<int>("test")        // assumes not-null
```
sungam3r, line 124:
```suggestion
Field<int>("test")        // by defaut assumes not-null
```

sungam3r, line 124: What about `Field<int?>("test")` ? Nullable or not? Maybe it is another discussion around nullability APIs (how to provide nullability info from caller).

sungam3r: изображение
3! sources of information about nullability 😮

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -468,4 +470,49 @@ public void should_not_throw_exception_on_valid_field_name_using_field_builder(s
...
+        type.Field<string?>("field10").Resolve(_ => "hello");
+        type.Fields.Find("field10").ShouldNotBeNull().Type.ShouldBe(typeof(NonNullGraphType<GraphQLClrOutputTypeReference<string>>));
+
+        type.Field<string?>("field11", false).Resolve(_ => "hello");
+        type.Fields.Find("field11").ShouldNotBeNull().Type.ShouldBe(typeof(NonNullGraphType<GraphQLClrOutputTypeReference<string>>));
+
+        type.Field<string?>("field12", true).Resolve(_ => "hello");
+        type.Fields.Find("field12").ShouldNotBeNull().Type.ShouldBe(typeof(GraphQLClrOutputTypeReference<string>));
+
+        // notice here that since the NRT attribute of 'string' cannot be read, it is assumed to be nullable
```
sungam3r, line 514: > notice here that since the NRT attribute of 'string' cannot be read

Why?

docs2/site/docs/migrations/migration7.md
Shane32: See the new tests which cover the `Field<int?>("test")` scenario.  Right or wrong, it matches the `Field(x => x.Amount)` syntax.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32: NRT annotations are stored on method signatures (a `MethodInfo` / similar) or generic class definitions.  An inline function doesn't have a `MethodInfo` to read, I believe, and a call to a generic function is directly compiled; no NRT attributes are stored anywhere.

So in the following example, NRT annotations are dropped by the compiler and there is no way to recover them:

```cs
Field<string?>("test");
```

I could be wrong, but that's my understanding.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -468,4 +470,49 @@ public void should_not_throw_exception_on_valid_field_name_using_field_builder(s
...
+        type.Field<int>("field1").Resolve(_ => 3);
+        type.Fields.Find("field1").ShouldNotBeNull().Type.ShouldBe(typeof(NonNullGraphType<GraphQLClrOutputTypeReference<int>>));
+
+        type.Field<int>("field2", false).Resolve(_ => 3);
+        type.Fields.Find("field2").ShouldNotBeNull().Type.ShouldBe(typeof(NonNullGraphType<GraphQLClrOutputTypeReference<int>>));
+
+        type.Field<int>("field3", true).Resolve(_ => 3);
+        type.Fields.Find("field3").ShouldNotBeNull().Type.ShouldBe(typeof(GraphQLClrOutputTypeReference<int>));
+
+        Should.Throw<ArgumentException>(() => type.Field<int?>("field4"));
```
sungam3r, line 488: Please add `.Message.ShouldBe(...)` to make this test more readable.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32: Because of the same issue, some class definitions cannot properly determine NRT status:

```cs
class ClassA<T>
{
    public T GetValue();
}

class ClassB : ClassA<string?> // these NRT attributes are stored
{
}

var valueA = new ClassA<string?>().GetValue();  // GetValue cannot determine NRT status of T
var valueB = new ClassB().GetValue();  // GetValue could determine NRT status of T
```

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32: done

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
sungam3r: изображение
😕

Shane32: Of course the compiler knows the type.  It's just not saved into the dll, so reflection cannot see it.

sungam3r: ok

sungam3r (approved):
👍

sungam3r:
Sorry for delay for such small PR. I really was busy.

PR #3248, opened 2022-07-22
Prevent double initialization of graph types

Shane32:


src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -189,25 +189,25 @@ protected void Initialize(ISchema schema, IServiceProvider serviceProvider, IEnu
...
                 if (type is not ScalarGraphType)
-                    AddType(type, ctx);
+                    AddTypeIfNotRegistered(type, ctx, schema);
             }
 
             // these fields must not have their field names translated by INameConverter; see HandleField
-            HandleField(null, SchemaMetaFieldType, ctx, false);
-            HandleField(null, TypeMetaFieldType, ctx, false);
-            HandleField(null, TypeNameMetaFieldType, ctx, false);
+            HandleField(null, SchemaMetaFieldType, ctx, false, schema);
```
sungam3r, line 204: Maybe add `Schema` into `TypeCollectionContext`?

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -1962,6 +1963,7 @@ namespace GraphQL.Types
         protected bool Equals(GraphQL.Types.IGraphType other) { }
         public override bool Equals(object? obj) { }
         public override int GetHashCode() { }
+        public virtual void Initialize(GraphQL.Types.ISchema schema) { }
```
sungam3r, line 1966: IMO it's not a responsibility of GraphType itself to track its "initialization flag" according to usage of that type in some schema. And what if I use one GraphType (singleton instance) within 10 schemas of my app? For example `Money` simple type with two fields - currency and amount?

Shane32: Graph types CANNOT be used within two schema instances.  It will ALWAYS bring memory leaks, and can lead to data corruption.  If you do so, you are doing so incorrectly.  This library does not support such a scenario.

In the example you gave, now the graph type will hold a reference to a `FloatGraphType` (or similar scalar) that is not part of the schema.  Meaning, this is a different reference than the `FloatGraphType` stored in the schema's `SchemaTypes` collection.  Hence, this a memory leak and is not the correct behavior.  And if there was a modification to the built-in `FloatGraphType` for one of the schemas, the other would behave incorrectly when referencing a field from the shared `Money` type.  Only scalars can be shared across schemas, and this is a supported scenario, and still supported with this change.

This change is necessary because of the high number of users that do not understand this concept and try to used scoped schemas incorrectly.  It almost always leads to frustration on their part and an issue that we need to answer.

Many services provide for an initialization flag or similar to prevent something from happening twice when it should not.  This is no different.  Our own `Schema` class prevents double initialization.

This PR also allows for schema-dependent initialization code, which was not possible previously.

Shane32: At least, that's my opinion.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: I thought of it, but the context is recreated a couple times in the initialization code.  I could put it in the context of course.

(Personally I think the entire class needs a rewrite, but it's more work than I've wanted to do so far.)

Shane32:
FYI, so far I can't seem to get the introspection query to return types in the same order as they were previously.  This is because introspection types were added more than once within `SchemaTypes`; now to prevent double initialization they are only added once.  And apparently this changes the order in which the types are read from the collection.  Since the introspection types are not sorted when read, I may have to update the tests for the introspection queries to get the tests to pass.

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
sungam3r: >  Only scalars can be shared across schemas

Is this mentioned in the documentation?

Shane32: > > Only scalars can be shared across schemas
> 
> Is this mentioned in the documentation?

No, but it could be.  We recommend that graph types are registered as transients in the documentation.  `.AddGraphTypes()` registers found graph types as transients and there is no overload/argument to register them otherwise.  This way if there are multiple schemas, or if the schema is registered as scoped, each schema will get their own instance of any shared graph types.  I think we have also tests to ensure that `SchemaTypes` will only create a single instance of each graph type required.

Shane32: Another problem, which this PR neither hurts nor solves, is that when using the `ResolveType` delegate of `UnionGraphType`, the delegate must return an instance of a graph type.  But since graph types are transients, it cannot be passed into the union graph type via DI.  (It can be, but it will be an un-initialized instance, so fields' `ResolvedType` properties will be null, causing a NRE.)  The easy solution is to manually register the referenced graph type as a singleton.  Having the graph type as a singleton prevents using multiple schemas with the same graph type, and prevents using the schema as a scoped instance -- the latter can cause the multithreading issues we have seen before.  Both of those scenarios will now throw an exception with this PR.

There are a couple solutions to this issue:
1. Pass `SchemaTypes` (or `ISchema`) to the delegate, so the proper graph type can be pulled from there.
2. Have the delegate return `Type` instances rather than `IGraphType` instances.  But this may prevent usage from dynamic schemas.  So I don't think that's an option either.
3. Have the delegate return a string representing the name of the graph type to use.  Not my preferred solution.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: ok done

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -14,13 +14,23 @@ public abstract class ComplexGraphType<TSourceType> : GraphType, IComplexGraphTy
...
         {
             Description ??= typeof(TSourceType).Description();
             DeprecationReason ??= typeof(TSourceType).ObsoleteMessage();
         }
 
+        /// <inheritdoc/>
+        public override void Initialize(ISchema schema)
+        {
+            if (_initialized)
+                throw new InvalidOperationException($"This graph type '{Name}' has already been initialized.");
```
sungam3r, line 30:
```suggestion
                throw new InvalidOperationException($"This graph type '{Name}' has already been initialized. Make sure that you do not use the same instance of graph type in multiple schemas");
```

`This graph type '{Name}' has already been initialized` message as is does not tell how to fix the problem.

src/GraphQL/Types/GraphType.cs
```diff
@@ -25,6 +25,11 @@ protected GraphType()
             }
         }
 
+        /// <inheritdoc/>
+        public virtual void Initialize(ISchema schema)
```
sungam3r, line 29: Why so? I guess because of `ScalarGraphType`. If so I suggest to move `throw` here and override `ScalarGraphType.Initialize` to be empty.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -177,8 +178,7 @@ protected void Initialize(ISchema schema, IServiceProvider serviceProvider, IEnu
             // CamelCaseNameConverter, as some fields are defined in pascal case - e.g. Field(x => x.Name)
             _nameConverter = CamelCaseNameConverter.Instance;
 
-            foreach (var introspectionType in _introspectionTypes.Values)
-                AddType(introspectionType, _context);
+            AddType(_introspectionTypes[typeof(__Schema)], _context);
```
sungam3r, line 181: `AddType` adds all types recursively, right?

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -386,6 +386,38 @@ public void does_not_throw_with_filtering_nameconverter(string fieldName)
...
+            var type = new ComplexType<string> { Name = "Query" };
+            type.Field<IntGraphType>("field1");
+            var schema = new Schema() { Query = type };
+            schema.Initialize();
+        }
+
+        {
+            var type = new ComplexType<string> { Name = "Query" };
+            type.Field<IntGraphType>("field1");
+            var schema = new Schema() { Query = type };
```
sungam3r, line 416:
```suggestion
            var schema = new Schema { Query = type };
```

code style

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Yes

src/GraphQL/Types/GraphType.cs
Shane32: Done

sungam3r (approved):

PR #3250, opened 2022-07-22
Update ResolveType to include schema

Shane32:
Another problem, which this PR neither hurts nor solves, is that when using the `ResolveType` delegate of `UnionGraphType`, the delegate must return an instance of a graph type.  But since graph types are transients, it cannot be passed into the union graph type via DI.  (It can be, but it will be an un-initialized instance, so fields' `ResolvedType` properties will be null, causing a NRE.)  The easy solution is to manually register the referenced graph type as a singleton.  Having the graph type as a singleton prevents using multiple schemas with the same graph type, and prevents using the schema as a scoped instance -- the latter can cause the multithreading issues we have seen before.  Both of those scenarios will now throw an exception with this PR.

There are a couple solutions to this issue:
1. Pass `SchemaTypes` (or `ISchema`) to the delegate, so the proper graph type can be pulled from there.
2. Have the delegate return `Type` instances rather than `IGraphType` instances.  But this may prevent usage from dynamic schemas.  So I don't think that's an option either.
3. Have the delegate return a string representing the name of the graph type to use.  Not my preferred solution.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/pull/3248#discussion_r927783159_

### DIFF
--- a/docs2/site/docs/getting-started/interfaces.md
+++ b/docs2/site/docs/getting-started/interfaces.md
@@ -117,27 +117,17 @@ that implements an Interface you are required to alter the Interface for that ne
 ```csharp
 public class CharacterInterface : InterfaceGraphType<StarWarsCharacter>
 {
-  public CharacterInterface(
-    DroidType droidType,
-    HumanType humanType)
+  public CharacterInterface()
   {
     Name = "Character";
 
     ...
 
-    ResolveType = obj =>
+    ResolveType = (obj, schema) => obj switch
     {
-        if (obj is Droid)
-        {
-            return droidType;
-        }
-
-        if (obj is Human)
-        {
-            return humanType;
-        }
-
-        throw new ArgumentOutOfRangeException($"Could not resolve graph type for {obj.GetType().Name}");
+        Droid _ => schema.AllTypes[typeof(DroidType)] as IObjectGraphType,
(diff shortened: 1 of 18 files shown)


joemcbride:
Just throwing this out there - isn't using `GraphQLTypeReference` in this case a simpler solution? I think in most cases that would make it so you don't have to use a switch. That reference should then get resolved to the concrete type in the framework.  That is basically the same thing as grabbing the type directly from the `ISchema`.

Shane32:
Yes. I didn’t realize it worked that way until I saw the schema builder code while writing this. It doesn’t allow to find by the CLR type of the graph type but it is a viable solution — just not strongly typed.

Shane32:
@joemcbride I could just update the documentation to show using `GraphQLTypeReference` and close this PR?

Either way is completely fine with me; I don't use `ResolveType` at all.  I just know that the sample doesn't work when registering schema types as transients, which is the current recommendation and default behavior.

joemcbride:
I think just updating the docs should work.

```csharp
ResolveType = obj => new GraphQLTypeReference(obj.GetType().Name);
 ```

Could also show an example of using an attribute on the model to fetch the associated GraphType name.

Shane32:
This PR has been replaced by #3254

PR #3253, opened 2022-07-27
Add IsNativelyAsync property

Shane32:
See https://github.com/graphql-dotnet/server/pull/733

### DIFF
--- a/src/GraphQL/Transport/IGraphQLSerializer.cs
+++ b/src/GraphQL/Transport/IGraphQLSerializer.cs
@@ -31,7 +31,8 @@ namespace GraphQL
 
         /// <summary>
         /// Indicates whether this serializer makes asynchronous calls to the underlying stream
-        /// while serializing or deserializing.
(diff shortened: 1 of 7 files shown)


src/GraphQL/Transport/IGraphQLSerializer.cs
```diff
@@ -28,5 +28,11 @@ public interface IGraphQLSerializer
         /// A <paramref name="value"/> of <see langword="null"/> returns <see langword="default"/>.
         /// </summary>
         T? ReadNode<T>(object? value);
+
+        /// <summary>
+        /// Indicates whether this serializer makes asynchronous calls to the underlying stream
+        /// while serializing or deserializing.
```
sungam3r, line 34:
```suggestion
        /// while serializing or deserializing. This property is auxiliary API in nature and may
        /// help to avoid an additional/unnecessary buffering at caller side.
```

sungam3r (approved):
Now I have some time to review, I will see the remaining mail tomorrow morning.

Issue #3233, opened 2022-07-18
GraphQL Serialization ObjectConverter NotSupportedException only online


julienFlexsoft:
## Description

I added an unhandledExceptionDelegate in my GraphQLMiddleware.  In this middleware I have logic that adds info to the logged error. This enriches the data we see in Azure App Insights. When I trigger errors locally everything works well: I get the error in Azure with all the extra info, and I get a clean graphQL error as response. But when I do the same thing online it doesn't work and returns a NotSupportedException from System.Text.Json.Serialization.Converters.ObjectConverter.

### Steps to reproduce

**GraphQLMiddleware.cs**
```
private async Task ExecuteAsync(HttpContext context, ISchema schema, IEnvHelper envHelper,
    DataLoaderDocumentListener dataLoaderDocumentListener, IRequestContext requestContext,
    ILogger<GraphQLMiddleware> logger)
{
    var start = DateTime.UtcNow;

    GraphQLRequest graphQLRequest = await _graphQLSerializer.ReadAsync<GraphQLRequest>(context.Request.Body, context.RequestAborted);
    var enableMetrics = envHelper.CurrentEnvironment() switch
    {
        EnvironmentEnum.Local or EnvironmentEnum.Dev or EnvironmentEnum.Staging => true,
        EnvironmentEnum.Production => false,
        _ => throw new NotImplementedException(),
    };
    var result = await _executer.ExecuteAsync(options =>
    {
        options.Schema = schema;
        options.Query = graphQLRequest.Query;
        options.OperationName = graphQLRequest.OperationName;
        options.Variables = graphQLRequest.Variables;
        options.UserContext = GenerateGraphQLUserContext(context);
        options.EnableMetrics = enableMetrics;
        options.RequestServices = context.RequestServices;
        options.CancellationToken = context.RequestAborted;
        options.Listeners.Add(dataLoaderDocumentListener);
        options.ThrowOnUnhandledException = false;
        options.ValidationRules = context.RequestServices.GetServices<IValidationRule>().Concat(DocumentValidator.CoreRules);
        options.UnhandledExceptionDelegate = ctx =>
        {
            return HandleErrors(context, ctx, graphQLRequest, requestContext, logger);
        };
    });

    if (enableMetrics)
    {
        result.EnrichWithApolloTracing(start);
    }

    await WriteResponseAsync(context, result, context.RequestAborted);
}

private async Task WriteResponseAsync(HttpContext context, ExecutionResult result,
[...]

PR #3256, opened 2022-07-31
Demonstrate issue 3233

sungam3r:
#3233 

Just to explicitly state the current status quo with serialization.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Issue3233.cs
+++ b/src/GraphQL.Tests/Bugs/Issue3233.cs
@@ -17,7 +17,7 @@ public class Issue3233
         result.AddError(errorWithData);
 
         var ex = Should.Throw<NotSupportedException>(() => serializer.Serialize(result));
-        if (OperatingSystem.IsWindows())
+        if (Environment.OSVersion.ToString().Contains("Windows"))
             ex.Message.ShouldBe("The type 'System.Object' is not a supported dictionary key using converter of type 'System.Text.Json.Serialization.Converters.ObjectConverter'. Path: $.");
         else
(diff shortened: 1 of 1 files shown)


Shane32 (approved):

Shane32:
What about changing ErrorInfoProvider so that data is not included by default?

Shane32:
> Do you mean `ErrorInfoProviderOptions.ExposeData` ? `ErrorInfoProvider` just uses its boolean value. I would not change ` = true` default for `ExposeData`.

Yes.  I really don't know why anyone would ever want `ExposeData` to be `true`, any more than they want to reveal stack traces.

sungam3r:
😕 I think we use `true` as default value because a long time ago it was used in old 2.4.0 version.

sungam3r:
In our backend we use this data dictionary inside error to control some specific auth behaviour across microservices.

sungam3r:
I want to say that this data has nothing to do with stack traces. It is some sort or user-defined structured information.

Shane32:
> I want to say that this data has nothing to do with stack traces. It is some sort or user-defined structured information.

But I believe the most common use case is internal information within C#. For example the MS SQL library adds data here. Somehow Azure is adding data there causing the problem related to this PR. In both of those cases, it is often not the intent of the developer to expose such data to the end user. 

I am not arguing that certain users, such as yourself, may wish to use the data field. But it seems the exception and not the rule.

To summarize, it seems the data field is used by other services and often unintentionally exposed to a caller of the API.

sungam3r:
I am inclined to agree. Do you want to change default?

Shane32:
> I am inclined to agree. Do you want to change default?

Completed in #3263

PR #3144, opened 2022-05-19
Code-First Federation 2 support

Mithras:
Subj.

Some details on what's in the PR:

1. Added [.devcontainer](https://github.com/Mithras/graphql-dotnet/tree/master/.devcontainer) because I don't want to install 9000 SDKs and packages on my host. OmniSharp (Intellisense) is half broken though because it works only with NET 6. This is not related to Federation and I can remove it if you don't want it. Or I can try to fix OmniSharp if you are interested in developing in containers.
2. Separate [GraphQL.Federation.Tests](https://github.com/Mithras/graphql-dotnet/tree/master/src/GraphQL.Federation.Tests) project with a couple INTEGRATION tests (run in 75ms though). I'm not sure you want integration tests and you might prefer 9000 unit tests instead so I made it separate so that it's easy to get rid of. I don't mind adding unit tests, just didn't want to waste time as I'm not sure this PR will ever get reviewed/merged. The tests are pretty simple and self explanatory, see [FederationFieldTests](https://github.com/Mithras/graphql-dotnet/blob/master/src/GraphQL.Federation.Tests/FederationFieldTests.cs).
3. A bunch of NEW stuff under [GraphQL.Federation](https://github.com/Mithras/graphql-dotnet/tree/master/src/GraphQL/Federation) namespace.
  3.1. `Attributes.*` - All Federation 2 [directives](https://www.apollographql.com/docs/federation/federated-types/federated-directives/) as attributes.
  3.2 [Extensions.FederationFieldExtensions](https://github.com/Mithras/graphql-dotnet/blob/master/src/GraphQL/Federation/Extensions/FederationFieldExtensions.cs) - All Federation 2 [directives](https://www.apollographql.com/docs/federation/federated-types/federated-directives/) as `FieldType`/`FieldBuilder` extensions.
  3.3 [Extensions.FederationGraphTypeExtensions](https://github.com/Mithras/graphql-dotnet/blob/master/src/GraphQL/Federation/Extensions/FederationGraphTypeExtensions.cs) - All Federation 2 [directives](https://www.apollographql.com/docs/federation/federated-types/federated-directives/) as `GraphType` extensions. Also, `.AddServices()` and `.AddEntities()` to add Federation fields explicitly/manually (e.g. when you want to specify `.AuthorizeWith()` on them).
[...]

PR #3257, opened 2022-08-01
Add ComplexScalarGraphType

### LINKED ISSUE #474: Support for JSON scalar type

Shane32:
Closes #474 

Alternative to #3246 

This scalar type is not specifically a "JSON" graph type so much as it is an 'any' graph type.  Meaning, it can represent any simple or complex object, either as a literal or a variable.

For example, with field `test` having query argument `arg` of type `ComplexScalarGraphType`:

```gql
query Example1 {
  test (arg: { randomField1: "hello", randomField2: [1,2,3], randomField3: { child1: "hello", child2: true, child3: null } }) {
    ...
  }
}

query Example2 ($arg: Complex!) {
  test (arg: $arg) {
    ...
  }
}
```

Variables for example 2:

```json
{
  "arg": {
    "randomField1": "hello",
    "randomField2": [1, 2, 3],
    "randomField3": {
      "child1": "hello"
      "child2": true,
      "child3": null,
    }
  }
}
```

This type can also be used as an output scalar as follows:

```cs
Field<ComplexScalarGraphType>("test")
    .Returns(_ => new
    {
        output1 = "hello",
        output2 = new[] { 1, 2, 3 },
        output3 = new
        {
            child1 = "hello",
            child2 = true,
            child3 = (string?)null,
        },
    });
```

Returns:

```json
{
  "data": {
    "test": {
        "output1": "hello",
        "output2": [1, 2, 3],
        "output3": {
          "child1": "hello",
          "child2": true,
          "child3": null
        }
    }
  }
}
```

Having a complex literal with nested nodes that reference a variable is not currently possible, due to limitations within the GraphQL.NET framework -- `ParseLiteral` does not have access to the request variables.  For example:

```gql
# impossible with current scalar framework design
query Example1 ($num: Int) {
  test (arg: { randomField1: "hello", randomField2: $num }) {
    ...
  }
}
```

This is just a proof-of-concept; it needs tests and review.  I do not have this PR as a goal for 7.0.  However, if we want to modify the `ScalarGraphType` so that `ParseLiteral` is also passed the request variables, then we should do so prior to 7.0.

This entire scalar could exist merely as sample code, however we have been asked for this feature numerous times and it seems reasonable to add.  We have already verified that it does not conflict with the GraphQL spec; scalars may be complex objects.

There is no direct dependency on either serialization engine.  However, makes some assumptions about how `InputsConverter` works, reading GraphQL literals in the same manner.


sungam3r:
> Having a complex literal with nested nodes that reference a variable is not currently possible

I doubt that it is worth supporting variables inside the scalars. Too narrow case.

src/GraphQL/Types/Scalars/ComplexScalarGraphType.cs
```diff
@@ -0,0 +1,175 @@
...
+            float n => new GraphQLFloatValue(n),
+            double n => new GraphQLFloatValue(n),
+            decimal n => new GraphQLFloatValue(n),
+
+            string s => new GraphQLStringValue(s),
+
+            IDictionary d => ConvertDictionary(d),
+            IEnumerable l => ConvertList(l),
+
+            _ => throw new NotImplementedException("Converting complex types to their AST representations is not supported.")
```
sungam3r, line 142:
```suggestion
            _ => throw new NotImplementedException($"Converting complex types ({value.GetType().Name}) to their AST representations is not supported.")
```

sungam3r:
I looked through the code and compared it with my rework of `SchemaPrinter` where I touched `AnyGraphType` as well. In essence, `ComplexGraphType` and `AnyGraphType` are the same and even implemented approximately the same. I propose to postpone introducing `Complex` scalar and review it along with #3144 after v7 release.

Shane32:
Updating for v8

Shane32:
Ref:
- #3760

src/GraphQL/Types/Scalars/ComplexScalarGraphType.cs
```diff
@@ -0,0 +1,118 @@
...
+                fields.Add(new GraphQLObjectField(
+                    new GraphQLName(keyString),
+                    ToAST(enumerator.Value)));
+            }
+            return new() { Fields = fields };
+        }
+
+        GraphQLListValue ConvertList(IEnumerable list)
+        {
+            List<GraphQLValue>? values = null;
```
gao-artur, line 110: It is worth checking if the `list is ICollection` and allocating the list with the initial capacity to avoid re-sizes

Shane32:
```suggestion
            List<GraphQLValue>? values = null;
            if (list is ICollection collection)
                values = new(collection.Count);
```

Shane32: Good idea

src/GraphQL/Types/Scalars/ComplexScalarGraphType.cs
```diff
@@ -0,0 +1,120 @@
...
+                    Fields = objectDictionary
+                        .Select(row => new GraphQLObjectField(
+                            new GraphQLName(row.Key),
+                            ToAST(row.Value)))
+                        .ToList(),
+                };
+            }
+
+            var fields = new List<GraphQLObjectField>(dictionary.Count);
+            var enumerator = dictionary.GetEnumerator();
```
gao-artur, line 96: Dispose the enumerator

gao-artur (approved):

PR #3269, opened 2022-08-07
Verify AutoRegisteringInputObjectGraphType works for init-only setters

Shane32:


sungam3r (approved):
set and init are the same things from runtime point of view

PR #3270, opened 2022-08-08
Add AutoRegisteringInterfaceGraphType

Shane32:
This adds `AutoRegisteringInterfaceGraphType` which is almost identical to `AutoRegisteringObjectGraphType`, except it inherits from `InterfaceGraphType<T>` rather than from `ObjectGraphType<T>`.  The following additional changes were necessary to finish proper support of this functionality:

1. Update `AutoRegisteringGraphTypeMappingProvider` to construct `AutoRegisteringInterfaceGraphType<T>` for interface CLR types
2. Support returning objects for an interface that does not have a matching `ObjectGraphType` for the type of object being returned.  This is accomplished by modifying the check for when a object is returned for an interface graph type, and it does not match on any `IsTypeOf` methods, to simply return the interface type.  This change should be allowed per the GraphQL spec, and so then if an interface CLR type gets automatically mapped to a interface graph type (rather than an object graph type), requests will still execute properly.
3. Allow `GraphQLAttributes` to be set on interfaces.
4. Add attribute to link CLR types to GraphQL interface types (either by referencing the CLR interface type or GraphQL interface type)


src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -154,7 +154,7 @@ protected virtual bool ShouldIncludeNode(ExecutionContext context, IHasDirective
         /// </summary>
         protected virtual void SetSubFieldNodes(ExecutionContext context, ObjectExecutionNode parent)
         {
-            var parentType = parent.GetObjectGraphType(context.Schema)!;
+            var parentType = parent.GetObjectGraphType(context.Schema) ?? parent.GraphType;
```
Shane32, line 157: This change makes it so that, for an interface, if `IsTypeOf` returns `false` for all configured object types, the interface graph type is used to resolve the query.  Previously some weird meaningless exception would occur.  This should not break GraphQL spec.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -428,7 +428,11 @@ private void AddType(IGraphType type, TypeCollectionContext context)
             {
                 foreach (var objectInterface in obj.Interfaces.List)
                 {
-                    if (AddTypeIfNotRegistered(objectInterface, context) is IInterfaceGraphType interfaceInstance)
+                    object typeOrError = RebuildType(objectInterface, false, context.ClrToGraphTypeMappings);
+                    if (typeOrError is string error)
+                        throw new InvalidOperationException($"The GraphQL implemented type for '{type.Name}' could not be derived implicitly. " + error);
+                    var objectInterface2 = (Type)typeOrError;
+                    if (AddTypeIfNotRegistered(objectInterface2, context) is IInterfaceGraphType interfaceInstance)
```
Shane32, line 435: This change is necessary to support `Interfaces.Add(typeof(GraphQLClrOutputTypeReference<MyInterfaceType>));`.  It just runs the type through the `RebuildType` code which runs it through the `IGraphTypeMappingProvider`.  Really more of a bug fix, I'd say.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -224,5 +224,143 @@ internal static LambdaExpression GetParameterExpression(Type parameterType, Quer
...
+        /// and <see cref="FieldType.StreamResolver"/>.
+        /// <br/><br/>
+        /// For fields and properties, no query arguments are added and the field resolver simply pulls the appropriate
+        /// member from <see cref="IResolveFieldContext.Source"/>.
+        /// <br/><br/>
+        /// For methods, method arguments are iterated and processed by
+        /// <see cref="AutoRegisteringObjectGraphType{TSourceType}.GetArgumentInformation{TParameterType}(FieldType, ParameterInfo)">GetArgumentInformation</see>, building
+        /// a list of query arguments and expressions as necessary. Then a field resolver is built around the method.
+        /// </summary>
+        internal static void BuildFieldTypeForOutput(
```
Shane32, line 240: This just consolidates code from `AutoRegisteringObjectGraphType` and `AutoRegisteringInterfaceGraphType`.  There are no code changes within this method.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -224,5 +224,143 @@ internal static LambdaExpression GetParameterExpression(Type parameterType, Quer
...
+        }
+
+        /// <summary>
+        /// Returns a list of properties, methods or fields that should have fields created for them.
+        /// <br/><br/>
+        /// Unless overridden, returns a list of public instance readable properties and public instance methods
+        /// that do not return <see langword="void"/> or <see cref="Task"/>
+        /// including properties and methods declared on inherited classes.
+        /// </summary>
+        internal static IEnumerable<MemberInfo> GetRegisteredMembersForOutput<TSourceType>(Expression<Func<TSourceType, object?>>[]? excludedProperties)
```
Shane32, line 328: This method consolidates code between `AutoRegisteringInterfaceGraphType` and `AutoRegisteringObjectGraphType`.  As can be seen here, typically there is no shared executable code between the two implementations, this consolidation does allow interfaces to be constructed with `AutoRegisteringObjectGraphType` and classes with `AutoRegisteringInterfaceGraphType`.  There seems to be no reason not to design it this way, and it does allow for interfaces to be used to separate the graph attributes from GraphQL object graph types.

src/GraphQL/Types/Composite/AutoRegisteringInterfaceGraphType.cs
```diff
@@ -0,0 +1,154 @@
...
+using System.Reflection;
+
+namespace GraphQL.Types
+{
+    /// <summary>
+    /// Allows you to automatically register the necessary fields for the specified type.
+    /// Supports <see cref="DescriptionAttribute"/>, <see cref="ObsoleteAttribute"/>, <see cref="DefaultValueAttribute"/> and <see cref="RequiredAttribute"/>.
+    /// Also it can get descriptions for fields from the XML comments.
+    /// </summary>
+    public class AutoRegisteringInterfaceGraphType<TSourceType> : InterfaceGraphType<TSourceType>
```
Shane32, line 13: So `AutoRegisteringInterfaceGraphType` and `AutoRegisteringObjectGraphType` are precisely identical, except one derives from `InterfaceGraphType<T>` and the other from `ObjectGraphType<T>`.  Duplicated code is minimal, as most of it is contained within the `AutoRegisteringHelper` internal class.

src/GraphQL/Types/Composite/AutoRegisteringInterfaceGraphType.cs
```diff
@@ -0,0 +1,154 @@
...
+            _excludedProperties = excludedProperties;
+            Name = typeof(TSourceType).GraphQLName();
+            ConfigureGraph();
+            foreach (var fieldType in ProvideFields())
+            {
+                _ = AddField(fieldType);
+            }
+        }
+
+        /// <inheritdoc cref="AutoRegisteringObjectGraphType{TSourceType}.ConfigureGraph"/>
```
Shane32, line 37: All of the comments are written suitably for interfaces just as well as classes, so to deduplicate the comments, I just pointed everything except the constructors to the other class.

Shane32:
@sungam3r I'd like to get this into 7.0 also.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -0,0 +1,809 @@
...
+using System.Reflection;
+using GraphQL.DataLoader;
+using GraphQL.Execution;
+using GraphQL.Types;
+using Microsoft.Extensions.DependencyInjection;
+using Moq;
+
+namespace GraphQL.Tests.Types;
+
+public class AutoRegisteringInterfaceGraphTypeTests
```
Shane32, line 14: Tests are mostly copied from AutoRegisteringObjectGraphTypeTests, modified for interfaces, with a few tests removed and a few added.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -0,0 +1,809 @@
...
+    [Theory]
+    [InlineData("{find(type:CAT){id name}}", @"{""data"":{""find"":{""id"":""10"",""name"":""Fluffy""}}}")]
+    [InlineData("{find(type:DOG){id name}}", @"{""data"":{""find"":{""id"":""20"",""name"":""Shadow""}}}")]
+    [InlineData("{find(type:CAT){id name ... on Cat { lives } ... on Dog { isLarge }}}", @"{""data"":{""find"":{""id"":""10"",""name"":""Fluffy"",""lives"":9}}}")]
+    [InlineData("{find(type:DOG){id name ... on Cat { lives } ... on Dog { isLarge }}}", @"{""data"":{""find"":{""id"":""20"",""name"":""Shadow"",""isLarge"":true}}}")]
+    [InlineData("{cat{id name}}", @"{""data"":{""cat"":{""id"":""10"",""name"":""Fluffy""}}}")]
+    [InlineData("{dog{id name}}", @"{""data"":{""dog"":{""id"":""20"",""name"":""Shadow""}}}")]
+    [InlineData("{cat{...frag}} fragment frag on IAnimal { id name }", @"{""data"":{""cat"":{""id"":""10"",""name"":""Fluffy""}}}")]
+    [InlineData("{dog{...frag}} fragment frag on IAnimal { id name }", @"{""data"":{""dog"":{""id"":""20"",""name"":""Shadow""}}}")]
+    public async Task EndToEndTest(string query, string expected)
```
Shane32, line 453: These tests demonstrate scenario described in the migration notes and tests typical behavior when writing a type-first schema.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -0,0 +1,809 @@
...
+            Query = query,
+            RequestServices = provider,
+        }).ConfigureAwait(false);
+        var serializer = provider.GetRequiredService<IGraphQLTextSerializer>();
+        var actual = serializer.Serialize(result);
+        actual.ShouldBeCrossPlatJson(expected);
+    }
+
+    [Fact]
+    public async Task ExecutesQueryWithInterfaceOnly()
```
Shane32, line 472: Verifies that an object can be returned from an interface even if it does not match any object graph types registered to the interface.  (Failed previously.)  With this change any graph can return an GraphQL interface even if the schema does not contain any object types that implement that GraphQL interface.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -0,0 +1,809 @@
...
+    }
+
+    public enum AnimalType
+    {
+        Cat,
+        Dog,
+    }
+
+    [Implements(typeof(IAnimal))]
+    public class Cat : IAnimal
```
Shane32, line 531: Notice the explicit interface dependency attribute.  I'm not sure what else to do here; how else to tell what interfaces should be mapped or not?  It would be nice to somehow notice what CLR types have been mapped to GraphQL interfaces and automatically enable those mappings... but that will have to be a future feature.

Shane32:
@sungam3r While I would like to merge this for v7, if you want to delay it, then I can post a separate PR with only the necessary infrastructure bits that need to be changed so I can use this code in my own environment.

sungam3r:
Reviewing...

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -681,11 +681,11 @@ public override void Modify(FieldType fieldType, bool isInputType)
     [Description("Test description")]
     private class TestClass_WithMultipleAttributes { }
 
-    private class TestOverrideDefaultName<T> : AutoRegisteringInputObjectGraphType<T>
+    private class TestOverrideDefaultName<T> : AutoRegisteringObjectGraphType<T>
```
sungam3r, line 684: 😄 👍

src/GraphQL/Attributes/ImplementsAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>
+        /// Returns the interface graph type that this graph type implements.
+        /// </summary>
+        public Type InterfaceGraphType { get; }
+
+        /// <inheritdoc/>
+        public override void Modify(IGraphType graphType)
+        {
+            if (graphType is not IImplementInterfaces graphType2)
+                throw new InvalidOperationException("The InterfaceAttribute can only be applied to graph types that implement IImplementInterfaces.");
```
sungam3r, line 40:
```suggestion
                throw new InvalidOperationException($"The {nameof(ImplementsAttribute)} can only be applied to graph types that implement {nameof(IImplementInterfaces)}.");
```

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: > for an interface, if IsTypeOf **returns false for all configured object types**, the interface graph type is used to resolve the query

Is it normal? It seems to me that this situation is a configuration error and when returning the interface we would just hide problem instead helping to solve it.

>  Previously some weird meaningless exception would occur. 

So maybe to track this case here and throw other exception with user-friendly error message?

sungam3r: I suppose `GetObjectGraphType` should always return something (not null) with the exception of root node.

src/GraphQL/Attributes/ImplementsAttribute.cs
```diff
@@ -0,0 +1,45 @@
+using GraphQL.Types;
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Marks the type as implementing a specified property.
```
sungam3r, line 6: property ?

src/GraphQL/Attributes/ImplementsAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                InterfaceGraphType = type;
+            }
+            else if (type.IsInterface)
+            {
+                InterfaceGraphType = typeof(GraphQLClrOutputTypeReference<>).MakeGenericType(type);
+            }
+            else
+            {
+                throw new ArgumentOutOfRangeException(nameof(type));
```
sungam3r, line 27:
```suggestion
                throw new ArgumentOutOfRangeException(nameof(type), "Can be a interface CLR type or an interface graph type.");
```

src/GraphQL/Attributes/ImplementsAttribute.cs
Shane32:
```suggestion
    /// Marks the type as implementing a specified interface.
```

src/GraphQL/Attributes/ImplementsAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace GraphQL
+{
+    /// <summary>
+    /// Marks the type as implementing a specified property.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public class ImplementsAttribute : GraphQLAttribute
+    {
+        /// <summary>
+        /// Marks the type as implementing a specified property.
```
Shane32, line 12:
```suggestion
        /// Marks the type as implementing a specified interface.
```

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: I do not see why the returned value must be any specific object graph type defined in the schema.  There is no GraphQL spec, as far as I can tell, that requires this.  Although such a configuration may be rare now, it may be more common if users use CLR types to generate their schema, and in so doing, map an interface for which they do not also map each and every CLR class that implements the interface.

Shane32: I compare this to a .NET public interface with one or more private implementations.

sungam3r: I think I understand your reasoning. Yes, field resolver returns some "runtime" object and then execution engine use this object against some graph type. For concrete graph type it is this type. For interfaces and unions this type should be somehow inferred. We have `ResolveType`+`IsTypeOf` to solve this task. If you use interface graph type as a "fallback" in case of `null`, then it will work I'm sure, but it is possible will work not at all as the author conceived. Imagine some query with fragment like
```graphql
query {
  user {
    ... on Administrator {
       login
       password
    }
    ... on Guest {
       image
    }
 }
}
```

`Query.user` is interface graph type `Person`. Schema misconfigured somehow (I don't know exactly how) so `GetObjectGraphType` returns `null` instead of `Administrator` (or `Guest`) and then you start using `Person` as resulting type. I think that `CollectFields` would return wrong selection set.

Shane32: I already added one test for this, but I'll add additional tests if you like.

Shane32: I added a whole bunch more tests.  Everything works exactly as expected.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: ok

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -428,7 +428,11 @@ private void AddType(IGraphType type, TypeCollectionContext context)
             {
                 foreach (var objectInterface in obj.Interfaces.List)
                 {
-                    if (AddTypeIfNotRegistered(objectInterface, context) is IInterfaceGraphType interfaceInstance)
+                    object typeOrError = RebuildType(objectInterface, false, context.ClrToGraphTypeMappings);
+                    if (typeOrError is string error)
+                        throw new InvalidOperationException($"The GraphQL implemented type for '{type.Name}' could not be derived implicitly. " + error);
```
sungam3r, line 433:
```suggestion
                        throw new InvalidOperationException($"The GraphQL implemented type '{objectInterface.Name}' for object graph type '{type.Name}' could not be derived implicitly. " + error);
```

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -224,5 +224,143 @@ internal static LambdaExpression GetParameterExpression(Type parameterType, Quer
...
+        {
+            if (typeof(TSourceType).IsInterface)
+            {
+                var types = typeof(TSourceType).GetInterfaces().Append(typeof(TSourceType));
```
sungam3r, line 332: Multiple interfaces may declare members with the same name and the same or different return type. The is no such problem for `typeof(TSourceType).GetProperties(flags)`.

docs2/site/docs/migrations/migration7.md
```diff
@@ -136,6 +136,85 @@ the graph type to be specified in order to define a field.
 As with the expression syntax or the `AutoRegisteringObjectGraphType`,
 CLR type mappings can be tailored via the `schema.RegisterTypeMapping()` methods.
 
+### 10. Interface graph types can be automatically built from CLR types
+
+Similar to how input and output types can be inferred from their CLR counterparts,
+now interface graph types can also be inferred from CLR types. This is possible
+with the new class `AutoRegisteringInterfaceGraphType<TSourceType>` which functions
+identically to `AutoRegisteringObjectGraphType<TSourceType>` except creates an
```
sungam3r, line 144:
```suggestion
identically to `AutoRegisteringObjectGraphType<TSourceType>` except it creates an
```

docs2/site/docs/migrations/migration7.md
```diff
@@ -136,6 +136,85 @@ the graph type to be specified in order to define a field.
...
+    AnimalType Type { get; }
+    string Name { get; }
+}
+
+public enum AnimalType { Cat, Dog }
+
+[Implements(typeof(IAnimal))]
+public class Cat : IAnimal
+{
+    [Id] public int Id => 10;
```
sungam3r, line 191: Shouldn't the attribute be inherited to not duplicate it here?

sungam3r: > It is important to ensure that any GraphQL attributes applied to members of the CLR types
are applied to both the interface and the classes alike, as the GraphQL.NET engine will build
distinct graph types for the interface and classes which implement those interfaces.

Well, I agree that GraphQL.NET engine will build distinct graph types but it _can_ use metainformation provided via interface/base class. By default all attributes have `Inherited = true`.

sungam3r: I mean `GetPropertyInfo("Id").GetCustomAttributes` do return 'Id' attribute. Am I wrong?

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
sungam3r: I see thought there is too much copy-pasted code in `AutoRegisteredInterfaceGraphType`. Over time, this will become a problem and we can forget that the code needs to be changed in two places. Now we are hostages of the situation with the existing `ObjectGraphType<TSourceType>` base class. I propose to add additional class or even make `AutoRegisteringHelper` non-static and compose it with `AutoRegisteredInterfaceGraphType/`AutoRegisteredObjectGraphType` delegating all real work to methods in `AutoRegisteringHelper`.

src/GraphQL/Attributes/ImplementsAttribute.cs
```diff
@@ -0,0 +1,45 @@
+using GraphQL.Types;
+
+namespace GraphQL
+{
+    /// <summary>
+    /// Marks the type as implementing a specified interface.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
```
sungam3r, line 8: Maybe add struct?

src/GraphQL/Attributes/ImplementsAttribute.cs
```diff
@@ -0,0 +1,45 @@
+using GraphQL.Types;
+
+namespace GraphQL
```
sungam3r, line 3: file-scoped namespaces for all new code

src/GraphQL/Attributes/ImplementsAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// Marks the type as implementing a specified interface.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public class ImplementsAttribute : GraphQLAttribute
```
sungam3r, line 9: I think you forgot to add `AllowMultiple = true` that is `false` by default.

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: I’m not sure that the tests are testing exactly what I'm talking about, but it is not so important to me now. Too narrow case.

src/GraphQL/Types/Composite/AutoRegisteringInterfaceGraphType.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// <inheritdoc cref="AutoRegisteringObjectGraphType{TSourceType}.BuildMemberInstanceExpression(MemberInfo)"/>
+        protected virtual LambdaExpression BuildMemberInstanceExpression(MemberInfo memberInfo)
+            => _sourceExpression;
+
+        private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression
+            = context => (TSourceType)(context.Source ?? ThrowSourceNullException());
+
+        private static object ThrowSourceNullException()
+        {
+            throw new NullReferenceException("IResolveFieldContext.Source is null; please use static methods when using an AutoRegisteringInterfaceGraphType as a root graph type or provide a root value.");
```
sungam3r, line 102: Throwing `NullReferenceException` is not a recommended practice, NRE is a runtime exception in fact and should not be thrown by any user code.

sungam3r:
I looked through all files so far.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: Well, I certainly don't want to change the OOP design of the class, if that's what you mean.  I'll deduplicate as much code as possible within that constraint.

docs2/site/docs/migrations/migration7.md
Shane32: I tested it.  Your attribute inheritance works for derived classes, but not classes which implement interfaces.

src/GraphQL/Attributes/ImplementsAttribute.cs
Shane32: Done

src/GraphQL/Attributes/ImplementsAttribute.cs
Shane32: Done

src/GraphQL/Attributes/ImplementsAttribute.cs
Shane32: Done + added test

src/GraphQL/Types/Composite/AutoRegisteringInterfaceGraphType.cs
Shane32: Changed to InvalidOperationException

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: That is true; it would cause an error when adding a second field with the same name to the interface.  I could write a test to verify this, but I did not see it worth the trouble to do so.  This of course assumes that the user does not tag the second field with the `[Name]` attribute to change the name within the generated graph type.

Note that if there is a complex inheritance structure, where multiple paths lead to the same member, this does not matter because `GetInterfaces()` return all inherited interfaces, and members cannot be redefined on children without the `new` keyword, which would of course break the graph because of a second field with the same name.  I added `GraphTypeCanImplementMultipleInterfaces` to demonstrate this, and there's already a multiple tests demonstrating inherited interfaces' members are used (as `IAnimal` inherits from `IObject`).

So unless you would like additional tests, I do not see any changes necessary to this code.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: I don't know if it's for the best, but I deduplicated every last possible line of code.  🤷‍♂️ 

As of now, I do not want to change the class design, such as doing anything like this:

```cs
AutoRegisteringInterfaceGraphType<THandler, TSource>
    where THelper : AutoRegisteringHelper
```

Doing so would add a whole layer of complexity to override something; and it is often enough to require one or a couple lines of code in one of the various methods.

Shane32: Unfortunately the deduplication changed almost all the code I'd already written, so now the changes are a bit more extensive.

sungam3r: I mean add private field of `AutoRegisteringHelper` type add call its methods. No changes in public API surface.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
sungam3r: It is enough for me that you are aware of the problem.

docs2/site/docs/migrations/migration7.md
```diff
@@ -136,6 +136,85 @@ the graph type to be specified in order to define a field.
...
+mapping provided by `AddAutoClrMappings()` or `AddAutoSchema()`, any CLR interface
+type is automatically mapped to a interface graph type rather than an object graph
+type.
+
+Note that auto-mapped CLR types do not automatically register or link
+to any GraphQL interfaces; such mapping needs to be specified via the new
+`ImplementsAttribute`. Similarly, CLR types not referenced directly in the schema
+need to be added to the schema manually or else no graph type will be generated for them.
+
+Below is a typcial example of how the new functionality can be used:
```
sungam3r, line 155:
```suggestion
Below is a typiсal example of how the new functionality can be used:
```

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
Shane32: Well then we can refactor the code later.  I've already wrote this entire PR twice over.

sungam3r (approved):

PR #3272, opened 2022-08-08
Rewrite IDocumentCache as IConfigureExecution

sungam3r:
fixes #3267

@Shane32 I need your help with tests using mocks. I'm not familiar with that type of tests and always spend enormous time to figure out what I need to do to fix them. I tried and reverted `GraphQLBuilderExtensionTests`.

Notes about PR:
1. `IDocumentCache` interface is removed. GraphQL.NET does not need it at all. `MemoryDocumentCache` as an implementation calls GetAsync/SetAsync methods itself. These methods still exist and `public` but only because of some tests. Would you like me to make them `protected`?
2. `DefaultDocumentCache` is not needed anymore. Default registration is removed.
3. GetAsync/SetAsync accept `ExecutionOptions` instead of `string query`. In my derived implementation of cache I want to use APQ hash as a key (from `ExecutionOptions.Extensions`) instead of potentially long query text. It should improve performance. 

TODO: change documentation, add migration notes

PR #3273, opened 2022-08-08
Add SortOrder property to IConfigureExecution

Shane32:
I think this is required for #3272

### DIFF
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -409,3 +409,18 @@ Field<MessageGraphType>("messages").ResolveStreamAsync(context => observable);
 ```
 
 Also `ComplexGraphType.Field<IntGraphType>("name")` now returns `FieldBuilder` instead of `FieldType`.
+
+### 11. `SortOrder` property added to `IConfigureExecution`
+
+If you have classes that implement `IConfigureExecution`, you will now need to also implement the
+added `SortOrder` property. The sort order determines the order that the `IConfigureExecution`
+instances are run, with the lowest value being run first.
+
+The default sort order of configurations are as follows:
+
+- 100: Apollo tracing
+- 200: Validation rules, document executers, and other 'initial' setting configurations
+- 300: User calls to ConfigureExecutionOptions
+- 400: Automatic Persisted Query processing
+- 500: Memory cache
+- 600: User calls to ConfigureExecution (configurable)
--- a/src/GraphQL.ApiTests/GraphQL.MemoryCache.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.MemoryCache.approved.txt
@@ -17,6 +17,7 @@ namespace GraphQL.Caching
     {
         public const string SUPPORTED_VERSION = "1";
         protected AutomaticPersistedQueriesExecutionBase() { }
+        public virtual float SortOrder { get; set; }
         protected virtual bool CheckHash(string hash, string query) { }
         protected virtual GraphQL.ExecutionResult CreateExecutionResult(GraphQL.ExecutionError error) { }
         public virtual System.Threading.Tasks.Task<GraphQL.ExecutionResult> ExecuteAsync(GraphQL.ExecutionOptions options, GraphQL.DI.ExecutionDelegate next) { }
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -343,7 +343,6 @@ namespace GraphQL
             where TConfigureExecution :  class, GraphQL.DI.IConfigureExecution { }
         public static GraphQL.DI.IGraphQLBuilder ConfigureExecutionOptions(this GraphQL.DI.IGraphQLBuilder builder, System.Action<GraphQL.ExecutionOptions> action) { }
         public static GraphQL.DI.IGraphQLBuilder ConfigureExecutionOptions(this GraphQL.DI.IGraphQLBuilder builder, System.Func<GraphQL.ExecutionOptions, System.Threading.Tasks.Task> action) { }
-        public static GraphQL.DI.IGraphQLBuilder ConfigureExecutionOptions(this GraphQL.DI.IGraphQLBuilder builder, System.Action<GraphQL.ExecutionOptions> action, float sortOrder) { }
         public static GraphQL.DI.IGraphQLBuilder ConfigureSchema(this GraphQL.DI.IGraphQLBuilder builder, System.Action<GraphQL.Types.ISchema> action) { }
         public static GraphQL.DI.IGraphQLBuilder ConfigureSchema(this GraphQL.DI.IGraphQLBuilder builder, System.Action<GraphQL.Types.ISchema, System.IServiceProvider> action) { }
         public static GraphQL.DI.IServiceRegister Register<TService>(this GraphQL.DI.IServiceRegister services, GraphQL.DI.ServiceLifetime serviceLifetime, bool replace = false)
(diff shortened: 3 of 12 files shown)


sungam3r:
I **really** doubt the correctness of this approach. Imagine that `IMiddleware` from ASP.NET has `SortOrder` property 😮  . I understand what problem you want to solve, but it seems to me that it is not necessary to try to solve it, introducing additional properties. The problem is solved by natural ordering on the caller side. Middlewares user can be mistaken in the same way and setup middlewares in the wrong order. I saw this a hundred times. What is solution? Just reorder `UseXXX` lines, that's all.

sungam3r:
The only drawback that I see here is the use of `AddXXX` methods instead of `UseXXX` similar to ASP.NET. The developers are accustomed to this terminology and for many of them the semantics are clear: order does not matter to `AddXXX` and does for `UseXXX`. I understand that in our situation the `UseXXX` naming is not very suitable 😕 .

Shane32:
> The only drawback that I see here is the use of `AddXXX` methods instead of `UseXXX` similar to ASP.NET. The developers are accustomed to this terminology and for many of them the semantics are clear: order does not matter to `AddXXX` and does for `UseXXX`. I understand that in our situation the `UseXXX` naming is not very suitable 😕 .

I think this design reinforces why this is necessary.  All configurations are simply mashed into a DI provider.  One user (was it you?) was talking about how multiple plugins in the same application call the same AddGraphQL registrations repeatedly within a larger application, so we needed to be sure the infrastructure could handle that.  (I was going to make another PR to enhance that aspect, after this is merged.)  We have no documentation of any nature that would indicate that the order of operations is important, none of the samples pay attention to the order, and the expectation from the `Add...` convention would be that any order is suitable.

With the conversion of memory cache to an `IConfigureExecution`, there is an immediate conflict between the memory cache and APQ -- the order must be correct or else the memory cache will be ineffective.  Also, any validation rules that do not add themselves to `CachedDocumentValidationRules` need to be installed prior to the memory cache, or else they will always run.

There is also precedent within other pluggable-type frameworks for a sort order.  Look at the open-source [NopCommerce](https://www.nopcommerce.com/en) application.  They support installable plugins (of which there are 1,500+ available), and the infrastructure is designed to run the `Startup.cs` file within each plugin according to the sort order defined by the developer.  Perhaps not identical to what we are doing, but having a sort order is not a new idea.

Additionally, having a sort order allows to create plugins that install themselves at multiple points within the execution.  This might be used for logging or Apollo Tracing purposes.

We may wish to continue with our development of `IConfigureExecution`-based features, such as converting `IDocumentBuilder` to an `IConfigureExecution` module as well.  As we do so, it will become more important to apply/install those "modules" in the correct order.  Having a sort order field makes this simple for the user to do so, handing over the ordering of these modules to the developer -- especially if we have any predefined (auto-installed) modules within the system.

Without a sort order, we will need to adjust all of our documentation to indicate _where_ to install the described feature, and answer the questions that inevitable will come up due to users having them in the wrong order.  I'm not looking forward to that.

sungam3r:
>  One user (was it you?) was talking about how multiple plugins in the same application call the same AddGraphQL registrations repeatedly within a larger application, so we needed to be sure the infrastructure could handle that

Yes, it was me. The semantics of `Add` APIs in DI should be idempotent.

> and the expectation from the Add... convention would be that any order is suitable.

Yes. So a side note of this discussion I propose to change names from `Add` to something else. Use?

> Additionally, having a sort order allows to create plugins that install themselves **at multiple points** within the execution.

How?

> We may wish to continue with our development of IConfigureExecution-based features, such as converting IDocumentBuilder to an IConfigureExecution module as well

I thought about this when looked into cache code.


Nevertheless, this proposal with the SortOrder property seems far from perfect to me. Maybe the problem can be solved somehow differently?

Shane32:
Something like this:
```cs
    public static IGraphQLBuilder AddLogger(this IGraphQLBuilder builder)
    {
        builder.ConfigureExecution(async (options, next) =>
        {
            options.RequestServices!.GetRequiredService<ILogger<IDocumentExecuter>>().LogTrace("starting execution");
            var ret = await next(options).ConfigureAwait(false);
            options.RequestServices!.GetRequiredService<ILogger<IDocumentExecuter>>().LogTrace("execution complete");
            return ret;
        }, -1);
        builder.ConfigureExecution((options, next) =>
        {
            options.RequestServices!.GetRequiredService<ILogger<IDocumentExecuter>>().LogTrace("configured options; caching query");
            return next(options);
        }, 350);
        builder.ConfigureExecution(async (options, next) =>
        {
            // log some information during execution through a document listener
            options.Listeners.Add(options.RequestServices!.GetRequiredService<MyLoggingListener>());
            // log unhandled exceptions
            var oldHandler = options.UnhandledExceptionDelegate;
            options.UnhandledExceptionDelegate = context =>
            {
                options.RequestServices!.GetRequiredService<ILogger<IDocumentExecuter>>().LogTrace("unhandled exception");
                return oldHandler(context);
            };
            options.RequestServices!.GetRequiredService<ILogger<IDocumentExecuter>>().LogTrace("executing query");
            var ret = await next(options).ConfigureAwait(false);
            options.RequestServices!.GetRequiredService<ILogger<IDocumentExecuter>>().LogTrace("done executing query");
            return ret;
        }, 999);
    }
```

Then `.AddLogging()` would inject itself into all the proper points of the execution.

Shane32:
> Nevertheless, this proposal with the SortOrder property seems far from perfect to me. Maybe the problem can be solved somehow differently?

I'm still listening, but changing `Add` to `Use` doesn't seem any better to me.  Now half the calls would be `Add` (e.g. `AddValidationRule` or `AddSerializer`) and half would be `Use` (e.g. `UseApolloTracing`).  And the `Use` commands would be wrapped within a call to `Add` (`AddGraphQL`).  And the semantics of `Use` -- the fact that they execute in a prescribed order -- is entirely dependent on whether the DI engine preserves the ordering of registrations.  Not to mention it's yet another change required to transition from v5 to v7.

Shane32:
I suppose it is an option to deprecate the old `Add` methods without removing them.

Shane32:
So would `AddMiddleware` be `Add` or `Use`?  It adds itself to the DI but also calls `FieldMiddleware.Use` ....

Shane32:
@sungam3r I'm looking at it, and it seems that `Use` would only apply to:
- UseMiddleware
- UseApolloTracing
- UseDocumentCache / UseMemoryCache
- UseAutomaticPersistedQueries

We already have 'ConfigureSchema` and `ConfigureExecution`, which do not start with "Add" and probably mesh fine with "Use".

I'll try a PR so we can look at it.

Shane32:
I'm trying it, but the problem is that `AddValidationRule` also uses `ConfigureExecutionOptions` to add the validation rule.  So essentially it because a `Use` rule also -- and indeed it matters in relation to a call to `UseMemoryCache` -- even though logically all the rules run simultaneously during document validation; there is no real 'order' to them.  Now I feel we are back to having a sort order so the validation rules can be added prior to `ConfigureExecution` calls.

Perhaps we do rename some methods to being with `Use`, but I still feel that the sort order property is added so that other `Add` calls (such as `AddValidationRule`) will run prior to any `Use` or `Configure` calls.  Perhaps all of the `Use`/`Configure` calls always execute together in order.  So the sort order is only:
- 100 -- all "AddValidationRule" and similar configuration calls
- 200 -- all "Use" and "Configure" calls, including UseApolloTracing and UseMemoryCache

So I stand by my recommendation to add a sort order.  We may not expose it to the user, except through the `IConfigureExecution` interface, but I would still add it nevertheless.

sungam3r:
Good, good observations and logger example. Indeed, now I doubt about `UseXXX`. Let me think some more about this topic. Maybe it's worth to think to rewrite `AddValidationRule` to **not use** ConfigureExecution ? I feel that we misuse our "configure execution tool" here and it conflicts with "real" features that can and should be chained in order-dependent sequence.

sungam3r (commented):
Superceded by #3277

PR #3275, opened 2022-08-09
Remove Arguments initialization from FieldBuilder

sungam3r:


Shane32 (approved):

Shane32:
@sungam3r Surprisingly enough, this broke a bunch of my code.  No big deal, but perhaps we should update the migration document in case other people have code like mine.

sungam3r:
Example?

Shane32:
Nothing I can share.  It was just code that created a field with a field builder, and then passed the FieldType instance to some other code which attempted to add an argument.

Shane32:
Just saying that sometimes the things you wouldn't expect to have any outside effect, does.

PR #3276, opened 2022-08-09
Introduce ErrorInfoProviderOptions.ExposeExceptionDetailsMode

### LINKED ISSUE #3268: What to do with exception type name in error message?
`ExposeExceptionStackTrace = true`:
```
GraphQL.ExecutionError: Exception of type 'System.ArgumentNullException' was thrown.
 ---> System.ArgumentNullException: Exception of type 'System.ArgumentNullException' was thrown.
 ---> System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   --- End of inner exception stack trace ---
   at GraphQL.Tests.Errors.ErrorInfoProviderTests.exposeExceptions_with_real_stack_trace() in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Errors\ErrorInfoProviderTests.cs:line 166
   --- End of inner exception stack trace ---
   at GraphQL.Tests.Errors.ErrorInfoProviderTests.exposeExceptions_with_real_stack_trace() in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Errors\ErrorInfoProviderTests.cs:line 170
```

`ExposeExceptionStackTrace = false`:
```
   Exception of type 'System.ArgumentNullException' was thrown.
```

We use `Exception.ToString()` when `ExposeExceptionStackTrace = true` and it appends not only stack trace in message but also prepend it with exception type like `GraphQL.ExecutionError: ` above. It worries me from time to time. In addition, I have been doubting for a long time that the stack trace should be displayed together with the error message, and not as separate field in extensions. 

For example error payload may look like in case of `ExposeExceptionStackTrace = true`
```json
{
    "errors": [
        {
            "message": "Exception of type 'System.ArgumentNullException' was thrown.",
            "extensions": {
                "stacktrace": "---> System.ArgumentNullException: Exception of type 'System.ArgumentNullException' was thrown.
 ---> System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   --- End of inner exception stack trace ---
   at GraphQL.Tests.Errors.ErrorInfoProviderTests.exposeExceptions_with_real_stack_trace() in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Errors\ErrorInfoProviderTests.cs:line 166
   --- End of inner exception stack trace ---
   at GraphQL.Tests.Errors.ErrorInfoProviderTests.exposeExceptions_with_real_stack_trace() in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Errors\ErrorInfoProviderTests.cs:line 170"
                "code": ""ARGUMENT_NULL"",
                "codes": [
                    ""ARGUMENT_NULL""
                ]
            }
        }
    ]
}
```

instead of 

```json
{
    "errors": [
        {
            "message": "GraphQL.ExecutionError: Exception of type 'System.ArgumentNullException' was thrown.
 ---> System.ArgumentNullException: Exception of type 'System.ArgumentNullException' was thrown.
 ---> System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   --- End of inner exception stack trace ---
   at GraphQL.Tests.Errors.ErrorInfoProviderTests.exposeExceptions_with_real_stack_trace() in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Errors\ErrorInfoProviderTests.cs:line 166
   --- End of inner exception stack trace ---
   at GraphQL.Tests.Errors.ErrorInfoProviderTests.exposeExceptions_with_real_stack_trace() in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Errors\ErrorInfoProviderTests.cs:line 170",
            "extensions": {
                "code": ""ARGUMENT_NULL"",
                "codes": [
                    ""ARGUMENT_NULL""
                ]
            }
        }
    ]
}
```

In other words I think it is worth to keep error messages "clean" without additional techical info - without stacktrace and exception type name.

sungam3r:
fixes #3268 

Example of `extensions.stacktrace`. Note that there are no error messages - only stacktraces and exception types.

```
GraphQL.ExecutionError:
   at GraphQL.Tests.Errors.ErrorInfoProviderTests.exposeExceptions_with_real_stack_trace_in_extensions() in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Errors\ErrorInfoProviderTests.cs:line 186
[inner] System.ArgumentNullException:
   at GraphQL.Tests.Errors.ErrorInfoProviderTests.exposeExceptions_with_real_stack_trace_in_extensions() in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Errors\ErrorInfoProviderTests.cs:line 182
[inner] System.ArgumentOutOfRangeException:
   <no stacktrace>
```

Or... should I just set `error.ToString()` into `extensions.stacktrace` ?


src/GraphQL/Execution/ErrorInfoProvider.cs
```diff
@@ -65,14 +69,32 @@ public virtual ErrorInfo GetInfo(ExecutionError executionError)
                         extensions.Add("number", number);
                     if (data != null)
                         extensions.Add("data", data);
+                    if (stacktrace != null)
+                        extensions.Add("stacktrace", stacktrace);
```
Shane32, line 73:
```suggestion
                        extensions.Add("stackTrace", stacktrace);
```
Names are camelCase, not lowercase, so I would think this should be "stackTrace"

src/GraphQL/Execution/ErrorInfoProviderOptions.cs
```diff
@@ -12,8 +12,11 @@ public class ErrorInfoProviderOptions
         /// </summary>
         public bool ExposeExceptionStackTrace { get; set; }
 
+        /// <inheritdoc cref="Execution.ExposeExceptionStackTraceMode"/>
+        public ExposeExceptionStackTraceMode ExposeExceptionStackTraceMode { get; set; }
+
         /// <summary>
-        /// Specifies whether the extensions property, including by default the 'code', 'codes' and 'data' properties, should be serialized.
+        /// Specifies whether the extensions property, including by default the 'code', 'codes', 'data' and 'stacktrace' properties, should be serialized.
```
Shane32, line 19:
```suggestion
        /// Specifies whether the extensions property, including by default the 'code', 'codes', 'data' and 'stackTrace' properties, should be serialized.
```

src/GraphQL/Execution/ErrorInfoProviderOptions.cs
```diff
@@ -34,4 +37,20 @@ public class ErrorInfoProviderOptions
...
+    /// </summary>
+    public enum ExposeExceptionStackTraceMode
+    {
+        /// <summary>
+        /// Exception stack trace is located along with exception message.
+        /// </summary>
+        Message,
+
+        /// <summary>
+        /// Exception stack trace is located in "extensions.stacktrace" separately from exception message itself.
```
Shane32, line 52:
```suggestion
        /// Exception stack trace is located in "extensions.stackTrace" separately from exception message itself.
```

src/GraphQL/Execution/ErrorInfoProvider.cs
```diff
@@ -53,8 +54,11 @@ public virtual ErrorInfo GetInfo(ExecutionError executionError)
                     codes = null;
                 var number = _options.ExposeCode && executionError is ValidationError validationError ? validationError.Number : null;
                 var data = _options.ExposeData && executionError.Data?.Count > 0 ? executionError.Data : null;
+                var stacktrace = _options.ExposeExceptionStackTrace && _options.ExposeExceptionStackTraceMode == ExposeExceptionStackTraceMode.Extensions
+                    ? BuildStackTrace(executionError)
```
Shane32, line 58:
```suggestion
                    ? executionError.ToString()
```
I would just store `ToString()` which contains all the details, including the type name and description of each inner exception.

Example:

```
System.NullReferenceException: Outer exception message
 ---> System.InvalidOperationException: Center exception message
 ---> System.ApplicationException: Innermost exception message
   at GraphQL.MicrosoftDI.Tests.AdapterTests.AllowNullProvider() in C:\Users\Shane\source\repos\graphql-dotnet\graphql-dotnet\src\GraphQL.MicrosoftDI.Tests\AdapterTests.cs:line 26
   --- End of inner exception stack trace ---
   at GraphQL.MicrosoftDI.Tests.AdapterTests.AllowNullProvider() in C:\Users\Shane\source\repos\graphql-dotnet\graphql-dotnet\src\GraphQL.MicrosoftDI.Tests\AdapterTests.cs:line 30
   --- End of inner exception stack trace ---
   at GraphQL.MicrosoftDI.Tests.AdapterTests.AllowNullProvider() in C:\Users\Shane\source\repos\graphql-dotnet\graphql-dotnet\src\GraphQL.MicrosoftDI.Tests\AdapterTests.cs:line 35
```

Perhaps the property name should be called "details" instead of "stackTrace", however.

src/GraphQL/Execution/ErrorInfoProvider.cs
```diff
@@ -65,14 +69,32 @@ public virtual ErrorInfo GetInfo(ExecutionError executionError)
...
+                {
+                    AppendEntry();
+                    ex = ex.InnerException!;
+                    prefix = "[inner] ";
+                }
+
+                return sb.ToString();
+
+                void AppendEntry() => sb.Append(prefix).Append(ex.GetType().FullName).AppendLine(":").AppendLine(ex.StackTrace ?? "   <no stacktrace>");
+            }
```
Shane32, line 97: This basically just replicates `ToString()` in a different manner and without the messages of exceptions (outer or inner exceptions).  I would just remove and replace with `ToString()`.

```suggestion
```

src/GraphQL/Execution/ErrorInfoProvider.cs
Shane32:
```suggestion
                        extensions.Add("details", stacktrace);
```
Or perhaps "details" as it will store the type names, messages and stack trace of outer and inner exceptions -- all information you'd want if enabling "stack traces".

sungam3r: So... if that piece of information is some sort of _details_ then maybe just use `exception.ToString()` and not bother with manual building formatted string?

src/GraphQL/Execution/ErrorInfoProvider.cs
sungam3r: Good. I would too. I just wanted to confirm this intent.

src/GraphQL/Execution/ErrorInfoProvider.cs
Shane32: I think that's what I suggested...  use "details" as the key and it equals `exception.ToString()`.

sungam3r: OK but then `ExposeExceptionStackTrace` is not good enough name and something like `ExposeExceptionDetails`/`ExposeExceptionDetailsMode` suits better.

sungam3r: StackTrace is just a StackTrace but we really have stack traces + exception types + exception messages

Shane32: Maybe just add `ExposeExceptionDetails` and deprecate the old property

Shane32: Or leave the old property as-is and add the new property...

sungam3r:
Done

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -2419,6 +2426,11 @@ namespace GraphQL.Types
         public System.Type Type { get; set; }
         public virtual void ApplyAttributes() { }
         public virtual System.Type ConstructGraphType() { }
+        public class f : System.IEquatable<GraphQL.Types.TypeInformation.f>
+        {
+            public f(int i) { }
+            public int i { get; set; }
+        }
```
Shane32, line 2433: Remove?

Shane32:
@sungam3r do you think Message should be the default mode?

We could have Extension be the default, and write code within the ExposeExceptionStackTrace property to set it equal to Message

src/GraphQL/Execution/ErrorInfoProviderOptions.cs
```diff
@@ -10,10 +10,25 @@ public class ErrorInfoProviderOptions
         /// <summary>
         /// Specifies whether stack traces should be serialized.
         /// </summary>
-        public bool ExposeExceptionStackTrace { get; set; }
+        [Obsolete("Use ExposeExceptionDetails property instead")]
+        public bool ExposeExceptionStackTrace
+        {
+            get => ExposeExceptionDetails;
+            set => ExposeExceptionDetails = value;
```
Shane32, line 17:
```suggestion
            set
            {
                ExposeExceptionDetails = value;
                if (value)
                    ExposeExceptionDetailsMode = Message;
            }
```
Just an idea.....

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
sungam3r: ooops, leak of local experiments

sungam3r: done

src/GraphQL/Execution/ErrorInfoProviderOptions.cs
sungam3r: > do you think Message should be the default mode?

for v5.4 yes, for v7 I would like to switch it to Extensions and remove `ExposeExceptionStackTrace`

Shane32: I don't see any good reason to remove the member now, immediately before v7 is released.  It can be a deprecated member for v7.

I also would not switch the default of the new property between versions.

I would do as I posted above -- setting the old property retains the old behavior, while setting the new property defaults to the new behavior.  There is no breaking change for 5.4 and no breaking change when 7 is released.  The [Obsolete] comment needs to be updated to reflect this of course.

sungam3r: ExposeExceptionDetailsMode defaults to `Message`. Why to reset it to Message again into setter ?

sungam3r: ExposeExceptionStackTrace is the same thing as new property. It is just a way to not renaming it and introduce breaking changes. I don't want to bring some special behavior into setter.

sungam3r: The goal is to eventually remove `ExposeExceptionStackTrace` and _maybe_ change default mode from Message to Extensions. 

I'm done for today, too tired to think more.

Shane32: > ExposeExceptionDetailsMode defaults to Message. Why to reset it to Message again into setter ?

You said you wanted Extension to be the default, which I agree with.

> I don't want to bring some special behavior into setter.

Then you can't release this in 5.4.  It must be v7.  If you release it in 5.4 then it will cause a breaking change that is not marked by an obsolete flag in v7.  There is no reason to do so with v7 so near to being released.

Either that or Message should continue to be the default after v7 is released.

Or you go with my 3rd option, where using the new option uses the new default, and using the old option uses the old default.

Shane32: I really don't see why not to do as I suggested.  It allows for non-API-breaking behavior in both v5.4 and v7, while having the new default value of 'Extension' for all new code.  Then we can remove the property in v8.

>  I don't want to bring some special behavior into setter.

Nor do I, but in this case, it is clearly the best choice for the scenario.  (in my opinion)

Shane32: > The goal is to eventually remove `ExposeExceptionStackTrace` and _maybe_ change default mode from Message to Extensions.

If there is no goal to either change the default from Message or to remove the old property in v7, then we can merge as-is.  We should think about it and decide one way or the other.

sungam3r: I was busy again. Now I looked at the code again and realized that the difference was not fundamental to me at all. As a result, the old property will eventually be deleted, and now I will not use it. I applied suggestion.

src/GraphQL/Execution/ErrorInfoProviderOptions.cs
```diff
@@ -10,10 +10,30 @@ public class ErrorInfoProviderOptions
...
 
         /// <summary>
-        /// Specifies whether the extensions property, including by default the 'code', 'codes' and 'data' properties, should be serialized.
+        /// Specifies whether detailed exception information (exception types, stack traces, inner exceptions)
+        /// should be serialized.
+        /// </summary>
+        public bool ExposeExceptionDetails { get; set; }
+
+        /// <inheritdoc cref="Execution.ExposeExceptionDetailsMode"/>
+        public ExposeExceptionDetailsMode ExposeExceptionDetailsMode { get; set; }
```
Shane32, line 32:
```suggestion
        public ExposeExceptionDetailsMode ExposeExceptionDetailsMode { get; set; } = ExposeExceptionDetailsMode.Extensions;
```

I thought we both agreed this would be better?  Note that with the code in the deprecated `ExposeExceptionStackTrace` property, this is a non-breaking change.

sungam3r: ok

sungam3r:
I retargeted to `develop` and added docs.

Shane32 (approved):
API approvals need to be updated, by the way.

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -1154,10 +1157,14 @@ namespace GraphQL.Execution
     public class ExecutionStrategyRegistration : System.IEquatable<GraphQL.Execution.ExecutionStrategyRegistration>
     {
         public ExecutionStrategyRegistration(GraphQL.Execution.IExecutionStrategy Strategy, GraphQLParser.AST.OperationType Operation) { }
```
sungam3r, line 1159:
```suggestion
        public ExecutionStrategyRegistration(GraphQL.Execution.IExecutionStrategy Strategy, GraphQLParser.AST.OperationType Operation) { }
        protected virtual System.Type EqualityContract { get; }
```

src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
```diff
@@ -1154,10 +1157,14 @@ namespace GraphQL.Execution
     public class ExecutionStrategyRegistration : System.IEquatable<GraphQL.Execution.ExecutionStrategyRegistration>
     {
         public ExecutionStrategyRegistration(GraphQL.Execution.IExecutionStrategy Strategy, GraphQLParser.AST.OperationType Operation) { }
```
sungam3r, line 1159:
```suggestion
        public ExecutionStrategyRegistration(GraphQL.Execution.IExecutionStrategy Strategy, GraphQLParser.AST.OperationType Operation) { }
        protected virtual System.Type EqualityContract { get; }
```

sungam3r:
I hope @stakx will accept fix in API approver to not bother anymore with this tiny property.

docs2/site/docs/migrations/migration7.md
```diff
@@ -228,6 +228,17 @@ methods defined on an interface. This functionality is available in C# 8.0 and l
 When building a graph type from an interface, methods are built for all inherited methods
 as well as the specified interface's methods. This is by design.
 
+### 11. Add `ErrorInfoProviderOptions.ExposeExceptionDetailsMode` property
```
sungam3r, line 231: @Shane32 Please review docs for grammar flaws.

Shane32: I looked through it briefly, but I thought perhaps I would take a pass over the entire document before we release 7.0 -- I was going to do that today (my time), after you merge this PR.  Perhaps you can review my changes tonight (your time) and if it's okay with you, I'll release 7.0 after that (tonight my time).

Mainly I'm thinking of re-ordering migration notes, either to make them flow better, or to draw attention to important notes.  Probably fairly light changes.

sungam3r: So I merge this one.

PR #3277, opened 2022-08-09
Change to Use calls with IConfigureExecution.SortOrder

Shane32:
Note that AddMemoryCache has not been changed to UseMemoryCache here because it still is based on IDocumentCache.

Sort order is added to `IConfigureExecution`, but is only used internally, unless a user writes a custom `IConfigureExecution` implementation.

Ordering is as follows:

1. All calls to `Add` and `ConfigureExecutionOptions`, in the order they are called.
2. All calls to `Use` and `ConfigureExecution`, in the order they are called.

This essentially acts as though there were two separate interfaces, `IConfigureExecutionOptions` and `IConfigureExecution`, which there was in v5, I believe.  We could revert the removal of `IConfigureExecutionOptions` but I would rather eliminate the interface and add the property.

We could also add a default implementation for `SortOrder` so that there is no breaking change for `IConfigureExecution` implementations (C# 8 or newer codebases).

Another option would be to sort as follows by default:
1. All calls to `Add`
2. All calls to `ConfigureExecutionOptions`
3. All calls to `Use` and `ConfigureExecution`, in the order they are called.

Or:
1. All calls to `Add`
2. All calls to `Use`, `ConfigureExecutionOptions` and `ConfigureExecution`, in the order they are called

However, the suggestion presented in this PR is simplest both internally, and for additional plugins, and conceptually.  In other words, the AddValidationRule extension just calls ConfigureExecutionOptions.  No need to call some private/internal member with a customized sort order value.  And third-party code similarly can easily hook into the pre or post configuration by simply calling either ConfigureExecutionOptions or ConfigureExecution.

This design also allows for user code to call ConfigureExecutionOptions to have code run prior to any other options being configured, and allows for third-party code to hook into any point of the sort order if desired.

Being that there is only two default sort orders now (100 and 200), custom logging or similar code that attempts to inject into the pipeline at multiple points would only be able to do so at three points -- before 100, between 100 and 200, and after 200.  This is not so bad; the logging example was just a hypothetical example anyway.

### DIFF
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -72,7 +72,6 @@ A list of the available extension methods is below:
 
 | Method    | Description / Notes | Library |
 |-----------|---------------------|---------|
-| `AddApolloTracing`      | Registers and enables metrics depending on the supplied arguments, and adds Apollo Tracing data to the execution result | |
 | `AddAutoClrMappings`    | Configures unmapped CLR types to use auto-registering graph types | |
 | `AddAutoSchema`         | Registers a schema based on CLR types | |
 | `AddClrTypeMappings`    | Scans the specified assembly for graph types intended to represent CLR types and registers them within the schema | |
@@ -87,7 +86,6 @@ A list of the available extension methods is below:
 | `AddGraphTypes`         | Scans the specified assembly for graph types and registers them within the DI framework | |
 | `AddGraphTypeMappingProvider` | Registers a graph type mapping provider for unmapped CLR types | |
 | `AddMemoryCache`        | Registers the memory document cache and configures its options | GraphQL.MemoryCache |
-| `AddMiddleware<>`       | Registers the specified middleware and configures it to be installed during schema initialization | |
 | `AddNewtonsoftJson`     | Registers the serializer that uses Newtonsoft.Json as its underlying JSON serialization engine | GraphQL.NewtonsoftJson |
 | `AddSchema<>`           | Registers the specified schema | |
 | `AddSelfActivatingSchema<>` | Registers the specified schema which will create instances of unregistered graph types during initialization | |
@@ -98,10 +96,14 @@ A list of the available extension methods is below:
 | `ConfigureExecutionOptions` | Configures execution options at runtime | |
 | `ConfigureSchema`       | Configures schema options when the schema is initialized | |
 | `Configure<TOptions>`   | Used by extension methods to configures an options class within the DI framework | |
+| `UseApolloTracing`      | Registers and enables metrics depending on the supplied arguments, and adds Apollo Tracing data to the execution result | |
+| `UseAutomaticPersistedQueries` | Enables Automatic Persisted Queries support | GraphQL.MemoryCache |
+| `UseMiddleware<>`       | Registers the specified middleware and configures it to be installed during schema initialization | |
 
 The above methods will register the specified services typically as singletons unless otherwise specified. Graph types and middleware are registered
 as transients so that they will match the schema lifetime. So with a singleton schema, all services are effectively singletons.
 
+Methods that start with `Configure` and `Use` execute in the order they appear, which may be important. For instance
 Custom `IGraphQLBuilder` extension methods typically rely on the `Services` property of the builder in order to register services
 with the underlying dependency injection framework. The `Services` property returns a `IServiceRegister` interface which has these methods:
 
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -103,7 +103,10 @@ A list of the available extension methods is below:
 The above methods will register the specified services typically as singletons unless otherwise specified. Graph types and middleware are registered
 as transients so that they will match the schema lifetime. So with a singleton schema, all services are effectively singletons.
 
-Methods that start with `Configure` and `Use` execute in the order they appear, which may be important. For instance
+Methods that start with `Configure` and `Use` execute in the order they appear, which may be important. For instance,
+calling `UseMemoryCache` prior to `UseAutomaticPersistedQueries` would result in the memory cache being unable to
+cache any APQ queries.
+
 Custom `IGraphQLBuilder` extension methods typically rely on the `Services` property of the builder in order to register services
 with the underlying dependency injection framework. The `Services` property returns a `IServiceRegister` interface which has these methods:
 
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -105,9 +105,8 @@ as transients so that they will match the schema lifetime. So with a singleton s
 
 Calls to `ConfigureExecutionOptions` and methods that start with `Add` will execute first, in the order they
 appear, followed by calls to `ConfigureExecution` and methods that start with `Use`. The order of the calls
-may be important. For instance,
-calling `UseMemoryCache` prior to `UseAutomaticPersistedQueries` would result in the memory cache being unable to
-cache any APQ queries.
+may be important. For instance, calling `UseMemoryCache` prior to `UseAutomaticPersistedQueries` would result in
+the memory cache being unable to cache any APQ queries.
 
 Custom `IGraphQLBuilder` extension methods typically rely on the `Services` property of the builder in order to register services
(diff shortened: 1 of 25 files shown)


sungam3r:
Was on road today and busy, 2 am now, I'll try to review thi and other pending PRs tomorrow.

src/GraphQL/DI/IConfigureExecution.cs
```diff
@@ -31,5 +46,7 @@ public ConfigureExecution(Func<ExecutionOptions, ExecutionDelegate, Task<Executi
 
         public Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
             => _action(options, next);
+
+        public virtual float SortOrder => GraphQLBuilderExtensions.SORT_ORDER_CONFIGURATION;
```
sungam3r, line 50:
```suggestion
        public float SortOrder => GraphQLBuilderExtensions.SORT_ORDER_CONFIGURATION;
```

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -68,10 +68,13 @@ public DocumentExecuter(IDocumentBuilder documentBuilder, IDocumentValidator doc
         private ExecutionDelegate BuildExecutionDelegate(IEnumerable<IConfigureExecution> configurations)
         {
             ExecutionDelegate execution = CoreExecuteAsync;
-            var configurationArray = configurations.ToArray();
-            for (var i = configurationArray.Length - 1; i >= 0; i--)
+
+            // OrderBy here performs a stable sort; that is, if the sort order of two elements are equal,
+            // the order of the elements are preserved. The order is reversed since each execution wraps
+            // the prior configured executions. OrderByDescending is not used because that would result
```
sungam3r, line 74: Do you mean `OrderByDescending ` is not stable?

src/GraphQL.MemoryCache/MemoryCacheGraphQLBuilderExtensions.cs
```diff
@@ -23,19 +23,27 @@ public static IGraphQLBuilder AddMemoryCache(this IGraphQLBuilder builder, Actio
...
+        => builder.UseAutomaticPersistedQueries(action == null ? null : (options, _) => action(options));
 
-    /// <inheritdoc cref="AddAutomaticPersistedQueries(IGraphQLBuilder, Action{AutomaticPersistedQueriesCacheOptions})"/>
-    public static IGraphQLBuilder AddAutomaticPersistedQueries(this IGraphQLBuilder builder, Action<AutomaticPersistedQueriesCacheOptions, IServiceProvider>? action)
+    /// <inheritdoc cref="UseAutomaticPersistedQueries(IGraphQLBuilder, Action{AutomaticPersistedQueriesCacheOptions})"/>
+    public static IGraphQLBuilder UseAutomaticPersistedQueries(this IGraphQLBuilder builder, Action<AutomaticPersistedQueriesCacheOptions, IServiceProvider>? action)
     {
-        builder.Services
-            .Configure(action)
-            .TryRegister<IConfigureExecution, AutomaticPersistedQueriesExecution>(ServiceLifetime.Singleton, RegistrationCompareMode.ServiceTypeAndImplementationType);
```
sungam3r, line 38: I see switching from TryRegister to Register semantics and I think it's fine in general case.

sungam3r (approved):
I must admit that the breaks in reviews make positive changes to my perception. Now the `SortOrder` property no longer seems bad to me. At least this is rational behavior especially with only two well-known and properly described "standard" values.

Good work.

sungam3r:
I'm sure that it is better than
изображение

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: No, it is stable, which produces different results compared to ascending reversed when it encounters entries with the same sort value.

Assuming B1 == B2:

| Original | Ascending | Ascending Reversed | Descending |
|-|-|-|-|
| A, D, B1, B2 | A, B1, B2, D | D, B2, B1, A | D, B1, B2, A |
| A, D, B2, B1 | A, B2, B1, D | D, B1, B2, A | D, B2, B1, A |

In this case we require a stable ascending sort, reversed.  Descending will not do.  Note that MS documented the fact that OrderBy performs a stable sort, so we can rely on this.

If you have a better way of explaining this, please suggest a revision to the comment - thanks!

src/GraphQL.MemoryCache/MemoryCacheGraphQLBuilderExtensions.cs
Shane32: If you remember, I suggested we do not implement `RegistrationCompareMode.ServiceTypeAndImplementationType` at all.  But since we have it, I was going to write a subsequent PR to change many of the `AddSomething<T>()` registrations where it would be logical to do so (not the `AddSomething<T>(T instance)` or `AddSomething<T>(factory)` calls).

As of now, this particular change here is made to be similar to the rest of the similar calls in the library -- but again, I was going to make another PR which would change this back.

Shane32:
> I must admit that the breaks in reviews make positive changes to my perception. Now the `SortOrder` property no longer seems bad to me. At least this is rational behavior especially with only two well-known and properly described "standard" values.
> 
> Good work.

I'm glad you like it.  I do find that we often do come up with a better proposal than either one of us originally started with.

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -98,10 +96,17 @@ A list of the available extension methods is below:
...
 | `ConfigureSchema`       | Configures schema options when the schema is initialized | |
 | `Configure<TOptions>`   | Used by extension methods to configures an options class within the DI framework | |
+| `UseApolloTracing`      | Registers and enables metrics depending on the supplied arguments, and adds Apollo Tracing data to the execution result | |
+| `UseAutomaticPersistedQueries` | Enables Automatic Persisted Queries support | GraphQL.MemoryCache |
+| `UseMiddleware<>`       | Registers the specified middleware and configures it to be installed during schema initialization | |
 
 The above methods will register the specified services typically as singletons unless otherwise specified. Graph types and middleware are registered
 as transients so that they will match the schema lifetime. So with a singleton schema, all services are effectively singletons.
 
+Methods that start with `Configure` and `Use` execute in the order they appear, which may be important. For instance,
```
Shane32, line 106:
```suggestion
Calls to `ConfigureExecution` and methods that start with `Use` execute in the order they appear, which may be important. For instance,
```

docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -98,10 +96,19 @@ A list of the available extension methods is below:
...
+| `UseMiddleware<>`       | Registers the specified middleware and configures it to be installed during schema initialization | |
 
 The above methods will register the specified services typically as singletons unless otherwise specified. Graph types and middleware are registered
 as transients so that they will match the schema lifetime. So with a singleton schema, all services are effectively singletons.
 
+Calls to `ConfigureExecutionOptions` and methods that start with `Add` will execute first, in the order they
+appear, followed by calls to `ConfigureExecution` and methods that start with `Use`. The order of the calls
+may be important. For instance,
+calling `UseMemoryCache` prior to `UseAutomaticPersistedQueries` would result in the memory cache being unable to
+cache any APQ queries.
```
Shane32, line 110:
```suggestion
may be important. For instance, calling `UseMemoryCache` prior to `UseAutomaticPersistedQueries` would result in
the memory cache being unable to cache any APQ queries.
```

Shane32:
@sungam3r I updated the documentation a little.  Let me know if you suggest any changes, and I'll do so in another PR.

Of note, I did not add migration notes for these changes.  The old methods are marked with Obsolete and a proper note describing the required change, and are fully functional.  I can do so if you think I should.

I also had to update the test due to the fact that SortOrder is not virtual anymore.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: understood

src/GraphQL.MemoryCache/MemoryCacheGraphQLBuilderExtensions.cs
sungam3r: I was thinking about it like about `UseDeveloperExceptionPage`. If one wants to use it multiple times - why not ?

sungam3r:
I already said somewhere that I don't treat obsoleting APIs as a breaking change. Deletion of APIs is a breaking change.

PR #3265, opened 2022-08-03
Feature/assembly strongname

anderslaub:
## Motivation

Many solutions based on .NET Framework (net4x) has both direct and inherited dependencies on GraphQl-dotnet.  

Since the currently released assemblies does not have a strongly typed name, developers who work on such solutions are forced to depend on a compatible version of graphql-dotnet as their inherited dependency. This often means v2.4.0 due to breaking changes introduced in later major versions.

By strong name signing the assemblies that target netstandard2.0/1, developers who are currently stuck on an old inherited version will be able to do codebase versioned assembly binding.

_example:_
```xml
            <dependentAssembly xdt:Transform="InsertIfMissing"  xdt:Locator="Condition(asmv1:assemblyIdentity/@name='GraphQl')">
                <assemblyIdentity name="GraphQl"  publicKeyToken="e6238258560628ee" />
                <codeBase version="5.0.0.0" href="bin\deps\GraphQL.dll" />
            </dependentAssembly>

            <dependentAssembly xdt:Transform="InsertIfMissing"  xdt:Locator="Condition(asmv1:assemblyIdentity/@name='GraphQl.DataLoader')">
                <assemblyIdentity name="GraphQl.DataLoader" publicKeyToken="e6238258560628ee" />
                <codeBase version="5.0.0.0" href="bin\deps\GraphQl.DataLoader.dll" />
            </dependentAssembly>

            <dependentAssembly xdt:Transform="InsertIfMissing"  xdt:Locator="Condition(asmv1:assemblyIdentity/@name='GraphQl.NewtonsoftJson')">
                <assemblyIdentity name="GraphQL.NewtonsoftJson" publicKeyToken="e6238258560628ee" />
                <codeBase version="5.0.0.0" href="bin\deps\GraphQL.NewtonsoftJson.dll" />
            </dependentAssembly>

            <dependentAssembly xdt:Transform="InsertIfMissing"  xdt:Locator="Condition(asmv1:assemblyIdentity/@name='GraphQLParser')">
                <assemblyIdentity name="GraphQLParser" publicKeyToken="e6238258560628ee" />
                <codeBase version="8.0.0.0" href="bin\deps\GraphQLParser.dll" />
            </dependentAssembly>
```

Signing the assemblies also enable other use-cases which has otherwise been prevented by legacy requirements for strong-named assembly (such as Visual Studio extension development).  

[...]

PR #3278, opened 2022-08-09
Remove dependency on Nito.AsyncEx.Context

Shane32:
It was only used, I believe, because of missing ConfigureAwait calls, which have been fixed, and currently there is a build error for any missing ConfigureAwaits.

Required by:
- #3265

### DIFF
--- a/.github/workflows/test-code.yml
+++ b/.github/workflows/test-code.yml
@@ -37,7 +37,7 @@ jobs:
           dotnet-version: |
             3.1.x
             5.0.x
-            6.0.x
+            6.0.303
           source-url: https://nuget.pkg.github.com/graphql-dotnet/index.json
         env:
           NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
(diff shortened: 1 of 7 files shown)


.github/workflows/test-code.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           dotnet-version: |
             3.1.x
             5.0.x
-            6.0.x
+            6.0.303
```
Shane32, line 40: Api approvals broke with the .NET 6.0.400 SDK released today.  Not sure why; they run fine on my PC.  For some reason it would seem that the api approvals generate differently on Linux than Windows.

Shane32: https://github.com/PublicApiGenerator/PublicApiGenerator/issues/245

Shane32:
@sungam3r I'm merging this to effectively continue with strong-name PR.  Let me know if you have any comments when you have time.

sungam3r:
I started to look into this PR today but then was busy and returned back only now. I'm fine with these changes but one question still remains. Why do we need to sign test assemblies?

Shane32:
I was thinking the same thing, but I thought that perhaps it “tests” that signed assemblies work?? Anyway it didn’t hurt to sign them so that’s the way it’s set up

sungam3r:
ok

PR #3281, opened 2022-08-10
[WIP] Simplified assembly strong-name signing

Shane32:


anderslaub (commented):
The only comment I have is the same as I wrote [here  ](https://github.com/graphql-dotnet/parser/pull/265/files#r942635223) regarding keeping the _ underscore prefix om `_FriendAssembliesPublicKey`. I am surprised that this naming is not recognized as showing the the intent / scope of the property.

Besides that all changes here are making it nice and simple. 

Using the `CustomBeforeMicrosoftCommonTargets` property to load the targets file was only for doing conditional signing on targetframework as I started out with - so good that it got cleaned when not needed.

Shane32:
Replaced by #3265

PR #3288, opened 2022-08-15
Restore IDocumentCache removal as deprecated members

Shane32:
Reduces amount of changes necessary for migration to v7 when a custom document cache implementation was in use.

This was not done for `IComplexityAnalyzer` because the options were removed from `ExecutionOptions` and so there is no simple way to provide a direct replacement.

The `DocumentExecuter` constructors have not been restored because it would be necessary to restore `IComplexityAnalyzer` also.

@sungam3r Thoughts?  We don't have to merge this PR, but I feel we should make an attempt at reducing API diff between major versions when it can be avoided.

### DIFF
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -334,16 +334,16 @@ Using the `IGraphQLBuilder` interface to configure the GraphQL.NET execution eng
 
 ### 5. Changes in document caching
 
-To make work with document cache more flexible and allow some advanced use-cases this component
-was moved out of the GraphQL.NET execution engine. There is no more `IDocumentCache` interface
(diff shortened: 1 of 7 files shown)


sungam3r:
I am skeptical about this. I would not. Firstly, I doubt that many people wrote their implementations of cache. I'm sure 99% consumers use memory cache provided out of the box. And even if they did it, then rewriting it is quite simple. Secondly, this is an additional burden for us to support the old and new solution, jumping around Git, changesets, documentation, review, remember to remove old APIs in future, etc.

Shane32:
Ok.  Generally I would try to preserve API if possible, but here I agree, and it is notable that we cannot provide 100% compatibility anyway; it's only partial.

PR #3294, opened 2022-08-16
Restore `Field<TGraphType>()` method

### LINKED ISSUE #3293: FieldBuilder no longer works  with Field<Type>().Name("foo")
## Description

There is no longer an overload for Field which takes 0 arguments. We used this version of the FieldBuilder extensively unfortunately. Was this removal intentional?

```
Field<Type>().Name("foo").Resolve(...
``` 

It seems it was removed in f1f8348c0f8ae45b71f06688dfd4dde2338d5f05

### Expected result

Either a gradual deprecation or continued functionalty

### Actual result

error CS1501: No overload for method 'Field' takes 0 arguments 

### Environment

7.0.0

Shane32:
Fixes #3293


Shane32:
I think we made a mistake in removing a `Field()` method without marking it obsolete first.  Please review.

sungam3r (approved):

Issue #3301, opened 2022-08-24
Can't get directive to work


ygetarts:
Hello,

I'm extremely new to graphql and  working on a new codebase and trying to implement custom directives. At the moment, I'm trying to implement a simple field validation, where a field called "count" has a min value of 1 and a max of 10, for example. 

I'm trying to follow the guide [here](https://graphql-dotnet.github.io/docs/getting-started/directives/), but when I call FindAppliedDirective, it always returns null.

```
public class MaxLength : Directive
  {
    public MaxLength()
      : base("maxLength", DirectiveLocation.Mutation, DirectiveLocation.InputFieldDefinition)
    {
        Description = "Used to specify the minimum and/or maximum length for an input field or argument.";
        Arguments = new QueryArguments(
            new QueryArgument<IntGraphType>
            {
              Name = "min",
              Description = "If specified, specifies the minimum length that the input field or argument must have."
            },
            new QueryArgument<IntGraphType>
            {
              Name = "max",
              Description = "If specified, specifies the maximum length that the input field or argument must have."
            }
      );
    }
  }

  public class MaxLengthDirectiveVistor : BaseSchemaNodeVisitor
  {
    public override async void VisitObjectFieldDefinition(FieldType field, IObjectGraphType type, ISchema schema)
    {
      var applied = field.FindAppliedDirective("maxLength");
      if (applied != null)
      {....
```
this code is getting hit, it's just "applied" is always null. I've also tried passing in many different params into FindAppliedDirective, like 'count' and 'name' and other things, just to see if anything would show up, but it always returns null.

And I'm setting the direct like:

```
public class BookSummaryCreateArgInputType : InputObjectGraphType<BookSummaryCreateArg>
{
  public BookSummaryCreateArgInputType()
  {
    this.Name = "BookSummaryCreateArg";
    this.Field(_ => _.Count)
                          .Directive("maxLength", x =>
                            x.AddArgument(new DirectiveArgument("min") { Name = "min", Value = 1 })
                              .AddArgument(new DirectiveArgument("max") { Name = "max", Value = 10}));
  }
}
```
Not sure where I'm going wrong here?

Shane32:
Custom directives that validate input values are not for the faint of heart!  But my first thought is that perhaps you are overriding the wrong method within your custom node visitor, as I think `VisitObjectFieldDefinition` are only for output fields.  Perhaps overloading `VisitInputObjectFieldDefinition` is what you are looking for.

[...]

PR #3307, opened 2022-08-25
Fix InputFieldsAndArgumentsOfCorrectLength validation rule

Shane32:
See #3301

### DIFF
--- a/src/GraphQL.Tests/Validation/InputFieldsAndArgumentsOfCorrectLengthRequiredTests.cs
+++ b/src/GraphQL.Tests/Validation/InputFieldsAndArgumentsOfCorrectLengthRequiredTests.cs
@@ -1,0 +1,358 @@
+using GraphQL.Validation.Rules;
+
+namespace GraphQL.Tests.Validation;
+
+public class InputFieldsAndArgumentsOfCorrectLengthRequiredTests : ValidationTestBase<InputFieldsAndArgumentsOfCorrectLength, ValidationSchema>
+{
+    // SCALAR INPUT FIELD (LITERAL + VARIABLE TESTS)
+
+    [Fact]
+    public void good_literal_input_field_length()
+    {
+        ShouldPassRule(@"
+            {
+              complicatedArgs {
+                complexArgField2(complexArg: { requiredField: true, stringField: ""aaaa"" })
+              }
+            }");
+    }
+
+    [Fact]
+    public void good_variable_input_field_length()
+    {
+        ShouldPassRule(_ =>
+        {
+            _.Query = @"query q($value: String)
+                {
+                  complicatedArgs {
+                    complexArgField2(complexArg: { requiredField: true, stringField: $value })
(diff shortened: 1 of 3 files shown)


src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
```diff
@@ -59,7 +59,7 @@ public override ValueTask VisitFieldAsync(ValidationContext context, GraphQLVari
             new MatchingNodeVisitor<GraphQLArgument>((arg, context) => CheckLength(arg, arg.Value, context.TypeInfo.GetArgument(), context)),
             new MatchingNodeVisitor<GraphQLObjectField>((field, context) =>
             {
-                if (context.TypeInfo.GetInputType(1) is IInputObjectGraphType input)
+                if (context.TypeInfo.GetInputType(1)?.GetNamedType() is IInputObjectGraphType input)
```
sungam3r, line 62: 🤦‍♂️

src/GraphQL.Tests/Validation/InputFieldsAndArgumentsOfCorrectLengthRequiredTests.cs
```diff
@@ -0,0 +1,358 @@
+using GraphQL.Validation.Rules;
+
+namespace GraphQL.Tests.Validation;
+
+public class InputFieldsAndArgumentsOfCorrectLengthRequiredTests : ValidationTestBase<InputFieldsAndArgumentsOfCorrectLength, ValidationSchema>
```
sungam3r, line 5: You don't want to put tests in InputFieldsAndArgumentsOfCorrectLengthTests.cs?

sungam3r (approved):

src/GraphQL.Tests/Validation/InputFieldsAndArgumentsOfCorrectLengthRequiredTests.cs
Shane32: Either way is fine with me.  It just seemed easier at the time to copy the whole file.  Go ahead and make any changes to the PR you wish.

sungam3r: It seemed to me we had tests and I really found a file with tests. The next question that I asked myself - how they differ? Only `non-null` type modifier?

Shane32: Right; the only difference is that it is testing fields with non-null modifiers.  The tests are all the same, I think.

sungam3r: ok

PR #3314, opened 2022-09-05
Switch to Nullability.Source

sungam3r:
изображение

### DIFF
--- a/src/GraphQL/GraphQL.csproj
+++ b/src/GraphQL/GraphQL.csproj
@@ -11,7 +11,7 @@
   <ItemGroup>
     <PackageReference Include="GraphQL-Parser" Version="8.1.0" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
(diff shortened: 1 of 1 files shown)


sungam3r:
@Shane32 By the way I always have some issues with NullabilityInfo package when working locally in VS 2022
изображение
The same for this new package. `dotnet build` from console works fine.
изображение
ping @SimonCropp

Shane32:
@sungam3r I have no problems with the VS IDE in this respect on any of my computers.  Is there any chance that you have set `SingleTestPlatform` and this is causing this issue?  (Just guessing because I can't think of any other reason.)

Peeking at the definition for me properly shows either the .NET reference or the NullabilityInfo instance depending on the platform selection in the dropdown.

image

Shane32:
This PR seems to work fine also for me.

image

Shane32 (approved):

Shane32:
Note that I have been using VS 2022 awhile and am using 17.3.2 now

PR #3316, opened 2022-09-06
Prevent using graphtype as model

Shane32:
Implemented within `ComplexGraphType<TSourceType>`

Applies to:
- `ObjectGraphType<TSourceType>`
- `InputObjectGraphType<TSourceType>`
- `AutoRegisteringObjectGraphType<TSourceType>`
- `AutoRegisteringInputObjectGraphType<TSourceType>`

### DIFF
--- a/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
+++ b/src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
@@ -268,21 +268,26 @@ public class BubbleNullSchema : Schema
         var query = new ObjectGraphType();
 
         query.Field<NonNullGraphType<DataGraphType>>("nonNullableDataGraph")
-            .Resolve(c => new DataGraphType { Data = c.Source as Data }
+            .Resolve(c => new DataModel { Data = c.Source as Data }
         );
 
         query.Field<DataGraphType>("nullableDataGraph")
-            .Resolve(c => new DataGraphType { Data = c.Source as Data }
+            .Resolve(c => new DataModel { Data = c.Source as Data }
         );
 
         query.Field<NonNullGraphType<ListGraphType<NonNullGraphType<DataGraphType>>>>("nonNullableListOfNonNullableDataGraph")
-            .Resolve(_ => new[] { new DataGraphType() });
+            .Resolve(_ => new[] { new DataModel() });
 
         Query = query;
     }
 }
 
-public class DataGraphType : ObjectGraphType<DataGraphType>
+public class DataModel
+{
+    public Data Data { get; set; }
+}
+
+public class DataGraphType : ObjectGraphType<DataModel>
 {
     public DataGraphType()
     {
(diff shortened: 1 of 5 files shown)


Shane32:
This PR doesn't work because `__Type` derives from `ObjectGraphType<IGraphType>`.  Options:

1. Change `__Type` to derive from `ObjectGraphType` rather than `ObjectGraphType<IGraphType>`.  This would technically be a breaking change
2. Move the checks into the auto-registering graph type constructors

Shane32:
Solution 2 implemented here

src/GraphQL.Tests/Bugs/BubbleUpTheNullToNextNullable.cs
```diff
@@ -268,21 +268,26 @@ public BubbleNullSchema()
...
 
         query.Field<NonNullGraphType<ListGraphType<NonNullGraphType<DataGraphType>>>>("nonNullableListOfNonNullableDataGraph")
-            .Resolve(_ => new[] { new DataGraphType() });
+            .Resolve(_ => new[] { new DataModel() });
 
         Query = query;
     }
 }
 
-public class DataGraphType : ObjectGraphType<DataGraphType>
```
sungam3r, line 285: Ha)

sungam3r: That means that in theory in very specific cases someone **may** want to inherit from `ObjectGraphType<OtherGraphType>`.

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -17,6 +17,9 @@ public abstract class ComplexGraphType<TSourceType> : GraphType, IComplexGraphTy
         /// <inheritdoc/>
         protected ComplexGraphType()
         {
+            if (typeof(IGraphType).IsAssignableFrom(typeof(TSourceType)) && GetType() != typeof(Introspection.__Type))
```
sungam3r, line 20: 🤔 Maybe in future we introduce some flag (GlobalSwitch?) to allow this behavior.

sungam3r: I understand that for 99,99% cases this check helps to solve developer problems.

src/GraphQL.Tests/Types/NonNullGraphTypeTests.cs
```diff
@@ -94,34 +94,35 @@ public NullableSchema()
...
+            .Resolve(c => new DataModel { Data = c.Source as ExampleContext });
         query.Field<NonNullableSchemaType>("nonNullable")
-            .Resolve(c => new NonNullableSchemaType { Data = c.Source as ExampleContext });
+            .Resolve(c => new DataModel { Data = c.Source as ExampleContext });
 
         Query = query;
     }
 }
 
-public class NullableSchemaType : ObjectGraphType<NullableSchemaType>
```
sungam3r, line 105: Ah, even one more place.

Shane32:
Maybe a global switch. I figured that (a) if there’s a problem then someone will raise an issue (b) people can use the non-generic version of ObjectGraphType if necessary in these rare circumstances. 

I suggest we wait until someone complains before adding a global switch.

src/GraphQL.Tests/Types/NonNullGraphTypeTests.cs
Shane32: These are tests, after all - would they really be coded like this for production use? I highly doubt it — running all the graph constructor code for a data model? Seems highly unlikely. 

Could be though - I’m not sure.

sungam3r: Tests, yes. We ourselves write a code that should not be written.

sungam3r:
Of course.

PR #3317, opened 2022-09-06
Prevent using graphtype as model - alt

Shane32:
Applies to auto-registering types only
- `AutoRegisteringObjectGraphType`
- `AutoRegisteringInputObjectGraphType`
- `AutoRegisteringInterfaceGraphType`


Shane32:
Closing in favor of #3316

sungam3r:
Agree that #3316 seems better.

PR #3323, opened 2022-09-12
Remove init-only properties from pre-.NET 5 targets

### LINKED ISSUE #3322: Cannot use ValidateAsync in .NET 4.7.1
Hi,
I am trying to use GraphQL 7.0.2 in a .NET 4.7.1 application (netstandard2.0).

However, when I try to use ValidateAsync method I am getting this error:

image

C# language version for .NET 4.7.1 (netstandard2.0)  is 7.3 and GraphQL which is mutli-target solution, is using some features of newer C#.

I am wondering if you have any solution to this problem. 

Also, out of my curiosity,  how come GraphQL got compiled in first place with C# 9.0 features to target netstandard2.0 without compile tags?

Thanks.

Shane32:
Fixes #3322

### DIFF
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -3022,6 +3022,7 @@ namespace GraphQL.Validation
     public readonly struct ValidationOptions
     {
         public ValidationOptions() { }
+        public ValidationOptions(GraphQL.Types.ISchema schema, GraphQLParser.AST.GraphQLDocument document, System.Collections.Generic.IEnumerable<GraphQL.Validation.IValidationRule>? rules, System.Collections.Generic.IDictionary<string, object?> userContext, GraphQL.Instrumentation.Metrics metrics, GraphQL.Inputs variables, GraphQL.Inputs extensions, GraphQLParser.AST.GraphQLOperationDefinition operation, System.IServiceProvider? requestServices, System.Security.Claims.ClaimsPrincipal? user, System.Threading.CancellationToken cancellationToken) { }
         public System.Threading.CancellationToken CancellationToken { get; set; }
         public GraphQLParser.AST.GraphQLDocument Document { get; set; }
         public GraphQL.Inputs Extensions { get; set; }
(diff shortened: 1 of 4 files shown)


src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -3022,6 +3022,7 @@ namespace GraphQL.Validation
     public readonly struct ValidationOptions
     {
         public ValidationOptions() { }
+        public ValidationOptions(GraphQL.Types.ISchema schema, GraphQLParser.AST.GraphQLDocument document, System.Collections.Generic.IEnumerable<GraphQL.Validation.IValidationRule>? rules, System.Collections.Generic.IDictionary<string, object?> userContext, GraphQL.Instrumentation.Metrics metrics, GraphQL.Inputs variables, GraphQL.Inputs extensions, GraphQLParser.AST.GraphQLOperationDefinition operation, System.IServiceProvider? requestServices, System.Security.Claims.ClaimsPrincipal? user, System.Threading.CancellationToken cancellationToken) { }
```
Shane32, line 3025: This is a read-only struct.  To allow C# 8 and prior languages to write to these properties, a constructor was added as a non-breaking change.  Changing the struct to a read/write struct is a breaking change and affects other code.

src/GraphQL/Attributes/Authorization/AuthorizeAttribute.cs
```diff
@@ -29,13 +29,29 @@ public AuthorizeAttribute(string policy)
...
     /// </summary>
-    public string? Policy { get; init; }
+    public string? Policy
+    {
+        get;
+#if NET5_0_OR_GREATER
+        init;
+#else
+        set;
+#endif
```
Shane32, line 39: Minimize impact here.  Perhaps just converting to `set;` would be better.

See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/init#breaking-changes

Shane32: Note that this class is not likely to be inherited in 3rd-party libraries, so non-source-breaking is probably acceptable here.

Shane32: Changed it to simply `set;` since we wouldn't want a class library compiled against .NET Standard 2.0 to assume that the property was an `init;` while the main project is compiled against .NET 6 and assumes the property was a `set;` therefor having a broken build without any 3rd party references.

PR #3324, opened 2022-09-13
Add missing authorization extensions from field and connection builders

Shane32:
See https://github.com/graphql-dotnet/server/discussions/925


Shane32:
In the next version of GraphQL.NET, perhaps the field and connection builders should implement `IMetadataProvider` so these methods can be removed.

sungam3r:
> In the next version of GraphQL.NET, perhaps the field and connection builders should implement IMetadataProvider so these methods can be removed.

Indeed. And I have a vague feeling that I had already tried to do so.

PR #3329, opened 2022-09-16
Support .NET 7.0

zcsizmadia:
.NET 7.0 just went RC (rc.1) and expected to be officially released in mid November 2022. This PR should be merged once .NET 7.0 is released on November, however in the meantime the build process can be validated.

Note:
Before merging `include-prerelease: true` lines should be removed from the github workflow `actions/setup-dotnet@v2` actions.


src/GraphQL.ApiTests/net6+net7/GraphQL.approved.txt
```diff
@@ -1177,7 +1177,6 @@ namespace GraphQL.Execution
     public class ExecutionStrategyRegistration : System.IEquatable<GraphQL.Execution.ExecutionStrategyRegistration>
     {
         public ExecutionStrategyRegistration(GraphQL.Execution.IExecutionStrategy Strategy, GraphQLParser.AST.OperationType Operation) { }
-        protected virtual System.Type EqualityContract { get; }
```
zcsizmadia, line 1180: Need advice with this change

src/GraphQL.ApiTests/netstandard20+netstandard21/GraphQL.approved.txt
```diff
@@ -1177,7 +1177,6 @@ namespace GraphQL.Execution
     public class ExecutionStrategyRegistration : System.IEquatable<GraphQL.Execution.ExecutionStrategyRegistration>
     {
         public ExecutionStrategyRegistration(GraphQL.Execution.IExecutionStrategy Strategy, GraphQLParser.AST.OperationType Operation) { }
-        protected virtual System.Type EqualityContract { get; }
```
zcsizmadia, line 1180: Need advice with this change

zcsizmadia:
.NET Core 3.1 will go EOL mid December 2022 and .NET 5.0 is already EOL (https://dotnet.microsoft.com/en-us/download/dotnet)

Those targets could be removed as well when NET 7 is released. Supporting those EOL frameworks will be ok using the netstandard2.x targets.

src/GraphQL.Tests/GraphQLValuesCacheTests.cs
```diff
@@ -13,7 +13,13 @@ public void GetInt()
         GraphQLValuesCache.GetInt("-1000000").ShouldBe(-1000000);
         GraphQLValuesCache.GetInt("1000000").ShouldBe(1000000);
 
-        Should.Throw<FormatException>(() => GraphQLValuesCache.GetInt("a")).Message.ShouldBe("Input string was not in a correct format.");
```
zcsizmadia, line 16: It is unfortunate that MS has changed the FormatException message from .NET 6 -> .NET 7. Maybe the strict exception message check could be removed and just check for throwing `FormatException`. In that case the nasty ifdef could be eliminated.

src/GraphQL.Tests/GraphQLValuesCacheTests.cs
```diff
@@ -25,6 +31,12 @@ public void GetLong()
         GraphQLValuesCache.GetLong("-1000000").ShouldBe(-1000000L);
         GraphQLValuesCache.GetLong("1000000").ShouldBe(1000000L);
 
-        Should.Throw<FormatException>(() => GraphQLValuesCache.GetLong("a")).Message.ShouldBe("Input string was not in a correct format.");
+#if NET7_0_OR_GREATER
```
zcsizmadia, line 34: It is unfortunate that MS has changed the FormatException message from .NET 6 -> .NET 7. Maybe the strict exception message check could be removed and just check for throwing `FormatException`. In that case the nasty `ifdef` could be eliminated.

src/GraphQL.Tests/Instrumentation/ApolloTracingTests.cs
```diff
@@ -68,23 +68,28 @@ public void extension_has_expected_format()
     public void serialization_should_have_correct_case(IGraphQLTextSerializer writer)
     {
         var trace = new ApolloTrace(new DateTime(2019, 12, 05, 15, 38, 00, DateTimeKind.Utc), 102.5);
-        var expected = @"{
+#if NET7_0_OR_GREATER
```
zcsizmadia, line 71: To eliminate the framework specific expected string, can the `ms` fraction be changed to `102`?  In that case the formatting would match all .NET frameworks.

Shane32: Let's leave it as you have it now.  I would prefer to investigate further and determine why the change occurred, and then determine if we should change the serializer implementation to format the value in a defined manner regardless of the target framework.

src/GraphQL/GraphQL.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;net6</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;net7;net6</TargetFrameworks>
```
Shane32, line 4: Please remove .NET 7 TFM until we have a ifdef that requires it.

src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;net5;net6;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;net7;net6;net5;netcoreapp3.1</TargetFrameworks>
```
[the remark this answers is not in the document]
zcsizmadia: Adding net7 there will use the STJ 7.0+, instead of (4.7,2) which is used in case of netstandard2.x target. MS claims performance gain in STJ 7 compared to earlier version.

src/GraphQL/GraphQL.csproj
zcsizmadia: Removed.

Shane32:
> .NET Core 3.1 will go EOL mid December 2022 and .NET 5.0 is already EOL (https://dotnet.microsoft.com/en-us/download/dotnet)
> 
> Those targets could be removed as well when NET 7 is released. Supporting those EOL frameworks will be ok using the netstandard2.x targets.

GraphQL.NET only targets .NET Standard 2.0, .NET Standard 2.1 and .NET 6, as there are no ifdefs for other target frameworks.  System.Text.Json includes a target for .NET Core 3.1 because System.Text.Json is included with .NET Core 3.0+.  As such, the dependency can be eliminated.

At present I see no reason to change the targeted frameworks.  I feel that .NET Core 3.1 is especially important because not only it is a LTS release, a number of the packages developed at that time are cross-platform compatible with .NET Framework -- for example, Entity Framework Core.  While perhaps these applications can all run on .NET 6, when using the 3.1.x components of that time, it may be more likely that applications written for .NET Core 3.1 designed to be cross compatible with .NET Framework may not yet be upgraded to .NET 6.  There is a similar situation that exists with .NET Core 2.1 for web applications -- hence why GraphQL.NET Server 7.x now supports .NET Framework and .NET Core 2.1 web applications.  (Not that in that situation, there is official support from Microsoft for ASP.NET Core 2.1.x when run on supported editions of .NET Framework.)

I would have little argument against removing .NET 5 code -- it was never a LTS release and is more of a preview of .NET 6, just like .NET Core 3.0 was compared to .NET Core 3.1.  But yet, there is no reason to at this time.

If and when it becomes troublesome to perform development/testing for older frameworks (e.g. trying to develop/test .NET Framework 4.5 applications today), we will likely drop those frameworks from the test suite and/or libraries at that time.

Thanks for the PR!  As you suggest, we likely will not merge the PR until .NET 7.0 is released.

src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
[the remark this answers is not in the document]
Shane32: System.Text.Json is not a dependency for .NET Core 3.1+ as it is released as part of the framework.  .NET 7 will use the .NET 6 TFM as it is the closest match.  As such, it should have all of the performance benefits available with .NET 7.

src/GraphQL.Tests/Instrumentation/ApolloTracingTests.cs
zcsizmadia: It seems to be the issue is in the DateTime formatting library:

```
DateTime dt = new DateTime(2019, 12, 05, 15, 38, 00, DateTimeKind.Utc);
dt = dt.AddMilliseconds(102.5);
Console.WriteLine($"{dt.ToString("O")}");
```

net 7.0 -> `2019-12-05T15:38:00.1025000Z`
net5.0, net6.0 -> `2019-12-05T15:38:00.1030000Z`

Shane32: Or perhaps `AddMilliseconds` does not retain the fractional portion of the milliseconds specified... ?

- https://github.com/dotnet/runtime/pull/73198

Looks like this line here:

```cs
return AddTicks((long)millis_double * TicksPerMillisecond);
```

should have been this:

```cs
return AddTicks((long)(millis_double * TicksPerMillisecond));
```

but they cleaned it all up for .NET 7.

Shane32: So we could ignore it, or we could change the call to `AddMilliseconds` to use `AddTicks` instead, which would fix the bug in .NET 6 and prior versions.

src/GraphQL.SystemTextJson/GraphQL.SystemTextJson.csproj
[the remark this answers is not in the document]
zcsizmadia: Roger. I always assumed. If I build a net 6.0 app  always the STJ latest 6.0.x will be package referenced and not the absolute latest (in this case 7.0.x). However I never tested that, so  I will remove net7 as suggested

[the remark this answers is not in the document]
zcsizmadia: Removed net7

src/GraphQL.Tests/Instrumentation/ApolloTracingTests.cs
Shane32: :+1:

src/GraphQL.ApiTests/net6+net7/GraphQL.approved.txt
sungam3r: @Shane32  I think there are no problems if we all move to > 6.0.402 SDK . @zcsizmadia Please rebase on `master`. I removed these lines in #3380 .

sungam3r:
> Those targets could be removed as well when NET 7 is released. Supporting those EOL frameworks will be ok using the netstandard2.x targets.

Well, I'm not so positive to quickly remove support for "old" TFMs especially if this support is given at low price.

src/GraphQL.Tests/GraphQLValuesCacheTests.cs
sungam3r: > It is unfortunate that MS has changed the FormatException message from .NET 6 -> .NET 7. 

Maybe yes, maybe no :) . Yes, I prefer to test exception messages, not only exception types. When I read the test code, I better understand the nature of the code tested. I would not call it a tough requirement, just a recommendation that I would line to adhere to.

> So we just update the tests - as you have already done - when necessary.

Yes. Update if necessary. Sometimes I use `StartsWith`/`Contains` instead of exact match.

src/GraphQL/Instrumentation/ApolloTrace.cs
```diff
@@ -13,7 +13,9 @@ public class ApolloTrace
         public ApolloTrace(DateTime start, double durationMs)
         {
             StartTime = start.ToUniversalTime();
-            EndTime = StartTime.AddMilliseconds(durationMs);
+            // https://github.com/dotnet/runtime/pull/73198 fixes AddMilliseconds, but it is only available NET7+
+            // Use AddTicks instead to fix it for all frameworks
+            EndTime = StartTime.AddTicks((long)(durationMs * TimeSpan.TicksPerMillisecond));
```
sungam3r, line 18: 👍

sungam3r:
> I would have little argument against removing .NET 5 code -- it was never a LTS release and is more of a preview of .NET 6, just like .NET Core 3.0 was compared to .NET Core 3.1. But yet, there is no reason to at this time.

> If and when it becomes troublesome to perform development/testing for older frameworks (e.g. trying to develop/test .NET Framework 4.5 applications today), we will likely drop those frameworks from the test suite and/or libraries at that time.

> Thanks for the PR! As you suggest, we likely will not merge the PR until .NET 7.0 is released.

Agree. Let's merge ASAP after NET7 release.

.github/workflows/build-artifacts-code.yml
```diff
@@ -39,10 +39,11 @@ jobs:
           name: Documentation
           path: docs2/public/**
           if-no-files-found: error
-      - name: Setup .NET Core 6.0 SDK
+      - name: Setup .NET 7.0 SDK
```
sungam3r, line 42:
```suggestion
      - name: Setup .NET SDK
```

.github/workflows/codeql-analysis.yml
```diff
@@ -16,10 +16,11 @@ jobs:
     - name: Checkout source
       uses: actions/checkout@v3
 
-    - name: Setup .NET Core SDK
+    - name: Setup .NET 7.0 SDK
```
sungam3r, line 19:
```suggestion
    - name: Setup .NET SDK
```

.github/workflows/publish-code.yml
```diff
@@ -46,10 +46,11 @@ jobs:
           name: Documentation
           path: docs2/public/**
           if-no-files-found: error
-      - name: Setup .NET Core 6.0 SDK
+      - name: Setup .NET 7.0 SDK
```
sungam3r, line 49:
```suggestion
      - name: Setup .NET SDK
```

sungam3r (changes requested):
I'm fine to merge this one after resolving conflicts and .NET7 releaase.

zcsizmadia:
@sungam3r Merged master into the PR:

1. Using EOL SDKs the build will generat several EOL warnings. As of now it is only 5.0, however 3.1 goes EOL soon, so that will do th same. Added `<CheckEolTargetFramework>false</CheckEolTargetFramework>` to Directory.build.props to disable disable that warning. (EOL SDKs are used only for tests and benchmarks)
2. `dotnet-quality: preview` should be removed from the github workflow yml files before merging this PR

sungam3r:
`CheckEolTargetFramework` is fine, we use it in the server project as well.

sungam3r (approved):

src/Directory.Build.props
```diff
@@ -23,6 +23,16 @@
     <_FriendAssembliesPublicKey>PublicKey=0024000004800000940000000602000000240000525341310004000001000100352162dbf27be78fc45136884b8f324aa9f1dfc928c96c24704bf1df1a8779b2f26c760ed8321eca5b95ea6bd9bb60cd025b300f73bd1f4ae1ee6e281f85c527fa013ab5cb2c3fc7a1cbef7f9bf0c9014152e6a21f6e0ac6a371f8b45c6d7139c9119df9eeecf1cf59063545bb7c07437b1bc12be2c57d108d72d6c27176fbb8</_FriendAssembliesPublicKey>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!--
+      GraphQL library is not compiled for EOL target frameworks, however tests and benchmarks are running against as many target frameworks (even EOL) as possible
+      Hide EOL warnings (e.g. net5):
+      "warning NETSDK1138: The target framework 'net5' is out of support and will not receive security updates in the future.
+      Please refer to https://aka.ms/dotnet-core-support for more information about the support policy."
+    -->
```
Shane32, line 32:
```suggestion
```

I would remove this comment; it is false.  For example, GraphQL.SystemTextJson is also compiled for .NET 5, and GraphQL.Server is also compiled for .NET Core 2.1.

Besides, the comment should be unnecessary -- the property name is self descriptive and can be easily looked up if not.

src/GraphQL/Instrumentation/ApolloTrace.cs
Shane32: This should be patched in a separate PR.  While perhaps not likely to matter, it can/will cause different behavior in the GraphQL runtime.

Shane32: - #3384

zcsizmadia:
@Shane32 
1. Removed the comments for CheckEolTargetFramework
2. Removed the ApolloTrace AddTicks fix. The tests in this PR will fail until #3384 is not merged.

src/GraphQL/Instrumentation/ApolloTrace.cs
zcsizmadia: Done

src/Directory.Build.props
zcsizmadia: Done

Shane32 (approved):

zcsizmadia:
NEt 7 was released today. Merged master into the PR branch and removed the dotnet-quality=preview lines from the github workflows.

Shane32:
Visual Studio 17.4 hasn't been released yet.  Probably it will be within a day or so and then we can merge this.

sungam3r:
I have run VS installer just now - 2.4 Gb update for 17.4 :)

sungam3r:
Successful build on my machine except broken localization in console
изображение

PR #3332, opened 2022-09-17
Add type integrity check when existing types are found

### LINKED ISSUE #3331: Casing causes FieldsOnCorrectTypeError on mapped fields
## Description

A type is registered like this.

```
public class MyObjectGraphType : ObjectGraphType<MyObject>
{
    public MyObjectGraphType()
    {
        Field(p => p.Id);
        Field(p => p.Name);
    }
}
```

There is also a generic wrapper that allows for skip, take and total to be added to any query (left out here for simplicity, added at the end to the unit tests).

The (dynamically generated) schema is using the camel case naming convention. 

When checking the schema (e.g. inspecting it via code, or via GraphiQL), it shows the correct lowercase members (applying camel case):

``` 
query {
  myObjects {
    data {
      id
      name
    }
    skip
    take
    total
  }
}

``` 

When executing this query, it shows this error:

``` 
{
  "errors": [
    {
      "message": "Cannot query field 'id' on type 'MyObject'. Did you mean 'Id'?",
      "locations": [
        {
          "line": 4,
          "column": 7
        }
      ],
      "extensions": {
        "code": "FIELDS_ON_CORRECT_TYPE",
        "codes": [
          "FIELDS_ON_CORRECT_TYPE"
        ],
        "number": "5.3.1"
      }
    },
    {
      "message": "Cannot query field 'name' on type 'MyObject'. Did you mean 'Name'?",
      "locations": [
        {
          "line": 5,
          "column": 7
        }
      ],
      "extensions": {
        "code": "FIELDS_ON_CORRECT_TYPE",
        "codes": [
          "FIELDS_ON_CORRECT_TYPE"
        ],
        "number": "5.3.1"
      }
    }
  ]
}
``` 

When checking the *introspection*, it returns this (seems the names are correct):

``` 
{
	"kind": "OBJECT",
	"name": "MyObject",
	"description": null,
	"fields": [
		{
			"name": "id",
			"description": null,
			"args": [],
			"type": {
				"kind": "NON_NULL",
				"name": null,
				"ofType": {
					"kind": "SCALAR",
					"name": "ID",
					"ofType": null
				}
			},
			"isDeprecated": false,
			"deprecationReason": null
		},
		{
			"name": "name",
			"description": null,
			"args": [],
			"type": {
				"kind": "NON_NULL",
				"name": null,
				"ofType": {
					"kind": "SCALAR",
					"name": "String",
					"ofType": null
				}
			},
			"isDeprecated": false,
			"deprecationReason": null
		}
	],
	"inputFields": null,
	"interfaces": [],
	"enumValues": null,
	"possibleTypes": null
}
```

**What could cause the schema to return it in camelCase, but the server validation requiring it in PascalCase?**

### Steps to reproduce

Hard to reproduce yet. I did write unit tests (see end of this post). This test case is a simplified version of a complex graphql schema, but somehow I cannot find why it's not applying the naming convention correctly here.

### Expected result

I would expect the query to return the correct results.

### Environment

GraphQL 7.0.2
GraphQL.Server 7.1.1

### Unit tests

```

public class MyBugTest : QueryTestBase<Schema>
{
    private readonly Schema _schema;

    public MyBugTest()
    {
        var schema = new Schema
        {
            NameConverter = new CamelCaseNameConverter()
        };

        var queryGraphType = new ObjectGraphType
        {
            Name = "Query"
        };

        RegisterGenericQueryGraphTypes(schema, queryGraphType, "MyObject", "MyObjects");

        schema.Query = queryGraphType;

        _schema = schema;
    }

    [Fact]
    public void Camel_Case_Convention_Should_Work()
    {
        var query = @"
query {
  myObjects {
    data {
      id
      name
    }
    skip
    take
    total
  }
}";

        var expected = @"{
    ""myObjects"": {
      ""data"": [
        {
          ""id"": ""96b8c7b5-d542-4d70-a0cf-b0bbc0db119f"",
          ""name"": ""test1""
        },
        {
          ""id"": ""b405cb58-b966-4926-a989-90a76217af66"",
          ""name"": ""test2""
        }
      ],
      ""skip"": 0,
      ""take"": 10,
      ""total"": 2
    }
  }";
        AssertQuerySuccess(query, expected, null);
    }

    public override Schema Schema => _schema;

    private static void RegisterGenericQueryGraphTypes(Schema schema, ObjectGraphType queryGraphType,
        string singular, string plural)
    {
        var objectGraphType = new MyObjectGraphType();
        schema.RegisterType(objectGraphType);

        queryGraphType.Field<MyObjectGraphType>(singular)
            .Arguments(new QueryArguments(
                new QueryArgument(typeof(Guid?).GetGraphTypeFromType(true)) { Name = "Id" }))
            .Resolve(context => new MyObject
            {
                Id = Guid.NewGuid(),
                Name = "test"
            });

        var objectDataResultGraphType = CreateDataResultObjectGraphType<MyObject>(singular, objectGraphType, null);
        schema.RegisterType(objectDataResultGraphType);

        queryGraphType.Field(plural, objectDataResultGraphType)
            .Arguments(new QueryArguments(
                new QueryArgument(typeof(int?).GetGraphTypeFromType(true)) { Name = "Skip" },
                new QueryArgument(typeof(int?).GetGraphTypeFromType(true)) { Name = "Take" }))
            .Resolve(context =>
            {
                var dataResult = new DataResult<MyObject>
                {
                    Skip = 0,
                    Take = 10,
                    Total = 2
                };

                dataResult.Data.AddRange(new[]
                {
                    new MyObject
                    {
                        Id = Guid.Parse("96b8c7b5-d542-4d70-a0cf-b0bbc0db119f"),
                        Name = "test1"
                    },
                    new MyObject
                    {
                        Id = Guid.Parse("b405cb58-b966-4926-a989-90a76217af66"),
                        Name = "test2"
                    },
                });

                return dataResult;
            });
    }

    private static IObjectGraphType CreateDataResultObjectGraphType<TEntity>(string name, IObjectGraphType objectGraphType, FuncFieldResolver<DataResult<MyObject>, List<TEntity>> fieldResolver)
    {
        var graphType = new ObjectGraphType
        {
            Name = $"{name}DataResult"
        };

        // Total
        graphType.AddField(new FieldType
        {
            Name = nameof(DataResult<MyObject>.Total),
            ResolvedType = new LongGraphType(),
        });

        // Offset
        graphType.AddField(new FieldType
        {
            Name = nameof(DataResult<MyObject>.Skip),
            ResolvedType = new IntGraphType(),
        });

        // Take
        graphType.AddField(new FieldType
        {
            Name = nameof(DataResult<MyObject>.Take),
            ResolvedType = new IntGraphType(),
        });

        // Data
        graphType.AddField(new FieldType
        {
            Name = nameof(DataResult<MyObject>.Data),
            ResolvedType = new ListGraphType(objectGraphType),
            Resolver = fieldResolver
        });

        return graphType;
    }
}

public class MyObject
{
    public Guid Id { get; set; }
}

public class MyObjectGraphType : ObjectGraphType<MyObject>
{
    public MyObjectGraphType()
    {
        Field(p => p.Id);
        Field(p => p.Name);
    }
}

public class DataResult<TResult>
{
    public DataResult()
    {
        Data = new List<TResult>();
    }

    public long Total { get; set; }

    public int Skip { get; set; }

    public int Take { get; set; }

    public List<TResult> Data { get; private set; }
}

public class DataResultGraphType<TType, TGraphType> : ObjectGraphType<DataResult<TType>>
    where TGraphType : IGraphType
{
    public DataResultGraphType()
    {
        Field(x => x.Skip);
        Field(x => x.Take);
        Field(x => x.Total);
        Field(x => x.Data);
    }
}
```

GeertvanHorrik:
This PR adds type integrity checks when a type is (accidentally) registered twice.

This fixes #3331. For example, in my case it was not applying the name conversion (camelCase) to the 2nd object causing runtime issues.

Example of an exception being thrown. This should give the developer some pointers to look for registration issues.

```
Type 'MyObjectGraphType' has field 'id', but this cannot be found on the new type. New type has:
- Id
- Name

Make sure to use the same type instances when registering the same type
```

Question: this PR causes 1 unit tests for a bug (Issue1927ResolvedTypeWithType, test_outputs) to fail. How would you like to deal with this?

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug3331.cs
+++ b/src/GraphQL.Tests/Bugs/Bug3331.cs
@@ -1,7 +1,6 @@
 using GraphQL.Conversion;
 using GraphQL.Resolvers;
 using GraphQL.Types;
-using Shouldly;
 
 namespace GraphQL.Tests.Bugs;
 
@@ -30,7 +29,7 @@ public class Bug3331
             .Message.ShouldBe("A different instance of the type 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.");
 
         // Must have 2 instances
-        Assert.Equal(3, MyObjectGraphType.SharedInstanceCounter);
+        MyObjectGraphType.SharedInstanceCounter.ShouldBe(3);
     }
 
     private static void RegisterGenericQueryGraphTypes(Schema schema, ObjectGraphType queryGraphType,
--- a/src/GraphQL.Tests/Bugs/Bug3331.cs
+++ b/src/GraphQL.Tests/Bugs/Bug3331.cs
@@ -2,50 +2,10 @@ using GraphQL.Conversion;
 using GraphQL.Resolvers;
 using GraphQL.Types;
 
-namespace GraphQL.Tests.Types.Collections;
+namespace GraphQL.Tests.Bugs;
 
-public partial class SchemaTypesTests
+public class Bug3331
 {
-    [Fact]
-    public void throws_exception_when_multiple_type_instances_exists_simple()
-    {
-        var schema = new Schema
-        {
-            NameConverter = new CamelCaseNameConverter()
-        };
-
-        var queryGraphType = new ObjectGraphType
-        {
-            Name = "Query"
-        };
-
-        schema.RegisterType(queryGraphType);
-
-        // Object 1
-        var graphType1 = new ObjectGraphType
-        {
-            Name = "MyObject"
-        };
-
-        graphType1.Field<IntGraphType>("int");
-
-        queryGraphType.Field("first", graphType1);
-
-        // Object 2
-        var graphType2 = new ObjectGraphType
-        {
-            Name = "MyObject"
-        };
-
-        graphType2.Field<IntGraphType>("int");
-        graphType2.Field<StringGraphType>("string");
-
-        queryGraphType.Field("second", graphType2);
-
-        // Test
-        Assert.Throws<InvalidOperationException>(() => schema.Initialize());
-    }
-
     [Fact]
     public void throws_exception_when_multiple_type_instances_exists_complex()
     {
--- a/src/GraphQL.Tests/Bugs/Bug3331.cs
+++ b/src/GraphQL.Tests/Bugs/Bug3331.cs
@@ -1,6 +1,7 @@
 using GraphQL.Conversion;
 using GraphQL.Resolvers;
 using GraphQL.Types;
+using Shouldly;
 
 namespace GraphQL.Tests.Bugs;
 
@@ -25,7 +26,8 @@ public class Bug3331
 
         schema.Query = queryGraphType;
 
-        Assert.Throws<InvalidOperationException>(() => schema.Initialize());
+        Should.Throw<InvalidOperationException>(() => schema.Initialize())
+            .Message.ShouldBe("A different instance of the type 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.");
 
         // Must have 2 instances
         Assert.Equal(3, MyObjectGraphType.SharedInstanceCounter);
--- a/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
+++ b/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
@@ -7,7 +7,7 @@ using Moq;
 
 namespace GraphQL.Tests.Types.Collections;
 
-public class SchemaTypesTests
+public partial class SchemaTypesTests
 {
     [Fact]
     public void does_not_request_instance_more_than_once()
--- a/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
+++ b/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
@@ -83,7 +83,8 @@ public class SchemaTypesTests
         queryGraphType.Field("second", graphType2);
 
         // Test
-        Assert.Throws<InvalidOperationException>(() => schema.Initialize());
+        Should.Throw<InvalidOperationException>(() => schema.Initialize())
+            .Message.ShouldBe("A different instance of the type 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.");
     }
 
     [Fact]
--- a/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
+++ b/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
@@ -1,3 +1,4 @@
+using GraphQL.Conversion;
 using GraphQL.DI;
 using GraphQL.StarWars;
 using GraphQL.StarWars.Types;
@@ -7,7 +8,7 @@ using Moq;
 
 namespace GraphQL.Tests.Types.Collections;
 
-public partial class SchemaTypesTests
+public class SchemaTypesTests
 {
     [Fact]
     public void does_not_request_instance_more_than_once()
@@ -43,6 +44,46 @@ public partial class SchemaTypesTests
         mock.Verify(x => x.GetService(typeof(IEnumerable<IConfigureSchema>)), Times.Once);
         mock.Verify(x => x.GetService(typeof(IEnumerable<IGraphTypeMappingProvider>)), Times.Once);
         mock.VerifyNoOtherCalls();
+    }
+
+    [Fact]
+    public void throws_exception_when_multiple_type_instances_exists()
+    {
+        var schema = new Schema
+        {
+            NameConverter = new CamelCaseNameConverter()
+        };
+
+        var queryGraphType = new ObjectGraphType
+        {
+            Name = "Query"
+        };
+
+        schema.RegisterType(queryGraphType);
+
+        // Object 1
+        var graphType1 = new ObjectGraphType
+        {
+            Name = "MyObject"
+        };
+
+        graphType1.Field<IntGraphType>("int");
+
+        queryGraphType.Field("first", graphType1);
+
+        // Object 2
+        var graphType2 = new ObjectGraphType
+        {
+            Name = "MyObject"
+        };
+
+        graphType2.Field<IntGraphType>("int");
+        graphType2.Field<StringGraphType>("string");
+
+        queryGraphType.Field("second", graphType2);
+
+        // Test
+        Assert.Throws<InvalidOperationException>(() => schema.Initialize());
     }
 
     [Fact]
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -2,6 +2,7 @@ using System.Collections;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Numerics;
+using System.Text;
 using GraphQL.Conversion;
 using GraphQL.Instrumentation;
 using GraphQL.Introspection;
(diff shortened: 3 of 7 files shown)


Shane32:
I’ll review in a bit. Ideally we would retain a test for whatever 1927 addressed, while fixing the issue that is interfering with the new patch. Feel free to edit the old test and I’ll review the changes.

Shane32:
I’d like to have your test case moved to a Bug3331.cs file while adding a test within SchemaTypesTests that is very narrow and tests the exact scenario we are fixing. It should be possible within a dozen lines of code or so - just add two fields to a query type that each have their ResolvedType properties set to different instances of an object graph type with the same name. It should fail upon calling schema.Initialize

GeertvanHorrik:
All the methods on the SchemaTypes were private. I will keep the simple test on the SchemaTypes unit tests and move the complex one to the bug file. Does that sound good?

Shane32:
Sounds good.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -646,11 +647,62 @@ private void AddTypeIfNotRegistered(IGraphType type, TypeCollectionContext conte
...
+                return;
+            }
+
+            if (existingType.GetType() != newType.GetType())
+            {
+                throw new InvalidOperationException($"Type '{existingType.GetType().Name}' is already registered but being re-registered using different type '{newType.GetType().Name}'");
+            }
+
+            // Ignore scalars
+            if (existingType is not IObjectGraphType objectKnownType || newType is not IObjectGraphType objectNewType)
```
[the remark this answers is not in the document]
Shane32: This check is good; scalars do not matter if multiple instances are being used, and it is likely enough that it happens throughout the codebase.  However, the same issue applies to `InterfaceGraphType` and `InputObjectGraphType` instances as well as `IObjectGraphType` instances

I'd probably just combine this with the `ReferenceEquals` check above but either way is fine.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -646,11 +647,62 @@ private void AddTypeIfNotRegistered(IGraphType type, TypeCollectionContext conte
...
+        private void EnsureTypeEquality(IGraphType existingType, IGraphType newType)
+        {
+            if (ReferenceEquals(existingType, newType))
+            {
+                return;
+            }
+
+            if (existingType.GetType() != newType.GetType())
+            {
+                throw new InvalidOperationException($"Type '{existingType.GetType().Name}' is already registered but being re-registered using different type '{newType.GetType().Name}'");
```
Shane32, line 670: I suggest we expand this error message a bit to include the GraphQL type name that is registered a second time.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -646,11 +647,62 @@ private void AddTypeIfNotRegistered(IGraphType type, TypeCollectionContext conte
...
+                    {
+                        exceptionStringBuilder.AppendLine($"- {newField.Name}");
+                    }
+
+                    exceptionStringBuilder.AppendLine();
+                    exceptionStringBuilder.Append("Make sure to use the same type instances when registering the same type");
+
+                    throw new InvalidOperationException(exceptionStringBuilder.ToString());
+                }
+            }
```
Shane32, line 705: Let's replace all of this with a simple exception.  What if all fields match?  It should still generate an exception.

```suggestion
            throw new InvalidOperationException($"A different instance of the type '{newType.Name}' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.");
```

Shane32:
It seems that the `Issue1927ResolvedTypeWithType` tests are easily fixed by adding a name to the inner type.

```cs
        var innerObject = new ObjectGraphType() { Name = "inner" };
```

However, it reveals an issue with `RegisterType` where registering the same type multiple times causes an issue -- probably the same issue you ran into.  See failing test `SchemaTests.registers_additional_duplicated_types` and `Bug1046.any_registration_order_should_work`, plus a bunch in `OverlappingFieldsCanBeMergedTest`.

I think we need to fix the underlying issue here, specifically that calling `RegisterType<T>` on a type that has already been registered (directly or indirectly) does not attempt to reregister the type.  Probably changing some call from `AddType` to `AddTypeIfNotRegistered` or similar within `SchemaTypes`.

GeertvanHorrik:
Thank you very much for the code review.

I've addressed all the issues. I tried adding the parent type info (because now it's still hard to pinpoint where it goes wrong), but it seems this context is not available in the context nor the types.

When changing the method to a more simplified version, it does cause more tests to fail (which I think is a good thing, this can lead to very hard to find issues). Will fix them once you approved my changes so far.

GeertvanHorrik:
> I think we need to fix the underlying issue here, specifically that calling RegisterType<T> on a type that has already been registered (directly or indirectly) does not attempt to reregister the type. Probably changing some call from AddType to AddTypeIfNotRegistered or similar within SchemaTypes.

Made the change to the calls (except where I think it didn't make much sense). There are now 23 failing tests, showing potential issues.

I think the parents (of both old and new) should be passed into the AddTypeIfNotRegistered. This will give more context. Would that be a good change?

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -646,11 +646,37 @@ private void AddTypeIfNotRegistered(IGraphType type, TypeCollectionContext conte
...
+
+            // Ignore scalars
+            if (existingType is ScalarGraphType && newType is ScalarGraphType)
+            {
+                return;
+            }
+
+            if (existingType.GetType() != newType.GetType())
+            {
+                throw new InvalidOperationException($@"Unable to register GraphType '{newType.GetType().FullName}' with the name '{newType.Name}'. The name '{newType.Name}' is already registered to '{existingType.GetType().FullName}'. Check your schema configuration.");
```
Shane32, line 675: We have a `GetFriendlyName()` extension method for `Type`s rather than `FullName` which we typically use as a matter of convention.
```suggestion
                throw new InvalidOperationException($@"Unable to register GraphType '{newType.GetType().GetFriendlyName()}' with the name '{newType.Name}'. The name '{newType.Name}' is already registered to '{existingType.GetType().GetFriendlyName()}'. Check your schema configuration.");
```

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
```diff
@@ -45,6 +46,46 @@ public void does_not_request_instance_more_than_once()
...
+            Name = "MyObject"
+        };
+
+        graphType2.Field<IntGraphType>("int");
+        graphType2.Field<StringGraphType>("string");
+
+        queryGraphType.Field("second", graphType2);
+
+        // Test
+        Assert.Throws<InvalidOperationException>(() => schema.Initialize());
```
Shane32, line 86: We use the Shouldly assertion library for tests, and as a matter of convention, test the message name (more so it can be seen what the failure is rather than because we want to ensure it does not change).

```suggestion
        Should.Throw<InvalidOperationException>(() => schema.Initialize())
            .Message.ShouldBe("...");
```

src/GraphQL/Types/Collections/SchemaTypes.cs
GeertvanHorrik: I copied this from another location. Will change it there as well.

GeertvanHorrik:
Thanks for the review. Points should be addressed. Only my previous comment remains: how to deal with the other unit tests?

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -189,7 +189,7 @@ protected void Initialize(ISchema schema, IServiceProvider serviceProvider, IEnu
                 {
                     if (this[name] == null)
                     {
-                        AddType(graphType, context);
+                        AddTypeIfNotRegistered(graphType, context);
```
Shane32, line 192: Well, unfortunately this isn't helping.  I did a little investigation and found that `GetSchemaTypes` is going to create instances for each of the types listed in `AdditionalTypes`.  That could be solved easily enough with `.Distinct()` but it does not cover the situation where an indirect reference to a type previously added via `RegisterType` has a separate reference from the subsequent one added via `RegisterType`.  I don't have time to look at it further at the moment.

Shane32: So might as well revert this line at this time.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -603,7 +603,7 @@ private void AddTypeWithLoopCheck(IGraphType resolvedType, TypeCollectionContext
             context.InFlightRegisteredTypes.Push(namedType);
             try
             {
-                AddType(resolvedType, context);
+                AddTypeIfNotRegistered(resolvedType, context);
```
Shane32, line 606: Here either.  Might as well revert.

Shane32:
We could probably make line 201 to pass a boolean to `AddTypeIfNotRegistered` so that it does not throw an exception if a duplicate is encountered.  Theoretically this should be fine, but this feels like a hack and not a solution.  For example, then it would then not throw an exception if two different instances were manually registered.

Shane32:
(At least the schema would be consistent)

Shane32:
Really, the types in `AdditionalTypes` should be calling the other overload of `AddTypeIfNotRegistered` that accepts a `Type` and not a `IGraphType`.

Shane32:
I suggest:

```cs

        private static (IEnumerable<IGraphType>, IEnumerable<Type>) GetSchemaTypes(ISchema schema, IServiceProvider serviceProvider)
        {
            return (
                GetSchemaTypeInstances(),
                schema.AdditionalTypes.Select(x => x.GetNamedType()).Where(x => !typeof(ScalarGraphType).IsAssignableFrom(x)));

            IEnumerable<IGraphType> GetSchemaTypeInstances()
            {
                // Manually registered AdditionalTypeInstances and AdditionalTypes should be handled first.
                // This is necessary for the correct processing of overridden built-in scalars.

                foreach (var instance in schema.AdditionalTypeInstances)
                    yield return instance;

                foreach (var type in schema.AdditionalTypes)
                {
                    var type2 = type.GetNamedType();
                    if (typeof(ScalarGraphType).IsAssignableFrom(type2))
                    {
                        yield return (IGraphType)serviceProvider.GetRequiredService(type2);
                    }
                }

                //TODO: According to the specification, Query is a required type. But if you uncomment these lines, then the mass of tests begin to fail, because they do not set Query.
                // if (Query == null)
                //    throw new InvalidOperationException("Query root type must be provided. See https://spec.graphql.org/October2021/#sec-Schema-Introspection");

                if (schema.Query != null)
                    yield return schema.Query;

                if (schema.Mutation != null)
                    yield return schema.Mutation;

                if (schema.Subscription != null)
                    yield return schema.Subscription;
            }
        }

//line 147

            var (typeinstances, types) = GetSchemaTypes(schema, serviceProvider);

//line 169

            // Add manually-added scalar types. To allow overriding of built-in scalars, these must be added
            // prior to adding any other types (including introspection types).
            foreach (var type in typeinstances)
            {
                if (type is ScalarGraphType)
                    AddType(type, _context);
            }

//line 198

            foreach (var type in typeinstances)
            {
                if (type is not ScalarGraphType)
                    AddTypeIfNotRegistered(type, ctx);
            }

            foreach (var type in types)
            {
                AddTypeIfNotRegistered(type, ctx);
            }
```

I believe that will fix all of the other tests

src/GraphQL.Tests/Bugs/Bug3331.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        RegisterGenericQueryGraphTypes(schema, queryGraphType, "MyObject", "MyObjects");
+
+        schema.Query = queryGraphType;
+
+        Should.Throw<InvalidOperationException>(() => schema.Initialize())
+            .Message.ShouldBe("A different instance of the type 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.");
+
+        // Must have 2 instances
+        Assert.Equal(3, MyObjectGraphType.SharedInstanceCounter);
```
Shane32, line 33:
```suggestion
        MyObjectGraphType.SharedInstanceCounter.ShouldBe(3);
```

Shane32:
Have to change line 240-241 also to this:

```cs
                foreach (var type in schema.AdditionalTypes)
                {
                    var type2 = type.GetNamedType();
                    if (typeof(ScalarGraphType).IsAssignableFrom(type2))
                    {
                        yield return (IGraphType)serviceProvider.GetRequiredService(type2);
                    }
                }
```

And then all tests should pass.

GeertvanHorrik:
That's it, everything seems to run fine now. Thanks for all the review work.

If you want anything changed, please let me know and I'll try to fix asap.

Shane32:
I think this is great; thanks for your help!

Shane32 (approved):

PR #3347, opened 2022-10-04
Fix a bug that cannot print `ListGraphType<GuidGraphType>`'s default value

moreal:
This pull request tries to fix a bug that cannot print `LIstGraphType<GuidGraphType>`'s default value well.
The `SchemaPrinter` will become to use non-generic `IEnumerable` and cast the elements with `IEnumerable.Cast<T>()` method.

## Description

If there is a field to have `ListGraphType<GuidGraphType>`-typed argument and it has `new List<Guid>()` default value, the `SchemaPrinter` throws an `InvalidCastException` exception because `List<Guid>` cannot be converted to `IEnumerable<object>`.

```
Exception of type 'System.InvalidCastException' was thrown: Unable to cast object of type 'System.Collections.Generic.List`1[[System.Guid, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]' to type 'System.Collections.Generic.IEnumerable`1[[System.Object, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]'
```


Shane32:
Ahh; probably because an `IEnumerable<out T>` of a value type does not inherit from `IEnumerable<object>`; only reference types do.

Shane32 (approved):

sungam3r:
@moreal Thanks!

PR #3349, opened 2022-10-05
Introduce VersionGraphType

sungam3r:


sungam3r:
Good for 7.2 I think. 

@Shane32 I am going to return to the support of the project in the second half of October. So far I am busy and outside the city.

sungam3r:
> this does not comply with SemVer 1.0.0 or SemVer 2.0.0. Namely, it does not allow strings but only numbers.

Yes, I know. I just want to give a base concept of SemVer - major.minor.patch . MS implementation differs from SemVer spec. `System.Version` is rather common type to use so I chose it. For my needs it is enough thought I'm fine to add "real" SemVer scalar. Look here https://www.nuget.org/packages?q=semver . It looks like https://github.com/maxhauser/semver project works well with SemVer versions. The main disadvantage for me is that it brings external dependency which is not related to GraphQL.NET. `System.Version` is a well-known common type and I like to use it. So maybe just change comments and remove mention about SemVer ? Let it be a scalar that allows you to work more conveniently with versions without touching all aspects of the SemVer specification.

Shane32:
> So maybe just change comments and remove mention about SemVer

Ok

Shane32:
See also: https://www.nuget.org/packages/NuGet.Versioning

src/GraphQL/Types/Scalars/VersionGraphType.cs
```diff
@@ -0,0 +1,59 @@
+using GraphQLParser.AST;
+
+namespace GraphQL.Types;
+
+/// <summary>
+/// The Version scalar graph type represents a version number according to SemVer.
```
Shane32, line 6: Please remove remarks that mention SemVer

sungam3r:
I realized that "real" SemVer scalar is required for my needs, so maybe I will add both and place "real" one in separate package under new Scalars folder in solution.

sungam3r:
What about naming? `VersionGraphType` with `Version` name and `SemVerGraphType` with `SemVer` name?

Shane32:
> I realized that "real" SemVer scalar is required for my needs, so maybe I will add both and place "real" one in separate package under new Scalars folder in solution.

I think we should discuss further, in that case.

The `Version` class is described as:

> Represents the version number of an assembly, operating system, or the common language runtime. This class cannot be inherited.

As such, it was designed for a certain purpose and may have limited uses outside of that.  For instance, you cannot represent "4.0" with the Version class, but rather must specify "4.0.0.0".

SemVer has a much wider support of typical version number patterns, including "4.0" and "4.0.1-beta", etc.  But then, we should ask ourselves questions along these lines:
- Why cannot the user simply use a string to hold the version number?
- Is there a great benefit to a custom scalar class?
- Should it not resolve to a custom class representing the SemVer vs a string?
- What class should it resolve to?
- Should we use one of the multiple existing SemVer packages, such as the one in NuGet.Versioning?
- Should we use our own custom SemVer class?
- What if the user has their own SemVer class or uses another SemVer package?
- How complicated is a custom scalar to implement, if we do not provide an implementation?

We must keep in mind that a completely valid implementation is as simple as the sample below, and does not require adding another published package:

```cs
    public class SemVerGraphType : ScalarGraphType
    {
        public override object? ParseLiteral(GraphQLValue value) => value switch
        {
            GraphQLStringValue s => NuGet.Versioning.SemanticVersion.Parse((string)s.Value),
            GraphQLNullValue _ => null,
            _ => ThrowLiteralConversionError(value)
        };

        public override object? ParseValue(object? value) => value switch
        {
            string s => NuGet.Versioning.SemanticVersion.Parse(s),
            null => null,
            _ => ThrowValueConversionError(value)
        };

        public override object? Serialize(object? value) => value switch
        {
            NuGet.Versioning.SemanticVersion sv => sv.ToNormalizedString(),
            null => null,
            _ => ThrowSerializationError(value)
        };
    }
```

Another option might be to make a helper scalar class for string-based scalars:

```cs
    public abstract class StringBasedGraphType<T> : ScalarGraphType
    {
        public override object? ParseLiteral(GraphQLValue value) => value switch
        {
            GraphQLStringValue s => Parse((string)s.Value),
            GraphQLNullValue _ => null,
            _ => ThrowLiteralConversionError(value)
        };

        public override bool CanParseLiteral(GraphQLValue value) => value switch
        {
            GraphQLStringValue s => TryParse((string)s.Value, out var _),
            GraphQLNullValue _ => true,
            _ => false,
        };

        public override object? ParseValue(object? value) => value switch
        {
            string s => Parse(s),
            T t => value,
            null => null,
            _ => ThrowValueConversionError(value)
        };

        public override bool CanParseValue(object? value) => value switch
        {
            string s => TryParse(s, out var _),
            T => true,
            null => null,
            _ => false
        };

        public override object? Serialize(object? value) => value switch
        {
            T t => Serialize(t),
            null => null,
            _ => ThrowSerializationError(value)
        };

        public abstract T Parse(string value);

        public virtual bool TryParse(string value, out T ret)
        {
            try
            {
                ret = Parse(value);
                return true;
            }
            catch
            {
                return false;
            }
        }

        public abstract string Serialize(T value);
    }

    public class SemVerGraphType : StringBasedGraphType<NuGet.Versioning.SemanticVersion>
    {
        public override SemanticVersion Parse(string value)
            => NuGet.Versioning.SemanticVersion.Parse(value);

        public override bool TryParse(string value, out SemanticVersion ret)
            => NuGet.Versioning.SemanticVersion.TryParse(value, out ret);

        public override string Serialize(SemanticVersion value)
            => value.ToNormalizedString();
    }
```

This helper class could be used for some of the internal scalars (Guid and date scalars), and make creating custom scalars simpler.

Shane32:
Keep in mind also that the value converter can be used to convert the string to a SemVer inside `GetArgument`.

sungam3r:
> Why cannot the user simply use a string to hold the version number?

I have a special personal case - my autogenerated GraphQL clients need a custom scalar type in schema to bind appropriate .NET type. In one of our services we use SemVer versions widely and I have become tired to convert strings to `Version` on client each time. 

> Is there a great benefit to a custom scalar here?

For ones no, for other maybe. Again - I need custom scalar to differentiate pure string from string of some taste (semver inside).

Regarding NuGet.Versioning I'm fine using it in separate package. Why not? We have directory for two serializers, why not to have one for custom scalars that are not so simple as others (built-in in SchemaTypes)?

> Should we use our own custom SemVer class?

😕 I would not. Not an interest of GraphQL.NET. 

> What if the user has their own SemVer class or uses another SemVer package?

The same question can be asked for each scalar that we have in addition to 5 declared in spec. I would answer "then user will post an issue here and we will consider this case". 

> How complicated is a custom scalar to implement, if we do not provide an implementation?

What?

Regarding `StringBasedGraphType`. I would not go this way. It looks a bit heavy. I would add self -sufficient new scalar:
  - no dependencies -> then in GraphQL.NET code package
  - with dependencies -> then in separate package inside this repo 

> Keep in mind also that the value converter can be used to convert the string to a SemVer inside GetArgument.

To be clear here - my interest is not only in server behavior but in client too. Of course I'm glad that server that uses new SemVer scalar can automatically validate inputs and outputs. The very fact of the existence of the new scalar **name** is still very important to me.

Shane32:
If we are going to start creating officially supported scalars that have dependencies, such as the NuGet.Versioning implementation of SemVer, it seems there is likely to be a separate nuget package for each unique scalar.  As such, we should define a very clear format for each scalar, such as this:

- `GraphQL.Scalars.(NuGet package here).(Class here, optional)`

In this instance it would be:

`GraphQL.Scalars.NuGet.Versioning` or `GraphQL.Scalars.NuGet.Versioning.SemanticVersion`

Another option might be to put the class name first, such as below, but then what if there are multiple classes in one dependency?

`GraphQL.Scalars.SemanticVersion.NuGet.Versioning`

I'm also not sure how the client portion fits into this.  The client code doesn't use any of the server code to operate.  We would need two separate NuGet packages, one for the client and one for the server.  We should have a naming strategy for those NuGet packages as well.

For all of the above scenarios, I would create another repository to contain these extra scalars.  I would not add it to the main repository, and personally I would not add them at all.  It is very easy to create a custom scalar.

> The same question can be asked for each scalar that we have in addition to 5 declared in spec. I would answer "then user will post an issue here and we will consider this case".

I don't think this is a fair comparison.  Most of the "additional" types are .NET intrinsic types, such as `short` or `decimal`.  They are not custom classes; they are root types of the .NET language.  This is true of our date/time scalars as well, although our implementation is our own definition.  The only other types we have is `Guid` and `Uri` scalars, which are extremely well known and often-used .NET Standard types.

Adding a custom scalar for the .NET Standard `Version` class doesn't fit with any of the above scalars.  Rather it's just one of a dozen (or dozens) of classes within the .NET library that might be of interest to a developer, and a rather uncommonly used one at that.  `IPAddress`, `IPEndpoint` and `DnsEndPoint` all fit in this category as well, and are probably used more often than `Version` is.  We _could_ also add scalars for base-64 encodings of data strings (either as byte arrays or MemoryStreams), and many others, all from the .NET Standard library.

But in this case, the scalar you're suggesting isn't any of these but rather for one out of thousands of classes in third-party packages.  We can't compare `NuGet.Versioning.SemanticVersion` to `short` or `DateTime` or `Guid`.

So in conclusion, I'm not entirely against adding it, but I think it should be done thoughtfully and logically, and in a separate repository.

Shane32:
> > How complicated is a custom scalar to implement, if we do not provide an implementation?
> 
> What?

I was trying to point out that the value of an "official" scalar for NuGet.Versioning.SemanticVersion has very little practical value.  It is going to contain about 2 dozen lines of code which could just as easily be posted in a FAQ.  It does not seem to me to be worth the trouble to maintain.  And if 2 dozen lines of code is just too complicated for end users, by adding `StringBasedGraphType` to the GraphQL.NET package, it could be only 1 dozen lines of code.  Do we really need to publish another NuGet package just for the sake of a dozen lines of code?  No.

sungam3r:
> GraphQL.Scalars.NuGet.Versioning or GraphQL.Scalars.NuGet.Versioning.SemanticVersion

OK. Honestly, I think it is not so nesessary to include **full** package name in scalar package name.

> We would need two separate NuGet packages, one for the client and one for the server

No need in client packages. Only server one. How some client handles custom scalar is a very different story.

> We could also add scalars for base-64 encodings of data strings (either as byte arrays or MemoryStreams), and many others, all from the .NET Standard library.

Yes.

>  We can't compare NuGet.Versioning.SemanticVersion to short or DateTime or Guid.

I was not going to compare them.

> So in conclusion, I'm not entirely against adding it, but I think it should be done thoughtfully and logically, and in a **separate repository**.

This will require more support compared to the addition to this repo and not consistently with the its current state: two custom serializers, dataloader, memory cache, di - all of these **can** be moved in other repos. Yeah, it's old conversation about how to host "core" code and all infrastructure around it. IIRC it was you who wanted to merge server project here. So... I just look now into solution tree and really don't feel any inconvenience to expand its ecosystem with one more direction - Scalars folder.

> Do we really need to publish another NuGet package just for the sake of a dozen lines of code? No.

The obvious answer may not be so obvious in the future. I do not want to speculate about the degree of usability. I'm ready to spend time and make PR even if only to see how it will look and feel and use preview builds for my purposes.

One more though - usability of custom scalars may be a cumulative effect. I know that at least for developers in my company the use of their custom scalars is not a thing that they are ready to engage in. They would prefer to take a finished proven design from "official" repo, albeit this design consists of 100 lines of code. So indeed, Base64/Json/IPAddress/etc scalars are the real candidates in my mind to appear in future as part of GraphQL.NET either in main package or in "satellite" ones under Scalars folder.

Shane32:
> I know that at least for developers in my company the use of their custom scalars is not a thing that they are ready to engage in.

Assuming you're talking about a large company, then I cannot explain how utterly silly and ludicrous this sounds to me.  I must be misunderstanding you.

Bottom line: I don't want to support a custom scalar _just_ because your company is too cheap to write/support 23 lines of code.  Let's see some other community members agree that we should be adding an additional NuGet package for these 23 (or 100) lines of code.

sungam3r:
I am satisfied with the situation when I will develop a branch with such scalars for some time even without merhing in `master`.

> Assuming you're talking about a large company, then I cannot explain how utterly silly and ludicrous this sounds to me. I must be misunderstanding you.

This is a matter of priority and business. Often business issues and technological development issues are on different scales. We have our own scalars of course.

Shane32:
It seems that in the past few weeks we have decided to close this PR?

sungam3r:
I will clean all that when I gather my thoughts.

PR #3352, opened 2022-10-07
Improve field description inheritance performance

cstichlberger:
On a large-ish GraphQL schema of mine producing the schema already quite a bit of time.

I've tracked down the culprit to the following code in __Field.cs which performs description inheritance (fills in empty field descriptions from corresponding interface fields, which was added in PR #1470 as a fix for issue #1004).

```
// We have to iterate over all schema types because FieldType has no reference to the GraphType to which it belongs.
foreach (var item in context.Schema.AllTypes.Dictionary)
{
   if (item.Value is IComplexGraphType fieldOwner && fieldOwner.Fields.Contains(context.Source))
```

Looping over the whole schema is pretty inefficient, increasing with O(numberOfFieldsWithoutDescription * numberOfTypes * numberOfFields).

This PR solves that by keeping a dictionary of the fields' owners and finding a field's owning type by dictionary lookup. It brings down schema generation time from 4 seconds to 1 seconds for my scenario (and more the bigger a schema gets).

The functionality in __Field.cs is already under test via Issue1004.cs, where I've taken the liberty of extending the test a bit by adding the counter-example of a field that's empty but isn't inherited from an interface.

Note that the new SchemaTypes.BuildFieldOwners method intentionally uses the dictionary indexer (and not Dictionary.Add) to avoid breaking existing schemas that reuse fields for different types (in case someone does this). The new test case in SchemaTypesTest aims at this scenario.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Issue1004.cs
+++ b/src/GraphQL.Tests/Bugs/Issue1004.cs
@@ -53,7 +53,7 @@ public class Issue1004Query : ObjectGraphType
         IsTypeOf = o => true;
         Field<StringGraphType>("field1").Resolve(_ => throw null);
         Field<StringGraphType>("field2").Description("Not so important").Resolve(_ => throw null);
-        Field<StringGraphType>("nonInterfaceField").Resolve(_ => throw null);
+        Field<StringGraphType>("nonInterfaceField").Resolve(_ => throw null); // https://github.com/graphql-dotnet/graphql-dotnet/pull/3352
         Interface<Issue1004Interface>();
     }
 }
--- a/src/GraphQL.Tests/Bugs/Issue1004.cs
+++ b/src/GraphQL.Tests/Bugs/Issue1004.cs
@@ -26,6 +26,9 @@ public class Issue1004 : QueryTestBase<DescriptionFromInterfaceSchema>
       },
       {
         ""description"": ""Not so important""
+      },
+      {
+        ""description"": null
       }
     ]
   }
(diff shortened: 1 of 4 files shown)


cstichlberger:
Good point, I've moved the field description inheritance fully to SchemaTypes - this also makes things simpler, now there's no need anymore for looking up a field's owner when we're working with the fieldOwners directly

Shane32 (approved):
Looks good to me

src/GraphQL.Tests/Bugs/Issue1004.cs
```diff
@@ -50,6 +53,7 @@ public Issue1004Query()
         IsTypeOf = o => true;
         Field<StringGraphType>("field1").Resolve(_ => throw null);
         Field<StringGraphType>("field2").Description("Not so important").Resolve(_ => throw null);
+        Field<StringGraphType>("nonInterfaceField").Resolve(_ => throw null);
```
sungam3r, line 56:
```suggestion
        Field<StringGraphType>("nonInterfaceField").Resolve(_ => throw null); // https://github.com/graphql-dotnet/graphql-dotnet/pull/3352
```

sungam3r:
> to avoid breaking existing schemas that reuse fields for different types (in case someone does this).

It's rather strange and error-prone use case. I doubt someone does so. And **if** someone does your changes **may** overwrite fields descriptions now 😄 .

cstichlberger:
> > to avoid breaking existing schemas that reuse fields for different types (in case someone does this).
> 
> It's rather strange and error-prone use case. I doubt someone does so. And **if** someone does your changes **may** overwrite fields descriptions now 😄 .

Yeah, true, it's a rather weird case, I just wanted to avoid throwing an ArgumentException due to the duplicate key just in case.

Thanks for the suggested changes, these make perfect sense.

sungam3r (approved):

PR #3362, opened 2022-10-16
Fix GraphQLName for classes named Type and GraphType

Shane32:
Previously, an empty string would be returned, which would throw during `Name = typeof(TSource).GraphQLName();` before the name was altered via attributes.

### DIFF
--- a/src/GraphQL.Tests/TypeExtensionTests.cs
+++ b/src/GraphQL.Tests/TypeExtensionTests.cs
@@ -17,4 +17,19 @@ public class TypeExtensionTests
     {
         type.IsNamedType().ShouldBe(expected);
     }
+
+    [Theory]
+    [InlineData(typeof(Type), "Type")]
+    [InlineData(typeof(Guid), "Guid")]
+    [InlineData(typeof(ScalarGraphType), "Scalar")]
+    [InlineData(typeof(NonNullGraphType<ListGraphType<IdGraphType>>), "Id")]
+    [InlineData(typeof(TestType), "Test")]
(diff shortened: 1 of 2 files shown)


Shane32:
@sungam3r Can you review this today?

Shane32:
I'll be merging today if there are no comments

src/GraphQL.Tests/TypeExtensionTests.cs
```diff
@@ -17,4 +17,19 @@ public void IsNamedGraphType(Type type, bool expected)
     {
         type.IsNamedType().ShouldBe(expected);
     }
+
+    [Theory]
+    [InlineData(typeof(Type), "Type")]
```
sungam3r, line 22:
```suggestion
    [InlineData(typeof(Type), "Type")]
    [InlineData(typeof(GraphType), "GraphType")]
```

sungam3r:
> before the name was altered via attributes.

By the way do you think that it is still worth to have
```c#
 var attr = type.GetCustomAttribute<GraphQLMetadataAttribute>();

            if (!string.IsNullOrEmpty(attr?.Name))
            {
                return attr!.Name!;
            }
```

at the very beginning of ` public static string GraphQLName(this Type type)` ? For some cases `ConfigureGraph();` is right after and it will  overwrite graph type name.

sungam3r (approved):

Shane32:
> By the way do you think that it is still worth to have

Perhaps; I have not considered it either way.  I consider that usage to be legacy/obsolete behavior.

PR #3379, opened 2022-11-03
Bump setup-dotnet action version

sungam3r:
https://github.com/marketplace/actions/setup-net-core-sdk

@zcsizmadia , regarding #3329 - no more `include-prerelease` option. Use `dotnet-quality` instead.


sungam3r:
Rel: #3361

Issue #2887, opened 2022-01-28
Drop support of old options in SchemaPrinterOptions


sungam3r:
@Shane32  I propose to drop this. Actual spec does not allow use ',' instead of '&amp;' when inheriting a type from multiple interfaces. Also support of this feature requires changes in parser project (little but...) .

Rel: #2868

Shane32:
I agree

sungam3r:
And even this one
![изображение](https://user-images.githubusercontent.com/21261007/151465251-057c5a8e-9f8a-4897-8351-c55570b8fd08.png)

8 years left...

Shane32:
Agree.  

In general, current version should support current-1 spec at most. People can use v4 if they need old spec support.

sungam3r:
Or inherit from `SDLWriter` and override desired parts.

PR #3382, opened 2022-11-05
Printer

### LINKED ISSUE #2868: Rewrite SchemaPrinter
Convert `Schema` into `GraphQLDocument` and then use `SDLWriter`.

sungam3r:
This is my very old PR that I wrote ~ 1 year ago but did not publish. It requires re-reading, because I remember almost nothing. I managed to rewrite schema printing moving it to use new classes from the parser project, but there were problems with the federation. Rel: #3144, it also has changes in printing. I was focused entirely on printing, not federation. These PRs may be reviewed together and may help each other to find a better way of doing printing/federation. Also I would like (if possible) to split PRs into smaller ones and merge step by step.

Closes #2868 , #2887


sungam3r:
@Mithras I would like you to help with this one in federation part. I don't use federation.

Shane32:
In general, the schema printer operation should be independent of whether a schema was built schema first or code first, whether it is a federated schema or not. 

But I think we should concentrate on (and merge) #3144 first. The existing schema printer, while perhaps inefficient, does work, and there have been multiple requests for code first federation plus an outstanding PR. 

I’d be glad to help review this PR - but let’s have the priority on #3144

sekizi:
It seems to be much more than just printer and changes seem to overlap with #3144. This PR is still only supports V1 Federation which is essentially deprecated. I think it's better to merge #3144 first because it adds support for 3 key features: Federation 2, code first and allows using DataLoader in Federated resolvers. If you want to merge printer before #3144, maybe we can split this PR in two: printer changes and Federation changes, and then merge only printer?

Mithras:
It seems to be much more than just printer and changes seem to overlap with https://github.com/graphql-dotnet/graphql-dotnet/pull/3144. This PR is still only supports V1 Federation which is essentially deprecated. I think it's better to merge https://github.com/graphql-dotnet/graphql-dotnet/pull/3144 first because it adds support for 3 key features: Federation 2, code first and allows using DataLoader in Federated resolvers. If you want to merge printer before https://github.com/graphql-dotnet/graphql-dotnet/pull/3144, maybe we can split this PR in two: printer changes and Federation changes, and then merge only printer?

Shane32:
Closing as finished by #3649 .  Branch can be restored if desired.

PR #3395, opened 2022-11-09
Introduce ExecutionError.Extensions

### LINKED ISSUE #3354: Add Extensions dictionary to ExecutionError object
## Summary

Consider adding an Extensions dictionary (String, object) to the ExecutionError object to honour the GraphQL spec:
https://spec.graphql.org/October2021/#sec-Errors
https://spec.graphql.org/October2021/#example-8b658

### Basic example

Developers can then add custom properties to the errors collection via the dictionary in the usual manner, eg:
```
try {
    ...
} catch(Exception ex) { 
    ExecutionError executionError = new ExecutionError("Oh dear, that went horribly wrong");
    executionError.Extensions.Add("customError", "Your custom error here")
    context.Errors.Add(executionError);
}
```
The output of which would be something like:
```
{
"data": {
    "someMutation": null
},
"errors": [{            
        "message": "Oh dear, that went horribly wrong",            
        "extensions": {
            "customError": "Your custom error here"                
        }            
    }]
}
```
### Motivation
This should allow for simplified exception handling with custom errors, and also honour the GraphQL spec.

PS: Have asked this question here too:
https://stackoverflow.com/questions/74012743/why-doesnt-graphql-net-honour-the-errors-extensions-schema

sungam3r:
fixes #3354

Also demonstrates raw string literals feature https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#raw-string-literals

### DIFF
--- a/src/GraphQL.Tests/Serialization/ExecutionErrorTests.cs
+++ b/src/GraphQL.Tests/Serialization/ExecutionErrorTests.cs
@@ -24,8 +24,8 @@ public class ExecutionErrorTests
     public void Extensions(IGraphQLTextSerializer serializer)
     {
         var error = new ExecutionError("some error 1")
-            .WithExtension("severity", "warn")
-            .WithExtension("rank", 42);
+            .AddExtension("severity", "warn")
+            .AddExtension("rank", 42);
 
         var expected = """
 {
--- a/src/GraphQL.Tests/Serialization/ExecutionErrorTests.cs
+++ b/src/GraphQL.Tests/Serialization/ExecutionErrorTests.cs
@@ -21,6 +21,29 @@ public class ExecutionErrorTests
 
     [Theory]
     [ClassData(typeof(GraphQLSerializersTestData))]
+    public void Extensions(IGraphQLTextSerializer serializer)
+    {
+        var error = new ExecutionError("some error 1")
+            .WithExtension("severity", "warn")
+            .WithExtension("rank", 42);
+
+        var expected = """
+{
+  "message": "some error 1",
+  "extensions": {
+    "severity": "warn",
+    "rank": 42
+  }
+}
+""";
+
+        var actual = serializer.Serialize(error);
+
+        actual.ShouldBeCrossPlatJson(expected);
+    }
+
+    [Theory]
+    [ClassData(typeof(GraphQLSerializersTestData))]
     public void Null(IGraphQLTextSerializer serializer)
     {
         var expected = "null";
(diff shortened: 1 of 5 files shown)


src/GraphQL.Tests/Serialization/ExecutionErrorTests.cs
```diff
@@ -19,6 +19,29 @@ public void Simple(IGraphQLTextSerializer serializer)
...
+    [Theory]
+    [ClassData(typeof(GraphQLSerializersTestData))]
+    public void Extensions(IGraphQLTextSerializer serializer)
+    {
+        var error = new ExecutionError("some error 1")
+            .WithExtension("severity", "warn")
+            .WithExtension("rank", 42);
+
+        var expected = """
+{
```
sungam3r, line 31: @Shane32 I would like to eventually switch all tests to use raw strings instead of `@` with `""`.

src/GraphQL/Execution/ErrorInfoProvider.cs
```diff
@@ -56,8 +56,9 @@ public virtual ErrorInfo GetInfo(ExecutionError executionError)
                 var details = _options.ExposeExceptionDetails && _options.ExposeExceptionDetailsMode == ExposeExceptionDetailsMode.Extensions
                     ? executionError.ToString()
                     : null;
+                var userExtensions = executionError.Extensions;
 
-                if (code != null || codes != null || data != null || details != null)
+                if (code != null || codes != null || data != null || details != null || userExtensions != null)
```
Shane32, line 61:
```suggestion
                if (code != null || codes != null || data != null || details != null || userExtensions?.Count > 0)
```

Add `?.Count > 0` or no?  Either way is fine with me.

src/GraphQL/Execution/Errors/ExecutionError.cs
```diff
@@ -107,5 +120,15 @@ public static TError AddLocation<TError>(this TError error, ASTNode? abstractNod
             error.AddLocation(Location.FromLinearPosition(document.Source, abstractNode.Location.Start));
             return error;
         }
+
+        /// <summary>
+        /// Adds an extension to <see cref="ExecutionError.Extensions"/>.
+        /// </summary>
+        public static TError WithExtension<TError>(this TError error, string key, object? value)
```
Shane32, line 127: I don't like the fact that now an error is going to have a `.AddExtension` and a `.WithExtension` method (essentially).  How can we fix this?  Make `AddExtension` internal or move the code into this method?  Remove this extension method?

sungam3r: I expected such comment. See `AddLocation` instance method and extension method. They have different signatures so there are no problems. I would like to have single instance method but then I could not chain calls. I removed instance method and rename extension method.

sungam3r: I mean could not chain calls **without loosing initial type of error**.

Shane32 (approved):

PR #3398, opened 2022-11-10
Introduce generic attributes

sungam3r:
Demonstration.

See https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#generic-attributes

As I understand these may be problems on machines with compiler version less than 17.1

### DIFF
--- a/src/GraphQL/Attributes/OutputTypeAttribute.cs
+++ b/src/GraphQL/Attributes/OutputTypeAttribute.cs
@@ -6,30 +6,14 @@ namespace GraphQL
     /// Specifies an output graph type mapping for the CLR class or property marked with this attribute.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Struct | AttributeTargets.Property | AttributeTargets.Method | AttributeTargets.Field)]
-    public class OutputTypeAttribute : GraphQLAttribute
+    public class OutputTypeAttribute<TGraphType> : GraphQLAttribute
+        where TGraphType : IGraphType
     {
-        private Type _outputType = null!;
-
-        /// <inheritdoc cref="OutputTypeAttribute"/>
-        public OutputTypeAttribute(Type graphType)
+        /// <inheritdoc cref="OutputTypeAttribute{TGraphType}"/>
+        public OutputTypeAttribute()
         {
-            OutputType = graphType;
-        }
-
-        /// <inheritdoc cref="OutputTypeAttribute"/>
-        public Type OutputType
-        {
-            get => _outputType;
-            set
-            {
-                if (value == null)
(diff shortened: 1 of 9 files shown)


src/GraphQL/Attributes/OutputTypeAttribute.cs
```diff
@@ -6,38 +6,22 @@ namespace GraphQL
     /// Specifies an output graph type mapping for the CLR class or property marked with this attribute.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Struct | AttributeTargets.Property | AttributeTargets.Method | AttributeTargets.Field)]
-    public class OutputTypeAttribute : GraphQLAttribute
+    public class OutputTypeAttribute<TGraphType> : GraphQLAttribute
```
sungam3r, line 9: Generic attributes save a couple lines here and provide better checks.

src/GraphQL/Extensions/TypeExtensions.cs
```diff
@@ -164,9 +164,9 @@ public static Type GetGraphTypeFromType(this Type type, bool isNullable = false,
                 }
                 else if (mode == TypeMappingMode.OutputType)
                 {
-                    var outputAttr = type.GetCustomAttribute<OutputTypeAttribute>();
-                    if (outputAttr != null)
-                        graphType = outputAttr.OutputType;
+                    //var outputAttr = type.GetCustomAttribute<OutputTypeAttribute>();
```
sungam3r, line 167: But also we have a problem when reading those attributes - should specify generic type.

sungam3r:
> As I understand these may be problems on machines with compiler version less than 17.1

Well, I think one can use generic attributes even for "old" TFMs like net5/netstandard2.0. The key here is compiler version, not TFM.

src/GraphQL/Attributes/OutputTypeAttribute.cs
sungam3r: > as it would require manually changing LangVersion (which is not always an acceptable answer). We have had issues before along the same lines with init;

I thought that adding LangVersion in csproj is not a big deal. Nevertheless it's only for demo purposes. I will try to change attributes to work in both cases.

src/GraphQL/Extensions/TypeExtensions.cs
sungam3r: Yeah, the resolution was obvious.

sungam3r:
изображение
Problems with net48 TFM, 178 tests failed.

sungam3r:
So generic attributes can not be used in tests for NET48.

sungam3r:
Fixed with `#ifdef`s.

Shane32 (approved):

PR #3401, opened 2022-11-12
Simplify versioning

Shane32:


.github/workflows/build-artifacts-code.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ env.NODE_VERSION }}
-        uses: actions/setup-node@v3.5.1
+        uses: actions/setup-node@v3
```
sungam3r, line 25: Does this mean that the latest 3.*.* version will be taken?

Shane32: Technically, no.  Practically, yes.

src/GraphQL.DataLoader.Tests/GraphQL.DataLoader.Tests.csproj
```diff
@@ -18,9 +18,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Bogus" Version="34.0.2" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="3.1.20" />
-    <PackageReference Include="System.Reactive" Version="5.0.0" />
+    <PackageReference Include="Bogus" Version="34.*" />
```
sungam3r, line 21: I am rather for such a change than against it, although it brings a certain degree of non-determinance in the build.

sungam3r (approved):

.github/workflows/build-artifacts-code.yml
Shane32: Technically this is the "v3" tag.  When GitHub issues a new release, they modify their old "v3" tag to point to the latest 3.x tag.  This is standard practice across the GitHub official repos that support workflows.  But if they were not to do this, the v3 tag would remain on the same commit.

Any developer that writes a workflow script can actually change the tag to point to anything they want.  So if for security you actually wanted to pin to a certain release/commit, without the possibility that the developer could change it on you, you'd use `@sha` where `sha` is the sha code of the commit.

sungam3r: Ok, thanks.

Issue #3374, opened 2022-10-28
Required service for type GraphQL.Types.LongGraphType not found occurred


mariusbancila:
While starting up my GraphQL app (using the latest version of graphql-dotnet, 7.1.1) I get this error in the output window:

```
Required service for type GraphQL.Types.LongGraphType not found occurred
```

This is followed by:
```
Error: This graph type 'Query' has already been initialized. Make sure that you do not use the same instance of a graph type in multiple schemas. It may be so if you registered this graph type as singleton; see https://graphql-dotnet.github.io/docs/getting-started/dependency-injection/ for more info. occurred
```
(This later one is quite misleading. You need to fix the previous ones and then this one goes away. I was tricked by this for a while.)

But what service could be missing for `LongGraphType`?

mariusbancila:
I found this in the sources:

```
        public static object GetRequiredService(this IServiceProvider provider, Type serviceType)
        {
            if (provider == null)
                throw new ArgumentNullException(nameof(provider));

            if (serviceType == null)
                throw new ArgumentNullException(nameof(serviceType));

            object? service = provider.GetService(serviceType);
            if (service != null)
                return service;

            throw new InvalidOperationException($"Required service for type {serviceType} not found");
        }
```

This is also misleading. It's not that a service is missing for the `LongGraphType`, the `LongGraphType` is a service that is missing.

Shane32:
We should modify the error message to match the typical .NET message for this scenario.

But I'd really like to discover why it triggers the `This graph type 'Query' has already been initialized` error.  Certainly it would seem that the two messages are unrelated.  If you can provide a sample that reproduces the error, it would be helpful.

Shane32:
If you can provide a sample, then I can investigate.

mariusbancila:
This happens when I removed obsolete calls:

```
var field = tableType.Field(
            columnTypeDT.GetType(),
            columnName,
            description: "...",
            resolve: ctx =>
```
With the following:
```
var field = tableType.AddField(
            new FieldType()
            {
            ResolvedType = columnTypeDT,
            Name = columnName,
            Description = "...",
            Resolver = new FuncFieldResolver<object>(ctx =>
```
When using `Type` instead of `ResolvedType` then the error is gone!

Shane32:
[...]

PR #3402, opened 2022-11-12
Change error message for GetRequiredService

sungam3r:
Inspired by #3374

### DIFF
--- a/src/GraphQL/Utilities/ServiceProviderExtensions.cs
+++ b/src/GraphQL/Utilities/ServiceProviderExtensions.cs
@@ -28,7 +28,7 @@ namespace GraphQL.Utilities
             if (service != null)
                 return service;
 
-            throw new InvalidOperationException($"Required service for type {serviceType} not found");
+            throw new InvalidOperationException($"No service for type '{serviceType}' has been registered.");
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/GraphQL/Utilities/ServiceProviderExtensions.cs
```diff
@@ -28,7 +28,7 @@ public static object GetRequiredService(this IServiceProvider provider, Type ser
             if (service != null)
                 return service;
 
-            throw new InvalidOperationException($"Required service for type {serviceType} not found");
+            throw new InvalidOperationException($"No service for type '{serviceType}' has been registered.");
```
sungam3r, line 31: Copied from MS implementation.

Shane32: I suggest we use the friendly name extension method we have for better support of generic types

Shane32:
```suggestion
            throw new InvalidOperationException($"No service for type '{serviceType.GetFriendlyName()}' has been registered.");
```

sungam3r: ok

Shane32 (approved):

PR #3404, opened 2022-11-13
Move custom built-in scalars into new package

sungam3r:
This is the first change I suggest to include for v8. I believe that _custom built-in_ scalars is a bit awkward term. Core repo should (IMO) implement only 5 scalars from spec. All other scalars can be plugged in by APIs that we had already provided. Not everyone needs custom scalars like `Long` or `Uri`. Moreover having those built-in eventually leads to conflicts with user-defined ones so we suggest ways to "reregister/override". Also this PR is slightly simplifies `SchemaTypes`.

My doubts now are:
1. Package/namespaces naming.
2. Failed tests - all tests from `GetGraphTypeFromType_Matrix` + some others. This is because `GetGraphTypeFromType` static method has no access to custom scalars now since `SchemaTypes.BuiltInScalarMappings` contains only mappings for 5 scalars.
3. Semantic shift for `BuiltInScalarMappings`. Now this collection represents ONLY types that map to 5 scalars from spec. Shoul be documented.

Options I see for failed tests:
1. Just remove failed tests 😄 
2. Provide static property like ones from `GlobalSwitches` to inject more scalars to be used from `GetGraphTypeFromType`.

Rel: #3349


src/GraphQL.Tests/Bugs/BugWithCustomScalarsInDirective.cs
```diff
@@ -24,6 +24,7 @@ public class BugWithCustomScalarsInDirectiveSchema : Schema
     public BugWithCustomScalarsInDirectiveSchema(IServiceProvider provider, BugWithCustomScalarsInDirectiveQuery query)
         : base(provider)
     {
+        this.RegisterBCLScalars();
```
sungam3r, line 27: Not all tests require this. For first run ~350 tests failed for 5 platforms so there are ~70 "real" tests that use custom scalars.

Shane32:
Sorry, but I am very very much against this PR. 

The custom scalars within GraphQL.NET mirror intrinsic .NET types. They may not be native to GraphQL but they are native to .NET.  Further, at least one will be needed by almost every schema. 

Separating into another package simply increases the burden on the developer to reference another package and provide additional setup.

It’s one thing to add a separate nuget package for custom scalars that require references to third party packages, but that’s not the case here. Not only that, these are not some unique custom data model within some random namespace, like the Version class we discussed in a separate PR, but rather these are .NET root value types. This is the GraphQL.NET (emphasis on .NET) implementation. We should support .NET intrinsic types.

Finally, I don’t know how this PR benefits anyone. Separation in this case simply isn’t a benefit.

joemcbride:
> Sorry, but I am very very much against this PR.

I have to agree with this. Providing a set of graph types for the .NET primitives in the core project is a huge benefit. Needing to add one more package would just be annoying.

sungam3r:
> Further, at least one will be needed by almost every schema.

Which one?

> Separating into another package simply increases the burden on the developer to reference another package and provide additional setup.

Only for those who really use some scalars above 5 built-in into spec. Even our own tests showed that this is required far from all cases. Additional setup is 1 line of code.

Copied from https://github.com/graphql-dotnet/graphql-dotnet/pull/3349#issuecomment-1289912437 :

> I don't think this is a fair comparison. Most of the "additional" types are .NET intrinsic types, such as short or decimal. They are not custom classes; they are **root types** of the .NET language. This is true of our date/time scalars as well, although our implementation is our own definition. The only other types we have is Guid and Uri scalars, which are extremely well known and often-used .NET Standard types.
Adding a custom scalar for the .NET Standard Version class doesn't fit with any of the above scalars. Rather it's just one of a dozen (or dozens) of classes within the .NET library that might be of interest to a developer, and a rather uncommonly used one at that. IPAddress, IPEndpoint and DnsEndPoint all fit in this category as well, and are probably used more often than Version is. We could also add scalars for base-64 encodings of data strings (either as byte arrays or MemoryStreams), and many others, all from the .NET Standard library.

So... moving to other package is bad. Adding new `Version` built-in scalar is bad as well, right?  The evaluation criterion is whether it is root types or not root seems to me very subjective. The general question from me is - what is a strategy for extension existing scalars set?
- Drop it into core project if this scalar is one of ones like `IPAddress` and `DnsEndPoint` ?
- Drop it into new project into this repo if this scalar is one of ones like `SemVer` that has some dependencies ?
- Just forbid to expand (seal) **root scalars** set here? This option seems to me dishonest in relation to other types ("non-root" if you like), because in the reasons for this attitude I see only a subjective opinions but technically nothing prevents you from expanding this scalars set further.

Shane32:
Which types are needed depend on the schema.  I personally would find it unlikely that very many schemas can live without some sort of date scalar.  Then also the underlying data format dictates the numeric formats; if using `decimal` with the data source, for example with the SQL Server `money` type, then you'd likely use the decimal scalar.  Similar for `short` and so on.

> The evaluation criterion is whether it is root types or not root seems to me very subjective. 

I don't think so.  .NET primitives are [well defined](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types).  Guid, Uri and the date types are perhaps not true .NET primitives, although are commonly used enough to be (subjectively) considered a primitive.  (`Date` actually is a VB.NET keyword mapping to the .NET `DateTime` type.)  And they accordingly all live in the `System` namespace.

`Version` is not a true primitive, nor is it a commonly used class, nor does it live in the `System` namespace.

> Drop it into core project if this scalar is one of ones like IPAddress and DnsEndPoint ?

If it's a .NET type, sure, we can add it to the core project.  We can decide how/where to add them, if indeed they should be added any differently than the rest.  The `Version` class/scalar fits into this category.

> Drop it into new project into this repo if this scalar is one of ones like SemVer that has some dependencies ?

Unpublished, as example code (with tests)?  Sure.  Published?  Well, it certainly makes no sense to have one additional repo with many dependencies because of many random scalars that were added over time.  I don't think it makes much sense to create many projects each having only a few lines of code in them.  Note that adding these additional dependencies into the main repo creates additional work for maintainers of the project, unpublished or not.  There's also versioning issues, as the custom scalar repo must now follow the GraphQL.NET versioning and also the third-party dependency versioning.

That is why I suggested a new repo with all these additional scalars that require dependencies, with clear naming conventions laid out for the (many) NuGet packages.

> Just forbid to expand (seal) root scalars set here? 

No; I see no reason to do so.  We just added `DateOnly` and `TimeOnly` support, as it fits within all the same guidelines as the rest of the GraphQL.NET built-in scalars.

sungam3r:
Regarding your reference to https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types . By your logic we have already brought a bunch of scalars, which do not represent "root types".

- BigIntGraphType - BigInteger from System.Numerics namespace
- DateGraphType - DateTime
- DateOnlyGraphType - DateOnly
- DateTimeGraphType - DateTime
- DateTimeOffsetGraphType - DateTimeOffset
- GuidGraphType - Guid
- TimeOnlyGraphType - TimeOnly
- TimeSpanMillisecondsGraphType  - custom scalar that works for many underlying .NET types
- TimeSpanSecondsGraphType - custom scalar that works for many underlying .NET types
- UriGraphType - Uri


>  Version is not a true primitive, nor is it a commonly used class, **nor does it live in the System namespace**.

Well...
изображение

Nevertheless, the summary is:
1. OK to add Version scalar into GraphQL/Types/Scalars
2. OK to add some other scalars in future if these scalars "look and feel" like Version/Guid/whatever from `System` namespace
3. NOT OK to add SemVer scalar into this repo even in separate project. Right? So I have to create my own private package.
4. NOT OK to move existing (root or non-root) scalars out of core project. Decline PR.

joemcbride:
Perhaps it is a bit subjective, but this project does have to draw the line somewhere. All of the added types you listed I have used in multiple projects (perhaps besides GUID, since I just always use ID for that). I have never needed version. This project won't always cover ever use case, but it can cover the "90%" use cases.

sungam3r:
> I have never needed version

I have never needed `Sbyte`, `UShort`, `ULong`, `BigInt`... but I do need `Version` :)

Shane32:
I believe we have decided to close this PR also.

PR #3414, opened 2022-11-21
Add test for interface with no underlying object type

Shane32:
Based on what I can see in the GraphQL spec, it is allowed by the spec to have an interface where there are no object types that implement the interface.

> GraphQL interfaces represent a list of named fields and their arguments. GraphQL objects can then implement these interfaces which requires that the object type will define all fields defined by those interfaces.

Note word "can".  It does say this also: 

> Result Coercion
>
> The interface type should have some way of determining which object a given result corresponds to. Once it has done so, the result coercion of the interface is the same as the result coercion of the object.

But I take this to mean in the context of the application.  (and it says "should", not "required")  Since we can define field resolvers on an interface, they can be executed on that basis, and so it is not _necessary_ to relate a result to the object type.

This logic is most useful in GraphQL.NET with the `AutoRegisteringInterfaceGraphType` class, I believe, as .NET's interfaces don't need to be mapped to a class to be useful, and this behavior mimics that.

This PR simply asserts the existing behavior with a new test, and does not change any code/behavior.  However, this code is at odds with the `throws_when_unable_to_determine_object_type` test.  The reason both tests pass currently is because the check for the code only executes when the field referencing the interface is a nullable field (which is a bug).

How should we address this?

1. Allow any type returned from a field to be passed to an interface resolver.  Bad for diagnosing issues, as a cast may occur within the execution framework before it hits a resolver, without a clear indication of what the trouble is.
2. Same as above, but add IsTypeOf to IInterfaceGraphType so that it can be assured that the type is compatible, with the default implementation to return `true` for anything that matches `<T>` in `InterfaceGraphType<T>`, and `false` otherwise.
3. Add IsTypeOf, but always return `false` by default, requiring that there is an object graph type for any returned value.  Within `AutoRegisteringInterfaceGraphType.IsTypeOf`, return `true` appropriately.
4. Always require an object graph type for any returned value.

I suggest option 3 above.  This is probably most compatible with our existing behavior.

### DIFF
--- a/src/GraphQL.Tests/Types/InterfaceTests.cs
+++ b/src/GraphQL.Tests/Types/InterfaceTests.cs
@@ -1,0 +1,47 @@
+using GraphQL.Types;
+
+namespace GraphQL.Tests.Types;
+
+public class InterfaceTests : QueryTestBase<InterfaceTests.MySchema>
+{
+    [Fact]
+    public void InterfaceWithNoObject()
+    {
+        AssertQuerySuccess("{ test { id } }", """
+            {
+                "test": {
+                    "id": "123"
+                }
+            }
+            """);
+    }
+
+    public class MySchema : Schema
+    {
+        public MySchema()
+        {
+            Query = new MyQuery();
+        }
+    }
+
+    public class MyQuery : ObjectGraphType
+    {
+        public MyQuery()
+        {
+            Field<NonNullGraphType<MyInterfaceType>>("test", resolve: _ => new MyObject());
+        }
+    }
+
+    public class MyInterfaceType : InterfaceGraphType<MyObject>
+    {
+        public MyInterfaceType()
+        {
+            Field("id", x => x.Id);
+        }
+    }
+
+    public class MyObject
+    {
+        public string Id { get; set; } = "123";
+    }
+}
--- a/src/GraphQL.Tests/Types/InterfaceTests.cs
+++ b/src/GraphQL.Tests/Types/InterfaceTests.cs
@@ -7,10 +7,46 @@ public class InterfaceTests : QueryTestBase<InterfaceTests.MySchema>
     [Fact]
     public void InterfaceWithNoObject()
     {
-        AssertQuerySuccess("{ test { id } }", """
+        AssertQuerySuccess(
+            """
             {
-                "test": {
+                noObjectTest {
+                    id                    # as there is no object type defined for this
+                                          # result, the interface resolver will be used
+                                          # to return "123_interface"
+                }
+            }
+            """,
+            """
+            {
+                "noObjectTest": {
                     "id": "123"
+                }
+            }
+            """);
+    }
+
+    [Fact]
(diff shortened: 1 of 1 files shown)


sungam3r:
>  Since we **can define field resolvers on an interface**, they **can be executed on that basis**, and so it is not necessary to relate a result to the object type.

😕  I consider defining resolvers on interfaces is a one of the main design flaws in GraphQL.NET. The same one as resolvers for input objects. Rel - #1176

> it is not necessary to relate a result to the object type

What about `__typename` ? What should return then?

sungam3r:
изображение

Shane32:
> What about `__typename` ? What should return then?

Good point.  Maybe we must require an object.

src/GraphQL.Tests/Types/InterfaceTests.cs
```diff
@@ -0,0 +1,110 @@
...
+public class InterfaceTests : QueryTestBase<InterfaceTests.MySchema>
+{
+    [Fact]
+    public void InterfaceWithNoObject()
+    {
+        AssertQuerySuccess(
+            """
+            {
+                noObjectTest {
+                    id                    # as there is no object type defined for this
```
sungam3r, line 14:
```suggestion
                    __typename
                    id                    # as there is no object type defined for this
```

try this

sungam3r:
You raised a question that remind me a long-forgotten topic about resolvers redesign. Just wondering - can we remove resolvers for inputs/interfaces in v8?

sungam3r:
And how do you feel about raw string literals?

Shane32:
> And how do you feel about raw string literals?

They are nice.  I personally wouldn't rewrite old tests to use them, but if you want to, sure.

sungam3r:
Somewhere in the middle of my path through thousands of changes, I began to doubt, but still finished migration).

Shane32:
> You raised a question that remind me a long-forgotten topic about resolvers redesign. Just wondering - can we remove resolvers for inputs/interfaces in v8?

I have not considered how much 'redesign' is necessary.  I'm not in favor of changing the underlying classes and interfaces "just because".  Here we would be required to separate FieldType for input object vs output object.  This can have a SIGNIFICANT impact on existing codebases.  Certainly it would break any dynamic codebase; perhaps all codebases.  I would favor adding a schema initialization check that validates that input object fields do not have resolvers assigned to them.

Interface fields fall into a similar category, although I'm sure they are not as often used.  Plus, at the moment, we have one test that ensures that interface field resolvers do function correctly (part of AutoRegisteringInterfaceGraphType).  But this PR is discussing whether we will revoke that functionality due to GraphQL spec incompatibility.

src/GraphQL.Tests/Types/InterfaceTests.cs
Shane32: Right now it crashes with a null reference exception.

Shane32:
See #3417

PR #3416, opened 2022-11-21
Fix SubFields property for interfaces and unions

Shane32:
Demonstrates that accessing `SubFields` fails from within a field resolver of a field that is of a union graph type.

### DIFF
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -204,8 +204,13 @@ namespace GraphQL.Execution
         /// <inheritdoc/>
         public virtual Dictionary<string, (GraphQLField field, FieldType fieldType)>? GetSubFields(ExecutionContext context, ExecutionNode node)
         {
+            //IGraphType specificType = node is ObjectExecutionNode parent
+            //    ? parent.GetObjectGraphType(context.Schema) ?? parent.GraphType
+            //    : node.FieldDefinition!.ResolvedType!;
+
             return node.Field?.SelectionSet?.Selections?.Count > 0
                 ? CollectFieldsFrom(context, node.FieldDefinition!.ResolvedType!, node.Field.SelectionSet, null)
+                //? CollectFieldsFrom(context, specificType, node.Field.SelectionSet, null)
                 : null;
         }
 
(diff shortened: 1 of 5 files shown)


Shane32:
@sungam3r I'm really really pressed for time, so if you want to look at this, please do so.  Otherwise I'll get back to it when I have time; not sure when that will be.

sungam3r:
Do you mean fix the bug using provided test? I think I can look into it tomorrow.

sungam3r:
I reproduced the bug and the problem is a bit wider. Actually there are 2.
1. Cast bug as is. We are trying to cast union type to complex type.
2. `IResolveFieldContext.SubFields` can't work properly in case of abstract graph types - interfaces and unions. We need to know `node.Result` first to be able to derive graph type from it.

Note that `SetSubFieldNodes` method has no such problem because is called after `node.Result` was set. I continue to look for a suitable solution. Cast exeption is just a consequence of a conceptual problem.

sungam3r:
Or other explanation. See here 
изображение

Actually we should add commented assert here. `specificType` should be specific type indeed to be able to collect right fields. Not union and not interface. By the way, `SubFields` in your test returns empty set. I do not think that this is much better than if a exception was generated.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -204,8 +204,13 @@ protected virtual void SetSubFieldNodes(ExecutionContext context, ObjectExecutio
         /// <inheritdoc/>
         public virtual Dictionary<string, (GraphQLField field, FieldType fieldType)>? GetSubFields(ExecutionContext context, ExecutionNode node)
         {
+            //IGraphType specificType = node is ObjectExecutionNode parent
```
sungam3r, line 207: I placed comments here because I'm not sure if it helps but at least I tried to align specificType argument value with one from SetSubFieldNodes

sungam3r:
Options regarding unions:
1. Throw error if one calls `SubFields` on union.
2. Return empty field set from union (as now).
3. Return field set from the first/last/random union member 🤯 .
4. Add new method with additional argument like `supposedResolverResult` to be used to chose appropriate union member 🤯 .

Options regarding interfaces:
1. Throw error if one calls `SubFields` on interface.
2. Return field set from interface (as now).

In any case, until we get a concrete runtime value from the resolver, we cannot say which fields will come in future in case of unions. For your test you may add another union member like `TypeB` with different field (name) and return appropriate runtime value from resolver. So your query 
```graphql
  __typename
                    ... on TypeA {
                        id
                    }
```
will return only one field - `__typename`.

Shane32:
So there’s a conceptual problem with both unions and interfaces. At the moment I’d opt for option 2 in both cases - no fields for unions (or just __typename) and interface fields for interfaces. 

Since we have ResolveFieldContext class which includes a copy constructor, it doesn’t make sense to throw an exception - unless we somehow patch the copy constructor. 

Besides, this makes (some) sense - until the type is known, it returns the only fields that could be resolved. 

I think it would be best for version 8 to have some more complex method of enumerating child fields. Perhaps we provide the graph type requirement or something. And we should be able to take a child field and easily enumerate its children. This has been repeatedly asked for and I could personally use it.

sungam3r:
> Since we have ResolveFieldContext class which includes a copy constructor, it doesn’t make sense to throw an exception - unless we somehow patch the copy constructor.

I don't get it.

Shane32:
> > Since we have ResolveFieldContext class which includes a copy constructor, it doesn’t make sense to throw an exception - unless we somehow patch the copy constructor.
> 
> I don't get it.

Well as of now, this code fails within resolvers of union fields:

```cs
var ctx2 = new ResolveFieldContext(context);
```

That is what is crashing the conventions project. It isn’t attempting to access subfields directly.  This is clearly a bug of GraphQL.NET - if we are going to provide a copy constructor then it should work for contexts of a union field type.

sungam3r:
Ah, I got it. I would not patch `ResolveFieldContext(IResolveFieldContext context)` in any way. I would only fix invalid cast and add some notes about unions and interfaces into xml comments.

Shane32:
Ok so ready for review/merge?

sungam3r:
Not yet. I want to add xml comments about unions tonight.

Shane32:
Mark it approved when you're done with it and I'll review and merge it.

sungam3r (approved):

Shane32:
Looks good!

PR #3417, opened 2022-11-21
Fix check for non-resolved interface type

Shane32:
Replaces #3414

This is a breaking change --- but (due to a bug) ONLY if the field that returns an interface type is not nullable.

### DIFF
--- a/src/GraphQL.Tests/Execution/AbstractTypeTests.cs
+++ b/src/GraphQL.Tests/Execution/AbstractTypeTests.cs
@@ -17,7 +17,7 @@ public class AbstractTypeErrorTests : QueryTestBase<AbstractSchema>
     [Fact]
     public void throws_when_unable_to_determine_object_type_nullable()
     {
-        var result = AssertQueryWithErrors("{ pets2 { name } }", """{ "pets2": null }""", expectedErrorCount: 1);
+        var result = AssertQueryWithErrors("{ pets2 { name } }", null, expectedErrorCount: 1);
         var error = result.Errors.First();
         error.Message.ShouldBe("Error trying to resolve field 'pets2'.");
         error.InnerException.ShouldNotBeNull();
--- a/src/GraphQL.Tests/Execution/AbstractTypeTests.cs
+++ b/src/GraphQL.Tests/Execution/AbstractTypeTests.cs
@@ -15,13 +15,53 @@ public class AbstractTypeErrorTests : QueryTestBase<AbstractSchema>
     }
 
     [Fact]
-    public void throws_when_unable_to_determine_object_type_nullable()
+    public void throws_when_unable_to_determine_object_type_nonnull()
     {
         var result = AssertQueryWithErrors("{ pets2 { name } }", null, expectedErrorCount: 1);
         var error = result.Errors.First();
         error.Message.ShouldBe("Error trying to resolve field 'pets2'.");
         error.InnerException.ShouldNotBeNull();
         error.InnerException.Message.ShouldBe("Abstract type Pet must resolve to an Object type at runtime for field Query.pets2 with value '{ name = Eli }', received 'null'.");
+    }
+
+    [Fact]
+    public void throws_when_unable_to_determine_object_type_list()
+    {
+        var result = AssertQueryWithErrors("{ pets3 { name } }", """{ "pets3": [ null ] }""", expectedErrorCount: 1);
+        var error = result.Errors.First();
+        error.Message.ShouldBe("Error trying to resolve field 'pets3'.");
+        error.InnerException.ShouldNotBeNull();
+        error.InnerException.Message.ShouldBe("Abstract type Pet must resolve to an Object type at runtime for field Query.pets3 with value '{ name = Eli }', received 'null'.");
+    }
+
+    [Fact]
+    public void throws_when_unable_to_determine_object_type_list_nonnull()
+    {
+        var result = AssertQueryWithErrors("{ pets4 { name } }", """{ "pets4": null }""", expectedErrorCount: 1);
+        var error = result.Errors.First();
+        error.Message.ShouldBe("Error trying to resolve field 'pets4'.");
+        error.InnerException.ShouldNotBeNull();
+        error.InnerException.Message.ShouldBe("Abstract type Pet must resolve to an Object type at runtime for field Query.pets4 with value '{ name = Eli }', received 'null'.");
+    }
+
+    [Fact]
+    public void throws_when_unable_to_determine_object_type_nonnull_list()
+    {
+        var result = AssertQueryWithErrors("{ pets5 { name } }", """{ "pets5": [ null ] }""", expectedErrorCount: 1);
+        var error = result.Errors.First();
+        error.Message.ShouldBe("Error trying to resolve field 'pets5'.");
+        error.InnerException.ShouldNotBeNull();
+        error.InnerException.Message.ShouldBe("Abstract type Pet must resolve to an Object type at runtime for field Query.pets5 with value '{ name = Eli }', received 'null'.");
+    }
+
+    [Fact]
+    public void throws_when_unable_to_determine_object_type_nonnull_list_nonnull()
+    {
+        var result = AssertQueryWithErrors("{ pets6 { name } }", null, expectedErrorCount: 1);
+        var error = result.Errors.First();
+        error.Message.ShouldBe("Error trying to resolve field 'pets6'.");
+        error.InnerException.ShouldNotBeNull();
+        error.InnerException.Message.ShouldBe("Abstract type Pet must resolve to an Object type at runtime for field Query.pets6 with value '{ name = Eli }', received 'null'.");
     }
 }
 
@@ -41,6 +81,10 @@ public class AbstractQueryType : ObjectGraphType
         Name = "Query";
         Field<PetInterfaceType>("pets").Resolve(_ => new { name = "Eli" });
         Field<NonNullGraphType<PetInterfaceType>>("pets2").Resolve(_ => new { name = "Eli" });
+        Field<ListGraphType<PetInterfaceType>>("pets3").Resolve(_ => new[] { new { name = "Eli" } });
+        Field<ListGraphType<NonNullGraphType<PetInterfaceType>>>("pets4").Resolve(_ => new[] { new { name = "Eli" } });
+        Field<NonNullGraphType<ListGraphType<PetInterfaceType>>>("pets5").Resolve(_ => new[] { new { name = "Eli" } });
+        Field<NonNullGraphType<ListGraphType<NonNullGraphType<PetInterfaceType>>>>("pets6").Resolve(_ => new[] { new { name = "Eli" } });
     }
 }
 
--- a/src/GraphQL.Tests/Execution/AbstractTypeTests.cs
+++ b/src/GraphQL.Tests/Execution/AbstractTypeTests.cs
@@ -13,6 +13,16 @@ public class AbstractTypeErrorTests : QueryTestBase<AbstractSchema>
         error.InnerException.ShouldNotBeNull();
         error.InnerException.Message.ShouldBe("Abstract type Pet must resolve to an Object type at runtime for field Query.pets with value '{ name = Eli }', received 'null'.");
     }
+
+    [Fact]
+    public void throws_when_unable_to_determine_object_type_nullable()
+    {
+        var result = AssertQueryWithErrors("{ pets2 { name } }", """{ "pets2": null }""", expectedErrorCount: 1);
+        var error = result.Errors.First();
+        error.Message.ShouldBe("Error trying to resolve field 'pets2'.");
+        error.InnerException.ShouldNotBeNull();
+        error.InnerException.Message.ShouldBe("Abstract type Pet must resolve to an Object type at runtime for field Query.pets2 with value '{ name = Eli }', received 'null'.");
+    }
 }
 
 public class PetInterfaceType : InterfaceGraphType
@@ -30,6 +40,7 @@ public class AbstractQueryType : ObjectGraphType
     {
         Name = "Query";
         Field<PetInterfaceType>("pets").Resolve(_ => new { name = "Eli" });
+        Field<NonNullGraphType<PetInterfaceType>>("pets2").Resolve(_ => new { name = "Eli" });
     }
 }
 
--- a/src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
@@ -613,7 +613,7 @@ public class AutoRegisteringInterfaceGraphTypeTests
         public string Name { get; set; } = null!;
         public bool IsLarge { get; set; }
     }
-    
+
     private class NoDefaultConstructorTestClass : NoDefaultConstructorTestInterface
     {
         public NoDefaultConstructorTestClass(bool value)
--- a/src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
@@ -254,22 +254,22 @@ public class AutoRegisteringInterfaceGraphTypeTests
     }
 
     [Theory]
-    [InlineData(nameof(ArgumentTestsInterface.WithNonNullString), "arg1", "hello", null, "hello")]
-    [InlineData(nameof(ArgumentTestsInterface.WithNullableString), "arg1", "hello", null, "hello")]
-    [InlineData(nameof(ArgumentTestsInterface.WithNullableString), "arg1", null, null, null)]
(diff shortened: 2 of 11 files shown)


src/GraphQL.Tests/Execution/AbstractTypeTests.cs
```diff
@@ -13,6 +13,16 @@ public void throws_when_unable_to_determine_object_type()
         error.InnerException.ShouldNotBeNull();
         error.InnerException.Message.ShouldBe("Abstract type Pet must resolve to an Object type at runtime for field Query.pets with value '{ name = Eli }', received 'null'.");
     }
+
+    [Fact]
+    public void throws_when_unable_to_determine_object_type_nullable()
```
Shane32, line 18: This test would have failed previously, since there was a bug in ExecutionStrategy that prevented checking values for nullable interface fields.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -623,14 +622,16 @@ protected virtual void ValidateNodeResult(ExecutionContext context, ExecutionNod
...
             }
 
             if (result == null)
             {
                 return;
             }
 
+            var objectType = fieldType as IObjectGraphType;
+
             if (fieldType is IAbstractGraphType abstractType)
```
Shane32, line 635: Previously this `fieldType` would not have been unwrapped by the above lines of code, and so would skip all these checks in this `if` block.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -26,14 +29,18 @@ internal static class AutoRegisteringOutputHelper
         FieldType fieldType,
         Func<MemberInfo, LambdaExpression> BuildMemberInstanceExpression,
         Func<Type, Func<FieldType, ParameterInfo, ArgumentInformation>> getTypedArgumentInfoMethod,
-        Action<ParameterInfo, QueryArgument> ApplyArgumentAttributes)
+        Action<ParameterInfo, QueryArgument> ApplyArgumentAttributes,
+        bool isInterface)
```
Shane32, line 33: This change (to an internal class) skips building field resolver for interface fields, and instead assigns a static field resolver implementation that throws an exception.

Skipping building the field resolvers saves a little time during generation of the schema, as compilation is relatively slow.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -9,6 +9,9 @@ namespace GraphQL.Types;
 /// </summary>
 internal static class AutoRegisteringOutputHelper
 {
+    private static readonly IFieldResolver _invalidFieldResolver = new FuncFieldResolver<object?>(_ => throw new InvalidOperationException("This field resolver should never be called. It is only used to prevent the default field resolver from being used."));
+    private static readonly ISourceStreamResolver _invalidStreamResolver = new SourceStreamResolver<object?>((Func<IResolveFieldContext, IObservable<object?>>)(_ => throw new InvalidOperationException("This source stream resolver should never be called. It is only used to prevent the default field resolver from being used.")));
```
Shane32, line 13: GitHub Copilot generated error message.  Feel free to change if desired.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -335,35 +336,23 @@ public void CanOverrideFieldList()
...
+        var ex = await Should.ThrowAsync<InvalidOperationException>(async () => await resolver.ResolveAsync(new ResolveFieldContext
         {
             Source = obj,
             FieldDefinition = new FieldType { Name = fieldName },
-        }).ConfigureAwait(false);
-        actual.ShouldBe(expected);
-    }
-
-    [Fact]
-    public async Task WorksWithNoDefaultConstructor()
```
Shane32, line 358: This removed test doesn't make any sense for an interface -- they never have a default constructor.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -372,35 +361,10 @@ public async Task ThrowsWhenSourceNull()
...
         (await Should.ThrowAsync<InvalidOperationException>(async () => await graphType.Fields.Find("Example1")!.Resolver!.ResolveAsync(context).ConfigureAwait(false)).ConfigureAwait(false))
-            .Message.ShouldBe("IResolveFieldContext.Source is null; please use static methods when using an AutoRegisteringInterfaceGraphType as a root graph type or provide a root value.");
+            .Message.ShouldBe("This field resolver should never be called. It is only used to prevent the default field resolver from being used.");
         (await Should.ThrowAsync<InvalidOperationException>(async () => await graphType.Fields.Find("Example2")!.Resolver!.ResolveAsync(context).ConfigureAwait(false)).ConfigureAwait(false))
-            .Message.ShouldBe("IResolveFieldContext.Source is null; please use static methods when using an AutoRegisteringInterfaceGraphType as a root graph type or provide a root value.");
-    }
-
-#if !NET48 // .NET Framework 4.8 does not support default interface implementation, so would not support static methods on an interface
-    [Fact]
-    public async Task WorksWithNullSource()
```
Shane32, line 382: This removed test doesn't really apply anymore; it's just going to throw the same exception as ThrowsWhenSourceNull.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -417,18 +381,6 @@ public void TestBasicClassNoExtraFields()
         graphType.Fields.Count.ShouldBe(2);
     }
 
-    [Fact]
-    public async Task CustomHardcodedArgumentAttributesWork()
```
Shane32, line 421: This test also doesn't really apply since resolvers don't execute.  The attribute capability is adequately tested within AutoRegisteringObjectGraphTypeTests

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -268,7 +268,7 @@ public void Argument_VerifyType(string fieldName, string argumentName, Type? arg
     [InlineData(nameof(ArgumentTests.IdIntArg), "arg1", 123, null, 123)]
     [InlineData(nameof(ArgumentTests.TypedArg), "arg1", "123", null, 123)]
     [InlineData(nameof(ArgumentTests.MultipleArgs), "arg1", "hello", 123, "hello123")]
-    public async Task Argument_ResolverTests_WithNonNullString(string fieldName, string arg1Name, object? arg1Value, int? arg2Value, object? expected)
+    public async Task Argument_ResolverTests(string fieldName, string arg1Name, object? arg1Value, int? arg2Value, object? expected)
```
Shane32, line 271: Fixed improper name

src/GraphQL.Tests/Types/AutoRegisteringObjectGraphTypeTests.cs
```diff
@@ -487,6 +488,18 @@ public void TestInheritedFields()
         graphType.Fields.Count.ShouldBe(5);
     }
 
+    [Fact]
+    public void CannotBuildWithoutMemberInstanceExpression()
```
Shane32, line 492: Increase code coverage

Shane32:
@sungam3r These changes are based on our most recent discussion; it requires returned objects from interface to resolve to a GraphQL type within the schema.  Do you have time to review?  I think it's about time for 7.2 to be released.

Shane32:
@sungam3r Whats the status here? I thought we already discussed this, and I rewrote the whole PR based on your comments.  What’s holding this up?

sungam3r:
In process of review now.

src/GraphQL/Types/Composite/AutoRegisteringInterfaceGraphType.cs
```diff
@@ -61,22 +61,15 @@ protected void BuildFieldType(FieldType fieldType, MemberInfo memberInfo)
...
-        => _sourceExpression;
-
-    private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression
-        = context => (TSourceType)(context.Source ?? ThrowSourceNullException());
-
-    private static object ThrowSourceNullException()
-    {
-        throw new InvalidOperationException("IResolveFieldContext.Source is null; please use static methods when using an AutoRegisteringInterfaceGraphType as a root graph type or provide a root value.");
-    }
+        => throw new NotSupportedException();
```
sungam3r, line 72:
```suggestion
        => throw new NotSupportedException("Interface graph types do not support field resolvers");
```

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -24,16 +27,25 @@ internal static class AutoRegisteringOutputHelper
     public static void BuildFieldType(
         MemberInfo memberInfo,
         FieldType fieldType,
-        Func<MemberInfo, LambdaExpression> BuildMemberInstanceExpression,
+        Func<MemberInfo, LambdaExpression>? BuildMemberInstanceExpression,
```
sungam3r, line 30:
```suggestion
        Func<MemberInfo, LambdaExpression>? buildMemberInstanceExpression,
```

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -84,6 +102,11 @@ internal static class AutoRegisteringOutputHelper
         {
             throw new ArgumentOutOfRangeException(nameof(memberInfo), "Member must be a field, property or method.");
         }
+        if (BuildMemberInstanceExpression == null)
```
Shane32, line 105:
```suggestion
        if (buildMemberInstanceExpressionFunc == null)
```

src/GraphQL/Execution/ExecutionStrategy.cs
sungam3r: So only `nonnull` is problem here, not `list` ?

Shane32: Hmm... I think I tested it and determined it was also a problem.  I rewrote this code a couple times and probably missed it on the rewrite.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -642,14 +641,16 @@ protected virtual void ValidateNodeResult(ExecutionContext context, ExecutionNod
                         + $" Field: {node.Name}, Type: {nonNullType}.");
                 }
 
-                objectType = nonNullType.ResolvedType as IObjectGraphType;
+                fieldType = nonNullType.ResolvedType;
             }
 
             if (result == null)
```
Shane32, line 647:
```suggestion
            if (result is ListGraphType listGraphType)
            {
                fieldType = listGraphType.ResolvedType;
            }

            if (result == null)
```

Shane32: That may not be right either; what about a list of non-null entries?  I will need to look on my PC and see.

sungam3r: изображение

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
sungam3r, line 13:
```suggestion
    private static readonly ISourceStreamResolver _invalidStreamResolver = new SourceStreamResolver<object?>((Func<IResolveFieldContext, IObservable<object?>>)(_ => throw new InvalidOperationException("This source stream resolver should never be called. It is only used to prevent the default source stream resolver from being used.")));
```

src/GraphQL.Tests/Execution/AbstractTypeTests.cs
sungam3r: That is, on the contrary - for non-nullable interface fields.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -294,7 +295,8 @@ public async Task Argument_ResolverTests_WithNonNullString(string fieldName, str
         using var provider = serviceCollection.BuildServiceProvider();
         context.RequestServices = provider;
         fieldType.Resolver.ShouldNotBeNull();
-        (await fieldType.Resolver!.ResolveAsync(context).ConfigureAwait(false)).ShouldBe(expected);
+        var ex = await Should.ThrowAsync<InvalidOperationException>(async () => await fieldType.Resolver!.ResolveAsync(context).ConfigureAwait(false)).ConfigureAwait(false);
```
sungam3r, line 298:
```suggestion
        var ex = await Should.ThrowAsync<InvalidOperationException>(fieldType.Resolver!.ResolveAsync(context)).ConfigureAwait(false);
```

Shane32: Does not compile -- ThrowAsync expects Task and ResolveAsync returns ValueTask

Shane32: Also, the suggested code would execute outside of the ThrowAsync, so any exceptions thrown before getting wrapped in a Task would not be caught by ThrowAsync.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
```diff
@@ -294,7 +295,8 @@ public async Task Argument_ResolverTests_WithNonNullString(string fieldName, str
...
+        ex.Message.ShouldBe("This field resolver should never be called. It is only used to prevent the default field resolver from being used.");
```
sungam3r, line 299: If now all the tests throw an exception, then what is the benefit of using `AutoRegisteringInterfaceGraphType`?

Shane32: All the resolver tests fail of course.  The class still builds all the fields and field arguments based on an interface, which is still very useful.  Attributes still work, except the ones that affect resolvers like `[Scoped]`.  Even `[FromServices]` is still useful as it will not generate an field argument for that method parameter (although `[Ignore]` is equally as useful for that scenario).

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
sungam3r: Understood. https://github.com/shouldly/shouldly/issues/753

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
sungam3r: Could one build `AutoRegisteringInterfaceGraphType` from interface and then modify resolvers to return something else?

sungam3r (approved):
"Semi-breaking" change here so this should be tracked in migration notes for v7 I suppose.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
sungam3r: They advice to use `Func<Task>`.

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
Shane32: Well you can, but validation checks in the execution strategy will prevent interface resolvers from being used, whether it's based on `AutoRegisteringInterfaceGraphType` or `InterfaceGraphType`, since if it resolves to an object type (as we determined was required so that `__typename` works) then the object's resolver is used instead.

sungam3r: One _may_ have a problem if the interface fields are returning now ( https://github.com/graphql-dotnet/graphql-dotnet/pull/3416#issuecomment-1324361913 ) from `SubFields` and for some reason he calls resolver manually.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: Added tests to verify validation occurs for lists

Shane32:
> "Semi-breaking" change here so this should be tracked in migration notes for v7 I suppose.

Done

src/GraphQL.Tests/Types/AutoRegisteringInterfaceGraphTypeTests.cs
Shane32: Migration notes added

PR #3423, opened 2022-11-26
New method to get variables values

### LINKED ISSUE #3418: ValidationContext.GetVariableValuesAsync has side effects
Today I faced one funny bug (side effect) in one of our services. The same validation error was listed 5 times:
изображение

This is "be design" of `ValidationContext.GetVariableValuesAsync`. This method parses inputs and coerces them into appropriate types. **Also** it reports validation errors via `ReportError` call (4 calls in ValidationContext.cs). OK? Yes... but this method is public and one of our custom validation rules calls it as well to get variables list. This rule does so inside `MathingNodeVisitor<GraphQLField>` so each time traversing algorithm visits some field `ValidationContext.GetVariableValuesAsync` method is called. In case of invalid variable **each** field in query yields new element in `ValidationContext.Errors` about the same problem. 5 fields -> 5 errors, 100 fields -> 100 errors. I think this is a design flaw. We should not add errors **inside** this method. Instead we should return them as a result along with variables. The only place where we use `GetVariableValuesAsync` is `DocumentValidator` just before returning `ValidationResult` to the called. The code easily may be rewritten to append returned collection of errors to existing errors in `context` variable:
изображение

But... it's a breaking change for `GetVariableValuesAsync`. Thoughts?

The current workaround for me now - make a clone of `ValidationContext` instance in our validation rule each time I want to call `GetVariableValuesAsync`.

sungam3r:
fixes #3418 

I just copied `GetVariableValuesAsync` body with little changes to return errors.

### DIFF
--- a/src/GraphQL.ApiTests/net5/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net5/GraphQL.approved.txt
@@ -3028,8 +3028,8 @@ namespace GraphQL.Validation
         public System.Collections.Generic.List<GraphQL.Validation.VariableUsage>? GetVariables<TNode>(TNode node)
             where TNode : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasSelectionSetNode { }
         [return: System.Runtime.CompilerServices.TupleElementNames(new string?[]?[] {
-                "variables",
-                "errors"})]
+                "Variables",
+                "Errors"})]
         public System.Threading.Tasks.ValueTask<System.ValueTuple<GraphQL.Validation.Variables, System.Collections.Generic.List<GraphQL.Validation.ValidationError>?>> GetVariablesValuesAsync(GraphQL.Validation.IVariableVisitor? visitor = null) { }
         public string? IsValidLiteralValue(GraphQL.Types.IGraphType type, GraphQLParser.AST.GraphQLValue valueAst) { }
         public void ReportError(GraphQL.Validation.ValidationError error) { }
--- a/src/GraphQL.ApiTests/net5/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net5/GraphQL.approved.txt
@@ -3022,9 +3022,15 @@ namespace GraphQL.Validation
         public System.Collections.Generic.List<GraphQL.Validation.VariableUsage>? GetRecursiveVariables(GraphQLParser.AST.GraphQLOperationDefinition operation) { }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentDefinition>? GetRecursivelyReferencedFragments(GraphQLParser.AST.GraphQLOperationDefinition operation) { }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentDefinition>? GetRecursivelyReferencedFragments(System.Collections.Generic.List<GraphQLParser.AST.GraphQLOperationDefinition> operations) { }
+        [System.Obsolete("This method can repeatedly add the same error into ValidationContext.Errors colle" +
+            "ction when called multiple times. Use GetVariablesValuesAsync method instead.")]
         public System.Threading.Tasks.ValueTask<GraphQL.Validation.Variables> GetVariableValuesAsync(GraphQL.Validation.IVariableVisitor? visitor = null) { }
         public System.Collections.Generic.List<GraphQL.Validation.VariableUsage>? GetVariables<TNode>(TNode node)
             where TNode : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasSelectionSetNode { }
+        [return: System.Runtime.CompilerServices.TupleElementNames(new string?[]?[] {
+                "variables",
+                "errors"})]
+        public System.Threading.Tasks.ValueTask<System.ValueTuple<GraphQL.Validation.Variables, System.Collections.Generic.List<GraphQL.Validation.ValidationError>?>> GetVariablesValuesAsync(GraphQL.Validation.IVariableVisitor? visitor = null) { }
         public string? IsValidLiteralValue(GraphQL.Types.IGraphType type, GraphQLParser.AST.GraphQLValue valueAst) { }
         public void ReportError(GraphQL.Validation.ValidationError error) { }
     }
--- a/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net6/GraphQL.approved.txt
@@ -3042,8 +3042,8 @@ namespace GraphQL.Validation
         public System.Collections.Generic.List<GraphQL.Validation.VariableUsage>? GetVariables<TNode>(TNode node)
             where TNode : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasSelectionSetNode { }
         [return: System.Runtime.CompilerServices.TupleElementNames(new string?[]?[] {
-                "variables",
-                "errors"})]
+                "Variables",
+                "Errors"})]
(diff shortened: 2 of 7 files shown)


src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -128,7 +128,10 @@ public List<GraphQLFragmentSpread> GetFragmentSpreads(GraphQLSelectionSet node)
 
         /// <summary>
         /// Returns all of the variable values defined for the operation from the attached <see cref="Variables"/> object.
+        /// Only correctly validated variables are returned. If the variable is specified incorrectly, then an instance of
+        /// <see cref="ValidationError"/> is added to the <see cref="Errors"/>.
         /// </summary>
+        [Obsolete("This method can repeatedly add the same error into ValidationContext.Errors collection when called multiple times. Use GetVariablesValuesAsync method instead.")]
         public async ValueTask<Variables> GetVariableValuesAsync(IVariableVisitor? visitor = null)
```
Shane32, line 135: Can this call the `GetVariablesValuesAsync` method?

sungam3r: yes)

Shane32:
I don't mind this PR as a temporary patch for 7.x, but I don't like this is as a long-term solution to the underlying issue.

sungam3r:
What do you suggest?

Actually I need a patch for v5.

src/GraphQL/Validation/ValidationContext.cs
```diff
@@ -128,16 +128,38 @@ public List<GraphQLFragmentSpread> GetFragmentSpreads(GraphQLSelectionSet node)
...
+
+            return variables;
+        }
+
+        /// <summary>
+        /// Returns all of the variable values defined for the operation from the attached <see cref="Variables"/> object.
+        /// Only correctly validated variables are returned. If the variable is specified incorrectly, then an instance of
+        /// <see cref="ValidationError"/> is returned within the list of errors.
+        /// </summary>
+        public async ValueTask<(Variables variables, List<ValidationError>? errors)> GetVariablesValuesAsync(IVariableVisitor? visitor = null)
```
Shane32, line 153:
```suggestion
        public async ValueTask<(Variables Variables, List<ValidationError>? Errors)> GetVariablesValuesAsync(IVariableVisitor? visitor = null)
```

src/GraphQL.Benchmarks/Benchmarks/DetailedBenchmark.cs
```diff
@@ -223,7 +223,7 @@ public Validation.Variables ParseVariables()
                 Schema = Schema,
                 Variables = Inputs ?? Inputs.Empty,
                 Operation = Operation,
-            }.GetVariableValuesAsync().Result;
+            }.GetVariablesValuesAsync().Result.variables;
```
Shane32, line 226:
```suggestion
            }.GetVariablesValuesAsync().Result.Variables;
```

src/GraphQL.ApiTests/net6/GraphQL.approved.txt
```diff
@@ -3036,9 +3036,15 @@ namespace GraphQL.Validation
...
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentDefinition>? GetRecursivelyReferencedFragments(GraphQLParser.AST.GraphQLOperationDefinition operation) { }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentDefinition>? GetRecursivelyReferencedFragments(System.Collections.Generic.List<GraphQLParser.AST.GraphQLOperationDefinition> operations) { }
+        [System.Obsolete("This method can repeatedly add the same error into ValidationContext.Errors colle" +
+            "ction when called multiple times. Use GetVariablesValuesAsync method instead.")]
         public System.Threading.Tasks.ValueTask<GraphQL.Validation.Variables> GetVariableValuesAsync(GraphQL.Validation.IVariableVisitor? visitor = null) { }
         public System.Collections.Generic.List<GraphQL.Validation.VariableUsage>? GetVariables<TNode>(TNode node)
             where TNode : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasSelectionSetNode { }
+        [return: System.Runtime.CompilerServices.TupleElementNames(new string?[]?[] {
+                "variables",
+                "errors"})]
```
Shane32, line 3046:
```suggestion
                "Variables",
                "Errors"})]
```

src/GraphQL.ApiTests/net5/GraphQL.approved.txt
```diff
@@ -3022,9 +3022,15 @@ namespace GraphQL.Validation
...
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentDefinition>? GetRecursivelyReferencedFragments(GraphQLParser.AST.GraphQLOperationDefinition operation) { }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLFragmentDefinition>? GetRecursivelyReferencedFragments(System.Collections.Generic.List<GraphQLParser.AST.GraphQLOperationDefinition> operations) { }
+        [System.Obsolete("This method can repeatedly add the same error into ValidationContext.Errors colle" +
+            "ction when called multiple times. Use GetVariablesValuesAsync method instead.")]
         public System.Threading.Tasks.ValueTask<GraphQL.Validation.Variables> GetVariableValuesAsync(GraphQL.Validation.IVariableVisitor? visitor = null) { }
         public System.Collections.Generic.List<GraphQL.Validation.VariableUsage>? GetVariables<TNode>(TNode node)
             where TNode : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasSelectionSetNode { }
+        [return: System.Runtime.CompilerServices.TupleElementNames(new string?[]?[] {
+                "variables",
+                "errors"})]
```
Shane32, line 3032:
```suggestion
                "Variables",
                "Errors"})]
```

Shane32 (changes requested):

Shane32:
> Actually I need a patch for v5.

We can backport this patch.

> What do you suggest?

Not sure offhand.  I would like to discuss it.  Here's one idea:

- New interface for validation rules that need to check variables.
- These rules run once standard validation rules have been run, and always run whether or not document caching is enabled.
- Field nodes are replaced with a derived class that also contains field arguments (using existing methods that read arguments and deserialize them, whether from variables or anything else).
- Then execution engine can use these field arguments instead of parsing them a second time.
- This has an added benefit of deserializing all argument literal values before any execution begins, which does not happen now, and could potentially throw an error, such as when passing an invalid date string to a DateTime field argument as a literal

Shane32:
Building on that basis, we could perhaps add validation methods to fields and scalars.  So:

1. A scalar can inspect directives defined on the field to determine whether the field passes validation; and
2. During code-first type initialization, a simple call can set a validation rule on a field

Ideally could be 'compiled' during schema initialization so that only the minimum of calls are made.

sungam3r:
It looks like a rather big redesign.

Shane32:
> It looks like a rather big redesign.

It's not so bad.  See #3424 for the first part which moves argument parsing to the document validator.  Next PR could add ways to hook into the validation.

Shane32 (approved):

sungam3r:
What is the process of patching v5.x ? I see 5.3.3 tag , origin/v533 branch, origin/v5 branch with merged (but not released) #3381 in it.

sungam3r:
I suppose I need to modify all workflows to support `v5` branch.

Shane32:
No. Just make a PR against the v5 branch. The one line to run the tests is already changed in that branch. No changes necessary to publish.

sungam3r:
Even `build-artifacts-code.yml` to generate
изображение
?

Shane32:
I think that's mainly used to publish beta builds.

PR #3424, opened 2022-11-27
Move argument parsing to document validator - part 1

Shane32:
Part 1: Move argument parsing from GetArgument to document validator, so errors with literals are caught proactively (see how a number of tests are "fixed")

Note: There are breaking changes to public but primarily-internal classes such as `ExecutionContext` and `IDocumentValidator`.  No changes to `IValidationRule` or similar that would break common libraries like Server, Authorization, Conventions, I'm guessing.

Speed and memory usage should be comparable.  AST tree does need to be iterated again for argument parsing.  Memory usage can actually decrease in certain scenarios.

- [ ] Part 2: Improve error reporting for literal values that fail parsing
- [ ] Part 3: Allow 3rd party validation rules to execute that include argument information


src/GraphQL.Tests/Execution/VariablesTests.cs
```diff
@@ -185,9 +185,8 @@ public void fail_on_incorrect_value()
               fieldWithObjectInput(input: ["foo", "bar", "baz"])
             }
             """;
-        var expected = "{ \"fieldWithObjectInput\": null }";
 
-        var result = AssertQueryWithErrors(query, expected, rules: Enumerable.Empty<IValidationRule>(), expectedErrorCount: 1, executed: true);
+        var result = AssertQueryWithErrors(query, null, rules: Enumerable.Empty<IValidationRule>(), expectedErrorCount: 1, executed: false);
         result.Errors[0].Message.ShouldBe("Error trying to resolve field 'fieldWithObjectInput'.");
         result.Errors[0].InnerException.Message.ShouldStartWith("Expected object value for 'TestInputObject', found not an object '[\"foo\", \"bar\", \"baz\"]'.");
```
Shane32, line 191: Tests like this we can see:
- Request is not executed if the argument literal cannot parse correctly
- Errors can now be improved upon (either in this PR or a follow-up PR) as they do not occur within a field resolver but rather during document validation

src/GraphQL.Tests/Validation/DefaultValuesOfCorrectTypeTests.cs
```diff
@@ -68,9 +68,6 @@ public void variables_with_invalid_default_values()
             _.Error(BadValueForDefaultArgMessage("a", "Int", "\"one\"", "Expected type 'Int', found \"one\"."), 2, 13);
             _.Error(BadValueForDefaultArgMessage("b", "String", "4", "Expected type 'String', found 4."), 3, 16);
             _.Error(BadValueForDefaultArgMessage("c", "ComplexInput", "\"notverycomplex\"", "Expected 'ComplexInput', found not an object."), 4, 22);
-            _.Error("Variable '$a' is invalid. Error coercing default value.", 2, 3);
-            _.Error("Variable '$b' is invalid. Error coercing default value.", 3, 3);
-            _.Error("Variable '$c' is invalid. Error coercing default value.", 4, 3);
```
Shane32, line 73: Also notice that we do not have redundant error messages as variables/literals do not attempt to be read if the document is invalid

src/GraphQL.Tests/Validation/OverlappingFieldsCanBeMergedTest.cs
```diff
@@ -255,7 +255,7 @@ public void Allows_different_args_where_no_conflict_is_possible_should_pass()
         const string query = """
             fragment conflictingArgs on Pet {
                 ... on Dog {
-                    name(surname: true)
+                    name(surname: "test")
```
Shane32, line 258: This test needed to be changed because `surname` is defined as a string, and is now validated along with the document before the field resolver hits.  The boolean value failed validation before.

src/GraphQL.Tests/Validation/VariablesAreInputTypesTests.cs
```diff
@@ -38,10 +38,6 @@ public void output_types_are_invalid()
                 message: VariablesAreInputTypesError.UndefinedVarMessage("c", "Pet"),
                 line: 1,
                 column: 40);
-            _.Error(
-               message: "Variable '$b' is invalid. No value provided for a non-null variable.",
-               line: 1,
-               column: 20);
```
Shane32, line 44: Again, this error is really inaccurate; variable $b was correctly and already reported as having an invalid type.  It should not report again as being not provided.

src/GraphQL/Execution/IExecutionContext.cs
```diff
@@ -78,6 +78,16 @@ public interface IExecutionContext : IProvideUserContext
...
 
+        /// <summary>
+        /// A dictionary of field argument values.
+        /// </summary>
+        IDictionary<GraphQLField, IDictionary<string, ArgumentValue>>? ArgumentValues { get; }
+
+        /// <summary>
+        /// A dictionary of directive argument values.
+        /// </summary>
+        IDictionary<GraphQLField, IDictionary<string, DirectiveInfo>>? DirectiveValues { get; }
```
Shane32, line 89: As of now, we provide means to retrieve directive from fields only, via `IResolveFieldContext`.  It would be an easy matter to change this dictionary to be ASTNode instead of GraphQLField and collective directives from anywhere in the AST, but (a) what's the point? and (b) where would the user access them from?

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -134,5 +156,131 @@ private async Task<(IValidationResult validationResult, Variables variables)> Va
...
+            }
+
+            protected override async ValueTask VisitFieldAsync(GraphQLField field, Context context)
+            {
+                var fieldType = field.Name.Value == context.Schema.TypeMetaFieldType.Name
+                    ? context.Schema.TypeMetaFieldType
+                    : context.Type?.Fields.Find(field.Name.Value);
+                if (fieldType == null)
+                    return;
+                if (field.Arguments?.Count > 0)
```
Shane32, line 217: If there were no arguments defined on the field within the query, GetArguments will use the precalculated dictionary of default field argument values, so a dictionary does not need to be allocated.

src/GraphQL/Types/Schema.cs
```diff
@@ -469,6 +469,7 @@ protected virtual void Validate()
             // Do not change the order of these validations.
             CoerceInputTypeDefaultValues();
             SchemaValidationVisitor.Instance.Run(this);
+            FieldTypeDefaultArgumentsVisitor.Instance.Run(this);
```
sungam3r, line 472: I would move `FieldTypeDefaultArgumentsVisitor` after `AppliedDirectivesValidationVisitor`.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
```diff
@@ -0,0 +1,29 @@
...
+        private FieldTypeDefaultArgumentsVisitor()
+        {
+        }
+
+        /// <inheritdoc/>
+        public override void VisitObjectFieldDefinition(FieldType field, IObjectGraphType type, ISchema schema)
+        {
+            if (field.Arguments?.Count > 0)
+            {
+                field.DefaultArguments = field.Arguments.ToDictionary(arg => arg.Name, arg => new ArgumentValue(arg.DefaultValue, ArgumentSource.FieldDefault));
```
sungam3r, line 25: Two questions.
1. override `VisitInterfaceFieldDefinition` as well?
2. Include only args where `arg.DefaultValue` is not null? As I understand now `field.DefaultArguments` will always contain all arguments regardless whether it has any default value or not.

sungam3r: Or if you do that intentionally then I would clarify what _Default_ Arguments means in that context.

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -71,7 +75,7 @@ public Task<(IValidationResult validationResult, Variables variables)> ValidateA
             return ValidateAsyncCoreAsync(context, options.Rules ?? CoreRules);
         }
 
-        private async Task<(IValidationResult validationResult, Variables variables)> ValidateAsyncCoreAsync(ValidationContext context, IEnumerable<IValidationRule> rules)
+        private async Task<(IValidationResult validationResult, Variables variables, IDictionary<GraphQLField, IDictionary<string, ArgumentValue>>? argumentValues, IDictionary<GraphQLField, IDictionary<string, DirectiveInfo>>? directiveValues)> ValidateAsyncCoreAsync(ValidationContext context, IEnumerable<IValidationRule> rules)
```
sungam3r, line 78: 4 elements in tuple, it's time to convert it into named class/structure

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -235,7 +237,7 @@ await listener.AfterExecutionAsync(context)
         /// <summary>
         /// Builds a <see cref="ExecutionContext"/> instance from the provided values.
         /// </summary>
-        protected virtual ExecutionContext BuildExecutionContext(ExecutionOptions options, GraphQLDocument document, GraphQLOperationDefinition operation, Variables variables, Metrics metrics)
+        protected virtual ExecutionContext BuildExecutionContext(ExecutionOptions options, GraphQLDocument document, GraphQLOperationDefinition operation, Variables variables, Metrics metrics, IDictionary<GraphQLField, IDictionary<string, ArgumentValue>>? argumentValues, IDictionary<GraphQLField, IDictionary<string, DirectiveInfo>>? directiveValues)
```
sungam3r, line 240: I would pack all arguments into struct and pass it with in modifier. Both variants binary backward incompatible but struct at least source backward compatible.

src/GraphQL/ResolveFieldContext/ReadonlyResolveFieldContext.cs
```diff
@@ -43,10 +43,10 @@ internal ReadonlyResolveFieldContext Reset(ExecutionNode? node, ExecutionContext
         }
 
         private IDictionary<string, ArgumentValue>? GetArguments()
-            => ExecutionHelper.GetArguments(_executionNode.FieldDefinition!.Arguments, _executionNode.Field!.Arguments, _executionContext.Variables);
+            => _executionContext.ArgumentValues?.TryGetValue(FieldAst, out var ret) ?? false ? ret : FieldDefinition.DefaultArguments;
```
sungam3r, line 46: `_arguments` and _directives` fields are not needed anymore. The same for both private methods.

src/GraphQL/Execution/IExecutionContext.cs
sungam3r: I would not be puzzled without obvious need.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
Shane32: > override VisitInterfaceFieldDefinition as well?

Good catch, but unnecessary in the end.  I wrote a test which proved that due to #3417 it will not be used.  I added it anyway though.

Shane32: > Include only args where `arg.DefaultValue` is not null? As I understand now `field.DefaultArguments` will always contain all arguments regardless whether it has any default value or not.

No; this would complicate the implementation with perhaps zero benefit.  Could be slower.  Default field arguments still need to be added to the dictionary, and cloning from the default argument values doesn't provide a material benefit, since the default value is already in native form.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: I made a change; let's review again before adding a struct.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
Shane32: > Or if you do that intentionally then I would clarify what _Default_ Arguments means in that context.

Can you help here with comments/naming/etc as you see fit?   As I think you understand, it is a dictionary of argument values with all values set to the field default value (`null` or otherwise), used to pass to field resolvers when no arguments are supplied to the field within the query document.  Note: non-null constraints are ignored for this purpose, although if there are any non-null field arguments, then (assuming validation has not been disabled) it should be impossible to need this `DefaultArguments` value.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: OK. I plan to make a full review tonight. Was busy with other stuff.

src/GraphQL/Validation/IValidationResult.cs
```diff
@@ -14,5 +17,20 @@ public interface IValidationResult
...
         /// </summary>
         ExecutionErrors Errors { get; }
+
+        /// <summary>
+        /// Returns the set of variables parsed from the inputs.
+        /// </summary>
+        Variables? Variables { get; }
+
+        /// <summary>
+        /// Returns a dictionary of fields with suppied arguments.
```
sungam3r, line 27:
```suggestion
        /// Returns a dictionary of fields with supplied arguments.
```

src/GraphQL/Validation/IValidationResult.cs
```diff
@@ -14,5 +17,20 @@ public interface IValidationResult
...
+        /// </summary>
+        IDictionary<GraphQLField, IDictionary<string, ArgumentValue>>? ArgumentValues { get; }
+
+        /// <summary>
+        /// Returns a dictionary of directives with suppied arguments.
```
sungam3r, line 32:
```suggestion
        /// Returns a dictionary of directives with supplied arguments.
```

src/GraphQL/Validation/ValidationResult.cs
```diff
@@ -9,23 +12,40 @@ public class ValidationResult : IValidationResult
         /// <param name="errors">Set of validation errors.</param>
         public ValidationResult(IEnumerable<ValidationError> errors)
         {
-            Errors.AddRange(errors);
+            if (errors.Any())
+            {
+                Errors = new ExecutionErrors();
```
sungam3r, line 17: `Errors` property was already initialized on line 39.

Shane32: Reworked ValidationResult

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -118,12 +118,36 @@ private async Task<(IValidationResult validationResult, Variables variables)> Va
                     await new BasicVisitor(visitors).VisitAsync(context.Document, new BasicVisitor.State(context)).ConfigureAwait(false);
                 }
 
+                if (context.HasErrors)
+                {
+                    return new ValidationResult(context.Errors);
```
sungam3r, line 123: Do you want fast return here even without reading variables? After this return the control will flow into `BuildExecutionContext` and `foreach (var listener in options.Listeners)` loop. That code can use variables in one or another way.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -245,7 +244,9 @@ protected virtual ExecutionContext BuildExecutionContext(ExecutionOptions option
                 UserContext = options.UserContext,
 
                 Operation = operation,
-                Variables = variables,
+                Variables = validationResult.Variables ?? Variables.None,
+                ArgumentValues = validationResult.ArgumentValues,
```
sungam3r, line 248: Design issue. Both `ArgumentValues` and `DirectiveValues` may be initialized here with some "null-object" like `Variables.None` or `Inputs.Empty`. All other collection/dictionary properties do so.

src/GraphQL/Validation/ValidationResult.cs
```diff
@@ -50,5 +67,14 @@ public sealed class SuccessfullyValidatedResult : IValidationResult
         /// Returns an empty list of execution errors.
         /// </summary>
         public ExecutionErrors Errors => EmptyExecutionErrors.Instance;
+
+        /// <inheritdoc/>
+        public Variables? Variables => null;
```
sungam3r, line 72: I doubt about these nulls. This class is public and now it can be used ONLY if `variables == null && argumentValues == null && directiveValues == null`. Existing code may break silently. I would make `SuccessfullyValidatedResult` internal.

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: I would prefer to leave as null when they have not been parsed.  If you agree, then they are nullable even when IsValid is true, and even if we set it to a null-like object such as Variables.None.  So then it seems there is little point to initializing to a null-like object.  Thoughts?

src/GraphQL/Validation/ValidationResult.cs
Shane32: > I would make `SuccessfullyValidatedResult` internal.

Agree

sungam3r: I mean the name of the class does not correspond to the contents. It is some sort of `SuccessfullyValidatedResultWithoutAnyVariablesAndArgumentsAndDirectives`.

src/GraphQL/Execution/DocumentExecuter.cs
sungam3r: > So then it seems there is little point to initializing to a null-like object.

Avoid null checks on client side? I'm fine to leave them as nullable if these properties are readonly by nature.  `OutputExtensions` and `Errors` (for example) are initialized with empty dictionary/collection because they get-only in `IExecutionContext` and caller should can change contents. If it is so then I would change type from `IDictionary` to `IReadonlyDictionary` to make explicit distinction beetween read-only properties and mutable properties of `IExecutionContext`.

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: I expected such an answer. Then documentation about variables should clear say that variables are not parsed because ... (your text from the comment above).

src/GraphQL/Execution/DocumentExecuter.cs
Shane32: We could do `IReadOnlyDictionary<GraphQLField, IDictionary<string, ArgumentValue>>` of course...

sungam3r: At least + place a general comment on `IExecutionContext` about distinction beetween read-only properties and mutable properties.

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
```diff
@@ -0,0 +1,135 @@
...
+                    context.Types.Push(type);
+                    await base.VisitInlineFragmentAsync(inlineFragment, context).ConfigureAwait(false);
+                    context.Types.Pop();
+                }
+            }
+        }
+
+        protected override async ValueTask VisitFragmentDefinitionAsync(GraphQLFragmentDefinition fragmentDefinition, Context context)
+        {
+            if (context.Schema.AllTypes[fragmentDefinition.TypeCondition.Type.Name.Value]?.GetNamedType() is IComplexGraphType type)
```
sungam3r, line 52: 🤔 Why all these `is IComplexGraphType` checks? Scalar fields **can** have arguments as well.

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
```diff
@@ -0,0 +1,135 @@
...
+
+public partial class DocumentValidator
+{
+    private class ParseArgumentVisitor : ASTVisitor<ParseArgumentVisitor.Context>
+    {
+        public static readonly ParseArgumentVisitor Instance = new();
+
+        private ParseArgumentVisitor() { }
+
+        protected override async ValueTask VisitOperationDefinitionAsync(GraphQLOperationDefinition operationDefinition, Context context)
```
sungam3r, line 16: I have a suspicion that this code should be called only for executed operation - that is returned from `DocumentExecuter.GetOperation`.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
sungam3r: >  it is necessary to provide a complete list of arguments to field resolvers, regardless of whether the argument contains a default value or is null.

To align behavior with `ExecutionHelper.GetArguments` that returns dictionary of all argument values?

Shane32: Yes (I was specifically thinking of `IResolveFieldContext.Arguments`, which just calls `ExecutionHelper.GetArguments`.)

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
Shane32: This is a visitor for a fragment definition.  [5.5.1.3](https://spec.graphql.org/June2018/#sec-Fragments-On-Composite-Types) says:

> The target type of fragment must have kind UNION, INTERFACE, or OBJECT.

We already have error checking for 5.5.1.3 (see `FragmentsOnCompositeTypes`), which will append the proper error to the response.  So here, if 5.5.1.3 was violated we just skip the fragment altogether -- but assuming it was not, it will always cast to an `IComplexGraphType`.

Please add a comment if you wish me to explain another piece of code.

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
Shane32: It will save cpu cycles for sure, but nor does it hurt.  While easy to do here, it is not quite so easy to selectively parse fragments that are in use.  I can write such logic, however.  Should I?

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
sungam3r: OK. I'm talking about all `is IComplexGraphType` checks in this class. I just chose the first check that came to hand.

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -245,7 +244,9 @@ protected virtual ExecutionContext BuildExecutionContext(ExecutionOptions option
                 UserContext = options.UserContext,
 
                 Operation = operation,
-                Variables = variables,
+                Variables = validationResult.Variables ?? Variables.None,
```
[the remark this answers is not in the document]
Shane32: Done

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
Shane32: Done

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -4,15 +4,16 @@ namespace GraphQL.Validation
 {
     /// <summary>
     /// Validates a document against a set of validation rules and returns a list of the errors found.
+    /// If the document passes validation, also returns the set of parsed variabes and argument values.
```
sungam3r, line 7:
```suggestion
    /// If the document passes validation, also returns the set of parsed variables and argument values.
```

src/GraphQL/Validation/IValidationResult.cs
```diff
@@ -1,7 +1,11 @@
+using GraphQL.Execution;
+using GraphQLParser.AST;
+
 namespace GraphQL.Validation
 {
     /// <summary>
     /// Contains a list of the validation errors found after validating a document against a set of validation rules.
+    /// If the document passes validation, this will also contain the set of parsed variabes and argument values.
```
Shane32, line 8:
```suggestion
    /// If the document passes validation, this will also contain the set of parsed variables and argument values.
```

Shane32:
@sungam3r I have made all the changes you requested a few weeks ago.  I also added code comments and revised xml comments.  No real urgency, but please review again when you have time.  I'd like to continue with part 2 and part 3 of this effort (when I have time).

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
```diff
@@ -0,0 +1,183 @@
...
+                    var arguments = ExecutionHelper.GetArguments(fieldType.Arguments, field.Arguments, context.Variables);
+                    if (arguments != null)
+                    {
+                        (context.ArgumentValues ??= new()).Add(field, arguments);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    // todo: report error properly
+                    context.ValidationContext.ReportError(new ValidationError($"Error trying to resolve field '{field.Name.Value}'.", ex));
```
Shane32, line 124: Note to self: Update this with a more appropriate error message.  Currently set this way because that was the error message that would have occurred previously.

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
```diff
@@ -0,0 +1,183 @@
...
+                    var directives = ExecutionHelper.GetDirectives(field, context.Variables, schema);
+                    if (directives != null)
+                    {
+                        (context.DirectiveValues ??= new()).Add(field, directives);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    // todo: report error properly
+                    context.ValidationContext.ReportError(new ValidationError($"Error trying to resolve field '{field.Name.Value}'.", ex));
```
Shane32, line 142: Ditto

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
sungam3r: It's the next on my todo/review list.

sungam3r: It's the next on my todo/review list.

sungam3r:
Will review tomorrow most probably.

src/GraphQL/Validation/IValidationResult.cs
```diff
@@ -1,7 +1,11 @@
+using GraphQL.Execution;
+using GraphQLParser.AST;
+
 namespace GraphQL.Validation
 {
     /// <summary>
     /// Contains a list of the validation errors found after validating a document against a set of validation rules.
+    /// If the document passes validation, this will also contain the set of parsed variables and argument values.
```
sungam3r, line 8:
```suggestion
    /// If the document passes validation, this will also contain the set of parsed variables and argument values for fields and applied directives.
```

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -4,15 +4,16 @@ namespace GraphQL.Validation
 {
     /// <summary>
     /// Validates a document against a set of validation rules and returns a list of the errors found.
+    /// If the document passes validation, also returns the set of parsed variables and argument values.
```
Shane32, line 7:
```suggestion
    /// If the document passes validation, also returns the set of parsed variables and argument values for fields and applied directives.
```

src/GraphQL/Validation/IValidationResult.cs
```diff
@@ -14,5 +18,26 @@ public interface IValidationResult
...
+        /// Returns the set of variables parsed from the inputs.
+        /// If the document did not pass validation, this value will be <see langword="null"/>.
+        /// If the document passed validation but contained no variables, this value will be <see cref="Variables.None"/>.
+        /// </summary>
+        Variables? Variables { get; }
+
+        /// <summary>
+        /// Returns a dictionary of fields, and for each field, a dictionary of arguments defined for the field with their values.
+        /// If the document did not pass validation, or if no field arguments were found, this value will be <see langword="null"/>.
+        /// Note that fields will not be present in this dictionary if they would only contain arguments with default values.
```
sungam3r, line 32:
```
 // if any arguments were supplied in the document for the field, load all defined arguments
            // for the field
            if (field.Arguments?.Count > 0)
```

Is this controlled by this piece of code? And the purpose is to return `FieldDefinition.DefaultArguments` here

> _executionContext.ArgumentValues?.TryGetValue(FieldAst, out var ret) ?? false ? ret : FieldDefinition.DefaultArguments;

without allocating additional dictionary. Right?

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
sungam3r: > Can you help here with comments/naming/etc as you see fit? As I think you understand, it is a dictionary of argument values with all values set to the field default value (null or otherwise), used to pass to field resolvers when no arguments are supplied to the field within the query document. 

Uhhh, naming is hard. I look at the code trying to feel the essence of this property and it seems to me more and more that it is more likely a sort of optimizarion. It does not carry an own semantic - just the same arguments as `field.Arguments` but grouped by name (selection of default values from arguments of course). 

By the way, it seems to be worth making it internal.

sungam3r: And **IF** it can be internal then I'll suggest to name it as `AllArgumentsDefaultValues`

sungam3r: `DefaultArguments` sounds ands feels like another concept. When I was reading the code for the first time, it confused me - all arguments vs only those that have defaults.

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
sungam3r: another pr?

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
sungam3r: another pr?

src/GraphQL/Validation/IValidationResult.cs
Shane32: I believe that's correct, preventing allocation of the dictionary and their entries as well.  So basically, we are optimizing for the idea that much of the time there are no field arguments specified, and simply the default set needs to be returned.  If I remember correctly, this is a performance improvement over the existing codebase.

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
Shane32: Yes; part 2, as noted in PR summary.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
```diff
@@ -0,0 +1,38 @@
+using GraphQL.Execution;
+using GraphQL.Types;
+
+namespace GraphQL.Utilities
+{
+    /// <summary>
+    /// Sets <see cref="FieldType.DefaultArguments"/> for each <see cref="FieldType"/>.
+    /// </summary>
+    public sealed class FieldTypeDefaultArgumentsVisitor : BaseSchemaNodeVisitor
```
sungam3r, line 9:
```suggestion
    internal sealed class FieldTypeDefaultArgumentsVisitor : BaseSchemaNodeVisitor
```

?

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -71,6 +72,13 @@ internal void SetName(string name, bool validate)
         /// <inheritdoc/>
         public QueryArguments? Arguments { get; set; }
 
+        /// <summary>
+        /// This property contains the argument values supplied to the field resolver if no arguments
+        /// to the field were supplied within the request.
+        /// Note that this value is automatically initialized during schema initialization.
```
sungam3r, line 78:
```suggestion
        /// This property contains the argument values supplied to the field resolver if no arguments
        /// to the field were supplied within the request. This property serves as an optimization in
        /// <see cref="ReadonlyResolveFieldContext.Arguments". So basically, we are optimizing for
        /// the idea that much of the time there are no field arguments specified, and simply the
        /// default set needs to be returned. 
        /// Note that this value is automatically initialized during schema initialization.
```

src/GraphQL/Validation/DocumentValidator.cs
```diff
@@ -131,9 +137,28 @@ private async Task<(IValidationResult validationResult, Variables variables)> Va
...
+                    return new ValidationResult(context.Errors) { Variables = variables };
+                }
+
+                // parse all field arguments
+                using var parseArgumentVisitorContext = new ParseArgumentVisitor.Context(context, variables);
+                await ParseArgumentVisitor.Instance.VisitAsync(context.Document, parseArgumentVisitorContext).ConfigureAwait(false);
+                var argumentValues = parseArgumentVisitorContext.ArgumentValues;
+                var directiveValues = parseArgumentVisitorContext.DirectiveValues;
+
+                // todo: execute validation rules that need to be able to read field arguments/directives
```
sungam3r, line 151: What rules? Existing rules, new rules?

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        protected override ValueTask VisitFragmentSpreadAsync(GraphQLFragmentSpread fragmentSpread, Context context)
+        {
+            // visit the fragment that this fragment spread references,
+            // unless we already visited this fragment spread
+            var fragmentName = fragmentSpread.FragmentName.Name.Value;
+            if ((context.VisitedFragments ??= new()).Add(fragmentName))
+            {
+                foreach (var definition in context.ValidationContext.Document.Definitions)
```
sungam3r, line 73:
```
var definition = context.ValidationContext.Document.FindFragmentDefinition(fragmentName);
if (definition != null) ...
```

src/GraphQL/Validation/DocumentValidator.cs
Shane32: User defined rules.

I was writing a rule recently which validates directive argument values; this is very difficult at present if it is going to support variables (which it should), but this should be easy.  Another sample might be a rule for complexity analysis, which may need to look at `first` or `last` arguments of connection fields.  Another sample might be if a validation rule were to validate, based on directives perhaps, the format of `ID` graph type values.  Although this could be in the scalar, sometimes the format of the string ID value changes in different parts of the schema, such as a `long` value in some parts and a `Guid` value in other parts.  This should be easy to implement as a custom validation rule.  And finally, our sample `InputFieldsAndArgumentsOfCorrectLength` rule is currently 100 lines of code and is a complex mess to correctly implement.  Even so, it does not run values through the scalar, which it should do.  It should be a very simple validation rule to write.

I've not fully thought through everything yet, but nothing can be done if argument parsing does not occur until the resolver executes.  Hence this PR.  In addition, scalar parsing failures should be known before execution, and with this PR that will now occur prior to execution, benefitting error responses as already seen in some of the tests.  So we have a twofold benefit potentially.

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -71,6 +72,16 @@ internal void SetName(string name, bool validate)
         /// <inheritdoc/>
         public QueryArguments? Arguments { get; set; }
 
+        /// <summary>
+        /// This property contains the argument values supplied to the field resolver if no arguments
+        /// to the field were supplied within the request. This property serves as an optimization in
+        /// <see cref="ReadonlyResolveFieldContext.Arguments". So basically, we are optimizing for
```
Shane32, line 78:
```suggestion
        /// <see cref="ReadonlyResolveFieldContext.Arguments"/>. So basically, we are optimizing for
```

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
```diff
@@ -0,0 +1,183 @@
...
+                {
+                    // the fragment name should be one of the definitions in the document, or the
+                    // document would not have passed validation
+                    if (definition is GraphQLFragmentDefinition fragment && fragment.FragmentName.Name.Value == fragmentName)
+                    {
+                        return VisitFragmentDefinitionAsync(fragment, context);
+                    }
+                }
```
Shane32, line 81:
```suggestion
                var definition = context.ValidationContext.Document.FindFragmentDefinition(fragmentName);
                if (definition != null)
                {
                    // the fragment name should be one of the definitions in the document, or the
                    // document would not have passed validation
                    return VisitFragmentDefinitionAsync(definition, context);
                }
```

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: That is, the rules of validation will say at what point they want to plug.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
Shane32: Well, without running `FieldTypeDefaultArgumentsVisitor` the schema would be improperly initialized and could not be used.  All other schema initialization visitors within `Schema.Validate()` are public, and `Validate` is `protected virtual`.  And besides, we allow custom `ISchema` implementations, and any custom implementation would also need to run the visitor.  I think we have to keep it public for now.

src/GraphQL/Validation/DocumentValidator.ParseArgumentVisitor.cs
```diff
@@ -0,0 +1,181 @@
...
+                catch (Exception ex)
+                {
+                    // todo: report error properly
+                    context.ValidationContext.ReportError(new ValidationError($"Error trying to resolve field '{field.Name.Value}'.", ex));
+                }
+            }
+            // if any directives were supplied in the document for the field, load all defined arguments
+            // for directvies on the field
```
sungam3r, line 126:
```suggestion
            // for directives on the field
```

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
Shane32: How about `DefaultArgumentValues`?

Shane32: Maybe have it `public` with `internal set;` ?

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: Like `IConfigureExecution.SortOrder` ?

Shane32: Yes, that's another point I have yet to consider.  Certainly I would think that validation rules that rely on argument/directive values should not be run if the values were unable to be parsed.

I was thinking we might make all rules required by GraphQL specifications run prior to any custom rules.  This would make it easier to write custom validation rules without having lots of error checks for invalid schema.

For example, I was writing a special visitor that needed to follow fragments which executed within a validation rule.  What if there is a missing fragment?  Or worse yet, a fragment cycle?  Normally this would be caught by the built-in rules.  But I had to write custom code to handle a fragment cycle or else it would have resulted in a stack overflow exception.

This needs to be weighed against any additional CPU time that may be required for this design.  Just an idea for now; I'll come up with a suggestion that we can review together.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
sungam3r: I feel some roughness in these requirements. Additional optimization has become not just optimization, but a necessary part of work with the schema. Nevertheless, I'm fine keeping it public.

src/GraphQL/Validation/DocumentValidator.cs
Shane32: Perhaps a sort order property.  If possible I'd like to retain binary compatibility with rules written for v7, which may preclude adding properties to `IValidationRule`.  So I'm not really sure yet.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
sungam3r: > DefaultArgumentValues

OK. I would not take this property into public API surface. IMO it's true implementation detail that we want to use to optimize allocations. We can do without this property at all. It serves like data cache.

Shane32: Perhaps `ArgumentsDefaultValues`?

What if we move the property underneath `QueryArguments` so it is less 'visible'?  Perhaps `QueryArguments.DefaultValues` or `DefaultValuesCollection`.

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: `SortPropertyAttribute` and the are no breaking changes :)

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
Shane32: Ok, I'm fine with it remaining `internal` until requested otherwise.

Well, after the rework, it would sorta be necessary unless using `ReadonlyResolveFieldContext` (directly or indirectly).  We surely will need to allow custom validation rules to pull the argument values dictionary, so if nothing else, we will need to add an extension method or something.

src/GraphQL/Validation/DocumentValidator.cs
Shane32: I like that idea... maybe all rules < 0 run before argument parsing, then all rules > 0 run after.  Now presently we run all rules together (simultaneously) so TypeInfo can be accessed by all rules.  So... maybe it's just a boolean attribute???  Maybe group by value in the attribute?  Needs more thought.

src/GraphQL/Utilities/Visitors/FieldTypeDefaultArgumentsVisitor.cs
sungam3r: > What if we move the property underneath QueryArguments so it is less 'visible'? Perhaps QueryArguments.DefaultValues or DefaultValuesCollection.

This changes nothing IMO. It would be more simple to keep it in `FieldType` as internal property.

Shane32: Ok done.

src/GraphQL/Validation/DocumentValidator.cs
sungam3r: TypeInfo and other shared stuff is an optimization detail. The core question is to give or not a possibility to order validation rules among each other and among different stages of validation - before parsing arguments and after.

sungam3r (approved):
This was a long one.

PR #3426, opened 2022-11-28
v5: New method to get variables values

sungam3r:
Backport of #3423


Shane32 (approved):
I would like to release 5.4 just prior to releasing 7.2 so that 7.2 shows up as the most recent release under Releases. 7.2 is waiting for #3417

Issue #3359, opened 2022-10-13
Custom JsonConverter for DateTime types not being called when serializing the response


mglace:
## Description

`JsonConverter<DateTime>` not working because DateTime values are returned as strings by the `ValueExecutionNode.ToValue()` method on line 46 of the ExecutionResultJsonConverter.

### Steps to reproduce

I created a new DateTimeJsonConverter as follows: 

```
public class DateTimeJsonConverter : JsonConverter<DateTime>
{
    public override bool CanConvert(Type typeToConvert) => typeToConvert == typeof(DateTime);

    public override DateTime Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
    {
        return DateTime.Parse(reader.GetString());
    }

    public override void Write(Utf8JsonWriter writer, DateTime value, JsonSerializerOptions options)
    {
        writer.WriteStringValue(value.ToString("MM-dd-yyyy'T'HH:mm:ss"));
    }
}
```

And I registered it with the GraphQLSerializer like so:

```
var options = new JsonSerializerOptions();

options.Converters.Add(new DateTimeJsonConverter());

var serializer = new GraphQLSerializer(options);
```

Using the Star Wars example, I added a DateTime property to the Droid class:

```
public class Droid
{
    public string Id { get; set; }
    public string Name { get; set; }
    public DateTime Timestamp { get; set; }
}

public class DroidType : ObjectGraphType<Droid>
{
    public DroidType()
    {
        Field(x => x.Id).Description("The Id of the Droid.");
        Field(x => x.Name).Description("The name of the Droid.");
        Field(x => x.Timestamp).Description("The timestamp the Droid was brought online.");
    }
}

public class StarWarsQuery : ObjectGraphType
{
    public StarWarsQuery()
    {
        Field<DroidType>("hero")
            .Resolve(context => new Droid { Id = "1", Name = "R2-D2", Timestamp = DateTime.UtcNow });
    }
}
```

### Expected result

When calling `await serializer.WriteAsync(stream, result);`, timestamp property on the DroidType should be formatted using the custom JsonConverter provided.

### Actual result

When I call `await serializer.WriteAsync(stream, result);`, it serializes my DroidType with no regard to my custom DateTimeJsonConverter.  Placing a breakpoint in the Write method of my converter confirms that it is not being called.  When I break on the CanConvert method, I see that the value of `typeToConvert` is System.String, not System.DateTime as expected.

### Further Investigation

[...]

PR #3445, opened 2022-12-07
Fix schema initialization exceptions

### LINKED ISSUE #3444: Initialization problems are not thrown when calling Initialize a second time
When calling Initialize and encountering a situation like two graph types with the same name, GraphQL.NET throws an appropriate exception and fails to load.  This is correct; however, calling Initialize again produces an error indicating that an encountered graph type was already initialized.  While technically true, this is very misleading and masks the cause of the issue.

There are obviously two solutions to this problem:

1. Rollback any changes to the graph types during the partial initialization; or
2. Cache the `Exception` and rethrow it upon subsequent calls to `Initialize()` for the same schema.

Option 1 is an issue because it's not just the initialization flag that is set during schema initialization.  Many `ResolvedType` fields are set, and determining which fields were changed would be a hassle.  And simply clearing the initialized flag would result in a similar exception, because the `ResolvedType` fields that were already set would reference obsolete copies of graph types during re-initialization.

For this reason, I'd opt for Option 2.  We would have to be sure we could preserve the stack trace when rethrowing the exception, but this should not be difficult.  We can probably just capture it within an `ExceptionDispatchInfo` instance.

Shane32:
Fixes #3444 

See reports of this issue:
- https://github.com/graphql-dotnet/graphql-dotnet/issues/3359#issuecomment-1314768150
- https://github.com/graphql-dotnet/graphql-dotnet/issues/3374
- https://github.com/graphql-dotnet/conventions/issues/234

Also cc @STRATZ-Ken who also experienced this issue

### DIFF
--- a/src/GraphQL.Tests/Bugs/BugReport.cs
+++ b/src/GraphQL.Tests/Bugs/BugReport.cs
@@ -1,0 +1,33 @@
+using GraphQL.Types;
+
+namespace GraphQL.Tests.Bugs;
+
+public class Bug3444
+{
+    [Fact]
+    public void Test_Rethrow_Schema_Initialization_Failure()
+    {
+        var queryType = new ObjectGraphType { Name = "Query" };
+        queryType.Field<TestOneType>("test1");
+        queryType.Field<TestTwoType>("test2");
+        var schema = new Schema() { Query = queryType };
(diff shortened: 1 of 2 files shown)


sungam3r (approved):
Nice fix.

PR #3456, opened 2022-12-16
Support IAsyncEnumerable for subscriptions

Shane32:


sungam3r:
Will review in coming days.

Shane32:
@sungam3r This PR has been outstanding for 2 weeks without comment.  I will merge soon if you have no comments.  I would particularly like your input on whether #3457 should be merged for the next minor bump along with this PR, or delayed until the next major bump.

sungam3r:
Just reviewed #3457. I think swithing to `Microsoft.Bcl.AsyncInterfaces` can be done even as patch, not minor.

Regarding review as is - sorry, little time. Every day there is some kind of work or household chores. For the rest, it remains a few hours after midnight. Tomorrow the last working day before the holidays and I think that I will spend it working with mail and reviewing code. I'm interested in AOT things. I have long wanted to see what it is.

Shane32:
PR #3457 merged into this PR

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
```diff
@@ -20,11 +20,47 @@ public DynamicScopedSourceStreamResolver(ISourceStreamResolver resolver)
...
         _resolverFunc = async context =>
         {
-            using var scope = context.RequestServicesOrThrow().CreateScope();
-            return await resolver.ResolveAsync(new ScopedResolveFieldContextAdapter<object>(context, scope.ServiceProvider)).ConfigureAwait(false);
+            var scope = context.RequestServicesOrThrow().CreateScope();
+            var scopedContext = new ScopedResolveFieldContextAdapter<object>(context, scope.ServiceProvider);
+            var observable = await resolver.ResolveAsync(scopedContext).ConfigureAwait(false)
+                ?? throw new InvalidOperationException("The source stream resolver returned null.");
+            // keep the service scope alive until the subscription has been disposed
+            return new ObservableMapper(observable, scope);
```
Shane32, line 28: The code in this method could be considered either a bug fix or feature, but typically is not necessary when the returned `IObservable` comes from a singleton service and is the same instance returned to other callers.  With `IAsyncEnumerable` / `yield` support, this feature becomes mandatory for `[Scoped]` to function as expected.

src/GraphQL/Resolvers/ObservableFromAsyncEnumerable.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// which returns <see cref="IAsyncEnumerable{T}"/>. Each execution will create a new
+    /// <see cref="ObservableFromAsyncEnumerable{T}"/> instance which can only be used once.
+    /// </summary>
+    public static Func<IResolveFieldContext, ValueTask<IObservable<object?>>> Create(Func<IResolveFieldContext, ValueTask<IAsyncEnumerable<T>>> func)
+    {
+        return async context =>
+        {
+            var cts = CancellationTokenSource.CreateLinkedTokenSource(context.CancellationToken);
+            var enumerable = await func(new OverrideCancellationContext(context, cts.Token)).ConfigureAwait(false);
+            return new ObservableFromAsyncEnumerable<T>(cts, enumerable);
```
Shane32, line 34: The underlying cancellation token is signaled if the HTTP connection is aborted.  However, the cts here may be signalled if the specific subscription is stopped within a websocket connection.  Hence a linked token is created and passed into the `IResolveFieldContext` which is used to populate the arguments of the delegate.

src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
```diff
@@ -55,15 +55,15 @@ public async void ScopedSubscriptionWorks()
         };
         var unscopedSubscriptionResolver = graphType.Fields.Find(nameof(TestClass.UnscopedAsyncSubscription))!.StreamResolver!;
         var scopedAsyncSubscriptionResolver = graphType.Fields.Find(nameof(TestClass.ScopedAsyncSubscription))!.StreamResolver!;
-        (await unscopedSubscriptionResolver.ResolveAsync(context).ConfigureAwait(false)).Subscribe(new SampleObserver("0 1"));
-        (await unscopedSubscriptionResolver.ResolveAsync(context).ConfigureAwait(false)).Subscribe(new SampleObserver("1 2"));
-        (await unscopedSubscriptionResolver.ResolveAsync(context).ConfigureAwait(false)).Subscribe(new SampleObserver("2 3"));
+        (await unscopedSubscriptionResolver.ResolveAsync(context).ConfigureAwait(false)).Subscribe(new SampleObserver("0 1")).Dispose();
```
sungam3r, line 58: Was it a necessity to work properly or just code style issue?

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
sungam3r: Firstly, I do not use subscriptions. Secondly, I use GraphType first approach. So... it's rather hard for me to look throught all these changes and track implementation details in mind, but I'll try.

If I understand correctly before these changes scope was ALWAYS destroyed immediately after creation of returned `IObservable` instance. Code of that observable bits **can** use service provider but it most cases this is root service provider so destroying (and even using) scoped provider does not matter. Then you added support for `IAsyncEnumerable` where scoped services **may** be used wider and in that case you lengthen lifetime of scope - the scope now will die immediately after subscription is disposed and not in `async context => {...}` delegate above. Right?

src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
Shane32: Necessary -- although in actual environment, the subscriptions would have been disposed.

src/GraphQL/Extensions/MethodInfoExtensions.cs
```diff
@@ -0,0 +1,14 @@
+#if !NET5_0_OR_GREATER
+
+using System.Reflection;
+
+namespace GraphQL;
+
+internal static class MethodInfoExtensions
```
sungam3r, line 7: 😄 I though this is some new optimized API, but it is just 
изображение

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
Shane32: That's essentially correct.

src/GraphQL/Extensions/MethodInfoExtensions.cs
sungam3r: Maybe remove target argument since all calls pass `null`?

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
sungam3r: Are subscriptions guaranteed to be destroyed?

Shane32: Yes.  If not it is a guaranteed memory leak.  Extreme care in the GraphQL.NET Server project was made to ensure proper disposal in any situation.

src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
sungam3r: Ah, I think this is because `Class1.DisposedCount.ShouldBe` checks. Now scope dies not in `ResolveAsync` but when calling `Dispose` on the returned subscription.

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
sungam3r: ok

src/GraphQL.MicrosoftDI.Tests/ScopedAttributeTests.cs
Shane32: Correct.

src/GraphQL/Extensions/MethodInfoExtensions.cs
Shane32: Ok done.

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
```diff
@@ -20,11 +20,47 @@ public DynamicScopedSourceStreamResolver(ISourceStreamResolver resolver)
     {
         _resolverFunc = async context =>
         {
-            using var scope = context.RequestServicesOrThrow().CreateScope();
-            return await resolver.ResolveAsync(new ScopedResolveFieldContextAdapter<object>(context, scope.ServiceProvider)).ConfigureAwait(false);
+            var scope = context.RequestServicesOrThrow().CreateScope();
```
sungam3r, line 23: This implementation is less stable in case of exceptions before newing `ObservableMapper`; `scope` variable will stay undisposed. I would place try/catch here before `new ObservableMapper(observable, scope);`

Shane32: Good catch

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
```diff
@@ -86,16 +109,48 @@ public SourceStreamMethodResolver(MethodInfo methodInfo, LambdaExpression instan
...
+                var innerType = bodyExpression.Type.GetGenericArguments()[0];
+                var method = _convertFromAsyncEnumerableMethodInfo.MakeGenericMethod(innerType);
+                var func = method.CreateDelegate<Func<Expression, ParameterExpression, Func<IResolveFieldContext, ValueTask<IObservable<object?>>>>>();
+                return func(bodyExpression, resolveFieldContextParameter);
+            }
 
             if (taskBodyExpression == null)
             {
-                throw new InvalidOperationException("Method must return a IObservable<T> or Task<IObservable<T>> where T is a reference type.");
+                throw new InvalidOperationException("Method must return a IObservable<T> or Task<IObservable<T>> where T is a reference type, or a IAsyncEnumerable<T> or Task<IAsyncEnumerable<T>>.");
```
sungam3r, line 123: \+ `ValueTask<IObservable<T>>` + `ValueTask<IAsyncEnumerable<T>>`

Shane32: I intentionally simplified the error message... do you think it should cover that also?

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
Shane32: Fixed

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
Shane32: Note that generally we encourage the use of synchronous or Task-based asynchronous calls throughout the GraphQL.NET API.  `ValueTask` support was added under the hood for performance since we are often wrapping synchronous calls with `ValueTask`, but we do not have builder methods and similar with `ValueTask`-based delegates.

So when I updated the auto-registering classes, I added support for `ValueTask` if someone were to use it, but I didn't mention `ValueTask` in the exception messages.  If you think I should do so, I can.

sungam3r, line 123:
```suggestion
                throw new InvalidOperationException("Method must return either: 1) a IObservable<T> or IAsyncEnumerable<T> or 2) Task<> or ValueTask<> containing types from 1. In all cases T must be a reference type.");
```

Something like that?

sungam3r: Better to mention ValueTask option in error messages IMO.

Shane32: Well, I'd say it looks a bit confusing.  And `IAsyncEnumerable<T>` does not require `T` to be a reference type.

sungam3r: Where reference constraint come from?

sungam3r: Where reference constraint come from?

Shane32: `IObservable<T>` is cast directly to `IObservable<object>`.  This kind of cast is only possible with reference types.  `IAsyncEnumerable<T>` performs a cast (including boxing if necessary) within the code that converts it directly to `IObservable<object>` -- specifically on ObservableFromAsyncEnumerable.cs:70 -- `observer.OnNext(z)` performs an implicit cast from T to object even if T is a value type.

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
```diff
@@ -61,10 +65,20 @@ public SourceStreamMethodResolver(MethodInfo methodInfo, LambdaExpression instan
                         taskBodyExpression = Expression.Call(_castFromTaskAsyncMethodInfo.MakeGenericMethod(innerType), bodyExpression);
                     }
                 }
+                // Task<IAsyncEnumerable<T>>
+                else if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(IAsyncEnumerable<>))
+                {
+                    var innerType = type.GetGenericArguments()[0];
+                    var method = _convertFromTaskAsyncEnumerableMethodInfo.MakeGenericMethod(innerType);
+                    var func = method.CreateDelegate<Func<Expression, ParameterExpression, Func<IResolveFieldContext, ValueTask<IObservable<object?>>>>>();
+                    return func(bodyExpression, resolveFieldContextParameter);
```
sungam3r, line 74: Thought about design. As you said `SourceStreamMethodResolver` was already complicated and now has become event more complicated. Thought each if-branch individually is simple. What I don't like here - a mix of return statements in individual (new) branches vs setting `taskBodyExpression` with returning result at the very end. I would switch to return in the each if-branch and remove `else`.  So only `throw new InvalidOperationException` would stay at the very end.

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
Shane32: Note that `ObservableFromAsyncEnumerable<T>` implements `IObservable<object?>`, not `IObservable<T>`.  This was written this way because GraphQL.NET requires `IObservable<object?>` and so I put the cast directly in the implementation which allows value types to be returned.

sungam3r: Ah, I see.

Shane32: I could write a wrapper for `IObservable<T>` value types so they are supported also.

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
Shane32: Ok done.

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
Shane32: > + `ValueTask<IObservable<T>>` + `ValueTask<IAsyncEnumerable<T>>`

I just did the most simple; add the two additional types to the error message.

sungam3r:
```c#
 private sealed class ObservableAdapter<T> : IObservable<object?>
        {
            private readonly IObservable<T> _observable;

            public ObservableAdapter(IObservable<T> observable)
            {
                _observable = observable;
            }

            public IDisposable Subscribe(IObserver<object?> observer) => _observable.Subscribe(new ObserverAdapter(observer));

            private sealed class ObserverAdapter : IObserver<T>
            {
                private readonly IObserver<object?> _observer;
                public ObserverAdapter(IObserver<object?> observer)
                {
                    _observer = observer;
                }
                public void OnCompleted() => _observer.OnCompleted();
                public void OnError(Exception error) => _observer.OnError(error);
                public void OnNext(T value) => _observer.OnNext(value);
            }
        }
```

sungam3r:
```c#
 private static async ValueTask<IObservable<object?>> CastFromValueTaskAsync<T>(ValueTask<IObservable<T>> task) where T : class
            => await task.ConfigureAwait(false);
        private static async ValueTask<IObservable<object?>> CastFromValueTask2Async<T>(ValueTask<IObservable<T>> task)
          => new ObservableAdapter<T>(await task.ConfigureAwait(false));
```

sungam3r: > But I'm not sure that's much better.

Agree. With removal of T constraint it will be even more better.

src/GraphQL/Resolvers/ObservableFromAsyncEnumerable.cs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private async void SubscribeInternalAsync(IObserver<object?> observer, IAsyncEnumerable<T> enumerable)
+    {
+        try
+        {
+            // enumerate the source and pass the items to the observer
+            await foreach (var z in enumerable.WithCancellation(_cancellationTokenSource.Token).ConfigureAwait(false))
+            {
+                observer.OnNext(z);
```
sungam3r, line 70:
```suggestion
                observer.OnNext(z); // note - boxing here in case of value types
```

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
Shane32: Thanks.  I'll do this is a subsequent PR.  It won't be hard but it will change enough that it will be easier to review separately.

sungam3r: OK

sungam3r: Should we somehow encourage to use ref-types instead of value-types for observables in error messages/docs ?

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
```diff
@@ -36,64 +36,113 @@ public SourceStreamMethodResolver(MethodInfo methodInfo, LambdaExpression instan
         /// </summary>
         protected virtual Func<IResolveFieldContext, ValueTask<IObservable<object?>>> BuildSourceStreamResolver(ParameterExpression resolveFieldContextParameter, Expression bodyExpression)
```
sungam3r, line 37: Looks better. I think after mentioned new PR with observable adapter all if-else blocks here could be flattened.

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
Shane32: No.  100% of GraphQL.NET boxes value types when returned from `IFieldResolver.ResolveAsync` and for storage within `ExecutionResult`.  Adding a wrapper here will not contribute any significant memory or CPU usage to processing time; it's a single allocation when subscribing only.  I also cannot find anything within the GraphQL spec that says that a scalar cannot be returned from a subscription rather than an object type.

sungam3r: OK. IMO in the new PR message(s) will be changed.

sungam3r:
Almost done, 7/9. I need to deal with some other work here but I think I will have time to complete review today.

src/GraphQL/Resolvers/SourceStreamMethodResolver.cs
Shane32: If you mean to remove the 'else' in each of the 'else if' statements, then I would not.  I'm not sure what else you could mean.  It's about as clean as it can get, and it has comments to make it extremely clear which code is matching on what types.  😕 

I could move more code into local functions, but each code block is only about 3-4 lines of code, and doing so might make it even harder to follow.

Let me know if there's something else you suggest I try.

sungam3r: I will show later, not in this PR.

src/GraphQL/Resolvers/ObservableFromAsyncEnumerable.cs
sungam3r: I prefer to place such words into code as comments.

src/GraphQL/Resolvers/ObservableFromAsyncEnumerable.cs
```diff
@@ -0,0 +1,140 @@
...
+        // this would only occur if IResolveFieldContext.CancellationToken has been signaled
+        _cancellationTokenSource.Token.ThrowIfCancellationRequested();
+
+        // iterate the async enumerable until the cancellation token is signaled via IDisposable.Dispose
+        SubscribeInternalAsync(observer, enumerable);
+
+        return this;
+    }
+
+    private async void SubscribeInternalAsync(IObserver<object?> observer, IAsyncEnumerable<T> enumerable)
```
sungam3r, line 63: `async void` - not a frequent guest...

Shane32: The usage was directly copied from a MS implementation within System.Reactive, so I kept it, although almost all the rest of code changed.

Shane32: System.Reactive.Linq, that is.  They have code to convert an IAsyncEnumerable to IObservable which is what i started with.

src/GraphQL.Tests/Types/AutoRegisteringObservableTests.cs
```diff
@@ -0,0 +1,365 @@
...
+        var observable = await streamResolver.ResolveAsync(context).ConfigureAwait(false);
+        observable.ToEnumerable().ShouldBe(new string[] { "ok2" });
+    }
+
+    [Theory]
+    [InlineData(nameof(TestClass.AsyncCancelToken1), true)]
+    [InlineData(nameof(TestClass.AsyncCancelToken1), false)]
+    [InlineData(nameof(TestClass.AsyncCancelToken2), true)]
+    [InlineData(nameof(TestClass.AsyncCancelToken2), false)]
+    public async Task AsyncEnumerable_Cancel1(string fieldName, bool cancelViaDispose)
```
sungam3r, line 117:
```suggestion
    public async Task AsyncEnumerable_Cancel(string fieldName, bool cancelViaDispose)
```

src/GraphQL/Resolvers/ObservableFromAsyncEnumerable.cs
Shane32:
```suggestion
            var cts = CancellationTokenSource.CreateLinkedTokenSource(context.CancellationToken);

            // The underlying cancellation token is signaled if the HTTP connection is aborted. However, the
            // cts here may be signalled if the specific subscription is stopped within a websocket
            // connection. Hence a linked token is created and passed into the IResolveFieldContext
            // which is used to populate the arguments of the delegate.

            var enumerable = await func(new OverrideCancellationContext(context, cts.Token)).ConfigureAwait(false);
            return new ObservableFromAsyncEnumerable<T>(cts, enumerable);
```

src/GraphQL/Resolvers/ObservableFromAsyncEnumerable.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// which returns <see cref="IAsyncEnumerable{T}"/>. Each execution will create a new
+    /// <see cref="ObservableFromAsyncEnumerable{T}"/> instance which can only be used once.
+    /// </summary>
+    public static Func<IResolveFieldContext, ValueTask<IObservable<object?>>> Create(Func<IResolveFieldContext, ValueTask<IAsyncEnumerable<T>>> func)
+    {
+        return async context =>
+        {
+            var cts = CancellationTokenSource.CreateLinkedTokenSource(context.CancellationToken);
+
+            // The underlying cancellation token is signaled if the HTTP connection is aborted. However, the
```
sungam3r, line 34: Strictly speaking we don't know transport details here, but it's OK to work in such terms.

sungam3r (approved):
It was not too difficult, although I should still work on understanding some places.

PR #3457, opened 2022-12-16
Add Microsoft.Bcl.AsyncInterfaces dependency for .NET Standard 2.0 projects

Shane32:
Enables IAsyncEnumerable subscription support for .NET Standard 2.0 projects for use with #3456 .

This PR targets #3456 currently.  If #3456 is merged then it will be updated to target master (7.3) or develop (8.0).

Besides the additional dependency, there are no breaking changes within this PR.

We can likely choose any version for the new `Microsoft.Bcl.AsyncInterfaces` dependency, as we only require the basic interface definitions.  We can also choose to remove the `System.Threading.Tasks.Extensions` dependency, as it is indirectly referenced by the `Microsoft.Bcl.AsyncInterfaces` dependency.  I left it in, just as Microsoft did with the `Microsoft.Extensions.DependencyInjection.Abstractions` 7.0.0 nuget package

### DIFF
--- a/src/GraphQL/GraphQL.csproj
+++ b/src/GraphQL/GraphQL.csproj
@@ -14,7 +14,6 @@
     <PackageReference Include="Nullability.Source" Version="2.1.0" PrivateAssets="all" />
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.3" Condition="'$(TargetFramework)' == 'netstandard2.1'" />
   </ItemGroup>
--- a/src/GraphQL/GraphQL.csproj
+++ b/src/GraphQL/GraphQL.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.1;netstandard2.0;net5;net6</TargetFrameworks>
@@ -15,6 +15,7 @@
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.3" Condition="'$(TargetFramework)' == 'netstandard2.1'" />
   </ItemGroup>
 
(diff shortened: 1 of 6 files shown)


src/GraphQL/GraphQL.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```
sungam3r, line 18: 🤔 1.1.1 ? Maybe 7.0.0 ? They all netstandard2.0 compatible.

sungam3r: 7.0.0
изображение


1.1.1
изображение

sungam3r:
I'm not a fan of dependencies so I would remove `System.Threading.Tasks.Extensions` from being listed.

sungam3r (approved):

src/GraphQL/GraphQL.csproj
Shane32: A couple thoughts on this:
1. I believe we are only using ** interfaces ** from this package; no code.  So it really makes no difference to us which version is referenced.
2. This reference is practically only required for .NET Framework or .NET Core 2.1 targets.  Version 1.1.1 is the last 1.x release, and 1.x would be in use for projects created in that era.  It's also the second-most popular (by downloads), where 1.0.0 is most popular.
3. Anyone can reference 7.0.0 or any newer version if they require it for their code.  Again, we just need the interfaces referenced.

src/GraphQL/GraphQL.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```
Shane32, line 17:
```suggestion
```

src/GraphQL/GraphQL.csproj
sungam3r: I understand.

>  So it really makes no difference to us which version is referenced. Anyone can reference 7.0.0 or any newer version if they require it for their code.

True. I just balance between the desire to use the minimum version as a library developer and maximum as its consumer, so as not to add packages on my own.

Shane32: When it matters, though, it really bites hard.  For instance, @SimonCropp keeps his libraries up to date with the latest versions of all dependencies always.  Most recently, MS released a **_completely_** broken copy of Microsoft.SqlClient (v5.0.1), which can crash servers within hours.  Since his LocalDb package references 5.0.1, my application pulls in the buggy version of SqlClient.  Fortunately it's only used in tests and does not matter there, but in a similar vein, the latest version of his GraphQL.EntityFramework package can only be used EF 7.0.1, even though the interfaces it uses are probably compatible as far back as EF Core 2 or 3.  I have a couple applications that run on .NET Framework, and so I'm stuck on EF Core 3 for certain server projects, precluding me from using GraphQL.EntityFramework, or any recent version anyway.

So my 'balance' here is an attempt to reference the same major version as would be done by other code for the same target framework.  For example, for a `net5.0` TFM, reference version `5.0.x` of `Microsoft.Extensions.Options`.  This way, the major frameworks likely to be used by the client application (ASP.NET Core, for example) are likely to be using the same major version of side packages as you are.  In this case, I felt that version 1.1.1 would apply the best.

Even better is to have no dependencies, which we have achieved nicely so far, but in this case we need one for the .NET Standard 2.0 TFM.

sungam3r: I completely understand and agree. I reason in the same way.

>  SimonCropp keeps his libraries up to date with the latest versions of all dependencies always. 

Yeah. I know :) I'm familiar with such an extreme and do not encourage that strategy.

Regarding `Bcl` package I just thought that it would not "bite". It looks rather "harmless". Nevertheless, I'm fine with `1.1.1`.

sungam3r: 😞 3 AM here but still did not review PRs about subscriptions / AOT, I'm done for today. Very-very-very good work in conventions repo.

PR #3471, opened 2022-12-31
Fix precision test

Shane32:
Test was only active for .NET Core 3.1 previously (contrary to comments), and it only fails due to a bad test.  Probably during deserialization of the expected JSON into an ExecutionResult prior to comparing it to the actual execution result.

It's also possible that our revised serializers have fixed any issues that previously existed.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
+++ b/src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
@@ -4,12 +4,8 @@ namespace GraphQL.Tests.Bugs;
 
 public class Bug138DecimalPrecisionTests : QueryTestBase<DecimalSchema>
 {
-#if NETCOREAPP3_1
     [Fact]
-#else
-    [Fact(Skip = "24.149999999999999 with .NET Core < 3.1")]
-#endif
-    public void double_to_decimal_does_not_lose_precision()
+    public async Task double_to_decimal_does_not_lose_precision()
     {
         var query = """
                 mutation SetState{
@@ -17,9 +13,16 @@ public class Bug138DecimalPrecisionTests : QueryTestBase<DecimalSchema>
                 }
             """;
 
-        var expected = """{ "set": 24.15 }""";
-
-        AssertQuerySuccess(query, expected);
+        var result = await new DocumentExecuter().ExecuteAsync(new()
+        {
+            Schema = Schema,
+            Query = query,
+        }).ConfigureAwait(false);
+        var expected = """{"data":{"set":24.15}}""";
+        var stjJson = new SystemTextJson.GraphQLSerializer().Serialize(result);
+        var nsjJson = new NewtonsoftJson.GraphQLSerializer().Serialize(result);
+        stjJson.ShouldBe(expected);
+        nsjJson.ShouldBe(expected);
     }
 }
 


sungam3r:
If I comment ifdefs and run this test for all TFMs then only NET48+STJ fails. All other combinations are green. On NET48 STJ serializes 24.15 decimal as 24.149999999999999 string.

Shane32:
It’s because the test deserializes the expected response into ExecutionResult incorrectly (part of the testing code). Has nothing to do with the parsing of the query or the serialization of the answer. That’s why my revision removes the buggy code from the test and only tests the GraphQL.NET library - which works for all TFMs

Shane32:
If you notice the failure says that “expected” is 24.1499999 - not “actual” is 24.149999 - indicating that the expected value changed from the constant entered into the test.

sungam3r:
The problem is indeed in `writer.Serialize((ExecutionResult)expectedExecutionResultOrJson);` where we have 24.15 double value that is serialized as 24.149999999999999. `runResult` is always serialized well since serializer works with execution nodes that have right type of values in their `Result` properties - `decimal` in out case. In case of `CreateQueryResult` from tests we have double value inside returned dictionary and STJ serializes it as 24.149999999999999. I tried to bump STJ version in GraphQL.SystemTextJson hoping it helps. Nope - no one version above 4.7.2 (from 5.0.0 to 7.0.1) helped, so I think the problem is in internals of .NET Framework runtime.

Shane32:
This doesn’t make sense. The test passes when you take AssertQuerySuccess out of the picture - on all TFMs!

sungam3r:
> This doesn’t make sense.

What exactly? I'm talking about that `CreateQueryResult + writer.Serialize` (of course from tests only) produces 24.15 on 3.1, 5.0, 6.0 and 7.0 but it produces 24.149999999 on NET48 for any STJ package version used. Whatever, see #3473.

src/GraphQL.Tests/Bugs/Bug138DecimalPrecisionTests.cs
```diff
@@ -4,22 +4,25 @@ namespace GraphQL.Tests.Bugs;
...
         var query = """
                 mutation SetState{
                     set(request:24.15)
                 }
             """;
 
-        var expected = """{ "set": 24.15 }""";
-
-        AssertQuerySuccess(query, expected);
+        var result = await new DocumentExecuter().ExecuteAsync(new()
```
sungam3r, line 16: #3473 allows to preserve all internal settings in executer and serializers and do not duplicate it here.

Shane32:
This is the minimum repo to create a failure on NET48:

```cs
JsonSerializer.Serialize(24.15).ShouldBe("24.15");
```

However, this test is a "decimal precision test".  The reason it's failing is because AssertQuerySuccess converts the result to a double within `InputsJsonConverter`.

sungam3r:
> The reason it's failing is because AssertQuerySuccess converts the result to a double within InputsJsonConverter.

Nope. I thought the same first. `AssertQuerySuccess` on all TFMs converts 24.15 to double. Just check it. The problem is on serialization part **from double to string**. net31,5,6,7 - all of them serialize 24.15 double to 24.15 string. But only NET48 serializes 24.15 double to 24.149999999999 string

Shane32:
> > The reason it's failing is because AssertQuerySuccess converts the result to a double within InputsJsonConverter.
> 
> Nope. I thought the same first. `AssertQuerySuccess` on all TFMs converts 24.15 to double. Just check it. The problem is on serialization part **from double to string**. net31,5,6,7 - all of them serialize 24.15 double to 24.15 string. But only NET48 serializes 24.15 double to 24.149999999999 string

Yes it is.  Do you want the stack trace?  Here's where AssertQuerySuccess converts the expected result via the InputsJsonConverter:

```
JsonStringExtensions.ToInputs(String json) line 42
JsonStringExtensions.ToDictionary(String json) line 31
JsonStringExtensions.ToExecutionResult(String json, ExecutionErrors errors, Boolean executed) line 19
QueryTestBase`2.CreateQueryResult(String result, ExecutionErrors errors, Boolean executed) line 227
QueryTestBase`2.AssertQuerySuccess(String query, String expected, Inputs variables, Object root, IDictionary`2 userContext, CancellationToken cancellationToken, IEnumerable`1 rules, INameConverter nameConverter) line 60
Bug138DecimalPrecisionTests.double_to_decimal_does_not_lose_precision() line 18
```

The InputsJsonConverter converts the value to a double, as I said.  Then it gets reserialized to a string, which STJ does incorrectly.

Shane32:
But the point is that it should not be converting to a double at all.  This is explicitly supposed to be a decimal test.

Shane32:
It's really questionable why it's getting deserialized and reserialized at all.  If just to reformat the string, it could convert to JsonElement and back.  If to normalize the data (in case of decimal for example), it should do so to both the expected and actual.

sungam3r:
> But the point is that it should not be converting to a double at all. 

True. Unforeseen fact.

> It's really questionable why it's getting deserialized and reserialized at all.

I think to normalize formatting (spaces, new lines, etc) + to introduce `data` property when serializing back to string (to not bother test author to write data wrapper in tests).

PR #3477, opened 2023-01-02
Reduce dependencies for .NET 5+

Shane32:
With this change, the only dependency for .NET 5+ is `GraphQL-Parser`.

### DIFF
--- a/src/GraphQL/GraphQL.csproj
+++ b/src/GraphQL/GraphQL.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.1;netstandard2.0;net5;net6</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


sungam3r (commented):
I like this PR. Less deps is better. I just doubt about first line. I have seen such changes for a long time. Do you have `autocrlf = true` in your gitconfig? I checked out raw csproj file by curl 
> curl https://raw.githubusercontent.com/graphql-dotnet/graphql-dotnet/master/src/GraphQL/GraphQL.csproj --output C:\Users\sunga\Downloads\1.txt

to be sure it contains only LF and so it is:
изображение

Shane32:
The file in `master` currently has the unicode BOM in front of it.

image

Since `charset = utf-8` and not `charset = utf-8-bom` for all files (`[*]`), there should be no BOM in the file

This PR correctly fixes the file to remove the BOM.

image

I'm not sure what keeps re-adding the BOM to the file.  Perhaps VS occasionally adds the BOM regardless of the editorconfig when making solution changes, like when using NuGet package manager to add references.

Note that since I'm on Windows, all end-of-line tags are converted to CRLF by git, and should remain as CRLF while on Windows.  I think that when pushing back to GitHub, git will convert them to LF-only.  This matches what is seen in my hex editor and your screenshot.

Shane32:
@sungam3r We could add a TFM for .NET Core 2.1 to remove the dependencies for .NET Core 2.1 and 3.1 also.  But although they are LTS releases, they are not currently supported, so I feel there's no need to do so at this point.

sungam3r:
> I'm not sure what keeps re-adding the BOM to the file.

Ah, BOM... Yes, known issue. It is VS when modifying csproj. I would not use netcoreapp2.1 TFM.

sungam3r (approved):

PR #3478, opened 2023-01-02
Add test to verify no BOM used

sungam3r:
Rel: #3477

I read a bunch of issues with complaints about VS behavior so here is a KISS solution.


src/GraphQL.ApiTests/BomTests.cs
```diff
@@ -0,0 +1,36 @@
+namespace GraphQL.ApiTests;
+
+public class BomTests
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/pull/3477
+    [Fact]
+    public void Files_Should_Not_Use_BOM()
+    {
+        var baseDir = new DirectoryInfo(AppDomain.CurrentDomain.BaseDirectory).Parent!.Parent!.Parent!.Parent!.Parent!;
```
Shane32, line 9: This kind of code makes for unreliable tests.  Perhaps it may build to a different folder with a different build script, or different operating system, or a new version of a SDK.  I can't approve this PR.

Besides, it is unnecessary.  "So what" if BOM is or is not there?

Shane32:
> I read a bunch of issues with complaints about VS behavior so here is a KISS solution.

I'm not complaining.  And this isn't a simple "solution".  It's a test with very specific assumptions.

src/GraphQL.ApiTests/BomTests.cs
sungam3r: I can change directory location algorithm in any appropriate way.

> "So what" if BOM is or is not there?

During the long time, I see that sometimes one or another file sometimes changes in PRs because of this. These are parasitic changes that should not be. It's like if we constantly change spaces count from PR to PR.

sungam3r: OK to use `GITHUB_WORKSPACE` from here https://docs.github.com/en/actions/learn-github-actions/environment-variables as root folder (if not null)?

sungam3r: `CallerFilePath` looks better

src/GraphQL.ApiTests/BomTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+public class BomTests
+{
+    // https://github.com/graphql-dotnet/graphql-dotnet/pull/3477
+    [Fact]
+    public void Files_Should_Not_Use_BOM()
+    {
+        string GetPath([CallerFilePath] string path = "") => path; // <GIT_ROOT>\src\GraphQL.ApiTests\BomTests.cs
+
+        var gitRoot = new DirectoryInfo(GetPath()).Parent!.Parent!.Parent!;
```
Shane32, line 13: Then before performing any other checks, we should ensure that the sln exists in this folder.

sungam3r: For what? There is no sln file in graphql root.

Shane32: We should just have some way of verifying we are reading the correct folder.

Shane32: Since the sln isn't in this folder, we should fix this so it isn't scanning the whole repo.  All the csproj and cs files are in the src folder right?  If that wasn't true, we should move the sln file.

sungam3r: I intentionally want to scan all repo files. Why to lock onto sln location? The "correctness" of  the whole path is guaranteed by Roslyn of course.

Shane32: I don't think this PR should be merged.  It serves no valuable purpose.  It makes tests more complicated and performs unnecessary scanning of files.  As it is now, it's scanning folders that do not need to be scanned, such as the documentation files, and if someone changes the folder tree, may scan files outside of the repo.

As far as I'm aware, we generally have only had issues with csproj files, and rarely so.  (Notice that for the test to pass only some csproj files needed to be updated.)  I don't see why we are expending ANY time and effort on this.  Certainly this is not a KISS solution.  Better to work on actual features rather than making our own lives more difficult with tests that do not affect functionality, such as forcing us to fix bom errors when they arise.  We already have enough problems with tests we created, such as now needing to add ifdefs to `using` statements when they are not needed by all TFMs.  While formatting tests may be good, there is overkill, and this is one such example.

There are additional reasons not to merge this.  What if there are files in excluded directories within the scanned folder?  What about stuff like Nullability.Source which is not part of the project but yet will bring in cs files?  And how are we expected to change files if the bom is wrong -- open a hex editor??

If you really want to merge this, confine it to the src tree with a verification that GraphQL.sln exists there.  I don't think it should be merged at all and will just create extra work for us to maintain the project.

sungam3r: >  It makes tests more complicated and performs unnecessary scanning of files.

😕  complicated? 14 lines of code and 0.3 sec for all repo scan. Nevertheless I can switch to scan src only, not a big deal.

Regarding the rest comments - I explained the reasoning before

>  These are parasitic changes that should not be. It's like if we constantly change spaces count from PR to PR.

Ideally dotnet-format tool should alert about BOM but it does not so I just filled a gap.

sungam3r: Last days I was on the road so had no much time to work on it or other pending reviews. Will continue tomorrow.

sungam3r: I thought one more time about your words about scanning and I think I have logical answer. Let's consider our editorconfig file:
```
[*]
insert_final_newline = true
charset = utf-8
```

1. It requires that ALL files should have utf-8 (no BOM) encoding without exceptions further down the file.
2. It requires this at the root repo level so it is intended to work even for docs, ci actions, whatever.

> confine it to the src tree

Then it would be logical to move editorconfig in `src` folder as well.

Shane32: All of my logic was explained in https://github.com/graphql-dotnet/graphql-dotnet/pull/3478#discussion_r1060301765 .  The editorconfig has no bearing on my reasoning.  I would not move the editorconfig.

I have no further comments.  As stated before, I think this test will simply add headache for the developers and should not be merged.

src/GraphQL.Harness/appsettings.Development.json
```diff
@@ -1,4 +1,4 @@
-﻿{
+{
```
sungam3r, line 1: ooops, BOM even here
https://stackoverflow.com/questions/4990095/json-specification-and-usage-of-bom-charset-encoding

Shane32: Don’t care. Works fine. Absolutely no effect on the functionality of this sample program, which isn’t even included in this library.

sungam3r:
I added the same sln check as in Auth repo.

Shane32 (approved):
ok

PR #3479, opened 2023-01-05
Fix nullable reference annotations for ValidationError constructors

Shane32:


Shane32:
That’s a breaking change. I am trying to prevent breaking validation rules’ API in v8.

sungam3r:
I know. Pretty harmless breaking change.

Shane32:
> I know. Pretty harmless breaking change.

Well, any custom validation rule likely relies on these constructors.  The server repo requires those constructors, and my own custom validation rules use those constructors as well.  I expect any custom validation rule would be fairly likely to use those constructors, as they are the only ones that allow indicating which node caused the failure.

So while I would say changing `DocumentValidator`'s constructors might be 'harmless', since people don't generally derive from `DocumentValidator`, in this case, I think we should make an effort to prevent breaking these constructors.

sungam3r (approved):

Issue #3480, opened 2023-01-05
EnsureTypeEquality enhancements


OpenSpacesAndPlaces:
Asks:
------------------
1. Would it be possible to get more verbose logging out of the terminal exception case?
2. Can AnyScalarGraphType be added to the EnsureTypeEquality check.

Details:
-------------------
Working on upgrading from 3.3 to 7.2 and I'm running into "A different instance of the GraphType" errors.
If I followed the code changes correctly, the old version would just overwrite anything that name matched with the last item in.
The part that is confusing me is that I'm already storing the types that have been created - so I'm not sure what difference it's finding.

Additionally, I think there was a second layer of protection related to:
https://github.com/graphql-dotnet/graphql-dotnet/issues/474

 My older code used:
 class JsonGraphType : ScalarGraphType
 
But the suggested change was:
AnyScalarGraphType

So the "// Ignore scalars" is also skipped.

```
    private void EnsureTypeEquality(IGraphType existingType, IGraphType newType)
        {
            if (ReferenceEquals(existingType, newType))
            {
                return;
            }

            // Ignore scalars
            if (existingType is ScalarGraphType && newType is ScalarGraphType)
            {
                return;
            }

            if (existingType.GetType() != newType.GetType())
            {
                throw new InvalidOperationException($"Unable to register GraphType '{newType.GetType().GetFriendlyName()}' with the name '{newType.Name}'. The name '{newType.Name}' is already registered to '{existingType.GetType().GetFriendlyName()}'. Check your schema configuration.");
            }

            // All other types are considered "potentially wrong" when being re-registered, throw detailed exception
            throw new InvalidOperationException($"A different instance of the GraphType '{newType.GetType().GetFriendlyName()}' with the name '{newType.Name}' has already been registered within the schema. Please use the same instance for all references within the schema, or use {nameof(GraphQLTypeReference)} to reference a type instantiated elsewhere.");
        }
```

Shane32:
Try GraphQL.NET 7.2.1 and see if there's any change.

Shane32:
or 7.2.2

sungam3r:
1. Of course it is possible but I don't understant the situation you want to log.
2. I think that it is a wrong way. Scalars were ignored intentionally.

Shane32:
If you’re still on 7.2.0 or prior 7.x version, very likely you are seeing the wrong error and the true error is being masked. Using 7.2.1 or newer will show the correct error.

[...]

PR #3488, opened 2023-01-06
Additional trace information in exception messages during schema initialization

sungam3r:
Helps to deal with #3480

### DIFF
--- a/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
+++ b/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
@@ -85,21 +85,7 @@ public class SchemaTypesTests
         // Test
         Should.Throw<InvalidOperationException>(() => schema.Initialize())
             .Message.ShouldBe(@"A different instance of the GraphType 'ObjectGraphType' with the name 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.
-Existing type trace:
-
-Loop over manually-added non-scalar types from AdditionalTypeInstances
-AddTypeIfNotRegistered(IGraphType, TypeCollectionContext) for type 'Query'
-Loop for fields of complex type 'Query'
-Field 'Query.first'
-AddTypeIfNotRegistered(IGraphType, TypeCollectionContext) for type 'MyObject'
-
-New type trace:
-
-Loop over manually-added non-scalar types from AdditionalTypeInstances
-AddTypeIfNotRegistered(IGraphType, TypeCollectionContext) for type 'Query'
-Loop for fields of complex type 'Query'
-Field 'Query.second'
-AddTypeIfNotRegistered(IGraphType, TypeCollectionContext) for type 'MyObject'");
+To view additional trace enable GlobalSwitches.TrackGraphTypeInitialization switch.");
     }
 
     [Fact]
--- a/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
+++ b/src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
@@ -84,7 +84,22 @@ public class SchemaTypesTests
 
         // Test
         Should.Throw<InvalidOperationException>(() => schema.Initialize())
-            .Message.ShouldBe("A different instance of the GraphType 'ObjectGraphType' with the name 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.");
+            .Message.ShouldBe(@"A different instance of the GraphType 'ObjectGraphType' with the name 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.
+Existing type trace:
+
+Loop over manually-added non-scalar types from AdditionalTypeInstances
+AddTypeIfNotRegistered(IGraphType, TypeCollectionContext) for type 'Query'
+Loop for fields of complex type 'Query'
+Field 'Query.first'
+AddTypeIfNotRegistered(IGraphType, TypeCollectionContext) for type 'MyObject'
+
+New type trace:
+
+Loop over manually-added non-scalar types from AdditionalTypeInstances
+AddTypeIfNotRegistered(IGraphType, TypeCollectionContext) for type 'Query'
+Loop for fields of complex type 'Query'
+Field 'Query.second'
+AddTypeIfNotRegistered(IGraphType, TypeCollectionContext) for type 'MyObject'");
     }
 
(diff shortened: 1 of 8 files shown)


src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -656,7 +694,7 @@ private IGraphType AddTypeIfNotRegistered(Type type, TypeCollectionContext conte
                 else if (_builtInCustomScalars.TryGetValue(namedType, out var builtInCustomScalar))
                 {
                     foundType = builtInCustomScalar;
-                    AddType(foundType, _context);
+                    AddType(foundType, context);
```
sungam3r, line 697: @Shane32 I changed this accidentally. Looks like a bug thought. 🤔

sungam3r: On the other side - `_context` is a field and this is the only usage outside of `Initialize` method.

Shane32, line 697: Why was this changed?  Bug?

src/GraphQL/GlobalSwitches.cs
```diff
@@ -65,6 +65,13 @@ public static class GlobalSwitches
     /// </summary>
     public static bool UseDeclaringTypeNames { get; set; } = false;
 
+    /// <summary>
+    /// Enable this switch to see additional debugging information in exception messages during schema initialization.
+    /// <br/>
+    /// By default disabled.
+    /// </summary>
+    public static bool TrackGraphTypeInitialization { get; set; } = true;
```
Shane32, line 73:
```suggestion
    public static bool TrackGraphTypeInitialization { get; set; } = false;
```

Shane32: I don't think all this extra trace activity and metadata logging should be enabled by default.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: Accidentally. Now I puzzle over whether `_context` was used here on purpose. Awkward SchemaTypes code. I did not remember.

src/GraphQL/GlobalSwitches.cs
sungam3r: I forgot to remote it 🤦‍♂️ .

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -629,6 +665,7 @@ private void AddTypeWithLoopCheck(IGraphType resolvedType, TypeCollectionContext
             context.InFlightRegisteredTypes.Push(namedType);
             try
             {
+                using var _ = context.Trace($"AddTypeWithLoopCheck for type '{namedType.Name}'");
```
Shane32, line 668: I would like to point out that these strings get created whether tracing is enabled or not.  Either we can leave it, which is fine, or we can add the name as a second argument, like so:

```cs
IDisposable Trace(string description, string? param = null)
{
    if (tracing is enabled)
        description = param == null ? description : $"{description} '{param}'";
    // other code
}

using var _ = context.Trace("AddTypeWithLoopCheck for type", namedType.Name);
```

Shane32: (obviously this is just an example)

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: "If it ain't broke, don't fix it".... unless you're sure.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: I know, can change it.

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
```diff
@@ -84,7 +84,8 @@ public void throws_exception_when_multiple_type_instances_exists()
 
         // Test
         Should.Throw<InvalidOperationException>(() => schema.Initialize())
-            .Message.ShouldBe("A different instance of the GraphType 'ObjectGraphType' with the name 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.");
+            .Message.ShouldBe(@"A different instance of the GraphType 'ObjectGraphType' with the name 'MyObject' has already been registered within the schema. Please use the same instance for all references within the schema, or use GraphQLTypeReference to reference a type instantiated elsewhere.
+To view additional trace enable GlobalSwitches.TrackGraphTypeInitialization switch.");
```
Shane32, line 88: :+1:

Setting global switches in tests is failure-prone without correct setup.  If you want to do so, be sure to apply `[Collection("StaticTests")]` to the class that contains the tests, so they do not run parallel to other tests.  (This is how we fixed the intermittent culture failures.)

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Yeah, I don't know if it was intentional or not.  As such, I personally wouldn't touch it, but I might add a comment.  I'll review the remainder of this PR and leave this change for you to decide.

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: If you plan to do so, then I'll review again after you do so.  Everything else looks fine.

sungam3r: done

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
sungam3r: I would keep switch disabled.

src/GraphQL/Types/TypeCollectionContext.cs
```diff
@@ -20,22 +20,61 @@ internal TypeCollectionContext(Func<Type, IGraphType> resolver, Action<string, I
...
+        {
+            return InitializationTrace == null
+                ? default
+                : new(this, string.Format(traceElement, arg1, arg2));
+        }
+    }
+
+    internal readonly struct TypeCollectionContextInitializationTrace : IDisposable
+    {
+        private readonly TypeCollectionContext _context;
```
Shane32, line 67:
```suggestion
        private readonly TypeCollectionContext? _context;
```

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: I reverted.

Shane32 (approved):

PR #3494, opened 2023-01-11
Fix issue with inherited record equals method being added to output

nicholashmoss:
Fix issue with inherited record equals method being added to output.

When auto-registering a record that inherits another record the inherited record's equals method was added to the output graph type.

This additionally created unwanted input graph types of the base record.


Shane32:
Looks great - thanks!!

PR #3461, opened 2022-12-23
Add support for AOT compilation

Shane32:
To demonstrate AOT compilation: within the `GraphQL.AotCompliationSample` folder, run `dotnet publish` and then run the executable within the `bin\Debug\net7.0\win-x64\publish` folder.

Note: this does not eliminate the many warnings you will receive when AOT-compiling (or more specifically, trimming).  However, it does provide the basic changes necessary that should generally make GraphQL.NET work with trimming.  Offhand, restrictions include:

- No use of `NameFieldResolver`
- Built-in connection types (GraphQL Relay types) are not supported, since they rely on `NameFieldResolver`
- No automatic CLR type mappings
- CLR type mappings in general have questionable support, since they rely on generics created at runtime
- Strongly recommend explicitly setting graph type on each field (e.g. `StringGraphType`)
- Strongly recommend explicitly setting resolver on each field with `.Resolve(...)`

Use of dependency injection is delicate, but seems to work.  May need to 'pin' some types so they can be constructed by the DI engine, particularly with options classes like `JsonSerializerOptions` or `ErrorInfoProviderOptions`.

To ensure any particular type is not trimmed during AOT, use this code:

```cs
// call Preserve with the type to prevent being trimmed
Preserve<MyType>();

static void Preserve<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T>()
    => GC.KeepAlive(typeof(T));
```

Any schema that is created with explicit references to graph type instances should work no problem.

I've also added an extremely basic test to ensure that the sample application runs correctly when trimmed and AOT-compiled on both Windows and Linux.

PR #3502, opened 2023-01-16
Introduce PolySharp

### LINKED ISSUE #3500: Investigate PolySharp package
https://github.com/Sergio0694/PolySharp

Rel : https://github.com/EasyNetQ/EasyNetQ.Management.Client/pull/213

As I see it allows to get rid of all our hacks with copying attributes.

sungam3r:
closes #3500


src/Directory.Build.props
```diff
@@ -25,6 +25,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="PolySharp" Version="1.10.0" Condition="'$(TargetFramework)' != 'net5' AND '$(TargetFramework)' != 'net6' AND '$(TargetFramework)' != 'net7'">
```
Shane32, line 28: Isn't the point here that PolySharp determines what polyfills are necessary based on which .NET version is being referenced?

```suggestion
    <PackageReference Include="PolySharp" Version="1.10.0">
```

Sergio0694: It does indeed. This condition should not be necessary.

src/Directory.Build.props
```diff
@@ -25,6 +25,10 @@
...
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
```
Sergio0694, line 30: Can simplify this a bit:
```suggestion
      <IncludeAssets>build; analyzers</IncludeAssets>
```

sungam3r: This is a default behavior when VS adds dev-only package. I would not change this.

src/Directory.Build.props
sungam3r: You may try. I tried it at first ang got compilation errors.

sungam3r: OK, I got the reason.

sungam3r:
@Shane32 I would not change in either way internal visibility for tests. Condition attribute looks like less evil.

Sergio0694:
Note, if you also want attributes such as `[DynamicallyAccessedMembers]` to be generated, to help make the code simpler for #3461, you'll need to also define `PolySharpIncludeRuntimeSupportedAttributes` to `"true"` in a `PropertyGroup` 🙂

PolySharp only generates language support attributes by default, not the dummy runtime support ones too.

Shane32:
Adding .NET 5.0 as an additional TFM for the MicrosoftDI and Dataloader projects solves the compilation issues so we can remove the conditions from the nuget package.  The issue has to do with the combination of having the InternalsVisibleTo attribute specified and the fact that some projects are targeting .NET Standard 2.0 while others are targeting .NET 5.0+ in the same build.

Shane32:
> Note, if you also want attributes such as `[DynamicallyAccessedMembers]` to be generated, to help make the code simpler for #3461, you'll need to also define `PolySharpIncludeRuntimeSupportedAttributes` to `"true"` in a `PropertyGroup` 🙂
> 
> PolySharp only generates language support attributes by default, not the dummy runtime support ones too.

Thanks for the tip @Sergio0694 !

Shane32:
@sungam3r Let's add this to `Tests.props` and remove the condition from the package reference:

```xml
<PackageReference Remove="PolySharp" />
```

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/pull/3502#issuecomment-1384376806

And that?

Shane32:
Your call - I suggest fixing it either by (1) adding .NET 5 as a TFM for those projects, or (2) excluding PolySharp from Tests.props.  I suggest the latter for 3 reasons:

1. I don't think we need the polyfills for the tests
2. We would be required to have .NET 5 as a TFM for every project even if a .NET 5 TFM was unnecessary - for example if we drop specific support for .NET 5 (allowing .NET Standard 2.0 fallback)
3. There may be other issues when `PolySharpIncludeRuntimeSupportedAttributes` is enabled

Shane32 (approved):

sungam3r:
I did not understand the issue with net5 magic fix but I would not add that TFM as well. I merge as is.

sungam3r:
OK. It's time to tune #3641 👍

PR #3503, opened 2023-01-16
Fix NRT metadata

Shane32:
The existing GraphQL.DataLoader and GraphQL.MicrosoftDI projects do not compile with a .NET 5 TFM because of NRT-related issues.  While `IServiceProvider.GetService` is obvious, the data loader issues are due to the underlying Dictionary having a `TKey` restriction of `notnull`.  For reference, calling `Dictionary<TKey, T>.Add` with `null` for a key [throws an exception](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.add?view=netstandard-2.0) as does [TryGetValue](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netstandard-2.0).  As such, this change will not affect working applications, but simply adds NRT information to the existing mechanics of the classes.


sungam3r:
Nice. 2 cents from me - we can try complile all published projects with the most modern TFM net7.0 just to test such features. Actual pack/publish process will not change, only one for tests.

sungam3r (approved):

sungam3r:
Something like `dotnet test -p:TargetFrameworks="netstandard2.0;net7.0"

PR #3509, opened 2023-01-18
Friendly error message when registering GraphType instead of ClrType

### LINKED ISSUE #3507: Nullable DateGraphType Argument Type Mapping Failing?
## Description

To preface this, I recently upgraded my solution from GraphQL v4 -> v7. This issue did not occur for me on v4. 

When using the type GraphQL.Types.DateGraphType as an argument and also passing nullable as true I get the following error.

The GraphQL type for argument 'DateMutations.updateDate.value' could not be derived implicitly. Could not find type mapping from CLR type 'GraphQL.Types.DateGraphType' to GraphType. Did you forget to register the type mapping with the 'ISchema.RegisterTypeMapping'?

### Steps to reproduce

If I write my example field like so I get the error:

```
            Field<BooleanGraphType>("updateDate")
                .Argument<DateGraphType>("newDate", true)
                .Argument<int>("id")
                .Resolve()
                .WithScope()
                .WithService<IMediator>()
                .ResolveAsync(UpdateDateResolver);
```

If I remove the nullable as true the schema generates successfully. 

```
            Field<BooleanGraphType>("updateDate")
                .Argument<DateGraphType>("newDate")
                .Argument<int>("id")
                .Resolve()
                .WithScope()
                .WithService<IMediator>()
                .ResolveAsync(UpdateDateResolver);
```

### Environment

.NET 7
GraphQL 7.2.2
GraphQL.DataLoader 7.2.2
GraphQL.MicrosoftDI 7.2.2
GraphQL.NewtonsoftJson 7.2.2
GraphQL.Server.All 7.2.0
GraphQL.Server.Transports.AspNetCore 7.2
GraphQL.Server.Ui.Playground 7.2.0

sungam3r:
fixes #3507


Shane32:
Before approving, I want to review adding the check in the `Field` and `Argument` methods so the exception occurs at the point it happens rather than during initialization.

Shane32:
I really don't think this is the correct way to solve this issue.  I think we need to add the check in `GetGraphTypeFromType` or somewhere so the exception occurs during `Field` or `Argument`.  We already made a similar restriction here:

image

There is no reason why users should be allowed to use graph types as clr types, and the exception should be triggered on the line of code it occurs on, not during schema initialization, when it can be helped.

sungam3r:
OK. Then the next thing I should do - list all public consumers of `GetGraphTypeFromType` (Field and Argument fluent methods are among them). `GetGraphTypeFromType` is public as well. Add check at each public level?

Shane32:
Well, the biggest potential for a problem is `Field<TClrType>(string name, bool nullable = false)` and the similar one for argument, because these it is very easy for a user to accidentally use the wrong overload without intending to do so.  All they would need to do is add a boolean to an otherwise correct bit of code.  So for those two specifically I would add an explicit exception within those methods so users know exactly what the resolution to the problem is.  For example, it should include helpful information within the exception message like `"You cannot specify a graph type with this overload.  Rather, please use the overload Field<TGraphType>(string name), using the NonNullGraphType<T> graph type if necessary"`.

All other uses can run through `GetGraphTypeFromType` with a more generic error message - perhaps along the lines of `new ArgumentOutOfRangeException("The supplied type is already a graph type.");`.  You can of course look up all other uses to see if there are any that are deemed likely to occur, but I do not think we need dedicated exception messages for every use case.  For example, it is very unlikely for `Field<TParameterType>(Expression<Func<TSourceType, TParameterType>>)` to be improperly called.

In either case, throwing the exception immediately allows for a stack trace to easily show exactly where the problematic use occurred and hence how to fix it.

Shane32:
@sungam3r Let me know if you would like me to post a PR with my suggestion.

sungam3r:
Go on if you have time. I will not touch this so far.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -791,7 +791,12 @@ private object GetGraphType(Type clrType, bool input, IEnumerable<IGraphTypeMapp
             var ret = GetGraphTypeFromClrType(clrType, input, typeMappings);
```
Shane32, line 791: Exception will be thrown here for `IGraphType` CLR types; additional code in this method not required.

src/GraphQL/Types/Schema.cs
```diff
@@ -366,7 +366,14 @@ public void RegisterTypes(params Type[] types)
         /// <inheritdoc/>
         public void RegisterTypeMapping(Type clrType, Type graphType)
         {
-            (_clrToGraphTypeMappings ??= new()).Add((clrType ?? throw new ArgumentNullException(nameof(clrType)), graphType ?? throw new ArgumentNullException(nameof(graphType))));
+            (_clrToGraphTypeMappings ??= new()).Add((CheckType(clrType ?? throw new ArgumentNullException(nameof(clrType))), graphType ?? throw new ArgumentNullException(nameof(graphType))));
+
+            Type CheckType(Type clrType)
+            {
+                return typeof(IGraphType).IsAssignableFrom(clrType)
+                    ? throw new ArgumentException($"{clrType.FullName}' is already a GraphType (i.e. not CLR type like System.DateTime or System.String). You must specify CLR type instead of GraphType.", nameof(clrType))
```
Shane32, line 374: Fine with this check, but should be `ArgumentOutOfRangeException`.

```suggestion
                    ? throw new ArgumentOutOfRangeException(nameof(clrType), $"{clrType.FullName}' is already a GraphType (i.e. not CLR type like System.DateTime or System.String). You must specify CLR type instead of GraphType.", nameof(clrType))
```

Shane32: Should we add a check to be sure the graph type is indeed a graph type?

src/GraphQL/Types/Collections/SchemaTypes.cs
Shane32: Sorry, thinking of wrong method.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -791,7 +791,12 @@ private object GetGraphType(Type clrType, bool input, IEnumerable<IGraphTypeMapp
...
 
             if (ret == null)
-                return $"Could not find type mapping from CLR type '{clrType.FullName}' to GraphType. Did you forget to register the type mapping with the '{nameof(ISchema)}.{nameof(ISchema.RegisterTypeMapping)}'?";
+            {
+                string additionalMessage = typeof(IGraphType).IsAssignableFrom(clrType)
+                    ? $" Note that '{clrType.FullName}' is already a GraphType (i.e. not CLR type like System.DateTime or System.String). Most likely you need to specify corresponding CLR type instead of GraphType."
+                    : "";
+                return $"Could not find type mapping from CLR type '{clrType.FullName}' to GraphType. Did you forget to register the type mapping with the '{nameof(ISchema)}.{nameof(ISchema.RegisterTypeMapping)}'?{additionalMessage}";
+            }
```
Shane32, line 799: This changes are fine, although hopefully are redundant.

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: When I wrote the code, I also got confused in them.

src/GraphQL/Types/Schema.cs
sungam3r: I remoced second `nameof(clrType)`

> to be sure the graph type is indeed a graph type?

изображение

?

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: _hopefully_

src/GraphQL/Types/Schema.cs
Shane32: I mean a verification that `graphType` IS assignable from `IGraphType` within `RegisterTypeMapping`.

Shane32: > I removed second `nameof(clrType)`

:+1:

sungam3r: I added second check. By the way yesterday I configured GPG signing so all my commits now are signed
изображение
I thought it was time to do this, because I have to work not only with this project, but also others.

Shane32 (approved):

PR #3517, opened 2023-01-29
Update sample application for AOT type-first

Shane32:
This change eliminates the use of `Preserve` in the type-first AOT compilation sample, and removes old comments that are not applicable to the sample.


Shane32:
Nice that we have tests for AOT compilation.

sungam3r (approved):

PR #3518, opened 2023-01-30
Graph types cannot be used as CLR types

Shane32:
Prevents using graph types as CLR types for `GetGraphTypeFromType` and indirectly `FieldBuilder` and `ArgumentBuilder`

Fixes:
- #3507

Replaces:
- #3509 

This change will cause an exception to be thrown at the exact line of code in error, rather than during schema validation.

### DIFF
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -518,11 +518,11 @@ public class ComplexGraphTypeTests
         type.Fields.Find("field3").ShouldNotBeNull().Type.ShouldBe(typeof(GraphQLClrOutputTypeReference<int>));
 
         var e1 = Should.Throw<ArgumentException>(() => type.Field<int?>("field4"));
-        e1.Message.ShouldBe("The GraphQL type for field 'TestObject.field4' could not be derived implicitly from type 'Nullable`1'.");
+        e1.Message.ShouldStartWith("The GraphQL type for field 'TestObject.field4' could not be derived implicitly from type 'Nullable`1'. Explicitly nullable type: Nullable<Int32> cannot be coerced to a non nullable GraphQL type.");
         e1.InnerException.ShouldNotBeNull().Message.ShouldStartWith("Explicitly nullable type: Nullable<Int32> cannot be coerced to a non nullable GraphQL type.");
 
         var e2 = Should.Throw<ArgumentException>(() => type.Field<int?>("field5", false));
-        e2.Message.ShouldBe("The GraphQL type for field 'TestObject.field5' could not be derived implicitly from type 'Nullable`1'.");
+        e2.Message.ShouldStartWith("The GraphQL type for field 'TestObject.field5' could not be derived implicitly from type 'Nullable`1'. Explicitly nullable type: Nullable<Int32> cannot be coerced to a non nullable GraphQL type.");
         e2.InnerException.ShouldNotBeNull().Message.ShouldStartWith("Explicitly nullable type: Nullable<Int32> cannot be coerced to a non nullable GraphQL type.");
 
         type.Field<int?>("field6", true).Resolve(_ => 3);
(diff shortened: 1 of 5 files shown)


src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -447,7 +447,7 @@ public virtual FieldType AddField(FieldType fieldType)
             }
             catch (ArgumentOutOfRangeException exp)
             {
-                throw new ArgumentException($"The GraphQL type for field '{Name ?? GetType().Name}.{name}' could not be derived implicitly from type '{typeof(TReturnType).Name}'.", exp);
+                throw new ArgumentException($"The GraphQL type for field '{Name ?? GetType().Name}.{name}' could not be derived implicitly from type '{typeof(TReturnType).Name}'. " + exp.Message, exp);
```
sungam3r, line 450: Are you sure it's a good practice to append inner exception message here? Generally I try to avoid this.

Shane32: In this case, the inner exception messages are always generated by our own code and explain the cause of the problem.  See tests which demonstrate how the exception messages are now more detailed.

Shane32: Notice also, we are not throwing an `ExecutionError` that may be passed to clients.  This is just rethrowing the exception with additional detail specifying which field/argument and type caused the problem.

Shane32: Basically, these errors will always be caught during development time (or graph initialization would fail), and the more detailed the exception message, the better.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -518,11 +518,11 @@ public void create_fieldbuilder_with_inferred_graph_type()
         type.Fields.Find("field3").ShouldNotBeNull().Type.ShouldBe(typeof(GraphQLClrOutputTypeReference<int>));
 
         var e1 = Should.Throw<ArgumentException>(() => type.Field<int?>("field4"));
-        e1.Message.ShouldBe("The GraphQL type for field 'TestObject.field4' could not be derived implicitly from type 'Nullable`1'.");
+        e1.Message.ShouldStartWith("The GraphQL type for field 'TestObject.field4' could not be derived implicitly from type 'Nullable`1'. Explicitly nullable type: Nullable<Int32> cannot be coerced to a non nullable GraphQL type.");
```
Shane32, line 521: More detailed message available.

src/GraphQL/Types/Composite/ComplexGraphType.cs
sungam3r: I see that. I mean that exception's information is "duplicated" across parent exception and inner exception. Not a big deal for me.

src/GraphQL.Tests/Utilities/GetGraphTypeFromTypeTests.cs
```diff
@@ -377,6 +391,23 @@ public void InputTypeIsDereferenced_InputField(Type referenceType, Type mappedTy
...
+    public void GraphTypesAsClrTypesAreDisallowed()
+    {
+        var expectedMessage = "The graph type 'IntGraphType' cannot be used as a CLR type.";
+
+        Should.Throw<ArgumentOutOfRangeException>(() => typeof(IntGraphType).GetGraphTypeFromType())
+            .Message.ShouldStartWith(expectedMessage);
+
+        var graphType = new ObjectGraphType();
+        Should.Throw<ArgumentException>(() => graphType.Field<IntGraphType>("test", true))
+            .Message.ShouldStartWith("The GraphQL type for field 'Object.test' could not be derived implicitly from type 'IntGraphType'. " + expectedMessage);
```
Shane32, line 404: More detailed message available now.

> The GraphQL type for field 'Object.test' could not be derived implicitly from type 'IntGraphType'. The graph type 'IntGraphType' cannot be used as a CLR type.

src/GraphQL/Types/Composite/ComplexGraphType.cs
Shane32: True, but here I think it is warranted.

Shane32: People do not always realize there is an inner exception.

sungam3r:
I have looked through this, OK, good. But what about merging this PR as not replacement of #3509 but as addition to it? I mean checks in #3509 make sense too. For example, `public void RegisterTypeMapping(Type clrType, Type graphType)` accepts **clrType** and added check is top-most one to catch such invalid usage.

sungam3r:
The same for `private object GetGraphType(Type clrType, bool input, IEnumerable<IGraphTypeMappingProvider>? typeMappings)`. I think about it as a protection on the last mile.

src/GraphQL/Types/Composite/ComplexGraphType.cs
sungam3r: ok

Shane32:
Makes sense for methods not covered by this change

sungam3r:
Then merge it first and I'll look at test results in my branch.

PR #3519, opened 2023-02-02
More AOT annotations

sungam3r:
I tried to follow AOT annotations guidelines and I have a bad feeling about this all.

Problems:
1. It is very toxic for code - even worse then async/await "infection" or NRT annotations.
2. I do not understand why we should manually write annotations - the analyzer already knows what is missing in specific cases and demands to put it into code. It looks strange. I hope that in NET8 this will be fixed and it will not be necessary to manually do the work. Though it may be the same story as for NRT - analizer knows but it's a developer responsibility to say the last word.
3. I do not understant what to do in 50% of cases when `Type` comes from some calculated variable.
4. It seems to me that we "excessive" demand trimmer to preserve some type's parts in methods with branch logic (for example, type.GetProperties() in the first branch and type.GetMethods() in the second).

I'm skeptical about such code annotation.

### DIFF
--- a/src/GraphQL/Extensions/AssemblyExtensions.cs
+++ b/src/GraphQL/Extensions/AssemblyExtensions.cs
@@ -22,6 +22,7 @@ namespace GraphQL
         /// Skips classes where the source type is <see cref="object"/>, or where the class is marked with
         /// the <see cref="DoNotMapClrTypeAttribute"/>.
         /// </summary>
+        [RequiresUnreferencedCode("Types containing in the provided assembly might be removed")]
         public static List<(Type ClrType, Type GraphType)> GetClrTypeMappings(this Assembly assembly)
         {
             //create a list of type mappings
--- a/src/GraphQL/Extensions/ObjectExtensions.cs
+++ b/src/GraphQL/Extensions/ObjectExtensions.cs
@@ -38,7 +38,10 @@ namespace GraphQL
         /// In case of configuring field as Field(x => x.FName).Name("FirstName") source dictionary
         /// will have 'FirstName' key but its value should be set to 'FName' property of created object.
         /// </param>
-        public static object ToObject(this IDictionary<string, object?> source, Type type, IGraphType? mappedType = null)
+        public static object ToObject(
+            this IDictionary<string, object?> source,
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.NonPublicProperties | DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields)] Type type,
+            IGraphType? mappedType = null)
         {
             // Given Field(x => x.FName).Name("FirstName") and key == "FirstName" returns "FName"
             string GetPropertyName(string key, out FieldType? field)
@@ -205,7 +208,14 @@ namespace GraphQL
         /// will have 'FirstName' key but its value should be set to 'FName' property of created object.
         /// </param>
         /// <remarks>There is special handling for strings, IEnumerable&lt;T&gt;, Nullable&lt;T&gt;, and Enum.</remarks>
-        public static object? GetPropertyValue(this object? propertyValue, Type fieldType, IGraphType? mappedType = null)
+        public static object? GetPropertyValue(
+            this object? propertyValue,
+#if NET6_0_OR_GREATER
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor | DynamicallyAccessedMemberTypes.Interfaces)] Type fieldType,
+#else
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] Type fieldType,
+#endif
+            IGraphType? mappedType = null)
         {
             // Short-circuit conversion if the property value already of the right type
             if (propertyValue == null || fieldType == typeof(object) || fieldType.IsInstanceOfType(propertyValue))
--- a/src/GraphQL/Extensions/SchemaExtensions.cs
+++ b/src/GraphQL/Extensions/SchemaExtensions.cs
@@ -128,7 +128,7 @@ namespace GraphQL
         /// <param name="schema">The schema for which the mapping is registered.</param>
         /// <typeparam name="TClrType">The CLR property type from which to infer the GraphType.</typeparam>
         /// <param name="mode">Which registering mode to use - input only, output only or both.</param>
-        public static void AutoRegister<TClrType>(this ISchema schema, AutoRegisteringMode mode = AutoRegisteringMode.Both)
+        public static void AutoRegister<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicMethods)] TClrType>(this ISchema schema, AutoRegisteringMode mode = AutoRegisteringMode.Both)
         {
             if (mode.HasFlag(AutoRegisteringMode.Output))
                 schema.RegisterTypeMapping<TClrType, AutoRegisteringObjectGraphType<TClrType>>();
--- a/src/GraphQL/Resolvers/NameFieldResolver.cs
+++ b/src/GraphQL/Resolvers/NameFieldResolver.cs
@@ -60,7 +60,9 @@ namespace GraphQL.Resolvers
         /// <param name="target"> The type from which you want to get the value. </param>
         /// <param name="name"> Property/method name. </param>
         /// <returns> Compiled field resolver. </returns>
-        private static IFieldResolver CreateResolver(Type target, string name)
+        private static IFieldResolver CreateResolver(
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.NonPublicProperties | DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods)] Type target,
+            string name)
         {
             var param = Expression.Parameter(typeof(IResolveFieldContext), "context");
             var source = Expression.MakeMemberAccess(param, _sourcePropertyInfo);
--- a/src/GraphQL/Types/Collections/IGraphTypeMappingProvider.cs
+++ b/src/GraphQL/Types/Collections/IGraphTypeMappingProvider.cs
@@ -12,5 +12,5 @@ public interface IGraphTypeMappingProvider
     /// <param name="clrType">The CLR type to be mapped.</param>
     /// <param name="isInputType">Indicates whether the type is an input type.</param>
     /// <param name="preferredGraphType">The graph type that is suggested for this CLR type.</param>
-    Type? GetGraphTypeFromClrType(Type clrType, bool isInputType, Type? preferredGraphType);
+    Type? GetGraphTypeFromClrType([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicMethods)] Type clrType, bool isInputType, Type? preferredGraphType);
 }
--- a/src/GraphQL/Types/Collections/ManualGraphTypeMappingProvider.cs
+++ b/src/GraphQL/Types/Collections/ManualGraphTypeMappingProvider.cs
@@ -15,6 +15,6 @@ internal class ManualGraphTypeMappingProvider : IGraphTypeMappingProvider
         _isOutputType = graphType.IsOutputType();
     }
 
-    public Type? GetGraphTypeFromClrType(Type clrType, bool isInputType, Type? preferredGraphType)
+    public Type? GetGraphTypeFromClrType([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicMethods)] Type clrType, bool isInputType, Type? preferredGraphType)
         => clrType == _clrType && (isInputType && _isInputType || !isInputType && _isOutputType) ? _graphType : preferredGraphType;
 }
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -220,7 +220,7 @@ namespace GraphQL.Types
             _introspectionTypes = CreateIntrospectionTypes(schema.Features.AppliedDirectives, schema.Features.RepeatableDirectives);
 
             _context = new TypeCollectionContext(
-               type => BuildGraphQLType(type, t => _builtInScalars.TryGetValue(t, out var graphType) ? graphType : _introspectionTypes.TryGetValue(t, out graphType) ? graphType : (IGraphType)Activator.CreateInstance(t)!),
+               type => BuildGraphQLType(type, ([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] t) => _builtInScalars.TryGetValue(t, out var graphType) ? graphType : _introspectionTypes.TryGetValue(t, out graphType) ? graphType : (IGraphType)Activator.CreateInstance(t)!),
                (name, type, ctx) =>
                {
                    SetGraphType(name, type);
@@ -394,7 +394,9 @@ namespace GraphQL.Types
         /// </summary>
         public int Count => Dictionary.Count;
 
-        private IGraphType BuildGraphQLType(Type type, IGraphType resolvedType)
+        private IGraphType BuildGraphQLType(
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] Type type,
+            IGraphType resolvedType)
             => BuildGraphQLType(type, _ => resolvedType);
 
         /// <summary>
@@ -405,10 +407,12 @@ namespace GraphQL.Types
         /// <see cref="IProvideResolvedType.ResolvedType"/> property is set to a new instance of
         /// the base (wrapped) type.
         /// </summary>
-        protected internal virtual IGraphType BuildGraphQLType(Type type, Func<Type, IGraphType> resolve)
+        protected internal virtual IGraphType BuildGraphQLType(
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] Type type,
+            Func<Type, IGraphType> resolve)
         {
             var local = resolve;
-            local ??= t => (IGraphType)Activator.CreateInstance(t)!;
+            local ??= ([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] t) => (IGraphType)Activator.CreateInstance(t)!;
             resolve = t => FindGraphType(t) ?? local(t);
 
             if (type.IsGenericType)
--- a/src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
+++ b/src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
@@ -133,7 +133,11 @@ internal static class AutoRegisteringOutputHelper
     /// that do not return <see langword="void"/> or <see cref="Task"/>
     /// including properties and methods declared on inherited classes.
     /// </summary>
-    public static IEnumerable<MemberInfo> GetRegisteredMembers<TSourceType>(Expression<Func<TSourceType, object?>>[]? excludedProperties)
+#if NET6_0_OR_GREATER
+    public static IEnumerable<MemberInfo> GetRegisteredMembers<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.Interfaces)] TSourceType>(Expression<Func<TSourceType, object?>>[]? excludedProperties)
+#else
+    public static IEnumerable<MemberInfo> GetRegisteredMembers<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicMethods)] TSourceType>(Expression<Func<TSourceType, object?>>[]? excludedProperties)
+#endif
     {
         if (typeof(TSourceType).IsInterface)
         {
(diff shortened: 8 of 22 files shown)


src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -405,10 +407,12 @@ private IGraphType BuildGraphQLType(Type type, IGraphType resolvedType)
...
         /// the base (wrapped) type.
         /// </summary>
-        protected internal virtual IGraphType BuildGraphQLType(Type type, Func<Type, IGraphType> resolve)
+        protected internal virtual IGraphType BuildGraphQLType(
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] Type type,
+            Func<Type, IGraphType> resolve)
         {
             var local = resolve;
-            local ??= t => (IGraphType)Activator.CreateInstance(t)!;
+            local ??= ([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] t) => (IGraphType)Activator.CreateInstance(t)!;
```
sungam3r, line 415: Weird. Trimmer warns even for lambda argument.

sungam3r:
One more.
5. Analyzer require to have the same annotations in virtual-override methods - https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-warnings/il2092. See `IGraphTypeMappingProvider`, `AutoRegisteringGraphTypeMappingProvider` and `ManualGraphTypeMappingProvider`. First I places annotations on `AutoRegisteringGraphTypeMappingProvider` - OK. Then on `IGraphTypeMappingProvider` - well... OK. But then I had to do it for `ManualGraphTypeMappingProvider` but there is no to do with `clrType` here - just a single compare operation!

src/GraphQL/Resolvers/NameFieldResolver.cs
```diff
@@ -60,7 +60,9 @@ public class NameFieldResolver : IFieldResolver
         /// <param name="target"> The type from which you want to get the value. </param>
         /// <param name="name"> Property/method name. </param>
         /// <returns> Compiled field resolver. </returns>
-        private static IFieldResolver CreateResolver(Type target, string name)
+        private static IFieldResolver CreateResolver(
```
Shane32, line 63: I might just mark this entire class with `[RequiresUnreferencedCode]` as the use of `NameFieldResolver` is not supported (although if the type was rooted, then it should work fine).  Marking the entire class will eliminate all warnings here and we can revert this line of code.

Shane32: Again, we will need to ignore warnings from callers.  (or else we will end up marking the entirety of GraphQL.NET with `RequiresUnreferencedCode` !)

src/GraphQL/Types/Collections/IGraphTypeMappingProvider.cs
```diff
@@ -12,5 +12,5 @@ public interface IGraphTypeMappingProvider
     /// <param name="clrType">The CLR type to be mapped.</param>
     /// <param name="isInputType">Indicates whether the type is an input type.</param>
     /// <param name="preferredGraphType">The graph type that is suggested for this CLR type.</param>
-    Type? GetGraphTypeFromClrType(Type clrType, bool isInputType, Type? preferredGraphType);
+    Type? GetGraphTypeFromClrType([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicMethods)] Type clrType, bool isInputType, Type? preferredGraphType);
```
Shane32, line 15: This is incorrect.  Simply having a CLR mapping to a graph type does not require all public methods/properties to be rooted.  Calls to `Field` will provide a static reference to the properties needed.

src/GraphQL/Types/Collections/ManualGraphTypeMappingProvider.cs
```diff
@@ -15,6 +15,6 @@ public ManualGraphTypeMappingProvider(Type clrType, Type graphType)
         _isOutputType = graphType.IsOutputType();
     }
 
-    public Type? GetGraphTypeFromClrType(Type clrType, bool isInputType, Type? preferredGraphType)
+    public Type? GetGraphTypeFromClrType([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicMethods)] Type clrType, bool isInputType, Type? preferredGraphType)
```
Shane32, line 18: This is incorrect. Simply having a CLR mapping to a graph type does not require all public methods/properties to be rooted.  Calls to `Field` will provide a static reference to the properties needed.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -220,7 +220,7 @@ protected void Initialize(ISchema schema, IServiceProvider serviceProvider, IEnu
             _introspectionTypes = CreateIntrospectionTypes(schema.Features.AppliedDirectives, schema.Features.RepeatableDirectives);
 
             _context = new TypeCollectionContext(
-               type => BuildGraphQLType(type, t => _builtInScalars.TryGetValue(t, out var graphType) ? graphType : _introspectionTypes.TryGetValue(t, out graphType) ? graphType : (IGraphType)Activator.CreateInstance(t)!),
+               type => BuildGraphQLType(type, ([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] t) => _builtInScalars.TryGetValue(t, out var graphType) ? graphType : _introspectionTypes.TryGetValue(t, out graphType) ? graphType : (IGraphType)Activator.CreateInstance(t)!),
```
Shane32, line 223: This is another situation where it might be more appropriate to mark the entire class as `[RequiresUnreferencedCode]`, and then ignore warnings from callers.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -133,7 +133,11 @@ private static bool IsObservableOrAsyncEnumerable(Type type)
     /// that do not return <see langword="void"/> or <see cref="Task"/>
     /// including properties and methods declared on inherited classes.
     /// </summary>
-    public static IEnumerable<MemberInfo> GetRegisteredMembers<TSourceType>(Expression<Func<TSourceType, object?>>[]? excludedProperties)
+#if NET6_0_OR_GREATER
```
Shane32, line 136: We could just use `[RequiresUnreferencedCode]` here.  It is not in a public class and warnings should be ignored by the caller, as we have the proper markings on the auto-registering types.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -303,7 +303,7 @@ protected virtual IObjectGraphType ToObjectGraphType(GraphQLObjectTypeDefinition
             return type;
         }
 
-        private void InitializeField(FieldConfig config, Type? parentType)
+        private void InitializeField(FieldConfig config, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.PublicProperties)] Type? parentType)
```
Shane32, line 306: This is for schema-first.

Shane32 (changes requested):
Thanks; I see you have found a number of attributes which should certainly be added, and others which the compiler should pick up automatically but are needed to eliminate warnings.  What we don't yet have are attributes indicating where it is proper to suppress warnings.  If you like, I can create a PR targeting this PR with my suggested changes from this review, plus some of the suppression attributes that I suggest we add.

Shane32:
> 1. It is very toxic for code - even worse then async/await "infection" or NRT annotations.

I have no problem at all with async/await or NRT annotations, and on the contrary appreciate the explicitness of async/await and NRT annotations.  In this case, the annotations are only needed where we are using reflection, and necessarily so if we want to provide the best experience for the user with trimming support.  Other libraries that do not use trimming would not need such annotations.  Honestly, there are no annotations within the majority of the codebase; nothing in DocumentExecuter, the execution strategies, schema printer, validation rules, and most other classes.

At the same time, we make extensive use of reflection, such that if we were to follow the suggested annotations based on build warnings, we would likely spread the 'infection' throughout the entire codebase, primarily due to a few specific methods -- namely `ToObject`, `GetGraphTypeFromType`, the `SchemaTypes` class and `NameFieldResolver`.  So, what we really need to is determine where the suppressions and warnings need to be placed, and all will be well.

> 2. I do not understand why we should manually write annotations - the analyzer already knows what is missing in specific cases and demands to put it into code. It looks strange. I hope that in NET8 this will be fixed and it will not be necessary to manually do the work. Though it may be the same story as for NRT - analizer knows but it's a developer responsibility to say the last word.

There certainly are a few annotations that the compiler is requiring that should not be necessary; namely static methods which calls other annotated methods.  They are certainly not required.  Perhaps for virtual methods they might be due to the methods being overridden.  I assume this must be to ease analysis of libraries built on these methods; perhaps at design time.  It is possible that with .NET 8 the analyzer rule will be updated to reduce this need.  However, most of the build warnings addressed in this PR do not fall in this category, and make sense to me.

> 3. I do not understant what to do in 50% of cases when `Type` comes from some calculated variable.

I can review.  I think there is a logical answer, but often that answer might to simply mark a whole method or class as `[RequiresUnreferencedCode]` and suppress the warning elsewhere, for ease of development more than anything else.

> 4. It seems to me that we "excessive" demand trimmer to preserve some type's parts in methods with branch logic (for example, type.GetProperties() in the first branch and type.GetMethods() in the second).

I think that is the case in a few places.  I have no problem with that; better the trimmer trims to little than too much.  Allowing the user to trim their assembly has a significant benefit, I believe, in certain circumstances, rather than requiring that the user's assembly be rooted for any use of GraphQL.NET.

> 5. Analyzer require to have the same annotations in virtual-override methods

Nowhere would I add annotations simply because the analyzer suggests we do so.  The analyzer may not be correct.  We may need to suppress warnings in some cases, either because the warning is incorrect, or because it is not applicable in the scenario.

sungam3r:
>  What we don't yet have are attributes indicating where it is proper to suppress warnings.

I agree. It's a bit tricky yet to figure out the right way to go with trimming, not much experience here.

sungam3r:
>  If you like, I can create a PR targeting this PR with my suggested changes from this review, plus some of the suppression attributes that I suggest we add.

Actually I think that I'll close this one to continue with the series of some smaller PRs. I'm not sure about all my changes here. I would like first to add trimming annotations in those places that 100% safe to add.

src/GraphQL/Types/Scalars/Enumeration/EnumerationGraphType.cs
```diff
@@ -118,11 +118,18 @@ public override GraphQLValue ToAST(object? value)
 /// Also it can get descriptions for enum fields from the XML comments.
 /// </summary>
 /// <typeparam name="TEnum"> The enum to take values from. </typeparam>
-public class EnumerationGraphType<TEnum> : EnumerationGraphType
+public class EnumerationGraphType<[DynamicallyAccessedMembers(GET_ALL_MEMBERS)] TEnum> : EnumerationGraphType
```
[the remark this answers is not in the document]
sungam3r: From MS sources
```
  [DynamicallyAccessedMembers(GetAllMembers)]
        public virtual MemberInfo[] GetMember(string name, BindingFlags bindingAttr) => GetMember(name, MemberTypes.All, bindingAttr);

  internal const DynamicallyAccessedMemberTypes GetAllMembers = DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields |
            DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods |
            DynamicallyAccessedMemberTypes.PublicEvents | DynamicallyAccessedMemberTypes.NonPublicEvents |
            DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.NonPublicProperties |
            DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.NonPublicConstructors |
            DynamicallyAccessedMemberTypes.PublicNestedTypes | DynamicallyAccessedMemberTypes.NonPublicNestedTypes;
```

I will not merge this particular change and this entire PR and I'll resolve all other conversations where I'm not 100% sure about correctness.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -36,7 +36,7 @@ public class InputObjectGraphType : InputObjectGraphType<object>
     }
 
     /// <inheritdoc cref="IInputObjectGraphType"/>
-    public class InputObjectGraphType<TSourceType> : ComplexGraphType<TSourceType>, IInputObjectGraphType
+    public class InputObjectGraphType<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.NonPublicProperties | DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields)] TSourceType> : ComplexGraphType<TSourceType>, IInputObjectGraphType
```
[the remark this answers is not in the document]
sungam3r: It's required by `value.ToObject` and inner calls from it.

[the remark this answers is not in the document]
Shane32: No, it isn't "required".  It's simply the warning on the inner calls within `ToObject` because .NET doesn't know that a static reference exists in the code.  We need not apply warnings to our methods where it is known that the warning does not apply.  Rather, it is correct and proper to suppress them.

src/GraphQL/Extensions/ObjectExtensions.cs
```diff
@@ -38,7 +38,10 @@ public static T ToObject<T>(this IDictionary<string, object?> source)
         /// In case of configuring field as Field(x => x.FName).Name("FirstName") source dictionary
         /// will have 'FirstName' key but its value should be set to 'FName' property of created object.
         /// </param>
-        public static object ToObject(this IDictionary<string, object?> source, Type type, IGraphType? mappedType = null)
+        public static object ToObject(
+            this IDictionary<string, object?> source,
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.NonPublicProperties | DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields)] Type type,
```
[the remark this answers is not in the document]
sungam3r: I'm not sure here. I just "bubbled up" requirements from called methods up to the stack. In the process of this, I began to guess that a deeper analysis of the code was required.

src/GraphQL/Extensions/SchemaExtensions.cs
```diff
@@ -128,7 +128,7 @@ public static void AutoRegister(this ISchema schema, Type clrType, AutoRegisteri
         /// <param name="schema">The schema for which the mapping is registered.</param>
         /// <typeparam name="TClrType">The CLR property type from which to infer the GraphType.</typeparam>
         /// <param name="mode">Which registering mode to use - input only, output only or both.</param>
-        public static void AutoRegister<TClrType>(this ISchema schema, AutoRegisteringMode mode = AutoRegisteringMode.Both)
+        public static void AutoRegister<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicMethods)] TClrType>(this ISchema schema, AutoRegisteringMode mode = AutoRegisteringMode.Both)
```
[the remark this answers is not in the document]
sungam3r: As I understand the general approach here is the same as with NRT - we should not hide information about "trimming requirements", instead we should expose it to the callers and so on and on - like async/await "infection". At least MS documentation talks about it. At some point type becomes fixed and this "infection" ends. Or one cansuppress warning if he knows usage context.

src/GraphQL/Resolvers/NameFieldResolver.cs
sungam3r: > Again, we will need to ignore warnings from callers.

I don't understand why so. If we suppress warnings, then this is provided by some guarantees in the code.

src/GraphQL/Types/Collections/IGraphTypeMappingProvider.cs
sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/pull/3519#issuecomment-1412933180

How to deal with warning here?

src/GraphQL/Types/Collections/ManualGraphTypeMappingProvider.cs
sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/pull/3519#issuecomment-1412933180

How to deal with warning here?

src/GraphQL/Types/Collections/SchemaTypes.cs
sungam3r: > This is another situation where it might be more appropriate to mark the entire class as [RequiresUnreferencedCode], 

While I slightly understand the reasoning here I do not undestand the next part:

> and then ignore warnings from callers.

src/GraphQL/Utilities/SchemaBuilder.cs
sungam3r: Do you mean that for entire schema-first parts of project we should not bother with all these annotations and just mark all top-level public classes with `[RequiresUnreferencedCode]`?

sungam3r:
>  In this case, the annotations are only needed where we are using reflection, and necessarily so if we want to provide the best experience for the user with trimming support.

I would say that the problem is about transitivity in trimming annotations. It is not obvious to me at what stage (method/type) these annotations should stop.

> Other libraries that do not use trimming would not need such annotations.

_use_ -> _support_ I would say. Apps can use (can require) trimming on publish as I understand, not libs.

> So, what we really need to is determine where the suppressions and warnings need to be placed, and all will be well.

I agree. From my side I still do not feel this edge.

>  but often that answer might to simply mark a whole method or class as [RequiresUnreferencedCode] and suppress the warning elsewhere, for ease of development more than anything else.

I see that you are constantly using a pair of something like "let's use [RequiresUnreferencedCode] here" immediately followed by something like "and suppress the warning elsewhere" 😄 . I don't get it. Why to add annotations at all if then suppress all usages of the marked APIs?

src/GraphQL/Types/Composite/InputObjectGraphType.cs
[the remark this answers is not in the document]
sungam3r: OK. I got the core idea, makes sense, but then the next question is - how to know when the warning does not **really** apply? I guess that the answer lies on the surface - one need to know well and understand the internal structure of code and its usage (usage _patterns_ I mean) by consumers. It turns out that this task is **much** more complicated than NRT annotations.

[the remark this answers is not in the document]
sungam3r: > Rather, it is correct and proper to suppress them.

One more note. I think that suppression messages should be as much detailed as possible to keep things working while project evolves. I mean this is some sort of implicit contract between layered APIs and their consumers. This implicit contract is quite easy to break invisible for everyone. In this regard, the question arises of comprehensive testing. It would be good to somehow run all the already written tests (OK, not all but most of them) in the "AOT-ed environment". I wonder about something like _xUnit AOT runner_ 🤔 .

sungam3r:
> We should be more concerned with "100% correct to add". 

Yes, I mean _correct_ instead of _safe_.

src/GraphQL/Extensions/AssemblyExtensions.cs
```diff
@@ -22,6 +22,8 @@ internal static class AssemblyExtensions
         /// Skips classes where the source type is <see cref="object"/>, or where the class is marked with
         /// the <see cref="DoNotMapClrTypeAttribute"/>.
         /// </summary>
+        [RequiresUnreferencedCode("
+Please ensure that the graph types used by your schema and their constructors are not trimmed by the compiler.")]
```
Shane32, line 26:
```suggestion
        [RequiresUnreferencedCode("Please ensure that the graph types used by your schema and their constructors are not trimmed by the compiler.")]
```

sungam3r:
@Shane32 I would not merge this. Let's keep it for some time as a reference/reminder. Instead I would like to add new AOT annotations one by one (1-5-10 APIs) so that each PR will only change a small part of the codebase.

Shane32:
I agree.  I also tried to make some minor updates yesterday to annotate just `ToObject` but found that there was no way to properly annotate the method, since it is recursive and operates on list types.  This restriction then applies to the calls to `Field` within `InputObjectGraphType` (inherited members from `ComplexGraphType`).  I feel we will find this type of issue persists throughout the codebase.  We have a choice of either suppressing all aberrations, "doing our best" and expecting it will work in most situations, or else in addition to "doing our best" we also mark the entry point(s) as `[RequiresUnreferencedCode]` similar to how the MS DI provider works.

Shane32:
@sungam3r As you know, .NET 8 has a heavy focus on AOT compatibility I've been reading up on their approach, and it seems that one technique they use often to provide an AOT-compatible package is to detect known incompatibilities and throw an exception.  For example, one of their goals was to provide an AOT-compatible dependency injection provider.  However, if a value type was registered (such as `MyStruct`), and then the enumerable type was requested (e.g. `IEnumerable<MyStruct>`), there would be a runtime error since static code analysis could not reveal that the array type of `MyStruct` was needed.  In their case the solution was to throw an exception if any value-type was registered, considering that the likelyhood of this need was rare.

In a similar manner, we could check the [`IsDynamicCodeSupported` property](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimefeature.isdynamiccodesupported?view=netstandard-2.1#system-runtime-compilerservices-runtimefeature-isdynamiccodesupported) and throw an exception for certain situations where it is not practical to simply mark the method as `[RequiresDynamicCode]`.  One such example is adding a schema-initialization check for any field resolvers that use `NameFieldResolver`.  We can throw an exception in this case, so it is clear to a developer that the code will not operate correctly in production when they are debugging it.  Another situation is input object CLR classes which include an `IEnumerable<T>` or `IList<T>` property of value types.  While `List<T>` of a value type would be able to be analyzed by static code analysis, an interface would not.

Of course other methods that simply fail to run correctly can be marked as `[RequiresDynamicCode]`

sungam3r:
Maybe, maybe. My general feeling is about trim analyzer **should** be smart enough to do 90% of things that we have to do now marking code with `[DynamicallyAccessedMemberTypes(..)]`. After all, the compiler "sees" all calls to reflection methods, invoking ctors, etc. Why should I as developer do all that things? It looks ridiculous. Maybe MS was not ready with that feature 100% for NET7 so they released it as is requiring 100500 hints from developers to guide trimmer.

I don't want to guide trimmer so much. This is a work for a compiler/trimmer itself, not a person.

PR #3520, opened 2023-02-06
Get Query and Mutation via constructor injection

EmmanuelPonnudurai:
If the items are registered in the DI container, then constructor injection can be preferred over explicit call to `GetRequiredService`.

### DIFF
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -31,11 +31,11 @@ You can override the default implementation by passing a `IServiceProvider` to t
 ```csharp
 public class StarWarsSchema : GraphQL.Types.Schema
 {
-    public StarWarsSchema(IServiceProvider provider)
+    public StarWarsSchema(IServiceProvider provider, StarWarsQuery query, StarWarsMutation mutation)
         : base(provider)
     {
-        Query = provider.GetRequiredService<StarWarsQuery>();
-        Mutation = provider.GetRequiredService<StarWarsMutation>();
+        Query = query;
+        Mutation = mutation;
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -155,8 +155,20 @@ public void ConfigureServices(IServiceCollection services)
 }
 ```
 
-If you previously pulled in your query, mutation and/or subscription classes via dependency injection, you will need
-to manually pull in those dependencies from the `SelfActivatingServiceProvider` via `GetRequiredService` as follows:
+If you previously pulled in your query, mutation and/or subscription classes via dependency injection, you can pull in those dependencies via constructor injection OR via `GetRequiredService` as follows:
```
Shane32, line 158: This is incorrect in the context provided.  See the code on line 154 of the `dependency-injection.md` file:

```cs
services.AddSingleton<ISchema, StarWarsSchema>(services => new StarWarsSchema(new SelfActivatingServiceProvider(services)));
```

As seen there, the DI setup is specifically calling `new StarWarsSchema(...)`, and providing it with a specific `IServiceProvider` instance.  This instance must be used to pull the Query and Mutation instances from dependency injection, as this example is describing specifically how to not-register any GraphQL types (which include the Query and Mutation types) in the DI engine.

However, if the `AddSelfActivatingSchema<TSchema>()` method were used on the `IGraphQLBuilder` interface (see line 89), then the query and mutation types could be pulled in via constructor injection.

We could also modify line 154 so that the query and mutation types may use constructor injection.  This would require this change to the sample:

```cs
{
    services.AddSingleton<ISchema, StarWarsSchema>(services =>
    {
        var selfActivatingServices = new SelfActivatingServiceProvider(provider);
        return ActivatorUtilities.CreateInstance<TSchema>(selfActivatingServices);
    });
}
```

Shane32:
Honestly we should rewrite this entire document based on the `IGraphQLBuilder` interface...

docs2/site/docs/getting-started/dependency-injection.md
EmmanuelPonnudurai: OK got it. It was not clear when I first saw it, my bad.

Shane32 (approved):

PR #3522, opened 2023-02-07
Prevent double registration in AddSelfActivatingSchema

sungam3r:
Parent issue: #2797


Shane32:
I would not change this until we have a consensus, and if deemed worthy, change all similar places.  There is no reason why this particular instance is more or less susceptible to double registration than `AddSchema`, for example.  We also have double registration in other places like the serializers and middleware.

PR #3021, opened 2022-03-22
Add 'Then' data loader extension method for lists of data loader results.

Shane32:
See need in issue #3019

PR #3525, opened 2023-02-08
Fix data loader Then method

Shane32:
Fixes poor design introduced in #3021.

The `Then` methods introduced in #3021 require that the input enumerable list has been iterated to work properly.  Although the xml comments describe this behavior, it is certainly not intuitive.  This was a bad design and should never have been a prerequisite for using these particular `Then` methods.  It's caused a 3-day outage in our servers due to this bad design.

This patch checks if the input list is an `IList` implementation (commonly an array or `List<T>`), indicating that the source list has been enumerated.  If it has not been, it does so via `ToList()`, which is correct and ** **required** ** for proper operation.

Note that another extension method already performs correct handling of lists, executing `LoadAsync` for each member and adding the resulting `IDataLoaderResult` instances to a list before passing it along to the remainder of the function:

```cs
public static IDataLoaderResult<T[]> LoadAsync<TKey, T>(this IDataLoader<TKey, T> dataLoader, IEnumerable<TKey> keys)
```

In my issue today, 100s of child nodes were supposed to be queued to load via a data loader, but instead were individually queued into a queue of 1 entry each, and each entry loaded from the database independently.  The amount of overhead for each data loader and database request, although small, was enough to max out the CPU on the server which generally is idle, without taxing the database itself too much.

### DIFF
--- a/src/GraphQL.DataLoader.Tests/Types/QueryType.cs
+++ b/src/GraphQL.DataLoader.Tests/Types/QueryType.cs
@@ -79,10 +79,7 @@ public class QueryType : ObjectGraphType
                     users.GetUsersByIdAsync);
 
                 var ids = ctx.GetArgument<IEnumerable<int>>("ids");
-                // note: does not work properly without ToList, because LoadAsync would not have
-                // been called, so the ids would not have been queued for execution prior to the
-                // first call to GetResultAsync
-                var ret = ids.Select(id => loader.LoadAsync(id)).ToList();
+                var ret = ids.Select(id => loader.LoadAsync(id));
                 var ret2 = ret.Then(values => values.Where(x => x != null));
                 return ret2;
             });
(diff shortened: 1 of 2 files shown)


src/GraphQL.DataLoader.Tests/Types/QueryType.cs
```diff
@@ -79,10 +79,7 @@ public QueryType(IDataLoaderContextAccessor accessor, IUsersStore users, IOrders
                     users.GetUsersByIdAsync);
 
                 var ids = ctx.GetArgument<IEnumerable<int>>("ids");
-                // note: does not work properly without ToList, because LoadAsync would not have
-                // been called, so the ids would not have been queued for execution prior to the
-                // first call to GetResultAsync
-                var ret = ids.Select(id => loader.LoadAsync(id)).ToList();
+                var ret = ids.Select(id => loader.LoadAsync(id));
```
Shane32, line 82: This test failed before this patch, as it was known to do at the time I wrote it.  This actually a perfect sample of the bad design, requiring users to add `ToList` when such behavior is required.

Shane32:
It is also significant to note that incorrect use does not 'break' operation, meaning that the data will be retrieved correctly (just not data-loaded), so the issue can be unnoticed for a long period of time, and then hard-to-find once it becomes a problem.

Shane32:
@sungam3r if you have no comments then I will merge this

sungam3r:
Working on it.

sungam3r (approved):

PR #3526, opened 2023-02-09
Use PreserveNewest

sungam3r:
https://github.com/dotnet/project-system/blob/main/docs/up-to-date-check.md#copytooutputdirectory-always-vs-preservenewest


Shane32:
I don’t understand how this change applies to us or why we would change this. If for speedier builds, MS already applied a “fix” in 17.2, which is rather old now.

sungam3r:
If there are no explicit reasons to copy the entire file each time, then the documentation advises to use `PreserveNewest` mode. Other two places in the codebase do so.

Shane32 (approved):

PR #3529, opened 2023-02-12
Fix of eternal loop in ComplexityAnalyzer in case of fragment cycles

### LINKED ISSUE #3527: ComplexityValidationRule deadlock
## Description

After adding `ComplexityValidationRule` to the `ValidationRules` as suggested in the migration (v7) documentation process of validation of query with `RecursiveFragment` ends in an infinite loop in `ComplexityAnalyzer`'s method `GetDependencies`.

My understanding of the issue:

In v5 the ComplexityValidation was done after ValidationRules so the `NoFragmentCycles` validation rule ended the validation process with error before complexity analyzer.

In v7 the ComplexityValidation was transformed into ValidationRule and it runs with all the other core / default validation rules asynchronously. So even though the `NoFragmentCycles` detects the cycle correctly the ComplexityValidationRule runs simultaneously and never ends.

### Steps to reproduce

1. Add `ComplexityValidationRule` to `ValidationRules` of `ExecutionOptions`
2. Execute query like this
```
                   {
                        type_All(limit: 20) { 
                            items { 
                                links(limit: 20) { 
                                    items { 
                                        ... RecursiveFragment
                                    }
                                }
                            } 
                        }
                    } 
                    fragment RecursiveFragment on Type   {
                       links(limit: 20) {
                            items {
                                ... RecursiveFragment
                            }
                        }
                    }
```

### Expected result

It should end with `Cannot spread fragment 'RecursiveFragment' within itself.` and end the process of validation somehow.

### Actual result

Deadlock in of the following method.

1. There is one `_selectionSetsToVisit`
2. It `Pop()` out
3. Then it `Push()` another one back in
4. Goes back to 1.

```
HashSet<GraphQLFragmentDefinition>? GetDependencies(GraphQLFragmentDefinition def)
            {
                HashSet<GraphQLFragmentDefinition>? dependencies = null;
                _selectionSetsToVisit.Push(def.SelectionSet);

                while (_selectionSetsToVisit.Count > 0)
                {
                    foreach (var selection in _selectionSetsToVisit.Pop().Selections)
                    {
                        if (selection is GraphQLFragmentSpread spread)
                        {
                            var frag = document.FindFragmentDefinition(spread.FragmentName.Name.Value);
                            if (frag != null)
                            {
                                (dependencies ??= new()).Add(frag);
                                _selectionSetsToVisit.Push(frag.SelectionSet);
                            }
                        }
                        else if (selection is IHasSelectionSetNode hasSet && hasSet.SelectionSet != null)
                        {
                            _selectionSetsToVisit.Push(hasSet.SelectionSet);
                        }
                    }
                }

                _selectionSetsToVisit.Clear();
                return dependencies;
            }
```

### Environment

Win10

sungam3r:
Fixes #3527.

There are 2 bugs in fact.
1. `ComplexityValidationRule` should run after all other standard rules. Before fix it was actually run (syncronously) **before** those rules that return non-null visitor. Note that @arguit wrote test with absolutely wrong schema, not StarWars.
2. `ComplexityValidationRule` indeed hangs in infinite while loop.

I successfully fixed the first. Also I added fast check to eliminate all possible problems with complexity analysis. Then I decided to fix the second. I "fixed" it but tests failed. It turned out that the code I added in `ComplexityAnalyzer.cs` is not enough. It was a naive approach. In fact, `NoFragmentCycles` rule should always be run if we use complexity analysis. First I thought to "force-inject" it into list of validation rules but then I changed my mind. It seems that it will be better to rollback changes in `ComplexityAnalyzer.cs` at all and go on with fix1. It is unlikely that someone uses complexity analysis when turning off the standard rules. Fix1 should solve #3527 so I believe fix2 is not needed. 

[UPDATED] I changed check to simple counter with limit of 2000.

### DIFF
--- a/src/GraphQL.Tests/Complexity/ComplexityValidationTests.cs
+++ b/src/GraphQL.Tests/Complexity/ComplexityValidationTests.cs
@@ -1,3 +1,5 @@
+using GraphQL.Execution;
+using GraphQL.Validation;
 using GraphQL.Validation.Complexity;
 using GraphQL.Validation.Errors;
 using GraphQL.Validation.Errors.Custom;
@@ -170,6 +172,6 @@ public class ComplexityValidationTest : ComplexityTestBase
 
         res.Result.Errors.ShouldNotBe(null);
         res.Result.Errors.Count.ShouldBe(1);
-        res.Result.Errors[0].ShouldBeOfType<NoFragmentCyclesError>().Message.ShouldBe("Cannot spread fragment 'RecursiveFragment' within itself.");
+        res.Result.Errors[0].ShouldBeOfType<ValidationError>().Message.ShouldBe("It looks like document has fragment cycle. Please make sure you are using standard validation rules especially NoFragmentCycles one.");
     }
 }
--- a/src/GraphQL.Tests/Complexity/ComplexityValidationTests.cs
+++ b/src/GraphQL.Tests/Complexity/ComplexityValidationTests.cs
@@ -1,9 +1,6 @@
-using GraphQL.Execution;
 using GraphQL.Validation;
 using GraphQL.Validation.Complexity;
-using GraphQL.Validation.Errors;
 using GraphQL.Validation.Errors.Custom;
-using GraphQL.Validation.Rules;
 
 namespace GraphQL.Tests.Complexity;
 
--- a/src/GraphQL.Tests/Complexity/ComplexityValidationTests.cs
+++ b/src/GraphQL.Tests/Complexity/ComplexityValidationTests.cs
@@ -1,5 +1,7 @@
 using GraphQL.Validation.Complexity;
+using GraphQL.Validation.Errors;
 using GraphQL.Validation.Errors.Custom;
+using GraphQL.Validation.Rules;
 
 namespace GraphQL.Tests.Complexity;
 
@@ -103,5 +105,71 @@ public class ComplexityValidationTest : ComplexityTestBase
         res.Result.Errors[0].ShouldBeOfType<ComplexityError>();
         res.Result.Errors[0].Message.ShouldBe("Query is too complex to execute. Complexity is 480, maximum allowed on this endpoint is 25. The field with the highest complexity is 'friends' with value 125.");
         res.Result.Errors[0].InnerException.ShouldBeNull();
+    }
+
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/3527
+    [Fact]
+    public void recursive_fragment_should_not_end_in_eternal_loop1()
+    {
+        const string query = """
+            {
+                type_All(limit: 20) {
+                    items {
+                        links(limit: 20) {
+                            items {
+                                ... RecursiveFragment
+                            }
+                        }
+                    }
+                }
+            }
+            fragment RecursiveFragment on Type   {
+               links(limit: 20) {
+                    items {
+                        ... RecursiveFragment
+                    }
+                }
+            }
+            """;
+
+        var complexityConfiguration = new ComplexityConfiguration();
+        var res = Execute(complexityConfiguration, query);
+
+        res.Result.Errors.ShouldNotBe(null);
+        res.Result.Errors.Count.ShouldBe(3);
+        res.Result.Errors.All(e => e is not ComplexityError);
+    }
+
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/3527
+    [Fact]
+    public void recursive_fragment_should_not_end_in_eternal_loop2()
+    {
+        const string query = """
+            {
+                type_All(limit: 20) {
+                    items {
+                        links(limit: 20) {
+                            items {
+                                ... RecursiveFragment
+                            }
+                        }
+                    }
+                }
+            }
+            fragment RecursiveFragment on Type   {
+               links(limit: 20) {
+                    items {
+                        ... RecursiveFragment
+                    }
+                }
+            }
+            """;
+
+        var complexityConfiguration = new ComplexityConfiguration();
+        var res = Execute(complexityConfiguration, query, onlyComplexityRule: true);
+
+        res.Result.Errors.ShouldNotBe(null);
+        res.Result.Errors.Count.ShouldBe(1);
+        res.Result.Errors[0].ShouldBeOfType<NoFragmentCyclesError>().Message.ShouldBe("Cannot spread fragment 'RecursiveFragment' within itself.");
     }
 }
--- a/src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
+++ b/src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
@@ -121,7 +121,6 @@ namespace GraphQL.Validation.Complexity
         private static Dictionary<GraphQLFragmentDefinition, HashSet<GraphQLFragmentDefinition>?> BuildDependencies(GraphQLDocument document)
         {
             Stack<GraphQLSelectionSet> _selectionSetsToVisit = new();
-            HashSet<ROM> _visitedFragments = new();
             Dictionary<GraphQLFragmentDefinition, HashSet<GraphQLFragmentDefinition>?> dependencies = new();
 
             foreach (var fragmentDef in document.Definitions.OfType<GraphQLFragmentDefinition>())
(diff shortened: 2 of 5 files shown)


src/GraphQL/Validation/Rules.Custom/ComplexityValidationRule.cs
```diff
@@ -35,16 +36,49 @@ public ComplexityValidationRule(ComplexityConfiguration complexityConfiguration,
 
     /// <inheritdoc/>
     public ValueTask<INodeVisitor?> ValidateAsync(ValidationContext context)
+        => new(new PostponeComplexityValidationVisitor(_complexityConfiguration, _complexityAnalyzer));
+
+    // https://github.com/graphql-dotnet/graphql-dotnet/issues/3527
+    private sealed class PostponeComplexityValidationVisitor : INodeVisitor
```
sungam3r, line 42: @Shane32 I think to revert all files except this one and `recursive_fragment_should_not_end_in_eternal_loop1`.

src/GraphQL/Validation/Rules.Custom/ComplexityValidationRule.cs
```diff
@@ -35,16 +36,49 @@ public ComplexityValidationRule(ComplexityConfiguration complexityConfiguration,
...
+
+        public ValueTask LeaveAsync(ASTNode node, ValidationContext context)
         {
-            context.ReportError(ex);
+            // Complexity analysis should run at the very end.
+            if (node is GraphQLDocument)
+            {
+                // Fast return here to avoid all possible problems with complexity analysis.
+                // For example, document may contain fragment cycles, see https://github.com/graphql-dotnet/graphql-dotnet/issues/3527
+                // Note, that ComplexityValidationRule/ComplexityAnalyzer still checks for fragment cycles since there may be no standard rules configured.
```
sungam3r, line 66: TODO: remove this comment

src/GraphQL.Tests/Complexity/ComplexityValidationTests.cs
```diff
@@ -104,4 +106,70 @@ public void fail_when_too_complex_and_nested()
...
+               links(limit: 20) {
+                    items {
+                        ... RecursiveFragment
+                    }
+                }
+            }
+            """;
+
+        var complexityConfiguration = new ComplexityConfiguration();
+        var res = Execute(complexityConfiguration, query, onlyComplexityRule: true);
```
Shane32, line 169: This should test with all rules and verify that only one error is produced, and no deadlock occurs.

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
```diff
@@ -140,6 +142,9 @@ private static GraphQLFragmentDefinition GetFirstFragmentWithoutPendingDependenc
                     {
                         if (selection is GraphQLFragmentSpread spread)
                         {
+                            if (!_visitedFragments.Add(spread.FragmentName.Name.Value))
+                                throw new NoFragmentCyclesError(document, spread.FragmentName.Name.StringValue);
```
Shane32, line 146: I didn’t review all the code here, but my gut feeling is that to match all of our other code this should not this an error, but silently ignore the situation. There are many situations which can break the complexity analyzer (or similarly the authorization rule), such as referencing a type, fragment or variable that does not exist.

Shane32: In v8, when we implement our plans to have built-in rules run first, then read arguments, then other rules, then we do not need to worry about these kinds of situations here, or in the authorization rule, or within third-party rules, since the built-in rules would prevent such a situation from occurring. Sorry I have not worked on it recently, but I’ll get it done.

src/GraphQL.Tests/Complexity/ComplexityValidationTests.cs
sungam3r: `recursive_fragment_should_not_end_in_eternal_loop1` is the one with all rules

Shane32: I guess reviewing from my cell phone is proving difficult….!!!

sungam3r: I review only the simplest PRs from cell phone. It is simply unrealistic to work with more serious edits.

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
sungam3r: I changed this check to the primitive one. It works for now.

src/GraphQL/Validation/Rules.Custom/ComplexityValidationRule.cs
sungam3r: Nope. KISS pattern made a job.

src/GraphQL/Validation/Rules.Custom/ComplexityValidationRule.cs
sungam3r: nope, simple check with counter works

Shane32:
> 1. `ComplexityValidationRule` should run after all other standard rules. Before fix it was actually run (syncronously) **before** those rules that return non-null visitor. Note that @arguit wrote test with absolutely wrong schema, not StarWars.

Can you explain why this is a bug?  What do other validation rules have to do with the complexity rule?

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
```diff
@@ -134,8 +134,14 @@ private static GraphQLFragmentDefinition GetFirstFragmentWithoutPendingDependenc
                 HashSet<GraphQLFragmentDefinition>? dependencies = null;
                 _selectionSetsToVisit.Push(def.SelectionSet);
 
+                int counter = 0;
+                const int MAX_ITERATIONS = 2000;
                 while (_selectionSetsToVisit.Count > 0)
                 {
+                    // https://github.com/graphql-dotnet/graphql-dotnet/issues/3527
+                    if (++counter > MAX_ITERATIONS)
+                        throw new ValidationError("It looks like document has fragment cycle. Please make sure you are using standard validation rules especially NoFragmentCycles one.");
```
Shane32, line 143: It would seem that with this check in place there should be no need to add the `PostponeComplexityValidationVisitor` and related code.

Shane32: Besides the fact that the error would be reported twice, that is.

sungam3r:
> Can you explain why this is a bug? What do other validation rules have to do with the complexity rule?

This is a bug because standard checks should run first to validate "base structure" of document. It does not make sense to start custom rules. The situation here is about the same as with parsing variables, when it makes no sense to try to parse them if the validation of the document failed. Remember, you recently wrote about this in your PR?

src/GraphQL/Validation/Complexity/ComplexityAnalyzer.cs
sungam3r: I strongly feel that complexity analysis should be run after all standard validation rules if any. Remember that we switched that code into validation rule as a internal techical aspect. One can wrap any code into validation rule, but this does not mean that it can and should be launched before the basic checks are performed.

Shane32:
> > Can you explain why this is a bug? What do other validation rules have to do with the complexity rule?
> 
> This is a bug because standard checks should run first to validate "base structure" of document. It does not make sense to start custom rules. The situation here is about the same as with parsing variables, when it makes no sense to try to parse them if the validation of the document failed. Remember, you recently wrote about this in your PR?

This is not the time or place to fix this issue, by adding a whole ton of overhead in this one rule because of a poor design.  Nor is it truly required; the authorization rule I wrote works fine given any document.  Let us fix the infinite loop issue temporarily - whether it produces and extra error or not - and resolve validation rules for v8 permanently.

Shane32:
Please consider anyway; I really don't care much about the complexity analyzer at present - I will certainly be writing a new one from scratch in the future, either for my company only, or as open source to replace this useless one.  I've stated the MANY deficiencies of the complexity analyzer many times and need not reiterate them.

If you're sure the current form of the PR is the best answer at present, that's fine.

sungam3r:
> by adding a whole ton of overhead in this one rule

What ton of overhead?

Shane32:
> > by adding a whole ton of overhead in this one rule
> 
> What ton of overhead?

All of the changes within `ComplexityValidationRule.cs` - the entire PostponeComplexityValidationVisitor class and more.

sungam3r:
From overhead point of view it is just +1 allocation. It's nothing compared to all actions done by complexity analyzer.

> I will certainly be writing a new one from scratch in the future

Good. And now only a fix is needed, not a global rewrite. Feel free to rewrite complexity analyzer as you see fit for v8.

> If you're sure the current form of the PR is the best answer at present, that's fine.

I think so. At least for v7.x timeline.

Shane32 (approved):

PR #3535, opened 2023-02-16
Fix NRT attributes in ObjectExtensions.cs

Shane32:
Fixes build problems with latest .NET SDK.  No effective code changes or public API changes.


Shane32:
Merging now since this is breaking other PRs/tests

sungam3r:
Ok

PR #3536, opened 2023-02-16
CI fix for .NET SDK 7.0.200

Shane32:
See:
- https://github.com/dotnet/sdk/issues/15607
- https://github.com/dotnet/sdk/issues/30624
- https://github.com/dotnet/sdk/issues/30625
- https://github.com/dotnet/sdk/issues/30624#issuecomment-1432118204

The new error will be downgraded to a warning in the release next month.  Until then we can explicitly specify the output folder as shown in this PR.  Note that the `-o` option will not produce an error for `dotnet pack` at all after the release next month.

### DIFF
--- a/.github/workflows/build-artifacts-code.yml
+++ b/.github/workflows/build-artifacts-code.yml
@@ -54,7 +54,8 @@ jobs:
         run: dotnet build --no-restore -c Release -p:VersionSuffix=$GITHUB_RUN_NUMBER
       - name: Pack solution [Release]
         working-directory: src
-        run: dotnet pack --no-restore --no-build -c Release -p:VersionSuffix=$GITHUB_RUN_NUMBER -o out
+        run: dotnet pack --no-restore --no-build -c Release -p:VersionSuffix=$GITHUB_RUN_NUMBER -p:PackageOutputPath=out
+        # See: https://github.com/dotnet/sdk/issues/30624#issuecomment-1432118204
       - name: Upload artifacts
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/publish-code.yml
+++ b/.github/workflows/publish-code.yml
@@ -61,7 +61,8 @@ jobs:
         run: dotnet build --no-restore -c Release -p:Version=$version
       - name: Pack solution [Release]
         working-directory: src
-        run: dotnet pack --no-restore --no-build -c Release -p:Version=$version -o out
+        run: dotnet pack --no-restore --no-build -c Release -p:Version=$version -p:PackageOutputPath=out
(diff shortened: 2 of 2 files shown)


sungam3r:
https://github.com/dotnet/sdk/issues/30624#issuecomment-1432203481

sungam3r:
Maybe just fix SDK with 7.0.102 instead 7.0.x?

Shane32:
This fix is permanent. Downgrading just means we need to fix it again later.

sungam3r:
https://github.com/dotnet/sdk/issues/30624#issuecomment-1435040613

sungam3r:
> We are working on a 7.0.201 hotfix release.  After internal discussions, we've decided that waiting until next month's servicing release would be too long to wait to provide relief to customers. We're hoping to release the hotfix next week

https://github.com/dotnet/sdk/issues/30624#issuecomment-1435408741

sungam3r:
изображение
Did you check that pack produces packages in the common directory after switching to the new key?

Shane32:
I tested it locally. It works. In the snippet above the out folder is a relative path and the working folder is src. 

If they are going to release a hot fix then maybe we just wait a week. But we can’t release 7.3 until then and no beta builds will occur.

sungam3r:
> If they are going to release a hot fix then maybe we just wait a week. 

I would do so.

> But we can’t release 7.3 until then and no beta builds will occur.

I understand. Nevertheless, I do not like that the preview version of the compiler is used. I faced other errors when assembling other projects. Who knows what surprises the preview version contains? Perhaps we just will get broken binary. I would wait for the official fix so as not to guess about quality of `17.5.0-preview-23061-01+040e2a90e`.

sungam3r:
https://github.com/dotnet/sdk/issues/30625#issuecomment-1445175355

Shane32:
I suggest we merge this fix.  It's been a week and we still cannot run CI scripts properly.  This PR should not have any negative consequences.  We can always revert it if MS fixes whatever's wrong.

> Nevertheless, I do not like that the preview version of the compiler is used. I faced other errors when assembling other projects. Who knows what surprises the preview version contains? Perhaps we just will get broken binary. 

MS has stated that this is merely cosmetic, and that a non-preview version number will be reflected in 7.0.202.  It is not an actual 'preview' version in that sense of the word.

Also, according to reports, pinning a prior version doesn't fix the problem because the latest installed MSBuild version will be used regardless, and 7.0.200 is preinstalled on the current version of the GitHub Actions runner.  It would be necessary to manually remove the 7.0.200 SDK from the machine.  (I did not test this.)

Shane32:
> I tested it locally. It works. 

Apparently it does not work the same locally as here 👎 .  Now I need to patch it again.

PR #3467, opened 2022-12-30
Add API for OpenTelemetry

Shane32:
Uses the System.Diagnostics.Activity API present in .NET 5+ to provide telemetry data to OpenTelemetry and/or other providers.

The code is agnostic but favors OpenTelemetry recommendations for recorded tags and data.

See:
- https://learn.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-instrumentation-walkthroughs
- https://opentelemetry.io/docs/instrumentation/net/getting-started/
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/graphql.md

Sample code:

```cs
var builder = WebApplication.CreateBuilder(args);

var serviceName = "MyCompany.MyProduct.MyService";
var serviceVersion = "1.0.0";

// Configure important OpenTelemetry settings, the console exporter, and instrumentation library
builder.Services.AddOpenTelemetry().WithTracing(b => b
    .AddSource(serviceName)
    .SetResourceBuilder(
        ResourceBuilder.CreateDefault()
            .AddService(serviceName: serviceName, serviceVersion: serviceVersion))
    .AddHttpClientInstrumentation()
    .AddAspNetCoreInstrumentation()
    .AddSqlClientInstrumentation()
    .AddConsoleExporter(o => o.Targets = ConsoleExporterOutputTargets.Debug))
    .StartWithHost();

builder.Services.AddGraphQL(b => b
    .AddAutoSchema<Query>()
    .AddSystemTextJson()
    .UseTelemetry()  // <-- added builder method here
);
```

PR #3539, opened 2023-02-19
Introduce EventSource

### LINKED ISSUE #3540: Catch exceptions for telemetry
By the way , see this https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/README.md#filter

> The Filter receives the HttpContext of the incoming request, and does not collect telemetry about the request if the Filter returns false **or throws exception**.

And somewhere in the OTel spec I've read that implementation should be forgiving for errors inside telemetry internals so not break apps. Given such example from MS `OpenTelemetry.Instrumentation.AspNetCore.Implementation.HttpInListener` sources
```c#
            // enrich Activity from payload only if sampling decision
            // is favorable.
            if (activity.IsAllDataRequested)
            {
                try
                {
                    if (this.options.Filter?.Invoke(context) == false)
                    {
                        AspNetCoreInstrumentationEventSource.Log.RequestIsFilteredOut(nameof(HttpInListener), nameof(this.OnStartActivity), activity.OperationName);
                        activity.IsAllDataRequested = false;
                        activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
                        return;
                    }
                }
                catch (Exception ex)
                {
                    AspNetCoreInstrumentationEventSource.Log.RequestFilterException(nameof(HttpInListener), nameof(this.OnStartActivity), activity.OperationName, ex);
                    activity.IsAllDataRequested = false;
                    activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
                    return;
                }
```

I suggest to do the same - wrap filter in try/catch. See #3539. We can go this way to write messages in case of exceptions, etc.

_Originally posted by @sungam3r in https://github.com/graphql-dotnet/graphql-dotnet/issues/3467#issuecomment-1435998129_

sungam3r:
Helps with #3467
fixes #3540

https://learn.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-getting-started

No ifdefs, no additional dependencies. Supported by all TFMs. Lightweight and cross-platform.

Trace (actually _events_) captured by `dotnet-trace collect --providers GraphQL -p 22316`:

изображение


src/GraphQL/GraphQLEventSource.cs
```diff
@@ -0,0 +1,49 @@
...
+
+/// <summary>
+/// EventSource events emitted from the project.
+/// </summary>
+[EventSource(Name = "GraphQL")]
+internal class GraphQLEventSource : EventSource
+{
+    public static readonly GraphQLEventSource Log = new();
+
+    [NonEvent]
```
sungam3r, line 14: https://stackoverflow.com/questions/27058931/why-the-restriction-on-parameter-type-and-count-on-eventsource-methods

sungam3r: message	ERROR: Exception in Command Processing for EventSource GraphQL: Unsupported type Schema in event source.

sungam3r:
@Shane32 I think `EventSource` may be a sort of _base_ tool to instrument GraphQL.NET. It's like logs (MEL) but more for more specific use cases - profiling.

sungam3r:
Also by design it's 100% internal thing, so we can change everything.

src/GraphQL/GraphQLEventSource.cs
Shane32: Why don't we either (a) implement it like `SchemaCreated(Schema schema) => SchemaCreated(schema.GetType().Name);` or (b) add an extension method that does the same?  From the little I've read here, I don't think we should add a method marked `[NonEvent]` and then have it write an event.

sungam3r: Just an example of doing some calculations. Actually both events were taken from my head without any analysis. I'm fine to add anything else to track GraphQL internals.

sungam3r: I mean that I of course can pass `schema.GetType().Name` from the caller and remove two overloads.

Shane32: That is also acceptable.

Shane32:
I need to review this PR in more detail and read up on EventSource, etc.

sungam3r:
For me, this is a new topic as well, I postponed the reading for a long time. The first thing to do - think about those events that we **want** to emit.

Shane32:
At the moment, I'm not sure I want ANY events written to my Windows Event Log, which I believe would be the default behavior.  I have not tested it yet.

sungam3r:
I think Windows Event Log does not work without some additional settings. `EventSource` is an abstraction for emitting events to any consumer. 

Quick search results - https://stackoverflow.com/questions/53791076/eventsource-doesnt-write-logs-in-windows-event-viewer

sungam3r:
Yes. 

> `EventSource` is excellent if you want tighter control over ETW or EventPipe integration

> but for **general purpose logging**, `ILogger` is more flexible and easier to use

My intent here is not general purpose logging but something more focused on metrics/counters. Moreover I see that MS/others use `EventSource` in OTel related code.

sungam3r:
I updated PR, making it closer to real needs.

src/GraphQL/GraphQLEventSource.cs
sungam3r: I removed overloads.

src/GraphQL/GraphQLEventSource.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        WriteEvent(1);
+    }
+
+    [NonEvent]
+    public void RequestFilterException(Exception ex)
+    {
+        if (IsEnabled(EventLevel.Error, EventKeywords.All))
+        {
+            RequestFilterException(ex.ToInvariantString());
```
sungam3r, line 45: > Call [IsEnabled()](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource.isenabled#system-diagnostics-tracing-eventsource-isenabled) before performing any resource intensive work related to firing an event, such as computing an expensive event argument that won't be needed if the event is disabled.

`ex.ToInvariantString()` is considered as _intensive work_.

sungam3r: See `OpenTelemetry.Instrumentation.AspNetCore.Implementation.AspNetCoreInstrumentationEventSource`

src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
```diff
@@ -50,10 +50,21 @@ public GraphQLTelemetryProvider(GraphQLTelemetryOptions options)
     /// <inheritdoc/>
     public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
     {
-        if (!_telemetryOptions.Filter(options))
+        try
+        {
+            if (_telemetryOptions.Filter?.Invoke(options) == false)
```
sungam3r, line 55: The same approach here as in `OpenTelemetry.Instrumentation.AspNetCore.Implementation.HttpInListener`
изображение

Shane32:
https://learn.microsoft.com/en-us/dotnet/core/diagnostics/compare-metric-apis

Shane32:
Just a quick thought: I think we have to consider the fact that the current implementation within MS's libraries are based on older technology simply because their library is older.  This does not necessarily mean it is currently recommended.

Secondly: I think we should try to prevent overlap if possible.  No need to write to every historical logger / metric codebase solely because it exists, especially if the code is on-by-default.

Let's consider what would be desired for a modern library today and go from that angle.  Note: these are just some initial thoughts; I still need to read up on this stuff.

Shane32:
> I updated PR, making it closer to real needs.

How so?  All it tracks are schema created and initialized events, which typically only occur once per application lifecycle!  So far this PR seems pointless.

> My intent here is not general purpose logging but something more focused on metrics/counters. 

Those events are certainly not metrics.

Shane32:
Also this is not configurable as designed.  Within #3467 we implemented it as opt-in so it can be configured or left disabled.  Here it's built-in.  Why the difference?

sungam3r:
> How so? All it tracks are schema created and initialized events, which typically only occur once per application lifecycle

sungam3r:
> Just a quick thought: I think we have to consider the fact that the current implementation within MS's libraries are based on older technology simply because their library is older. This does not necessarily mean it is currently recommended.
> 
> Secondly: I think we should try to prevent overlap if possible. No need to write to every historical logger / metric codebase solely because it exists, especially if the code is on-by-default.
> 
> Let's consider what would be desired for a modern library today and go from that angle. Note: these are just some initial thoughts; I still need to read up on this stuff.

I'm fine either way. In case of modern lib it will cost us one additional conditional dependency. In case of "old" EventSource it costs nothing.

Shane32:
I don't get it.  Perhaps in the PR description you should state what this PR's intent is and/or what it does.  I certainly would not add `EventSource` just to log exceptions from bad code within `UseTelemetry` options...????

Shane32:
I'm so pressed for time I may not have read this PR fully.

sungam3r:
> Also this is not configurable as designed. Within https://github.com/graphql-dotnet/graphql-dotnet/pull/3467 we implemented it as opt-in so it can be configured or left disabled. Here it's built-in. Why the difference?

Yes. I look at `EventSource` as a built-in feature. This is a more general API than OpenTelemetry and does not require additional dependencies and overhead.

 I would like GraphQL.NET  to provide event counters out of the box. For example, see how it was done here https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1027 . To implement counters `EventSource` first should be created so I went this way. In addition, as you have seen, `EventSource` can come in handy for sending events (something like a `ILogger`, but not so general-purpose) in case of exceptional situations around OTel filtering and other stuff.

I found `dotnet-counters` tool rather usefull (and other .NET monitoring tools too) so the overall plan is:
1. Add `EventSource` as an internal entity into GrpahQL.NET
2. Fill it with events (1-5-10). It will take time. For now I see single event for filtration in `GraphQLTelemetryProvider`. Of course we can find more cases over time.
3. Fill it with counters. For example:
   - total resolvers executed
   - current executions
   - total dataloaders executed
   - current dataloaders
   - ExecutionNode creation rate
   - whatever else we want to count! 

I would say that I don't look at EventSource/EventCounters as an "old" API that should be avoided. It just works and it costs nothing in terms of performance.

> I certainly would not add EventSource just to log exceptions from bad code within UseTelemetry options..

Yep. Not **just** to log exceptions. It so happened that I presented not the main application of this tool. Its main application for me is counters. I can continue working on counters in this PR or in new one - I'm fine either way.

So the questions so far are:

1. Are you OK with such a feature as _counters_ as a built-in feature of the whole lib?
2. Are you OK with `EventSource/EventCounters` as implementation?

sungam3r:
@Shane32 Just googled it, see this https://github.com/dotnet/runtime/search?q=PollingCounter for reference. These are examples of counters/events in .NET runtime.

sungam3r:
@Shane32 I enabled this setting
изображение
and now there are such options
изображение

Shane32:
If we want built-in counters and etc, perhaps we should consider adding a dependency on [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource) for v8 and using the most modern framework rather than using an outdated pattern.

PR #3541, opened 2023-02-20
Check Activity.Status for NET5

sungam3r:


Shane32:
This PR makes no sense.

image

These are just tests, after all.

sungam3r:
> This PR makes no sense

Come on!  Old code checks status only for net6+. New code checks status also for net5. Does it really make no sense?

Shane32 (approved):
This looks like overkill to me.  Sure, it's just tests.

PR #3543, opened 2023-02-21
Require query type on schemas

Shane32:
Adds a global switch `RequireRootQueryType` to enable the new schema check, which defaults to `true`.  The global switch is marked obsolete and can be removed in v9.

Tests were added to test the schema check and that it can be disabled.


Shane32:
@sungam3r  FYI, I'm fine with or without the global switch.  It's a couple lines of code to fix any schema to comply with the new requirement, and about just as many to bypass it (setting the global switch and ignoring the build warning).

docs2/site/docs/migrations/migration8.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Breaking Changes
+
+### 1. Query type is required
+
+Pursuant to the GraphQL specification, a query type is required for any schema.
+This is enforced during schema validation but may be bypassed as follows:
+
+```csharp
+GlobalSwitches.AllowSchemaWithoutQuery = true;
```
sungam3r, line 16:
```suggestion
GlobalSwitches.RequireRootQueryType = false;
```

docs2/site/docs/migrations/migration8.md
```diff
@@ -0,0 +1,20 @@
...
+```
+
+Future versions of GraphQL.NET will not contain this property.
```
sungam3r, line 19:
```suggestion
Future versions of GraphQL.NET will not contain this property and each schema will always be required to have root Query type to comply with the GraphQL specification.
```

sungam3r (approved):

PR #3544, opened 2023-02-21
Mark IsConcrete as obsolete

Shane32:


src/GraphQL/Extensions/TypeExtensions.cs
```diff
@@ -19,6 +19,7 @@ public static class TypeExtensions
         /// <returns>
         ///   <see langword="true"/> if the specified type is neither abstract nor an interface; otherwise, <see langword="false"/>.
         /// </returns>
+        [Obsolete("This method will be removed in a future version of GraphQL.NET.")]
         public static bool IsConcrete(this Type type)
```
Shane32, line 23: This method is not used by GraphQL.NET.  Moreover, the logic is questionable.  `IsAbstract` already checks if the type is an interface, and this method does not check if it is a generic type definition.  So this method currently is equivalent to `!type.IsAbstract` and does nothing else.

> The [IsAbstract](https://learn.microsoft.com/en-us/dotnet/api/system.type.isabstract?view=netstandard-2.0) property returns true in the following cases:

>
> - The current type is abstract; that is, it cannot be instantiated, but can only serve as the base class for derived classes. In C#, abstract classes are marked with the [abstract](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/abstract) keyword; in F#, they are marked with the [AbstractClass](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/abstract-classes) attribute; in Visual Basic, they are marked with the [MustInherit](https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/modifiers/mustinherit) keyword.

>
> - The current type is an interface.

Shane32: Testing shows that `IsAbstract` returns `false` for `typeof(List<>)`, by the way.

sungam3r (approved):

sungam3r:
I' юm now in the hospital after a planned operation, so I can only answer simple letters for 1-2 days.

Shane32:
> I'm now in the hospital after a planned operation, so I can only answer simple letters for 1-2 days.

Hopefully all went well, and you recover quickly.

PR #3547, opened 2023-02-21
Fix execution context for subscription resolvers

### LINKED ISSUE #3546: Scoped dependencies for messages initiated on the server side?
## Summary
Do we have a DI **scope** bound to the web socket so we can resolve dependencies for sending server messages into the subscription using _that scope_? 

## Relevant information
I have some dependencies that rely on HttpContext so we can get things like "host", "currentUser" etc. Those are resolved ok on the subscription requests. Understandably, that we can't have a HttpContext when we fire some events from the server. 
I hoped that IResolveFieldContext.RequestServices are resolved in the scope of subscription requests and we can safely get, for example, `CurrentUser`. 
For sure, relying on the `HttpContext` may be wrong in the first place, but I'm just trying to describe the whole picture and the reasons I'm facing.
I'm sure there's a way. I'd like to ask, how to solve it correctly.

### Environment (if relevant)
GraphQL 7.2.2
.NET 6

Shane32:
Fixes #3546 

This PR captures the System.Threading.ExecutionContext during subscription connection request (specifically, when `ExecutionResult.Streams[0].Value.Subscribe()` is called), and then restores it during processing of field resolvers.  This allows code that relies on `AsyncLocal` such as `HttpContextAccessor` to access the client's execution context rather than that of the event source.  Effectively, a field resolver which examines `IHttpContextAccessor.HttpContext` will now return the proper `HttpContext` of the client rather than that of the event source, as would be expected and would work for any other field resolver.

This is classified as a bug because security-related code which relies on `IHttpContextAccessor.HttpClient.User` would return the wrong `ClaimsPrincipal` being that of the event source's user (if the event source was initiated by another HTTP request) rather than the client.  This is especially important as graph types are often not specific to queries, mutations or subscriptions, and it would be expected for `IHttpContextAccessor` to work the same within a specific graph type, regardless of whether it was a query, mutation or subscription.

@sungam3r When you get a chance, please review; let's merge this asap.

See also:
- #3546 
- https://github.com/graphql-dotnet/server/pull/991


src/GraphQL.Tests/Subscription/SubscriptionExecutionStrategyTests.cs
```diff
@@ -492,6 +492,43 @@ public async Task MultipleSubscriptions_NoValidation()
         result.Executed.ShouldBeTrue();
     }
 
+    private static readonly AsyncLocal<string?> _asyncLocal = new();
+
+    [Fact]
+    public async Task NetExecutionContextIsCarriedOver()
+    {
+        // This verifies that the System.Thread.ExecutionContext is preserved during data events after the initial subscription.
```
Shane32, line 500:
```suggestion
        // This verifies that the System.Threading.ExecutionContext is preserved during data events after the initial subscription.
```

src/GraphQL.Tests/Subscription/SubscriptionExecutionStrategyTests.cs
```diff
@@ -492,6 +492,43 @@ public async Task MultipleSubscriptions_NoValidation()
...
 
+    private static readonly AsyncLocal<string?> _asyncLocal = new();
+
+    [Fact]
+    public async Task NetExecutionContextIsCarriedOver()
+    {
+        // This verifies that the System.Threading.ExecutionContext is preserved during data events after the initial subscription.
+        // This means that AsyncLocal instances will contain the values they had during the initial subscription during the
+        // execution of field resolvers. As HttpContextAccessor is based on AsyncLocal, this allows subscription field resolvers
+        // that references IHttpContextAccessor.HttpContext to access the correct HttpContext of the connected client, rather
```
Shane32, line 503:
```suggestion
        // that reference IHttpContextAccessor.HttpContext to access the correct HttpContext of the connected client, rather
```

Shane32:
@sungam3r I hope you’re doing well. Perhaps you could review this PR this morning. Thanks!

Shane32:
I'll be merging this tomorrow unless there are objections.

sungam3r:
OK. So far I do not check the mail every day. As usual, I will check all the changes as soon as I get to them.

PR #3550, opened 2023-02-24
Add schema validation visitor for AOT scenarios

Shane32:
This is the approach taken by MS for .NET 8 in some scenarios.

Note that `RuntimeFeature.IsDynamicCodeSupported` returns `true` while debugging.

In this PR, attempting to run the AOT sample produces this error when running the compiled app:

```
Sample of AOT compilation of a GraphQL query

Executing request: { hero { id name } }

Unhandled Exception: System.InvalidOperationException: NameFieldResolver is not supported in scenarios when dynamic compilation is not supported by the runtime. Please fix CharacterInterfaceConnection.totalCount.
   at GraphQL.Utilities.AotValidationVisitor.VisitObjectFieldDefinition(FieldType, IObjectGraphType, ISchema) + 0x348
   at GraphQL.SchemaExtensions.Run(ISchemaNodeVisitor, ISchema) + 0x861
   at GraphQL.Types.Schema.Validate() + 0x9c
   at GraphQL.Types.Schema.CreateAndInitializeSchemaTypes() + 0x16f
   at GraphQL.Types.Schema.Initialize() + 0x13e
```

This is correct.  In fact, disabling the visitor and attempting to access the property referenced by the above error produces this runtime error:

```
Sample of AOT compilation of a GraphQL query

Executing request: { hero { id name friendsConnection { totalCount } } }

Unhandled Exception: System.InvalidOperationException: Expected to find property or method 'totalCount' on type 'Connection`1' but it does not exist.
   at GraphQL.Resolvers.NameFieldResolver.CreateResolver(Type, String) + 0x2a0
   at GraphQL.Resolvers.NameFieldResolver.<>c.<Resolve>b__6_0(ValueTuple`2 t) + 0x2d
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey, Func`2) + 0x1c8
   at GraphQL.Resolvers.NameFieldResolver.Resolve(IResolveFieldContext, String) + 0x1d3
   at GraphQL.Resolvers.NameFieldResolver.ResolveAsync(IResolveFieldContext context) + 0x61
```

So we can see that the schema validation visitor added in this PR is doing its job correctly by identifying incorrectly configured fields within the schema.

### DIFF
--- a/src/GraphQL/Extensions/ObjectExtensions.cs
+++ b/src/GraphQL/Extensions/ObjectExtensions.cs
@@ -19,6 +19,7 @@ namespace GraphQL
         /// <typeparam name="T">The type to create.</typeparam>
         /// <param name="source">The source of values.</param>
         /// <returns>T.</returns>
+        [RequiresUnreferencedCode("The type of the object to be converted to must be statically discoverable.")]
         public static T ToObject<T>(this IDictionary<string, object?> source)
             where T : class
             => (T)ToObject(source, typeof(T));
@@ -38,6 +39,7 @@ namespace GraphQL
         /// In case of configuring field as Field(x => x.FName).Name("FirstName") source dictionary
         /// will have 'FirstName' key but its value should be set to 'FName' property of created object.
         /// </param>
+        [RequiresUnreferencedCode("The type of the object to be converted to must be statically discoverable.")]
         public static object ToObject(this IDictionary<string, object?> source, Type type, IGraphType? mappedType = null)
         {
             // Given Field(x => x.FName).Name("FirstName") and key == "FirstName" returns "FName"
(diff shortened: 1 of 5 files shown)


Shane32:
@sungam3r This code can't be merged because of the way connections are constructed with GraphQL.NET 7.  I was going to try to add a little code to remove or modify the connection field, but am hampered at every turn:
- Fields cannot be removed from graphs because we never exposed a `Remove` method on `TypeFields` or `ComplexGraphType`.
- A class derived from `ConnectionType<TNodeType, TEdgeType>` cannot be registered within DI, as code within `SchemaTypes` directly instantiates connection types, `PageInfo`, and built-in scalars.
- Connection types should use an interface or similar to directly access the proper fields, rather than using `NameFieldResolver`

I propose making some or all of the above changes in GraphQL.NET 8.  Of course we can independently review each change as it proposed, but I just wanted your input before I started creating a number of issues/PRs.

On a separate note, graphql-wg has indicated that the proposal for oneOf input types will not be merged to the spec until it has been implemented, even though the review process is essentially complete.  I'm not sure if this is 'putting the cart before the horse', but in any case, I propose we split the inheritance of input types from output types for GraphQL.NET 8, and within or after the course of this change, add oneOf input types as an 'experimental' feature.  Once we can demonstrate support for oneOf input types within GraphQL.NET, it will be one step closer to merging into the official spec.

PR #3562, opened 2023-03-07
Use MinVer

sungam3r:
See https://github.com/adamralph/minver

It allows to not bother with setting version manually in code/csproj. I tested it in 4 repos. So far it works.


sungam3r:
https://stackoverflow.com/questions/75537987/how-do-i-get-the-github-actions-section-in-the-visual-studio-solution-view

src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>7.3.0-preview</VersionPrefix>
+    <MinVerDefaultPreReleaseIdentifiers>preview</MinVerDefaultPreReleaseIdentifiers>
```
sungam3r, line 4: https://github.com/adamralph/minver#can-i-change-the-default-pre-release-identifiers-from-alpha0-to-something-else

.github/workflows/build-artifacts-code.yml
```diff
@@ -21,6 +21,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
sungam3r, line 25: MinVer need tags to work. It's a known issue with `actions/checkout` - this action has no explicit option to fetch tags. `fetch-depth: 0` effectivelty fetches all tags. Yep, it fetches all commits as well 😞 .

Shane32:
I don't favor this.  First, I don't like having to retrieve the entire commit history, further delaying test runs.  But even more importantly, I don't like introducing another build-time dependency when it is not necessary.  We already have too many dependencies here that constantly "need" updating.  It also makes it more difficult to understand the versioning process by looking at the csproj and workflow files, whereas now it is clear and precise.

What are the benefits?  The numbering for dev builds is more "accurate"?  Does not matter; they require quite a lot of work to use, and the few users there are (if any) should not care.

So in summary, I see this as something that increases, not decreases, the maintenance required for GraphQL.NET.  As an open-source project maintained by only two of us, both of which are extremely pressed for time, I cannot endorse it.

Shane32:
Another option is to leave the version in directory.build.props at 20.0.0-preview. It will always generate a newer version than the latest published version. (This only matters if using another published dependency in conjunction with a dev build.)

Shane32:
If I have time perhaps I will test MinVer in one of my own repos

sungam3r:
If I would choose between some kind of fictitious version and the absense of version in sources, then I will prefer the latter.

sungam3r:
> If I have time perhaps I will test MinVer in one of my own repos

👍 Honestly it took me a couple of **years** to finally try, I have long known about this package and saw its use in other famous repositories.

.github/workflows/build-artifacts-code.yml
sungam3r: fixed in https://github.com/actions/checkout/pull/579

Shane32:
I think what we have already works perfectly.

Shane32:
I'll close it for now then.  We can reopen if we wish to discuss further.

PR #3568, opened 2023-03-30
Fix UnhandledError when resolving top-level subscription field

sungam3r:
Also see #3213

Yesterday one of our team members found a bug when using field middleware and calling subscription.

Before fix:
```
    Shouldly.ShouldAssertException : errors
        should be null but was
    [GraphQL.Execution.UnhandledError: Error trying to resolve field 'newMessageContent'.
     ---> System.InvalidOperationException: Expected to find property or method 'newMessageContent' on type 'String' but it does not exist.
       at GraphQL.Resolvers.NameFieldResolver.CreateResolver(Type target, String name) in C:\_GITHUB_\graphql-dotnet\src\GraphQL\Resolvers\NameFieldResolver.cs:line 83
       at GraphQL.Resolvers.NameFieldResolver.<>c.<Resolve>b__6_0(ValueTuple`2 t) in C:\_GITHUB_\graphql-dotnet\src\GraphQL\Resolvers\NameFieldResolver.cs:line 38
       at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
       at GraphQL.Resolvers.NameFieldResolver.Resolve(IResolveFieldContext context, String name) in C:\_GITHUB_\graphql-dotnet\src\GraphQL\Resolvers\NameFieldResolver.cs:line 38
       at GraphQL.Resolvers.NameFieldResolver.ResolveAsync(IResolveFieldContext context) in C:\_GITHUB_\graphql-dotnet\src\GraphQL\Resolvers\NameFieldResolver.cs:line 28
       at GraphQL.Tests.Subscription.SubscriptionTests.MyMiddleware.ResolveAsync(IResolveFieldContext context, FieldMiddlewareDelegate next) in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Subscription\SubscriptionTests.cs:line 58
       at GraphQL.Tests.Subscription.SubscriptionTests.<>c__DisplayClass3_1.<SubscribeToContent>b__1(IResolveFieldContext context) in C:\_GITHUB_\graphql-dotnet\src\GraphQL.Tests\Subscription\SubscriptionTests.cs:line 84
       at GraphQL.Resolvers.FuncFieldResolver`1.ResolveAsync(IResolveFieldContext context) in C:\_GITHUB_\graphql-dotnet\src\GraphQL\Resolvers\FuncFieldResolver.cs:line 40
       at GraphQL.Execution.ExecutionStrategy.ExecuteNodeAsync(ExecutionContext context, ExecutionNode node) in C:\_GITHUB_\graphql-dotnet\src\GraphQL\Execution\ExecutionStrategy.cs:line 477
       --- End of inner exception stack trace ---]
```

`ApplyMiddleware` method sets `NameFieldResolver.Instance` for all fields, even for subscriptions. I have no ideas why we haven't found it before.

@Shane32 I need your help. I do not like current implementation. The main problem is `complex == schema?.Subscription` comparison. `SchemaTypes` has no reference to schema but in fact only `Schema` calls all methods on `SchemaTypes`, so I just added two methods and delegated work to them. I would like to change design - either adding reference to `Schema` or in some other way.

### DIFF
--- a/src/GraphQL/Types/Schema.cs
+++ b/src/GraphQL/Types/Schema.cs
@@ -451,7 +451,7 @@ namespace GraphQL.Types
             {
                 // At this point, Initialized will return false, and Initialize will still lock while waiting for initialization to complete.
                 // However, AllTypes and similar properties will return a reference to SchemaTypes without waiting for a lock.
-                _allTypes.ApplyMiddleware(FieldMiddleware);
+                _allTypes.ApplyMiddleware(FieldMiddleware, this);
 
                 foreach (var visitor in GetVisitors())
                     visitor.Run(this);
--- a/src/GraphQL/Types/Schema.cs
+++ b/src/GraphQL/Types/Schema.cs
@@ -451,7 +451,7 @@ namespace GraphQL.Types
             {
                 // At this point, Initialized will return false, and Initialize will still lock while waiting for initialization to complete.
                 // However, AllTypes and similar properties will return a reference to SchemaTypes without waiting for a lock.
-                _allTypes.ApplyMiddleware(FieldMiddleware, this);
+                _allTypes.ApplyMiddleware(FieldMiddleware);
 
                 foreach (var visitor in GetVisitors())
                     visitor.Run(this);
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -5,6 +5,7 @@
     <PackageReleaseNotes>See https://github.com/graphql-dotnet/graphql-dotnet/releases and https://graphql-dotnet.github.io/docs/migrations/migration7</PackageReleaseNotes>
     <Nullable>enable</Nullable>
     <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
+    <!--<EnableAotAnalyzer>true</EnableAotAnalyzer> PublishAot property enables it-->
     <!--<ILLinkTreatWarningsAsErrors>false</ILLinkTreatWarningsAsErrors> does not work-->
     <WarningsNotAsErrors>$(WarningsNotAsErrors);IL2026;IL2055;IL2067;IL2070;IL2072;IL2075;IL2090;IL2091;IL2092;IL2095</WarningsNotAsErrors>
   </PropertyGroup>
--- a/src/GraphQL.Tests/Subscription/SubscriptionTests.cs
+++ b/src/GraphQL.Tests/Subscription/SubscriptionTests.cs
@@ -1,4 +1,5 @@
 using System.Reactive.Linq;
+using GraphQL.Instrumentation;
 
 namespace GraphQL.Tests.Subscription;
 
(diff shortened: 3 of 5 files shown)


Shane32:
How about this?

```cs
var inner = field.Resolver ?? (field.StreamResolver != null) ? SourceFieldResolver.Instance : NameFieldResolver.Instance);
```

Shane32:
> `ApplyMiddleware` method sets `NameFieldResolver.Instance` for all fields, even for subscriptions. I have no ideas why we haven't found it before.

I believe the schema-first and auto-registering code both set the Resolver property correctly.  It is only code-first that would have an issue.

Now, it is not mandatory that the SourceFieldResolver is used.  The event data from the IObservable is passed in as the root value at SubscriptionExecutionStrategy.cs:182.  However, setting a custom resolver could apply a transform to this data.  It is much more intuitive to apply the transform to the IObservable, and as far as I know this is how subscriptions are typically used.  NameFieldResolver would never be used.

Also if we look at the `FieldSubscribe` and `FieldSubscribeAsync` code, we note that we have a todo comment to remove the resolver.  We should probably remove it.  Perhaps for now we should set it to the proper default if it is null (although you are doing this through the middleware code).  People can manually set the Resolver property if they need to override SourceFieldResolver.Instance, which I believe would be rare.

Also within those same two methods, note that the `resolve` is typed as `Func<IResolveFieldContext<TSourceType>, object?>`.  Specifically consider `TSourceType`.  This is most definitely incorrect; it would match the `IObservable` event data type, not the type of the graph type's source on which the field was added.  Since few people use the ExecutionOptions.Root property, and hence are unlikely to strongly type their root Subcription type, TSourceType is probably almost always `object`, matching `IObservable<object>` used here.

src/GraphQL/Types/Schema.cs
```diff
@@ -451,7 +451,7 @@ IEnumerable<ISchemaNodeVisitor> GetVisitors()
             {
                 // At this point, Initialized will return false, and Initialize will still lock while waiting for initialization to complete.
                 // However, AllTypes and similar properties will return a reference to SchemaTypes without waiting for a lock.
-                _allTypes.ApplyMiddleware(FieldMiddleware);
+                _allTypes.ApplyMiddleware(FieldMiddleware, this);
```
Shane32, line 454: I don't mind the added method, but I dislike that we are calling an internal method.  Better to mark the old method as obsolete and expose the new method publicly.

sungam3r: I reverted it.

sungam3r:
Thanks. I would like to publish v7.3.1, OK?

Shane32 (approved):

Shane32:
> Thanks. I would like to publish v7.3.1, OK?

Sure

PR #3569, opened 2023-03-30
Set FieldType.Resolver properly when calling a field builder's ResolveStream methods

Shane32:
See:
- #3568 

I'd consider this a bug with the design of the new field builder methods.


sungam3r:
Hah, interesting addition. But it does not replace #3568, right?

sungam3r (approved):

Shane32:
> Hah, interesting addition. But it does not replace #3568, right?

Correct

PR #3571, opened 2023-04-03
Allow deprecation of input values

### LINKED ISSUE #1112: Deprecating a field on an InputObjectGraphType does not reflect on the Introspection query response.
## Description

This issue is related to the below issues:

https://github.com/graphql/graphql-spec/issues/197 (open)

The Introspection query response does not include the `isDeprecated` and `deprecationReason` fields for `InputFields` marked as deprecated.

### Steps to reproduce

This is a simple mutation in test:

                FieldAsync<SampleResponse>(
                    "SampleMutation",
                        new QueryArgument<SampleInput>
                        {
                            Name = "Input"
                        }
                    ),
                    resolve: async context =>
                    {
                        return null;
                    }
                );

The below is the query argument type with a list of integers that is marked as deprecated.

```
    public class SampleInput : InputObjectGraphType
    {
        public SampleInput()
        {
            Name = "SampleInput";
           Field<ListGraphType<IntGraphType>>("someIntList", deprecationReason: "Deprecated").
        }
   }
```

The introspection query fired by the GraphQL client does not include the `isDeprecated` and `deprecationReason` for  `InputFields` (Looks like this is not even supported).

```
query IntrospectionQuery {
  __schema {
    queryType {
      name
    }
    mutationType {
      name
    }
    subscriptionType {
      name
    }
    types {
      ...FullType
    }
    directives {
      name
      description
      locations
      args {
        ...InputValue
      }
    }
  }
}

fragment FullType on __Type {
  kind
  name
  description
  fields(includeDeprecated: true) {
    name
    description
    args {
      ...InputValue
    }
    type {
      ...TypeRef
    }
    isDeprecated
    deprecationReason
  }
  inputFields {
    ...InputValue
  }
  interfaces {
    ...TypeRef
  }
  enumValues(includeDeprecated: true) {
    name
    description
    isDeprecated
    deprecationReason
  }
  possibleTypes {
    ...TypeRef
  }
}

fragment InputValue on __InputValue {
  name
  description
  type {
    ...TypeRef
  }
  defaultValue
}

fragment TypeRef on __Type {
  kind
  name
  ofType {
    kind
    name
    ofType {
      kind
      name
      ofType {
        kind
        name
        ofType {
          kind
          name
          ofType {
            kind
            name
            ofType {
              kind
              name
              ofType {
                kind
                name
              }
            }
          }
        }
      }
    }
  }
}
```

### Expected result

The introspection query response should get the below subset of the response representing `sampleInput` and thus enabling to query on the `isDeprecated` and `deprecationReason` fields on  `InputFields`.

```
        {
          "kind": "INPUT_OBJECT",
          "name": "SampleInput",
          "description": null,
          "fields": null,
          "inputFields": [
            {
              "name": "SampleInput",
              "description": null,
              "isDeprecated": true,
              "deprecationReason": "Deprecated",
              "type": {
                "kind": "LIST",
                "name": null,
                "ofType": { "kind": "SCALAR", "name": "Int", "ofType": null }
              },
              "defaultValue": "null"
            }
          ],
          "interfaces": null,
          "enumValues": null,
          "possibleTypes": null
        }
```

### Actual result

But the actual subset of the response obtained from the server on the introspection query is as below (without the `isDeprecated` and `deprecationReason` fields.)

```
        {
          "kind": "INPUT_OBJECT",
          "name": "SampleInput",
          "description": null,
          "fields": null,
          "inputFields": [
            {
              "name": "SampleInput",
              "description": null,
              "type": {
                "kind": "LIST",
                "name": null,
                "ofType": { "kind": "SCALAR", "name": "Int", "ofType": null }
              },
              "defaultValue": "null"
            }
          ],
          "interfaces": null,
          "enumValues": null,
          "possibleTypes": null
        }
```

### Environment

.Net Core 2.1
GraphQL version 2.4.0
graphql-playground-react@1.7.10

### LINKED ISSUE #3570: Directive for Input fields
How can I mark input field as deprecated? If we have old not optimised filters by field and I add new filter field but I wanna mark old field as obsolete/deprecated. 

input FilterOperationSet {
  oldFilter: Boolean <-- this must be delete in future
  newFilter: Boolean
}
I trying use deprecated but @deprecated(reason: String = "No longer supported") on FIELD_DEFINITION | ENUM_VALUE and INPUT_FIELD_DEFINITION not supported.
Then I trying use custom directive with INPUT_FIELD_DEFINITION, but in schema I haven`t mark:
https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Utilities/SchemaPrinter.cs#L306
public string PrintInputValue(FieldType field)
{
    Schema?.Initialize();
    var argumentType = field.ResolvedType!;
    var description = $"{FormatDescription(field.Description, "  ")}  {field.Name}: {argumentType}";
    if (field.DefaultValue != null)
    {
        description += " = {0}".ToFormat(FormatDefaultValue(field.DefaultValue, argumentType));
    }
    return description;
}

sungam3r:
Conforms to https://github.com/graphql/graphql-spec/pull/525

fixes #1112
fixes #3570

In the second commit I want to push changes required by https://github.com/graphql/graphql-spec/pull/805


Shane32:
@sungam3r  I'm on board with your changes, and have confidence in your ability to implement them.  Confirm when you're ready for me to do final review.  (now?)

sungam3r:
I need to push second commit with changes in schema validation. I need some time to find tests.

src/GraphQL.Tests/Files/SDL/SchemaPrinter.graphql
```diff
@@ -0,0 +1,198 @@
+
+schema {
```
sungam3r, line 2: I moved it to file to use syntax highlighting.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -33,6 +33,7 @@ internal class TestObject
         [Description("Super secret")]
         public string? someString { get; set; }
         [Obsolete("Use someInt")]
+        [DefaultValue(true)]
```
sungam3r, line 36: Ooops
>  System.InvalidOperationException : The required input field 'someBoolean' of an Input Object 'TestObject' has not default value so `@deprecated` directive must not be applied to this input field. To deprecate an input field, it must first be made optional by either changing the type to nullable or adding a default value.

And here is an example of breaking changes that I talked about. Not tooling, just "bad-written" schema that was valid before and is not valid now.

src/GraphQL/Types/Collections/SchemaDirectives.cs
```diff
@@ -25,6 +25,9 @@ public class SchemaDirectives : IEnumerable<Directive>
         /// </summary>
         public virtual DeprecatedDirective Deprecated { get; } = new DeprecatedDirective();
 
+        // TODO: remove this one when the next version of the spec will be released
+        internal DeprecatedDirective DeprecatedDraft { get; } = new DeprecatedDirective(deprecationOfInputValues: true);
```
sungam3r, line 29: I do not like it and I will glad to remove it wit new spec release.

sungam3r:
@Shane32 Ready for review. There are not so many changes and they turned out to be quite simple.

src/GraphQL/Types/Collections/SchemaDirectives.cs
Shane32: Waiting for the GraphQL WG to merge anything is really quite hopeless.  It's very sad they do not explicitly state in the spec that graphql clients should not break when built-in types have properties they do not understand, or additional built-in types are present.  This prevents GraphQL itself from adding features to the spec as non-breaking changes.  Instead all the common tooling breaks for each feature like this because of this poor design decision.  It also causes problems when attempting to implement experimental features.  So they won't add features to the spec because nobody supports them, which is because the common tools are explicitly programmed to disallow experimental features.  It's a recursively bad loop.

sungam3r: Yep.

src/GraphQL/ExperimentalFeatures.cs
```diff
@@ -19,4 +19,12 @@ public class ExperimentalFeatures
...
     /// </summary>
     public bool RepeatableDirectives { get; set; } = false;
+
+    /// <summary>
+    /// Enables deprecation of input values - arguments on a field or input fields on an input type.
+    /// <br/><br/>
+    /// This feature is from a working draft of the specification (not experimental, just not released yet).
+    /// </summary>
+    public bool DeprecationOfInputValues { get; set; } = false;
+
```
Shane32, line 29: remove blank line?

src/GraphQL/Types/Collections/SchemaDirectives.cs
Shane32: I agree I don't like this design.  Can we make an internal method to overwrite the value of the public property during schema initialization or something?  Rather than exposing the 'wrong' directive when the experimental feature is enabled?  Honestly I'm not exactly sure how this class is used.

src/GraphQL/Types/Schema.cs
```diff
@@ -125,7 +125,7 @@ public Schema(IServiceProvider services, IEnumerable<IConfigureSchema> configura
             _services = services;
 
             Directives = new SchemaDirectives();
-            Directives.Register(Directives.Include, Directives.Skip, Directives.Deprecated);
+            Directives.Register(Directives.Include, Directives.Skip, Features.DeprecationOfInputValues ? Directives.DeprecatedDraft : Directives.Deprecated);
```
Shane32, line 128: Is `Features` a static property?

src/GraphQL/Types/Collections/SchemaDirectives.cs
Shane32: Maybe add a public constructor with the property, then in the Schema constructor, it passes the value from Features.whatever.  Then the RegisterDirectives call doesn't need the ternary

sungam3r: > Can we make an internal method to overwrite the value of the public property during schema initialization or something?

It would conflict in case of multiple schemas with different settings in single app.

Shane32: I don't think so.  These are not static properties and they are not in a static instance.

See here:

image

Just change it to

```cs
Directives = new SchemaDirectives(Features.DeprecationOfInputValues);
Directives.Register(Directives.Include, Directives.Skip, Directives.Deprecated);
```

image

However, I don't understand how the code works either with or without that change.  How can `Features.DeprecationOfInputValues` be changed prior to the code in `Schema`'s constructor be called?  `Features` is not static.

image

src/GraphQL.Tests/Files/SDL/SchemaPrinter.graphql
```diff
@@ -0,0 +1,198 @@
...
+  query: Root
+}
+
+# Marks an element of a GraphQL schema as no longer supported.
+directive @deprecated(
+  reason: String = "No longer supported"
+) on FIELD_DEFINITION | ENUM_VALUE
```
Shane32, line 9: This is incorrect.

src/GraphQL/Types/Collections/SchemaDirectives.cs
Shane32: You can see from the introspection result test that the wrong directive is being applied.

src/GraphQL/Types/Schema.cs
Shane32: Ok then it is impossible to initialize the proper directives with this code because `Features` cannot be accessed prior to the constructor being executed.  See comments above.

src/GraphQL/Types/Collections/SchemaDirectives.cs
sungam3r: Oops, my bad, fixed.

src/GraphQL/Types/Schema.cs
sungam3r: Fixed. Overlooked it.

src/GraphQL.Tests/Files/SDL/SchemaPrinter.graphql
sungam3r: fixed

Shane32 (approved):

PR #3574, opened 2023-04-04
Only root fields of subscriptions should set StreamResolver; only object output types should set Resolver

### LINKED ISSUE #1176: InputObjectGraphType resolve function ignored

### LINKED ISSUE #2994: Add schema validation that subscription resolvers are only set on valid fields
It's worth to add additional check into schema validation.

_Originally posted by @sungam3r2022 in https://github.com/graphql-dotnet/graphql-dotnet/pull/2993#r819915648_

sungam3r:
fixes #2994
also fixes #1176

@Shane32 I suggest to consider this as a "fix" for #1176. I do not believe that we will split `FieldType` for inputs and outputs or redesign API in some other way. At least I'm sure that it will be opened for years without any progress. Schema validation allows to find all invalid usages during initialization, i.e. early thought in runtime, not at compile time, so I think this is a good compromise and we do not need to break the whole design.

### DIFF
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -603,7 +603,7 @@ in version 7.2+.
 
 Starting with version 7.4, the `StreamResolver` property should be set exclusively for root fields
 in subscriptions, while the `Resolver` property should only be assigned to object output types'
-fields. Fields within union types, interface types and input object types should not set these
+fields. Fields within interface types and input object types should not set these
 properties. This modification may lead to potential disruptions in your applications, as the
 schema may trigger an exception during initialization. To eliminate this exception, simply
 cease assigning these properties for the fields specified in the exception message.
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -598,3 +598,10 @@ require them to resolve to GraphQL Object types. GraphQL.NET 7.2+ now enforces t
 Due to the above validation, resolvers defined on fields of interface types are never be used by
 GraphQL.NET. As such, resolvers are not generated for fields of `AutoRegisteringInterfaceGraphType`s
 in version 7.2+.
+
+### 18. `SchemaValidationVisitor` has more runtime checks
+
+From version 7.4 on, only root fields of subscriptions should set `StreamResolver` and only object
+output types should set `Resolver` properties. This change may be potentially breaking for your apps,
+because schema will throw exception during initialization. To get rid of that exception just stop
+setting these properties for fields mentioned in exception message.
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -601,7 +601,9 @@ in version 7.2+.
 
 ### 18. `SchemaValidationVisitor` has more runtime checks
 
-From version 7.4 on, only root fields of subscriptions should set `StreamResolver` and only object
-output types should set `Resolver` properties. This change may be potentially breaking for your apps,
-because schema will throw exception during initialization. To get rid of that exception just stop
-setting these properties for fields mentioned in exception message.
+Starting with version 7.4, the `StreamResolver` property should be set exclusively for root fields
+in subscriptions, while the `Resolver` property should only be assigned to object output types'
+fields. Fields within union types, interface types and input object types should not set these
+properties. This modification may lead to potential disruptions in your applications, as the
+schema may trigger an exception during initialization. To eliminate this exception, simply
+cease assigning these properties for the fields specified in the exception message.
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -469,9 +469,9 @@ namespace GraphQL.Types
 
             foreach (var item in Dictionary)
             {
-                if (item.Value is IComplexGraphType complex && item.Value is not IInputObjectGraphType)
+                if (item.Value is IObjectGraphType obj)
                 {
-                    foreach (var field in complex.Fields.List)
+                    foreach (var field in obj.Fields.List)
                     {
                         var inner = field.Resolver ?? (field.StreamResolver == null ? NameFieldResolver.Instance : SourceFieldResolver.Instance);
 
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -469,7 +469,7 @@ namespace GraphQL.Types
 
             foreach (var item in Dictionary)
             {
-                if (item.Value is IComplexGraphType complex)
+                if (item.Value is IComplexGraphType complex && item.Value is not IInputObjectGraphType)
                 {
                     foreach (var field in complex.Fields.List)
                     {
--- a/src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
+++ b/src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
@@ -9,7 +9,6 @@ namespace GraphQL.Types;
 /// </summary>
 internal static class AutoRegisteringOutputHelper
 {
-    private static readonly IFieldResolver _invalidFieldResolver = new FuncFieldResolver<object?>(_ => throw new InvalidOperationException("This field resolver should never be called. It is only used to prevent the default field resolver from being used."));
     /// <summary>
     /// Configures query arguments and a field resolver for the specified <see cref="FieldType"/>, overwriting
     /// any existing configuration within <see cref="FieldType.Arguments"/>, <see cref="FieldType.Resolver"/>
--- a/src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
+++ b/src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
@@ -10,7 +10,7 @@ namespace GraphQL.Types;
 internal static class AutoRegisteringOutputHelper
 {
     private static readonly IFieldResolver _invalidFieldResolver = new FuncFieldResolver<object?>(_ => throw new InvalidOperationException("This field resolver should never be called. It is only used to prevent the default field resolver from being used."));
-    private static readonly ISourceStreamResolver _invalidStreamResolver = new SourceStreamResolver<object?>((Func<IResolveFieldContext, IObservable<object?>>)(_ => throw new InvalidOperationException("This source stream resolver should never be called. It is only used to prevent the default source stream resolver from being used.")));
+    //private static readonly ISourceStreamResolver _invalidStreamResolver = new SourceStreamResolver<object?>((Func<IResolveFieldContext, IObservable<object?>>)(_ => throw new InvalidOperationException("This source stream resolver should never be called. It is only used to prevent the default source stream resolver from being used.")));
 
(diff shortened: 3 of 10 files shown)


src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -469,7 +469,7 @@ public void ApplyMiddleware(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegat
 
             foreach (var item in Dictionary)
             {
-                if (item.Value is IComplexGraphType complex)
+                if (item.Value is IComplexGraphType complex && item.Value is not IInputObjectGraphType)
```
sungam3r, line 472: ~130 tests failed
also this saves a bit of memory

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -510,11 +510,13 @@ public virtual FieldType AddField(FieldType fieldType)
 
             var builder = CreateBuilder<TProperty>(type)
                 .Name(name)
-                .Resolve(new ExpressionFieldResolver<TSourceType, TProperty>(expression))
                 .Description(expression.DescriptionOf())
                 .DeprecationReason(expression.DeprecationReasonOf())
                 .DefaultValue(expression.DefaultValueOf());
 
+            if (this is not IInputObjectGraphType)
```
sungam3r, line 517: ~270 tests failed
also this saves a bit of memory

sungam3r:
@Shane32 I need your help with grammar in messages.

src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
```diff
@@ -129,6 +132,9 @@
                 throw new InvalidOperationException($"The field '{field.Name}' of an Interface type '{type.Name}' must be an output type.");
 
             ValidateFieldArgumentsUniqueness(field, type);
+
+            if (field.StreamResolver != null && type != schema.Subscription)
```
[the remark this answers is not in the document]
Shane32: Let's add the same check for the Resolver property of interface types.  We already determined that we must not have resolvers on interfaces because according to spec the interface must resolve to a type, and that way __typename can be resolved correctly, etc.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -105,7 +105,7 @@ internal static class AutoRegisteringOutputHelper
         if (buildMemberInstanceExpressionFunc == null)
         {
             fieldType.Resolver = _invalidFieldResolver;
```
Shane32, line 107:
```suggestion
            // interface types do not set resolvers
            fieldType.Resolver = null;
```

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -10,7 +10,7 @@ namespace GraphQL.Types;
 internal static class AutoRegisteringOutputHelper
 {
     private static readonly IFieldResolver _invalidFieldResolver = new FuncFieldResolver<object?>(_ => throw new InvalidOperationException("This field resolver should never be called. It is only used to prevent the default field resolver from being used."));
-    private static readonly ISourceStreamResolver _invalidStreamResolver = new SourceStreamResolver<object?>((Func<IResolveFieldContext, IObservable<object?>>)(_ => throw new InvalidOperationException("This source stream resolver should never be called. It is only used to prevent the default source stream resolver from being used.")));
+    //private static readonly ISourceStreamResolver _invalidStreamResolver = new SourceStreamResolver<object?>((Func<IResolveFieldContext, IObservable<object?>>)(_ => throw new InvalidOperationException("This source stream resolver should never be called. It is only used to prevent the default source stream resolver from being used.")));
```
Shane32, line 14:
```suggestion
```

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -10,7 +10,7 @@ namespace GraphQL.Types;
 internal static class AutoRegisteringOutputHelper
 {
     private static readonly IFieldResolver _invalidFieldResolver = new FuncFieldResolver<object?>(_ => throw new InvalidOperationException("This field resolver should never be called. It is only used to prevent the default field resolver from being used."));
```
Shane32, line 12:
```suggestion
```

Shane32: Another option would be to have these 'invalid resolvers' set by the schema validator to ensure that interface resolvers are not used.

sungam3r: I understand. I would prefer remove.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
sungam3r: good

Shane32:
Do we have schema validation tests to ensure that `StreamResolver` IS set for the `Schema.Subscription` graph type?

Shane32:
For v8, I would also like to require that `Resolver` is null when `StreamResolver` is not null.  This would certainly be a breaking change, but it should clarify behavior, as implementing a field resolver on top of a stream resolver is not intuitive and introduces unnecessary complexity.  Moreover, if a mapping was needed from the `IObservable` type to the resolved type, a simple `.Select` (via System.Reactive extensions) allows for nearly identical behavior.

sungam3r:
> Do we have schema validation tests to ensure that StreamResolver IS set for the Schema.Subscription graph type?

No. I can add it in new PR, don't want to mix one more check.

Shane32:
> > Do we have schema validation tests to ensure that StreamResolver IS set for the Schema.Subscription graph type?
> 
> No. I can add it in new PR, don't want to mix one more check.

Sounds good.  If this were not the case, the schema would be improperly configured, so this is not a breaking change.

sungam3r:
I prefer option 4 - merge as is and wait for feedback. If anyone will be affected, then push changes for option 3 in v7.5+

docs2/site/docs/migrations/migration7.md
```diff
@@ -598,3 +598,10 @@ require them to resolve to GraphQL Object types. GraphQL.NET 7.2+ now enforces t
 Due to the above validation, resolvers defined on fields of interface types are never be used by
 GraphQL.NET. As such, resolvers are not generated for fields of `AutoRegisteringInterfaceGraphType`s
 in version 7.2+.
+
+### 18. `SchemaValidationVisitor` has more runtime checks
+
+From version 7.4 on, only root fields of subscriptions should set `StreamResolver` and only object
+output types should set `Resolver` properties. This change may be potentially breaking for your apps,
+because schema will throw exception during initialization. To get rid of that exception just stop
+setting these properties for fields mentioned in exception message.
```
Shane32, line 607:
```suggestion
Starting with version 7.4, the `StreamResolver` property should be set exclusively for root fields
in subscriptions, while the `Resolver` property should only be assigned to object output types'
fields. Fields within union types, interface types and input object types should not set these
properties. This modification may lead to potential disruptions in your applications, as the
schema may trigger an exception during initialization. To eliminate this exception, simply
cease assigning these properties for the fields specified in the exception message.
```

I suggest adding an explicit notice that union/interface/input objects cannot have resolver set.  Then I asked ChatGPT to rewrite it.

sungam3r: > Fields within union types

Strictly speaking unions have no fields. They consist of other object graph types. 

Leave as as?

Shane32: Whatever you think.  I'll approve now and you can take out 'unions' out of there if you like.

Shane32 (approved):

docs2/site/docs/migrations/migration7.md
```diff
@@ -598,3 +598,12 @@ require them to resolve to GraphQL Object types. GraphQL.NET 7.2+ now enforces t
 Due to the above validation, resolvers defined on fields of interface types are never be used by
 GraphQL.NET. As such, resolvers are not generated for fields of `AutoRegisteringInterfaceGraphType`s
 in version 7.2+.
+
+### 18. `SchemaValidationVisitor` has more runtime checks
+
+Starting with version 7.4, the `StreamResolver` property should be set exclusively for root fields
+in subscriptions, while the `Resolver` property should only be assigned to object output types'
+fields. Fields within union types, interface types and input object types should not set these
```
sungam3r, line 606:
```suggestion
fields. Fields within interface types and input object types should not set these
```

andreyrd:
Funny enough, this just broke some really old code I'm trying to upgrade 🤣 

I cannot apply the fix in https://github.com/graphql-dotnet/conventions/pull/259 because the fields are created in a way where I don't have any kind of `IsInterfaceType` available. And I don't understand it well enough at the moment to do a proper refactor.

But this really should never have been shipped in 7.x because at the end of the day it was a breaking change whether it was easy to fix in conventions or not 🫣

Shane32:
Yeah, see discussion from last April.  As a general rule of thumb, we do not hesitate to make these kinds of changes:
- Error messages or exact classes thrown from error conditions.
- Breaking changes that are considered bug fixes -- the code returned invalid data, or formatted incorrect, etc.
- Additional validation against invalid schema configurations.

The last one is a bit of a gray area.  Technically a schema would probably work for most people even with the incorrect configuration.  However, when the configuration is incorrect, depending on the scenario, it sometimes can lead to corrupted data.  The most obvious example is using shared graph instance types across different schema instances.  It might work great 99% of the time but it will really mess up the application the other 1%.

In this case, allowing resolvers on interface types can lead people to believe that they would be executed, when they never were.  We ended up deciding to apply the change mid-cycle in 7.x.  Perhaps this decision was in error.

In any case, if you want to write a 'safe' patch for 7.x, you can write a `ISchemaNodeVisitor` implementation that removes any resolvers from interface types, and register it with `Schema.RegisterVisitor`.  Alternatively you could override the method that validates the schema, but I would not suggest that approach or it may disable other checks.

andreyrd:
That's fair and makes sense. I'll take a look at those analyzers. Thank you!

PR #3583, opened 2023-04-07
[WIP] Strong typed API for building fields

sungam3r:
Rel: #3574 and #1176.

This is my attempt to solve the problem of excessive `StreamResolver` property for `FieldType` class and API around it.

TODO:
- Add APIs to use `ResolveStream` method.


src/GraphQL.DataLoader.Tests/Types/OrderItemType.cs
```diff
@@ -14,8 +14,7 @@ public OrderItemType(IDataLoaderContextAccessor accessor, IProductsStore product
         Field(x => x.Quantity);
         Field(x => x.UnitPrice);
 
-        Field<ProductType, Product>()
-            .Name("Product")
+        Field<ProductType, Product>("Product")
```
sungam3r, line 17: It's not related directly to my changes, I just stop using obsoleted overload in some places.

src/GraphQL.Tests/Initialization/SchemaInitializationTests.cs
```diff
@@ -494,9 +495,13 @@ public class SchemaWithFieldStreamResolverOnNonRootSubscriptionField : Schema
     public SchemaWithFieldStreamResolverOnNonRootSubscriptionField()
     {
         var type = new ObjectGraphType { Name = "MyQuery" };
-        type.Field<StringGraphType>("str")
-            .ResolveStream(_ => new Subscription.SampleObservable<string>())
-            .Resolve(_ => "abc");
+        type.AddField(new SubscriptionRootFieldType
```
sungam3r, line 498: @Shane32 I switched to explicitly construct fields here because there are no methods to call `ResolveStream` yet, `Field` method returns `FieldBuilder`, not `SubscriptionRootFieldBuilder`. As you see user is allowed to use `SubscriptionRootFieldType` in wrong places. Good or bad - my goal here is preserve tests behavior to check error messages. We can design top-level APIs in such a way to prevent all invalid usages like that one.

src/GraphQL/Builders/SubscriptionRootFieldBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+            Name = name,
+            Type = type,
+        };
+        return new SubscriptionRootFieldBuilder<TSourceType, TReturnType>(fieldType);
+    }
+
+    /// <summary>
+    /// Sets a source stream resolver for the field.
+    /// </summary>
+    public virtual FieldBuilder<TSourceType, TReturnType> ResolveStream(Func<IResolveFieldContext<TSourceType>, IObservable<TReturnType?>> sourceStreamResolver)
```
sungam3r, line 50: Both `ResolveStream` and `ResolveStreamAsync` are unused now.

src/GraphQL/Execution/SubscriptionExecutionStrategy.cs
```diff
@@ -90,13 +90,8 @@ public override async Task<ExecutionResult> ExecuteAsync(ExecutionContext contex
...
         {
-            var resolver = node.FieldDefinition?.StreamResolver;
-
-            if (resolver == null)
-            {
-                // todo: this should be caught by schema validation
-                throw new InvalidOperationException($"Stream resolver not set for field '{node.Field.Name}'.");
-            }
+            var resolver = (node.FieldDefinition as SubscriptionRootFieldType)?.StreamResolver
+                ?? throw new InvalidOperationException($"Stream resolver not set for field '{node.Field.Name}'."); // TODO: this should be caught by schema validation
```
sungam3r, line 94: I'll post PR into `master` to fix this.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -70,13 +69,12 @@ internal static class AutoRegisteringOutputHelper
                 {
                     var resolver = new SourceStreamMethodResolver(methodInfo, memberInstanceExpression, expressions);
                     fieldType.Resolver = resolver;
-                    fieldType.StreamResolver = resolver;
+                    (fieldType as SubscriptionRootFieldType ?? throw new InvalidOperationException("BOOM!")).StreamResolver = resolver;
```
sungam3r, line 72: @Shane32 What would you like to do here?

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -399,6 +399,22 @@ public virtual FieldType AddField(FieldType fieldType)
             });
         }
 
+        /// <summary>
+        /// Creates a field builder used by SubscriptionField() methods.
+        /// </summary>
+        protected virtual SubscriptionRootFieldBuilder<TSourceType, TReturnType> CreateSubscriptionRootBuilder<TReturnType>([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type type)
```
sungam3r, line 405: Both new methods here are unused.

Shane32:
I took a look at this today.  So far I don't see this as a good solution if it requires people to use AddField rather than the field builders.  This undermines much of our recent design changes, which encourage use of field builders and allows type inference and extension method use.

Now if there was a new base object type (e.g. `SubscriptionObjectGraphType<T>`) and the `Field<T>` methods on it all returned subscription field builders, that could be considered.  But I fear that it will be a massive amount of duplicated code.

Another option would be to add `SubscriptionField<T>` methods that return a subscription field builder, but that doesn't seem to be much different than what we have now.

So far, I just don't see how the benefits outweigh the negatives.  I'll be glad to review any options you may have.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
Shane32: `AutoRegisteringHelper.CreateField` creates the `FieldType`, but this is entirely separate from the this code that builds the resolvers.  We would have to cross the logic between the two, so `AutoRegisteringHelper.CreateField` calls `AutoRegisteringOutputHelper.IsObservableOrAsyncEnumerable` to determine what kind of `FieldType` to add.

Shane32:
It's also possible that I misunderstand your changes from my very-brief scan of the changes.  I would like to see a sample of what changes are required for a typical subscription schema, both with auto registering graph types, and code-first.  I'm not specifically seeing it yet.

src/GraphQL.Tests/Subscription/SubscriptionSchema.cs
```diff
@@ -23,14 +23,14 @@ public class ChatSubscriptions : ObjectGraphType
     public ChatSubscriptions(IChat chat)
     {
         _chat = chat;
-        AddField(new FieldType
+        AddField(new SubscriptionRootFieldType
         {
             Name = "messageAdded",
             Type = typeof(MessageType),
             StreamResolver = new SourceStreamResolver<Message>(Subscribe)
         });
```
Shane32, line 31: As we discourage this design style, this doesn't seem applicable to understanding typical required changes.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
Shane32: Which it looks like you've already done.

src/GraphQL.Tests/Initialization/SchemaInitializationTests.cs
Shane32: I understand.  As such, this is also not a good example of typical code changes required by users.

Shane32:
It seems that you wrote another field builder, but unless I'm wrong, there's no way to use it.  Let's assume you fix that issue.  It seems like a lot of extra code and trouble for very little gain.  In order to be worth it, this PR should prevent users from creating subscription fields at compile time.

For example, let's say you add `SubscriptionField` which creates a subscription field builder.  There's nothing to prevent a user from calling `Field` rather than `SubscriptionField`.  So we are back exact to where we are now.  The only difference is that the user has to discern between `Field` and `SubscriptionField` versus `Resolve` and `ResolveStream`.  Given the two, I prefer the latter.  Intellisense will show `ResolveStream` as an autocompletion when `Resolve` is beginning to be typed, whereas this is not the case with `Field`.  Of course we could name it `FieldSubscription` but in the end, the user still needs to choose the right method.

We could take it even a step further, as I described above, where there is a dedicated `SubsciptionGraphType`.  Then we could set `Schema.Subscription` to only allow setting a `SubscriptionGraphType` instance.  The `Field` methods could be shadowed, so that the default ones are not accessible.  Then we achieve compile-time checks.

So I would not merge yet.  Certainly not as-is, with no way to use the subscription field builders.

Shane32:
All my comments apply equally to changes towards the goal of having input types disallow resolvers (which I endorse): the goal is met if compile-time restrictions are in place.  Otherwise we have missed the goal.

Shane32:
Another thing I'm thinking, is that the subscription field type inherits from `FieldType`, so it could be assigned anywhere that `FieldType` is allowed.  In other words, a stream resolver could be set for a non-root-subscription field.  Maybe we need some type of inheritance:

- `FieldType` - abstract
  - `ObjectFieldType` - for output object types (adds `FieldResolver`)
  - `InterfaceFieldType` - for interface object types
  - `InputFieldType` - for input types
  - `SubscriptionFieldType` - for root subscription types (adds `StreamResolver`)

However, these changes might be drastic to implement. I think we should sit back when complete with whatever changes we come up with and consider: are the changes worth it?  Probably, I think, but let's not presuppose that the answer is yes.

src/GraphQL.Tests/Subscription/SubscriptionSchema.cs
sungam3r: This is the minimum change for the code to be compiled. I did not aim to rewrite everything.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
sungam3r: I mean would you like to throw exception or not. If yes then which type and message?

src/GraphQL.Tests/Initialization/SchemaInitializationTests.cs
sungam3r: Yes. For me it was good way to make tests green again.

sungam3r:
> However, these changes might be drastic to implement.

I needed to start with something and I took the first step. Inheritance hierarchy looks like the second. I understand that in the end it may not work or will be too complex to programm. Nevertheless, after adding the latest changes to schema validator, I got the impression that eventually we can get a working solution.

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
Shane32: I suggest either:

1. `new ArgumentException("The member is a stream resolver, but the specified fieldType is not a SubscriptionRootFieldType.", nameof(fieldType));`
2. `new InvalidOperationException("The member is a stream resolver, but the specified fieldType is not a SubscriptionRootFieldType.");`

Shane32:
> > However, these changes might be drastic to implement.
> 
> I needed to start with something and I took the first step. Inheritance hierarchy looks like the second. I understand that in the end it may not work or will be too complex to programm. Nevertheless, after adding the latest changes to schema validator, I got the impression that eventually we can get a working solution.

:+1:

Please ping me when this PR is ready for further review.

sungam3r:
It is ready for further review. I added typed fields and new builders. In fact, the changes are not as dramatic as I thought.

TODOs/questions and pain points:
1) Connection builders. See commented code block in `InterfaceGraphType.cs`. Now `ConnectionBuilder` works only with `ObjectFieldType` but I think it should also support `InterfaceFieldType`. Also I see `// TODO: Remove in v5` note in code.
2) Jumping around generics in `TypeFields<T>`/`ITypeFields`. Each derived field type has its own typed `Fields` property but also there is some code working with `IComplexGraphType`. See `ComplexGraphTypeExtensions.Fields` extension method, `.Fields().AsEnumerable()` usages and note about CS0695 error in `TypeFields.cs`.
3) Some aux methods have appeared like `internal static QueryArguments? Arguments(this FieldType fieldType) => fieldType is IFieldTypeWithArguments ftwa ? ftwa.Arguments : null;`
4) `IComplexGraphType` interface became empty. `ComplexGraphType<TSourceType>` class contains only ctor.
5) `ObjectGraphType.CreateSubscriptionRootBuilder` is still unused. I'm not sure how to be with `SubscriptionRootFieldType` - inherit it from `ObjectFieldType` or from `FieldType`. I tend to the former.

Shane32:
I need time to review.  Looking over only the API changes, I was thinking that perhaps there is an opportunity to turn some field builder methods into extension methods and perform some deduplication of code.  Perhaps the field builders are merely interfaces.  I have to review further.

Shane32:
Note that I think this PR is on the right track, but we should be careful to merge without a full review.

sungam3r:
>  to turn some field builder methods into extension methods and perform some deduplication of code

I would not even if code is duplicated. The reason is different return types of builders to allow chaining.

Shane32:
Extension methods can still do this

sungam3r:
Yes, they can. I have not yet evaluated what is better - to have separate builders or try to reduce the code to one class through extension methods.

sungam3r:
I remembered why I did not use extension methods - all methods of builders are virtual.

PR #3602, opened 2023-04-22
Use IClassFixture for OpenTelemetryTests

sungam3r:
It allows to share context across all tests in class - https://xunit.net/docs/shared-context.


Shane32:
What is the purpose of sharing the context across tests here? Usually it would not be desired. And I see you had a call to Reset. At first glance it seems an improper use.

Shane32:
Here we see the same pattern. The code that was in the constructor was moved to the fixture. The tests should not re-use a shared context, and the initialization code was already deduplicated by way of the constructor. 

There is no code cleanup in this PR and context should not be shared here. I do not believe the changes should be made as proposed here.

PR #3606, opened 2023-04-24
Scoped lifetime support for AddValidationRule

### LINKED ISSUE #2947: Lifetime of validation rule in AddValidationRule
@Shane32  I see that rule is added as singleton and user cannot configure this. Why?

Shane32:
fixes #2947


Shane32:
I don't see the need for migration notes here.  It's source backwards compatible and there already is adequate documentation regarding DI.  It could be listed as a new feature however.

PR #3607, opened 2023-04-24
Catch exceptions during Subscribe and pass to the unhandled exception delegate

### LINKED ISSUE #3010: User exceptions in subscription's IObservable.Subscribe are not handled by the unhandled error delegate
When a subscription is executed, Rule 5.2.3.1 ensures that a single root field is requested; not zero and not more than one.  The GraphQL DocumentExecuter in conjunction with the SubscriptionExecutionStrategy returns an IObservable.  But when IObservable.Subscribe is called to set up the subscription (within GraphQL.Server), exceptions are not handled by the unhandled exception handler (since this is outside of 'execution').  Rather, GraphQL.Server needs to handle the exception in some fashion.

Shane32:
Fixes #3010

### DIFF
--- a/src/GraphQL.Tests/Subscription/ObservableExtensionsTests.cs
+++ b/src/GraphQL.Tests/Subscription/ObservableExtensionsTests.cs
@@ -272,6 +272,20 @@ public class ObservableExtensionsTests
         });
     }
 
+    [Fact]
+    public async Task ExceptionsDuringSubscribeProduceOnError()
+    {
+        var errorObservable = new ErrorObservable();
+        var observable = errorObservable.SelectCatchAsync<string, string>((s, _) => new(s), (ex, _) => new(new ExecutionError(ex.Message)));
+        using var subscription = observable.Subscribe(Observer);
+        await Observer.WaitForAsync("Error 'ExecutionError'. ").ConfigureAwait(false);
+    }
+
+    private class ErrorObservable : IObservable<string>
+    {
+        public IDisposable Subscribe(IObserver<string> observer) => throw new NotImplementedException("This is a test");
+    }
+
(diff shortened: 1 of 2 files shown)


src/GraphQL/Extensions/ObservableExtensions.cs
```diff
@@ -48,12 +48,33 @@ public SelectCatchAsyncObservable(IObservable<TIn> source, Func<TIn, Cancellatio
...
+                {
+                    try
+                    {
+                        ex = await _transformError(ex, default).ConfigureAwait(false);
+                    }
+                    finally
+                    {
+                        observer.OnError(ex);
+                    }
+                });
```
sungam3r, line 76:
```suggestion
                newObserver.OnError(ex);
```

Will it work?

src/GraphQL.Tests/Subscription/ObservableExtensionsTests.cs
```diff
@@ -272,6 +272,20 @@ public void NullArgumentsThrow()
         });
     }
 
+    [Fact]
+    public async Task ExceptionsDuringSubscribeProduceOnError()
+    {
+        var errorObservable = new ErrorObservable();
+        var observable = errorObservable.SelectCatchAsync<string, string>((s, _) => new(s), (ex, _) => new(new ExecutionError(ex.Message)));
+        using var subscription = observable.Subscribe(Observer);
+        await Observer.WaitForAsync("Error 'ExecutionError'. ").ConfigureAwait(false);
```
sungam3r, line 281: Maybe modify `SampleObserver.OnError` to track text for exception? So this test will check for "Error 'ExecutionError'. This is a test" - that would be more readable.

sungam3r (approved):

PR #3609, opened 2023-04-24
Simplify metadata builder methods

### LINKED ISSUE #3328: Add `IMetadataBuilder` interface
In the next version of GraphQL.NET, perhaps the field and connection builders should implement `IMetadataProvider` so these methods can be removed.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/issues/3324#issuecomment-1244737081_

Shane32:
Closes #3328 

This eliminates redundant field and connection builder methods for adding metadata such as directives, authentication information, and complexity impact data.  With this design there is no additional methods/properties on field builders for end users to see (i.e. no new `.Metadata` property).  Metadata can still be accessed through the `FieldType` property.  It also does not expose methods to read metadata which may be confusing.

This PR also solves many of the missing metadata methods (directives, complexity impact, etc) on the connection builders.

This also unifies the method name used to apply metadata to a field.  So `WithComplexityImpact` is now standardized whether using a `FieldType` directly or via a field builder.  I know that does not fit the naming pattern of the 'native' field builder methods, but I feel this is not a bad thing and should be consistent for all metadata methods.


src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -2209,18 +2202,12 @@ namespace GraphQL.Types
...
         string? Description { get; set; }
     }
-    public interface IProvideMetadata
-    {
-        System.Collections.Generic.Dictionary<string, object?> Metadata { get; }
-        TType GetMetadata<TType>(string key, System.Func<TType> defaultValueFactory);
-        TType GetMetadata<TType>(string key, TType defaultValue = default);
-        bool HasMetadata(string key);
-    }
+    public interface IProvideMetadata : GraphQL.Types.IMetadataBuilder { }
```
sungam3r, line 2205: What is the purpose of two identical interfaces?

src/GraphQL/Types/IProvideMetadata.cs
```diff
@@ -5,36 +5,7 @@ namespace GraphQL.Types
     /// This interface is implemented by numerous descendants like <see cref="GraphType"/>,
     /// <see cref="FieldType"/>, <see cref="Schema"/> or others.
     /// </summary>
-    public interface IProvideMetadata
+    public interface IProvideMetadata : IMetadataBuilder
```
sungam3r, line 8: From `IMetadataBuilder`:
> This interface is implemented by field builders for configuring metadata.

So each extension method that accepts `IMetadataBuilder` can accept `IProvideMetadata` as well. I do not see advantages in addition of second interface from the caller point of view. I understand what this PR about but jumping around changes `IProvideMetadata` <->`IMetadataBuilder` looks unnecessary.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
Shane32: How else to provide writing methods to builders and not reading methods?

Shane32: I can unnecessarily move some methods from one interface to the other but why?

src/GraphQL/Types/IProvideMetadata.cs
Shane32: We certainly do not want to add methods such as IsAnonymous or GetDirective to field builders

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
sungam3r: > How else to provide writing methods to builders and not reading methods?

I do not understand.

src/GraphQL/Types/IProvideMetadata.cs
sungam3r: [GetDirective -> GetAppliedDirectives ?] Well, now I better understand intention but still - why not? We do not **add** all that methods. They just exist. Honestly, it's harder for me to read code with two nearly identical interfaces for different purposes. OK, if you **really** want to split usage of builders from usage of "pure" `IProvideMetadata` let's consider such design:
```cs
public interface IBuilder { } // nothing here for now, just marker interface for where constraint 
public class AnyBuilder<T, U, whatever> : IBuilder, .... { ... } // add IBuilder for every builder class in repo

// AllowAnonymous - example of usage
public static TMetadataProvider AllowAnonymous<TMetadataProvider>(this TMetadataProvider provider)
            where TMetadataProvider : IProvideMetadata
            where TMetadataProvider : IBuilder // new constraint
```

so builder-releated methods are available only for classes marked with `IBuilder` interface.

Shane32: No, it should not be simply builder classes but any classes that allow writing metadata.  Consider:

```cs
var field = new FieldType(...);
field.AllowAnonymous();
```

It would not be logical for `FieldType` to be marked with `IBuilder`.  At least not at this time.  Also, it's the reader methods that need an additional constraint, not the writing methods.  Otherwise we would end up with the same problem with the builder methods.

sungam3r: > It would not be logical for `FieldType` to be marked with `IBuilder`

Yes. I didn't want that.


So the whole intent is to split read APIs (informational) from write APIs (builders)?

Shane32: Yes, or more specificially, separate read/write from write-only.  GraphQL.NET doesn't have read-only classes as a general rule, like for example `FieldType` is read/write -- I always wish we would have an `IFieldType` that's read only.

So the extension methods for `IMetadataBuilder` is for writing only whereas `IProvideMetadata` is read/write.

Shane32: If you remember, I spent quite a lot of time attempting to make `ISchema` and all descendants read-only, so you 'build' the schema and then it is read-only for its lifetime.  This didn't work out as the design changes necessary were so complex that the concept was nixed.  If it were so, we would have read-only extension methods and write-only extension methods.  Some classes may be read-write.  But at the moment the infrastructure is either write-only (builder classes) or read/write (everything else).

sungam3r: > So the extension methods for `IMetadataBuilder` is for writing only whereas `IProvideMetadata` is read/write.

OK, now I completely understand. Give me time to think about it, I'm still in doubt. It looks unusual.

Shane32: We can have `IMetadataReader` and `IMetadataWriter` which both inherit from `IProvideMetadata`.

sungam3r: I mean that we begin to artificially separate APIs by read/write applicability and this can harm when both required.

Shane32: I added `MetadataReader` property to pull reader from writer interface.

src/GraphQL/Types/IGraphType.cs
```diff
@@ -40,7 +40,7 @@ public interface IProvideDeprecationReason
     /// <summary>
     /// Represents a graph type within the GraphQL schema.
     /// </summary>
-    public interface IGraphType : IProvideMetadata, IProvideDescription, IProvideDeprecationReason, INamedType
+    public interface IGraphType : IMetadataReader, IMetadataWriter, IProvideDescription, IProvideDeprecationReason, INamedType
```
Shane32, line 43: Note: `IGraphType` is a read/write interface (unfortunately).  So for example GraphQLAttribute implementations use `Modify(IGraphType)` and metadata needs to be writable.  Also the other inherited interfaces are writable such as `IProvideDescription`.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
sungam3r: Understood.

docs2/site/docs/migrations/migration8.md
```diff
@@ -5,6 +5,28 @@ See [issues](https://github.com/graphql-dotnet/graphql-dotnet/issues?q=milestone
...
+`IMetadataReader` reader from the `IMetadataWriter.MetadataReader` property. Here is an example:
+
+```csharp
+public static TMetadataBuilder RequireAdmin<TMetadataBuilder>(this TMetadataBuilder builder)
+    where TMetadataBuilder : IMetadataWriter
+{
+    if (builder.MetadataReader.GetRoles?.Contains("Guests"))
+        throw new InvalidOperationException("Cannot require admin and guest access at the same time.");
+    builder.AuthorizeWithRoles("Administrators");
+    return builder;
```
sungam3r, line 22:
```suggestion
    return builder.AuthorizeWithRoles("Administrators");
```

docs2/site/docs/migrations/migration8.md
```diff
@@ -18,3 +40,12 @@ GlobalSwitches.RequireRootQueryType = false;
...
 Future versions of GraphQL.NET will not contain this property and each schema will always be required to have a root Query type to comply with the GraphQL specification.
 
+### 2. Use `ApplyDirective` instead of `Directive` on field builders
+
+The `Directive` method on field builders has been renamed to `ApplyDirective` to better fit with
+other field builder extension methods.
+
+### 3. Use `WithComplexityImpact` instead of `ComplexityImpact` on field builders
+
+The `ComplexityImpact` method on field builders has been renamed to `WithComplexityImpact` to better fit with
```
sungam3r, line 50: `With` and `Apply` words are synonims here. Maybe unify? I tend to use apply over with.

src/GraphQL/Builders/ConnectionBuilder.cs
```diff
@@ -362,5 +362,14 @@ private static void CheckForErrors(IResolveConnectionContext<TSourceType> contex
                 throw new ArgumentException("Cannot use `last` with unidirectional connections.");
             }
         }
+
+        // Allow metadata builder extension methods to read/write to the underlying field type without unnecessarily
```
sungam3r, line 366:
```suggestion
        // Allows metadata builder extension methods to read/write to the underlying field type without unnecessarily
```
?

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -338,11 +316,9 @@ protected FieldBuilder(FieldType fieldType)
         /// then their default values (if any) will be used.
         /// </summary>
         /// <param name="name">Directive name.</param>
+        [Obsolete("Please use the ApplyDirective method")]
```
sungam3r, line 319: You know my preferences - I would remove these methods to not bother with them when doing further work in v8.

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -366,31 +340,34 @@ protected FieldBuilder(FieldType fieldType)
...
         /// <param name="impact">Field's complexity impact.</param>
+        [Obsolete("Please use the WithComplexityImpact method")]
         public virtual FieldBuilder<TSourceType, TReturnType> ComplexityImpact(double impact)
-        {
-            FieldType.WithComplexityImpact(impact);
-            return this;
-        }
+            => this.WithComplexityImpact(impact);
+
+        // Allow metadata builder extension methods to read/write to the underlying field type without unnecessarily
```
sungam3r, line 364:
```suggestion
        // Allows metadata builder extension methods to read/write to the underlying field type without unnecessarily
```

src/GraphQL/Extensions/AuthorizationExtensions.cs
```diff
@@ -190,166 +189,5 @@ public static TMetadataProvider AuthorizeWithRoles<TMetadataProvider>(this TMeta
...
-        /// <summary>
-        /// Adds authorization policy to the specified field builder. If the underlying field already contains
-        /// a policy with the same name, then it will not be added twice.
-        /// </summary>
-        /// <typeparam name="TSourceType"></typeparam>
-        /// <typeparam name="TReturnType"></typeparam>
-        /// <param name="builder"></param>
-        /// <param name="policy"> Authorization policy name. </param>
-        /// <returns> The reference to the specified <paramref name="builder"/>. </returns>
-        public static FieldBuilder<TSourceType, TReturnType> AuthorizeWithPolicy<TSourceType, TReturnType>(
```
sungam3r, line 203: Nice cleanup.

sungam3r (approved):

src/GraphQL/Types/IProvideMetadata.cs
sungam3r: ok

docs2/site/docs/migrations/migration8.md
Shane32: Separate PR

src/GraphQL/Builders/FieldBuilder.cs
Shane32: We can always just make these into obsolete extension methods so as to not impede your field builder refactoring work. I really don’t like deleting field builder methods outright, as v5 it us expected that a user may have many of these calls in their codebase. Let’s consider the best way to implement the obsolete methods as you refactor the field builders

docs2/site/docs/migrations/migration8.md
Shane32: (Since WithComplexityImpacf already exists)

PR #3615, opened 2023-04-25
Add GetInputExtension method

sungam3r:
Why does `IResolveFieldContext.InputExtensions` has `IReadOnlyDictionary<string, object?>` type instead of `Inputs`? Because of this I have to cast extensions
```
 // actually context.InputExtensions is of type GraphQL.Inputs
                if (context.InputExtensions is not IDictionary<string, object?> values || values.Count == 0)
                    return null;
```

### DIFF
--- a/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
+++ b/src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
@@ -118,31 +118,10 @@ namespace GraphQL
             if (context == null)
                 throw new ArgumentNullException(nameof(context));
 
-            if (context.InputExtensions == null || context.InputExtensions.Count == 0)
-                return null;
-
-            // actually context.InputExtensions is of type GraphQL.Inputs
-            if (context.InputExtensions is not IDictionary<string, object?> values || values.Count == 0)
-                return null;
-
-            if (path.IndexOf('.') != -1)
-            {
-                string[] keys = path.Split(_separators);
-
-                for (int i = 0; i < keys.Length - 1; ++i)
-                {
-                    if (values.TryGetValue(keys[i], out object? v) && v is IDictionary<string, object?> d)
-                        values = d;
-                    else
-                        return null;
-                }
-
-                return values.TryGetValue(keys[keys.Length - 1], out object? result) ? result : null;
-            }
-            else
-            {
-                return values.TryGetValue(path, out object? result) ? result : null;
-            }
+            // Actually context.InputExtensions is of type GraphQL.Inputs : ReadOnlyDictionary<string, object?> and
+            // ReadOnlyDictionary<string, object?> implements IDictionary<string, object?> so this cast should never
+            // return null for majority of cases.
+            return GetByPath(context.InputExtensions as IDictionary<string, object?>, path, true);
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


Shane32 (approved):

sungam3r:
Change `IReadOnlyDictionary<string, object?>` to  `Inputs` for v8?

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -107,6 +107,44 @@ public static IResolveFieldContext<TSourceType> As<TSourceType>(this IResolveFie
...
+            if (context == null)
+                throw new ArgumentNullException(nameof(context));
+
+            if (context.InputExtensions == null || context.InputExtensions.Count == 0)
+                return null;
+
+            // actually context.InputExtensions is of type GraphQL.Inputs
+            if (context.InputExtensions is not IDictionary<string, object?> values || values.Count == 0)
+                return null;
+
```
Shane32, line 127: remove

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -107,6 +107,44 @@ public static IResolveFieldContext<TSourceType> As<TSourceType>(this IResolveFie
...
+            if (path.IndexOf('.') != -1)
+            {
+                string[] keys = path.Split(_separators);
+
+                for (int i = 0; i < keys.Length - 1; ++i)
+                {
+                    if (values.TryGetValue(keys[i], out object? v) && v is IDictionary<string, object?> d)
```
Shane32, line 134:
```suggestion
                    if (values.TryGetValue(keys[i], out object? v) && v is IReadOnlyDictionary<string, object?> d)
```

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
sungam3r: does not compile

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
```diff
@@ -118,12 +135,14 @@ public static IResolveFieldContext<TSourceType> As<TSourceType>(this IResolveFie
             if (context == null)
                 throw new ArgumentNullException(nameof(context));
 
-            if (context.OutputExtensions == null || context.OutputExtensions.Count == 0)
-                return null;
+            return GetByPath(context.OutputExtensions, path, true);
+        }
 
-            lock (context.OutputExtensions)
+        private static object? GetByPath(IDictionary<string, object?>? dictionary, string path, bool useLock)
```
Shane32, line 141: I would suggest changing the signature here to `IReadOnlyDictionary` here and the `is IDictionary` line.  It probably makes little difference, as the `ImmutableDictionary` class and even the brand-new `FrozenDictionary` class supports `IDictionary` as well as `IReadOnlyDictionary`.

Shane32 (approved):

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
Shane32: Well you can’t literally remove it. I meant remove the cast.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
Shane32: It would have compiled if you removed the cast. But I see you rewrote it now.

src/GraphQL/ResolveFieldContext/ResolveFieldContextExtensions.cs
sungam3r: done

PR #3621, opened 2023-05-02
Include license in package

Shane32:
Hopefully then the license should appear in Visual Studio, similar to MS packages:

image

Right now it looks like this:

image

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -21,6 +21,7 @@
   <ItemGroup Condition="'$(IsPackable)' == 'true'">
     <None Include="..\..\assets\logo.64x64.png" Pack="true" PackagePath="\" Visible="false" />
     <None Include="..\..\README.md" Pack="true" PackagePath="\" Visible="false" />
+    <None Include="..\..\LICENSE.md" Pack="true" PackagePath="\" Visible="false" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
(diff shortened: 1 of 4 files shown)


sungam3r:
A long time ago we switched from PackageLicenseFile to PackageLicenseExpression. Also it was a recommended way by MS, compiler reported a warning. Many projects did so. So why to bring PackageLicenseFile back again? As I understand VS (with some plugin installed?) shows package contents. OK. It is just an auxilary functionality of IDE. Perhaps later VS will learn to show license using PackageLicenseExpression. But now it cannot and it should not follow from this that the license file must be added to the package.

Shane32:
I'm fine removing PackageLicenseFile.  Nevertheless, I think the license.md file should be bundled within the nuget.

src/Directory.Build.targets
```diff
@@ -1,6 +1,7 @@
 <Project>
 
   <PropertyGroup Condition="'$(IsPackable)' == 'true'">
+    <PackageLicenseFile>LICENSE.md</PackageLicenseFile>
```
Shane32, line 4:
```suggestion
```

sungam3r (approved):
Well, generally I'm not against it as well but have you seen any such usages besides MS packages? Existence of those 3 files may be explained by MS corporate politics and does not apply to us as well as majority of other open source projects.

Shane32:
Well the license agreement explicitly says:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

“Should we not follow our own rules?” was my thinking

Shane32:
Note that the license “expression” has no legal bearing on the copyright status (in my mind), as we (the authors) specified an explicit license / copyright notice for this package.

sungam3r:
Legal stuff is not in my strong skills.

PR #3628, opened 2023-05-22
Prevent double registration of configuration types

Shane32:
When using a type to configure a schema or execution (`IConfigureSchema` or `IConfigureExecution`) this change prevents it from being registered in DI twice.  Does not affect `.ConfigureExecution()` with an instance or factory type.

Within GraphQL.NET, this only affects the following methods:
- `.ConfigureExecution<T>()` (not the other two overloads)
- `.UseTelemetry(...)`
- `.UseMemoryCache(...)`
- `.UseAutomaticPersistedQueries(...)`

With this change, now calling any of those methods more than once will not result in the execution added to the pipeline more than once.  Note that the configuration delegate (if provided) does get run for each call, as it should.

Does not affect `.UseApolloTracing()`, `.UseMiddleware()`, `.ConfigureExecutionOptions()` or any other method.

We should consider reconfiguring `.UseMiddleware<T>()` in the same fashion.  For better or worse, it would prevent installing the same middleware class twice.  So for example, calling `.UseApolloTracing()` twice would not install the `InstrumentFieldsMiddleware` middleware twice.  Note that `.UseApolloTracing()` would need additional changes to use the `.ConfigureExecution<T>()` overload so the execution does not get run twice.

In addition to the above, `.ConfigureSchema<T>()` was added to mimic the same behavior.


Shane32:
Reasons against merging this: Calling any of the `.Use___()` methods implies configuration of the execution pipeline typically in the order specified when calling the method.  So for example, within ASP.NET, is there code to prevent calling `app.UseDeveloperExceptionPage()` more than once?  And, if a `.Use___()` method was called twice, which order is used when executing a document?  The order matters sometimes, such as when adding APQ support in the pipeline!

PR #3630, opened 2023-05-23
Better error message for ThrowSerializationError

sungam3r:


src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
```diff
@@ -339,6 +350,72 @@ enum PetKind {
...
+
+            enum PetKindType {
+              CAT
+              DOG
+            }
+            """;
+
+        var schema = Schema.For(definitions, c =>
+        {
+            c.Types.Include<PetKindType>(); // type.GraphQLName() inside truncates 'Type' suffix from 'PetKindType'
```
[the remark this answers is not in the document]
Shane32: Well, this should be a [processing error](https://graphql-dotnet.github.io/docs/getting-started/errors/#a-nameprocessingerrorsaprocessing-errors) which do not inherit from `ExecutionError` and should be caught by the default unhandled exception handler and so the only thing returned would be "Error trying to resolve field ...".

This design allows us to include "sensitive" information in the error message for use in logging, as this is a server-side error not a client-side error.  Many of our existing error messages expose this type of information.

So, in this case, regardless of whether the information is sensitive, it is not exposed unless the server is specifically programmed to expose server errors to the client.

As such, I'm happy to expand these error messages to include any pertinent information that can help the server operator determine the cause of the bug and fix it.  (The error messages should not be tailored toward clients, as they are not the intended recipient.)

Shane32 (approved):

src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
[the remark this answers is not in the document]
sungam3r: > Well, this should be a [processing error](https://graphql-dotnet.github.io/docs/getting-started/errors/#a-nameprocessingerrorsaprocessing-errors) which do not inherit from ExecutionError and should be caught by the default unhandled exception handler and so the only thing returned would be "Error trying to resolve field ...".

As I understand so it is now. Did you mean I can expand error message safely?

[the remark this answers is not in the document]
Shane32: `AssertQueryWithError` inspects the `ExecutionResult` and compares the inner exception and message to the expected output.  That is why the error message needs to be updated in the tests.  But the serialized result does not contain the message.

src/GraphQL/Types/Scalars/Enumeration/EnumerationGraphType.cs
```diff
@@ -97,7 +97,9 @@ public void Add(EnumValueDefinition value)
         var foundByValue = Values.FindByValue(value);
 
         return foundByValue == null
-            ? value == null ? null : ThrowSerializationError(value)
+            ? value == null
+                ? null
+                : throw new InvalidOperationException($"Unable to serialize '{value}' value of type '{value.GetType().GetFriendlyName()}' to the enumeration type '{Name}'. Enumeration does not contain such value. Available values: {string.Join(", ", Values.Select(v => $"'{v.Value}' of type '{v.Value?.GetType().GetFriendlyName()}'"))}.")
```
sungam3r, line 102: OK?

Shane32: It's okay.  Probably could be better, such as using `"DOG"` rather than `'DOG' of type 'String'` and `Bug1773Enum.Hello` rather than `'Hello' value of type 'Bug1773Enum'` -- seems a bit wordy now.  Feel free to merge or tweak further as desired.

Shane32: Ultimately this exception should typically only occur during development, and rarely during production.  So whether the exception message is perfect or not makes little difference so long as the programmer understands the issue and can fix it.  Meaning, 'whatever you think is fine' :smile:

src/GraphQL.Tests/Utilities/SchemaBuilderTests.cs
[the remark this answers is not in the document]
sungam3r: OK

src/GraphQL/Types/Scalars/Enumeration/EnumerationGraphType.cs
sungam3r: "DOG" is good but value may be of any type - that is why I added `of type....`

sungam3r: The same for `'Hello' value of type 'Bug1773Enum'` instead of `Bug1773Enum.Hello` . I'll keep message as is for now.

Shane32: Quotes could denote a string like it does in C#

Shane32: > The same for `'Hello' value of type 'Bug1773Enum'` instead of `Bug1773Enum.Hello` . I'll keep message as is for now.

Ok

sungam3r: > Quotes could denote a string like it does in C#

Of course, but special code is required to check type and concatenate proper text.

PR #3631, opened 2023-05-23
Support OpenTelemetry auto instrumentation

Shane32:


src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -1193,7 +1193,7 @@ public static IGraphQLBuilder UseTelemetry(this IGraphQLBuilder builder, Action<
     public static IGraphQLBuilder UseTelemetry(this IGraphQLBuilder builder, Action<GraphQLTelemetryOptions, IServiceProvider>? configure)
     {
         builder.Services.Configure(configure);
-        builder.ConfigureExecution<GraphQLTelemetryProvider>();
+        builder.Services.TryRegister<IConfigureExecution, GraphQLTelemetryProvider>(ServiceLifetime.Singleton, RegistrationCompareMode.ServiceTypeAndImplementationType);
```
Shane32, line 1196: Not necessary if #3628 is merged

Shane32:
Before merging, this PR should link to the applicable documentation at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation which indicates this is the correct pattern to support automatic instrumentation.  So far I am unable to find any.

Shane32:
@RassK Using the sample code provided in the issue you linked, I created a test, which does not pass because (a) the binding flag needs to be public, not non-public, and (b) the arguments were not passed to the method.

Can you provide a link to the reflection code that calls this `EnableAutoInstrumentation` method?  Then we can be sure the implementation here matches the expected implementation within the OpenTelemetry library -- at least for the current version of OpenTelemetry,

RassK:
> @RassK Using the sample code provided in the issue you linked, I created a test, which does not pass because (a) the binding flag needs to be public, not non-public, and (b) the arguments were not passed to the method.
> 
> Can you provide a link to the reflection code that calls this `EnableAutoInstrumentation` method? Then we can be sure the implementation here matches the expected implementation within the OpenTelemetry library -- at least for the current version of OpenTelemetry,

Updated [the issue](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2520) in OpenTelemetry AutoInstrumentation about the correct usage.

The final implementation is here > 
https://github.com/RassK/opentelemetry-dotnet-instrumentation/pull/954/files#diff-8b2aefd647deef50f042e7b6d65abf19f8acb24c29f26dce47937051380d06d0R23

We do have end to end tests, but at least rc package has to be available in the Nuget to test it.

Shane32:
I updated the tests in this PR to align as closely as possible with [this PR](https://github.com/RassK/opentelemetry-dotnet-instrumentation/pull/954) .

PR #3642, opened 2023-06-07
Don't use ConfigureAwait in docs

EmmanuelPonnudurai:
I have removed the ConfiureAwait calls because it is generally not advisable and also in this case it also is not really needed.


Shane32 (approved):
Looks good. I concur that user code should not typically contain ConfigureAwait - only library code - and this is a sample of user code.

PR #3649, opened 2023-06-19
[WIP] Add SchemaExporter

Shane32:
Waiting on:
- https://github.com/graphql-dotnet/parser/pull/326
- https://github.com/graphql-dotnet/parser/pull/325
- Added tests
- Deprecate/remove SchemaPrinter
- Add/update docs

PR #3660, opened 2023-07-10
Add GraphQL schema-first federation sample and tests

Shane32:


Shane32:
@joemcbride I have added a schema-first federation sample and configured a GitHub Action runner to start the subgraphs, build the supergraph, host the router, and execute some queries.  This would be a first step towards better testing and additional features for federation support.  Would you be interested in helping me expand on this federation sample so that it is properly representative of your/typical use case and/or currently supported features?  I do not use schema-first nor federation, so I'm sorta going blind here.

The purpose is multi-faceted:
1. Ensure that future code changes do not break existing schema-first federation support
2. Allow for updating the schema printer to a new codebase - see PR #3649
3. Allow for federation 2 support and testing
4. Allow for code-first federation -- see PR #3144 
5. Provide a sample setup for a federated schema and/or recommended practices

A working sample may allow me or others to also:
1. Add XML comments to the classes and methods pertaining to federation features
2. Provide documentation for using federation with GraphQL.NET

The most important question I have at the moment is whether Federation 1 support is used/mandatory/necessary/obsolete.  It seems that it would affect the schema printer notably, so it is important to have a proper sample here if I am to update the `FederatedSchemaPrinter` class.  For Federation 2, we can use the standard schema printer, as it does not hurt to print the federation-specific directives, nor is use of the `extend` keyword required.  I am anticipating that for v8 we may either (a) rename the federated schema printer to `Federation1SchemaPrinter`, or (b) append an option to print the schema in Federation v1 format, or similar.

cc: @Mithras @sungam3r

PR #3662, opened 2023-07-11
Allow nullable variables for non-null arguments that have defaults; allow scalar variable to list types

Shane32:
See:
- https://spec.graphql.org/June2018/#sec-Type-System.List

Sample query which would fail validation:

```gql
query ($arg: String) {
  dummy(dummyArg: $arg)
}
```

where `dummyArg` was defined as `[String]`.  However, these scenarios worked correctly:

```gql
query {
  dummy(dummyArg: "test")
}
```

and

```gql
query ($arg: [String]) {
  dummy(dummyArg: $arg)
}
```

where variables were:

```json
{
  "arg": "test"
}
```

Now the original query at the top passes validation and works as described in the GraphQL spec.

Note that the spec is slightly in conflict with the description of rule All Variable Usages Are Allowed.

See:
- https://github.com/graphql/graphql-spec/issues/1033
- https://github.com/graphql/graphql-spec/pull/1043

This PR also fixes some out-of-spec behavior during coercion of arguments that have default values:

```gql
# this should pass validation, providing that null is not provided to the arg variable
query ($arg: String = "abc") {
  dummy (nonNullArg: $arg)
}

# also this should pass validation, providing that null is not provided to the arg variable
query ($arg: String) {
  dummy (nonNullArgWithDefaultValue: $arg)
}

# also this should pass validation, providing that null is not provided to the arg variable
query ($arg: String = "abc") {
  dummy (inputArg: { nonNullValue: $arg })
}

# also this should pass validation, providing that null is not provided to the arg variable
query ($arg: String) {
  dummy (inputArg: { nonNullValueWithDefaultValue: $arg })
}
```

And I added a lot more tests to cover various of these scenarios.


Shane32:
Todo: double check we have tests that ensure that `[1,2,3]` doesn't match `[[Int]]`, either as literal, or variable, or with this new feature.

Shane32:
Todo: fix final failing test to prevent allowing null for a non-null input object field

src/GraphQL.Tests/Validation/ValidationContextTests.cs
```diff
@@ -0,0 +1,340 @@
...
+    [InlineData("query q08 ($arg: String!) { dummyNoDefault(arg: $arg) }", "{\"arg\":null}",
+        "Variable '$arg' is invalid. Received a null input for a non-null variable.")]
+    // q09 should fail (passing test) because null was explicitly passed to a non-null argument,
+    //   regardless of whether there is a variable default
+    [InlineData("query q09 ($arg: String = \"varDefault\") { dummyNoDefault(arg: $arg) }", "{\"arg\":null}",
+        "Variable '$arg' is invalid. Received a null input for a non-null variable.")]
+    [InlineData("query q10 { dummyList(arg: null) }", null,
+        "Argument 'arg' has invalid value. Expected '!', found null.")]
+    [InlineData("query q11 { dummyNestedList(arg: null) }", null,
+        "Argument 'arg' has invalid value. Expected '!', found null.")]
```
Shane32, line 298: These error messages could be cleaned up.

Shane32:
@sungam3r This PR contains bug fixes for type validation of variables:
- When a non-null argument (or input object field) has a default value, the variable type should allow a nullable type -- except with the `null` value is passed as the variable.
- And, if the value is not specified (with no variable default), it should use the argument default.

It also allows passing `String` variables to `[String]` arguments, but since this technically is not part of the spec, it is locked behind an experimental feature flag.  See:
- https://github.com/graphql/graphql-spec/issues/1033
- https://github.com/graphql/graphql-spec/pull/1043

I will merge this shortly unless you have comments.

PR #3664, opened 2023-07-11
Add NRT annotations to tests

Shane32:
This adds NRT annotations for all test projects.  Mostly just adding `?` or `!` where needed, but sometimes adding `.ShouldNotBeNull()` or `.ShouldHaveSingleItem()` or similar, as appropriate.  Did not attempt to add `.ShouldNotBeNull()` everywhere to eliminate all uses of `!`.

Only 3 annotation changes to GraphQL.NET library code for internal methods used by tests.  No functional changes.


src/GraphQL/Validation/Errors/KnownArgumentNamesError.cs
```diff
@@ -37,7 +37,7 @@ public KnownArgumentNamesError(ValidationContext context, GraphQLArgument node,
         {
         }
 
-        internal static string UnknownArgMessage(string argName, string fieldName, string type, string[] suggestedArgs)
+        internal static string UnknownArgMessage(string argName, string fieldName, string type, string[]? suggestedArgs)
```
Shane32, line 40: Change 1

src/GraphQL/Validation/Errors/KnownArgumentNamesError.cs
```diff
@@ -47,7 +47,7 @@ internal static string UnknownArgMessage(string argName, string fieldName, strin
             return message;
         }
 
-        internal static string UnknownDirectiveArgMessage(string argName, string directiveName, string[] suggestedArgs)
+        internal static string UnknownDirectiveArgMessage(string argName, string directiveName, string[]? suggestedArgs)
```
Shane32, line 50: Change 2

src/GraphQL/Validation/Errors/KnownTypeNamesError.cs
```diff
@@ -17,7 +17,7 @@ public KnownTypeNamesError(ValidationContext context, GraphQLNamedType node, str
         {
         }
 
-        internal static string UnknownTypeMessage(string type, string[] suggestedTypes)
+        internal static string UnknownTypeMessage(string type, string[]? suggestedTypes)
```
Shane32, line 20: Change 3

src/GraphQL.DataLoader.Tests/BatchDataLoaderTests.cs
```diff
@@ -383,4 +384,5 @@ public async Task Returns_Null_For_Null_Value_Types()
         var loader = new BatchDataLoader<int?, string>((_, _) => throw new Exception());
```
Shane32, line 384: This won't compile with NRT enabled because of the `where TKey : notnull` constraint.  So NRT is disabled here.

src/GraphQL.DataLoader.Tests/CollectionBatchLoaderTests.cs
```diff
@@ -160,4 +161,5 @@ public async Task Returns_Null_For_Null_Value_Types()
         var loader = new CollectionBatchDataLoader<int?, string>((_, _) => throw new Exception());
```
Shane32, line 161: Ditto

src/GraphQL.Tests/Builders/ConnectionBuilderTests.cs
```diff
@@ -16,21 +16,11 @@ public void should_have_name()
...
-        // race condition with does_not_throw_with_filtering_nameconverter test
-        try
-        {
-            exception = Should.Throw<ArgumentOutOfRangeException>(() => type.Connection<ObjectGraphType>(fieldName));
-        }
-        catch (ShouldAssertException)
-        {
-            System.Threading.Thread.Sleep(100); // wait a bit and retry
-            exception = Should.Throw<ArgumentOutOfRangeException>(() => type.Connection<ObjectGraphType>(fieldName));
-        }
```
Shane32, line 32: `does_not_throw_with_filtering_nameconverter` currently has `[Collection("StaticTests")]` defined on the class.  It should not have a race condition in the future with any code.  Very likely this was a patch that did not work and was subsequently fixed by adding the class to that collection.

src/GraphQL.Tests/Shared/JsonStringExtensions.cs
```diff
@@ -23,14 +21,13 @@ public static ExecutionResult ToExecutionResult(this string? json, ExecutionErro
...
 
+    [return: NotNullIfNotNull("json")]
     public static Dictionary<string, object?>? ToDictionary(this string? json)
     {
         if (json == null)
             return null;
 
         var ret = json.ToInputs();
-        if (ret == null)
-            return null;
```
Shane32, line 33: Removed condition is impossible to occur.  `ToInputs` returns Inputs.Empty if `json` was null, and `json` has already been verified to be not-null.

src/Tests.props
```diff
@@ -6,6 +6,8 @@
     <SingleTestPlatform Condition="'$(SingleTestPlatform)' == ''">false</SingleTestPlatform>
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
     <IsPackable>false</IsPackable>
+    <Nullable>enable</Nullable>
+    <NoWarn>$(NoWarn);CS8618</NoWarn>
```
Shane32, line 10: CS8618 warns when a non-nullable property can contain a null value due to not being initialized in the constructor -- e.g.:

```cs
class Test
{
    public string Abc { get; set; }
}
```

src/GraphQL/Validation/Errors/KnownArgumentNamesError.cs
gao-artur: Looks like `params string[]` fit better here

src/Tests.props
gao-artur: Shouldn't they be marked as nullable then?

gao-artur:
Resharper shows me some more errors in your branch. For example, in the `AllScalarGraphTypeTests`
image
This is fixed by changing `(short)default` to `default(short)`

In `DataLoaderSubscriptionTest`
image
Makes sense because of `FirstOrDefaultAsync`.

Also, for some reason, I'm getting this error in `JsonStringExtensions.ToDictionary` (net48 TFM only)
image

src/GraphQL.DataLoader.Tests/Types/OrderItemType.cs
```diff
@@ -17,7 +17,7 @@ public OrderItemType(IDataLoaderContextAccessor accessor, IProductsStore product
         Field<ProductType, Product>("Product")
             .ResolveAsync(ctx =>
             {
-                var loader = accessor.Context.GetOrAddBatchLoader<int, Product>("GetProductById",
+                var loader = accessor.Context!.GetOrAddBatchLoader<int, Product>("GetProductById",
```
gao-artur, line 20: I believe these will be reverted in `develop` branch? https://github.com/graphql-dotnet/graphql-dotnet/pull/3792

src/Tests.props
```diff
@@ -6,6 +6,8 @@
     <SingleTestPlatform Condition="'$(SingleTestPlatform)' == ''">false</SingleTestPlatform>
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
     <IsPackable>false</IsPackable>
+    <Nullable>enable</Nullable>
```
gao-artur, line 9: Please remove this property from the test projects. Or, maybe NRT can be enabled in the `Directory.Build.targets` unconditionally and removed from other places?

src/GraphQL/Validation/Errors/KnownArgumentNamesError.cs
Shane32: done

src/Tests.props
Shane32: No, it's just an input model.  Finally now we can mark `required`, but that didn't exist in prior versions of .NET.

src/GraphQL.DataLoader.Tests/Types/OrderItemType.cs
Shane32: They are unnecessary in the develop branch, right.

src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
```diff
@@ -575,13 +575,14 @@ public void toAST_ok(Type graphType, object value)
                 BigInteger bi => new GraphQLIntValue(bi),
                 float f => new GraphQLFloatValue(f),
                 double d => new GraphQLFloatValue(d),
-                _ => null
+                _ => throw new InvalidOperationException("Invalid value type")
```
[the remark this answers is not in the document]
Shane32: This is a test.  The exception only exists because it won't compile without a `_ => ...` clause.  The exception will never hit (or the test is broken).

src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
```diff
@@ -617,11 +618,11 @@ public void parseLiteral_other_ok(Type graphType, object value, object parsed)
             bool b => b ? new GraphQLTrueBooleanValue() : new GraphQLFalseBooleanValue(),
             double f => new GraphQLFloatValue(f),
             string s => new GraphQLStringValue(s),
-            _ => null
+            _ => throw new InvalidOperationException("Invalid value type")
```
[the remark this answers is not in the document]
Shane32: Ditto.  It might as well be `throw null!` and it would be just as effective.

src/Tests.props
Shane32: > Please remove this property from the test projects. 

Good catch; done.

> Or, maybe NRT can be enabled in the `Directory.Build.targets` unconditionally and removed from other places?

Maybe; let's get the bulk of the changes merged first.

Shane32:
I changed `(short)default` to `(short)0`.  And the `message.Data.ShouldNotBeNull();` error.

Shane32:
I don't understand why the NotNullIfNotNull would give you a problem.  Although it doesn't exist in .NET 4.8, PolySharp should be providing the source for this.  It works fine on my PC and compiles fine in the CI.  Can you compile from the command line?  Is it a VS issue or a Resharper issue?

src/GraphQL.Tests/Types/AllScalarGraphTypeTests.cs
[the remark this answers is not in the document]
Shane32: but i changed it anyway...

gao-artur:
> Is it a VS issue or a Resharper issue?

Probably Resharper. The build works fine.

gao-artur (approved):
Huge work! Looks great!

PR #3667, opened 2023-07-18
Fix SchemaBuilder so it reads directives from extension types

Shane32:


src/GraphQL/Utilities/SchemaBuilderExtensions.cs
```diff
@@ -49,20 +58,23 @@ public static bool AstTypeHasFields(this IProvideMetadata type)
...
                 }
             }
-
             return provider;
         }
 
         public static bool HasExtensionAstTypes(this IProvideMetadata type)
         {
-            return GetExtensionAstTypes(type).Count > 0;
+            return type.HasMetadata(EXTENSION_AST_METAFIELD);
```
Shane32, line 66:
```suggestion
            return type.HasMetadata(EXTENSION_AST_METAFIELD) && GetExtensionAstTypes(type).Count > 0;
```

PR #3691, opened 2023-08-14
Record data-loader execution time within Apollo Tracing results

Shane32:
Currently a resolver's `duration` property starts from the moment the resolver starts executing, and ends when the result is returned, even if the result is an `IDataLoaderResult`.  It does not include the amount of time that the data loader spent executing.  So, looking at trace results you can see gaps in the timeline, which is where the execution strategy executes the data loaders.

This PR modifies the per-resolver instrumentation to record the duration to include up to the point in time that the data loader has finished.  This means it is recording 'dead' time between the point that the resolver has finished and the point that the data loader starts.  (The execution engine tries to run as many resolvers as possible before executing data loaders.)  It also means that when batch data loaders are in use (the typical use case), multiple resolvers will be recording time within the `duration` property for the execution of the same data loader.

Ideally, the data loader execution time would be recorded separately, and only once for each batch data loader in use.  This presents two problems: (1) the [obsolete](https://github.com/apollographql/apollo-tracing) Apollo Tracing data format does not specify how to record this information, and so viewers like Playground would not be able to parse it, and (2) the execution engine is not aware of which `IDataLoaderResult` instances belong to which batch data loaders.

I am hoping for some feedback before merging this PR - thanks!


Shane32:
@gao-artur Let me know if you have any thoughts on this PR.  I'm not sure if the result is better, or worse!

gao-artur:
I have never used Apollo tracing and have never seen its graphical representation. But it sounds reasonable to include the DL duration in the field resolver because the field is not resolved until the DL completes.

src/GraphQL/Instrumentation/InstrumentFieldsMiddleware.cs
```diff
@@ -25,8 +27,36 @@ public class InstrumentFieldsMiddleware : IFieldMiddleware
                 { "path", context.ResponsePath },
             };
 
-            using (context.Metrics.Subject("field", name, metadata))
-                return await next(context).ConfigureAwait(false);
+            var marker = context.Metrics.Subject("field", name, metadata);
+            var ret = await next(context).ConfigureAwait(false);
```
gao-artur, line 31: Can this throw?

Shane32: Yes. Looks like it needs a try catch block…

gao-artur (approved):

src/GraphQL.DataLoader.Tests/ApolloTracingTests.cs
```diff
@@ -0,0 +1,49 @@
...
+            .AddSystemTextJson()
+            .UseApolloTracing());
+        var services = serviceCollection.BuildServiceProvider();
+        var serializer = services.GetRequiredService<IGraphQLTextSerializer>();
+        var executer = services.GetRequiredService<IDocumentExecuter>();
+        var ret = await executer.ExecuteAsync(new()
+        {
+            Query = "{test}",
+            RequestServices = services,
+        }).ConfigureAwait(false);
```
Shane32, line 43:
```suggestion
        });
```

src/GraphQL.DataLoader.Tests/ApolloTracingTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    // and the execution engine is not aware of which queued data loaders run as a single batch operation or not,
+    // making implementation difficult
+    [Fact]
+    public async Task DataLoaderTimeRecorded()
+    {
+        var query = new ObjectGraphType() { Name = "Query" };
+        query.Field<StringGraphType>("test")
+            .Resolve(_ => new SimpleDataLoader<string>(async _ =>
+            {
+                await Task.Delay(2000).ConfigureAwait(false);
```
Shane32, line 27:
```suggestion
                await Task.Delay(2000);
```

src/GraphQL.DataLoader.Tests/ApolloTracingTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    // note that (a) for batch data loaders, the time spent in the data loader is counted for all resolvers
+    // waiting on that data loader, and (b) any delay caused by the execution engine is counted within the
+    // resolver execution time
+    //
+    // perhaps it would be more ideal to separately record the data loader execution time within the apollo
+    // tracing output (e.g. 'extensions.tracing.execution.dataloaders'), but this does not currently happen,
+    // and the execution engine is not aware of which queued data loaders run as a single batch operation or not,
+    // making implementation difficult
+    [Fact]
+    public async Task DataLoaderTimeRecorded()
```
gao-artur, line 21: This test is [unstable](https://github.com/graphql-dotnet/graphql-dotnet/actions/runs/6846713537/job/18613753677)
```bash
[xUnit.net 00:00:04.41]     GraphQL.DataLoader.Tests.ApolloTracingTests.DataLoaderTimeRecorded [FAIL]
  Failed GraphQL.DataLoader.Tests.ApolloTracingTests.DataLoaderTimeRecorded [2 s]
  Error Message:
   Shouldly.ShouldAssertException : resolverData.Duration
    should be greater than or equal to
2000000000L
    but was
1999818500L
```

PR #3695, opened 2023-08-22
Add custom validation rule for denying introspection queries

Shane32:
This seems to be an often-asked question, so this PR adds a validation rule for denying introspection queries to the codebase.


Shane32:
No idea. I didn’t see a codecov report for the last few PRs we were doing either.

PR #3697, opened 2023-08-26
Add GraphQL.Analyzers

### LINKED ISSUE #3623: Roslyn analyzer + automatic code fixes
A few areas in this library usage could be improved using static code analysis to catch incorrect usages and provide automatic code fixes. Some examples:
-  When a type implements an interface, the static code analyzer can verify all fields from the interface are implemented and have the correct type. The automatic code fix can add missing fields with empty resolvers and suggest changing the field type on base or derived graph type in case of a mismatch
```c#
public class PersonGraphType : InterfaceGraphType
{
    public PersonGraphType()
    {
        Field<StringGraphType>("FirstName");
        Field<StringGraphType>("LastName");
        Field<IntGraphType>("Age");
    }
}

public class WorkerGraphType : ObjectGraphType<Worker>
{
    public WorkerGraphType()
    {
        Interface<PersonGraphType>();

        Field<StringGraphType>("FirstName").Resolve(x => x.Source.FirstName);
        // missing "LastName" field
        Field<IdGraphType>("Age");  // wrong type. Should be IntGraphType
    }
}
```

- The resolver of the field returns an incompatible type. For example, returning `int` for `ListGraphType<T>` field
```c#
//incompatible type, suggest changing return type to IntGraphType
Field<ListGraphType<IntGraphType>>("Numbers").Resolve(x => 5);
```

A few deprecations were made in the latest versions that are great candidates for automatic code fixes. For example:
- [A bunch of FieldXXX APIs were deprecated](https://graphql-dotnet.github.io/docs/migrations/migration7/#11-a-bunch-of-fieldxxx-apis-were-deprecated) can be fixed automatically (and this is the main reason I started thinking about Roslyn code fixes)
```c#
// GraphQL 5.x
Field<NonNullGraphType<StringGraphType>>(
  "name",
  "Argument name",
  resolve: context => context.Source!.Name);

// GraphQL 7.x
Field<NonNullGraphType<StringGraphType>>("name")
  .Description("Argument name")
  .Resolve(context => context.Source!.Name);
```
- The `Field<T>()` method without arguments was deprecated and can be fixed automatically
```c#
Field<StringGraphType>().Name("FirstName");

Field<StringGraphType>("FirstName");
```
- #3574. With Roslyn analyzer, the runtime error can be detected and automatically fixed in compile time.

I can think of more examples like missing `Resolve` method or returning `Task<T>` from the sync `Resolve` method.

gao-artur:
Closes #3623.

This PR focuses on rewriting obsolete methods + verifying resolvers are called on output types only. The best way to see what's done is by checking unit tests. In general, there are three analyzers:
- `FieldBuilderAnalyzer` replaces the `Field(name: "name", description..., resolver: ...` with `Field("name").Description(...).Resolve(...)`
- `FieldNameAnalyzer` replaces the `Field().Name("name")` with `Field("name")`
- `ResolverAnalyzer` detects `Resolve*` method usage on the non-output graph types, marks them as errors, and provides a code fix to remove them.

In code fixes, I tried to keep the original formatting as much as possible. For example, this code:
```c#
Field<StringGraphType>("name1", "description1", null,
    context => "text1");
```
Will be rewritten as:
```c#
Field<StringGraphType>("name1").Description("description1").Arguments(null)
    .Resolve(context => "text1");
```
But I also provided two configuration keys to reformat and skip `null` values. This should be configured in the `.editorconfig` file
```
dotnet_diagnostic.GQL004.reformat = true
dotnet_diagnostic.GQL004.skip_nulls = true
```
This will lead to the following result
```
Field<StringGraphType>("name1")
    .Description("description1")
    .Resolve(context => "text1");
```


src/GraphQL.Analyzers.CodeFixes/FieldBuilderCodeFixProvider.cs
```diff
@@ -0,0 +1,252 @@
...
+
+        foreach (var diagnostic in context.Diagnostics)
+        {
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+            bool isAsyncField = diagnostic.Properties.ContainsKey(Constants.Properties.IsAsync);
+            bool isDelegate = diagnostic.Properties.ContainsKey(Constants.Properties.IsDelegate);
+
+            var fieldInvocationExpression = (InvocationExpressionSyntax)root!.FindNode(diagnosticSpan);
+
+            const string codeFixTitle = "Rewrite obsolete 'Field' method";
```
gao-artur, line 40: This is the title that will be shown in the yellow lightbulb as the fix suggestion. I need help with choosing good words, both, for detected diagnostics and fix suggestions.

src/GraphQL.Analyzers.CodeFixes/GraphQL.Analyzers.CodeFixes.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <RootNamespace>GraphQL.Analyzers</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.7.*" />
```
gao-artur, line 10: Do you have any guidelines for version fixing?

src/GraphQL.Analyzers.Tests/Verifiers/CSharpAnalyzerVerifier`1+Test.cs
gao-artur, line 1: All files under the `Verifiers` folder were generated by the default template. I only changed the namespace to include `Verifiers` to stop resharper warnings and added some changes to satisfy this repo stylecop rules

src/GraphQL.Analyzers.Tests/VerifiersExtensions/CSharpCodeFixVerifier.cs
gao-artur, line 1: The files in `VerifiersExtensions` folder were added to add common customizations: 
https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/README.md#overriding-the-default-verifier-behavior

Alternatively it can be done directly in the `Verifiers` folder.

src/GraphQL.Analyzers/FieldBuilderAnalyzer.cs
```diff
@@ -0,0 +1,121 @@
...
+namespace GraphQL.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class FieldBuilderAnalyzer : DiagnosticAnalyzer
+{
+    // Violation: Field<T>("name", "description", ...)
+    // Fixed:     Field<T>("name").Description("description")...
+    public static readonly DiagnosticDescriptor DoNotUseObsoleteFieldMethods = new(
+        id: "GQL004",
+        title: "Don't use obsolete Field methods",
```
gao-artur, line 16: The title can be seen when disabling diagnostics. For example:
```c#
#pragma warning disable GQL004 // Don't use obsolete Field methods
Field<StringGraphType>(name: "name", resolve: context => "test");
#pragma warning restore GQL004 // Don't use obsolete Field methods
```

src/GraphQL.Analyzers/FieldBuilderAnalyzer.cs
```diff
@@ -0,0 +1,121 @@
...
+        messageFormat: "Use correct Field method overload",
```
gao-artur, line 17: This is the message that the user sees in the build output and warning window. Note it can include placeholders that need to be filled when reporting diagnostics.

docs2/site/docs/analyzers/GQL001_DefineTheNameInFieldMethod.md
gao-artur, line 1: [Documenting your analyzers.md](https://github.com/dotnet/roslyn-analyzers/blob/main/docs/Documenting%20your%20analyzers.md)

Shane32:
@gao-artur I can't wait to check out this PR!  Unfortunately, we had a storm and lost power this last weekend, so I didn't look at it.  As I'm pretty busy with work, I'm hoping to review it next weekend.

I did have a couple basic questions, which perhaps may be more evident if I had time to look at the code, but here goes:
- How are these analyzers used -- would they be referenced from the main GraphQL package, or must they be installed separately as another NuGet package?
- Once referenced, either directly or indirectly, do they automatically analyze the user's source code, or is configuration required?
- I assume that no changes to user's code is made until the user selects 'fix' or 'fix all' from Visual Studio?
- When publishing an application, is the analyzer DLLs excluded from the built/published project?

Perhaps we should add a page to the docs outlining the analyzers included and how to configure them.

Shane32:
I'd also recommend against using special characters such as ` and + within filenames.

src/GraphQL.Analyzers.Tests/Verifiers/CSharpAnalyzerVerifier`1+Test.cs
Shane32: Can we rename the filenames to match the class names?

gao-artur:
Sure, I'm not in a hurry. Take your time 🙂

Basically, the default analyzer template provides 3 projects: one for the analyzer: `GraphQL.Analyzers`, one for the automatic code fixes: `GraphQL.Analyzers.CodeFixes`, and one to pack them into a stand-alone nuget: `GraphQL.Analyzers.Package`. Alternatively, the content of the `Package` project can be added to the `GraphQL` project. This way, analyzers and code fixes will be installed with the main nuget, which is, in my opinion, preferred over the separate package. The analyzer is added to the `analyzers/dotnet/cs` folder, so it will be a compile-time dependency only and will be excluded from the run-time dependencies.

Nothing happens before the user decides to apply the code fixes. Any diagnostic rule can be disabled in `.editorconfig` if users are not ready yet to apply the changes.

I have added doc files for every rule as advised by the roslyn best practices. Perhaps, need to add a section about how to suppress the diagnostic. Anyway, documentation is not my strongest skill, so I'll need your help here.

gao-artur:
> I'd also recommend against using special characters such as ` and + within filenames.

These were added by the template, but I can rename files if you prefer.

Shane32:
> I have added doc files for every rule as advised by the roslyn best practices. Perhaps, need to add a section about how to suppress the diagnostic. Anyway, documentation is not my strongest skill, so I'll need your help here.

No problem.  Or ask [ChatGPT](https://chat.openai.com/).  If you've never used it, try this: type in "write a documentation page in markdown that explains what analzyers are now included with GraphQL.NET and how to configure them, based on this pull request summary:" and then paste all the text from your PR summary, with no changes.

docs2/site/docs/analyzers/Overview.md
gao-artur, line 1: @Shane32 Yeah, Chat is really good at writing docs. I'm finally going to adopt this tool in my daily job 🙂
The result needs tunning but I posted the initial version, so you can tell what to improve.

gao-artur: But before polishing the documentation, I'd like to ask you to review codeFixTitle's and `DiagnosticDescriptor`'s configuration, as it may affect the final documentation.

gao-artur:
There is something wrong with the `Microsoft.CodeAnalysis.*.XUnit` packages. When there is a mismatch in expected and actual diagnostics, I get an exception
```
System.MissingMethodException
Method not found: 'Void Xunit.Sdk.EqualException..ctor(System.Object, System.Object)'.
   at Microsoft.CodeAnalysis.Testing.Verifiers.EqualWithMessageException..ctor(Object expected, Object actual, String userMessage)
   at Microsoft.CodeAnalysis.Testing.Verifiers.XUnitVerifier.Equal[T](T expected, T actual, String message) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Testing.Verifiers.XUnit/XUnitVerifier.cs:line 49
   at Microsoft.CodeAnalysis.Testing.AnalyzerTest`1.VerifyDiagnosticResults(IEnumerable`1 actualResults, ImmutableArray`1 analyzers, DiagnosticResult[] expectedResults, IVerifier verifier) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs:line 360
   at Microsoft.CodeAnalysis.Testing.AnalyzerTest`1.VerifyDiagnosticsAsync(EvaluatedProjectState primaryProject, ImmutableArray`1 additionalProjects, DiagnosticResult[] expected, IVerifier verifier, CancellationToken cancellationToken) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs:line 273
   at Microsoft.CodeAnalysis.Testing.CodeFixTest`1.RunImplAsync(CancellationToken cancellationToken) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs:line 264
   at Microsoft.CodeAnalysis.Testing.AnalyzerTest`1.RunAsync(CancellationToken cancellationToken) in /_/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs:line 181
```

When I'm downgrading xUnit in `Tests.props` from `2.5.0` to `2.3.0` everything works as expected.

Shane32:
> The issue is described [here](https://github.com/dotnet/roslyn-sdk/issues/1099). Are you ok to temporarily downgrade the xUnit version until the issue is resolved? Alternatively, I can use NUnit or MSTest adaptors instead of xUnit, but it will require usings and attribute adjustments.

Yes, that would be fine.  I assume that the build fails with v2.5.0, and we can just monitor dependabot's merge requests when a new version of the affected packages are released.

src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
```diff
@@ -0,0 +1,613 @@
...
+        const string fix = """
+            using GraphQL.Types;
+
+            namespace Sample.Server;
+
+            public class MyGraphType : ObjectGraphType
+            {
+                public MyGraphType()
+                {
+                    Field<StringGraphType>("name").Description("description").Arguments(null)
```
Shane32, line 435: Can we eliminate `.Arguments(null)` here when `null` was passed to `arguments`?

src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
```diff
@@ -0,0 +1,613 @@
...
+
+            public class MyGraphType : ObjectGraphType
+            {
+                public MyGraphType()
+                {
+                    Field<StringGraphType>("name")
+                        .Arguments(new QueryArguments(new QueryArgument<StringGraphType>
+                        {
+                            Name = "argName"
+                        }))
```
Shane32, line 263: Have you considered an analyzer to convert arguments to the new syntax also?  I'm thinking probably a separate analyzer.  It would convert this old syntax to this:

```cs
                        .Argument<StringGraphType>("argName")
                        // additional arguments on separate lines
```

src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
Shane32: I see you already have that when `SkipNullsOption` is `true`.

src/GraphQL.Analyzers/HelpLinks.cs
```diff
@@ -0,0 +1,11 @@
+namespace GraphQL.Analyzers;
+
+public static class HelpLinks
+{
+    public const string DOCS_URL = "https://graphql-dotnet.github.io/docs/analizers";
```
Shane32, line 5:
```suggestion
    public const string DOCS_URL = "https://graphql-dotnet.github.io/docs/analyzers";
```

src/GraphQL.Analyzers/AnalyzerReleases.Shipped.md
```diff
@@ -0,0 +1,3 @@
+﻿; Shipped analyzer releases
+; https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
```
Shane32, line 2: I'm looking at the docs referenced here, and I'm not sure that these comments are supposed to remain here, but rather are just placeholder text when auto-creating these files.  Do you know?  It would certainly be preferred if they remained (certainly in some fashion), but they are not markdown comments, as far as I can tell.

Shane32:
I'd like to add a test that fails with XUnit 2.5.0, such as this:

```cs
using VerifyCS = GraphQL.Analyzers.Tests.VerifiersExtensions.CSharpCodeFixVerifier<
    GraphQL.Analyzers.ResolverAnalyzer,
    GraphQL.Analyzers.ResolverCodeFixProvider>;

namespace GraphQL.Analyzers.Tests;

public class XUnitTests
{
    [Fact]
    public async Task VerifyProperExceptionReturned()
    {
        // see https://github.com/dotnet/roslyn-sdk/issues/1099
        // problem exists when using xunit 2.5.0 with Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit 1.1.1

        string source = $$"""
            using System;
            using System.Threading.Tasks;
            using GraphQL;
            using GraphQL.MicrosoftDI;
            using GraphQL.Types;

            namespace Sample.Server;

            public class MyGraphType : InterfaceGraphType
            {
                public MyGraphType() =>
                    Field<StringGraphType, string>("Test").Resolve(_ => "");
            }
            """
        ;

        try
        {
            await VerifyCS.VerifyAnalyzerAsync(source).ConfigureAwait(false);
        }
        catch (MissingMethodException)
        {
            throw;
        }
        catch
        {
            return;
        }
        Assert.Fail("An exception should have been thrown.");
    }
}
```

ok?

docs2/site/docs/analyzers/GQL004_DoNotUseObsoleteFieldMethods.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+### Configure `null` values handling
+
+The `skip_nulls` option can be used to remove unnecessary `null` values assignments
+
+```ini
+[*.cs]
+dotnet_diagnostic.GQL004.skip_nulls = true
+```
```
Shane32, line 122: What is default?  Can we configure so `skip_nulls` is `true` by default?  Generally I would expect that null was provided only to skip an argument to the Field method.

src/GraphQL.Analyzers.CodeFixes/GraphQL.Analyzers.CodeFixes.csproj
Shane32: So far in this solutions we have fixed versions for all published packages and tests.  For sample projects, we let the version float like you show here.  Dependabot will create pull requests for each version bump, and we can ensure that the bump passes tests before merging.

src/GraphQL.Analyzers.Package/tools/install.ps1
```diff
@@ -0,0 +1,272 @@
+param($installPath, $toolsPath, $package, $project)
```
Shane32, line 1: Is this file a typical file copied from somewhere?  The signature block was included with the file, I assume?

src/GraphQL.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,12 @@
+; Unshipped analyzer release
+; https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+GQL001 | FieldNameDefinition | Warning | FieldNameAnalyzer, [Documentation](https://graphql-dotnet.github.io/docs/analizers/GQL001_DefineTheNameInFieldMethod)
+GQL002 | FieldNameDefinition | Warning | FieldNameAnalyzer, [Documentation](https://graphql-dotnet.github.io/docs/analizers/GQL002_NameMethodInvocationCanBeRemoved)
+GQL003 | FieldNameDefinition | Warning | FieldNameAnalyzer, [Documentation](https://graphql-dotnet.github.io/docs/analizers/GQL003_DifferentNamesDefinedByFieldAndNameMethods)
```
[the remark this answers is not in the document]
Shane32: Do you want to create a PR for deprecating the Name method?  What do you think?

Shane32:
I've reviewed everything; looks good.  I had a few comments we should answer before merging; see above.

@joemcbride is it okay to publish a new package called GraphQL.Analyzers ?

src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
gao-artur: Yeah, I thought about it but didn't implement it yet because I didn't need it in our code base. But I can do it before we merge this PR.

src/GraphQL.Analyzers/AnalyzerReleases.Shipped.md
gao-artur: This was added automatically when these files were created (also automatically). I think we can remove them because the built-in analyzer detects any change to analyzer descriptors and provides a fix to update these files.

gao-artur:
> I'd like to add a test that fails with XUnit 2.5.0, such as this:
> 
> ```cs
> using VerifyCS = GraphQL.Analyzers.Tests.VerifiersExtensions.CSharpCodeFixVerifier<
>     GraphQL.Analyzers.ResolverAnalyzer,
>     GraphQL.Analyzers.ResolverCodeFixProvider>;
> 
> namespace GraphQL.Analyzers.Tests;
> 
> public class XUnitTests
> {
>     [Fact]
>     public async Task VerifyProperExceptionReturned()
>     {
>         // see https://github.com/dotnet/roslyn-sdk/issues/1099
>         // problem exists when using xunit 2.5.0 with Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit 1.1.1
> 
>         string source = $$"""
>             using System;
>             using System.Threading.Tasks;
>             using GraphQL;
>             using GraphQL.MicrosoftDI;
>             using GraphQL.Types;
> 
>             namespace Sample.Server;
> 
>             public class MyGraphType : InterfaceGraphType
>             {
>                 public MyGraphType() =>
>                     Field<StringGraphType, string>("Test").Resolve(_ => "");
>             }
>             """
>         ;
> 
>         try
>         {
>             await VerifyCS.VerifyAnalyzerAsync(source).ConfigureAwait(false);
>         }
>         catch (MissingMethodException)
>         {
>             throw;
>         }
>         catch
>         {
>             return;
>         }
>         Assert.Fail("An exception should have been thrown.");
>     }
> }
> ```
> 
> ok?

Good idea. I'd just remove the dependency on gql analyzers, so this test won't change when these analyzers are removed.
```c#
const string source = """
    using System;

    namespace Sample.Server;

    {
    }
    """;

// error CS0246: The type or namespace name 'ObjectGraphType' could not be found (are you missing a using directive or an assembly reference?)
await Assert
    .ThrowsAsync<EqualWithMessageException>(async () => await VerifyCS.VerifyAnalyzerAsync(source).ConfigureAwait(false))
    .ConfigureAwait(false);
```

Shane32:
> Good idea. I'd just remove the dependency on gql analyzers, so this test won't change when these analyzers are removed

Good - the simpler the better.

Shane32:
You’ve directly referenced EqualWithMessageException - will that exception change if MS references the xunit source repo?

docs2/site/docs/analyzers/GQL004_DoNotUseObsoleteFieldMethods.md
gao-artur: The `null` could also be provided to the `description` argument. For example
```c#
{
    public MyGraphType()
    {
        Field<StringGraphType>(
            "name",
            null,
            null,
            context => "text");
    }
}
```
But I agree, that removing redundant nulls by default sounds like a better choice. Do you prefer to keep the `skip_nulls` option name or rename it to `preserve_nulls` in that case?

src/GraphQL.Analyzers.Package/tools/install.ps1
gao-artur: This file was added by the default template. I didn't really try to understand this script.

gao-artur:
> You’ve directly referenced EqualWithMessageException - will that exception change if MS references the xunit source repo?

I don't think so because it will be a breaking change. This exception is defined by MS, and it's derived from `EqualException` defined by xunit.

src/GraphQL.Analyzers/AnalyzerReleases.Unshipped.md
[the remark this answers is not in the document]
gao-artur: Sure. Will do.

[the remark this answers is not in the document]
gao-artur: For the categories: I tried to find the best practices for category naming but found nothing. Categories are global so that the same names can be used by different analyzers. Maybe it's better to create a single category for all GraphQL analyzers? Or maybe prefix category names with `graphql-` to distinguish them from other analyzers?

docs2/site/docs/analyzers/GQL004_DoNotUseObsoleteFieldMethods.md
Shane32: > Do you prefer to keep the `skip_nulls` option name or rename it to `preserve_nulls` in that case?

Either way; your call.

src/GraphQL.Analyzers/AnalyzerReleases.Unshipped.md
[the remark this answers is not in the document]
Shane32: Well, looking at built-in analyzer rules, I see categories named "Design", "Style", "CodeQuality", etc.  Offhand I'd probably select one of these categories.  ChatGPT has this to say:

> When creating a custom .NET code analyzer, you can choose to categorize your analyzers under any category name you see fit; however, it's often beneficial to use or be inspired by the categories used by the built-in analyzer rules to keep things consistent and intuitive for other developers who might use your analyzers. Below are some of the common categories used by built-in analyzer rules in .NET:
> 
> 1. **Design** - Issues related to API design, such as naming or inheritability.
> 2. **Performance** - Issues that could potentially impact the performance of the code, such as unnecessary allocations or inefficient algorithms.
> 3. **Usage** - Issues related to the usage of APIs, such as passing incorrect arguments to a method.
> 4. **Documentation** - Issues related to XML documentation comments, such as missing or improperly formatted comments.
> 5. **Maintainability** - Issues that might make the code harder to maintain, such as complexity or dead code.
> 6. **Reliability** - Issues that might affect the reliability of the code, such as improper exception handling.
> 7. **Style** - Issues related to coding style, such as naming conventions or formatting.
> 8. **Globalization** - Issues that might affect the global usability of the code, such as not supporting multiple cultures or languages.
> 9. **Interoperability** - Issues that might affect the interoperability of the code with other components or technologies, such as P/Invoke or COM.
> 10. **Security** - Issues that might introduce security vulnerabilities, such as SQL injection or buffer overflows.
> 11. **Correctness** - Issues that relate to the correctness of the code, such as logic errors or incorrect algorithm implementations.
> 
> These categories correspond to the different rule categories used by .NET analyzers, and each category usually contains several different rules that are related to the general theme of the category. It is good practice to choose a category name that accurately describes the kind of issues that your analyzer will detect.
> 
> For an analyzer that suggests alternatives to obsolete method calls, it would be most appropriate to categorize it under the Usage category. This category is designed to encompass issues related to the incorrect or suboptimal usage of APIs, which includes using obsolete APIs when better alternatives are available.
> 
> By categorizing your analyzer under the "Usage" category, you help to signal to developers that the analyzer is identifying issues related to how APIs are being used in the code, which can include using APIs that have been marked as obsolete. This can facilitate a better understanding of the purpose and function of your analyzer.

Is there any particular reason not to use a standard category name?  You already prefixed the rule id with GQL and I like that, so the rules should be distinct from built-in rules.  My second choice would be simply GraphQL as the category name.  I'll defer to you, however.

src/GraphQL.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,12 @@
+; Unshipped analyzer release
+; https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+GQL001 | FieldNameDefinition | Warning | FieldNameAnalyzer, [Documentation](https://graphql-dotnet.github.io/docs/analizers/GQL001_DefineTheNameInFieldMethod)
```
Shane32, line 8:
```suggestion
GQL001 | FieldNameDefinition | Warning | FieldNameAnalyzer, [Documentation](https://graphql-dotnet.github.io/docs/analyzers/GQL001_DefineTheNameInFieldMethod)
```
and similar for the next few lines

src/GraphQL.Analyzers/AnalyzerReleases.Unshipped.md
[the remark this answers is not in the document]
gao-artur: Then, I'll change everything to `Usage`

src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
gao-artur: I think I'll postpone this to a separate PR. Since the old syntax is not deprecated, this should probably be implemented as a `CodeRefactoringProvider` rather than `CodeFixProvider`. The difference is that refactoring isn't backed by diagnostic rules and doesn't produce compiler warnings. It will appear under the "Quick Actions" menu similar to this
image

Shane32: > I think I'll postpone this to a separate PR. Since the old syntax is not deprecated, this should probably be implemented as a `CodeRefactoringProvider` rather than `CodeFixProvider`.

Sounds good.

docs2/site/docs/analyzers/GQL004_DoNotUseObsoleteFieldMethods.md
```diff
@@ -0,0 +1,130 @@
...
+
+```c#
+Field<StringGraphType>("name")
+    .Description("description")
+    .Resolve(context => "text");
+```
+
+### Configure `null` values handling
+
+The `skip_nulls` option can be set to `false` to preserve `null` values assignments. The default value if `true`.
```
Shane32, line 116:
```suggestion
The `skip_nulls` option can be set to `false` to preserve `null` values assignments. The default value is `true`.
```

Shane32:
Based on our discussion [here](https://github.com/graphql-dotnet/graphql-dotnet/pull/3713#issuecomment-1722474266), let's configure GraphQL to reference GraphQL.Analyzers, ensuring that there are no dependencies listed for GraphQL.Analyzers and there is no additional code added to runtime builds.

> Look at [Microsoft.VisualStudio.Threading](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading) for example. It depends on [Microsoft.VisualStudio.Threading.Analyzers](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers/). The analyzers package shows no dependencies even though [it has few](https://github.com/microsoft/vs-threading/blob/main/src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj#L24-L30). When I install the `GraphQL.Analyzers` package, the VS also doesn't show any dependencies for this package. It also doesn't appear in the `bin` or `publish`.

gao-artur:
I tried applying the same [trick](https://github.com/microsoft/vs-threading/blob/main/src/Microsoft.VisualStudio.Threading/Microsoft.VisualStudio.Threading.csproj#L51-L57) MS does in `Microsoft.VisualStudio.Threading` package. The code builds without errors in VS but fails to build/pack in cli with package downgrade errors
```
MSBuild version 17.7.0-preview-23281-03+4ce2ff1f8 for .NET
  Determining projects to restore...
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605: Warning As Error: Detected package downgrade: System.Runtime.CompilerServices.Unsafe from 6.0.0 to 4.5.3. Reference the package directly from the project to select a different version. 
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605:  GraphQL -> GraphQL.Analyzers.CodeFixes -> Microsoft.CodeAnalysis.CSharp.Workspaces 4.7.0 -> Microsoft.CodeAnalysis.Common 4.7.0 -> System.Runtime.CompilerServices.Unsafe (>= 6.0.0) 
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605:  GraphQL -> System.Runtime.CompilerServices.Unsafe (>= 4.5.3)
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605: Warning As Error: Detected package downgrade: System.Runtime.CompilerServices.Unsafe from 6.0.0 to 4.5.3. Reference the package directly from the project to select a different version. 
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605:  GraphQL -> GraphQL.Analyzers.CodeFixes -> GraphQL.Analyzers_Internal -> Microsoft.CodeAnalysis.CSharp 4.7.0 -> Microsoft.CodeAnalysis.Common 4.7.0 -> System.Runtime.CompilerServices.Unsafe (>= 6.0.0) 
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605:  GraphQL -> System.Runtime.CompilerServices.Unsafe (>= 4.5.3)
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605: Warning As Error: Detected package downgrade: Microsoft.Bcl.AsyncInterfaces from 7.0.0 to 1.1.1. Reference the package directly from the project to select a different version. 
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605:  GraphQL -> GraphQL.Analyzers.CodeFixes -> Microsoft.CodeAnalysis.CSharp.Workspaces 4.7.0 -> Microsoft.CodeAnalysis.Workspaces.Common 4.7.0 -> Microsoft.Bcl.AsyncInterfaces (>= 7.0.0) 
C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj : error NU1605:  GraphQL -> Microsoft.Bcl.AsyncInterfaces (>= 1.1.1)
  Failed to restore C:\github\graphql-dotnet\src\GraphQL\GraphQL.csproj (in 269 ms).
  2 of 3 projects are up-to-date for restore.

```

gao-artur:
Ok, I see what they did there. They use `CodeFixes` project instead of the `Package` project with all the hacks applied there. Trying to reproduce here...

gao-artur:
I don't know why, but when I produce a local analyzers package and install it somewhere, VS doesn't see any diagnostics rule inside it
image
That's how it should look like (installed from our internal nuget repo)
image

Shane32:
For Microsoft.VisualStudio.Threading.Analyzers, the files are in the folder `analyzers/cs` whereas in GraphQL.Analyzers they are in `analyzers/dotnet/cs`

image

image

gao-artur:
The problem was `Microsoft.CodeAnalysis.CSharp.*` packages version. It doesn't work with 4.7.0 but works with 4.6.0. No idea why. Maybe it depends on the installed SDK version. With the latest [commit](https://github.com/graphql-dotnet/graphql-dotnet/pull/3697/commits/5a731c20bf97981dafaa90ae956a6f0884c669f3) the nuspec files look like this now:

<details>
  <summary>GraphQL.nuspec (GraphQL.Analyzers included)</summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
  <metadata>
    <id>GraphQL</id>
    <version>7.4.1-preview</version>
    <authors>Joe McBride</authors>
    <license type="expression">MIT</license>
    <licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
    <icon>logo.64x64.png</icon>
    <readme>README.md</readme>
    <description>GraphQL for .NET</description>
    <releaseNotes>See https://github.com/graphql-dotnet/graphql-dotnet/releases and https://graphql-dotnet.github.io/docs/migrations/migration7</releaseNotes>
    <copyright>Copyright (c) 2015-2023 Joseph T. McBride, Ivan Maximov, Shane Krueger, et al. All rights reserved.</copyright>
    <tags>GraphQL json api</tags>
    <repository type="git" url="https://github.com/gao-artur/graphql-dotnet.git" commit="5a731c20bf97981dafaa90ae956a6f0884c669f3" />
    <dependencies>
      <group targetFramework="net5.0">
        <dependency id="GraphQL.Analyzers" version="7.4.1-preview" include="All" />
        <dependency id="GraphQL-Parser" version="8.4.0" exclude="Build,Analyzers" />
      </group>
      <group targetFramework="net6.0">
        <dependency id="GraphQL.Analyzers" version="7.4.1-preview" include="All" />
        <dependency id="GraphQL-Parser" version="8.4.0" exclude="Build,Analyzers" />
      </group>
      <group targetFramework=".NETStandard2.0">
        <dependency id="GraphQL.Analyzers" version="7.4.1-preview" include="All" />
        <dependency id="GraphQL-Parser" version="8.4.0" exclude="Build,Analyzers" />
        <dependency id="Microsoft.Bcl.AsyncInterfaces" version="1.1.1" exclude="Build,Analyzers" />
        <dependency id="Microsoft.CSharp" version="4.7.0" exclude="Build,Analyzers" />
        <dependency id="System.ComponentModel.Annotations" version="4.7.0" exclude="Build,Analyzers" />
        <dependency id="System.Memory" version="4.5.5" exclude="Build,Analyzers" />
      </group>
      <group targetFramework=".NETStandard2.1">
        <dependency id="GraphQL.Analyzers" version="7.4.1-preview" include="All" />
        <dependency id="GraphQL-Parser" version="8.4.0" exclude="Build,Analyzers" />
        <dependency id="Microsoft.CSharp" version="4.7.0" exclude="Build,Analyzers" />
        <dependency id="System.ComponentModel.Annotations" version="4.7.0" exclude="Build,Analyzers" />
        <dependency id="System.Runtime.CompilerServices.Unsafe" version="4.5.3" exclude="Build,Analyzers" />
      </group>
    </dependencies>
  </metadata>
</package>
```

</details>

<details>
  <summary>GraphQL.Analyzers .nuspec (no dependencies section)</summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
  <metadata>
    <id>GraphQL.Analyzers</id>
    <version>7.4.1-preview</version>
    <authors>GraphQL.Analyzers.Package</authors>
    <developmentDependency>true</developmentDependency>
    <license type="expression">MIT</license>
    <licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
    <icon>logo.64x64.png</icon>
    <readme>README.md</readme>
    <description>GraphQL.Analyzers</description>
    <releaseNotes>See https://github.com/graphql-dotnet/graphql-dotnet/releases and https://graphql-dotnet.github.io/docs/migrations/migration7</releaseNotes>
    <tags>GraphQL.Analyzers, analyzers</tags>
    <repository type="git" url="https://github.com/gao-artur/graphql-dotnet.git" commit="f2fe96f6cb0133299d4995b6deaa69972e9c9bd3" />
  </metadata>
</package>
```

</details>

src/GraphQL.Analyzers/AnalyzerReleases.Shipped.md
gao-artur: Tried to change it to markdown but got this error

> Analyzer release file 'AnalyzerReleases.Shipped.md' has a missing or invalid release header '# Shipped analyzer releases'

Shane32:
Looking good.  I think the only thing remaining to be done here is to align some of the nuspec's xml tags like 'copyright' with the ones in the GraphQL file.  We can do that as a separate PR or in this one.

gao-artur:
Will do. Btw, why not set them in the `Directory.Build.props` file?

Shane32:
> Will do. Btw, why not set them in the `Directory.Build.props` file?

🤷‍♂️ Seems logical to me.

Shane32 (approved):

gao-artur:
@Shane32 I have another [analyzer](https://github.com/gao-artur/graphql-dotnet/compare/analyzer...gao-artur:graphql-dotnet:input-graphtype-analyzer) that is based on this PR's code. Is it ok to merge it here, or are you planning to merge this PR anytime soon?

Shane32:
I’ll merge this and you can post a new PR

PR #3713, opened 2023-09-16
Deprecate Name method

### LINKED ISSUE #3712: Deprecate the Name method
Do you want to create a PR for deprecating the Name method?  What do you think?

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/pull/3697#discussion_r1327979935_

gao-artur:
Closes #3712 
A couple of questions.
1. There are the following [methods](https://github.com/graphql-dotnet/graphql-dotnet/blob/c2cc6671bf5eea6977f8865d436e2cbbe0659ca8/src/GraphQL/Builders/FieldBuilder.cs#L55)
```c#
public static FieldBuilder<TSourceType, TReturnType> Create(IGraphType type, string name = "default")
public static FieldBuilder<TSourceType, TReturnType> Create([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type? type = null, string name = "default")
```
After deprecating the `Name` the `"default"` value should be removed, but it will be a breaking change. Should I open an additional issue to do that in v8/9/whatever? Or should I deprecate it too and create an overload with the `name` as a first argument?
```c#
public static FieldBuilder<TSourceType, TReturnType> Create(string name, IGraphType type)
```
The second option will be more consistent with the `Field(string name, IGraphType type)` and `Field(string name, Type type)` methods.
2. [ConnectionBuilder](https://github.com/graphql-dotnet/graphql-dotnet/blob/c2cc6671bf5eea6977f8865d436e2cbbe0659ca8/src/GraphQL/Builders/ConnectionBuilder_Typed.cs) is not consistent with the `FieldBuilder`. Should it also be reworked to require `name` as a mandatory argument? If the answer is yes, it also needs an analyzer with the code fix.

### DIFF
--- a/src/GraphQL/Builders/FieldBuilder.cs
+++ b/src/GraphQL/Builders/FieldBuilder.cs
@@ -52,6 +52,7 @@ namespace GraphQL.Builders
         /// </summary>
         /// <param name="type">The graph type of the field.</param>
         /// <param name="name">The name of the field.</param>
+        [Obsolete("Please use the overload that accepts the name as the first argument.")]
         public static FieldBuilder<TSourceType, TReturnType> Create(IGraphType type, string name = "default")
         {
             var fieldType = new FieldType
@@ -62,8 +63,35 @@ namespace GraphQL.Builders
             return new FieldBuilder<TSourceType, TReturnType>(fieldType);
         }
 
+        /// <summary>
+        /// Returns a builder for a new field.
+        /// </summary>
+        /// <param name="name">The name of the field.</param>
+        /// <param name="type">The graph type of the field.</param>
+        public static FieldBuilder<TSourceType, TReturnType> Create(string name, IGraphType type)
+        {
+            var fieldType = new FieldType
+            {
+                Name = name,
+                ResolvedType = type,
+            };
+            return new FieldBuilder<TSourceType, TReturnType>(fieldType);
+        }
+
         /// <inheritdoc cref="Create(IGraphType, string)"/>
+        [Obsolete("Please use the overload that accepts the name as the first argument.")]
         public static FieldBuilder<TSourceType, TReturnType> Create([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type? type = null, string name = "default")
+        {
+            var fieldType = new FieldType
+            {
+                Name = name,
+                Type = type,
+            };
+            return new FieldBuilder<TSourceType, TReturnType>(fieldType);
+        }
+
+        /// <inheritdoc cref="Create(string, IGraphType)"/>
+        public static FieldBuilder<TSourceType, TReturnType> Create(string name, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type? type = null)
         {
             var fieldType = new FieldType
             {
--- a/src/GraphQL/Builders/FieldBuilder.cs
+++ b/src/GraphQL/Builders/FieldBuilder.cs
@@ -86,6 +86,7 @@ namespace GraphQL.Builders
         /// <summary>
         /// Sets the name of the field.
         /// </summary>
+        [Obsolete("Please configure the filed name by providing the name as an argument to the 'Field' method.")]
         public virtual FieldBuilder<TSourceType, TReturnType> Name(string name)
         {
             FieldType.Name = name;
--- a/src/GraphQL/Builders/FieldBuilder.cs
+++ b/src/GraphQL/Builders/FieldBuilder.cs
@@ -114,7 +114,7 @@ namespace GraphQL.Builders
         /// <summary>
         /// Sets the name of the field.
         /// </summary>
-        [Obsolete("Please configure the filed name by providing the name as an argument to the 'Field' method.")]
+        [Obsolete("Please configure the field name by providing the name as an argument to the 'Field' method.")]
         public virtual FieldBuilder<TSourceType, TReturnType> Name(string name)
         {
             FieldType.Name = name;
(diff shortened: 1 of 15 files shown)


Shane32:
So long as both can occur without breaking binary changes, it sounds good to me.

gao-artur:
I didn't remove the default names from the `ConnectionBuilder.Create` methods. Removing them won't break the binary compatibility but will break the source compatibility. The automatic code fix can help to migrate the name from the `Name` method, but it's probably not ok to do so in non-major version change. It also depends on how you will ship the analyzer package: if it will be a stand-alone package that needs to be installed explicitly, then it is better to wait with this change for v8. But if the analyzer package will be referenced by the main package and installed transitively, then we probably can do it earlier.
What do you think?

Shane32:
> But if the analyzer package will be referenced by the main package and installed transitively, then we probably can do it earlier.

Can you explain how this would work?  Just reference GraphQL.Analyzers from GraphQL?

Historically GraphQL.NET has put an emphasis on eliminating dependencies.  The current version only has a single dependency when used with .NET 5.0 or later, that being GraphQL-Parser, which itself has no dependencies for .NET Standard 2.1+.  Even when using the server library, the only non-GraphQL.NET dependency (with .NET 6) is Microsoft.Extensions.Options >= 2.0.0.

So by referencing this from the main package, does this create an indirect dependency on Microsoft.CodeAnalysis.CSharp ?

How do other mainstream libraries deal with this question?  Are there any MS docs that discuss this?  What effect do you think that would have on users?

Shane32:
I think you partially answered my question already here:

> Alternatively, the content of the Package project can be added to the GraphQL project. This way, analyzers and code fixes will be installed with the main nuget, which is, in my opinion, preferred over the separate package. The analyzer is added to the `analyzers/dotnet/cs` folder, so it will be a compile-time dependency only and will be excluded from the run-time dependencies.

gao-artur:
Look at [Microsoft.VisualStudio.Threading](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading) for example. It depends on [Microsoft.VisualStudio.Threading.Analyzers](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers/). The analyzers package shows no dependencies even though [it has few](https://github.com/microsoft/vs-threading/blob/main/src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj#L24-L30). When I install the `GraphQL.Analyzers` package, the VS also doesn't show any dependencies for this package. It also doesn't appear in the `bin` or `publish`.

Shane32:
> Look at [Microsoft.VisualStudio.Threading](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading) for example. It depends on [Microsoft.VisualStudio.Threading.Analyzers](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers/). The analyzers package shows no dependencies even though [it has few](https://github.com/microsoft/vs-threading/blob/main/src/Microsoft.VisualStudio.Threading.Analyzers/Microsoft.VisualStudio.Threading.Analyzers.csproj#L24-L30). When I install the `GraphQL.Analyzers` package, the VS also doesn't show any dependencies for this package. It also doesn't appear in the `bin` or `publish`.

That sounds perfect!

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -86,6 +114,7 @@ protected FieldBuilder(FieldType fieldType)
         /// <summary>
         /// Sets the name of the field.
         /// </summary>
+        [Obsolete("Please configure the filed name by providing the name as an argument to the 'Field' method.")]
```
Shane32, line 117:
```suggestion
        [Obsolete("Please configure the field name by providing the name as an argument to the 'Field' method.")]
```

Shane32:
Looks good to me.  @gao-artur All set with this PR ?

Shane32 (approved):

gao-artur:
What did you decide regarding default name values in `ConnectionBuilder.Create` methods?

Shane32:
> What did you decide regarding default name values in `ConnectionBuilder.Create` methods?

I'm not exactly sure what the question is.  All of the methods that have default names, or no names, or the name as the second argument instead of the first, have been deprecated with no changes (for binary and source compatibility), and replaced by methods with the name as the first argument.  Sounds good to me or did I miss something?

Shane32:
Hmm... looks like the apiapprovals file is out of date.

gao-artur:
> I'm not exactly sure what the question is

I'm talking about `Create` methods in `ConnectionBuilder<TSourceType>` class. [Here](https://github.com/gao-artur/graphql-dotnet/blob/deprecate-name-method/src/GraphQL/Builders/ConnectionBuilder.cs#L126), [here](https://github.com/gao-artur/graphql-dotnet/blob/deprecate-name-method/src/GraphQL/Builders/ConnectionBuilder.cs#L135) and [here](https://github.com/gao-artur/graphql-dotnet/blob/deprecate-name-method/src/GraphQL/Builders/ConnectionBuilder.cs#L146).
All the other builders forward their calls to these methods. I have two options:
1. Leave them as is and remove the default values in v8
2. Remove the default values now and create a code fix to move the name from the `Name` method to `Create` before releasing this change. It will be binary-compatible but not source-compatible.

By the way, this class has a comment [TODO: Remove in v5](https://github.com/gao-artur/graphql-dotnet/blob/deprecate-name-method/src/GraphQL/Builders/ConnectionBuilder.cs#L94) but it's not obsolete.

Shane32:
If removing the default value is not a binary breaking change, then perhaps we do this:
1. Remove the default value (binary compatible), and
2. Add obsolete overloads without a name argument (source compatible)

This way when we are ready to release v8, we can just look for all obsolete methods and delete them, rather than having to remember to make this change.

gao-artur:
Sounds good!

Shane32 (approved):

Shane32:
I'll leave these PRs up for a few days before merging, in case anyone else has comments.

PR #3720, opened 2023-09-25
InputGraphTypeAnalyzer

gao-artur:
Please see the `GQL006` readme file for description.


gao-artur:
It also makes sense to analyze types with open generic with constraints, but I gave up after ~10 hours of failed attempts.
For example
```c#
public class MyBaseInputObjectGraphType<TSourceType> : InputObjectGraphType<TSourceType>
    where TSourceType : MyBaseSourceType
{
    public MyBaseInputObjectGraphType()
    {
        Field<StringGraphType>("NameNotExistsOnMyBaseSourceType");
    }
}
```

So, for now, such types are not analyzed.

docs2/site/docs/analyzers/GQL006_InvalidInputField.md
```diff
@@ -0,0 +1,115 @@
...
+## Rule description
+
+This diagnostic is reported when the input field name cannot be mapped to any field, property, or constructor parameter of the `TSourceType` type. The name comparison is case-insensitive. The input field can be mapped when all the following rules are respected:
+
+### Mapping rules for property/field
+
+- Must have the exact same name (case-insensitive)
+- Must be public
+- Must not be static
+- Must be settable (have a public setter / not a constant)
```
Shane32, line 24: Currently, I believe the `ToObject` code allows deserializing to get-only properties that are set via a constructor argument.  Have you checked that?

Shane32: Oh, I see you have 'rules for constructor parameter' below, so I assume you have.

gao-artur: Btw, it should be possible to replace the `ToObject` implementation with a source generator similar to how MS did with a [Configure](https://andrewlock.net/exploring-the-dotnet-8-preview-using-the-new-configuration-binder-source-generator/#looking-at-the-generated-code) method in .NET 8. It will remove the use of reflection and significantly improve the deserialization [performance](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/#configuration).

src/GraphQL.Analyzers/Extensions.cs
```diff
@@ -117,6 +93,11 @@ public static Location GetMethodInvocationLocation(this MemberAccessExpressionSy
             TextSpan.FromBounds(methodNameLocation.SourceSpan.Start, argsLocation.SourceSpan.End));
     }
 
+    public static HashSet<T> ToHashSet<T>(this IEnumerable<T> source) => new(source);
+
+    public static HashSet<T> ToHashSet<T>(this IEnumerable<T> source, IEqualityComparer<T> comparer)
+        => new(source, comparer);
+
```
gao-artur, line 100: .NET Standard 2.0 doesn't have a built-in `ToHashSet` extension method, only the `ToImmutableHashSet`. But `ImmutableHashSet` is significantly [slower](https://steven-giesel.com/blogPost/34e0fd95-0b3f-40f2-ba2a-36d1d4eb5601) than `HashSet`, both in terms of creation and lookup.

Shane32 (approved):

Shane32:
@gao-artur Is this PR just for string checks like `Field("Name")` or does it also check that expressions like `Field(x => x.Name)` have a public setter or constructor argument?

Shane32:
If you're done with this PR, I'll merge it - lmk.

gao-artur:
> @gao-artur Is this PR just for string checks like `Field("Name")` or does it also check that expressions like `Field(x => x.Name)` have a public setter or constructor argument?

It only check names. But you are right, need to check the property on expression is settable. Will add this later this week.

Shane32:
Ok, want me to merge this now or do you want to add to this PR?

gao-artur:
Yeah, let's merge it. I'll create a new PR. Thanks.

gao-artur:
Thinking again, it's probably should be 2 separate diagnostics: the property doesn't exist and the property doesn't have a public setter. So please wait with the merge. I'll split it in a few days.

gao-artur:
Summary of the PR.
There are two new diagnostics:
**GQL006** checks the input field name has a matching field, property, or public ctor parameter on the source type. I covered the following scenarios:
- `Field<>(name)`
- `Field(name)`
- `Field(name, expression)` (the `Field(expression)` is not checked because the property name defines the input field name)
- The `name` can be defined as a literal or const string (it can be a local const: `FirstName` or const defined in another class `FieldNames.FirstName`). Read-only fields or any kind of get-name methods are not supported.
- The `name` can be defined as a public ctor parameter of the source type
- If the source type is not closed but has generic constraints, the constraints are checked. For example
```c#
public abstract class MyBaseInputObjectGraphType<TSource> : InputObjectGraphType<TSource>
    where TSource : MyBaseSource, ISource
```
I'm checking the `MyBaseSource` and `ISource` fields and properties. If generic has no constraints, all the input fields are reported because we can't verify they can be mapped to any source type field.

**GQL007** checks whether the source type field/property can be set. I covered the following scenarios:
- private, static, const, readonly, private setter
- `Field(name, expression)` and `Field(expression)` - both are verified

The diagnostic is reported on the `name` argument when possible. In the case of `Field(x => x.Prop)` - on the `x.Prop` part of the expression.
I didn't cover the usage of the obsolete `Name` method.

**Possible additional enhancements**
I didn't verify if the defined set of fields is enough to construct an instance of the source type if it doesn't have a default constructor or has `required` properties. For example, the `SourceType` needs an additional `LastName` input field name:
```c#
Field<StringGraphType>("FirstName");

public class SourceType
{
    public SourceType(string firstName, string lastName) { ... }
}
```
This scenario is a bit difficult because the missing input field name can be defined on the base or derived input graph type. But this is possible and can be added later if needed.

I also didn't cover the scenario of ambiguous properties/fields. For example, if the input field name matches two properties with the same name but different casing, or the property on the derived source type hides another property with the same name on the base source type.

Shane32:
Looks great!  I'll leave it for a few days in case anyone has any comments before merging.

gao-artur:
Btw, it could be great if you can test the analyzer preview on your code base. It can help to find use cases I didn't think about that may cause the analyzer to throw exceptions (in this case VS shows an error message and disables the analyzer).

Shane32:
Easy enough after it's merged.  If there are bugs, I'll post an issue.

Shane32:
`GraphQL.7.4.1-preview-857` contains this PR, available from GitHub Packages

Shane32:
Tested against two of my larger GraphQL projects.  One had no warnings, but I tested the analyzer to be sure it was working correctly, and it was.  The other had a large number of suggestions to rewrite `Field`.  I 'fixed all' and reviewed the changed code.  It compiled correctly immediately.  One thing I notice is that when rewriting the QueryArguments, it consistently indented the code improperly, even while retaining the general format of the code.  Not sure if anything can be done about that, and it's certainly not a big deal (`dotnet format` might fix it anyway).  But overall, very impressive!!!! 🥇 💯 🥇 💯 🥇 

Here's some snippets of the indenting FYI:

Sample 1: (notice here that Resolve is indented also)
image

Sample 2:
image

Sample 3:
image

Sample 4:
image

Shane32:
Here's an idea for another analyzer: rewrite `builder.Argument` for obsolete overloads:

image

Shane32:
Having these analyzers makes it easier to remove obsolete methods in new major versions. :+1: I had planned to leave some of these common methods available throughout v8, but with the analyzer users can fix all their code at once when upgrading.

Shane32:
@gao-artur I've tested the new version.  Definitely better, but still a couple weird formatting situations.  If you want me to, I can provide more samples, etc.  On the other hand, I'm not sure how easy/hard/useful it is to get this perfect.  It does compile fine, which is the most important part.

gao-artur:
@Shane32, please post all the issues you find. I'll check how much work it is to fix them. Sorry I don't have much free time now, so if these issues are minor cosmetics, I'd prefer to invest the time in new analyzers.

Shane32:
> if these issues are minor cosmetics, I'd prefer to invest the time in new analyzers.

Fully agree.

> please post all the issues you find

Anytime the `);` was on its own line:

image

Then I had a couple files which consistently formatted a bit weird throughout the whole file -- but I don't understand why these would be any different than any other files.  For instance, compare to the snip above.

image

But generally, very good results considering the number of changes.  For reference, the PR (with only its changes) had about 200 lines of code changed.  19 lines "needed" to be tweaked because of the above issues, half being the `);` issue, and the other half only being across two files.  So certainly proved to be extremely useful!

PR #3729, opened 2023-10-11
Fix leading trivia when rewriting field builder

gao-artur:


gao-artur:
Strange. The `CarriageReturnLineFeed` used to work in both OS before this change.

Shane32:
Maybe `Environment.NewLine` ?

Shane32:
Another thing to consider is that Git may decode line feeds differently on Linux vs Windows.  So if there's a text literal within the code like this: `"Test\nTest2"`, that will be the same on both operating systems, but when you write `@"Test` with `Test2"` on the next line, then the literal itself might change based on the OS it was compiled on.  Just something to keep in mind when writing tests.

src/GraphQL.Analyzers.CodeFixes/FieldBuilderCodeFixProvider.cs
```diff
@@ -77,6 +77,17 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+        const int oneLevelIndentation = 4;
+        int fieldInvocationIndentation = newFieldInvocationExpression
+            .GetLeadingTrivia()
+            .Where(trivia => trivia.IsKind(SyntaxKind.WhitespaceTrivia))
+            .Sum(trivia => trivia.FullSpan.Length);
+
+        var whitespaceTrivia = Whitespace(new string(' ', fieldInvocationIndentation + oneLevelIndentation));
+        var newLineTrivia = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+            ? CarriageReturnLineFeed
+            : LineFeed;
```
Shane32, line 89: Is this not equivalent to `Environment.NewLine` ?

https://learn.microsoft.com/en-us/dotnet/api/system.environment.newline?view=net-7.0

> `\r\n` for non-Unix platforms, or `\n` for Unix platforms.

gao-artur: Will try this
```c#
EndOfLine(Environment.NewLine)
```

src/GraphQL.Analyzers.CodeFixes/FieldBuilderCodeFixProvider.cs
```diff
@@ -236,18 +246,16 @@ static SyntaxToken GetIdentifier(SimpleNameSyntax simpleNameSyntax, string name)
...
                             Argument(argumentExpression))));
 
         return exp;
     }
-
-    private static SyntaxNode Format(SyntaxNode syntaxNode, Workspace workspace)
-    {
-        syntaxNode = syntaxNode.WithAdditionalAnnotations(Formatter.Annotation);
-        return Formatter.Format(syntaxNode, Formatter.Annotation, workspace);
-    }
```
gao-artur, line 252: This method was normalizing the line endings. But it also makes wrong indentations.

Shane32: Do we have a new test added for the scenario fixed by this PR?

gao-artur: I have adjusted an old test. This is the result before the fix
image

Shane32: oh i see.  :+1:

Shane32 (approved):

Shane32:
All set to merge @gao-artur  ?

gao-artur:
Yep

Shane32:
Available in `GraphQL.7.6.2-preview-859`

PR #3740, opened 2023-10-17
Fix formatting when the line before Field invocation contains whitespaces

gao-artur:
This PR fixes the second issue reported [here](https://github.com/graphql-dotnet/graphql-dotnet/pull/3720#issuecomment-1764981528)
image

The fix for `);` on a separate line is more complicated, so I'll leave it as is.

### DIFF
--- a/src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
+++ b/src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
@@ -455,6 +455,55 @@ public class FieldBuilderAnalyzerTests
     }
 
     [Fact]
+    public async Task EmptyLineWithWhiteSpacesBeforeField_ArgumentsFormattingPreservedAsync()
+    {
+        const string source = """
+            using GraphQL.Types;
+
+            namespace Sample.Server;
+
+            public class MyGraphType : ObjectGraphType
+            {
+                public MyGraphType()
+                {
+                    // next line contains whitespaces
+                    
+                    Field<StringGraphType>("name", "description", null,
+                        resolve: context => Resolve(
+                            "text"
+                        ));
+                }
+
+                public string Resolve(string s1) => s1;
+            }
+            """;
+
+        const string fix = """
+            using GraphQL.Types;
+
+            namespace Sample.Server;
+
+            public class MyGraphType : ObjectGraphType
+            {
+                public MyGraphType()
+                {
+                    // next line contains whitespaces
+                    
+                    Field<StringGraphType>("name").Description("description")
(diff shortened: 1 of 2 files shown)


gao-artur:
There is a nice VS extension to visualize trailing whitespaces in the code. It also allows to automatically remove them on document save. Unfortunately, it also removes whitespaces from the raw strings, so I had to disable the auto-remove feature to write this test.
https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer

src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
```diff
@@ -454,6 +454,55 @@ public MyGraphType()
...
+                        .Resolve(context => Resolve(
+                            "text"
+                        ));
+                }
+
+                public string Resolve(string s1) => s1;
+            }
+            """;
+
+        var expected = VerifyCS.Diagnostic(FieldBuilderAnalyzer.DoNotUseObsoleteFieldMethods).WithSpan(11, 9, 14, 15);
```
Shane32, line 502:
```suggestion
        // ensure whitespace was not removed from input sample
        source.Replace("\r", "").ShouldContain("// next line contains whitespaces\n    ");

        var expected = VerifyCS.Diagnostic(FieldBuilderAnalyzer.DoNotUseObsoleteFieldMethods).WithSpan(11, 9, 14, 15);
```

Shane32:
I looked back at my code, and although there isn't whitespace on the end of the previous line, there is a comment there, which is likely equivalent to whitespace:

image

In addition to the test you just added, perhaps we also add a similar test with a comment on the prior line instead of only whitespace.

src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
Shane32: I'm thinking that it is likely enough that some other formatting tool in the future may remove this whitespace, and the test should still pass.  So with a couple extra lines of code we can ensure the test is correct.

Shane32:
> The fix for `);` on a separate line is more complicated, so I'll leave it as is.

:+1:

src/GraphQL.Analyzers.Tests/FieldBuilderAnalyzerTests.cs
gao-artur: Makes sense.

Shane32 (approved):

Shane32:
Tested on my codebase; works great!

PR #3741, opened 2023-10-17
Change Analyzers Verifiers and upgrade xUnit to 2.5.2

gao-artur:
Based on [this suggestion](https://github.com/dotnet/roslyn-sdk/issues/1099#issuecomment-1723487931), I replaced `Testing.XUnit` dependencies with generic `Testing` and copied missing classes to keep the same exception types as previously.
Then, I upgraded xUnit to the latest 2.5.2 and fixed all the breaking changes in the copied types.

Additionally, the new xUnit version brings with it many analyzers.

Added `LongLivedMarshalByRefObject` as a base class to `MessageSinkWrapper`
image

Removed all the `.ConfigureAwait(false)` in tests projects and added `RCS1090` to `NoWarn` section in `Tests.props`
image

The open question is `xUnit1031`. Any reason why these tests can't be made async to remove the blocking calls?
image


Shane32:
We added `ConfigureAwait(false)` in tests due to dataloader tests deadlocking (as some tests assumed multithreaded behavior).  So we made it a requirement for all tests.

We can proceed if xUnit recommends against that, but we will need to keep an eye on the data loader tests to be sure they are stable.  If they start failing again, we may need to add `ConfigureAwait(false)` back into the data loader tests.

Shane32:
> The open question is `xUnit1031`. Any reason why these tests can't be made async to remove the blocking calls?

Nope

gao-artur:
> If they start failing again, we may need to add ConfigureAwait(false) back into the data loader tests.

Got it! Do you want to revert this change already, or do you prefer to wait and see if the issue still exists?

> > The open question is `xUnit1031`. Any reason why these tests can't be made async to remove the blocking calls?
> 
> Nope

Done. There are a few places that couldn't be converted. Look for `#pragma warning disable xUnit1031`

Shane32:
> Got it! Do you want to revert this change already, or do you prefer to wait and see if the issue still exists?

Well, it was intermittent, so we probably won't know until a while after it's merged.  Either we just merge it as-is, or revert the changes to only the data-loader tests project, having a global warning ignore for that project only.  Perhaps even make the changes in a separate PR so it is easy to revert if it doesn't work out.

gao-artur:
> Perhaps even make the changes in a separate PR so it is easy to revert if it doesn't work out.

Not sure what you mean. What change should go in a separate PR?

gao-artur:
The `xUnit1030` fix in all projects?

Shane32:
xUnit1030 in the DataLoader project

gao-artur:
Reverted. Do you have an idea how to enforce the `RCS1090` rule in DataLoader project if it's globally suppressed in `Tests.props`?

src/GraphQL.Analyzers.Tests/GraphQL.Analyzers.Tests.csproj
```diff
@@ -9,13 +9,13 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.7.*" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit" Version="1.*" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.XUnit" Version="1.*" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeRefactoring.Testing.XUnit" Version="1.*" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing.XUnit" Version="1.*" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing.XUnit" Version="1.*" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeRefactoring.Testing.XUnit" Version="1.*" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.6.0" />
```
Shane32, line 12: Going back from 4.7 to 4.6?

Shane32 (approved):

src/GraphQL.Analyzers.Tests/GraphQL.Analyzers.Tests.csproj
gao-artur: This is the version I use in the analyzers project. Just aligned them. I had problem with. 4.7 - VS didn't recognize any analyzers in the package.

gao-artur:
Would you like me to create a PR for DataLoader tests?

Shane32:
That would be great!  We can merge it, and then if it ends up that we are again having trouble with data loader tests failing intermittently, hopefully we can fix it "correctly", or if not we can revert the PR.

PR #3747, opened 2023-10-21
Rewrite the obsolete Name method on ConnectionBuilder

gao-artur:
The `FieldNameAnalyzer` was extended to support `ConnectionBuilder.Name` in addition to `FieldBuilder.Name`.
The `ComplexGraphType.Connection` and `ConnectionBuilder<TSourceType>.Create` methods were tested on all the use cases the `ComplexGraphType.Field` was tested. The `ConnectionBuilder.Create` (non-generic builder class) has a single test but it should work the same way the `ConnectionBuilder<TSourceType>.Create` works.

### DIFF
--- a/src/GraphQL.Analyzers.Tests/FieldNameAnalyzerTests.cs
+++ b/src/GraphQL.Analyzers.Tests/FieldNameAnalyzerTests.cs
@@ -19,7 +19,7 @@ public class FieldNameAnalyzerTests
     [Theory]
     [InlineData(Constants.MethodNames.Field)]
     [InlineData(Constants.MethodNames.Connection)]
-    public async Task FieldAndNameMethodsCalled_NotGraphQLBuilder_NoDiagnostics(string builderName)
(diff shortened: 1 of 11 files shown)


src/GraphQL.Analyzers.Tests/FieldNameAnalyzerTests.cs
```diff
@@ -92,12 +122,13 @@ public class MyGraphType : ObjectGraphType
             {
                 public MyGraphType()
                 {
-                    Field<StringGraphType>().Resolve(context => "Test").Name("Text");
+                    {{builderName}}<StringGraphType>().Description("description").Name("Text");
```
gao-artur, line 125: The `ConnectionBuilder.Resolve` returns void in contrast to `FieldBuilder.Resolve`, which returns the builder instance, so I had to replace it with the `Description` method. Is this inconsistency intended?

Shane32: No, but it would be a breaking change to change the signature in v7.  We can fix v8 (the `develop` branch).

Shane32 (approved):

PR #3752, opened 2023-10-27
Fix xml docs

gao-artur:
Before this change, Resharper was showing 116 errors in 44 files. Not really 116 errors. They are just multiplied by the number of supported target frameworks.
image
 
After the fix, there are still 12 errors in 8 files (actually two errors). But I think it's just a bug in Resharper.
image

The [first error](https://github.com/graphql-dotnet/graphql-dotnet/blob/a107165f2f1594ea78feb73ddd2b949b1f6caa67/src/GraphQL/Extensions/ObservableExtensions.cs#L304) is `<typeparamref name="TOut"/>`. This comment is applied to the inner record struct, but the `TOut` is defined by the outer class.
The [second error](https://github.com/graphql-dotnet/graphql-dotnet/blob/a107165f2f1594ea78feb73ddd2b949b1f6caa67/src/GraphQL/Validation/ValidationOptions.cs#L9) is `<see cref="IDocumentValidator.ValidateAsync(in ValidationOptions)`. Resharper doesn't love the `in` modifier.


Shane32:
@gao-artur FYI: As a team member, I'd encourage you to create PR branches directly within the graphql-dotnet repository.  Not a big deal, but it makes it easier for the rest of us to pull your branch when needed.  We have not been particular about the branch names; most are just named whatever the PR is (e.g. `fix_xml_docs`).  We could use prefixes like `issues/` or `features/` or whatever, but so far the team is small enough that it doesn't matter much.

Shane32 (approved):

Shane32:
I don't use ReSharper, but all changes look good 👍 

Feel free to merge after approval (which I have).

gao-artur:
> I'd encourage you to create PR branches directly within the graphql-dotnet repository

Sure, no problem. I started this particular branch before I was invited to the team, so I continued with it. I already started another analyzer implementation on this repo branch and not on my fork.

> I don't use ReSharper

How? I feel handicapped without ReSharper 😄
Btw they have a free license for [OSS maintainers](https://www.jetbrains.com/community/opensource/#support)

Shane32:
Also FYI, when I create PRs, I usually do the following, and I'd encourage you to do so also:
- Assign myself to the PR
- Update the tags, typically removing the ones that are auto-added
- Add milestone for the proper version (in this case 7.7 for the master branch) when it seems appropriate.  I usually only do so for bug fixes or new features.  Code cleanup type stuff including dependabot I usually don't.

Also, any documentation pushed to master will immediately publish with no delay.  Just FYI to keep that in mind.  This means that sometimes we have docs that are ahead of the nuget release, but this hasn't yet been a problem.

Shane32:
> How? I feel handicapped without ReSharper 😄
> Btw they have a free license for [OSS maintainers](https://www.jetbrains.com/community/opensource/#support)

Nice! I'll take a look.  I just have never done so -- and, if I start using a tool, then often I need to buy it for all my devs, so I favor using built-in tools.  Thankfully, VS has much better refactoring tools than it used to.  But if a tool is worth it, it's worth it.

Shane32:
Btw, we always use "squash and merge" and have the PR title as the commit title.  This should be the default, so you shouldn't need to worry about it.

PR #3758, opened 2023-11-05
[WIP] Add input resolvers

### LINKED ISSUE #1176: InputObjectGraphType resolve function ignored

Shane32:
Allows input "resolvers" in input object graph types, to coerce or validate input values.

Possible uses:
- Validate a field value
- Coerce the field value to another type

This feature is not strictly "necessary"; customizing the `ParseDictionary` method for an input object type would result in nearly equivalent behavior.

This would be a non-breaking change for v7, I believe, although it currently targets v8.


src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -91,5 +91,12 @@ internal void SetName(string name, bool validate)
         /// Gets or sets a subscription resolver for the field. Only applicable to the root fields of subscription.
         /// </summary>
         public ISourceStreamResolver? StreamResolver { get; set; }
+
+        /// <summary>
+        /// Validates and/or parses the value received from the client.
+        /// Throw an exception if necessary to indicate a problem.
+        /// Only applicable to fields of input graph types.
+        /// </summary>
+        public Func<object?, object?> ParseValue { get; set; } = static value => value;
```
Shane32, line 100: Should this be an interface with a property, like `IFieldResolver` is?  Not sure exactly what the benefit would be.  At least it would match the `Resolver` and `StreamResolver` properties.

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -132,6 +132,15 @@ protected FieldBuilder(FieldType fieldType)
...
         }
 
+        /// <summary>
+        /// Sets the input resolver for the field.
+        /// </summary>
+        public virtual FieldBuilder<TSourceType, TReturnType> ParseValue(Func<object?, object?> parseValue)
+        {
+            FieldType.ParseValue = parseValue;
+            return this;
+        }
```
Shane32, line 142: What should the signature of this method be?  The following might be better, although then is only useful for validation:

`ParseValue(Func<TReturnType?, TReturnType?> parseValue)`

Or perhaps it should be this:

`ParseValue(Func<object?, TReturnType?> parseValue)`

src/GraphQL.Tests/Types/InputObjectGraphTypeTests.cs
```diff
@@ -20,4 +23,53 @@ public void should_throw_an_exception_if_object_graph_type_contains_Input_object
...
         exception.Message.ShouldContain("Output type 'Object' can have fields only of output types: ScalarGraphType, ObjectGraphType, InterfaceGraphType, UnionGraphType or EnumerationGraphType.");
     }
+
+    [Fact]
+    public async Task input_resolver_works()
+    {
+        // demonstrates having a StringGraphType field that accepts a Uri as input
+        // the string value is coerced to a Uri prior to beginning execution of the request
+        var inputType = new InputObjectGraphType<Class1>();
+        inputType.Field<StringGraphType, Uri>("url")
```
Shane32, line 33: Now here I was unable to write this:

```suggestion
        inputType.Field(x => x.Url)
            .Type<StringGraphType>()
```

since the `Type` method only accepted an implementation and not a type.  Perhaps an overload to the `Type` method should be added.  Notice that `TReturnType` will be `Uri` as that is what matches the type of the property, regardless of what data type is returned from the scalar.

Shane32:
@gao-artur @sungam3r I would appreciate your comments on this PR.

gao-artur:
The referenced issue #1176 discusses the "reverse resolve": `field => property` rather than `property => field` for the output type. I also assumed the resolver on the input type would work this way and found the hard way it was just ignored. So, while the `ParseValue` is a nice enhancement, it doesn't fix the referenced issue.
Since the resolver for input fields should only point to the property rather than execute the code, the resolver signature should accept `Expression`
```c#
FieldBuilder<TSourceType, TReturnType> Resolve(Expression<Func<IResolveFieldContext<TSourceType>, TReturnType?>> resolve)
```
Maybe the `Resolve` name is confusing here? That's why @sungam3r said resolvers don't make sense on input types? Maybe it should be called `Map` or `Project` in the case of input types?

Shane32:
You're right; the original issue in #1176 is mapping; but this has been supported for a long time via `Field("test", x => x.Url)`.  So maybe this is a new feature.  And/or maybe this feature isn't needed and should be dropped.

src/GraphQL.Tests/Types/InputObjectGraphTypeTests.cs
Shane32: I should have written `inputType.Field(x => x.Url, type: typeof(StringGraphType))`

Shane32:
@gao-artur I'd like to include some supported way of defining validation for input fields (at a minimum).  This is my main goal for v8.  It seems to me that the implementation in this PR also functions as input coercion, which can also be quite valuable.

The most prominent use case for myself is when some fields in a schema with a `ID` graph type need to be coerced to an `int` and others to a `Guid` or other type.  While GraphQL.NET will coerce during `GetArgument`, this occurs after execution has started, and not during the validation phase of the request.

With the PR as it sits now, users can define an 'input resolver' (or parser or whatever you want to call it) to coerce the string value from the client during the validation phase, and in so doing raise any coercion errors prior to execution.

Points to be considered with the current implementation:
- Interface or delegate property?
- Should there be separate connection points for coercion vs validation?  If not, this design pattern can be followed:
  - For coercion, coerce first then call prior delegate
  - For validation, call prior delegate first, then perform validation
- Should the field builder have two different methods to add coercion or validation as shown above?  Then access the FieldType directly to 'reset' it if necessary.
- What about for arguments and not just object fields?

I'm going to probably create another PR with the suggested changes along with updates to the auto-registering types to make use of the new feature with no additional changes to user code.

PR #3759, opened 2023-11-05
[WIP] Add 'map' and 'type' to field builder

Shane32:
See #1176 and #3758

### DIFF
--- a/src/GraphQL/Builders/FieldBuilder.cs
+++ b/src/GraphQL/Builders/FieldBuilder.cs
@@ -1,5 +1,4 @@
 using System.Linq.Expressions;
-using System.Reflection;
 using GraphQL.Resolvers;
 using GraphQL.Types;
 using GraphQL.Validation.Complexity;
--- a/src/GraphQL/Builders/FieldBuilder.cs
+++ b/src/GraphQL/Builders/FieldBuilder.cs
@@ -1,4 +1,5 @@
 using System.Linq.Expressions;
+using System.Reflection;
 using GraphQL.Resolvers;
 using GraphQL.Types;
 using GraphQL.Validation.Complexity;
@@ -107,8 +108,37 @@ namespace GraphQL.Builders
         /// <param name="type">The graph type of the field.</param>
         public virtual FieldBuilder<TSourceType, TReturnType> Type(IGraphType type)
         {
+            FieldType.Type = null;
             FieldType.ResolvedType = type;
             return this;
+        }
+
+        /// <summary>
+        /// Sets the graph type of the field.
+        /// </summary>
+        public virtual FieldBuilder<TSourceType, TReturnType> Type<TGraphType>()
+            where TGraphType : IGraphType
+        {
+            FieldType.Type = typeof(TGraphType);
+            FieldType.ResolvedType = null;
+            return this;
+        }
+
+        /// <summary>
+        /// For input object fields, sets the mapping of the field to the source type's member.
+        /// </summary>
+        [Obsolete("Please configure the field mapping by providing an expression as an argument to the 'Field' method.")]
+        public virtual FieldBuilder<TSourceType, T> Map<T>(Expression<Func<TSourceType, T>> expression)
+        {
+            if (expression.Body is MemberExpression expr)
+            {
+                FieldType.Metadata[ComplexGraphType<T>.ORIGINAL_EXPRESSION_PROPERTY_NAME] = expr.Member.Name;
+            }
+            else
+            {
+                throw new ArgumentException("Expression must directly reference a member on the source type.", nameof(expression));
+            }
+            return Returns<T>();
         }
 
         /// <summary>
--- a/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
@@ -966,6 +966,9 @@ namespace GraphQL.Builders
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Directive(string name, System.Action<GraphQL.Types.AppliedDirective> configure) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Directive(string name, string argumentName, object? argumentValue) { }
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Directive(string name, string argument1Name, object? argument1Value, string argument2Name, object? argument2Value) { }
+        [System.Obsolete("Please configure the field mapping by providing an expression as an argument to t" +
+            "he \'Field\' method.")]
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, T> Map<T>(System.Linq.Expressions.Expression<System.Func<TSourceType, T>> expression) { }
         [System.Obsolete("Please configure the field name by providing the name as an argument to the \'Fiel" +
             "d\' method.")]
         public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> Name(string name) { }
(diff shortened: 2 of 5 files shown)


src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -107,10 +108,39 @@ protected FieldBuilder(FieldType fieldType)
...
+            FieldType.Type = typeof(TGraphType);
+            FieldType.ResolvedType = null;
+            return this;
+        }
+
+        /// <summary>
+        /// For input object fields, sets the mapping of the field to the source type's member.
+        /// </summary>
+        [Obsolete("Please configure the field mapping by providing an expression as an argument to the 'Field' method.")]
+        public virtual FieldBuilder<TSourceType, T> Map<T>(Expression<Func<TSourceType, T>> expression)
```
Shane32, line 131: This method should work perfectly with all existing code, since ToObject is already aware of the `ORIGINAL_EXPRESSION_PROPERTY_NAME` metadata.  But I don't think we should add this method; rather people should use the `Field("testing", x => x.Sample)` when defining their field.  Thoughts?

gao-artur: > rather people should use the Field("testing", x => x.Sample)

This syntax is less discoverable. I wasn't aware of it before you showed it to me when working on analyzers.

gao-artur: Also, I don't know how others define their schema, but we always start with `Field<>(name)...`. With the syntax above, the input field definition is not consistent with other object fields.

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -107,10 +107,39 @@ protected FieldBuilder(FieldType fieldType)
...
+        {
+            if (expression.Body is MemberExpression expr)
+            {
+                FieldType.Metadata[ComplexGraphType<T>.ORIGINAL_EXPRESSION_PROPERTY_NAME] = expr.Member.Name;
```
gao-artur, line 134: Ensure it has a public setter?

src/GraphQL/Builders/FieldBuilder.cs
gao-artur: The `ResolverAnalyzer` (GQL005) will remove the `Resolve` from the input fields. I think it is worth adding a fix to convert it to `Map` instead. So whoever has not upgraded to [7.4](https://github.com/graphql-dotnet/graphql-dotnet/pull/3574) yet will get the desired behavior.

Shane32: The syntax below has been the primary way of adding fields to both input and output objects since 2.x.  I really don't think we should change it.

```cs
Field(x => x.Name); // add a field for the Name property, non-nullable, with an auto-mapped graph type
Field(x => x.Name, true); // similar but nullable
Field("FullName", x => x.Name); // similar but with "fullName" as the name, non-nullable
Field("FullName", x => x.Name, true); // similar but nullable
```

If this PR was merged with no changes, the equivalent code to the first line would be this:

```cs
Field<NonNullGraphType<GraphQLClrOutputTypeReference<string>>, string>("Name").Map(x => x.Name);
```

It is much desirable to allow users to easily create graph types such as this:

```cs
public class CustomerType : ObjectGraphType<Customer>
{
    public CustomerType()
    {
        Field(x => x.FirstName);
        Field(x => x.LastName);
        Field("Name", x => x.FullName);
        Field(x => x.Company, true);
    }
}
```

The code would look almost identical for an input type.

gao-artur: We never used this overload. All our fields defined in a more verbose but explicit way
```c#
public class CustomerType : ObjectGraphType<Customer>
{
    public CustomerType()
    {
        Field<NonNullGraphType<StringGraphType>>("FirstName).Resolve(ctx => ctx.Source.FirstName);
        Field<NonNullGraphType<StringGraphType>>("LastName).Resolve(ctx => ctx.Source.LastName);
        Field<NonNullGraphType<StringGraphType>>("Name).Resolve(ctx => ctx.Source.FullName);
        Field<StringGraphType>("Company).Resolve(ctx => ctx.Source.Company);
    }
}
```

We do it mainly to decouple graph field names from the source property names. But I agree that the same can be achieved with a more compact form.
```c#
public class CustomerType : ObjectGraphType<Customer>
{
    public CustomerType()
    {
        Field("FirstName", x => x.FirstName);
        Field("LastName", x => x.LastName);
        Field("Name", x => x.FullName);
        Field("Company", x => x.Company, true);
    }
}
```

Maybe we should adopt this.

Shane32:
I don't think this should be merged.  Rather the existing `Type(IGraphType)` method should be deprecated.

gao-artur:
I don't like that for output types, the verbose syntax gives you the most flexibility (and performance), while for input types - the expression syntax.
I wanted to create an analyzer + code fix to set preferred syntax for input and output types so users can keep their schema definition consistent.
Something like this (config names just for example):
```ini
# require explicit name to make sure the schema doesn't change when you rename the property
graphql_short_syntax_require_explicit_names = true

# the default preferred syntax
# for input types
#    short: always use expression syntax
#    verbose: always use verbose syntax. When graphql_short_syntax_require_explicit_names = true, use Map() method to set the name
# for output types
#    short: use expression syntax when the resolver is a simple property accessor
#    verbose: always use verbose syntax
graphql_prefered_syntax = none|short|verbose

# configuration per object type
graphql_input_fields_prefered_syntax = default|short|verbose
graphql_output_fields_prefered_syntax = default|short|verbose
graphql_interface_fields_prefered_syntax = default|short|verbose
```

When using short syntax for output fields, the performance can be improved with code generation to avoid dynamic compilation: https://github.com/graphql-dotnet/graphql-dotnet/issues/3801

Shane32:
I think there’s one major difference between input and output types, that being that for input types you have to define a resolver, which is often a multi-line code block, and for input types you define a mapping, which is typically a very short lambda. And if you only need a mapping for a field of an output type then it can be written in the exact same format as the input field. Once compiled the lambda will run virtually the same as defining a resolver. 

I personally find that the easiest code to write and maintain is no code, so I always attempt to write my graph types to use implicit graph/clr types. Doing so helps ensure that any clr type changes carry through to the graph type and show up as changes to the schema in my schema verification test. 

This I value above code consistency, so in my code you’ll find a mixture of different output type definition patterns as follows:
- For any data model where mappings can be inferred, I mark the clr type with `[Map]` and let the auto registering graph type classes handle it.
- For any database models I typically define a graph type class to ensure the proper fields are mapped and in the right format. The database models come from another assembly anyway. But the fields are defined as simply as possible; a single line like input fields if possible. 
- For mutation graphs, I almost always use type-first code via the auto registering graph type classes, as it essentially “eliminates” the graph type definition - the field definitions are derived from the method signatures. 

Of course that’s just how I write code and you or other users may have different preferences.

Shane32:
One reason we may wish to ‘push’ one form of writing graph types over another may be due to AOT compilation. If we truly want to provide AOT compatible infrastructure then we need to consider what method signatures will trigger what warnings and where they can be safely suppressed. 

For instance, a field definition that includes the exact graph type and a resolver can perhaps be safely suppressed. If this form is in a separate overload then it’s easy to mark the overload as safe to use vs others which trigger a warning. 

Or we can write a bunch of analyzers by hand to determine this, I suppose.

Shane32:
> I wanted to create an analyzer + code fix to set preferred syntax for input and output types so users can keep their schema definition consistent. Something like this (config names just for example):

Makes sense to me.  It's a good idea.

gao-artur:
> Makes sense to me. It's a good idea.

The explicit field names should be possible with verbose input fields, too (hence, the name `graphql_short_syntax_require_explicit_names` is not correct; it should be `graphql_filed_require_explicit_names`). Otherwise, we can break the field mapping if we rename the property. More than that, the different filed and property names can be desired, but they are only possible with expression syntax. I think both syntaxes should be equivalent in the features they provide.

gao-artur:
> we need to consider what method signatures will trigger what warnings

They probably can be solved by https://github.com/graphql-dotnet/graphql-dotnet/issues/3801 or another kind of code generation.

Issue #3757, opened 2023-11-04
Rewrite `ObjectExtensions.ToObject with` source generator


gao-artur:
Just FYI, here's a few comments that may be helpful:

When `InputObjectGraphType<T>` is in use, `ObjectExtensions.ToObject` is called from `InputObjectGraphType<T>.ParseDictionary` (unless overridden).  When `GetArgument` is later called, the object isn't an `IDictionary` and so it does not call `ObjectExtensions.ToObject` again.  The advantage here is that deserialization errors are caught during variable parsing.  This scenario is probably the case for most code-first and type-first code.

When `InputObjectGraphType` (untyped) is in use, the dictionary itself is the stored for later use by `GetArgument`.  The field resolver's call to `GetArgument` triggers `ObjectExtensions.ToObject` to convert it to the specified type.  This is scenario is probably the case for schema-first code. (not sure though)

In the past, I had been thinking about replacing the `ToObject` code with dynamically-compiled code.  This doesn't help for AOT compilation scenarios, however.  I think I may have even tried it, but the current design of `ToObject` is such that it may call a different constructor if an optional field was not provided.  (Going from memory here, so I could be wrong.)  In the end, the existing code worked fine, so I spent my dev time elsewhere.

`AutoRegisteringInputObjectGraphType` only adds fields with settable properties.  Some people have requested that it add all properties, assuming that get-only properties can be set through the constructor.  I would say that it should inspect the constructors and include get-only properties as applicable, but I have not spent any time on it.  Either way, the design of `AutoRegisteringInputObjectGraphType` (and probably `InputObjectGraphType<T>`) is such that an applicable constructor could be selected and code generated (dynamically compiled or source generated) that always behaves in a predefined way for that type.  It may represent a change in behavior, in that all fields are always set on the deserialized object; optional fields would be set to null if not provided and no default field value exists.  I would favor this, as it should simplify the code that executes during `ParseDictionary`.  But it should be made for v8 if it changes existing behavior (even if only in rare circumstances.)

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/issues/3732#issuecomment-1793467429_

gao-artur:
The current idea (draft and oversimplified). For this source code

```c#
public class MyInput
{
    public MyInput(string field1)
    {
        Field1 = field1;
    }

    public string Field1 { get; }
    public string? Field2 { get; init; }
[...]

PR #3761, opened 2023-11-10
ToObject compilation

Shane32:
This is an experiment to make ToObject compile a delegate at runtime to be used instead of reflection.  It's not finished and still doesn't quite work right in some scenarios.

Update: it now has only one functional difference from the reflection-based ToObject: only specific list types are supported:
1. Arrays
2. interfaces supported by `List<T>`
3. `List<T>`

The reflection-based code also supports any list class which implements `IList`, such as any class derived from `List<T>`.

This PR also introduces a global switch to enable/disable ToObject compilation, which defaults to `true` but is automatically set to `false` if `.AddSchema` is called specifying a scoped registration.


Shane32:
Progress so far:

### No compilation

| Method             | UseCompilation | Mean        | Error    | StdDev   | Gen0   | Allocated |
|------------------- |--------------- |------------:|---------:|---------:|-------:|----------:|
| Person_Populated   | False          |   481.49 ns | 2.563 ns | 2.272 ns | 0.0343 |     216 B |
| Person_Undefined   | False          |   256.31 ns | 2.404 ns | 2.249 ns | 0.0305 |     192 B |
| Company_Populated  | False          | 1,652.34 ns | 4.658 ns | 4.357 ns | 0.1106 |     704 B |
| Company_Undefined  | False          |   848.33 ns | 3.847 ns | 3.410 ns | 0.0706 |     448 B |
| Employee_Populated | False          | 2,031.73 ns | 8.089 ns | 7.171 ns | 0.1183 |     760 B |
| Employee_Undefined | False          | 1,148.54 ns | 6.729 ns | 6.294 ns | 0.1106 |     704 B |
| Salary_Populated   | False          |   845.79 ns | 2.352 ns | 2.085 ns | 0.0610 |     384 B |
| Salary_Undefined   | False          |   538.91 ns | 3.980 ns | 3.528 ns | 0.0610 |     384 B |

### With compilation

| Method             | UseCompilation | Mean        | Error    | StdDev   | Gen0   | Allocated |
|------------------- |--------------- |------------:|---------:|---------:|-------:|----------:|
| Person_Populated   | True           |    79.28 ns | 0.301 ns | 0.281 ns | 0.0050 |      32 B |
| Person_Undefined   | True           |    17.31 ns | 0.187 ns | 0.166 ns | 0.0051 |      32 B |
| Company_Populated  | True           |   218.43 ns | 1.247 ns | 1.106 ns | 0.0408 |     256 B |
| Company_Undefined  | True           |    43.20 ns | 0.883 ns | 0.826 ns | 0.0102 |      64 B |
| Employee_Populated | True           |   277.27 ns | 3.072 ns | 2.873 ns | 0.0253 |     160 B |
| Employee_Undefined | True           |    64.77 ns | 0.331 ns | 0.293 ns | 0.0153 |      96 B |
| Salary_Populated   | True           |   116.88 ns | 0.517 ns | 0.483 ns | 0.0088 |      56 B |
| Salary_Undefined   | True           |    32.95 ns | 0.146 ns | 0.137 ns | 0.0089 |      56 B |

Todo:

- [ ] Don't use `IEnumerable` interface to loop through lists when indexing is available.
- [ ] Call `List<T>` constructor with count when available
- [ ] When possible, write directly to an array rather than a list

Shane32:
Further progress when list types are not specifically `List<T>` (e.g. `IList<T>` or `IEnumerable<T>` or `T[]`):



| Method             | UseCompilation | Mean        | Error     | StdDev    | Gen0   | Allocated |
|------------------- |--------------- |------------:|----------:|----------:|-------:|----------:|
| Person_Populated   | True           |    77.22 ns |  0.329 ns |  0.275 ns | 0.0050 |      32 B |
| Person_Undefined   | True           |    15.98 ns |  0.121 ns |  0.101 ns | 0.0051 |      32 B |
| Company_Populated  | True           |   208.04 ns |  1.155 ns |  1.081 ns | 0.0408 |     256 B |
| Company_Undefined  | True           |    43.21 ns |  0.278 ns |  0.260 ns | 0.0102 |      64 B |
| Employee_Populated | True           |   244.52 ns |  3.307 ns |  2.931 ns | 0.0191 |     120 B |
| Employee_Undefined | True           |    61.84 ns |  0.849 ns |  0.794 ns | 0.0153 |      96 B |
| Salary_Populated   | True           |   105.00 ns |  0.517 ns |  0.458 ns | 0.0088 |      56 B |
| Salary_Undefined   | True           |    32.83 ns |  0.298 ns |  0.249 ns | 0.0089 |      56 B |

Shane32:
This PR now represents 10x speedup and 5x less memory.

gao-artur:
Amazing improvement!

Shane32:
I've made the appropriate changes so that `ToObject` gets compiled during schema initialization for singleton schemas.  This should catch deserialization problems before runtime and is an improvement over prior versions of GraphQL.NET.

src/GraphQL/Extensions/ObjectExtensions.cs
```diff
@@ -184,20 +223,15 @@ public struct MemberFieldInfo
                 if (index == -1)
                 {
                     if (ctorParam.IsOptional)
-                        ctorFields[i] = new() { ParameterInfo = ctorParam };
+                        ctorFields[i] = new(null, ctorParam, null);
```
gao-artur, line 226: Usually, I add argument names when passing constants. It's hard to understand who these `null` arguments are without checking the method/constructor signature. And absolutely impossible when reading the code in web browser. Compare with this:
```c#
new(key: null, ctorParam, graphType: null)
```
IMO much cleaner and easier to read.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -50,12 +52,46 @@ internal InputObjectGraphType(InputObjectGraphType<TSourceType>? cloneFrom)
...
+                // check the value converter for a conversion from dictionary to this object type
+                var conv = ValueConverter.GetConversion(typeof(IDictionary<string, object?>), typeof(TSourceType));
+                if (conv != null)
+                {
+                    _parseDictionary = conv;
+                }
+                else if (GlobalSwitches.DynamicallyCompileToObject)
+                {
+                    // validate and compile the conversion from dictionary to object
+                    _parseDictionary = ObjectExtensions.CompileToObject(typeof(TSourceType), this);
```
gao-artur, line 79: So the user now must override the `Initialize` method when overriding `ParseDictionary` to avoid the `CompileToObject` penalty? Fills smelly...

src/GraphQL/Extensions/ObjectExtensions.Compiled.cs
```diff
@@ -0,0 +1,456 @@
...
+    /// <summary>
+    /// Compiles a function to convert a dictionary to an object based on a specified <see cref="ReflectionInfo"/> instance.
+    /// </summary>
+    private static Func<IDictionary<string, object?>, object> CompileToObject(ReflectionInfo info)
+    {
+        var bestConstructor = info.Constructor;
+        var ctorFields = info.CtorFields;
+        var members = info.MemberFields;
+
+        // T obj;
```
gao-artur, line 37: This was the last line I understood in this file. This looks Chinees for me 😅

gao-artur:
> but is automatically set to `false` if `.AddSchema` is called specifying a scoped registration.

We must find a solution for this. I don't think we need to fallback to reflection when scoped schema is used.

Shane32:
> > but is automatically set to `false` if `.AddSchema` is called specifying a scoped registration.
> 
> We must find a solution for this. I don't think we need to fallback to reflection when scoped schema is used.

Well, we could cache the graph type along with the compiled expression to see if the fields match or something. But I don’t think we should always compile. Some users I know dynamically create a different GraphQL schema for each request and compiling expression trees is an extremely slow process. Also keep in mind that this optimization really isn’t critical. We’ve brought down the time from 2 millionths of a second to 200 billionths for a second. Not really a big deal, even as fun as it is to code this stuff. Finally, GraphQL.NET is specifically optimized for a singleton schema. Scoped schemas are supported but always secondary. If we can find a reasonable solution, sure, but our focus is on singleton schemas.

gao-artur:
> Some users I know dynamically create a different GraphQL schema for each request

In this case, they should explicitly disable dynamic compilation anyway, so they are not a concern.

> Scoped schemas are supported but always secondary. If we can find a reasonable solution, sure, but our focus is on singleton schemas.

Agree.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
gao-artur: Maybe `_parseDictionary` can be changed to `Lazy<>`? However, this will take us a step back in the startup validation story.

Shane32: No I can do that because I split ReflectionInfo from the compilation. So long as it can gather the reflection info then it can lazy compile.  I forgot about that scenario. Note that the user will need to override Initialize if they are overriding ParseDictionary for a type that fails our validation rules.

src/GraphQL/Extensions/ObjectExtensions.cs
Shane32: Good idea.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
Shane32: It just adds a couple more allocations per input type.

Shane32: Take a look now.  So now it examines the type to determine if ParseDictionary has been overridden.  Solves multiple problems at once; now nobody needs to override Initialize when the override ParseDictionary.  Test added.  Still compiles ahead of time when ParseDictionary has not been overridden.

src/GraphQL/Extensions/ObjectExtensions.Compiled.cs
Shane32: I did try to make it as readable as possible.  Each method has a sample of the code that the 'chinese' compiles to. 😉 One thing that might not be immediately apparent is that the dictionary parameter expression is static and so doesn't get passed to the other methods. (`_dictionaryParam`)

Shane32: i also wrote the code (mostly) in one long multi-line statement, as i think it reads more like the code it compiles to, rather than compiling the innermost expression first, storing it in a variable, and so on.  Then it reads a bit backwards from how it executes.

If you like, I can add comments to each line or so to make it more readable.

Shane32: Everything here is currently `private` so we can rename method names or restructure this code as desired in the future.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
gao-artur: Nice trick!

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -50,12 +52,56 @@ internal InputObjectGraphType(InputObjectGraphType<TSourceType>? cloneFrom)
...
             // if (cloneFrom == null) { /* initialization logic */ }
+
+            if (typeof(TSourceType) == typeof(object))
+            {
+                // for InputObjectGraphType just return the dictionary
+                _parseDictionary = static x => x;
+            }
+        }
+
+        private static readonly Type[] _parseDictionaryParameterTypes = [typeof(IDictionary<string, object?>)];
```
gao-artur, line 63: Setting this in a static field won't save any computations because static fields in generic classes are unique for different generic arguments. For example, `InputObjectGraphType<Cat>` and `InputObjectGraphType<Dog>` will have different `_parseDictionaryParameterTypes` instances. And since `Initialize` is called only once, it is better to convert it to a local variable to not hold this array in memory for an entire application lifetime.

Shane32: You’re right; I didn’t think of that.

gao-artur (approved):

Shane32:
@gao-artur I've been adding tests, trying to get the % coverage of the new code as high as possible.  Everything seems to be working smoothly.

I did make one change: with the old reflection-based ToObject code, when trying to coerce an object to a list type, it would just skip if the object could not be cast to `IEnumerable`.  Now it will throw and matches the exception that the precompiled ToObject would throw.  This should be very rare scenario, as GraphQL variable coercion occurs first and makes a well-formed dictionary to be passed to ToObject.

Shane32:
I see you approved it @gao-artur so I'll merge it but of course if you see any other changes that should be made, I'll make a follow-up PR.

src/GraphQL.Tests/ObjectExtensionsTest.cs
```diff
@@ -419,4 +432,208 @@ private class MyInput8
...
+    }
+
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public void toobject_throws_for_constructor_with_unknown_parameters(bool compile)
+    {
+        var inputs = """{ "name": "tom" }""".ToInputs();
+        Should.Throw<InvalidOperationException>(() => inputs.ToObject<MyInput11>(compile))
+            .Message.ShouldBe("Cannot find field named 'age' on graph type 'MyInput11' to fulfill constructor parameter for CLR type 'MyInput11'.");
```
gao-artur, line 498: Didn't you say you [want](https://github.com/graphql-dotnet/graphql-dotnet/issues/3757#issuecomment-1807297711) to coerce this to default value in this case?

Shane32: In this case `age` isn't defined on the graph type.  So, no, if it encounters an unrecognized constructor parameter without a default value, it throws.

PR #3763, opened 2023-11-12
Analyzers/awaitable resolver analyzer

gao-artur:
Detect the awaitable delegates in sync resolve methods and provide a code fix.

```c#
// before
Field<StringGraphType>("Title").Resolve(ctx => GetTitleAsync());

// after
Field<StringGraphType>("Title").ResolveAsync(async ctx => await GetTitleAsync());
```

Note that there are some not supported scenarios. Some can be fixed later, others are edge cases that will just make the code much more complicated.

For example, I can detect and fix cases like this
```c#
Field<StringGraphType>("Test").Resolve(ctx =>
{
    if (condition)
        return GetAsync();
    else
        return GetAnotherAsync();
});
```

But I decided not to support trinary operators for now because there can be multiple nested levels
```c#
Field<StringGraphType>("Test").Resolve(ctx =>
{
    return condition
        ? GetAsync();
        : anotherCondition
            ? GetAnotherAsync()
            : null;
});
```

Also, I detect and warn about awaitable return types like this
```c#
Field<StringGraphType, Task<string>>("Title")
    .Resolve(ctx => Task.FromResult("developer"));
```

But I don't provide a code fix because of the complexity of implementing it and because I believe people don't write such a code.

I also don't currently provide an automatic code fix for `ResolveScoped` because of the complexity. For example:
```C#
//before
Field<StringGraphType>("Title").ResolveScoped(ctx => GetTitleAsync());

// after
Field<StringGraphType, string>("Title").ResolveScopedAsync(async ctx => await GetTitleAsync());
```
Awaiting is not enough. Need also to define the return type. But I hope to fill this gap in the future.

And the most complex scenario:
```c#
// before
Field<StringGraphType>("Title", object).ResolveScoped(ctx => ResolveTitleAsync(ctx));

// after
Field<StringGraphType>("Title", object).ResolveScopedAsync<Person, object>(async ctx =>await ResolveTitleAsync(ctx));
```
Need to define the source and the result types on the `ResolveScopedAsync`. But again, I don't believe people set an `object` or `dynamic` explicitly as the return type.

### DIFF
--- a/docs2/site/docs/analyzers/GQL009_UseAsyncResolver.md
+++ b/docs2/site/docs/analyzers/GQL009_UseAsyncResolver.md
@@ -1,0 +1,139 @@
+# GQL009: Use async resolver
+
+|                        | Value  |
+| ---------------------- | ------ |
+| **Rule ID**            | GQL009 |
+| **Category**           | Usage  |
+| **Default severity**   | Error  |
+| **Enabled by default** | Yes    |
+| **Code fix provided**  | Yes    |
+
+## Cause
+
+This rule triggers when the sync version of the `Resolve` or
+`ResolveScoped` methods is used with awaitable delegate.
+
+## Rule description
+
+`ResolveAsync` and `ResolveScopedAsync` methods should be used to
+register awaitable delegates. The most common awaitable types are
+`System.Threading.Tasks.Task` and `System.Threading.Tasks.ValueTask`,
+but any type that defines a valid `GetAwaiter()` method, which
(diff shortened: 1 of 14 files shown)


Shane32 (approved):

Shane32:
FYI, any improper use of Resolve / ResolveAsync will result in runtime errors with GraphQL.NET v5+.  But this will greatly assist users of older versions of GraphQL.NET during the upgrade process. 🥇

gao-artur:
That's exactly the reason why I created this analyzer 🙂

PR #3769, opened 2023-11-14
ObjectExtensions.ToObject requires mappedType

Shane32:
Prior to making any planned changes to `ToObject`, I think it is necessary to require a reference to the input object graph type.  This will simplify the design and changes to `ToObject` in subsequent PRs.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug956.cs
+++ b/src/GraphQL.Tests/Bugs/Bug956.cs
@@ -25,9 +25,6 @@ public sealed class Bug956QueryType : ObjectGraphType
                 string asString = ctx.GetArgument<string>("base64");
                 asString.ShouldBe("R3JhcGhRTCE="); // GraphQL!
 
-                var asList = ctx.GetArgument<List<byte>>("base64");
-                asList.Count.ShouldBe(12);
-
                 byte[] asBinary = ctx.GetArgument<byte[]>("base64");
                 Encoding.UTF8.GetString(asBinary).ShouldBe("GraphQL!");
 
--- a/src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
@@ -184,10 +184,10 @@ public class AutoRegisteringInputObjectGraphTypeTests
     [InlineData(nameof(FieldTests.NotNullEnumerableNotNullIntField), typeof(NonNullGraphType<ListGraphType<NonNullGraphType<GraphQLClrInputTypeReference<int>>>>))]
     [InlineData(nameof(FieldTests.NullableEnumerableNullableIntField), typeof(ListGraphType<GraphQLClrInputTypeReference<int>>))]
     [InlineData(nameof(FieldTests.NullableEnumerableNotNullIntField), typeof(ListGraphType<NonNullGraphType<GraphQLClrInputTypeReference<int>>>))]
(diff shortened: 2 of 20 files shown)


src/GraphQL.Tests/Bugs/Bug956.cs
```diff
@@ -25,9 +25,6 @@ public Bug956QueryType()
                 string asString = ctx.GetArgument<string>("base64");
                 asString.ShouldBe("R3JhcGhRTCE="); // GraphQL!
 
-                var asList = ctx.GetArgument<List<byte>>("base64");
-                asList.Count.ShouldBe(12);
```
Shane32, line 29: This test made no sense.  Converting to `List<byte>` produced 12 bytes (conversion of the `string` to `char[]` then to `List<int>`), but converting to `byte[]` produced 8 bytes (decoding the `string` from a base64 string).

As such, there is no practical change here; converting a string to a `byte[]` is assumed to be a base64-decoding operation and still functions.

src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
```diff
@@ -357,10 +386,10 @@ private class FieldTests
...
         public IEnumerable<int?>? NullableEnumerableNullableIntField { get; set; }
         public IEnumerable<int>? NullableEnumerableNotNullIntField { get; set; }
-        public Tuple<int, string>?[] NotNullArrayNullableTupleField { get; set; } = null!;
-        public Tuple<int, string>[] NotNullArrayNotNullTupleField { get; set; } = null!;
-        public Tuple<int, string>?[]? NullableArrayNullableTupleField { get; set; }
-        public Tuple<int, string>[]? NullableArrayNotNullTupleField { get; set; }
+        public InputTuple<int, string>?[] NotNullArrayNullableTupleField { get; set; } = null!;
+        public InputTuple<int, string>[] NotNullArrayNotNullTupleField { get; set; } = null!;
+        public InputTuple<int, string>?[]? NullableArrayNullableTupleField { get; set; }
+        public InputTuple<int, string>[]? NullableArrayNotNullTupleField { get; set; }
```
Shane32, line 392: Note: `Tuple<T1, T2>` is still supported here same as before, but is not supported by the test framework because the auto-registering input type won't generate members for constructor arguments, and there is no input graph type defined for `Tuple<int, string>`.  So to make it easy I just created an `InputTuple<T1, T2>` which has settable properties.  The other option would have been to create an input object graph type definition for `Tuple<T1, T2>`

src/GraphQL/Extensions/ObjectExtensions.cs
```diff
@@ -166,7 +171,8 @@ string GetPropertyName(string key, out FieldType? field)
 
                 if (propertyInfo != null && propertyInfo.CanWrite)
                 {
-                    object? value = GetPropertyValue(item.Value, propertyInfo.PropertyType, field?.ResolvedType);
+                    object? value = GetPropertyValue(item.Value, propertyInfo.PropertyType, field.ResolvedType
+                        ?? throw new InvalidOperationException($"Could not get ResolvedType for field '{field.Name}' of type '{inputGraphType}'."));
```
gao-artur, line 175: You use two different messages to say the same thing here and in line 133:
```
"Could not find resolved type for field '{fieldName}' of type '{inputGraphType}'."
"Could not get ResolvedType for field '{field.Name}' of type '{inputGraphType}'."
```

gao-artur (approved):

src/GraphQL/Extensions/ObjectExtensions.cs
Shane32: Fixed

PR #3773, opened 2023-11-15
Add GraphQLConstructor attribute for ToObject

Shane32:
When deserializing objects via `ToObject` (and indirectly by `ParseDictionary`), the constructor is chosen from the available public constructors as follows:

1. the only constructor
2. a constructor marked with `[GraphQLConstructor]`
3. the parameterless constructor
4. otherwise, an exception is thrown


Shane32:
@gao-artur Assuming you approve of this PR, a good candidate for an analyzer for v8 would be as follows: look for classes that inherit from `InputObjectGraphType<T>` where the following is true:
1. The class is not a generic class
2. `ParseDictionary` is not overridden
3. `T` has multiple public constructors
4. `T` has no parameterless public constructor
5. `T` has no public constructor marked with `[GraphQLConstructor]`

I would simply flag as an error with no automatic code fix; let the developer mark the proper constructor or otherwise resolve the issue.  This would certainly be valuable when upgrading from v7 to v8.

docs2/site/docs/migrations/migration8.md
```diff
@@ -99,3 +99,6 @@ This change was made to prevent duplicate registrations of the same service with
 - `ObjectExtensions.ToObject` requires input object graph type for conversion.
 - Only public constructors are eligible candidates while selecting a constructor.
 - The public default constructor is used when available.
```
Shane32, line 101:
```suggestion
- The public parameterless constructor is used when available.
```

docs2/site/docs/migrations/migration8.md
```diff
@@ -98,4 +98,7 @@ This change was made to prevent duplicate registrations of the same service with
 - `ObjectExtensions.ToObject<T>` was removed; it was only used by internal tests.
 - `ObjectExtensions.ToObject` requires input object graph type for conversion.
 - Only public constructors are eligible candidates while selecting a constructor.
-- The public default constructor is used when available.
+- The public parameterless constructor is used when available.
+- If only a single public constructor is available, it is used.
+- If a public constructor is marked with `[GraphQLConstructor]`, it is used.
+- Otherwise an exception is thrown during deserialization.
```
Shane32, line 104:
```suggestion
- If only a single public constructor is available, it is used.
- If a public constructor is marked with `[GraphQLConstructor]`, it is used.
- Otherwise the public parameterless constructor is used if available.
- Otherwise an exception is thrown during deserialization.
```

gao-artur (approved):

PR #3778, opened 2023-11-17
Analyzers code and docs reorganization

gao-artur:
This PR just makes some code reorganization, adds documentation to extension methods, and reformats markdown files to break line after ~80 column. No changes in logic.

I used [Prettier](https://prettier.io/) extension for VS Code to automatically break lines. But it modified [packages files](https://github.com/graphql-dotnet/graphql-dotnet/commit/087b777aa0de1eb02ed92a77a0029ad248967848). I have no idea what are these files, so let me know if you want to revert this change. Or if you want to move the `.prettierrc.json` file to the root docs folder.


src/GraphQL.sln.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/Abbreviations/=QL/@EntryIndexedValue">QL</s:String></wpf:ResourceDictionary>
```
Shane32, line 2: What's this for?  User preferences or project settings?

gao-artur: This is ReSharper config file to stop complaining about QL in capital letters. The config says it's abbreviation.

Shane32 (approved):

src/GraphQL.sln.DotSettings
gao-artur: I'll revert this file. And also packages. It looks like CI doesn't like these changes.

Shane32:
Tests fail so have to check that but looks good in general

PR #3779, opened 2023-11-19
Verify input source type constructors

### LINKED ISSUE #3774: Analyzer should verify input object constructors
> @gao-artur Assuming you approve of this PR, a good candidate for an analyzer for v8 would be as follows: look for classes that inherit from `InputObjectGraphType<T>` where the following is true:
> 1. The class is not a generic class
> 2. `ParseDictionary` is not overridden
> 3. `T` has multiple public constructors
> 4. `T` has no parameterless public constructor
> 5. `T` has no public constructor marked with `[GraphQLConstructor]`
> 
> I would simply flag as an error with no automatic code fix; let the developer mark the proper constructor or otherwise resolve the issue.  This would certainly be valuable when upgrading from v7 to v8.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/issues/3773#issuecomment-1811762073_

gao-artur:
Closes #3774.

Notes:
- I made the `InputGraphTypeAnalyzerTests` partial and moved the fields validation into `.Fields` and constructor validation into `.Constructors`. Nothing changed in the fields validation logic.
- the analyzer diagnostic is reported also on `abstract` source types.
- the analyzer is skipped on an open generic source type
```c#
 public class CustomInputObjectGraphType<TSource> : InputObjectGraphType<TSource>
     where TSource : MySource
```
- the diagnostic also reported when multiple public constructors annotated with `[GraphQLConstructor]`
- there are no tests but I tested manually that everything works also with structs, records, and records with primary constructors

### DIFF
--- a/src/GraphQL.ApiTests/BomTests.cs
+++ b/src/GraphQL.ApiTests/BomTests.cs
@@ -12,8 +12,9 @@ public class BomTests
 
         var gitRoot = new DirectoryInfo(GetPath()).Parent!.Parent!.Parent!;
         // Protection from situations when this test is copied to another repo or the folder structure changed, etc.
-        if (!File.Exists(Path.Combine(gitRoot.FullName, "src", "GraphQL.sln")))
-            throw new InvalidOperationException("Unable to find repository root");
+        var slnFile = Path.Combine(gitRoot.FullName, "src", "GraphQL.sln");
+        if (!File.Exists(slnFile))
+            throw new InvalidOperationException($"Unable to find repository root - '{slnFile}' missing.");
 
         byte[] buffer = new byte[3];
         int counter = 0;
--- a/src/GraphQL.ApiTests/BomTests.cs
+++ b/src/GraphQL.ApiTests/BomTests.cs
@@ -10,11 +10,12 @@ public class BomTests
     {
         string GetPath([CallerFilePath] string path = "") => path; // <GIT_ROOT>\src\GraphQL.ApiTests\BomTests.cs
 
-        var gitRoot = new DirectoryInfo(GetPath()).Parent!.Parent!.Parent!;
+        var callerFilePath = GetPath();
+        var gitRoot = new DirectoryInfo(callerFilePath).Parent!.Parent!.Parent!;
         // Protection from situations when this test is copied to another repo or the folder structure changed, etc.
         var slnFile = Path.Combine(gitRoot.FullName, "src", "GraphQL.sln");
         if (!File.Exists(slnFile))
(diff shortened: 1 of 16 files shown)


gao-artur:
@Shane32 do you know why it fails?

Shane32:
I’ll look at it today.

Shane32:
Probably had to do with a new .NET SDK being installed.

gao-artur:
Did you consider using `global.json` to ensure the same SDK version is used in CI and dev machines?

Shane32:
No, we usually let the .NET SDK version float to the newest available.  Generally we want the source to be compatible with the newest SDK and would rather just fix issues as they come up.  I use global.json for corporate projects where I can control or specify the version of dev tools in use.

Shane32:
In this case runner image 20231030.2.0 worked fine but runner image 20231115.7.0 did not.  But it was due to what was really a bug in our code.  GitHub hasn't updated their repo that says what's in the new runner image yet; it should be posted here:

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

That link is for the 20231030.2.0 version currently.  [Permalink](https://github.com/actions/runner-images/blob/436da67f4bd24acde9d9119870203f9dbbcf3bbe/images/ubuntu/Ubuntu2204-Readme.md)

Shane32:
Yeah, now v8 is included.

https://github.com/actions/runner-images/commit/1a6c5278e25cf94eb39bc702b3745aef75381475

src/GraphQL.ApiTests/BomTests.cs
```diff
@@ -10,10 +10,12 @@ public void Files_Should_Not_Use_BOM()
     {
         string GetPath([CallerFilePath] string path = "") => path; // <GIT_ROOT>\src\GraphQL.ApiTests\BomTests.cs
 
-        var gitRoot = new DirectoryInfo(GetPath()).Parent!.Parent!.Parent!;
+        var callerFilePath = GetPath();
+        var gitRoot = new FileInfo(callerFilePath).Directory!.Parent!.Parent!;
```
Shane32, line 14: This change is a bug fix in our tests.  `GetPath()` returns a file not a directory.

Shane32: (the comment in line 11 is still an accurate representation of what is returned from `GetPath()`)

Shane32 (approved):

src/GraphQL.ApiTests/BomTests.cs
gao-artur: Thanks!

PR #3782, opened 2023-11-20
AutoRegisteringInputObjectGraphType includes certain read-only properties

Shane32:
AutoRegisteringInputObjectGraphType includes read-only properties when the property name matches a parameter in the chosen constructor.

### DIFF
--- a/src/GraphQL.sln
+++ b/src/GraphQL.sln
@@ -179,7 +179,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "analyzers", "analyzers", "{
 		..\docs2\site\docs\analyzers\GQL007_CanNotSetSourceField.md = ..\docs2\site\docs\analyzers\GQL007_CanNotSetSourceField.md
 		..\docs2\site\docs\analyzers\GQL008_DoNotUseObsoleteArgumentMethod.md = ..\docs2\site\docs\analyzers\GQL008_DoNotUseObsoleteArgumentMethod.md
 		..\docs2\site\docs\analyzers\GQL009_UseAsyncResolver.md = ..\docs2\site\docs\analyzers\GQL009_UseAsyncResolver.md
-		..\docs2\site\docs\analyzers\GQL010_CanNotResolveInputSourceTypeConstructor.md = ..\docs2\site\docs\analyzers\GQL010_CanNotResolveInputSourceTypeConstructor.md
 		..\docs2\site\docs\analyzers\Overview.md = ..\docs2\site\docs\analyzers\Overview.md
 	EndProjectSection
 EndProject
--- a/src/GraphQL.sln
+++ b/src/GraphQL.sln
@@ -179,6 +179,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "analyzers", "analyzers", "{
 		..\docs2\site\docs\analyzers\GQL007_CanNotSetSourceField.md = ..\docs2\site\docs\analyzers\GQL007_CanNotSetSourceField.md
 		..\docs2\site\docs\analyzers\GQL008_DoNotUseObsoleteArgumentMethod.md = ..\docs2\site\docs\analyzers\GQL008_DoNotUseObsoleteArgumentMethod.md
 		..\docs2\site\docs\analyzers\GQL009_UseAsyncResolver.md = ..\docs2\site\docs\analyzers\GQL009_UseAsyncResolver.md
+		..\docs2\site\docs\analyzers\GQL010_CanNotResolveInputSourceTypeConstructor.md = ..\docs2\site\docs\analyzers\GQL010_CanNotResolveInputSourceTypeConstructor.md
 		..\docs2\site\docs\analyzers\Overview.md = ..\docs2\site\docs\analyzers\Overview.md
 	EndProjectSection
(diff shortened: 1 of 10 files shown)


src/GraphQL.sln
```diff
@@ -179,7 +179,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "analyzers", "analyzers", "{
 		..\docs2\site\docs\analyzers\GQL007_CanNotSetSourceField.md = ..\docs2\site\docs\analyzers\GQL007_CanNotSetSourceField.md
 		..\docs2\site\docs\analyzers\GQL008_DoNotUseObsoleteArgumentMethod.md = ..\docs2\site\docs\analyzers\GQL008_DoNotUseObsoleteArgumentMethod.md
 		..\docs2\site\docs\analyzers\GQL009_UseAsyncResolver.md = ..\docs2\site\docs\analyzers\GQL009_UseAsyncResolver.md
-		..\docs2\site\docs\analyzers\GQL010_CanNotResolveInputSourceTypeConstructor.md = ..\docs2\site\docs\analyzers\GQL010_CanNotResolveInputSourceTypeConstructor.md
 		..\docs2\site\docs\analyzers\Overview.md = ..\docs2\site\docs\analyzers\Overview.md
```
Shane32, line 182:
```suggestion
		..\docs2\site\docs\analyzers\GQL010_CanNotResolveInputSourceTypeConstructor.md = ..\docs2\site\docs\analyzers\GQL010_CanNotResolveInputSourceTypeConstructor.md
		..\docs2\site\docs\analyzers\Overview.md = ..\docs2\site\docs\analyzers\Overview.md
```

Shane32:
@gao-artur  The change to BomTest was completely ineffective btw.  It just happened to pick up an old runner to pass the test.  Probably `[CallerFilePath]` used to return the entire path and now just has a relative path.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -303,5 +303,22 @@ internal static TypeInformation GetTypeInformation(ParameterInfo parameterInfo)
...
+        /// parameterless constructor, or the only public contructor, or returns <see langword="null"/> otherwise.
+        /// </summary>
+        internal static ConstructorInfo? GetConstructor<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] TSourceType>()
+        {
+            var constructors = typeof(TSourceType).GetConstructors();
+            return constructors.Length switch
+            {
+                0 => null,
+                1 => constructors[0],
+                _ => Array.Find(constructors, c => c.GetCustomAttribute<GraphQLConstructorAttribute>() != null)
```
gao-artur, line 319: `Array.Find` is equivalent to `FirstOrDefault`. I think it should throw if multiple constructors are annotated with `GraphQLConstructorAttribute`. The same in the `ToObject` extension method. The analyzer already detects it as an error.

Shane32: In this situation I don’t want it to throw because if someone has overridden ParseDictionary then it should not throw. Note that this is an internal method. 

Yes I wrote the code with FirstOrDefault but the .NET Analyzer suggested changing it to Array.Find so I accepted its suggestion.

Shane32: I could rename the method “OrDefault”?

Shane32: Oh you said if multiple constructors are marked. Yes I can change that.

gao-artur: > I don’t want it to throw because if someone has overridden ParseDictionary

But I thought this feature is to avoid the manual definition of `InputObjectGraphType`'s, so there is no place to override `ParseDictionary`

Shane32: Overriding `ParseDictionary` is still possible.  It is certainly viable to derive from `AutoRegisteringInputObjectGraphType<MyInputClass>` and only override `ParseDictionary` if that's the only change that the developer wants to make.  They would not need to define all the fields.

src/GraphQL/Types/Composite/AutoRegisteringHelper.cs
```diff
@@ -303,5 +303,65 @@ internal static TypeInformation GetTypeInformation(ParameterInfo parameterInfo)
...
+                return constructors[0];
+            // if there are multiple public constructors, return the one marked with GraphQLConstructorAttribute, or the parameterless constructor, or null
+            ConstructorInfo? match = null;
+            ConstructorInfo? parameterless = null;
+            foreach (var constructor in constructors)
+            {
+                if (constructor.GetCustomAttribute<GraphQLConstructorAttribute>() != null)
+                {
+                    if (match != null)
+                        throw new InvalidOperationException($"Multiple constructors marked with {nameof(GraphQLConstructorAttribute)} found on type '{sourceType.GetFriendlyName()}'.");
```
Shane32, line 337: This is always a design error and should not occur, regardless of whether `ParseDictionary` is overridden.

gao-artur (approved):

PR #3787, opened 2023-11-20
Fix source paths for CallerFilePath

Shane32:


Shane32:
Who knows; maybe this is a bug in the SDK.  Maybe not.  Anyway, tests pass now with the 8.0 SDK within CI.

.github/workflows/test-code.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         run: dotnet build --no-restore -c Release
       - name: Build solution [Debug]
         working-directory: src
-        run: dotnet build --no-restore -c Debug
+        run: dotnet build --no-restore -c Debug -p:DeterministicSourcePaths=false
```
gao-artur, line 54: Maybe it's related to the fact that [SourceLink](https://github.com/dotnet/sourcelink#using-source-link-in-net-projects) is embedded and enabled by default in .NET 8 SDK. Even though it says

> Referencing any Source Link package in a .NET SDK 8+ project suppresses SourceLink that is included in the SDK.

the SourceLink in this repo is added under conditional `ItemGroup`
```xml
<ItemGroup Condition="'$(IsPackable)' == 'true'">
```
So it consumes the built-in version. If the .NET 8 is missing this condition, it can affect the test project.

gao-artur (approved):

Issue #3777, opened 2023-11-17
IDataLoaderContextAccessor.Context and IResolveFieldContext.RequestServices nullability are not very usable


ccurrens:
## Description

The nullable annotations for `IDataLoaderContextAccessor.Context` are correct, but not very usable for developers. It is correct to be nullable, since it only has a value during document execution and is null at other times.

The problem is that any code that has nullable references enabled and is defining fields that use data loaders either have to add null checks on `Context` or use the null forgiveness operator (e.g. `accessor.Context!.GetOrAddBatchLoader`).

One suggestion is that the `Context` property could be made non-null, and throw if the underlying asynclocal is null. Framework guidelines says to avoid throwing from getters, but it may be acceptable here since it would only be thrown if the getter is used outside of a context when it is valid -- normal use will never see this exception thrown.

We'd like to avoid having the null forgiveness operator everywhere in our code, but also if there are valid cases where this property can be null while executing field resolvers, we'd want to know that to, so we can ensure our code is correct.

### Steps to reproduce

Write code like this:

```csharp
        Field<ListGraphType<StringGraphType>>("description")
            .Resolve(context =>
            {
                var loader = accessor.Context.GetOrAddBatchLoader<Guid, string>(
                    "key",
                    storage.GetDecriptionByIdsAsync);

                return loader.LoadAsync(context.Source.ItemGuid);
            })
```

### Expected result

No CS8604 warning about possible null reference argument for parameter 'context' in 'GetOrAddCollectionBatchLoader'

### Actual result

You get a CS8604 warning about possible null reference argument for parameter 'context' in 'GetOrAddCollectionBatchLoader'

### Environment

C# with nullable reference types enabled

Shane32:
Makes sense.  In the meantime, you could add an extension method similar to this:

```cs
public static DataLoaders(this IResolveFieldContext context)
    => (context.RequestServices
        ?? throw new InvalidOperationException("RequestServices not set"))
        .GetRequiredService<IDataLoaderContextAccessor>().Context
        ?? throw new InvalidOperationException("Data loader context not initialized";

// sample
        Field<ListGraphType<StringGraphType>>("description")
            .Resolve(context =>
            {
                var loader = context.DataLoaders().GetOrAddBatchLoader<Guid, string>(
                    "key",
                    storage.GetDecriptionByIdsAsync);

                return loader.LoadAsync(context.Source.ItemGuid);
            });
```

Shane32:
[...]

PR #3792, opened 2023-11-23
Make IDataLoaderContextAccessor.Context non-null

ccurrens:
As discussed in #3777, this PR changes the null annotations for `IDataLoaderContextAccessor.Context` to be non-null. This supports the nullability contract when the property is used correctly (e.g., inside of a document execution) and when the `DataLoaderDocumentListener` is registered

### DIFF
--- a/src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
+++ b/src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
@@ -1,3 +1,5 @@
+using System.Diagnostics;
+
 namespace GraphQL.DataLoader;
 
 /// <inheritdoc cref="IDataLoaderContextAccessor"/>
@@ -9,7 +11,11 @@ public class DataLoaderContextAccessor : IDataLoaderContextAccessor
     [AllowNull]
     public DataLoaderContext Context
     {
-        get => _current.Value!;
+        get
+        {
+            Debug.Assert(_current.Value != null, "DataLoaderContext is null. Ensure that DataLoaderDocumentListener is registered in the IoC container");
+            return _current.Value!;
+        }
         set => _current.Value = value;
     }
 }
--- a/src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
+++ b/src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
@@ -6,9 +6,10 @@ public class DataLoaderContextAccessor : IDataLoaderContextAccessor
     private static readonly AsyncLocal<DataLoaderContext?> _current = new();
 
     /// <inheritdoc/>
-    public DataLoaderContext? Context
+    [AllowNull]
+    public DataLoaderContext Context
     {
-        get => _current.Value;
+        get => _current.Value!;
         set => _current.Value = value;
     }
 }
--- a/src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
+++ b/src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
@@ -15,5 +15,5 @@ public class DataLoaderContextAccessor : IDataLoaderContextAccessor
 
     [DoesNotReturn]
     private DataLoaderContext ThrowMissingContext()
-        => throw new InvalidOperationException("DataLoaderContext is null. Ensure that DataLoaderDocumentListener is registered in the IoC container");
+        => throw new InvalidOperationException("DataLoaderContext is null. Ensure that you have called IGraphQLBuilder.AddDataLoader() or otherwise have added an instance of DataLoaderDocumentListener to ExecutionOptions.Listeners.");
 }
--- a/src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
+++ b/src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
@@ -1,5 +1,3 @@
-using System.Diagnostics;
-
 namespace GraphQL.DataLoader;
 
 /// <inheritdoc cref="IDataLoaderContextAccessor"/>
@@ -11,11 +9,11 @@ public class DataLoaderContextAccessor : IDataLoaderContextAccessor
     [AllowNull]
     public DataLoaderContext Context
     {
-        get
-        {
-            Debug.Assert(_current.Value != null, "DataLoaderContext is null. Ensure that DataLoaderDocumentListener is registered in the IoC container");
-            return _current.Value!;
-        }
+        get => _current.Value ?? ThrowMissingContext();
         set => _current.Value = value;
     }
+
+    [DoesNotReturn]
+    private DataLoaderContext ThrowMissingContext()
+        => throw new InvalidOperationException("DataLoaderContext is null. Ensure that DataLoaderDocumentListener is registered in the IoC container");
 }
--- a/src/GraphQL.DataLoader/DataLoaderDocumentListener.cs
+++ b/src/GraphQL.DataLoader/DataLoaderDocumentListener.cs
@@ -26,7 +26,7 @@ public class DataLoaderDocumentListener : IDocumentExecutionListener
     /// <inheritdoc/>
     public Task BeforeExecutionAsync(IExecutionContext context)
     {
-        _accessor.Context ??= new();
+        _accessor.Context = new();
 
         return Task.CompletedTask;
     }
--- a/src/GraphQL.DataLoader/IDataLoaderContextAccessor.cs
+++ b/src/GraphQL.DataLoader/IDataLoaderContextAccessor.cs
@@ -8,5 +8,6 @@ public interface IDataLoaderContextAccessor
     /// <summary>
     /// The current <seealso cref="DataLoaderContext"/>
     /// </summary>
-    DataLoaderContext? Context { get; set; }
+    [AllowNull]
+    DataLoaderContext Context { get; set; }
 }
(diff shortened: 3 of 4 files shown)


src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
```diff
@@ -6,9 +6,10 @@ public class DataLoaderContextAccessor : IDataLoaderContextAccessor
     private static readonly AsyncLocal<DataLoaderContext?> _current = new();
 
     /// <inheritdoc/>
-    public DataLoaderContext? Context
+    [AllowNull]
+    public DataLoaderContext Context
     {
-        get => _current.Value;
+        get => _current.Value!;
```
ccurrens, line 12: I'm a little unsure about this. On the one hand, using `!` here will result in the least amount of code when the data loader feature is used correctly. But when used incorrectly, it returns null rather than having a helpful exception message. I have a hard time penalizing the correct case for this, though.

An alternative would be to use a throw helper to throw when it is null. The property could still be inlined, but there's still a conditional check...

ccurrens: Ah, another option I just thought of: `Debug.Assert`. That will help developers ensure it is setup correctly while not penalizing production use.

Shane32: Debug.Assert are for internal testing; the compiler removes these from the compiled nuget packages. So developers won’t get any benefit from Debug.Assert.

ccurrens: 🤦‍♂️ of course. I'll remove that. Do you think I should add a throw helper or leave it without one?

Shane32: I probably would. I can’t think of a reason you’d need to know the value was null. However I’d consider it a breaking change and push the change to v8

Shane32: Adding the throw helper assists the developer in knowing how to fix the issue which is a great benefit.

ccurrens: Great, I'll change the base to the develop branch and add the throw helper.

src/GraphQL.DataLoader/DataLoaderContextAccessor.cs
```diff
@@ -6,9 +6,14 @@ public class DataLoaderContextAccessor : IDataLoaderContextAccessor
...
+    public DataLoaderContext Context
     {
-        get => _current.Value;
+        get => _current.Value ?? ThrowMissingContext();
         set => _current.Value = value;
     }
+
+    [DoesNotReturn]
+    private DataLoaderContext ThrowMissingContext()
+        => throw new InvalidOperationException("DataLoaderContext is null. Ensure that DataLoaderDocumentListener is registered in the IoC container");
```
Shane32, line 18: A few comments here:
- By default, neither GraphQL.NET nor GraphQL.NET Server pull all `IDocumentExecutionListener` instances from DI, so registering the listener in the IoC container is not enough to fix the problem.  While it is possible with custom code to do so, the current suggested mechanism is to use `.AddDataLoaders()` to install the data loader document listener.  This will register `DataLoaderDocumentListener` in the DI container and install a `IConfigureExecution` instance which is set up to pull the listener from DI and add it to `ExecutionOptions.Listeners`.
- By convention, we end all exception messages with a period.

```suggestion
        => throw new InvalidOperationException("DataLoaderContext is null. Ensure that you have called IGraphQLBuilder.AddDataLoader() or otherwise have added an instance of DataLoaderDocumentListener to ExecutionOptions.Listeners.");
```

src/GraphQL.DataLoader/IDataLoaderContextAccessor.cs
```diff
@@ -8,5 +8,6 @@ public interface IDataLoaderContextAccessor
     /// <summary>
     /// The current <seealso cref="DataLoaderContext"/>
     /// </summary>
-    DataLoaderContext? Context { get; set; }
+    [AllowNull]
+    DataLoaderContext Context { get; set; }
```
Shane32, line 12: 👍

ccurrens:
Whoops, looks like there are a few test failures. Didn't run it after adding the thrown exception. I'll take a look

src/GraphQL.DataLoader/DataLoaderDocumentListener.cs
```diff
@@ -26,15 +26,15 @@ public Task AfterValidationAsync(IExecutionContext context, IValidationResult va
     /// <inheritdoc/>
     public Task BeforeExecutionAsync(IExecutionContext context)
     {
-        _accessor.Context ??= new();
```
ccurrens, line 29: The tests were failing because this did a null check before setting to a new instance. That no longer works now that we throw if it's `null`.

The obvious fix was to add a function named `Initialize` that could bypass the getter, accessing the `AsyncLocal` directly and setting it to a new instance if it was `null`. I then also decided to remove the setter altogether and introduce a function to reset the context back to `null`.

I'm not sure the setter makes sense anymore since you can't do a conditional set based on whether or not it is `null`. This is why I went ahead and removed it and replace it with a `Reset()` function to set it back to `null`.

As I write this, I realize it's a larger breaking change because consumers can no longer set the context to an arbitrary value. So maybe a) the setter should remain or b) `Reset` should take in a `DataLoaderContext?` parameter so allow consumers to set it to whatever they want.

Shane32: Maybe we should reconsider simply changing the NRT attributes. 

Also consider that this interface is modeled after `IHttpContextAccessor` and we should review/consider their implementation. 

https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.ihttpcontextaccessor.httpcontext?view=aspnetcore-7.0#microsoft-aspnetcore-http-ihttpcontextaccessor-httpcontext

Shane32: Another option is to leave existing property as is and introduce a new property like RequiredContext (similar in vein to GetRequiredService). Sorta wish it was an extension method though

ccurrens: Regarding it being modeled against `IHttpContextAccessor`, I think a good question to ask is what are valid use cases for it. In one [documentation example](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/use-http-context?view=aspnetcore-7.0#httpcontext-isnt-thread-safe) the `IHttpContextAccessor` is used in a `HttpClient` delegating handler, where depending on when the `HttpClient` is called, could result in being outside of the request flow and having the `HttpContext` nullable.

Is it considered valid usage of the `DataLoaderContext` for it to be accessed outside of document execution? If it is, the property should stay nullable and something like `RequiredContext`/`GetRequiredContext` could be introduced. If it's not valid to use it outside of document execution, then I think it's better to have it modeled against the valid use case (non-null).

----------------

Another interesting discussion on nullability is [here](https://github.com/dotnet/aspnetcore/pull/22928#discussion_r440513476), when `ISession` is changed to be non-null, even though technically custom implementations of `HttpContext` could cause it to return a non-null instance. I like the comment that there needs to be pragmatism applied to nullability attributes and focus on the 99% use case, and I think their [implementation in `DefaultHttpContext`](https://github.com/dotnet/aspnetcore/blob/744cb017a7d7d14a5e25076b9a9a8d3e4702f5d8/src/Http/Http/src/DefaultHttpContext.cs#L198) is similar to what I'm attempting here in the previous commit.

So another possible option is to revert the last commit that adds `Reset` and `Initialize`, which effectively disallows the `??=` operator. But this is not really different from what was done with `DefaultHttpContext.Session`. This makes the assumption that changing the way `DataLoaderDocumentListener` sets the context is safe:

```csharp
// before
_accessor.Context ??= new();

// after
_accessor.Context = new();
```
It seems like it's safe to skip the null check and always set it (I assume each `IDocumentExecutionListener` is only executed once). If a custom listener sets their own, it would require that they ensure `DataLoaderDocumentListener` is not added.

Shane32: Well, to think about it a bit, having `IHttpContextAccessor` return a nullable context allows code to operate differently based on whether it was null, and is useful for logging and similar scenarios.  But pulling the `IDataLoaderContext` is really only useful if you're going to use the data loader, which implies returning an `IDataLoaderResult<T>` from the method.  Sure it's theoretically possible to detect that the context is null and immediately retrieve the values, but you'd still need to wrap the values in an `IDataLoaderResult<T>`.  As such, the only practical use would be for tests, and it's much simpler to mock the data loader context in that scenario.

So I'd have to say that it's not really 'valid usage' to use `IDataLoaderContextAccessor` without the listener in use.

Agree?

Shane32: Let's go with your suggestion above, reverting the last commit and changing this:

```cs
// before
_accessor.Context ??= new();

// after
_accessor.Context = new();
```

If someone wants to be able to read `null` values from the context, they can write their own implementation of `DataLoaderContextAccessor` which allows null values or provides `Initialize` and `Reset` methods or similar.

Shane32: (Which is how the PR is now.)

Shane32:
You'll need to run the tests local so that the ApiApprovals test data is updated

src/GraphQL.DataLoader/DataLoaderDocumentListener.cs
ccurrens: > So I'd have to say that it's not really 'valid usage' to use IDataLoaderContextAccessor without the listener in use.

I very much agree. I was thinking about it yesterday and came to the same conclusion about the data loader result.

Shane32 (approved):

Shane32:
Looks good to me.  I'll leave open for a few days in case any else has any comments before merging.

cc: @gao-artur

gao-artur:
Looks good 👍

PR #3795, opened 2023-11-24
NotAGraphTypeAttribute analyzer

### LINKED ISSUE #3783: Create analyzer to simulate a "not-a-graph-type-constraint"
> Another analyzer idea: prevent invalid uses of `IGraphType`-derived classes, such as: (I did this today)
> 
> ```cs
> //valid
> GraphQLClrInputTypeReference<string>
> //invalid
> GraphQLClrInputTypeReference<StringGraphType>
> ```
> 
> Ditto for `GraphQLClrOutputTypeReference<T>` and `IDataLoaderResult<T>`.  Unfortunately there's no type constraint that says 'is not IGraphType'.  This idea could be expanded to other use cases like return types of FieldBuilder and `.Returns<T>()`.
> 
> We do have type constraints for `NonNullGraphType<T>` and `ListGraphType<T>` so no analyzer is needed there.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/issues/3732#issuecomment-1818277622_

gao-artur:
Fixes #3783.


src/GraphQL/Types/QueryArgument.cs
```diff
@@ -83,7 +83,7 @@ public string Name
         /// <summary>
         /// Gets or sets the default value of the argument.
         /// </summary>
-        public object? DefaultValue { get; set; }
+        public object? DefaultValue { get; set; } // TODO: [NotAGraphType]?
```
Shane32, line 86: This property contains an actual value (such as "abc"), not a type, so probably not necessary.

Shane32:
This looks great!  Request a review when ready.

gao-artur:
@Shane32 I'm ready. Please review.
I only annotated types and methods where you must provide type arguments explicitly. For example, in `Returns` method, you must provide the type:
```c#
Field<StringGraphType>.Returns<string>();
```

But in `DataLoaderExtensions.ResolveAsync` they are derived from the builder type implicitly and can be removed. So the validation happens on the `Field` method only:
```c#
Field<StringGraphType, string>.ResolveAsync<StringGraphType, string>();
```

docs2/site/docs/analyzers/GQL011_MustNotBeConvertibleToGraphType.md
```diff
@@ -0,0 +1,111 @@
...
+In this example, the `StringGraphType` is incorrectly used as the return type of
+the resolver
+
+```c#
+public class MyGraphType : ObjectGraphType
+{
+    public MyGraphType()
+    {
+        Field<StringGraphType>("name")
+            .Returns<string>()
```
Shane32, line 52: Is this sample correct? (Which should be a sample of incorrect code)

gao-artur: Yep, I put here the correct sample, and in the fix section, the incorrect...

Shane32 (approved):

PR #3803, opened 2023-12-03
[WIP] InputObjectParserGenerator

gao-artur:
Initial POC.

TODO:
- [x] Block namespace
- [x] Filescoped namespace
- [x] Global namespace
- [x] Parameterless constructor
- [x] Simple properties
- [ ] Parameterized constructors
- [ ] Coerces undefined fields to default for constructor arguments
- [ ] `init` and `required` properties
- [ ] Coerces undefined fields to default for required/init-only props
- [ ] Complex properties (lists and other input types)
- [ ] Support fields defined with expression `Filed(x => x.FamilyName)`
- [ ] Reuse code from `InputGraphTypeAnalyzer`
- [ ] Support NRT
- [ ] Enable generator by msbuild property
- [ ] Add `partial` analyzer and code fix when the generator is enabled
- [ ] Support `AutoRegisteringInputObjectGraphType<TSourceType>`
- [ ] Test generated code. Ideally, run the same tests set with and without the generator
- [ ] Improve [performance](https://www.thinktecture.com/en/net/roslyn-source-generators-performance/). How to measure?
- [ ] Create schema visitor to ensure all the fields available in runtime were discovered in compile time
- [ ] `ValueConverter.TryConvertTo(source, type, out object? result, typeof(IDictionary<string, object>)`?


Shane32:
Now we should have a very well-defined operation to pattern this source generator from.  No ambiguity for the constructor selection.  Constructor parameters and supported in a defined manner.  Init-only and required properties are set during initialization.  Non-public members are not supported.  Readonly public fields are not supported.

Shane32:
And as fast as dynamic compilation is, it doesn't work for scoped schemas (currently) or when the project is AOT compiled.

PR #3824, opened 2023-12-14
Fix GQL009 code fix formatting

gao-artur:
The formatting was broken when lambda's parameter or body was on a different line.

### DIFF
--- a/src/GraphQL.Analyzers.CodeFixes/AwaitableResolverCodeFixProvider.cs
+++ b/src/GraphQL.Analyzers.CodeFixes/AwaitableResolverCodeFixProvider.cs
@@ -71,11 +71,12 @@ public class AwaitableResolverCodeFixProvider : CodeFixProvider
                 // Resolve(async ctx => await AsyncMethod() or statement)
                 var newLambda = lambda
                     .WithAsyncKeyword(
-                        Token(SyntaxKind.AsyncKeyword).WithLeadingTrivia(Space))
+                        Token(SyntaxKind.AsyncKeyword).WithLeadingTrivia(lambda.GetLeadingTrivia()))
                     .WithExpressionBody(
                         AwaitExpression(
-                            Token(SyntaxKind.AwaitKeyword).WithTrailingTrivia(Space),
-                            lambda.ExpressionBody));
+                                Token(SyntaxKind.AwaitKeyword),
+                                lambda.ExpressionBody)
+                            .WithLeadingTrivia(lambda.ExpressionBody.GetLeadingTrivia()));
 
                 docEditor.ReplaceNode(lambda, newLambda);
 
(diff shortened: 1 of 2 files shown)


Shane32:
lmk if I should issue a release

Shane32 (approved):

gao-artur:
This was the only problem I found with our schema. Can you check on your schema to see if there are any formatting issues? If all good, I think we can release a fix.

Shane32:
I had an issue where installing GraphQL 7.7.1 broke my CI script, with the following error:

```
The analyzer assembly '/home/runner/.nuget/packages/graphql.analyzers/7.7.1/analyzers/dotnet/cs/GraphQL.Analyzers.dll' references version '4.6.0.0' of the compiler, which is newer than the currently running version '4.3.0.0'.
```

However, changing `actions/setup-dotnet@v1` to `v4` solved the problem.  (FYI this project uses .NET 6)

Shane32:
Only formatting thing I ran into:

image

It really did a great job with everything, making changes to over 200 lines of code perfectly!

gao-artur:
>Only formatting thing I ran into:

I think we have already talked about this problem. I'll try to take a look this weekend and see how difficult will be the fix.

gao-artur:
> I had an issue where installing GraphQL 7.7.1 broke my CI script, with the following error:
> 
> ```
> The analyzer assembly '/home/runner/.nuget/packages/graphql.analyzers/7.7.1/analyzers/dotnet/cs/GraphQL.Analyzers.dll' references version '4.6.0.0' of the compiler, which is newer than the currently running version '4.3.0.0'.
> ```
> 
> However, changing `actions/setup-dotnet@v1` to `v4` solved the problem. (FYI this project uses .NET 6)

Unfortunately, I can't find any release notes for `Microsoft.CodeAnalysis` packages. Should I downgrade the package so it can be built with older SDKs? What is the minimal SDK version we want to support?

Shane32:
I'm not sure.  What are the consequences if you simply reference 4.3.0.0 ?  If we can reference the older version with no consequence then I would do so.  One of my projects uses ASP.NET Core 2.1 (on .NET Framework) because it runs on an operating system that is not supported by .NET Core 3.1 or newer.  (Technically this combination is still supported by MS, at least until the OS goes out of support.)  But other users may have other reasons for using older TFMs/SDKs.

Shane32:
Note that I was able to upgrade to GraphQL.NET 7.7.1 in the above scenario just fine, perhaps because the CI script installs SDKs for 2.1, 3.1, 5.0, 6.0 and 7.0.

gao-artur:
I'll have to test it with older versions. I have no idea what changed between versions, as MS doesn't want to manage any release notes
https://github.com/dotnet/roslyn/issues/7150

PR #3827, opened 2023-12-14
Deprecated fields are returned when feature is disabled

### LINKED ISSUE #3825: Deprecated input fields break the schema in 7.4.0
Related issue #3658.

I'm upgrading from 7.3.0 to 7.7.1.
We have a few input types with all fields marked as deprecated. With a new version, I'm getting the following error even when `schema.Features.DeprecationOfInputValues = false;`

> {
>   "error": "Invalid schema Error:\nInput Object type ContentRecordsQueryParams must define one or more fields."
> }

With this config key set to `false` (default), I expect the field depreciation will be ignored, just like it was in the older version, but instead, the field is deprecated and not returned from the introspection query and breaks previously valid graph type.

Shane32:
Closes #3825

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug3825.cs
+++ b/src/GraphQL.Tests/Bugs/Bug3825.cs
@@ -1,0 +1,86 @@
+using GraphQL.Types;
+
+namespace GraphQL.Tests.Bugs;
+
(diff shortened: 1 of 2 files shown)


gao-artur (approved):
Thank you!

PR #3839, opened 2023-12-17
Change default type naming

### LINKED ISSUE #3465: Graph types' automatic names are not distinct for generic classes
This is a problem when there's a type such as `Connection<News>` which is generated as `Connection` rather than `ConnectionNews` or `NewsConnection`.  This then conflicts with `Connection<Article>` and similar.  Obviously we could hard-code detection for `Connection<T>` and `Edge<T>`, but I feel we should solve this for any generic type.  This is more distinctly a problem for auto-registering graph types, (and auto-registering schemas) where there typically is no class definition and so cannot be easily solved by adding a single line of code.  Since the object type is a generic type, one cannot add an attribute to the class to set the auto-generated name either.

Suggestions:

1. Include the generic type name(s) of the class when generating a GraphQL name for a class, in the order they appear.  So this means:

| Class Name | Graph Name |
|-|-|
| `News` | `News` |
| `Connection<News>` | `ConnectionNews` |
| `Data<Connection<News>>` | `DataConnectionNews` |
| `Data<News, Obj>` | `DataNewsObj` |

2. Same, but reverse order

| Class Name | Graph Name |
|-|-|
| `News` | `News` |
| `Connection<News>` | `NewsConnection` |
| `Data<Connection<News>>` | `NewsConnectionData` |
| `Data<News, Obj>` | `ObjNewsData` |

3. Introduce a global switch to configure the naming algorithm, with three options: current, option 1 above, and option 2 above.

Option 1 makes the most sense at first glance, although the current use case (relay types) would indicate that option 2 would be preferred -- in fact, required to meet GraphQL relay spec (connection types must end in "Connection").  Option 3 also has quite a bit of merit, as it could be introduced now in version 7 with the option of changing the default for version 8, while also allowing for easy backwards-compatible behavior in v8 if we change the default.  It also allows users to choose between option 1 and option 2.  However, I generally prefer to choose an specific answer over adding a global switch.  Global switches add more code to be maintained, and are unlikely for users to realize they even exist.

So at the moment, I'd just go with option 2 and put the feature in v8.  My second choice is option 3, with option 2 being the default setting in v8.  I'd also be happy to introduce a global switch for v7 to use the v8 naming strategy, removing the global switch for v8.

Shane32:
Closes #3465

### DIFF
--- a/src/GraphQL.Tests/Extensions/TypeExtensionTests.cs
+++ b/src/GraphQL.Tests/Extensions/TypeExtensionTests.cs
@@ -1,0 +1,96 @@
+using GraphQL.Types;
+using GraphQL.Types.Relay;
+using GraphQL.Types.Relay.DataObjects;
+
+namespace GraphQL.Tests.Extensions;
+
+[Collection("StaticTests")]
+public class TypeExtensionTests
+{
+    [Theory]
+
+    // v8 naming
+    [InlineData(false, false, typeof(Class1), "Class1")]
+    [InlineData(false, false, typeof(Class1Type), "Class1")]
+    [InlineData(false, false, typeof(Class1Graph), "Class1Graph")]
+    [InlineData(false, false, typeof(Class1GraphType), "Class1")]
+    [InlineData(false, false, typeof(GenericClass<>), "TGenericClass")]
+    [InlineData(false, false, typeof(GenericClassType<>), "TGenericClass")]
+    [InlineData(false, false, typeof(GenericClassGraph<>), "TGenericClassGraph")]
+    [InlineData(false, false, typeof(GenericClassGraphType<>), "TGenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1>), "Class1GenericClass")]
+    [InlineData(false, false, typeof(GenericClassType<Class1>), "Class1GenericClass")]
+    [InlineData(false, false, typeof(GenericClassGraph<Class1>), "Class1GenericClassGraph")]
+    [InlineData(false, false, typeof(GenericClassGraphType<Class1>), "Class1GenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1Type>), "Class1GenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1Graph>), "Class1GraphGenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1GraphType>), "Class1GenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1, string>), "Class1StringGenericClass")]
+    [InlineData(false, false, typeof(GenericClassType<Class1, string>), "Class1StringGenericClass")]
+    [InlineData(false, false, typeof(GenericClassGraph<Class1, string>), "Class1StringGenericClassGraph")]
+    [InlineData(false, false, typeof(GenericClassGraphType<Class1, string>), "Class1StringGenericClass")]
+    [InlineData(false, true, typeof(Class1), "TypeExtensionTests_Class1")]
+    [InlineData(false, false, typeof(NonNullGraphType<StringGraphType>), "String")]
+    [InlineData(false, false, typeof(ListGraphType<StringGraphType>), "String")]
+    [InlineData(false, false, typeof(NonNullGraphType<ListGraphType<NonNullGraphType<StringGraphType>>>), "String")]
+    [InlineData(false, false, typeof(NonNullGraphType<ListGraphType<NonNullGraphType<InputObjectGraphType<Class1>>>>), "Class1InputObject")]
+    [InlineData(false, false, typeof(ConnectionType<StringGraphType>), "StringConnection")]
+    [InlineData(false, false, typeof(Connection<string>), "StringConnection")]
+
+    // legacy naming
+    [InlineData(true, false, typeof(Class1), "Class1")]
+    [InlineData(true, false, typeof(Class1Type), "Class1")]
+    [InlineData(true, false, typeof(Class1Graph), "Class1Graph")]
+    [InlineData(true, false, typeof(Class1GraphType), "Class1")]
+    [InlineData(true, false, typeof(GenericClass<>), "GenericClass")]
+    [InlineData(true, false, typeof(GenericClassType<>), "GenericClass")]
+    [InlineData(true, false, typeof(GenericClassGraph<>), "GenericClassGraph")]
+    [InlineData(true, false, typeof(GenericClassGraphType<>), "GenericClass")]
+    [InlineData(true, false, typeof(GenericClass<Class1>), "GenericClass")]
+    [InlineData(true, false, typeof(GenericClassType<Class1>), "GenericClass")]
+    [InlineData(true, false, typeof(GenericClassGraph<Class1>), "GenericClassGraph")]
+    [InlineData(true, false, typeof(GenericClassGraphType<Class1>), "GenericClass")]
+    [InlineData(true, false, typeof(GenericClass<Class1Type>), "GenericClass")]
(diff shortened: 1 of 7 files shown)


Shane32:
@gao-artur please review.  Once the design is finished, I'll add a myriad of tests and we can review again.

src/GraphQL/Extensions/TypeExtensions.cs
```diff
@@ -63,12 +64,67 @@ internal static bool IsNamedType(this Type type)
...
+                return NameOf(type)
+                    .Replace('@', '_'); // F# anonymous class support
+
+                static string NameOf(Type type)
+                {
+                    type = type.GetNamedType();
+                    var attr = type.GetCustomAttribute<GraphQLMetadataAttribute>();
+                    if (!string.IsNullOrEmpty(attr?.Name))
+                    {
+                        return attr!.Name!;
```
gao-artur, line 86: The `!` after `attr` is redundant.

src/GraphQL/Types/GraphType.cs
```diff
@@ -51,6 +51,9 @@ private string GetDefaultName()
         {
             var type = GetType();
```
gao-artur, line 52: Should it be
```c#
return type.GraphQLName();
```
after this line, and the rest of the method removed?

gao-artur (commented):
Looks good 👍

src/GraphQL/Extensions/TypeExtensions.cs
```diff
@@ -63,12 +64,67 @@ internal static bool IsNamedType(this Type type)
...
+                            name = name.Substring(0, name.Length - "Type".Length);
+                        }
+                    }
+
+                    if (GlobalSwitches.UseDeclaringTypeNames)
+                    {
+                        var parent = type.DeclaringType;
+                        while (parent != null)
+                        {
+                            name = parent.Name + "_" + name;
```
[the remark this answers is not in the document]
Shane32:
```suggestion
                            name = $"{parent.Name}_{name}";
```

src/GraphQL/Extensions/TypeExtensions.cs
Shane32:
```suggestion
                        return attr.Name!;
```

src/GraphQL/Extensions/TypeExtensions.cs
```diff
@@ -85,8 +141,10 @@ public static string GraphQLName(this Type type)
                 typeName = typeName.Substring(0, typeName.IndexOf('`'));
             }
 
+#pragma warning disable CS1718 // Comparison made to same variable
```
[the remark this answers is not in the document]
Shane32: No, I was testing some code and forgot to remove it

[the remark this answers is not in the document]
Shane32:
```suggestion
```

src/GraphQL/Extensions/TypeExtensions.cs
```diff
@@ -87,6 +143,7 @@ public static string GraphQLName(this Type type)
 
             if (typeName == nameof(GraphType) || typeName == nameof(Type))
                 return typeName;
+#pragma warning restore CS1718 // Comparison made to same variable
```
Shane32, line 146:
```suggestion
```

src/GraphQL/Types/GraphType.cs
Shane32: No, because the legacy code worked different than the legacy version of `.GraphQLName()`.

src/GraphQL/Extensions/TypeExtensions.cs
Shane32: Unfortunately it doesn't work when targeting .NET Standard 2.0

src/GraphQL/Extensions/TypeExtensions.cs
```diff
@@ -63,12 +64,67 @@ internal static bool IsNamedType(this Type type)
...
+                return NameOf(type)
+                    .Replace('@', '_'); // F# anonymous class support
+
+                static string NameOf(Type type)
+                {
+                    type = type.GetNamedType();
+                    var attr = type.GetCustomAttribute<GraphQLMetadataAttribute>();
+                    if (!string.IsNullOrEmpty(attr?.Name))
+                    {
+                        return attr.Name!;
```
Shane32, line 86:
```suggestion
                        return attr!.Name!;
```

src/GraphQL.Tests/Extensions/TypeExtensionTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    [InlineData(false, false, typeof(GenericClassGraphType<Class1>), "Class1GenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1Type>), "Class1GenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1Graph>), "Class1GraphGenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1GraphType>), "Class1GenericClass")]
+    [InlineData(false, false, typeof(GenericClass<Class1, string>), "Class1StringGenericClass")]
+    [InlineData(false, false, typeof(GenericClassType<Class1, string>), "Class1StringGenericClass")]
+    [InlineData(false, false, typeof(GenericClassGraph<Class1, string>), "Class1StringGenericClassGraph")]
+    [InlineData(false, false, typeof(GenericClassGraphType<Class1, string>), "Class1StringGenericClass")]
+    [InlineData(false, true, typeof(Class1), "TypeExtensionTests_Class1")]
+    [InlineData(false, true, typeof(GenericClass<Class1>), "TypeExtensionTests_Class1TypeExtensionTests_GenericClass")]
```
Shane32, line 33: Hmmm.....

src/GraphQL.Tests/Extensions/TypeExtensionTests.cs
```diff
@@ -0,0 +1,100 @@
...
+public class TypeExtensionTests
+{
+    [Theory]
+
+    // v8 naming
+    [InlineData(false, false, typeof(Class1), "Class1")]
+    [InlineData(false, false, typeof(Class1Type), "Class1")]
+    [InlineData(false, false, typeof(Class1Graph), "Class1Graph")]
+    [InlineData(false, false, typeof(Class1GraphType), "Class1")]
+    [InlineData(false, false, typeof(GenericClass<>), "TGenericClass")]
```
gao-artur, line 17: Does it make sense to create a name for an open generic?

src/GraphQL.Tests/Extensions/TypeExtensionTests.cs
gao-artur: Yeah, this is ugly... Maybe the declaring type name should be applied only once? On the other hand, the generic class and generic argument can be declared in different parent classes, and users will expect to get different prefixes. Any combination of generic and nested classes seems to give an ugly solution. So, maybe just blame the users if they do that?

Shane32: Original issue was #2457 

I'm fine to leave it as shown here, letting users come up with their own solution if they run into problems.  `UseDeclaringTypeNames` always seemed to be a very edge-case fix; it doesn't help if the same class is in different namespaces, for instance.  I wouldn't have merged the feature to begin with, but it does not really hurt to leave it there.

src/GraphQL.Tests/Extensions/TypeExtensionTests.cs
Shane32: I can't imagine a situation where `GetGraphQLName` is useful for an open generic type.  The test just demonstrates that if you were to do so, it will not crash and come up with the best name it can, based on the name of the type parameter, which is in this case "T".

gao-artur (approved):

PR #3843, opened 2023-12-20
Updates to new GraphQLName function

Shane32:
After working on #3842, I realized some additional changes need to be made for proper support of the new feature.

### DIFF
--- a/src/GraphQL.Tests/Types/ObjectGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ObjectGraphTypeTests.cs
@@ -2,6 +2,7 @@ using GraphQL.Types;
 
 namespace GraphQL.Tests.Types;
 
+[Collection("StaticTests")]
 public class ObjectGraphTypeTests
 {
     private class TestInterface : InterfaceGraphType { }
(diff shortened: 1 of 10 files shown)


src/GraphQL.Tests/Types/ObjectGraphTypeTests.cs
```diff
@@ -51,10 +52,21 @@ public void should_throw_on_invalid_graphtype_name()
         }
     }
 
-    [Fact]
-    public void should_ignore_graphqlmetadata_attribute()
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public void should_ignore_graphqlmetadata_attribute(bool useLegacyTypeNaming)
```
Shane32, line 58: This test specifically ensured that the GraphQLMetadata attribute was ignored when naming a class.  This is correct; the metadata attribute is not designed for use by graph classes, but only for model classes.  So I've changed the GraphQLName logic to ignore checking for the attribute for graph classes.

src/GraphQL/Introspection/__Type.cs
```diff
@@ -27,7 +27,7 @@ public __Type(bool allowAppliedDirectives = false)
         /// </param>
         public __Type(bool allowAppliedDirectives = false, bool deprecationOfInputValues = false)
         {
-            Name = nameof(__Type);
+            SetName(nameof(__Type), validate: false);
```
Shane32, line 30: These changes align all of the introspection types to set their own type name in the same fashion.  (Some other existing introspection types already set the name this way.)  Previously this code was not necessary as the default name matched the set name and so skipped validation anyway.

gao-artur (approved):

PR #3848, opened 2023-12-23
Allowed derived telemetry providers

Shane32:


src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
```diff
@@ -44,11 +45,33 @@ public GraphQLTelemetryProvider(GraphQLTelemetryOptions options)
...
     /// <inheritdoc/>
     public virtual float SortOrder => GraphQLBuilderExtensions.SORT_ORDER_CONFIGURATION;
 
     /// <inheritdoc/>
-    public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
+    public virtual Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
+    {
+        return !_onlyWhenAutoTelemetryEnabled || OpenTelemetry.AutoInstrumentation.Initializer.Enabled
+            ? ExecuteInternalAsync(options, next)
+            : next(options);
```
Shane32, line 71: We can't change AddTelemetry to run before GraphQLBuilderBase adds the telemetry to the pipeline, but we can have the auto-added telemetry implementation check Enabled at runtime, so if it is subsequently disabled by a call to AddTelemetry, the auto-added telemetry is an allocation-free noop.

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -1206,10 +1207,31 @@ public static IGraphQLBuilder UseTelemetry(this IGraphQLBuilder builder, Action<
...
     public static IGraphQLBuilder UseTelemetry(this IGraphQLBuilder builder, Action<GraphQLTelemetryOptions, IServiceProvider>? configure)
     {
+        OpenTelemetry.AutoInstrumentation.Initializer.Enabled = false;
         builder.Services.Configure(configure);
         builder.Services.TryRegister<IConfigureExecution, GraphQLTelemetryProvider>(ServiceLifetime.Singleton, RegistrationCompareMode.ServiceTypeAndImplementationType);
         return builder;
     }
+
+    /// <inheritdoc cref="UseTelemetry(IGraphQLBuilder, Action{GraphQLTelemetryOptions}?)"/>
+    public static IGraphQLBuilder UseTelemetry<TProvider>(this IGraphQLBuilder builder)
```
gao-artur, line 1217: How do you pass the `GraphQLTelemetryOptions`? And if `TProvider : GraphQLTelemetryProvider`, it may be useful to make `TOptions : GraphQLTelemetryOptions`, so the options can be extended to support additional configuration for the extended provider.

src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
Shane32: Nope, RegisterDefaultServices occurs after the builder delegate.

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
Shane32: Same issue as `ErrorInfoProvider` and other builder methods.  The user will have to configure options with `Configure()` to configure the options class they are using.

gao-artur: Why no provide an overload with options?

Shane32: This scenario it might make sense to add `TOptions : GraphQLTelemetryOptions` since if someone wanted a fully custom telemetry provider they should use `ConfigureExecution<TProvider>`

Shane32: Whereas `IErrorInfoProvider` does not require any options; just the default implementation allows for certain options.  Users should be able to write their own implementation without necessarily using options.

Shane32: Or here we could just add two more overloads........

src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
```diff
@@ -44,11 +45,33 @@ public GraphQLTelemetryProvider(GraphQLTelemetryOptions options)
...
+    /// </summary>
+    private GraphQLTelemetryProvider()
+    {
+        _onlyWhenAutoTelemetryEnabled = true;
+        _telemetryOptions = OpenTelemetry.AutoInstrumentation.Initializer.Options ?? new GraphQLTelemetryOptions();
+    }
+
+    // make sure no accidental use of the default constructor by DI by making it private,
+    // accessible only through AutoTelemetryProvider, a shared static instance
+    internal static GraphQLTelemetryProvider? _autoTelemetryProvider;
```
gao-artur, line 60: Should be `private`?

Shane32: yes

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
Shane32: I don't mind more overloads.  But realize that now someone needs to list both the provider class and options class:

```cs
b.UseTelemetry<MyProvider, MyProviderOptions>(o => { /* options */ });

// or
b.UseTelemetry<MyProvider>();
b.Configure(new MyProviderOptions { /* options */ });
```

Shane32: @gao-artur comments?

gao-artur: It depends on how many overloads are you ready to provide 😁
There can be an overload with `TProvider` and `GraphQLTelemetryOptions` and another with `TProvider` and `TOptions`

Shane32: Sometimes simpler is better.  How about an xml comment that says how to configure options!?!?  Do we really need to add 4 more options here?

```cs
    public static IGraphQLBuilder UseTelemetry<TProvider>(this IGraphQLBuilder builder, Action<GraphQLTelemetryOptions> options)
    public static IGraphQLBuilder UseTelemetry<TProvider>(this IGraphQLBuilder builder, Action<GraphQLTelemetryOptions, IServiceProvider> options)
    public static IGraphQLBuilder UseTelemetry<TProvider, TOptions>(this IGraphQLBuilder builder, Action<TOptions> options)
    public static IGraphQLBuilder UseTelemetry<TProvider, TOptions>(this IGraphQLBuilder builder, Action<TOptions, IServiceProvider> options)
```

It seems like overkill.  Nobody has ever asked for more overloads for AddErrorInfoProvider

Shane32: It's not like it really hurts.  It's just a larger codebase to maintain.  Right now I think there are tests covering every overload (I didn't do the added ones in this PR yet).  Adding 4 more overloads means 4 more tests, too.

gao-artur: I think these 4 are the required minimum. There may be a need for a factory method to create `TProvider`, but we can wait till someone requests it.

Shane32: We don't have similar methods for `UseMemoryCache` or `UseAutomaticPersistedQueries` either.

gao-artur: Thinking again, who will need an `IServiceProvider` to construct options? I don't think I ever needed something like that.

Shane32: It's really useful when you want to inspect IHttpContext before setting options.  So for UseErrorInfoProvider you can use it to look for a specific header or IP address or similar before enabling detailed error reports.

gao-artur: Ok, then. Let's keep only the first two?

Shane32: It's also useful when you set options from `IConfiguration` (environment variables, app settings or user secrets); you just pull the options class from DI.  But of course usually `IConfiguration` can be had prior and so it's not necessary to pull from DI.

Shane32: Well, basically I'd just do all 4 together on request.  But if you're requesting them then I'd just add all 4 now.

gao-artur: Yeah, you are right. I recall now that I used this pattern when need to create options with dependencies
```c#
services.AddOptions<IConfiguration>()
    .Configure<GraphQLTelemetryOptions>((configuration, options) => { });
```

Shane32: Another way to think of it is that people can just `AddProvider<T>` and they should be quite familiar with DI and the fact that if you have an options class, they need to `AddSingleton` to register it

Shane32: Having these extra overloads should not be confusing for any user.

gao-artur: > Well, basically I'd just do all 4 together on request. But if you're requesting them then I'd just add all 4 now.

Please add them. Because I just copied the `UseTelemetry` method and updated it to register the derived type, because I didn't find a simpler way to do that.

gao-artur (approved):

Shane32:
I still need to add tests for the builder overloads but i think this is everything

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -1202,23 +1202,45 @@ public static IGraphQLBuilder AddExecutionStrategy<TExecutionStrategy>(this IGra
     /// When applicable, place after calls to UseAutomaticPersistedQueries to ensure that the query document is recorded properly.
     /// </remarks>
     public static IGraphQLBuilder UseTelemetry(this IGraphQLBuilder builder, Action<GraphQLTelemetryOptions>? configure = null)
-        => UseTelemetry(builder, configure != null ? (opts, _) => configure(opts) : null);
+        => UseTelemetry<GraphQLTelemetryProvider>(builder, configure);
```
gao-artur, line 1205: Please update the link to semantic conventions. It moved here
https://opentelemetry.io/docs/specs/semconv/database/graphql/

Shane32: Done

PR #3850, opened 2023-12-23
Update the `GraphQLTelemetryOptions.Filter` delegate to not create any downstream traces when `false`

### LINKED ISSUE #3847: GraphQLTelemetryProvider improvements
**Filter:**
In the current implementation, `GraphQLTelemetryOptions.Filter` acts more like `Skip` or `Suppress`. It only skips the GraphQL span creation, but all the following spans will be created as usual. The `Filter` option is usually used to "cut" the span creation from this point, so, for example, if the graph resolver makes a DB roundtrip, this won't be recorded. See my [implementation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1440/files#diff-607b3cc0c6a27a031377ba9f187ef72a13c9a3c7c623088a604e3d1f92bb0336) in the `Hangfire` instrumentation, for example (which is based on how this feature is implemented in other instrumentations).
When is this useful? For example, we have a few mutations that, in some cases, can create a huge trace (20k+ spans). Parts of these traces can be dropped because of memory limits, breaking the trace and making it useless. We want to be able to filter these traces and not create them at all based on mutation names.

**EnrichWith\*:**
Currently, `EnrichWith*` delegates are executed regardless of the `Activity.IsAllDataRequested` state which is a waste of CPU. When the `IsAllDataRequested = false`, the span (activity) won't be recorded anyway, so adding tags is useless. Note that the recommended way to add tags is by calling [SetTag](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.settag?view=net-8.0) method that performs tags dedup, which is fast but not free.

**Set\*TagsAsync:**
Similar to `EnrichWith*`, the `Set*TagsAsync` should be called only when `Activity.IsAllDataRequested = true`.

gao-artur:
Fixes #3847.

### DIFF
--- a/src/GraphQL/Telemetry/GraphQLTelemetryOptions.cs
+++ b/src/GraphQL/Telemetry/GraphQLTelemetryOptions.cs
@@ -24,7 +24,10 @@ public class GraphQLTelemetryOptions
     public Func<ExecutionOptions, string?> SanitizeDocument { get; set; } = options => options.Query;
 
     /// <summary>
-    /// Indicates whether the GraphQL execution is included in the telemetry.
+    /// Indicates whether to collect telemetry about the GraphQL query execution.
+    /// If filter returns <see langword="true" />, the telemetry is collected.
+    /// If filter returns <see langword="false" />, the telemetry about the GraphQL query
+    /// and all the downstream calls is not collected.
     /// </summary>
     public Func<ExecutionOptions, bool> Filter { get; set; } = _ => true;
 
--- a/src/GraphQL/Telemetry/GraphQLTelemetryOptions.cs
+++ b/src/GraphQL/Telemetry/GraphQLTelemetryOptions.cs
@@ -26,7 +26,7 @@ public class GraphQLTelemetryOptions
     /// <summary>
     /// Indicates whether to collect telemetry about the GraphQL query execution.
     /// If filter returns <see langword="true" />, the telemetry is collected.
-    /// If filter returns <see langword="false" />, the telemetry about the GraphQL query
+    /// If filter returns <see langword="false" />, the telemetry for the GraphQL query
     /// and all the downstream calls is not collected.
     /// </summary>
     public Func<ExecutionOptions, bool> Filter { get; set; } = _ => true;
--- a/src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
+++ b/src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
@@ -85,21 +85,8 @@ public class GraphQLTelemetryProvider : IConfigureExecution
         if (!activity.IsAllDataRequested)
             return await next(options).ConfigureAwait(false);
 
-        try
-        {
-            if (!_telemetryOptions.Filter(options))
-            {
-                activity.IsAllDataRequested = false;
-                activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
-                return await next(options).ConfigureAwait(false);
-            }
-        }
-        catch
-        {
-            activity.IsAllDataRequested = false;
-            activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
+        if (!Filter(options, activity))
             return await next(options).ConfigureAwait(false);
-        }
 
         // record the requested operation name and optionally the GraphQL document
         await SetInitialTagsAsync(activity, options).ConfigureAwait(false);
@@ -210,6 +197,31 @@ public class GraphQLTelemetryProvider : IConfigureExecution
         }
         _telemetryOptions.EnrichWithExecutionResult(activity, executionOptions, result);
         return Task.CompletedTask;
+    }
+
+    private bool Filter(ExecutionOptions options, Activity activity)
+    {
+        try
+        {
+            if (!_telemetryOptions.Filter(options))
+            {
+                activity.IsAllDataRequested = false;
+                activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
+                return false;
+            }
+        }
+        catch (Exception ex)
+        {
+#if NET6_0_OR_GREATER
+            activity.SetStatus(ActivityStatusCode.Error);
+#else
+            activity.SetTag("otel.status_code", "ERROR");
+#endif
+            _telemetryOptions.EnrichWithException(activity, ex);
+            throw;
+        }
+
+        return true;
     }
 
     // note: this could be implemented as a validation rule with no public API changes
--- a/src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
+++ b/src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
@@ -50,9 +50,6 @@ public class GraphQLTelemetryProvider : IConfigureExecution
     /// <inheritdoc/>
     public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
     {
-        if (!_telemetryOptions.Filter(options))
-            return await next(options).ConfigureAwait(false);
-
         // start the Activity, in fact Activity.Stop() will be called from within Activity.Dispose() at the end of using block
 #pragma warning disable CS0618 // Type or member is obsolete
         using var activity = await StartActivityAsync(options).ConfigureAwait(false);
@@ -61,6 +58,25 @@ public class GraphQLTelemetryProvider : IConfigureExecution
         // do not record any telemetry if there are no listeners or it decided not to sample the current request
         if (activity == null)
             return await next(options).ConfigureAwait(false);
+
+        if (!activity.IsAllDataRequested)
+            return await next(options).ConfigureAwait(false);
+
+        try
+        {
+            if (!_telemetryOptions.Filter(options))
+            {
+                activity.IsAllDataRequested = false;
+                activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
+                return await next(options).ConfigureAwait(false);
+            }
+        }
+        catch
+        {
+            activity.IsAllDataRequested = false;
+            activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
+            return await next(options).ConfigureAwait(false);
+        }
 
         // record the requested operation name and optionally the GraphQL document
         await SetInitialTagsAsync(activity, options).ConfigureAwait(false);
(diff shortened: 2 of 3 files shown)


src/GraphQL/Telemetry/GraphQLTelemetryOptions.cs
```diff
@@ -24,7 +24,10 @@ public class GraphQLTelemetryOptions
     public Func<ExecutionOptions, string?> SanitizeDocument { get; set; } = options => options.Query;
 
     /// <summary>
-    /// Indicates whether the GraphQL execution is included in the telemetry.
+    /// Indicates whether to collect telemetry about the GraphQL query execution.
+    /// If filter returns <see langword="true" />, the telemetry is collected.
+    /// If filter returns <see langword="false" />, the telemetry about the GraphQL query
```
Shane32, line 29:
```suggestion
    /// If filter returns <see langword="false" />, the telemetry for the GraphQL query
```

src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs
```diff
@@ -62,6 +59,25 @@ public virtual async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options
...
+                activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
+                return await next(options).ConfigureAwait(false);
+            }
+        }
+        catch
+        {
+            activity.IsAllDataRequested = false;
+            activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
+            return await next(options).ConfigureAwait(false);
+        }
```
Shane32, line 79:
```suggestion
        if (!_telemetryOptions.Filter(options))
        {
            activity.IsAllDataRequested = false;
            activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
            return await next(options).ConfigureAwait(false);
        }
 ```
I don't think we should catch exceptions within the filter delegate.  Also with the prior code, exceptions thrown from `next(options)` would trigger running `next(options)` again.

Comment (gao-artur) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:79]: That how it's done in other instrumentations. The philosophy of OTel says telemetry bugs, shouldn't break the application.

Comment (gao-artur) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:79]: You are right about the "trigger again". Will check it tommorow.

Comment (Shane32) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:79]: These are not telemetry bugs.  This is user code.  Just as if there was a bug in any other user code.  Sure if you were OpenTelemetry writing a plugin, you wouldn't want auto-enabled telemetry in GraphQL to then crash the user's application.  But the filter delegate is user code.

**Comment by Shane32** [comment]
My only remaining concern is that SetStatus won't be called if IsAllDataRequested is false.  It seems like it should at least set the status to Error at the appropriate times.

File: src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs (hunk 2 above)
Comment (gao-artur) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:79]: You are right, this is user code. But this code is related to telemetry and not to business logic. It's ok to fail to start the application if telemetry is misconfigured, but failing the already running application when emitting traces is less acceptable.
Literally every instrumentation in core and contrib OTel repos catching errors from the filter delegate, and users will expect the same from GraphQL:
[AspNet](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/6bb57e25b284ed500981a91afa6156c8ed0fa4d4/src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInListener.cs#L52-L74)
[AspNetCore](https://github.com/open-telemetry/opentelemetry-dotnet/blob/0889e8dc32badbb4c9be8fd24a974fb7e61fa779/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L155-L171)
[Http](https://github.com/open-telemetry/opentelemetry-dotnet/blob/0889e8dc32badbb4c9be8fd24a974fb7e61fa779/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L120-L136)
[EF Core](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/6bb57e25b284ed500981a91afa6156c8ed0fa4d4/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L169-L189)
[SqlClient](https://github.com/open-telemetry/opentelemetry-dotnet/blob/0889e8dc32badbb4c9be8fd24a974fb7e61fa779/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs#L76-L92)
[Hangfire](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/6bb57e25b284ed500981a91afa6156c8ed0fa4d4/src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs#L55-L69)

The only difference is they are not silently swallowing these exceptions but logging `EventSource` events. We can do the same if this PR is merged https://github.com/graphql-dotnet/graphql-dotnet/pull/3539.

I can add a note about this behavior in XML docs to make it clear. Let me know what is your final decision on this, and I'll change the code accordingly.

**Comment by gao-artur** [comment]
> My only remaining concern is that SetStatus won't be called if IsAllDataRequested is false

This reminds me about this [comment](https://github.com/open-telemetry/opentelemetry-dotnet/issues/3290#issuecomment-1215993120) that wasn't addressed. It's probably related to specs rather than implementation.
I checked multiple instrumentations, and `SetStatus` is always called under `if (activity.IsAllDataRequested)`. I understand your concerns about coupling to OTel, but I have zero experience with other tracing systems and can answer only in the context of OTel. I think it's better to start with a specific tracing system compliance and extend it when a request to support other systems comes.

File: src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs (hunk 2 above)
Comment (Shane32) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: Well, if we don't have any MS references to copy, then we will need to refer to OTel sources.  Maybe we should review #3539 first.

Comment (Shane32) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: It just seems like a critical design error to swallow exceptions.  I can't imagine why tracing would be disabled when an exception was encountered either.  Wouldn't that just make it harder to track down what went wrong?  I'm okay following a standardized pattern, but just saying...

Comment (Shane32) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: At least with #3539 then it gets logged....somewhere...we hope...

**Comment by Shane32** [comment]
> > My only remaining concern is that SetStatus won't be called if IsAllDataRequested is false
> 
> This reminds me about this [comment](https://github.com/open-telemetry/opentelemetry-dotnet/issues/3290#issuecomment-1215993120) that wasn't addressed. It's probably related to specs rather than implementation. I checked multiple instrumentations, and `SetStatus` is always called under `if (activity.IsAllDataRequested)`. I understand your concerns about coupling to OTel, but I have zero experience with other tracing systems and can answer only in the context of OTel. I think it's better to start with a specific tracing system compliance and extend it when a request to support other systems comes.

Ok

File: src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs (hunk 2 above)
Comment (gao-artur) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: To be honest, `EventSource` feels like very old technology that is only old .NET Framework developers are familiar with 😅
I'd prefer to use `ILogger` for internal logging. If you don't want to get dependent on `Microsoft.Extensions.Logging.Abstuctions` package, we can introduce an internal logging interface + adapter for `ILogger` similar to what you did with `IServiceRegister => IServiceCollection`

Comment (Shane32) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: My thoughts also https://github.com/graphql-dotnet/graphql-dotnet/pull/3539#issuecomment-1437030956

Comment (Shane32) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: What about removing the exception swallowing for now until we introduce some type of logging mechanism?

Comment (Shane32) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: I'd be open to adding some dependencies to GraphQL.NET for v8 - specifically:

- [Microsoft.Extensions.DependencyInjection.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection.Abstractions)
- [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource)
- [Microsoft.Extensions.Logging.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions)

Each of these have around 4 billion downloads.  They are not going anywhere and have minimal API exposure.  We can reference older versions to ensure that there are no versioning conflicts.

Comment (gao-artur) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: > What about removing the exception swallowing for now until we introduce some type of logging mechanism?

Ok

Comment (gao-artur) [reply src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs:102]: Some thoughts about the subject. TL;DR: use `ILogger`
https://github.com/dotnet/diagnostics/issues/1605

File: src/GraphQL/Telemetry/GraphQLTelemetryProvider.cs [hunk 3]
```diff
@@ -196,6 +199,31 @@ protected virtual Task SetResultTagsAsync(Activity activity, ExecutionOptions ex
...
+            }
+        }
+        catch (Exception ex)
+        {
+#if NET6_0_OR_GREATER
+            activity.SetStatus(ActivityStatusCode.Error);
+#else
+            activity.SetTag("otel.status_code", "ERROR");
+#endif
+            _telemetryOptions.EnrichWithException(activity, ex);
```
gao-artur, line 220: Many instrumentation libraries provide `RecordException` boolean to [record the exception information](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md) on the span. `OpenTelemetry.Api` provides a convenient extension method to implement this functionality. I'd like to add this option, too, but since we don't reference the `OpenTelemetry.Api` package, we will have to copy their code. Strictly saying, this is not really necessary because the user can do that in `EnrichWithException` delegate. Boolean is just more convenient. What do you think?

Shane32: We could create a protected method with default functionality.  It seems to me that the default delegate in the options class should do nothing, and only be used for adding additional tags.

Shane32 (approved):

PR #3855, opened 2023-12-27
Add other .NET 8 tests

Shane32:


Shane32:
We can probably drop .NET 5 testing for the `develop` branch, I think.

src/GraphQL.DataLoader.Tests/GraphQL.DataLoader.Tests.csproj
```diff
@@ -2,15 +2,15 @@
...
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(SingleTestPlatform)' != 'true' and '$(OS)' == 'Windows_NT'">
-    <TargetFrameworks>net7.0;net6.0;net5.0;netcoreapp3.1;net48</TargetFrameworks>
+    <TargetFrameworks>net8.0;net7.0;net6.0;net5.0;netcoreapp3.1;net48</TargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(SingleTestPlatform)' != 'true' and '$(OS)' != 'Windows_NT'">
-    <TargetFrameworks>net7.0;net6.0;net5.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net8.0;net7.0;net6.0;net5.0;netcoreapp3.1</TargetFrameworks>
```
gao-artur, line 13: Did you consider moving these definitions to `Test.props` so there will be a single file to maintain?

Shane32: Well, it works if we only use `<TargetFrameworks>` and never `<TargetFramework>`

gao-artur: Do you mean this doesn't work when defined in `Test.props`?
```xml
  <PropertyGroup Condition="'$(SingleTestPlatform)' == 'true'">
    <TargetFramework>net8.0</TargetFramework>
  </PropertyGroup>
```
And this in a test project
```
  <PropertyGroup>
    <SingleTestPlatform>true</SingleTestPlatform>
  </PropertyGroup>
```

Shane32: No, I mean to override the default set of frameworks with net8.0, such as for the api checks, you have to use `TargetFrameworks`.  Maybe you want to play with it some more; I didn't spend too much time on it.  I did try to add a condition so that if TargetFramework was set it would clear TargetFrameworks but that didn't work either.  the .NET build tools gave lots of errors.

Shane32: your example wouldn't work, however, because Tests.props is imported at the top of the props file for each test project.  we would have to move the import or add a test.targets or something

gao-artur: > Tests.props is imported at the top of the props file

yeah, you are right

src/GraphQL.Harness.Tests/GraphQL.Harness.Tests.csproj
```diff
@@ -2,11 +2,11 @@
...
 
   <PropertyGroup Condition="'$(SingleTestPlatform)' == 'true'">
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFrameworks>$(SingleFramework)</TargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(SingleTestPlatform)' != 'true'">
-    <TargetFrameworks>net7.0;net6.0;net5.0</TargetFrameworks>
+    <TargetFrameworks>net8.0;net7.0;net6.0;net5.0</TargetFrameworks>
   </PropertyGroup>
```
gao-artur, line 10: Why these are still in the test project? These tests can't run on `net48`?

Shane32: No, GraphQL.Harness is designed for ASP.NET Core 3.x and above.  ASP.NET Core 2.x (which supports netfx) is notably different, with the libraries resting in nuget packages instead of within the shared framework.  I'm sure it could be made compatible with extra effort, although the testing library Alba currently used by GraphQL.Harness.Tests does not support ASP.NET Core 2.x

gao-artur: Not compatible packages. I see.

gao-artur (approved):

src/GraphQL.Harness.Tests/GraphQL.Harness.Tests.csproj
Shane32: You may be interested in reviewing the GraphQL.Server csproj and how we made that work for both netfx and .NET Core:

https://github.com/graphql-dotnet/server/blob/master/src/Transports.AspNetCore/Transports.AspNetCore.csproj

Note that although it may be possible to run ASP.NET Core 2.x on .NET Core 3.x, that scenario is not supported by the Server project.  It also does not target .NET 3.0, which means it's probably not even possible to use GraphQL.NET Server on .NET 3.0 at all.

There are also separate sample projects there for netfx vs .NET Core, which is where the end-to-end tests reside.

Shane32:
@gao-artur Moving the `<TargetFrameworks>` out of the csproj into the tests.props is really wreaking havoc on my PC.  After restarting VS two or three times, switching branches a few times, clearing and rebuilding a bunch, it finally seems I can again work on a PR.  Doesn't seem stable though.  Let me know if this is a fluke or if you're experiencing problems also.  Maybe we need to move the `<TargetFrameworks>` back to the csproj files.

Shane32:
It's causing problems on my laptop and desktop both.

gao-artur:
I think, I know what you are talking about. Usually, when different branches have different target frameworks, VS goes a bit crazy when switch from one to another. Then I close VS, remove all the bin and obj folders and rebuild. Maybe, moving out the target frameworks makes this problem to appear more frequently. I'll check this tommorow and update you.

Shane32:
Yeah, let's just keep an eye on it.  It's a real pain when the development environment is difficult to work in.  Hopefully the issue was just temporary.

gao-artur:
@Shane32 after removing `bin/obj` folders, I no longer see errors when switching branches.

PR #3857, opened 2023-12-27
Add input value coercion and validation (v2)

Shane32:
Replaces:
- #3758 

See also:
- #1176


Shane32:
@gao-artur Perhaps when you get a chance you could review the migration doc to see what the purpose and scope of the changes in this PR are, and lmk any comments you have.

gao-artur:
> review the migration doc

Great explanation! I finally understand the purpose of the feature, and it looks great!
A couple of questions:
- Is it possible to override the `Parser` when using expression syntax? For example, I may want to use a custom collection implementation for `ListGraphType<>` or construct an object using an internal contractor of a factory method.
- Is it possible to define a "xor" validator? For example, I may want to validate that only one argument was specified: `Id` or `Name`. I think this is achievable if you add some context to pass metadata between validators.

Shane32:
> * Is it possible to override the `Parser` when using expression syntax?

Yes; calling ParseValue replaces Parser with the provided delegate. The expression syntax merely sets a default parser since it knows the CLR type that the value must be coerced to. 

> * Is it possible to define a "xor" validator

By overriding ParseDictionary it is possible. I’d prefer to keep these methods more “pure” and in keeping with the context-less style of all other parsing in the library. Scalars’ ParseValue and objects’ ParseDictionary do not have context.

Adding context may be a benefit in any of those situations, like if you wanted parsing to change based on a directive set on the field. But so far it doesn’t exist anywhere else so I wouldn’t add it only here.

In your example, validation would need access to the full dictionary of values anyway, making it similar to ParseDictionary in that respect. Oh and it would need to run after all values are coerced, complicating the execution. 

By the way, this PR could be compared somewhat to the sample validation rule included in GraphQL.NET which allows a maximum length of a string set on a field. But the validation rule incurs a dictionary lookup for every field to determine if the directive was set. And it’s a real pain to get working in every scenario (literal, variable, etc). Whereas this PR is easy to use and has a very minimal performance impact.

gao-artur:
> By overriding ParseDictionary it is possible

Yeah, but this is much more complex than using a delegate 🙂

> validation would need access to the full dictionary of values

Not necessarily. The argument validate delegate can check if the exclusive flag is already set on the context (simple `Dictionary<string, object>`). If it is, throw a validation exception. If not, set the flag.

But keeping this feature simple is a good enough reason for not adding the context.

Shane32:
The next biggest question I have is whether the new delegates should run for null values? Now it does but it seems like maybe that’s pointlessly complicating this functionality. Validation rules are as a general rule of thumb going to check for disallowed null values. 

The only reason I’d even consider it is because GraphQL.NET does allow a scalar having a non-null GraphQL representation coerce to null within .NET. (Probably never used functionality.)

A 3rd option is to make the ParseValue and Validate methods create a lambda which skips for null and runs the passed function otherwise, and sets Parser or Validator to the new lambda. So you can directly set Parser or Validator if you really want to handle nulls.

Shane32:
I’d probably recommend that Parser and Validator do not run for null ever. Or “option 3”.

gao-artur:
I have no opinion here. Sorry...

Shane32:
Then I’ll probably change that. It will simplify using this feature.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -660,6 +660,11 @@ namespace GraphQL
     {
         public OutputTypeAttribute() { }
     }
+    public static class QueryArgumentExtensions
+    {
+        public static GraphQL.Types.QueryArgument ParseValue(this GraphQL.Types.QueryArgument argument, System.Func<object, object> parseValue) { }
+        public static GraphQL.Types.QueryArgument Validate(this GraphQL.Types.QueryArgument argument, System.Action<object> validator) { }
+    }
```
Shane32, line 667: These are extension methods directly on `QueryArgument` because we have no builder classes for arguments.  Rather, the configure delegate of `.Argument<int>("age", configure: arg => { })` passes the `QueryArgument` directly.

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -132,6 +132,38 @@ protected FieldBuilder(FieldType fieldType)
...
+        public virtual FieldBuilder<TSourceType, TReturnType> Validate(Action<object> validation)
+        {
+            var oldValidator = FieldType.Validator;
+            FieldType.Validator = oldValidator == FieldType.DefaultValidator
+                ? validation
+                : value =>
+                {
+                    oldValidator(value);
+                    validation(value);
+                };
```
gao-artur, line 162: This `else` section can be simplified
```c#
: FieldType.Validator + validation;
```

src/GraphQL/Types/QueryArgument.cs
```diff
@@ -117,6 +117,32 @@ public string Name
             return type;
         }
 
+        /// <summary>
+        /// Parses the value received from the client when the value is not <see langword="null"/>.
+        /// Occurs during validation prior to <see cref="Validator"/>.
+        /// Throw an exception if necessary to indicate a problem.
+        /// Only applicable to fields of input graph types.
```
gao-artur, line 124: Is there a plan to finish the #3583 for v8? If not, it may be worth creating an analyzer to error on incorrect usage. Or at least throw a runtime exception during initialization.

src/GraphQL/Extensions/QueryArgumentExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    public static QueryArgument Validate(this QueryArgument argument, Action<object> validator)
+    {
+        var oldValidator = argument.Validator;
+        argument.Validator = oldValidator == QueryArgument.DefaultValidator
+            ? validator
+            : value =>
+            {
+                oldValidator(value);
+                validator(value);
+            };
```
gao-artur, line 35: Can be simplified
```c#
: argument.Validator + validator;
```

src/GraphQL/Builders/FieldBuilder.cs
Shane32: Nice!

src/GraphQL/Types/QueryArgument.cs
Shane32: > Is there a plan to finish the https://github.com/graphql-dotnet/graphql-dotnet/pull/3583 for v8?

I don't think so.  I don't have the time to work on it, and it would probably cause a number of breaking changes anyway, depending on the final implementation.

> If not, it may be worth creating an analyzer to error on incorrect usage. Or at least throw a runtime exception during initialization.

I agree; maybe even both.  So throw an exception during initialization if `fieldType.Parser` or `fieldType.Validator` was set for an input type?

FYI the advantage of not throwing an exception during initialization (and only using an analyzer) is so that dynamic code (such as the auto-registering types' code) need not know if it's an input field or not.  So far within GraphQL.NET this isn't a problem.

I'll add the initialization check and you can comment if you think it's better without it.  If you have time you can write an analyzer, or we can just stick with the initialization check.

Shane32: We already have checks to be sure Resolver and StreamResolver are not set for input fields, so certainly we should add these checks.

gao-artur: I think the exception should be there because no one loves silent errors. This is a new feature, so when people start using it, they will immediately find they used it wrongly and will read the documentation more carefully. Adding an exception in a later stage is usually a problem because the code that is used to work will start throwing after version upgrading. But the analyzer can be added even after the v8.0 release because it will cause compile time errors only on newly added parsers and validators but not on existing ones.

Shane32: And you can always silence an analyzer warning/error.

Shane32: Schema initialization checks have been added.

gao-artur (approved):

PR #3861, opened 2023-12-29
Use markup key to indicate expected diagnostics location

gao-artur:


Shane32 (approved):

Shane32:
FYI, the only time I use the `test` label is when the PR _only_ applies to tests.  Since typically any programming change should add/change tests, so I don't label those with the `test` label.

Shane32:
This PR looks good!

PR #3864, opened 2023-12-30
Change solution to use file-scoped namespaces

Shane32:
We should probably only merge this if/when we are sure we are not making additional changes to v7, so as to minimize merge conflicts.  Now?

All changes in this PR were done automatically by VS.  No other changes were done, besides a single change to `.editorconfig` to require file-scoped namespaces on all files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -194,7 +194,7 @@ csharp_preserve_single_line_blocks = true
 csharp_preserve_single_line_statements = false
 
 # C# formatting settings - Namespace options
-csharp_style_namespace_declarations = file_scoped:suggestion
+csharp_style_namespace_declarations = file_scoped:error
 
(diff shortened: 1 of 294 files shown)


.editorconfig
```diff
@@ -194,7 +194,7 @@ csharp_preserve_single_line_blocks = true
 csharp_preserve_single_line_statements = false
 
 # C# formatting settings - Namespace options
-csharp_style_namespace_declarations = file_scoped:suggestion
+csharp_style_namespace_declarations = file_scoped:error
```
Shane32, line 197: Only change besides a single command in VS to fix all IDE0161 (file-scoped namespace suggestion) in the solution.

gao-artur:
I'm happy you are merging this. It means v8 with Federation 2 is closer than ever 😄

Issue #3823, opened 2023-12-14
Ensure input object fields default don't conflict with source members default


gao-artur:
In the sample you provided, it would be more likely that the field would be required and have a default value of an empty list.

Then, to avoid confusion, we must have an analyzer to ensure "legal" defaults are applied on graph and source types. Please correct me if this table is wrong:
|Graph type field|Source type member default|
|-|-|
| Nullable without default | Nothing or null |
| Nullable with default | Nothing or same as on graph |
| Not nullable without default | Nothing |
| Not nullable with default | Nothing or same as on graph |

If we create such an analyzer, then we can implement this note:

> We could set ALL undefined properties to their default values, not just init-only ones, making reflection, source-generated and dynamically-compiled versions simpler.

Because there won't be runtime surprises for the developer.

_Originally posted by @gao-artur in https://github.com/graphql-dotnet/graphql-dotnet/issues/3804#issuecomment-1853750299_

Shane32:
A simpler alternative here might just be to have the analyzer:
1. Locate graph input fields on CLR types, either by matching by name or lambda
2. If they (a) exist in the constructor, or (b) are init-only or required and have a default value, display a warning that the value will be overwritten when etc etc
3. If they are read-only (fields) or have no public setter (properties), and there is no corresponding value in the constructor, display an error that the value cannot be written to.  (separate rule?)

But of course more analysis would not hurt either 😄 

Keep in mind that this rule should apply to all properties of CLR types consumed by `AutoRegisteringInputObjectGraphType<T>` also.  (It does not by default auto-generate graph fields for CLR fields.)

gao-artur:
> 3. If they are read-only (fields) or have no public setter (properties), and there is no corresponding value in the constructor, display an error that the value cannot be written to. (separate rule?)

Already exists: [GQL007](https://graphql-dotnet.github.io/docs/analyzers/gql007/). Need to extend it to `AutoRegisteringInputObjectGraphType`.

FYI, I can only produce diagnostics on the graph fields but not the source type members because there is no guarantee that the source type's source code is available or located in the project where the analyzer is referenced. In the `AutoRegisteringInputObjectGraphType<T>` case, all the warnings will be reported on the `<T>`.

gao-artur:
TODO for me: fix the `it's doesn't` message as part of this story.

[...]

PR #3865, opened 2023-12-30
Extend GQL010 rule for `AutoRegisteringInputObjectGraphType`

gao-artur:
Partially fixes https://github.com/graphql-dotnet/graphql-dotnet/issues/3823


src/GraphQL.Analyzers.Tests/InputGraphTypeAnalyzerTests.Constructors.cs
```diff
@@ -31,6 +34,18 @@ public CustomInputObjectGraphType()
...
+              }
+
+              public class MyGraphType : ObjectGraphType
+              {
+                  public MyGraphType()
+                  {
+                      Field<AutoRegisteringInputObjectGraphType<{|#2:MySource|}>>();
+                  }
+              }
+
```
gao-artur, line 48: I extended one test to include a type derived from `AutoRegisteringInputObjectGraphType` and usage of `AutoRegisteringInputObjectGraphType`. The code to analyze these cases is the same as before. The only difference is the diagnostic location.

Shane32 (approved):

Issue #3680, opened 2023-07-31
DataLoader & scoped dependencies concurrent access with parallel execution strategy


br3aker:
## Description

DataLoader does not work as expected with scoped dependencies. [Documentation provides rather misleading example](ScopedDependency) when scoped dependency can be accessed from the `DataLoader`, i.e. database access with Entity Framework.

Current server implementation can guarantee single thread access if and only if:
1. `loaderKey` is set, i.e. only one dataloader exists for such access
2. `maxBatchSize` property is less or equal to dataloader  'load requests', i.e. only one dataloader exists for such access

`.Resolve().WithScope()` can't help here as their scope is created for entire node execution time, problem is that `IDataLoaderResult` are resolved **after** regular node execution -> [scope is destroyed](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.MicrosoftDI/ScopedFieldResolver.cs#L23) -> `ObjectDisposedException` or any other problems with dependencies from disposed scope.

We also can't use 'root' scope for entire HTTP call, [`IDataLoaderResult` are resolved asynchronously](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Execution/ExecutionStrategy.cs#L515), this call [is not awaited at the execution strategy level](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Execution/ParallelExecutionStrategy.cs#L95) -> concurrent access exception from Entity Framework `DbContext`, for example.

The more of a hack than a solution that comes to mind is manually creating scope before the actual resolver execution and disposing it in the dataloader `Then` chained call. Something like this:

<details>
  <summary>Simple schema for reproductino</summary>

```cs
public static class FieldBuilderExtensions
{
    public static FieldBuilder<TSourceType, IDataLoaderResult<TReturnType>> ResolveWithDataLoader<TSourceType, TReturnType, T1>(
        this FieldBuilder<TSourceType, IDataLoaderResult<TReturnType>> builder,
        Func<IResolveFieldContext<TSourceType>, DataLoaderContext, T1, IDataLoaderResult<TReturnType>> func)
    {
        Func<IResolveFieldContext<TSourceType>, IDataLoaderResult<TReturnType>> resolver = context =>
        {
            IServiceScope? scope = default;
            try
            {
                scope = context.RequestServices.CreateScope();
                var dataLoader = scope.ServiceProvider
[...]

PR #3866, opened 2023-12-31
[WIP] Add DI data loader extension methods

Shane32:
See:
- #3680

### DIFF
--- a/src/GraphQL.ApiTests/GraphQL.DataLoader.approved.txt
+++ b/src/GraphQL.ApiTests/GraphQL.DataLoader.approved.txt
@@ -1,3 +1,16 @@
+namespace GraphQL
+{
+    public static class ConfigureDataLoaderExtensions
+    {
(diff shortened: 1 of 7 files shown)


Shane32:
Todo: update documentation

Shane32:
I think this concept has merit, specifically the `AddBatchDataLoader` method, but the `AddCollectionBatchDataLoader` method doesn't make sense, as there may be multiple data loaders needed for the same returned type.  For instance:

```gql
# assume IDs are ints internally

type Author {
  id: ID!
  name: String!
  books: [Book!]!   # needs a collection batch loader based on the author id
}

type Category {
  id: ID!
  name: String!
  books: [Book!]!   # needs a collection batch loader based on the category id
}

type Book {
  id: ID!
  title: String!
  author: Author!
  category: Category!
}
```

With this PR as it is now, both calls to `AddCollectionBatchLoader` would each register as `IDataLoader<int, IEnumerable<Book>>` and so it would not be possible to use DI to select the proper data loader.

PR #3867, opened 2024-01-01
Schema builder fixes; add schema validation of subscription fields

### LINKED ISSUE #3580: Ensure that StreamResolver IS set for the Schema.Subscription graph type
See #3574

Shane32:
Fixes #3580 

Adds schema validation to ensure that `StreamResolver` is set for fields of root subscription types.  This is a non-breaking change because without the validation, attempting to resolve these fields would fail at runtime.

Trying to write a fix for #3580 I found two bugs in the schema-first SchemaBuilder (as far as I can tell).
- `Include` does not include configurations via `ForAll`
- Subscriptions don't work properly when the `schema` definition is listed after the subscription definition

### DIFF
--- a/src/GraphQL/Utilities/SchemaBuilder.cs
+++ b/src/GraphQL/Utilities/SchemaBuilder.cs
@@ -111,6 +111,7 @@ Schema contains a redefinition of these types: {string.Join(", ", duplicates.Sel
 
             var directives = new List<Directive>();
 
+            // process the schema definition first
             foreach (var def in document.Definitions)
             {
                 if (def is GraphQLSchemaDefinition schemaDef)
@@ -118,7 +119,11 @@ Schema contains a redefinition of these types: {string.Join(", ", duplicates.Sel
                     _schemaDef = schemaDef;
                     schema.SetAstType(schemaDef);
                 }
-                else if (def is GraphQLObjectTypeDefinition objDef)
+            }
+
+            foreach (var def in document.Definitions)
+            {
+                if (def is GraphQLObjectTypeDefinition objDef)
                 {
                     var type = ToObjectGraphType(objDef);
                     _types[type.Name] = type;
--- a/src/GraphQL/Utilities/TypeSettings.cs
+++ b/src/GraphQL/Utilities/TypeSettings.cs
@@ -74,7 +74,7 @@ namespace GraphQL.Utilities
         /// </summary>
         public void Include(string name, Type type)
         {
-            _typeConfigurations[name].Type = type;
+            For(name).Type = type;
         }
 
         /// <summary>
@@ -113,7 +113,7 @@ namespace GraphQL.Utilities
         /// </summary>
         public void Include(string name, Type type, Type typeOfType)
         {
-            var config = _typeConfigurations[name];
+            var config = For(name);
             config.Type = type;
             config.IsTypeOfFunc = obj => obj?.GetType().IsAssignableFrom(typeOfType) ?? false;
         }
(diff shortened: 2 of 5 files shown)


src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -111,14 +111,19 @@ private Schema BuildSchemaFrom(GraphQLDocument document)
 
             var directives = new List<Directive>();
 
+            // process the schema definition first
             foreach (var def in document.Definitions)
             {
                 if (def is GraphQLSchemaDefinition schemaDef)
                 {
                     _schemaDef = schemaDef;
```
Shane32, line 119: `_schemaDef` must be set first or setting subscription resolvers don't get set properly.  Who knows maybe more stuff.

src/GraphQL/Utilities/TypeSettings.cs
```diff
@@ -74,7 +74,7 @@ public void Include(Type type)
         /// </summary>
         public void Include(string name, Type type)
         {
-            _typeConfigurations[name].Type = type;
+            For(name).Type = type;
```
Shane32, line 77: `ForAll` only triggers when using `For`, but seems to clearly indicate that it should operate on all types.  So I've changed `Include` to call `For` and thereby execute any configured `ForAll` code.

src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
```diff
@@ -238,6 +238,14 @@ public override void VisitSchema(ISchema schema)
             var n3 = schema.Subscription?.Name;
             if (n1 == n2 && n1 != null || n1 == n3 && n1 != null || n2 == n3 && n2 != null)
                 throw new InvalidOperationException("The query, mutation, and subscription root types must all be different types if provided.");
+            if (schema.Subscription != null)
+            {
+                foreach (var field in schema.Subscription.Fields.List)
+                {
+                    if (field.StreamResolver == null)
+                        throw new InvalidOperationException($"The field '{field.Name}' of the subscription root type '{schema.Subscription.Name}' must have StreamResolver set. You should set StreamResolver only all root fields of subscriptions.");
```
gao-artur, line 246: Typo? Should it be `on` instead of `all`?

gao-artur: In general, fixing these errors one by one is pretty hard because it requires restarting the application multiple times. I think gathering all the errors and throwing single/aggregated exception after the full validation can provide a better experience. This is usually not a problem during development but is annoying during the version upgrade.

Shane32: Ok let's make that change for v8.  Do you want me to continue with this PR or just add this validation into v8?

Shane32:
```suggestion
                        throw new InvalidOperationException($"The field '{field.Name}' of the subscription root type '{schema.Subscription.Name}' must have StreamResolver set.");
```

Shane32: > Ok let's make that change for v8. Do you want me to continue with this PR or just add this validation into v8?

Well, it's not like the schema will work properly if this were to fail validation, so might as well leave it in here.

gao-artur: Throwing aggregated error is a job for separate PR

gao-artur (approved):

src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
Shane32: #3875

PR #3871, opened 2024-01-01
Mix errors from validation and listener

### LINKED ISSUE #3089: Mix errors from validation and listener?
Just found this code in `Documentexecuter`
```csharp
 context = BuildExecutionContext(options, document, operation, variables, metrics);

                foreach (var listener in options.Listeners)
                {
                    await listener.AfterValidationAsync(context, validationResult) // TODO: remove ExecutionContext or make different type ?
                        .ConfigureAwait(false);
                }

                if (!validationResult.IsValid)
                {
                    return new ExecutionResult
                    {
                        Errors = validationResult.Errors,
                        Perf = metrics.Finish()
                    };
                }

                if (context.Errors.Count > 0)
                {
                    return new ExecutionResult
                    {
                        Errors = context.Errors,
                        Perf = metrics.Finish()
                    };
                }

                executionOccurred = true;
```

Questions:
1. If there were validation errors, then is it worth calling `AfterValidationAsync` at all?
2. If yes then why do we have two independent return blocks for validation errors and context errors? They should be mixed.

Shane32:
Closes #3089 
- #3089

### DIFF
--- a/src/GraphQL/Execution/DocumentExecuter.cs
+++ b/src/GraphQL/Execution/DocumentExecuter.cs
@@ -146,23 +146,14 @@ namespace GraphQL
                         .ConfigureAwait(false);
                 }
 
-                if (!validationResult.IsValid)
+                if (!validationResult.IsValid || context.Errors.Count > 0)
                 {
-                    var errors = validationResult.Errors;
-                    if (context.Errors.Count > 0)
+                    var errors = !validationResult.IsValid ? validationResult.Errors : context.Errors;
+                    if (!validationResult.IsValid && context.Errors.Count > 0)
                         errors.AddRange(context.Errors);
                     return new ExecutionResult
                     {
                         Errors = errors,
-                        Perf = metrics.Finish()
-                    };
-                }
-
-                if (context.Errors.Count > 0)
-                {
-                    return new ExecutionResult
-                    {
-                        Errors = context.Errors,
                         Perf = metrics.Finish()
                     };
                 }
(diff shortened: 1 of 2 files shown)


src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -148,9 +148,12 @@ private async Task<ExecutionResult> CoreExecuteAsync(ExecutionOptions options)
 
                 if (!validationResult.IsValid)
```
gao-artur, line 149: Can it be `if (!validationResult.IsValid || context.Errors.Count > 0)` and the lines 161-168 removed?

Shane32: Was trying to reduce allocations, but i can still simplify it some

Shane32: Simplified; harder to understand now but less code

gao-artur (approved):

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -146,20 +146,14 @@ private async Task<ExecutionResult> CoreExecuteAsync(ExecutionOptions options)
                         .ConfigureAwait(false);
                 }
 
-                if (!validationResult.IsValid)
+                if (!validationResult.IsValid || context.Errors.Count > 0)
                 {
+                    var errors = !validationResult.IsValid ? validationResult.Errors : context.Errors;
+                    if (!validationResult.IsValid && context.Errors.Count > 0)
+                        errors.AddRange(context.Errors);
```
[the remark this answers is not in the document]
Shane32: Working with @sungam3r we always tried real hard to minimize allocations.  Perhaps overly so.  But it makes for a "better" codebase - depending on your viewpoint.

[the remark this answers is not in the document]
Shane32: This is an error path, not a hot path, so it really is a little overkill.

PR #3872, opened 2024-01-01
Remove IComplexityAnalyzer

Shane32:
`IComplexityAnalyzer` has been obsolete since 7.0.0, marked to be removed in v8.  (See #3179)  The changes here do not change functionality in the slightest.  (It still performs as poorly/wacky as before.)  Just removing the obsolete interface.  As with v7, functionality is still completely sealed within GraphQL.NET.  As with the recommendation for v7, users wishing different behavior for a complexity analyzer should create a new validation rule.

Note that while the changes within ComplexityValidationRule looks a little confusing, the methods are pretty much direct copies of the methods previously contained within ComplexityAnalyzer, with no changes.


src/GraphQL/Validation/Rules.Custom/ComplexityValidationRule.cs
```diff
@@ -1,84 +1,205 @@
...
+            {
+                context.ReportError(ex);
+            }
+        }
+
+        return default;
+    }
+
+    /// <inheritdoc/>
+    public void Validate(GraphQLDocument document, ISchema schema)
```
Shane32, line 56: The previous signature was:

```cs
public void Validate(GraphQLDocument document, ComplexityConfiguration complexityParameters, ISchema? schema = null)
```

But the schema is always passed from the validation rule, and the complexity parameters are pulled from the constructor.  So this signature was slightly updated to make more sense.

Shane32: Unfortunately this name "conflicts" with `ValidateAsync` from `IValidationRule.ValidateAsync`, being that they have completely different signatures.  I just left it as-is.

gao-artur (approved):

PR #3873, opened 2024-01-02
[WIP] Validation rules can read coerced argument values

Shane32:
This allows validation rules to inspect (or validate) argument values.  This is most specifically will be useful when designing a new complexity analyzer / rate limiter, which should be able to inspect connection types to determine how many records will be returned based on the passed 'first' or 'last' field arguments.  As a reference see [GitHub rate limits and node limits for the GraphQL API](https://docs.github.com/en/graphql/overview/rate-limits-and-node-limits-for-the-graphql-api) where the number of requested rows is a factor of the rate limiter.

Another way to implement this would be to add `bool IValidationRule.AfterVariableParsing` (or similar), but that would break `IValidationRule` implementations.  I would prefer to keep `IValidationRule` intact for v8, so compatibility is retained to the server project and other 3rd party code that uses the basic `IValidationRule` interface.  In this case the breaking change is to the lesser-used `IVariableVisitorProvider` interface.

We can also implement a default implementation of the new `ValidateArgumentsAsync` interface method.  This works for .NET Standard 2.1 and later; we would use conditional compilation so that for .NET Standard 2.0 targets, the method is required.  However, when writing a class definition, right-clicking on the interface and selecting 'implement all methods' does not implement the methods with default interface implementations.  So it may be more difficult for users to realize that the method exists.

The changed interface method definition would look like this:
```cs
        ValueTask<INodeVisitor?> ValidateArgumentsAsync(ValidationContext context)
#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP3_0_OR_GREATER
            => default
#endif
            ;
```


docs2/site/docs/migrations/migration8.md
```diff
@@ -354,3 +359,13 @@ You can also set global attributes that will be applied to all auto-registering
 if you wish to define your own naming logic.
 
 The `UseLegacyTypeNaming` option is deprecated and will be removed in GraphQL.NET v9.
+
+### 9. `IVariableVisitorProvider.ValidateArgumentAsync` added
```
Shane32, line 363:
```suggestion
### 9. `IVariableVisitorProvider.ValidateArgumentsAsync` added
```

docs2/site/docs/migrations/migration8.md
```diff
@@ -354,3 +359,13 @@ You can also set global attributes that will be applied to all auto-registering
 if you wish to define your own naming logic.
 
 The `UseLegacyTypeNaming` option is deprecated and will be removed in GraphQL.NET v9.
+
+### 9. `IVariableVisitorProvider.ValidateArgumentsAsync` added
+
+The `IVariableVisitorProvider` interface has a new method `ValidateArgumentAsync` that is
```
Shane32, line 365:
```suggestion
The `IVariableVisitorProvider` interface has a new method `ValidateArgumentsAsync` that is
```

Shane32:
@gao-artur Let me know when you have a chance to review.  As powerful as `Validator` and `Parser` are, this PR is not as beneficial as it would be without it.  However, for a complexity analzyer / rate limiter, this could be very useful.  This is my last personal goal here in v8.  We could also consider revamping `IVariableVisitor` to basically perform the same hooks as `Validator` and `Parser` do now.  But it seems better to use a schema visitor to tie these hooks directly into the field definition, as shown in #3885 so there are no lookups during coercion.  So I'm leaving `IVariableVisitor` alone for now.  In fact, maybe we should deprecate it entirely.

gao-artur:
I tried to read it multiple times but understood nothing. Maybe it's a late hour. Will try again tomorrow.

Shane32:
> I tried to read it multiple times but understood nothing. Maybe it's a late hour. Will try again tomorrow.

No problem.  I'd suggest only reading the changes to the migration document.  Be sure to comment on anything that you don't understand and I'll rewrite it.  It's really great to have a second set of eyes to give another perspective on the documentation from someone who didn't write the code.

Once you understand the goals outlined in the migration doc, we can move to the code, but if I failed to make the migration doc understandable, then I gotta fix that first!

Thanks!!

docs2/site/docs/migrations/migration8.md
```diff
@@ -246,6 +246,80 @@ At this time GraphQL.NET does not directly support the `MaxLength` and similar a
 implement your own attributes as shown above, or call the `Validate` method to set a validation
 function.
 
+### 6. Validation rules can read or validate field arguments and directive arguments
+
+Validation rules can now read or validate field arguments and directive arguments. This is useful
+for edge cases, such as when a complexity analyzer needs to read the value of a field argument
+to determine the complexity of the field. While it could also be used within custom validation rules
+to validate the value of field arguments or directive arguments, perhaps based on schema directives,
+it cannot be used to validate fields of input object graph types.
```
gao-artur, line 255: Some parts of the text look like a formal math definition. I had to slow down and read it again. For example, this is too verbose
> Validation rules can now read or validate field arguments and directive arguments.

IMO, this is easier to read
> Validation rules can now read or validate field and directive arguments.

And this happens in many places along the text

```suggestion
Validation rules can now read or validate field and directive arguments. This is useful
for edge cases, such as when a complexity analyzer needs to read the value of a field argument
to determine the complexity of the field. While it could also be used within custom validation rules
to validate the value of field or directive arguments, perhaps based on schema directives,
it cannot be used to validate fields of input objects.
```

docs2/site/docs/migrations/migration8.md
```diff
@@ -246,6 +246,80 @@ At this time GraphQL.NET does not directly support the `MaxLength` and similar a
...
+
+Note that with the addition of `Validator` and `Parser` delegates, validation rules are no longer
+required to validate input values. If you plan to store validation rules within schema directives,
+such that they may be read by clients (see `ExperimentalFeatures.AppliedDirectives`), you may wish
+to consider adding a schema initialization visitor to walk the schema, applying the validation rules
+to the appropriate fields arguments via the `Validate` method, rather than using a validation rule
+to validate the field arguments. This will be much more efficient, as the validation rule will not
+be called for every field argument on every field, nor will it require a dictionary lookup to find
+the validation rule for the field argument.
```
gao-artur, line 264: Add a note that both approach example is provided below, so the reader won't waste time trying to understand all the details.

docs2/site/docs/migrations/migration8.md
```diff
@@ -246,6 +246,80 @@ At this time GraphQL.NET does not directly support the `MaxLength` and similar a
...
+
+To implement a validation rule that validates field arguments or directive arguments, implement
+`IValidationRule` and `IVariableVisitorProvider`. The `IVariableVisitorProvider` interface has a
+new method `ValidateArgumentsAsync` that is called during validation. This method is called after
+all field arguments and directive arguments have been parsed, allowing a validation rule to validate
+parsed argument values. It will run even if there are no arguments to validate, so you should check
+for this case and return early if there are no arguments to validate. However it will not run if
+errors occur prior to parsing the arguments, such as if the field or directive is not found, or
+if an argument fails coercion or validation.
```
gao-artur, line 273:
```suggestion
To implement a validation rule that validates field or directive arguments, implement
`IValidationRule` and `IVariableVisitorProvider`. The `IVariableVisitorProvider` interface has a
new method, `ValidateArgumentsAsync`, that is called during validation. This method is called after
all the relevant arguments have been parsed, allowing a validation rule to validate
parsed values. It will run even if there are no arguments to validate, so you should check
for this case and return early if there is nothing to validate. However, it will not run if
errors occur prior to parsing the arguments, such as if the field or directive is not found, or
if an argument fails coercion or validation.
```

docs2/site/docs/migrations/migration8.md
```diff
@@ -246,6 +246,80 @@ At this time GraphQL.NET does not directly support the `MaxLength` and similar a
...
+        return default;
+    }
+
+    ValueTask INodeVisitor.LeaveAsync(ASTNode node, ValidationContext context) => default;
+}
+```
+
+The above is sample functionality, although in this situation it would be more efficient to
+have a schema visitor apply this validation rule to the appropriate fields via the `Validate`
+method during schema initialization.
```
gao-artur, line 321: Can you add an example here to clarify the difference between the two approaches?

docs2/site/docs/migrations/migration8.md
```diff
@@ -369,3 +443,13 @@ Please see the v7 migration document regarding the new `schema.ToAST()` and
 For federated schemas, the `ServiceGraphType`'s `sdl` field will now use the
 new implementation to print the schema. Please raise an issue if this causes
 a problem for your federated schema.
+
+### 11. `IVariableVisitorProvider.ValidateArgumentsAsync` added
+
+The `IVariableVisitorProvider` interface has a new method `ValidateArgumentsAsync` that is
+called during validation. This method is called after all field arguments and directive
+arguments have been parsed, allowing a validation rule to validate parsed argument values.
```
gao-artur, line 451:
```suggestion
The `IVariableVisitorProvider` interface has a new method, `ValidateArgumentsAsync`, that is
called during validation. This method is called after all field and directive
arguments have been parsed, allowing a validation rule to validate parsed argument values.
```

gao-artur (commented):
After a good sleep it was easier to understand 🙂

gao-artur:
> We can also implement a default implementation of the new `ValidateArgumentsAsync` interface method.

I wouldn't do that. The default interface implementation is a hack for legacy systems where updating all the implementations requires too much effort. I don't think it's a good practice to do that.

Shane32:
Good comments.  I'll work on it

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
```diff
@@ -29,6 +29,10 @@ public override void VisitInputObjectFieldDefinition(FieldType field, IInputObje
     public override void VisitObjectFieldArgumentDefinition(QueryArgument argument, FieldType field, IObjectGraphType type, ISchema schema)
         => argument.Validator += GetValidator(argument);
 
+    /// <inheritdoc/>
+    public override void VisitInterfaceFieldArgumentDefinition(QueryArgument argument, FieldType field, IInterfaceGraphType type, ISchema schema)
+        => field.Validator += GetValidator(argument);
```
Shane32, line 34: Bug fix

docs2/site/docs/migrations/migration8.md
Shane32: Done

docs2/site/docs/migrations/migration8.md
Shane32: Done

src/GraphQL/Validation/IVariableVisitorProvider.cs
```diff
@@ -9,6 +9,13 @@ public interface IVariableVisitorProvider
...
         /// </summary>
-        IVariableVisitor GetVisitor(ValidationContext context);
+        IVariableVisitor? GetVisitor(ValidationContext context);
+
+        /// <summary>
+        /// Prepares and returns a node visitor to be used to validate a document (via a node walker) against this
+        /// validation rule after the document has parsed all arguments. Validation failures are added then by this
+        /// visitor to a list stored within <see cref="ValidationContext.Errors"/>.
+        /// </summary>
+        ValueTask<INodeVisitor?> ValidateArgumentsAsync(ValidationContext context);
```
gao-artur, line 19: The method name is a bit confusing. The existing `GetVisitor` fits well for the `IVariableVisitorProvider`, but `ValidateArgumentsAsync` looks inappropriate. I don't have a better suggestion, though. Maybe it should be placed in a new interface? Also, it's called `Validate` but returns visitor, which is not intuitive at all.

Shane32: What about if we add a whole new interface for validation rules?  We can then re-think and consolidate the entire design without breaking compatibility with rules designed for v7.

gao-artur: > What about if we add a whole new interface for validation rules?

Maybe. I need to dive into the validation code to understand how it's implemented today.

Shane32: Sure.  I'd appreciate it if you could help me pick the best approach.  FYI this is what happens now (effectively):

1. `DocumentExecuter.ExecuteAsync(ExecutionOptions options)` runs with two relevant options:
  a. `IEnumerable<IValidationRule>? ValidationRules`
  b. `IEnumerable<IValidationRule>? CachedDocumentValidationRules`
2. If `ValidationRules` is null, then `ValidationRules = CoreRules` which contains the default set of rules
3. Adding rules via `AddValidationRule<T>(bool useForCachedDocuments = false)` will append to the proper property
4. If using document caching, and a cache hit occurs, `ValidationRules = CachedDocumentValidationRules`
5. `CachedDocumentValidationRules` is not used after this point
6. The document is parsed (if not using a cached instance)
7. Each instance of `IValidationRule.ValidateAsync` is executed.  Any non-null returned `INodeVisitor` instances are added to a list.  It is possible for validation to occur here and not use a node visitor.  But typically this method just returns a node visitor, often a new instance to hold temporary information during the validation.
8. Assuming there are any `INodeVisitor` instances, then `TypeInfo` is prepended to the list and the node visitor executes.  Each node is walked, and for each node, all the configured node visitors run for that node.  So the validation rules all run 'concurrently' rather than one after another.  The `TypeInfo` populates `ValidationContext` with information that assists validation rules to run efficiently.
9. If there are any errors, they are returned and execution stops
10. For each `IValidationRule`, `if (rule is IVariableVisitorProvider)` then add returned `IVariableVisitor` to a list of variable visitors (if not null)
11. Input values are parsed (and `Parser` and `Validator` delegates run if configured).  Any configured variable visitors are called at the appropriate time.  Unfortunately this only occurs with variables and not literals.  Ideally, I would like to change this to support both ... or deprecate this feature entirely in lieu of `Parser` and `Validator` delegates.
12. If any errors have occurred, they are returned and execution stops
13. NEW FEATURE:
  a. For each `IValidationRule`, `if (rule is IVariableVisitorProvider)`, call `ValidateArgumentsAsync` and add the returned `INodeVisitor` to a list (if not null)
  b. If any visitors, `TypeInfo` is prepended and these visitors execute in the same fashion as the first pass of validation rules.  The only difference is that it occurs after input value parsing has occurred.
14. If any errors have occurred, they are returned and execution stops
15. Document is executed

Shane32: Ok I’ll come up with an alternative solution and present it for review

Shane32:
@gao-artur I found that changes to `IMetadataProvider` are going to prevent GraphQL.Server from working with GraphQL.NET 8 anyway.  As such I was thinking we just forget trying to retain the `IValidationRule` signature, with all of the other validation changes that are being made.

So what about this for `IValidationRule` in v8 ?

```cs
    public interface IValidationRule
    {
        // gets the node visitor to use before field arguments and variables are parsed
        ValueTask<INodeVisitor?> GetPreNodeVisitorAsync(ValidationContext context);
        // gets the visitor to validate variables as they are being parsed
        ValueTask<IVariableVisitor?> GetVariableVisitorAsync(ValidationContext context);
        // gets the node visitor to use after field arguments and variables are parsed
        ValueTask<INodeVisitor?> GetPostNodeVisitorAsync(ValidationContext context);
    }
```

 Feel free to suggest changes to these names.

gao-artur:
> So what about this for `IValidationRule` in v8?

Looks like this is going to replace the existing `IValidationRule` and also `IVariableVisitorProvider`, right? Assuming that most of the users will rarely need all 3 methods, it may be worth creating an abstract `ValidationRule` class with all methods returning `null` by default.
On the other hand, a rule that needs more than 1 of these methods should probably be two independent rules. Do you have an example of the validation that will want to implement at least two methods?

Shane32:
In this PR I added ValidationRuleBase which is similar to DocumentListenerBase and how you described where it returns default unless overridden

Shane32:
I think it’s rare but possible. The existing sample of max length for string arguments use two of the three.

Shane32:
I think having a single interface for validation rules is clearer and easier to configure with the GraphQLBuilder. This also allows the most flexibility so basically any validation rule can hook in anywhere needed in the pipeline. I consider writing custom validation rules an advanced scenario.

Shane32:
> In this PR I added ValidationRuleBase which is similar to DocumentListenerBase and how you described where it returns default unless overridden

I mean in #3905 sorry. The revised proposal.

Shane32:
Replaced by #3905

PR #3878, opened 2024-01-02
AllowedOnAttribute to force correct FieldBuilder methods usage

### LINKED ISSUE #3860: Analyzer to ensure Parser and Validator delegates are only set on input fields
Complementary analyzer for PR #3857

gao-artur:
Fixes #3860
- https://github.com/graphql-dotnet/graphql-dotnet/issues/3860

@Shane32, before I proceed with analyzer implementation and tests, can you please verify the methods are correctly annotated?
For summary:
- `Resolve*` methods are only allowed on `IObjectGraphType`
- `Argument` on `IObjectGraphType` and `IInterfaceGraphType`
- `ParseValue` and `Validate` on `IInputObjectGraphType`


src/GraphQL.MicrosoftDI/DynamicScopedFieldResolver.cs
```diff
@@ -26,5 +27,6 @@ public DynamicScopedFieldResolver(IFieldResolver resolver)
     }
 
     /// <inheritdoc/>
+    [AllowedOn<IObjectGraphType>]
```
Shane32, line 30: This is just a `IFieldResolver` implementation, so I don't think we need to set it here, right?

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
```diff
@@ -39,6 +40,7 @@ public DynamicScopedSourceStreamResolver(ISourceStreamResolver resolver)
     }
 
     /// <inheritdoc/>
+    [AllowedOn<IObjectGraphType>]
```
Shane32, line 43: or here

Shane32:
> Fixes #3860
> 
> @Shane32, before I proceed with analyzer implementation and tests, can you please verify the methods are correctly annotated? For summary:
> 
> * `Resolve*` methods are only allowed on `IObjectGraphType`
> * `Argument` on `IObjectGraphType` and `IInterfaceGraphType`
> * `ParseValue` and `Validate` on `IInputObjectGraphType`

I think that's right.

Would annotating the FieldType.Resolver, FieldType.StreamResolver, FieldType.Parser, and FieldType.Validator properties have any effect?

src/GraphQL.MicrosoftDI/DynamicScopedFieldResolver.cs
gao-artur: Yeah, you are right

gao-artur:
> have any effect?

Do you mean should we annotate them? Probably we do. If using methods is restricted, there is no reason to use the property too.

Shane32:
> > have any effect?
> 
> Do you mean should we annotate them? Probably we do. If using methods is restricted, there is no reason to use the property too.

👍

src/GraphQL.Analyzers/AllowedOnAnalyzer.cs
```diff
@@ -122,11 +118,28 @@ private void AnalyzeMemberAccessExpression(SyntaxNodeAnalysisContext context)
...
         MemberAccessExpressionSyntax memberAccessExpressionSyntax,
-        DiagnosticDescriptor diagnosticDescriptor)
+        DiagnosticDescriptor diagnosticDescriptor,
+        ImmutableArray<ITypeSymbol> allowedTypes)
     {
+        string typesString = allowedTypes.Length switch
+        {
+            1 => allowedTypes[0].Name,
+            2 => $"{allowedTypes[0].Name} or {allowedTypes[1].Name}",
+            _ => $"{string.Join(", ", allowedTypes.Take(allowedTypes.Length - 1).Select(t => t.Name))} or {allowedTypes[^1].Name}"
```
gao-artur, line 128: We currently support up to 2 allowed types, but if more is needed, the analyzer is ready. Just define a new attribute with more generic arguments.

src/GraphQL.Analyzers/NotAGraphTypeAnalyzer.cs
```diff
@@ -85,10 +84,10 @@ private void AnalyzeGenericName(SyntaxNodeAnalysisContext context)
 
         bool HasNotAGraphTypeAttribute(ISymbol typeParam)
         {
-            notAGraphTypeAttribute ??= context.Compilation.GetTypeByMetadataName(Constants.MetadataNames.NotAGraphTypeAttribute);
             return typeParam.GetAttributes()
-                .Any(data => SymbolEqualityComparer.Default.Equals(data.AttributeClass, notAGraphTypeAttribute));
+                .Any(data => data.AttributeClass?.MetadataName == Constants.MetadataNames.NotAGraphTypeAttribute);
```
gao-artur, line 88: `GetTypeByMetadataName` returns `null` when the same name is defined in multiple referenced assemblies. Since code analysis attributes are now added as linked sources to multiple projects, this approach doesn't work anymore.
There is another method `GetAllTypesByMetadataName` that will return all the types with matching metadata name but it's not available in the used version.

src/GraphQL/CodeAnalysis/AllowedOnAttribute.cs
```diff
@@ -0,0 +1,15 @@
+using GraphQL.Types;
+
+namespace GraphQL;
+
+/// <summary>
+/// Marker attribute for analyzer usage to specify
+/// graph types where the annotated method can be used.
+/// The generic argument should be an interface.
```
gao-artur, line 8: Technically, this limitation can be removed. It will be a little bit more expensive to iterate base classes.

gao-artur:
> If using methods is restricted, there is no reason to use the property too.

I decided to remove the attributes from the properties for now. The analyzer will only work when the property is accessed while enclosed in the restricted graph type. Something like this:

```c#
public class DroidType : ObjectGraphType<Droid>
{
    public DroidType(StarWarsData data)
    {
        Field<StringGraphType>("name").FieldType.Validator = ...
    }
}
```

I don't think it's likely someone will write such a code.

gao-artur:
I messed with merge...

src/GraphQL.Analyzers/AllowedOnAnalyzer.cs
```diff
@@ -8,37 +8,21 @@
...
     // Violation: any of _forbiddenMethodNames used on type that doesn't implement _allowedInterfaces
     // Fixed:     remove the illegal method
-    public static readonly DiagnosticDescriptor IllegalResolverUsage = new(
-        id: DiagnosticIds.ILLEGAL_RESOLVER_USAGE,
-        title: "Illegal resolver usage",
-        messageFormat: "Resolvers are not allowed on non-output graph types",
+    public static readonly DiagnosticDescriptor IllegalMethodUsage = new(
+        id: DiagnosticIds.ILLEGAL_METHOD_USAGE,
+        title: "Illegal method usage",
+        messageFormat: "'{0}' invocation is only allowed on types implementing {1}",
```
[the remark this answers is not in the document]
Shane32: Seems fine to me.

src/GraphQL/CodeAnalysis/AllowedOnAttribute.cs
Shane32: Seems fine to me.  This is only an internal attribute so it only needs to support the types we plan to use it with.

src/GraphQL.Analyzers/NotAGraphTypeAnalyzer.cs
Shane32: I've not read through the code, so I'm not sure exactly how this affects usage.

docs2/site/docs/analyzers/gql005.md
```diff
@@ -7,6 +7,9 @@
 | **Default severity**   | Error  |
 | **Enabled by default** | Yes    |
 | **Code fix provided**  | Yes    |
+| **Removed in**         | v8.0   |
```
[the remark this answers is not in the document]
Shane32: Your call.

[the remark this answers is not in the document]
Shane32: If this analyzer rule has been removed, do we really want it listed in the docs at all?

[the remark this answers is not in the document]
gao-artur: Will do in separate PR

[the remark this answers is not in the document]
Shane32: If we do want it, so the link exists from older versions, maybe we remove it from the site map?

docs2/site/sitemap.yml
```diff
@@ -114,6 +114,8 @@
             file: gql010.md
           - title: GQL011
             file: gql011.md
+          - title: GQL012
+            file: gql012.md
```
Shane32, line 118: Yeah I think i'd just remove GQL005 from the site map if it is being removed in this version.  This is in the develop branch so the docs won't publish btw.

Shane32 (approved):

docs2/site/sitemap.yml
gao-artur: But if someone will use an older version after v8 release, he won't be able to navigate to the GQL005 help link. I'd prefer to keep the old documentation.

Shane32: Fair enough.

Issue #3881, opened 2024-01-04
How add the UnhandledExceptionDelegate in the startup?


julienFlexsoft:
I am using version 7.2.2.

Right now I am using a custom GraphQLMiddleware. It looks like this:
```csharp
var result = await _executer.ExecuteAsync(options =>
{
    options.Schema = schema;
    options.Query = graphQLRequest.Query;
    options.OperationName = graphQLRequest.OperationName;
    options.Variables = graphQLRequest.Variables;
    options.UserContext = GenerateGraphQLUserContext(context);
    options.User = context.User;
    options.EnableMetrics = enableMetrics;
    options.RequestServices = context.RequestServices;
    options.CancellationToken = context.RequestAborted;
    options.Listeners.Add(dataLoaderDocumentListener);
    options.ThrowOnUnhandledException = false;
    options.ValidationRules = context.RequestServices.GetServices<IValidationRule>().Concat(DocumentValidator.CoreRules);
    options.UnhandledExceptionDelegate = ctx =>
    {
        return HandleErrors(context, ctx, graphQLRequest, requestContext, envHelper, logger);
    };
});
private GraphQLUserContext GenerateGraphQLUserContext(HttpContext context)
{
    GraphQLUserContext result = new()
    {
        { "User", context.User },
        { "CustomKey", "" },
        { "CustomKey1", "" }
    };
    result.User = context.User;
    return result;
}
private Task HandleErrors(HttpContext context, UnhandledExceptionContext exceptionCtx, GraphQLRequest graphQLRequest,
    IRequestContext requestContext, IEnvHelper env, ILogger<GraphQLMiddleware> logger)
{
     // Custom logic here (ex: special logging, metrics)
}
```

and I am changing it to this:
```csharp
services.AddGraphQL(x => x
    .AddUserContextBuilder(context => {
        GraphQLUserContext result = new()
        {
            { "User", context.User },
            { "CustomKey", "" },
            { "CustomKey1", "" }
        };
        result.User = context.User;
        return result;
    })
);
```
But I am wondering how I can get my UnhandledExceptionDelegate logic in there?
I am switching away from the middleware because someone here told me that this was the preferred way to initialize GraphQL, mainly because it will require less work when updating later. Is this still accurate?
For example, the options.ValidationRules line caused some issues when updating GraphQL last time, which wouldn't have been an issue if I used the startup method.

Shane32:
```cs
[...]

PR #3882, opened 2024-01-04
Add 'AddUnhandledExceptionHandler' builder extension method

Shane32:
See:
- #3881

### DIFF
--- a/docs2/site/docs/getting-started/dependency-injection.md
+++ b/docs2/site/docs/getting-started/dependency-injection.md
@@ -89,6 +89,7 @@ A list of the available extension methods is below:
 | `AddSelfActivatingSchema<>` | Registers the specified schema which will create instances of unregistered graph types during initialization | |
 | `AddSerializer<>`       | Registers the specified serializer | |
 | `AddSystemTextJson`     | Registers the serializer that uses System.Text.Json as its underlying JSON serialization engine | GraphQL.SystemTextJson |
+| `AddUnhandledExceptionHandler` | Configures the unhandler exception handler | |
 | `AddValidationRule<>`   | Registers the specified validation rule and configures it to be used at runtime | |
 | `ConfigureExecution`    | Configures execution middleware to monitor or modify both options and the result | |
 | `ConfigureExecutionOptions` | Configures execution options at runtime | |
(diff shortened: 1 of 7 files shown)


docs2/site/docs/getting-started/dependency-injection.md
```diff
@@ -89,6 +89,7 @@ A list of the available extension methods is below:
 | `AddSelfActivatingSchema<>` | Registers the specified schema which will create instances of unregistered graph types during initialization | |
 | `AddSerializer<>`       | Registers the specified serializer | |
 | `AddSystemTextJson`     | Registers the serializer that uses System.Text.Json as its underlying JSON serialization engine | GraphQL.SystemTextJson |
+| `AddUnhandledExceptionHandler` | Configures the unhandler exception handler | |
```
Shane32, line 92: So far we use Add or Use consistently depending on if it's configuring execution options (order doesn't usually matter) vs wrapping the execution (order matters).

In this case the last call to `AddUnhandledExceptionHandler` is going to be the only effective one, just as the last call to `AddSerializer` is the only one effective.

I also called it `AddUnhandledExceptionHandler` vs `AddUnhandledExceptionDelegate` as this seemed more fitting.

src/GraphQL.Tests/DI/GraphQLBuilderExtensionTests.cs
```diff
@@ -898,6 +898,82 @@ public void ConfigureExecution_Null()
...
+        e2.ShouldBe(e);
+    }
+
+    [Fact]
+    public void AddUnhandledExceptionHandler_Null()
+    {
+        Should.Throw<ArgumentNullException>(() => _builder.AddUnhandledExceptionHandler((Action<UnhandledExceptionContext>)null!));
+        Should.Throw<ArgumentNullException>(() => _builder.AddUnhandledExceptionHandler((Action<UnhandledExceptionContext, ExecutionOptions>)null!));
+        Should.Throw<ArgumentNullException>(() => _builder.AddUnhandledExceptionHandler((Func<UnhandledExceptionContext, Task>)null!));
+        Should.Throw<ArgumentNullException>(() => _builder.AddUnhandledExceptionHandler((Func<UnhandledExceptionContext, ExecutionOptions, Task>)null!));
```
Shane32, line 973: The 4 overloads are shown here.  Basically, synchronous or asynchronous, with or without ExecutionOptions.

gao-artur (approved):

PR #3885, opened 2024-01-05
Add PatternMatchingDirective and PatternMatchingVisitor sample

Shane32:
This PR includes a new/better sample for directive-based validation, which may be useful.  Then I noticed that we were missing `DirectiveAttribute` for applying directives to type-first schemas.  So I added that in #3886 .

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -258,7 +258,11 @@ services.AddGraphQL(b => b
     .AddSchema<MyQuery>()
     .ConfigureSchema(s =>
     {
+        // add the directive to the schema
         s.Directives.Register(new PatternMatchingDirective());
+
+        // add the visitor to the schema, which will apply validation rules to all field
+        // arguments and input fields that have the @pattern directive applied
         s.RegisterVisitor(new PatternMatchingVisitor());
     }));
 ```
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -246,6 +246,44 @@ At this time GraphQL.NET does not directly support the `MaxLength` and similar a
 implement your own attributes as shown above, or call the `Validate` method to set a validation
 function.
 
+### 6. `@pattern` custom directive added for validating input values against a regular expression pattern
+
+This directive allows for specifying a regular expression pattern to validate the input value.
+It can also be used as sample code for designing new custom directives, and is now the preferred
+design over the older `InputFieldsAndArgumentsOfCorrectLength` validation rule.
+This directive is not enabled by default, and must be added to the schema as follows:
+
+```csharp
+services.AddGraphQL(b => b
+    .AddSchema<MyQuery>()
+    .ConfigureSchema(s =>
+    {
+        s.Directives.Register(new PatternMatchingDirective());
+        s.RegisterVisitor(new PatternMatchingVisitor());
+    }));
+```
+
+You can then apply the directive to any input field or argument as follows:
+
+```csharp
+Field(x => x.FirstName)
+    .ApplyDirective("pattern", "regex", "[A-Z]+"); // uppercase only
+```
+
+### 7. DirectiveAttribute added to support applying directives to type-first graph types and fields
+
+For example:
+
+```csharp
+private class Query
+{
+    public static string Hello(
+        [Directive("pattern", "regex", "[A-Z]+")] // uppercase only
+        string arg)
+        => arg;
+}
+```
+
 ## Breaking Changes
 
 ### 1. Query type is required
--- a/src/GraphQL/Attributes/DirectiveAttribute.cs
+++ b/src/GraphQL/Attributes/DirectiveAttribute.cs
@@ -1,0 +1,91 @@
+using GraphQL.Types;
+using GraphQL.Utilities;
+
+namespace GraphQL.Attributes;
+
+/// <summary>
+/// Applies a directive to part of a schema.
+/// </summary>
+[AttributeUsage(AttributeTargets.Enum | AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Struct | AttributeTargets.Property | AttributeTargets.Method | AttributeTargets.Field | AttributeTargets.Parameter, AllowMultiple = true)]
+public class DirectiveAttribute : GraphQLAttribute
+{
+    /// <inheritdoc cref="DirectiveAttribute"/>
+    public DirectiveAttribute(string name)
+    {
+        Name = name;
+    }
+
+    /// <inheritdoc cref="DirectiveAttribute"/>
+    public DirectiveAttribute(string name, string argumentName, object argumentValue)
+    {
+        Name = name;
+        Arguments.Add(argumentName, argumentValue);
+    }
+
+    /// <inheritdoc cref="DirectiveAttribute"/>
+    public DirectiveAttribute(string name, string argumentName1, object argumentValue1, string argumentName2, object argumentValue2)
+    {
+        Name = name;
+        Arguments.Add(argumentName1, argumentValue1);
+        Arguments.Add(argumentName2, argumentValue2);
+    }
+
+    /// <inheritdoc cref="DirectiveAttribute"/>
+    public DirectiveAttribute(string name, string argumentName1, object argumentValue1, string argumentName2, object argumentValue2, string argumentName3, object argumentValue3)
+    {
+        Name = name;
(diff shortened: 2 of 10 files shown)


src/GraphQL/GraphQL.csproj
```diff
@@ -21,9 +21,7 @@
   <ItemGroup>
     <!-- We don't consume the analyzers in this library itself,
          but we reference the project to add a package dependency so users of this library will automatically get the analyzers. -->
-    <ProjectReference Include="..\GraphQL.Analyzers.Package\GraphQL.Analyzers.Package.csproj"
-                      PrivateAssets="none"
-                      Condition="'$(ExcludeRestorePackageImports)' != ''"  />
+    <ProjectReference Include="..\GraphQL.Analyzers.Package\GraphQL.Analyzers.Package.csproj" PrivateAssets="none" Condition="'$(ExcludeRestorePackageImports)' != ''" />
```
Shane32, line 24: Must have auto-reformatted itself....

Shane32:
Waiting on #3886 first:
- #3886

src/GraphQL/Attributes/DirectiveAttribute.cs
```diff
@@ -0,0 +1,91 @@
+using GraphQL.Types;
+using GraphQL.Utilities;
+
+namespace GraphQL.Attributes;
+
+/// <summary>
+/// Applies a directive to part of a schema.
+/// </summary>
+[AttributeUsage(AttributeTargets.Enum | AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Struct | AttributeTargets.Property | AttributeTargets.Method | AttributeTargets.Field | AttributeTargets.Parameter, AllowMultiple = true)]
+public class DirectiveAttribute : GraphQLAttribute
```
Shane32, line 10: This file will not show as added once #3886 is merged in.

docs2/site/docs/migrations/migration8.md
```diff
@@ -246,6 +246,48 @@ At this time GraphQL.NET does not directly support the `MaxLength` and similar a
 implement your own attributes as shown above, or call the `Validate` method to set a validation
 function.
 
+### 6. `@pattern` custom directive added for validating input values against a regular expression pattern
+
+This directive allows for specifying a regular expression pattern to validate the input value.
+It can also be used as sample code for designing new custom directives, and is now the preferred
+design over the older `InputFieldsAndArgumentsOfCorrectLength` validation rule.
```
gao-artur, line 253: Why is this preferred over validation rules?

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
```diff
@@ -0,0 +1,53 @@
...
+///
+/// field.Field(x => x.Email)
+///     .ApplyDirective("pattern", "regex", @".+\@.+\..+");
+/// </code>
+/// </summary>
+public class PatternMatchingVisitor : BaseSchemaNodeVisitor
+{
+    /// <inheritdoc/>
+    public override void VisitInputObjectFieldDefinition(FieldType field, IInputObjectGraphType type, ISchema schema)
+        => field.Validator += GetValidator(field);
```
gao-artur, line 26: Oh, your new `Validator` feature gets useful. Nice!

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
```diff
@@ -0,0 +1,53 @@
...
+        // look for @pattern directive applied to the field argument or input field, and if found, use Validate to set the validation method
+        var applied = fieldArgOrInputField.FindAppliedDirective("pattern");
+        if (applied?.FindArgument("regex")?.Value is not string regex)
+            return null;
+
+        // if GlobalSwitches.DynamicallyCompileToObject is true, then compile
+        // the regex (or pull from the static cache if already compiled)
+        // (note that GlobalSwitches.DynamicallyCompileToObject is false for AOT schemas and scoped schemas)
+        var regexObject = new Regex($"^{regex}$",
+            GlobalSwitches.DynamicallyCompileToObject ? RegexOptions.Compiled : RegexOptions.None);
```
gao-artur, line 43: Is there any trick to make it work with [GeneratedRegexAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.generatedregexattribute?view=net-8.0) that will generate the regex implementation which will be used here?

docs2/site/docs/migrations/migration8.md
gao-artur: Found the answer here
https://github.com/graphql-dotnet/graphql-dotnet/pull/3873/files#diff-72602e5d1f15e15b9dcb3dc7964b37f622e56e31de537a3494941121d2e4abadR257

Shane32: Well, without #3873 , it's really difficult to read argument values.  All you have access to is the raw AST, before it has hit any of the configured scalars or input objects and been coerced to .NET types.  The existing sample is provided in `InputFieldsAndArgumentsOfCorrectLength`.

But let's say we implement #3873 which was my original concept for v8.  Then at least we can write validation rules much easier that are capable of inspecting coerced argument values for any field.  This in fact doesn't fully meet my original goal of universally validating any input value, but let's say we fix that (probably by extending `IVariableVisitor` to work for both literals and variables post-coercion).  Ok so now we can perform the same task in a validation rule with much less code.

So given all those assumptions, now let's compare to how a schema visitor would function.  The complexity of implementation is probably roughly the same now -- if you take a look at `PatternMatchingVisitor` it's pretty simple.  And I'm guessing that after changing `IVariableVisitor` it could get down to about that simple.

But what happens if we implement as a validation rule?  Every execution has to scan the entire incoming request's AST to match on each field argument and see if there are any directives applied.  Pulling the directives list is a dictionary lookup (we should think about eliminating that for v8), and then scanning the applied directives dictionary (if it exists) for an applied directive is another dictionary lookup.  If it finds a match, another dictionary lookup for the value of the argument(s).  When most input fields or field arguments have no attached rules, it seems like a lot of wasted CPU time.  In this regex sample, it would incur regenerating the regex (or pulling from the static dictionary), another costly operation.  Versus with a schema visitor, the exact code necessary to run the validation is built right into the coercion code (into `Validator`) along with the precompiled dependencies - in this case the compiled regex function.

So it becomes obvious that the schema visitor approach is much more optimized because there is no wasted CPU time on portions of the AST that have no validation rules, and any validation rule for any specific field can be 'precompiled' directly into the validation function with no dictionary lookups during runtime.

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
Shane32: I don't think so.  An optimized implementation for code-first would be like:

```cs
    [GeneratedRegex(
        @"^(ht|f)tp(s?)\:\/\/[0-9a-zA-Z]([-.\w]*[0-9a-zA-Z])*(:(0-9)*)*(\/?)([a-zA-Z0-9\-\.\?\,\'\/\\\+&amp;%\$#_]*)?$", 
        RegexOptions.CultureInvariant, 
        matchTimeoutMilliseconds: 1000)]
    private static partial Regex MatchIfValidUrl();

//type def
Field(x => x.Url)
    .Validate(value =>
    {
        if (!MatchIfValidUrl.IsMatch((string)value))
            throw new "Invalid URL";
    });
```

Unless we add a dedicated attribute for regex (this PR doesn't), and if somehow we can chain source builders so our source builder notices the attribute and links into the regex source builder to generate the code for that regex .....

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
Shane32: Yeah, it's getting me very excited.  I've long been wanting to have an easy way to validate values per-argument-definition before they reach the resolver.

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
gao-artur: Looks like no generic solution. Only manual validator configuration like in your example.

docs2/site/docs/migrations/migration8.md
gao-artur: Got it. The code snippet is very useful for understanding.

gao-artur (approved):

PR #3886, opened 2024-01-05
Add DirectiveAttribute

Shane32:
Samples and tests in PR #3885

This new directive mirrors the existing MetadataAttribute which allows adding arbitrary metadata to type-first schemas.

### DIFF
--- a/src/GraphQL/Attributes/DirectiveAttribute.cs
+++ b/src/GraphQL/Attributes/DirectiveAttribute.cs
@@ -1,0 +1,91 @@
+using GraphQL.Types;
+using GraphQL.Utilities;
+
+namespace GraphQL.Attributes;
+
+/// <summary>
+/// Applies a directive to part of a schema.
+/// </summary>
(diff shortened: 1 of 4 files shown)


src/GraphQL/Attributes/DirectiveAttribute.cs
```diff
@@ -0,0 +1,91 @@
...
+        Arguments.Add(argumentName2, argumentValue2);
+        Arguments.Add(argumentName3, argumentValue3);
+    }
+
+    /// <inheritdoc cref="DirectiveAttribute"/>
+    /// <remarks>
+    /// The <paramref name="argsAndValues"/> parameter must contain an even number of elements, where
+    /// the first element of each pair is the argument name and the second element is the argument value.
+    /// </remarks>
+    public DirectiveAttribute(string name, params object[] argsAndValues)
```
gao-artur, line 47: Is it worth adding an analyzer to ensure an even number of elements and the argument names are strings?

gao-artur (approved):

src/GraphQL/Attributes/DirectiveAttribute.cs
Shane32: Probably pretty low on the priority list.  Sure that would be nice, and it could check that the first element of each pair was a string too.  Keep people from doing stuff like `[Directive("maxlength", 5)]` when they meant to do `[Directive("maxlength", "value", 5)]`

PR #3905, opened 2024-01-27
Refactor validation rules

Shane32:
Replaces #3873 

This allows for post-variable-coercion validation, necessary for the next-gen complexity analyzer (to be done).  It also combines the `IVariableVisitorProvider` into `IValidationRule` which previously was a hidden optional interface layered on top of `IValidationRule`.  It will break any existing `IValidationRule` implementations.

I believe this refactor is long overdue, and due to the other breaking changes around `IMetadataProvider`, might as well be now.

Having all potential validation tie-in points within a single interface will simplify configuration of complex validation rules.  For example, users still just write this in their startup:

```cs
.AddValidationRule<MyValidationRule>()
```

and the validation rule can tie into the GraphQL validation pipeline at all/any necessary points.


src/GraphQL/Validation/Rules.Custom/InputFieldsAndArgumentsOfCorrectLength.cs
```diff
@@ -18,7 +18,7 @@ namespace GraphQL.Validation.Rules
     /// Please copy the newer <see cref="PatternMatchingDirective"/> and <see cref="PatternMatchingVisitor"/>
     /// when designing new custom directives that can be used to validate input fields and arguments.
     /// </remarks>
-    public class InputFieldsAndArgumentsOfCorrectLength : IValidationRule, IVariableVisitorProvider
+    public class InputFieldsAndArgumentsOfCorrectLength : ValidationRuleBase
```
Shane32, line 21: This is the sample custom validation rule included with GraphQL.NET that uses two tie-in points within the updated `IValidationRule` interface.  Both previously and now, it only requires a single `.AddValidationRule<InputFieldsAndArgumentsOfCorrectLength>()` line of code to install.  The main difference is that previously the `IVariableVisitorProvider` was hidden, in the sense that you'd have to know how GraphQL.NET works to know that it checks if `rule is IVariableVisitorProivder` and make use of its functionality.  Now, anyone that inspects `IValidationRule` can see the method on the interface.

docs2/site/docs/migrations/migration8.md
```diff
@@ -288,6 +288,33 @@ private class Query
...
+}
+```
+
+This allows for a single validation rule to validate AST structure, validate variable values,
+and/or validate coerced field and directive arguments.
+
+To simplify the creation of validation rules, the `ValidationRuleBase` class has been added,
+which implements the `IValidationRule` interface and provides default implementations for
+all three methods.
+
```
gao-artur, line 317: Are you going to include the comparison with `Validator` and `Parser` delegates mentioned in the previous PR? I think it was important. https://github.com/graphql-dotnet/graphql-dotnet/pull/3873

gao-artur (commented):
Looks good!

docs2/site/docs/migrations/migration8.md
Shane32: I've added the new documentation for creating custom validation rules into the official GraphQL.NET docs (none existed before), referenced from the migration document.  Ok?  It includes various samples you mentioned.

docs2/site/docs/getting-started/query-validation.md
```diff
@@ -16,3 +26,328 @@ await schema.ExecuteAsync(_ =>
...
+Here are some examples:
+
+```csharp
+// for an input object graph type
+
+Field(x => x.FirstName)
+    .Validate(value =>
+    {
+        if (((string)value).Length >= 10)
+            throw new ArgumentException("Length must be less than 10 characters.");
```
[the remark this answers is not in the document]
Shane32: Yes, and currently no special handling of `ValidationError` or `ExecutionError` exists, so all exceptions are wrapped.  Elsewhere generally throughout GraphQL.NET we allow exceptions that inherit from `ExecutionError` to pass through to the response without being wrapped.  I'm thinking about writing a PR to change this.

docs2/site/docs/getting-started/query-validation.md
```diff
@@ -16,3 +26,328 @@ await schema.ExecuteAsync(_ =>
...
+
+Field(x => x.Password)
+    .Validate(value =>
+    {
+        VerifyPasswordComplexity((string)value);
+    });
+```
+
+The `Validator` delegate is called during the validation stage, prior to execution of the request.
+Null values are not passed to the validation function. Supplying an invalid value will produce
```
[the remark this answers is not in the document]
Shane32: Yes; that's clearer.

gao-artur (approved):

PR #3907, opened 2024-01-30
Remove some deprecated methods

Shane32:
The deprecated methods removed in this PR, I believe, should be removed in v9 due to these reasons:

- Some `Add...` IGraphQLBuilder methods were replaced by `Use...` methods.  (1) there is an important distinction between `Add` and `Use`, and (2) changes are very easy to fix and all consolidated in the user's startup file
- `IsConcrete` has no useful purpose; it is functionally identical to `!type.IsAbstract`
- Telemetry provider is very new and now is a good time to change `StartActivityAsync` back to `StartActivity` (which is necessary since otherwise it would not be compatible with `AsyncLocal` internals)
- Validation code is all changing anyway

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -427,7 +427,7 @@ The following methods have been removed:
 
 | Method | Comment |
 |--------|---------|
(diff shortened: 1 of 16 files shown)


gao-artur:
> should be removed in v9

Did you mean v8?

Shane32:
> > should be removed in v9
> 
> Did you mean v8?

Yes, whoops.

PR #3913, opened 2024-02-01
Add CaseInsensitiveEnumerationGraphType and allow custom GraphType implementations in schema-first approach using DI

### LINKED ISSUE #3105: allow case-insensitive parsing of enumeration values
## Summary

When using enumeration values in mutation arguments, I want to have an option to configure, if the value is parsed case-sensitive or case-insensitive.
(I prefer case-insensitive).

This behavior seems to have changed with v5. 
After upgrading to 5.1.1 (from 4.7.1), my integration tests failed because of that.

### Basic example

Given this enum for example:
`public enum LengthSystem
{
    Metric = 0,
    Imperial = 1,
}`

I want to change that enum value via a mutation, by passing the new value as a variable in the request.
I want to make it work, even if the value is written "imperial" or "Imperial".

At the moment (v5.1.1), the mutation only works if the value is exactly "IMPERIAL".

### Motivation

To prevent typos or hard-to-find bugs. Just want to make things more robust for the caller of my GraphQL API.

PSanetra:
This PR implements two features:

1. Add a CaseInsensitiveEnumerationGraphType which enables developers to allow passing case-insensitive enum variable values
2. Allow custom GraphType implementation in schema-first approach

Fixes #3105


Shane32:
My initial thought is that this is a breaking change, since non-generic graph types are not typically registered within DI.  So any existing users that don't use the GraphQL builder methods would have a broken build.  We could fix that by using `(FieldType?)ServiceProvider.GetService(typeof(FieldType)) ?? new FieldType()` to minimize the likelyhood of an issue, although technically it's still breaking as seen in the below paragraph.

Second, we have tests to ensure that graph types are only pulled from DI once when the schema is built.  This allows use of `AddSingleton` for graph types, which is very common.  The code only works if the graph type is registered as a transient, as singleton / scoped would cause it to re-use the same graph type instance repeatedly.  This would seriously break a schema, and probably would not be apparent what the cause of the problem is.  In other words, it would be required that users only register via `AddTransient`, and the only use case is to provide a different base implementation such as using a case-insensitive enumeration graph type, and only for schema-first.  (On the other hand, if you use `AddSingleton` for a scoped schema, that's a major problem also.)  We would probably need additional documentation about this in the DI documentation page.

Third, tests don't yet pass.  However, it looks like the ApiTests need to run and a single test is failing due to the added registrations in `GraphQLBuilderBase`.  Should be an easy fix.

I really don't think we should include the "case-insensitive" implementation.  First, case-insensitivity is specifically against the GraphQL spec, and we should not be 'encouraging' such design.  I'd rather include it as a sample in the Q&A.  Further, the implementation here, although named "case-insensitive" really only allows for constant-case and upper-case.  These are multiple ways of spelling, not a difference in case.  So "JohnDoe" becomes either "JOHNDOE" or "JOHN_DOE".  You can't use "johndoe" or "johnDoe" or "john_doe".

I'd like input from other of the maintainers here.  If there are other solutions to this issue, I'd like to discuss/review them as well.

Shane32:
> You can't use "johndoe" or "johnDoe" or "john_doe".

My mistake. I see you convert the input to uppercase.  I'd suggest you just use a case-insensitive comparer in the dictionary constructors.

src/GraphQL.Tests/Utilities/SchemaBuilderExecutionTests.cs
```diff
@@ -293,6 +294,67 @@ public void can_execute_resolver()
...
+            .AddTransient<EnumerationGraphType, CaseInsensitiveEnumerationGraphType>()
+            .AddGraphQL(_ => { })
+            .AddSingleton<ISchema>(services =>
+            {
+                Builder.ServiceProvider = services;
+                Builder.Types.For("Dog").IsTypeOf<Dog>();
+                Builder.Types.For("Cat").IsTypeOf<Cat>();
+                Builder.Types.Include<PetQueryType>();
+                return Builder.Build(defs);
+            });
```
Shane32, line 340:
```suggestion
            .AddGraphQL(b => b
                .AddSchema(services =>
                {
                    Builder.ServiceProvider = services;
                    Builder.Types.For("Dog").IsTypeOf<Dog>();
                    Builder.Types.For("Cat").IsTypeOf<Cat>();
                    Builder.Types.Include<PetQueryType>();
                    return Builder.Build(defs);
                }));
```

src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
```diff
@@ -23,7 +23,17 @@ public ExecutionResult AssertQuery(Action<ExecuteConfig> configure)
         configure(config);
 
         var schema = Builder.Build(config.Definitions);
+
+        return AssertQuery(configure, schema);
```
Shane32, line 27: This calls the other overload, which runs ```configure(config)``` a second time -- ???

src/GraphQL/Types/Scalars/Enumeration/CaseInsensitiveEnumValues.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace GraphQL.Types;
+
+/// <summary>
+/// A class that represents a case insensitive set of enumeration definitions.
+/// </summary>
+public class CaseInsensitiveEnumValues : EnumValues
+{
+    private readonly IDictionary<string, EnumValueDefinition> _upperValues = new Dictionary<string, EnumValueDefinition>();
+    private readonly IDictionary<string, EnumValueDefinition> _constantCaseValues = new Dictionary<string, EnumValueDefinition>();
```
Shane32, line 11:
```suggestion
    private readonly Dictionary<string, EnumValueDefinition> _upperValues = new Dictionary<string, EnumValueDefinition>(StringComparer.InvariantCultureIgnoreCase);
    private readonly Dictionary<string, EnumValueDefinition> _constantCaseValues = new Dictionary<string, EnumValueDefinition>(StringComparer.InvariantCultureIgnoreCase);
```

src/GraphQL/Types/Scalars/Enumeration/CaseInsensitiveEnumValues.cs
```diff
@@ -0,0 +1,44 @@
+using GraphQLParser;
+
+namespace GraphQL.Types;
+
+/// <summary>
+/// A class that represents a case insensitive set of enumeration definitions.
+/// </summary>
+public class CaseInsensitiveEnumValues : EnumValues
```
Shane32, line 8: Hmmm, so inheriting from `EnumValues` so that `FindByValue` does not need to be overridden.... but it duplicates the list and all the conversions to constant case ...

src/GraphQL/Types/Scalars/Enumeration/CaseInsensitiveEnumValues.cs
```diff
@@ -0,0 +1,44 @@
...
+        var value = base.FindByName(name);
+
+        if (value != null)
+        {
+            return value;
+        }
+
+        var strName = name.ToString();
+
+        _upperValues.TryGetValue(strName.ToUpperInvariant(), out value);
```
Shane32, line 33:
```suggestion
        _upperValues.TryGetValue(strName, out value);
```

src/GraphQL/Types/Scalars/Enumeration/CaseInsensitiveEnumValues.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        if (value != null)
+        {
+            return value;
+        }
+
+        _constantCaseValues.TryGetValue(strName.ToConstantCase(), out value);
```
Shane32, line 40:
```suggestion
        _constantCaseValues.TryGetValue(strName, out value);
```
I would just match the input to both dictionaries.

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -268,7 +268,9 @@ protected virtual IObjectGraphType ToObjectGraphType(GraphQLObjectTypeDefinition
 
             var type = _types.TryGetValue(name, out var t)
                 ? t as ObjectGraphType ?? throw new InvalidOperationException($"Type '{name} should be ObjectGraphType")
-                : new ObjectGraphType { Name = name };
+                : ServiceProvider.GetRequiredService<ObjectGraphType>();
```
Shane32, line 271:
```suggestion
                : ServiceProvider.GetService<ObjectGraphType>() ?? new ObjectGraphType();
```
Probably need to add the necessary extension method within `Utilities\ServiceProviderExtensions.cs`

src/GraphQL/Utilities/SchemaBuilder.cs
```diff
@@ -370,13 +372,12 @@ protected virtual FieldType ToFieldType(string parentTypeName, GraphQLFieldDefin
...
             AssertKnownField(fieldConfig, typeConfig);
 
-            var field = new FieldType
-            {
-                Name = fieldConfig.Name,
-                Description = fieldConfig.Description ?? fieldDef.Description?.Value.ToString() ?? fieldDef.MergeComments(),
-                ResolvedType = ToGraphType(fieldDef.Type!),
-                Resolver = fieldConfig.Resolver
-            };
+            var field = ServiceProvider.GetRequiredService<FieldType>();
```
Shane32, line 375:
```suggestion
            var field = ServiceProvider.GetService<FieldType>() ?? new FieldType();
```
and similar

src/GraphQL/DI/GraphQLBuilderBase.cs
```diff
@@ -57,6 +57,14 @@ protected virtual void RegisterDefaultServices()
             Services.TryRegister(typeof(ConnectionType<,>), typeof(ConnectionType<,>), ServiceLifetime.Transient);
             Services.TryRegister<PageInfoType>(ServiceLifetime.Transient);
 
+            // configure non-generic graph types
+            Services.TryRegister(typeof(EnumerationGraphType), typeof(EnumerationGraphType), ServiceLifetime.Transient);
+            Services.TryRegister(typeof(InputObjectGraphType), typeof(InputObjectGraphType), ServiceLifetime.Transient);
+            Services.TryRegister(typeof(ObjectGraphType), typeof(ObjectGraphType), ServiceLifetime.Transient);
+            Services.TryRegister(typeof(InterfaceGraphType), typeof(InterfaceGraphType), ServiceLifetime.Transient);
+            Services.TryRegister(typeof(UnionGraphType), typeof(UnionGraphType), ServiceLifetime.Transient);
+            Services.TryRegister(typeof(FieldType), typeof(FieldType), ServiceLifetime.Transient);
```
Shane32, line 66: With the change from GetRequiredService to GetService, this isn't necessary...

src/GraphQL/Types/Scalars/Enumeration/CaseInsensitiveEnumValues.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    private readonly IDictionary<string, EnumValueDefinition> _upperValues = new Dictionary<string, EnumValueDefinition>();
+    private readonly IDictionary<string, EnumValueDefinition> _constantCaseValues = new Dictionary<string, EnumValueDefinition>();
+
+    /// <inheritdoc />
+    public override void Add(EnumValueDefinition value)
+    {
+        base.Add(value);
+        _upperValues[value.Name.ToUpperInvariant()] = value;
+        _constantCaseValues[value.Name.ToConstantCase()] = value;
```
gao-artur, line 18: I think you can use a single dictionary for both keys. Moreover, for a single-word enum, both values will produce the same key.

Shane32: > I think you can use a single dictionary for both keys. Moreover, for a single-word enum, both values will produce the same key.

Good point

src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
PSanetra: Yeah, it is not really clean, but the problem is, that the overloading methods already need the configured config, but need to pass on the configure Action. In the existing code this was already implicitly the case with the existing Overloads. (Slightly hidden in the `var runResult = Executer.ExecuteAsync(options).Result;` line). (So actually at the moment the configure Action is called three times :sweat_smile: )

PSanetra:
@Shane32 I have removed the public `CaseInsensitiveEnumerationGraphType`, and have simplified the approach with the introduction of an optional `IGraphTypeFactory` interface. Please have another look.

src/GraphQL.Tests/Utilities/SchemaBuilderTestBase.cs
Shane32: Not too worried about how the tests run so that’s fine.

Shane32:
> @Shane32 I have removed the public `CaseInsensitiveEnumerationGraphType`, and have simplified the approach with the introduction of an optional `IGraphTypeFactory` interface. Please have another look.

I like this a lot better.  The name `Factory` indicates that it is used to create new instances, similar to `IServiceScopeFactory`.

But how about this for the interface:

```cs
public interface IGraphTypeFactory
{
    // add xml comments here stating that the supported types are ObjectGraphType, EnumerationGraphType, FieldType, etc
    object CreateInstance(Type type);
}

public static GraphTypeFactoryExtensions
{
    public static T CreateInstance<T>(this IGraphTypeFactory factory)
        where T : class
        => (T)factory.CreateInstance(typeof(T));
}
```

This design makes it so there is only one instance of `GraphTypeFactory` installed in the DI engine.

PSanetra:
@Shane32 I like it better to have one Factory Service per GraphType. This way it is possible to customize only specific GraphType implementations. Otherwise I would need a single factory to support all GraphTypes.

Shane32:
Well you can use inheritance:

```cs
public MyFactory : DefaultGraphTypeFactory
{
    public object CreateInstance(Type type)
    {
        if (type == typeof(EnumerationGraphType))
            return new MyEnumGraphType();
        return base.CreateInstance(type);
    }
}
```

although honestly this isn't much different:
```cs
public MyFactory : IGraphTypeFactory
{
    public object CreateInstance(Type type)
        => type == typeof(EnumerationGraphType) ? new MyEnumGraphType() : Activator.CreateInstance(type);
}
```

PSanetra:
Yes it is not impossible to find ways, but if possible I would like to avoid coupling between different Factory implementations.

Shane32:
@gao-artur any comments here?

gao-artur:
Looks good 👌

Shane32:
@gao-artur any thoughts on `IGraphTypeFactory<T>.Create` vs what's basically `IGraphTypeFactory.Create<T>` ?

gao-artur:
In general, a single factory makes more sense for me, but I understand @PSanetra's logic.

Shane32:
I just feel that this is not a good design pattern.  Are there other examples of this design pattern in practice?  As a single `IGraphTypeFactory` it's much like `IServiceProvider`, so it feels like we are reinventing the wheel here to do it differently.  And you can always decouple it on the user side if you wanted to with an implementation of `IGraphTypeFactory` that pulled the requested service from the underlying service provider.

PSanetra:
@Shane32 I think a good example of this pattern is the `IAzureClientFactory<TClient>` which is the generic factory interface to use to create client instances in all Azure SDKs.

See https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.azure.iazureclientfactory-1?view=azure-dotnet

src/GraphQL/Types/GenericGraphTypeFactory.cs
```diff
@@ -0,0 +1,13 @@
+namespace GraphQL.Types;
+
+/// <summary>
+/// A generic factory to create instances of specific TGraphType implementations from parameterless constructors.
+/// </summary>
+public class GenericGraphTypeFactory<TGraphType> : IGraphTypeFactory<TGraphType> where TGraphType : IGraphType
```
gao-artur, line 6: IMO, `DefaultGraphTypeFactory` is a better name.

src/GraphQL/Types/GenericGraphTypeFactory.cs
```diff
@@ -0,0 +1,13 @@
+namespace GraphQL.Types;
+
+/// <summary>
+/// A generic factory to create instances of specific TGraphType implementations from parameterless constructors.
```
gao-artur, line 4:
```suggestion
/// A generic factory to create instances of specific <typeparamref name="TGraphType"/> implementations from parameterless constructors.
```

src/GraphQL/Types/GenericGraphTypeFactory.cs
```diff
@@ -0,0 +1,13 @@
...
+/// </summary>
+public class GenericGraphTypeFactory<TGraphType> : IGraphTypeFactory<TGraphType> where TGraphType : IGraphType
+{
+    /// <summary>
+    /// Create a new instance of TGraphType. Requires a parameterless constructor.
```
gao-artur, line 9:
```suggestion
    /// Create a new instance of <typeparamref name="TGraphType"/>. Requires a parameterless constructor.
```

src/GraphQL/Types/IGraphTypeFactory.cs
```diff
@@ -0,0 +1,13 @@
+namespace GraphQL.Types;
+
+/// <summary>
+/// A factory to create instances of specific TGraphType implementations.
```
gao-artur, line 4:
```suggestion
/// A factory to create instances of specific <typeparamref name="TGraphType"/> implementations.
```

src/GraphQL/Types/IGraphTypeFactory.cs
```diff
@@ -0,0 +1,13 @@
...
+/// </summary>
+public interface IGraphTypeFactory<out TGraphType> where TGraphType : IGraphType
+{
+    /// <summary>
+    /// Create a new instance of TGraphType
```
gao-artur, line 9:
```suggestion
    /// Create a new instance of <typeparamref name="TGraphType"/>
```

Shane32:
Sounds good. I’ll review again tonight

src/GraphQL/Types/GenericGraphTypeFactory.cs
Shane32: Agree.  It also matches `DefaultServiceProvider` and `DefaultNameConverter` and stuff

Shane32:
Let's make the changes @gao-artur suggested and then merge.

PSanetra:
@gao-artur @Shane32 Alright I have applied the suggested changes and squashed the commits.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -2207,6 +2207,12 @@ namespace GraphQL.Types
         public override object? ParseLiteral(GraphQLParser.AST.GraphQLValue value) { }
         public override object? ParseValue(object? value) { }
     }
+    public class GenericGraphTypeFactory<TGraphType> : GraphQL.Types.IGraphTypeFactory<TGraphType>
```
gao-artur, line 2210: @PSanetra run `GraphQL.ApiTests.csproj` tests locally to update the approved API

PSanetra: Done :slightly_smiling_face:

Shane32 (approved):

PSanetra:
@Shane32 @gao-artur Thank you for merging! Do you have an ETA when this change might be released? Is it possible to get a pre-release containing this change?

Shane32:
When reasonable, I usually issue a release when asked; I can do it now.

You can get prerelease versions from the GitHub Package Registry.  Unfortunately it requires authentication so it's a bit of a pain to use.  See https://github.com/graphql-dotnet/graphql-dotnet/pkgs/nuget/GraphQL .  Version [7.6.2-preview-980](https://github.com/orgs/graphql-dotnet/packages/nuget/GraphQL/174978698) would be the latest 7.x version (newer than 7.7).

PR #3921, opened 2024-02-26
Code-First & Type-First Federation 2 Support

Shane32:
Based on #3144 by @Mithras

- New code is in namespace GraphQL.Federation
- Old code in namespace GraphQL.Utilities.Federation has been marked as obsolete
- Federation code is unified into `.AddFederation()` for both schema-first, code-first and type-first use cases.
- Old schema-first syntax is marked as obsolete, but will function without any (or very few) changes
- Code first supported, mostly by @Mithras 
- Type first supported
- Apollo Router now runs an actual test against four live subgraph schemas (aka demo apps) to form a single supergraph:
  - Schema first, new syntax, federation 1 compatible
  - Schema first, old syntax, federation 1 compatible
  - Code first, federation 2 compatible
  - Type first, federation 2 compatible

Breaking changes:
- A couple classes were moved.
- The `__typename` field is not implicitly added to federation requests

Prerequisites:
- [x] #3952 
- [x] #3949 
- [x] #3951
- [x] #3965 
- [x] #3975

Split into separate PRs:
- [x] #3957 
- [x] #3958 
- [x] #3961 
- [x] #3960
- [x] #3963 
- [x] #3964 
- [x] Add sample applications / tests -- this PR

To-do / nice-to-have:
- [ ] Analyzer rule for new `[FederationResolver]` attribute to ensure it is used correctly
- [ ] #3962 
- [x] New tests for Sample3 and Sample4 web applications
- [x] Additional federation testing for Sample3 and Sample4
- [x] Move new extension methods to GraphQL.Federation namespace rather than GraphQL.Federation.Extensions
- [x] Add documentation
- [x] Break PR into smaller pieces and merge each piece separately
- [x] Merge master/develop into PR
- [x] Add xml comments to all federation classes/etc

Moved out of scope:
- [ ] New validation rule - see https://github.com/graphql/graphql-spec/pull/1098
- [ ] Consider: auto-hide any unreferenced types in the schema (??)
- [ ] Consider: post-initialization schema visitor


src/GraphQL.sln
```diff
@@ -132,6 +132,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GraphQL.MicrosoftDI.Tests",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GraphQL.DataLoader", "GraphQL.DataLoader\GraphQL.DataLoader.csproj", "{466AD883-96BD-458E-9872-3C14E49D18F8}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GraphQL.Federation.Tests", "GraphQL.Federation.Tests\GraphQL.Federation.Tests.csproj", "{54F047AD-9227-4894-8188-6E8DE08DFFC1}"
```
Shane32, line 135:
```suggestion
```

src/GraphQL.sln
```diff
@@ -255,6 +256,10 @@ Global
 		{466AD883-96BD-458E-9872-3C14E49D18F8}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{466AD883-96BD-458E-9872-3C14E49D18F8}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{466AD883-96BD-458E-9872-3C14E49D18F8}.Release|Any CPU.Build.0 = Release|Any CPU
+		{54F047AD-9227-4894-8188-6E8DE08DFFC1}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{54F047AD-9227-4894-8188-6E8DE08DFFC1}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{54F047AD-9227-4894-8188-6E8DE08DFFC1}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{54F047AD-9227-4894-8188-6E8DE08DFFC1}.Release|Any CPU.Build.0 = Release|Any CPU
```
Shane32, line 262:
```suggestion
```

Shane32:
@gao-artur How involved do you want to be with this PR?

So far I've taken what others have done before me and made the following modifications:
- `ToObject` design was changed for v8, requiring the necessary functionality rewritten for federation
- Rewrite `IFederationResolver` to return `ValueTask<object?>` and properly support asynchronous tasks
- Fix entity resolver to support serial and parallel execution strategies
- Repair tests and other misc for v8 compatibility
- Remove built-in camel-case conversion of field names within attributes (this means `nameof` won't work, but now it is compatible with fields of any case, as camel-case is not mandated by graphql)
- Utilize new schema printer, fix where directives are stored
- Fix extension methods for v8 to use `IMetadataWriter` and similar (new for v8)
- Remove `NeverType` from codebase (requires #3927 )

Note: I read a note in the code that types' permissions are not validated from the federation resolver.  This is true for any union type; the union type itself is validated but not the specific type returned, as it is not known until the resolver executes.  We should consider fixing this for v8, not just for federation, but for unions or interfaces in general (or else be sure to add the applicable notes to the documentation that this occurs).

Shane32:
cc: @Mithras

gao-artur:
Hi @Shane32, sorry I was a bit off last month. Changes at work, and I hardly have free time now.
I'm not sure I have enough knowledge of the GraphQL.NET internals to review this PR, but I definitely want to know how it works because I want to adopt the Federation shortly after it is released. So, I'll review all your PR's anyway, even after you merge them.
And I'll try to finish the input graph type analyzers during this month because I think they must be ready for v8.0.

Shane32:
@Mithras @michael-watson @mac2000  I'm going to try to put time into this PR over the next couple weekends, so GraphQL.NET v8 can be released.  It's at the point now where I think the code is really close to what it should be.  Let me know if you'd like to spend time reviewing it or assisting also.

Shane32:
@joemcbride I've been working on GraphQL.NET Federation support for code-first and "type-first" approaches in addition to the schema-first approach already available, basing my efforts on prior work done by @Mithras and others.  I've also unified the schema-first setup with the other approaches, while still supporting (for v8) any legacy schema-first code (with applicable code marked as obsolete).  Let me know if you'd like to review the changes here.  I know there's a lot but if you're interested, I can guide you what will affect schema-first.  Mainly I'd look at these four subgraph projects, which each are part of a single supergraph (tested together within CI):
- GraphQL.Federation.SchemaFirst.Sample1 - schema-first with new syntax
- GraphQL.Federation.SchemaFirst.Sample2 - schema-first with old syntax
- GraphQL.Federation.CodeFirst.Sample3 - code-first
- GraphQL.Federation.TypeFirst.Sample4 - "type-first"

@sungam3r @gao-artur Let me know if you'd like to participate in this PR also.

I'd really like to get GraphQL.NET v8 released with code-first federation support, plus the other new enhancements that are ready to be released.

mac2000:
@Shane32 in our case we have bazillion of services running on graphql-net with self made federation introduced back then in issue which I can not find anymore

side by side with new services based on Hotchocolate

I will ask our backend engineer to see if he will be able to get up and running one of the services based on this branch and come back with feedback if everything workedout

gao-artur:
I have yet to learn the theory, but I'm trying to understand how the federated schemas should work based on the sample projects. I see that the `Sample3` project defines the `UserGraphType` with two fields: `id` and `reviews`. The `Sample4` project defines the `User` type with fields `id` and `username`. So, if I start and configure the Apollo client, I will see the `user` type with 3 fields: `id`, `username` and `reviews`? Is that correct?
Btw, why the Sample3 and 4 projects are not in the solution file?

Shane32:
> Btw, why the Sample3 and 4 projects are not in the solution file?

Weird.  Fixed.

> So, if I start and configure the Apollo client, I will see the user type with 3 fields: id, username and reviews? Is that correct?

Yes, that is correct.  Conceptually, the samples are microservices as follows:

1. Categories microservice.  Has a Category type but doesn't depend on any other microservices.
2. Products microservice.  Has a Product type, and extends Category to view all products in that category.
3. Reviews microservice.  Has a Review type, and extends Product and User types.
4. Users microservice.  Has a User type but doesn't depend on any other microservices.

See `federation-request-4.json` and `federation-response-4.json` for a sample query that runs a query across all services, where the Apollo Router will need to wait for an answer from certain subgraphs to be able to forward additional requests to other subgraphs, until it can assemble the entire response.

gao-artur:
So, if I want to resolve `username` from `review.author`, Apollo will send an HTTP request with the user id to the Users microservice? What if there is a list of `users`? Does it support DataLoader to issue a single HTTP request?

Shane32:
> So, if I want to resolve `username` from `review.author`, Apollo will send an HTTP request with the user id to the Users microservice? What if there is a list of `users`? Does it support DataLoader to issue a single HTTP request?

Yes, Apollo will send a list of entities that need resolving together in a single HTTP request.  Within GraphQL.NET we can use data loaders to issue a single request to the underlying database.

gao-artur:
Thanks! I'll try to read more about Federation on the weekend and try to help with the review.

gao-artur:
I'm still learning federation basics, but I noticed the federation subgraph can define a schema with [types only](https://www.apollographql.com/docs/federation/federated-types/overview/#subgraph-schemas) (`Reviews` schema) and without any query field. Will GraphQL.NET allow such schemas when `.AddFederation()` is used?

Shane32:
It should.  Technically there should still be a minimum of one field, `_services`, which is automatically added to the query type.  `AddFederation` should add this before running schema validation checks.  We should add a test for this scenario, however.

Chances are that `AddFederation` does not yet handle the scenario where if there are zero resolvable types, then the `_entity` field is not added to the query type.  I should add a test for that as well.

Shane32:
> I'm still learning federation basics, but I noticed the federation subgraph can define a schema with [types only](https://www.apollographql.com/docs/federation/federated-types/overview/#subgraph-schemas) (`Reviews` schema) and without any query field. Will GraphQL.NET allow such schemas when `.AddFederation()` is used?

I've revised Sample3 (the 'Reviews' schema) to remove the field added to the Query type.  I've also added support for subgraphs with no resolvable entities.  For those subgraphs, the `_Entity` type and `_entities` field is removed from the schema.  Finally, I added additional tests for Sample3 and Sample4.

src/GraphQL/Federation/Extensions/FederationHelper.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// <summary>
+    /// Applies a <c>@link</c> directive to the specified GraphQL schema based on the provided <see cref="FederationDirectiveEnum"/> values.
+    /// </summary>
+    /// <param name="schema">The GraphQL schema to build the link extension for.</param>
+    /// <param name="import">The <see cref="FederationDirectiveEnum"/> values specifying which directives to include in the link extension.</param>
+    public static void ApplyLinkDirective(this ISchema schema, FederationDirectiveEnum import)
+    {
+        schema.ApplyDirective("link", d =>
+        {
+            d.AddArgument(new DirectiveArgument(URL_ARGUMENT) { Value = "https://specs.apollo.dev/federation/v2.0" });
```
gao-artur, line 154: Is there a plan to support higher [specs](https://www.apollographql.com/docs/federation/federation-versions/) versions?
From what I understand, higher specs require clients to upgrade their router version. So, maybe this version can be set dynamically based on the features used. For example, if the `@interfaceObject` directive was applied during the schema definition, the version here should be updated to v2.3.

Shane32:
@vadymkutsenko Thanks for testing - it's very encouraging!

> app.UseGraphQL doesn't work

Makes sense - probably because the GraphQL.Server.Transports.AspNetCore package referenced ISchema inside the NuGet GraphQL.dll and you have compiled your own ISchema.  Chances are that you can solve this by using the beta build from GitHub Packages, if you so desired to do so.  This will all be cleaned up when v8 is released of course.

src/GraphQL/Federation/Extensions/FederationHelper.cs
Shane32: I think I'll change the `AddFederation` method argument to be a builder method, so we can add features or configuration settings going forwards.  Perhaps the federation version should be a string property?

gao-artur: Yep, I thought about such config as well. It can be a "higher supported spec version" config. Then, we can add a validation that will throw if the user used a feature from a higher version. So, I can see two approaches here:
1. Automatic detection of the minimal required version based on used features
2. Explicit higher version set by configuration

I think both are useful.

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -67,6 +67,7 @@ public void entity_query()
         const string query = """
                 query ($_representations: [_Any!]!) {
                     _entities(representations: $_representations) {
+                        __typename
```
Shane32, line 70: `__typename` is no longer automatically injected into the query

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -88,7 +89,6 @@ public void entity_query()
     }
 
     [Theory]
-    [InlineData("...on User { id }", false)]
```
Shane32, line 91: `__typename` is no longer automatically injected into the query.

src/GraphQL.Tests/Utilities/FederatedSchemaBuilderTests.cs
```diff
@@ -214,6 +214,7 @@ public void resolve_reference_is_not_trying_to_await_for_each_field_individialy_
         const string query = """
             {
                 _entities(representations: [{__typename: "User", id: "1" }, {__typename: "User", id: "2" }]) {
+                    __typename
```
Shane32, line 217: `__typename` is no longer automatically injected into the query.

src/GraphQL/Federation/Types/LinkImportGraphType.cs
```diff
@@ -0,0 +1,15 @@
+namespace GraphQL.Federation.Types;
+
+/// <summary>
+/// Represents a scalar type for link imports in GraphQL Federation.
+/// Used to define imported directives.
+/// The name of this graph type is "link__Import".
+/// </summary>
+public class LinkImportGraphType : AnyScalarGraphType
```
gao-artur, line 8: Why does it derive from Any and not String?

Shane32:
@gao-artur This PR now only contains additional tests, some created by me, others by @Mithras .

src/GraphQL/Federation/Extensions/FederationHelper.cs
Shane32: > Yep, I thought about such config as well. It can be a "higher supported spec version" config. Then, we can add a validation that will throw if the user used a feature from a higher version. So, I can see two approaches here:
> 
> 1. Automatic detection of the minimal required version based on used features
> 2. Explicit higher version set by configuration
> 
> I think both are useful.

Completed

gao-artur:
This is incredible work! Thank you and @Mithras for making this possible!

gao-artur (approved):
This is incredible work! Thank you and @Mithras for making this possible!

PR #3927, opened 2024-03-05
Federation prep: Remove union check

Shane32:
For:
- #3921 

This check already exists within the schema validator code and is not necessary here.  It must be removed to allow the federation PR to dynamically build the union graph type after the schema has been initialized.  It must be built after the schema is initialized as the initialization process recursively discovers all of output graph types referenced by the schema.


Shane32:
image

gao-artur (approved):

PR #3931, opened 2024-03-11
Add IListConverterFactory

### LINKED ISSUE #3930: Support more collection type conversions from input list
[ObjectExtensions.GetPropertyValue](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Extensions/ObjectExtensions.cs#L225) current behaviour:

When the `fieldType` is `ISet<T>`, `IReadOnlySet<T>`, `IImmutableSet<T>`, or any of the concrete implementations it returns `List<T>`. I believe it should be considered a bug.

For `ImmutableArray<T>` it throws
> System.NotSupportedException
> Specified method is not supported.
>    at System.Collections.Immutable.ImmutableArray`1.System.Collections.IList.Add(Object value)

For `ImmutableList<T>` it throws
> System.MissingMethodException
> No parameterless constructor defined for type 'System.Collections.Immutable.ImmutableList`1[System.Double]'.

Shane32:
Closes #3930 

Adds:
- `ValueConverter.RegisterListConverterFactory` to register a factory instance for a generic type, like for `IImmutableList<>`
- `ValueConverter.RegisterListConverterFactory` to register an interface generic type to a concrete generic type, like for `IEnumerable<T>` to `List<T>` instead of `T[]`
- `ValueConverter.GetListConverterFactory` to pull a factory instance from the registered list
- `ValueConverter.GetListConverter` to get a converter for a specific type (e.g. `IImmutableList<int>`)
- `IListConverterFactory` - an interface for creating list converters for a specific generic type (e.g. `ImmutableList<>`)
- `IListConverter` - an interface for converting an object array (`object[]`) to a specific type (e.g. `ImmutableList<int>`)
- `ListConverter` - a class to hold a conversion delegate and element type properties
- `ListConverterFactoryBase` - an abstract class containing some helper code for most factory implementations

Sample custom list converter:

```cs
internal sealed class ImmutableListConverterFactory : ListConverterFactoryBase
{
    public override Func<object?[], object> Create<T>()
        => list => ImmutableList.CreateRange(list.Cast<T>());
}
```

Sample registration:

```cs
var immutableListConverter = new ImmutableListConverterFactory();
RegisterListConverterFactory(typeof(ImmutableList<>), immutableListConverter);
RegisterListConverterFactory(typeof(IImmutableList<>), immutableListConverter);
```

Reflection-based list converters are provided by default, so these list types do not need a dedicated converter:

```cs
public class MyList<T>
{
    public MyList(IEnumerable<T> list) { /* ... */ }
}

public class MyList2<T>
{
    public void Add(T value) { /* ... */ }
}
```

You can also register an interface type like this:

```cs
RegisterListConverterFactory(typeof(IMyList<>), typeof(MyList<>));
```

Open generics with multiple arguments are not supported.

Default behavior is to use an array whenever possible to minimize allocations, so `IEnumerable<T>` and `IList<T>` and so on will return an array without a `List<T>` wrapper.

Note that the default implementations will cast a null value to `default(T)`.  So an input graph type of `Int` containing `[null, 2, 3]` will not throw an error but will coerce to `[0, 2, 3]` for `int[]`, while for `int?[]` it would be `[null, 2, 3]`.  This must be done by each implementation, so the immutable list converter factory code provided above will not do this and would throw an exception.  The reason is that this was the general behavior in the past of `GetArgument<T>`, and so this legacy behavior is maintained.


src/GraphQL/Conversion/ListConverterFactories/CustomListConverterFactory.cs
```diff
@@ -0,0 +1,153 @@
...
+/// <summary>
+/// Dynamically constructs a list of the specified type using reflection.
+/// The type must have a public constructor that takes a single parameter compatible
+/// with an array of the element type, such as <see cref="IEnumerable{T}"/>,
+/// or a public parameterless constructor and a public method named "Add" that
+/// takes a single parameter of the element type.
+/// Alternatively, the type may implement <see cref="IList"/> and have a public
+/// parameterless constructor.
+/// </summary>
+internal sealed class CustomListConverterFactory : IListConverterFactory
```
Shane32, line 17: This class replaces the previous reflection-based approach available in previous versions of GraphQL.NET.  While this works for Queue and Stack, it does not work for the Immutable classes because (1) there is no public constructor that accepts `IEnumerable<T>`, and (2) there is no public default constructor along with a public method `Add<T>(T item)`.

src/GraphQL/Conversion/IListConverter.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <summary>
+    /// The type of the list element.
+    /// </summary>
+    Type ElementType { get; }
+
+    /// <summary>
+    /// Converts a list of objects to a specific list type, such as <see cref="HashSet{T}"/>.
+    /// </summary>
+    object Convert(object?[] list);
+}
```
Shane32, line 17: The list converter itself primarily just has a `Convert` method to convert a list of objects (`object[]`) into the intended instance; in my example, to `HashSet<int>`.

src/GraphQL/Conversion/ListConverterFactories/ListConverterFactoryBase.cs
```diff
@@ -0,0 +1,60 @@
...
+                throw; //unreachable
+            }
+        });
+    }
+
+    /// <summary>
+    /// Returns a converter which will convert items from a given <c>object[]</c> list which contains
+    /// items of the specified <typeparamref name="T"/> into a list instance.
+    /// </summary>
+    public abstract Func<object?[], object> Create<T>();
```
Shane32, line 59: The abstract `ListConverterFactoryBase` class provides three functions:
1. Analyzes the generic type parameters and extracts the element type (e.g. `int` for `HashSet<int>`)
2. Provides caching of generated converter instances
3. Calls a typed `Create<T>` method to make it easier to write list converter factories.  (`T` is the element type)

src/GraphQL/Extensions/ObjectExtensions.cs
```diff
@@ -272,6 +272,7 @@ public MemberFieldInfo(string key, MemberInfo member, bool isInitOnly, bool isRe
     /// will have 'FirstName' key but its value should be set to 'FName' property of created object.
     /// </param>
     /// <remarks>There is special handling for strings, IEnumerable&lt;T&gt;, Nullable&lt;T&gt;, and Enum.</remarks>
+    [RequiresUnreferencedCode("Recursively converts propertyValue to fieldType.")]
     public static object? GetPropertyValue(this object? propertyValue, Type fieldType, IGraphType mappedType)
```
Shane32, line 276: One of the benefits of this new design is that the amount of code within `ObjectExtensions.GetPropertyValue` is notably simplified.  Also the similar code within `ObjectExtensions.Compiled.cs` is greatly simplified.  In general, all the reflection-based code got moved to `CustomListConverterFactory` (where it is much easier understood), and all the code dedicated specifically to array or `List<T>` generation was removed as well, being replaced with `ArrayListConverterFactory` and `DefaultListConverterFactory`, respectively.  Finally, the type analysis code was also removed and replaced with a dictionary lookup within `ValueConverter.GetListConverterFactory` and the generic type analysis within `ListConverterFactoryBase`.

In short, the design is better regardless of whether it allows for custom list converters or not.

src/GraphQL/Extensions/TypeExtensions.cs
```diff
@@ -14,54 +14,37 @@ namespace GraphQL;
...
-        typeof(ICollection<>),
-        typeof(IReadOnlyCollection<>),
-        typeof(IReadOnlyList<>),
-    ];
-
     /// <summary>
-    /// Returns the list type of the indicated type.
+    /// Converts the specified <see cref="IEnumerable"/> to an <see cref="Array"/> of type <see cref="object"/>.
     /// </summary>
-    internal static (bool IsArray, bool IsList, Type ElementType) GetListType(this Type type)
```
Shane32, line 35: `GetListType` also got removed.

src/GraphQL/Conversion/ListConverterFactories/ArrayListConverterFactory.cs
```diff
@@ -0,0 +1,24 @@
+namespace GraphQL.Conversion;
+
+internal sealed class ArrayListConverterFactory : ListConverterFactoryBase
```
Shane32, line 3: All the default factories are internal.  If someone wants a reference they can call `ValueConverter.GetListConverterFactory(typeof(int[]))` or similar.  I don't see why anyone would need a reference besides accessing it through `ValueConverter.GetListConverterFactory`.

src/GraphQL/Conversion/ListConverterFactories/ArrayListConverterFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    private ArrayListConverterFactory()
+    {
+    }
+
+    public static ArrayListConverterFactory Instance { get; } = new ArrayListConverterFactory();
```
Shane32, line 9: Most of the internal factories do not have an `Instance` property, since they are only constructed once in the static constructor of `ValueConverter`.  Each of them should effectively be a singleton as they cache created list converters.

Another design would be to eliminate caching within the factories, and provide a global cache within `ValueConverter`.  So you call `ValueConverter.GetListConverter(Type listType)` and only incur a single dictionary lookup, rather than `ValueConverter.GetListConverterFactory(Type listType).Create(Type listType)` which incurs two dictionary lookups.

src/GraphQL/Conversion/ListConverterFactories/ImmutableListConverters.cs
```diff
@@ -0,0 +1,23 @@
...
+    public override Func<object?[], object> Create<T>()
+        => list => ImmutableList.CreateRange(list.Cast<T>());
+}
+
+internal sealed class ImmutableSetListConverterFactory : ListConverterFactoryBase
+{
+    public override Func<object?[], object> Create<T>()
+        => list => ImmutableHashSet.CreateRange(list.Cast<T>());
+}
+#endif
```
Shane32, line 23: Sample immutable implementations.  (I'm removing from this PR for the moment; we can discuss further.)

src/GraphQL/Conversion/ValueConverter.cs
```diff
@@ -136,6 +142,36 @@ static ValueConverter()
...
+#if NETCOREAPP1_0_OR_GREATER
+        var immutableListConverter = new ImmutableListConverterFactory();
+        RegisterListConverterFactory(typeof(ImmutableList<>), immutableListConverter);
+        RegisterListConverterFactory(typeof(IImmutableList<>), immutableListConverter);
+        var immutableSetConverter = new ImmutableSetListConverterFactory();
+        RegisterListConverterFactory(typeof(ImmutableHashSet<>), immutableSetConverter);
+        RegisterListConverterFactory(typeof(IImmutableSet<>), immutableSetConverter);
+        var immutableArrayConverter = new ImmutableArrayListConverterFactory();
+        RegisterListConverterFactory(typeof(ImmutableArray<>), immutableArrayConverter);
+#endif
```
Shane32, line 174: Sample registrations to enable the immutable factories.

src/GraphQL/Conversion/ListConverterFactories/CustomListConverterFactory.cs
```diff
@@ -0,0 +1,215 @@
...
+
+    /// <summary>
+    /// Creates a delegate that calls the specified <paramref name="constructor"/> after casting
+    /// the delegate's parameter to <see cref="IEnumerable{T}"/> of the <paramref name="elementType"/>.
+    /// </summary>
+    private static Func<object?[], object> CreateLambdaViaConstructor(Type elementType, ConstructorInfo constructor)
+    {
+        var methodInfo = _castMethodInfo.MakeGenericMethod(elementType);
+
+        // todo: directly call the constructor when GlobalSwitches.DynamicallyCompileToObject is false (??)
```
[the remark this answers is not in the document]
Shane32: Yes, but in an AOT scenario it is impossible to compile code.  So then it is running an expression tree via interpretation and reflection, which would be slower than just calling the method via reflection to begin with.

The thing is that `GlobalSwitches.DynamicallyCompileToObject` is more for scoped schemas, not just for AOT scenarios.

So for now the reflection-based code is commented out.

src/GraphQL/Conversion/ValueConverter.cs
```diff
@@ -136,6 +140,25 @@ static ValueConverter()
...
+        RegisterListConverterFactory(typeof(IList), ArrayListConverterFactory.Instance);
+        RegisterListConverterFactory(typeof(IList<>), ArrayListConverterFactory.Instance);
+        RegisterListConverterFactory(typeof(IEnumerable<>), ArrayListConverterFactory.Instance);
+        RegisterListConverterFactory(typeof(ICollection<>), ArrayListConverterFactory.Instance);
+        RegisterListConverterFactory(typeof(IReadOnlyList<>), ArrayListConverterFactory.Instance);
+        RegisterListConverterFactory(typeof(IReadOnlyCollection<>), ArrayListConverterFactory.Instance);
+        // types that return a List<T>
+        RegisterListConverterFactory(typeof(List<>), new DefaultListConverterFactory());
+        // types that return a HashSet<T>
+        var hashSetConverter = new HashSetListConverterFactory();
```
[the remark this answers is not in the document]
Shane32: Sure.

Shane32:
So how it works is that during argument parsing, everything will be one of these:
- for list types, `object?[]`
- for input objects, return of `InputObjectGraphType.ParseDictionary`, which is `Dictionary<string, object?>` for `InputObjectGraphType<object>` and for strongly typed objects will by default call `ToObject` with the dictionary.
- for scalars, the return value of `ScalarGraphType.ParseValue` (or `.ParseLiteral`)

Now for any input object fields or output field arguments, it processes ParseValue as it is constructing this object tree - sort of in a reverse manner from the bottom of the tree.

Now all the code in this PR affects ToObject / GetPropertyValue, which isn't actually used directly in the argument parsing process, except in certain scenarios:

1. Within `InputObjectGraphType<T>.ParseDictionary` as noted above

2. If you use `Field(x => x.MyList)` for a input object field definition, the method detects the Type of `MyList` (let's say it is `HashSet<int>`) and then automatically executes this:

```cs
builder.ParseValue(value => value is TProperty ? value : value.GetPropertyValue(typeof(TProperty), builder.FieldType.ResolvedType!)!);
```

3. If you use "type-first" schemas, the same thing happens; the coercion to the proper type is automatically added via ParseValue.  Subsequent calls to ParseValue would replace this 'default' behavior.

However, if you use `Field("MyList")` then the type is not known and no call to ParseValue occurs by default.  The code in this PR would only be hit when you called `GetArgument<HashSet<int>>()` -- then the list (which still would be `object[]`) would get converted to `HashSet<int>` by this code.

Umm, did I make any clearer?

So back to your questions:

> Does `ParseValue` work for collections at all?

Well, yes.  It operates on a input field / query argument level.  So if the field is a `[Int]` then ParseValue can be used for the collection but will not operate on the individual values within the collection.

> Also, if we have a list converter for a specific type, will the `ParseValue` have precedence?

Generally, yes.  Take this example:

```cs
Field(x => x.MyList); //which is HashSet<int>
    .ParseValue(value =>
    {
        // <---
    });
```

At that point in the code `value` is an `object[]` because although the default implementation of `ParseValue` in this instance would have called `GetPropertyValue` (and the list converter), the `ParseValue` delegate was _replaced_ with user code.

> Does that mean the `ParseValue` is the only option to support immutable types?

No ... but the `CustomListProviderFactory` cannot do it.  You need to register a custom list provider factory, like this:

```cs
public class ImmutableListConverterFactory : ListConverterFactoryBase
{
    public override Func<object?[], object> Create<T>()
        => list => ImmutableList.CreateRange(list.Cast<T>());
}

RegisterListConverterFactory(typeof(ImmutableList<>), new ImmutableListConverterFactory());
```

gao-artur:
Great explanation, thanks!

The only scenario I think is missing is allowing more granular list conversion control. For example, I can register a case-insensitive `HashSet` converter globally
```
list => new HashSet<string>(list, StringComparer.InvariantCultureIgnoreCase);
```

But I can't register a case-sensitive conversion for a specific field without manually parsing the inner elements.

Shane32:
Well, you can just use ParseValue on the specific field.

```cs
Field(x => x.AlternateNames)
    .ParseValue(list => new HashSet<string>(list.Cast<string>()));
```

I think a consideration also is whether you can register a closed generic?

```cs
RegisterListConversionFactory(typeof(HashSet<string>), new CaseInsensitiveHashSetListFactory());
```

This will cause an additional dictionary lookup for each registered type.  Still might be a good idea though.

gao-artur:
Ah, right, string requires just casting. Comlplex objects require manual parsing. But it's good enough.
Looks great! Thank you!

gao-artur (approved):

PR #3932, opened 2024-03-12
AOT compatibility for IListConverter; add migration notes

Shane32:
This PR does two things:

1. Provides code paths for AOT compilation

`ListConverterFactoryBase` used `MakeGenericMethod` to call a typed version of `Create` on its derived class.  This would fail for any value type in AOT scenarios because, for example, the compiler would not know to compile for `Create<int>` or `Create<long?>`, etc.  Essentially it would completely break ToObject.  I considered adding direct links effectively pinning the create method for native value types, but this would consequently pin and compile for all those list types as well.  (For all value types and their nullable counterparts, this would be a couple dozen versions of List and HashSet).  So instead the code relies on the fact that types passed in are pinned by the caller, as just provides an optimized code path for that scenario.

2. Allows for typed list converters to be installed.

Allows for a typed list converter to be added easily, like this:

```cs
ValueConverter.RegisterListConverter<HashSet<string>, string>(values => new HashSet<string>(values));
```


docs2/site/docs/migrations/migration8.md
```diff
@@ -319,6 +319,50 @@ Documentation has been added to the [Query Validation](https://graphql-dotnet.gi
...
+}
+
+ValueConverter.RegisterListConverterFactory(typeof(IImmutableList<>), new ImmutableListConverterFactory());
+```
+
+Finally, if you simply need to map an interface list type to a concrete list type, you can do so as follows:
+
+```csharp
+ValueConverter.RegisterListConverterFactory(typeof(IList<>), typeof(List<>)); // default mapping is T[]
+```
```
gao-artur, line 364: I think we need a dedicated doc page for `ValueConverter`. Currently, it is only mentioned in the [Custom Scalar](https://graphql-dotnet.github.io/docs/getting-started/custom-scalars/#valueconverter) page. But it has a wider usage now.

gao-artur (approved):

docs2/site/docs/migrations/migration8.md
Shane32: #3933

PR #3947, opened 2024-06-02
Add IsPrivate and SkipTypeCheck properties to IGraphType / IFieldType

Shane32:
These are prerequisites for GraphQL Federation support in #3921 .  This PR targets v8.

## IsPrivate

Allows schema visitors to set `IGraphType.IsPrivate` or `IFieldType.IsPrivate` to effectively remove a graph type or field from the schema.  Introspection queries cannot see these graph types or fields, and requests cannot reference them.  This is the only viable solution currently because `SchemaTypes` builds a read-only list during initialization, and similarly the fields cannot be removed from graph types.

## SkipTypeCheck

For schema-first schemas, skips `IsTypeOf` checks, while allowing to set `IsTypeOf` for other scenarios.

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -363,6 +363,18 @@ Finally, if you simply need to map an interface list type to a concrete list typ
 ValueConverter.RegisterListConverterFactory(typeof(IList<>), typeof(List<>)); // default mapping is T[]
 ```
 
+### 10. `IGraphType.IsPrivate` and `IFieldType.IsPrivate` properties added
+
+Allows to set a graph type or field as private within a schema visitor, effectively removing it from the schema.
+Introspection queries will not be able to query the type/field, and queries will not be able to reference the type/field.
+
+### 11. `IObjectGraphType.SkipTypeCheck` property added
+
+Allows to skip the type check for a specific object graph type during resolver execution. This is useful
+for schema-first schemas where the CLR type is not defined while the resolver is built, while allowing
+`IsTypeOf` to be set automatically for other use cases. Schema-first schemas will automatically set this
+property to `true` for all object graph types to retain the existing behavior.
+
 ## Breaking Changes
 
 ### 1. Query type is required
@@ -532,3 +544,8 @@ public interface IValidationRule
 
 It is recommended to inherit from `ValidationRuleBase` for custom validation rules
 and override only the methods you need to implement.
+
+### 13. New properties added to `IGraphType`, `IFieldType` and `IObjectGraphType`
+
+See the new features section for details on the new properties added to these interfaces.
+Unless you directly implement these interfaces, you should not be impacted by these changes.
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -367,6 +367,12 @@ ValueConverter.RegisterListConverterFactory(typeof(IList<>), typeof(List<>)); //
 
 Allows to set a graph type or field as private within a schema visitor, effectively removing it from the schema.
 Introspection queries will not be able to query the type/field, and queries will not be able to reference the type/field.
+Exporting the schema as a SDL (or printing it) will not include the private types or fields.
+
+Private types are fully 'resolved', and it is possible to reference the private types through the `SchemaTypes` dictionary.
+This makes it possible to create a private type used within the schema but not exposed to the client. For instance,
+it is possible to dynamically create input object types to deserialize GraphQL Federation entity representations, which are
+normally sent via the `_Any` type.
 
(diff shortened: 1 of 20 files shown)


Shane32:
I should probably add some tests before merging.

gao-artur:
Looks like maintenance hell...
I think I don't understand the purpose of these additions. The `IsPrivate`, for example. Is it to allow the object to be printed by the federation schema printer but excluded from the introspection?

Shane32:
No, it’s to remove the type from the schema. An alternative design would be to add some type of Remove functionality to SchemaTypes and Fields.

gao-artur:
I don't understand the logic behind these flags (not because they are bad but because I don't know all the library internals). But except that the `IsPrivate` is too viral I have no other concerns.

Shane32:
There are two specific use cases for GraphQL Federation, where types/fields must be removed after the schema is initiailized:

1. `_entities` and `_Entity` must be removed if there are not resolvable types.

If there are no resolvable types, per GraphQL Federation specification, the `_Entity` union must not be added to the schema.  I believe the implication is that the `_entities` field must also not be added to the schema.

PR #3921 adds `_Entity` to the schema prior to initialization, so that accessing `schema.AllTypes["_Entity"]` will return the type, and introspection works properly and etc.  Adding `_Entity` within the schema visitor is not possible because `schema.AllTypes` is a fixed dictionary.  If we skip that and just add the field directly to the Query type, while the query would execute federation requests properly, but introspection and schema printing would not function correctly.  Allowing types to be added after schema initialization is, I believe, a bad idea because none of the initialization code (e.g. setting `ResolvedType` properties) will run on the added type.  Which is why calling `RegisterType` throws an error now after initialization.

You might be thinking: could we add a way to add code that executes prior to initialization?  This is not possible as most of the types are implicitly discovered by `SchemaTypes`, rather than having been explicitly added to the schema.  And we cannot walk the schema without instantiating the discovered types to read their field lists.  Essentially we would have to duplicate all of the initialization logic again.

2. Type-first schemas leverages AutoRegistering helper code to create the federation entity resolver

Right now in PR #3921 I have this sample code:

```cs
[Key("id")]
public class User
{
    [Id]
    public int Id { get; set; }
    public required string Username { get; set; }

    [FederationResolver]
    public static Task<User?> GetUserByIdAsync([Id] int id, [FromServices] Data data)
    {
        return data.GetUserByIdAsync(id);
    }
}
```

I think the sample as shown above makes the most logical sense from a CLR standpoint, in conjunction with the existing way to use type-first; there is a static member to create itself.  In order to implement it, I am relying on the auto-registering graph code to build a 'fake' field resolver, which later we assign to the federation resolver.  In order to do so, I must add the field to the graph type and let `SchemaTypes` initialize it, or else each parameter will have unresolvable types assigned to them such as `GraphQLClrInputTypeReference<int>`.  After `SchemaTypes` initializes these fields, they can be discarded from the graph, only used by federation to resolve these types.  In the sample above, `IdGraphType` is explicitly specified, but even then, the code does not know if the user has overridden the default `IdGraphType` behavior within the schema configuration.  I often do so for my own schemas because I know that all ID values take a certain form (e.g. `Guid` or `int`) and I can provide input validation on these values (prior to request execution) if I replace `IdGraphType` with a custom implementation.  Again, letting `SchemaTypes` initialize this field solves the problem.  And finally, it is possible for `@key` to have some type of complex requirement, such as `id author { id }`.  Supporting `IsPrivate` solves this problem quite nicely, as again, all of the requisite types are initialized, but the ones unnecessary by the schema can be discarded.

### Alternate proposals

1. If we add a method to remove a type/field from an initialized schema, I could write a schema validation visitor to scan for types/fields for specific metadata, and call `SchemaTypes.Remove(type)` or `TypeFields.Remove(field)` to remove the type/field from the schema at that time.  This has a distinct disadvantage that the 'removal' capability is not exposed to users.  So in the complex key sample above (`id author { id }`), where type-first will generate an input type for `author` (but not hide it), it would be more difficult for a user to hide that type; they would have to write their own schema validation visitor.

Also note that with this solution, we cannot perform a schema initialization check to ensure all referenced types are not private.  Worse, we would have to deal with multithreading issues, where someone may try to remove a type, after initialization, while another thread is executing a request.  Ugh.  That alone might make this infeasible.

2. We could retain `IsPrivate`, but simply remove the types/fields automatically after schema initialization via an internal method.  This means that any references to these types/fields would need to be captured and stored somewhere by a custom schema validation visitor beforehand (which for #3921 probably already occurs).  So then requesting the type by name from `SchemaTypes` would fail for private types, and iterating `TypeFields` would not return any private fields. This would remove the need to add runtime checks against `IsPrivate` throughout the codebase.

I would endorse this alternative implementation method.

Shane32:
Closed in favor of #3949

PR #3949, opened 2024-06-02
Add IsPrivate and SkipTypeCheck properties to IGraphType / IFieldType

Shane32:
These are prerequisites for GraphQL Federation support in #3921 .  This PR targets v8.

## IsPrivate

Allows schema visitors to set `IGraphType.IsPrivate` or `IFieldType.IsPrivate` to remove a graph type or field from the schema after initialization.  Introspection queries cannot see these graph types or fields, and requests cannot reference them.  This is the only viable solution currently because `SchemaTypes` builds a read-only list during initialization, and similarly the fields cannot be removed from graph types.

## SkipTypeCheck

For schema-first schemas, skips `IsTypeOf` checks, while allowing to set `IsTypeOf` for other scenarios.

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -363,6 +363,28 @@ Finally, if you simply need to map an interface list type to a concrete list typ
 ValueConverter.RegisterListConverterFactory(typeof(IList<>), typeof(List<>)); // default mapping is T[]
 ```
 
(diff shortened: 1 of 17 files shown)


Shane32:
@gao-artur Still have not added tests but this should eliminate additional spaghetti-code throughout the codebase.

gao-artur:
Yeah, this approach looks muuuch better! 👍

gao-artur (approved):

PR #3953, opened 2024-06-02
Add OneOf Input Object support

Shane32:
## Summary

OneOf Input Objects are a special variant of Input Objects where the type system asserts that exactly one of the fields must be set and non-null, all others being omitted. This is useful for representing situations where an input may be one of many different options.

See proposed spec:
- https://github.com/graphql/graphql-spec/pull/825

## Todo

- [x] Add `IsOneOf` to `IInputObjectGraphType`
- [x] Add `isOneOf` to `__Type`
- [x] Append `@oneOf` to OneOf Input Object types when printing the SDL
- [x] Support schema-first schemas that use `@oneOf`
- [x] Ensure input object literals for OneOf Input Objects do not have multiple entries
- [x] Ensure input object values for OneOf Input Objects are not null
- [x] Ensure variables mapped to OneOf Input Objects have a single non-null entry
- [x] Add tests for validating input object literals
- [x] Add tests for validating input object variables
- [x] Add tests for various combinations as demonstrated in the RFC
- [x] Add tests for schema introspection
- [x] Add tests for SDL export
- [x] Add tests for schema-first importing
- [x] Add type-first attribute to mark an input type as a OneOf Input Object
- [x] Add migration notes


Shane32:
@gao-artur I’m not done yet but let me know if you have any comments. I think there’s enough for you to see where I’m aiming. I’ll request another review when finished. (Or you can wait until then.)

gao-artur:
This is great! It solves my [problem](https://github.com/graphql-dotnet/graphql-dotnet/pull/3857#issuecomment-1870921458) with "xor" argument validation 🎉

docs2/site/docs/migrations/migration8.md
```diff
@@ -389,6 +389,21 @@ property to `true` for all object graph types to retain the existing behavior.
...
+asserts that exactly one of the fields must be set and non-null, all others
+being omitted. This is useful for representing situations where an input may be
+one of many different options.
+
+See: https://github.com/graphql/graphql-spec/pull/825
+
+To use this feature, set the `IsOneOf` property on your `InputObjectGraphType` to `true`.
+
+Note: the feature is still a draft and has not made it into the official GraphQL spec yet.
+It is expected to do once it has been implemented in multiple libraries and proven to be useful.
```
Shane32, line 404:
```suggestion
It is expected to be added once it has been implemented in multiple libraries and proven to be useful.
```

Shane32:
Ok I think it's feature-complete and just needs tests added.

src/GraphQL/Utilities/Visitors/SchemaValidationVisitor.cs
```diff
@@ -237,6 +237,18 @@ public override void VisitInputObjectFieldDefinition(FieldType field, IInputObje
 
         if (field.Resolver != null)
             throw new InvalidOperationException($"The field '{field.Name}' of an Input Object type '{type.Name}' must not have Resolver set. You should set Resolver only for fields of object output types.");
+
+        //OneOf Input Objects
+        // RULE: If the original Input Object is a OneOf Input Object then:
+        // - All fields of the Input Object type extension must be nullable.
+        // - All fields of the Input Object type extension must not have default values.
```
gao-artur, line 244: - #3954

docs2/site/docs/migrations/migration8.md
```diff
@@ -389,6 +389,24 @@ property to `true` for all object graph types to retain the existing behavior.
...
+See: https://github.com/graphql/graphql-spec/pull/825
+
+To use this feature:
+- **Code-First**: Set the `IsOneOf` property on your `InputObjectGraphType` to `true`.
+- **Schema-First**: Use the `@oneOf` directive on the input type in your schema definition.
+- **Type-First**: Use the `[OneOf]` directive on the CLR class.
+
+Note: the feature is still a draft and has not made it into the official GraphQL spec yet.
+It is expected to be added once it has been implemented in multiple libraries and proven to be useful.
+It is not expected to change from the current draft.
```
gao-artur, line 408: Should it be behind the experimental switch then?

Shane32: I wouldn't.  They're going to merge it into the released GraphQL version now (see https://github.com/graphql/graphql-spec/pull/825#issuecomment-2148072272 ), and besides, so long as a user doesn't use the feature, then it's like it doesn't exist.  The only exception is if someone defined `@oneOf` in their own schema-first schema.  In which case it's going to break their schema soon enough.  But even if a code-first user defined their own `@oneOf`, it won't change anything because the flag on the input type won't be set.

Shane32: The feature, in its current state, is already being merged into numerous implementations.

- https://github.com/graphql/graphql-spec/pull/825#issuecomment-2135724148
- https://github.com/graphql/graphql-spec/pull/825#issuecomment-2143721271

gao-artur (approved):

docs2/site/docs/migrations/migration8.md
Shane32: Oh, and introspecting `__Type` will show a new field.

Shane32:
I'll wait a few days before merging, just in case someone else wants to look at the PR.

Shane32:
@gao-artur Based on the response to my comment in the RFC PR, I've updated this PR to require variable usages for OneOf Input Object fields to require non-null types.  To change it back would be non-breaking, so better to be conservative at this point.  Please review my latest commit - thanks!

See:
- https://github.com/graphql/graphql-spec/pull/825#issuecomment-2148965796

gao-artur:
Good catch!

gao-artur:
I believe Playground and other UI's don't support this feature yet and won't help with correct variable definition?

Shane32:
Probably. Chances are that they use GraphQL.js under the hood, and until GraphQL.js is released with OneOf support, we won’t see it in those tools.

PR #3957, opened 2024-06-07
Add GraphQL Federation v2 required types

Shane32:
Part 1 of #3921

### DIFF
--- a/src/GraphQL/Federation/Types/LinkImportGraphType.cs
+++ b/src/GraphQL/Federation/Types/LinkImportGraphType.cs
@@ -1,0 +1,15 @@
+namespace GraphQL.Federation.Types;
+
+/// <summary>
+/// Represents a scalar type for link imports in GraphQL Federation.
+/// Used to define imported directives.
+/// The name of this graph type is "link__Import".
+/// </summary>
+public class LinkImportGraphType : AnyScalarGraphType
+{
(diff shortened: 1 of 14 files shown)


src/GraphQL/Federation/Types/LinkImportGraphType.cs
```diff
@@ -0,0 +1,15 @@
+namespace GraphQL.Federation.Types;
+
+/// <summary>
+/// Represents a scalar type for link imports in GraphQL Federation.
+/// Used to define imported directives.
+/// The name of this graph type is "link__Import".
+/// </summary>
+public class LinkImportGraphType : AnyScalarGraphType
```
gao-artur, line 8: Why does it derive from Any and not String?

Shane32: https://specs.apollo.dev/link/v1.0/#Import

> [Import](https://specs.apollo.dev/link/v1.0/#Import) can take the form of a string ... or an object with name and (optionally as)

gao-artur (approved):

src/GraphQL/Federation/Types/LinkImportGraphType.cs
gao-artur: Shouldn't it have a validation for these two forms only?

PR #3958, opened 2024-06-07
Add federation metadata extension methods and attributes

Shane32:
Part 2 for #3921 

Does not include the directive definitions

- [x] Add documentation
- [x] Scope extension methods to the proper types
- [x] Add url links to Apollo docs


src/GraphQL/Federation/Extensions/FederationMetadataExtensions.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    /// <summary>
+    /// Adds the "@key" directive to a GraphQL type.
+    /// </summary>
+    /// <param name="graphType">The GraphQL type to which the directive is added.</param>
+    /// <param name="fields">An array of field names that form the key.</param>
+    /// <param name="resolvable">Indicates whether the key is resolvable. Default is true.</param>
+    /// <typeparam name="TMetadataWriter">The type of the metadata writer.</typeparam>
+    /// <returns>The modified metadata writer.</returns>
+    public static TMetadataWriter Key<TMetadataWriter>(this TMetadataWriter graphType, string[] fields, bool resolvable = true)
```
gao-artur, line 20: The use of extension methods in constructor looks somewhat ugly because it requires using `this.`. Are the any other alternatives?

gao-artur:
A lot of work for analyzers 🥴

Shane32:
I've got some ideas....

src/GraphQL/Federation/Extensions/FederationMetadataExtensions.cs
Shane32: > The use of extension methods in constructor looks somewhat ugly because it requires using `this.`. Are the any other alternatives?

The only other option is to add methods directly on the class.  This requires duplicating all of the methods that apply to graph types within `ComplexGraphType`.

Personally I'd like to treat federation as separate, as if it was in a separate NuGet package, which may be the case in the future.  In that vein, everything is in the GraphQL.Federation namespace and should be almost completely separable from the main GraphQL.NET codebase, similar to the design of the data loaders.

Shane32:
Ok the extension methods and attributes should all be scoped to the appropriate targets.

Shane32:
Have not yet added url links within the xml comments

src/GraphQL/Federation/Extensions/FederationMetadataExtensions.cs
Shane32: > I think it may be useful to add a link to the directive explanation in the XML docs https://www.apollographql.com/docs/federation/federated-types/federated-directives

done

src/GraphQL/Federation/Extensions/FederationMetadataExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// </summary>
+    /// <param name="graphType">The GraphQL type to which the directive is added.</param>
+    /// <param name="fields">An array of field names that form the key.</param>
+    /// <param name="resolvable">Indicates whether the key is resolvable. Default is true.</param>
+    /// <typeparam name="TMetadataWriter">The type of the metadata writer.</typeparam>
+    /// <returns>The modified metadata writer.</returns>
+    /// <remarks>
+    /// See <see href="https://www.apollographql.com/docs/federation/federated-types/federated-directives#key"/>.
+    /// </remarks>
+    public static TMetadataWriter Key<TMetadataWriter>(this TMetadataWriter graphType, string[] fields, bool resolvable = true)
```
gao-artur, line 23: This extension is useful to define compound or nested field keys. But I guess a single field as a key should be the most frequently used. Then need another extension based on `IFieldMetadataWriter`

gao-artur: Actually, FieldBuilder implements `IMetadataWriter`, so it works for all cases.

gao-artur: But it doesn't implement `IComplexGraphType` so it doesn't work...

Shane32: I don't follow you. `@key` is only valid on an object or interface, which is what `IComplexGraphType` is.

```gql
directive @key(fields: FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
```

https://www.apollographql.com/docs/federation/federated-types/federated-directives#key

Shane32: As is implemented now, these methods simply add directives to the schema.  They are not 'intelligent'; for example, you can't put `[Key]` on a field and have it automatically add the proper directive to the graph type.  While that's possible, it's not the current design.

So for `@key`, you can do `@key(fields: "id name pricing { id }")` by specifying `this.Key("id name pricing { id }")`

Shane32: Note that trying to make `@key` 'intelligent' presents a problem where the field case is not determined until the schema is initialized.  So when you write `Field(x => x.Id)` that will initially have an name of "Id" which would probably not be valid as an argument to `@key`.  It would need to be "id".  But for schemas using pascal case, it would need to be "Id".  So I decided to ignore that set of problems entirely.  And that ignores trying to handle complex keys with child properties.

gao-artur: I mean, probably, `@key(fields: "id")` is the most commonly used. Then I'd expect to be able to do this
```
Field(x => x.Id).Key();
```
And the field name will be taken from the FieldBuilder's `Name`.

Shane32: Note that technically `IInputObjectGraphType` inherits `IComplexGraphType` also.  I'm not too concerned about it; schema validation will ensure all directives are placed appropriately.  We could split the method, as I did with some of the other extension methods, having one target `IObjectGraphType` and the other target `IInterfaceGraphType`.  I would need to add another class.

Thoughts?

gao-artur: > It would need to be "id"

I believe people will be using `nameof()` to get the field name anyway, so this problem won't go anywhere

Shane32: Well, it would be nice. Maybe we could add a schema visitor to build the keys after the schema is initialized.

gao-artur: >having one target IObjectGraphType and the other target IInterfaceGraphType

Sounds better than implementing an analyzer to prevent incorrect usage. We can probably use the `[AllowedOn<>]`, but I'd prefer strong API over static analysis.

Shane32: In your example `Field(x => x.Id).Key();` doesn’t even have access to the graph type, so it would have to be something like a schema visitor.

gao-artur: I see. Then, it can be enchantment in 8.1 if you prefer.

gao-artur: There is still plenty of space for analyzers. For example, validating field names really exists. Or key field doesn't return union, interface or takes arguments.

Shane32: > I see. Then, it can be enchantment in 8.1 if you prefer.

Yes let’s just get a workable version pushed out

Shane32: > > having one target IObjectGraphType and the other target IInterfaceGraphType
> 
> Sounds better than implementing an analyzer to prevent incorrect usage. We can probably use the `[AllowedOn<>]`, but I'd prefer strongly typed API over static analysis.

I’ll make the change

Shane32: > > I see. Then, it can be enchantment in 8.1 if you prefer.
> 
> Yes let’s just get a workable version pushed out

Or at least merged in so we can review what further enhancements we want to make

gao-artur: Btw key on interface is only supported sine v2.3. Adding it now will require adding version detection/configuration + validation.

Shane32: Yeah, we need to work on that (not yet in #3921 )

Shane32: Generally, I'm planning to have the proper directives added based on the version.  Schema validation will already match up the applied directives to configured directives.  So it will catch this error if it's the wrong version.

Shane32: > > > having one target IObjectGraphType and the other target IInterfaceGraphType
> > 
> > 
> > Sounds better than implementing an analyzer to prevent incorrect usage. We can probably use the `[AllowedOn<>]`, but I'd prefer strongly typed API over static analysis.
> 
> I’ll make the change

Done

gao-artur (approved):

Issue #3954, opened 2024-06-03
Add static analyzer for OneOf


gao-artur:
Detect by:
 - InputObjectGraphType.IsOneOf = true;
 - [OneOfAttribute]

Rules:
 - all fields must be nullable (respect NRT)
 - No default values

gao-artur:
@Shane32 are these all possible nullability combinations?

Code first:
```c#
// Good
Field<StringGraphType>("name");
Field(x => x.Name, nullable: true);
Field(x => x.Name, type: typeof(StringGraphType));

// Bad
Field<NonNullGraphType<StringGraphType>>("name");
Field(x => x.Name);
Field(x => x.Name, nullable: false);
Field(x => x.Name, type: typeof(NonNullGraphType<StringGraphType>));
```

Type first:
```c#
[OneOf]
public class MyInput
{
    // Good
    public int? NullableValue { get; set; }
    // Bad
    public int NonNullableValue { get; set; }

#nullable disable
    // Good
    public string NullableRefValue1 { get; set; }

#nullable enable
    // Good
    public string? NullableRefValue2 { get; set; }

    // Bad
    public string NonNullableRefValue { get; set; } = null!;
}
```

Are there other options to control nullability?

Shane32:
Looks about right to me

PR #3967, opened 2024-06-23
OneOf analyzer

gao-artur:
Fixes 
- #3954 

Verify all `OneOf` fields are nullable and have no default values.
Support code-first and type-first.
Don't report when `[Ignore]` is used.


src/GraphQL.StarWars/.editorconfig
```diff
@@ -0,0 +1,2 @@
+[*.cs]
+resharper_virtual_member_call_in_constructor_highlighting = none
```
gao-artur, line 2: This is to avoid [Virtual member call in constructor](https://www.jetbrains.com/help/resharper/VirtualMemberCallInConstructor.html) warning by Resharper
image

src/GraphQL.Analyzers.Tests/OneOfAnalyzerTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                  public int? NullableField = {|#1:1|};
+
+              #nullable disable
+                  public string NullableRefProp1 { get; set; } = {|#2:"xxx"|};
+                  public string NullableRefField1 = {|#3:"xxx"|};
+
+              #nullable enable
+                  public string? NullableRefProp2 { get; set; } = {|#4:"yyy"|};
+                  public string? NullableRefField2 = {|#5:"yyy"|};
+              }
```
Shane32, line 193: Default values for properties defined in this manner are not copied to GraphQL (since reflection cannot read these values), and arguably should not cause a diagnostic.  Default values are defined as follows:

```cs
[DefaultValue(true)]
public bool Prop { get; set; }
```

gao-artur: Oh, good to know. Will fix.

Shane32:
Btw, merging with `develop` should clear up the commit history for this PR, I think...

gao-artur:
I created this branch from your `OneOf` branch before you merged it. Just didn't have time to finish the analyzer implementation until today. Need to rebase to clear the history but a little bit scary because I never did it but heard some scary tales 😅

Shane32:
It doesn’t matter. Ok with me if you just leave it. We squash and merge anyway so it won’t show up in the develop history.

Issue #3969, opened 2024-06-24
How to define field of type [string!] with expression


gao-artur:
I'm trying to use expressions instead of verbose syntax to describe the fields. How do I convert this definition to get `[string!]` type?
```c#
Field<ListGraphType<NonNullGraphType<StringGraphType>>>("Names").Resolve(x => x.Source.Names);
```

This leads to `[string]!`
```c#
Field(x => x.Names);
```

This to `[string]`
```c#
Field(x => x.Names, nullable: true);
```

I want a nullable list with non-nullable elements. The only option I found is defining the `type` parameter:
```c#
Field(x => x.Names, type: typeof(ListGraphType<NonNullGraphType<StringGraphType>>));
```

But it kind of misses the goal because typing-wise, I only save the `"Names"` string.

gao-artur:
Using NRT doesn't help either.

Shane32:
I don't have an answer here.  I run into similar issues using the ID graph type.  Usually I create an extension method within my application to facilitate the goal.  So for example:

```cs
this.IdField(x => x.Id);   // uses custom extension method
```

Shane32:
We could make this change for v8, so that by default all members directly referenced are analyzed for NRT.  But analyzing NRT isn't possible for computed fields or anything that's not a direct member access, as the generic type in the `Field<T>` will not carry any NRT data.  So the only way is to, like we do for name detection, look for directly referenced members (e.g. `x => x.Name`) and pull NRT data from the member.

Note that I expect this would be a very major breaking change.  One example:

```cs
class MyClass
{
    public string? Name { get; set; }
}

Field(x => x.Name);  // v7: non-null .... v8: nullable

// how to fix???   'nullable' argument already defaults to 'false'
```

We may need to change method signatures (e.g. `bool? nullable`), and we may need to design a global switch to revert behavior.

Keep in mind also that GraphQL encourages all fields to be nullable, so that errors do not bubble.  I don't design my graphs this way, but it certainly some developers do so, especially if they rely on errors not-bubbling in their client applications.

gao-artur:
Yes, I thought about all these problems as well. But they look solvable.
As you have said, the global switch can be introduced to revert the behavior. The default value from the `nullable` argument can be removed, and additional overload can be introduced without this argument. This overload will call the overload with `nullable` depending on NRT analysis + global switch value.
[...]

PR #3970, opened 2024-06-24
[WIP] Infer field nullability from NRT annotations

gao-artur:
POC.

Fixes
- #3969

The `Field` methods accepting `Expression` were changed to prevent incompatible configurations. It's not possible anymore to configure `nullable` when `type` is defined because it will be ignored anyway.

All the overloads preserve their old behavior except the one accepting only expression.
```c#
Field(x => x.Name);
```
In this case, the type and its nullability will be inferred from NRT annotations. The old behavior can be configured with the global switch:
```c#
GlobalSwitches.InferFieldNullabilityFromNRTAnnotations = false;
```
Or by passing explicit `false` to `nullable` parameter
```c#
Field(x => x.Name, nullable: false);
```

Note that most of the changes are source-compatible. The only scenario in which the user will need to adjust the code is when the `nullable` and `type` are defined together. The user will need to remove the `nullable` argument.

### DIFF
--- a/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
@@ -218,6 +218,7 @@ namespace GraphQL
         public static bool EnableReadDescriptionFromAttributes { get; set; }
         public static bool EnableReadDescriptionFromXmlDocumentation { get; set; }
         public static bool EnableReflectionCaching { get; set; }
+        public static bool InferFieldNullabilityFromNRTAnnotations { get; set; }
         [System.Obsolete("The query root operation type must be provided and must be an Object type. See ht" +
             "tps://spec.graphql.org/October2021/#sec-Root-Operation-Types")]
         public static bool RequireRootQueryType { get; set; }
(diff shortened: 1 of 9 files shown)


gao-artur:
@Shane32, please share your thoughts.

gao-artur:
Thinking again, I could implement it without changing methods signature when both `nullable` and `type` are `null`. Then it will be possible to deliver in v7 with a legacy flag.

Shane32:
> Note that most of the changes are source-compatible. The only scenario in which the user will need to adjust the code is when the `nullable` and `type` are defined together. The user will need to remove the `nullable` argument.

Can we add an obsolete member for this?

Shane32:
What about `Field("Total", x => x.Subtotal + x.Tax);` (for output types)?

Currently this should compile.  Seems like it will fail here with a cast failure to MemberExpression; please confirm.  I think it should fallback to the old behavior, which for value types will correctly infer the nullability, and for reference types will assume non-null.

Shane32:
In general, I think this is looking good...

gao-artur:
> > Note that most of the changes are source-compatible. The only scenario in which the user will need to adjust the code is when the `nullable` and `type` are defined together. The user will need to remove the `nullable` argument.
> 
> Can we add an obsolete member for this?

I think there will be some overloads hiding each other. Anyway, this is an illegal combination, that was possible because API allowed it. It should be a simple fix, so I'd prefer just to remove this overload instead of maintaining obsolete methods.

gao-artur:
> What about `Field("Total", x => x.Subtotal + x.Tax);` (for output types)?
> 
> Currently this should compile. Seems like it will fail here with a cast failure to MemberExpression; please confirm. I think it should fallback to the old behavior, which for value types will correctly infer the nullability, and for reference types will assume non-null.

You are right. Will fix it tomorrow.

Shane32:
> > > Note that most of the changes are source-compatible. The only scenario in which the user will need to adjust the code is when the `nullable` and `type` are defined together. The user will need to remove the `nullable` argument.
> > 
> > 
> > Can we add an obsolete member for this?
> 
> I think there will be some overloads hiding each other. Anyway, this is an illegal combination, that was possible because API allowed it. It should be a simple fix, so I'd prefer just to remove this overload instead of maintaining obsolete methods.

I know this may be a debatable point, but as a general rule I attempt to keep the GraphQL.NET basic methods (the ones that developers will use by the hundreds) stable, so they are not removed until around 2 years after they are marked obsolete, which usually is two major versions. With analyzers, assuming they work even when the method has been removed, we could reduce this to a single major version perhaps. But I cannot endorse removing the method when we can provide a method for backwards compatibility marked obsolete.

Shane32:
If we do provide the obsolete method signature as I suggested, does this code cause a problem with the compiler?  Assuming not...

```cs
Field(x => x.Name, type: typeof(StringGraphType));  //note the use of "type:"
```

If it does, then skip the obsolete member.

Issue #1290, opened 2019-09-05
FieldBuilder ignores nullable argument when it is passed along with type


vkukarau:
## Description
FieldBuilder ignores `nullable` argument if `type` is specified.

### Steps to reproduce
Define GraphQL type with fields as follows:
```
    public class AppointmentType : ObjectGraphType<Appointment>
    {
        public AppointmentType()
        {
            Field(x => x.Date, false, typeof(DateTimeGraphType));
            Field(x => x.Status, false, typeof(EnumerationGraphType<AppointmentStatus>));
            Field(x => x.Description, false, typeof(StringGraphType));
            Field(x => x.Title, false);
       }
    }   
```
Note that all fields are set as non-nullable, first 3 fields have `type` argument, and the last one doesn't have it.

### Expected result
Fields in generated GraphQL schema should look like:
```
date: DateTime!
status: AppointmentStatus!
description: String!
title: String!
```

### Actual result
Fields in generated GraphQL schema look like:
```
date: DateTime
status: AppointmentStatus
description: String
title: String!
```
Note that all fields are specified as non-nullable, but only `title` is properly defined in the resulting schema

### Environment
nuget GraphQL 2.4.0
Windows

joemcbride:
This is expected behavior.  If you are choosing to override the GraphType, that means you have full control over the exact definition.

MisterJimson:
While that makes sense, it's not a great developer experience to explicitly pass `false` to a parameter called 'nullable' and have the result not reflect that.

Do you think a reasonable change could be to wrap the type in a NonNullGraphType if `false` is set here to avoid this pitfall?

joemcbride:
`nullable` is an optional parameter to that method.

https://github.com/graphql-dotnet/graphql-dotnet/blob/444cc529a15526aa9b70376ec54bbe5b6d33fdcb/src/GraphQL/Types/ComplexGraphType.cs#L278-L282

What do you tell the person who then does this?

```csharp
Field(x => x.Description, type: typeof(NonNullGraphType<StringGraphType>));
```

And now ends up with:

```csharp
description: String!!
```

This would be a breaking behavior change.  Someone has to change their code.  Either you or them.

We could update the API so there are two methods, one that only accepts the type and one that only accepts the nullable, so there is never the combination.

MisterJimson:
Well if the type is already a `NonNullGraphType`, we wouldn't need to wrap it in a `NonNullGraphType` when nullable is false.

[...]

PR #3971, opened 2024-06-25
Obsolete Field method receiving expression and optional nullable and type arguments

gao-artur:
Fixes
- #1290

@Shane32 do you know if new methods can cause binary compatibility issues? The new overloads hide the old overload with both optional parameters. Should I target the `develop` branch instead?

### DIFF
--- a/src/GraphQL/Types/Composite/ComplexGraphType.cs
+++ b/src/GraphQL/Types/Composite/ComplexGraphType.cs
@@ -559,6 +559,7 @@ namespace GraphQL.Types
         /// <param name="expression">The property of the source object represented within an expression.</param>
         /// <param name="nullable">Indicates if this field should be nullable or not. Ignored when <paramref name="type"/> is specified.</param>
         /// <param name="type">The graph type of the field; if <see langword="null"/> then will be inferred from the specified expression via registered schema mappings.</param>
+        [Obsolete("Please use another overload that receives only one of the 'nullable' or 'type' arguments. This method will be removed in v9.")]
         public virtual FieldBuilder<TSourceType, TProperty> Field<TProperty>(
             string name,
             Expression<Func<TSourceType, TProperty>> expression,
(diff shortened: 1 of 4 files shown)


src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -559,6 +559,7 @@ public virtual FieldType AddField(FieldType fieldType)
         /// <param name="expression">The property of the source object represented within an expression.</param>
         /// <param name="nullable">Indicates if this field should be nullable or not. Ignored when <paramref name="type"/> is specified.</param>
         /// <param name="type">The graph type of the field; if <see langword="null"/> then will be inferred from the specified expression via registered schema mappings.</param>
+        [Obsolete("Please use another overload that receives only one of the 'nullable' or 'type' arguments. This method will be removed in v9.")]
```
gao-artur, line 562: I suggest making it private in v9 instead of removing it so there won't be code duplicates between other overloads.

Shane32:
> @Shane32 do you know if new methods can cause binary compatibility issues? The new overloads hide the old overload with both optional parameters. Should I target the `development` branch instead?

No; as long as the signature has not changed, there will be no binary compatibility issues.  Keep in mind that optional parameters are compiled into the caller's code, so:

```cs
// old
Field(x => x.Name);
// would be compiled as:
Field(x => x.Name, false, null); 
```

This means that if you change the default value, already-compiled code will not 'see' the change.  Also, if you change an optional parameter to be required, it does not break binary compatibility.  (This is a scenario where you could do that.)

So in the above sample, compiling the same old code with v7.9 would call a different method than previously.

gao-artur:
Sometimes, these tests fail in a local run with this error. Then, I need to rebuild to get them green. Do you have an idea why it happens?
```
System.IO.FileNotFoundException
Could not load file or assembly 'Microsoft.Bcl.AsyncInterfaces, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
   at GraphQL.Types.TypeInformation.ProcessTypeTree(List`1 typeTree, Boolean isInput)
   at GraphQL.Types.TypeInformation..ctor(MethodInfo methodInfo) in C:\github\graphql-dotnet\src\GraphQL\Types\TypeInformation.cs:line 151
   at GraphQL.Types.AutoRegisteringHelper.GetTypeInformation(MemberInfo memberInfo, Boolean isInputType) in C:\github\graphql-dotnet\src\GraphQL\Types\Composite\AutoRegisteringHelper.cs:line 283
   at GraphQL.Types.AutoRegisteringObjectGraphType`1.GetTypeInformation(MemberInfo memberInfo) in C:\github\graphql-dotnet\src\GraphQL\Types\Composite\AutoRegisteringObjectGraphType.cs:line 163
   at GraphQL.Types.AutoRegisteringHelper.CreateField(IGraphType graphType, MemberInfo memberInfo, Func`2 getTypeInformation, Action`2 buildFieldType, Boolean isInputType) in C:\github\graphql-dotnet\src\GraphQL\Types\Composite\AutoRegisteringHelper.cs:line 146
   at GraphQL.Types.AutoRegisteringObjectGraphType`1.CreateField(MemberInfo memberInfo) in C:\github\graphql-dotnet\src\GraphQL\Types\Composite\AutoRegisteringObjectGraphType.cs:line 106
   at GraphQL.Types.AutoRegisteringHelper.ProvideFields(IEnumerable`1 members, Func`2 createField, Boolean isInputType)+MoveNext() in C:\github\graphql-dotnet\src\GraphQL\Types\Composite\AutoRegisteringHelper.cs:line 263
   at GraphQL.Types.AutoRegisteringObjectGraphType`1..ctor(AutoRegisteringObjectGraphType`1 cloneFrom, Expression`1[] excludedProperties, Boolean cache) in C:\github\graphql-dotnet\src\GraphQL\Types\Composite\AutoRegisteringObjectGraphType.cs:line 57
   at GraphQL.Types.AutoRegisteringObjectGraphType`1..ctor(Expression`1[] excludedProperties) in C:\github\graphql-dotnet\src\GraphQL\Types\Composite\AutoRegisteringObjectGraphType.cs:line 38
   at GraphQL.Types.AutoRegisteringObjectGraphType`1..ctor() in C:\github\graphql-dotnet\src\GraphQL\Types\Composite\AutoRegisteringObjectGraphType.cs:line 31
```

image

Shane32:
No; it's happening for me as well.  We should probably dump .NET Core 3.1 testing for `develop`.

Shane32 (approved):

PR #3972, opened 2024-06-25
Add native support for @link

Shane32:
This does the following:
- Adds `schema.LinkSchema()` to import definitions from another schema.  All options from `@link` are available.  Automatically imports `@link` and supporting types also, if it is not already imported.
- Reads `@link` definitions applied to schema-first schemas.  Does not actually import the definitions; the SDL must still contain the definitions or the schema will not pass validation.
- Adds print option to remove imported definitions from being printed.  Useful for printing federation schemas.

Todo:
- [x] Add tests

Subsequent PR:
- [ ] Add schema visitor to rename types/directives that were imported to their imported alias.  Must do so for both the directive definitions and the applied directives.


src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -1695,22 +1696,6 @@ namespace GraphQL.Federation.Types
     {
         public FieldSetGraphType() { }
     }
-    public class LinkImportGraphType : GraphQL.Federation.Types.AnyScalarGraphType
```
Shane32, line 1698: These API changes are new for v8 anyway, so not breaking from v7.

Shane32:
@gao-artur This PR splits off `@link` support, and then we can separately integrate this into the Apollo Federation logic.  What do you think?

src/GraphQL/Extensions/SchemaExtensions.cs
```diff
@@ -405,6 +410,91 @@ private void Replace(IProvideResolvedType provider)
...
+                }
+            }
+        }
+        if (!linkInstalled && config.Url != LinkConfiguration.LINK_URL)
+        {
+            schema.AddLinkDirectiveSupport();
+        }
+
+        schema.ApplyDirective("link", config.ConfigureAppliedDirective);
+    }
```
gao-artur, line 497: I don't understand the code in this class. `AddLinkDirectiveSupport` is the only new public method, and it's called from the private `LinkSchema`, which is called from the public `AddLinkDirectiveSupport`...
Also I don't see any usage of the other private `LinkSchema` overload (with `url`).
If the `LinkConfiguration.LINK_URL` is the only URL currently supported, why is there an option to configure it in this version?
`LINK_URL` points to `v1.0`, so no support for `v2.0`?
I'm somewhat confused...

src/GraphQL/Extensions/SchemaExtensions.cs
```diff
@@ -405,6 +410,91 @@ private void Replace(IProvideResolvedType provider)
...
+        schema.Directives.Register(linkDirective);
+    }
+
+    /// <summary>
+    /// Adds a @link directive to the schema, specifying a URL and optional configuration for the link.
+    /// </summary>
+    /// <param name="schema">The schema to which the directive will be added.</param>
+    /// <param name="url">The URL of the linked schema.</param>
+    /// <param name="configuration">An optional action to configure the link further.</param>
+    private static void LinkSchema(this ISchema schema, string url, Action<LinkConfiguration>? configuration = null)
```
Shane32, line 462: This was supposed to be public.

src/GraphQL/Extensions/SchemaExtensions.cs
Shane32: See below. Only version that exists is 1.0 currently. 

https://specs.apollo.dev/link/v1.0/

gao-artur: It's a `@link` spec version, not a federation version 🤦‍♂️

gao-artur: Wait. Shouldn't it be a federation spec URL?
```
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3",
        import: ["@key", "@shareable"])
```
https://www.apollographql.com/docs/federation/federated-types/federated-directives/

src/GraphQL/Utilities/LinkConfiguration.cs
```diff
@@ -0,0 +1,298 @@
...
+    /// </summary>
+    public LinkConfiguration(string url)
+    {
+        Url = url;
+    }
+
+    /// <summary>
+    /// The URL of the linked schema.
+    /// </summary>
+    public string Url { get; set; }
```
gao-artur, line 26: If the URL is required via constructor, then this property probably should be get-only.

src/GraphQL/Utilities/LinkConfiguration.cs
```diff
@@ -0,0 +1,298 @@
...
+    /// <summary>
+    /// Configures the specified applied directive with the settings from this configuration.
+    /// </summary>
+    /// <remarks>
+    /// This is called by <see cref="SchemaExtensions.LinkSchema(ISchema, LinkConfiguration)"/> to apply the configuration to the schema.
+    /// </remarks>
+    internal void ConfigureAppliedDirective(AppliedDirective o)
+    {
+        if (o.Name != "link")
+            throw new ArgumentException("The supplied directive is not an @link directive.", nameof(o));
```
gao-artur, line 92:
```suggestion
            throw new ArgumentException("The supplied directive is not a @link directive.", nameof(o));
```

src/GraphQL/Extensions/SchemaExtensions.cs
Shane32: The goal is to support linking in the federation spec. First it links the link spec though because that’s not part of the federation spec. So each federated schema is going to have two links on it.

src/GraphQL/Utilities/LinkConfiguration.cs
Shane32: Ok I’ll change that.

src/GraphQL/Extensions/SchemaExtensions.cs
```diff
@@ -405,6 +410,102 @@ private void Replace(IProvideResolvedType provider)
...
+        }
+
+        var config = new LinkConfiguration(url);
+        configuration?.Invoke(config);
+        schema.ApplyDirective("link", config.ConfigureAppliedDirective);
+    }
+
+    /// <summary>
+    /// Gets the applied @link directives from the schema.
+    /// Results are only valid after the schema has been initialized.
```
gao-artur, line 500: Should throw then if `schema.Initialized == false`?

src/GraphQL/Utilities/LinkConfiguration.cs
```diff
@@ -0,0 +1,318 @@
...
+
+    /// <summary>
+    /// A dictionary of types and directives to import from the linked schema, specifying the alias to use for each imported member.
+    /// To use the imported member without an alias, specify the same name for the key and value.
+    /// </summary>
+    /// <remarks>
+    /// For example:
+    /// <code>
+    /// link.Imports.Add("@key", "@key");                 // imports @key without an alias
+    /// link.Imports.Add("@shareable", "@fed_shareable"); // imports @shareable as @fed_shareable
```
gao-artur, line 70: What if I import it with a double underscore `"@fed__shareable"`? Is it legal?

src/GraphQL/Utilities/LinkConfiguration.cs
```diff
@@ -0,0 +1,318 @@
...
+        if (Purpose != null)
+            o.AddArgument(new DirectiveArgument("purpose") { Value = Purpose.Value });
+
+        if (Imports.Count > 0)
+        {
+            var importList = new List<object>();
+            foreach (var import in Imports)
+            {
+                if (import.Key == null)
+                    throw new InvalidOperationException("No name specified for an import.");
```
gao-artur, line 141: `null` is an illegal key for dictionaries, so it can never happen

src/GraphQL/Utilities/LinkConfiguration.cs
```diff
@@ -0,0 +1,318 @@
...
+                var alias = import.Value ?? import.Key;
+                if (import.Key.StartsWith("@", StringComparison.Ordinal))
+                {
+                    NameValidator.ValidateName(import.Key.Substring(1), NamedElement.Directive);
+                    if (!alias.StartsWith("@", StringComparison.Ordinal))
+                        throw new InvalidOperationException("An alias for a directive must start with '@'.");
```
gao-artur, line 147: Please specify the directive name in the error message

src/GraphQL/Utilities/Visitors/RemoveImportedTypesVisitor.cs
```diff
@@ -0,0 +1,113 @@
...
+                }
+            }
+        }
+
+        var context = new Context
+        {
+            ImportedNamespaces = importedNamespaces,
+            ImportedTypes = importedTypes,
+        };
+        _instance.VisitAsync(node, context).GetAwaiter().GetResult();
```
gao-artur, line 52: image

docs2/site/docs/migrations/migration8.md
```diff
@@ -617,6 +617,75 @@ public class Widget
...
 
+### 18. Support added for the Apollo `@link` directive
+
+This directive indicates that some types and/or directives are to be imported from another schema.
+Types and directives can be explicitly imported, either with their original name or with an alias.
+Any types or directives that are not explicitly imported will be assumed to be named with a specified
+namespace, which is derived from the URL of the linked schema if not set explicitly.
+Visit https://specs.apollo.dev/link/v1.0/ for more information.
+
+To link another schema, use this code like this in your schema constructor or `ConfigureSchema` call:
```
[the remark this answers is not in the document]
Shane32:
```suggestion
To link another schema, use code like this in your schema constructor or `ConfigureSchema` call:
```

src/GraphQL/Extensions/SchemaExtensions.cs
Shane32: I added checks

src/GraphQL/Utilities/LinkConfiguration.cs
Shane32: Yes, as far as I can tell, it is legal.

src/GraphQL/Utilities/LinkConfiguration.cs
Shane32: Removed null check

src/GraphQL/Utilities/LinkConfiguration.cs
Shane32: Updated

src/GraphQL/Utilities/Visitors/RemoveImportedTypesVisitor.cs
Shane32: `RemoveImportedTypesVisitor` is a sealed class and none of its methods call truly asynchronous methods.  In other words, it will always run synchronously.  Use of `.GetAwaiter().GetResult()` will unwrap any exceptions if they were to occur.

PR #3974, opened 2024-06-26
[WIP] Infer nullability from NRT

gao-artur:
Replaces #3970.

With this implementation, we can also use a legacy flag to deliver the new feature in v7

### DIFF
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -163,14 +163,62 @@ public class ComplexGraphTypeTests
     [Fact]
     public void accepts_property_expressions()
     {
+        bool old = GlobalSwitches.InferFieldNullabilityFromNRTAnnotations;
+        GlobalSwitches.InferFieldNullabilityFromNRTAnnotations = false;
+
+        try
+        {
+            var schema = new Schema();
+            var type = new ComplexType<Droid>();
+            _ = type.Field(d => d.Name);
+            schema.Query = type;
+            schema.Initialize();
+
+            type.Fields.Last().Name.ShouldBe("name");
+            type.Fields.Last().Type.ShouldBe(typeof(NonNullGraphType<StringGraphType>));
+        }
+        finally
+        {
+            GlobalSwitches.InferFieldNullabilityFromNRTAnnotations = old;
+        }
+    }
+
+    [Fact]
+    public void infer_nullability_from_nrt()
+    {
         var schema = new Schema();
-        var type = new ComplexType<Droid>();
-        _ = type.Field(d => d.Name);
+        var type = new ComplexType<NrtTest>();
+        _ = type.Field(d => d.Str1);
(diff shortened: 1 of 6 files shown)


src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -162,15 +162,63 @@ public void auto_register_input_object_graph_type()
 
     [Fact]
     public void accepts_property_expressions()
+    {
+        bool old = GlobalSwitches.InferFieldNullabilityFromNRTAnnotations;
+        GlobalSwitches.InferFieldNullabilityFromNRTAnnotations = false;
```
gao-artur, line 167: TODO: move to `"StaticTests"`

Shane32: lol - i checked already - `ComplexGraphTypeTests` is already marked with `[Collection("StaticTests")]`

Shane32:
> Replaces #3970.
> 
> With this implementation, we can also use a legacy flag to deliver the new feature in v7

Let's just have a PR target `master`, then after it's merged, add a PR to change the default flag (and add migration notes) for v8.

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -561,15 +561,30 @@ public virtual FieldType AddField(FieldType fieldType)
...
     {
         try
         {
-            if (type == null)
-                type = typeof(TProperty).GetGraphTypeFromType(nullable, this is IInputObjectGraphType ? TypeMappingMode.InputType : TypeMappingMode.OutputType);
+            if (type == null && nullable == null && GlobalSwitches.InferFieldNullabilityFromNRTAnnotations)
+            {
+                var typeInfo = AutoRegisteringHelper.GetTypeInformation(((MemberExpression)expression.Body).Member, this is IInputObjectGraphType);
+                type = typeInfo.ConstructGraphType();
+            }
```
Shane32, line 582: I think we need a little more code in here, along these lines -- see example below: (unfinished)

```suggestion
            if (type == null && nullable == null && GlobalSwitches.InferFieldNullabilityFromNRTAnnotations)
            {
                if (expression.Body is MemberExpression memberExpression)
                {
                    var typeInfo = AutoRegisteringHelper.GetTypeInformation(memberExpression.Member, this is IInputObjectGraphType);
                    type = typeInfo.ConstructGraphType();
                }
                else if (typeof(TProperty).IsValueType)
                {
                    nullable = Nullable.GetUnderlyingType(typeof(TProperty)) != null;
                    // call GetGraphTypeFromType
                }
                else // non-member expression such as: x => "Sample"
                {
                    nullable = false; // fallback to legacy behavior
                    // call GetGraphTypeFromType
                }
            }
```

Shane32:
@gao-artur  This PR is closed so I'm deleting the branch.  You can always restore it if need be.  Ditto with other old branches.

PR #3975, opened 2024-06-28
Allow applied directives to be imported from another schema

Shane32:
Now that schemas can be linked, the next step is to make it so that directives that are imported get automatically renamed properly.  This PR adds `FromSchemaUrl` to `AppliedDirective` to allow specifying the schema url that the directive is imported from.  During schema initiailization, any directive that has a `FromSchemaUrl` that matches an `@link` directive will get the proper alias applied to it.  By design, if no match, it is unaffected.  All Federation directives have been updated.

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -688,7 +688,7 @@ options to the same url, or with a separate URL to link multiple schemas.
 
 ### 19. `FromSchemaUrl` added to `AppliedDirective`
 
-This supports using a directive that was separately imported via `@link`. After importing the schema as described
+This property supports using a directive that was separately imported via `@link`. After importing the schema as described
 above, apply imported directives to your schema similar to the example below:
 
 ```csharp
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -686,6 +686,21 @@ schema
 Note that you may call `LinkSchema` multiple times with the same URL to apply additional configuration
 options to the same url, or with a separate URL to link multiple schemas.
 
+### 19. `FromSchemaUrl` added to `AppliedDirective`
+
+This supports using a directive that was separately imported via `@link`. After importing the schema as described
+above, apply imported directives to your schema similar to the example below:
+
+```csharp
+graphType.ApplyDirective("shareable", s => s.FromSchemaUrl = "https://specs.apollo.dev/federation/"); // applies to any version
+// or
+graphType.ApplyDirective("shareable", s => s.FromSchemaUrl = "https://specs.apollo.dev/federation/v2.3"); // only version 2.3
+```
+
+During schema initialization, the name of the applied directive will be resolved to the fully-qualified name.
+In the above example, if `@shareable` was imported, the directive will be applied as `@shareable`, but if not, it will
+be applied as `@federation__shareable`. Aliases are also supported.
+
 ## Breaking Changes
 
 ### 1. Query type is required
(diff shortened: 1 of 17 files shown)


src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -321,6 +321,25 @@ protected FieldBuilder(FieldType fieldType)
...
     }
 
+    /// <summary>
+    /// Adds an argument to the field.
+    /// </summary>
+    /// <param name="type">The graph type of the argument.</param>
+    /// <param name="name">The name of the argument.</param>
+    /// <param name="configure">A delegate to further configure the argument.</param>
+    [AllowedOn<IObjectGraphType, IInterfaceGraphType>]
+    public virtual FieldBuilder<TSourceType, TReturnType> Argument(IGraphType type, string name, Action<QueryArgument>? configure = null)
```
Shane32, line 331: This overload was missing -- explicit graph type instance rather than CLR type or typeof(T) for graph type.  I've run into this issue before.  So I added it.  We have methods like this for `Field` so it seems that it was simply an oversight.

Shane32: The only question is the order of parameters.  To match the existing `Argument(Type type, string name)` I did it this way.

src/GraphQL/Extensions/SchemaExtensions.cs
```diff
@@ -484,7 +484,7 @@ public static void LinkSchema(this ISchema schema, string url, Action<LinkConfig
                         link.ConfigureAppliedDirective(appliedDirective);
                         return;
                     }
-                    if (url == LinkConfiguration.LINK_URL)
+                    if (urlMatch == LinkConfiguration.LINK_URL)
```
Shane32, line 487: Bugfix

docs2/site/docs/migrations/migration8.md
```diff
@@ -686,6 +686,21 @@ schema
 Note that you may call `LinkSchema` multiple times with the same URL to apply additional configuration
 options to the same url, or with a separate URL to link multiple schemas.
 
+### 19. `FromSchemaUrl` added to `AppliedDirective`
+
+This supports using a directive that was separately imported via `@link`. After importing the schema as described
```
gao-artur, line 691: The beginning of the sentence seems to miss some words. Did you mean to say "This property supports"?

src/GraphQL/Builders/FieldBuilder.cs
gao-artur: It could be better to have `name` as a first argument to be consistent with a `Field` methods, but it will require changing the arguments order in `Argument(Type graphType, string name)`. Looks good to me.

src/GraphQL/Utilities/Visitors/RenameImportedDirectivesVisitor.cs
```diff
@@ -0,0 +1,93 @@
...
+    public override void VisitUnion(UnionGraphType type, ISchema schema) => VisitNode(type);
+
+    /// <summary>
+    /// Scan all applied directives to see if any were imported from another schema, and if so,
+    /// rename them with the appropriate alias as defined by the @link directive.
+    /// </summary>
+    private void VisitNode(IMetadataReader metadataReader)
+    {
+        var appliedDirectives = metadataReader.GetAppliedDirectives();
+        if (appliedDirectives == null || appliedDirectives.List == null)
```
gao-artur, line 75:
```suggestion
        if (appliedDirectives?.List == null)
```

docs2/site/docs/migrations/migration8.md
Shane32:
```suggestion
This property supports using a directive that was separately imported via `@link`. After importing the schema as described
```

gao-artur (approved):

PR #3976, opened 2024-07-01
Infer field nullability from NRT

gao-artur:
Replaces #3974.
Fixes
- #3969

### DIFF
--- a/docs2/site/docs/migrations/migration7.md
+++ b/docs2/site/docs/migrations/migration7.md
@@ -290,6 +290,53 @@ Please note that the `SchemaExporter` class will export type defintions as type 
 
 Please see the `SchemaExtensions.PrintAsync` source code implementation specifics.
 
+### 14. Infer field nullability from NRT annotations
+
+Since 7.9 you can enable the inference of the field nullability from the Null Reference Types (NRT)
+annotations when defining the field with expression. To enable the feature set the  
+`GlobalSwitches.InferFieldNullabilityFromNRTAnnotations` to `true`.  
+
+For example, given the following code
+
+```c#
+public class Person
+{
+    public string FullName { get; set; }
+    public string? SpouseName { get; set; }
+    public IList<string>? Children { get; set; }
+}
+
+public class PersonGraphType : ObjectGraphType<Person>
+{
+    public PersonGraphType()
+    {
+        Field(p => p.FullName);
+        Field(p => p.SpouseName);
+        Field(p => p.Children);
+    }
+}
+```
+
+When `InferFieldNullabilityFromNRTAnnotations` is `false` (default), the result is:
+
+```graphql
+type Person {
+  fullName: String!
+  spouseName: String!
+  children: [String]!
+}
+```
+
+When `InferFieldNullabilityFromNRTAnnotations` is `true`:
+
+```graphql
+type Person {
+  fullName: String!
+  spouseName: String
+  children: [String!]
+}
+```
+
 ## Breaking Changes
 
 ### 1. `DataLoaderPair<TKey, T>.Loader` property removed
--- a/src/GraphQL.Tests/Subscription/ShouldlyExtensions.cs
+++ b/src/GraphQL.Tests/Subscription/ShouldlyExtensions.cs
@@ -1,4 +1,5 @@
 using GraphQL.SystemTextJson;
+using Should = Shouldly.Should;
 
 namespace GraphQL.Tests.Subscription;
 
@@ -56,5 +57,17 @@ public static class ShouldlyExtensions
         result.Errors.ShouldBeNull();
         result.Executed.ShouldBeTrue();
         return result;
+    }
+
+    public static void ShouldBeWhen(this object? obj, bool condition, object onTrue, object onFalse) =>
+        obj.ShouldBe(condition ? onTrue : onFalse);
+
+    public static void ShouldThrowWhen<TException>(bool condition, Action action)
+        where TException : Exception
+    {
+        if (condition)
+            Should.Throw<TException>(action);
+        else
+            Should.NotThrow(action);
     }
 }
--- a/src/GraphQL.Tests/Subscription/ShouldlyExtensions.cs
+++ b/src/GraphQL.Tests/Subscription/ShouldlyExtensions.cs
@@ -1,5 +1,4 @@
 using GraphQL.SystemTextJson;
-using Should = Shouldly.Should;
 
 namespace GraphQL.Tests.Subscription;
 
@@ -57,17 +56,5 @@ public static class ShouldlyExtensions
         result.Errors.ShouldBeNull();
         result.Executed.ShouldBeTrue();
         return result;
-    }
-
-    public static void ShouldBeWhen(this object? obj, bool condition, object onTrue, object onFalse) =>
-        obj.ShouldBe(condition ? onTrue : onFalse);
-
-    public static void ShouldThrowWhen<TException>(bool condition, Action action)
-        where TException : Exception
-    {
-        if (condition)
-            Should.Throw<TException>(action);
-        else
-            Should.NotThrow(action);
     }
 }
--- a/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
@@ -163,14 +163,61 @@ public class ComplexGraphTypeTests
     [Fact]
     public void accepts_property_expressions()
     {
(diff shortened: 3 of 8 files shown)


src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
```diff
@@ -551,6 +561,85 @@ public void create_fieldbuilder_with_inferred_graph_type()
...
+            ShouldlyExtensions.ShouldThrowWhen<ArgumentException>(!infer, () => type.Field(d => d.Int2));
+
+            schema.Query = type;
+            schema.Initialize();
+
+            var field = type.Fields.FirstOrDefault(f => f.Name == "str1").ShouldNotBeNull();
+            field.Type.ShouldBe(typeof(NonNullGraphType<StringGraphType>));
+
+            field = type.Fields.FirstOrDefault(f => f.Name == "str2").ShouldNotBeNull();
+            field.Type.ShouldBeWhen(infer, typeof(StringGraphType), typeof(NonNullGraphType<StringGraphType>));
```
gao-artur, line 593: I think without the ability to gradually migrate the schema type-by-type, this feature will stay disabled forever for large schemas...

src/GraphQL/Types/Composite/ComplexGraphType.cs
```diff
@@ -564,12 +564,54 @@ public virtual FieldType AddField(FieldType fieldType)
...
+        /// <param name="nullable">Indicates if this field should be nullable or not. Ignored when <paramref name="type"/> is specified.</param>
+        /// <param name="type">The graph type of the field; if <see langword="null"/> then will be inferred from the specified expression via registered schema mappings.</param>
+        /// <remarks>
+        /// When <paramref name="nullable"/> and <paramref name="type"/> are both <see langword="null"/>
+        /// the field's nullability depends on <see cref="GlobalSwitches.InferFieldNullabilityFromNRTAnnotations"/> and <paramref name="expression"/> type.
+        /// When set to <see langword="true"/> and expression is <see cref="MemberExpression"/>,
+        /// the result field nullability will match the Null Reference Type annotations of the member represented by the expression.
+        /// If expression is not <see cref="MemberExpression"/> and <typeparamref name="TProperty"/> is of value type
+        /// the graph type nullability will match the <typeparamref name="TProperty"/> nullability. Otherwise, the field will be not nullable.
+        /// </remarks>
```
gao-artur, line 588: Please review the `remarks` section

src/GraphQL.Tests/Subscription/ShouldlyExtensions.cs
```diff
@@ -57,4 +58,16 @@ public static ExecutionResult ShouldBeSuccessful(this ExecutionResult result)
         result.Executed.ShouldBeTrue();
         return result;
     }
+
+    public static void ShouldBeWhen(this object? obj, bool condition, object onTrue, object onFalse) =>
+        obj.ShouldBe(condition ? onTrue : onFalse);
```
Shane32, line 63: Why?

```cs
obj.ShouldBeWhen(condition, trueObj, falseObj);
// vs
obj.ShouldBe(condition ? trueObj : falseObj);
```
Seems the second option is at least as preferable as the first, especially since `ShouldBeWhen` isn't a generic method.  It's at least as readable, perhaps more, and it's shorter.

src/GraphQL.Tests/Subscription/ShouldlyExtensions.cs
```diff
@@ -57,4 +58,16 @@ public static ExecutionResult ShouldBeSuccessful(this ExecutionResult result)
...
+
+    public static void ShouldThrowWhen<TException>(bool condition, Action action)
```
Shane32, line 65: I wouldn't really do this one either, personally, mainly because it can't be an extension method.  So we can't use `Should.Thow` like we can anywhere else.  But ok if you want to for your tests.

I would comment that typically within GraphQL.NET we examine the message from the exception to verify that it is the one intended.  I'm not firm on this (just stating what the existing code is like), but you could modify `void` to be `TException?` along with an annotation that indicates it's not-null when `condition` is `true`.  Then it could be used like this:

```cs
ShouldThrowWhen(condition, () => DoStuff())
    ?.Message.ShouldBe("abcd");
```

Personally I think verifying the exception type is often suitable, but sometimes verifying the message is important when the exception type is rather generic.

src/GraphQL.Tests/Subscription/ShouldlyExtensions.cs
```diff
@@ -57,4 +58,16 @@ public static ExecutionResult ShouldBeSuccessful(this ExecutionResult result)
...
+        where TException : Exception
+    {
+        if (condition)
+            Should.Throw<TException>(action);
+        else
+            Should.NotThrow(action);
```
Shane32, line 71: I would not call `Should.NotThrow`; I would call the action directly.  If it throws, the stack trace and error will show up in the test explorer.  But if you have a reason for doing so, that's fine; whatever you think.

src/GraphQL.Tests/Subscription/ShouldlyExtensions.cs
```diff
@@ -57,4 +58,16 @@ public static ExecutionResult ShouldBeSuccessful(this ExecutionResult result)
...
+    }
```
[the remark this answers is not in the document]
Shane32: Nope

Shane32 (approved):

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
gao-artur: I think you are right. If NRT is enabled and user has a `string?`, for example, they probably use `nullable: true` to set the correct field nullability. Enabling this feature will change nothing in this case. If they have a non-nullable `string`, then nothing changes again. And if they have NRT disabled, why would they want to enable this feature?

docs2/site/docs/migrations/migration7.md
```diff
@@ -290,6 +290,53 @@ Please note that the `SchemaExporter` class will export type defintions as type
...
+When `InferFieldNullabilityFromNRTAnnotations` is `true`:
+
+```graphql
+type Person {
+  fullName: String!
+  spouseName: String
+  children: [String!]
+}
+```
+
```
gao-artur, line 339: @Shane32 please review migration notes update

Shane32: Migration notes look perfect! I do suggest we move/copy these to the 8.0 migration notes in a separate PR targeting develop, along with changing the default value and appropriate changes to the notes to reflect this. 

This really is more of a new feature for 8.0, backported to 7.9, and I’d like to emphasis this feature in the new features section of 8.0. Decent chance are that both 7.9 and 8.0 will be released together. Or that any user upgrading to 7.9 is probably just doing so because of dependabot and not reading any migration notes since they are already on 7.x. 

Sometimes I do add minor version new features to the migration notes but usually only when it’s early in the major version’s lifecycle.

src/GraphQL.Tests/Types/ComplexGraphTypeTests.cs
Shane32: Yup. And if they wanted all nullable fields, as some devs do to prevent null bubbling, again they would have nullable: true so the feature can be safely enabled.

SimonCropp:
how do i use this feature with an async resolve?

Shane32:
Not really applicable, although you certainly can use ResolveAsync on the returned builder method if you like.  This only works when using the expression syntax to define a field, and only when the expression syntax is a simple pointer to a member, which usually implies that you plan to return or populate the field, cases where a field resolver is not used.

SimonCropp:
but from what i can see there is no code path where we can get the null infer to work without having to provide a sync resolve

Shane32:
I don’t think you understand. Nullability can’t be inferred from a lambda. The compiler doesn’t store the annotations anywhere so it’s not possible to write code to infer them. Sync/async doesn’t matter. You can’t get it to work with a synchronous resolver any better than an asynchronous resolver. 

This PR reads expression trees to find a pointer to a member from which it can read the NRT attributes. Properties and fields are not typically asynchronous.  This PR can’t read nullability for anything more than a simple property or field pointer, since expression trees also do not store nullability information.

Shane32:
The AutoRegisteringObjectGraphType code exclusively creates fields from member signatures. So it can and does infer nullability for synchronous and asynchronous members alike. 

There was a FieldDelegate method. Not sure if it still exists. Potentially it could be programmed to infer nullability from the delegate that it pointed to. I don’t think this is a typical programming pattern however.

SimonCropp:
@Shane32 ok. thanks for the clarification

PR #3977, opened 2024-07-02
Infer nullability from NRT by default

gao-artur:


docs2/site/docs/migrations/migration8.md
```diff
@@ -969,3 +1016,50 @@ Unless you directly implement this interface, you should not be impacted by this
 This is required for OneOf Input Object support and is used to determine if a variable is required.
 Unless you have a custom validation rule that uses `VariableUsage`, you should not be impacted
 by this change.
+
+### 20. Infer field nullability from NRT annotations is enabled by default
+
+When defining the field with expression, the graph type nullability will be inferred from
```
Shane32, line 1022: Can we shorten this to a sentence and reference the new feature? Perhaps here would be a great place to add a sample line of code to switch back to the old default however.

Shane32 (approved):

PR #3982, opened 2024-07-09
Use ThreadStatic for cached fields

Shane32:
The `IResolveFieldContext` instance and dictionary are heavily used and so currently use `Interlocked.Exchange` to reuse a cached instance.  The instance is cached per `ExecutionContext` so that concurrent executions each have their own shared field.  This enhances the concept so that it is per thread rather than per execution.  Performance shows nearly the same results with existing test suite (which execute only synchronous fields), but the code is better organized.

Before:

| Method        | UseCaching | Mean       | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|-------------- |----------- |-----------:|----------:|----------:|-------:|-------:|----------:|
| Introspection | False      | 425.171 us | 8.1294 us | 9.6774 us | 9.7656 | 2.9297 | 187.21 KB |
| Hero          | False      |   9.897 us | 0.0465 us | 0.0435 us | 0.3357 |      - |   6.37 KB |
| Introspection | True       | 249.638 us | 1.6388 us | 1.4528 us | 7.8125 | 1.9531 | 144.19 KB |
| Hero          | True       |   1.853 us | 0.0214 us | 0.0190 us | 0.1354 |      - |   2.52 KB |

After:

| Method        | UseCaching | Mean       | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|-------------- |----------- |-----------:|----------:|----------:|-------:|-------:|----------:|
| Introspection | False      | 411.722 us | 1.7013 us | 1.5914 us | 9.7656 | 2.4414 | 185.97 KB |
| Hero          | False      |   9.893 us | 0.0517 us | 0.0459 us | 0.3204 |      - |   6.07 KB |
| Introspection | True       | 243.321 us | 1.4652 us | 1.3705 us | 7.3242 | 1.9531 | 142.95 KB |
| Hero          | True       |   1.818 us | 0.0325 us | 0.0304 us | 0.1183 |      - |   2.23 KB |

### DIFF
--- a/src/GraphQL/Execution/ExecutionStrategy.cs
+++ b/src/GraphQL/Execution/ExecutionStrategy.cs
@@ -140,6 +140,9 @@ public abstract class ExecutionStrategy : IExecutionStrategy
         return true;
     }
 
+    [ThreadStatic]
+    private static Dictionary<string, (GraphQLField, FieldType)>? _reusableFields;
+
     /// <summary>
     /// Creates execution nodes for child fields of an object execution node. Only run if
     /// the object execution node result is not <see langword="null"/> or object execution
@@ -150,23 +153,30 @@ public abstract class ExecutionStrategy : IExecutionStrategy
         Debug.Assert(parent.Result != null || parent is RootExecutionNode);
 
         var parentType = parent.GetObjectGraphType(context.Schema) ?? parent.GraphType;
-        var fields = Interlocked.Exchange(ref context.ReusableFields, null);
-        fields = CollectFieldsFrom(context, parentType, parent.SelectionSet!, fields);
-
-        var subFields = new ExecutionNode[fields.Count];
-
-        int i = 0;
-        foreach (var kvp in fields)
+        var fields = _reusableFields ??= new();
+        try
         {
-            (var field, var fieldDefinition) = kvp.Value;
-            var node = BuildExecutionNode(parent, fieldDefinition.ResolvedType!, field, fieldDefinition);
-            subFields[i++] = node;
+            fields = CollectFieldsFrom(context, parentType, parent.SelectionSet!, fields);
+
+            var subFields = new ExecutionNode[fields.Count];
+
+            int i = 0;
+            foreach (var kvp in fields)
+            {
+                (var field, var fieldDefinition) = kvp.Value;
+                var node = BuildExecutionNode(parent, fieldDefinition.ResolvedType!, field, fieldDefinition);
+                subFields[i++] = node;
+            }
+
+            parent.SubFields = subFields;
+
+            fields.Clear();
         }
-
-        parent.SubFields = subFields;
-
-        fields.Clear();
-        Interlocked.CompareExchange(ref context.ReusableFields, fields, null);
+        catch
+        {
+            _reusableFields = null;
+            throw;
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -150,23 +153,30 @@ protected virtual void SetSubFieldNodes(ExecutionContext context, ObjectExecutio
         Debug.Assert(parent.Result != null || parent is RootExecutionNode);
 
         var parentType = parent.GetObjectGraphType(context.Schema) ?? parent.GraphType;
-        var fields = Interlocked.Exchange(ref context.ReusableFields, null);
-        fields = CollectFieldsFrom(context, parentType, parent.SelectionSet!, fields);
+        var fields = _reusableFields ??= new();
```
gao-artur, line 156: Will this work better?
```c#
[ThreadStatic]
private static Dictionary<string, (GraphQLField, FieldType)> _reusableFields = new();

protected virtual void SetSubFieldNodes(ExecutionContext context, ObjectExecutionNode parent)
{
    var fields = _reusableFields;
    try
    {
        ...
    }
    finally
    {
        fields.Clear()
    }
}

File: src/GraphQL/Execution/ExecutionStrategy.cs [hunk 2]
```diff
@@ -442,6 +452,9 @@ async Task SetArrayItemNode(object? d)
     protected virtual IFieldResolver SelectResolver(ExecutionNode node, ExecutionContext context)
         => node.FieldDefinition.Resolver ?? (node.Parent is RootExecutionNode && context.Operation.Operation == OperationType.Subscription ? SourceFieldResolver.Instance : NameFieldResolver.Instance);
 
+    [ThreadStatic]
+    private static ReadonlyResolveFieldContext? _readonlyResolveFieldContext;
```
gao-artur, line 456: Don't you think `ObjectPool` will give better results? With `ThreadStatic`, you can reuse no more than number-of-threads objects, while there might be order of magnitude more concurrent tasks that need the context objects. It also can happen that one instance was promoted to a higher Gen but will be thrown out on the next cycle because another instance, probably from a younger generation, took its place. Then this young instance will be promoted to higher gen and thrown out again.

Shane32: No I don’t think so. At least, not here. The object in question only contains a two (?) values and is probably the minimum size of any allocation.  So it’s the same as any lambda or Task instance. ObjectPool is costly and only recommended when the cost to allocate is high.  And if the user needs to run asynchronous code to execute this field, one more allocation won’t perceptibly change the speed. ThreadStatic is merely to help when resolving many scalars from a source object - e.g. x => x.Name - which always happens synchronously. 

I know I’ve seen performance benchmarks for ObjectPool in the past but I can’t find it now.

src/GraphQL/Execution/ExecutionStrategy.cs
Shane32: ThreadStatic fields cannot have initializers. It would compile, but they will only run once and all other threads will not be initialized.

src/GraphQL/Execution/ExecutionStrategy.cs
gao-artur: Actually, I talked more about memory allocations. The `ReadonlyResolveFieldContext` is needed per resolved field, right? When resolving many fields in parallel, using ObjectPool can reduce the amount of work for GC. Of course, the pool size should be limited, and maybe a custom policy is needed to sharink it after some idle time.

Shane32: Yes, per resolved field. But multiple are only needed for resolving many asynchronous fields in parallel. If the field doesn’t await a truly asynchronous method (like calling a database), fields are executed serially even with ParallelExecutionStrategy.

Shane32: There is one benefit to the current design: for serial execution strategy, the context will always be re used because it has a field per execution context rather than per thread. But with thread static it may be that the initial allocation is free.

gao-artur (approved):

PR #3989, opened 2024-07-15
Rewrite complexity analyzer

Shane32:
** This PR depends on the following PR which must be merged first **
- #3991

New features:
- Fields can define the impact for the execution of that field (e.g. how long it will take to execute the ResolveAsync method) separate from a multiplier applied to the children fields (e.g. how many rows it returns for list fields)
- Default configuration allows for setting three default values:
  1. Scalar impact (impact to use for scalar fields)
  2. Object impact (impact to use for object fields)
  3. Default child multiplier for list fields (multiplier to use for list fields)
- Same as before, the `first` and `last` arguments are taken into account if defined and provided
- `first` and `last` work properly when supplied by variables (ignored before)
- `first` and `last` properly work for connection semantics where the argument is on the parent node
- Same as before, if `id` is specified on a list field, the child multiplier is set to 1
- Fields may define their own complexity algorithms, overriding all default behavior described above.  Easy for these algorithms to inspect parent fields and their coerced arguments.
- Meta fields are properly supported
- All calculations are logical and easy to follow
- Configuring a schema to ignore complexity for introspection fields is now as easy as setting the child impact multiplier for the three introspection meta fields to zero. (Not so before) An extension method for `ISchema` is included to do this.
- Easy to write asynchronous code to implement per-user, per-IP, or throttling limits
- Executes faster with less memory

Todo
- [x] Split to multiple PRs
- [x] Add code comments
- [x] Add XML comments
- [x] Write extensive tests
- [x] Perform benchmarks
- [x] Write documentation
- [x] Write migration notes


Shane32:
Here's some thoughts regarding the new complexity analyzer:
1. Should the field impact delegate by synchronous or async? All the supporting code is async, so it would be quite easy to make it async.  However, 99.9% of the use case will be synchronous; I cannot imagine where an async field impact delegate would be needed.  Further, it complicates making derived field impact delegates.
2. Name of various properties in the configuration class?  I feel like some could be improved but not sure how.
3. Property to ignore depth calculation of nodes where multiplier is zero?  For instance, when disabling the complexity calculation of introspection requests, this would then have the depth calculated as zero as well.  See new docs where this is mentioned.
4. Behavior of `.WithComplexityImpact(num)` - is this the field complexity or the child multiplier?  I think it was the child multiplier in previous versions (along with some bad math).  But if there's another overload with `impact, childMultiplier` then the simpler overload just only have the first number which is `impact`.  And perhaps it makes more sense this way.
5. Should the scalar impact be multiplied by the child multiplier when the field is a list of scalars?  Currently it does not do this.  The impact is just how long it takes to do whatever the resolver is.  I think this is most straightforward and understandable to users.

src/GraphQL.Tests/DI/GraphQLBuilderExtensionTests.cs
```diff
@@ -346,23 +346,23 @@ public void AddDocumentExecuter_Null()
     [InlineData(false, true)]
     public void AddComplexityAnalyzer(bool withAction, bool withAction2)
```
Shane32, line 347: todo: duplicate this test for the new complexity analyzer

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -391,8 +411,8 @@ public static IGraphQLBuilder AddComplexityAnalyzer(this IGraphQLBuilder builder
         return builder;
     }
 
-    /// <inheritdoc cref="AddComplexityAnalyzer(IGraphQLBuilder, Action{ComplexityConfiguration})"/>
-    public static IGraphQLBuilder AddComplexityAnalyzer(this IGraphQLBuilder builder, Action<ComplexityConfiguration, IServiceProvider?>? action)
+    /// <inheritdoc cref="AddLegacyComplexityAnalyzer(IGraphQLBuilder, Action{LegacyComplexityConfiguration})"/>
```
Shane32, line 414:
```suggestion
    /// <inheritdoc cref="AddComplexityAnalyzer(IGraphQLBuilder, Action{ComplexityConfiguration})"/>
```

src/GraphQL/Validation/Complexity/FieldImpactContext.cs
```diff
@@ -0,0 +1,80 @@
...
+using GraphQL.Types;
+using GraphQL.Validation.Rules.Custom;
+using GraphQLParser.AST;
+
+namespace GraphQL.Validation.Complexity;
+
+/// <summary>
+/// Provides contextual information about the field being analyzed.
+/// </summary>
+public struct FieldImpactContext
```
Shane32, line 11: Using a struct here with limited amount of internal fields to speed complexity analysis and minimize allocations, as this is constructed for every analyzed field.  Using a struct means that it cannot be derived from, but additional fields can be added to a derived `ComplexityConfiguration` instance, which is accessible from this struct.

src/GraphQL/Validation/Complexity/FieldImpactContext.cs
```diff
@@ -0,0 +1,80 @@
...
+        ? null
+        : new FieldImpactContext
+        {
+            FieldDefinition = PeekElement(VisitorContext.FieldDefinitions, _parentDepth)!,
+            FieldAst = PeekElement(VisitorContext.FieldAsts, _parentDepth)!,
+            VisitorContext = VisitorContext,
+            _parentDepth = _parentDepth + 1,
+        };
+
+    private static T PeekElement<T>(Stack<T> from, int index)
```
Shane32, line 59: Similar to existing code within `ValidationContext`.  perhaps it would be best to make this an extension method, but the implementations are a little bit different.  This code is optimized for `index == 0` aka only inspecting the parent one level up.

src/GraphQL/Validation/ValidationContext.GetRecursivelyReferencedFragments.cs
```diff
@@ -63,7 +63,7 @@ public partial class ValidationContext
     /// will not reflect the change. Also ignores any overridden behavior within
     /// <see cref="Execution.ExecutionStrategy.ShouldIncludeNode{TASTNode}(Execution.ExecutionContext, TASTNode)">ExecutionStrategy.ShouldIncludeNode</see>.
     /// </remarks>
-    protected virtual bool ShouldIncludeNode(ASTNode node)
+    public virtual bool ShouldIncludeNode(ASTNode node)
```
Shane32, line 66: New method from #3981; didn't realize I would need this public to be useful!

src/GraphQL/Validation/Complexity/FieldImpactContext.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    internal ComplexityValidationRule.VisitorContext VisitorContext;
+    private int _parentDepth;
+
+    private IDictionary<string, ArgumentValue>? _arguments;
+    /// <inheritdoc cref="IResolveFieldContext.Arguments"/>
+    public IDictionary<string, ArgumentValue>? Arguments => _arguments ??= (VisitorContext.Arguments?.TryGetValue(FieldAst, out var args) == true ? args : null);
+
+    /// <inheritdoc cref="ResolveFieldContextExtensions.GetArgument{TType}(IResolveFieldContext, string, TType)"/>
+    public TType GetArgument<TType>(string name, TType defaultValue = default!) => Arguments?.TryGetValue(name, out var value) == true ? (TType)value.Value! : defaultValue;
```
Shane32, line 38: `IResolveFieldContext.GetArgument` uses the schema's defined name converter to first convert the argument.  Should that be done here also?  It's going to slow down the execution of all analysis.  Maybe just add a note?  Maybe add it here, but also add an internal implementation which doesn't, and use that internally?

Shane32:
Benchmarks show this:

| Method                   | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------------------------- |---------:|---------:|---------:|-------:|----------:|
| LegacyComplexityAnalyzer | 58.36 us | 0.704 us | 0.659 us | 0.3662 |   6.78 KB |
| NewComplexityAnalyzer    | 44.71 us | 0.398 us | 0.353 us | 1.2207 |  23.46 KB |

Although I think this is skewed because it only has a single shared instance, and the benchmark will run many threads concurrently (and always concurrently).  Assuming the typical use case is that the complexity analysis will only need to run on one thread at a time (since most of the processing power is during execution of the request, not the complexity analysis), the amount of memory needed should be minimal.

Shane32:
Nope it was just 'bad code'.  Fixed two places that were causing an allocation to occur.

| Method                   | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------------------------- |---------:|---------:|---------:|-------:|----------:|
| LegacyComplexityAnalyzer | 56.18 us | 0.250 us | 0.209 us | 0.3662 |   6.78 KB |
| NewComplexityAnalyzer    | 42.64 us | 0.126 us | 0.112 us | 0.1221 |   2.93 KB |

gao-artur:
Hi @Shane32, This topic looks complex. I'll need time to understand and review it. But I'm not available until the 22nd. Sorry.

Shane32:
Not a problem and thanks for the update!

Shane32:
@gao-artur No rush, but just FYI when you're available: I've split this PR into (1) renaming the old complexity analyzer, and (2) adding the new one.  It should make for an easier review.  You'll find that almost all of the actual logic is contained within the `ComplexityVisitor` and `ComplexityConfiguration` classes, with other relevant files being `ComplexityValidationRule`, `ComplexityVisitorContext` and `FieldImpactContext`.  I've extensively commented these classes, so it hopefully is easy to follow and review.  I've also added extensive documentation for the new complexity analyzer, which mostly consist of real-life examples.  You may want to read it first, so you can see the goals of the new complexity analyzer and how it can be used.

src/GraphQL/Validation/Complexity/FieldImpactContext.cs
```diff
@@ -0,0 +1,80 @@
...
+    internal ComplexityVisitorContext VisitorContext;
+    private int _parentDepth;
+
+    private IDictionary<string, ArgumentValue>? _arguments;
+    /// <inheritdoc cref="IResolveFieldContext.Arguments"/>
+    public IDictionary<string, ArgumentValue>? Arguments => _arguments ??= (VisitorContext.Arguments?.TryGetValue(FieldAst, out var args) == true ? args : null);
+
+    /// <inheritdoc cref="ResolveFieldContextExtensions.GetArgument{TType}(IResolveFieldContext, string, TType)"/>
+    /// <remarks>Does not coerce the argument name via <see cref="ISchema.NameConverter"/>.</remarks>
+    public TType GetArgument<TType>(string name, TType defaultValue = default!) => Arguments?.TryGetValue(name, out var value) == true ? (TType)value.Value! : defaultValue;
```
Shane32, line 38: Alternate logic:
```suggestion
    /// <inheritdoc cref="ResolveFieldContextExtensions.GetArgument{TType}(IResolveFieldContext, string, TType)"/>
    public TType GetArgument<TType>(string name, TType defaultValue = default!)
    {
        // see logic in ResolveFieldContextExtensions.GetArgument
        var arguments = Arguments;
        if (arguments == null)
            return defaultValue;
        var isIntrospectionField = FieldDefinition.Name.StartsWith("__", StringComparison.Ordinal) || ParentType.Name.StartsWith("__", StringComparison.Ordinal);
        var argumentName = isIntrospectionField ? name : (ValidationContext.Schema.NameConverter.NameForArgument(name, (IComplexGraphType)ParentType, FieldDefinition) ?? name);
        if (!arguments.TryGetValue(argumentName, out var value) || value.Value == null)
            return defaultValue;
        if (value is TType typedValue)
            return typedValue;
        var resolvedType = FieldDefinition.Arguments?.Find(argumentName)?.ResolvedType
            ?? throw new InvalidOperationException($"Could not obtain graph type instance for argument '{argumentName}'");
#pragma warning disable IL2087 // Target parameter argument does not satisfy 'DynamicallyAccessedMembersAttribute' in call to target method. The generic parameter of the source method or type does not have matching annotations.
        if (value.Value is IDictionary<string, object?> inputObject)
        {
            return (TType)inputObject.ToObject(typeof(TType), resolvedType);
        }
        return (TType)value.Value.GetPropertyValue(typeof(TType), resolvedType)!;
#pragma warning restore IL2087 // Target parameter argument does not satisfy 'DynamicallyAccessedMembersAttribute' in call to target method. The generic parameter of the source method or type does not have matching annotations.
    }
```

Shane32: This logic mirrors `IResolveFieldContext.GetArgument` exactly, first coercing the argument name via the name converter, then converting to the requested type via `GetPropertyValue`

Shane32:
No problem; thanks!!

docs2/site/docs/guides/complexity-analyzer.md
```diff
@@ -0,0 +1,346 @@
...
+    {
+        context.ReportError(new ExecutionError($"Query complexity of {complexity} exceeded throttling limit. Remaining: {remaining}"));
+    }
+};
+```
+
+### 6. Throttling Users Based on Execution Time
+
+While the complexity analyzer does not directly measure execution time, you can write the following code
+to control the maximum execution time of a query:
```
gao-artur, line 313: It looks unrelated to the topic. But it may be worth adding a dedicated property for the timeout configuration. And a dedicated docs page to mention different ways to configure the timeout: this property for simple config, `ConfigureExecution` for more complex scenarios, and APS.NET Core [middleware](https://learn.microsoft.com/en-us/aspnet/core/performance/timeouts?view=aspnetcore-8.0) when you don't need the GraphQL context.

docs2/site/docs/guides/complexity-analyzer.md
```diff
@@ -0,0 +1,346 @@
...
+
+### Configurable Options
+
+| Option                          | Description                                                    | Default Value |
+|---------------------------------|----------------------------------------------------------------|---------------|
+| MaxDepth                        | Limits the maximum depth of a query.                           | null          |
+| MaxComplexity                   | Limits the total complexity of a query.                        | null          |
+| DefaultScalarImpact             | Specifies the default complexity impact for scalar fields.     | 1             |
+| DefaultObjectImpact             | Specifies the default complexity impact for object fields.     | 1             |
+| DefaultListImpactMultiplier     | Specifies the average number of items returned by list fields. | 5             |
```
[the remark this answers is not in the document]
Shane32: Yes; I'm open to any suggestions.  I think we should choose a default.  Maybe we should go larger rather than smaller?  10?  20?

docs2/site/docs/guides/complexity-analyzer.md
```diff
@@ -0,0 +1,346 @@
...
+### 1. Estimating the Total Number of Nodes and Maximum Depth
+
+To configure the complexity analyzer to estimate the total number of nodes returned and/or the maximum depth,
+you can use the default configuration, or customize the default impact multiplier, or customize the impact
+multiplier used for specific fields. The default configuration assumes that list fields return an average of 5 items.
+
+#### Configuring Child Impact Multiplier for Specific Fields
+
+```csharp
+usersField.WithComplexityImpact(1, 100); // Assume the users field returns 100 items on average
```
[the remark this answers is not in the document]
Shane32: Good idea.

docs2/site/docs/guides/complexity-analyzer.md
```diff
@@ -0,0 +1,346 @@
...
+complexityConfig.ValidateComplexityDelegate = async (context, complexity, depth) =>
+{
+    // RequestServices may be used to access scoped services within the DI container
+    var logger = context.RequestServices!.GetRequiredService<ILogger<MySchema>>();
+    logger.LogInformation($"Query Complexity: {complexity}, Depth: {depth}");
+};
+```
+
+To log failed complexity checks, set the `MaxDepth` and `MaxComplexity` properties to `null` and provide
+your own validation delegate:
```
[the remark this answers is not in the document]
Shane32: Right now we have:

```cs
public Func<ValidationContext, double, int, Task>? ValidateComplexityDelegate { get; set; }
```

Rather than add yet another delegate, how about if this becomes:

```cs
public Func<ComplexityResult, Task>? ValidateComplexityDelegate { get; set; }

public class ComplexityResult
{
    public ValidationContext ValidationContext { get; }
    public ComplexityOptions ComplexityOptions { get; }
    public double TotalImpact { get; }
    public int MaxDepth { get; }
    public ComplexityError? Error { get; set; } // null if no error
}

// when validation occurs, initialize the ComplexityResult class, setting the Error
//   property if the impact or depth has been exceeded; then call the delegate.
// if the error has been cleared, continue as if no problem, otherwise report
//   the error within the Error property
```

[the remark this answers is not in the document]
Shane32: The `Error` property could be named something else, of course, like `Result`.  But as the class is `ComplexityResult`, I probably wouldn't.  Also we could allow `ValidationError` instead of `ComplexityError`.

docs2/site/docs/guides/complexity-analyzer.md
Shane32: Where are you thinking a timeout would be added?  It's not really anything to do with the complexity configuration.  Is it in `ExecutionOptions`?  But then the timeout would only be able to wrap the actual execution, since the options will not be configured if we tried to wrap the entire chain.  I could add a PR and we can look at it.

docs2/site/docs/guides/complexity-analyzer.md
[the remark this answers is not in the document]
gao-artur: Too low value will cause a false negative, too high - a false positive. False negative is a security issue, false positive is a UX issue. I was always told that security comes before UX, so perhaps it is better to start with a higher number to address security. Then user can set a lower number per field when needed. In our application, we have many fields that return hundreds of objects, but I don't think this is a common use case. Maybe 20 is a good choice.

docs2/site/docs/guides/complexity-analyzer.md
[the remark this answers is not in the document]
gao-artur: How about getting inspiration form [CookieAuthenticationEvents](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.cookies.cookieauthenticationevents?view=aspnetcore-8.0)? Every method in this class receives its own context type. The `ValidatePrincipal` receives [CookieValidatePrincipalContext](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.cookies.cookievalidateprincipalcontext?view=aspnetcore-8.0). There is a property called `Principal` and two methods: `RejectPrincipal` and `ReplacePrincipal`. When you see methods like Reject and Replace, you intuitively understand how to use the context and don't have to guess what to do with the `Principal` property (like setting it to `null`, for example).

In our case, it can be `ValidateComplexityContext`, with `ValidationResult` property and `FailValidation(ValidationError)` + `RejectValidation`
```csharp
public class ValidateComplexityContext
{
    public ValidationContext ValidationContext { get; }
    public ComplexityOptions ComplexityOptions { get; }
    public double TotalImpact { get; }
    public int MaxDepth { get; }
    public ValidationError? ValidationResult { get; private set;}

    public void FailValidation(ValidationError error) => ValidationResult = error;
    public void RejectValidation() => ValidationResult = null; // better name?
}
```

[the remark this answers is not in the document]
gao-artur: Btw, can it return `ValueTask` instead of `Task`?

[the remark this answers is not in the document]
Shane32: I think it would be better to mesh with code already familiar to GraphQL.NET users rather than copy a design from another class.  I for one have no familiarity with `CookieAuthenticationEvents` or `CookieValidatePrincipalContext`, and none of my devs do either.  I also don't find `Fail` and `Reject` to be clear; they seem to be synonyms.

We could also forgo the `ValidationResult` field altogether and do this:

```cs
public Func<ComplexityResult, Task> ValidateComplexityDelegate { get; set; } = result =>
{
    // default implementation
}


public class ComplexityResult
{
    public ValidationContext ValidationContext { get; }
    public ComplexityOptions ComplexityOptions { get; }
    public double TotalImpact { get; }
    public int MaxDepth { get; }
}
```

The default implementation calls `ValidationContext.AddError` to add an error if the impact or depth is exceeded.  However, this is a pain when overriding as now the default logic needs to be re-implemented, and/or the code needs to examine `ValidationContext` to determine if any complexity errors were added.  So I really don't like this.

Yet another option is this:

```cs
public Func<ComplexityResult, ValueTask<ComplexityError?>> ValidateComplexityDelegate { get; set; } = result =>
{
    // default implementation
}

public struct ComplexityResult
{
    public ValidationContext ValidationContext { get; }
    public ComplexityOptions ComplexityOptions { get; }
    public double TotalImpact { get; }
    public int MaxDepth { get; }
}
```

User-defined delegates can call the original implementation and then modify the result.  One benefit is that no allocations are typically necessary.  Again, this is more difficult to utilize than the approach used by `UnhandledErrorDelegate`; the simplest implementation would be:

```cs
var originalDelegate = options.ValidateComplexityDelegate;
options.ValidateComplexityDelegate = async (result) =>
{
    var error = await originalDelegate(result);
    if (error != null)
        // log error
    return error;
};
```

Whereas with the suggestion above, it would be:

```cs
options.ValidateComplexityDelegate = async (result) =>
{
    if (result.Error != null)
        // log error
}
```

I think the majority of the use case here is either:
1. Logging analysis results (perhaps only failures, perhaps all for statistical analysis)
2. Applying additional throttling restrictions based on the calculated total impact

As such, we should focus on those use cases, making it easy for the developer.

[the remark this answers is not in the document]
Shane32: > Btw, can it return `ValueTask` instead of `Task`?

There is little-to-no value when not returning `Task<T>`, as the returned `Task` will be a cached instance anyway.  I've tried to mask all use of `ValueTask` from the developer whenever possible.  However, here it is critical to use `ValueTask` for `CalculateComplexityAsync` as this is executed for each field and must return a value in each case.  Some of the GraphQL.NET interfaces do use `ValueTask` when `Task` would have been equally sufficient, just to maintain parity with the rest of the methods within the same class.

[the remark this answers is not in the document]
Shane32: See: "Should I use ValueTask for new asynchronous APIs?" here https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/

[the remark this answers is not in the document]
gao-artur: > I think it would be better to mesh with code already familiar to GraphQL.NET

Completely understandable. Then this one is the best of all the proposals because it's similar to `UnhandledExceptionDelegate`
```c#
public Func<ComplexityResult, Task>? ValidateComplexityDelegate { get; set; }
```
Just the names. Again, for consistency with `UnhandledExceptionDelegate`
```c#
public Func<ComplexityValidationContext, Task>? ValidateComplexityDelegate { get; set; }

public class ComplexityValidationContext
{
    public ValidationContext ValidationContext { get; }
    public ComplexityOptions ComplexityOptions { get; }
    public double TotalImpact { get; }
    public int MaxDepth { get; }
    public ValidationError? Error { get; set; } // null if no error
}
```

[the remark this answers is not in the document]
gao-artur: > See: "Should I use ValueTask for new asynchronous APIs?" here https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/

But then there is another opinion from another shinning name in the .NET world
[Prefer ValueTask to Task, always; and don't await twice](https://blog.marcgravell.com/2019/08/prefer-valuetask-to-task-always-and.html)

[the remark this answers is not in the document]
Shane32: Done, but I need to update the docs.

docs2/site/docs/guides/complexity-analyzer.md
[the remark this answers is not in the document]
Shane32: Updated to 20.

docs2/site/docs/guides/complexity-analyzer.md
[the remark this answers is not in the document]
Shane32: Docs updated

Shane32:
Renamed ComplexityConfiguration to ComplexityOptions as it seemed to fit better with other classes like ExecutionOptions.

docs2/site/docs/guides/complexity-analyzer.md
Shane32: See:
- #3998

docs2/site/docs/guides/complexity-analyzer.md
```diff
@@ -0,0 +1,351 @@
...
+- Other fields are scalar fields and use the default scalar impact of 1, multiplied by the
+  multiplier calculated for that level of the graph.
+
+## Example Scenarios
+
+### 1. Estimating the Total Number of Nodes and Maximum Depth
+
+To configure the complexity analyzer to estimate the total number of nodes returned and/or the maximum depth,
+you can use the default configuration, or customize the default impact multiplier, or customize the impact
+multiplier used for specific fields. The default configuration assumes that list fields return an average of 5 items.
```
gao-artur, line 112:
```suggestion
multiplier used for specific fields. The default configuration assumes that list fields return an average of 20 items.
```

docs2/site/docs/guides/complexity-analyzer.md
gao-artur: Should we remove this section now?

Shane32: Updated

docs2/site/docs/guides/complexity-analyzer.md
```diff
@@ -0,0 +1,321 @@
...
+
+- Fields can define the impact for the execution of that field (e.g., how long it will take to execute the ResolveAsync
+  method) separate from a multiplier applied to the children fields (e.g., how many rows it returns for list fields).
+- Default configuration allows for setting three default values:
+  - Scalar impact (impact to use for scalar fields)
+  - Object impact (impact to use for object fields)
+  - Default child multiplier for list fields (multiplier to use for list fields)
+- Per-field configurable behavior to determine field impact and child impact multipliers; by default it uses the
+  scalar impact for scalar fields, object impact for object fields, and the default child multiplier for list fields.
+- Default behavior considers connection semantics, such as `first`, `last`, and `id` arguments, to adjust the child
```
gao-artur, line 19: `limit` + `offset` are also a very popular solution for pagination (even though not the best practice). Can they also be added to special handling list? Alternativly, can we expose an additional overload for `WithComplexityImpact` with access to `IResolveFieldContext` so that user can create his own logic based on arguments?

Shane32: The functionality already exists but I didn't add documentation. I'll revise the documentation.

gao-artur: Oh, wait. We already have such an overload with `FieldImpactContext`, right?

Shane32: Yes

src/GraphQL/Validation/Complexity/FieldImpactContext.cs
gao-artur: Sounds good for me

docs2/site/docs/guides/complexity-analyzer.md
Shane32: Done

Shane32: The default logic can also be redefined within the options, if desired.

docs2/site/docs/guides/complexity-analyzer.md
```diff
@@ -0,0 +1,340 @@
...
+See the following documentation for more information:
+
+https://graphql-dotnet.github.io/docs/migrations/migration8/#24-execution-timeout-support
+
+## Advanced configurations
+
+### Defining Custom Complexity Calculations
+
+To set custom complexity calculations for specific fields, you can use the `WithComplexityImpact` overload
+that defines a calculation delegate as demonstrated in the following eample:
```
gao-artur, line 328:
```suggestion
that defines a calculation delegate as demonstrated in the following example:
```

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -3831,6 +3825,12 @@ namespace GraphQL.Validation.Complexity
         public double TotalImpact { get; }
         public GraphQL.Validation.ValidationContext ValidationContext { get; }
     }
+    public struct FieldComplexityResult
+    {
+        public double ChildImpactMultiplier;
+        public double FieldImpact;
+        public FieldComplexityResult(double fieldImpact, double childImpactMultiplier) { }
+    }
```
Shane32, line 3833: In this case the struct exists primarily to provide names to what was ValueTuple fields.  Should they be fields or props?

Shane32: The `ValueTuple<T1, T2>` struct uses fields, not properties.

https://learn.microsoft.com/en-us/dotnet/standard/value-tuples

docs2/site/docs/guides/complexity-analyzer.md
gao-artur: Thanks for the example!
We use `limit` in many places, so we will need to apply this logic a lot. Is it possible to make the list of supported pagination arguments extensible? So by default, it will contain `first` and `last`, but the user will be able to add `limit` or whatever if needed.
Alternatively, I think, the user can do that with the schema visitor by looking for fields with a `limit` argument and applying the logic from the example.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
gao-artur: If the type is public, I'd make them properties. I don't think the performance impact is noticeable.

docs2/site/docs/guides/complexity-analyzer.md
Shane32: Just redefine the default behavior within the options.  Copy the original code and make a few tweaks.  No need to use a schema visitor.

I don't know how extensible we want the default delegate.  Every field needs to run the code, and if it needs to search another list to know what arguments to look for, it will impact performance further.

Thoughts?

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
Shane32: Updated to props as suggested.

docs2/site/docs/guides/complexity-analyzer.md
gao-artur: Copy+adjust is also an option. The only drawback: if you fix a bug in the original code, I'll be stuck with buggy code 🙂

gao-artur (approved):

gao-artur:
Looks great!

Shane32:
> Looks great!

Thanks! I've wanted to add this feature for a couple years now.  I'm pretty happy with how it turned out and I really appreciate being able to collaborate with you on this!

Shane32:
Part of this PR is only possible in v8 with the new way that variable values are coerced, so it feels like a number of the features added in v8 are coming together to make a better library.

PR #3993, opened 2024-07-22
Add persisted document support

Shane32:
Adds "Persisted Documents" support based on the spec here:
- https://github.com/graphql/graphql-over-http/pull/264

### DIFF
--- a/src/GraphQL/PersistedDocuments/InvalidRequestError.cs
+++ b/src/GraphQL/PersistedDocuments/InvalidRequestError.cs
@@ -1,0 +1,14 @@
+using GraphQL.Execution;
+
+namespace GraphQL.PersistedDocuments;
+
+/// <summary>
+/// Represents an error that occurs when a request contains both query and documentId parameters.
+/// </summary>
+public class InvalidRequestError : RequestError
+{
+    /// <inheritdoc cref="InvalidRequestError"/>
+    public InvalidRequestError() : base("The request must not have both query and documentId parameters.")
+    {
+    }
+}
--- a/src/GraphQL/PersistedDocuments/PersistedDocumentHandler.cs
+++ b/src/GraphQL/PersistedDocuments/PersistedDocumentHandler.cs
@@ -1,0 +1,120 @@
+using GraphQL.DI;
+
+namespace GraphQL.PersistedDocuments;
+
+/// <summary>
+/// Handles persisted document requests.
+/// </summary>
+public class PersistedDocumentHandler : IConfigureExecution
+{
+    private readonly PersistedDocumentOptions _options;
+
+    /// <summary>
+    /// Initializes a new instance with the specified options.
+    /// </summary>
+    public PersistedDocumentHandler(PersistedDocumentOptions options)
+    {
+        _options = options;
+    }
+
+    /// <summary>
+    /// Initializes a new instance with the default options.
+    /// Useful when only sha256 document identifiers are in use and
+    /// <see cref="IPersistedDocumentLoader"/> has been registered to
+    /// load documents.
+    /// </summary>
+    public PersistedDocumentHandler()
+        : this(new PersistedDocumentOptions())
+    {
+    }
+
+    /// <summary>
+    /// Create <see cref="ExecutionResult"/> with provided error.
+    /// Override this method to change the provided error responses.
+    /// </summary>
+    protected virtual ExecutionResult CreateExecutionResult(ExecutionError error) => new(error);
+
+    private static readonly Dictionary<string, Func<string, bool>> _documentIdValidators = new()
+    {
+        {
+            "sha256",
+            payload =>
+            {
+                // validate sha256 hash length
+                if (payload.Length != 64)
+                    return false;
+
+                // validate only lowercase hexadecimal characters are present
+                for (int i = 0; i < payload.Length; i++)
+                {
+                    var c = payload[i];
+                    if (!((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f')))
+                        return false;
+                }
+
+                return true;
+            }
+        }
+    };
+
+    /// <inheritdoc/>
+    public async Task<ExecutionResult> ExecuteAsync(ExecutionOptions options, ExecutionDelegate next)
+    {
+        // ensure that both documentId and query are not provided
+        if (options.DocumentId != null && options.Query != null)
+            return CreateExecutionResult(new InvalidRequestError());
+
+        // if query is provided, but documentId is not, then we need to execute or refuse the request
+        if (options.Query != null)
+        {
+            if (!_options.AllowNonpersistedDocuments)
+                return CreateExecutionResult(new DocumentIdMissingError());
+            else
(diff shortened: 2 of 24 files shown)


gao-artur:
Interesting. How is it supposed to work with the Federation? Only the the Apollo Router will have this feature enabled?

Shane32:
> Interesting. How is it supposed to work with the Federation? Only the the Apollo Router will have this feature enabled?

I expect so. Doesn’t make sense for a subgraph to only allow whitelisted queries since the router creates them dynamically. Apollo currently supports their own version which we support by the UseAutomaticPersistedQueries call. 

https://www.apollographql.com/docs/kotlin/advanced/persisted-queries/

This PR is to be compatible with the proposed standard. It’s just a draft appendix to a draft spec so it’s subject to change. But the way that the GraphQL committee works is that nothing is merged until it is used in public libraries first. So the PR there hasn’t changed in a while and I thought we could merge in support here.

src/GraphQL/PersistedDocuments/InvalidRequestError.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace GraphQL.PersistedDocuments;
+
+/// <summary>
+/// Represents an error that occurs when a request contains both query and documentId parameters.
+/// </summary>
+public class InvalidRequestError : RequestError
+{
+    /// <inheritdoc cref="InvalidRequestError"/>
+    public InvalidRequestError() : base("The request must not have both query and documentId parameters.")
```
gao-artur, line 11: I'm not sure this is correct
> Note: When persisting a document for which the identifier has been derived by the client, it is generally good practice for the client to issue both the GraphQL Document and the document identifier to the server; the server could then regenerate the document identifier from the GraphQL Document independently, and check that the identifiers match before storing the Document.

Shane32: Where is that from? Are you sure you’re not reading docs on APQ support?

src/GraphQL/PersistedDocuments/PersistedDocumentHandler.cs
```diff
@@ -0,0 +1,120 @@
...
+        var query = await _options.GetQueryDelegate(options, prefix, payload).ConfigureAwait(false);
+
+        // if the document id was not found, return an appropriate error
+        if (query == null)
+            return CreateExecutionResult(new DocumentNotFoundError());
+
+        // set the query string and continue execution
+        options.Query = query;
+        return await next(options).ConfigureAwait(false);
+    }
```
gao-artur, line 105: Note for .NET 9: can reduce memory allocations here. https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview6/libraries.md#collection-lookups-with-spans

src/GraphQL/PersistedDocuments/InvalidRequestError.cs
gao-artur: The last paragraph under [Persisting a Document](https://github.com/graphql/graphql-over-http/blob/52d56fb36d51c17e08a920510a23bdc2f6a720be/spec/Appendix%20A%20--%20Persisted%20Documents.md)

gao-artur: Here: https://github.com/graphql/graphql-over-http/pull/264/files#diff-9be5577e05ae2112d2b8f95584b162d0dec01453bf6c85df58bf5db4f2c9727aR122-R129

src/GraphQL/PersistedDocuments/PersistedDocumentOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+/// </summary>
+public class PersistedDocumentOptions
+{
+    private const string ERROR_NO_REQUESTSERVICES = $"{nameof(ExecutionOptions)}.{nameof(ExecutionOptions.RequestServices)} or {nameof(PersistedDocumentOptions)}.{nameof(GetQueryDelegate)} must be set.";
+    private const string ERROR_NO_LOADER = $"{nameof(IPersistedDocumentLoader)} must be registered with your DI provider or {nameof(PersistedDocumentOptions)}.{nameof(GetQueryDelegate)} must be set.";
+
+    /// <summary>
+    /// Indicates if requests that are not persisted document requests should be allowed to be executed.
+    /// </summary>
+    public bool AllowNonpersistedDocuments { get; set; }
```
gao-artur, line 14: I believe using this feature for performance is a more common use case than for the allow list. So, I believe this should be `true` by default. Or renamed to `AllowOnlyPersistedDocuments`.

src/GraphQL/PersistedDocuments/InvalidRequestError.cs
Shane32: I believe this relates to the preceding note, discussing "automatic persisted queries (APQ)" support, where the client knows the text of the request and determines a hash for it, sending it instead of or addition to the original request.  We support this with the `UseAutomaticPersistedQueries` method, supporting Apollo's specification listed [here](https://www.apollographql.com/docs/apollo-server/performance/apq/), which differs from this specification.

Unfortunately the current spec does not provide enough information to make a working APQ implementation.  Without specifying the result in the case of a failed match against a hash, the client is unable to conclusively identify the failure reason and resend the request with the query string.  As such, both the preceding comment and this one have no purpose at this time.

src/GraphQL/PersistedDocuments/PersistedDocumentOptions.cs
Shane32: See comments above; APQ support is not possible with the current state of the draft RFC appendix.  As such, I believe the primary purpose of this PR is for trusted document support.

Shane32:
Another option would be to just introduce the `DocumentId` property to the `GraphQLRequest` class and skip adding the persisted document handler altogether.  This allows people to write their own handler if they want, which will be integrated to GraphQL.NET Server for GET/POST or Websocket calls.

src/GraphQL/PersistedDocuments/InvalidRequestError.cs
gao-artur: Understood. Then all looks good

gao-artur (approved):

Shane32:
Still okay with `AllowNonpersistedDocuments`?  I'm not a big fan of it, but it was the best I could think of.  We could use `AllowOnlyPersistedDocuments` and set default to `true` if you think that's better.  @gao-artur

src/GraphQL/PersistedDocuments/PersistedDocumentHandler.cs
Shane32: cool!

gao-artur:
> Still okay with `AllowNonpersistedDocuments`? I'm not a big fan of it, but it was the best I could think of. We could use `AllowOnlyPersistedDocuments` and set default to `true` if you think that's better. @gao-artur

I'm ok with both. But I think the default should be documented on the `UsePersistedDocuments` method to make it clear that this feature will disable arbitrary queries.

PR #4000, opened 2024-07-28
Add complexity attribute

### LINKED ISSUE #3200: Add an attribute for setting the complexity factor on fields of type-first/schema-first schemas.
Note: we should add an attribute for setting the complexity factor on fields of type-first/schema-first schemas.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/issues/3195#issuecomment-1164760968_

Shane32:
Closes #3200

### DIFF
--- a/src/GraphQL/Attributes/ComplexityAttribute.cs
+++ b/src/GraphQL/Attributes/ComplexityAttribute.cs
@@ -108,7 +108,6 @@ public class ComplexityAttribute<T> : GraphQLAttribute
     private readonly Func<FieldImpactContext, FieldComplexityResult> _func = new T().Analyze;
 
     /// <inheritdoc/>
-    [Complexity(fieldImpact: 1, childImpactMultiplier: 100)]
     public override void Modify(FieldConfig field)
         => field.WithComplexityImpact(_func);
 
--- a/src/GraphQL/Attributes/ComplexityAttribute.cs
+++ b/src/GraphQL/Attributes/ComplexityAttribute.cs
@@ -1,0 +1,115 @@
+using GraphQL.Types;
+using GraphQL.Utilities;
+using GraphQL.Validation.Complexity;
+
+namespace GraphQL.Attributes;
+
+/// <summary>
+/// Specifies the complexity impact and/or child impact multiplier of a field.
+/// Not applicable to input fields.
+/// </summary>
+[AttributeUsage(AttributeTargets.Method | AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false)]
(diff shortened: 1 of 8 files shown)


src/GraphQL/Attributes/ComplexityAttribute.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// <inheritdoc/>
+    public override void Modify(FieldConfig field)
+        => field.WithComplexityImpact(_func);
+
+    /// <inheritdoc/>
+    public override void Modify(FieldType fieldType, bool isInputType)
+    {
+        if (!isInputType)
+            fieldType.WithComplexityImpact(_func);
```
gao-artur, line 118: Silent ignore? Should throw instead? Do we need an analyzer similar to `AllowedOn<>` for code first?

src/GraphQL/Validation/Complexity/IFieldComplexityAnalyzer.cs
```diff
@@ -0,0 +1,16 @@
+using GraphQL.Attributes;
+
+namespace GraphQL.Validation.Complexity;
+
+/// <summary>
+/// Represents a field complexity analyzer. Used with the <see cref="ComplexityAttribute"/> to specify
+/// a custom implementation for calculating the complexity of a field. Classes that implement this
+/// interface must have a public parameterless constructor.
```
gao-artur, line 8: Need analyzer?

src/GraphQL/Attributes/ComplexityAttribute.cs
Shane32: Type-first classes may often be used for both input and output types. I would not throw or analyze.

src/GraphQL/Validation/Complexity/IFieldComplexityAnalyzer.cs
Shane32: Ideally that would be good. If developers use the generic version of the attribute then it already has the necessary constraint, but generic attributes are not supported in the .NET Framework. I don’t think it’s a big deal but if you want to add it - either to this pr or as a separate one, I’m all for it.

gao-artur (approved):

src/GraphQL/Validation/Complexity/IFieldComplexityAnalyzer.cs
gao-artur: Will add in separate PR

Shane32: Thanks!!!

PR #4001, opened 2024-07-29
Optimize scalar lists

Shane32:
This change makes it so that when serializing a list of intrinsic types, such as `int[]` for `ListGraphType<IntGraphType>`, the scalar can serialize the entire list at once.  This eliminates at least two allocations per element and it should be much faster to boot. Of course the use case is narrow, but is certainly advantageous if you want to return a large list of metrics or similar values out of GraphQL.  Note that no changes are made for input types.

## Performance

Given this type-first schema, where each field returns 100 elements:

```cs
public class MyQuery
{
    public static int[] Ints { get; } = Enumerable.Range(1, 100).ToArray();
    public static string[] Strings { get; } = Enumerable.Range(1, 100).Select(x => x.ToString()).ToArray();
    public static Guid[] Guids { get; } = Enumerable.Range(1, 100).Select(x => Guid.NewGuid()).ToArray();
}
```

And one of three queries (cached, so query parsing & validation is not timed):

```gql
query Ints { ints }
query Strings { strings }
query Guids { guids }
```

Below metrics include either:
1. Only execution, or
2. Execution and serialization

### Before

| Method  | Serialize | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|-------- |---------- |----------:|----------:|----------:|-------:|-------:|----------:|
| Ints    | False     |  9.740 us | 0.0540 us | 0.0505 us | 0.6104 | 0.0153 |  11.43 KB |
| Strings | False     |  7.168 us | 0.1009 us | 0.0944 us | 0.4883 | 0.0076 |   9.09 KB |
| Guids   | False     | 10.275 us | 0.1981 us | 0.2281 us | 1.1597 | 0.0610 |  21.59 KB |
| Ints    | True      | 12.197 us | 0.1425 us | 0.1333 us | 0.6561 | 0.0153 |  12.06 KB |
| Strings | True      | 10.457 us | 0.1583 us | 0.1481 us | 0.5493 | 0.0153 |  10.12 KB |
| Guids   | True      | 14.122 us | 0.1393 us | 0.1235 us | 1.5869 | 0.0916 |  29.27 KB |

### After

| Method  | Serialize | Mean     | Error     | StdDev    | Gen0   | Allocated |
|-------- |---------- |---------:|----------:|----------:|-------:|----------:|
| Ints    | False     | 1.075 us | 0.0088 us | 0.0083 us | 0.1087 |   2.01 KB |
| Strings | False     | 1.056 us | 0.0123 us | 0.0103 us | 0.1087 |   2.01 KB |
| Guids   | False     | 1.088 us | 0.0158 us | 0.0148 us | 0.1106 |   2.05 KB |
| Ints    | True      | 4.226 us | 0.0602 us | 0.0563 us | 0.2670 |   5.02 KB |
| Strings | True      | 3.300 us | 0.0404 us | 0.0378 us | 0.1640 |   3.07 KB |
| Guids   | True      | 4.807 us | 0.0538 us | 0.0503 us | 1.0376 |  19.11 KB |

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -947,6 +947,15 @@ rethrow the `OperationCanceledException` to the caller, regardless of the `Timeo
 Please review the documentation for the new complexity analyzer to understand how to use it and how to configure it.
 See the 'Complexity Analzyer' document with the 'Guides' section of the documentation for more information.
 
+### 26. Optimization when returning lists of scalars
+
+When returning a list of scalars, specifically of intrinsic types such as `int`, `string`, and `bool`, the matching
+scalar type (e.g. `IntGraphType`, `StringGraphType`, `BooleanGraphType`) will be used to serialize the list as a whole
+rather than each individual item. This can result in a significant performance improvement when returning large lists
+of scalars. Be sure the returned list type (e.g. `IEnumerable<int>`) matches the scalar type (e.g. `IntGraphType`)
+to take advantage of this optimization. Scalar types that require conversion, such as `DateTimeGraphType` are not
+currently optimized in this way.
+
 ## Breaking Changes
 
 ### 1. Query type is required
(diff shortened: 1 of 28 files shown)


gao-artur (approved):
Cool!

Issue #1497, opened 2020-01-11
Allow interfaces to implement other interfaces


sungam3r:
New feature from graphql spec: graphql/graphql-spec#373

darrenhull:
Any info on this being implemented? Its not great that this framework is lagging this far behind the specification.

sungam3r:
Not yet. Maybe in v5.

sungam3r:
Parser v8 (dev) supports that feature. @Shane32 it may be a challenge to support it here.

Issue #3875, opened 2024-01-02
Throw aggregated exception for schema validation errors


Shane32:
> In general, fixing these errors one by one is pretty hard because it requires restarting the application multiple times. I think gathering all the errors and throwing single/aggregated exception after the full validation can provide a better experience. This is usually not a problem during development but is annoying during the version upgrade.

_Originally posted by @gao-artur in https://github.com/graphql-dotnet/graphql-dotnet/pull/3867#discussion_r1439079811_

Shane32:
Done (mostly) in v8

PR #4008, opened 2024-08-10
Allow interfaces to implement other interfaces

Shane32:
Closes:
- #1497


src/GraphQL.Tests/Types/InterfaceGraphTypeTests.cs
```diff
@@ -46,6 +47,193 @@ public void VerifyDefaultArgumentsWork()
...
+        ValidateInterfaceInheritance(schema);
+    }
+
+    [Fact]
+    public void ReturnInterfacesRequiresResolver()
+    {
+        var schema = new MySchema2();
+        // the schema can be initialized without this field because no object graph type returns the interface type Image.
+        // here, since a field has been added that returns the interface type Image, the interface type must either
+        //   provide a resolveType function, or each possible type must provide an IsTypeOf function.
```
gao-artur, line 158: Intentional spaces?

gao-artur:
I haven't checked the existing tests for interfaces, so maybe this is already covered: should you also add tests to check that an interface can extend the implemented interface's field argument list and redefine the base fields with derived return types?

gao-artur:
I mean verifying the `IsValidImplementation(type, implementedType)` when `type` and `implementedType` are both interfaces.

Shane32:
Good thinking. I’m working on it. No tests existed directly for the applicable methods and there was a missing bit of new code. Argument testing for interface inheritance is not tested at all.

src/GraphQL.Tests/Validation/ValidationSchema.cs
```diff
@@ -47,7 +47,7 @@ public class Dog : ObjectGraphType
     public Dog()
     {
         Field<StringGraphType>("name")
-            .Argument<StringGraphType>("surname");
+            .Argument<BooleanGraphType>("surname");
```
gao-artur, line 50: Interesting how many mistakes like this we have in our schema 🥴
Must create an analyzer for interfaces. Otherwise, upgrading to v8 may be a hell...

src/GraphQL/Extensions/GraphQLExtensions.cs
```diff
@@ -174,24 +174,63 @@ public static Type GetNamedType(this Type type)
...
+            }
+
+            // any other arguments defined on the object field must be optional
+            if (field.Arguments?.Count > 0)
+            {
+                foreach (var arg in field.Arguments)
+                {
+                    if (arg.ResolvedType is NonNullGraphType && implementedField.Arguments?.Find(arg.Name) == null)
+                    {
+                        return throwError ? throw new ArgumentException($"Type {type.GetType().GetFriendlyName()} with name '{type.Name}' does not implement interface {implementedType.GetType().GetFriendlyName()} with name '{implementedType.Name}'. Field '{implementedField.Name}' argument '{arg.Name}' is not defined in the interface.") : false;
```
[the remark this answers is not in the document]
Shane32: Good idea

src/GraphQL.Tests/Types/InterfaceGraphTypeTests.cs
```diff
@@ -84,4 +300,212 @@ public MyObject()
...
+    {
+        public class Query
+        {
+            public Picture? Img => null!;
+        }
+
+        [Implements(typeof(INode))]
+        [Implements(typeof(IResource))]
+        [Implements(typeof(IImage))]
+        public class Picture : IImage
```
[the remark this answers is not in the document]
Shane32: There’s a few ways to attack this and I don’t know that I have the right answer. Currently we don’t enforce that for type-first that the .net types utilize inheritance. This makes things like adding arguments to an inherited method possible. (Although you could implement the original method privately or mark it with Ignore to mask it from GraphQL.) Also I’ve tried to keep attributes in the same relative location to the GraphQL schema. So while we could certainly and easily do what you suggested, someone reading the code would need to check all of the implemented interfaces of the CLR types to see if they are marked with GraphQLInterface or not. Further, since we don’t require marking with GraphQLInterface for other interface uses, this may be slightly confusing why it builds the interface but then doesn’t link any references to it. (Perhaps in the same way that it won’t now without ImplementsAtrribute…)

One question people might think is why doesn’t it just automatically generate GraphQL interfaces for all discovered interfaces on a CLR class. I don’t know about other people but this certainly wouldn’t work for me. I use inheritance within my code for database support and stuff that I don’t really want to expose to GraphQL. For example IHasDeleted marks any model that supports soft deletion - something that is oblivious to GraphQL callers. 

I don’t use GraphQL interfaces much. But I can see how your suggestion may be preferred, so you don’t need to constantly write Implements throughout your code, which is a bit redundant. We could add it as another option - people don’t have to use either one.

src/GraphQL.Tests/Validation/ValidationSchema.cs
Shane32: I’m not so concerned. If someone has a bug of this nature, it’s critically breaking their schema. The exceptions they will receive will describe the location of the error, and I would expect they would be happy to know about it. 

Also FYI, our code is slightly more lenient than the spec. It allows non-nullable args to be used for a nullable arg on an implemented field. (The spec says “invariant”.)

gao-artur: What is the behavior in v7 when you have such a bug in the schema? Will it fail the query? If it fails and users discover the mistake only during the upgrade to v8, it means the argument is not really used by the clients, so it's not a problem to fix the schema. However, if it works in v7 and stops working in v8, it requires fixing clients before upgrading to v8.
It would be great to add this validation to v7 and write it into logs as a warning to allow users to fix their schema before the upgrade, but unfortunately, we don't have a logger dependency. Then, an analyzer may be the best option for an easier upgrade. Report it as a warning in v7 and an error in v8. But the urgency depends on the answer to the question whether it fails the query or not.

src/GraphQL.Tests/Types/InterfaceGraphTypeTests.cs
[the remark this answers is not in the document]
gao-artur: I see. Yeah, it may be an enhancement to v8.1 then.

gao-artur (approved):

src/GraphQL.Tests/Validation/ValidationSchema.cs
Shane32: It will now fail initialization.  People can remove schema initialization checks if need be; execution will be the same.  So for example if the field had a non-null argument but the interface didn't, clients could execute a request on the interface via a fragment, which will cause the field resolver to execute without the required argument -- same as v7.

For reference, always in the past when we found similar missing validation checks, which would cause a breaking bug in the expected operation in certain cases, we have added the validation checks without hesitation, even sometimes on a minor version.  Basically, we consider such findings as bugs within GraphQL.NET, eligible for immediately fixing.  Obviously we prefer to make such changes on a major version.

Shane32: This perhaps would be quite valuable if many schema initialization errors are expected:
- #3875 

Do you think we should consider adding for v8?

gao-artur: > So for example if the field had a non-null argument but the interface didn't

But what about the wrong field type? I assume it's the same, right?

Shane32: It would execute the same as before.  In the situation here, it would read the boolean, and then fail during the call to `.GetArgument<string>`, I think, or however the `ValueConverter` is configured to deal with that conversion.

PR #4013, opened 2024-08-12
Reorder attribute method arguments

Shane32:
This additional `Modify` overload for GraphQLAttribute was added in #3952 targeting 7.9 as a prerequisite for federation changes in 8.0.  I think the order of arguments should be modified so that the overloads for field modification appear together - e.g. when typing `override` and looking at the intellisense popup.  This is not a breaking change as neither 7.9 nor 8.0 have been released yet.

With this change:

```
// modifies the graph type
public virtual void Modify(GraphQL.Types.IGraphType graphType) { }
public virtual void Modify(GraphQL.Types.IGraphType graphType, System.Type sourceType) { }

// modifies the field type
public virtual void Modify(GraphQL.Types.FieldType fieldType, bool isInputType) { }
public virtual void Modify(GraphQL.Types.FieldType fieldType, bool isInputType, GraphQL.Types.IGraphType graphType, System.Reflection.MemberInfo memberInfo, ref bool ignore) { }

// modifies the field argument
public virtual void Modify(GraphQL.Types.QueryArgument queryArgument) { }
public virtual void Modify(GraphQL.Types.QueryArgument queryArgument, System.Reflection.ParameterInfo parameterInfo) { }
```

Note that in #4011 I tried to add additional arguments to the last definition, but there's multiple reasons why it doesn't work [easily].  So in the end I was stuck with only the `queryArgument` and `parameterInfo`.  At least from the `ParameterInfo` you can get the defining method, and from that the defining type.

### DIFF
--- a/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
@@ -246,7 +246,7 @@ namespace GraphQL
         public virtual void Modify(GraphQL.Types.FieldType fieldType, bool isInputType) { }
         public virtual void Modify(GraphQL.Types.IGraphType graphType, System.Type sourceType) { }
         public virtual void Modify(GraphQL.Types.QueryArgument queryArgument, System.Reflection.ParameterInfo parameterInfo) { }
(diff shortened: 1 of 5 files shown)


gao-artur:
Looks good!

gao-artur (approved):

PR #4016, opened 2024-08-13
Aggregate schema initialization errors

### LINKED ISSUE #3875: Throw aggregated exception for schema validation errors

Shane32:
Closes #3875 

Note that until the schema passes basic validation, the applied directives' validation is not run.  And there are a few exceptions that are thrown outside of either validation class.  But for the most part, the validation occurs within those two schema visitors.

### DIFF
--- a/src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
+++ b/src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
@@ -8,13 +8,28 @@ namespace GraphQL.Utilities;
 /// </summary>
 public sealed class AppliedDirectivesValidationVisitor : ISchemaNodeVisitor
 {
-    /// <summary>
-    /// Returns a static instance of the <see cref="AppliedDirectivesValidationVisitor"/> class.
-    /// </summary>
-    public static readonly AppliedDirectivesValidationVisitor Instance = new();
+    private readonly List<Exception> _exceptions = new();
 
     private AppliedDirectivesValidationVisitor()
     {
+    }
+
+    /// <inheritdoc cref="AppliedDirectivesValidationVisitor"/>
+    public static void Run(ISchema schema)
+    {
+        var visitor = new AppliedDirectivesValidationVisitor();
+        visitor.Run(schema);
+        if (visitor._exceptions.Count > 0)
+            throw visitor._exceptions.Count == 1
+                ? visitor._exceptions[0]
+                : new AggregateException(
+                    "The schema is invalid. See inner exceptions for details.",
+                    visitor._exceptions);
+    }
+
+    private void ReportError(Exception ex)
+    {
+        _exceptions.Add(ex);
     }
 
     /// <inheritdoc/>
@@ -92,20 +107,25 @@ public sealed class AppliedDirectivesValidationVisitor : ISchemaNodeVisitor
             foreach (var directive in schema.Directives.List)
             {
                 if (!directive.Repeatable && appliedDirectives.Count(applied => applied.Name == directive.Name) > 1)
-                    throw new InvalidOperationException($"Non-repeatable directive '{directive.Name}' is applied to {GetElementDescription()} more than one time.");
+                    ReportError(new InvalidOperationException($"Non-repeatable directive '{directive.Name}' is applied to {GetElementDescription()} more than one time."));
             }
 
             foreach (var appliedDirective in appliedDirectives)
             {
-                var schemaDirective = schema.Directives.Find(appliedDirective.Name)
-                    ?? throw new InvalidOperationException($"Unknown directive '{appliedDirective.Name}' applied to {GetElementDescription()}.");
+                var schemaDirective = schema.Directives.Find(appliedDirective.Name);
+
+                if (schemaDirective == null)
+                {
+                    ReportError(new InvalidOperationException($"Unknown directive '{appliedDirective.Name}' applied to {GetElementDescription()}."));
+                    continue;
+                }
 
                 if (location != null && !schemaDirective.Locations.Contains(location.Value))
                 {
                     // TODO: think about strict check; needs to rewrite some tests (5)
                     // This is a temporary solution for @deprecated directive that we actually allow to more schema elements.
                     if (schemaDirective.Name != "deprecated")
-                        throw new InvalidOperationException($"Directive '{schemaDirective.Name}' is applied in the wrong location '{location.Value}' to {GetElementDescription()}. Allowed locations: {string.Join(", ", schemaDirective.Locations)}");
+                        ReportError(new InvalidOperationException($"Directive '{schemaDirective.Name}' is applied in the wrong location '{location.Value}' to {GetElementDescription()}. Allowed locations: {string.Join(", ", schemaDirective.Locations)}"));
                 }
 
                 if (schemaDirective.Arguments?.Count > 0)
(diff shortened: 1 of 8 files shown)


src/GraphQL/Utilities/Visitors/AppliedDirectivesValidationVisitor.cs
```diff
@@ -8,15 +8,30 @@ namespace GraphQL.Utilities;
...
     }
 
+    /// <inheritdoc cref="AppliedDirectivesValidationVisitor"/>
+    public static void Run(ISchema schema)
+    {
+        var visitor = new AppliedDirectivesValidationVisitor();
+        visitor.Run(schema);
+        if (visitor._exceptions.Count > 0)
+            throw visitor._exceptions.Count == 1
+                ? visitor._exceptions[0]
```
gao-artur, line 24: This exception will lose its original stack trace. If you don't want to wrap a single exception with `AggrigateException`, it would be better to change the `_exceptions` to `List<ExceptionDispatchInfo>`. But I'd just wrap it with `AggrigateException` so it's consistent.

gao-artur: Oh, actually, none of these exceptions were thrown, so they don't have any stack trace at all. Is that what we want?

gao-artur: After a bit more thinking, the stack trace of inner exceptions is not interesting at all. What's important it's an error message. So, probably, the inner exception type also has no value for the user, and we can just throw an exception with a list of error messages. What do you think?

Shane32: I’d much prefer to capture the stack trace but I didn’t want to try/catch repeatedly. Then again I realized that the inner stack traces are only useful for debugging GraphQL.NET itself, and the message can be used to determine the thrown exception point. 

I didn’t think of using try catch inside ReportError to generate the stack trace - that should work. Apart from that I could not find an official way to set StackTrace (without throwing).

I like the idea of using AggregateException. It’s a . NET way of indicating that multiple exceptions/errors occurred. I considered always throwing AggregateExceptiom, like some .NET APIs do, but there are a number of other schema exceptions that can be thrown and not wrapped. So instead I opted to unwrap by default unless multiple errors were encountered within the visitor. ExceptionDispatchInfo can be used to preserve the stack trace if it had been generated but since I’m not throwing it, the stack traces are never created, so there’s nothing to preserve, and it just throws the exception instance. 

In our codebases, we’ve used various different techniques to report a list of errors back to the caller. Now looking back on all the various error handling code we’ve written, I wish we just used AggregateException. So perhaps I’m a little biased and I’d like to hear your thoughts.

gao-artur: You will get the stack trace for the `AggregateException` pointing to this visitor class, which is good enough. Other exceptions will just look like this
```
System.InvalidOperationException: Non-repeatable directive 'xxx' is applied to yyy more than one time.
```

All the exceptions are created internally by this visitor, so their stack traces are redundant. You can always look for a message to find the exact location where they were thrown. And the `System.InvalidOperationException` adds no value, too.
Also, it can be difficult to see error messages in a long list of exceptions with stack traces. I'd prefer to see something like this

> The schema is invalid. Review and fix these issues:
> 1. Non-repeatable directive 'xxx' is applied to yyy more than one time.
> 2. Directive 'aaa' applied to bbb does not specify required argument 'ccc' that has no default value.
> 3. Directive 'xxx' applied to yyy does not specify required argument 'zzz' that has no default value.
>
> [ Stack trace of the outer exception goes here ]

Shane32: I'm a bit confused.  Isn't that pretty much what it does now?  `AggregateException.Message` returns text like this currently:

```
The schema is invalid. See inner exceptions for details. (The field 'in' of an Input Object type 'InputString' must have non-null 'ResolvedType' property for all types in the chain.) (The field 'not' of an Input Object type 'InputString' must have non-null 'ResolvedType' property for all types in the chain.)
```

The `ToString` method produces this output:

```
System.AggregateException: The schema is invalid. See inner exceptions for details. (The field 'in' of an Input Object type 'InputString' must have non-null 'ResolvedType' property for all types in the chain.) (The field 'not' of an Input Object type 'InputString' must have non-null 'ResolvedType' property for all types in the chain.)
 ---> System.InvalidOperationException: The field 'in' of an Input Object type 'InputString' must have non-null 'ResolvedType' property for all types in the chain.
   --- End of inner exception stack trace ---
   at GraphQL.Utilities.SchemaValidationVisitor.Run(ISchema schema) in C:\Users\Shane\source\repos\graphql-dotnet\graphql-dotnet\src\GraphQL\Utilities\Visitors\SchemaValidationVisitor.cs:line 25
   at GraphQL.Types.Schema.Validate() in C:\Users\Shane\source\repos\graphql-dotnet\graphql-dotnet\src\GraphQL\Types\Schema.cs:line 497
   at GraphQL.Types.Schema.CreateAndInitializeSchemaTypes() in C:\Users\Shane\source\repos\graphql-dotnet\graphql-dotnet\src\GraphQL\Types\Schema.cs:line 460
   at GraphQL.Types.Schema.Initialize() in C:\Users\Shane\source\repos\graphql-dotnet\graphql-dotnet\src\GraphQL\Types\Schema.cs:line 197
   at GraphQL.Tests.Initialization.SchemaInitializationTestBase.<>c__3`1.<ShouldThrowMultiple>b__3_0() in C:\Users\Shane\source\repos\graphql-dotnet\graphql-dotnet\src\GraphQL.Tests\Initialization\SchemaInitializationTestBase.cs:line 28
   at Shouldly.Should.ThrowInternal[TException](Action actual, String customMessage, String shouldlyMethod) in /_/src/Shouldly/ShouldStaticClasses/ShouldThrow.cs:line 26
 ---> (Inner Exception #1) System.InvalidOperationException: The field 'not' of an Input Object type 'InputString' must have non-null 'ResolvedType' property for all types in the chain.<---
```

I really don't think we should have the `Message` property return more than one line of text; it always seems problematic for both tests and users.

gao-artur: I didn't know that `AggregateException.Message` concatenates all the messages.
This `ToString` looks a bit weird... 
The first line is the AggregateException.Message 
The second line is the first inner exception
Then comes the AggregateException stack trace
And the last line is the second inner exception but called `(Inner Exception #1)` for some reason...

Shane32: > This `ToString` looks a bit weird...

Agreed, but it gets the job done.

gao-artur: Ok

gao-artur (approved):

PR #4017, opened 2024-08-13
Remove/review obsolete methods

Shane32:
Goals, to the extent reasonable:

- Field builder APIs should be marked as obsolete for 2 major versions
- Commonly-used infrastructure APIs should be marked as obsolete for 1 major version (e.g. legacy complexity analyzer, `IGraphQLBuilder` extension methods)
- Public APIs supporting internal infrastructure can be changed/removed during a major version bump (e.g. `IInterfaceGraphType`)

I've briefly reviewed all obsolete methods/classes in the codebase.  Most fell into a few different groups:

1. Field builder methods that did not require name parameter, marked as obsolete in v7.6.  These were noted to be removed in v9.  Note that this is a bit less than "2 major versions", but users can now apply fixes from analyzers.
2. Old field builder methods which returned `FieldType` instead of `FieldBuilder`.  These were marked as obsolete in 7, so allowing for 2 major versions (version 7 and version 8), means they should be removed in version 9.
3. Other APIs that were marked as obsolete in v8, such as the legacy complexity analyzer.  These should be removed in v9.

Notes:
- There were a very few APIs that were marked as obsolete in v5.  Those have been removed in this PR.
- I did not see any field builder methods marked as obsolete in v5 still remaining in the codebase.
- All obsolete messages have been updated to include the version we expect to remove them from the codebase.
- It would be good to check all obsolete field builder methods to see if they all have analyzers to expedite upgrading.  I did not do so.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Issue2275.cs
+++ b/src/GraphQL.Tests/Bugs/Issue2275.cs
@@ -71,7 +71,7 @@ public class Issue2275
         }
     }
 
-    [GraphQLMetadata(InputType = typeof(FilterInputGraphType))]
+    [InputType(typeof(FilterInputGraphType))]
     public class Filter
     {
         public string Key { get; set; }
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -1351,3 +1351,11 @@ If you have custom implementations of these interfaces, you may need to update t
 
 Verfies that the arguments defined on fields of interfaces are also defined on fields of implementing types,
 pursuant to GraphQL specifications.
+
+### 28. APIs marked obsolete since v5 have been removed
+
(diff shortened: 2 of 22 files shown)


src/GraphQL.Tests/Bugs/Issue2275.cs
```diff
@@ -71,7 +71,7 @@ public FilterInputGraphType()
         }
     }
 
-    [GraphQLMetadata(InputType = typeof(FilterInputGraphType))]
+    [InputType(typeof(FilterInputGraphType))]
```
gao-artur, line 74: Please update the documentation to use these attributes

Shane32: I don't see any reference to the removed `InputType` property within the documentation -- ??

gao-artur: Oh right, my mistake!

gao-artur (approved):

PR #4025, opened 2024-08-19
Add IResolveFieldContext.ExecutionContext

Shane32:
With the new variable parsing in #3424 , the `ExecutionHelper.GetArguments` signature has changed, and it is not preferred to use that method anymore since all the arguments have already been parsed -- it's essentially an internal detail now.  Because of this, code that synthesizes an `IResolveFieldContext` for child fields within a resolver, or any similar code which needs to examine arguments of child fields, needs modification.  This PR adds `IResolveFieldContext.ExecutionContext` so that code can easily pull already-parsed argument values for any part of the document.

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -27,7 +27,7 @@ GraphQL.NET v8 is a major release that includes many new features, including:
 - Optimize scalar lists (e.g. `ListGraphType<IntGraphType>`)
 - Allow GraphQL interfaces to implement other GraphQL interfaces (based on spec)
 
-Some of these features require changes to the infrastrucutre, which can cause breaking changes during upgrades.
+Some of these features require changes to the infrastructure, which can cause breaking changes during upgrades.
 Most notably, if your server uses any of the following features, you are likely to encounter migration issues:
 
 - Custom validation rules, schema node visitors, or IResolveFieldContext implementations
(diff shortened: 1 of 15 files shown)


Shane32:
@gao-artur The number of changes in v8, and their benefits, makes it feel more like jumping from v7 to v10.  Did you know that the migration readme is 1500 lines of text, larger than any prior migration document?!?

So far, all of my internal GraphQL helper NuGet packages need to be updated for v8.  Even the most basic validation rule doesn't work with v8 anymore because of the signature changes to `IValidationRule`, for example.  So far all the changes have been very easy and fairly small -- but all the individual modules do need to be updated, so it's not a 5-minute version bump in the main repo.

On the other hand, version 7.x has been very stable; it was first released almost exactly 2 years ago with nearly 8 million downloads.  Quite pleased with that -- 2 years with no breaking changes.  Version 5.x only was around for 6 months, version 4.x for 1 year, and version 3.x for 6 months.  If we do a good job with the initial release here, perhaps 8.x will also have a long lifetime.

gao-artur:
> The number of changes in v8, and their benefits, makes it feel more like jumping from v7 to v10. Did you know that the migration readme is 1500 lines of text, larger than any prior migration document?!?

Impressive and frightening at the same time 😄

For us, a major version upgrade was never a five-minute job. It always took at least a day to make compile changes and 2-5 days to fix runtime errors. So, there's nothing special here...

gao-artur (approved):

PR #4029, opened 2024-08-21
Pin .NET 8.0.400 or later 8.0.x SDK

SimonCropp:
SDK 9 is more strict with vulnerability checks.

eg

```
  GraphQL.SystemTextJson.csproj: [NU1904] Warning As Error: Package 'System.Text.Encodings.Web' 4.7.1 has a known critical severity vulnerability, https://github.com/advisories/GHSA-ghhp-997w-qr28
```

this PR forces sdk 8 to unblock people to contribute


src/global.json
```diff
@@ -0,0 +1,7 @@
+﻿{
+  "sdk": {
+    "version": "8.0.401",
+    "allowPrerelease": false,
+    "rollForward": "latestFeature"
+  }
+}
```
Shane32, line 7:
```suggestion
}
```

Shane32: Can we include the terminating lf ?

src/global.json
```diff
@@ -0,0 +1,7 @@
+﻿{
+  "sdk": {
+    "version": "8.0.401",
```
Shane32, line 3:
```suggestion
    "version": "8.0.400",
```
The lastest version of Visual Studio only installs 8.0.400.  Let's use 8.0.400 or a prior version.  (I was running 8.0.302 a few minutes ago.)

Shane32 (commented):
I don't usually limit to a specific SDK, but this looks fine, with the rollfoward setting enabled.  :+1:

src/global.json
SimonCropp: 8.0.401 is a security patch

SimonCropp:
not that this is only temporary. the cves will beak the build when the project eventually moves to .net 9

Shane32:
What about if we just do #4030 instead, fixing the problem for both .NET 8 and .NET 9?  Or both PRs?

Shane32 (approved):

PR #4030, opened 2024-08-21
Set NuGet audit mode

Shane32:


Shane32:
Other options:
- Disable audits for test projects
- Add a dependency in GraphQL.SystemTextJson to System.Text.Encodings.Web 4.7.2 (for .NET Standard builds), and set mode to all

gao-artur (approved):

PR #4033, opened 2024-08-21
Fix naming bug

Shane32:
Fixes bugs:
1. `MyClass<T>.Class1` throws an exception because Class1 is generic (compiled as `Class1<T>` under the hood) but does not have a name of "Name`1"
2. Generic parent types do not strip backtick when `useDeclaringTypeNames` is `true`.


src/GraphQL.Tests/Extensions/TypeExtensionTests.cs
```diff
@@ -10,66 +10,75 @@ public class TypeExtensionTests
...
+    [InlineData(85, true, false, typeof(ConnectionType<StringGraphType>), "Connection")]
+    [InlineData(86, true, false, typeof(Connection<string>), "Connection")]
+    [InlineData(87, true, false, typeof(Test1<string>.Test2), "Test2")]
+    [InlineData(88, true, false, typeof(Test1<string>.Test3<Class1>), "Test3")]
+    [InlineData(89, true, true, typeof(Test1<string>.Test2), "Test2")]
+    [InlineData(90, true, true, typeof(Test1<string>.Test3<Class1>), "Test3")]
 
     // note: cannot test F# anonymous class names within C# tests
-    public void GraphQLNameTest(bool useLegacyTypeNaming, bool useDeclaringTypeNames, Type type, string expected)
+    public void GraphQLNameTest(int i, bool useLegacyTypeNaming, bool useDeclaringTypeNames, Type type, string expected)
```
Shane32, line 79: Added numbering.  Added tests 37-40 and 87-90.  Tests 37, 39, 87, and 89 failed because of bug 1 before.  Tests 38 and 40 failed because of bug 2 before.

PR #4034, opened 2024-08-21
Add global assembly Description attribute

gao-artur:
Add a global assembly Description attribute, which will be used by analyzers to ensure the analyzed symbol is defined by the GraphQL.NET libraries and not by the user code.


Shane32 (approved):

Shane32:
@gao-artur Check out #4035 before merging

PR #4035, opened 2024-08-22
use public key for GraphQL* assembly check

SimonCropp:


SimonCropp:
@Shane32 this should be good to go

gao-artur:
Sorry, I'm not familiar with assembly signing. How does it work? Does it happen durring the build or after the build? I mean, will it be signed when debugging the code in VS?

Shane32:
Assembly signing applies some kind of digital signature to the file when building.  This lets .NET Framework applications load two versions of the same assembly at the same time and stuff -- but it's pretty pointless in .NET Core.  For the purposes here, once the assembly is built, it will be digitally signed with the snk file, and we can inspect loaded assemblies to see if it has the matching signature.

As for debugging the GraphQL.NET solution, maybe we just write a test to ensure that it correctly identifies the GraphQL.NET assemblies.  What I don't know is if it will identify the non-packable projects also (samples, etc).

Shane32:
Maybe @SimonCropp can provide additional insight

gao-artur:
It looks like it's controlled by the [SignAssembly](https://github.com/graphql-dotnet/graphql-dotnet/blob/00457da576a6f23dc9aeccf764611c3fe2e530db/src/Directory.Build.targets#L17-L19) property which is [true](https://github.com/graphql-dotnet/graphql-dotnet/blob/00457da576a6f23dc9aeccf764611c3fe2e530db/src/Directory.Build.props#L23) for all projects except [GraphQL.Harness](https://github.com/graphql-dotnet/graphql-dotnet/blob/00457da576a6f23dc9aeccf764611c3fe2e530db/src/GraphQL.Harness/GraphQL.Harness.csproj#L7) and [GraphQL.Harness.Tests](https://github.com/graphql-dotnet/graphql-dotnet/blob/00457da576a6f23dc9aeccf764611c3fe2e530db/src/GraphQL.Harness.Tests/GraphQL.Harness.Tests.csproj#L13). So I believe it will be signed when debugging.

There is a test [FieldAndNameMethodsCalled_NotGraphQLBuilder_NoDiagnostics](https://github.com/graphql-dotnet/graphql-dotnet/blob/00457da576a6f23dc9aeccf764611c3fe2e530db/src/GraphQL.Analyzers.Tests/FieldNameAnalyzerTests.cs#L23) that checks that `Field<T>().Name()` and `Connection<T>().Name()` methods defined by non GraphQL.NET builder types are not analyzed. It would be better to create a separate test, not coupled with a specific analyzer. I think it means manually creating `Workspace,` `Compilation`, and friends. @SimonCropp, do you know how to do that?

SimonCropp:
sorry. no i dont

gao-artur (approved):

gao-artur:
Ok, I'll do it later. Thank you!

PR #4039, opened 2024-08-24
[WIP] POC - PatternMatchingVisitor support AOT-compiled Regex instances

### LINKED ISSUE #4037: Support argument of Regex type by PatternMatchingVisitor
Currently, the `PatternMatchingVisitor` only supports `string` argument

https://github.com/graphql-dotnet/graphql-dotnet/blob/22052e73d734d3f0b60b52a04897afb12656c55a/src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs#L40-L41

It can be useful also to add support for `Regex` arguments
```c#
Field<string>("Hello")
    .Argument<StringGraphType>("arg", argument => argument.ApplyDirective("pattern", "regex", Patterns.AlphabeticalPattern()))
    .Resolve(ctx => ctx.GetArgument<string>("arg"));

public static partial class Patterns
{
    [GeneratedRegex("[A-Z]+")]
    public static partial Regex AlphabeticalPattern();
}
```

Shane32:
Closes #4037

### DIFF
--- a/src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
+++ b/src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
@@ -37,21 +37,46 @@ public class PatternMatchingVisitor : BaseSchemaNodeVisitor
     {
         // look for @pattern directive applied to the field argument or input field, and if found, use Validate to set the validation method
         var applied = fieldArgOrInputField.FindAppliedDirective("pattern");
-        if (applied?.FindArgument("regex")?.Value is not string regex)
-            return null;
+        var value = applied?.FindArgument("regex")?.Value;
 
-        // if GlobalSwitches.DynamicallyCompileToObject is true, then compile
-        // the regex (or pull from the static cache if already compiled)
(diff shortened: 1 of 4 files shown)


src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
```diff
@@ -37,21 +37,41 @@ public override void VisitInterfaceFieldArgumentDefinition(QueryArgument argumen
...
-            GlobalSwitches.DynamicallyCompileToObject ? RegexOptions.Compiled : RegexOptions.None);
+        var regexObject = value switch
+        {
+            // if GlobalSwitches.DynamicallyCompileToObject is true, then compile
+            // the regex (or pull from the static cache if already compiled)
+            // (note that GlobalSwitches.DynamicallyCompileToObject is false for AOT schemas and scoped schemas)
+            string regex => new Regex($"^{regex}$",
+                GlobalSwitches.DynamicallyCompileToObject ? RegexOptions.Compiled : RegexOptions.None),
+            Regex r => r,
+            _ => null
```
gao-artur, line 50: I think it should throw if the value is of not supported type rather than silently skipping the directive (same is in the current version)

PR #4041, opened 2024-08-25
Fix SerializeList code

### LINKED ISSUE #4040: Fail to serialize list of strings into `ListGraphType<IdGraphType>`
Caused by 
- #4001

```c#
Field<ListGraphType<IdGraphType>>("test").Resolve(_ => new[] { "a", "b" });
```

Fails with exception
> System.NotSupportedException: Specified method is not supported.
   at GraphQL.Types.IdGraphType.SerializeList(IEnumerable list) in /_/src/GraphQL/Types/Scalars/IdGraphType.cs:line 101
   at GraphQL.Execution.ExecutionStrategy.SetArrayItemNodesAsync(ExecutionContext context, ArrayExecutionNode parent) in /_/src/GraphQL/Execution/ExecutionStrategy.cs:line 430
   at GraphQL.Execution.ExecutionStrategy.CompleteNodeAsync(ExecutionContext context, ExecutionNode node) in /_/src/GraphQL/Execution/ExecutionStrategy.cs:line 587

Shane32:
Closes #4040

### DIFF
--- a/src/GraphQL.Tests/Types/BooleanGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/BooleanGraphTypeTests.cs
@@ -1,3 +1,5 @@
+using System.Collections;
+using System.Numerics;
 using GraphQL.Types;
 
 namespace GraphQL.Tests.Types;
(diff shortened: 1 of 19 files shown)


Shane32:
Pfff.  `CanSerializeList` doesn't even work properly.  That's what you get when I don't write sufficient tests, I guess.

gao-artur:
Wow! That's what you call sufficient tests? 😆

gao-artur (approved):

Shane32:
This is rather critical code.

gao-artur:
I think it would be better to improve the error message too. The`System.NotSupportedException: Specified method is not supported` wasn't very friendly and it took time to find out what exactly failed.

Shane32:
> I think it would be better to improve the error message too. The`System.NotSupportedException: Specified method is not supported` wasn't very friendly and it took time to find out what exactly failed.

True, but it shouldn’t be possible to reach that line of code under any circumstances.

PR #4044, opened 2024-08-26
Validate regex directive argument type

gao-artur:


src/GraphQL.Tests/Utilities/Visitors/PatternMatchingVisitorTests.cs
```diff
@@ -57,4 +57,21 @@ public static string Hello(
...
     }
+
+    [Theory]
+    [InlineData(null, "Pattern directive 'regex' argument must have non-null value")]
+    [InlineData(123, "Pattern directive 'regex' argument must be of 'string' type")]
+    public void InvalidRegexValue(object? argumentValue, string message)
+    {
+        var query = new ObjectGraphType { Name = "Query" };
+        query.Field<string>("Hello")
+            .Argument<StringGraphType>("arg", argument => argument.ApplyDirective("pattern", "regex", argumentValue))
```
gao-artur, line 68: Btw, I noticed that `argumentValue` in `ApplyDerective` is defined as nullable, but in `DirectiveAttribute` as non-nullable...

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
```diff
@@ -37,9 +37,16 @@ public override void VisitInterfaceFieldArgumentDefinition(QueryArgument argumen
     {
         // look for @pattern directive applied to the field argument or input field, and if found, use Validate to set the validation method
         var applied = fieldArgOrInputField.FindAppliedDirective("pattern");
-        if (applied?.FindArgument("regex")?.Value is not string regex)
+        var regexArgument = applied?.FindArgument("regex");
+        if (regexArgument == null)
             return null;
 
+        if (regexArgument.Value == null)
+            throw new ArgumentException("Pattern directive 'regex' argument must have non-null value");
```
Shane32, line 45: Do you want to include the location as well?

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
```diff
@@ -37,9 +37,16 @@ public override void VisitInterfaceFieldArgumentDefinition(QueryArgument argumen
...
+
+        if(regexArgument.Value is not string regex)
+            throw new ArgumentException("Pattern directive 'regex' argument must be of 'string' type");
```
Shane32, line 48: FYI, I think most of our XML comments and error messages end with a period.  (Not all yet though.)

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
Shane32: Should be easy if you pass in a few more args into `GetValidator` I think.

Shane32: This throws during schema validation so a location would probably be valuable.

Shane32: If we wanted to get fancier we could catch and rewrap exceptions caught in the regex constructor (in case of invalid patterns) to add the location as well.

Shane32 (approved):
See above comments, although whatever you want is fine with me.  I'll release 8.0.2 when done.

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
```diff
@@ -37,9 +37,16 @@ public override void VisitInterfaceFieldArgumentDefinition(QueryArgument argumen
...
         var applied = fieldArgOrInputField.FindAppliedDirective("pattern");
-        if (applied?.FindArgument("regex")?.Value is not string regex)
+        var regexArgument = applied?.FindArgument("regex");
+        if (regexArgument == null)
             return null;
 
+        if (regexArgument.Value == null)
+            throw new ArgumentException("Pattern directive 'regex' argument must have non-null value");
+
+        if(regexArgument.Value is not string regex)
```
Shane32, line 47:
```suggestion
        if (regexArgument.Value is not string regex)
```

src/GraphQL/Utilities/Visitors/Custom/PatternMatchingVisitor.cs
gao-artur: How do I do that?

Shane32: I'm thinking something like this:

Change line 36 & 37 to:

```cs
private static Action<object?>? GetValidator(QueryArgument? argument, FieldType field, IInterfaceGraphType type)
{
    var fieldArgOrInputField = argument ?? field;
```

Then change the exception message to:

```cs
throw new ArgumentException($"Pattern directive 'regex' argument at {GetLocation()} must be of 'string' type");
```

with a local method

```cs
string GetLocation()
    => argument != null
        ? $"{type.Name}.{field.Name}.{argument.Name}"
        : $"{type.Name}.{field.Name}";
```

Shane32 (approved):

PR #4052, opened 2024-09-03
Add attributes for Parser, Validator and ValidateArguments

Shane32:


src/GraphQL/Attributes/ParserAttribute.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// The method must exist on the declaring type of the field and be declared as static.
+    /// The method may be private or public and must have the signature 'object <paramref name="parserMethodName"/>(object value)'.
+    /// If it is public, you may wish to mark the method with <see cref="IgnoreAttribute"/> to prevent it from being exposed in the schema.
+    /// </summary>
+    [RequiresUnreferencedCode("Please ensure the specified method is not trimmed or use an alternative constructor.")]
+    public ParserAttribute(string parserMethodName)
+    {
+        _parserMethodName = parserMethodName
+            ?? throw new ArgumentNullException(nameof(parserMethodName));
+        _includePrivate = true;
```
gao-artur, line 28: Why not allow `protected` as well? Also, allowing `internal` when the type is specified sounds useful.

gao-artur:
Do you want an analyzer for these attributes?

docs2/site/docs/migrations/migration8.md
```diff
@@ -341,6 +411,29 @@ and throwing other exceptions will be caught by the unhandled exception handler.
...
+validator as shown in the below example:
+
+```csharp
+// sample for argument validation
+public class OutputClass3
+{
+    public static string Hello1(string str1, string str2) => str1 + str2;
+
+    [ValidateArguments(nameof(ValidateHelloArguments))]
+    public static string Hello2(string str1, string str2) => str1 + str2;
```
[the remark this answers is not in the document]
Shane32: The example was shortened for clarity as it seemed repetitious; it supports the same constructors as the other attributes.

src/GraphQL/Attributes/ParserAttribute.cs
Shane32: BindingFlags are public and non public. This statement actually allows all non public including protected and internal. 

Normally a reference like this would only reference public members, but since with type-first public members are exposed through GraphQL, I allowed non public members for this constructor. The other constructors can refer to internal or private classes but the method must be declared as public.

Shane32:
> Do you want an analyzer for these attributes?

I was hoping you’d ask! I think that would be really useful.

docs2/site/docs/migrations/migration8.md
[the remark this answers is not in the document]
Shane32: Should I repeat the other examples here also?

Shane32:
One more question: Should we rename ValidateArguments? Would ArgumentValidator seem more consistent with Parser and Validator? We have not released 8.1 yet so we can still change it.

docs2/site/docs/migrations/migration8.md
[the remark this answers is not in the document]
gao-artur: Maybe this is enough. We have intellisense after all.

gao-artur:
> > Do you want an analyzer for these attributes?
> 
> I was hoping you’d ask! I think that would be really useful.

Will do 🙂

> One more question: Should we rename ValidateArguments? Would ArgumentValidator seem more consistent with Parser and Validator? We have not released 8.1 yet so we can still change it.

Do you mean rename the `FieldType`'s property and argument? Sounds reasonable.

gao-artur:
Thinking again, `Validator` and `ArgumentValidator` names are confusing. It sounds like they do the same thing...

Shane32:
> Thinking again, `Validator` and `ArgumentValidator` names are confusing. It sounds like they do the same thing...

So, leave it as-is?

Shane32:
> I don't know. Do you have better names in mind?

Not really. 

ArgumentsValidator (plural)

Or leave as-is

gao-artur:
As is then

gao-artur (approved):

src/GraphQL/Attributes/ParserAttribute.cs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /// <summary>
+    /// Specifies a custom parser method for a field of an input object in a GraphQL schema using the specified method from a specified type.
+    /// The method must exist on the specified type and must be static and public.
+    /// The method must have the signature 'object <paramref name="parserMethodName"/>(object value)'.
+    /// </summary>
+    public ParserAttribute(
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+        Type parserType, string parserMethodName)
```
gao-artur, line 52: @Shane32, I think, when the `parserType` defines the same type where `Parser` attribute is applied (even though it can be omitted), it should also set `_includePrivate = true`

Shane32: Sure; makes sense

src/GraphQL/Attributes/ParserAttribute.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// The method must exist on the specified type and must be static and public.
+    /// The method must have the signature 'object Parser(object value)'.
+    /// </summary>
+    public ParserAttribute(
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+        Type parserType)
+    {
+        _parserType = parserType
+            ?? throw new ArgumentNullException(nameof(parserType));
+        _parserMethodName = nameof(FieldType.Parser);
```
gao-artur, line 42: @Shane32 I think the default name should be changed to verb `Parse` instead of noun `Parser`

Shane32: Ok

PR #4054, opened 2024-09-05
Change Facebook reference

mishmanners:
The old Facebook reference now links to: https://github.com/graphql/graphql-spec, meaning this is now hosted directly by GraphQL and no longer makes mention of Facebook.


MrHPotter:
@Shane32 if u don't appreciate being @`ed let me know and i will stop.
Otherwise i checked the link and op is right. Should be a quick and easy win to merge and close :)

Shane32:
> @Shane32 if u don't appreciate being @`ed let me know and i will stop. Otherwise i checked the link and op is right. Should be a quick and easy win to merge and close :)

I don't mind at all.  I just have not had time to work on this PR.

It's correct that the link now just takes you to the spec.  But, consider the changed sentence along with the following one:

>This is an implementation of [Facebook's GraphQL](https://github.com/facebook/graphql) in .NET.
>
> Now the [specification](https://github.com/graphql/graphql-spec) is being developed by the [GraphQL Foundation](https://foundation.graphql.org/).

The first sentence is providing a historic fact and the second is providing current status.  Changing the first sentence as proposed would then make the sentence redundant as the second sentence already provides this information along with the current link.

I think a change is needed -- at very least, the dead link should be removed.  Perhaps also the context of the sentence should be clarified.

Shane32:
Fixed by #4109 

Thanks for the input !

PR #4063, opened 2024-09-15
Add DefaultAstValueAttribute to set default values of complex objects in type-first schemas

Shane32:
This attribute sets the default value of an input object's field or an output type's field argument to a `GraphQLValue` instance parsed from the provided string.  Already there is functionality within the schema initialization code to create an instance of the correct object type from the `GraphQLValue` instance and apply it during schema initialization.

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -1215,7 +1215,7 @@ public class MyInputObject2
     public MyInputObject1 Json { get; set; }
 }
 
-// demonstrates setting a default value for an output field arguments
+// output type
 public class MyOutputObject
 {
     // typical way to set a default value of an output field argument
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -1194,6 +1194,38 @@ The `ExecutionContext` property has been added to the `IResolveFieldContext` int
 underlying execution context. This is useful for accessing the parsed arguments and directives from the operation
 via `IExecutionContext.GetArguments` and `GetDirectives`.
 
+### 29. `[DefaultAstValue]` attribute added (v8.1 and newer) to set default values for complex types with type-first schemas
+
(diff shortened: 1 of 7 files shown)


src/GraphQL/Attributes/DefaultAstValueAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Parameter)]
+public class DefaultAstValueAttribute : GraphQLAttribute
+{
+    private readonly GraphQLValue _value;
+    private static readonly ParserOptions _options = new() { Ignore = IgnoreOptions.All };
+
+    /// <inheritdoc cref="DefaultAstValueAttribute"/>
+    public DefaultAstValueAttribute(string astValue)
+    {
+        _value = Parser.Parse<GraphQLValue>(astValue, _options);
```
gao-artur, line 20: I can try doing this in analyzer to verify the correctness of the value. But this may be challenging because I need to reference the same version as the analyzed project...

docs2/site/docs/migrations/migration8.md
```diff
@@ -1194,6 +1194,38 @@ The `ExecutionContext` property has been added to the `IResolveFieldContext` int
...
+{
+    [DefaultValue("value")]
+    public required string Field1 { get; set; }
+}
+
+// demonstrates setting a default value for an input field that has a complex type
+public class MyInputObject2
+{
+    [DefaultAstValue("{ field1: \"value\" }")]
+    public MyInputObject1 Json { get; set; }
```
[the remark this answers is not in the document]
Shane32: Yes that would be required. The AST is parsed after the schema is initialized.

src/GraphQL/Attributes/DefaultAstValueAttribute.cs
Shane32: If you like. While it would be nice, I don’t think it’s necessary.

gao-artur (approved):

PR #4065, opened 2024-09-15
Bug fix for issue 4064: object cannot be coerced to a list of objects

Shane32:
Already works for ints and strings, but IDictionary is incorrectly recognized as a list


gao-artur (approved):

Shane32:
@gao-artur Just wondering if you want to add more features into 8.1 or if we should release it now or wait for a bit?  No rush necessarily, but if it was released now, I'd make use of the new features and bug fix.

gao-artur:
Let's releases it. I have nothing to add.

Shane32:
@gao-artur released.  You can move the new analyzers to 'shipped' now.

PR #4068, opened 2024-09-18
Add .git-blame-ignore-revs to improve git blame

MikeAlhayek:
The `.git-blame-ignore-revs` file is used in Git to specify a list of commits that should be ignored when running the `git blame` command.


Shane32 (approved):

gao-artur:
Nice!

PR #3870, opened 2024-01-01
Better error handling of requests with invalid operation names

Shane32:
Closes #3318

PR #4078, opened 2024-10-10
Bugfix: Treat operation name with empty string as null

### LINKED ISSUE #4077: DocumentExecuter.GetOperation() does not check for empty operation name
The DocumentExecuter.GetOperation() does the following check:

```
protected virtual GraphQLOperationDefinition GetOperation(string? operationName, GraphQLDocument document)
    {
        if (operationName == null)
        {
            GraphQLOperationDefinition? match = null;
            foreach (var def in document.Definitions)
            {
                if (def is GraphQLOperationDefinition op)
                {
                    if (match != null)
                        throw new NoOperationNameError();
                    match = op;
                }
            }

            return match ?? throw new NoOperationError();
        }

        foreach (var def in document.Definitions)
        {
            if (def is GraphQLOperationDefinition op && op.Name == operationName)
                return op;
        }

        throw new InvalidOperationNameError(operationName);
    }
```

This means, if `operationName` is an empty string (not `null`) if would through `InvalidOperationNameError`. I don't see the reason for this. Is should check empty string too. (`if (string.IsNullOrEmpty(operationName))`) .

In 7.x, this used to be as follows:
```
var operation = GetOperation(options.OperationName, document);
if (operation == null)
{
   throw new InvalidOperationError($"Query does not contain operation '{options.OperationName}'.");
}
```

```
protected virtual GraphQLOperationDefinition? GetOperation(string? operationName, GraphQLDocument document)
            => document.OperationWithName(operationName);
```
Which checked for empty string too.

Shane32:
Closes #4077 

Regression introduced in:
- #3870

### DIFF
--- a/src/GraphQL.Tests/Bugs/Bug1772.cs
+++ b/src/GraphQL.Tests/Bugs/Bug1772.cs
@@ -26,7 +26,6 @@ public class Bug1772 : QueryTestBase<Bug1772Schema>
     }
 
     [Theory]
-    [InlineData("")]
     [InlineData("thirdQuery")]
     [InlineData("query")]
     [InlineData("test")]
--- a/src/GraphQL.Tests/Bugs/Bug4078.cs
+++ b/src/GraphQL.Tests/Bugs/Bug4078.cs
@@ -2,7 +2,8 @@ using GraphQL.Types;
 
 namespace GraphQL.Tests.Bugs;
 
-public class Issue4077
+// https://github.com/graphql-dotnet/graphql-dotnet/pull/4078
+public class Bug4078
 {
     [Fact]
     public async Task empty_string_for_operation_name_is_treated_as_null()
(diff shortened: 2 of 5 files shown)


src/GraphQL.Tests/Bugs/Bug1772.cs
```diff
@@ -26,7 +26,6 @@ public async Task DocumentExecuter_works_for_valid_operation(string? operationNa
     }
 
     [Theory]
-    [InlineData("")]
```
Shane32, line 29: The code previously tested to ensure that passing `""` would throw an operation-not-found exception.

gao-artur (approved):
Hi @Shane32. Sorry for the delay in reviews. I'm on vacation far from civilization and without my laptop. I'll review other PRs next week because it's hard to read more than a couple of lines from mobile screen.

Shane32:
> Hi @Shane32. Sorry for the delay in reviews. I'm on vacation far from civilization and without my laptop. I'll review other PRs next week because it's hard to read more than a couple of lines from mobile screen.

No problem; have a good vacation!  There is no pressing urgency, and I've been extremely busy myself.

PR #4079, opened 2024-10-15
Add parser options in IDocumentBuilder

vincent-duret:
When using DI, it might be useful to be able to set the options using the interface directly without having to cast it to GraphQLDocumentBuilder.

Also in some scenarii, it might be convenient to be able to pass the parser options to the Build method.

### DIFF
--- a/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
@@ -1394,10 +1394,15 @@ namespace GraphQL.Execution
         public bool IgnoreLocations { get; set; }
         public int? MaxDepth { get; set; }
         public GraphQLParser.AST.GraphQLDocument Build(string body) { }
+        public GraphQLParser.AST.GraphQLDocument Build(string body, GraphQLParser.ParserOptions options) { }
     }
     public interface IDocumentBuilder
     {
+        bool IgnoreComments { get; set; }
+        bool IgnoreLocations { get; set; }
+        int? MaxDepth { get; set; }
         GraphQLParser.AST.GraphQLDocument Build(string body);
+        GraphQLParser.AST.GraphQLDocument Build(string body, GraphQLParser.ParserOptions options);
     }
     public interface IDocumentExecutionListener
     {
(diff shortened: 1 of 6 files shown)


Shane32:
The purpose of the interface is to provide a way for an instance of `IDocumentExecuter` to build the document, which only requires a single `Build` method. The interface specifically does not include options or additional methods as they are outside the scope of the interface. A good example from the .NET framework is the `IComparer<T>` interface, which also follows this principle. `IComparer<T>` only defines a single `Compare` method for comparing two objects and does not expose any additional configuration. Any options, such as case sensitivity or culture-specific comparisons, are handled by specific implementations, not the interface itself. Similarly, `IDocumentBuilder` keeps its focus on document construction, leaving configuration to the implementing class, like `GraphQLDocumentBuilder`.

You can use these techniques to customize the document builder:

```cs
services.AddSingleton<IDocumentBuilder>(new GraphQLDocumentBuilder() { IgnoreComments = true });

// or, assuming you create a GraphQLDocumentBuilderOptions class:
services.AddSingleton<IDocumentBuilder>(provider =>
{
    var options = provider.GetRequiredService<IOptions<GraphQLDocumentBuilderOptions>>().Value;
    return new GraphQLDocumentBuilder()
    {
        IgnoreComments = options.IgnoreComments,
    };
});
```

Since `ParserOptions` is a `struct`, adding a constructor that accepts `ParserOptions` is not useful for MS options extensions, since options classes must be a reference type.

Upon request we may add a builder method so that you can do this:

```cs
services.AddGraphQL(b => b
    .AddAutoSchema<QueryType>()
    .AddDocumentBuilder(options => options.IgnoreComments = true)
);

//or

services.AddGraphQL(b => b
    .AddAutoSchema<QueryType>()
    .AddDocumentBuilder((options, serviceProvider) => options.IgnoreComments = true)
);

//or

services.AddGraphQL(b => b
    .AddAutoSchema<QueryType>()
    .AddDocumentBuilder<MyDocumentBuilder>()
);
```

Shane32:
> Also in some scenarii, it might be convenient to be able to pass the parser options to the Build method.

In such cases, it may be better for the caller to utilize the underlying GraphQL-Parser library directly.

vincent-duret:
```
Please let me know if you think this extension method should be added.
``` 
I am not using those extensions method in my current project, but it might be useful for others.

Shane32:
I'll skip for now then.  The `IGraphQLBuilder` extension methods are designed for the majority of use cases, not every use case.  To reconfigure the document builder's default set of options seems to me to be a very rare need, as (1) comments are unnecessary to execute a request and add no benefit (2) locations are enabled by default and permit error responses that meet spec (3) maximum depth defaults to 128, which in an absurd amount of levels (System.Text.Json only defaults to 64).

If someone has a need we can always add the method.

PR #4080, opened 2024-10-15
Incorrect ResolvedType property initialization

vincent-duret:
If you use: new NonNullGraphType<< StringGraphType >>, the ResolvedType property is set to null.
It will cause the HasFullSpecifiedResolvedType method to return false.

But if you use new NonNullGraphType(new StringGraphType()), the method returns true.


Shane32:
`ResolvedType` is typically `null` until the schema is initialized, at which time the `ResolvedType` properties are all populated by `SchemaTypes`.

The `NonNullGraphType<T>()` constructor is not typically used by user code; if you need to explicitly specify the `ResolvedType` property, you will want to use the `NonNullGraphType(IGraphType graphType)` constructor and pass in an instance of the underlying type.  The only use case for the `NonNullGraphType<T>()` constructor is a custom implementation of `SchemaTypes` which performs schema initialization.

Issue #4060, opened 2024-09-11
Argument parsing breaking changes in v8


mariusbancila:
I have implemented a directive called `@export` that exports the value of a field to a variable, which can later be used in another part.

Here is an example:
- first read an account with a specific account number and export its creation date to a variable
- then read all the accounts with the same created date
```
query read($cid : Int!, $crdt : Int = 0)
{  
    generalLedgerAccount(filter:{accountNo : {_eq : 1000}})
    {
      items
      {
        accountNo
        name
        createdDate @export(as: "crdt")
      }
    }
    
    second : generalLedgerAccount(filter:{createdDate : {_eq : $crdt}})
    {
      items
      {
        accountNo
        name
        createdDate
      }
    }
}
```
In versions previous to 8, this worked well. The first generalLedgerAccount would be evaluated and the value of `createdDate` exported to `crdt`, then when the `second` node is evaluated, the argument is parsed with the updated value of `$crdt`.

I'm reading it like this:
```
var filterArgumentDef = ctx.FieldDefinition.Arguments?.FirstOrDefault(arg => arg.Name == "filter");
if (filterArgumentDef != null && ctx.Arguments.TryGetValue("filter", out var filterObjectArg) && 
   filterObjectArg.Value is object)
{
   // ...
}
```

However, in version 8, this no longer works. The `filterObjectArg` contains the evaluated argument prior to the updated variables. It's like both of them are parsed at the beginning, before any evaluation begins (so basically `_eq` is 0, and not the exported value). This breaks my implementation making the directive useless.

Shane32:
This is going to be difficult to implement in v8.  Here's how execution works in v8:

1. Parse the document
2. Execute most validation rules (`GetPreNodeVisitorAsync`)
3. Parse all field arguments and directive arguments
4. Execute other validation rules (`GetPostNodeVisitorAsync`)
5. Execute the request

If an error occurs at any stage, execution stops at that point.  This has the main benefit that input coercion rules can be defined on a per-argument (or per-input-field) level, causing a coercion error prior to the execution stage.  The most common example is when you use integer identifiers, but you are using the ID graph type.  In the past, passing "abc" would only fail during execution, but now you can define a coercion rule to cast the string to an integer prior to the execution stage.  (This coercion rule is defined automatically for you in certain cases.)

[...]

PR #4083, opened 2024-10-17
Fix argument parsing for merged selection sets

Shane32:
Fixes queries such as these:

```graphql
query heroQuery($id: Int!) {
    ...fragment1
    ...fragment2
}

fragment fragment1 on Query {
  hero(id: $id) {
    id
  }
}

fragment fragment2 on Query {
  hero(id: $id) {
    name
  }
}
```

Unfortunately this fix is not compatible with the patch described in #4060 - but it makes it no worse either.  See changes to `Issue4060.cs` tests and below notes.

### DIFF
--- a/src/GraphQL.Tests/Bugs/Issue4060.cs
+++ b/src/GraphQL.Tests/Bugs/Issue4060.cs
@@ -1,4 +1,5 @@
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Immutable;
 using GraphQL.Execution;
 using GraphQL.SystemTextJson;
@@ -58,6 +59,60 @@ public class Issue4060
             """);
     }
 
+    [Fact]
+    public async Task SampleCodeWorksWhenMergingSelectionSets()
+    {
+        var heroType = new ObjectGraphType() { Name = "Hero" };
+        heroType.Field<IntGraphType>("id");
+        heroType.Field<StringGraphType>("name");
+        var query = new ObjectGraphType() { Name = "Query" };
+        query.Field("hero", heroType)
+            .Argument<IntGraphType>("id")
+            .Resolve(context => new { id = context.GetArgument<int>("id"), name = "John Doe" });
+        var schema = new Schema { Query = query };
+        schema.Initialize();
+        var options = new ExecutionOptions
+        {
+            Schema = schema,
+            Query = """
+                query heroQuery($id: Int!) {
+                    ...fragment1
+                    ...fragment2
+                }
+                fragment fragment1 on Query {
+                    hero(id: $id) {
+                        id
+                    }
+                }
+                fragment fragment2 on Query {
+                    hero(id: $id) {
+                        name
+                    }
+                }
+                """,
+            Variables = new Inputs(new Dictionary<string, object?>
+                {
+                    { "id", 123 }
+                }),
+            ValidationRules = DocumentValidator.CoreRules.Append(new FieldMappingValidationRule()),
+        };
+        options.Listeners.Add(new DynamicVariableExecutionListener());
+
+        var result = await new DocumentExecuter().ExecuteAsync(options);
+        var resultJson = new GraphQLSerializer().Serialize(result);
+
+        resultJson.ShouldBeCrossPlatJson("""
+            {
+                "data": {
+                    "hero": {
+                        "id": 123,
+                        "name": "John Doe"
+                    }
+                }
+            }
+            """);
+    }
+
     public class FieldMappingValidationRule : ValidationRuleBase
     {
         public override ValueTask<INodeVisitor?> GetPostNodeVisitorAsync(ValidationContext context) => new(new MyVisitor());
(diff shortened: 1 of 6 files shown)


src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -292,8 +293,8 @@ protected virtual ExecutionContext BuildExecutionContext(ExecutionOptions option
 
             Operation = operation,
             Variables = validationResult.Variables ?? Variables.None,
-            ArgumentValues = validationResult.ArgumentValues,
-            DirectiveValues = validationResult.DirectiveValues,
+            ArgumentValues = validationResult.ArgumentValues == null ? null : new ConcurrentDictionary<GraphQLField, IDictionary<string, ArgumentValue>>(validationResult.ArgumentValues),
+            DirectiveValues = validationResult.DirectiveValues == null ? null : new ConcurrentDictionary<ASTNode, IDictionary<string, DirectiveInfo>>(validationResult.DirectiveValues),
```
Shane32, line 297: For merged selection sets, GraphQL.NET creates a faux `GraphQLField` containing the merged selection set as the only change.  While one solution would be to overwrite the `SelectionSet` property of the first `GraphQLField`, this will change the original representation of the query and would mess up document caching at the very least, so this is not possible.  So we are updating the dictionaries containing the arguments and directives to point to the original field arguments.  Unfortunately this occurs at runtime, and possibly within a parallel execution strategy.  So a `ConcurrentDictionary` is used.  We cannot create a `ConcurrentDictionary` at inception because the `ValidationContext` is hard typed to a `Dictionary<TKey, TValue>`.  So we end up copying the dictionary within the document executer into a concurrent dictionary.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -319,20 +319,36 @@ static bool Contains(ROM[] array, ROM item, int count)
...
+                var argumentValues = context.ArgumentValues;
+                if (argumentValues != null && argumentValues.TryGetValue(original.Field, out var arguments))
+                {
+                    ((IDictionary<GraphQLField, IDictionary<string, ArgumentValue>>)argumentValues)[newField] = arguments;
+                }
+                var directiveValues = context.DirectiveValues;
+                if (directiveValues != null && directiveValues.TryGetValue(original.Field, out var directives))
+                {
+                    ((IDictionary<ASTNode, IDictionary<string, DirectiveInfo>>)directiveValues)[newField] = directives;
+                }
```
Shane32, line 345: This code contains the cast.  Essentially, `ExecutionContext.ArgumentValues` and `ExecutionContext.DirectiveValues` must implement `IDictionary<TKey, TValue>` or an `InvalidCastException` will occur if there are selection sets being merged where the parent field has arguments.  In a v9 of GraphQL.NET we should make the above-referenced properties `IDictionary` instead of `IReadOnlyDictionary`.

src/GraphQL.Tests/Bugs/Issue4060.cs
```diff
@@ -97,13 +152,14 @@ public override Task BeforeExecutionAsync(IExecutionContext context)
             return Task.CompletedTask;
         }
 
-        private class FieldArgumentDictionary : IReadOnlyDictionary<GraphQLField, IDictionary<string, ArgumentValue>>
+        private class FieldArgumentDictionary : IReadOnlyDictionary<GraphQLField, IDictionary<string, ArgumentValue>>, IDictionary<GraphQLField, IDictionary<string, ArgumentValue>>
```
Shane32, line 155: In issue #4060 , the developer wanted to be able to redefine variable values at runtime.  This sample code within `Issue4060.cs` was provided to revert GraphQL.NET v8 to perform variable parsing during execution (vs during validation) as was done in v7, solving his issue.  To properly support merged selection sets, this class must now additionally support `IDictionary<TKey, TValue>` as shown here.  However, the changes shown within this `FieldArgumentDictionary` class are not sufficient to properly support merged selection sets while also allowing runtime changes to variables.

Shane32:
@gao-artur This issue is causing a problem for some users.  Let me know when you are back at a PC and can review it.  Thanks!

gao-artur (approved):

src/GraphQL/Execution/DocumentExecuter.cs
```diff
@@ -310,6 +311,12 @@ protected virtual ExecutionContext BuildExecutionContext(ExecutionOptions option
 
         context.ExecutionStrategy = SelectExecutionStrategy(context);
 
+        if (context.ExecutionStrategy is not SerialExecutionStrategy)
+        {
+            context.ArgumentValues = context.ArgumentValues == null ? null : new ConcurrentDictionary<GraphQLField, IDictionary<string, ArgumentValue>>(context.ArgumentValues);
+            context.DirectiveValues = context.DirectiveValues == null ? null : new ConcurrentDictionary<ASTNode, IDictionary<string, DirectiveInfo>>(context.DirectiveValues);
+        }
```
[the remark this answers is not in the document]
Shane32: I agree. I’ll add a comment and open an issue.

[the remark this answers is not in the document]
Shane32: - #4085

PR #4095, opened 2024-11-15
Enhance GraphQL MemoryCache to be compatible with persisted documents

Shane32:
When using Persisted Documents, in the case where the persisted document store is in a database or other 'remote' storage, it is likely that the developer will want to use local caching to speed requests.  This PR allows GraphQL.NET's memory cache to working in conjunction with it's persisted document handler:

1. Modifies the existing MemoryCache to return a persisted document error if Query and DocumentId are both set (same behavior as the existing PersistedDocumentHandler)
2. Modifies the existing MemoryCache to match on the `DocumentId` property of `ExecutionOptions` if present, otherwise the `Query` property.
3. Modifies the persisted document handler to skip processing where the `Document` (parsed AST) is set on `ExecutionOptions`.

Together, this allows configuring as such:

```cs
services.AddGraphQL(b => b
    .UseMemoryCache()
    .UsePeristedDocuments<MyLoader>(GraphQL.DI.ServiceLifetime.Scoped)
);
```

In this fashion, after a persisted document is successfully executed, it will be cached by the memory cache by its document id, and then used in future executions without hitting the persisted document handler.


docs2/site/docs/migrations/migration8.md
```diff
@@ -1052,6 +1052,20 @@ Sample persisted document request:
 }
 ```
 
+#### Caching (v8.3.0+)
+
+The persisted document handler does not provide caching by default. You may implement your own caching mechanism
+within the `GetQueryAsync` method to cache the query strings based on the document identifier. Alternatively, you
+may add the `.UseMemoryCache()` method to enable in-memory caching. Be sure to call `UseMemoryCache` before calling
```
gao-artur, line 1059: It is worth mentioning that the `GraphQL.MemoryCache` package is required.

src/GraphQL.MemoryCache/MemoryDocumentCache.cs
```diff
@@ -59,7 +60,7 @@ protected MemoryDocumentCache(IMemoryCache memoryCache, bool disposeMemoryCache,
     /// </summary>
     protected virtual MemoryCacheEntryOptions GetMemoryCacheEntryOptions(ExecutionOptions options)
     {
-        return new MemoryCacheEntryOptions { SlidingExpiration = _options.SlidingExpiration, Size = options.Query!.Length };
+        return new MemoryCacheEntryOptions { SlidingExpiration = _options.SlidingExpiration, Size = options.Query?.Length ?? options.DocumentId?.Length ?? 100 };
```
gao-artur, line 63: Won't setting 100 as a default be a breaching change?

Shane32: No as Query should be set for any existing code, and if it wasn’t then the code would break. Secondly, either the query or document id should always exist, or there would be nothing to cache against.

However the number is supposed to be representative of the amount of memory required in the cache. The length of the hash isn’t representative. I should write an ast visitor to compute the amount of memory required.

Shane32: Well, unfortunately `GetMemoryCacheEntryOptions` does not pass the `GraphQLDocument` in, so that will have to be a v9 change.  Once it does, we can just grab the `Source.Length` from the document and use that.  So for now, all we have is the document id length to measure.

Shane32: Added overload and marked old one as obsolete.  No breaking change, as if the Query property exists (as it must or the code would have thrown an exception before), it will use the old method.

docs2/site/docs/migrations/migration8.md
Shane32: updated

gao-artur (approved):

Shane32:
Added comments

Issue #1211, opened 2019-07-08
Sample application for using DataLoader


kktam:
## Summary
Provide a set of sample applications to illustrate how dataloader is used.

## Relevant information

Currently the dataloader documentation is few and often does not work. Need to have a set of application to easily illustrate how dataloader is setup in a real scenario, and provide sample on how joins are optimized.

corstian:
I have written several examples on how to use the data loader. [This one is an example on how to implement it combined with connections](https://github.com/corstian/graphql-dataloader-connections-demo). Then there's [this one which describes how to solve the data-loader problem in a more generic way for entity framework](
http://corstianboerman.com/posts/2019-02-12/generic-data-loaders-for-entity-framework-in-graphql.html).

Joins as known from database environment are not being used in GraphQL. The dynamic nature of GraphQL depends on the ability to resolve each and every type on its own. Dataloaders in this way are more of a batching mechanism which can be used to combine all individual requests in one call.

The example will probably help you further.

macco3k:
Is this still relevant? Since I've been exploring data loader concepts myself recently, I would like to give this a try.

Shane32:
Yes, it is.

Shane32:
I've assigned the bounty to you, marking it as `in-progress`, based on our bounty program @macco3k .

MrHPotter:
Hi since this seemed a bit forgotten and i thought it a nice evening activity, i tried my Hand at getting the Bounty: Here my PR with an Example Project Showcasing, Both Dataloaders for Single Entities and Sub-Graph Lists: 
 https://github.com/graphql-dotnet/graphql-dotnet/pull/4096

Shane32:
Please send an invoice for the data loader sample to the open collective account as described by our bounty program here: https://github.com/graphql-dotnet/graphql-dotnet/blob/master/BOUNTY.md

Please reply here when you've done so - thanks!

MrHPotter:
Thank you.
I Submitted an Invoice with the Title of this Issue: Sample application for using DataLoader #1211

PR #4096, opened 2024-11-22
Dataloader Sample Project - Bounty

MrHPotter:
As the Bounty looked a bit forgotten
, and the person last assigned the Bounty quite inactive, i thought i'd try my Hand.
https://github.com/graphql-dotnet/graphql-dotnet/issues/1211
appreciate feedback :)


Shane32:
I have not forgotten about this but been quite swamped. You’ll have to excuse me if it takes me a bit longer before I can look at it. I only very briefly looked at the code, but it seemed that you set up a db which is nice because it demonstrates a typical user setup. I was wondering (since I didn’t test the code) if it will run immediately on a users pc - like through sql localdb or in memory database or what? I was also thinking about whether this sample should be in the examples repo or not. Perhaps you can provide your thoughts in the meantime until I can take a look. Thanks!

MrHPotter:
Hi, no worries. this is an open Source project and i can see you have been the main Maintainer at this point. 
I will not rush you and will see if there is other spots of the project i can make some contributions to.

To your point, i exactly tried to provide that. A typcal setup GraphQl is used for.
This Sample Project is using SqlLite through EFCore. I Set it up in a way, that it should work immediately, when pressing start/ using dotnet run with the Project (It migrates and seeds some Data as part of the startup)
I Also injected the Most Recent Version of GraphQlServer and Altair Ui, for interested Users to Play arround with and Try out the Schema, adapt it etc.

I have no big opinion in which repo this should be. I just put it next to the other example Projects i could find :D which were in the GraphQl Repo. 
If it were up to me i would probably expect Example Projects in a Subfolder in this Main Repo, next to the src Folder i.e. '/Examples' as this would be the easiest to find, for people new to the Library.

Shane32:
> and will see if there is other spots of the project i can make some contributions to.

Sounds great!  Most of the bounties are for documentation as it is rather poor.  If there's any other documentation you want to work on, let me know.  Two notes:
- The migration documents should be very accurate, as I maintain those as each new or changed feature is merged.
- I've nothing against using generative AI to help compose the documentation.  But it should all be relevant and specific, documenting the nuances of GraphQL.NET, tailored for developers and with accurate samples -- we don't need a copy of what AI bots can already help people with.

MrHPotter:
so where are we on this PR and the Bounty?
I could i.e. Move it into the new samples folder, when we merge that PR.
If there are any unliked Patterns, i'm happy to adapt the Project to be up to the highest standards, as i would like it to be a really nice starter reference for anyone trying to set up a GQL Server using Dataloaders :)

MrHPotter:
i guess now it'd be a good Idea to move this Project into the new structure

MrHPotter:
Move it, downgraded to .Net 8 
fixed any WarningsAsError Issues
and did another small test. Still works :) 
Can just be started with the included launchsettings and there is some seedData in the Database to query via /ui/altair

samples/GraphQL.DataLoader.Example/DataLoaderGql.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
Shane32, line 1: This file needs formatting.  Tabs are not uniform.

samples/GraphQL.DataLoader.Example/DataLoaderGql.csproj
```diff
@@ -0,0 +1,26 @@
...
+      <IsTestProject>false</IsTestProject>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="GraphQL" Version="8.2.1" />
+        <PackageReference Include="GraphQL.DataLoader" Version="8.2.1" />
+        <PackageReference Include="GraphQL.MicrosoftDI" Version="8.2.1" />
+        <PackageReference Include="GraphQL.Server.All" Version="8.1.0" />
+        <PackageReference Include="GraphQL.Server.Ui.Altair" Version="8.1.0" />
+        <PackageReference Include="GraphQL.SystemTextJson" Version="8.2.1" />
```
Shane32, line 17: Needs to be ProjectReference not PackageReference

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace DataLoaderGql;
+
+public class DealershipDbContext : DbContext
+{
+
+    public DbSet<Car> Cars => Set<Car>();
+    public DbSet<Salesperson> Salespeople => Set<Salesperson>();
+    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+    {
+        optionsBuilder.UseSqlite(new SqliteConnectionStringBuilder()
```
Shane32, line 14: `.UseSqlite` should be on new line when chained with additional commands like `.UseSeeding`

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                Mode = SqliteOpenMode.ReadWriteCreate,
+                BrowsableConnectionString = true,
+                DataSource = "my.db"
+            }.ToString()).UseSeeding((ctx, _) =>
+        {
+            if(ctx.Set<Salesperson>().Any()) return;
+            ctx.Set<Salesperson>().AddRange(SeedSalespeople);
+            ctx.SaveChanges();
+        })
```
Shane32, line 24: tabbed inconsistently

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
```diff
@@ -0,0 +1,72 @@
...
+        .UseAsyncSeeding(async (ctx, _, ct) =>
+        {
+            if (await ctx.Set<Salesperson>().AnyAsync(cancellationToken: ct).ConfigureAwait(false)) return;
+            await ctx.Set<Salesperson>().AddRangeAsync(SeedSalespeople, cancellationToken: ct).ConfigureAwait(false);
+            await ctx.SaveChangesAsync(ct).ConfigureAwait(false);
+        });
```
Shane32, line 30: use `token` or `cancellationToken` for clarity

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    protected override void OnModelCreating(ModelBuilder modelBuilder)
+    {
+        modelBuilder.Entity<Salesperson>().HasMany(sp => sp.AssignedCars).WithOne()
+            .HasForeignKey(car => car.SalesPersonId);
+    }
+
+    private static readonly Salesperson[] SeedSalespeople =
```
Shane32, line 39: needs reformatting

samples/GraphQL.DataLoader.Example/GraphQl/CarsGraphType.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace DataLoaderGql.GraphQl;
+
+public class CarsGraphType : ObjectGraphType<Car>
+{
+    public CarsGraphType()
+    {
+        Field(x => x.Id, type: typeof(IntGraphType)).Description("Id of the car");
+        Field(x => x.Model, type: typeof(StringGraphType)).Description("Model of the car");
+        Field(x => x.Price, type: typeof(IntGraphType)).Description("Price of the car");
+        Field(x => x.ModelYear, type: typeof(IntGraphType)).Description("Model year of the car");
```
Shane32, line 13: use inferred graph types rather than explicit per recommended pattern

samples/GraphQL.DataLoader.Example/GraphQl/CarsGraphType.cs
```diff
@@ -0,0 +1,15 @@
...
+    }
+}
```
Shane32, line 15: files must end with newline.  note: `dotnet format` should fix a lot of these formatting errors, but tests were broken and didn't run.  see #4101

samples/GraphQL.DataLoader.Example/GraphQl/DealerShipQuery.cs
```diff
@@ -0,0 +1,17 @@
+﻿using GraphQL;
+using GraphQL.Types;
+
+namespace DataLoaderGql.GraphQl;
+
+public class DealerShipQuery : ObjectGraphType
+{
+    public DealerShipQuery()
+    {
+        Field<SalesmanGraphType>("salespeople").Argument<string>("name").AllowAnonymous().Resolve(ctx =>
```
Shane32, line 10: each chained method should be on new line.  `AllowAnonymous` is part of authentication and out of scope for this example.

samples/GraphQL.DataLoader.Example/GraphQl/SalespeopleByNameDataLoader.cs
```diff
@@ -0,0 +1,24 @@
+﻿using DataLoaderGql.Types;
+using GraphQL.DataLoader;
+using Microsoft.EntityFrameworkCore;
+
+namespace DataLoaderGql.GraphQl;
+
+public class SalespeopleByNameDataLoader(DealershipDbContext db) : DataLoaderBase<string, Salesperson>
```
Shane32, line 7: this example promotes the use of data loaders registered as services.  While this technique is documented, and in fact I use it often in my own schemas, we don't promote this as the primary way to make data loaders.  What's your take on this?  Perhaps we should demonstrate two different methods of making data loaders here within this sample?  One using the `IDataLoaderContextAccessor` (typical) and also via DI?  My foremost thought is that incorrect configuration of DI-based data loaders can lead to memory leaks.  It also is slightly more complex to set up, and may require a serial execution strategy, which is not often the typical use case.  Your thoughts?

samples/GraphQL.DataLoader.Example/GraphQl/SalespeopleByNameDataLoader.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    protected override async Task FetchAsync(IEnumerable<DataLoaderPair<string, Salesperson>> list, CancellationToken cancellationToken)
+    {
+        IQueryable<Salesperson> salesmen = db.Salespeople;
+        var names = list.Select(pair => pair.Key);
+
+        var lookup = await salesmen
+            .Where(sm => names.Contains(sm.Name))
+            .GroupBy(sm => sm.Name)
+            .ToDictionaryAsync(group => group.Key, cancellationToken).ConfigureAwait(false);
```
Shane32, line 17: Combining GroupBy and ToDictionary is the equivalent of ToLookup, and this could be written more like `.ToListAsync(...).ToLookup()` which would be a better representation of how the data is used.

We should also consider if the name is unique, in which case a dictionary without the groupby would be a better choice, eliminating Single() below and will provide a more streamlined example.

Shane32: Another option is to change how this is used to be a collection data loader, which just returns the lookup, so a name may return multiple entries.  I just feel like Single() is a bad database design, and would rather not include that in the sample, favoring a better design and/or simpler.

samples/GraphQL.DataLoader.Example/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+// Learn more about configuring OpenAPI at https://aka.ms/aspnet/openapi
+builder.Services.AddDbContext<DealershipDbContext>();
+
+builder.Services.AddGraphQL(builder =>
+{
+    builder
+        .AddSystemTextJson()
+        .AddGraphTypes(typeof(CarsGraphType).Assembly)
+        .AddSchema<DealershipSchema>();
+});
```
Shane32, line 19: This should be formatted like all of our examples, as follows:
```suggestion
builder.Services.AddGraphQL(b => b
    .AddSystemTextJson()
    .AddGraphTypes(typeof(CarsGraphType).Assembly)
    .AddSchema<DealershipSchema>()
);
```

samples/GraphQL.DataLoader.Example/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+var app = builder.Build();
+
+app.UseGraphQL();
+app.UseGraphQLAltair();
+app.MapGraphQLAltair().AllowAnonymous();
```
Shane32, line 24: This line is redundant and should be removed
```suggestion
```

samples/GraphQL.DataLoader.Example/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+await app
+    .Services
+    .CreateAsyncScope()
+    .ServiceProvider
+    .GetRequiredService<DealershipDbContext>()
+    .Database
+    .MigrateAsync()
+    .ConfigureAwait(false);
```
Shane32, line 32: This code is missing a `using` around the scope that is created.  I suggest a slight modification to make it easier to understand:

```suggestion
await using (var scope = app.Services.CreateAsyncScope())
{
    var dbContext = scope.ServiceProvider.GetRequiredService<DealershipDbContext>();
    await dbContext.Database.MigrateAsync().ConfigureAwait(false);
}
```

samples/GraphQL.DataLoader.Example/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+
+await app
+    .Services
+    .CreateAsyncScope()
+    .ServiceProvider
+    .GetRequiredService<DealershipDbContext>()
+    .Database
+    //For making sure seed Data gets applied
+    .EnsureCreatedAsync()
+    .ConfigureAwait(false);
```
Shane32, line 42: I'm not sure why both `Migrate` and `EnsureCreated` are in use, but chances are that either (a) one can be eliminated, or (b) we can just add this to the above:

```cs
await dbContext.Database.EnsureCreatedAsync().ConfigureAwait(false);
```

If both are required, it would seem that EnsureCreated would have to come first ????  But I could be wrong.

samples/GraphQL.DataLoader.Example/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+app.Run();
```
Shane32, line 43: considering that the rest of the file has async methods, we should probably make this async also

```suggestion
await app.RunAsync();
```

Shane32 (changes requested):

Shane32:
If you merge with master, it will run the build checks properly (I hope).

MrHPotter:
oh i see there is still conflicts left.

samples/GraphQL.DataLoader.Example/Program.cs
MrHPotter: i was not even aware that this exists and have a hard time imagining, what it could be good for, as setting up the Host, at least in my Head, would most likely be quite sequential. But that's likely a question for the dotnet team :D i can sure add this change :) Will commit it locally tho, as the Analyzer wants a ConfigureAwait

samples/GraphQL.DataLoader.Example/Program.cs
MrHPotter: i call ensureCreated only for the Seeding to run. This is something they seemingly changed in EFCore 9, where EnsureCreated actually didn't Create the DB Schema for me anymore and told me to migrate. I can try if Migrate also calls the Seeding Methods, but i think it does sadly not.

MrHPotter: i guess it was both other things i had messed up in the middle of building this. We have free choice of which one we want to use. So i guess we go with EnsureCreated, because it's the simpler version and we can leave out the Migrations.

samples/GraphQL.DataLoader.Example/DataLoaderGql.csproj
MrHPotter: this actually seems problematic, as the Server Package already imports these Dependencies as Nugets and is not available here for Project Reference.

Shane32: Right, that is why I suggested we might make a very-simple server project replacement ("polyfill") so that the sample is representative.  Or we have to move the sample out of the repo.  But it defeats the purpose to have the sample here if changes to the codebase are not reflected in the execution of the sample project.

samples/GraphQL.DataLoader.Example/Program.cs
MrHPotter: I removed the Migrations and left it with EnsureCreated with your suggested Style from above

samples/GraphQL.DataLoader.Example/DataLoaderGql.csproj
MrHPotter: Should be good now

samples/GraphQL.DataLoader.Example/Program.cs
Shane32: Sounds great!  Makes the sample simpler, and keeps it focused on demonstrating the data loader, not demonstrating EF.

MrHPotter:
Applied all suggested Changes except ProjectRef Change, due to the reason given in the Comment @Shane32 .

samples/GraphQL.DataLoader.Example/DataLoaderGql.csproj
MrHPotter: i will look to do this today. so what were you thinking? Just Copy the Code from server Project and make clear it's a pollyfill by naming and Project Config? Or would it be needed to be trimmed down a bunch?

Shane32: Look in the Harness sample and see the middleware file. It’s very simple. Use that along with an extension to link up UseGraphQL to the middleware so that the same code in the sample projects can be used with the server project.

MrHPotter: Have a look. i stole the minimum part from Harness Project and added an extra Comment in Program.cs.
Still works as expected locally

Shane32: Here, I've extracted it to a separate project: https://github.com/graphql-dotnet/graphql-dotnet/pull/4102

MrHPotter: that's a really nice Idea and can help making future example Projects simpler to get up and running.

samples/GraphQL.DataLoader.Example/GraphQl/SalespeopleByNameDataLoader.cs
MrHPotter: I think one of each is a good idea.
I dont think these Risks of DI dataloader jump out of the docs tho. Should probably added in scope of the Dataloader Docs Bounty Issue

samples/GraphQL.DataLoader.Example/GraphQl/SalespeopleByNameDataLoader.cs
MrHPotter: I couldnt figure out how to make a Collection DataLoader like this work. But maybe i can transform this one into the other Pattern?

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
```diff
@@ -0,0 +1,83 @@
...
+    public DbSet<Car> Cars => Set<Car>();
+    public DbSet<Salesperson> Salespeople => Set<Salesperson>();
+
+    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+    {
+        optionsBuilder
+            .UseSqlite(
+                new SqliteConnectionStringBuilder()
+                {
+                    Mode = SqliteOpenMode.ReadWriteCreate,
```
gao-artur, line 18: Wouldn't it be better to use a `Memory` mode for easier sample running?

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
```diff
@@ -0,0 +1,83 @@
...
+                    BrowsableConnectionString = true,
+                    DataSource = "my.db"
+                }
+                    .ToString()
```
gao-artur, line 22: Looks like needs formatting

samples/GraphQL.DataLoader.Example/GraphQl/CarsBySalespersonDataLoader.cs
```diff
@@ -0,0 +1,25 @@
...
+using GraphQL.DataLoader;
+using Microsoft.EntityFrameworkCore;
+
+namespace DataLoaderGql.GraphQl;
+
+public class CarsBySalespersonDataLoader(DealershipDbContext db) : DataLoaderBase<int, List<Car>>
+{
+    protected override async Task FetchAsync(IEnumerable<DataLoaderPair<int, List<Car>>> list, CancellationToken cancellationToken)
+    {
+        var keys = list.Select(pair => pair.Key).ToHashSet();
```
gao-artur, line 11: This `ToHashSet` would be helpful when used in the `Where` extension based on `IEnumerable`. But when it's based on `IQueariable`, you can leave it as non-enumerated `list.Select(pair => pair.Key)` without the risk of multiple enumerations. It would be enumerated only once to create the SQL `where in (...)` clause.

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
gao-artur: Yeah, I completely forgot this "little" detail...

samples/GraphQL.DataLoader.Example/appsettings.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "AllowedHosts": "*",
+  "GraphQlPath": "/graphql",
+  "EnableMetrics": true,
```
gao-artur, line 10: Metrics can have a very bad impact on performance. Would it be a good idea to enable them in a sample project?

samples/GraphQL.DataLoader.Example/DataLoaderGql.csproj
MrHPotter: ok, shouldn't i first switch to using #4102 first?

samples/GraphQL.DataLoader.Example/GraphQl/CarsGraphType.cs
MrHPotter: do i need something special up? dotnet format on the csproj doesnt change anything when i run it.

samples/GraphQL.DataLoader.Example/appsettings.json
MrHPotter: yeah i guess not. With the new extracted Polyfill @Shane32  set up, this wasn't used anymore anyway.

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
MrHPotter: also i personally prefer sqlite for a sample project.
As the Db gets automatically created, i don't see any harm to ease of use. And the User of the sample can even browse the little db, to maybe confirm how he thinks it all works.

samples/GraphQL.DataLoader.Example/GraphQl/CarsGraphType.cs
MrHPotter: i think it does end with a newline now tho

samples/GraphQL.DataLoader.Example/DealershipDbContext.cs
MrHPotter: is it good now?

samples/GraphQL.DataLoader.Example/GraphQl/CarsBySalespersonDataLoader.cs
```diff
@@ -0,0 +1,26 @@
...
+                => keys
+                .Contains(car.SalesPersonId))
+                .ToListAsync(cancellationToken: cancellationToken)
+                .ConfigureAwait(false))
+                .ToLookup(car => car.SalesPersonId);
+
+
+        foreach (var pair in list)
+        {
+            pair.SetResult(carsLookup[pair.Key].ToList());
```
Shane32, line 23:
```suggestion
            pair.SetResult(carsLookup[pair.Key]);
```

samples/GraphQL.DataLoader.Example/GraphQl/SalesmanGraphType.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace DataLoaderGql.GraphQl;
+
+public class SalesmanGraphType : ObjectGraphType<Salesperson>
+{
+    public SalesmanGraphType()
+    {
+        Field(x => x.Id, type: typeof(IntGraphType)).Description("Id of the salesman");
+        Field(x => x.Name, type: typeof(StringGraphType)).Description("Name of the salesman");
+        Field<ListGraphType<CarsGraphType>, List<Car>>("assignedCars").Description("Assigned cars").ResolveAsync(ctx =>
```
Shane32, line 13:
```suggestion
        Field<IEnumerable<Car>>("assignedCars", false).Description("Assigned cars").ResolveAsync(ctx =>
```

samples/GraphQL.DataLoader.Example/GraphQl/SalespeopleByNameDataLoader.cs
Shane32: The CarsBySalespersonDataLoader is an accurate representation of a collection data loader

samples/GraphQL.DataLoader.Example/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+        .AddSystemTextJson()
+        .AddGraphTypes(typeof(CarsGraphType).Assembly)
+        .AddSchema<DealershipSchema>()
+);
+
+var app = builder.Build();
+
+//This GraphQl Middleware is a polyfill to Operate without Dependencies on the Seperate Server Repository
+// For Setting up a real GraphQl WebApi, be sure to have a look at: https://github.com/graphql-dotnet/server
+app.UseGraphQL<DealershipSchema>();
```
Shane32, line 21:
```suggestion
app.UseGraphQL();
```

samples/GraphQL.DataLoader.Example/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddScoped<SalespeopleByNameDataLoader>().AddScoped<CarsBySalespersonDataLoader>();
+builder.Services.AddDbContext<DealershipDbContext>();
+
+
+builder.Services.AddGraphQL(b => b
+        .AddSystemTextJson()
+        .AddGraphTypes(typeof(CarsGraphType).Assembly)
+        .AddSchema<DealershipSchema>()
```
Shane32, line 14:
```suggestion
    .AddSystemTextJson()
    .AddGraphTypes(typeof(CarsGraphType).Assembly)
    .AddSchema<DealershipSchema>()
```

samples/GraphQL.DataLoader.Example/README.md
```diff
@@ -0,0 +1,6 @@
+﻿# GraphQl.Net DataLoader Refeference Implementation
+
+## Source:
+https://github.com/graphql-dotnet/graphql-dotnet/issues/1211
+
```
Shane32, line 5:
```suggestion
```
Not sure that the issue is relevant after it's merged... ??

samples/GraphQL.DataLoader.Example/README.md
```diff
@@ -0,0 +1,6 @@
...
+A small sample Project. Where we have a very rudamentary Car Dealership, managed in SQLite and we use the Dataloader to Load the Salespeople by Name, in Case we want multiple, aswell as their SubGraph of Assigned Cars, exposing this as a GraphQl Server with the GraphQL.Server.All Package and Altair Ui.
```
Shane32, line 6:
```suggestion
A sample project showcasing a basic car dealership system. The dealership is managed using SQLite, with data loading optimized by employing data loaders to fetch salespeople by name. This also supports retrieving their associated subgraph of assigned cars. The project is exposed as a GraphQL server using the `GraphQL.Server.All` package, and includes the Altair UI for interactive querying.
```

Shane32 (commented):
Few comments.  Last thing: need the data loader documentation updated to reference this sample project.  Use ChatGPT / etc if desired to ensure quality text.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/docs2/site/docs/guides/dataloader.md

samples/GraphQL.DataLoader.Example/GraphQl/SalespeopleByNameDataLoader.cs
MrHPotter: ah yes, i remember. I think this is why i looked to make this one a non-Collection One.
 I can make the Name a unique Index and FirstOrDefault as more usual Db Pattern. 
 Maybe we do a second little Project, that shares the overall Structure of this but uses the Other way of using Dataloaders, and Reference Both in the Docs.
 What do you think about that?
 I think we wouldnt want to recommend to mix them, as i think in the Docs it sais 'If u use this, you can leave the other stuff out' about the DI version

samples/GraphQL.DataLoader.Example/README.md
MrHPotter: I'd think it can give Context, to where the need is from and can give insights to interested Patterns and other Metadata to interested Parties.
If u prefer, i can remove it, but i also think it can't Hurt. Especially if we ensure this PR is linked to the Issue.

samples/GraphQL.DataLoader.Example/GraphQl/SalespeopleByNameDataLoader.cs
Shane32: Good idea

MrHPotter: Pushed it :) Also Referenced the DataLoader Docs. 
Those will ofcourse need to be a bit reworked and should then probably reference Code from the Sample Projects as far as possible, but i'd really prefer to do that in a seperate PR, after this one is closed. This one is already quite big and long running.

MrHPotter:
> Few comments. Last thing: need the data loader documentation updated to reference this sample project. Use ChatGPT / etc if desired to ensure quality text.
> 
> https://github.com/graphql-dotnet/graphql-dotnet/blob/master/docs2/site/docs/guides/dataloader.md

Could you take a look and see if what i did looks alright to you?

samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs
```diff
@@ -0,0 +1,90 @@
...
+                }
+                .ToString()
+            )
+            .UseSeeding((ctx, _) =>
+            {
+                if (ctx.Set<Salesperson>().Any())
+                    return;
+                ctx.Set<Salesperson>().AddRange(_seedSalespeople);
+                ctx.SaveChanges();
+            })
```
Shane32, line 30: duplicate
```suggestion
```

samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                if (await ctx
+                        .Set<Salesperson>()
+                        .AnyAsync(cancellationToken)
+                        .ConfigureAwait(false))
+                    return;
+                await ctx
+                    .Set<Salesperson>()
+                    .AddRangeAsync(_seedSalespeople, cancellationToken)
+                    .ConfigureAwait(false);
```
Shane32, line 41:
```suggestion
                ctx.Set<Salesperson>()
                    .AddRange(_seedSalespeople);
```
Calls to Add generally should not be async.  See MS docs.

> This method is async only to allow special value generators, such as the one used by 'Microsoft.EntityFrameworkCore.Metadata.SqlServerValueGenerationStrategy.SequenceHiLo', to access the database asynchronously. For all other cases the non async method should be used.

samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs
```diff
@@ -0,0 +1,90 @@
...
+                    return;
+                ctx.Set<Salesperson>().AddRange(_seedSalespeople);
+                ctx.SaveChanges();
+            })
+            .UseAsyncSeeding(async (ctx, _, cancellationToken) =>
+            {
+                if (await ctx
+                        .Set<Salesperson>()
+                        .AnyAsync(cancellationToken)
+                        .ConfigureAwait(false))
```
Shane32, line 36: As a general rule, all calls to ConfigureAwait within application code should be removed.  Not sure if one of the formatting rules needs to be changed for these sample projects.  `ConfigureAwait(false)` should be used for all library calls.

For reference, ConfigureAwait is used to return calls back to the original synchronization context.  However, there is none within ASP.NET Core, so `ConfigureAwait(false)` is a no-op.  This isn't exactly the reason for the above guidelines, but it does mean that it is unnecessary in any ASP.NET Core application, which these samples are.  So it will be cleaner if we remove the unnecessary code from the sample projects.

Here's one of many docs on the use of ConfigureAwait:
- https://devblogs.microsoft.com/dotnet/configureawait-faq/

samples/GraphQL.DataLoader.Sample.DI/README.md
```diff
@@ -0,0 +1,7 @@
+﻿# GraphQl.Net DataLoader Refeference Implementation
+
+## Source:
+https://github.com/graphql-dotnet/graphql-dotnet/issues/1211
+
+A small sample Project. Where we have a very rudamentary Car Dealership, managed in SQLite and we use the Dataloader to Load the Salespeople by Name, in Case we want multiple, aswell as their SubGraph of Assigned Cars, exposing this as a GraphQl Server with the GraphQL.Server.All Package and Altair Ui.
+In this one we showcase the usage of DI/Class based DataLoaders.
```
Shane32, line 7:
```suggestion
A simple sample project featuring a basic car dealership managed with SQLite. It demonstrates the use of DataLoader to load salespeople by name, including their assigned cars in a subgraph, and exposes this data as a GraphQL server using the `GraphQL.Server.All` package with the Altair UI. The project highlights dependency injection (DI) and class-based DataLoader usage.
```

docs2/site/docs/guides/dataloader.md
```diff
@@ -30,12 +33,9 @@ In the example above, a using a DataLoader will allow us to batch together all o
 
 ## Setup
 
-1. Register `IDataLoaderContextAccessor` in your IoC container.
-2. Register `DataLoaderDocumentListener` in your IoC container.
-
+1. Register the Dataloader Services
```
Shane32, line 36: Renumber the rest of the numbering in the sequence since 2 was deleted

docs2/site/docs/guides/dataloader.md
```diff
@@ -30,12 +33,9 @@ In the example above, a using a DataLoader will allow us to batch together all o
...
 ``` csharp
-services.AddSingleton<IDataLoaderContextAccessor, DataLoaderContextAccessor>();
-services.AddSingleton<DataLoaderDocumentListener>();
+services.AddDataLoader();
```
Shane32, line 38:
```suggestion
services.AddGraphQL(b => b
    .AddDataLoader()
    // other configurations
);
```

docs2/site/docs/guides/dataloader.md
```diff
@@ -3,6 +3,9 @@
 GraphQL.NET includes an implementation of Facebook's [DataLoader](https://github.com/facebook/dataloader) within the
 [`GraphQL.DataLoader`](https://www.nuget.org/packages/GraphQL.DataLoader) NuGet package.
 
+Sample projects for the Default Dataloaders described at the Top of this Page and the [DI based DataLoaders](#di-based-data-loaders) can be found in the [GraphQL.Net source repository](https://github.com/graphql-dotnet/graphql-dotnet).
+
```
Shane32, line 7:
```suggestion
Sample projects demonstrating both default data loaders (as described below on this page) and [DI-based data loaders](#di-based-data-loaders) are available in the [GraphQL.NET source repository](https://github.com/graphql-dotnet/graphql-dotnet/samples). These projects cover examples of data loaders accessed via `IDataLoaderContextAccessor` as well as those implemented using dependency injection (DI).
```

samples/GraphQL.DataLoader.Sample.Default/GraphQl/SalespersonGraphType.cs
```diff
@@ -0,0 +1,25 @@
+using GraphQL.DataLoader.Di.Sample.Types;
+using GraphQL.Types;
+using Microsoft.EntityFrameworkCore;
+
+namespace GraphQL.DataLoader.Di.Sample.GraphQl;
```
Shane32, line 5: namespace needs to match project

samples/GraphQL.DataLoader.Sample.Default/GraphQl/SalespersonGraphType.cs
```diff
@@ -0,0 +1,25 @@
...
+
+public sealed class SalespersonGraphType : ObjectGraphType<Salesperson>
+{
+    public SalespersonGraphType(IDataLoaderContextAccessor accessor, DealershipDbContext dbContext)
+    {
+        Field(x => x.Id, type: typeof(IntGraphType)).Description("Id of the salesman");
+        Field(x => x.Name, type: typeof(StringGraphType)).Description("Name of the salesman");
```
Shane32, line 12: remove type

samples/GraphQL.DataLoader.Sample.Default/GraphQl/SalespersonGraphType.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        Field<ListGraphType<CarsGraphType>, IEnumerable<Car>>("assignedCars").Description("Assigned cars")
```
Shane32, line 14:
```suggestion
        Field<IEnumerable<Car>>("assignedCars", false).Description("Assigned cars")
```
i think this is correct...

samples/GraphQL.DataLoader.Sample.Default/GraphQl/SalespersonGraphType.cs
Shane32: easiest way to fix it here is to right-click the "Di" and rename it, and it will change all the files

samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs
MrHPotter: Actually in their Docs they recommend implementing both UseSeeding and UseSeedingAsync together.

>  When using this feature, it is recommended to implement both UseSeeding and UseAsyncSeeding methods using similar logic, even if the code using EF is asynchronous.

samples/GraphQL.DataLoader.Sample.Default/README.md
```diff
@@ -0,0 +1,8 @@
+﻿# GraphQl.Net DataLoader Refeference Implementation
+
+## Source:
+https://github.com/graphql-dotnet/graphql-dotnet/issues/1211
+
+
+A small sample Project. Where we have a very rudamentary Car Dealership, managed in SQLite and we use the Dataloader to Load the Salespeople by Name, in Case we want multiple, aswell as their SubGraph of Assigned Cars, exposing this as a GraphQl Server with the GraphQL.Server.All Package and Altair Ui.
+In this one we showcase the Default Dataloader Usage, defined in the Field Resolvers.
```
Shane32, line 8:
```suggestion
A simple sample project featuring a basic car dealership managed with SQLite. It demonstrates the use of DataLoader accessed through `IDataLoaderContextAccessor` to load salespeople by name, including their assigned cars in a subgraph, and exposes this data as a GraphQL server using the `GraphQL.Server.All` package with the Altair UI. The project highlights the practical use of `IDataLoaderContextAccessor` for managing DataLoader instances.```

**Comment by Shane32** [comment]
Looks good.  Let's finish these last few updates and merge it in.

File: samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs (hunk 29 above)
Comment (MrHPotter) [reply samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs]: ah ok thx. i actually didnt realize there is no Syncronization Context in Asp.NetCore. In my Last Job adding this everywhere (in App Level Code of Asp.NetCore WebApis) was actually enforced with warnings. So at this point i thought it good practice.
I'll remove it from the sample Projects then.

File: samples/GraphQL.DataLoader.Sample.Default/GraphQl/SalespersonGraphType.cs (hunk 35 above)
Comment (Shane32) [reply samples/GraphQL.DataLoader.Sample.Default/GraphQl/SalespersonGraphType.cs:12]:
```suggestion
        Field(x => x.Id).Description("Id of the salesman");
        Field(x => x.Name).Description("Name of the salesman");
```

samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs
MrHPotter: ye im getting Build Errors about removing the Configure await now @Shane32 could you help me out where to exclude the 2 new Sample Projects from this Rule?

MrHPotter:
im getting Compiler Errors now from missing ConfigureAwait(false) :D => This is also why the Test Step is failing right now, i checked.

 other then that and the UseSeeding+UseAsyncSeeding thing where the Docs suggest we do both, i think we should be good - at least from my side

Shane32:
> im getting Compiler Errors now from missing ConfigureAwait(false) :D => This is also why the Test Step is failing right now, i checked.
> 
> other then that and the UseSeeding+UseAsyncSeeding thing where the Docs suggest we do both, i think we should be good - at least from my side

i'll fix it

samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs
Shane32: :+1:

samples/GraphQL.DataLoader.Sample.Default/DealershipDbContext.cs
Shane32: i added it to directory.build.targets -- see commit.

MrHPotter:
thx for adding the Compiler exclusion. from my side feel free to merge whenever :)

docs2/site/docs/guides/dataloader.md
```diff
@@ -30,15 +32,16 @@ In the example above, a using a DataLoader will allow us to batch together all o
...
-services.AddSingleton<DataLoaderDocumentListener>();
+
+services.AddGraphQL(b => b
+    .AddDataLoader()
+    // other configurations
+);
 ```
 
-3. Hook up your GraphQL schema to your IoC container.
+2Hook up your GraphQL schema to your IoC container.
```
Shane32, line 44:
```suggestion
2. Hook up your GraphQL schema to your IoC container.
```

docs2/site/docs/guides/dataloader.md
```diff
@@ -30,15 +32,16 @@ In the example above, a using a DataLoader will allow us to batch together all o
...
 ## Setup
 
-1. Register `IDataLoaderContextAccessor` in your IoC container.
-2. Register `DataLoaderDocumentListener` in your IoC container.
-
+1. Register the Dataloader Services
 ``` csharp
-services.AddSingleton<IDataLoaderContextAccessor, DataLoaderContextAccessor>();
-services.AddSingleton<DataLoaderDocumentListener>();
+
```
Shane32, line 37:
```suggestion
```

docs2/site/docs/guides/dataloader.md
```diff
@@ -30,15 +32,15 @@ In the example above, a using a DataLoader will allow us to batch together all o
 
 ## Setup
 
-1. Register `IDataLoaderContextAccessor` in your IoC container.
-2. Register `DataLoaderDocumentListener` in your IoC container.
-
+1. Register the Dataloader Services
```
Shane32, line 35:
```suggestion
1. Register the DataLoader services

```

samples/GraphQL.DataLoader.Sample.DI/GraphQl/CarsBySalespersonDataLoader.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        var keys = list.Select(pair => pair.Key);
+        var carsLookup =
+            (await db.Cars.Where(car
+                => keys
+                .Contains(car.SalesPersonId))
+                .ToListAsync(cancellationToken: cancellationToken)
+                )
+                .ToLookup(car => car.SalesPersonId);
+
```
Shane32, line 18:
```suggestion
        var carsLookup = (await db.Cars
            .Where(car => keys.Contains(car.SalesPersonId))
            .ToListAsync(cancellationToken: cancellationToken))
            .ToLookup(car => car.SalesPersonId);
```

Shane32:
I made a few changes, as follows:
- DbContext can't be injected to a singleton service in the 'default' sample
- CLR types were not mapped to graph types
- Query graph type should typically be injected from the schema definition
- The schema must pass IServiceProvider to the base constructor
- Changed salespeople (plural) to salesPerson (singular) for the non-collection loader sample

Shane32 (approved):

Issue #2406, opened 2021-03-17
Add sample to documentation for custom validation rules


Shane32:
**This issue participates in bounty program, see https://github.com/graphql-dotnet/graphql-dotnet/blob/master/BOUNTY.md**

**bounty amount: $100**

Tasks to be done:
1. Add new sample project demonstrating custom validation rule.
2. Add documentation page regarding samples provided by this repo. Currently there are 2 samples for AOT, 2 StarWars projects and GraphQL.Harness. They all may be considered as examples. 
3. Update sitemap.yaml and include page into solution tree. 
4. ~Also update migration document to show transition from EnterLeaveListener~ tracked by #2463

joelmandell:
Do you have a quick example that you could provide here in this issue?

joelmandell:
Okay I think I figured it out:

Migrated from:
```
 public Task<INodeVisitor> ValidateAsync(ValidationContext context)
        {
            var userContext = (context.UserContext["ident"]) as ClaimsIdentity;

            return new EnterLeaveListener(a =>
            {
                if (!userContext.IsAuthenticated)
                {
                    a.Match((Operation op) =>
                    {
                        if (((Field)op.SelectionSet.Selections[0]).Name != "login")
                        {
                            context.ReportError(new ValidationError(context.OriginalQuery + "___", "auth-required", "login"));
                        }
                    });
                }
            }).ToTask();
        }
```
To this in 4.x:


```
  public class RequiresAuthValidationRule : IValidationRule
    {

        private static readonly Task<INodeVisitor> _nodeVisitor = new MatchingNodeVisitor<Operation>((op, context) =>
        {
            var userContext = (context.UserContext["ident"]) as ClaimsIdentity;
            if (!userContext.IsAuthenticated)
            {
               
                if (((Field)op.SelectionSet.Selections[0]).Name != "login")
                {
                    context.ReportError(new ValidationError(context.Document.OriginalQuery + "___", "auth-required", "login"));
                }
            }
        }).ToTask();

        public Task<INodeVisitor> ValidateAsync(ValidationContext context) => _nodeVisitor;
    }
```

Shane32:
Looks good.  Also look at current sample:

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Validation/Rules/PossibleFragmentSpreads.cs

versus previous sample here:

[...]

PR #4099, opened 2024-11-26
Rearrange the sample projects into a separate root folder

MrHPotter:
My idea as i Described here: https://github.com/graphql-dotnet/graphql-dotnet/issues/2406#issuecomment-2501757637

I added a new Table To the Readme Showing where all the Example Projects are distributed and which one are Part of the Project or not.
In this PR i pulled out all but the Star wars example Projects out of the source Folder and Arranged them into a new Samples Solution within the Examples Folder.
In the Samples solution i also added a Solution Folder, referencing all Projects from src/ for ease of building the Example Projects.

Im still getting a weird Build Error in the Harness.Tests Project about there not being a Definition for TOExecutionResult in GraphQLAssertion.cs and SuccessResultAssertion.cs . 
Maybe someone knows why?


Shane32:
> Im still getting a weird Build Error in the Harness.Tests Project about there not being a Definition for TOExecutionResult in GraphQLAssertion.cs and SuccessResultAssertion.cs .
> Maybe someone knows why?

If you have the latest Visual Studio installed, you might want to add a global.json file in the root pointing to .NET 8 until I have time to update the repo to be compatible with the .NET 9 SDK.  Or maybe that's not the problem, but I know a bunch of repos are not compatible anymore because of outdated dependency issues.

MrHPotter:
im not sure. i don't even have dotnet 9 installed at this point. if u like the idea and we want to go with it, i'll compare with master, to see what's going on. Also ofcourse, it looks like this is breaking github actions. so i'd have to reroute the test commands for the sample projects there.

Shane32:
> if u like the idea and we want to go with it

Well, right now src is all the .NET code, and documentation and other stuff isn't in src.  I'm generally okay with this, but I think if we do this, we should move all the files in the src folder (sln, *.props, global.json, etc) into the repo root.  And I suggest using `samples` lowercase rather than `Samples` to conform with the rest of the repo's root folders.

Shane32:
@gao-artur Do you have any comments?

MrHPotter:
Oh thx i will look into slngen.
 Didnt actually mean to commit the ProjectAdder. I just tried to make the adding of the projects a  little less tedious and C# i am best at. especially as i was bootet into windows :D

So as official notes/playback:
Move props files etc into root and reference those from slns
Remove ProjectAdder
Fix SharedFolderPath in GraphQL.Harness.Tests

Thanks both of you for the great feedback

gao-artur:
Note that if you are working on Windows and renaming the `Samples` folder to lowercase `samples`, you may later find later both cases in GitHub with a mix of files in both folders. This is because Windows paths are case-insensitive. The simplest way to fix it, is to remove the folder, commit, restore the folder, fix the case, commit.

README.md
```diff
@@ -42,6 +42,14 @@ Provides the following packages:
...
+|----------------------------|----------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------|
+| GraphQL.Net/Samples        | Sample Projects focussed on the Features of the Core GraphQL Library (this) an Implementation of the GraphQL Spec                      | https://github.com/graphql-dotnet/graphql-dotnet                        |
+| GraphQL.Net.Server/Samples | Sample Projects focussed on the Features of the Server Package (Utilities to hooking up a GraphQL Server  in tune with Dotnet Web APIs | https://github.com/graphql-dotnet/server/tree/master/samples            |
+| Examples/Src               | Community Provided Examples. Not necessary Patterns Supported by this project. Provide a way to see how other people use the Lib       | https://github.com/graphql-dotnet/examples/tree/master/src              |
+| GraphQl.Net.Client/Examples| Example Implementations for the GraphQL Client implementation.								              | https://github.com/graphql-dotnet/graphql-client/tree/master/examples   |
```
Shane32, line 51:
```suggestion
## Example Projects and Sibling Repositories

| Project/Route               | Description                                                                                                            |
|-----------------------------|------------------------------------------------------------------------------------------------------------------------|
| [GraphQL.Net/Samples](https://github.com/graphql-dotnet/graphql-dotnet)                              | Sample projects focused on showcasing features of the core GraphQL library, an implementation of the GraphQL specification.           |
| [GraphQL.Net.Server/Samples](https://github.com/graphql-dotnet/server/tree/master/samples)           | Sample projects highlighting features of the server package, including utilities for integrating a GraphQL server with .NET Web APIs. |
| [Examples/Src](https://github.com/graphql-dotnet/examples/tree/master/src)                           | Community-provided examples. These may not represent officially supported patterns but show how others use the library.               |
| [GraphQL.Net.Client/Examples](https://github.com/graphql-dotnet/graphql-client/tree/master/examples) | Example implementations for the GraphQL client library.                                                                               |
```

samples/Samples.sln
```diff
@@ -0,0 +1,189 @@
+
+Microsoft Visual Studio Solution File, Format Version 12.00
```
Shane32, line 2: Please remove Samples.sln; all projects should be referenced by the sln in the root.

src/GraphQL.sln
```diff
@@ -6,13 +6,13 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".Solution Items", ".Solutio
 	ProjectSection(SolutionItems) = preProject
```
Shane32, line 6: Please move the sln to the root and revert removal of all sample projects

Shane32 (changes requested):
Generally looks good.  I would like to maintain a single sln for the repo, so let's put a sln in the root with all projects referenced.  I also suggested a few changes to the readme.

MrHPotter:
@Shane32 Moved everything into the GraphQL Solution and that into the Repo. Might have to adjust some working-directories in the Piplines with this in mind tho.
when moving from src toroot, do i just remove the working-directory part in the yaml or is it '.' or '/' ?  Im new to github pipes, coming from gitlab/devops and teamcity :D

MrHPotter:
so i checked in their docs and it seems '.' is fine, so i tried adjusting them on my own. It's in a seperate commit to revert, if it serves up any issues

Shane32:
Looks pretty close.  need to fix this error in the tests:
```
System.InvalidOperationException : Unable to find repository root - 'D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.sln' missing; caller file path 'D:\a\graphql-dotnet\graphql-dotnet\src\GraphQL.ApiTests\BomTests.cs'.
```

MrHPotter:
i'll have a look

MrHPotter:
@Shane32 it looks to me like i got it fixed :)

Shane32 (approved):

Issue #4106, opened 2024-12-17
Add schema validation for transitively referenced interfaces


Shane32:
We should write a schema validation rule to validate that the schema has been configured properly.  Perhaps also write an optional schema node visitor to apply the transitively implemented interfaces to all implementing types.  Perhaps only the latter.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/graphql-dotnet/issues/4105#issuecomment-2548601842_

Issue #4105, opened 2024-12-17
Use interface implementing interface lost opportunity to query by fragment


KvaKoZyaBBrr:
Hello! I dont know is this a bug or not so it is a question
## Summary
In version 7.2.2 when interfaces cannot implementing another interfaces, we have next schema:
```GraphQl
# Type
type Car implements ICar & _ICar & ICarExt
{
    id: Int
    name: String
    extField: String
}

# Base Interface. Not included ext fields
interface ICar
{
    id: Int
    name: String
}

# Ext interface
interface ICarExt
{
    extField: String
}

# Final interface for implementing
interface _ICar
{
    id: Int
    name: String
    extField: String
}
```
In version 8.2.1 we can do something more pretty:
```GraphQl
# Type
type Car implements _ICar
{
    id: Int
    name: String
    extField: String
}

# Base Interface. Not included ext fields
interface ICar
{
    id: Int
    name: String
}

# Ext interface
interface ICarExt
{
    extField: String
}

# Final interface for implementing
interface _ICar implements ICar & ICarExt
{
    id: Int
    name: String
    extField: String
}
```
But we get error on query
```GraphQl
query{
    cars{
        ... on ICar{
            name
        }
    }
}
```
Error text: "Fragment cannot be spread here as objects of type "_ICar" can never be of type "ICar"".
But I think we can get results as any implementations any deep. Is this bug by new feature or expect behaviour?

Shane32:
The schema configuration is illegal when configured in this manner.  It is required for the implementing types to implement all transitively implemented interfaces.  See spec: https://spec.graphql.org/October2021/#sel-EAHbhBDHAACLA-uB

> Transitively implemented interfaces (interfaces implemented by the interface that is being implemented) must also be defined on an implementing type or interface. For example, `Image` cannot implement `Resource` without also implementing `Node`:
>
> ```gql
> interface Node {
>   id: ID!
> }
> 
> interface Resource implements Node {
>   id: ID!
>   url: String
> }
> 
> interface Image implements Resource & Node {
>   id: ID!
>   url: String
>   thumbnail: String
> }
> ```

We should write a schema validation rule to validate that the schema has been configured properly.  Perhaps also write an optional schema node visitor to apply the transitively implemented interfaces to all implementing types.

Shane32:
For now you will need to explicitly implement `ICar`, `_ICar` and `ICarExt` on `Car`, similar to v7.

[...]

PR #4108, opened 2024-12-18
Auto-add transitively implemented interfaces to object and interface types

Shane32:
There is no schema validation rule to validate that transitively implemented interfaces are directly implemented by all implementing types.  There are two approaches to fix the problem, both of which are available to us in version 8:

### Auto-adding transitively implemented interfaces (this PR)
  - This would be considered a new feature, widening the supported requests, not narrowing
  - Easier to use / more C#-like interface implementation pattern
  - Resolvers can only execute against a concrete object type, so no problems there
  - Would change the introspection query (for affected schemas)

### Add schema validation to prevent an invalid schema from being initialized
  - Within minor version bumps, we allow changes that better enforce the graphql specification, even if doing so "breaks" an existing schema
  - Requires the developer to directly reference all transitively-referenced interfaces
  - This would break any implementation that uses an invalidly-configured schema in this manner

This PR represents the former option, being that it is likely what was desired by the developer in the first place, and prevents a minor version bump from breaking code.

See:
- https://github.com/graphql-dotnet/graphql-dotnet/issues/4105#issuecomment-2548601842
- #4106

### DIFF
--- a/src/GraphQL/Utilities/Visitors/TransitiveInterfaceVisitor.cs
+++ b/src/GraphQL/Utilities/Visitors/TransitiveInterfaceVisitor.cs
@@ -54,8 +54,9 @@ public sealed class TransitiveInterfaceVisitor : BaseSchemaNodeVisitor
             {
                 foreach (var transitiveInterface in iface.ResolvedInterfaces.List)
                 {
+                    // Ignore circular references (will be caught in SchemaValidationVisitor)
                     if (transitiveInterface == baseType)
-                        throw new InvalidOperationException($"'{baseType.Name}' cannot implement interface '{iface.Name}' because it creates a circular reference.");
+                        continue;
 
                     transitiveInterfaces.Add(transitiveInterface);
                 }
--- a/src/GraphQL/Utilities/Visitors/TransitiveInterfaceVisitor.cs
+++ b/src/GraphQL/Utilities/Visitors/TransitiveInterfaceVisitor.cs
@@ -1,0 +1,66 @@
+using GraphQL.Types;
+
+namespace GraphQL.Utilities;
+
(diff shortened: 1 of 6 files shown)


Shane32:
@gao-artur I'd appreciate your comments on the two approaches described in the PR description above; lmk if you agree that option 1 is a better choice or think option 2 would be better.

src/GraphQL/Utilities/Visitors/TransitiveInterfaceVisitor.cs
```diff
@@ -0,0 +1,67 @@
...
+    private void FindTransitiveInterfaces(IImplementInterfaces baseType, IImplementInterfaces type, HashSet<IInterfaceGraphType> transitiveInterfaces, HashSet<IInterfaceGraphType> checkedInterfaces)
+    {
+        foreach (var iface in type.ResolvedInterfaces.List)
+        {
+            if (checkedInterfaces.Add(iface))
+            {
+                foreach (var transitiveInterface in iface.ResolvedInterfaces.List)
+                {
+                    // Ignore circular references (will be caught in SchemaValidationVisitor)
+                    if (transitiveInterface == baseType)
```
gao-artur, line 58: Will it also work for non-singleton registrations?

gao-artur:
Option 1 looks more native for C# devs. Option 2 requires knowing the spec details. My vote for 1.

src/GraphQL/Utilities/Visitors/TransitiveInterfaceVisitor.cs
Shane32: Yes

gao-artur (approved):

PR #4116, opened 2025-01-15
Add additional annotations for proper AOT analysis

Shane32:
Generic types passed down to another function must inherit the `DynamicallyAccessedMembers` attribute or the native AOT analysis doesn't work properly.  These warnings did not show up with the .NET 6 analysis rules but do with the .NET 8 analysis rules, and even then, do not always show up.  Confirmed by testing.

Note: includes #4117 so that tests will run.  Once #4117 is merged, I'll merge from master to reduce the diff.


gao-artur:
These attributes are such a mess. You hardly see the code behind them...

gao-artur (approved):

Shane32:
> These attributes are such a mess. You hardly see the code behind them...

I agree.  Let me know if you have any ideas, while preserving AOT compatibility.

PR #4121, opened 2025-02-03
Fix issue with UseMemoryCache returning documents from the wrong schema

cotillion:
When multiple schemas are used in combination with UseMemoryCache documents which have passed validation in one schema will skip validation when used against another schema.

These queries then result in InvalidOperationExceptions from the ExecutionStrategy.

This PR adds Schema to the CacheItem which appears to fix it.

### DIFF
--- a/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCache.cs
@@ -1,5 +1,6 @@
 using GraphQL.DI;
 using GraphQL.PersistedDocuments;
+using GraphQL.Types;
 using GraphQL.Validation;
 using GraphQLParser.AST;
 using Microsoft.Extensions.Caching.Memory;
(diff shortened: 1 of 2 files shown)


Shane32:
Good catch!  However, this change will make the cache ineffective when used with scoped schemas.  Any thoughts or ideas?

cotillion:
> Good catch! However, this change will make the cache ineffective when used with scoped schemas. Any thoughts or ideas?

We don't use scoped schemas so I don't know much about them but as any difference in either types or ExampleExperimentalFeatures on the schema should result in a cache miss i don't see a great solution. 

One approach is extending MemoryCacheEntryOptions with a delegate which can return the cache isolation value, defaulting to executionOptions.Schema perhaps. Those doing dynamic schemas could put something in metadata to identify them.

Might be a good idea to allow setting a unique id field on a schema in the future.

Shane32:
> One approach is extending MemoryCacheEntryOptions with a delegate which can return the cache isolation value, defaulting to executionOptions.Schema perhaps. Those doing dynamic schemas could put something in metadata to identify them.

Let's do this, except have the default be `schema.GetType()`, at least for code-first schemas.  I'm going to fork from your PR and try it.

Shane32:
@cotillion Please review PR #4127 and lmk your comments - thanks!

cotillion:
Closing in favor of #4127

PR #4124, opened 2025-02-12
Add NoClrMapping option for input fields

Shane32:
Allows the developer to exclude specific fields from the graph from being mapped to the CLR type automatically.  The developer must override ParseDictionary to apply the input to the CLR object.  Note that IsValidDefault and ToAST methods also skip these fields, and if necessary, the developer should override those methods as well.

See discussion:
- https://github.com/graphql-dotnet/graphql-dotnet/discussions/4123

Added doc below:

---

### 30. NoClrMapping Extension for Input Fields with Custom CLR Mapping (from version 8.4.0)

A new extension method, `NoClrMapping`, has been introduced to give you fine-grained control over how input object fields are mapped to CLR types. In many scenarios, your input types may include fields that are computed or require custom processing before being set on your CLR model. For example, you might have a CLR type with a single `FullName` property but want your input object to accept separate `firstName` and `lastName` fields. By marking these fields with `NoClrMapping()`, they are excluded from the automatic binding process. You can then override the `ParseDictionary` method to manually combine these values into the CLR's `FullName` property, while letting the base implementation handle any automatically mapped fields (such as `Age`). The following example illustrates this pattern:

```csharp
public class Person
{
    public string FullName { get; set; } = string.Empty;
    public int Age { get; set; }
}

public class PersonInputType : InputObjectGraphType<Person>
{
    public PersonInputType()
    {
        Name = "PersonInput";

        // Define fields for 'firstName' and 'lastName' that are not automatically mapped.
        Field<NonNullGraphType<StringGraphType>>("firstName").NoClrMapping();
        Field<NonNullGraphType<StringGraphType>>("lastName").NoClrMapping();

        // Define the 'age' field which is automatically mapped to Person.Age.
        Field(x => x.Age);
    }

    public override object ParseDictionary(IDictionary<string, object?> value)
    {
        // Use base.ParseDictionary to handle automatic mapping (e.g., Age).
        var person = (Person)base.ParseDictionary(value);

        // Manually combine firstName and lastName into FullName.
        if (value.TryGetValue("firstName", out var firstNameObj) && firstNameObj is string firstName &&
            value.TryGetValue("lastName", out var lastNameObj) && lastNameObj is string lastName)
        {
            person.FullName = $"{firstName} {lastName}";
        }

        return person;
    }
}
```

In this example, the input object defines three fields: `firstName`, `lastName`, and `age`. The `firstName` and `lastName` fields are marked with `NoClrMapping()` so that they are not automatically bound to any CLR properties. Instead, the overridden `ParseDictionary` method combines these two fields into the `FullName` property of the `Person` CLR type. Meanwhile, the `Age` field is processed via the standard mapping mechanism by calling `base.ParseDictionary`. This approach allows you to seamlessly mix automatic and custom binding, providing greater flexibility in how your input objects are processed and mapped.

### DIFF
--- a/docs2/site/docs/migrations/migration8.md
+++ b/docs2/site/docs/migrations/migration8.md
@@ -1240,6 +1240,64 @@ public class MyOutputObject
 }
 ```
 
+### 30. NoClrMapping Extension for Input Fields with Custom CLR Mapping (from version 8.4.0)
+
+A new extension method, `NoClrMapping`, has been introduced to give you fine-grained control
+over how input object fields are mapped to CLR types. In many scenarios, your input types may
+include fields that are computed or require custom processing before being set on your CLR model.
+For example, you might have a CLR type with a single `FullName` property but want your input
+object to accept separate `firstName` and `lastName` fields. By marking these fields with
+`NoClrMapping()`, they are excluded from the automatic binding process. You can then override
+the `ParseDictionary` method to manually combine these values into the CLR's `FullName` property,
+while letting the base implementation handle any automatically mapped fields (such as `Age`). The
+following example illustrates this pattern:
+
+```csharp
+public class Person
+{
+    public string FullName { get; set; } = string.Empty;
+    public int Age { get; set; }
+}
+
+public class PersonInputType : InputObjectGraphType<Person>
+{
+    public PersonInputType()
+    {
+        Name = "PersonInput";
+
+        // Define fields for 'firstName' and 'lastName' that are not automatically mapped.
+        Field<NonNullGraphType<StringGraphType>>("firstName").NoClrMapping();
+        Field<NonNullGraphType<StringGraphType>>("lastName").NoClrMapping();
+
+        // Define the 'age' field which is automatically mapped to Person.Age.
+        Field(x => x.Age);
+    }
+
+    public override object ParseDictionary(IDictionary<string, object?> value)
+    {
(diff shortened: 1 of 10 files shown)


Shane32:
Note: federation tests are failing as Apollo Router has discontinued support of federation 1 supergraphs.  Need to fix separately.

gao-artur:
Nice! I have another use case for this feature. Sometimes, we want to remove the CLR property but keep the graphql field for backward compatibility. Before v8, we added the `.DeprecationReason("Ignored")` to the field to allow clients to update their codebase before eventually removing the field. But in v8 it's not possible anymore because of the dynamic compilation. I guess we need to add exclusion to [GQL006](https://graphql-dotnet.github.io/docs/analyzers/gql006/) when `NoClrMapping` is applied but `ParseDictionary` is not overridden.

gao-artur (approved):

gao-artur:
@Shane32 when do you plan to release this?

Shane32:
Well, I have no real plans for 8.4.0 besides what we already merged, so I'm good releasing it anytime.  I was thinking maybe 1mo after 8.3.x (like in 1 week), but delaying only makes a difference if we add more features.  Shall I release it now?  Have anything you want to add to 8.4.0?

gao-artur:
Yes, please release it. I really need the `NoClrMapping`.

Shane32:
> Yes, please release it. I really need the `NoClrMapping`.

Done!

PR #4127, opened 2025-02-12
Make memory cache unique per schema

Shane32:
This fix/change prevents the document cache from matching on a query previously cached for one schema when the same query is requested for another schema.

See:
- #4121

Added docs:

---

### 31. Unique Cache Keys for Multi-Schema Environments (from version 8.4.0)

Document caching now supports unique cache keys for applications using multiple or dynamic schemas. The cache key has been enhanced to include an extra property computed by the new `AdditionalCacheKeySelector` delegate. By default, this delegate distinguishes between schema-first and code-first implementations by returning the schema instance for schema-first and dynamic schemas (those that are direct implementations of `Schema`), and the schema type for code-first and type-first schemas (schemas that derive from `Schema`). This change ensures that cached documents are uniquely identified even when multiple schemas are configured in the same application, while still supporting caching for scoped code-first and type-first schemas.

To migrate, you may continue using the default behavior. However, if you need further customization -- such as incorporating additional context to identify unique dynamic schemas -- you can override the default selector as shown in the example below:

```csharp
services.AddGraphQL(b => b
    .UseMemoryCache(options =>
    {
        options.AdditionalCacheKeySelector = execOptions =>
        {
            if (execOptions.UserContext is IDictionary<string, object> context &&
                context.TryGetValue("CustomHeader", out var header))
            {
                return header;
            }
            return null;
        };
    })
);
```

These changes help ensure that cached documents are correctly associated with the appropriate schema when multiple schemas are in use.

### DIFF
--- a/src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
@@ -24,8 +24,9 @@ public class MemoryDocumentCacheOptions : MemoryCacheOptions, IOptions<MemoryDoc
 
     /// <summary>
     /// A delegate that supplies an extra value to be included in the cache key.
-    /// By default, this returns the <see cref="ExecutionOptions.Schema"/>. You can override
-    /// this behavior to incorporate additional uniqueness.
+    /// When not configured, the extra value is computed from <see cref="ExecutionOptions.Schema"/> as follows:
+    /// The schema instance is used for schema-first and dynamic schemas (when the type <see cref="Types.Schema"/>),
+    /// and the schema type is used for type-first and code-first schemas (when the type is derived from <see cref="Types.Schema"/>).
     /// </summary>
     public Func<ExecutionOptions, object?>? AdditionalCacheKeySelector { get; set; }
 
--- a/src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
+++ b/src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
@@ -22,5 +22,12 @@ public class MemoryDocumentCacheOptions : MemoryCacheOptions, IOptions<MemoryDoc
     /// </summary>
     public TimeSpan? SlidingExpiration { get; set; }
 
+    /// <summary>
+    /// A delegate that supplies an extra value to be included in the cache key.
+    /// By default, this returns the <see cref="ExecutionOptions.Schema"/>. You can override
+    /// this behavior to incorporate additional uniqueness.
+    /// </summary>
+    public Func<ExecutionOptions, object?>? AdditionalCacheKeySelector { get; set; }
+
     MemoryDocumentCacheOptions IOptions<MemoryDocumentCacheOptions>.Value => this;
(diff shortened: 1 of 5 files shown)


src/GraphQL.MemoryCache/MemoryDocumentCacheOptions.cs
```diff
@@ -22,5 +22,13 @@ public MemoryDocumentCacheOptions()
     /// </summary>
     public TimeSpan? SlidingExpiration { get; set; }
 
+    /// <summary>
+    /// A delegate that supplies an extra value to be included in the cache key.
+    /// When not configured, the extra value is computed from <see cref="ExecutionOptions.Schema"/> as follows:
+    /// The schema instance is used for schema-first and dynamic schemas (when the type <see cref="Types.Schema"/>),
```
Shane32, line 28:
```suggestion
    /// The schema instance is used for schema-first and dynamic schemas (when the type is <see cref="Types.Schema"/>),
```

src/GraphQL.Tests/Caching/MemoryCacheTests.cs
```diff
@@ -194,4 +197,71 @@ public async Task ExecuteAsync(bool querySet, bool documentIdSet, bool docSet, b
         memoryDocumentCacheMock.Protected().Verify("GetAsync", getCalled ? Times.Once() : Times.Never(), ItExpr.IsAny<ExecutionOptions>());
         memoryDocumentCacheMock.Protected().Verify("SetAsync", setCalled ? Times.Once() : Times.Never(), ItExpr.IsAny<ExecutionOptions>(), ItExpr.IsAny<GraphQLDocument>());
     }
+
+    [Theory]
+    [InlineData("schema1", "query1", "schema1", "query1", true)]
+    [InlineData("schema1", "query1", "schema2", "query1", false)]
+    [InlineData("schema2", "query2", "schema2", "query2", true)]
+    [InlineData("schema2", "query2", "schema1", "query2", false)]
+    [InlineData("schema1", "query1", "schema2", "query2", false)]
```
Shane32, line 206: Ensures cache is unique across different `Schema` instances, where `schema.GetType() == typeof(Schema)`.

src/GraphQL.Tests/Caching/MemoryCacheTests.cs
```diff
@@ -194,4 +197,71 @@ public async Task ExecuteAsync(bool querySet, bool documentIdSet, bool docSet, b
...
+    [InlineData("starwars1", "query1", "starwars1", "query1", true)]
+    [InlineData("starwars1", "query1", "starwars1", "query2", false)]
+    [InlineData("starwars1", "query1", "starwars2", "query1", true)]
+    [InlineData("starwars1", "query1", "starwars2", "query2", false)]
```
Shane32, line 210: Ensures cache is shared across different instances of derived `Schema` instances (where `schema.GetType()` is the same) -- in this case, two different instances of `StarWarsSchema`.

gao-artur (approved):

cotillion:
This looks good!

Issue #4129, opened 2025-02-24
Compiled ToObject fails to create KeyValuePair


gao-artur:
```c#
    [Theory]
    [InlineData(false)]
    [InlineData(true)]
    public void toobject_key_value_pair(bool compile)
    {
        var inputs = """{ "key": "k", "value": "v" }""".ToInputs();
        var value = inputs.ToObject<KeyValuePair<string, string>>(compile);
        value.Key.ShouldBe("k");
        value.Value.ShouldBe("v");
    }
```

> System.InvalidCastException: 'Unable to cast object of type 'GraphQL.Types.NonNullGraphType`1[GraphQL.Types.AutoRegisteringInputObjectGraphType`1[System.Collections.Generic.KeyValuePair`2[System.String,System.String]]]' to type 'GraphQL.Types.IInputObjectGraphType'.'

gao-artur:
Interesting... This test passes on .NET 6 but fails on other targets

gao-artur:
All structs have implicit parameterless constructor. Should the mapping logic ignore it when other constructors are available?

Shane32:
> All structs have implicit parameterless constructor. Should the mapping logic ignore it when other constructors are available?

I can't remember our logic, although I think we considered this question already.  Offhand, I'd say not necessarily, as it should be attempting to find the 'best fit', which may be the parameterless constructor combined with setting some properties.

gao-artur:
> To fix the test, change GraphQL.Tests.ObjectExtensions.ToObject.cs:20 as follows to use .GetNamedType(), stripping away the NonNullGraphType<T> wrapper

Yep. You are right.

gao-artur:
> I can't remember our logic, although I think we considered this question already. Offhand, I'd say not necessarily, as it should be attempting to find the 'best fit', which may be the parameterless constructor combined with setting some properties.

Then, this [description](https://graphql-dotnet.github.io/docs/migrations/migration8/#6-code-classlanguage-textobjectextensionstoobjectcode-changes-impacts-code-classlanguage-textinputobjectgraphtypecode) is not accurate:
> - Constructor is selected based on the following rules:
>   - If only a single public constructor is available, it is used.
>   - If a public constructor is marked with [GraphQLConstructor], it is used.
>   - Otherwise the public parameterless constructor is used if available.
>   - Otherwise an exception is thrown during deserialization.


[...]

PR #4131, opened 2025-02-25
Allow implicit struct constructors

Shane32:
Currently `AutoRegisteringInputObjectGraphType` cannot deserialize to a struct that has no explicit constructor.  This patch fixes the logic so if no explicitly defined constructors are found, the deserialization logic will use the implicit constructor.  This would be natural for code such as this, which was previously unsupported due to no explicitly defined constructors:

```cs
public struct MyStruct
{
    public int Id { get; set; }
}
```

This change does not affect any existing schemas, and as such is considered a bugfix.

See:
- #4129


Shane32:
I did not change the migration doc, as I feel this sentence "Otherwise the public parameterless constructor is used if available" implies the use of implicit default constructors, whether for classes or structs.

gao-artur (approved):

PR #4143, opened 2025-04-02
Add IGraphQLBuilder.VerifyServices extension method

Shane32:
The PR modifies `[FromServices]` and the resolver builder code to mark `FieldType` instances with metadata indicating which DI services are necessary to execute the resolver.  Then by adding `.ValidateServices()` to the startup, all those services can be validated that they can be serviced by the DI provider, reducing runtime errors due to missing services.

`.ValidateServices()` uses reflection to obtain a handle to `IServiceProviderIsService` as this type does not exist in the current dependency on Microsoft.Extensions.DependencyInjection.Abstractions 2.0.0.  Perhaps in v9 we should bump the dependency.

`.ValidateServices()` is opt-in and throws an exception when used with older versions of MSDI which do not contain this functionality.  This mimics the functionality of MSDI's BuildServiceProvider, where it is opt-in to verify the service chain.  It is also important to help reduce initialization time for scoped schemas.

### DIFF
--- a/src/GraphQL.MicrosoftDI/MicrosoftDIGraphQLBuilderExtensions.cs
+++ b/src/GraphQL.MicrosoftDI/MicrosoftDIGraphQLBuilderExtensions.cs
@@ -88,4 +88,14 @@ public static class MicrosoftDIGraphQLBuilderExtensions
             GraphQLParser.AST.OperationType.Subscription);
         return builder;
     }
+
+    /// <summary>
+    /// Verifies that all injected services can be created during GraphQL field execution.
+    /// Requires support for IServiceProviderIsService by the dependency injection framework.
+    /// </summary>
+    public static IGraphQLBuilder ValidateServices(this IGraphQLBuilder builder)
+    {
+        builder.Services.TryRegister<IConfigureSchema, ValidateServicesSchemaConfigurator>(ServiceLifetime.Singleton);
+        return builder;
+    }
 }
--- a/src/GraphQL.MicrosoftDI/ValidateServicesSchemaValidator.cs
+++ b/src/GraphQL.MicrosoftDI/ValidateServicesSchemaValidator.cs
@@ -1,0 +1,92 @@
+using System.Reflection;
+using GraphQL.Types;
+using GraphQL.Utilities;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GraphQL.MicrosoftDI;
+
+/// <summary>
+/// A schema validator that ensures all services required by <see cref="FromServicesAttribute"/> are registered.
+/// </summary>
+internal sealed class ValidateServicesSchemaValidator : BaseSchemaNodeVisitor
+{
+    private readonly Func<Type, bool> _isValidService;
+    private List<string>? _errors;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ValidateServicesSchemaValidator"/> class.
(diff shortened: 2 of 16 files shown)


src/GraphQL.MicrosoftDI/ValidateServicesSchemaValidator.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            foreach (var serviceType in serviceTypes)
+            {
+                if (!_isValidService(serviceType))
+                {
+                    _errors ??= new();
+                    _errors.Add($"The service '{serviceType.FullName}' required by '{type.Name}.{field.Name}' is not registered.");
+                }
+            }
+        }
```
gao-artur, line 38: I think it's also worth clearing the `REQUIRED_SERVICES_METADATA` to allow the GC. Maybe in `PostVisitSchema`

src/GraphQL.MicrosoftDI/MicrosoftDIGraphQLBuilderExtensions.cs
```diff
@@ -88,4 +88,14 @@ public static IGraphQLBuilder AddScopedSubscriptionExecutionStrategy(this IGraph
             GraphQLParser.AST.OperationType.Subscription);
         return builder;
     }
+
+    /// <summary>
+    /// Verifies that all injected services can be created during GraphQL field execution.
+    /// Requires support for IServiceProviderIsService by the dependency injection framework.
+    /// </summary>
+    public static IGraphQLBuilder ValidateServices(this IGraphQLBuilder builder)
```
gao-artur, line 96: Maybe add a boolean argument to enable/disable the validation. So the user would be able to enable it in dev or test environment but disable in prod to not hurt the startup time.
```c#
services.AddGraphQL(builder => builder
        .AddSchema<VerifyServicesTestSchema>()
        .ValidateServices(IHostEnvironment.IsDevelopment()));
```

Shane32: Good idea.  Done.

src/GraphQL.MicrosoftDI/ValidateServicesSchemaValidator.cs
Shane32: Not so sure about this.  We don't do this kind of cleanup anywhere else in the codebase, rather leaving the metadata static for the lifetime of the schema.  For example, every field that is defined by an expression, such as `Field(x => x.Name)`, stores a reference to the `PropertyInfo` used by by the expression.  It is not used by the schema after the field resolver is compiled.  In addition, the memory consumption of this metadata should be extremely minimal compared to the memory footprint of the schema, and speed isn't a concern as metadata isn't used during execution.  Do you really think this is worthwhile?

gao-artur: No, not really. This would be a micro optimization and if we do that here, we need to do that everywhere.

Shane32: We could optimize the memory requirements for this metadata by storing the Type directly for a single entry, and only creating a List when there is a second entry.  Another optimization could be to always store an array, at the cost of recreating the array for each additional member.

gao-artur (approved):

PR #4145, opened 2025-04-05
Add Type/Types to InterfaceGraphType; add [PossibleType]

Shane32:
Adds these members to `InterfaceGraphType`:
- `IEnumerable<Type> Types`
- `Type(Type type)`
- `Type<TType>()`

These members hae exactly the same signature as those that already exist on `UnionGraphType` and serve the same purpose - to provide a uninstantiated reference to a possible graph type for an abstract graph type.  These members should be added to `IAbstractGraphType`, but that would be a breaking change, so I'll do so for v9.

In addition, the `[PossibleType(Type type)]` attribute was added to mirror the `[Implements(Type type)]` attribute, such that an interface can point to a possible type instead of requiring the object type to point to the interface.  As with `[Implements]`, the provided type can be a CLR or graph type.


src/GraphQL/Types/Composite/InterfaceGraphType.cs
```diff
@@ -7,8 +7,16 @@ public interface IInterfaceGraphType : IAbstractGraphType, IComplexGraphType, II
...
+internal interface IInterfaceGraphType2 : IInterfaceGraphType
+{
+    public IEnumerable<Type> Types { get; set; }
+    public void Type(Type type);
+    public void Type<TType>() where TType : IObjectGraphType;
+}
+
 /// <inheritdoc cref="InterfaceGraphType"/>
-public class InterfaceGraphType<[NotAGraphType] TSource> : ComplexGraphType<TSource>, IInterfaceGraphType
+public class InterfaceGraphType<[NotAGraphType] TSource> : ComplexGraphType<TSource>, IInterfaceGraphType2
```
Shane32, line 19: Non-breaking change

docs2/site/docs/migrations/migration8.md
```diff
@@ -1362,6 +1362,43 @@ Field<StringGraphType>("myField")
...
+
+GraphQL.NET allows developers to configure object graph types to reference interfaces that they
+implement via the `Interface<T>()` or `AddResolvedInterface(IInterfaceGraphType type)`.
+Alternatively, you may reference the object graph type from the interface by using
+`AddPossibleType(IObjectGraphType type)`. However, this requires a concrete reference to the
+instantiated class rather than just the `Type` to be resolved during schema initialization.
+
+In version 8.5.0+, you may use `Type<TObjectType>()` or `Type(Type objectType)` to reference
+the object type from the interface type. This matches the same methods already available on
+`UnionGraphType`, unifying the API between unions and interfaces. The new methods will be
```
gao-artur, line 1375: Do you think it's a good idea to unify these two types? For `union`, it makes sense to declare possible types on the union type itself because union is a closed set of options. But interface is an abstraction and I don't think abstraction should know its derived types.

Shane32: My thoughts:
1. GraphQL.NET already allows this, but only for instantiated types. This PR just extends it to abstract types, if you know what I mean. 
2. When the object types are not directly referenced in the schema, you currently have to directly add a reference in the startup or else they don’t get registered.  #4145 is an issue where my sample would be better if this PR would be merged, due to this issue. 
3. GraphQL’s execution logic, by definition in the spec, requires resolving the returned object type to a concrete type. This is handled by the union or interface definition.
4. Unions and interfaces are by their nature quite similar, except interfaces can define fields and unions cannot. I understand there are some usage and syntax differences but the implementation is quite similar. You could argue that a union is nothing more than an interface with no fields.

Shane32: Also, there’s no way to implement this feature by a third party; the functionality needs to be present within SchemaTypes.

Shane32: I guess I’m thinking that GraphQL.NET has always been very flexible and this is one area where the underlying library is preventing the desired behavior from being added.

gao-artur (approved):

PR #4148, opened 2025-04-09
Support federation interface resolution

Shane32:
This allows for resolving interfaces within GraphQL federation.  Unfortunately it requires breaking changes.  While possible, it does not seem practical to implement the required functionality as non-breaking changes.

The biggest breaking change is to `IFederationResolver` which now uses `IComplexGraphType` instead of `IObjectGraphType`.  This will break custom implementations of the resolver as seen in Sample1, either if the interface is directly implemented, or if they derive from `FederationResolverBase`.

Extension methods to configure the schema have not changed, so it is possible or even likely that most federation schemas will not break with these changes.

### DIFF
--- a/samples/GraphQL.Federation.SchemaFirst.Sample1/MyFederatedResolver.cs
+++ b/samples/GraphQL.Federation.SchemaFirst.Sample1/MyFederatedResolver.cs
@@ -20,10 +20,10 @@ public class MyFederatedResolver<T> : IFederationResolver
 
     public bool MatchKeys(IDictionary<string, object?> representation) => true;
 
-    public object ParseRepresentation(IObjectGraphType graphType, IDictionary<string, object?> representation)
+    public object ParseRepresentation(IComplexGraphType graphType, IDictionary<string, object?> representation)
         => (int)Convert.ChangeType(representation["id"], typeof(int), CultureInfo.InvariantCulture)!;
 
-    public async ValueTask<object?> ResolveAsync(IResolveFieldContext context, IObjectGraphType graphType, object parsedRepresentation)
+    public async ValueTask<object?> ResolveAsync(IResolveFieldContext context, IComplexGraphType graphType, object parsedRepresentation)
     {
         int id = (int)parsedRepresentation;
         var data = context.RequestServices!.GetRequiredService<Data>();
--- a/src/GraphQL.ApiTests/net60/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net60/GraphQL.approved.txt
@@ -1742,7 +1742,7 @@ namespace GraphQL.Federation
         public System.Threading.Tasks.ValueTask<GraphQL.Validation.IVariableVisitor?> GetVariableVisitorAsync(GraphQL.Validation.ValidationContext context) { }
         public System.Threading.Tasks.ValueTask LeaveAsync(GraphQLParser.AST.ASTNode node, GraphQL.Validation.ValidationContext context) { }
     }
-    [System.AttributeUsage(System.AttributeTargets.Class, AllowMultiple=true)]
+    [System.AttributeUsage(System.AttributeTargets.Class | System.AttributeTargets.Interface, AllowMultiple=true)]
     public class KeyAttribute : GraphQL.GraphQLAttribute
     {
         public KeyAttribute(string fields) { }
@@ -2173,9 +2173,9 @@ namespace GraphQL.Resolvers
     }
     public class MemberResolver : GraphQL.Resolvers.IFieldResolver
     {
-        public MemberResolver(System.Reflection.FieldInfo fieldInfo, System.Linq.Expressions.LambdaExpression instanceExpression) { }
-        public MemberResolver(System.Reflection.PropertyInfo propertyInfo, System.Linq.Expressions.LambdaExpression instanceExpression) { }
-        public MemberResolver(System.Reflection.MethodInfo methodInfo, System.Linq.Expressions.LambdaExpression instanceExpression, System.Collections.Generic.IList<System.Linq.Expressions.LambdaExpression> methodArgumentExpressions) { }
+        public MemberResolver(System.Reflection.FieldInfo fieldInfo, System.Linq.Expressions.LambdaExpression? instanceExpression) { }
+        public MemberResolver(System.Reflection.PropertyInfo propertyInfo, System.Linq.Expressions.LambdaExpression? instanceExpression) { }
+        public MemberResolver(System.Reflection.MethodInfo methodInfo, System.Linq.Expressions.LambdaExpression? instanceExpression, System.Collections.Generic.IList<System.Linq.Expressions.LambdaExpression> methodArgumentExpressions) { }
         protected virtual System.Func<GraphQL.IResolveFieldContext, System.Threading.Tasks.ValueTask<object?>> BuildFieldResolver(System.Linq.Expressions.ParameterExpression resolveFieldContextParameter, System.Linq.Expressions.Expression bodyExpression) { }
         public virtual System.Threading.Tasks.ValueTask<object?> ResolveAsync(GraphQL.IResolveFieldContext context) { }
     }
--- a/src/GraphQL.ApiTests/net60/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net60/GraphQL.approved.txt
@@ -1673,10 +1673,23 @@ namespace GraphQL.Federation
     }
     public static class FederationGraphTypeExtensions
     {
+        public static void ResolveReference<TSourceType>(this GraphQL.Types.IInterfaceGraphType<TSourceType> graphType, GraphQL.Federation.Resolvers.IFederationResolver resolver) { }
+        public static void ResolveReference<TSourceType>(this GraphQL.Types.IInterfaceGraphType<TSourceType> graphType, System.Func<GraphQL.IResolveFieldContext, TSourceType, GraphQL.DataLoader.IDataLoaderResult<TSourceType?>> resolver)
+            where TSourceType : new() { }
+        public static void ResolveReference<TSourceType>(this GraphQL.Types.IInterfaceGraphType<TSourceType> graphType, System.Func<GraphQL.IResolveFieldContext, TSourceType, System.Threading.Tasks.Task<TSourceType?>> resolver)
+            where TSourceType : new() { }
+        public static void ResolveReference<TSourceType>(this GraphQL.Types.IInterfaceGraphType<TSourceType> graphType, System.Func<GraphQL.IResolveFieldContext, TSourceType, TSourceType?> resolver)
+            where TSourceType : new() { }
         public static void ResolveReference<TSourceType>(this GraphQL.Types.ObjectGraphType<TSourceType> graphType, GraphQL.Federation.Resolvers.IFederationResolver resolver) { }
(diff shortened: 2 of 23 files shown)


samples/GraphQL.Federation.SchemaFirst.Sample1/MyFederatedResolver.cs
```diff
@@ -20,10 +20,10 @@ public MyFederatedResolver(Func<Data, int, Task<T?>> resolver)
 
     public bool MatchKeys(IDictionary<string, object?> representation) => true;
 
-    public object ParseRepresentation(IObjectGraphType graphType, IDictionary<string, object?> representation)
+    public object ParseRepresentation(IComplexGraphType graphType, IDictionary<string, object?> representation)
         => (int)Convert.ChangeType(representation["id"], typeof(int), CultureInfo.InvariantCulture)!;
 
-    public async ValueTask<object?> ResolveAsync(IResolveFieldContext context, IObjectGraphType graphType, object parsedRepresentation)
+    public async ValueTask<object?> ResolveAsync(IResolveFieldContext context, IComplexGraphType graphType, object parsedRepresentation)
```
Shane32, line 26: If any changes were needed to an application, it would be approximately this complex.

src/GraphQL.ApiTests/net60/GraphQL.approved.txt
```diff
@@ -2752,6 +2765,7 @@ namespace GraphQL.Types
         GraphQLParser.AST.GraphQLValue ToAST(object value);
     }
     public interface IInterfaceGraphType : GraphQL.Types.IAbstractGraphType, GraphQL.Types.IComplexGraphType, GraphQL.Types.IGraphType, GraphQL.Types.IImplementInterfaces, GraphQL.Types.IMetadataReader, GraphQL.Types.IMetadataWriter, GraphQL.Types.INamedType, GraphQL.Types.IProvideDeprecationReason, GraphQL.Types.IProvideDescription, GraphQL.Types.IProvideMetadata { }
+    public interface IInterfaceGraphType<in TObject> : GraphQL.Types.IAbstractGraphType, GraphQL.Types.IComplexGraphType, GraphQL.Types.IGraphType, GraphQL.Types.IImplementInterfaces, GraphQL.Types.IInterfaceGraphType, GraphQL.Types.IMetadataReader, GraphQL.Types.IMetadataWriter, GraphQL.Types.INamedType, GraphQL.Types.IProvideDeprecationReason, GraphQL.Types.IProvideDescription, GraphQL.Types.IProvideMetadata { }
```
Shane32, line 2768: Creating this interface with `in TObject` was desired for the extension method to work properly.  We should review the use of `in` here.

src/GraphQL.ApiTests/net60/GraphQL.approved.txt
```diff
@@ -1811,18 +1824,18 @@ namespace GraphQL.Federation.Resolvers
...
-        System.Threading.Tasks.ValueTask<object?> ResolveAsync(GraphQL.IResolveFieldContext context, GraphQL.Types.IObjectGraphType graphType, object parsedRepresentation);
+        object ParseRepresentation(GraphQL.Types.IComplexGraphType graphType, System.Collections.Generic.IDictionary<string, object?> representation);
+        System.Threading.Tasks.ValueTask<object?> ResolveAsync(GraphQL.IResolveFieldContext context, GraphQL.Types.IComplexGraphType graphType, object parsedRepresentation);
     }
     public class Representation : System.IEquatable<GraphQL.Federation.Resolvers.Representation>
     {
-        public Representation(GraphQL.Types.IObjectGraphType GraphType, GraphQL.Federation.Resolvers.IFederationResolver Resolver, object Value) { }
-        public GraphQL.Types.IObjectGraphType GraphType { get; init; }
+        public Representation(GraphQL.Types.IComplexGraphType GraphType, GraphQL.Federation.Resolvers.IFederationResolver Resolver, object Value) { }
+        public GraphQL.Types.IComplexGraphType GraphType { get; init; }
```
Shane32, line 1838: `Representation` changed but it used only by `EntityResolver` and is practically an internal class.

src/GraphQL.Tests/Federation/InterfaceEntityTests.cs
```diff
@@ -0,0 +1,282 @@
+using GraphQL.Federation;
+using GraphQL.SystemTextJson;
+using GraphQL.Types;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GraphQL.Tests.Federation;
+
+public class InterfaceEntityTests
```
Shane32, line 8: Schema-first and Code-first tests were added to ensure it works properly.

src/GraphQL/Federation/Extensions/FederationGraphTypeExtensions.cs
```diff
@@ -89,4 +89,90 @@ public static void ResolveReference<TSourceType>(this ObjectGraphType<TSourceTyp
...
+    }
+
+    /// <summary>
+    /// Configures synchronous resolution of a reference using a resolver function with a return type.
+    /// </summary>
+    /// <typeparam name="TSourceType">The CLR type of the entity type, populated by the representation.</typeparam>
+    /// <param name="graphType">The graph type to apply the resolver to.</param>
+    /// <param name="resolver">The function used to resolve the source representation.</param>
+    public static void ResolveReference<TSourceType>(this IInterfaceGraphType<TSourceType> graphType, Func<IResolveFieldContext, TSourceType, TSourceType?> resolver)
+        where TSourceType : new()
```
Shane32, line 139: Here the `: new()` clause is added which will force type inference to break for interfaces, which is necessary because the keys are mapped into `TSourceType` before the resolver returns the object type.

In other words, it wouldn't be able to map `id` into `IMedia.Id` because it can't create an `IMedia`.  So we need to supply a implementing type like `Media` so that it can pass in `Media.Id` so that the resolver can create `Book` which also implements `IMedia`.

Shane32:
GraphQL.NET has generally had fairly strict backwards compatibility requirements; we almost never release a minor version with breaking changes.  So, either:
1. Release soon as 9.0.0
2. Release with version 9 when we have more to add (v7 was current for 2 years; v5 for 5 months; v4 for 1 year)
3. Release soon as 8.5.0

Considerations why to release with 8.5.0:
1. The changes only affect users using Federation
2. The Federation support (as it exists now) is new to v8, and this feature _could_ be considered somewhat of a bugfix
3. The breaking changes are not likely to impact Federation users
4. The breaking changes are even less likely to impact 3rd-party libraries
5. Fixing an application to work with 8.5.0 is probably altering two lines of code (if any).

The only thing I'm really concerned about is breaking 3rd-party library support.  I feel that the chance that a 3rd-party library directly implements the modified interface or derives from the modified classes is extremely low.  Whereas if someone needs to change two lines of code when they bump versions, that doesn't seem to be a big deal.

On the other hand, we can just bump to v9, even though it's only been 8 months.  I don't have any specific plans to add additional features to v9 so I don't see the point in waiting.

So, any thoughts @gao-artur or anyone else?

Shane32:
cc: @prowe

gao-artur:
We planned to add source generation in v9, but I've been extremely busy for the last year. I hope to start working on it sometime in October. Of course, it could also go into v10.
Another option is to release the v8.5-preview with a notice about breaking changes to try to get feedback from users. I'm not sure, though, if 3rd party library authors check the preview versions at all.
If you decide to go with a major version bump, then it's a great time to add .NET 9 support, remove obsolete APIs and cleanup some analyzers.

prowe:
I agree this might be a breaking change for specific corner cases where people are creating custom resolvers. I think it is a fairly direct translation for even those folks (swap a couple types) since they wouldn't be doing this with an interface anyway.

I would consider this a bug-fix as the Federation spec declares that resolving an interface should work. So we have to support it if we want to say we are compliant with federation v2.3.

My team would like to pull this in as quickly as possible so I'd be more than willing to pull a pre-release build of 8.5 and test it out

src/GraphQL.Tests/Federation/InterfaceEntityTests.cs
Shane32: And type-first, which required even more changes (but not more breaking changes)

Shane32:
@gao-artur @prowe Let's release this with 8.5.0 as a bug fix.  Devs that use custom resolvers will encounter a breaking change, requiring two simple changes to fix.  However, it should be seamless for the majority of federation users, and no change for any other users.

@prowe I would really appreciate it if you could pull a preview build.  We push preview builds to GitHub Packages.  Would that work for you?  Are you familiar with how to pull from there?  Note that preview builds are automatically generated once the PR is merged (so it won't be there yet).

Could I get a review from each of you when you have time?

In short, if I recall correctly, I had to make these changes:
- Entity resolver was tied only to IObjectGraphType; modified to work for interfaces
- Federation resolvers definitions modified to use `IComplexGraphType` which is object or interfaces (or input objects actually)
- `[Key]` attribute modified to allow placing on interfaces
- `AutoRegisteringInterfaceGraphType` modified to generate resolvers for static methods, required to support `[FederationResolver]`
- Extension methods added for interface graph types to allow defining resolvers on them. These extension methods will fail the `where TSource : new()` check because they are interfaces, requiring an explicit type, which needs to be a concrete class so it can be populated with the key fields
- In order fo the above extension method to work, `IInterfaceGraphType<in T>` was added so that .NET can type-check the coercion from the concrete class to the interface
- Tests added for code-first, schema-first and type-first.  Each of the three tests generates an identical schema/sdl (which is verified), and then attempts entity resolution both by interface and by object.  The schema used is Sample 1 from the Apollo Federation website regarding interface resolution.

prowe:
I will pull the preview build sometime in the next couple days and test it out.. Thank you

src/GraphQL/Types/Composite/AutoRegisteringOutputHelper.cs
```diff
@@ -99,12 +99,6 @@ public static void BuildFieldType(
         {
             throw new ArgumentOutOfRangeException(nameof(memberInfo), "Member must be a field, property or method.");
         }
-        if (buildMemberInstanceExpressionFunc == null)
-        {
-            // interface types do not set resolvers
-            fieldType.Resolver = null;
-            fieldType.StreamResolver = null;
-        }
```
Shane32, line 107: Moved to AutoRegisteringInterfaceGraphType.CreateField so that the attributes (specifically IsPrivate) gets applied first

Shane32:
@gao-artur I'll merge now so the preview build is available for @prowe .  Please review this when you get a chance regardless, and I'll make a follow-up PR if you have any comments.  Thanks!

Shane32:
The preview build is available at GitHub Packages as `GraphQL.8.5.0-preview-1130`

prowe:
@Shane32 I deployed this to our dev across 2 services and it worked out great!.. Thank you so much for the quick turnaround and I look forward to the GA release

Shane32:
8.5.0 released

dariuszkuc:
> This allows for resolving interfaces within GraphQL federation.

Just to double check - this is not related to the `@interfaceObject`? i.e. while `@interfaceObject` does refer to interface in the supergraph, in this specific subgraph it will be a regular object type (as GraphQL resolvers need to return a concrete type)

Shane32:
> > This allows for resolving interfaces within GraphQL federation.
> 
> Just to double check - this is not related to the `@interfaceObject`? i.e. while `@interfaceObject` does refer to interface in the supergraph, in this specific subgraph it will be a regular object type (as GraphQL resolvers need to return a concrete type)

Correct

Shane32:
@dariuszkuc Please let me know if you have any issues with `@interfaceObject` or anything else with Federation

PR #4175, opened 2025-09-03
Add QOL validation rule and schema visitor to detect use of deprecated elements

Shane32:
Two new classes have been added to help manage deprecated schema elements more effectively:
[`DeprecatedTypeReferenceVisitor`](src/GraphQL/Utilities/Visitors/Custom/DeprecatedTypeReferenceVisitor.cs) and
[`DeprecatedElementsValidationRule`](src/GraphQL/Validation/Rules.Custom/DeprecatedElementsValidationRule.cs).
These classes work together to provide comprehensive deprecation management at both the schema and query validation levels.

The [`DeprecatedTypeReferenceVisitor`](src/GraphQL/Utilities/Visitors/Custom/DeprecatedTypeReferenceVisitor.cs) is a schema visitor that identifies non-deprecated fields that directly reference deprecated types during schema initialization. This helps maintain schema consistency by flagging potential issues where deprecated types are still being referenced by active fields.

The [`DeprecatedElementsValidationRule`](src/GraphQL/Validation/Rules.Custom/DeprecatedElementsValidationRule.cs) is an abstract validation rule that identifies deprecated fields, arguments, and types referenced in GraphQL documents during query validation. It provides abstract methods that can be overridden to implement custom handling of deprecated element usage, such as logging warnings or collecting metrics.

```csharp
// Create a custom validation rule to handle deprecated element usage during query validation
public class LoggingDeprecatedElementsRule : DeprecatedElementsValidationRule
{
    private readonly ILogger<LoggingDeprecatedElementsRule> _logger;

    public LoggingDeprecatedElementsRule(ILogger<LoggingDeprecatedElementsRule> logger)
    {
        _logger = logger;
    }

    protected override ValueTask OnDeprecatedFieldReferencedAsync(ValidationContext context, GraphQLField fieldNode, FieldType fieldDefinition, IGraphType parentType)
    {
        _logger.LogWarning("Deprecated field '{ParentType}.{FieldName}' was used: {DeprecationReason}",
            parentType.Name, fieldDefinition.Name, fieldDefinition.DeprecationReason);
        return default;
    }

    protected override ValueTask OnDeprecatedFieldArgumentReferencedAsync(ValidationContext context, GraphQLArgument argumentNode, QueryArgument argumentDefinition, FieldType fieldDefinition, IGraphType parentType)
    {
        _logger.LogWarning("Deprecated argument '{ParentType}.{FieldName}.{ArgumentName}' was used: {DeprecationReason}",
            parentType.Name, fieldDefinition.Name, argumentDefinition.Name, argumentDefinition.DeprecationReason);
        return default;
    }

    protected override ValueTask OnDeprecatedDirectiveArgumentReferencedAsync(ValidationContext context, GraphQLArgument argumentNode, QueryArgument argumentDefinition, Directive directiveDefinition)
    {
        _logger.LogWarning("Deprecated directive argument '@{DirectiveName}.{ArgumentName}' was used: {DeprecationReason}",
            directiveDefinition.Name, argumentDefinition.Name, argumentDefinition.DeprecationReason);
        return default;
    }

    protected override ValueTask OnDeprecatedTypeReferencedAsync(ValidationContext context, GraphQLNamedType typeConditionNode, IGraphType typeDefinition)
    {
        _logger.LogWarning("Deprecated type '{TypeName}' was used: {DeprecationReason}",
            typeDefinition.Name, typeDefinition.DeprecationReason);
        return default;
    }
}

// Register the schema visitor and validation rule
services.AddGraphQL(b => b
    .AddSchema<MySchema>()
    .AddSchemaVisitor<DeprecatedTypeReferenceVisitor>()
    .AddValidationRule<LoggingDeprecatedElementsRule>()
);
```

### DIFF
--- a/src/GraphQL.Tests/Utilities/Visitors/DeprecatedTypeReferenceVisitorTests.cs
+++ b/src/GraphQL.Tests/Utilities/Visitors/DeprecatedTypeReferenceVisitorTests.cs
@@ -1,0 +1,273 @@
+using GraphQL.Types;
+using GraphQL.Utilities.Visitors.Custom;
+
+namespace GraphQL.Tests.Utilities.Visitors;
+
+public class DeprecatedTypeReferenceVisitorTests
+{
+    [Fact]
+    public void Should_Throw_When_NonDeprecated_Field_References_Deprecated_Type()
+    {
+        // Arrange
+        var deprecatedType = new ObjectGraphType
+        {
+            Name = "DeprecatedUser",
+            DeprecationReason = "Use NewUser instead"
+        };
+        deprecatedType.Field<StringGraphType>("id");
+
+        var activeType = new ObjectGraphType { Name = "Post" };
+        activeType.Field<StringGraphType>("title");
+        // Non-deprecated field referencing deprecated type
+        activeType.Field<ObjectGraphType>("author").Type(deprecatedType);
+
+        var schema = new Schema
+        {
+            Query = activeType
+        };
+        schema.RegisterType(deprecatedType);
+        schema.RegisterVisitor(new DeprecatedTypeReferenceVisitor());
+
+        // Act & Assert
+        var exception = Should.Throw<InvalidOperationException>(() => schema.Initialize());
+        exception.Message.ShouldBe("Non-deprecated field 'Post.author' references deprecated type 'DeprecatedUser'.");
+    }
+
+    [Fact]
+    public void Should_Not_Throw_When_Deprecated_Field_References_Deprecated_Type()
+    {
+        // Arrange
+        var deprecatedType = new ObjectGraphType
+        {
+            Name = "DeprecatedUser",
+            DeprecationReason = "Use NewUser instead"
+        };
+        deprecatedType.Field<StringGraphType>("id");
+
+        var activeType = new ObjectGraphType { Name = "Post" };
+        activeType.Field<StringGraphType>("title");
+        // Deprecated field referencing deprecated type - this should be allowed
+        activeType.Field<ObjectGraphType>("author")
+            .Type(deprecatedType)
+            .DeprecationReason("Use newAuthor field instead");
+
+        var schema = new Schema
+        {
+            Query = activeType
+        };
+        schema.RegisterType(deprecatedType);
+        schema.RegisterVisitor(new DeprecatedTypeReferenceVisitor());
+
+        // Act & Assert
+        schema.Initialize();
+    }
+
+    [Fact]
+    public void Should_Throw_AggregateException_When_Multiple_Violations()
+    {
+        // Arrange
+        var deprecatedType1 = new ObjectGraphType
+        {
+            Name = "DeprecatedUser",
+            DeprecationReason = "Use NewUser instead"
+        };
+        deprecatedType1.Field<StringGraphType>("id");
+
+        var deprecatedType2 = new ObjectGraphType
+        {
+            Name = "DeprecatedCategory",
+            DeprecationReason = "Use NewCategory instead"
+        };
+        deprecatedType2.Field<StringGraphType>("name");
+
+        var activeType = new ObjectGraphType { Name = "Post" };
+        activeType.Field<StringGraphType>("title");
+        activeType.Field<ObjectGraphType>("author").Type(deprecatedType1);
+        activeType.Field<ObjectGraphType>("category").Type(deprecatedType2);
+
+        var schema = new Schema
+        {
+            Query = activeType
+        };
+        schema.RegisterType(deprecatedType1);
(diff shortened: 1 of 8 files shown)


gao-artur:
Oh wow! That's a great addition! It will definitely help us to fix the inconsistency in the schema definition.
I think your defaults are reasonable. Whoever wants to log the visitor's output can always inherit it and override the `PostVisitSchema` method (that's what I'll probably do until I fix all the fields). It could be great to have a boolean flag in the constructor to switch between logging and exception, but unfortunately, we don't have the `ILogger` dependency.

src/GraphQL.Tests/Utilities/Visitors/DeprecatedTypeReferenceVisitorTests.cs
```diff
@@ -0,0 +1,273 @@
...
+        };
+        schema.RegisterType(activeUserType);
+        schema.RegisterVisitor(new DeprecatedTypeReferenceVisitor());
+
+        // Act & Assert
+        schema.Initialize();
+    }
+
+    [Fact]
+    public void Should_Throw_When_ObjectField_References_Deprecated_Type()
```
gao-artur, line 101: Duplicate of `Should_Throw_When_NonDeprecated_Field_References_Deprecated_Type` (except `title` field)

src/GraphQL.Tests/Utilities/Visitors/DeprecatedTypeReferenceVisitorTests.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        var schema = new Schema { Query = activeType };
+        schema.RegisterType(deprecatedType);
+        schema.RegisterVisitor(new DeprecatedTypeReferenceVisitor());
+
+        // Act & Assert
+        var exception = Should.Throw<InvalidOperationException>(() => schema.Initialize());
+        exception.Message.ShouldBe("Non-deprecated field 'Post.authors' references deprecated type 'DeprecatedUser'.");
+    }
+}
```
gao-artur, line 245: What happens when all type fields are deprecated? Should the type be marked as deprecated too?

src/GraphQL.Tests/Utilities/Visitors/DeprecatedTypeReferenceVisitorTests.cs
Shane32: removed duplicate

src/GraphQL.Tests/Utilities/Visitors/DeprecatedTypeReferenceVisitorTests.cs
Shane32: Not sure.  I was sorta copying the C# analyzer for obsolete classes/members.  In general, I suppose I would stick with the C# analyzer behavior, and not trigger a warning in the described instance.  It seems that a developer would have to go through a lot of trouble to mark every member as deprecated without marking the type as deprecated.  The only benefit I suppose is if they were ignorant of the fact that types could be marked as deprecated.  I think I'd leave it as-is.

Shane32:
> Is it worth implementing a static analyzer to mimic the `DeprecatedTypeReferenceVisitor`? It can confuse the user as to why he needs to register the visitor if an analyzer already detects potential problems. But an analyzer can't be as comprehensive as a visitor, and it may have some false negatives.

As nice as it sounds, it seems like it would be complex to implement.  You'd have to detect and flag in schema-first, code-first, and type-first scenarios.  You'd have to detect type mappings, so that the proper graph type is examined when fields infer the graph type from the CLR type (e.g. `Field(x => x.Name)`).  I use some helper libraries to building my schemas, such as [GraphQL.DI](https://www.nuget.org/packages/Shane32.GraphQL.DI), so these solutions may not even work for me.

Shane32:
> It could be great to have a boolean flag in the constructor to switch between logging and exception, but unfortunately, we don't have the `ILogger` dependency.

Side topic, but maybe in a future version we should make the base GraphQL.NET package consume a few standard MS packages such as:
- Microsoft.Extensions.DependencyInjection.Abstractions
- Microsoft.Extensions.Logging.Abstractions
- System.Diagnostics.DiagnosticSource
- System.Text.Json

On the other hand, I like the fact that GraphQL.NET has no dependencies.  Another option is to have a framework reference to `Microsoft.AspNetCore.App` for .NET 5+ which should include all of the above dependencies.  So GraphQL.NET will remain "dependency-free" for .NET 5+ while still supporting .NET Framework.  I'd just want to be sure this doesn't cause a problem with AOT compilation scenarios.

gao-artur (approved):

PR #4191, opened 2025-10-19
Use collection initializers

gao-artur:
Another try to fix `Use collection initializers or expressions`, but this time using [IDE0028](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0028) and [IDE0306](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0306) code fixes that are much less aggressive than ReSharper. Note that `dotnet_style_collection_initializer = true:error` was already configured in `.editorconfig`, I only added `dotnet_style_prefer_collection_expression = when_types_exactly_match` to avoid semantic changes.

### DIFF
--- a/samples/GraphQL.Federation.CodeFirst.Sample3/Program.cs
+++ b/samples/GraphQL.Federation.CodeFirst.Sample3/Program.cs
@@ -1,7 +1,6 @@
 using GraphQL.Federation.CodeFirst.Sample3.Schema;
 using GraphQL.Transport;
 using GraphQL.Types;
-using GraphQL.Utilities;
 
 namespace GraphQL.Federation.CodeFirst.Sample3;
 
--- a/samples/GraphQL.Federation.TypeFirst.Sample4/Program.cs
+++ b/samples/GraphQL.Federation.TypeFirst.Sample4/Program.cs
@@ -1,7 +1,6 @@
 using GraphQL.Federation.TypeFirst.Sample4.Schema;
 using GraphQL.Transport;
 using GraphQL.Types;
-using GraphQL.Utilities;
 
 namespace GraphQL.Federation.TypeFirst.Sample4;
 
--- a/src/GraphQL.MicrosoftDI.Tests/AdapterTests.cs
+++ b/src/GraphQL.MicrosoftDI.Tests/AdapterTests.cs
@@ -1,6 +1,5 @@
 using GraphQL.Execution;
 using GraphQL.Types;
-using GraphQL.Validation;
 using GraphQLParser.AST;
 using Moq;
 
--- a/src/GraphQL.MicrosoftDI.Tests/ConnectionAdapterTests.cs
+++ b/src/GraphQL.MicrosoftDI.Tests/ConnectionAdapterTests.cs
@@ -1,7 +1,6 @@
 using GraphQL.Builders;
 using GraphQL.Execution;
 using GraphQL.Types;
-using GraphQL.Validation;
 using GraphQLParser.AST;
 using Moq;
 
--- a/src/GraphQL.MicrosoftDI.Tests/GraphQLBuilderExtensionTests.cs
+++ b/src/GraphQL.MicrosoftDI.Tests/GraphQLBuilderExtensionTests.cs
@@ -1,5 +1,4 @@
 using GraphQL.Types;
-using GraphQL.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace GraphQL.MicrosoftDI.Tests;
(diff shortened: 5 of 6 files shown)


src/GraphQL/Execution/Errors/ExecutionErrors.cs
```diff
@@ -17,7 +17,7 @@ internal ExecutionErrors(int capacity)
 
     internal ExecutionErrors(ExecutionError error)
     {
-        List = new(1) { error };
+        List = [error];
```
Shane32, line 20: I would assume this is optimized, `List<T>` being an "intrinsic" type, but hard to see.  Not a big deal.

src/GraphQL/Federation/FederationHelper.cs
```diff
@@ -246,7 +246,7 @@ void RegisterDirective(string directiveName,
                 return;
             var directive = new Directive(link.NameForDirective(directiveName))
             {
-                Arguments = arguments != null ? new QueryArguments(arguments) : null,
+                Arguments = arguments != null ? [.. arguments] : null,
```
Shane32, line 249: Again, performance is unclear here.  Does it call the constructor that accepts an enumerable or call Add a bunch of times?  Again, no big deal.

src/GraphQL/Execution/ExecutionStrategy.cs
```diff
@@ -164,7 +164,7 @@
         Debug.Assert(parent.Result != null || parent is RootExecutionNode);
 
         var parentType = parent.GetObjectGraphType(context.Schema) ?? parent.GraphType;
-        var fields = _reusableFields ??= new();
+        var fields = _reusableFields ??= [];
```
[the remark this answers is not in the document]
Shane32: The static field is marked `[ThreadStatic]` so it is thread safe as-is.

src/GraphQL/Execution/Errors/ExecutionErrors.cs
gao-artur: This code
```c#
List<string> l = ["xx"];
```
is [translated](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQAUoKgW2QYQHsAbYiAYxgEtCA7ZAHwAEAmARgFgAoJgBgAImbACwBubtyYBmQS375+Ab279VgmU2H8AsgAoAlMq5qT/ADJVkMADxDeAPn7F+AXn4BtAEQAPb54C64sZqAL7cIUA=) to this
```c#
List<string> list = new List<string>();
CollectionsMarshal.SetCount(list, 1);
Span<string> span = CollectionsMarshal.AsSpan(list);
int num = 0;
span[num] = "xx";
num++;
```

So it creates a list with a default capacity of 0. I would say, it has similar performance.

src/GraphQL/Federation/FederationHelper.cs
gao-artur: The `QueryArguments` constructor also calls `Add` in a loop, so unless it's changed, it performs similarly.
This
```c#
public class C {
    public void M(string[]? arguments){
        QueryArguments? l = arguments != null ? [.. arguments] : null;
    }
}
```
[Translates](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQAUoKgW2QYQHsAbYiAYxgEtCA7ZAHwAEAmARgFgAoJgBgAImbACwBubn0FsAdEVIVqdZOK4SAzIJb98/AN7d+hwRqbD+AWQAUQ3gG0AugH5+WAOZwcEWjGQBKfVxGQfwAinAQCACeAIII7p7eyM7E/AC8LnEeXj78AITptHCk/M620tIZ8dnI9vwg/IWkKkEAvtxtqjwmWmERMZkJOfVCagA8NgB83AFBTBq9UbFViZYADli4yFJ2tW5+BkYzwYZghEhQ5AAW/JYAbliV/FS0lfuBx3oHH0bRACa/ljcvmaHw6rS+xn4AEkAKKFTzYGBncZsXgTfgAcQgMDhWURZ0svjS6JglwQhAA7g0IFSAHKEGBQnCrMiDCC/GEAD3IEFWilohJUENh8IisDOUjU0ixONF+IQhOJmOxuIR4oVwIhELmgjMfwBNkqWW8/g6LSAA===) to this
```c#
public class C
{
    public void M([Nullable(new byte[] { 2, 1 })] string[] arguments)
    {
        if (arguments != null)
        {
            QueryArguments queryArguments = new QueryArguments();
            int num = 0;
            while (num < arguments.Length)
            {
                string argument = arguments[num];
                queryArguments.Add(argument);
                num++;
            }
            QueryArguments queryArguments2 = queryArguments;
        }
        else
        {
            QueryArguments queryArguments2 = null;
        }
    }
}
....
```

PR #4193, opened 2025-10-21
Add field-level middleware and resolve field context accessor

Shane32:


src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -1087,6 +1088,7 @@ namespace GraphQL.Builders
     {
         protected FieldBuilder(GraphQL.Types.FieldType fieldType) { }
         public GraphQL.Types.FieldType FieldType { get; }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ApplyMiddleware(GraphQL.Instrumentation.IFieldMiddleware middleware) { }
```
Shane32, line 1095: Name?

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -1517,6 +1519,10 @@ namespace GraphQL.Execution
     {
         System.Collections.Generic.IDictionary<string, object?> UserContext { get; }
     }
+    public interface IResolveFieldContextAccessor
+    {
+        GraphQL.IResolveFieldContext Context { get; set; }
+    }
```
Shane32, line 1532: Currently within GraphQL.Execution namespace

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -1567,6 +1573,11 @@ namespace GraphQL.Execution
         public RequestError(string message) { }
         public RequestError(string message, System.Exception? innerException) { }
     }
+    public class ResolveFieldContextAccessor : GraphQL.Execution.IResolveFieldContextAccessor
+    {
+        public ResolveFieldContextAccessor() { }
+        public GraphQL.IResolveFieldContext Context { get; set; }
+    }
```
Shane32: Public or private?

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -1567,6 +1573,11 @@ namespace GraphQL.Execution
         public RequestError(string message) { }
         public RequestError(string message, System.Exception? innerException) { }
     }
+    public sealed class ResolveFieldContextAccessor : GraphQL.Execution.IResolveFieldContextAccessor
+    {
+        public GraphQL.IResolveFieldContext Context { get; set; }
+        public static GraphQL.Execution.ResolveFieldContextAccessor Instance { get; }
+    }
```
Shane32, line 1587: Public ?

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -3072,6 +3084,7 @@ namespace GraphQL.Types
         public GraphQL.Types.IObjectGraphType? Mutation { get; set; }
         public GraphQL.Conversion.INameConverter NameConverter { get; set; }
         public GraphQL.Types.IObjectGraphType Query { get; set; }
+        public GraphQL.Execution.IResolveFieldContextAccessor? ResolveFieldContextAccessor { get; set; }
```
Shane32, line 3100: Could use a boolean here - for example, `Schema.UseResolveFieldContextAccessor`.  Since the accessor implementation is always a singleton, it hardly needs to have the instance specified.  But currently this is set to an instance pulled from dependency injection.

src/GraphQL.ApiTests/net50/GraphQL.MicrosoftDI.approved.txt
```diff
@@ -150,11 +150,21 @@ namespace GraphQL.MicrosoftDI
         public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ResolveScoped<TSourceType, TReturnType>(this GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> builder, System.Func<GraphQL.IResolveFieldContext<TSourceType>, TReturnType?> resolver) { }
         public static GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ResolveScopedAsync<TSourceType, TReturnType>(this GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> builder, System.Func<GraphQL.IResolveFieldContext<TSourceType>, System.Threading.Tasks.Task<TReturnType?>> resolver) { }
     }
+    public class ScopedFieldMiddleware : GraphQL.Instrumentation.IFieldMiddleware
+    {
+        public static readonly GraphQL.MicrosoftDI.ScopedFieldMiddleware Instance;
+        public ScopedFieldMiddleware(GraphQL.Instrumentation.IFieldMiddleware? innerMiddleware) { }
+        public System.Threading.Tasks.ValueTask<object?> ResolveAsync(GraphQL.IResolveFieldContext context, GraphQL.Instrumentation.FieldMiddlewareDelegate next) { }
+    }
```
Shane32, line 158: Now scoped fields just apply middleware to the field with this class.  It could be internal, however, it should replace ScopedFieldResolver so I left it public.

src/GraphQL.ApiTests/net50/GraphQL.MicrosoftDI.approved.txt
Shane32, line 172: ScopedFieldResolver is now obsolete and unused by GraphQL.MicrosoftDI.

src/GraphQL.MicrosoftDI/DynamicScopedSourceStreamResolver.cs
Shane32, line 1: Since I did not add field middleware functionality to `SourceStreamResolver`s, this class still remains.

src/GraphQL.MicrosoftDI/ScopedFieldBuilderExtensions.cs
```diff
@@ -9,18 +9,37 @@ namespace GraphQL.MicrosoftDI;
 /// </summary>
 public static class ScopedFieldBuilderExtensions
 {
+    /// <summary>
+    /// Applies scoped middleware to the field. A dependency injection scope is created for the duration of the field resolver's execution
+    /// and the scoped service provider is passed within <see cref="IResolveFieldContext.RequestServices"/>.
+    /// </summary>
+    [AllowedOn<IObjectGraphType>]
+    internal static FieldBuilder<TSourceType, TReturnType> Scoped<TSourceType, TReturnType>(this FieldBuilder<TSourceType, TReturnType> builder)
```
Shane32, line 17: Currently internal.  Could be public, but we already have ResolveScoped plus the Resolve() builder methods such as `Resolve().WithScope().WithService<T>` etc

src/GraphQL.MicrosoftDI/ScopedFieldMiddleware.cs
```diff
@@ -0,0 +1,42 @@
+using GraphQL.Instrumentation;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GraphQL.MicrosoftDI;
+
+/// <summary>
+/// Field middleware that creates a dependency injection scope for the duration of the field resolver's execution.
+/// </summary>
+public class ScopedFieldMiddleware : IFieldMiddleware
```
Shane32, line 9: Notice that within ScopedFieldMiddleware, there is no need to adjust the context accessor (unlike DynamicSourceStreamResolver)

src/GraphQL.MicrosoftDI/ScopedFieldResolver.cs
```diff
@@ -9,6 +10,7 @@ namespace GraphQL.MicrosoftDI;
 /// to create a dependency injection scope. Then it calls a predefined <see cref="Func{T, TResult}"/>, passing the scoped service provider
 /// within <see cref="IResolveFieldContext.RequestServices"/>, and returns the result.
 /// </summary>
+[Obsolete("Please use ScopedFieldMiddleware instead. This class will be removed in a future release.")]
 public class ScopedFieldResolver<TReturnType> : FuncFieldResolver<TReturnType>
```
Shane32, line 14: ScopedFieldResolver is unused by GraphQL.MicrosoftDI now.  However if a developer did use it (for instance within GraphQL.Conventions), it will adjust the context accessor properly, which is pretty important.  Most of the other resolver wrappers are just type casting, and since the accessor only ever returns a non-strongly-typed `IResolveFieldContext`, it matters little whether it has the initial reference or the one with type casting.

src/GraphQL/Execution/ResolveFieldContextAccessor.cs
```diff
@@ -0,0 +1,23 @@
...
+    private static readonly AsyncLocal<IResolveFieldContext?> _context = new();
+    private ResolveFieldContextAccessor() { }
+
+    /// <summary>
+    /// Singleton instance of the <see cref="ResolveFieldContextAccessor"/>.
+    /// </summary>
+    public static ResolveFieldContextAccessor Instance { get; } = new();
+
+    /// <inheritdoc/>
+    public IResolveFieldContext? Context
```
Shane32, line 18: Unlike `IDataLoaderContextAccessor` this can return `null`.  This allows for it to be used within services (e.g. for logging), where it may or may not be called from a GraphQL field resolver.

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -652,6 +652,33 @@ public static IGraphQLBuilder AddDocumentListener<TDocumentListener>(this IGraph
...
+    /// <c>IHttpContextAccessor</c> in ASP.NET Core. Registers <see cref="IResolveFieldContextAccessor"/>
+    /// as a singleton and configures the schema to populate the accessor during field execution.
+    /// </summary>
+    public static IGraphQLBuilder AddResolveFieldContextAccessor(this IGraphQLBuilder builder)
+    {
+        // Register the accessor as a singleton
+        builder.Services.TryRegister<IResolveFieldContextAccessor>(ResolveFieldContextAccessor.Instance);
+
+        // Configure the schema to use the accessor
+        builder.ConfigureSchema<ConfigureResolveFieldContextAccessor>();
```
Shane32, line 668: The design allows for `AddResolveFieldContextAccessor` to be called any number of times.

src/GraphQL/Resolvers/FuncFieldResolver.cs
```diff
@@ -107,7 +107,7 @@ public FuncFieldResolver(Func<IResolveFieldContext<TSourceType>, ValueTask<TRetu
                 {
                     adapter.Set(context);
                 }
-                var ret = resolver(adapter);
+                object? ret = resolver(adapter);
```
Shane32, line 110: In the rare circumstance that this code is executed, prevents value types from being boxed twice in the same method - once during the call to `CanReuseContextForValue` and once during `return ret;`

src/GraphQL/Resolvers/ResolveFieldContextAccessorResolver.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        _accessor.Context = context;
+        object? ret = null;
+        try
+        {
+            ret = await _innerResolver.ResolveAsync(context).ConfigureAwait(false);
+        }
+        finally
+        {
+            _accessor.Context = null;
```
Shane32, line 29: I was only going to have `_accessor.Context = context;` and forgo setting it back to null.  More notably it would allow removal of the `await` which duplicates the async context.  But performance testing did not show any notable difference or additional memory use.

src/GraphQL/Types/Collections/SchemaTypes.cs
```diff
@@ -514,7 +508,27 @@ public void ApplyMiddleware(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegat
...
+                    else if (transform != null)
+                    {
+                        // Only schema-wide middleware
+                        fieldMiddlewareDelegate = transform(inner.ResolveAsync);
+                    }
+                    else
+                    {
+                        // No middleware at all, skip this field
+                        continue;
+                    }
```
Shane32: The changes here effectively just allow for field-specific middleware and have nothing to do with ResolveFieldContextAccessor.

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -93,6 +94,11 @@ public Type? Type
     /// </summary>
     public ISourceStreamResolver? StreamResolver { get; set; }
 
+    /// <summary>
+    /// Gets or sets field middleware for the field. Only applicable to fields of output graph types.
+    /// </summary>
+    public IFieldMiddleware? Middleware { get; set; }
```
Shane32: There is a single new field defined on `FieldType`.  It reuses the same `IFieldMiddleware` interface that already exists within the project.  The extension method `ApplyMiddleware` will chain middleware together if called multiple times.  But for the majority of use cases, this will be null or set to `ScopedFieldMiddleware.Instance`.

src/GraphQL/Types/Schema.cs
```diff
@@ -169,6 +171,13 @@ public static Schema For<TSchemaBuilder>(string typeDefinitions, Action<TSchemaB
     /// <inheritdoc/>
     public IFieldMiddlewareBuilder FieldMiddleware { get; internal set; } = new FieldMiddlewareBuilder();
 
+    /// <summary>
+    /// Gets or sets the <see cref="IResolveFieldContextAccessor"/> instance to be used for populating
+    /// the current field context during execution. When set, middleware will be automatically applied
+    /// to all fields to populate the accessor.
+    /// </summary>
+    public IResolveFieldContextAccessor? ResolveFieldContextAccessor { get; set; }
```
Shane32, line 178: Again, this field on `Schema` could be a boolean or something instead of a reference.  Alternatively, it could be a built-in feature of GraphQL.NET.

src/GraphQL/Types/Schema.cs
```diff
@@ -466,6 +475,34 @@ IEnumerable<ISchemaNodeVisitor> GetVisitors()
         {
             // At this point, Initialized will return false, and Initialize will still lock while waiting for initialization to complete.
             // However, AllTypes and similar properties will return a reference to SchemaTypes without waiting for a lock.
+
+            // Wrap resolvers with context accessor if configured
+            if (ResolveFieldContextAccessor != null)
```
Shane32, line 479: The code here is what enables the resolve field context accessor functionality.  It is essentially a schema visitor, looking for fields of object graph types, and wrapping their resolvers.  It skips default resolvers, as they do not require the context accessor.  Then the middleware is applied as it always has been, which gets applied over (before) the resolver runs.  We can't just add schema-level middleware to do this, as schema-level middleware is always called before field-level middleware (with this PR).  I could rewrite this code in a separate method for organization, or within a schema visitor for clarity.

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -652,6 +652,33 @@ public static IGraphQLBuilder AddDocumentListener<TDocumentListener>(this IGraph
...
+
+        return builder;
+    }
+
+    private sealed class ConfigureResolveFieldContextAccessor : IConfigureSchema
+    {
+        public void Configure(ISchema schema, IServiceProvider serviceProvider)
+        {
+            ((Schema)schema).ResolveFieldContextAccessor = serviceProvider.GetRequiredService<IResolveFieldContextAccessor>();
```
Shane32, line 677: Since properties cannot be added to `ISchema` without a breaking change, there is a cast here.  v9 of GraphQL.NET should include the property on `ISchema`.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
gao-artur: I like this one

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -488,4 +489,47 @@ public virtual FieldBuilder<TSourceType, TReturnType> ComplexityImpact(double im
     [AllowedOn<IObjectGraphType>]
     public virtual FieldBuilder<TSourceType, TReturnType> DependsOn<TService>()
         => this.DependsOn(typeof(TService));
+
+    /// <summary>
+    /// Applies middleware to the field. If middleware is already set, the new middleware will be chained after the existing middleware.
```
gao-artur, line 501: What about the global middleware? Will it come before or after field middleware?

src/GraphQL/Resolvers/ResolveFieldContextAccessorResolver.cs
```diff
@@ -0,0 +1,34 @@
...
+        _innerResolver = innerResolver ?? throw new ArgumentNullException(nameof(innerResolver));
+    }
+
+    public async ValueTask<object?> ResolveAsync(IResolveFieldContext context)
+    {
+        _accessor.Context = context;
+        object? ret = null;
+        try
+        {
+            ret = await _innerResolver.ResolveAsync(context).ConfigureAwait(false);
```
gao-artur: Any specific reason for using the `ret` local var and not returning here?

src/GraphQL/Resolvers/ResolveFieldContextAccessorResolver.cs
gao-artur: I'd say setting it to `null` is important to prevent accessor misuse. Check this discussion https://github.com/dotnet/aspnetcore/issues/3581.
But setting `null` may not be enough, because if someone started a concurrent task from the resolver, it will continue to see the context reference from its own copy of the `AsyncLocal`. That's why this object indirection is needed https://github.com/dotnet/aspnetcore/blob/main/src/Http/Http/src/HttpContextAccessor.cs#L37-L40

src/GraphQL/Types/Schema.cs
gao-artur: I think putting it in a schema visitor is cleaner

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -488,4 +489,47 @@ public virtual FieldBuilder<TSourceType, TReturnType> ComplexityImpact(double im
...
+    /// </summary>
+    /// <param name="middleware">The middleware to apply.</param>
+    [AllowedOn<IObjectGraphType>]
+    public virtual FieldBuilder<TSourceType, TReturnType> ApplyMiddleware(IFieldMiddleware middleware)
```
gao-artur, line 505: Is it possible to add another overload?
```c#
public virtual FieldBuilder<TSourceType, TReturnType> ApplyMiddleware<TFieldMiddleware >()
    where TFieldMiddleware : IFieldMiddleware
```

And resolve it from DI?

gao-artur: And also one that accepts a delegate from simple middlewares

src/GraphQL/Resolvers/ResolveFieldContextAccessorResolver.cs
Shane32: Makes sense except we don’t provide any similar guarantees for IResolveFieldContext, which is already reused by the execution strategy (except for certain cases) and so essentially suffers from this problem already. It seems like making this change would further consume memory, and in this scenario where 99% of the time it’s just a simple field access, it would be detrimental to performance. 

How important do you think this is to do?

gao-artur: We can document a warning about this issue instead.

Shane32:
> Introspection with cache allocates more memory than without cache? 🤔

🤷‍♂️ I re-ran the tests again to be sure, and that's how it comes out.  I only added `.AddResolveFieldContextAccessor()` to the existing benchmark test.

src/GraphQL/Builders/FieldBuilder.cs
Shane32: we can't have it resolve from DI as currently designed.  we would need to change the FieldType.Middleware signature to be a resolver transformation delegate which receives a copy of IServiceProvider.  then at initialization, it would pass the service provider through the transformation delegates along with the original resolver to create the transformed resolver.  should I do that?

Shane32: > And also one that accepts a delegate from simple middlewares

???  I guess I'm not exactly sure what you mean here.

gao-artur: > should I do that

I think without it, the solution is incomplete. When injecting dependencies into the graph's constructor, you can use a middleware instance to set up fields
```c#
public class UserGraphType : ObjectGraphType<User>
{
    public UserGraphType(MyMiddleware myMiddleware)
    {
        Field(...).ApplyMiddleware(myMiddleware);
        Field(...);
        Field(...).ApplyMiddleware(myMiddleware);
    }
}
```

But if you need to use multiple different middlewares, especially in types with many fields, using DI looks more elegant
```c#
public class UserGraphType : ObjectGraphType<User>
{
    public UserGraphType()
    {
        Field(...).ApplyMiddleware<MyMiddleware>();
        Field(...);
        Field(...).ApplyMiddleware<AnotherMiddleware>();
    }
}
```

Also, when using `WithService` extension, you have a kind of configuration mix
```c#
public class UserGraphType : ObjectGraphType<User>
{
    public UserGraphType(MyMiddleware myMiddleware)
    {
        Field(...)
            .Resolve()
            .WithService<IService>()
            .Resolve((ctx, service) => ...)
            .ApplyMiddleware(myMiddleware);
    }
}
```

The `IService` is provided with a resolver delegate, while `myMiddleware` is injected via the constructor.

> ??? I guess I'm not exactly sure what you mean here.

I mean the one you demonstrated in the documentation
```c#
{
    public MyGraphType()
    {
        Field<StringGraphType>("myField")
            .Resolve(context => "Hello World")
            .ApplyMiddleware(async (context, next) =>
            {
                // Code before resolver execution
                var result = await next(context);
                // Code after resolver execution
                return result;
            });
    }
}
```

src/GraphQL/Resolvers/ResolveFieldContextAccessorResolver.cs
Shane32: Fixed

src/GraphQL/Builders/FieldBuilder.cs
Shane32: Executes in this order:

1. Global middleware, in the order it was added
2. Field middleware, in the order it was added
3. Set resolve field context accessor
4. Transform `IResolveFieldContext` to typed e.g. `IResolveFieldContext<TSource>`
5. Run user code

src/GraphQL/Types/Schema.cs
Shane32: Done

src/GraphQL/Resolvers/ResolveFieldContextAccessorResolver.cs
Shane32: Fixed as you originally suggested.

src/GraphQL/Builders/FieldBuilder.cs
Shane32: oh, right. 🤦‍♂️

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -1087,6 +1088,10 @@ namespace GraphQL.Builders
     {
         protected FieldBuilder(GraphQL.Types.FieldType fieldType) { }
         public GraphQL.Types.FieldType FieldType { get; }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ApplyMiddleware(GraphQL.Instrumentation.IFieldMiddleware middleware) { }
+        public virtual GraphQL.Builders.FieldBuilder<TSourceType, TReturnType> ApplyMiddleware(System.Func<GraphQL.IResolveFieldContext, GraphQL.Instrumentation.FieldMiddlewareDelegate, System.Threading.Tasks.ValueTask<object?>> middleware) { }
```
Shane32: This overload differs in its signature from `schema.FieldMiddleware.Use(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> transform)`

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -2645,6 +2665,7 @@ namespace GraphQL.Types
         public string? DeprecationReason { get; set; }
         public string? Description { get; set; }
         public bool IsPrivate { get; set; }
+        public System.Func<System.IServiceProvider, GraphQL.Instrumentation.IFieldMiddleware>? MiddlewareFactory { get; set; }
```
Shane32: Renamed to `MiddlewareFactory` to more accurately represent its functionality.

src/GraphQL.Benchmarks/GraphQL.Benchmarks.csproj
```diff
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.15.2" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="9.0.10" />
```
Shane32, line 11: No idea why, but the benchmarks won't run without this anymore.

src/GraphQL.Tests/Instrumentation/MiddlewareOrderTests.cs
```diff
@@ -0,0 +1,722 @@
+using GraphQL.Instrumentation;
+using GraphQL.Types;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GraphQL.Tests.Instrumentation;
+
+/// <summary>
+/// Tests to verify the order in which schema and field middleware are applied and executed.
+/// </summary>
+public class MiddlewareOrderTests
```
Shane32, line 10: AI went a bit bonkers with the tests here.... but they work....

src/GraphQL/Builders/FieldBuilder.cs
Shane32: done

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -488,4 +497,93 @@ public virtual FieldBuilder<TSourceType, TReturnType> ComplexityImpact(double im
...
+
+        public ChainedFieldMiddleware(IFieldMiddleware first, IFieldMiddleware second)
+        {
+            _first = first;
+            _second = second;
+        }
+
+        public ValueTask<object?> ResolveAsync(IResolveFieldContext context, FieldMiddlewareDelegate next)
+        {
+            return _first.ResolveAsync(context, ctx => _second.ResolveAsync(ctx, next));
```
Shane32: This line of code will induce an allocation every time it is executed.  I would like to fix

Shane32: Fortunately, the only way this would happen is if a user defines two different field middlewares, one of which could be a scoped field middleware.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -257,6 +257,10 @@ namespace GraphQL
     }
     public static class FieldExtensions
     {
+        public static void ApplyMiddleware(this GraphQL.Types.FieldType fieldType, GraphQL.Instrumentation.IFieldMiddleware middleware) { }
+        public static void ApplyMiddleware(this GraphQL.Types.FieldType fieldType, System.Func<GraphQL.Instrumentation.FieldMiddlewareDelegate, GraphQL.Instrumentation.FieldMiddlewareDelegate> middleware) { }
+        public static void ApplyMiddleware<TMiddleware>(this GraphQL.Types.FieldType fieldType)
+            where TMiddleware : GraphQL.Instrumentation.IFieldMiddleware { }
```
Shane32, line 263: Added extension methods to FieldType so it can be used from ScopedAttribute and similar uses where there is no field builder available.

Shane32:
Final benchmarks with middleware installed:

| Method        | UseCaching | Mean       | Error     | StdDev    | Gen0    | Gen1   | Allocated |
|-------------- |----------- |-----------:|----------:|----------:|--------:|-------:|----------:|
| Introspection | False      | 360.605 us | 7.0865 us | 7.8766 us | 12.2070 | 3.9063 | 224.52 KB |
| Hero          | False      |   8.808 us | 0.0958 us | 0.0896 us |  0.3510 |      - |   6.48 KB |
| Introspection | True       | 205.639 us | 3.0805 us | 2.8815 us | 10.0098 | 2.9297 | 187.35 KB |
| Hero          | True       |   1.573 us | 0.0314 us | 0.0323 us |  0.1469 |      - |   2.73 KB |

src/GraphQL/Builders/FieldBuilder.cs
Shane32: added tests to ensure middleware order is correct

src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -488,4 +496,38 @@ public virtual FieldBuilder<TSourceType, TReturnType> ComplexityImpact(double im
...
+        FieldType.ApplyMiddleware<TMiddleware>();
+        return this;
+    }
+
+    /// <summary>
+    /// Applies middleware to the field using a delegate. If middleware is already set, the new middleware will be chained after the existing middleware.
+    /// </summary>
+    /// <param name="middleware">The middleware delegate to apply.</param>
+    [AllowedOn<IObjectGraphType>]
+    public virtual FieldBuilder<TSourceType, TReturnType> ApplyMiddleware(Func<FieldMiddlewareDelegate, FieldMiddlewareDelegate> middleware)
```
Shane32, line 528: Signature adjusted to match schema.FieldMiddleware.Use

src/GraphQL/Types/Fields/FieldType.cs
```diff
@@ -93,6 +94,14 @@ public Type? Type
     /// </summary>
     public ISourceStreamResolver? StreamResolver { get; set; }
 
+    /// <summary>
+    /// Gets or sets a transform function for field middleware. This function takes an IServiceProvider
+    /// and a FieldMiddlewareDelegate (the next delegate in the chain), and returns a new FieldMiddlewareDelegate
+    /// that wraps the next delegate with middleware logic.
+    /// Only applicable to fields of output graph types.
+    /// </summary>
+    public Func<IServiceProvider, FieldMiddlewareDelegate, FieldMiddlewareDelegate>? Middleware { get; set; }
```
Shane32, line 103: This is now a transform; name ok?

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
Shane32: renamed back

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
Shane32: fixed

src/GraphQL/Builders/FieldBuilder.cs
Shane32: fixed

Shane32:
@gao-artur Sorry about how giant this PR has gotten... not my first choice, but it seemed practical to design the features together.  If you want, I can break apart this PR into two separate PRs, one for field-level middleware, and one for the field context accessor.

src/GraphQL/Builders/FieldBuilder.cs
gao-artur: Btw, it's possible to get rid of the `.Resolve()` before applying `.WithService<>()`? It's not intuitive and difficult to discover. I remembered there was an option to use `.WithService`, but I couldn't figure out why it was not popping up in Intellisense. I had to look for examples in tests to understand how to do it.

Shane32: The problem is that it would require a bunch of variations of FieldBuilder like `FieldBuilder<TSource, TReturn, TService1, TService2, TService3>`.  Additionally, the resolver builders are in the MicrosoftDI package, allowing for the scoped option, separate from the field builders in the main package...

So for now we start the resolver builders with `Resolve()`, then chain together the options and stuff, ending the resolver builders with `ResolveAsync(...)` or `Resolve(...)` which brings you back to the base field builder.

We can look at changing this for v9 if you like, but I don't have a suggestion how to do so.

PR #4195, opened 2025-10-22
Add field-level middleware support

Shane32:


src/GraphQL/Builders/FieldBuilder.cs
```diff
@@ -488,4 +489,38 @@ public virtual FieldBuilder<TSourceType, TReturnType> ComplexityImpact(double im
     [AllowedOn<IObjectGraphType>]
     public virtual FieldBuilder<TSourceType, TReturnType> DependsOn<TService>()
         => this.DependsOn(typeof(TService));
+
+    /// <summary>
+    /// Applies middleware to the field. If middleware is already set, the new middleware will be chained after the existing middleware.
```
gao-artur: Can you add a note that global middleware is executed before field middleware?

src/GraphQL/Utilities/Visitors/FieldMiddlewareVisitor.cs
```diff
@@ -0,0 +1,41 @@
...
+        if (middlewareTransform != null)
+        {
+            var inner = field.Resolver ?? (field.StreamResolver == null ? NameFieldResolver.Instance : SourceFieldResolver.Instance);
+
+            // Apply the middleware transform to wrap the resolver
+            FieldMiddlewareDelegate wrappedDelegate = middlewareTransform(_serviceProvider, inner.ResolveAsync);
+            field.Resolver = new FuncFieldResolver<object>(wrappedDelegate.Invoke);
+
+            // Clear the middleware transform as it has been applied
+            field.Middleware = null;
```
gao-artur: Is that necessary? I think it may confuse during debugging, when you know you set a middleware, but apparently it's null

Shane32: No; removed.

src/GraphQL/Builders/FieldBuilder.cs
Shane32: Yes; done.

gao-artur:
Yeah, all good. I have no more comments. I don't have anything for 8.7. Hopefully will start working on federation analyzers this weekend so they will be part of 8.8.

PR #4204, opened 2025-10-30
Eliminate allocations during DataLoaderContext.GetOrAdd

Shane32:
This PR:
- Adds an overload to DataLoaderContext.GetOrAdd that provides an argument to the factory, just like ConcurrentDictionary.GetOrAdd in .NET Standard 2.1
- Updates the data loader extensions to use this.
- Changes the dictionary in DataLoaderContext to be a concurrent dictionary
- Update GraphQL.DataLoader to target .NET Standard 2.1 so the overload for concurrent dictionary's GetOrAdd exists
- Change _loaders to be lazy loaded (thread safe)
- Eliminate locks around _loaders

### DIFF
--- a/src/GraphQL.DataLoader/DataLoaderContext.cs
+++ b/src/GraphQL.DataLoader/DataLoaderContext.cs
@@ -10,7 +10,7 @@ public class DataLoaderContext
     private ConcurrentDictionary<string, IDataLoader>? _loaders;
 
     private ConcurrentDictionary<string, IDataLoader> GetLoaders()
-        => LazyInitializer.EnsureInitialized(ref _loaders!, static () => new ConcurrentDictionary<string, IDataLoader>());
+        => LazyInitializer.EnsureInitialized(ref _loaders)!;
 
     /// <summary>
     /// Add a new data loader if one does not already exist with the provided key


src/GraphQL.DataLoader/DataLoaderContext.cs
```diff
@@ -23,18 +42,31 @@ public TDataLoader GetOrAdd<TDataLoader>(string loaderKey, Func<TDataLoader> dat
...
+            throw new ArgumentNullException(nameof(loaderKey));
 
-                _loaders.Add(loaderKey, loader);
-            }
-        }
+        if (dataLoaderFactory == null)
+            throw new ArgumentNullException(nameof(dataLoaderFactory));
 
-        return (TDataLoader)loader;
+#if NETSTANDARD2_1 || NETCOREAPP2_0_OR_GREATER
```
Shane32, line 52: Reflects the supported TFMs for this overload, thereby futureproofing this code in case the project targets additional TFMs in the future.

src/GraphQL.DataLoader/DataLoaderContext.cs
```diff
@@ -1,11 +1,30 @@
...
+            // Atomically initialize _loaders if it's still null.
+            // CompareExchange returns the original value of _loaders:
+            //   - null if we won the initialization (and set _loaders = newLoaders)
+            //   - the existing dictionary if another thread initialized it first
+            // '?? newLoaders' makes 'loaders' reference whichever instance is now in _loaders.
+
+            var newLoaders = new ConcurrentDictionary<string, IDataLoader>();
+            loaders = Interlocked.CompareExchange(ref _loaders, newLoaders, null) ?? newLoaders;
+        }
+        return loaders;
```
gao-artur: Consider using [LazyInitializer.EnsureInitialized](https://learn.microsoft.com/en-us/dotnet/api/system.threading.lazyinitializer.ensureinitialized?view=net-9.0)

src/GraphQL.DataLoader/DataLoaderContext.cs
```diff
@@ -23,18 +42,31 @@ public TDataLoader GetOrAdd<TDataLoader>(string loaderKey, Func<TDataLoader> dat
...
+    /// Add a new data loader if one does not already exist with the provided key
+    /// </summary>
+    /// <typeparam name="TDataLoader">The type of <seealso cref="IDataLoader"/></typeparam>
+    /// <typeparam name="TValue">The type of the factory argument</typeparam>
+    /// <param name="loaderKey">Unique string to identify the <seealso cref="IDataLoader"/> instance</param>
+    /// <param name="dataLoaderFactory">Function to create the TDataLoader instance if it does not already exist. The factory receives a value of type TValue as a parameter.</param>
+    /// <param name="factoryArgument">The argument to pass to the factory function</param>
+    /// <returns>Returns an existing TDataLoader instance or a newly created instance if it did not exist already</returns>
+    public TDataLoader GetOrAdd<TDataLoader, TValue>(string loaderKey, Func<TValue, TDataLoader> dataLoaderFactory, TValue factoryArgument)
+        where TDataLoader : IDataLoader
```
gao-artur, line 44: While I understand that this argument order matches `ConcurrentDictionary.GetOrAdd`, I find it very inconvenient. You need to define the factory that depends on `TValue` before defining the `TValue`. The intellisence won't help you define the `Func`, until you provide the `factoryArgument`.
```c#
dataLoaderContext.GetOrAdd("key", arg => arg.{no intellicence here}...
```
By switching the `dataLoaderFactory` and `factoryArgument` you get a much better experience
```C#
dataLoaderContext.GetOrAdd("key", arg, arg => arg.{intellicence knows what is arg})
```

gao-artur: Also, the factory can also depend on `loaderKey`, so you probably need another overload with `Func<string, TValue, TDataLoader>` (or add it to this overload)

Shane32: well, in my experience, switching arguments across overloads is almost always a negative experience.  but what you say makes sense; perhaps this is an exception.  I can switch it, even though it doesn't match the MS experience.

I figured if someone really wants the loader key, they can include it in TValue using a ValueTuple.

```cs
var key = "key";
var myData = "abc";
dataLoaderContext.GetOrAdd(key, (key, myData), args => new MyDataLoader(args.key, args.myData));
```

Shane32: ok?

gao-artur: > they can include it in TValue using a ValueTuple.

They are, but it's just a repetition of the information already provided to `GetOrAdd`. Moreover, dictionary gives you the key in the factory (the argument you are discarding below)

gao-artur (approved):

src/GraphQL.DataLoader/DataLoaderContext.cs
marc81-marc: Big thanks .sean

PR #4206, opened 2025-11-04
Switch to npm from yarn

Shane32:
This PR is just so it is easier for me to manage, as I do not typically have yarn installed on my Windows PC.

I plan to try updating the deps to eliminate the GitHub security warnings, if possible.  (Even though these deps are not part of the distributed package and only affect building the documentation.)


Shane32:
Note: dependency review step failed only because it tried to post a very very large summary of changes that exceeded the maximum length.  So it would appear to be working correctly.

gao-artur (approved):

PR #4207, opened 2025-11-06
Prevent duplicate registrations

Shane32:
Previously, calling `AddDataLoader()` twice would double-register the document listener, and similar for other extension methods built on these fundamentals.  Changed methods:
- AddGraphTypeMappingProvider
- AddDocumentListener
- UseMiddleware
- AddValidationRule
- AddSchemaVisitor

Indirectly affected methods, which now may be called multiple times without ill effect:
- AddDataLoader
- AddLegacyComplexityAnalyzer
- AddComplexityAnalyzer

(maybe more)

### DIFF
--- a/src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
+++ b/src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
@@ -612,7 +612,7 @@ public static class GraphQLBuilderExtensions // TODO: split
     {
         builder.Services.TryRegister<IDocumentExecutionListener, TDocumentListener>(serviceLifetime, RegistrationCompareMode.ServiceTypeAndImplementationType);
         builder.Services.Register<TDocumentListener>(serviceLifetime);
-        builder.Services.TryRegister<IConfigureExecution, AddDocumentListenerConfiguration<TDocumentListener>>(ServiceLifetime.Singleton, RegistrationCompareMode.ServiceTypeAndImplementationType);
+        builder.ConfigureExecution<AddDocumentListenerConfiguration<TDocumentListener>>();
         return builder;
     }
 
@@ -718,7 +718,7 @@ public static class GraphQLBuilderExtensions // TODO: split
         builder.Services.TryRegister<IFieldMiddleware, TMiddleware>(serviceLifetime, RegistrationCompareMode.ServiceTypeAndImplementationType);
         builder.Services.Register<TMiddleware>(serviceLifetime);
         if (install)
-            builder.Services.TryRegister<IConfigureSchema, UseMiddlewareConfiguration<TMiddleware>>(ServiceLifetime.Singleton, RegistrationCompareMode.ServiceTypeAndImplementationType);
+            builder.ConfigureSchema<UseMiddlewareConfiguration<TMiddleware>>();
         return builder;
     }
 


src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -610,11 +610,24 @@ public static IGraphQLBuilder AddAutoClrMappings(this IGraphQLBuilder builder, b
     public static IGraphQLBuilder AddDocumentListener<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] TDocumentListener>(this IGraphQLBuilder builder, ServiceLifetime serviceLifetime = ServiceLifetime.Singleton)
         where TDocumentListener : class, IDocumentExecutionListener
     {
-        builder.Services.RegisterAsBoth<IDocumentExecutionListener, TDocumentListener>(serviceLifetime);
-        builder.ConfigureExecutionOptions(options => options.Listeners.Add(options.RequestServicesOrThrow().GetRequiredService<TDocumentListener>()));
+        builder.Services.TryRegister<IDocumentExecutionListener, TDocumentListener>(serviceLifetime, RegistrationCompareMode.ServiceTypeAndImplementationType);
+        builder.Services.Register<TDocumentListener>(serviceLifetime);
```
gao-artur, line 614: Why this one is `Register` and not `TryRegister`?

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -610,11 +610,24 @@ public static IGraphQLBuilder AddAutoClrMappings(this IGraphQLBuilder builder, b
...
+        builder.ConfigureExecution<AddDocumentListenerConfiguration<TDocumentListener>>();
```
gao-artur, line 615: What's the difference between configuring with a delegate and with a class?

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
Shane32: `TryRegister` with RegistrationCompareMode.ServiceTypeAndImplementationType will always register unless the exact combination of implementation and service type has already been registered.  It's used here because in legacy applications users may inject `IEnumerable<IDocumentExecutionListener>` and it shouldn't retrieve the same listener twice.  Now, GraphQL.NET's code doesn't actually use that at all; instead it registers a `IConfigureExecution` to install the listener which pulls the exact type from DI.

`Register` for the exact service type is always used when `IEnumerable<TService>` isn't in play (vs `TryRegister`), which effectively overwrites any previous registration.

src/GraphQL/Extensions/GraphQLBuilderExtensions.cs
Shane32: `TryRegister` with `RegistrationCompareMode.ServiceTypeAndImplementationType` can't be used on a delegate because each time the code runs it will return a different delegate so the implementation will never match.  With a class, it's the exact same type (unique per TDocumentListener) and can be matched on to prevent duplicate registrations.

Shane32:
This PR is a minor issue for me because I developed a GraphQL.NET extension which is dependent on various parts of GraphQL.NET, including the data loader.  My install extension is a generic method designed to be called multiple times for multiple types.  As it is now, if I include `AddDataLoader` within the extension method, it will be called twice if the user calls my method twice.  So now I must instruct users to install it separately, even though it is required to be installed for my extension to work.

PR #4229, opened 2025-12-12
Fix IResolveFieldContextAccessor for streams

Shane32:
This fixes an issue with IResolveFieldContextAccessor where the accessor's context is not set during the call to Subscribe, which is where typical setup code exists to start the stream, as demonstrated in the type-first code in the test using `IAsyncEnumerable` as well as code-first code with `IObservable`.

Additionally, the code was very tricky to get to work correctly since our `IResolveFieldContextAccessor` implementation clears held instances:

```cs
    /// <inheritdoc/>
    public IResolveFieldContext? Context
    {
        get => _context.Value?.Context;
        set
        {
            if (_context.Value != null)
                _context.Value.Context = null;

            if (value != null)
            {
                _context.Value = new ContextHolder { Context = value };
            }
        }
    }

    private sealed class ContextHolder
    {
        public IResolveFieldContext? Context;
    }
```

It is very important for streaming data that the accessor not be cleared.  While this could be solved simply by not setting `_accessor.Context = null;`, this could unintentionally allow the accessor's value to be accessible by the caller.  This PR's implementation handles all scenarios correctly.

### DIFF
--- a/.github/workflows/build-artifacts-code.yml
+++ b/.github/workflows/build-artifacts-code.yml
@@ -34,7 +34,7 @@ jobs:
         working-directory: docs2
         run: npm run build
       - name: Publish docs artifacts
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: Documentation
           path: docs2/public/**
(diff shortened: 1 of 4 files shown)


Shane32:
See:
- #4197

PR #4233, opened 2025-12-22
1- Move SchemaTypes logic to LegacySchemaTypes

Shane32:
This PR moves all the logic for SchemaTypes to a derived class LegacySchemaTypes in preparation for new logic to be added in a subsequent PR.  It also adds BuiltInScalars and BuiltInScalarsByName static protected members to replace the static private members that previously existed.  The logic is identical; this PR is mostly just refactoring.  Review the API tests to understand what was moved.

### DIFF
--- a/src/GraphQL.Tests/Initialization/SchemaInitializationTests.cs
+++ b/src/GraphQL.Tests/Initialization/SchemaInitializationTests.cs
@@ -280,37 +280,6 @@ public class SchemaWithArgumentsOnInputField : Schema
     }
 }
 
-// https://github.com/graphql-dotnet/graphql-dotnet/issues/2675
-public class SchemaWithNotFullSpecifiedResolvedType : Schema
-{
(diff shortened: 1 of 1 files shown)


src/GraphQL.Tests/Federation/FederationResolverTests.cs
```diff
@@ -467,33 +466,13 @@ private static T TestResolver<T>(IObjectGraphType objectGraphType, string repres
...
-
-        protected internal override Dictionary<ROM, IGraphType> Dictionary => _types;
-    }
-
-    private IFederationResolver GetResolver(IGraphType objectGraphType)
-    {
-        return GetResolvers(objectGraphType).ShouldHaveSingleItem();
-    }
-
-    private IEnumerable<IFederationResolver> GetResolvers(IGraphType objectGraphType)
```
Shane32, line 488: `GetResolver` and `GetResolvers` are not referenced in the codebase, so I removed them.

PR #4234, opened 2025-12-23
3-Add new SchemaTypes implementation

Shane32:
Depends on the following which must be merged first:
- #4233 
- #4246 

This PR rewrites the entirety of the SchemaTypes logic.  The logic is simplified, but I've added extensive documentation so the actual number of lines of code is similar.  The code is organized very cleanly with separate areas of responsibility as best I could.

One thing I am trying hard to do with these changes is to maintain binary compatibility for `Schema.AllTypes` which returns `SchemaTypes` and provides and indexer, `Count`, and `IEnumerable`.  This allows any 3rd party tooling for GraphQL.NET to remain compatible assuming it needs basic access to `Schema.AllTypes`.

Given that, the most significant question with this PR is whether the implementation should be put into a derived class or made part of SchemaTypes.  I'm 50-50 on this.  I've already moved the old logic to `LegacySchemaTypes` in #4233 which is available in case there is found to be any lacking functionality in the new design.  To be clear, the options are:

1. Create a dervied class (e.g. `NewSchemaTypes`) which derives from `SchemaTypes` with the needed functionality.  The main benefit is that we can make methods `protected virtual` if we want to allow users to override functionality.  The biggest negative is that I cannot think of an appropriate name for the class.

2. Same, but change the base type to `SchemaTypesBase`.  This breaks binary compatibility for code relying on `Schema.AllTypes`, but perhaps that's okay.  This PR represents this option.  

3. Add a public constructor to `SchemaTypes` which uses internal/private code to initialize the schema.  The logic is in a `private readonly ref struct NewSchemaType` which performs the initialization.  This is clear for users; `new SchemaTypes(...)` is still used, as before, to initialize a schema.  The biggest negative is that there are no `protected virtual` methods that can be overridden to alter initialization behavior.  Specifically `BuildGraphQLType` was protected previously and is not available now.  However, the addition of `IGraphTypeMappingProvider` (available since 5.0 via #2999) and other bug fixes have all but removed the need to override `SchemaTypes`.

### DIFF
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -1,0 +1,890 @@
+using GraphQL.Conversion;
+using GraphQL.Introspection;
+using GraphQL.Utilities;
+using GraphQLParser;
+
+namespace GraphQL.Types;
+
+/*
+ * ================================================================================================
+ * NewSchemaTypes - GraphQL Schema Type Collection and Initialization
+ * ================================================================================================
+ *
+ * This class manages the complete lifecycle of type discovery, registration, and initialization
+ * for a GraphQL schema. It implements a three-phase approach to ensure all types are properly
+ * discovered, processed, and finalized before the schema becomes operational.
+ *
+ * ================================================================================================
+ * THREE-PHASE INITIALIZATION PROCESS
+ * ================================================================================================
+ *
+ * PHASE 1: DISCOVERY
+ * ------------------
+ * The discovery phase collects known/root types from various sources WITHOUT processing them.
+ * This includes introspection types, manually-added types, root operation types, and additional
+ * type references. These types are registered but not processed, deferring all recursive
+ * field/argument processing until Phase 2.
+ *
+ * Types are discovered in the following order (which establishes registration priority):
+ *
+ *   1. Introspection Types (__Schema, __Type, __Field, __InputValue, __EnumValue, __Directive,
+ *      __DirectiveLocation, __TypeKind, and optionally __AppliedDirective, __DirectiveArgument)
+ *      - These are registered first to ensure they're available for schema introspection
+ *      - They use specific feature flags to determine which types to include
+ *
+ *   2. Manually-Added Type Instances (Schema.AdditionalTypeInstances)
+ *      - Pre-instantiated types added via Schema.RegisterType(instance)
+ *      - These take precedence over types resolved from the DI container
+ *      - Specifically useful for overriding built-in scalars
+ *
+ *   3. Root Operation Types (Query, Mutation, Subscription)
+ *      - The Query type is required; Mutation and Subscription are optional
+ *      - These define the entry points for GraphQL operations
+ *
+ *   4. Additional Type References (Schema.AdditionalTypes)
+ *      - Types added via Schema.RegisterType<T>()
+ *      - These are CLR types that will be resolved to GraphType instances
+ *      - Useful for ensuring types are included even if not directly referenced
+ *
+ * During discovery, types are added to the Dictionary (keyed by GraphQL type name) and to the
+ * _typeDictionary (keyed by CLR type). The skipProcessing flag is set to true, deferring all
+ * recursive field/argument processing until Phase 2.
+ *
+ * PHASE 2: PROCESSING
+ * -------------------
+ * The processing phase walks through all discovered types and resolves their dependencies,
+ * discovering additional types as needed. Before processing each type, the _onBeforeInitialize
+ * hook is called (if provided), allowing for last-minute modifications to the type before it's
+ * locked in. This phase then handles:
+ *
+ *   - Field Type Resolution: Converts Type references to ResolvedType instances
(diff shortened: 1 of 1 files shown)


src/GraphQL.Tests/Types/SchemaTests.cs
```diff
@@ -240,6 +241,50 @@ public void OnBeforeInitializeType_cannot_access_AllTypes()
         Should.Throw<InvalidOperationException>(() => schema.Initialize())
             .Message.ShouldBe("Cannot access AllTypes while schema types are being created. AllTypes is not available during OnBeforeInitializeType execution.");
     }
+
+    [Fact]
+    public void does_not_double_initialize_types()
```
Shane32, line 246: Passes now; did not before.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -2974,12 +2978,13 @@ namespace GraphQL.Types
...
         public void Register(params GraphQL.Types.Directive[] directives) { }
     }
-    public abstract class SchemaTypes : System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphType>, System.Collections.IEnumerable
+    public class SchemaTypes : System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphType>, System.Collections.IEnumerable
     {
         protected static readonly System.Collections.ObjectModel.ReadOnlyDictionary<System.Type, GraphQL.Types.ScalarGraphType> BuiltInScalars;
         protected static readonly System.Collections.ObjectModel.ReadOnlyDictionary<string, GraphQL.Types.ScalarGraphType> BuiltInScalarsByName;
         protected SchemaTypes() { }
         protected SchemaTypes(System.Collections.Generic.Dictionary<GraphQLParser.ROM, GraphQL.Types.IGraphType> dictionary) { }
+        public SchemaTypes(GraphQL.Types.ISchema schema, System.IServiceProvider serviceProvider, System.Collections.Generic.IEnumerable<GraphQL.Types.IGraphTypeMappingProvider>? graphTypeMappings = null, System.Action<GraphQL.Types.IGraphType>? onBeforeInitialize = null) { }
```
Shane32: This constructor is restored exactly as in v8.  (Other v8 constructors and methods have not been restored.)

src/GraphQL.Tests/Types/SchemaTests.cs
```diff
@@ -125,10 +126,10 @@ public void handles_cycle_field_type()
     }
 
     [Fact]
-    public void handles_stackoverflow_exception_for_cycle_field_type()
+    public void handles_cycle_field_type_2()
     {
         var schema = new ACyclingDerivingSchema(new FuncServiceProvider(t => t == typeof(AbstractGraphType) ? new ConcreteGraphType() : null));
-        Should.Throw<InvalidOperationException>(() => schema.AllTypes["abcd"]);
+        schema.Initialize();
```
Shane32, line 132: It is unclear why there was cycle checks during initialization previously, or why this test failed.  The new code handles cycles within types properly and this schema can be initialized.

src/GraphQL.Tests/Types/Collections/SchemaTypesTests.cs
```diff
@@ -222,6 +221,40 @@ public async Task can_override_built_in_types()
...
+        schema.RegisterType(new MyBooleanGraphType());
+        var query = new ObjectGraphType();
+        query.Field<IntGraphType>("sample")
+          .Argument<BooleanGraphType>("argNewBehavior");
+        schema.Query = query;
+        schema.Initialize();
+    }
+
+    [Fact]
+    public void cant_have_mixed_scalars_then_replace_them()
```
Shane32, line 237: the new tests in this file support the custom-scalars.md documentation

gao-artur:
I believe the binary compatibility is a nice-to-have, but not a necessity when bumping the major version. Users should not expect a 3rd-party library to continue working without recompilation, especially when we have many other breaking changes anyway. The source compatibility is more important. So, I vote for the second option (base class) and call the new implementation with the old name `SchemaTypes`.
I went over the PR, and it looks good: very clean and readable. However, there is a lot of code, so I couldn't stay focused while reading it. Maybe I missed something, but overall it looks good.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
Shane32: n/a

Shane32:
> I believe the binary compatibility is a nice-to-have, but not a necessity when bumping the major version. Users should not expect a 3rd-party library to continue working without recompilation, especially when we have many other breaking changes anyway. The source compatibility is more important. So, I vote for the second option (base class) and call the new implementation with the old name `SchemaTypes`. I went over the PR, and it looks good: very clean and readable. However, there is a lot of code, so I couldn't stay focused while reading it. Maybe I missed something, but overall it looks good.

`SchemaTypesBase` is now the base type, with `SchemaTypes` and `LegacySchemaTypes` as derived classes

gao-artur:
Yeah, I got what you are doing. Looks good!

gao-artur (approved):

PR #4244, opened 2025-12-28
Change base type to SchemaTypesBase

Shane32:
Builds on #4234 

Now organized as follows:
- SchemaTypesBase - returned by ISchema.AllTypes
- SchemaTypes - new implementation (currently actual logic is 100% private)
- LegacySchemaTypes - old implementation


src/GraphQL/Types/Collections/NewSchemaTypes.cs
```diff
@@ -185,14 +185,35 @@ namespace GraphQL.Types;
...
+    /// <param name="onBeforeInitialize">Pre-initialization hook called before each type is initialized.</param>
+    public SchemaTypes(
+        ISchema schema,
+        IServiceProvider serviceProvider,
+        IEnumerable<IGraphTypeMappingProvider>? graphTypeMappings = null,
+        Action<IGraphType>? onBeforeInitialize = null) : base()
+    {
+        new SchemaTypesInitializer(this, schema, serviceProvider, graphTypeMappings, onBeforeInitialize)
+            .Initialize();
+    }
```
Shane32, line 209: With this design we can collapse SchemaTypesInitializer functionality into SchemaTypes, giving the option to add protected members to allow overriding functionality.  However, here the entire initialization logic is private, so the class has been sealed and I've left the initialization logic as a readonly ref struct for optimization purposes

gao-artur (approved):

PR #4250, opened 2026-01-03
Honor Schema.Comparer in SchemaExporter.ToAST()

Shane32:
## Summary

Makes `SchemaExporter.ToAST()` respect the `Schema.Comparer` property when exporting schemas to AST format. This ensures consistent sorting behavior between introspection queries and SDL exports.

## Changes

- Modified `SchemaExporter` to apply sorting from `Schema.Comparer` throughout the export process:
  - Sorts directives and types in `Export()`
  - Sorts fields in `ExportInputObjectTypeDefinition()`, `ExportInterfaceTypeDefinition()`, and `ExportObjectTypeDefinition()`
  - Sorts enum values in `ExportEnumTypeDefinition()`
  - Sorts arguments in `ExportArgumentsDefinition()` (signature updated to accept `IFieldType?` parameter)

- Added test `Export_Should_Honor_Schema_Comparer()` with approved output validation demonstrating alphabetical sorting when `AlphabeticalSchemaComparer` is configured

- Updated migration documentation (migration9.md) noting this as a breaking change

## Breaking Change

When `Schema.Comparer` is set (e.g., to `AlphabeticalSchemaComparer`), the exported AST will now be sorted according to the comparer's rules. Previously, the export maintained natural ordering regardless of the comparer setting.

**Impact:**
- Direct calls to `schema.ToAST()` will now produce sorted output when a comparer is configured
- Federation SDL requests (`_service { sdl }`) will return sorted schemas when a custom comparer is set
- Schemas using `DefaultSchemaComparer` (the default) are unaffected

## Motivation

This change provides consistency between introspection and SDL exports. When users configure a schema comparer (typically for alphabetical sorting), they expect that sorting to apply to all schema representations, not just introspection queries.

### DIFF
--- a/docs2/site/docs/analyzers/gql021.md
+++ b/docs2/site/docs/analyzers/gql021.md
@@ -1,0 +1,77 @@
+# GQL021: Nullable reference type argument should specify nullable parameter
+
+|                        | Value   |
+| ---------------------- | ------- |
+| **Rule ID**            | GQL021  |
+| **Category**           | Usage   |
+| **Default severity**   | Warning |
+| **Enabled by default** | Yes     |
+| **Code fix provided**  | Yes     |
+| **Introduced in**      | v9.0    |
+
+## Cause
+
+This rule is triggered when a nullable reference type is used as a type argument
+for the `Argument` method without explicitly specifying the `nullable` parameter.
+
+## Rule description
+
+When using nullable reference types (e.g., `string?`) with the `Argument<T>`
+method, you should explicitly specify `nullable: true` to make the nullability
+intent clear and consistent with the GraphQL schema.
+
+## How to fix violations
+
+Add the `nullable: true` parameter to the `Argument` method call.
+
+## Example of a violation
+
+```c#
+public class MyGraphType : ObjectGraphType
+{
+    public MyGraphType()
+    {
+        Field<StringGraphType>("hello")
+            .Argument<string?>("name")
+            .Resolve(context => $"Hello {context.GetArgument<string?>("name")}");
+    }
+}
+```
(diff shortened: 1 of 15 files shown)


Shane32:
@gao-artur Technically this does not break the API, and could be integrated to v8.  However if users are using code generators that rely on the order of field arguments from the printed sdl, then this would be a breaking change.  As such, it's targeting v9 here.

src/GraphQL.Tests/Utilities/SchemaExporterTests.Export_Should_Honor_Schema_Comparer.approved.txt
```diff
@@ -0,0 +1,30 @@
+schema {
+  query: Type1
+  mutation: Type3
+}
+
+directive @zdir1 on FIELD | OBJECT
+
+directive @zdir2 on INLINE_FRAGMENT | OBJECT
+
+directive @zdir3(arg3: [ID], arg1: Float, arg2: Int!) on FRAGMENT_SPREAD | FIELD | INLINE_FRAGMENT | OBJECT
```
gao-artur, line 10: Shouldn't it be `arg1, arg2, arg3`?

Shane32: `ISchemaComparer` doesn't currently support sorting directive arguments.  We can do a subsequent PR for that capability (needs modification to `ISchemaComparer.cs` and `__Directive.cs` and so on).

gao-artur (approved):

Issue #4176, opened 2025-09-04
Add ability to control ISchemaNodeVisitor execution order


EvilVir:
## Summary

Currently execution of schema visitors is nondeterministic, in theory it goes over the underlying list so should be in order of registration but that's not given.

### Basic example

Add optional `Priority` property or attribute to `ISchemaNodeVisitor` and `OrderBy` that property/attribute when executing registered visitors during schema initialization stage.

### Motivation

In our project we have couple of schema visitors, some of them are adding new fields while other are enhancing all the fields on the schema. If enhancement visitor runs before fields adding one, the resulting schema is "broken". In some extreme scenarios it might even result in being unable to initialize schema altogether.

Shane32:
It's true that currently the order of not guaranteed for future versions, although for the current version it is deterministic and follows this exact order:

1. The schema is initiailized via the protected `CreateSchemaTypes()` method.
2. Field middleware is applied to each field resolver
3. Each visitor instance is executed in the order they were registered via `RegisterVisitor(ISchemaNodeVisitor visitor)` calls
4. Each visitor type is pulled from DI and executed in the order they were registered via `RegisterVisitor(Type type)` and `RegisterVisitor<TVisitor>()` calls
5. Built-in visitors are run

Note that the `AddSchemaVisitor<TVisitor>()` and `AddSchemaVisitor(ISchemaNodeVisitor visitor)` builder methods currently map to corresponding `RegisterVisitor` methods, which execute within a `ConfigureSchema` block; those execute before the schema constructor starts in the order they were registered.  This order depends on the DI engine preserving the order of registrations.  This is guaranteed by Microsoft's DI engine; see https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection which states:

> Services appear in the order they were registered when resolved via `IEnumerable<{SERVICE}>`.

Importantly, this means that any changes to the schema within an `ISchemaNodeVisitor` occurs after the schema is initialized.  So you can't, for example, add a field using only the graph type or CLR type because `ResolvedType` wouldn't be set on the field.  Nor can you add a new graph type as it would not be properly registered within `SchemaTypes`.

Example:

```cs
services.AddGraphQL(b => b
    .AddSchema<MySchema>()                          // Constructor executes after other registrations here
    .AddSchemaVisitor<MySchema4>()                  // Executes 4th overall

    .ConfigureSchema((sp, schema) =>
    {
[...]

PR #4251, opened 2026-01-03
Add SortOrder property to IConfigureSchema interface

Shane32:
## Summary
This PR adds a `SortOrder` property to the `IConfigureSchema` interface, bringing it into parity with the existing `IConfigureExecution` interface. This allows schema configurations to be executed in a predictable order based on their priority. Allows for #4176 in a future PR.

## Changes
- **Added `SortOrder` property to `IConfigureSchema` interface** - Determines the execution order of schema configurations (lower values execute first)
- **Updated `Schema` constructor** - Now sorts configurations by `SortOrder` before executing them using a stable sort
- **Implemented `SortOrder` in all implementations** - All built-in implementations return the default value of `100` (`SORT_ORDER_OPTIONS`)
- **Updated migration documentation** - Added section 10 to `migration9.md` documenting this breaking change

## Breaking Change
This is a **breaking change** for users with custom `IConfigureSchema` implementations. The interface now requires a `SortOrder` property to be implemented.

**Impact:** Typically no changes are required to user code unless you have a custom implementation of `IConfigureSchema`. If you do, you'll need to add the `SortOrder` property to your class. For most cases, returning `100` (the default value) is appropriate.

## Implementation Details
- Default sort order: `100` (matching `SORT_ORDER_OPTIONS` constant)
- Sorting behavior: Stable sort in ascending order (lowest executes first)
- Configurations with equal sort order execute in registration order (assuming DI provider preserves registration order)

## Files Changed
- `src/GraphQL/DI/IConfigureSchema.cs` - Added `SortOrder` property to interface and internal implementation
- `src/GraphQL/Types/Schema.cs` - Updated constructor to sort configurations before execution
- `src/GraphQL.MicrosoftDI/ValidateServicesSchemaConfigurator.cs` - Implemented `SortOrder` property
- `src/GraphQL/Extensions/GraphQLBuilderExtensions.cs` - Implemented `SortOrder` in 3 internal classes
- `src/GraphQL.Tests/DI/GraphQLBuilderExtensionTests.cs` - Updated test class implementation
- `docs2/site/docs/migrations/migration9.md` - Added migration documentation

## Testing
- Build completed successfully with no new errors
- All existing implementations updated to include the new property

### DIFF
--- a/src/GraphQL.MicrosoftDI/ValidateServicesSchemaConfigurator.cs
+++ b/src/GraphQL.MicrosoftDI/ValidateServicesSchemaConfigurator.cs
@@ -25,5 +25,5 @@ internal sealed class ValidateServicesSchemaConfigurator : IConfigureSchema
     }
 
     /// <inheritdoc/>
-    public float SortOrder => 100;
+    public float SortOrder => GraphQLBuilderExtensions.SORT_ORDER_OPTIONS;
 }
--- a/docs2/site/docs/migrations/migration9.md
+++ b/docs2/site/docs/migrations/migration9.md
@@ -234,6 +234,6 @@ public class CustomSchemaComparer : AlphabeticalSchemaComparer
 
 ### 13. `IConfigureSchema` interface now includes `SortOrder` property
 
-The `IConfigureSchema` interface now includes a `SortOrder` property, similar to the existing `IConfigureExecution` interface. This property determines the order in which schema configurations are executed, with lower values executing first. The default sort order for all built-in implementations is `100`.
+The `IConfigureSchema` interface now includes a `SortOrder` property, similar to the existing `IConfigureExecution` interface. This property determines the order in which schema configurations are executed, with lower values executing first. The default sort order for all built-in implementations is `100` (`GraphQLBuilderExtensions.SORT_ORDER_OPTIONS`).
 
 Typically no changes are required to user code unless you have a custom implementation of `IConfigureSchema`. If you do have a custom implementation, you will need to add the `SortOrder` property to your class. For most cases, returning `100` (the default value) is appropriate.
--- a/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
@@ -324,6 +324,8 @@ namespace GraphQL
     }
     public static class GraphQLBuilderExtensions
     {
+        public const float SORT_ORDER_CONFIGURATION = 200F;
+        public const float SORT_ORDER_OPTIONS = 100F;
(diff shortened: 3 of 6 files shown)


src/GraphQL.MicrosoftDI/ValidateServicesSchemaConfigurator.cs
```diff
@@ -23,4 +23,7 @@ public void Configure(ISchema schema, IServiceProvider serviceProvider)
         else
             throw new InvalidOperationException("Could not create a service validator. Ensure that the dependency injection framework supports IServiceProviderIsService.");
     }
+
+    /// <inheritdoc/>
+    public float SortOrder => 100;
```
gao-artur: Would it be appropriate to make `GraphQLBuilderExtensions.SORT_ORDER_OPTIONS` public to allow reusing it here? It may also be useful for users who implement their own `IConfigureSchema`'s if they want to reuse the default order or use relative expressions to run their configurators before/after the built-in configurators
```c#
// run before built-in configurators
public float SortOrder => GraphQLBuilderExtensions.SORT_ORDER_OPTIONS - 1;

// run after built-in configurators
public float SortOrder => GraphQLBuilderExtensions.SORT_ORDER_OPTIONS + 1;
```

src/GraphQL/Types/Schema.cs
```diff
@@ -131,7 +131,9 @@ public Schema(IServiceProvider services, IEnumerable<IConfigureSchema> configura
         Directives = new SchemaDirectives();
         Directives.Register(Directives.Include, Directives.Skip, Directives.Deprecated);
 
-        foreach (var configuration in configurations ?? Array.Empty<IConfigureSchema>())
+        // OrderBy here performs a stable sort; that is, if the sort order of two elements are equal,
+        // the order of the elements are preserved.
+        foreach (var configuration in (configurations ?? Array.Empty<IConfigureSchema>()).OrderBy(x => x.SortOrder))
```
gao-artur:
```suggestion
        foreach (var configuration in configurations?.OrderBy(x => x.SortOrder) ?? Enumerable.Empty<IConfigureSchema>())
```
To prevent the allocation of `OrderedEnumerable` if `configurations` is null

src/GraphQL.MicrosoftDI/ValidateServicesSchemaConfigurator.cs
Shane32: Okay but where do we expose it?  static extension class seems hardly the place, yet I don't know of a better one.  just change to public and leave where it is?

gao-artur: No idea... Maybe `GraphQLBuilderBase`? Not better than `GraphQLBuilderExtensions` though.

Shane32: Left it in its current place; changed to public.

PR #4252, opened 2026-01-04
Add automatic nullable value type detection for FieldBuilder.Argument<T>

Shane32:
### Summary
This PR enhances the `FieldBuilder.Argument<T>` method to automatically detect nullable value types, making it easier to work with nullable value types like `int?`, `DateTime?`, etc. without explicitly specifying the `nullable` parameter.

### Changes
- Changed the `nullable` parameter from `bool` (default `false`) to `bool?` (default `null`)
- When `nullable` is `null` (default), the method automatically detects if `T` is a nullable value type:
  - Nullable value types (`int?`, `DateTime?`, etc.) → automatically nullable fields
  - Non-nullable value types (`int`, `DateTime`, etc.) → non-null fields
  - Reference types (`string`, etc.) → non-null by default (must explicitly set `nullable: true` for optional)
- Updated both `Argument<T>` method overloads for consistency
- Added comprehensive migration documentation

### Example Usage
```csharp
// Before - had to explicitly specify nullable: true for nullable value types
Field<StringGraphType>("myField")
    .Argument<int?>("nullableArg", nullable: true)
    .Argument<int>("requiredArg", nullable: false);

// After - automatic detection when nullable parameter is omitted
Field<StringGraphType>("myField")
    .Argument<int?>("nullableArg")     // Automatically nullable
    .Argument<int>("requiredArg");     // Automatically non-null

### DIFF
--- a/docs2/site/docs/migrations/migration9.md
+++ b/docs2/site/docs/migrations/migration9.md
@@ -40,6 +40,24 @@ schema.Features.DeprecationOfInputValues = false;
 
 Note that these properties must be set before schema initialization.
 
+### 2. Automatic Nullable Value Type Detection in Field Arguments
+
+The `Argument<T>` method on field builders now supports automatic nullable value type detection. When the `nullable` parameter is not specified (defaults to `null`), nullable value types like `int?`, `DateTime?`, etc. will automatically be treated as nullable fields in the GraphQL schema.
+
+```csharp
+// Before v9 - explicit nullable parameter required
+Field<StringGraphType>("myField")
+    .Argument<int?>("nullableArg", nullable: true)  // Had to explicitly set nullable: true
+    .Argument<int>("requiredArg", nullable: false); // Explicitly non-null
+
+// v9 - automatic detection when nullable parameter is omitted
+Field<StringGraphType>("myField")
+    .Argument<int?>("nullableArg")     // Automatically nullable (nullable value type)
+    .Argument<int>("requiredArg");     // Automatically non-null (non-nullable value type)
+```
+
+This feature makes it easier to work with nullable value types without having to explicitly specify the `nullable` parameter. Note that reference types like `string` still default to non-null and require explicitly setting `nullable: true` to make them optional. You can always explicitly set `nullable: true` or `nullable: false` to override the automatic behavior if needed.
+
 ## Breaking Changes
 
 ### 1. Removal of Obsolete Members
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by Shane32** [comment]
@gao-artur Of note, this change cannot break schemas, as the existing behavior of `.Argument<int?>("test")` is to throw an exception.  See TypeExtensions.cs:212 :

```cs
        if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>))
        {
            type = type.GetGenericArguments()[0];
            if (!isNullable)
            {
                throw new ArgumentOutOfRangeException(nameof(isNullable),
                    $"Explicitly nullable type: Nullable<{type.Name}> cannot be coerced to a non nullable GraphQL type.");
            }
        }
```

docs2/site/docs/migrations/migration9.md
```diff
@@ -40,6 +40,24 @@ schema.Features.DeprecationOfInputValues = false;
...
+    .Argument<int?>("nullableArg", nullable: true)  // Had to explicitly set nullable: true
+    .Argument<int>("requiredArg", nullable: false); // Explicitly non-null
+
+// v9 - automatic detection when nullable parameter is omitted
+Field<StringGraphType>("myField")
+    .Argument<int?>("nullableArg")     // Automatically nullable (nullable value type)
+    .Argument<int>("requiredArg");     // Automatically non-null (non-nullable value type)
+```
+
+This feature makes it easier to work with nullable value types without having to explicitly specify the `nullable` parameter. Note that reference types like `string` still default to non-null and require explicitly setting `nullable: true` to make them optional. You can always explicitly set `nullable: true` or `nullable: false` to override the automatic behavior if needed.
```
gao-artur, line 59: Can the `InferFieldNullabilityFromNRTAnnotations` be extended to reference-type field arguments?

gao-artur (approved):

docs2/site/docs/migrations/migration9.md
Shane32: Impossible, at least impossible without help from source generators or the like. The compiled code doesn’t contain NRT metadata for method calls. Types and members yes but code no. Value types are actually different types when nullable so the type can be read. 

So when we have `Firld(x => x.Name)` which does read nullability, it’s reading it from x.Name versus the implied generic type T passed to Field. That’s why it also can’t read nullability of a formula like `Field(“Name”, x => x.Name?.ToLower())` which returns a reference type. 

Ultimately arguments don’t point to properties to read NRT data from.

gao-artur: I see. Then an analyzer with an automatic code fix is really an optimal solution. Unless we want to use the interceptor feature to replace the
```C#
.Argument<string?>("name")
```
With
```C#
.Argument<string?>("name", nullable: true)
```
behind the scenes.

Shane32: Yeah I’d like to do an interceptor to fully align behavior with Field. But I forgot the name and ChatGPT insisted it wasn’t possible, and I couldn’t find the docs on it.  (lol)

Shane32: Ugh, requires opt-in via `<InterceptorsNamespaces>`

gao-artur: Not anymore starting SDK 9.0.2
https://github.com/dotnet/roslyn/pull/76312
So we can use multitargeting to enable it at least to some users
https://andrewlock.net/supporting-multiple-sdk-versions-in-analyzers-and-source-generators/

PR #4253, opened 2026-01-04
Add GQL021 analyzer for nullable reference type arguments

Shane32:
### Summary
This PR adds a new analyzer (GQL021) that warns developers when they use nullable reference types in `Argument<T>` calls without explicitly specifying `nullable: true`. This complements the automatic nullable value type detection feature added in the previous PR.

### Changes
- **New Analyzer (GQL021)**: `NullableReferenceTypeArgumentAnalyzer`
  - Detects when `Argument<string?>` or other nullable reference types are used without explicit nullable parameter
  - Warns when `nullable: null` or `nullable: default` is passed (which defaults to non-null behavior)
  - Does NOT warn when nullable is explicitly set to `true` or `false`
  - Does NOT warn when nullable is a variable (can't determine value at compile time)
  
- **Code Fix Provider**: `NullableReferenceTypeArgumentCodeFixProvider`
  - Automatically adds `nullable: true` to fix the warning
  - Replaces existing `null`/`default` values with `true`
  - Preserves code structure (named vs positional arguments)

- **Comprehensive Test Suite**: 15 tests covering all scenarios
  - Tests for detection of nullable reference types
  - Tests for proper handling of variables and expressions
  - Tests for non-nullable types (no false positives)
  - Tests for generic type parameters
  - All tests passing ✅

### Why This Analyzer?
With the new automatic nullable value type detection (where `int?` automatically becomes nullable), there's a potential confusion point: nullable reference types like `string?` still default to non-null unless explicitly specified. This analyzer helps developers catch this mistake early.

### Example
```csharp
// ❌ Warning GQL021: Nullable reference type 'string?' should explicitly specify nullable: true
Field<StringGraphType>("test")
    .Argument<string?>("arg");

// ✅ Fixed
Field<StringGraphType>("test")
    .Argument<string?>("arg", nullable: true);

### DIFF
--- a/docs2/site/docs/analyzers/gql021.md
+++ b/docs2/site/docs/analyzers/gql021.md
@@ -65,8 +65,7 @@ your source file to disable and then re-enable the rule.
 #pragma warning restore GQL021
 ```
 
-To disable the rule for a file, folder, or project, set its severity to `none`
-in the
+To disable the rule for a file, folder, or project, set its severity to `none` in the
 [configuration file](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files).
 
 ```ini

### CODE REVIEW

**Comment by Shane32** [comment]
This PR depends on, which must be merged first:
- #4252

**Comment by Shane32** [comment]
@gao-artur I don't know much about analyzers, so AI wrote this whole PR with very little guidance from me.  Based on the tests, it seems to work, but it would be good if you reviewed it.  Feel free to make further improvements directly to the branch.

File: src/GraphQL.Analyzers/Helpers/HelpLinks.cs [hunk 1]
```diff
@@ -22,4 +22,5 @@ public static class HelpLinks
     public const string PARSER_METHOD_MUST_BE_VALID = $"{DOCS_URL}/gql018";
     public const string VALIDATOR_METHOD_MUST_BE_VALID = $"{DOCS_URL}/gql019";
     public const string VALIDATE_ARGUMENTS_METHOD_MUST_BE_VALID = $"{DOCS_URL}/gql020";
+    public const string NULLABLE_REFERENCE_TYPE_ARGUMENT_SHOULD_SPECIFY_NULLABLE = $"{DOCS_URL}/gql021";
```
gao-artur, line 25: Need to add a documentation file `gql021.md`

gao-artur:
That's it! You don't need me anymore 😅
Looks good!
Depending on the answer to [this question](https://github.com/graphql-dotnet/graphql-dotnet/pull/4252#discussion_r2659491858), we may need to add a recommendation to disable this rule when `InferFieldNullabilityFromNRTAnnotations = true`

Shane32:
The change to `[Theory]` was 💯 .  I should have told AI to do that - it never does, and bugs me when it creates 100 tests that are all nearly the same.

docs2/site/docs/analyzers/gql021.md
```diff
@@ -0,0 +1,78 @@
...
+your source file to disable and then re-enable the rule.
+
+```csharp
+#pragma warning disable GQL021
+// The code that's violating the rule is on this line.
+#pragma warning restore GQL021
+```
+
+To disable the rule for a file, folder, or project, set its severity to `none`
+in the
```
Shane32:
```suggestion
To disable the rule for a file, folder, or project, set its severity to `none` in the
```

PR #4258, opened 2026-01-06
Fix NativeAOT compatibility for array serialization in ExecutionResultJsonConverter

Shane32:
## Summary
This PR improves NativeAOT compilation compatibility by handling cases where value list types (e.g., `List<int>`) may not be registered with the `JsonSerializationContext`.

## Changes
Modified [`ExecutionResultJsonConverter.WriteExecutionNode()`](src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs:76) to handle serialization of array execution nodes more robustly:

1. **Added null check**: Explicitly handle null `serializedResult` values by writing a null JSON value

2. **Special handling for byte arrays**: Detect `byte[]` types and manually serialize them as JSON arrays of numbers to avoid writing as a base64 string

3. **Framework-specific serialization logic**:
   - **Pre-.NET 8**: Use explicit `JsonSerializer.Serialize<object?>()` to eliminate boxing when iterating over `IEnumerable`
   - **.NET 8+**: Attempt to retrieve type info from `JsonSerializerOptions` using `TryGetTypeInfo()`. If type info is not available (common for arrays of primitive types in NativeAOT scenarios), fall back to manual iteration and serialization of individual items

## Why This Matters
In NativeAOT compilation, not all generic types are automatically registered with the JSON serialization context. This change ensures that array serialization works correctly even when specific list types haven't been pre-registered, preventing runtime serialization failures in NativeAOT-compiled applications.

### DIFF
--- a/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
+++ b/src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
@@ -1,6 +1,3 @@
-#if NET8_0_OR_GREATER
-using System.Diagnostics;
-#endif
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using GraphQL.Execution;


src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
```diff
@@ -76,7 +76,43 @@ private static void WriteExecutionNode(Utf8JsonWriter writer, ExecutionNode node
...
+                    {
+                        writer.WriteNumberValue(b);
+                    }
+                    writer.WriteEndArray();
+                }
+                else
+                {
+#if !NET8_0_OR_GREATER
+                    // Specify object? to eliminate boxing when iterating over IEnumerable
+                    JsonSerializer.Serialize<object?>(writer, serializedResult, options);
```
Shane32, line 98: Previous code did not specify `<object?>` and would pick up `<IEnumerable?>` automatically.  This means that STJ would iterate over the `IEnumerable`, causing boxing for each value before written.  With the new solution, STJ will (hopefully) compile a customized serializer at runtime which does not require boxing.  However, it will detect `byte[]` as something that should be converted to base64 (see STJ docs).  So we have to add the preceding code to manually handle `byte[]` cases.

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
```diff
@@ -76,7 +76,43 @@ private static void WriteExecutionNode(Utf8JsonWriter writer, ExecutionNode node
             var items = arrayExecutionNode.Items;
             if (items == null)
             {
-                JsonSerializer.Serialize(writer, arrayExecutionNode.SerializedResult, options);
+                var serializedResult = arrayExecutionNode.SerializedResult;
+                if (serializedResult == null)
+                {
+                    writer.WriteNullValue();
+                }
+                else if (serializedResult.GetType() == typeof(byte[]))
```
Shane32, line 84: Cannot use `is byte[]` here because `sbyte[]` matches on `is byte[]`.  So need to use `GetType()` for exact type matching.  Also, no need to match on `List<byte>` or other enumerable `byte` classes; STJ does not convert them to base64.

Shane32: We do have test cases for `byte[]` and `sbyte[]`, I think, in the codebase already.

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
```diff
@@ -76,7 +76,43 @@ private static void WriteExecutionNode(Utf8JsonWriter writer, ExecutionNode node
...
+                else
+                {
+#if !NET8_0_OR_GREATER
+                    // Specify object? to eliminate boxing when iterating over IEnumerable
+                    JsonSerializer.Serialize<object?>(writer, serializedResult, options);
+#else
+                    if (options.TryGetTypeInfo(serializedResult.GetType(), out var typeInfo))
+                    {
+                        JsonSerializer.Serialize(writer, serializedResult, typeInfo);
+                    }
```
Shane32, line 103: This will always match when using reflection-based serialization and in that case is functionally the same as the pre-.NET 8 logic.

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
```diff
@@ -76,7 +76,43 @@ private static void WriteExecutionNode(Utf8JsonWriter writer, ExecutionNode node
...
+                    else
+                    {
+                        // Arrays of primitive types may not have type info available, so fall back to manual serialization
+                        writer.WriteStartArray();
+                        foreach (var item in serializedResult)
+                        {
+                            JsonSerializer.Serialize(writer, item, options);
+                        }
+                        writer.WriteEndArray();
+                    }
```
Shane32, line 113: So this branch is only used for NativeAOT deployments with a source-generated JsonSerializationContext (and where the enumerable type is one that cannot be natively handled).  This will incur boxing for each element for value types.

gao-artur (approved):

src/GraphQL.SystemTextJson/ExecutionResultJsonConverter.cs
Shane32: Users will be able to create a custom JsonSerializationContext and add special types if they want to optimize list types like `List<short>` which we won't include native handling for.

PR #4272, opened 2026-01-13
Refactor type mapping system with dedicated provider classes for AOT compatibility

Shane32:
## Summary

Refactors the type mapping system by introducing dedicated `IGraphTypeMappingProvider` implementations for built-in scalars and enums, improving Native AOT compatibility and code organization.

### New Classes
- **`BuiltInScalarMappingProvider`**: Maps CLR types to built-in scalar graph types (e.g., `int` → `IntGraphType`, `string` → `StringGraphType`)
- **`EnumGraphTypeMappingProvider`**: Maps CLR enum types to `EnumerationGraphType<TEnum>`

### Breaking Changes
- **Moved `SchemaTypesBase.BuiltInScalarMappings`** to `BuiltInScalarMappingProvider.BuiltInScalarMappings`
- Built-in providers are now automatically prepended to custom mapping providers during schema initialization via `Schema.PrependBuiltInProviders()`

### Improvements
- **Native AOT Compatibility**: Dynamic code generation is now isolated to specific provider classes with proper `RequiresDynamicCode` and `UnconditionalSuppressMessage` attributes
- **Better Separation of Concerns**: Built-in scalar and enum mappings are now handled by dedicated provider classes
- **Cleaner Code**: Removed duplicate mapping logic from `SchemaTypes` and `AutoRegisteringGraphTypeMappingProvider`

### Migration Notes
- Custom mapping providers will now receive built-in mappings via the `preferredType` parameter
- Users referencing `SchemaTypesBase.BuiltInScalarMappings` should update to `BuiltInScalarMappingProvider.BuiltInScalarMappings`
- No changes required for most users as the schema initialization automatically includes these providers
- `LegacySchemaTypes` continues to function as in v8 with no changes to mapping providers

### API Changes
- Updated API approval files for all target frameworks (.NET 5.0, 6.0, 8.0, and .NET Standard 2.0/2.1)
- Added `RequiresDynamicCode` attributes to `AutoRegisteringGraphTypeMappingProvider` and `EnumGraphTypeMappingProvider` constructors
- Added `UnconditionalSuppressMessage` to `AutoRegisteringGraphTypeMappingProvider.GetGraphTypeFromClrType()`

### Documentation
- Updated migration guide with detailed explanation of changes and migration path
- Added notes about `preferredType` parameter behavior for custom providers

### DIFF
--- a/src/GraphQL/Types/Collections/SchemaTypes.cs
+++ b/src/GraphQL/Types/Collections/SchemaTypes.cs
@@ -265,7 +265,6 @@ public sealed partial class SchemaTypes : SchemaTypesBase
             foreach (var (mappedClrType, mappedGraphType) in _schema.TypeMappings)
             {
                 // note that a mapped graph type might be both an input and output graph type
-
                 if (mappedGraphType.IsInputType())
                     _inputClrTypeMappingCache[mappedClrType] = mappedGraphType;
 
@@ -871,7 +870,7 @@ public sealed partial class SchemaTypes : SchemaTypesBase
         /// </summary>
         private Type GetGraphTypeFromClrType(Type clrType, bool isInputType)
         {
-            // Check cache first
+            // Check cache first, which includes mappings within schema.TypeMappings
             var cache = isInputType ? _inputClrTypeMappingCache : _outputClrTypeMappingCache;
             if (cache.TryGetValue(clrType, out var cachedType))
                 return cachedType;


src/GraphQL/Types/Collections/BuiltInScalarMappingProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+
+/// <summary>
+/// Maps CLR types to built-in scalar graph types.
+/// </summary>
+public sealed class BuiltInScalarMappingProvider : IGraphTypeMappingProvider
+{
+    /// <summary>
+    /// Returns a dictionary of default CLR type to graph type mappings for a set of built-in (primitive) types.
+    /// </summary>
+    public static ReadOnlyDictionary<Type, Type> BuiltInScalarMappings { get; } = new(new Dictionary<Type, Type>
```
gao-artur, line 14: Would it make sense to return an `IReadOnlyDictionary` instead of a specific type and initialize it with `FrozenDictionary` for .NET 8+ users?

src/GraphQL/Types/Collections/EnumGraphTypeMappingProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <inheritdoc/>
+    public Type? GetGraphTypeFromClrType(Type clrType, bool isInputType, Type? preferredType)
+    {
+        if (preferredType != null)
+            return preferredType;
+
+        // Auto-generate EnumerationGraphType<T> for enum types
+        if (clrType.IsEnum)
+            return CreateEnumerationGraphTypeNoWarn(clrType);
```
gao-artur, line 24: How often is this method called? Worth caching the created type?

src/GraphQL/Types/Collections/BuiltInScalarMappingProvider.cs
Shane32: Uhh, well, what benefit would that have?  Is there any practical difference?

gao-artur: The Frozen collection is a bit expensive to create, but faster to lookup. So, if this dictionary on a hot path, it would improve the performance. Otherwise doesn't worth it.

Shane32: I'll create a cache to satisfy all the comments here, in addition to caching "auto-registering" types, and then it won't be worth it to optimize this.

src/GraphQL/Types/Collections/EnumGraphTypeMappingProvider.cs
Shane32: Added cache for GetGraphTypeFromClrType which will cover all mapping providers

PR #4282, opened 2026-01-18
Remove unnecessary null-coalescing in FieldMiddlewareBuilder.Build()

Shane32:
This PR removes the unnecessary null-coalescing operator and default delegate fallback in FieldMiddlewareBuilder.Build().

### Changes
- Removed the `_defaultDelegate` static field and its initialization
- Removed the `using GraphQL.Resolvers;` directive (no longer needed)
- Simplified the Build() method to return `start => _middleware(start)` instead 
  of `start => _middleware(start ?? _defaultDelegate)`

### Reasoning
With Nullable Reference Types (NRT) enabled, the `start` parameter in the returned delegate is non-nullable (`FieldMiddlewareDelegate`), making the null-coalescing operator redundant. The schema builder infrastructure ensures that a valid starting delegate is always provided when building the schema, so the fallback to a default delegate is unnecessary. This change simplifies the code while maintaining the same runtime behavior.

### DIFF
--- a/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
+++ b/src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
@@ -1,4 +1,5 @@
 using GraphQL.Instrumentation;
+using GraphQL.Resolvers;
 using GraphQL.Types;
 using GraphQLParser.AST;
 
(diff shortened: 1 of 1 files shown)


src/GraphQL.Tests/Instrumentation/FieldMiddlewareBuilderTests.cs
```diff
@@ -190,6 +191,6 @@ internal static class TestExtensions
     public static FieldMiddlewareDelegate? BuildResolve(this FieldMiddlewareBuilder builder)
     {
         var transform = builder.Build();
-        return transform != null ? transform(null!) : null;
+        return transform != null ? transform(NameFieldResolver.Instance.ResolveAsync) : null;
```
Shane32, line 194: The prior use (passing `null`) was invalid and not used in the codebase.

gao-artur (approved):

Shane32:
This PR should have been part of #4277

PR #4285, opened 2026-01-18
Add AotSchema base class for AOT compilation support

Shane32:
This PR introduces a new `AotSchema` abstract base class to support Ahead-Of-Time (AOT) compilation scenarios in GraphQL.NET.

### Changes

- **New [`AotSchema`](src/GraphQL/Types/AotSchema.cs:1) class**: Abstract base class that extends [`Schema`](src/GraphQL/Types/Schema.cs:1) and implements `IServiceProvider`
  - Provides AOT-friendly type registration via [`AddAotType<TGraphType>()`](src/GraphQL/Types/AotSchema.cs:29) method
  - Maintains a dictionary of type factories for AOT-registered types
  - Overrides service provider to check AOT types first before delegating to the underlying service provider
  - Includes minimal built-in type mappings (int, string, bool, double) optimized for AOT scenarios

- **Refactored [`Schema`](src/GraphQL/Types/Schema.cs:1) constructor**: 
  - Made [`BuiltInTypeMappings`](src/GraphQL/Types/Schema.cs:409) property virtual to allow overriding in derived classes
  - Added internal constructor to support customization of default field resolver, value converter, and built-in mapping providers
  - Refactored built-in mapping provider initialization to be injectable

- **Updated API approval tests**: All target frameworks (net50, net60, net80, netstandard20+21) now include the new [`AotSchema`](src/GraphQL/Types/AotSchema.cs:1) public API surface

### Benefits

- Enables AOT-friendly schema definitions without relying on runtime reflection
- Provides a foundation for source generators to create AOT-compatible schemas
- Maintains backward compatibility with existing [`Schema`](src/GraphQL/Types/Schema.cs:1) usage

### DIFF
--- a/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
@@ -2590,8 +2590,8 @@ namespace GraphQL.Types
         public override object? Serialize(object? value) { }
         public override GraphQLParser.AST.GraphQLValue ToAST(object? value) { }
     }
-    public class EnumerationGraphType<TEnum> : GraphQL.Types.EnumerationGraphType
-        where TEnum : System.Enum
+    public class EnumerationGraphType<[System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicFields)]  TEnum> : GraphQL.Types.EnumerationGraphType
+        where TEnum :  struct, System.Enum
     {
         public EnumerationGraphType() { }
         public override bool CanParseValue(object? value) { }
--- a/src/GraphQL.ApiTests/net60/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net60/GraphQL.approved.txt
@@ -2597,8 +2597,8 @@ namespace GraphQL.Types
         public override object? Serialize(object? value) { }
         public override GraphQLParser.AST.GraphQLValue ToAST(object? value) { }
     }
-    public class EnumerationGraphType<TEnum> : GraphQL.Types.EnumerationGraphType
-        where TEnum : System.Enum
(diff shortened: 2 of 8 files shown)


docs2/site/docs/migrations/migration9.md
```diff
@@ -132,6 +132,12 @@ Notably, the AOT builder does **not** register open generic types such as `AutoR
 
 Additional services needed for your application should be registered explicitly through the builder's configuration delegate. This approach gives you full control over which services are included, ensuring compatibility with Native AOT constraints.
 
+### 5. AotSchema Base Class for AOT-Compatible Schemas
+
+A new `AotSchema` abstract base class has been added to support creating AOT-compatible schemas without relying on runtime reflection or dependency injection for graph type resolution. This class is designed to work with source generators that can produce AOT-friendly schema definitions.
+
+Please see the documentation located (todo) for creating AOT-friendly schemas.
```
Shane32, line 139: docs in future pr after source generator has been written

gao-artur (approved):

src/GraphQL/Types/AotSchema.cs
```diff
@@ -0,0 +1,80 @@
...
+    protected AotSchema(IServiceProvider services, IEnumerable<IConfigureSchema> configurations) : base(services, null, null, new ValueConverterAot(), [])
+    {
+        _services = services;
+        OnPreConfigure(services);
+        foreach (var configuration in configurations.OrderBy(x => x.SortOrder))
+        {
+            configuration.Configure(this, services);
+        }
+        Configure(services);
+        OnPostConfigure(services);
```
gao-artur, line 24: @Shane32 there is a problem with virtual methods called from the constructor
https://www.jetbrains.com/help/resharper/VirtualMemberCallInConstructor.html

Shane32: Right, I'm familiar with this warning.  It's saying that the OnPreConfigure and OnPostConfigure calls will run before the derived class's constructor runs.  So if you use DI injection with a custom constructor, the private readonly variable will still be null during the call to OnPreConfigure or OnPostConfigure.  Maybe we eliminate the methods entirely, and have the generated constructor call Configure.  Consequence is that if the developer writes their own constructor, they must call Configure.  But they can control initialization order.

Shane32: #4293

Issue #3968, opened 2024-06-23
Federation 2 analyzers


gao-artur:
- [ ] `@key`
  - Not null or empty
  - Only allowed on Output and Interface types (type first)
  - Verify fields exist
      - Simple key
      - Compound key
      - Nested fields
- [ ] `@requires`
   - Not null or empty
   - Verify fields exist
   - Required fields must be `@external`
- [ ] `[FederationResolver]`
  - Static
      - Must have resolver per `@key`
      - Return type should match object type
      - Argument name should match `@key` name
      - Allow only arguments marked with `[FromServices]`, key and `CancellationToken`
      - Key argument type should match the field type (including `[Id]` attribute)
  - Instance
      - Only one instance resolver is allowed
      - Return type should match object type
      - Allow only arguments marked with `[FromServices]` and `CancellationToken`
- [ ] `ResolveReference`
  - Only one resolver is allowed

Shane32:
Looks pretty good.  Couple comments:

> Argument name should match `@key` name
> Allow only arguments marked with `[FromServices]`, key and `CancellationToken`
> Allow only arguments marked with `[FromServices]` and `CancellationToken`

I wouldn't add these restrictions.  Users might use `[FromUserContext]`.  They might use `[Name]` to modify the name of the attribute from the local variable.  Or they might have their own attributes that perform a similar task to `[FromServices]`.  Or a global attribute may be registered that does something similar.  For example, a global attribute can be registered to detect `MyDbContext` and when matched, pull it from the service provider, to eliminate the redundancy of using `[FromServices]` everywhere that the database is used.

We could add the analyzer as a 'info' level warning, but I don't think I'd even add it as a 'warning' level warning.

PR #4290, opened 2026-01-18
GQLFED006: Key field must not be an interface or union type

gao-artur:
Contributes to
- #3968

Waits for
 - #4289

### DIFF
--- a/GraphQL.slnx
+++ b/GraphQL.slnx
@@ -62,6 +62,7 @@
     <File Path="docs2/site/docs/analyzers/gqlfed003.md" />
     <File Path="docs2/site/docs/analyzers/gqlfed004.md" />
     <File Path="docs2/site/docs/analyzers/gqlfed005.md" />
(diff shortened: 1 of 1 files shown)


src/GraphQL.Analyzers/SDK/GraphQLGraphType.cs
```diff
@@ -189,6 +203,34 @@ private bool CheckIsOutputType()
         return typeSymbol?.AllInterfaces.Any(i => i.Name is "IObjectGraphType" or "IInterfaceGraphType") == true;
     }
 
+    private bool CheckIsInterfaceType()
+    {
+        var typeSymbol = _typeSymbol.Value;
+        return typeSymbol?.AllInterfaces.Any(i => i.Name == "IInterfaceGraphType") == true;
```
gao-artur, line 209: I'll open a separate PR to make a propert symbols comparison

PR #4313, opened 2026-01-21
Add MemberScanAttribute for controlling member type scanning in auto-registering graph types

Shane32:
## Summary

Adds a new `MemberScanAttribute` that allows developers to control which member types (properties, fields, and/or methods) are scanned when building auto-registering graph types (`AutoRegisteringObjectGraphType<T>`, `AutoRegisteringInterfaceGraphType<T>`, and `AutoRegisteringInputObjectGraphType<T>`).

### New Types
- **`MemberScanAttribute`**: A class-level attribute that accepts a `ScanMemberTypes` enum to specify which members to scan
- **`ScanMemberTypes`**: A flags enum with `Properties`, `Fields`, and `Methods` options

### Default Behavior
- **Output types** (`AutoRegisteringObjectGraphType<T>`, `AutoRegisteringInterfaceGraphType<T>`): Scans properties and methods (fields excluded by default)
- **Input types** (`AutoRegisteringInputObjectGraphType<T>`): Scans properties only (fields and methods excluded by default)

### Implementation Details
- Updated `AutoRegisteringOutputHelper.GetRegisteredMembers<T>()` to respect the `MemberScanAttribute` and conditionally scan properties, fields, and/or methods based on the attribute value
- Updated `AutoRegisteringInputObjectGraphType<T>.GetRegisteredMembers()` to respect the attribute for properties and fields
- Added validation that prevents methods from being scanned for input types (throws `InvalidOperationException`)
- The attribute is inherited by derived classes

## Usage Examples

```csharp
// Scan only methods, excluding properties
[MemberScan(ScanMemberTypes.Methods)]
public class QueryMethodsOnly
{
    public string Property1 { get; set; }  // Not scanned
    public string GetData() => "data";     // Scanned
}

// Scan properties and fields, excluding methods
[MemberScan(ScanMemberTypes.Properties | ScanMemberTypes.Fields)]
public class OutputWithFields
{
    public string Property1 { get; set; } = "prop";  // Scanned
    public string Field1 = "field";                  // Scanned
    public string Method1() => "method";             // Not scanned
}

// Scan only properties for input types
[MemberScan(ScanMemberTypes.Properties | ScanMemberTypes.Fields)]
public class InputWithFields
{
    public string Property1 { get; set; } = "prop";  // Scanned
    public string Field1 = "field";                  // Scanned
}
```

## Tests Added
- Comprehensive unit tests for all combinations of member types across:
  - `AutoRegisteringObjectGraphType<T>`
  - `AutoRegisteringInterfaceGraphType<T>`
  - `AutoRegisteringInputObjectGraphType<T>`
- Tests for inheritance behavior
- Tests for validation (e.g., methods on input types)

## Documentation
- Added section to migration guide explaining the new feature with examples
- Updated XML documentation comments for `GetRegisteredMembers()` methods
- API approval files updated

### DIFF
--- a/src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
+++ b/src/GraphQL.Tests/Types/AutoRegisteringInputObjectGraphTypeTests.cs
@@ -650,11 +650,35 @@ public class AutoRegisteringInputObjectGraphTypeTests
     }
 
     [Fact]
-    public void MemberScanAttribute_InputType_ThrowsWhenMethodsSpecified()
+    public void MemberScanAttribute_InputType_MethodsAreSkipped()
     {
-        Should.Throw<InvalidOperationException>(
-            () => new AutoRegisteringInputObjectGraphType<InputMethodsClass>())
-            .Message.ShouldContain("Methods cannot be scanned for input types");
+        // Methods should be silently ignored for input types, so when Methods-only is specified,
+        // no fields will be present (since properties are not included in the scan)
+        var graphType = new AutoRegisteringInputObjectGraphType<InputMethodsClass>();
+        graphType.Fields.Count.ShouldBe(0); // No fields should be present
+        graphType.Fields.Find("Property1").ShouldBeNull();
+        graphType.Fields.Find("GetData").ShouldBeNull(); // Method should be skipped
+    }
+
+    [Fact]
+    public void MemberScanAttribute_InputType_PropertiesAndMethodsSkipsMethods()
+    {
+        // When both Properties and Methods are specified, methods should be silently ignored
+        var graphType = new AutoRegisteringInputObjectGraphType<InputPropertiesAndMethodsClass>();
+        graphType.Fields.Find("Property1").ShouldNotBeNull();
+        graphType.Fields.Find("Property2").ShouldNotBeNull();
+        graphType.Fields.Find("GetData").ShouldBeNull(); // Method should be skipped
+        graphType.Fields.Find("Field1").ShouldBeNull(); // Field should not be included
+    }
+
+    [Fact]
+    public void MemberScanAttribute_InputType_AllMembersSkipsMethods()
+    {
+        // When all member types are specified, methods should be silently ignored
+        var graphType = new AutoRegisteringInputObjectGraphType<InputAllMembersClass>();
+        graphType.Fields.Find("Property1").ShouldNotBeNull();
+        graphType.Fields.Find("Field1").ShouldNotBeNull();
+        graphType.Fields.Find("GetData").ShouldBeNull(); // Method should be skipped
     }
 
     [MemberScan(ScanMemberTypes.Properties)]
@@ -684,6 +708,24 @@ public class AutoRegisteringInputObjectGraphTypeTests
     private class InputMethodsClass
     {
         public string Property1 { get; set; } = "prop1";
+        public string GetData() => "data";
+    }
+
+    [MemberScan(ScanMemberTypes.Properties | ScanMemberTypes.Methods)]
+    private class InputPropertiesAndMethodsClass
+    {
+        public string Property1 { get; set; } = "prop1";
+        public string Property2 { get; set; } = "prop2";
+        public string Field1 = "field1";
+        public string GetData() => "data";
+    }
+
+    [MemberScan(ScanMemberTypes.Properties | ScanMemberTypes.Fields | ScanMemberTypes.Methods)]
+    private class InputAllMembersClass
+    {
+        public string Property1 { get; set; } = "prop1";
+        public string Field1 = "field1";
+        public string GetData() => "data";
     }
 
     [MemberScan(ScanMemberTypes.Fields)]


Shane32:
@gao-artur I'd like to be able to test my schemas with the new AOT functionality -- mainly in order to give the source generator a thorough workout and be able to test running it in production.  In order to do so, I'll need to add a couple attributes so the source generator can write code compatible with my schemas.  These new attributes should have obvious benefits; they are not user-specific.  This is one of them; also `[InstanceSource]` which controls how the instance is generated for the generated resolver, and `[MapClrType]` to override clr type mapping.

docs2/site/docs/migrations/migration9.md
```diff
@@ -138,6 +138,41 @@ A new `AotSchema` abstract base class has been added to support creating AOT-com
...
 
+### 6. MemberScanAttribute for Auto-Registering Types
+
+A new `MemberScanAttribute` has been added to control which member types (properties, fields, and/or methods) are scanned when building auto-registering graph types. This attribute provides more control over which members of your CLR types become GraphQL fields.
+
+The attribute accepts a `ScanMemberTypes` enum value (flags enum) that can be combined:
+
+- **`Properties`** - Scan public properties
+- **`Fields`** - Scan public fields
+- **`Methods`** - Scan public methods (not applicable for input objects)
```
gao-artur, line 170: Is it possible to register the same CRL type for input and output? What happens in this case?

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
```diff
@@ -98,25 +98,57 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
+                typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(predicate),
+                _excludedProperties);
+            members.AddRange(properties);
+        }
+
+        if (scanFields)
+        {
+            // get public instance fields
+            var fields = typeof(TSourceType).GetFields(BindingFlags.Public | BindingFlags.Instance);
+            members.AddRange(fields);
```
gao-artur, line 147: Would it include readonly fields?

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
```diff
@@ -98,25 +98,57 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
-        return AutoRegisteringHelper.ExcludeProperties(
-            typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(predicate),
-            _excludedProperties);
+        // Check for MemberScanAttribute to determine which member types to scan
+        // Default for input types is Properties only (not methods)
+        var memberScanAttr = typeof(TSourceType).GetCustomAttribute<MemberScanAttribute>();
+        var memberTypes = memberScanAttr?.MemberTypes ?? ScanMemberTypes.Properties;
+
+        // For input types, methods are not supported
+        if ((memberTypes & ScanMemberTypes.Methods) == ScanMemberTypes.Methods)
```
gao-artur: This can be simplified
```c#
if (memberTypes.HasFlag(ScanMemberTypes.Methods))
```

docs2/site/docs/migrations/migration9.md
Shane32: Yes and this is an important existing feature. The same model (eg Person) may be used as an output type and also an input type in a mutation. Features that apply to only one type or the other are ignored when the non-applicable auto registering logic builds the type.

gao-artur: If user relying on defaults, he is safe. But what if the user explicitly set Methods? It looks like it will fail for input type. I understand that he can just remove the attribute, but it may be confusing.
Maybe I just overcomplicating this and user won't probably have a scenario to set the attribute on the type used for input and ouput simultaneously.

Shane32: I see what you're saying, and you're absolutely right.  We should silently ignore methods because the same type may be used for an input mapping.

Shane32: > I see what you're saying, and you're absolutely right. We should silently ignore methods because the same type may be used for an input mapping.

Done

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
Shane32: Fixed

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
Shane32: Done

src/GraphQL/Types/Composite/AutoRegisteringInputObjectGraphType.cs
```diff
@@ -101,25 +101,53 @@ protected virtual IEnumerable<FieldType> ProvideFields()
...
-        // return the list of properties, excluding ones specifically specified in the AutoRegisteringInputGraphType constructor
-        return AutoRegisteringHelper.ExcludeProperties(
-            typeof(TSourceType).GetProperties(BindingFlags.Public | BindingFlags.Instance).Where(predicate),
-            _excludedProperties);
+        // Check for MemberScanAttribute to determine which member types to scan
+        // Default for input types is Properties only (not methods)
+        var memberScanAttr = typeof(TSourceType).GetCustomAttribute<MemberScanAttribute>();
+        var memberTypes = memberScanAttr?.MemberTypes ?? ScanMemberTypes.Properties;
+
+        // Note: for input types, methods are not skipped regardless of the attribute setting, as the same attribute may be used for output types
```
gao-artur: Did you mean "methods are skipped"?

Shane32:
```suggestion
        // Note: for input types, methods are skipped regardless of the attribute setting, as the same attribute may be used for output types
```

Shane32: 🙄 of course lol

PR #4295, opened 2026-01-20
Refactor parameter resolution from ArgumentInformation.Expression to ParameterAttribute-based approach

Shane32:
### Summary
Refactors parameter resolution in auto-registering graph types by moving the logic from `ArgumentInformation.Expression` to a `ParameterAttribute`-based approach. This simplifies the `ArgumentInformation` class and provides a more consistent and extensible mechanism for custom parameter resolution.

#### Core Refactoring
- **`ArgumentInformation`**: Removed `Expression` property, `SetDelegate<TParameterType>()`, and `SetDelegateWithCast()` methods
  - `ConstructQueryArgument()` now returns only `QueryArgument` (not a tuple with `LambdaExpression`)
  - Built-in resolvers for `IResolveFieldContext` and `CancellationToken` moved out of this class
  
- **`AutoRegisteringHelper`**: Added new methods for parameter resolution:
  - `GetParameterResolver()`: Checks for `ParameterAttribute` or built-in parameter types and returns appropriate resolver
  - `ConvertFuncToExpression()`: Converts `Func<IResolveFieldContext, T>` delegates to `LambdaExpression`
  - Updated `BuildFieldResolver_BuildMethodArguments()` to use the new resolver approach

- **`AutoRegisteringOutputHelper.BuildFieldType()`**: Added optional `getTypedParameterResolverMethod` parameter to support custom parameter resolvers

#### Graph Type Updates
- **`AutoRegisteringObjectGraphType<TSourceType>`**: 
  - Added `GetParameterResolver<TParameterType>()` virtual method for extensibility
  - Added `GetParameterResolverInternal<TParameterType>()` helper method
  - Updated `BuildFieldType()` to pass parameter resolver method to `AutoRegisteringOutputHelper`

- **`AutoRegisteringInterfaceGraphType<TSourceType>`**:
  - Applied same pattern as `AutoRegisteringObjectGraphType` for consistency
  - Added `GetParameterResolver<TParameterType>()` virtual method
  - Added `GetParameterResolverInternal<TParameterType>()` helper method
  - Updated `BuildFieldType()` to pass parameter resolver method to `AutoRegisteringOutputHelper`

#### Tests
- Updated `HardcodedValueAttribute` test to inherit from `ParameterAttribute<int>` instead of `GraphQLAttribute`
- Removed `ArgumentInformationTests.cs` as the tested functionality no longer exists

### Benefits
1. **Simplified `ArgumentInformation`**: No longer responsible for storing expression delegates
[...]

PR #4267, opened 2026-01-11
Convert ValueConverter to instance; add ValueConverterAot; many related changes

Shane32:
## Description

This PR refactors `ValueConverter` from a static class to an instance class, with each `ISchema` having its own `ValueConverter` property. This change enables per-schema value conversion customization and eliminates shared static state.

### 1. ValueConverter is now an instance property on ISchema

**Before:**
```csharp
ValueConverter.Register<string, int>(value => int.Parse(value));
var result = ValueConverter.ConvertTo("123", typeof(int));
```

**After:**
```csharp
schema.ValueConverter.Register<string, int>(value => int.Parse(value));
var result = schema.ValueConverter.ConvertTo("123", typeof(int));
```

### 2. Extension methods now require IValueConverter parameter

Methods like `ToObject`, `GetPropertyValue`, and `CompileToObject` now require an `IValueConverter` parameter:

```csharp
// Before
var obj = dictionary.ToObject(typeof(MyType), graphType);

// After
var obj = dictionary.ToObject(typeof(MyType), graphType, schema.ValueConverter);
```

### 3. Parser delegate signature changed

The `Parser` property on `FieldType` and `QueryArgument` now includes an `IValueConverter` parameter:

```csharp
// Before
fieldType.Parser = value => /* transform value */;

// After
fieldType.Parser = (value, valueConverter) => /* transform value */;
```

**Note:** Backward compatibility overloads are provided in `FieldBuilder.ParseValue()` and `QueryArgumentExtensions.ParseValue()` that accept the old `Func<object, object>` signature.

### 4. IFederationResolver.ParseRepresentation signature changed

```csharp
// Before
object ParseRepresentation(IComplexGraphType graphType, IDictionary<string, object?> representation)

// After
object ParseRepresentation(IComplexGraphType graphType, IDictionary<string, object?> representation, IValueConverter valueConverter)
```

### 5. ExecutionHelper methods updated

`ExecutionHelper.GetArguments` and `CoerceValue` now include an `IValueConverter valueConverter` parameter.

## Benefits

- **Isolation**: Different schemas can have different conversion rules
- **Testability**: Easier to test with isolated converters per schema
- **Thread-safety**: Eliminates shared static state between schemas
- **Flexibility**: Enables customization of value conversion per schema instance

## Migration Guide

A comprehensive migration guide has been added to `docs2/site/docs/migrations/migration9.md` and `site/docs/migrations/migration9.md` with detailed examples for all breaking changes.

## Files Changed

[...]

PR #4332, opened 2026-01-23
Add AOT auto-registering base types & other misc AOT prep

Shane32:
These base types will be used by the source generator to create classes that derive from `AutoRegisteringObjectGraphType<T>` and similar without actually carrying over any functionality (completely trims out code requiring reflection).  This allows referencing auto-generated types from AOT compatible code in typical fashion, using `AutoRegisteringObjectGraphType<T>` and similar, allowing the source generator to generate derived classes that are source generated.  This is important if you want to reference an interface type directly (not by its CLR type) -- mostly necessary when type mapping is turned off for the particular graph type (e.g. for the root query type).  I would need this capability if I were to run my production applications with the source generator.  Besides, it will make the source-generated code simpler.

Note: `EnumerationGraphType<T>` is AOT friendly already (when it's a closed generic) and I don't anticipate rewriting its logic in the source generator.

### DIFF
--- a/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
+++ b/src/GraphQL.ApiTests/net50/GraphQL.approved.txt
@@ -110,6 +110,18 @@ namespace GraphQL
         Output = 2,
         Both = 3,
     }
+    [System.AttributeUsage(System.AttributeTargets.Method | System.AttributeTargets.Property | System.AttributeTargets.Field | System.AttributeTargets.Parameter)]
+    public class BaseGraphTypeAttribute : GraphQL.GraphQLAttribute
+    {
+        public BaseGraphTypeAttribute(System.Type graphType) { }
+        public System.Type BaseGraphType { get; set; }
+        public override void Modify(GraphQL.Types.TypeInformation typeInformation) { }
+    }
+    public class BaseGraphTypeAttribute<TGraphType> : GraphQL.BaseGraphTypeAttribute
+        where TGraphType : GraphQL.Types.IGraphType
+    {
+        public BaseGraphTypeAttribute() { }
+    }
     public static class BoolBox
     {
         public static readonly object False;
@@ -663,10 +675,9 @@ namespace GraphQL
         object ToObject(System.Collections.Generic.IDictionary<string, object?> source, System.Type type, GraphQL.Types.IInputObjectGraphType inputGraphType);
     }
     [System.AttributeUsage(System.AttributeTargets.Method | System.AttributeTargets.Property | System.AttributeTargets.Field | System.AttributeTargets.Parameter)]
-    public class IdAttribute : GraphQL.GraphQLAttribute
+    public class IdAttribute : GraphQL.BaseGraphTypeAttribute<GraphQL.Types.IdGraphType>
     {
         public IdAttribute() { }
-        public override void Modify(GraphQL.Types.TypeInformation typeInformation) { }
     }
     [System.AttributeUsage(System.AttributeTargets.Method | System.AttributeTargets.Property | System.AttributeTargets.Field)]
     public class IgnoreAttribute : GraphQL.GraphQLAttribute
@@ -681,6 +692,18 @@ namespace GraphQL
         public System.Type InterfaceGraphType { get; }
         public override void Modify(GraphQL.Types.IGraphType graphType) { }
     }
+    [System.AttributeUsage(System.AttributeTargets.Property | System.AttributeTargets.Field | System.AttributeTargets.Parameter)]
+    public class InputBaseTypeAttribute : GraphQL.GraphQLAttribute
+    {
+        public InputBaseTypeAttribute(System.Type graphType) { }
+        public System.Type InputBaseType { get; set; }
+        public override void Modify(GraphQL.Types.TypeInformation typeInformation) { }
+    }
+    public class InputBaseTypeAttribute<TGraphType> : GraphQL.InputBaseTypeAttribute
+        where TGraphType : GraphQL.Types.IGraphType
+    {
+        public InputBaseTypeAttribute() { }
+    }
     [System.AttributeUsage(System.AttributeTargets.Class | System.AttributeTargets.Struct | System.AttributeTargets.Property | System.AttributeTargets.Field | System.AttributeTargets.Interface | System.AttributeTargets.Parameter)]
     public class InputNameAttribute : GraphQL.GraphQLAttribute
     {
@@ -690,7 +713,7 @@ namespace GraphQL
         public override void Modify(GraphQL.Types.QueryArgument queryArgument) { }
         public override void Modify(GraphQL.Types.FieldType fieldType, bool isInputType) { }
     }
-    [System.AttributeUsage(System.AttributeTargets.Class | System.AttributeTargets.Struct | System.AttributeTargets.Property | System.AttributeTargets.Field | System.AttributeTargets.Interface | System.AttributeTargets.Parameter)]
+    [System.AttributeUsage(System.AttributeTargets.Class | System.AttributeTargets.Struct | System.AttributeTargets.Method | System.AttributeTargets.Property | System.AttributeTargets.Field | System.AttributeTargets.Interface | System.AttributeTargets.Parameter)]
     public class InputTypeAttribute : GraphQL.GraphQLAttribute
     {
         public InputTypeAttribute([System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicConstructors)] System.Type graphType) { }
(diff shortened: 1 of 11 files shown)


src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -2492,7 +2492,7 @@ namespace GraphQL.Types
         protected virtual GraphQL.Types.TypeInformation GetTypeInformation(System.Reflection.MemberInfo memberInfo) { }
         protected virtual System.Collections.Generic.IEnumerable<GraphQL.Types.FieldType> ProvideFields() { }
     }
-    public class AutoRegisteringInterfaceGraphType<[System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.None | System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicMethods | System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicProperties)]  TSourceType> : GraphQL.Types.InterfaceGraphType<TSourceType>
+    public class AutoRegisteringInterfaceGraphType<TSourceType> : GraphQL.Types.InterfaceGraphType<TSourceType>
```
Shane32, line 2543: As previously discussed, AOT compatibility will now be provided using the source generator, and not through these attributes, which did not always work because they insufficiently walked the type tree.

src/GraphQL.ApiTests/net50/GraphQL.approved.txt
```diff
@@ -2503,10 +2503,10 @@ namespace GraphQL.Types
...
         [System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with \'RequiresDynamicCodeAttribute\' may break fu" +
             "nctionality when AOT compiling.", Justification="The constructor is marked with RequiresDynamicCodeAttribute.")]
-        protected void BuildFieldType(GraphQL.Types.FieldType fieldType, System.Reflection.MemberInfo memberInfo) { }
+        protected virtual void BuildFieldType(GraphQL.Types.FieldType fieldType, System.Reflection.MemberInfo memberInfo) { }
         protected virtual void ConfigureGraph() { }
         protected virtual GraphQL.Types.FieldType? CreateField(System.Reflection.MemberInfo memberInfo) { }
-        protected virtual GraphQL.Types.ArgumentInformation GetArgumentInformation<TParameterType>(GraphQL.Types.FieldType fieldType, System.Reflection.ParameterInfo parameterInfo) { }
+        protected virtual GraphQL.Types.ArgumentInformation GetArgumentInformation(GraphQL.Types.FieldType fieldType, System.Reflection.ParameterInfo parameterInfo) { }
```
Shane32, line 2557: With the changes in #4295 the generic type here is unnecessary.  The new `GetParameterResolver<T>` is typed and provides the replacement functionality.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -46,6 +46,7 @@ public class InputObjectGraphType : InputObjectGraphType<object>
 public class InputObjectGraphType<[NotAGraphType][DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.PublicProperties | DynamicallyAccessedMemberTypes.PublicFields)] TSourceType> : ComplexGraphType<TSourceType>, IInputObjectGraphType
 {
```
Shane32, line 47: I left the `DynamicallyAccessedMembers` on this because it would be needed if someone tried to write an input object 'by hand' without using the source generator in an AOT schema.  While it's easy to write a field resolver for output types, developers are unlikely to implement `ParseDictionary`, `ToAst` and `IsValidDefault` which all need reflection to operate correctly.  So leaving these attributes in place makes it 'likely' that the code would work as intended (even if not optimized).  The `AotAutoRegisteringInputObjectGraphType` does not include the attribute and suppresses the warning as it is unnecessary with source generation.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -67,46 +74,52 @@ internal InputObjectGraphType(InputObjectGraphType<TSourceType>? cloneFrom)
             // for InputObjectGraphType just return the dictionary
             _parseDictionary = static (x, _) => x;
         }
+        else
+        {
+            // set initialization function
+            _initialize = (schema) =>
```
Shane32, line 80: Setting an initialization delegate allows for trimming this code out when using the source generator

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -67,46 +74,53 @@ internal InputObjectGraphType(InputObjectGraphType<TSourceType>? cloneFrom)
...
+                        else
+                        {
+                            // if they have, validate and compile upon first use (if any)
+                            _parseDictionary = (data, valueConverter) => (_parseDictionary = ObjectExtensions.CompileToObject(typeof(TSourceType), this, schema.ValueConverter))(data, valueConverter);
+                        }
+                    }
+                    else
+                    {
+                        // use reflection to convert the dictionary to object
+                        _parseDictionary = null;
```
Shane32, line 107: Changed to `null` here with same effect; see below.

src/GraphQL/Types/Composite/InputObjectGraphType.cs
```diff
@@ -127,13 +141,10 @@ public virtual object ParseDictionary(IDictionary<string, object?> value, IValue
...
-        // remainder of this method should not occur unless the user has overridden Initialize
-        return ParseDictionaryViaReflection(value, valueConverter);
+        // fallback to using reflection-based conversion (or as defined by the value converter)
+        return valueConverter.ToObject(value, typeof(TSourceType), this);
     }
 
-    private object ParseDictionaryViaReflection(IDictionary<string, object?> value, IValueConverter valueConverter)
-        => valueConverter.ToObject(value, typeof(TSourceType), this);
-
     /// <inheritdoc/>
```
Shane32, line 148: So now instead of setting `_parseDictionary` to `ParseDictionaryViaReflection` which just mapped to the `valueConverter` anyway after #4267 , we just set it to null and the fallback logic takes over, which is the same.

src/GraphQL/Types/Composite/AutoRegisteringInterfaceGraphType.cs
```diff
@@ -16,14 +16,23 @@ internal static class AutoRegisteringInterfaceGraphType
...
+    /// Initializes a new instance of the <see cref="AutoRegisteringInterfaceGraphType{TSourceType}"/> class without adding any fields.
+    /// </summary>
+    /// <param name="configureGraph">When true, sets the name and processes all attributes defined for the source type.</param>
+    internal AutoRegisteringInterfaceGraphType(bool configureGraph)
+    {
+        if (configureGraph)
+        {
+            Name = typeof(TSourceType).GraphQLName();
+            ConfigureGraph();
+        }
```
Shane32, line 33: The bool is needed so the signature is different than the other ctors.  I anticipate using this logic - I don't want to reinvent the wheel with the source generator.

src/GraphQL/Types/Aot/AotAutoRegisteringInputObjectGraphType.cs
```diff
@@ -0,0 +1,24 @@
...
+    public AotAutoRegisteringInputObjectGraphType() : base(true)
+    {
+    }
+
+    /// <inheritdoc/>
+    public override GraphQLValue ToAST(object? value) => throw new NotImplementedException("ToAST must be implemented by the derived class.");
+    /// <inheritdoc/>
+    public override bool IsValidDefault(object value) => throw new NotImplementedException("IsValidDefault must be implemented by the derived class.");
+    /// <inheritdoc/>
+    protected override IEnumerable<MemberInfo> GetRegisteredMembers() => throw new NotImplementedException("GetRegisteredMembers must be implemented by the derived class if used.");
```
gao-artur, line 23: Do you mean the user should implement these methods, or would they be generated?

Shane32: Generated if the generator needs them.

src/GraphQL/Types/Composite/AutoRegisteringObjectGraphType.cs
```diff
@@ -176,19 +182,21 @@ protected virtual LambdaExpression BuildMemberInstanceExpression(MemberInfo memb
...
     [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "The constructor is marked with RequiresUnreferencedCode.")]
     [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "The constructor is marked with RequiresDynamicCode.")]
-    private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression = AutoRegisteringOutputHelper.BuildSourceExpression<TSourceType>();
+    private static readonly Expression<Func<IResolveFieldContext, TSourceType>> _sourceExpression
+        = AutoRegisteringOutputHelper.BuildSourceExpression<TSourceType>();
 
-    private static readonly MethodInfo _getArgumentInformationInternalMethodInfo = typeof(AutoRegisteringObjectGraphType<TSourceType>).GetMethod(nameof(GetArgumentInformationInternal), BindingFlags.NonPublic | BindingFlags.Instance)!;
-    private ArgumentInformation GetArgumentInformationInternal<TParameterType>(FieldType fieldType, ParameterInfo parameterInfo)
-        => GetArgumentInformation<TParameterType>(fieldType, parameterInfo);
+    internal static object ThrowSourceNullException()
```
Shane32, line 188: todo: move this to a helper method since it isn't used here anymore

